PR #730, opened 2020-10-01
Improved README.md

yashmit178:


YaelRiv:
Thanks @yashmit178, but we're going to stick with the original here for more clarity.

PR #732, opened 2020-10-01
Update lakefs_cli.md

AtieveWadhwa:
Removed broken text


arielshaqed (commented):
Thanks!

This is actually Markdown that gets post-processed, and the `{...}` becomes properties on the link.  Please see the result at https://docs.lakefs.io/quickstart/lakefs_cli.html.

I'll go ahead and close this, as it seems to be working as intended.

Thanks!

PR #733, opened 2020-10-01
Redefined README with syntax highlighted

SwarnadeepGhosh:
## Description
Syntax Highlighting is extremely necessary for the viewers because it gives ease and faster reading of the code fences.
For example :
```
class Myclass(){
  public static void main(){}
}
```
The above one is without syntax highlighting. And the below one is with syntax highlighted.
```java
class Myclass(){
  public static void main(String args[]){}
}
```
Now we can see the difference in how much syntax highlighting is important for users.

## Deploy Notes
I have used markdown syntax highlighting guidelines for this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ For more information see the [Official Documentation](https://docs.lakefs.io).
 
 2. Run the following command in PowerShell:
 
-   ```shell script
+   ```bash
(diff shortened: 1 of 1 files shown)


SwarnadeepGhosh:
Please review

arielshaqed:
Thanks!

This actually produces identical output, as you can see when clicking "Display the rich diff" button at the top of "Files".  I'm going to go ahead and close this, as it does not change appearances.

PR #737, opened 2020-10-01
Updated TOC, fixed markdown of links

alenros:

PR #739, opened 2020-10-01
Fixed TOC and links

alenros:


arielshaqed:
Please see comments on #737, I believe that they apply here too.

Thanks!

alenros:
You are right in both cases!

PR #745, opened 2020-10-01
Update docker-compose.yaml

impramodsargar:
Please remove these it is exposing docker's credentials

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -22,3 +22,4 @@ services:
     environment:
       POSTGRES_USER: lakefs
       POSTGRES_PASSWORD: lakefs
+Note: this file containing sensitive data which is user and password 


arielshaqed (commented):
This should probably be either as a fix or as an issue.

However you can see the Postgres password and username are exposed only inside the two containers that require them: `lakefs` and `postgres`.  Note that this is a single-machine configuration with local-only storage and a local containerised database that is not backed up.  This configuration is indeed vulnerable through these passwords -- if unmodified -- to a local unprivileged attacker.  So a fix would be to randomise them somehow, and an issue would be to describe the above.

I shall be closing this PR as it is not a valid code change -- it adds a top-level key to the configuration YaML object that is not AFAICT read by the code.

Thanks!

PR #750, opened 2020-10-02
General guidelines for Hacktoberfest on the front page

arielshaqed:
This should be removed on or shortly after 2020-11-01, of course.

Explain what PRs will be rejected, and how to avoid falling into one of those categories.

Includes an attempt at styling, but that appears invisible in GFM in a PR.


itaiad200 (approved):

arielshaqed:
Thanks!

PR #751, opened 2020-10-02
Addition of config endpoint to api that currently retrieves only blockstore type and using it in the client

### LINKED ISSUE #676: On repo creation from the UI, storage namespace placeholder hint should match lakeFS blockstore type
When creating a repository from the lakeFS UI, you need to specify a "storage namespace". This is the URI pointing to where the data will actually be stored.

Currently the hint always starts with "s3://...". However, since lakeFS can operate with various backing storage services, the hint should be suitable to the current storage type.
e.g. when using Google Storage, the hint should begin with "gs://..." etc.

Related to #502

eylonronen:
resolves #676


api/api_controller.go
```diff
@@ -2297,3 +2303,11 @@ func (c *Controller) ImportFromS3InventoryHandler() repositories.ImportFromS3Inv
 		})
 	})
 }
+
+func (c *Controller) ConfigGetConfigHandler() configop.GetConfigHandler {
+	return configop.GetConfigHandlerFunc(func(params configop.GetConfigParams, user *models.User) middleware.Responder {
+		return configop.NewGetConfigOK().WithPayload(&models.Config{
```
arielshaqed, line 2309: This one is easy to miss: Please add a call to `c.setupRequest` here for authorization.  It's OK to add a new auth type if there is no suitable one already.  Have a look all over this file, every handler func here has one.

swagger.yml
```diff
@@ -2021,3 +2027,16 @@ paths:
...
+      tags:
+        - config
+      operationId: getConfig
+      description: retrieve the lakefs config
+      responses:
+        200:
+          description: the lakefs config
+          schema:
+            $ref: "#/definitions/config"
+
```
arielshaqed, line 2042: Please also add a `401` response in order to be able to fail authorization.

webui/src/actions/config.js
```diff
@@ -0,0 +1,6 @@
+import * as api from "./api";
+import {AsyncActionType} from "./request";
+
+export const CONFIG_GET = new AsyncActionType('CONFIG_GET');
+
+export const getConfig = () => CONFIG_GET.execute(api.config.get);
```
arielshaqed, line 6: nit: newline at end of file.

webui/src/components/RepositoryCreateForm.js
```diff
@@ -52,7 +54,7 @@ export const RepositoryCreateForm = connect(({ repositories }) => {
             <Form.Group as={Row}>
                 <Form.Label column sm={fieldNameOffset}>Storage Namespace</Form.Label>
                     <Col sm={sm}>
-                        <Form.Control type="text" ref={storageNamespaceField} placeholder="i.e. s3://example-bucket/" onChange={checkValidity}/>
+                        <Form.Control type="text" ref={storageNamespaceField} placeholder={`i.e. ${config.payload['blockstore.type']}://example-bucket/`} onChange={checkValidity}/>
```
arielshaqed, line 57:
```suggestion
                        <Form.Control type="text" ref={storageNamespaceField} placeholder={`${config.payload['blockstore.type']}://example-bucket/`} onChange={checkValidity}/>
```
I know we have *too many of these PRs* for switching "e.g." and "i.e." (@johnnyaug), but let's fix this one while we're here.

arielshaqed (changes requested):
Thanks, this is nice work!

I'm sure someone with more UI knowledge than me will step in with an additional review.

But we *will* need to add authorization here, please see comments.  (This is a requested change)

The use of our async reducer seems a bit wasteful, AFAICT configuration seems never to be explicitly reloaded.  Wouldn't just using `useEffect` directly to load the configuration "once" (per app reload) be sufficient here?  (This is a question, I am _not_ currently requesting a change here.)

eylonronen:
I wondered about using your async reducer, but because I believe that since React 16.3 (and the arrival of Context API) redux is not necessary and is actually just overheads.
However, some store mechanism should be used since the entire app might need the config. If you want me to rewrite with a 
context I'd be happy to. It'll take about 5 minutes and will be much neater. Just tell me what you prefer

api/api_controller.go
eylonronen: I also wrote it in the slack help channel but I'll write here as well for the slight chance you'll read it first. 
I don't understand how to test it and would appreciate your guidance on this

api/api_controller.go
```diff
@@ -51,6 +53,8 @@ const (
 	UserContextKey        contextKey = "user"
 )
 
+var cfg = config.NewConfig()
```
johnnyaug, line 56: While this would work, creating a new config object here creates a coupling between the api controller and the config, which is not ideal. Since you don't need the entire config, just the blockstore type for now, I would just take it explictly from `deps.BlockAdapter.BlockstoreType()`
(you get the `deps` once you add the `setupRequest` call like @arielshaqed suggested.

johnnyaug (commented):
This is really nice work! Thank you for your help.
See my comment

api/api_controller.go
eylonronen: I changed it. Now it uses deps, however it works even without the setupRequest so I'm a little confused

api/api_controller_test.go
```diff
@@ -1311,6 +1312,31 @@ func TestHandler_RetentionPolicyHandlers(t *testing.T) {
...
+	t.Run("Get config (currently only block store type)", func(t *testing.T) {
+		resp, err := clt.Config.GetConfig(&config.GetConfigParams{}, bauth)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		got := resp.GetPayload()
+
+		if got.BlockstoreType != "mem" {
+			t.Errorf("expected to read back the same policy, got %s", got.BlockstoreType)
```
johnnyaug, line 1335: Change error message

api/api_controller_test.go
```diff
@@ -1311,6 +1312,31 @@ func TestHandler_RetentionPolicyHandlers(t *testing.T) {
 	})
 }
 
+func TestHandler_ConfigHandlers(t *testing.T) {
+	handler, deps := getHandler(t)
```
johnnyaug, line 1316: Nice test!

You rely on the blockstore type to be "mem", which is the default. It's ok that we're using mem (in-memory) type for the test, but it would be better to specify it explicitly, rather than relying on it as a default.

I suggest adding a `blockstoreType` arg to `getHandler` (before the opts).
If this arg is specified, the handler will be created with the given blockstoreType.

Currently, this value is taken from an environment variable (see NewBlockAdapterByEnv in handler_test.go) - keep this as the fallback option if the new argument is empty.

api/api_controller.go
```diff
@@ -2297,3 +2300,17 @@ func (c *Controller) ImportFromS3InventoryHandler() repositories.ImportFromS3Inv
 		})
 	})
 }
+
+func (c *Controller) ConfigGetConfigHandler() configop.GetConfigHandler {
+	return configop.GetConfigHandlerFunc(func(params configop.GetConfigParams, user *models.User) middleware.Responder {
+		c.setupRequest(user, params.HTTPRequest, []permissions.Permission{
```
johnnyaug, line 2306: See how the other methods call `setupRequest`:
```
deps, err := c.setupRequest(...)
```
then handle the error, and use the returned `deps` rather than `c.deps`

johnnyaug (commented):
Thanks @eylonronen , this is really getting there.
See my comment.s

I saw your messages on Slack, I hope I clarified everything. Let me know here or on Slack if you need anything else.

api/api_controller.go
eylonronen: Works great! (Finally realized how to test it)

api/api_controller.go
```diff
@@ -2297,3 +2300,23 @@ func (c *Controller) ImportFromS3InventoryHandler() repositories.ImportFromS3Inv
 		})
 	})
 }
+
+func (c *Controller) ConfigGetConfigHandler() configop.GetConfigHandler {
+	return configop.GetConfigHandlerFunc(func(params configop.GetConfigParams, user *models.User) middleware.Responder {
+		deps, err := c.setupRequest(user, params.HTTPRequest, []permissions.Permission{
+			{
+				// Should use repository creation permission but it is coupled to a repo id
+				Action:   permissions.ListRepositoriesAction,
```
arielshaqed, line 2309: This action not particularly relevant to the operation, if this is the best we have then I would rather add a new action.  However that's *out of scope* for this PR, so instead I opened #764.
```suggestion
				// TODO(#764): Add a new action for reading configs?
				Action:   permissions.ListRepositoriesAction,
```

testutil/db.go
```diff
@@ -204,3 +204,37 @@ func NewBlockAdapterByEnv(t *testing.T, translator block.UploadIDTranslator) blo
...
+	case gs.BlockstoreType:
+		ctx := context.Background()
+		client, err := storage.NewClient(ctx)
+		if err != nil {
+			t.Fatal("Google Storage new client", err)
+		}
+		return gs.NewAdapter(client, gs.WithTranslator(translator))
+
+	case lakefsS3.BlockstoreType:
+		awsRegion, regionOk := os.LookupEnv(envKeyAwsRegion)
```
arielshaqed, line 219: Not sure I understand.  `config` is exactly a mechanism to allow configuring everything from defaults or a config file or environment variables.  This code seems to bypass that (and `BuildBlockAdapter`) in favour of what looks like _another_ mechanism to allow configuring everything from defaults or environment variables.  Could we use `config` and the existing config, possibly just changing some default values for testing?

arielshaqed (changes requested):
Thanks!

Unhappy with how we build block adapters in testutil.go.  We refactored that part to build from configuration, I would rather we used that than write a new version.  (If too much refactoring is required for that, please open a new issue to do that so we remember to fix it!)

testutil/db.go
johnnyaug: I think the main issue here is that this code is almost a duplication of the function above it, `NewBlockAdapterByEnv` (which may suffer from what @arielshaqed is saying but it's out of the scope of this PR to fix it).

@eylonronen, instead I think `NewBlockAdapterByEnv` should be changed to `NewBlockAdapterByType`. Its callers will extract the type from the env variable (the function will still lookup some other stuff in the environment variables like `envKeyAwsSecretKey` and that is ok for now).

arielshaqed: That would also work for me, thanks!

johnnyaug (approved):
Looks good! Thank you @eylonronen

arielshaqed (approved):
Cool, thanks!

@johnnyaug I'm happy with `webui/`, but am far from a React/Redux expert.  Assuming you're cool with those changes.  (Or if not, we can fix them later).

johnnyaug:
@eylonronen there are some conflicts to resolve

arielshaqed:
@eylonronen your turn to review.  I fixed the conflict on fdf8c31, could you please take a quick look and click here that it's OK, and I'll pull?
Thanks!

eylonronen (commented):
Looks good

arielshaqed:
# :point_right:  Thanks!  :point_left:

arielshaqed:
> I wondered about using your async reducer, but because I believe that since React 16.3 (and the arrival of Context API) redux is not necessary and is actually just overheads.
> However, some store mechanism should be used since the entire app might need the config. If you want me to rewrite with a
> context I'd be happy to. It'll take about 5 minutes and will be much neater. Just tell me what you prefer

https://github.com/treeverse/lakeFS/issues/771 so we can track it.  You were of course right not to put it in a PR.

PR #668, opened 2020-09-29
Add ExtendTaskDeadline

arielshaqed:
Added to requirements, so needed here.

PR #648, opened 2020-09-25
Aligned to secret_access_key

Pr1-xD:
Closed #501 
Aligned all access_secret_key appearances in the code base to secret_access_key.

PR #765, opened 2020-10-05
Make parade an interface

arielshaqed:
To make testing easier.


arielshaqed:
Also fixes a race (eb7e4bb00729d434cd4d9d9ab5cbc0366bb69e12) on the database when listening.

arielshaqed:
Also fixes test DB raciness on #648 and even #641.  It wasn't manifesting before, which is why I could create the PR, but it is now.  If I pull them into each other in reverse order (first this one, then #648, then #641) then we should reach the desired state without breaking CI.

guy-har (approved):

arielshaqed:
Thanks!  Pulling (was stacked behind #641 and #668)...

PR #772, opened 2020-10-06
Wrap all up/down DB migration scripts in a transaction

arielshaqed:
This prevents half-successful migration scripts from leaving the DB in an unusable state.

Bad migration scripts (e.g. missing DDL removal in a down script) may still break things and
require manual fixing.  To fix that we would need to use the single-command `CREATE TABLE` form
with `FOREIGN KEY`s and `CONSTRAINT`s.  We initially made an undocumented decision not to do so,
and this commit continues that tradition.


nopcoder (approved):
cool

PR #790, opened 2020-10-08
Implement cataloger current diff using scanners

nopcoder:
Use a DB entity scanner to go over parent/child branches to perform diff operation.
The scanner interface operations as an iterator that scan a branch or a lineage.
The diff implementation uses the scanner to compare the two branches and produce diff result into a temporary table.
The iteration support starting after 'path' and the diff loop can limit the output to X records to handle cases where we need pagination.


catalog/cataloger_diff.go
```diff
@@ -112,45 +171,130 @@ func (c *cataloger) getDiffSummary(ctx context.Context, tx db.Tx) (map[Differenc
...
+		records++
 	}
-	return nil
+	if err := parentScanner.Err(); err != nil {
+		return err
+	}
+	return batch.Flush()
+}
+
+func lineageCommitIDByBranchID(lineage []lineageCommit, branchID int64) CommitID {
```
tzahij, line 254: Isn't there another very simmilar function?
Anyway - I think map is faster and simpler

catalog/cataloger_diff.go
```diff
@@ -175,17 +319,152 @@ func getDiffDifferences(ctx context.Context, tx db.Tx, limit int, after string)
...
+		}
+		if matchedParent.BranchID == parentBranchID {
+			// check if parent did any change to the entity
+			if matchedParent.MinCommit > effectiveCommits.ParentEffectiveCommit {
+				return DifferenceTypeConflict
+			}
+			return DifferenceTypeRemoved
+		}
+		// check if the parent saw this entry at the time
+		if matchedParent.MinCommit >= effectiveCommits.ParentEffectiveCommitByBranchID(matchedParent.BranchID) {
```
tzahij, line 417: I do not understand this one

catalog/cataloger_diff.go
```diff
@@ -175,17 +319,152 @@ func getDiffDifferences(ctx context.Context, tx db.Tx, limit int, after string)
...
+		if matchedParent.IsDeleted() {
+			return DifferenceTypeAdded
+		}
+		if matchedParent.BranchID == parentBranchID {
+			// check if parent did any change to the entity
+			if matchedParent.MinCommit > effectiveCommits.ParentEffectiveCommit {
+				return DifferenceTypeConflict
+			}
+			return DifferenceTypeChanged
+		}
```
tzahij, line 433: I think else would make it more clear.

catalog/db_lineage_scanner.go
```diff
@@ -0,0 +1,123 @@
...
+	commitID CommitID
+	scanners []*DBBranchScanner
+	nextRow  []*DBScannerEntry
+	ended    bool
+	err      error
+	value    *DBScannerEntry
+	opts     DBScannerOptions
+}
+
+func NewDBLineageScanner(tx db.Tx, branchID int64, commitID CommitID, opts *DBScannerOptions) *DBLineageScanner {
```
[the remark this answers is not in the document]
nopcoder: Added filterDeleted as scanner option

catalog/db_scanner.go
```diff
@@ -0,0 +1,28 @@
...
+
+type DBScanner interface {
+	Next() bool
+	Value() *DBScannerEntry
+	Err() error
+}
+
+func ScanDBEntryUntil(s DBScanner, ent *DBScannerEntry, p string) (*DBScannerEntry, error) {
+	for ent == nil || ent.Path < p {
+		_ = s.Next()
```
arielshaqed, line 18: This scanner seems to return end by `nil` value, rather than by returning `false` from `Scanner.Next`.  Isn't this a change from regular iterator semantics?

arielshaqed (commented):
Just requests to clarify some oddities that make it hard for me to read. Nothing blocking.

catalog/cataloger_diff.go
```diff
@@ -12,19 +12,36 @@ import (
...
-	contextDiffResultsKey      contextKey = "diff_results_key"
+	diffResultsTableNamePrefix = "catalog_diff_results"
+	diffResultsInsertBatchSize = 64
+
+	contextDiffResultsKey contextKey = "diff_results_key"
 )
 
+type diffEffectiveCommits struct {
+	ParentEffectiveCommit  CommitID `db:"parent_effective_commit"` // last commit parent synchronized with child. If non - it is the commit where the child was branched
+	ChildEffectiveCommit   CommitID `db:"child_effective_commit"`  // last commit child synchronized to parent. if never - than it is 1 (everything in the child is a change)
```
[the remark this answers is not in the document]
nopcoder: updated the doc

catalog/cataloger_diff.go
```diff
@@ -92,6 +110,47 @@ func (c *cataloger) doDiffByRelation(ctx context.Context, tx db.Tx, relation Rel
 	}
 }
 
+func (k contextKey) String() string {
+	return string(k)
+}
+
+func (c *diffEffectiveCommits) ParentEffectiveCommitByBranchID(branchID int64) CommitID {
+	for _, l := range c.ParentEffectiveLineage {
```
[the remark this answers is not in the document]
nopcoder: [gist](https://gist.github.com/nopcoder/123b58870d0e34613707d66135282590)
The lineage of 10 items vs 100 - slice will be faster
```
# 10 items
BenchmarkKVItemSlice-16         494677552                2.11 ns/op
BenchmarkKVStringMap-16         172534639                6.94 ns/op

# 100 items
BenchmarkKVItemSlice-16         100000000               29.1 ns/op
BenchmarkKVStringMap-16         182201276                8.50 ns/op
```

catalog/cataloger_diff.go
nopcoder: we pass the parent lineage at the effective commit - we use it to check if the entry existed that that time to tell the difference between removed and none.

catalog/cataloger_diff.go
nopcoder: what do you mean - anding 'else ' before each 'if'?

catalog/db_scanner.go
nopcoder: Next() returns boolean if we read the next value  - false on end-of-file or err. I abused the fact that at the end we return nil as Value().  but I can do better...

catalog/db_lineage_scanner_test.go
```diff
@@ -0,0 +1,201 @@
...
+
+	for _, bufSize := range bufferSizes {
+		_, _ = conn.Transact(func(tx db.Tx) (interface{}, error) {
+			// test lineage scanner
+			for branchNo := range objSkip {
+				branchName := "b" + strconv.Itoa(branchNo)
+				branchID := int64(branchNo + 1)
+				scanner := NewDBLineageScanner(tx, branchID, UncommittedID, &DBScannerOptions{BufferSize: bufSize})
+				for i := 0; scanner.Next(); i++ {
+					o := scanner.Value()
```
arielshaqed, line 36: IIUC we determine end-of-iteration by looking for `scanner.Value() == nil`, right?  So best test that it goes `nil` after `Next` returns `false`.

arielshaqed (commented):
Thanks!  Clearer now.

catalog/db_lineage_scanner_test.go
```diff
@@ -0,0 +1,201 @@
...
+	"context"
+	"strconv"
+	"testing"
+
+	"github.com/treeverse/lakefs/db"
+	"github.com/treeverse/lakefs/testutil"
+)
+
+func TestDBLineageScanner(t *testing.T) {
+	const numberOfObjects = 10
```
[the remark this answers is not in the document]
nopcoder: @arielshaqed saying that `bufferSizes` should include sizes less than `numberOfObjects`?

catalog/db_lineage_scanner_test.go
nopcoder: We determine end-of-iteration by Next() returning false, so I'll test Value() must return non-nil and after iteration nil

catalog/db_scanner.go
```diff
@@ -0,0 +1,27 @@
+package catalog
+
+type DBScannerOptions struct {
```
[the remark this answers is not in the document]
nopcoder: Customized by AdditionalWhere - where we can control by additional where part.
Still need to think about how to handle specific commit ranges without getting this Options too dirty

catalog/cataloger_diff.go
```diff
@@ -66,22 +83,23 @@ func (c *cataloger) Diff(ctx context.Context, repository string, leftBranch stri
...
 }
 
-func (c *cataloger) doDiff(ctx context.Context, tx db.Tx, leftID, rightID int64) error {
+func (c *cataloger) doDiff(ctx context.Context, tx db.Tx, leftID, rightID int64, limit int, after string) error {
 	relation, err := getBranchesRelationType(tx, leftID, rightID)
 	if err != nil {
 		return err
 	}
-	return c.doDiffByRelation(ctx, tx, relation, leftID, rightID)
+	return c.doDiffByRelation(ctx, tx, relation, leftID, rightID, limit, after)
```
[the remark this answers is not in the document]
nopcoder: The rule is to limit+1 at the entry function where it supposes to get the additional one record.
So because `Diff` returns "has more" it needs +1.
I'll update the code that will pass newLimit - it should be cleaner.

arielshaqed (approved):
Cool, thanks!

You may want someone more qualified to approve too, I mostly used this review to clear up my ignorance

catalog/db_branch_scanner.go
```diff
@@ -0,0 +1,126 @@
...
+	if s.after != "" {
+		q = q.Where("path > ?", s.after)
+	}
+	if s.commitID == CommittedID {
+		q = q.Where("min_commit < ?", MaxCommitID)
+	} else if s.commitID > 0 {
+		q = q.Where("min_commit between 1 and ?", s.commitID)
+	}
+	if s.opts.FilterDeleted {
+		q = q.Where("max_commit = ?", MaxCommitID)
```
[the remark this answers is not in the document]
nopcoder: used sq.Sqlizer for additional where part of the query

catalog/db_lineage_scanner_test.go
nopcoder: 1. check the value returned from Next
2. added test to validate that when we have error we expect Value to be nil

catalog/cataloger_diff.go
```diff
@@ -112,45 +185,130 @@ func (c *cataloger) getDiffSummary(ctx context.Context, tx db.Tx) (map[Differenc
...
+			matchedChild = childEnt
+		}
+		// diff between entries
+		diffType := evaluateFromParentElementDiffType(childID, childLastFromParentCommitID, childLineage, parentEnt, matchedChild)
+		if diffType == DifferenceTypeNone {
+			continue
+		}
+
+		diffRec := &diffResultRecord{
+			SourceBranch: parentID,
```
arielshaqed, line 244: I don't understand: isn't this ID fixed on each record, and equal to a parameter that the caller passed in?  (If so, can we just not include it?)

catalog/cataloger_diff.go
```diff
@@ -175,17 +333,152 @@ func getDiffDifferences(ctx context.Context, tx db.Tx, limit int, after string)
...
+	return DifferenceTypeAdded
+}
+
+func insertDiffResultsBatch(exerciser sq.Execer, tableName string, batch []*diffResultRecord) error {
+	if len(batch) == 0 {
+		return nil
+	}
+	ins := psql.Insert(tableName).Columns("source_branch", "diff_type", "path", "entry_ctid")
+	for _, rec := range batch {
+		ins = ins.Values(rec.SourceBranch, rec.DiffType, rec.Entry.Path, rec.EntryCtid)
```
arielshaqed, line 463: I don't understand.  You write a CTID of another table into a _table_?  Then everything up to the top-level exported functions has to carry a comment saying it is only useful inside a transaction (with an appropriately high isolation level), otherwise this is unsafe in the presence of concurrent VACUUMing FULL.  At the very least, add warning lines to all system documentation never to VACUUM FULL the tables.
Similarly, this imposes immediate restrictions either on any implementation of metadata retention (which deletes entries) or of all code that uses the result -- the CTID might point at nothing.
Also, please document the restrictions on restoring from backups: I assume most backup methods will have to invalidate ctid on restore, so the result of writing these diff entries must be trashed during backup/restore.

The manual [has this to say](https://www.postgresql.org/docs/12/ddl-system-columns.html):
> ctid
>
>    The physical location of the row version within its table. Note that although the ctid can be used to locate the row version very quickly, a row's ctid will change if it is updated or moved by VACUUM FULL. Therefore ctid is useless as a long-term row identifier. A primary key should be used to identify logical rows.

Our reliance on CTIDs is becoming a danger to data integrity.  This technical debt might be acceptable because currently there is no breakage that you or I can see.  But at the very least it requires precise exact documentation.

catalog/cataloger_diff.go
```diff
@@ -194,56 +487,37 @@ func (c *cataloger) diffFromChild(ctx context.Context, tx db.Tx, childID, parent
 		PlaceholderFormat(sq.Dollar).
```
arielshaqed, line 487: The canonical way to use Squirrel is to create a [single global variable `psql`](https://github.com/Masterminds/squirrel/blame/master/README.md#L77) (sorry for linking to a blame, dunno how to link to a line inside a Markdown file on GitHub).

catalog/cataloger_diff.go
```diff
@@ -194,56 +487,37 @@ func (c *cataloger) diffFromChild(ctx context.Context, tx db.Tx, childID, parent
...
 	if err != nil && !effectiveCommitsNotFound {
-		return fmt.Errorf("select effective commit: %w", err)
+		return nil, err
 	}
 	if effectiveCommitsNotFound {
 		effectiveCommits.ChildEffectiveCommit = 1 // we need all commits from the child. so any small number will do
 		parentEffectiveQuery, args, err := psql.Select("commit_id as parent_effective_commit").
 			From("catalog_commits").
-			Where("branch_id = ? AND merge_source_branch = ?", childID, parentID).
+			Where("branch_id = ? AND merge_source_branch = ? AND merge_type = 'from_parent'", childID, parentID).
```
arielshaqed, line 502: Is this a bug-fix or a result of splitting a function?  If the former, can we add a test?

catalog/db_lineage_scanner_test.go
arielshaqed: Yay! Thanks...

arielshaqed (changes requested):
Thanks!

We're storing CTIDs inside tables.  This is very scary, because TFM warns against using them for anything long-term.  I *know* that the _current implementation_ only writes them to an effectively-temporary table so it might be OK, but we do need a *tonne* of documentation around this:
1. All functions that might write a CTID to a table some time during their execution should carry a warning.  Specifically when creating such a dangerous table.
1. User-level documentation must warn *never _ever_ to run `VACUUM FULL` on the database if there is any chance of a concurrent merge*.  The fact that we would never do such a thing does not mean everyone follows the same ops guidelines, so we have to tell everyone they have to do it our way.  E.g. user X might have their own small private lakeFS instance (we actually encourage this) and naïvely suppose that this is safe because it is fast.  We should ensure that they know enough to be safe.
1. If there is any chance of such tables surviving and being re-used then we must document backup strategy etc.

Very sorry, but I am requesting changes with this as the primary blocker.  It cannot be solved by code because it is not about code, it is about dangerous future code or dangerous current ops.  I am treating warning all devs calling these functions, and any users running `VACUUM FULL`,  as a threat to user data integrity.  So: if they are *not* such please let me know and I will reconsider.

catalog/cataloger_diff.go
```diff
@@ -12,21 +12,49 @@ import (
...
+	// ParentEffectiveLineage lineage at the ParentEffectiveCommit
+	ParentEffectiveLineage []lineageCommit
+}
+
+type contextKey string
+
+type diffResultRecord struct {
+	SourceBranch int64
+	DiffType     DifferenceType
+	Entry        Entry
```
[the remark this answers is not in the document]
nopcoder: yes, I'll add field doc

catalog/cataloger_diff.go
```diff
@@ -12,21 +12,49 @@ import (
...
-	diffResultsTableNamePrefix            = "catalog_diff_results"
-	contextDiffResultsKey      contextKey = "diff_results_key"
+	diffResultsTableNamePrefix = "catalog_diff_results"
+	diffResultsInsertBatchSize = 1024
+
+	contextDiffResultsKey contextKey = "diff_results_key"
 )
 
+type diffEffectiveCommits struct {
+	// ParentEffectiveCommit last commit parent synchronized with child.
```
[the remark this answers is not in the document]
nopcoder: will use merged with

catalog/cataloger_diff.go
```diff
@@ -12,21 +12,49 @@ import (
...
+	ParentEffectiveLineage []lineageCommit
+}
+
+type contextKey string
+
+type diffResultRecord struct {
+	SourceBranch int64
+	DiffType     DifferenceType
+	Entry        Entry
+	EntryCtid    *string
```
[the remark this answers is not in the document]
nopcoder: In this PR I just used scanners as a replacement for the current SQL queries - in the next steps we can read the complete entry or generate a unique identifier so we can perform a merge without lookup the entry by ID or ctid

.all-contributorsrc
```diff
@@ -0,0 +1,140 @@
+{
```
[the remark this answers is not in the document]
nopcoder: sorry - I'll do it next time

catalog/cataloger_diff.go
```diff
@@ -12,21 +12,49 @@ import (
...
+
+type diffResultsBatchWriter struct {
+	tx                   db.Tx
+	DiffResultsTableName string
+	Records              []*diffResultRecord
+}
+
 var ErrMissingDiffResultsIDInContext = errors.New("missing diff results id in context")
 
+// Diff return list of differences between leftBranch and rightBranch.
```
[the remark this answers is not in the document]
nopcoder: No - this is just the reimplementation of the current support. Branch level - direct parent or direct child.

catalog/cataloger_diff.go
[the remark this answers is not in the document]
arielshaqed: OK, so let's document _that_!
```suggestion
	// CTID of current entry.  Do not use outside of catalog diff-by-iterators, this is going away soon!
	// TODO(issue number): Remove this by 2020-10-19
	EntryCtid    *string
```
(possibly using a larger value of 19.)

catalog/cataloger_diff.go
nopcoder: right, it is a bug - should take the branch ID from the parent entry's branch

catalog/cataloger_diff.go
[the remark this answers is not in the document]
nopcoder: opened an issue and updated the comment

catalog/cataloger_diff.go
nopcoder: This is the same mechanism as today - the cataloger's Merge calls the diff implementation inside the merge transaction and for entries that needs to update/added as part of the diff, we use the ctid to select the full entry information.

catalog/cataloger_diff.go
nopcoder: @tzahij suggested being very explicit - there is no way you will match merge_source_branch with a different merge type between these branches.

catalog/cataloger_diff.go
nopcoder: no, this is great - updated the code.

nopcoder:
> You may want someone more qualified to approve too, I mostly used this review to clear up my ignorance

Thanks for everything - I have also Tzahi going over the implementation to check that it matches the current SQL one

catalog/cataloger_diff.go
```diff
@@ -39,8 +39,8 @@ type contextKey string
 type diffResultRecord struct {
 	SourceBranch int64
 	DiffType     DifferenceType
-	Entry        Entry
-	EntryCtid    *string
+	Entry        Entry   // Partially filled. Path is always set.
+	EntryCtid    *string // CTID of the modified/added entry. Do not use outside of catalog diff-by-iterators. https://github.com/treeverse/lakeFS/issues/831
```
arielshaqed, line 43: Cool!

arielshaqed (approved):
Thanks!  This sets the R number for CTID-20 exposure to <0.8, meaning we should be able to recover from this plague.  :-)

catalog/cataloger_diff.go
arielshaqed: "Drop a marker and keep going" :-)

PR #792, opened 2020-10-08
Include UI tests as part of the Go test workflow

nopcoder:
Remove the UI test as a job, making the Go job as our build.
It will reduce the build time as we do not require to set up a VM for the
job.
After the merge make sure if master branch protection needs an update - remove "Test React App", "Run Test Suite" will test the same.


itaiad200:
@nopcoder How does it reduce the build time? Isn't it tested in parallel with the go parallel?
It's worth doing because it saves resources, but it increases the time, not vice versa.

nopcoder:
> @nopcoder How does it reduce the build time? Isn't it tested in parallel with the go parallel?
> It's worth doing because it saves resources, but it increases the time, not vice versa.

you are right, it was my mistake that it will make the build shorter - the longest job defines that.
this one just reduces the resources needed to run the build.

moving the job will just simplify the github flows as will not have two places where we configure/build nodejs env - one for testing and one for code generation.

PR #797, opened 2020-10-11
add copy to adapter

guy-har:


block/adapter.go
```diff
@@ -56,6 +56,7 @@ type Adapter interface {
 	GetRange(obj ObjectPointer, startPosition int64, endPosition int64) (io.ReadCloser, error)
 	GetProperties(obj ObjectPointer) (Properties, error)
 	Remove(obj ObjectPointer) error
+	Copy(sourceObj, destinationObj ObjectPointer) error
```
arielshaqed, line 59: Don't these need to be `ObjectPointer`s?  Otherwise how do we handle namespaces?

block/gs/adapter.go
```diff
@@ -173,6 +173,25 @@ func (a *Adapter) Remove(obj block.ObjectPointer) error {
 	return nil
 }
 
+func (a *Adapter) Copy(sourceObj, destinationObj block.ObjectPointer) error {
+	var err error
+	defer reportMetrics("Copy", time.Now(), nil, &err)
+	qualifiedDestinationKey, err := resolveNamespace(destinationObj)
+	if err != nil {
+		return err
```
arielshaqed, line 181:
```suggestion
		return fmt.Errorf("resolve destination: %w", err)
```
and similar below gives context on what name failed.

block/local/adapter.go
```diff
@@ -93,6 +93,27 @@ func (l *Adapter) Remove(obj block.ObjectPointer) error {
...
+	dest := l.getPath(destinationObj.Identifier)
+	destinationFile, err := os.Create(dest)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_ = destinationFile.Close()
+	}()
+	source := l.getPath(sourceObj.Identifier)
+	sourceFile, err := os.OpenFile(source, os.O_RDONLY, 0755)
```
arielshaqed, line 106: Why not `os.Open`?  (You anyway pass weird permissions, that are then ignored.)

block/local/adapter.go
```diff
@@ -93,6 +93,27 @@ func (l *Adapter) Remove(obj block.ObjectPointer) error {
 	return os.Remove(p)
 }
 
+func (l *Adapter) Copy(sourceObj, destinationObj block.ObjectPointer) error {
+	dest := l.getPath(destinationObj.Identifier)
+	destinationFile, err := os.Create(dest)
```
arielshaqed, line 98: Create after opening the source - opening the source is a cheaper operation.

block/local/adapter.go
```diff
@@ -93,6 +93,27 @@ func (l *Adapter) Remove(obj block.ObjectPointer) error {
...
+	if err != nil {
+		return err
+	}
+	defer func() {
+		_ = destinationFile.Close()
```
arielshaqed, line 103: 1. This ignores errors.
1. This leaves a bad (empty) file when the copy fails (or even if the source cannot be opened, because of create-before-open above).  It should instead delete it.

arielshaqed (changes requested):
Cool, thanks!
We do need to test this, understandably such integration testing is not part of this PR.  So please do open an issue to do that.

The only *required* change is to pass `ObjectPointer`s in the adapter interface, just like all other methods do.  That also means getting rid of the icky `resolve` calls in the adapter I think.

block/adapter.go
guy-har: They are `ObjectPointer`s

arielshaqed: Aaah, evidently I can copy but I cannot read.

arielshaqed (approved):
Yikes, sorry!
Please see other change requests, nothing there is in any way blocking.

block/local/adapter.go
guy-har: changed it, Thanks

block/local/adapter.go
guy-har: 👍

Issue #534, opened 2020-08-26
Export lakeFS branch to S3 continuously


arielshaqed:
Cover issue for visibility.

arielshaqed:
Was done waaay back but also being removed today.

PR #795, opened 2020-10-11
[parade] Allow deleting aborted and completed tasks

arielshaqed:
Now that it is part of the requested interface, have to fix a bug that double-counts done tasks during delete.

Also refactor and cleanups

PR #803, opened 2020-10-12
Generate tasks from diffs for exporting

arielshaqed:
Part of #534.

Reviewers:
* @guy-har for **everything**
* @nopcoder for changes to diff format, and for Go, and I hope also everything.

### DIFF
--- a/parade/ddl.go
+++ b/parade/ddl.go
@@ -78,6 +78,33 @@ const (
 	TaskInvalid TaskStatusCodeValue = "[invalid]"
 )
 
+var ErrBadTypeConversion = errors.New("bad type")
+
+// nolint: stylecheck
+func (dst *TaskStatusCodeValue) Scan(src interface{}) error {
+	var sc TaskStatusCodeValue
+	switch s := src.(type) {
+	case string:
+		sc = TaskStatusCodeValue(strings.ToLower(s))
+	case []byte:
+		sc = TaskStatusCodeValue(strings.ToLower(string(s)))
+	default:
+		return fmt.Errorf("cannot convert %T to TaskStatusCodeValue: %w", src, ErrBadTypeConversion)
+	}
+
+	if !(sc == TaskPending || sc == TaskInProgress || sc == TaskAborted || sc == TaskCompleted) {
+		// not a failure, "just" be a newer enum value than known
+		*dst = TaskInvalid
+		return nil
+	}
+	*dst = sc
+	return nil
+}
+
+func (src TaskStatusCodeValue) Value() (driver.Value, error) {
+	return string(src), nil
+}
+
 // TaskData is a row in table "tasks".  It describes a task to perform.
 type TaskData struct {
 	ID     TaskID `db:"task_id"`
--- a/parade/ddl.go
+++ b/parade/ddl.go
@@ -110,17 +110,18 @@ type TaskData struct {
 	ID     TaskID `db:"task_id"`
 	Action string `db:"action"`
 	// Body is JSON-formatted
-	Body               *string           `db:"body"`
-	Status             *string           `db:"status"`
-	StatusCode         string            `db:"status_code"`
-	NumTries           int               `db:"num_tries"`
-	MaxTries           *int              `db:"max_tries"`
-	TotalDependencies  *int              `db:"total_dependencies"`
-	ActorID            ActorID           `db:"actor_id"`
-	ActionDeadline     *time.Time        `db:"action_deadline"`
-	PerformanceToken   *PerformanceToken `db:"performance_token"`
-	ToSignalAfter      []TaskID          `db:"to_signal_after"`
-	NotifyChannelAfter *string           `db:"notify_channel_after"`
+	Body               *string             `db:"body"`
+	Status             *string             `db:"status"`
+	StatusCode         TaskStatusCodeValue `db:"status_code"`
+	NumTries           int                 `db:"num_tries"`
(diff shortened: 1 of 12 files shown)


arielshaqed:
hypnotoad says: ignore test failures, review code!

parade/parade_test.go
```diff
@@ -133,6 +133,29 @@ func intAddr(i int) *int {
 	return &i
 }
 
+func scanIDs(t *testing.T, prefix string) []parade.TaskID {
+	t.Helper()
+	rows, err := db.Query(`SELECT id FROM tasks WHERE id LIKE format('%s%%', $1::text)`, prefix)
+	if err != nil {
+		t.Errorf("[I] select remaining IDs for prefix %s: %s", prefix, err)
```
nopcoder, line 140: Fatalf

parade/parade_test.go
```diff
@@ -133,6 +133,29 @@ func intAddr(i int) *int {
...
+	defer func() {
+		if err := rows.Close(); err != nil {
+			t.Fatalf("[I] remaining ids iterator close: %s", err)
+		}
+	}()
+	gotIDs := make([]parade.TaskID, 0)
+	for rows.Next() {
+		var id parade.TaskID
+		if err = rows.Scan(&id); err != nil {
+			t.Errorf("[I] scan ID value: %s", err)
```
nopcoder, line 152: Fatalf - not sure you can continue to scan after that

parade/parade_test.go
```diff
@@ -208,6 +231,40 @@ func TestTaskDataIterator_Values(t *testing.T) {
...
+			var dst parade.TaskStatusCodeValue
+			if err := dst.Scan(c.in); err != nil {
+				if !errors.Is(err, c.err) {
+					t.Errorf("got err %v, expected %v", err, c.err)
+				}
+			} else {
+				if dst != c.out {
+					t.Errorf("expected %s, got %s", c.out, dst)
+				}
+			}
```
nopcoder, line 263:
```suggestion
			err := dst.Scan(c.in)
			if !errors.Is(err, c.err) {
				t.Fatalf("got err %v, expected %v", err, c.err)
			}
			if err == nil && dst != c.out {
				t.Fatalf("expected %s, got %s", c.out, dst)
			}
```

nopcoder (changes requested):

parade/parade_test.go
arielshaqed: Why not?  It's just a broken row (e.g. parse failure).  `rows.Next` is welcome to fail if the database is :fire:  on fire.

So you're actually right, the code is wrong, and I added a test that rows were successfully scanned.

parade/parade_test.go
arielshaqed: Nice!  Taking it but keeping it an error rather than a fatal.  It gives me less indentation (which is good for my 80-column brain) and it tests the case where an error was expected but not received.  Thanks!

arielshaqed (commented):
Thanks, @nopcoder !  PTAL...

nopcoder (approved):

parade/ddl.go
```diff
@@ -78,22 +78,50 @@ const (
...
+	switch s := src.(type) {
+	case string:
+		sc = TaskStatusCodeValue(strings.ToLower(s))
+	case []byte:
+		sc = TaskStatusCodeValue(strings.ToLower(string(s)))
+	default:
+		return fmt.Errorf("cannot convert %T to TaskStatusCodeValue: %w", src, ErrBadTypeConversion)
+	}
+
+	if !(sc == TaskPending || sc == TaskInProgress || sc == TaskAborted || sc == TaskCompleted) {
```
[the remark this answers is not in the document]
nopcoder: Some of the tasks here got wrong code-doc comment - ex: TaskAborted is ABORTED
```
	// TaskPending indicates a task is waiting for an actor to perform it (new or being
	// retried)
	TaskPending TaskStatusCodeValue = "pending"
	// IN_PROGRESS indicates a task is being performed by an actor.
	TaskInProgress TaskStatusCodeValue = "in-progress"
	// ABORTED indicates an actor has aborted this task with message, will not be reissued
	TaskAborted TaskStatusCodeValue = "aborted"
	// TaskCompleted indicates an actor has completed this task with message, will not reissued
	TaskCompleted TaskStatusCodeValue = "completed"
	// TaskInvalid is used by the API to report errors
	TaskInvalid TaskStatusCodeValue = "[invalid]"
```

parade/parade_test.go
nopcoder: You are right that calling Next after Scan that failed will probably exit. The code reader that sees Errorf will think that this error is something we can live with and continue - in this case adding the id to the slice, which is not what we wanted.

parade/ddl.go
```diff
@@ -78,22 +78,50 @@ const (
 	TaskInvalid TaskStatusCodeValue = "[invalid]"
 )
 
+var ErrBadTypeConversion = errors.New("bad type")
+
+// nolint: stylecheck
+func (dst *TaskStatusCodeValue) Scan(src interface{}) error {
```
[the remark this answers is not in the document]
nopcoder: more than one letter - but I don't think for the same type have a different name for the receiver (src, dst)

guy-har (approved):

arielshaqed (commented):
Thanks!

Pulling _without_ #795.  @nopcoder ended up reviewing that too, so it doesn't necessarily need much further review.  But it's still better off as a separate PR.

PR #804, opened 2020-10-12
Improvement/import tool progress indication

### LINKED ISSUE #798: Import manifest progress indication
1. When the import step is complete the progress doesn't show 100% complete - it is missing ~1000 files.
2. When we enter the 'commit' changes step - there is no indication. The general progress looks like the import of add/change/delete continues.

johnnyaug:
Closes #798

### DIFF
--- a/cmdutils/progress.go
+++ b/cmdutils/progress.go
@@ -5,7 +5,6 @@ import (
 	"time"
 
 	"github.com/jedib0t/go-pretty/text"
-
 	"github.com/vbauerster/mpb/v5"
 	"github.com/vbauerster/mpb/v5/decor"
 )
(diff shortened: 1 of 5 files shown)


cmdutils/progress.go
```diff
@@ -4,29 +4,41 @@ import (
...
 	Progress() []*Progress
 }
 
 type Progress struct {
-	label   string
-	current *int64
-	total   *int64
+	label        string
+	current      *int64
+	total        *int64
```
nopcoder, line 38: Think that it will be better to have these as `int64` and when we call the atomic increment get the address of the variable - so we will not be required to handle nil.

nopcoder (commented):
related to the previous talk about this code -

The doc about why you need to use atomic load when you use atomic API - https://preshing.com/20130618/atomic-vs-non-atomic-operations/

nopcoder (approved):
Cool

Issue #441, opened 2020-08-06
SQL error is shown to the user


itaiad200:
When trying to create a repository with a name of an existing repository, the following error is printed:

`Error executing command: error creating repository: insert repository: ERROR: duplicate key value violates unique constr
aint "catalog_repositories_name_uindex" (SQLSTATE 23505)`

This should be changed to a user-friendly error.

Example command to reproduce:
`lakectl repo create lakefs://existing-repo s3://example-bucket`

kzuri:
I would like to take a shot at this

ari-hacks:
👋🏾 I'd like to take this one

YaelRiv:
Hello @ari-hacks! It's all yours. 
If you need any assistance - please join our slack channel:
https://join.slack.com/t/lakefs/shared_invite/zt-g86mkroy-186GzaxR4xOar1i1Us0bzw

ari-hacks:
Awesome, thanks!

avats-dev:
@ari-hacks are you working on this one or can I try?

ari-hacks:
@avats-dev yep, still assigned. 🙂

PR #806, opened 2020-10-12
fixed sql-error-441

ari-hacks:
#441 

New Output: 
Screen Shot 2020-10-22 at 7 38 59 PM

### DIFF
--- a/catalog/cataloger_create_repository.go
+++ b/catalog/cataloger_create_repository.go
@@ -38,7 +38,9 @@ func (c *cataloger) CreateRepository(ctx context.Context, repository string, sto
 
 		// create repository with ref to branch
 		if _, err := tx.Exec(`INSERT INTO catalog_repositories (id,name,storage_namespace,creation_date,default_branch)
-			VALUES ($1,$2,$3,transaction_timestamp(),$4)`, repoID, repository, storageNamespace, branchID); err != nil {
+			VALUES ($1,$2,$3,transaction_timestamp(),$4)`, repoID, repository, storageNamespace, branchID); db.IsUniqueViolation(err) {
+			return nil, fmt.Errorf("%s %w", repository, db.ErrAlreadyExists)
+		} else if err != nil {
 			return nil, fmt.Errorf("insert repository: %w", err)
 		}
 
--- a/cmd/lakectl/cmd/repo.go
+++ b/cmd/lakectl/cmd/repo.go
@@ -96,7 +96,6 @@ var repoCreateCmd = &cobra.Command{
 		if err != nil {
 			DieErr(err)
 		}
-		
 		repo, err := clt.GetRepository(context.Background(), u.Repository)
 		if err != nil {
 			DieErr(err)
--- a/cmd/lakectl/cmd/repo.go
+++ b/cmd/lakectl/cmd/repo.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
-	"strings"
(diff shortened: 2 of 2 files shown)


cmd/lakectl/cmd/repo.go
```diff
@@ -93,10 +94,11 @@ var repoCreateCmd = &cobra.Command{
 			DefaultBranch:    defaultBranch,
 			ID:               &u.Repository,
 		})
-		if err != nil {
+		if err != nil && strings.Contains(err.Error(), "duplicate") {
+			DieFmt("Repository '%s' already exists.\n", u.Repository)
```
arielshaqed, line 98: Thanks!  This unfortunately ties the client to the server implementation; that would be unfortunate -- e.g. if the server implementation changed, or even if the PostgreSQL error message changed.

Also, doing it here on this client means that any other program that uses the [lakeFS API](https://docs.lakefs.io/reference/api.html) will not benefit from this change.

I would prefer that you detect this type of error on the server side.  E.g. the create-repo operation inside `catalog/` should identify this duplicate key condition (because -- unlike the client -- it is necessarily acquainted with the database), and report a new error type.  The client probably does not even need to be changed.

arielshaqed (changes requested):
Thanks!

Please make a change on the server side, details inside.  Do tag me (@ariels) [on our Slack]( https://join.slack.com/t/lakefs/shared_invite/zt-g86mkroy-186GzaxR4xOar1i1Us0bzw) if you have any questions!

cmd/lakectl/cmd/repo.go
ari-hacks: @arielshaqed, thanks that makes sense! I will make those changes

catalog/cataloger_create_repository.go
```diff
@@ -38,7 +38,9 @@ func (c *cataloger) CreateRepository(ctx context.Context, repository string, sto
 
 		// create repository with ref to branch
 		if _, err := tx.Exec(`INSERT INTO catalog_repositories (id,name,storage_namespace,creation_date,default_branch)
-			VALUES ($1,$2,$3,transaction_timestamp(),$4)`, repoID, repository, storageNamespace, branchID); err != nil {
+			VALUES ($1,$2,$3,transaction_timestamp(),$4)`, repoID, repository, storageNamespace, branchID); db.IsUniqueViolation(err) {
+			return nil, fmt.Errorf("%s %w", repository, db.ErrAlreadyExists)
```
arielshaqed, line 42:
```suggestion
			return nil, fmt.Errorf("create repository %s: %w", repository, db.ErrAlreadyExists)
```
(details the operation on the error message).

arielshaqed (commented):
This looks good!  Hadn't realized it would be this short :-)

Does this fix the error message?  If you tested manually and could copy+paste the new error message into the PR, that would be great and I'll approve and pull over the weekend.  Otherwise I'll do these steps on Sunday.

arielshaqed:
Sorry if the below is unclear:

> This looks good! Hadn't realized it would be this short :-)

Short is _good_ and contains fewer places for bugs to show up later.  I like "short".  "Short" is not the same as "easy".  So thanks!

catalog/cataloger_create_repository.go
ari-hacks: Yup, I updated the comment above with a screenshot of the new error message. 🙂

arielshaqed (approved):
This is great, thanks!

PR #807, opened 2020-10-13
add export action and action manager

guy-har:
draft


action/action_manager.go
```diff
@@ -0,0 +1,143 @@
...
+func (a *workerPool) Close() {
+	close(a.exit)
+	close(a.ch)
+	a.wg.Wait()
+}
+
+func (a *workerPool) start() {
+	a.wg.Add(a.workers)
+	for i := 0; i < a.workers; i++ {
+		go func() {
```
arielshaqed, line 132: Can we tell each worker its ID?  Then we can include it in the logging context, and I can see whether a bunch of errors all come from the same worker or from different ones.  (Maybe logrus already has a goroutine ID??)

action/action_manager.go
```diff
@@ -0,0 +1,143 @@
...
+			defer a.wg.Done()
+			for task := range a.ch {
+				res := a.handler.Handle(task.Action, task.Body)
```
arielshaqed, line 135: TODO: periodically extend task ownership here.

action/action_manager.go
```diff
@@ -0,0 +1,143 @@
...
+				err := a.parade.ReturnTask(task.ID, task.Token, res.Status, res.StatusCode)
+				if err != nil {
+					logging.Default().WithField("action", task.Action).Errorf("failed to return task: %w", err)
```
arielshaqed, line 138: Please also include the task ID and the statuses.

action/action_manager.go
```diff
@@ -0,0 +1,143 @@
+package action
```
arielshaqed, line 1: Can these be under `parade/`?  It's a utility library that I think @tzahij already asked for, so it seems reasonable to live there.

action/action_manager_test.go
```diff
@@ -0,0 +1,179 @@
...
+	returnCalled := int32(0)
+	return mockParade{
+		ownCalled:    &ownCalled,
+		returnCalled: &returnCalled,
+		tasks:        tasks,
+	}
+}
+
+func getKTasks(k int, action string) []parade.OwnedTaskData {
+	var tasks []parade.OwnedTaskData
```
arielshaqed, line 33:
```suggestion
	tasks := make([]parade.OwnedTaskData, 0, k)
```

action/action_manager_test.go
```diff
@@ -0,0 +1,179 @@
...
+			},
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			a := NewAction(tt.mh, tt.mp, tt.properties)
+			time.Sleep(tt.sleepTime)
+			a.Close()
+			returnCalled := atomic.LoadInt32(tt.mp.returnCalled)
```
arielshaqed, line 169: I don't understand why we are using the subtleties of `atomic`.  There are a channel operation and a wait between the write to the variable and this read.  By e.g. https://golang.org/ref/mem#tmp_7 it happens before.  Does the race detector complain about a regular read here?
In any case, without these synchronizations I am not even sure why an atomic load would happen after a regular store.

action/handler.go
```diff
@@ -0,0 +1,14 @@
...
+
+import "github.com/treeverse/lakefs/parade"
+
+type HandlerResult struct {
+	Status     string
+	StatusCode parade.TaskStatusCodeValue
+}
+
+type TaskHandler interface {
+	Handle(action string, body *string) HandlerResult
```
arielshaqed, line 11: Please document this type and its methods.

export/export_handler.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func NewHandler(adapter block.Adapter) *Handler {
+	return &Handler{
+		adapter: adapter,
+	}
+}
+
+type TaskBody struct {
+	DestinationNamespace string
```
arielshaqed, line 33: We should align fields with #803, of course.  I also think we should prefer to keep some fields like in that PR.  First, because it keeps the namespace lookups more localised in code.  And secondly, it makes the message easier to read and understand -- useful when you have a bunch of broken tasks and need to understand what should have happened and didn't.
In particular, when copying I don't understand what the "destination namespace" and "destination ID" will be.  I expect the lakeFS object named `/a/b/c` to be stored in some storage namespace with a random-looking ID, but with a destination `s3://my-bucket/branch-foo-export-dir/a/b/c` -- why talk about namespaces here?

arielshaqed (commented):
Thanks!  Very nice stuff.

Some operational comments on the manager.  Also I think the block adapter copy needs some operation: for export it is *not* writing to a destination namespace and hash, just to some path.

action/action_manager.go
guy-har: I considered both options, 
I agree, changing it.

export/export_handler_test.go
```diff
@@ -0,0 +1,81 @@
...
+
+func TestExportHandler_Handle(t *testing.T) {
+	tests := []struct {
+		name           string
+		Body           TaskBody
+		Action         string
+		blockstoreType string
+	}{
+		{
+			name:   "copy on mem",
```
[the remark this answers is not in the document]
guy-har: it's on mem.
used misleading paths.
I will change them

parade/action_manager.go
```diff
@@ -0,0 +1,156 @@
...
+	defaultErrWaitTime = time.Millisecond * 300
+	defaultMaxDuration = time.Minute * 30 // Todo(guys): change this
+)
+
+// ManagerProperties defines the configuration properties of an ActionManager
+type ManagerProperties struct {
+	Workers     int            // number of goroutines handling tasks
+	ChannelSize int            // size of the channel containing tasks for workers
+	MaxTasks    int            // max tasks requested in every ownTasks request
+	WaitTime    *time.Duration // time to wait if OwnTasks returned no tasks.
```
arielshaqed, line 24: Not sure why these are pointers and the `int`s are values, identified by being zero.

parade/action_manager.go
```diff
@@ -0,0 +1,156 @@
...
+	ErrWaitTime *time.Duration // time to wait if OwnTasks returned err.
+	MaxDuration *time.Duration // maxDuration passed to parade.OwnTasks
+}
+
+// A ActionManager manages the process of requesting and returning tasks for a specific TaskHandler
+// The manager requests tasks, sends the tasks to workers through a channel, the workers then handle the task and return it
```
arielshaqed, line 30:
```suggestion
// ActionManager manages the process of requesting and returning tasks for a specific TaskHandler.
// The manager requests tasks, sends the tasks to workers through a channel, the workers then handle the task and return it.
```

parade/action_manager.go
```diff
@@ -0,0 +1,156 @@
...
+	go func() {
+		for {
+			select {
+			case <-a.quit:
+				return
+			default:
+				ownedTasks, err := a.parade.OwnTasks(a.handler.Actor(), a.properties.MaxTasks, a.handler.Actions(), a.properties.MaxDuration)
+				if err != nil {
+					logging.Default().WithField("actor", a.handler.Actor()).Errorf("manager failed to receive tasks: %s", err)
+					time.Sleep(*a.properties.WaitTime)
```
arielshaqed, line 97:
```suggestion
					time.Sleep(*a.properties.ErrWaitTime)
					continue
```
?

The first line is in order to wait the time I think you meant.  The second line is in order not to wait `WaitTime` after waiting `ErrWaitTime`.

parade/action_manager.go
```diff
@@ -0,0 +1,156 @@
...
+	a.wg.Add(a.workers)
+	for i := 0; i < a.workers; i++ {
+		go func() {
+			workerID := uuid.New()
+			defer a.wg.Done()
+			for task := range a.ch {
+				res := a.handler.Handle(task.Action, task.Body)
+				err := a.parade.ReturnTask(task.ID, task.Token, res.Status, res.StatusCode)
+				if err != nil {
+					logging.Default().WithFields(logging.Fields{
```
arielshaqed, line 145: Can we maybe add these fields to a logger and use that on all the handlers?
(Something to consider, certainly not in any way required for this PR!)

arielshaqed (approved):
Thanks!
I'm approving, but would _really_ like to see improved numbers on the benchmark from using this!  That can happen in a future PR, too.

parade/action_manager.go
guy-har: @arielshaqed Because 0 is a valid argument for the pointers. 
And I didn't wan't to override it with a default value.

PR #812, opened 2020-10-14
Fix task status docs

arielshaqed:
Wrong constant names.


arielshaqed:
I think it should be possible to check for these automatically.  Will look.

nopcoder (approved):

PR #814, opened 2020-10-14
loadtest db entry create

nopcoder:
- remove old get repo test


tzahij (approved):
Looks fine to me.
1. Run it on a large DB. so that we can see the impact of updating the primary key (which is scattered on a large area). It is much better if the paths are naturally distributed, and not just consecutive paths that will go to the same index block.
2. I guess write batching is always on. How many entries per write? per transaction?

PR #837, opened 2020-10-18
Add new ReadConfigAction auth action for reading config

### LINKED ISSUE #764: Add new auth action for reading config
See [here](https://github.com/treeverse/lakeFS/pull/751/files#diff-448a1eb83c464f524edead8bd8d82e97R2309) for context.  #751 adds a REST endpoint for (reading) configuration.  It uses "list repositories" as the action, which is... odd.  Create a new relevant auth action for this.
This is more interesting than it seems: GET config endpoints only ever grow, and different users are allowed to read different parts of the config.  I imagine we would have some structure on the returned config object fields, so for now (probably) enough to allow/deny all config values.  (Right now it's just a single field, so equally well allow just this one field)

daniel-shuy:
This PR creates `ReadConfigAction`, a new auth action for reading config, instead of `ListRepositoriesAction`.

Closes #764

### DIFF
--- a/api/api_controller.go
+++ b/api/api_controller.go
@@ -2230,9 +2230,7 @@ func (c *Controller) ConfigGetConfigHandler() configop.GetConfigHandler {
 	return configop.GetConfigHandlerFunc(func(params configop.GetConfigParams, user *models.User) middleware.Responder {
 		deps, err := c.setupRequest(user, params.HTTPRequest, []permissions.Permission{
 			{
-				// Should use repository creation permission but it is coupled to a repo id
-				// TODO(#764): Add a new action for reading configs?
-				Action:   permissions.ListRepositoriesAction,
+				Action:   permissions.ReadConfigAction,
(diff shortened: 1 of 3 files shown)


ozkatz (changes requested):
Thanks for this - PR looks great! 
Just a couple of small things that I think are missing:
1. documentation under docs/reference/authorization.md (there's a table listing all available permission actions that needs to be updated)
2. under auth/setup.go - the built-in "Admins" group should probably come with the appropriate policy that allows to perform this action

daniel-shuy:
> documentation under docs/reference/authorization.md (there's a table listing all available permission actions that needs to be updated)

Done!

> under auth/setup.go - the built-in "Admins" group should probably come with the appropriate policy that allows to perform this action

`AuthFullAccess` already imports all `auth` actions (`auth:*`)

ozkatz (approved):
Looks great! Thanks for the contribution :partying_face:

PR #841, opened 2020-10-19
Callhome to collect heartbeats

### LINKED ISSUE #839: Callhome - Send heartbeat event

itaiad200:
closes #839

### DIFF
--- a/stats/collector.go
+++ b/stats/collector.go
@@ -12,9 +12,12 @@ import (
 )
 
 const (
-	DefaultCollectorEventBufferSize = 1024 * 1024
-	DefaultFlushInterval            = time.Second * 600
-	DefaultSendTimeout              = time.Second * 5
+	collectorEventBufferSize = 1024 * 1024
+	flushInterval            = time.Second * 600
+	sendTimeout              = time.Second * 5
+
+	// heartbeatInterval is the interval between 2 heartbeat events.
+	heartbeatInterval = 60 * time.Minute
 )
 
 type Collector interface {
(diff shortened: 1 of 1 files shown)


stats/collector.go
```diff
@@ -206,6 +213,20 @@ func (s *BufferedCollector) CollectMetadata(accountMetadata *Metadata) {
 	}
 }
 
+func (s *BufferedCollector) collectHeartbeat(ctx context.Context) {
+	// consistent id for all heartbeats of this run
+	const idLength = 16
```
ozkatz, line 218: there's already a process ID that is generated on startup and sent with all events (see sender.go 103), no need to create another one here.

stats/collector.go
```diff
@@ -206,6 +213,20 @@ func (s *BufferedCollector) CollectMetadata(accountMetadata *Metadata) {
...
+	uniqueRunID := randstr.Hex(idLength)
+	for {
+		select {
+		case <-time.After(heartbeatInterval):
```
ozkatz, line 222: I would add a small jitter to avoid thundering herd from large installations

stats/collector.go
```diff
@@ -206,6 +213,20 @@ func (s *BufferedCollector) CollectMetadata(accountMetadata *Metadata) {
...
+			s.CollectEvent("heartbeat", uniqueRunID)
```
ozkatz, line 223: I would add this event as class = "global" (which already exists for events that aren't tied to user actions) and action = "heartbeat" which in any case should remain a static constant to allow filtering by it.

ozkatz (changes requested):

stats/collector.go
itaiad200: Done.

stats/collector.go
itaiad200: Writes to this channel don't determine the time when events are sent, it is controlled by the `flushTicker`.
Large installation collects events from other sources too, so that problem will remain even if we add jitter here.
If you think it's a real issue I'll open a ticket and add jitter to the `flushTicker`.

stats/collector.go
itaiad200: Done.

stats/collector.go
ozkatz: True.

ozkatz (approved):
LGTM!

PR #843, opened 2020-10-19
Update README.md

Sumindar:
Adding Hindi translation of the entire document.
Thanks

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,100 +14,95 @@
 </p>
 
 
-## What is lakeFS
+## लेकफस क्या है
 
-lakeFS is an open source layer that delivers resilience and manageability to object-storage based data lakes.
+लकेफस एक ओपन सोर्स लेयर है जो ऑब्जेक्ट स्टोरेज के डेटा लक्स को तन्यकता एवं प्रबंधनीयता प्रदान करती है।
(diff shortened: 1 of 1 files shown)


arielshaqed (changes requested):
<p dir="rtl">שלום סומינדר!</p><p dir="rtl">גם אנחנו משתמשים בשפות שאינן אנגלית - כיף לראות שיש ביקוש לתרגום מסמכים טכניים!</p>

Hi @Sumindar !

We too use some non-English languages -- it's good to see demand for translated technical documentation!

Thanks for contributing a translation!  Unfortunately we probably do not have any Hindi speakers on staff.  While I search for a reviewer (luckily I do have some leads), could you please do the following?
1. Do *not* remove the original README.md file, it will probably remain in English.  Instead, place the translation under `docs/in/README.md` (or another more suitable place).
2. Add a link to that translation to README.md.
3. Most importantly, suggest and document (in the newly-added directory from step 1) how we can keep the translations up-to-date.  We currently update this file at least once a month, it would be a shame to have to delete translations because of our inability to keep them updated.

arielshaqed:
Continuing on #850.

Issue #445, opened 2020-08-06
Interactive tool for generating a lakeFS configuration file


itaiad200:
The `lakefs` binary uses a yaml configuration file.

As a lakeFS user, I would like a command line utility to generate this config file interactively.

Upon running "lakefs config", the user should be asked to fill in the basic information for a lakefs server to run.
A minimal configuration file example can be found [here](https://docs.lakefs.io/deploying/install.html#docker) in the docs.

Example of information to get from the user:
1. Database connection string
2. Blockstore type.
3. If the blockstore type is S3, choose the region and authentication method (aws profile, aws key pair, or using the instance role)
4. S3 gateway domain  name

The output should be a yaml file containing the configuration. The user should be able to specify an output destination for the file. If not specified, it should be saved to the default location: $HOME/.lakefs.yaml (with override protection).

DarthCucumber:
Hey, I'm a open source newbie. I would like to work on this issue.

YaelRiv:
Hello DarthCucumber! 

It's all yours. Thanks for giving us a hand 💪

Feel free to join our slack channel if you have any questions. 
https://join.slack.com/t/lakefs/shared_invite/zt-g86mkroy-186GzaxR4xOar1i1Us0bzw

Many thanks, 
Yael

johnnyaug:
Hey @DarthCucumber , should I assign you to this issue?

DarthCucumber:
@johnnyaug yea sure I started working on it. If I face any issue can I ask on slack ?

johnnyaug:
Of course, please do!

johnnyaug:
Hey @DarthCucumber , still here? :slightly_smiling_face:

DarthCucumber:
@johnnyaug yep, I'm almost done. I'll make the PR in 1-2 days

PR #846, opened 2020-10-20
Interactive tool for generating a lakeFS configuration file

DarthCucumber:
Please review this PR and point out the necessary changes required or issues in the PR.

For config file I referred the [configuration reference](https://docs.lakefs.io/reference/configuration.html#reference) page

### DIFF
--- a/cmd/lakefs/cmd/config.go
+++ b/cmd/lakefs/cmd/config.go
@@ -1,0 +1,220 @@
+package cmd
+
+import (
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/manifoldco/promptui"
+	"github.com/mitchellh/go-homedir"
+
+	"io/ioutil"
+
+	"encoding/json"
+
+	"github.com/spf13/cobra"
+	"gopkg.in/yaml.v2"
+)
+
+// configuration model
+
+type Database struct {
+	ConnectionString string `yaml:"connection_string"`
+}
+
+type Gateways struct {
+	S3 struct {
+		DomainName string `yaml:"domain_name"`
+		Region     string `yaml:"region"`
+	} `yaml:"s3"`
+}
+
+type Configuration struct {
+	DB         Database               `yaml:"database"`
+	Blockstore map[string]interface{} `yaml:"blockstore"`
+	GW         Gateways               `yaml:"gateways"`
+}
+
+// configCmd represents the config command
+var configCmd = &cobra.Command{
+	Use:   "config",
+	Short: "User Interactive Config File generator",
+	Long:  `long description here...`,
+	Run: func(cmd *cobra.Command, args []string) {
+
+		var config Configuration
+
+		//loop for menu based interaction
+
+		var isEdit string
+		isEdit = "Yes"
+
+		for {
+
+			//Database
+			//connection string
+			prompt := promptui.Prompt{
+				Label:    "Database->connection string",
+				Validate: isEmptyInput,
+			}
+
+			connectionString, _ := prompt.Run()
+			config.DB.ConnectionString = connectionString
+
+			//Blockstore
+			config.Blockstore = make(map[string]interface{})
+			prompt = promptui.Prompt{
+				Label:    "Blockstore->Type",
+				Validate: isEmptyInput,
+			}
+
+			blockstoreType, _ := prompt.Run()
+			config.Blockstore["type"] = blockstoreType
+
+			if blockstoreType == "s3" {
+				S3 := make(map[string]string)
+
+				//s3 region
+				prompt = promptui.Prompt{
+					Label:    "Blockstore->s3->region",
+					Validate: isEmptyInput,
+				}
+
+				blockstoreRegion, _ := prompt.Run()
+				S3["region"] = blockstoreRegion
+
+				//s3 credentials file
+				prompt := promptui.Prompt{
+					Label:    "Blockstore->s3->Credentials File",
+					Validate: isEmptyInput,
+				}
+				credFile, _ := prompt.Run()
+				S3["credentials_file"] = credFile
+
+				//s3 profile
+				prompt = promptui.Prompt{
+					Label:    "Blockstore->s3->profile",
+					Validate: isEmptyInput,
+				}
+				prof, _ := prompt.Run()
+				S3["profile"] = prof
+
+				config.Blockstore["s3"] = S3
+
+			} else {
+
+				Loc := make(map[string]string)
+
+				//path
+				prompt := promptui.Prompt{
+					Label:    "Blockstore->Local->Path",
+					Validate: isEmptyInput,
+				}
+				localPath, _ := prompt.Run()
+				Loc["path"] = localPath
+
+				config.Blockstore["local"] = Loc
+
+			}
+
+			//Gateways
+
+			//domain_name
+			prompt = promptui.Prompt{
+				Label:    "Gateways->Domain Name",
+				Validate: isEmptyInput,
+			}
+
+			domainName, _ := prompt.Run()
+			config.GW.S3.DomainName = domainName
+
+			//Region
+			prompt = promptui.Prompt{
+				Label:    "Gateways->Region",
+				Validate: isEmptyInput,
+			}
+
+			s3Region, _ := prompt.Run()
+			config.GW.S3.Region = s3Region
+
+			//printing the config
+
+			fmt.Println("------------config---------------")
+			PrintConfig(config)
+
+			selectPrompt := promptui.Select{
+				Label: "Want to edit?",
+				Items: []string{"Yes", "No"},
+			}
+
+			_, isEdit, _ = selectPrompt.Run()
+
+			//if not wants to edit then exit loop
+			if isEdit == "No" {
+				break
+			}
+
+		}
+		//menu loop ends
+
+		//config file path
+		homeDir, _ := homedir.Dir()
+		prompt := promptui.Prompt{
+			Label: "Save config file to (default-" + homeDir + "/.lakefs.yaml)",
+		}
+
+		fileDir, _ := prompt.Run()
+
+		if len(fileDir) == 0 {
+			fileDir = homeDir + "/.lakefs.yaml"
+		}
+
+		err := saveConfig(config, fileDir)
+
+		if err != nil {
+			fmt.Println("failed to save the file:", err)
+			os.Exit(1)
+		} else {
+			fmt.Println("Config File Saved Successfully at", fileDir)
+		}
+
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(configCmd)
+}
+
+//utility functions
+
+//function to save config file to provided destination
+func saveConfig(c Configuration, filename string) error {
+
+	bytes, err := yaml.Marshal(c)
+
+	if err != nil {
+		fmt.Println("Error occured while saving file:", err)
+		os.Exit(1)
+	}
+
+	return ioutil.WriteFile(filename, bytes, 0644)
+
+}
+
+//PrintConfig function to pretty print the configuration structure
+func PrintConfig(c Configuration) {
+
+	configData, _ := json.MarshalIndent(c, "", " ")
+	fmt.Printf("configuration:\n%s", string(configData))
+
+}
+
+// input validation functions
+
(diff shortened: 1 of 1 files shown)


arielshaqed (changes requested):
Hi @DarthCucumber!

Sorry, turns out that #445 is an actual requirement for this.  I missed that (thanks, @guy-har !).

@itaiad200 could you assign a reviewer who understand the requirement?

~Thanks for this contribution.  The configuration file format is currently far from stable, and none of our users have expressed a desire for an interactive configuration model.  Indeed many of them prefer the text format.  Meanwhile the configuration format itself is evolving on a monthly basis, making maintenance of a UI a possible burden.~
~That said, there is a long and useful tradition of external configuration tools -- the most notable one is probably [Linux `menuconfig`](https://en.wikipedia.org/wiki/Menuconfig).
So I would like to propose the following:~
~1. Please place your tool under a new directory underneath `contrib`.
2. Add analytics for its usage (@guy-har can probably help).
3. If it sees continued usage, we will be able to adopt it as a supported tool.
WDYT?~

guy-har:
Thanks @arielshaqed, 
we should add analytics for it's usage

@DarthCucumber, can you please add analytics?
You could see here an example https://github.com/treeverse/lakeFS/blob/6c6351b25c7b716b5f59fe44a042c837c6c693e6/cmd/lakefs/cmd/init.go#L54-L64

We would like to send an event on `config`.

Let me know if you have any questions

cmd/lakefs/cmd/config.go
```diff
@@ -0,0 +1,220 @@
...
+		for {
+
+			//Database
+			//connection string
+			prompt := promptui.Prompt{
+				Label:    "Database->connection string",
+				Validate: isEmptyInput,
+			}
+
+			connectionString, _ := prompt.Run()
```
johnnyaug, line 61: Need to handle the errors returned from promptui. In particular, pressing ctrl+c should exit immediately.

cmd/lakefs/cmd/config.go
```diff
@@ -0,0 +1,220 @@
...
+
+		var isEdit string
+		isEdit = "Yes"
+
+		for {
+
+			//Database
+			//connection string
+			prompt := promptui.Prompt{
+				Label:    "Database->connection string",
```
johnnyaug, line 57: I would remove the "->" from the prompts

cmd/lakefs/cmd/config.go
```diff
@@ -0,0 +1,220 @@
...
+					Validate: isEmptyInput,
+				}
+
+				blockstoreRegion, _ := prompt.Run()
+				S3["region"] = blockstoreRegion
+
+				//s3 credentials file
+				prompt := promptui.Prompt{
+					Label:    "Blockstore->s3->Credentials File",
+					Validate: isEmptyInput,
```
johnnyaug, line 89: This one is not mandatory

cmd/lakefs/cmd/config.go
```diff
@@ -0,0 +1,220 @@
...
+				Validate: isEmptyInput,
+			}
+
+			connectionString, _ := prompt.Run()
+			config.DB.ConnectionString = connectionString
+
+			//Blockstore
+			config.Blockstore = make(map[string]interface{})
+			prompt = promptui.Prompt{
+				Label:    "Blockstore->Type",
```
johnnyaug, line 67:
```suggestion
				Label:    "Blockstore type [s3|gs|local|mem|transient]",
```
Also validate that the value is one of those

cmd/lakefs/cmd/config.go
```diff
@@ -0,0 +1,220 @@
...
+		fmt.Println("Error occured while saving file:", err)
+		os.Exit(1)
+	}
+
+	return ioutil.WriteFile(filename, bytes, 0644)
+
+}
+
+//PrintConfig function to pretty print the configuration structure
+func PrintConfig(c Configuration) {
```
johnnyaug, line 205: can be private

cmd/lakefs/cmd/config.go
```diff
@@ -0,0 +1,220 @@
...
+
+	configData, _ := json.MarshalIndent(c, "", " ")
+	fmt.Printf("configuration:\n%s", string(configData))
+
+}
+
+// input validation functions
+
+// to check if the input is empty
+func isEmptyInput(input string) error {
```
johnnyaug, line 215: More like "isNotEmpty"

cmd/lakefs/cmd/config.go
```diff
@@ -0,0 +1,220 @@
...
+	if len(input) < 1 {
```
johnnyaug, line 216:
```suggestion
	if len(input) == 0 {
```

cmd/lakefs/cmd/config.go
johnnyaug: for now, maybe just remove this one, let's only allow profile using this tool

cmd/lakefs/cmd/config.go
johnnyaug:
```suggestion
				Label:    "Connection string for Postgres database to use lakeFS with",
```

cmd/lakefs/cmd/config.go
```diff
@@ -0,0 +1,220 @@
...
+				prompt := promptui.Prompt{
+					Label:    "Blockstore->s3->Credentials File",
+					Validate: isEmptyInput,
+				}
+				credFile, _ := prompt.Run()
+				S3["credentials_file"] = credFile
+
+				//s3 profile
+				prompt = promptui.Prompt{
+					Label:    "Blockstore->s3->profile",
```
johnnyaug, line 96:
```suggestion
					Label:    "AWS profile name to use for authentication",
```

cmd/lakefs/cmd/config.go
```diff
@@ -0,0 +1,220 @@
...
+					Validate: isEmptyInput,
+				}
+				prof, _ := prompt.Run()
+				S3["profile"] = prof
+
+				config.Blockstore["s3"] = S3
+
+			} else {
```
johnnyaug, line 104: need to handle other blockstore types.

for mem,transient - no need for extra input.
for gs, see docs

cmd/lakefs/cmd/config.go
```diff
@@ -0,0 +1,220 @@
...
+
+				Loc := make(map[string]string)
+
+				//path
+				prompt := promptui.Prompt{
+					Label:    "Blockstore->Local->Path",
```
johnnyaug, line 110:
```suggestion
					Label:    "Local path to use for object storage",
```

cmd/lakefs/cmd/config.go
```diff
@@ -0,0 +1,220 @@
...
+
+				config.Blockstore["local"] = Loc
+
+			}
+
+			//Gateways
+
+			//domain_name
+			prompt = promptui.Prompt{
+				Label:    "Gateways->Domain Name",
```
johnnyaug, line 124: Domain name to use for lakeFS's S3-compatible API.
Also add a default value here: `s3.local.lakefs.io:8000`

cmd/lakefs/cmd/config.go
johnnyaug: Also, need to handle all errors in general

cmd/lakefs/cmd/config.go
```diff
@@ -0,0 +1,220 @@
...
+			s3Region, _ := prompt.Run()
+			config.GW.S3.Region = s3Region
+
+			//printing the config
+
+			fmt.Println("------------config---------------")
+			PrintConfig(config)
+
+			selectPrompt := promptui.Select{
+				Label: "Want to edit?",
```
johnnyaug, line 146: If you're giving the option to edit, allow the user to edit just some of the fields.
You can do it by adding the values entered by the user as default values. This why they can press Enter for fields they don't want to edit

Another option is not to give an edit option.

cmd/lakefs/cmd/config.go
```diff
@@ -0,0 +1,220 @@
...
+				Items: []string{"Yes", "No"},
+			}
+
+			_, isEdit, _ = selectPrompt.Run()
+
+			//if not wants to edit then exit loop
+			if isEdit == "No" {
+				break
+			}
+
```
johnnyaug, line 156: many redundant new lines accross the file

cmd/lakefs/cmd/config.go
```diff
@@ -0,0 +1,220 @@
...
+			Label: "Save config file to (default-" + homeDir + "/.lakefs.yaml)",
+		}
+
+		fileDir, _ := prompt.Run()
+
+		if len(fileDir) == 0 {
+			fileDir = homeDir + "/.lakefs.yaml"
+		}
+
+		err := saveConfig(config, fileDir)
```
johnnyaug, line 172: if file exists, confirm overwriting it

cmd/lakefs/cmd/config.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	return ioutil.WriteFile(filename, bytes, 0644)
+
+}
+
+//PrintConfig function to pretty print the configuration structure
+func PrintConfig(c Configuration) {
+
+	configData, _ := json.MarshalIndent(c, "", " ")
```
johnnyaug, line 207: Better to output the configuration here as yaml, so it will look like the end result and have the same keys as in our docs.

cmd/lakefs/cmd/config.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+func init() {
+	rootCmd.AddCommand(configCmd)
+}
+
+//utility functions
+
+//function to save config file to provided destination
+func saveConfig(c Configuration, filename string) error {
```
johnnyaug, line 191: c can be a pointer. also in other places in this file

cmd/lakefs/cmd/config.go
```diff
@@ -0,0 +1,220 @@
...
+	"gopkg.in/yaml.v2"
+)
+
+// configuration model
+
+type Database struct {
+	ConnectionString string `yaml:"connection_string"`
+}
+
+type Gateways struct {
```
johnnyaug, line 25: Instead of using dedicated structs, I would use a map.

johnnyaug (changes requested):
Thank you for this contribution, @DarthCucumber! 

I've left some comments, and also note that some of our linters failed - you can see these failures as comments as well.

DarthCucumber:
@guy-har @johnnyaug @arielshaqed  thank you for reviewing and  pointing out issues. I'll surely add the feature and fix the issues @johnnyaug pointed out. If I get stuck I'll ask here or in the slack.

DarthCucumber:
@guy-har  I tried understanding the code you provided but I couldn't figure out how it's working and how it's related to the config except of the access id and stuff. Is it like generating the access id and other credentials and sharing it with the config file? Correct me if I'm wrong.

guy-har:
@DarthCucumber The code I provided is an example of how we collect usage analytics.
We would also like to collect analytics for config.
So basically you need to use the same stats collector from the example with `stats.CollectEvent("global", "config")`

cmd/lakefs/cmd/config.go
```diff
@@ -0,0 +1,220 @@
...
+	DB         Database               `yaml:"database"`
+	Blockstore map[string]interface{} `yaml:"blockstore"`
+	GW         Gateways               `yaml:"gateways"`
+}
+
+// configCmd represents the config command
+var configCmd = &cobra.Command{
+	Use:   "config",
+	Short: "User Interactive Config File generator",
+	Long:  `long description here...`,
```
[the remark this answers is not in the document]
DarthCucumber: pl suggest a long description, or things I have to include in long description.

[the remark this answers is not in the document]
johnnyaug: You can remove it for now

cmd/lakefs/cmd/config.go
DarthCucumber: So can I use only one structure `Configuration` and for rest I can use map, or should I use map for all of em?

itaiad200:
Closed due to inactivity - @DarthCucumber feel free to reopen if you wish to resume this PR.

PR #851, opened 2020-10-20
diff with additional fields support

nopcoder:


catalog/cataloger_diff.go
```diff
@@ -84,11 +85,15 @@ func (c *cataloger) Diff(ctx context.Context, repository string, leftReference s
 		return nil, false, fmt.Errorf("right reference: %w", err)
 	}
 
-	if limit < 0 || limit > DiffMaxLimit {
-		limit = DiffMaxLimit
```
arielshaqed, line 88: Nit: I prefer this type of logic: it's fewer lines and no `else`, also the common path is straight-line code and the `if` fixes an issue if there is ne.

catalog/cataloger_diff.go
```diff
@@ -168,11 +174,26 @@ func (c *diffEffectiveCommits) ParentEffectiveCommitByBranchID(branchID int64) C
...
+func newDiffResultsBatchWriter(tx db.Tx, tableName string, additionalFields []string) *diffResultsBatchWriter {
+	// scan Entry db metadata and map addinalFields names into the struct field index
+	fieldsByIndex := make(map[int]string)
+	if len(additionalFields) > 0 {
+		ft := reflect.TypeOf(Entry{})
+		for _, fieldName := range additionalFields {
+			for i, l := 0, ft.NumField(); i < l; i++ {
+				// lookup db tag value on field - when matched we keep the field position at the same index of the field name
+				if ft.Field(i).Tag.Get("db") == fieldName {
+					fieldsByIndex[i] = fieldName
```
arielshaqed, line 186: This fails silently when a field name is not found,  I would much rather fail noisily, even to the extent of killing the SQL query with a nasty SQL error.  At the very least, if nothing is found, log an error here.

catalog/cataloger_diff.go
```diff
@@ -439,6 +473,12 @@ func createDiffResultsTable(ctx context.Context, executor sq.Execer) (string, er
 		diff_type integer NOT NULL,
 		path character varying COLLATE "C" NOT NULL,
 		entry_ctid tid
+		physical_address character varying,
```
arielshaqed, line 476: What is the cost of adding 6 potentially NULL fields?

arielshaqed (commented):
I don't understand `fieldsIndex`.  Why do we need our own reflection?  I would expect *always* to pass a full `struct` with all possible fields (basically `Entry`), and "just" change `Select` to have those additional columns (and then `StructScan` them in) and "just" change `Insert` to have *all* possible columns (and then pass them in, making them show up as NULL if missing).
Or is that INSERT going to be too expensive??

catalog/cataloger_diff.go
nopcoder: from what I understood in this case 1 byte per record - having bit mask per record to the fields used

nopcoder:
> I don't understand `fieldsIndex`. Why do we need our own reflection? I would expect _always_ to pass a full `struct` with all possible fields (basically `Entry`), and "just" change `Select` to have those additional columns (and then `StructScan` them in) and "just" change `Insert` to have _all_ possible columns (and then pass them in, making them show up as NULL if missing).
> Or is that INSERT going to be too expensive??

Had a couple of issues and I can tune the implementation - I will start with the end result I want possibly in the next PR - depends on how much time I need to write a blog:

I don't want a temporary diff results table for the internal diff API - I want the diff to return an iterator with diff records - like a scanner. The merge implementation can use this API to produce diff results.
This scanner will be configured - like the current scanners what to read and all be happy as the diff public API is limited the max records it can hold, there is no size issue here and other implementations like export probably will process the output and throw the result. The downside is that while the iterator is working the transaction is alive, which is not a problem in most use cases, but I can also accept a current working transaction to work with.

Back to our story - sqlx support DB reflection, but squirrel doesn't. In this case, I add the right columns to the Scanner select + pass entry and get the right fields filled.  But when I get to batch the data and build an insert query I get no help from the builder to map just these fields to the insert.

Looking at inserting all the fields when batching will require creating another type where all the additional fields are optional or the option (the stong candidate) was to map the entry default field value to null instead of using reflection.

I can go with the last option, it will remove the use of reflection and make the batch insert include all the possible fields -  make sense?

arielshaqed:
> Looking at inserting all the fields when batching will require creating another type where all the additional fields are optional or the option (the stong candidate) was to map the entry default field value to null instead of using reflection. also the read part will have to read all the possible fields - no reflection into the entry as the type can be null now :)
> 
> I can go with the last option, it will remove the use of reflection and make the batch insert include all the possible fields - make sense?

I think that's the best option: This is Go.  TypeScript is on the other side of Begin at every startup in WeWork.

nopcoder:
Updated the way we insert entry information by checking defaults - I can use switch case with the field name - but it will duplicate the metadata lookup by reflection...

arielshaqed (approved):
Thanks!

nopcoder:
thank you

PR #852, opened 2020-10-20
add filtering by relevant commits to branch scanned

tzahij:
The motivation for this PR came from trying to merge an import branch of 178 million entries into an empty master branch.
The import branch is the parent of the  master branch, so it appears in its lineage with a maximum commitID of 100.
the import branch has only entries with commitID 133. but an older import branch from a previous repository has 178 million entries with commitID 34.
When the optimizer builds the plan, it can see that there are 178 million entries with commitID 34. It has no information that they are all irrelevant. so even though it has an index on branchID-min_commit - it will not access it, but scan the unique index for entries that satisfy those conditions. since there are no entries that satisfy it - it will just scan for more than 50 minutes (at which stage we stopped it).
The original query was:
SELECT DISTINCT ON (branch_id,path) branch_id, path, min_commit, max_commit, ctid, checksum 
FROM catalog_entries WHERE branch_id = 34 AND min_commit between 1 and 100
ORDER BY branch_id, path, min_commit desc LIMIT 1024

The query was modify to include only the relevant commits of the import branch (which is the commit number where the import branch was created)
SELECT DISTINCT ON (branch_id,path) branch_id, path, min_commit, max_commit, ctid, checksum 
FROM catalog_entries WHERE branch_id = 34 AND min_commit in (100)
ORDER BY branch_id, path, min_commit desc LIMIT 1024.

This query returns with no rows within some milliseconds


catalog/db_branch_scanner.go
```diff
@@ -35,9 +42,36 @@ func NewDBBranchScanner(tx db.Tx, branchID int64, commitID CommitID, opts *DBSca
...
+	}
+	sql := "SELECT commit_id::text FROM catalog_commits WHERE branch_id = $1 AND commit_id <= $2 ORDER BY commit_id"
+	err := tx.Select(&commits, sql, branchID, maxCommitId)
+	if err != nil {
+		panic(err)
+	}
+	if commitID == UncommittedID {
+		commits = append(commits, strconv.FormatInt(int64(MaxCommitID), 10))
+	}
+	if len(commits) > MaxCommitsInFilter {
```
arielshaqed, line 67: Isn't this condition backwards?  I'd expect to want to have all the commits if there are *few* commits, otherwise use a range.  As written `commitsWhere` has unbounded length, which is odd.

catalog/db_branch_scanner.go
```diff
@@ -35,9 +42,36 @@ func NewDBBranchScanner(tx db.Tx, branchID int64, commitID CommitID, opts *DBSca
...
+func getRelevantCommitsCondition(tx db.Tx, branchID int64, commitID CommitID) (string, error) {
+	var maxCommitId CommitID
+	var commits []string
+	var commitsWhere string
+	if commitID == UncommittedID || commitID == CommittedID {
+		maxCommitId = MaxCommitID
+	} else {
+		maxCommitId = commitID
+	}
+	sql := "SELECT commit_id::text FROM catalog_commits WHERE branch_id = $1 AND commit_id <= $2 ORDER BY commit_id"
```
arielshaqed, line 59: This changes the SQL when `commitID == CommittedID`, from checking `commit_id < MaxCommitId` to checking `commit_id <= MaxCommitId`.

catalog/db_branch_scanner.go
```diff
@@ -35,9 +42,36 @@ func NewDBBranchScanner(tx db.Tx, branchID int64, commitID CommitID, opts *DBSca
...
+	err := tx.Select(&commits, sql, branchID, maxCommitId)
+	if err != nil {
+		panic(err)
```
arielshaqed, line 62: Please don't panic on a DB failure, it might even just be transient.  Just fail the transaction, other things might continue to work,

arielshaqed (changes requested):
It looks like this is special-cased to speed up the first few (100) operations on a branch.  That's not blocking.

Maybe (probably...) I don't understand the code well enough, but it looks to me like there are 2 behavioural changes.  In particular it seems that the condition for filtering is backwards, and will result in very long queries.  How did we test that this indeed speeds things up with these extremely long queries?  And how do we protect ourselves against breaking with a *really* long query?

catalog/db_branch_scanner.go
tzahij: Thanks - you are right.Could have saved me time.   I found it with the debugger :-(
Very impressive

catalog/db_branch_scanner.go
tzahij: This is a query on catalog_commits. where MaxCommitId will never appear. If looking for uncomitted - then that number ia appended

catalog/db_branch_scanner.go
tzahij: About the general comment - F2F

catalog/db_branch_scanner.go
arielshaqed: Is this still relevant?

catalog/db_branch_scanner.go
```diff
@@ -48,28 +48,29 @@ func NewDBBranchScanner(tx db.Tx, branchID int64, commitID CommitID, opts *DBSca
...
+	sql := "SELECT commit_id::text as str_commit_id FROM catalog_commits WHERE branch_id = $1 AND commit_id <= $2 ORDER BY commit_id"
+	err := tx.Select(&commits, sql, branchID, branchMaxCommitID)
 	if err != nil {
 		panic(err)
 	}
 	if commitID == UncommittedID {
 		commits = append(commits, strconv.FormatInt(int64(MaxCommitID), 10))
 	}
-	if len(commits) > MaxCommitsInFilter {
+	if len(commits) < MaxCommitsInFilter {
```
arielshaqed, line 68:
```suggestion
	if len(commits) <= MaxCommitsInFilter {
```
better matches my expectations of "max".

catalog/db_branch_scanner.go
arielshaqed: Thanks!  This makes a lot more sense.  (Seeing "> max" is usually a red flag.)  But this version (without the fix) could not have been any faster -- it yielded essentially the same SQL  So we must have a repeatable benchmark for this code to know that it is faster -- e.g. it would probably have caught this error.

After flipping the comparison, the my comment about an unbounded query length is no longer relevant of course.

arielshaqed (commented):
Hope my comment on the fixed code still shows up (it might not, because it's attached to a now-outdated line.  In case it isn't, here's an important bit about knowing that this code is faster:

> Thanks! This makes a lot more sense. (Seeing "> max" is usually a red flag.) But this version (without the fix) could not have been any faster -- it yielded essentially the same SQL So I think we should have a repeatable benchmark for this code to know that it is faster -- e.g. it would probably have caught this error, and it will allow us to tune the "max" value as hardware and RDS instances evolve.

catalog/db_branch_scanner.go
```diff
@@ -35,9 +42,37 @@ func NewDBBranchScanner(tx db.Tx, branchID int64, commitID CommitID, opts *DBSca
 		s.opts.BufferSize = DBScannerDefaultBufferSize
 	}
 	s.buf = make([]*DBScannerEntry, 0, s.opts.BufferSize)
+	commitsWhere, _ := getRelevantCommitsCondition(tx, branchID, commitID)
```
nopcoder, line 45: 1. error ignored
2. can be called once in `readBufferIfNeeded` or part of query build and not on iterator creation

catalog/db_branch_scanner.go
```diff
@@ -1,23 +1,30 @@
...
-	idx      int
-	after    string
-	ended    bool
-	err      error
-	value    *DBScannerEntry
+	opts         DBScannerOptions
+	tx           db.Tx
+	branchID     int64
+	commitID     CommitID
+	commitsWhere string
```
nopcoder, line 21: we use squerrel to build the query in this scanner - we can sq.Sqlizer to hold where part expression

catalog/db_branch_scanner.go
```diff
@@ -1,23 +1,30 @@
...
+	"strings"
+
 	sq "github.com/Masterminds/squirrel"
 	"github.com/treeverse/lakefs/db"
 )
 
-const DBScannerDefaultBufferSize = 1024
+const (
+	DBScannerDefaultBufferSize = 1024
+	MaxCommitsInFilter         = 100
```
nopcoder, line 13: MaxCommitIDsInBranchScanner or something like that

catalog/db_branch_scanner.go
```diff
@@ -35,9 +42,37 @@ func NewDBBranchScanner(tx db.Tx, branchID int64, commitID CommitID, opts *DBSca
...
+	var branchMaxCommitID CommitID
+	var commits []string
+	var commitsWhere string
+	if commitID == UncommittedID || commitID == CommittedID {
+		branchMaxCommitID = MaxCommitID
+	} else {
+		branchMaxCommitID = commitID
+	}
+	// commit_id name is changed so that sorting will be performed on the numeric value, not the string value (where "10" is less than "2")
+	sql := "SELECT commit_id::text as str_commit_id FROM catalog_commits WHERE branch_id = $1 AND commit_id <= $2 ORDER BY commit_id"
```
nopcoder, line 60: 1. why do we need the string value we can just use the value in numeric golang slice and give it to our builder.
2. missing the limit or limit +1 in order not to read more than we need

catalog/db_branch_scanner.go
```diff
@@ -106,16 +141,17 @@ func (s *DBBranchScanner) buildQuery() sq.SelectBuilder {
...
 		Limit(uint64(s.opts.BufferSize))
 	if s.after != "" {
 		q = q.Where("path > ?", s.after)
 	}
-	if s.commitID == CommittedID {
-		q = q.Where("min_commit < ?", MaxCommitID)
-	} else if s.commitID > 0 {
-		q = q.Where("min_commit between 1 and ?", s.commitID)
-	}
+	//if s.commitID == CommittedID {
```
nopcoder, line 150: remove code comments

catalog/db_branch_scanner.go
```diff
@@ -35,9 +42,37 @@ func NewDBBranchScanner(tx db.Tx, branchID int64, commitID CommitID, opts *DBSca
...
+	sql := "SELECT commit_id::text as str_commit_id FROM catalog_commits WHERE branch_id = $1 AND commit_id <= $2 ORDER BY commit_id"
+	err := tx.Select(&commits, sql, branchID, branchMaxCommitID)
+	if err != nil {
+		panic(err)
+	}
+	if commitID == UncommittedID {
+		commits = append(commits, strconv.FormatInt(int64(MaxCommitID), 10))
+	}
+	if len(commits) < MaxCommitsInFilter {
+		commitsWhere = "min_commit in (" + strings.Join(commits, `,`) + ")"
```
arielshaqed, line 69: Does this work when no commits are found?

arielshaqed (commented):
Small issue: what if no commits are found?  I think it doesn't work.

catalog/db_branch_scanner.go
tzahij: yes

arielshaqed (commented):
I  do not understand when this kind of thing works and when it does not.

E.g. I tried a smallish table on my machine:

```
ariels@postgres=# \d xyzzy
               Table "public.xyzzy"
 Column |  Type   | Collation | Nullable | Default 
--------+---------+-----------+----------+---------
 e      | text    |           |          | 
 n      | integer |           |          | 
Indexes:
    "xyzzy_ix" btree (e)

ariels@postgres=# \d xyzzy_ix 
      Index "public.xyzzy_ix"
 Column | Type | Key? | Definition 
--------+------+------+------------
 e      | text | yes  | e
btree, for table "public.xyzzy"
```

Now I inserted rows with values from `generate_series`, so there is 1 of each value of `e`. 
```
ariels@postgres=# select count(distinct e) from xyzzy;
  count   
----------
 10000000
(1 row)
```
And then I repeated some of the `e` values so that they have multiple `n` values:
```
ariels@postgres=# select e, c from (select e, count(*) c from xyzzy group by e) r where c > 1;
     e      |    c     
------------+----------
 0000005000 |   100001
 0000004000 |   100001
 0000002000 | 10000001
(3 rows)
```

So we have 10_000_000 distinct values, but 3 of them are much more common.

I also created the attached file.
[selector.sql.gz](https://github.com/treeverse/lakeFS/files/5417230/selector.sql.gz)
It analyzes a query which is functionally identical to 
```sql
select sum(n) from xyzzy where e <= '0000007005';
```
but lists all the `e`'s separately.  Annoyingly, it *consistently* (I ran it repeatedly, without "EXPLAIN ANALYZE") runs slightly (10%) slower -- the analyzer goes for a nonparallel index scan for the long "selector" query:
```
QUERY PLAN
Aggregate  (cost=641774.08..641774.09 rows=1 width=8) (actual time=27943.124..27943.134 rows=1 loops=1)
  ->  Index Scan using xyzzy_ix on xyzzy  (cost=0.56..610559.82 rows=12485704 width=4) (actual time=2.046..14629.009 rows=10207005 loops=1)
        Index Cond: (e = ANY ('{0000000000,0000000001,blah blah blah ,0000006981,0000006982,0000006983,0000006984,0000006985,0000006986,0000006987,0000006988,0000006989,0000006990,0000006991,0000006992,0000006993,0000006994,0000006995,0000006996,0000006997,0000006998,0000006999,0000007000,0000007001,0000007002,0000007003,0000007004,0000007005}'::text[]))
Planning Time: 9.370 ms
JIT:
  Functions: 5
  Options: Inlining true, Optimization true, Expressions true, Deforming true
  Timing: Generation 0.446 ms, Inlining 2.910 ms, Optimization 15.908 ms, Emission 7.559 ms, Total 26.822 ms
Execution Time: 27943.654 ms
(9 rows)
```

A wise man once said,
"it hurt itself in its confusion!"

Here's what the fast query does, just parallel scan:
```
ariels@postgres=# explain analyze select sum(n) from xyzzy where e <= '0000007005';
QUERY PLAN
Finalize Aggregate  (cost=225940.37..225940.38 rows=1 width=8) (actual time=9314.943..9317.971 rows=1 loops=1)
  ->  Gather  (cost=225940.15..225940.36 rows=2 width=8) (actual time=9314.376..9317.945 rows=3 loops=1)
        Workers Planned: 2
        Workers Launched: 2
        ->  Partial Aggregate  (cost=224940.15..224940.16 rows=1 width=8) (actual time=9287.774..9287.782 rows=1 loops=3)
              ->  Parallel Seq Scan on xyzzy  (cost=0.00..214398.04 rows=4216845 width=4) (actual time=256.249..5132.478 rows=3402335 loops=3)
                    Filter: (e <= '0000007005'::text)
                    Rows Removed by Filter: 3330998
Planning Time: 0.383 ms
JIT:
  Functions: 17
  Options: Inlining false, Optimization false, Expressions true, Deforming true
  Timing: Generation 3.155 ms, Inlining 0.000 ms, Optimization 1.215 ms, Emission 16.025 ms, Total 20.395 ms
Execution Time: 9320.372 ms
(14 rows)
Time: 9321.415 ms (00:09.321)
```

Now I'm *_NOT_* saying this will always happen.  But I *AM* saying that this query needs to be tested on our simulated customer DB to determine just how much faster or slower this is.

ddl/000008_commit_idx.down.sql
```diff
@@ -0,0 +1,8 @@
+BEGIN;
```
arielshaqed: This file probably needs to be 000010 right now, we have some files in 8 and 9.

ddl/000008_commit_idx.down.sql
```diff
@@ -0,0 +1,8 @@
...
+DROP INDEX IF EXISTS entries_uncommitted_branch_min_commit;
+DROP INDEX IF EXISTS entries_branch_min_commit;
+CREATE INDEX entries_uncommitted_branch_min_commit
```
arielshaqed: How does this affect CreateEntry performance?

ddl/000008_commit_idx.down.sql
```diff
@@ -0,0 +1,8 @@
...
+    ON catalog_entries USING btree
+        (branch_id ASC NULLS LAST, min_commit ASC NULLS LAST)
+    WHERE min_commit = '1000000000000000000'::bigint;
```
arielshaqed: Don't understand: if `min_commit` is constant, why is it included in the table?

ddl/000010_commit_idx.down.sql
```diff
@@ -0,0 +1,8 @@
+BEGIN;
+DROP INDEX IF EXISTS entries_uncommitted_branch_min_commit;
+DROP INDEX IF EXISTS entries_branch_min_commit;
+CREATE INDEX entries_uncommitted_branch_min_commit
```
arielshaqed, line 4: How does this affect performance of CreateEntry?

ddl/000010_commit_idx.down.sql
```diff
@@ -0,0 +1,8 @@
...
+    ON catalog_entries USING btree
+        (branch_id ASC NULLS LAST, min_commit ASC NULLS LAST)
```
arielshaqed, line 6: I don't understand why the index is on `min_commit`, if that is constant.

arielshaqed (changes requested):
I am unsure how the added commits df16ece and/or 5b4df03 relate to [my comment (above)](https://github.com/treeverse/lakeFS/pull/852#pullrequestreview-513876360).

To clarify: This PR makes life more difficult in order to increase performance.  So it needs to come with a benchmark that shows the performance increase.

This matters to me because I don't understand when the new code would be faster than the old code and when it would be slower.  That is because I don't understand how the query planner would do a good job even with a moderate number of commits: it is really hard (for me -- I tried and I don't know!) to know whether to do a full scan or use an index as soon as we have several thousand commit IDs: there are just so many unknowns (e.g. what to do when _some_ of the desired commit IDs are in the "most common" and _others_ are not?  what to do in the regions of the filesystem namespace where the "most common" commit IDs are not common?).

So I measured and gave an example of a query that looks similar and runs _slower_ when this technique is applied.  If you believe that that query is a *bad* example (certainly possible!) then please explicitly dismiss that comment and measure a *good* example.

arielshaqed (approved):
OK... so we saw many cases where it appears to speed things up on a similar query, and a few where it slows things down but not by much.   As stated it doesn't slow down `CreateEntry` -- so I think we can pull this.
I'd still like to limit it to <100 commits (or even fewer) -- I don't understand what the query planner is doing here, and I feel more comfortable with giving it fewer cases in the IN clause.
Thanks for the hard work profiling this -- it's the only way to introduce some confidence in what the planner will do to us!

PR #854, opened 2020-10-21
fix(docs): small grammar fixes & not correct words

daniloff200:
Bunch of small grammar fixes, which were found during the docs learning and researching

### DIFF
--- a/docs/reference/import.md
+++ b/docs/reference/import.md
@@ -27,7 +27,7 @@ Unfortunately, copying data is not always feasible for the following reasons:
 3. It requires you to switch to using the lakeFS endpoint in all places at once.
 
 ## Using lakeFS import tool
-To solve this, we offer an import tool which will not copy any data and allow for a more gradual onboarding process.
+To solve this we offer an import tool that does not copy any data, allowing for a more gradual onboarding process.
 
 The lakeFS import tool will use the [S3 Inventory](https://docs.aws.amazon.com/AmazonS3/latest/dev/storage-inventory.html) feature to create lakeFS metadata.
 The imported metadata will be reflected in lakeFS under a special branch, called `import-from-inventory`.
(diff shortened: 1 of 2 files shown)


docs/reference/authorization.md
```diff
@@ -47,7 +47,7 @@ See [this example for authenticating with the AWS CLI](../using/aws_cli.md).
 
 Access to resources is managed very much like [AWS IAM](https://docs.aws.amazon.com/IAM/latest/UserGuide/intro-structure.html){:target="_blank"}.
 
-There are 4 basic components to the system:
+There are 4 basic components of the system:
```
arielshaqed, line 50: Please revert this change, "to" is correct here and is common.  In this sentence "of" would imply that there can be other components.

docs/reference/import.md
```diff
@@ -27,7 +27,7 @@ Unfortunately, copying data is not always feasible for the following reasons:
 3. It requires you to switch to using the lakeFS endpoint in all places at once.
 
 ## Using lakeFS import tool
-To solve this, we offer an import tool which will not copy any data and allow for a more gradual onboarding process.
+To solve this, we offer an import tool that will not copy any data and allow for a more gradual onboarding process.
```
arielshaqed, line 30: Nice.  Mind trying this?
```suggestion
To solve this we offer an import tool that does not copy any data, allowing for a more gradual onboarding process.
```

arielshaqed (commented):
Great, thanks!

Like any grammar fix PR I have 2 small requests, inline.

docs/reference/authorization.md
daniloff200: done

docs/reference/import.md
daniloff200: Yeah, I think we can use that.
Thanks for the suggestion!

arielshaqed (approved):
Great, thanks!  fixes a bug and more readable.

arielshaqed:
Waiting for tests (which should anyway pass...) to pull.

PR #863, opened 2020-10-25
Diffs are now on references not branches

arielshaqed:
See identical change to naming (but with a very long actual implementation) in 99e1bd1180.


itaiad200 (approved):

arielshaqed:
change parameter name in interface, coverage drops

PR #865, opened 2020-10-26
Use pgx instead of sqlx

arielshaqed:
Needed to allow parade (which needs pgx features not available in sql/sqlx) to use the same transaction with the cataloger.

### DIFF
--- a/catalog/cataloger_get_commit.go
+++ b/catalog/cataloger_get_commit.go
@@ -28,7 +28,7 @@ func (c *cataloger) GetCommit(ctx context.Context, repository, reference string)
 				LEFT JOIN catalog_branches bb ON bb.id = c.merge_source_branch
 			WHERE b.id=$1 AND c.commit_id=$2`
 		var rawCommit commitLogRaw
-		if err := tx.Get(&rawCommit, query, branchID, ref.CommitID); err != nil {
+		if err := tx.GetStruct(&rawCommit, query, branchID, ref.CommitID); err != nil {
 			return nil, err
 		}
 		commit := convertRawCommit(&rawCommit)
@@ -41,12 +41,16 @@ func (c *cataloger) GetCommit(ctx context.Context, repository, reference string)
 }
 
 func convertRawCommit(raw *commitLogRaw) *CommitLog {
+	metadata := raw.Metadata
+	if metadata == nil {
+		metadata = make(Metadata)
+	}
 	c := &CommitLog{
 		Reference:    MakeReference(raw.BranchName, raw.CommitID),
 		Committer:    raw.Committer,
 		Message:      raw.Message,
 		CreationDate: raw.CreationDate,
-		Metadata:     raw.Metadata,
+		Metadata:     metadata,
 	}
 	if raw.MergeSourceBranchName != "" && raw.MergeSourceCommit > 0 {
 		reference := MakeReference(raw.MergeSourceBranchName, raw.MergeSourceCommit)
--- a/catalog/cataloger_get_commit.go
+++ b/catalog/cataloger_get_commit.go
@@ -28,7 +28,7 @@ func (c *cataloger) GetCommit(ctx context.Context, repository, reference string)
 				LEFT JOIN catalog_branches bb ON bb.id = c.merge_source_branch
 			WHERE b.id=$1 AND c.commit_id=$2`
 		var rawCommit commitLogRaw
-		if err := tx.GetStruct(&rawCommit, query, branchID, ref.CommitID); err != nil {
+		if err := tx.Get(&rawCommit, query, branchID, ref.CommitID); err != nil {
 			return nil, err
 		}
 		commit := convertRawCommit(rawCommit)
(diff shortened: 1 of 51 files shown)


arielshaqed:
Sorry for size: while pgx is entirely like sql and scany is entirely like the additions of sqlx, both of these statements are untrue.  OTOH not sure we have a choice.

arielshaqed:
Consider reviewing commit-by-commit, you will review some lines twice but I have tried to arrange the commits in logical order.

go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/davecgh/go-spew v1.1.1
 	github.com/dgryski/go-gk v0.0.0-20200319235926-a69029f61654 // indirect
 	github.com/dlmiddlecote/sqlstats v1.0.1
+	github.com/georgysavva/scany v0.2.6
```
nopcoder, line 19: run go mod tidy to remove the sqlx

catalog/cataloger_list_commits.go
```diff
@@ -71,10 +71,11 @@ func (c *cataloger) ListCommits(ctx context.Context, repository, branch string,
 	return commits, hasMore, err
 }
 
-func convertRawCommits(rawCommits []*commitLogRaw) []*CommitLog {
+func convertRawCommits(rawCommits []commitLogRaw) []*CommitLog {
 	commits := make([]*CommitLog, len(rawCommits))
 	for i, commit := range rawCommits {
-		commits[i] = convertRawCommit(commit)
+		copy := commit
```
nopcoder, line 77: why do we need copy? we do not use the pointer - if we are passing references I prefer convertRawCommit to take a struct and not a pointer so we will not remember to make a copy

catalog/cataloger_get_commit.go
```diff
@@ -41,12 +41,16 @@ func (c *cataloger) GetCommit(ctx context.Context, repository, reference string)
 }
 
 func convertRawCommit(raw *commitLogRaw) *CommitLog {
+	metadata := raw.Metadata
+	if metadata == nil {
+		metadata = make(Metadata)
+	}
```
nopcoder, line 47: which issue did you have with nil map?

db/database.go
```diff
@@ -81,49 +84,57 @@ func (d *SqlxDatabase) performAndReport(fields logging.Fields, fn func() (interf
 	return ret, err
 }
 
-func (d *SqlxDatabase) Get(dest interface{}, query string, args ...interface{}) error {
+func (d *PgxDatabase) GetStruct(dest interface{}, query string, args ...interface{}) error {
```
nopcoder, line 87: The only issue that I have is that all the other methods behave like xxxStruct and this one got the explicit name.
Looks like Get should get the explicit name in this case.

db/database.go
```diff
@@ -273,6 +288,23 @@ func (d *SqlxDatabase) getAuroraVersion() (string, error) {
...
+		Idle:            int(stat.IdleConns()),
+		// WaitCount is the number of connections for which sql had to wait.
+		// EmptyAcquireCount is the number of connections that pgx acquired after
+		// waiting.  The two are close enough (but race each other).
+		WaitCount: stat.EmptyAcquireCount(),
+		// Time to acquire is close enough to time spent waiting; fudge.
+		WaitDuration:      stat.AcquireDuration(),
+		MaxIdleClosed:     0, // TODO(ariels): Does pgx have this anywhere?
+		MaxIdleTimeClosed: 0, // TODO(ariels): Could generate this with post-connect/release hooks
+		MaxLifetimeClosed: 0, // TODO(ariels): Is this even possible in pgx?
```
nopcoder, line 308: Check with Yoni

db/migration.go
```diff
@@ -158,7 +158,7 @@ func MigrateTo(p params.Database, version uint) error {
 		return err
 	}
 	defer func() {
-		_ = mdb.Close()
+		mdb.Close()
 	}()
```
nopcoder, line 162: can remove the anonymous func

nopcoder (commented):
minor comments - a lot of work - good job!

db/database.go
arielshaqed: Changing to `Get` and `GetPrimitive` (it's not really a primitive, Go doesn't have those, but it sort-of is, and "scalar" would also be inappropriate).

db/database.go
arielshaqed: Paging @johnnyaug for advice here?

db/migration.go
arielshaqed: Good catch, missed that one, thanks!

arielshaqed (commented):
Thanks!

Need @johnnyaug to advise about DB statistics, but if that takes a while might pull first and ask questions later.

nopcoder (approved):

arielshaqed:
Thanks!  Pulling once tests pass.

db/database.go
arielshaqed: Cleared by @johnnyaug : not part of callhome, no known users of field on monitoring (or indeed cause to use it), at least one of these dbstats fields appears to have been added since written (but no easy way to tell - `database/sql` appears to have no version in `go.sum` because it's part of Go).

PR #873, opened 2020-10-28
change migrate behaviour

### LINKED ISSUE #871: Change migrate behaviour
Change migrate behaviour:
1. move auto migrate to setup
2. fail to run lakeFS if migration no up to date
3. add an option to migrate down to a specific version if does not exist

guy-har:
closes #871

### DIFF
--- a/cmd/lakefs/cmd/run.go
+++ b/cmd/lakefs/cmd/run.go
@@ -53,8 +53,10 @@ var runCmd = &cobra.Command{
 		// validate service names and turn on the right flags
 		dbParams := cfg.GetDatabaseParams()
 
-		if err := db.ValidateSchemaUpToDate(dbParams); err != nil {
-			logger.WithError(err).Fatal("Failed on schema validation")
+		if err := db.ValidateSchemaUpToDate(dbParams); errors.Is(err, db.ErrSchemaNotCompatible) {
+			logger.WithError(err).Fatal("Migration version mismatch")
+		} else if err != nil {
+			logger.WithError(err).Warn("Failed on schema validation")
 		}
 		dbPool := db.BuildDatabaseConnection(dbParams)
 		defer func() {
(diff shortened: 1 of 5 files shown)


cmd/lakefs/cmd/run.go
```diff
@@ -53,6 +53,9 @@ var runCmd = &cobra.Command{
 		// validate service names and turn on the right flags
 		dbParams := cfg.GetDatabaseParams()
 
+		if err := db.ValidateSchemaUpToDate(conf.GetDatabaseParams()); err != nil {
+			logger.WithError(err).Error("Failed on schema validation")
```
nopcoder, line 57: Need to exit or use Fatal

db/connect.go
```diff
@@ -47,15 +47,6 @@ func ConnectDB(p params.Database) (Database, error) {
 	conn.SetMaxIdleConns(p.MaxIdleConnections)
 	conn.SetConnMaxLifetime(p.ConnectionMaxLifetime)
 
-	if !p.DisableAutoMigrate {
```
nopcoder, line 50: Missing the code that removes the member from the struct and the loading from the configuration

cmd/lakefs/cmd/run.go
```diff
@@ -53,6 +53,9 @@ var runCmd = &cobra.Command{
 		// validate service names and turn on the right flags
 		dbParams := cfg.GetDatabaseParams()
 
+		if err := db.ValidateSchemaUpToDate(conf.GetDatabaseParams()); err != nil {
```
nopcoder, line 56: 1. `conf` and probably it is not you - but all the commands in lakefs should use `cfg`
2. you don't need to call cfg.GetDatabaseParams() you have them in `dbParams` from line 54

nopcoder (changes requested):

cmd/lakefs/cmd/run.go
guy-har: Considered doing that,
But then no one could run a previous version even if they wanted.

nopcoder (approved):
Cool 👍

PR #877, opened 2020-10-28
Instead of running merge after diff completes, a diff scanner sends changes to merge, that writes them in batches

tzahij:
Changing the diff and merge processes so that they can be interleaved. Diff was changed to be a scanner hat returns diff lines in a Next() loop. and merge is done in batches.
Next step will be to run the merge update operation as a series of small batches. I believe this will solve the problem we have with running SimmilarWeb repository on a 32 GB memory and 64 GB temporary storage.


catalog/cataloger_merge_test.go
```diff
@@ -138,18 +136,19 @@ func TestCataloger_Merge_FromParentConflicts(t *testing.T) {
...
-	}
-	if res.Summary[DifferenceTypeConflict] != len(expectedDifferences) {
-		t.Fatalf("Merge summary conflicts=%d, expected %d", res.Summary[DifferenceTypeConflict], len(expectedDifferences))
-	}
-	differences, _, err := c.Diff(ctx, repository, "master", "branch1", DiffParams{Limit: -1})
-	testutil.MustDo(t, "diff merge changes", err)
-	if !differences.Equal(expectedDifferences) {
-		t.Errorf("Merge differences = %s, expected %s", spew.Sdump(differences), spew.Sdump(expectedDifferences))
-	}
+	//expectedDifferences := Differences{
```
arielshaqed, line 139: No code in comments please.  E.g. https://kentcdodds.com/blog/please-dont-commit-commented-out-code  Also below.
What does this test test?

arielshaqed (commented):
Thanks!
I'm here just for the `sqlx` -> `pgx` + `scany` changes, and it looks great.  It actually should get better -- `pgx` lets you interpolate an SQL array by passing a slice, so at least one place (I commented) that manually constructs a VALUES clause from a slice should work directly.  I just hope Squirrel won't try to destroy that functionality, but given that it's quite a lot of work for Squirrel to do that I think it's worth a try!

catalog/cataloger_merge.go
```diff
@@ -128,128 +165,99 @@ func hasCommitDifferences(tx db.Tx, leftID, rightID int64) (bool, error) {
...
+		_, err = tx.Exec(sql, args...)
+		if err != nil {
+			return err
+		}
 	}
-	if err != nil {
-		return 0, err
+	// insert tombsotnes into parent branch that has a removed entry in its lineage
+	if len(tombstonePaths) > 0 {
+		values := "(VALUES ('" + strings.Join(tombstonePaths, "'),('") + "')) AS t(path)"
```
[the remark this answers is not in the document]
tzahij: Yes, That is certainly great. Tests I did when writing my blog post showed pgx array handling improved performance significantly

catalog/cataloger_merge.go
```diff
@@ -128,128 +165,99 @@ func hasCommitDifferences(tx db.Tx, leftID, rightID int64) (bool, error) {
...
+			return err
+		}
+		_, err = tx.Exec(sql, args...)
+		if err != nil {
+			return err
+		}
 	}
-	if err != nil {
-		return 0, err
+	// insert tombsotnes into parent branch that has a removed entry in its lineage
```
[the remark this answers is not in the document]
tzahij: 10x

catalog/cataloger_merge.go
```diff
@@ -128,128 +165,99 @@ func hasCommitDifferences(tx db.Tx, leftID, rightID int64) (bool, error) {
...
+	if err != nil {
 		return err
 	}
-
-	childNewLineage := strconv.FormatInt(int64(parentLastCommitID), 10)
-	if len(parentLastLineage) > 0 {
-		childNewLineage += "," + parentLastLineage
+	if relation == RelationTypeFromParent {
+		err = tx.Get(&parentLastLineage, `SELECT DISTINCT ON (branch_id) ARRAY_TO_STRING(lineage_commits,',') FROM catalog_commits
+												WHERE branch_id = $1 AND merge_type = 'from_parent' ORDER BY branch_id,commit_id DESC`, leftID)
```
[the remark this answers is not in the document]
tzahij: Do no know why github added spaces - It is nicely alligned in my source in goland

catalog/cataloger_merge_test.go
tzahij: yep

catalog/cataloger_merge.go
```diff
@@ -261,3 +269,7 @@ func (c *cataloger) mergeNonDirect(_ context.Context, _ db.Tx, previousMaxCommit
 	}).Debug("Merge non direct - feature not supported")
 	return ErrFeatureNotSupported
 }
+
+func int64Str(l int64) string {
```
nopcoder, line 273: too much - I think it will be clear to just see the function you call or just use fmt.Sprintf

catalog/cataloger_merge.go
```diff
@@ -128,128 +165,99 @@ func hasCommitDifferences(tx db.Tx, leftID, rightID int64) (bool, error) {
...
 	}
-
-	// do the merge based on the relation
-	previousMaxCommitID, err := getLastCommitIDByBranchID(tx, rightID)
-	if err != nil {
-		return 0, err
+	// apply changes
+	if len(paths) > 0 {
+		// set entries that exist in the right branch as deleted by entries that were removed or changed
+		setMaxCommit := sq.Update("catalog_entries").
```
nopcoder, line 188: use `psql`

catalog/cataloger_merge.go
```diff
@@ -128,128 +165,99 @@ func hasCommitDifferences(tx db.Tx, leftID, rightID int64) (bool, error) {
...
-	case RelationTypeNotDirect:
-		err = c.mergeNonDirect(ctx, tx, previousMaxCommitID, nextCommitID, leftID, rightID, committer, msg, metadata)
-	default:
-		return 0, ErrUnsupportedRelation
+	if len(ctidArray) > 0 {
+		// copy entries from left to right
+		internalSelect := sq.Select(int64Str(rightID), "path", "physical_address", "creation_date", "size", "checksum", "metadata", int64Str(int64(nextCommitID))).
+			From("catalog_entries").
+			Where(sq.Eq{"ctid": ctidArray})
+		copyEntries := sq.Insert("catalog_entries").
```
nopcoder, line 206: psql

catalog/cataloger_merge.go
```diff
@@ -128,128 +165,99 @@ func hasCommitDifferences(tx db.Tx, leftID, rightID int64) (bool, error) {
...
-		err = c.mergeFromParent(ctx, tx, previousMaxCommitID, nextCommitID, leftID, rightID, committer, msg, metadata)
-	case RelationTypeFromChild:
-		err = c.mergeFromChild(ctx, tx, previousMaxCommitID, nextCommitID, leftID, rightID, committer, msg, metadata)
-	case RelationTypeNotDirect:
-		err = c.mergeNonDirect(ctx, tx, previousMaxCommitID, nextCommitID, leftID, rightID, committer, msg, metadata)
-	default:
-		return 0, ErrUnsupportedRelation
+	if len(ctidArray) > 0 {
+		// copy entries from left to right
+		internalSelect := sq.Select(int64Str(rightID), "path", "physical_address", "creation_date", "size", "checksum", "metadata", int64Str(int64(nextCommitID))).
```
nopcoder, line 203: 1. why do we need a static column with rightID if we have the value and we can use it in the outer query where part
2. nextCommitID got a type CommitID - you can add String() method to it so you can capture this one once.

catalog/cataloger_merge.go
```diff
@@ -128,128 +165,99 @@ func hasCommitDifferences(tx db.Tx, leftID, rightID int64) (bool, error) {
...
+		}
 	}
-	if err != nil {
-		return 0, err
+	// insert tombsotnes into parent branch that has a removed entry in its lineage
+	if len(tombstonePaths) > 0 {
+		values := "(VALUES ('" + strings.Join(tombstonePaths, "'),('") + "')) AS t(path)"
+		sql := `INSERT INTO catalog_entries (branch_id,path,physical_address,size,checksum,metadata,min_commit,max_commit)
+				SELECT $1,path,'',0,'','{}',$2,0 FROM ` + values
+		_, err := tx.Exec(sql, rightID, previousMaxCommitID)
```
nopcoder, line 223: why not use squirrel in this case - preferred over the string concat

catalog/cataloger_diff.go
```diff
@@ -115,12 +106,24 @@ func (c *cataloger) Diff(ctx context.Context, repository string, leftReference s
 				AdditionalFields: params.AdditionalFields,
 			},
 		}
-		err = c.doDiff(ctx, tx, diffParams)
+		scanner, err := c.newDiffScanner(tx, diffParams)
 		if err != nil {
 			return nil, err
 		}
+		clearChecksumField := !stringIsInSlice(params.AdditionalFields, DBEntryFieldChecksum)
```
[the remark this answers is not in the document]
tzahij: By default, the scanner reads only the fields that exist in the unique key - branch_id,path, min and max commit.
Checksum are needed for the diff, so we have to always retrieve them when using diff.
If the caller of the diff did not ask for them - he/she will not get then

catalog/cataloger_merge.go
```diff
@@ -128,128 +165,99 @@ func hasCommitDifferences(tx db.Tx, leftID, rightID int64) (bool, error) {
...
+		if len(parentLastLineage) > 0 {
+			*childNewLineage += "," + parentLastLineage
+		}
 	}
-
 	_, err = tx.Exec(`INSERT INTO catalog_commits (branch_id, commit_id, previous_commit_id,committer, message, creation_date, metadata, merge_type, merge_source_branch, merge_source_commit,
                      lineage_commits)
-		VALUES ($1,$2,$3,$4,$5,transaction_timestamp(),$6,'from_parent',$7,$8,string_to_array($9,',')::bigint[])`,
-		childID, nextCommitID, previousMaxCommitID, committer, msg, metadata, parentID, parentLastCommitID, childNewLineage)
+		VALUES ($1,$2,$3,$4,$5,transaction_timestamp(),$6,$10,$7,$8,string_to_array($9,',')::bigint[])`,
```
nopcoder, line 253: I would keep the $ numbering with the parameter order

catalog/cataloger_merge.go
```diff
@@ -128,128 +165,99 @@ func hasCommitDifferences(tx db.Tx, leftID, rightID int64) (bool, error) {
...
-	childNewLineage := strconv.FormatInt(int64(parentLastCommitID), 10)
-	if len(parentLastLineage) > 0 {
-		childNewLineage += "," + parentLastLineage
+	if relation == RelationTypeFromParent {
+		err = tx.Get(&parentLastLineage, `SELECT DISTINCT ON (branch_id) ARRAY_TO_STRING(lineage_commits,',') FROM catalog_commits
+												WHERE branch_id = $1 AND merge_type = 'from_parent' ORDER BY branch_id,commit_id DESC`, leftID)
+		if err != nil && !errors.As(err, &db.ErrNotFound) {
+			return err
+		}
+		t := int64Str(int64(leftLastCommitID))
```
nopcoder, line 245: Avoid use of pointers

catalog/cataloger_merge.go
nopcoder, line 253: can't we pass a slice of []int64 as childNewLineage?

nopcoder (changes requested):

catalog/cataloger_diff.go
```diff
@@ -130,24 +133,25 @@ func (c *cataloger) Diff(ctx context.Context, repository string, leftReference s
...
 }
 
-// doDiff internal implementation of the actual diff. limit <0 will scan the complete branch
-func (c *cataloger) doDiff(ctx context.Context, tx db.Tx, params *doDiffParams) error {
+func (c *cataloger) newDiffScanner(tx db.Tx, params *doDiffParams) (*diffScanner, error) {
+	var scanner *diffScanner
+	var err error
 	if params == nil {
-		return fmt.Errorf("doDiff params are required: %w", ErrInvalidValue)
+		return nil, fmt.Errorf("doDiff params are required: %w", ErrInvalidValue)
```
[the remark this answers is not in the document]
nopcoder: True, same comment I had - prefer not to use pointers in this case unless it is a big structure, but this one will not blow our stack and as far as I know, there is no guaranty it will copy (performance) but we should assume we got a copy.

catalog/cataloger_diff.go
```diff
@@ -115,12 +106,24 @@ func (c *cataloger) Diff(ctx context.Context, repository string, leftReference s
...
+		clearChecksumField := !stringIsInSlice(params.AdditionalFields, DBEntryFieldChecksum)
+		differences := make(Differences, 0, params.Limit+1)
+		for scanner.Next() {
+			v := scanner.Value()
+			var d Difference
+			d.Entry = v.Entry
+			d.Type = v.DiffType
+			if clearChecksumField {
+				d.Entry.Checksum = ""
+			}
```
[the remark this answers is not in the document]
tzahij: OK

catalog/cataloger_diff.go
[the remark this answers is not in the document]
tzahij: OK

catalog/cataloger_merge.go
tzahij: Then I can not use debugSqelizer - non-starter for me

catalog/cataloger_merge_test.go
```diff
@@ -138,18 +136,19 @@ func TestCataloger_Merge_FromParentConflicts(t *testing.T) {
...
+	//}
+	//todo: fix test
+	//if res.Summary[DifferenceTypeConflict] != len(expectedDifferences) {
+	//	t.Fatalf("Merge summary conflicts=%d, expected %d", res.Summary[DifferenceTypeConflict], len(expectedDifferences))
+	//}
+	//differences, _, err := c.Diff(ctx, repository, "master", "branch1", DiffParams{Limit: -1})
+	//testutil.MustDo(t, "diff merge changes", err)
+	//if !differences.Equal(expectedDifferences) {
+	//	t.Errorf("Merge differences = %s, expected %s", spew.Sdump(differences), spew.Sdump(expectedDifferences))
+	//}
```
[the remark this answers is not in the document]
tzahij: Merge behavior changed as we talked - we do not return the summary on conflict

catalog/errors.go
```diff
@@ -28,4 +28,7 @@ var (
 	ErrReadEntryTimeout              = errors.New("read entry timeout")
 	ErrMissingDiffResultsIDInContext = errors.New("missing diff results id in context")
 	ErrInvalidValue                  = errors.New("invalid value")
+	ErrNonDirectNotSupported         = errors.New("non direct diff not supported")
+	ErrNonDirectMergeNotSupported    = errors.New("merge suported only between branches that are parent-child or child-parent")
+	ErrSameBranchMergeNotSupported   = errors.New("same branch merge not supported")
```
nopcoder, line 33: 1. Use ErrFeatureNotSupported (%w) when you build these errors as all re from this type
2. typo: supported
3. ErrNonDirectNotSupported - missing "branch"

catalog/cataloger_diff.go
```diff
@@ -115,455 +102,271 @@ func (c *cataloger) Diff(ctx context.Context, repository string, leftReference s
...
+			return nil, ErrNonDirectNotSupported
+		}
+		scanner, err := NewDiffScanner(tx, diffParams, relation)
+		if err != nil {
+			return nil, err
+		}
+		differences := make(Differences, 0, params.Limit+1)
+		for scanner.Next() {
+			v := scanner.Value()
+			differences = append(differences, Difference{v.Entry, v.DiffType})
```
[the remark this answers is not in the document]
tzahij: change to embedded type

catalog/cataloger_diff.go
```diff
@@ -718,23 +485,38 @@ func (c *cataloger) getRefsRelationType(tx db.Tx, params *doDiffParams) (Relatio
...
-		// source exists, target not - removed
-		return DifferenceTypeRemoved
 	}
-	// source exists and no target - added
-	if targetEnt == nil || targetEnt.IsDeleted() {
-		return DifferenceTypeAdded
+	return c.ParentEffectiveCommit
+}
+
+func stringIsInSlice(slice []string, val string) bool {
```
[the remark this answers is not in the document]
tzahij: remove this function

catalog/cataloger_merge.go
tzahij: Done. Removed function int64Str

catalog/cataloger_merge.go
```diff
@@ -128,136 +160,101 @@ func hasCommitDifferences(tx db.Tx, leftID, rightID int64) (bool, error) {
...
-		return err
-	}
-
-	// DifferenceTypeChanged or DifferenceTypeAdded - create entries into this commit based on parent branch
-	_, err = tx.Exec(`INSERT INTO catalog_entries (branch_id,path,physical_address,creation_date,size,checksum,metadata,min_commit)
-				SELECT $1,path,physical_address,creation_date,size,checksum,metadata,$2 AS min_commit
-				FROM catalog_entries e
-				WHERE e.ctid IN (SELECT entry_ctid FROM `+diffResultsTableName+` WHERE diff_type IN ($3,$4))`,
-		parentID, nextCommitID, DifferenceTypeAdded, DifferenceTypeChanged)
+func InsertMergeCommit(tx db.Tx, relation RelationType, leftID int64, rightID int64, nextCommitID CommitID, previousMaxCommitID CommitID, committer string, msg string, metadata Metadata) error {
```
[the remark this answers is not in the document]
tzahij: Done

catalog/cataloger_merge.go
tzahij: Removed

catalog/cataloger_merge.go
```diff
@@ -128,136 +160,101 @@ func hasCommitDifferences(tx db.Tx, leftID, rightID int64) (bool, error) {
...
-		"commit_id":      previousMaxCommitID,
-		"next_commit_id": nextCommitID,
-		"left_id":        leftID,
-		"right_id":       rightID,
-		"committer":      committer,
-		"msg":            msg,
-	}).Debug("Merge non direct - feature not supported")
-	return ErrFeatureNotSupported
+func int64Str(l int64) string {
+	return strconv.FormatInt(l, 10)
```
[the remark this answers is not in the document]
tzahij: Used it, looks good

catalog/cataloger_merge.go
tzahij: Used your version

catalog/cataloger_merge.go
```diff
@@ -128,136 +160,101 @@ func hasCommitDifferences(tx db.Tx, leftID, rightID int64) (bool, error) {
...
-
-	// DifferenceTypeChanged or DifferenceTypeAdded - create entries into this commit based on parent branch
-	_, err = tx.Exec(`INSERT INTO catalog_entries (branch_id,path,physical_address,creation_date,size,checksum,metadata,min_commit)
-				SELECT $1,path,physical_address,creation_date,size,checksum,metadata,$2 AS min_commit
-				FROM catalog_entries e
-				WHERE e.ctid IN (SELECT entry_ctid FROM `+diffResultsTableName+` WHERE diff_type IN ($3,$4))`,
-		parentID, nextCommitID, DifferenceTypeAdded, DifferenceTypeChanged)
+func InsertMergeCommit(tx db.Tx, relation RelationType, leftID int64, rightID int64, nextCommitID CommitID, previousMaxCommitID CommitID, committer string, msg string, metadata Metadata) error {
+	var err error
+	var childNewLineage *string
```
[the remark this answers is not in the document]
tzahij: Used your version

catalog/cataloger_merge.go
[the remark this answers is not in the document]
arielshaqed: ## :warning: Given that path is user-controlled, this looks like a vector for an SQL injection attack vector by any user with permission to write a file to a branch.

Can you expand on this:
> Tried to use pgx array support, but it did not work for me

I discovered that it works in the export config stuff, e.g. https://github.com/treeverse/lakeFS/blob/29ae1cab03a8ef921b54696ad3a8179aa121d829/catalog/cataloger_export.go#L131 interpolates a `StringArray` type into an `INSERT`.
But failing that, we should wrap [`pgx.Connection.QuoteString`](https://pkg.go.dev/gopkg.in/jackc/pgx.v1#Connection.QuoteString) as part of `DB` to allow for safer quoting.  This version is unsafe (because of the quotes) and hard to read (because I need to wrap my head around the open and close brackets and quotes at each part of `strings.Join`.

catalog/cataloger_merge.go
arielshaqed: It is hard to tell without further details.  However if I might guess, interpolating a `StringArray` gives you VALUES of an row shape and you would like it to have a columns shape.  One way around it is:
```sql
ariels@postgres=# select unnest(array [2,3,4,5]);
 unnest 
--------
      2
      3
      4
      5
(4 rows)
```

Another is to use `CopyFrom` support (which is kinda why we have to use `pgx` anyway).

catalog/cataloger_merge.go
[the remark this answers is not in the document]
arielshaqed: Seems like you didn't push that commit: still a capital `i`.

catalog/cataloger_merge.go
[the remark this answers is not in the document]
arielshaqed: It's probably a tab.  Go solved the tabs vs. spaces debate by proclaiming that all tabs are 8 spaces (see e.g. this documentary https://youtu.be/cowtgmZuai0?t=95).
all-time winner of the editor wars

catalog/cataloger_merge_test.go
arielshaqed: Re-open: no code in comments please.

arielshaqed (changes requested):
Thanks!
1. I think from some of your comment replies that you may have missed pushing some commits.
1. Must-have changes: apparent SQL injection vector for any user with permission to write a file to lakeFS.

catalog/cataloger_diff.go
```diff
@@ -11,13 +10,6 @@ import (
...
 const DiffMaxLimit = 1000
 
-type diffResultRecord struct {
-	TargetEntryNotInDirectBranch bool // the entry is reflected via lineage, NOT in the branch itself
-	DiffType                     DifferenceType
-	Entry                        Entry   // Partially filled. Path is always set.
-	EntryCtid                    *string // CTID of the modified/added entry. Do not use outside of catalog diff-by-iterators. https://github.com/treeverse/lakeFS/issues/831
-}
-
 type diffEffectiveCommits struct {
```
nopcoder, line 13: should also move to DiffScanner

catalog/db_diff_scanner.go
```diff
@@ -0,0 +1,366 @@
...
+type doDiffParams struct {
+	DiffParams
+	Repository    string
+	LeftCommitID  CommitID
+	LeftBranchID  int64
+	RightCommitID CommitID
+	RightBranchID int64
+}
+
+type diffEvaluator func(s *DiffScanner, leftEntry *DBScannerEntry, rightEntry *DBScannerEntry) DifferenceType
```
nopcoder, line 19: The first function argument - 's' `*DiffScanner` should be removed

catalog/db_diff_scanner.go
```diff
@@ -0,0 +1,366 @@
...
+func NewDiffScanner(tx db.Tx, params doDiffParams) (*DiffScanner, error) {
+	relation, err := getRefsRelationType(tx, params)
+	if err != nil {
+		return nil, err
+	}
+	if relation == RelationTypeNotDirect {
+		return nil, ErrNonDirectNotSupported
+	}
+	scanner := &DiffScanner{Relation: relation,
+		params: params}
```
nopcoder, line 43:
```suggestion
	scanner := &DiffScanner{
		Relation: relation,
		params: params,
	}
```

catalog/db_diff_scanner.go
```diff
@@ -0,0 +1,366 @@
...
+		AdditionalFields: prepareDiffAdditionalFields(params.AdditionalFields),
+	}
+	s.leftScanner = NewDBLineageScanner(tx, params.LeftBranchID, params.LeftCommitID, &scannerOpts)
+	s.rightScanner = NewDBLineageScanner(tx, params.RightBranchID, params.RightCommitID, &scannerOpts)
+	return s, nil
+}
+
+func (s *DiffScanner) Next() bool {
+	for s.leftScanner.Next() {
+		// is parent element is relevant
```
nopcoder, line 115: think we can remove this comment - not relevant, right?

catalog/diff.go
```diff
@@ -11,10 +11,16 @@ const (
 )
 
 type Difference struct {
-	Entry                // Partially filled.
+	Entry                // Partially filled. Path is always set.
 	Type  DifferenceType `db:"diff_type"`
 }
 
+type diffResultRecord struct {
```
nopcoder, line 18: The type is still private, other packages will have odd behavior while trying to use the iterator returned Value()

nopcoder (changes requested):

catalog/db_scanner.go
```diff
@@ -15,12 +15,11 @@ type DBScanner interface {
...
-	for ent == nil || ent.Path < p {
+func ScanDBEntriesUntil(s DBScanner, p string) (*DBScannerEntry, error) {
+	for s.Value() == nil || s.Value().Path < p {
 		if !s.Next() {
 			return nil, s.Err()
 		}
-		ent = s.Value()
 	}
-	return ent, nil
+	return s.Value(), nil
```
nopcoder, line 24: 1. Can reduce the checks done calling Value multiple times by capture the value
2. No need to return the entry we just moved the scanner - caller can call Value()
```
ent := s.Value()
for ent == nil || ent.Path < p {
  if !s.Next() {
    return s.Err()
  }
  ent = s.Value()
}
return nil
```

catalog/cataloger_merge.go
```diff
@@ -209,10 +208,9 @@ func applyDiffChangesToRightBranch(tx db.Tx, mergeBatch mergeBatchRecords, previ
 	}
 	// insert tombstones into parent branch that has a removed entry in its lineage
 	if len(tombstonePaths) > 0 {
-		values := "(VALUES ('" + strings.Join(tombstonePaths, "'),('") + "')) AS t(path)"
 		sql := `INSERT INTO catalog_entries (branch_id,path,physical_address,size,checksum,metadata,min_commit,max_commit)
-				SELECT $1,path,'',0,'','{}',$2,0 FROM ` + values
-		_, err := tx.Exec(sql, rightID, previousMaxCommitID)
+				SELECT $1,path,'',0,'','{}',$2,0 FROM(SELECT * FROM UNNEST($3::text []))t(path)`
```
nopcoder, line 212: Nice trick - I usually use a builder to build the expression

nopcoder (changes requested):
Small change to consider - but looks much better now

nopcoder (approved):

catalog/db_branch_scanner.go
```diff
@@ -35,13 +42,49 @@ func NewDBBranchScanner(tx db.Tx, branchID int64, commitID CommitID, opts *DBSca
...
+	var commits []string
+	var commitsWhere string
+	if commitID == UncommittedID || commitID == CommittedID {
+		branchMaxCommitID = MaxCommitID
+	} else {
+		branchMaxCommitID = commitID
+	}
+	// commit_id name is changed so that sorting will be performed on the numeric value, not the string value (where "10" is less than "2")
+	sql := "SELECT commit_id::text as str_commit_id FROM catalog_commits WHERE branch_id = $1 AND commit_id <= $2 ORDER BY commit_id limit $3"
+	err := tx.Select(&commits, sql, branchID, branchMaxCommitID, BranchScannerMaxCommitsInFilter+1)
```
arielshaqed, line 62: Doesn't this include #852?  That one still has outstanding comments -- can we please resolve those first?

arielshaqed (changes requested):
Please let's finish #852 before including it in another commit.

arielshaqed (approved):
#852 will be going in, so my objection (that this would introduce that change) is gone; I am no longer requesting changes to this PR, only that it go in *after* #852.
Thanks for your patience!

tzahij:
After doing merge of 178 million entries - both as child to parent and parent to child

PR #879, opened 2020-10-28
Fix/clientside validation in repo creation 833

### LINKED ISSUE #833: Wrong UI error when repo name contains underscore (on repo creation)

Sufiyan1997:
Closes #833 

Added client side validation while creating a repository as follows:

- Repository name should match `^[a-z0-9][a-z0-9-]{2,62}$`
- Storage namespace should match `^(s3|gs|mem|local|transient):\/.*$`
- Default branch name should not be empty


webui/src/components/RepositoryCreateForm.js
```diff
@@ -51,20 +64,35 @@ export const RepositoryCreateForm = connect(
             <Form.Group as={Row} controlId="id">
                 <Form.Label column sm={fieldNameOffset}>Repository ID</Form.Label>
                 <Col sm={sm}>
-                    <Form.Control type="text" autoFocus ref={repoIdField} onChange={checkValidity}/>
+                    <Form.Control type="text" autoFocus ref={repoIdField} onChange={checkRepoValidity}/>
+                    {!repoValid &&
+                        <Form.Text className="text-danger">
+                            Min 2 characters. Only Alpahnumeric characters and '-' allowed.
```
itaiad200, line 70: Add 'lowercase' to the error message.

webui/src/components/RepositoryCreateForm.js
```diff
@@ -18,20 +18,33 @@ export const RepositoryCreateForm = connect(
...
-            storageNamespaceField.current.value.length === 0 ||
-            defaultBranchField.current.value.length === 0) {
-            setFormValid(false);
-            return;
-        }
-        setFormValid(true);
+    const checkRepoValidity = () => {
+        const isRepoValid = repoValidityRegex.test(repoIdField.current.value)
+        setRepoValid(isRepoValid);
+        setFormValid(isRepoValid);
```
itaiad200, line 35: Here and elsewhere.
This is wrong, form validity should be AND of all three, not the state of the last validity function running.

itaiad200 (changes requested):

webui/src/components/RepositoryCreateForm.js
```diff
@@ -18,20 +18,33 @@ export const RepositoryCreateForm = connect(
     };
   })(({ error, onSubmit, onCancel, create, sm = 6, config }) => {
     const fieldNameOffset = 3;
+    const repoValidityRegex = /^[a-z0-9][a-z0-9-]{2,62}$/;
+    const storageNamespaceValidityRegex = /^(s3|gs|mem|local|transient):\/.*$/;
 
     const [formValid, setFormValid] = useState(false);
+    const [repoValid, setRepoValid] = useState(true);
```
itaiad200, line 25: This should start as false since it's empty, right?

webui/src/components/RepositoryCreateForm.js
```diff
@@ -18,20 +18,33 @@ export const RepositoryCreateForm = connect(
...
+    const [storageNamespaceValid, setStorageNamespaceValid] = useState(true);
```
itaiad200, line 26: Here too.

webui/src/components/RepositoryCreateForm.js
Sufiyan1997: Ideally yes. But if it starts as false then the error messages will be displayed the first time the modal opens. And as we have initialized `formValid` as false, it will anyways prevent the user from submitting the form without any input.

itaiad200: Gotcha, that will do.

itaiad200 (approved):

PR #880, opened 2020-10-29
Add current export state DB API to cataloger

arielshaqed:
Add DB cataloger API to manage current export state.  This takes an actual _transaction_, to allow adding tasks within that same transaction.  Otherwise racing occurs when 2 exports are triggered concurrently.

### DIFF
--- a/ddl/000009_export_current.down.sql
+++ b/ddl/000009_export_current.down.sql
@@ -1,0 +1,6 @@
+BEGIN;
+
+DROP TABLE catalog_branches_export_state;
+DROP TYPE catalog_branches_export_status;
+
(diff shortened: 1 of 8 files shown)


arielshaqed:
Thanks!  Fixed and will pull once tests pass.

ddl/000009_export_current.down.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+
+DROP TABLE catalog_branches_export_state;
```
guy-har, line 3: If exists

guy-har (approved):
Looks good 👍

PR #883, opened 2020-10-30
Update README.md

vinayak017:
There were some spelling mistakes in this file. So I updated it.


arielshaqed:
Null change dropped.  Makes MD worse not better.

PR #884, opened 2020-10-30
Update NOTICE

vinayak017:
There were some spelling mistakes in this file. So I updated it.


arielshaqed:
This is the Apache license, verbatim.  We are unable to accept changes to it as they may have legal and other implications.
If you believe that the Apache should be changed, please suggest changes top level to the Apache project.
Thanks!

PR #887, opened 2020-10-31
Diff working in a separate go routine from merge. Parallel diff and merge

tzahij:


catalog/cataloger_merge.go
```diff
@@ -52,49 +54,46 @@ func (c *cataloger) Merge(ctx context.Context, repository, leftBranch, rightBran
...
 		var rowsCounter int
-		for scanner.Next() {
-			v := scanner.Value()
-			summary[v.Type]++
-			rowsCounter++
-			if v.Type == DifferenceTypeConflict {
-				return nil, ErrConflictFound
+		for buf := range mergeBatchChan {
+			if buf.err != nil {
+				workerExitFlag = true
```
arielshaqed, line 78: This is a race condition per https://golang.org/ref/mem: there is no happens-before relationship between the write here and the read on [line 149](https://github.com/treeverse/lakeFS/pull/887/files#diff-c9a39b7db07f5cc3f023b9c866919ecf8227e2a8fe1702ffcd941f38b9c08fe1R149).
The easiest way to do this is to pass a channel and close it here, and have the worker shut down when the channel closes.  You might use an atomic variable here, but that is more difficult.

catalog/cataloger_merge.go
```diff
@@ -128,6 +125,43 @@ func (c *cataloger) Merge(ctx context.Context, repository, leftBranch, rightBran
...
 }
 
+func drainChannel(channel chan mergeBatchRecords) {
+	// If merge just exited on error, than the diffWorker could continue sending to channel until the buffer
+	// got full, and diffWorker would stay in memory, in the middle of a transaction - something that
+	// may create problems after a while.
+	// channel draining is needed so that diffWorker does not get stuck, and  can exit the transaction
+	for ok := true; ok; { // drain the channel, till it closes
+		_, ok = <-channel
+	}
```
arielshaqed, line 135:
```suggestion
	for _ = <-channel {
	}
```

arielshaqed (changes requested):
Exit worker variable seems racy.  @nopcoder any idea why `go test --race` doesn't pick it up?

arielshaqed:
> Exit worker variable seems racy. @nopcoder any idea why `go test --race` doesn't pick it up?

n/m, not an issue with race detection.  https://codecov.io/gh/treeverse/lakeFS/pull/887/diff?src=pr&el=tree#D2-150 says that line never ran in a test.  We should have a test that does run it.

catalog/cataloger_merge.go
```diff
@@ -52,49 +54,46 @@ func (c *cataloger) Merge(ctx context.Context, repository, leftBranch, rightBran
 				Limit: -1,
 			},
 		}
-
-		scanner, err := NewDiffScanner(tx, params)
+		relation, err := getRefsRelationType(tx, params)
```
nopcoder, line 57: guess this code need to get the update from the parent branch first - we can get the refs relation from the scanner

catalog/cataloger_merge.go
nopcoder:
```go
for range channel {
}
```

catalog/cataloger_merge.go
```diff
@@ -11,10 +11,14 @@ import (
 )
 
 const (
-	MergeBatchSize = 256
+	MergeBatchSize       = 256
+	MergeBatchChanBuffer = 10
 )
 
-type mergeBatchRecords []*diffResultRecord
+type mergeBatchRecords struct {
```
nopcoder, line 18: why use channel of batch operations - we have diff iterators that produce a channel of changes, we can batch them as we did before (if we like) - we already as part of one big transaction, so the batching can be done per kind of operation on the merge level.

catalog/cataloger_merge.go
```diff
@@ -52,49 +54,46 @@ func (c *cataloger) Merge(ctx context.Context, repository, leftBranch, rightBran
...
 		}
-		// do the merge based on the relation
 		previousMaxCommitID, err := getLastCommitIDByBranchID(tx, rightID)
 		if err != nil {
 			return nil, err
 		}
-
-		differences := make(mergeBatchRecords, 0, MergeBatchSize)
+		mergeBatchChan := make(chan mergeBatchRecords, MergeBatchChanBuffer)
+		var workerExitFlag bool
```
nopcoder, line 73: using variables without locking to communicate state between goroutine is a race issue

catalog/cataloger_merge.go
```diff
@@ -52,49 +54,46 @@ func (c *cataloger) Merge(ctx context.Context, repository, leftBranch, rightBran
...
+		go c.diffWorker(params, mergeBatchChan, &workerExitFlag)
```
nopcoder, line 74: the worker needs to accept the call 'ctx' and check/honor it in case of cancelation.

nopcoder, line 74: Consider - moving this code into a function/method that will make channels - one for differences and the second for an error. The function will make the channels, start the goroutine and defer close both channels when the worker is done.
We should pass a cancelation context based on the request context.
The worker should check the context while processing the diff scanners, on any error sending the error on an errors channel and quit.
The function will return the two channels to the caller and you can process the differences (batch or not batch in a transaction).
If the merge execution fails with an error, it would cancel the context given the diff.

catalog/errors.go
```diff
@@ -31,4 +31,5 @@ var (
 	ErrNonDirectNotSupported         = errors.New("non direct diff not supported")
 	ErrNonDirectMergeNotSupported    = errors.New("merge suported only between branches that are parent-child or child-parent")
 	ErrSameBranchMergeNotSupported   = errors.New("same branch merge not supported")
+	ErrMergeWorkerRollback           = errors.New("merge worker exit because of diff error")
```
nopcoder, line 34: can remove this one?

catalog/cataloger_merge.go
```diff
@@ -128,6 +125,43 @@ func (c *cataloger) Merge(ctx context.Context, repository, leftBranch, rightBran
...
+	// got full, and diffWorker would stay in memory, in the middle of a transaction - something that
+	// may create problems after a while.
+	// channel draining is needed so that diffWorker does not get stuck, and  can exit the transaction
+	for ok := true; ok; { // drain the channel, till it closes
+		_, ok = <-channel
+	}
+}
+
+func (c *cataloger) diffWorker(params doDiffParams, mergeBatchChan chan mergeBatchRecords, exitFlag *bool) {
+	_, _ = c.db.Transact(func(tx db.Tx) (interface{}, error) {
```
nopcoder, line 139: General question - does the fact that we do diff in one isolation and apply the changes in the different one can it cause in-consistency issues? example: the diff find a conflict and we applied a change in a transaction we opened that there is a conflict already there.

catalog/cataloger_merge.go
```diff
@@ -128,6 +125,43 @@ func (c *cataloger) Merge(ctx context.Context, repository, leftBranch, rightBran
 	return mergeResult, err
 }
 
+func drainChannel(channel chan mergeBatchRecords) {
```
nopcoder, line 128: If you will communicate the diff scanner work throw 2 channels - data and error (size 1) - the optimization to drain the channel in all the places the scanner failed will be gone.
The merge/caller will check both channels and in case of an error, it will just exit.

nopcoder (changes requested):

catalog/errors.go
tzahij: removed it and two others that are not needed

catalog/cataloger_merge.go
tzahij: An ancient Chinese proverb say that if you think about it for a long enough  - you will find inconsistencies in transactions.

on the entries I see no problem. only committed entries are read from source(left Branch). committed entries are changed only by commit or merge (which is a special commit).
The read transaction takes a snapshot, so even if the source is committed while a merge is running - it will not be reflected.
there is a possibility that a source parent will be merged with grand parent while a merge is running (quite a strange situation). Then there may be a problem because the lineage of the parent will be updated without the diffWorker "seeing" those changes. Then when the merge finishes - It may read the new commited lineage. (depends on serialization mode. As long as we are in serializable mode - it has to happen in the millisecond between start of the merge transaction, and the start of the diffWorker)
we have 3 possible solutions:
1. read the parent lineage by diffWorker, and passing it to the merge.
2. Lock the possibility of cascading from-parent merges
3. Decide that since lineage will go the way of the Dodo - ignore it until we change architecture
I vote 3

catalog/cataloger_merge.go
tzahij: AFAIU, Ctx helps when it is combined with other wait states - e.g. wait on a channel and context. The diffWorker is very simple and has no select statement

catalog/cataloger_merge.go
arielshaqed: This is a race condition.  You are quoting when "a read r of a variable v is _allowed_ to observe a write w to v", and "allowed" is italicised.  The very next paragraph states when it _must_ observe the write: " r is _guaranteed_ to observe w if both of the following hold: (1) w happens before r.[...]".  I dropped (2) because (1) doesn't hold: the write to this memory location does not happen before the read.
The burden of proof to show that the write happens-before is on the author not the reviewer.  This is made difficult by this:
> When multiple goroutines access a shared variable v, they must use synchronization events to establish happens-before conditions that ensure reads observe the desired writes
The only synchronization events that I can see are the channel, but that goes in the wrong direction to establish the desired conditions.

It seems to me that to resolve this, we can either run this code through a test with `-race` (coverage says no test covers this case) or we can find a sequence of happens-before conditions in the memory model that shows that the write is before the read or we can rewrite it to use an explicit synchronization mechanism such as closing a channel.

arielshaqed:
Here's a simplified example of the race condition:
```go
package main

import "testing"

type Task struct {
	Exit   bool
	Output chan int
}

func (t *Task) Run() {
	for i := 0; !t.Exit; i++ {
		t.Output <- i
	}
	close(t.Output)
}

func TestRace(t *testing.T) {
	task := Task{Output: make(chan int, 9999999)}

	go task.Run()

	for i := 0; i < 10000; i++ {
		_ = <-task.Output
	}

	count := 0
	task.Exit = true
	for range task.Output {
		count++
	}

	t.Log(count, " before stop")
}
```
(also attached)

```
ariels@ariels:~/Dev/Tests$ go test -race stop_test.go 
==================
WARNING: DATA RACE
Write at 0x00c000118490 by goroutine 7:
  command-line-arguments.TestRace()
      /home/ariels/Dev/Tests/stop_test.go:27 +0x13c
  testing.tRunner()
      /snap/go/6633/src/testing/testing.go:1123 +0x202

Previous read at 0x00c000118490 by goroutine 16:
  command-line-arguments.(*Task).Run()
      /home/ariels/Dev/Tests/stop_test.go:11 +0x87

Goroutine 7 (running) created at:
  testing.(*T).Run()
      /snap/go/6633/src/testing/testing.go:1168 +0x5bb
  testing.runTests.func1()
      /snap/go/6633/src/testing/testing.go:1439 +0xa6
  testing.tRunner()
      /snap/go/6633/src/testing/testing.go:1123 +0x202
  testing.runTests()
      /snap/go/6633/src/testing/testing.go:1437 +0x612
  testing.(*M).Run()
      /snap/go/6633/src/testing/testing.go:1345 +0x3b3
  main.main()
      _testmain.go:43 +0x236

Goroutine 16 (running) created at:
  command-line-arguments.TestRace()
      /home/ariels/Dev/Tests/stop_test.go:20 +0xdc
  testing.tRunner()
      /snap/go/6633/src/testing/testing.go:1123 +0x202
==================
--- FAIL: TestRace (0.01s)
    stop_test.go:32: 3425  before stop
    testing.go:1038: race detected during execution of test
FAIL
FAIL	command-line-arguments	0.014s
FAIL
```
[stop_test.go.gz](https://github.com/treeverse/lakeFS/files/5480098/stop_test.go.gz)

catalog/cataloger_merge.go
tzahij: Of course the go routine has to read constantly - and it always reads before w wrote. Once w updates it - it will be read and acted upon.(exit the go routine)
A race condition can happen only when more than one party can do something.(one runner can not race himself). This is definitely not the case here.

nopcoder (changes requested):
Discussed F2F

catalog/cataloger_merge.go
nopcoder: This is an exit flag - it means that one goroutine will read the value and one will update the value.
Any unsynchronized access to shared variables is a race.

catalog/cataloger_merge.go
```diff
@@ -71,30 +68,40 @@ func (c *cataloger) Merge(ctx context.Context, repository, leftBranch, rightBran
 		if err != nil {
 			return nil, err
 		}
-		mergeBatchChan := make(chan mergeBatchRecords, MergeBatchChanBuffer)
-		var workerExitFlag bool
-		go c.diffWorker(params, mergeBatchChan, &workerExitFlag)
+
+		mergeCtx, cancel := context.WithCancel(ctx)
```
nopcoder, line 72: Extract this method it will make the code safer by using return and defer to handle the cancellation:
```go
func (c *cataloger) doMerge(ctx context.Context, tx db.Tx, params doDiffParams, mergeResult *MergeResult, previousMaxCommitID CommitID, nextCommitID CommitID, relation RelationType) (int, error) {
	ctx, cancel := context.WithCancel(ctx)
	defer cancel()

	mergeBatchChan, errChan := c.initDiffWorker(ctx, params)
	var rowsCounter int
	for {
		select {
		case buf, ok := <-mergeBatchChan:
			if !ok {
				return rowsCounter, nil
			}
			for _, d := range buf {
				rowsCounter++
				mergeResult.Summary[d.Type]++
				if d.Type == DifferenceTypeConflict {
					return 0, ErrConflictFound
				}
			}
			err := applyDiffChangesToRightBranch(tx, buf, previousMaxCommitID, nextCommitID, params.RightBranchID, relation)
			if err != nil {
				return 0, err
			}
		case err := <-errChan:
			return rowsCounter, err
		}
	}
}
```

catalog/cataloger_merge.go
```diff
@@ -103,9 +110,11 @@ func (c *cataloger) Merge(ctx context.Context, repository, leftBranch, rightBran
 		if rowsCounter == 0 {
 			commitDifferences, err := hasCommitDifferences(tx, leftID, rightID)
 			if err != nil {
+				cancel()
```
nopcoder, line 113: at this point, we already had to call cancel

catalog/cataloger_merge.go
```diff
@@ -103,9 +110,11 @@ func (c *cataloger) Merge(ctx context.Context, repository, leftBranch, rightBran
...
 				return nil, err
 			}
 			if !commitDifferences {
+				cancel()
```
nopcoder, line 117: at this point, we already had to call cancel

nopcoder (approved):

Issue #490, opened 2020-08-16
Diagnostic collector for lakeFS


nopcoder:
Design and implement diagnostics collect for lakeFS.
Collect requirements and update a design document first.
- collect data from lakeFS daemon
- collect schema version information
- collect tables and index sizes
- dump everything into an archive for easy transport

nopcoder:
The first PR handles the part that collects information from the database.
The data we collect helps in finding issues related to schema version, gives size information and etc.
Additional information from the running service still need to be collected in order to integrate the information as part of the diagnostics.

nopcoder:
Closing this one, when new requirements on data we need to collect will be added, we will open an issue and implement.

PR #889, opened 2020-11-01
lakefs diagnostics command

nopcoder:
Part of #490 

lakefs diagnostics - command line that will connect to the database collect information about the lakeFS to help in diagnostics.

Additional information will be collected as part of lakeFS process and will be collected by this command.

### DIFF
--- a/diagnostics/collector.go
+++ b/diagnostics/collector.go
@@ -97,52 +97,55 @@ func (c *Collector) writeQueryContent(ctx context.Context, writer *zip.Writer, n
 }
 
 func (c *Collector) writeRawQueryContent(ctx context.Context, writer *zip.Writer, name string, query string, args ...interface{}) error {
-	rows, err := c.db.Query(query, args...)
-	if err != nil {
-		return fmt.Errorf("execute query: %w", err)
-	}
-	defer rows.Close()
-
-	filename := name + csvFileExt
-	w, err := writer.Create(filename)
-	if err != nil {
-		return fmt.Errorf("new file for table %s - %w", name, err)
-	}
-	csvWriter := csv.NewWriter(w)
-	defer csvWriter.Flush()
-
-	first := true
-	for rows.Next() {
-		if err := ctx.Err(); err != nil {
-			return err
-		}
-		if first {
-			first = false
-			descriptions := rows.FieldDescriptions()
-			if len(descriptions) == 0 {
-				return errNoColumnsFound
-			}
-			cols := make([]string, len(descriptions))
-			for i, fd := range descriptions {
-				cols[i] = string(fd.Name)
-			}
-			if err := csvWriter.Write(cols); err != nil {
-				return fmt.Errorf("write csv header for %s: %w", name, err)
-			}
-		}
-		v, err := rows.Values()
+	_, err := c.db.Transact(func(tx db.Tx) (interface{}, error) {
+		rows, err := tx.Query(query, args...)
 		if err != nil {
-			return err
+			return nil, fmt.Errorf("execute query: %w", err)
 		}
-		record := make([]string, len(v))
-		for i := range v {
-			record[i] = fmt.Sprintf("%v", v[i])
+		defer rows.Close()
+
+		filename := name + csvFileExt
+		w, err := writer.Create(filename)
+		if err != nil {
+			return nil, fmt.Errorf("new file for table %s - %w", name, err)
 		}
-		if err := csvWriter.Write(record); err != nil {
-			return err
+		csvWriter := csv.NewWriter(w)
+		defer csvWriter.Flush()
+
+		first := true
+		for rows.Next() {
+			// write csv header row
+			if first {
+				first = false
+				descriptions := rows.FieldDescriptions()
+				if len(descriptions) == 0 {
+					return nil, errNoColumnsFound
+				}
+				cols := make([]string, len(descriptions))
+				for i, fd := range descriptions {
+					cols[i] = string(fd.Name)
+				}
+				if err := csvWriter.Write(cols); err != nil {
+					return nil, fmt.Errorf("write csv header for %s: %w", name, err)
+				}
+			}
+
+			// format and write data row
+			v, err := rows.Values()
+			if err != nil {
+				return nil, err
+			}
+			record := make([]string, len(v))
+			for i := range v {
+				record[i] = fmt.Sprintf("%v", v[i])
+			}
+			if err := csvWriter.Write(record); err != nil {
+				return nil, err
+			}
 		}
-	}
-	return rows.Err()
+		return nil, rows.Err()
+	}, db.WithContext(ctx), db.ReadOnly())
+	return err
 }
 
 func (c *Collector) writeTableContent(ctx context.Context, writer *zip.Writer, name string) error {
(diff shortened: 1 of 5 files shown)


diagnostics/collector.go
```diff
@@ -0,0 +1,170 @@
...
+	var errs []error
+	contentFromTables := []string{
+		"auth_installation_metadata",
+		"schema_migrations",
+		"pg_stat_database",
+	}
+	for _, tbl := range contentFromTables {
+		err = c.writeTableContent(ctx, writer, tbl)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("write table content for %s %w", tbl, err))
```
itaiad200, line 45:
```suggestion
			errs = append(errs, fmt.Errorf("write table content for %s: %w", tbl, err))
```

diagnostics/collector.go
```diff
@@ -0,0 +1,170 @@
...
+	countFromTables := []string{
+		"catalog_branches",
+		"catalog_commits",
+		"catalog_repositories",
+		"auth_users",
+	}
+	for _, tbl := range countFromTables {
+		err = c.writeTableCount(ctx, writer, tbl)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("write table count for %s %w", tbl, err))
```
itaiad200, line 58:
```suggestion
			errs = append(errs, fmt.Errorf("write table count for %s: %w", tbl, err))
```

diagnostics/collector.go
```diff
@@ -0,0 +1,170 @@
...
+		}
+	}
+
+	err = c.writeRawQueryContent(ctx, writer, "table_sizes", `
+SELECT *, pg_size_pretty(total_bytes) AS total
```
itaiad200, line 63: @tzahij for a quick glance..

diagnostics/collector.go
```diff
@@ -0,0 +1,170 @@
...
+func (c *Collector) writeTableCount(ctx context.Context, writer *zip.Writer, name string) error {
+	q := sq.Select("COUNT(*)").From(name)
+	return c.writeQueryContent(ctx, writer, name+"_count", q)
+}
+
+func (c *Collector) writeErrors(ctx context.Context, writer *zip.Writer, errs []error) error {
+	if len(errs) == 0 {
+		return nil
+	}
+	w, err := writer.Create("errors.log")
```
itaiad200, line 159: If any error occurred, you'll always be writing 2 files, the `output.log` and the `errors.log`. It would be nice if they could be written to the same folder.

diagnostics/collector.go
```diff
@@ -0,0 +1,170 @@
...
+	if err != nil {
+		return err
```
itaiad200, line 161: Add more text, like: "can't write errors file", o.w the returned error itself seems to indicate that everything else was fine.

diagnostics/collector.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	for _, err := range errs {
+		if ctx.Err() != nil {
```
itaiad200, line 164: Why do you check this inside the loop?

diagnostics/collector.go
```diff
@@ -0,0 +1,170 @@
...
+				return fmt.Errorf("write csv header for %s %w", name, err)
+			}
+		}
+		v, err := rows.Values()
+		if err != nil {
+			return err
+		}
+		record := make([]string, len(v))
+		for i := range v {
+			record[i] = fmt.Sprintf("%v", v[i])
```
itaiad200, line 136: Is this csv formatted? Don't you need a `","` between the columns of each row?

diagnostics/collector.go
```diff
@@ -0,0 +1,170 @@
...
+func (c *Collector) writeQueryContent(ctx context.Context, writer *zip.Writer, name string, selectBuilder sq.SelectBuilder) error {
+	q := selectBuilder.Limit(maxRecordsPerQueryCollect)
+	query, args, err := q.ToSql()
+	if err != nil {
+		return fmt.Errorf("query build: %w", err)
+	}
+	return c.writeRawQueryContent(ctx, writer, name, query, args...)
+}
+
+func (c *Collector) writeRawQueryContent(ctx context.Context, writer *zip.Writer, name string, query string, args ...interface{}) error {
```
itaiad200, line 96:
```suggestion
func (c *Collector) writeRawQueryContent(ctx context.Context, zipWriter *zip.Writer, name string, query string, args ...interface{}) error {
```

itaiad200: since you have csvWriter...

diagnostics/collector.go
```diff
@@ -0,0 +1,170 @@
...
+type Collector struct {
+	db db.Database
+}
+
+const (
+	maxRecordsPerQueryCollect = 1000
+	csvFileExt                = ".csv"
+)
+
+var ErrNoColumnsFound = errors.New("no columns found")
```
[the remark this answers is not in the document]
itaiad200: Or even use it at all..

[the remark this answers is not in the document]
nopcoder: Right, at the end I didn't expose all the functionality of this module - so this error is required, but it will only be serialized into errors.log

diagnostics/collector.go
```diff
@@ -0,0 +1,170 @@
...
+	"encoding/csv"
+	"errors"
+	"fmt"
+	"io"
+
+	sq "github.com/Masterminds/squirrel"
+	"github.com/treeverse/lakefs/db"
+)
+
+type Collector struct {
```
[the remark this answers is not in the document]
nopcoder: Adding code doc for the struct and the public method - about the TODOs. I think we discussed to create issues and not add TODO in code, so I'll leave them in the issue

diagnostics/collector.go
nopcoder: @tzahij can you comment if you have or not have any comment related to this query?

diagnostics/collector.go
nopcoder: what is output.log?

diagnostics/collector.go
nopcoder: I can check it once or not pass any context to this function - usually when I pass ctx and the function does something that takes time or in a loop - I condition the context check.
If it is too much I can skip this part and just always write the errors.

diagnostics/collector.go
nopcoder: The csv writer does the CSV formatting -- it will also handle value escape like using ',' in the value.

diagnostics/collector.go
nopcoder: It will be fewer changes to the code if tomorrow I'll add support for tgz and just pass an interface ;)

nopcoder:
@itaiad200 thanks for the review - I've noticed that I added code without tests, so I can do better. Please wait until I'll add some cover.

diagnostics/collector.go
itaiad200: Sorry, I meant 'lakefs-diagnostics.zip', it's better if it's always next to the errors, no?

nopcoder: want to get a zip file and if some of the queries didn't run - I want to see it in the errors.txt found there.

diagnostics/collector.go
itaiad200: Non blocking: you're kind of using the `ctx.Done()` behaviour for loops, which isn't very common.

diagnostics/collector.go
itaiad200: Gotcha

itaiad200 (approved):

diagnostics/collector.go
nopcoder: I'll update the code to check Done before checking an error

PR #892, opened 2020-11-02
Remove unused GCS parameters

nopcoder:
This PR is just to remove the unused parameters, thanks @arielshaqed.
I'm in favor that if we all agree that viper will be the configuration solution for the project each module that will require a configuration will use viper to extract the configuration.
In order not to have it all over the place, and have our services configured by Params/Config struct - we can have a function in the module scope that will use viper and initialized these structs.


arielshaqed (approved):
Yay!  It decreases safely size, what's not to like?
Thanks!

PR #897, opened 2020-11-06
Update README.md

rishabh710:
Addition of table of content and contributing section


arielshaqed:
Thanks!
Our main page lakefs.io should already include a table of contents. It gets auto-generated from this file. So for the time being we shall not be accepting changes that manually add a table of contents.

PR #898, opened 2020-11-08
Fail a task if it retries too many times

arielshaqed:
Previously it would just never be owned or trigger any of its successors, preventing (easy)
failure of the following tasks.


guy-har (approved):
👍

arielshaqed:
Thanks!

PR #901, opened 2020-11-08
Feature/single export

guy-har:
Draft,
Contains todos
Need to add ddl to migration and change to work with one DB instance.


export/export.go
```diff
@@ -0,0 +1,61 @@
...
+func getExportID(repo, branch, commitRef string) (string, error) {
+	nid, err := nanoid.Nanoid()
+	if err != nil {
+		return "", err
+	}
+	return fmt.Sprintf("%s-%s-%s-%s", repo, branch, commitRef, nid), nil
+}
+
+// ExportBranchStart inserts a start task on branch and changes branch export state to pending
+// if export on  already in progress will return error
```
arielshaqed, line 23: Could you rephrase this comment please?  I found it a bit hard to follow -- maybe just adding punctuation will make it easier for me!

export/export_handler.go
```diff
@@ -41,6 +49,121 @@ func PathToPointer(path string) (block.ObjectPointer, error) {
...
+		taskData, err := tasksGenerator.Add(diffs)
+		if err != nil {
+			return err
+		}
+		// add taskData tasks
+		err = h.parade.InsertTasks(context.Background(), taskData)
+		if err != nil {
+			return err
+		}
+		if hasMore && len(diffs) > 0 {
```
arielshaqed, line 103: Can we have `hasMore && len(diffs) == 0`?!  (Note that in that case it looks like an infinite loop, so not sure what this extra caution does here.  If it's a worry, it should be an error message -- maybe even fatal.)

export/tasks_generator.go
```diff
@@ -105,12 +122,19 @@ func (exportID taskIDGenerator) finishedTaskID() parade.TaskID {
 	return parade.TaskID(fmt.Sprintf("%s:finished", exportID))
 }
 
-func (exportID taskIDGenerator) copyTaskID(physicalAddress string) parade.TaskID {
-	return parade.TaskID(fmt.Sprintf("%s:copy:%s", exportID, physicalAddress))
+func (exportID taskIDGenerator) startedTaskID() parade.TaskID {
+	return parade.TaskID(fmt.Sprintf("%s:started", exportID))
```
arielshaqed, line 126: This is a simple past-tense verb, whereas in other task IDs we use an imperative verb.  Could you do that here, too?  (Not sure which imperative form is intended here!)

export/tasks_generator.go
```diff
@@ -236,32 +260,59 @@ type TasksGenerator struct {
...
+		idGen:                 taskIDGenerator(exportID),
+		successTasksGenerator: NewSuccessTasksTreeGenerator(exportID, generateSuccessFor, makeDestination, finishBody),
 	}
 }
 
 // Add translates diffs into many tasks and remembers "generate success" tasks for Finish.  It
 // returns some tasks that can already be added.
 func (e *TasksGenerator) Add(diffs catalog.Differences) ([]parade.TaskData, error) {
-	const initialSize = 1_000
+	const initialSize = 1000
```
arielshaqed, line 315: Why?  I _really_ like this form, it is much easier for me to read large numbers when their digits are grouped in 3s.

export/tasks_generator_test.go
```diff
@@ -327,59 +335,64 @@ func TestTasksGenerator_Simple(t *testing.T) {
 	}
 }
 
+func getGeneratedID(path string) string {
```
arielshaqed, line 338: This copies code from the system under test.  Consider just exporting it -- maybe inside a `ForTesting` object.  Or strip the generated "random" values before comparing.

parade/ddl.sql
```diff
@@ -8,7 +8,7 @@ CREATE TYPE task_status_code_value AS ENUM (
 );
 
 CREATE TABLE IF NOT EXISTS tasks (
-    id VARCHAR(64) NOT NULL PRIMARY KEY, -- nanoid
+    id VARCHAR(128) NOT NULL PRIMARY KEY, -- nanoid
```
arielshaqed, line 11: Note that this "128"\ appears in numerous other places.  And let's get rid of `nanoid` in the comment -- it never was.

Thinking about it, we should probably get rid of the length limitation.  Per https://www.postgresql.org/docs/12/datatype-character.html there is actually no advantage to this limitation, it only saves space when it truncates :scream: .

arielshaqed (commented):
Cool stuff!

I think it's really clean good code.

export/tasks_generator.go
guy-har: GoLands linter threw an exception (it isn't updated to work with this form yet), which caused the auto formatter to stop working. so because it was a small number I allowed myself to change it.
I could change it back

api/api_controller.go
```diff
@@ -2222,6 +2228,29 @@ func (c *Controller) ExportGetContinuousExportHandler() exportop.GetContinuousEx
 	})
 }
 
+func (c *Controller) ExportExecuteContinuousExport() exportop.ExecuteContinuousExportHandler {
+	return exportop.ExecuteContinuousExportHandlerFunc(func(params exportop.ExecuteContinuousExportParams, user *models.User) middleware.Responder {
+		deps, err := c.setupRequest(user, params.HTTPRequest, []permissions.Permission{
+			{
+				Action:   permissions.CreateBranchAction,
```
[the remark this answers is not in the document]
guy-har: Agree, changing it. 
Thanks

catalog/cataloger_export.go
```diff
@@ -136,6 +136,23 @@ func (c *cataloger) PutExportConfiguration(repository string, branch string, con
...
 var ErrExportFailed = errors.New("export failed")
 
+func (c *cataloger) ExportState(repo, branch, newRef string, cb func(oldRef string, state CatalogBranchExportStatus) (newState CatalogBranchExportStatus, newMessage *string, err error)) error {
+	_, err := c.db.Transact(func(tx db.Tx) (interface{}, error) {
+		oldRef, state, err := c.ExportStateMarkStart(tx, repo, branch, newRef)
+		if err != nil {
+			return nil, err
+		}
+
+		newState, newMsg, err := cb(oldRef, state)
```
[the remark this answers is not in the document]
guy-har: In case that `newState != ExportStatusFailed` 
The DB will return an error in case that `msg != nil`

export/export_handler.go
guy-har: I missed that,
Thanks

export/export.go
guy-har: Done

export/tasks_generator_test.go
guy-har: 👍

api/api_controller.go
```diff
@@ -2222,6 +2228,29 @@ func (c *Controller) ExportGetContinuousExportHandler() exportop.GetContinuousEx
...
+				WithPayload(responseErrorFrom(err))
+		}
+		deps.LogAction("execute_continuous_export")
+
+		err = export.ExportBranchStart(deps.Parade, deps.Cataloger, params.Repository, params.Branch)
+		if err != nil {
+			return exportop.NewExecuteContinuousExportDefault(http.StatusInternalServerError).
+				WithPayload(responseErrorFrom(err))
+		}
+		return exportop.NewExecuteContinuousExportCreated()
```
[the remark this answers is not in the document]
guy-har: sounds good, I added it.

api/api_controller.go
```diff
@@ -191,7 +197,7 @@ func (c *Controller) Configure(api *operations.LakefsAPI) {
 
 	api.ExportGetContinuousExportHandler = c.ExportGetContinuousExportHandler()
 	api.ExportSetContinuousExportHandler = c.ExportSetContinuousExportHandler()
-
+	api.ExportExecuteContinuousExportHandler = c.ExportExecuteContinuousExport()
```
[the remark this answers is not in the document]
guy-har: 👌

export/tasks_generator.go
guy-har: 👌

parade/ddl.sql
guy-har: Done, I removed the length limitation

guy-har:
@arielshaqed PTAL

export/export_handler_test.go
```diff
@@ -150,3 +151,45 @@ func TestTouch(t *testing.T) {
...
+	}{
+		{
+			name:                   "one regex",
+			lastKeysInPrefixRegexp: []string{".*\\.success$"},
+			expectTrue:             []string{"a.success", "other.success"},
+			expectFalse:            []string{"dfd", "a.suc", "a.successer"},
+			want:                   nil,
+		},
+		{
+			name:                   "two regexes",
```
arielshaqed, line 172:
```suggestion
			name:                   "two regexps",
```
(one advantage of "regexp" is that you don't have smart alecs suggesting "regexen" as the plural.

arielshaqed (approved):
Thanks!

I think we're there, right?

PR #902, opened 2020-11-08
[parade] Move DDL under ddl/

arielshaqed:
Requires some changes, most notably an ability to get a the pgx pool from a database.  This
should eventually vanish: we could either use pgx more directly, or export enough functionality
from Database to enable Parade directly on top.


guy-har (approved):

arielshaqed:
Thanks!

PR #903, opened 2020-11-08
DB scanner options by value with option to update AdditionalWhere

nopcoder:


tzahij (approved):
A very technical change. Looks good

tzahij (approved):

PR #904, opened 2020-11-08
"mkdir -p" destination directories for local adapter Put, Copy

arielshaqed:
Also try to do it reasonably efficiently: try to create, then try again if the directory might
not exist.  This avoids numerous system calls and speeds up the case of writing numerous files
into the same directory.


guy-har (approved):

arielshaqed:
THANKS! :-)

PR #906, opened 2020-11-09
merge from parent - remove ancestor branches that did not change from last commit

tzahij:
…parent branch

### DIFF
--- a/catalog/db_diff_scanner.go
+++ b/catalog/db_diff_scanner.go
@@ -93,7 +93,7 @@ func (s *DiffScanner) diffFromParent(tx db.Tx, params doDiffParams, scannerOpts 
 	}
 	rightLineage, err := getLineage(tx, params.RightBranchID, UncommittedID)
 	if err != nil {
-		return nil, fmt.Errorf("get chaild lineage failed on :%w", err)
+		return nil, fmt.Errorf("get right branch lineage: %w", err)
 	}
 	// If some ancestor branch commit id is the same for parent and child - then the parent does not need to read it
 	// so it is trimmed from the parent lineage
--- a/catalog/db_diff_scanner.go
+++ b/catalog/db_diff_scanner.go
@@ -95,14 +95,14 @@ func (s *DiffScanner) diffFromParent(tx db.Tx, params doDiffParams, scannerOpts 
 	}
 	// If some ancestor branch commit id is the same for parent and child - then the parent does not need to read it
 	// so it is trimmed from the parent lineage
-	if len(leftLineage) >= 1 {
-		for i := range leftLineage {
-			if leftLineage[i].CommitID == rightLineage[i+1].CommitID {
-				leftLineage = leftLineage[:i]
-				break
-			}
+
+	for i := range leftLineage {
+		if leftLineage[i].CommitID == rightLineage[i+1].CommitID {
+			leftLineage = leftLineage[:i]
+			break
 		}
 	}
+
 	scannerOpts.Lineage = leftLineage
 	s.leftScanner = NewDBLineageScanner(tx, params.LeftBranchID, CommittedID, scannerOpts)
 	scannerOpts.Lineage = rightLineage
--- a/catalog/db_diff_scanner.go
+++ b/catalog/db_diff_scanner.go
@@ -89,7 +89,7 @@ func (s *DiffScanner) diffFromParent(tx db.Tx, params doDiffParams, scannerOpts 
 	}
 	leftLineage, err := getLineage(tx, params.LeftBranchID, CommittedID)
 	if err != nil {
-		return nil, fmt.Errorf("get parent last commit failed: %w", err)
+		return nil, fmt.Errorf("get left branch lineage: %w", err)
 	}
 	rightLineage, err := getLineage(tx, params.RightBranchID, UncommittedID)
 	if err != nil {
--- a/catalog/db_diff_scanner.go
+++ b/catalog/db_diff_scanner.go
@@ -89,15 +89,17 @@ func (s *DiffScanner) diffFromParent(tx db.Tx, params doDiffParams, scannerOpts 
 	}
 	leftLineage, err := getLineage(tx, params.LeftBranchID, CommittedID)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("get parent last commit failed: %w", err)
 	}
 	rightLineage, err := getLineage(tx, params.RightBranchID, UncommittedID)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("get chaild lineage failed on :%w", err)
 	}
 	// If some ancestor branch commit id is the same for parent and child - then the parent does not need to read it
 	// so it is trimmed from the parent lineage
-
+	if len(rightLineage)-len(leftLineage) != 1 {
+		return nil, fmt.Errorf("corrupted lineage definitions:%w", ErrLineageCorrupted)
+	}
 	for i := range leftLineage {
 		if leftLineage[i].CommitID == rightLineage[i+1].CommitID {
 			leftLineage = leftLineage[:i]
--- a/catalog/db_diff_scanner.go
+++ b/catalog/db_diff_scanner.go
@@ -98,7 +98,7 @@ func (s *DiffScanner) diffFromParent(tx db.Tx, params doDiffParams, scannerOpts 
 	// If some ancestor branch commit id is the same for parent and child - then the parent does not need to read it
 	// so it is trimmed from the parent lineage
 	if len(rightLineage)-len(leftLineage) != 1 {
-		return nil, fmt.Errorf("corrupted lineage definitions:%w", ErrLineageCorrupted)
+		return nil, ErrLineageCorrupted
 	}
 	for i := range leftLineage {
 		if leftLineage[i].CommitID == rightLineage[i+1].CommitID {
--- a/catalog/db_diff_scanner.go
+++ b/catalog/db_diff_scanner.go
@@ -55,19 +55,23 @@ func NewDiffScanner(tx db.Tx, params doDiffParams) (*DiffScanner, error) {
 		Relation: relation,
 		params:   params,
 	}
+	scannerOpts := DBLineageScannerOptions{DBScannerOptions: DBScannerOptions{
+		After:            params.After,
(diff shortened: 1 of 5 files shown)


catalog/db_scanner.go
```diff
@@ -7,6 +7,7 @@ type DBScannerOptions struct {
 	After            string
 	AdditionalFields []string
 	AdditionalWhere  sq.Sqlizer
+	TrimmedLineage   []lineageCommit
```
nopcoder, line 10: The DB scanner options are common options used by both Scanners - using lineage as part of options that we pass to branch scanner just break the 'common' idea.
Also the term TrimmedLineage - is the implementation detail of a specific use-case, we should keep the scanner option general as much as possible.

In this case, the DBLineageScanner need to accept LineageScannerOptions that has Lineage as an optional field - this will enable us to create scanner and override the default lineage.

catalog/db_diff_scanner.go
```diff
@@ -86,12 +85,31 @@ func (s *DiffScanner) diffFromParent(tx db.Tx, params doDiffParams) (*DiffScanne
...
 		AdditionalFields: prepareDiffAdditionalFields(params.AdditionalFields),
 	}
-	s.leftScanner = NewDBLineageScanner(tx, params.LeftBranchID, CommittedID, scannerOpts)
-	s.rightScanner = NewDBLineageScanner(tx, params.RightBranchID, UncommittedID, scannerOpts)
-	s.childLineage, err = getLineage(tx, params.RightBranchID, UncommittedID)
+	parentLineage, err := getLineage(tx, params.LeftBranchID, CommittedID)
 	if err != nil {
 		return nil, err
 	}
+	childLineage, err := getLineage(tx, params.RightBranchID, UncommittedID)
```
nopcoder, line 92: lets keep the names align - `leftLineage` and `rightLineage`

catalog/db_diff_scanner.go
```diff
@@ -86,12 +85,31 @@ func (s *DiffScanner) diffFromParent(tx db.Tx, params doDiffParams) (*DiffScanne
...
+	if err != nil {
+		return nil, err
+	}
+	// If some ancestor branch commit id is the same for parent and child - then the parent does not need to read it
+	// so it is trimmed from the parent lineage
+	if len(parentLineage) >= 1 {
```
nopcoder, line 98: 1. no need to check len >= 1 if it was 0 the loop was skipped
2. need to verify that len(right lineage) > len(left lineage)

catalog/db_diff_scanner.go
```diff
@@ -86,12 +85,31 @@ func (s *DiffScanner) diffFromParent(tx db.Tx, params doDiffParams) (*DiffScanne
...
+		for i := range parentLineage {
+			if parentLineage[i].CommitID == childLineage[i+1].CommitID {
+				parentLineage = parentLineage[:i]
+				break
+			}
+		}
+	}
+	parentOpts := scannerOpts
+	parentOpts.TrimmedLineage = parentLineage
+	childOpts := scannerOpts
```
nopcoder, line 108: we don't need scannerOpts anymore - just define each scanner options

catalog/db_diff_scanner.go
```diff
@@ -86,12 +85,31 @@ func (s *DiffScanner) diffFromParent(tx db.Tx, params doDiffParams) (*DiffScanne
...
+	childOpts.TrimmedLineage = childLineage
+	s.leftScanner = NewDBLineageScanner(tx, params.LeftBranchID, CommittedID, parentOpts)
+	s.rightScanner = NewDBLineageScanner(tx, params.RightBranchID, UncommittedID, childOpts)
+	s.childLineage = childLineage
```
nopcoder, line 112: We still need to get the lineage from the scanner when possible

nopcoder (changes requested):

catalog/db_diff_scanner.go
tzahij: done

catalog/db_diff_scanner.go
tzahij: done

catalog/db_diff_scanner.go
tzahij: not sure I understand. I have the lineage in a var. LineageScanner.ReadLineage will access the DB

catalog/db_scanner.go
tzahij: Done

catalog/db_diff_scanner.go
```diff
@@ -56,19 +55,23 @@ func NewDiffScanner(tx db.Tx, params doDiffParams) (*DiffScanner, error) {
 		Relation: relation,
 		params:   params,
 	}
+	scannerOpts := DBLineageScannerOptions{DBScannerOptions: DBScannerOptions{
+		After:            params.After,
+		AdditionalFields: prepareDiffAdditionalFields(params.AdditionalFields)},
+	}
```
nopcoder, line 61: Format the struct initialize - DBScannerOptions like any other field in the struct

```suggestion
	scannerOpts := DBLineageScannerOptions{
		DBScannerOptions: DBScannerOptions{
			After:            params.After,
			AdditionalFields: prepareDiffAdditionalFields(params.AdditionalFields),
		},
	}
```

catalog/db_diff_scanner.go
```diff
@@ -82,44 +85,48 @@ func (s *DiffScanner) diffFromParent(tx db.Tx, params doDiffParams) (*DiffScanne
 	if err != nil {
 		return nil, fmt.Errorf("get child last commit failed: %w", err)
 	}
-	scannerOpts := DBScannerOptions{
-		After:            params.After,
-		AdditionalFields: prepareDiffAdditionalFields(params.AdditionalFields),
+	leftLineage, err := getLineage(tx, params.LeftBranchID, CommittedID)
+	if err != nil {
+		return nil, err
```
nopcoder, line 90: use fmt.Errorf to describe the error as it was from getting lineage of the left branch

catalog/db_diff_scanner.go
```diff
@@ -82,44 +85,48 @@ func (s *DiffScanner) diffFromParent(tx db.Tx, params doDiffParams) (*DiffScanne
...
 	}
-	s.leftScanner = NewDBLineageScanner(tx, params.LeftBranchID, CommittedID, scannerOpts)
-	s.rightScanner = NewDBLineageScanner(tx, params.RightBranchID, UncommittedID, scannerOpts)
-	s.childLineage, err = getLineage(tx, params.RightBranchID, UncommittedID)
+	rightLineage, err := getLineage(tx, params.RightBranchID, UncommittedID)
```
nopcoder, line 92: use fmt.Errorf to describe the error as it was from getting lineage of the right branch

catalog/db_diff_scanner.go
```diff
@@ -82,44 +85,48 @@ func (s *DiffScanner) diffFromParent(tx db.Tx, params doDiffParams) (*DiffScanne
...
 	if err != nil {
 		return nil, err
 	}
+	// If some ancestor branch commit id is the same for parent and child - then the parent does not need to read it
+	// so it is trimmed from the parent lineage
+
+	for i := range leftLineage {
```
nopcoder, line 99: you removed the len check - don't want to verify that the child right side is bigger?

tzahij: done

catalog/db_diff_scanner.go
tzahij: done

catalog/db_diff_scanner.go
```diff
@@ -82,44 +87,50 @@ func (s *DiffScanner) diffFromParent(tx db.Tx, params doDiffParams) (*DiffScanne
 	if err != nil {
 		return nil, fmt.Errorf("get child last commit failed: %w", err)
 	}
-	scannerOpts := DBScannerOptions{
-		After:            params.After,
-		AdditionalFields: prepareDiffAdditionalFields(params.AdditionalFields),
+	leftLineage, err := getLineage(tx, params.LeftBranchID, CommittedID)
+	if err != nil {
+		return nil, fmt.Errorf("get parent last commit failed: %w", err)
```
nopcoder, line 92:
```suggestion
		return nil, fmt.Errorf("get left branch lineage: %w", err)
```

catalog/db_diff_scanner.go
```diff
@@ -82,44 +87,50 @@ func (s *DiffScanner) diffFromParent(tx db.Tx, params doDiffParams) (*DiffScanne
...
 	}
-	s.leftScanner = NewDBLineageScanner(tx, params.LeftBranchID, CommittedID, scannerOpts)
-	s.rightScanner = NewDBLineageScanner(tx, params.RightBranchID, UncommittedID, scannerOpts)
-	s.childLineage, err = getLineage(tx, params.RightBranchID, UncommittedID)
+	rightLineage, err := getLineage(tx, params.RightBranchID, UncommittedID)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("get chaild lineage failed on :%w", err)
```
nopcoder, line 96:
```suggestion
		return nil, fmt.Errorf("get right branch lineage: %w", err)
```

catalog/errors.go
```diff
@@ -29,4 +29,5 @@ var (
 	ErrInvalidValue                = errors.New("invalid value")
 	ErrNonDirectNotSupported       = errors.New("non direct diff not supported")
 	ErrSameBranchMergeNotSupported = errors.New("same branch merge not supported")
+	ErrLineageCorrupted            = errors.New("child and father lineage does not match")
```
nopcoder, line 32:
```suggestion
	ErrLineageCorrupted            = errors.New("lineage corrupted")
```

catalog/db_diff_scanner.go
```diff
@@ -82,44 +87,50 @@ func (s *DiffScanner) diffFromParent(tx db.Tx, params doDiffParams) (*DiffScanne
...
 	}
+	// If some ancestor branch commit id is the same for parent and child - then the parent does not need to read it
+	// so it is trimmed from the parent lineage
+	if len(rightLineage)-len(leftLineage) != 1 {
+		return nil, fmt.Errorf("corrupted lineage definitions:%w", ErrLineageCorrupted)
```
nopcoder, line 101:
```suggestion
		return nil, ErrLineageCorrupted
```

nopcoder (approved):

PR #907, opened 2020-11-09
Create extension pgcrypto in "public" schema

arielshaqed:
The issue is that extension pgcrypto might already exist in the database, and we still need to
access it.  So we cannot choose to locate the extension in a schema of our choice.  On the other
hand it must be in the *same* schema each time because each _test_ uses a different schema, and
it is not possible to redefine the same extension there.  So `public` is a good choice.


guy-har (approved):

arielshaqed:
Well, that failed testing quite miserably, so PTAL.

guy-har:
Looks good, thanks

PR #914, opened 2020-11-12
fail export in case export is currently in progress

guy-har:


arielshaqed (approved):
Cool!
We will have to figure out how to schedule the next export as an additional parade task in order to do continuous export.  That can come after this PR, this PR already makes things much better, so it is obviously approved.
Good safe catch!

nopcoder (approved):

PR #915, opened 2020-11-12
Document export and unhide from CLI

guy-har:


ddl/000009_export_current.down.sql
```diff
@@ -1,6 +1,6 @@
 BEGIN;
 
 DROP TABLE IF EXISTS catalog_branches_export_state;
-DROP TYPE catalog_branches_export_status;
+DROP TYPE If EXISTS catalog_branch_export_status;
```
arielshaqed, line 4:
```suggestion
DROP TYPE IF EXISTS catalog_branch_export_status;
```
SQL is case-insensitive-ish in places like this, so it works, but it's not pretty.

arielshaqed (approved):
Thanks!

Before you pull, please change the PR title also to state that it documents `export` and shows (unhides) it.

PR #917, opened 2020-11-12
webui get config called only on create repository

### LINKED ISSUE #916: /setup route triggers a call to config API which cannot succeed
When setting up a new lakeFS installation, going to `/setup` which is happens before logging in, triggers an API request to `/api/v1/config` - since there's no logged-in user, the response is 401 Unauthorized - which triggers the browser to ask for credentials.

nopcoder:
Fix #916


guy-har (approved):
👍

PR #920, opened 2020-11-13
fix export to s3 not working due to wrong paths

guy-har:


export/export_handler.go
```diff
@@ -221,18 +231,19 @@ func (h *Handler) done(body *string, signalledErrors int) error {
...
-	path, err := PathToPointer(fmt.Sprintf("%s/%s", finishData.StatusPath, fileName))
-	if err != nil {
-		return err
-	}
-	data := fmt.Sprintf("status: %s, signalled_errors: %d\n", status, signalledErrors)
-	reader := strings.NewReader(data)
-	err = h.adapter.Put(path, reader.Size(), reader, block.PutOpts{})
-	if err != nil {
-		return err
+	if len(finishData.StatusPath) > 0 {
```
nopcoder, line 235: can you extract this block to a method, something like:

```go
	err = h.updateStatusPath(finishData, status, signalledErrors)
	if err != nil {
		return err
	}
```

```go
func (h *Handler) updateStatusPath(finishData FinishData, status catalog.CatalogBranchExportStatus, signalledErrors int) error {
	if finishData.StatusPath == "" {
		return nil
	}
	fileName := fmt.Sprintf("%s-%s-%s", finishData.Repo, finishData.Branch, finishData.CommitRef)
	path, err := PathToPointer(fmt.Sprintf("%s/%s", finishData.StatusPath, fileName))
	if err != nil {
		return err
	}
	data := fmt.Sprintf("status: %s, signalled_errors: %d\n", status, signalledErrors)
	reader := strings.NewReader(data)
	return h.adapter.Put(path, reader.Size(), reader, block.PutOpts{})
}
```

export/export_handler.go
```diff
@@ -39,14 +39,20 @@ type TaskBody struct {
...
 func PathToPointer(path string) (block.ObjectPointer, error) {
 	u, err := url.Parse(path) // TODO(guys): add verify path on create task
 	if err != nil {
 		return block.ObjectPointer{}, err
 	}
 	return block.ObjectPointer{
-		StorageNamespace: fmt.Sprintf("%s://%s", u.Scheme, u.Host),
-		Identifier:       u.Path,
+		StorageNamespace: fmt.Sprintf("%s://%s/", u.Scheme, u.Host),
+		Identifier:       removeLeadingSlash(u.Path),
```
nopcoder, line 55: stings.TrimPrefix or stringsTrimLeft

nopcoder (approved):
Cool

PR #921, opened 2020-11-13
Diff performance improvement by trim lineage 

tzahij:
does two things with lineage scanning of of left(source) branches of diff. in parent-to-child diff
1) removes the end of the left lineage list if it is the same as the right lineage list.
2) remove items from source if they were commited already in the previous lineage list


nopcoder (commented):
expressed my review in PR #928

nopcoder (approved):

PR #927, opened 2020-11-15
Allow access key configuration from "lakefs superuser" command

### LINKED ISSUE #908: Add ability to specify Access Key ID and Secret Key in setup endpoint
We need this for integrating lakeFS with CI: an instance of lakeFS will be created before a build is run and be destroyed after that. Every time lakeFS is set up, it will create different set up Access Key ID / Secret Key so it is hard to inject those values to the test.

arielshaqed:
Makes writing some integrations much easier.  Fixes #908.


arielshaqed:
Manually tested.  It works, with the following caveat.  AWS signing s3v4 (which gets used by default) apparently enforces a particular format for the access key ID.  I was able to get the `aws` command-line tool to work by adding a superuser with an access key ID that was generated by AWS (on the real AWS system)  Any secret key works, even "foo".  So this is not a security issue, but it _is_ a potential inconvenience.

arielshaqed:
@keimoon hope this will work for you!

docs/assets/js/swagger.yml
```diff
@@ -2026,10 +2043,10 @@ paths:
         - export
         - branches
       operationId: run
-      summary: export branch
```
itaiad200, line 2029: merge issues?

auth/setup.go
```diff
@@ -229,18 +229,34 @@ func AddAdminUser(authService Service, user *model.User) (*model.Credential, err
...
-	if err != nil {
-		return nil, fmt.Errorf("create credentials for %s %w", user.Username, err)
+	var creds *model.Credential
+	if user.AccessKeyID == "" {
+		// Generate and return a key pair
+		creds, err = authService.CreateCredentials(user.Username)
+		if err != nil {
+			return nil, fmt.Errorf("create credentials for %s %w", user.Username, err)
+		}
+	} else {
```
itaiad200, line 239: Are we ok with zero checks for the access secret key? e.g. we roll with an empty string?

auth/setup.go
```diff
@@ -229,18 +229,34 @@ func AddAdminUser(authService Service, user *model.User) (*model.Credential, err
...
-	// Generate and return a key pair
-	creds, err := authService.CreateCredentials(user.Username)
-	if err != nil {
-		return nil, fmt.Errorf("create credentials for %s %w", user.Username, err)
+	var creds *model.Credential
+	if user.AccessKeyID == "" {
+		// Generate and return a key pair
+		creds, err = authService.CreateCredentials(user.Username)
+		if err != nil {
+			return nil, fmt.Errorf("create credentials for %s %w", user.Username, err)
```
itaiad200, line 237: Here and elsewhere, go format standard for errors is:

```suggestion
			return nil, fmt.Errorf("create credentials for %s: %w", user.Username, err)
```

auth/setup.go
```diff
@@ -229,18 +229,34 @@ func AddAdminUser(authService Service, user *model.User) (*model.Credential, err
...
+		}
+	} else {
+		creds, err = authService.AddCredentials(user.Username, user.AccessKeyID, user.SecretAccessKey)
+		if err != nil {
+			return nil, fmt.Errorf("add credentials for %s %w", user.Username, err)
```
itaiad200, line 242:
```suggestion
			return nil, fmt.Errorf("add credentials for %s: %w", user.Username, err)
```

auth/setup.go
```diff
@@ -229,18 +229,34 @@ func AddAdminUser(authService Service, user *model.User) (*model.Credential, err
...
-	adminUser := &model.User{
+	return CreateInitialAdminUserWithKeys(authService, metadataManger, username, nil, nil)
+}
+
+func CreateInitialAdminUserWithKeys(authService Service, metadataManger MetadataManager, username string, accessKeyID *string, secretAccessKey *string) (*model.Credential, error) {
+	adminUser := &model.SuperuserConfiguration{User: model.User{
 		CreatedAt: time.Now(),
 		Username:  username,
+	}}
+	if accessKeyID != nil && secretAccessKey != nil {
```
itaiad200, line 257: What happens if only one of them is nil?

.github/workflows/nessie.yaml
```diff
@@ -137,6 +140,24 @@ jobs:
         run: |
             cd nessie/ops
             docker-compose ps -q postgres && docker-compose exec -T postgres pg_dumpall --username=lakefs | gzip | aws s3 cp - s3://nessie-system-testing/${{ github.run_number }}/dump.gz
+      - name: Run lakeFS S3 to use with local API key
+        env:
+          TAG: ${{ steps.version.outputs.tag }}
+          # Setting Account_ID as a secret as a way to avoid specifying it here
+          REPO: ${{ secrets.AWS_ACCOUNT_ID }}.dkr.ecr.us-east-1.amazonaws.com
+          LAKEFS_STATS_ENABLED: "false"
+          LAKEFS_BLOCKSTORE_TYPE: s3
```
itaiad200, line 149: Suggestion: Test it on a different db (mem, local, transient, MINIO?), we'll add more coverage (s3 is already tested)

api/api_controller.go
```diff
@@ -265,7 +265,12 @@ func (c *Controller) SetupLakeFSHandler() setupop.SetupLakeFSHandler {
 		c.deps.Collector.CollectEvent("global", "init")
 
 		username := swag.StringValue(setupReq.User.Username)
-		cred, err := auth.CreateInitialAdminUser(c.deps.Auth, c.deps.MetadataManager, username)
+		var cred *model.Credential
+		if setupReq.User.Key == nil {
+			cred, err = auth.CreateInitialAdminUser(c.deps.Auth, c.deps.MetadataManager, username)
+		} else {
```
itaiad200, line 271: I hate swagger and its defaults for required/optional, but I remember there's a way you'll get empty strings here, right?
If so, need to verify here.

api/api_controller_test.go
```diff
@@ -1444,72 +1444,97 @@ func TestHandler_ContinuousExportHandlers(t *testing.T) {
...
-	defer srv.Close()
-
 	name := "admin"
-	user := models.Setup{
-		Username: &name,
-	}
-	req, err := json.Marshal(user)
-	if err != nil {
-		t.Fatal("JSON marshal request", err)
+	cases := []struct {
```
itaiad200, line 1448: Consider adding tests for edge cases: I'm mostly worried about empty strings here.

.github/workflows/nessie.yaml
```diff
@@ -78,6 +78,9 @@ jobs:
     name: Run latest lakeFS app
     needs: deploy-image
     runs-on: ubuntu-latest
+    env:
+      LAKEFS_AUTH_ENCRYPT_SECRET_KEY: some random secret string
+      LAKEFS_DATABASE_CONNECTION_STRING: postgres://lakefs:lakefs@postgres/postgres?sslmode=disable
```
itaiad200, line 83: Revery

itaiad200 (changes requested):

api/api_controller_test.go
arielshaqed: Done!

auth/setup.go
arielshaqed: 1. Done!
2. This continues the previous format.
3. So I didn't fix the rest of the file. :shrug:

docs/assets/js/swagger.yml
arielshaqed: Yup; reverted.  Thanks!

arielshaqed (commented):
Thanks!

PTAL...

docs/assets/js/swagger.yml
arielshaqed: Actually... this is just the broken half-unused copy of swagger.yml we use for JS.  @nopcoder can explain why intersecting the broken contradictory build chains of Makefile, golang and React require this.

arielshaqed (commented):
Thanks!

PTAL...

nopcoder:
> Thanks!
> 
> PTAL...

Can switch the swagger update of the docs with verification that they match as a lint error, will it be a better option?
The idea was not to forget to update our documentation which is under docs/ with the latest swagger - which is found in root.

itaiad200 (approved):
Noice!

arielshaqed:
Thanks!  Pulling...

arielshaqed:
[...]
@nopcoder :
> Can switch the swagger update of the docs with verification that they match as a lint error, will it be a better option?
> The idea was not to forget to update our documentation which is under docs/ with the latest swagger - which is found in root.

I guess that would work.  Note that the current version does not guarantee anything -- and, indeed, the two files often fall out of sync with one another.

PR #938, opened 2020-11-19
local adapter - respect namespace in objects full path

guy-har:


block/local/adapter.go
```diff
@@ -70,9 +70,24 @@ func NewAdapter(path string, opts ...func(a *Adapter)) (*Adapter, error) {
 	}
 	return adapter, nil
 }
+func resolveNamespace(obj block.ObjectPointer) (block.QualifiedKey, error) {
+	qualifiedKey, err := block.ResolveNamespace(obj.StorageNamespace, obj.Identifier)
+	if err != nil {
+		return qualifiedKey, err
+	}
+	if qualifiedKey.StorageType != block.StorageTypeLocal {
+		return qualifiedKey, block.ErrInvalidNamespace
```
arielshaqed, line 79: How can this happen?

block/local/adapter.go
```diff
@@ -70,9 +70,24 @@ func NewAdapter(path string, opts ...func(a *Adapter)) (*Adapter, error) {
...
 
-func (l *Adapter) getPath(identifier string) string {
-	return path.Join(l.path, identifier)
+func (l *Adapter) getPath(identifier block.ObjectPointer) (string, error) {
+	obj, err := resolveNamespace(identifier)
+	if err != nil {
+		return "", err
+	}
+	p := fmt.Sprintf("%s/%s", obj.StorageNamespace, obj.Key)
+	return path.Join(l.path, p), nil
```
arielshaqed, line 90: I think this should work:
```suggestion
	return path.Join(l.path, obj.StorageNamespace, obj.Key), nil
```

arielshaqed (changes requested):
Great, thanks!  All the code is nice, tested, makes sense, and obviously approved.

There may be backwards-compatiblity issues that we should discuss.  I suggest we put it out as a release note but not more - local storage is _not_ currently suitable for production use.  Semver says it's a new minor version (because we're pre-1.0.0).  @ozkatz WDYT?

block/local/adapter.go
guy-har: @arielshaqed 
This could happen if the repository is configured with a different namespace.
Will happen due to wrong configuration (that we currently aloud).
Or, for example, an old repository that was configured when we where using S3, and then we changed to local

arielshaqed (approved):
Approved once release note is ready to go.

PR #943, opened 2020-11-19
Feature/export repair

guy-har:


catalog/cataloger.go
```diff
@@ -168,24 +168,17 @@ type ExportConfigurator interface {
 	PutExportConfiguration(repository string, branch string, conf *ExportConfiguration) error
 }
 
+type stateCB func(oldRef string, state CatalogBranchExportStatus) (newState CatalogBranchExportStatus, newMessage *string, err error)
```
arielshaqed, line 171: 1.  If it's useful to us and appears in an exported interface, it could be useful for others.
   ```suggestion
   type ExportStateCallback func(oldRef string, state CatalogBranchExportStatus) (newState CatalogBranchExportStatus, newMessage *string, err error)
   ```
2. Please document the function, here or where it is used.

catalog/cataloger.go
```diff
@@ -168,24 +168,17 @@ type ExportConfigurator interface {
...
-	// the other is done.
-	ExportStateMarkStart(tx db.Tx, repo string, branch string, newRef string) (string, CatalogBranchExportStatus, error)
-	// ExportStateMarkEnd verifies that the current export is of ref and ends an export operation
-	// on branch of repo.
-	ExportStateMarkEnd(tx db.Tx, repo string, branch string, ref string, newState CatalogBranchExportStatus, newMessage *string) error
-	// ExportStateDelete deletes any export state for repo.  Mostly useful in tests: in a
-	// living system the export state is part of the state of the world.
-	ExportStateDelete(tx db.Tx, repo string, branch string) error
+	// sets the new values with the values returned from the callback
+	// in case newRef == "" will will not change ref
```
arielshaqed, line 177:
```suggestion
	// ExportStateSet sets the new values with the values returned from the callback.  It sets the ref when newRef != ""
```

catalog/cataloger.go
```diff
@@ -168,24 +168,17 @@ type ExportConfigurator interface {
...
+	ExportState(repo, branch, newRef string, cb stateCB) error
```
arielshaqed, line 178:
```suggestion
	ExportStateSet(repo, branch, newRef string, cb stateCB) error
```

Also it might be clearer to return `newRef` from the callback.

catalog/cataloger_export.go
```diff
@@ -139,107 +140,88 @@ func (c *cataloger) PutExportConfiguration(repository string, branch string, con
...
-	if err != nil && !missing {
-		err = fmt.Errorf("ExportStateuMarkStart: failed to get existing state: %w", err)
-		return oldRef, state, err
-	}
-	oldRef = res.CurrentRef
-	state = res.State
+			branchID)
+		missing := errors.Is(err, db.ErrNotFound)
+		if err != nil && !missing {
+			err = fmt.Errorf("ExportStateuMarkStart: failed to get existing state: %w", err)
```
arielshaqed, line 188: *tpyo:
```suggestion
			err = fmt.Errorf("ExportStateSet: failed to get existing state: %w", err)
```

ddl/000013_export_repair.down.sql
```diff
@@ -0,0 +1,3 @@
+update catalog_branches_export_state
+set state='exported-successfully'
+where state = 'export-repaired'
```
arielshaqed, line 3: 1. Please use caps for SQL keywords.
2. Please end with a semicolon (so it is safe to add additional commands) and a newline (so editors like it).

Also on the next SQL file.

docs/assets/js/swagger.yml
```diff
@@ -2024,6 +2024,10 @@ paths:
         name: branch
         required: true
         type: string
+      - in: query
+        description: will change export status from export-failed to repaired and start a new export
```
arielshaqed, line 2028:
```suggestion
        description: if set, change export status from export-failed to repaired and start a new export
```

docs/assets/js/swagger.yml
```diff
@@ -2024,6 +2024,10 @@ paths:
...
+        name: repair
+        type: boolean
```
arielshaqed, line 2030: This doesn't feel like a boolean value to me, but rather like a different command.  So it should either be an enum ("`action`"?) or an entirely separate endpoint.

export/export.go
```diff
@@ -63,3 +66,20 @@ func ExportBranchDone(cataloger catalog.Cataloger, status catalog.CatalogBranchE
 	})
 	return err
 }
+
+var ErrRepairNonFailed = errors.New("branch status is not export-failed, could not be repaired")
```
arielshaqed, line 70:
```suggestion
var ErrRepairWrongStatus = errors.New("incorrect status")
```
and then state the specifics on a wrapping error.  Otherwise we could have many more errors of very similar type -- and the client needs to handle them all in exactly the same way.

This might even return as a separate HTTP code.

catalog/cataloger_export.go
```diff
@@ -139,107 +140,88 @@ func (c *cataloger) PutExportConfiguration(repository string, branch string, con
...
 
-	var query string
-	if missing {
-		query = `
+		// run callback
+		newState, newMsg, err := cb(oldRef, state)
+		if err != nil {
+			return nil, err
+		}
+		// in case newRef was not set leave oldRef
```
arielshaqed, line 199: Yeah, so probably best to return `newRef` from the callback as well.  Otherwise it's a bit tricky to understand what a call does.  (This was also an issue with my previous version, of course!)

catalog/cataloger.go
```diff
@@ -168,24 +168,17 @@ type ExportConfigurator interface {
...
+
+	// GetExportState returns the current Export state params
+	GetExportState(repo string, branch string) (ExportState, error)
```
arielshaqed, line 181: I don't think this function is ever safe to call: there is always a race with `ExportState`(`Set`).  And indeed I see that it is only used in tests.  Note that if you provide just `ExportStateSet` then you only need to test that function.  And you can do that by calling it _twice_, and seeing that the second callback receives the value you put there the first time.
WDYT?

arielshaqed (changes requested):
Thanks!

catalog/cataloger.go
guy-har: @arielshaqed The reason I chose to add this function even that it was only used in tests, was because I assumed we will need to provide a way to get the current status.

catalog/cataloger_export.go
guy-har: 👍

ddl/000013_export_repair.down.sql
guy-har: Changed, Thanks

catalog/cataloger_export.go
```diff
@@ -139,107 +140,84 @@ func (c *cataloger) PutExportConfiguration(repository string, branch string, con
...
-	tag, err := tx.Exec(`DELETE FROM catalog_branches_export_state WHERE branch_id=$1`, branchID)
-	if err != nil {
-		return err
-	}
-	if tag.RowsAffected() != 1 {
-		return fmt.Errorf("[I] ExportStateDelete: could not delete single row %s: %w", tag, ErrEntryNotFound)
-	}
-	return nil
+		tag, err := tx.Exec(query,
+			branchID, newRef, newState, newMsg)
```
arielshaqed, line 213: nit:
```suggestion
		tag, err := tx.Exec(query, branchID, newRef, newState, newMsg)
```

catalog/cataloger_export_test.go
```diff
@@ -177,106 +174,39 @@ func TestExportState(t *testing.T) {
...
-	for _, tt := range cases {
-		pool, err := pgxpool.Connect(ctx, c.DbConnURI)
-		if err != nil {
-			t.Fatalf(err.Error())
+
+	if err := c.ExportStateSet(repo, defaultBranch, insertStart); err != nil {
+		t.Fatal(err)
+	}
+
+	var InProgressToSuccess ExportStateCallback
```
arielshaqed, line 186: 1. Why not define it as a regular function, outside of the test?
2. In any case, why is it capitalized?

cmd/lakectl/cmd/export.go
```diff
@@ -104,12 +101,26 @@ var exportExecuteCmd = &cobra.Command{
 	},
 }
 
+var exportRepairCmd = &cobra.Command{
+	Use:   "repair",
+	Short: "marks failed export as repaired",
```
arielshaqed, line 106:
```suggestion
	Short: "mark failed export as repaired",
```
We usually keep documentation of commands in the imperative mode, e.g. see above `Short`s.  Go follows the Google oddity of using "_descriptive verbs in the indicative mood_" (https://google.github.io/styleguide/cppguide.html#Function_Comments), no reason to punish our users...

ddl/000014_export_repair.down.sql
```diff
@@ -0,0 +1,3 @@
+UPDATE catalog_branches_export_state
+SET state='exported-successfully'
+WHERE state = 'export-repaired';
```
arielshaqed, line 3: nit: newline at end of file (also on next file).

export/export.go
```diff
@@ -34,32 +34,53 @@ func ExportBranchStart(paradeDB parade.Parade, cataloger catalog.Cataloger, repo
...
+		}
+		return oldRef, status, statusMsg, nil
 	})
 	return err
 }
+
+var ErrRepairWrongStatus = errors.New("incorrect status")
+
+// ExportBranchRepair will change state from Failed To Repair and start a new export
+// will return Error in Case current state is not Failed
```
arielshaqed, line 78:
```suggestion
// ExportBranchRepair changes state from Failed To Repair and starts a new export.
// It fails if the current state is not ExportStatusFailed.
```

arielshaqed (commented):
Cool, thanks!

Please go over the comments (all nit-picking, sorry) and then pull.  Also I expect a conflict with #949, so whoever of us goes last will have to fix.

arielshaqed (approved):
I had one job...

cmd/lakectl/cmd/export.go
guy-har: 👍

PR #947, opened 2020-11-22
SSTable interface for committed data

itaiad200:
Interface for the SSTable handling of the committed data.


rocks3/catalog.go
```diff
@@ -0,0 +1,206 @@
+package rocks3
```
itaiad200, line 1: Please ignore this file.

ozkatz: I will ignore it when you remove it 🤭

tree/sstable/writer.go
```diff
@@ -0,0 +1,85 @@
...
+
+	return nil
+}
+
+func (dw *DiskWriter) Close() (*WriteResult, error) {
+	if err := dw.w.Close(); err != nil {
+		return nil, fmt.Errorf("sstable file write: %w", err)
+	}
+
+	sstableID := SSTableID(fmt.Sprintf("%x", dw.sha256.Sum(nil)))
```
ozkatz, line 72: would also dependency inject the naming part - let's have a single place and a single way of turning a hash into a name. otherwise we'll quickly find ourselves with hex in one place, base64 in anoter and no easy way for an external observer to correlate both

tree/sstable/writer.go
```diff
@@ -0,0 +1,85 @@
...
+	if err := os.Rename(dw.tempPath, sstableLocalPath(dw.tablesDirPath, sstableID)); err != nil {
+		return nil, fmt.Errorf("temp file rename: %w", err)
+	}
+
+	// TODO: check if this sstable already exists
+	// TODO: write to s3
```
ozkatz, line 78: would also leave that to a tiered storage manager...

tree/tree.go
```diff
@@ -0,0 +1,131 @@
...
+	"encoding/base64"
+	"encoding/json"
+	"hash/fnv"
+	"io/ioutil"
+	"sort"
+
+	"github.com/treeverse/lakefs/tree/sstable"
+)
+
+const SplitFactor = 200_000 // avarage number of entries in a part file. used to compute modulo on hash of path
```
ozkatz, line 13: fine for now, should be a configuration

tree/tree.go
```diff
@@ -0,0 +1,131 @@
...
+
+type TreePartType struct {
+	PartName string `json:"part_name"`
```
ozkatz, line 16: why json?

tree/tree.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+func (trees TreesRepoType) NewScanner(tree TreeID, start string) (*treeScanner, error) {
+	t, err := trees.loadTreeIfNeeded(tree)
+	if err != nil {
+		return nil, err
+	}
+	sst := findSST(t, start)
+	if sst == nil {
+		return nil, ErrPathBiggerThanMaxPath
```
ozkatz, line 78: not sure when this error value is useful, its name is hard to understand

tree/tree.go
```diff
@@ -0,0 +1,131 @@
...
+	pos := sort.Search(n, func(i int) bool {
+		return tree.TreeSlice[i].MaxPath >= path
+	})
+	if pos < n {
+		return &tree.TreeSlice[pos].PartName
+	}
+	return nil
+}
+
+/*
```
ozkatz, line 98: remove plz

ozkatz (changes requested):

tree/tree.go
tzahij: Scanner started with path that is above the maximum path in the tree

tree/tree.go
tzahij: BTW - this file belongs in my implementation of the tree. should be removed

tzahij (commented):
remove the tree file.

tree/sstable/writer.go
```diff
@@ -0,0 +1,85 @@
...
+
+	dw.sha256.Sum(pathBytes)
+	dw.sha256.Sum(entryBytes)
+
+	return nil
+}
+
+func (dw *DiskWriter) Close() (*WriteResult, error) {
+	if err := dw.w.Close(); err != nil {
+		return nil, fmt.Errorf("sstable file write: %w", err)
```
tzahij, line 69: file close, not write

tzahij (commented):
Hi, I need those interfaces so I can proceed withe the tree part.
We should separate our files so our PRs are not mixed together

PR #948, opened 2020-11-22
make postgres connection string optional

johnnyaug:


db/migration.go
```diff
@@ -107,8 +118,12 @@ func getMigrate(params params.Database) (*migrate.Migrate, error) {
 	defer func() {
 		_ = src.Close()
 	}()
-
-	m, err := migrate.NewWithSourceInstance("httpfs", src, params.ConnectionString)
+	var m *migrate.Migrate
+	if params.ConnectionString == "" {
+		m, err = getMigrateUsingEnvVariables(src)
```
arielshaqed, line 123: (Trying to learn migrate...) This ends up calling `NewWithInstance`, the other branch calls `NewWithSourceInstance`.  Can't we do the same in both and have a smaller `getMigrateUsingEnvVariables`?

arielshaqed (approved):
Cool, thanks!

How was this tested?  (Can we add it to Nessie?  Do we want to??)

arielshaqed:
> How was this tested? (Can we add it to Nessie? Do we want to??)

[Not in this PR, of course...]

db/migration.go
```diff
@@ -107,7 +105,9 @@ func getMigrate(params params.Database) (*migrate.Migrate, error) {
 	defer func() {
 		_ = src.Close()
 	}()
-
+	if params.ConnectionString == "" {
```
nopcoder, line 108: can you do this one on a temp var so we will not modify the caller data

nopcoder (approved):
cool!

PR #952, opened 2020-11-24
[bugfix] return ErrNotFound correctly

arielshaqed:
Package `scany` depends on a different version of `pgx` than the rest of lakeFS.  So
`errors.Is(err, pgx.ErrNoRows)` fails.  Luckily it (sort-of) knows of this issue and
wraps this call inside it as `pgxscan.NotFound`.

Also make `ErrNotFound` wrap `pgx.ErrNoRows` rather than a new error.

### DIFF
--- a/db/errors.go
+++ b/db/errors.go
@@ -2,10 +2,13 @@ package db
 
 import (
 	"errors"
+	"fmt"
+
+	"github.com/jackc/pgx/v4"
 )
 
 var (
-	ErrNotFound      = errors.New("not found")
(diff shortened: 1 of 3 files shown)


arielshaqed:
I think this bug can be seen in various places, where the pgx error may become visible instead of a nice readable error.

arielshaqed:
Not quite right :-( @nopcoder please hold off from reviewing for now.

nopcoder (approved):

PR #959, opened 2020-11-24
Rocks catalog interface

guy-har:


catalog/rocks/catalog.go
```diff
@@ -0,0 +1,241 @@
+package rocks3
```
ozkatz, line 1: don't forget to change :)

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,241 @@
...
+	DiffTypeChanged
+	DiffTypeConflict
+)
+
+type Diff struct {
+	Path Path
+	Type DiffType
+}
+
+// StorageNamespace is the URI to the storage location
```
ozkatz, line 57: let's add a note about validation here (for example, that functions/methods receiving a StorageNamespace, RepositoryID etc as a type can assume it has passed validation and is a legal one.

ozkatz (changes requested):

nopcoder (approved):

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,241 @@
...
+
+// internal structures used by Catalog
+type RepositoryIterator interface {
+	First() (RepositoryID, Repository)
+	Next() (RepositoryID, Repository)
+	SeekGE(BranchID) (RepositoryID, Repository)
+	Close()
+}
+
+type EntryIterator interface {
```
tzahij, line 128: I strongly prefer the iterators(scanners) we have in the MVCC catalog. I do not think we need to change the interfaces we use because rocksDB use them. Most our iterators are logical. Let us have a talk on this.

tzahij (changes requested):

catalog/rocks/catalog.go
nopcoder: Paste your suggested interface and we can discuss it.
As for pros and cons - the previous scanner will no longer be available, so we are not adding another variant.
The above interface is similar to how we communicate with pebble so I assume it will just make the code more similar in concept.
If we prefer the scanner interface we also should consider defining a value that captures the key and the value and not returning 3 values.

tzahij:
```suggestion
type EntryIterator interface {
	Next() bool
	Value() (*Path, *Entry)
	Err() error
	SeekGE(Path) (bool)
}
```

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// Branch is a pointer to a commit.
+type Branch struct {
+	CommitID CommitID
+	// nolint: structcheck, unused
+	stagingToken StagingToken
+}
+
+// Diff represents a changed state for a given entry (added, removed, changed, conflict)
```
itaiad200, line 42:
```suggestion
// DiffType represents a changed state for a given entry (added, removed, changed, conflict)
```

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,244 @@
...
+	*Entry
+}
+
+// Interfaces
+type Catalog interface {
+	// entries
+	GetEntry(ctx context.Context, repositoryID RepositoryID, ref Ref, path Path) (*Entry, error)
+	SetEntry(ctx context.Context, repositoryID RepositoryID, branchID BranchID, path Path, entry Entry) error
+	DeleteEntry(ctx context.Context, repositoryID RepositoryID, branchID BranchID, path Path) error
+	ListEntries(ctx context.Context, repositoryID RepositoryID, ref Ref, prefix, from, delimiter string, amount int) ([]Listing, bool, error)
```
itaiad200, line 102: Needs documentation for the returned bool, what does false/true mean here?

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,244 @@
...
+
+	// refs
+	CreateBranch(ctx context.Context, repositoryID RepositoryID, branchID BranchID, ref Ref) (Branch, error)
+	GetBranch(ctx context.Context, repositoryID RepositoryID, branchID BranchID) (Branch, error)
+	Dereference(ctx context.Context, repositoryID RepositoryID, ref Ref) (CommitID, error)
+	Log(ctx context.Context, repositoryID RepositoryID, commitID CommitID, amount int) ([]Commit, bool, error)
+	ListBranches(ctx context.Context, repositoryID RepositoryID, from BranchID, amount int) ([]Branch, bool, error)
```
itaiad200, line 109: These 2 needs documentation too.

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,244 @@
...
+	DeleteBranch(ctx context.Context, repositoryID RepositoryID, branchID BranchID) error
+
+	// commits
+	Commit(ctx context.Context, repositoryID RepositoryID, branchID BranchID, commit Commit) (CommitID, error)
+	Reset(ctx context.Context, repositoryID RepositoryID, branchID BranchID) error
+	Revert(ctx context.Context, repositoryID RepositoryID, branchID BranchID, ref Ref) error
```
itaiad200, line 115: We often confuse Reset/Revert terms, it needs documentation to declare the desired behaviour.

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,244 @@
...
+
+	// diffs and merges
+	Merge(ctx context.Context, repositoryID RepositoryID, from Ref, to BranchID) (CommitID, error)
+	DiffUncommitted(ctx context.Context, repositoryID RepositoryID, branchID BranchID, from Path, amount int) ([]Diff, bool, error)
+	Diff(ctx context.Context, repositoryID RepositoryID, left, right Ref, from Path, amount int) ([]Diff, bool, error)
```
itaiad200, line 120: Same here with the bool.

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,244 @@
...
+}
+
+// internal structures used by Catalog
+type RepositoryIterator interface {
```
itaiad200, line 124: Let's document the expected behaviour for each one of the First, Next, SeekGE , pending the iterator discussion.

itaiad200, line 124: Same here.

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,244 @@
...
+	First() (RepositoryID, Repository)
+	Next() (RepositoryID, Repository)
+	SeekGE(BranchID) (RepositoryID, Repository)
+	Close()
+}
+
+type EntryIterator interface {
```
itaiad200, line 131: Same here

itaiad200: And below.

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,244 @@
...
+	// it tracks changes from left to right, returning an iterator of Diff entries
+	Diff(ctx context.Context, ns StorageNamespace, left, right, base TreeID, from Path) (DiffIterator, error)
+
+	// Merge receives two trees and a 3rd merge base tree used to resolve the change type
+	// it applies that changes from left to right, resulting in a new tree that
+	// is expected to be immediately addressable
+	Merge(ctx context.Context, ns StorageNamespace, left, right, base TreeID) (TreeID, error)
+
+	// Apply is the act of taking an existing tree (snapshot) and applying a set of changes to it.
+	// A change is either an entity to write/overwrite, or a tombstone to mark a deletion
```
itaiad200, line 224: Define what a tombstone will look like for entries.

catalog/rocks/catalog.go
guy-har: Thanks

catalog/rocks/catalog.go
nopcoder: code updated

nopcoder (approved):

tzahij (approved):

catalog/rocks/catalog.go
```diff
@@ -7,53 +7,17 @@ import (
...
-	CommitID CommitID
-	// nolint: structcheck, unused
-	stagingToken StagingToken
-}
-
-// Diff represents a changed state for a given entry (added, removed, changed, conflict)
+// DiffType represents a changed state for a given entry (added, removed, changed, conflict)
 type DiffType uint8
 
+//goland:noinspection GoUnusedConst
```
itaiad200, line 13: goland??

itaiad200 (approved):

catalog/rocks/catalog.go
```diff
@@ -73,10 +73,6 @@ type Entry struct {
 	ETag         string
 }
 
-func (e *Entry) IsTombstone() bool {
```
arielshaqed: Are we going to have different types `Entry` in mutable and immutable?  Because, while the immutables interface never needs a tombstone, AFAICT the mutables interface _must_ have one if we want to layer mutables on top of immutables.  (The only alternative I can see is to have mutables know about immutables and overlay for itself… which just ends up making a different internal "entry" type there.)

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,338 @@
...
+	DiffTypeChanged
+	DiffTypeConflict
+)
+
+// function/methods receiving the following basic types could assume they passed validation
+type (
+	// StorageNamespace is the URI to the storage location
+	StorageNamespace string
+
+	// RepositoryID is an identifier for a repo
```
arielshaqed, line 25:
```suggestion
	// RepositoryID is an identifier for a repo
```
I think we enforce allowed characters here.

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,338 @@
...
+	RepositoryID string
+
+	// Path represents a logical path for an entry
+	Path string
```
arielshaqed, line 29: Does it start with `/`?

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,338 @@
...
+
+	// Ref could be a commit ID, a branch name, a Tag
```
arielshaqed, line 31:
```suggestion
	// Ref locates a tree at a certain point in time.  It may be a commit ID, a branch name, or a tag
```

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,338 @@
...
+	Ref string
+
+	// TagID represents a named tag pointing at a commit
```
arielshaqed, line 34: Is it immutable?  (I'd expect half, just like in git)

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,338 @@
...
+	TagID string
+
+	// CommitID is a content addressable hash representing a Commit object
+	CommitID string
+
+	// BranchID is an identifier for a branch
+	BranchID string
```
arielshaqed, line 41: Legal chars?

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,338 @@
...
+
+	// TreeID represents a snapshot of the tree, referenced by a commit
```
arielshaqed, line 43:
```suggestion
	// TreeID represents a snapshot of the tree.  It can be referenced by a commit.
```

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,338 @@
...
+}
+
+type RepositoryRecord struct {
+	RepositoryID RepositoryID
+	*Repository
+}
+
+// Entry represents metadata or a given object (modified date, physical address, etc)
+type Entry struct {
+	LastModified time.Time
```
arielshaqed, line 70:
```suggestion
	CreationDate time.Time
```
Can we ever change an entry?

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,338 @@
...
+	CommonPrefix string
+
+	// Metadata key/value strings to holds metadata information on entry and commit
+	Metadata map[string]string
+)
+
+// Repository represents repository metadata
+type Repository struct {
+	StorageNamespace StorageNamespace
+	CreationDate     time.Time
```
arielshaqed, line 59: We should decide if we want to call it a "date".  I prefer `CreatedAt` or `CreationTime`; preferably the former because it does not repeat the type.

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,338 @@
...
+
+// Diff represents a change in path
+type Diff struct {
+	Path Path
+	Type DiffType
+}
+
+// Interfaces
+type Catalog interface {
+	// GetEntry returns entry from repository / reference by path, nil entry is a valid value for tombstone
```
arielshaqed, line 125: "Tombstone" still here.

catalog/rocks/catalog.go
```diff
@@ -0,0 +1,338 @@
...
+	// returns error if entry does not exist
```
arielshaqed, line 126: Let's specify which error (we had weirdness on this in the MVCC implementation).

catalog/rocks/catalog.go
arielshaqed: Relevant because we removed `IsTombstone` :-/

arielshaqed (commented):
Sorry, started these concurrently with pulling.

PR #962, opened 2020-11-25
Immutable tiered storage

itaiad200:
Initial draft of the tiered storage.


pyramid/file.go
```diff
@@ -0,0 +1,37 @@
+package pyramid
```
arielshaqed, line 1: I like the pyramid at Uxmal
:smile:

pyramid/pyramid.go
```diff
@@ -0,0 +1,22 @@
...
+const (
+	StorageTypeSSTable StorageType = iota
+	StorageTypeTreeManifest
+)
+
+// FS is pyramid abstraction of filesystem where the persistent storage-layer is the block storage.
+// Files on the Local disk are transient and might be cleaned up by the eviction policy.
+type FS interface {
+	// Store uploads the file to the block storage.
+	// It may remove the file from the local storage.
```
arielshaqed, line 16:
```suggestion
	// It takes ownership of file and might remove it from local storage.
```

pyramid/pyramid.go
```diff
@@ -0,0 +1,22 @@
...
+	Store(t StorageType, filename string) error
```
arielshaqed, line 17: Can you give a good way to get a useful filename to use here?  I _do_ want my client to use "the right" directory.  E.g. if I pick my own filename it may be on the wrong filesystem.

pyramid/pyramid.go
```diff
@@ -0,0 +1,22 @@
+package pyramid
+
+// StorageType is enumeration of the different tiered storage types
+// supported by pyramid.
+type StorageType int
+
+const (
+	StorageTypeSSTable StorageType = iota
+	StorageTypeTreeManifest
```
arielshaqed, line 9: I think comments here are needed.  E.g.: is this the tier?  Are high numbers slower?

Or is it the file type?  In that case, I'd want the _client_ to control the type just by holding one pyramid for each time.  (Consider the case of a RefsManager managing commits)

pyramid/tierFS.go
```diff
@@ -0,0 +1,143 @@
...
+type storageProps struct {
+	t                  StorageType
+	localDir           string
+	blockStoragePrefix string
+}
+
+// ImmutableTierFS is a filesystem where written files are never edited.
+// All files are stored in the block storage. Local paths are treated as a
+// cache layer that will be evicted according to the given eviction algorithm.
+type ImmutableTierFS struct {
```
arielshaqed, line 22: AFAICS this FS concept is _anyway_ only suitable for immutable files.

pyramid/tierFS.go
```diff
@@ -0,0 +1,143 @@
...
+	adaptor block.Adapter
+	config  map[StorageType]storageProps
+
+	// TODO: use refs anc last-access for the eviction algorithm
```
arielshaqed, line 26: We have a `cache` package in this project that uses golang-lru and could probably be useful here.  However caches tend to have particular semantics, consider picking one cache and coding around that one from the start (to avoid surprises).

pyramid/file.go
```diff
@@ -0,0 +1,37 @@
...
+
+import "os"
+
+type File struct {
+	fh      *os.File
```
arielshaqed, line 6: Can this be a ReadWriteSeeker or somesuch?  Not everything is a file.  (E.g. for testing a pure-memory fake file might be really useful).

pyramid/tierFS.go
```diff
@@ -0,0 +1,143 @@
...
+		return fmt.Errorf("open file: %w", err)
+	}
+	defer f.Close()
+
+	stat, err := f.Stat()
+	if err != nil {
+		return fmt.Errorf("file stat: %w", err)
+	}
+
+	return tfs.adaptor.Put(tfs.objPointer(t, ns, filename), stat.Size(), f, block.PutOpts{})
```
arielshaqed, line 73: Having to stat just to get the size of the file is really unfortunate -- e.g. it means we require the `Stat`.  (It can also be less safe than reading to the end.)

pyramid/tierFS.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+// Load returns the a file descriptor to the local file.
+// If the file is missing from the local disk, it will try to fetch it from the block storage.
+func (tfs *ImmutableTierFS) Load(t StorageType, ns, filename string) (*File, error) {
+	localPath := tfs.localpath(t, filename)
+	fh, err := os.Open(localPath)
+	if err != nil {
```
arielshaqed, line 81: This might be easier to read in an early-return style.  In this case it might be
```go
fh, err := os.Open(localPath)
if err == nil {
  return /* something with fh */, nil
}
if !os.IsNotExist(err) {
  return nil, fmt.Errorf("...%w", err)
}
fh, err = tfs.readFBS(t, ns, filename)
if err != nil {
// ...
```

pyramid/tierFS.go
```diff
@@ -0,0 +1,143 @@
...
+	defer reader.Close()
+
+	localPath := tfs.localpath(t, filename)
+	writer, err := os.Create(localPath)
+	if err != nil {
+		return nil, fmt.Errorf("creating file: %w", err)
+	}
+	defer writer.Close()
+
+	if _, err := io.Copy(writer, reader); err != nil {
```
arielshaqed, line 119: Why copy an immutable file?

arielshaqed (commented):
Nice!

Will use this to store commits.

pyramid/pyramid.go
itaiad200: Nope, you can't reference files from a different filesystem, that's guaranteed.

pyramid/file.go
itaiad200: Not really, `pebble` sstable reader requires the Entire File interface. see https://github.com/cockroachdb/pebble/blob/e6d62a79c8e3f3bec3cc1f49c93ad4a311985d40/sstable/reader.go#L2200

pyramid/tierFS.go
itaiad200: Let me know if this comment is still valid assuming we must use files.

pyramid/tierFS.go
itaiad200: I'm copying from the block adapter response to the local file, should I do it in some other way?

pyramid/tierFS.go
itaiad200: Why do I need cache in here? I need to keep the access of all opened files, it doesn't make sense to drop any record here.

arielshaqed (commented):
Thanks!

Probably need to unit-test everything here; in particular I suspect file creation might not do what we want (comments marked).

pyramid/file.go
```diff
@@ -0,0 +1,48 @@
...
+	f.access.touch(f.rPath)
+	return f.fh.Read(p)
+}
+
+func (f *File) ReadAt(p []byte, off int64) (n int, err error) {
+	f.access.touch(f.rPath)
+	return f.fh.ReadAt(p, off)
+}
+
+func (f *File) Write(p []byte) (n int, err error) {
```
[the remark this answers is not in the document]
itaiad200: pyramid isn't an immutable Read-Only FS. You can write files to it, there's an API which is similar to `os.Create` which returns a file handle you can write to. It will only become immutable when that file is closed. 
However you're right about being able to write to a file which is being read. I'll add a bool to return err in such cases.

pyramid/tierFS.go
```diff
@@ -0,0 +1,244 @@
...
+	fileRef := tfs.newLocalFileRef(namespace, filename)
+	if err := os.Rename(originalPath, fileRef.fullPath); err != nil {
+		return fmt.Errorf("rename file: %w", err)
+	}
+
+	tfs.eviction.store(fileRef.fsRelativePath, stat.Size())
+	return nil
+}
+
+func (tfs *ImmutableTierFS) Create(namespace, filename string) (*File, error) {
```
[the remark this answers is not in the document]
itaiad200: Files being created, i.e. opened for write but not closed, aren't evicted.

pyramid/tierFS.go
```diff
@@ -0,0 +1,244 @@
...
+
+	return &File{
+		fh:     fh,
+		rPath:  fileRef.fsRelativePath,
+		access: tfs.eviction,
+		close:  tfs.adapterStore(fileRef, fh),
+	}, nil
+}
+
+func (tfs *ImmutableTierFS) adapterStore(fileRef localFileRef, reader io.Reader) func(size int64) error {
```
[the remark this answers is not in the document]
itaiad200: Right, thx - this piece of code is deleted.

pyramid/tierFS.go
```diff
@@ -0,0 +1,343 @@
...
+
+func (tfs *TierFS) validateArgs(namespace, filename string) error {
+	if err := tfs.validateNamespace(namespace); err != nil {
+		return err
+	}
+
+	return tfs.validateFilename(filename)
+}
+
+func (tfs *TierFS) validateFilename(filename string) error {
```
arielshaqed, line 282: [discussed f2f] As discussed, we actually *must* have separators inside S3 filenames: for scaling it likes small "directories", so you should split files named as digests into (say) 2 levels, with digest 444455556666aaaaffffeeee00002222 going into path 4444/55556666aaaaffffeeee00002222 :vomiting_face:

pyramid/tierFS.go
```diff
@@ -0,0 +1,343 @@
...
+}
+
+func (tfs *TierFS) validateNamespace(ns string) error {
+	return tfs.containsSeparator(ns)
+}
+
+var errSeparatorInFS = errors.New("path contains separator")
+
+func (tfs *TierFS) containsSeparator(str string) error {
+	if strings.ContainsRune(str, os.PathSeparator) {
```
arielshaqed, line 293: When called from `validateFilename` this doesn't work for Windows :vomiting_face: (I think), because S3 pathnames use `/` as a separator and Windows uses `\` :shrug:.

arielshaqed (commented):
Thanks!
Sorry it's taking so long.  As discussed the largest change is probably to write files *outside* the scope of the same files as reading them.  Specifically we need to be able to close a to-be-written-to-S3 file _without_ copying it out to S3, e.g. to abort it when Pebble has a write error.  And _that_ can even happen when Pebble is closing the file, at which point it commits to close its received file.

pyramid/ro_file.go
```diff
@@ -0,0 +1,32 @@
...
+	return f.fh.Read(p)
+}
+
+func (f *ROFile) ReadAt(p []byte, off int64) (n int, err error) {
+	f.eviction.touch(f.rPath)
+	return f.fh.ReadAt(p, off)
+}
+
+func (f *ROFile) Stat() (os.FileInfo, error) {
+	return f.fh.Stat()
```
arielshaqed, line 27: I do want to touch here.  But... as long as we collect useful metrics from pyramid (_not necessarily in this PR!_), that's OK either way.

pyramid/tierFS.go
```diff
@@ -84,93 +88,71 @@ func NewFS(c *Config) (FS, error) {
...
-		}
-	}
-
 	return nil
 }
 
 func (tfs *TierFS) removeFromLocal(rPath relativePath) {
-	removeFromLocal(tfs.fsLocalBaseDir, rPath)
+	// This will be called by the cache eviction mechanism during entry insert.
+	// We don't want to wait while the file is being removed from the local disk.
```
arielshaqed, line 118: Neat!  But it _does_ mean we will need to be very careful how we handle errors there.

pyramid/tierFS.go
```diff
@@ -84,93 +88,71 @@ func NewFS(c *Config) (FS, error) {
...
-
-// Store adds the local file to the FS.
-// on successful operation, file will no longer be available under the originalPath.
-func (tfs *TierFS) Store(namespace, originalPath, filename string) error {
-	if err := tfs.validateArgs(namespace, filename); err != nil {
-		return fmt.Errorf("invalid args: %w", err)
+func removeFromLocal(logger logging.Logger, fsLocalBaseDir string, rPath relativePath) {
+	p := path.Join(fsLocalBaseDir, string(rPath))
+	if err := os.Remove(p); err != nil {
+		logger.WithError(err).Errorf("Removing file %s", p)
```
arielshaqed, line 125:
```suggestion
		logger.WithError(err).WithField("path", p).Errorf("failed to remove file")
```
Logs ideally have no formatting, everything goes in fields.  That allows me e.g. to estimate a histogram of eviction frequencies by file.

pyramid/tierFS.go
```diff
@@ -320,6 +304,10 @@ func (tfs *TierFS) newLocalFileRef(namespace, filename string) localFileRef {
 }
 
 func (tfs *TierFS) objPointer(namespace, filename string) block.ObjectPointer {
+	if runtime.GOOS == "windows" {
+		filename = strings.ReplaceAll(filename, `\\'`, "/")
```
arielshaqed, line 308: This change is fundamentally unsafe, e.g. if/when MacOS decides it has a new way of writing file paths.  I think that in all files that are not local, we should avoid `filepath.Join` in favour of URL-based joiners or our very own, and write our own `/`s.

pyramid/file_test.go
```diff
@@ -0,0 +1,134 @@
...
+	"path"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/google/uuid"
+)
+
+func TestPyramidWriteFile(t *testing.T) {
+	filename := uuid.Must(uuid.NewRandom()).String()
```
arielshaqed, line 15: Not a fan of UUIDs or of random data in tests, but if we're going to use them then this form is documented as _exactly_ the above:
```suggestion
	filename := uuid.New().String()
```

pyramid/file_test.go
```diff
@@ -0,0 +1,134 @@
...
+	filepath := path.Join("/tmp", filename)
```
arielshaqed, line 16: This is not a safe place.  Why not just [`ioutil.TempFile`](https://golang.org/pkg/io/ioutil/#TempFile)?

arielshaqed (changes requested):
This is really nice, and a lot simpler!

Some issues:
1. I don't think size-based LRU will make good decisions (I am not familiar with any literature about it either way, though).  I think that it always makes cache thrashing worse.  This is a criticism of our fork of the LRU cache, I just think it is particularly relevant here in our use-case: most user workloads will have a mix of very large and very small files (say, commits vs. trees), and frequent access to a small file should prefer to displace small files to large ones.  We _can_ partly overcome this with multiple pyramids, but then we need to configure each one separately.
2. Need to delete empty directories from the cache.  Otherwise a long-lived system can end up with high fan-out, which is even less efficient on POSIX than on S3.
3. Can we handle paths with zero-length subdirectories?  In S3 the paths `s3://foo/a////b` and `s3://foo/a///b` are distinct, whereas in POSIX `/a////b`, `/a///b` and `/a/b` all point to the same place.
   If we never use pyramid to cache general user files on disk, then we can just fail pathnames with an empty directory component.  Otherwise we will need metadata.

pyramid/tierFS.go
itaiad200: Not sure I follow - how can we prepare for a scenario when MacOS decides to use '^' as the separator? Filenames can contain the new separator.

itaiad200:
> This is really nice, and a lot simpler!
> 
> Some issues:
> 
> 1. I don't think size-based LRU will make good decisions (I am not familiar with any literature about it either way, though).  I think that it always makes cache thrashing worse.  This is a criticism of our fork of the LRU cache, I just think it is particularly relevant here in our use-case: most user workloads will have a mix of very large and very small files (say, commits vs. trees), and frequent access to a small file should prefer to displace small files to large ones.  We _can_ partly overcome this with multiple pyramids, but then we need to configure each one separately.
> 2. Need to delete empty directories from the cache.  Otherwise a long-lived system can end up with high fan-out, which is even less efficient on POSIX than on S3.
> 3. Can we handle paths with zero-length subdirectories?  In S3 the paths `s3://foo/a////b` and `s3://foo/a///b` are distinct, whereas in POSIX `/a////b`, `/a///b` and `/a/b` all point to the same place.
>    If we never use pyramid to cache general user files on disk, then we can just fail pathnames with an empty directory component.  Otherwise we will need metadata.

1. FS is already per usage (e.g. separate FS for trees and sstables). There are several reasons for this choice and you mentioned some. Other reasons are the need for separation of the disk usage, dynamic FS allocations and more. I also don't think LRU cache is optimised for variable cost entries - we'll have metrics to understand exactly by how much.
I'm making 2 assumptions here, that I'll think will hold for the first version:
- Disk capacity per FS is much bigger than a single file (>1000)
- Files are roughly the same size.
2. Done.
3. Good point. Currently there's no use-case for user based files and we control all paths. Added verification to file paths.

pyramid/file_test.go
itaiad200: Changed - need to some randomness so that subsequent runs after a fail test won't look at the same data (if failed to be deleted)

pyramid/ro_file.go
itaiad200: Added a touch - metrics in next PR.

pyramid/tierFS.go
itaiad200: Nothing much we can do, right?
For now logging it, will add metrics later on.

arielshaqed (approved):
Great, thanks!

2 small requests for follow on PRs:
1. Rename tierFS.go to tier_fs.go.
2. Monitoring. It's a cache, *and* we've already disagreed about lru.

pyramid/tierFS.go
```diff
@@ -0,0 +1,364 @@
...
+		return errSeparatorInFS
+	}
+	return nil
+}
+
+// relativePath is the path of the file under TierFS
+type relativePath string
+
+// localFileRef consists of all possible local file references
+type localFileRef struct {
```
tzahij, line 320: I suggest that the two last properties will be methods on localFileRef

pyramid/file.go
```diff
@@ -0,0 +1,65 @@
...
+	closed    bool
+	persisted bool
+	store     func(string) error
+}
+
+func (f *File) Read(p []byte) (n int, err error) {
+	return f.fh.Read(p)
+}
+
+func (f *File) ReadAt(p []byte, off int64) (n int, err error) {
```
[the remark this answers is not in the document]
itaiad200: I would rather keep the same os.File terminology :)

pyramid/file.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (f *File) Stat() (os.FileInfo, error) {
+	return f.fh.Stat()
+}
+
+func (f *File) Sync() error {
+	return f.fh.Sync()
+}
+
+func (f *File) Close() error {
```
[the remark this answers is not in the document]
itaiad200: I would love to unify them. But `Close` may be called by a different layer, e.g. while closing sstable writer.

pyramid/file_test.go
```diff
@@ -0,0 +1,134 @@
...
+	defer os.Remove(filepath)
```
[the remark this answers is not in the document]
itaiad200: outdated comment.

pyramid/ro_file.go
```diff
@@ -0,0 +1,32 @@
+package pyramid
+
+import (
+	"os"
+)
+
+// ROFile is pyramid wrapper for os.file that implements io.ReadCloser
+// with hooks for updating evictions.
+type ROFile struct {
```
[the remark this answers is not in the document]
itaiad200: Why? it's also readable.

pyramid/file.go
```diff
@@ -0,0 +1,65 @@
+package pyramid
+
+import (
+	"fmt"
+	"os"
+)
+
+// File is pyramid wrapper for os.file that triggers pyramid hooks for file actions.
+type File struct {
```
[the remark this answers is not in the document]
itaiad200: Temp file name, which is only relevant during the write itself, is held by the os.File handle.
The final filename is passed during `Store`.

pyramid/tierFS.go
itaiad200: Good point. But then I'll have to keep a pointer to tfs, which is a bit messy.
I don't feel strongly about it, willing to change if you want.

Issue #655, opened 2020-09-25
WebUI delete losing focus on refresh


nopcoder:
Delete refresh updates the UI and focus back at the top of the listing.
When you have to scroll to locate your item, the delete operation will trigger a refresh that will get you back to the top of the list.

hi-18-K:
I can solve this issue! Could you assign it to me?

arielshaqed:
> I can solve this issue! Could you assign it to me?

Cool, thanks!  Let us know if you need help.  And (of course) looking forward to your PRs.

iamRishabh07:
Hi, I would like to work on this issue,

YaelRiv:
Hey @iamRishabh07! Thanks for your interest and willingness to help. 
This is already assigned to @hi-18-K.

Let us know if you need help finding another suitable issue :)

hi-18-K:
sir/mam, I have tried a solution of saving the scroll position on click of delete button and trigrring a scrollTo() method after a time lag(ie after refresh is done after deleting an object) . I have done this on my js file hosted locally. I am trying to install lakefs via docker with the help of given tutorial on youtube(having installed docker and pulling official lakefs image).But it is showing this error:-
![Screenshot from 2020-10-03 13-06-50](https://user-images.githubusercontent.com/46860508/94986090-68a8b280-0579-11eb-948d-8cada73655bd.png)

hi-18-K:
I am first time using docker so please explain in a bit simple way what should I do..

johnnyaug:
Hey @hi-18-K, still working on this?

hi-18-K:
Yes sir, actually I had a project submission pending(on 11th oct), initially I tried using js scrollTo which I need to modify a bit, also this is my first open source contribution in full scale project that's why it took so much time... I will submit this work on or before 17th oct.

johnnyaug:
No worries :slightly_smiling_face:, take your time, join our Slack if you need some help

nopcoder:
@hi-18-K why updates or partial work related to this issue that you have?

hi-18-K:
I have created PR. Was much confused how and where to place code. I have changed webui -> src -> actions -> objects.js.
The code is running well on my webpage in localhost environment, please let me know if it is doing something wrong in LakeFS.
I was confused about where to place the snippet and how to test...which is why it took so much time.

PR #966, opened 2020-11-26
Feature/added scrollto after object is deleted (issue #655)

hi-18-K:
added scrollTo feature. When any object is deleted its scroll position is stored and after it is deleted scrollTo is triggered to restore previous scroll state. 
Uses react scrollTo from "react-scroll" library.
issue #655 

This feature update is working fine in my local environment.
If there is any error, while running , please let me know. 
Also if the code snippet is not placed optimally, please tell me about that too, I was very much confused about that.

---
This change is [Reviewable](https://reviewable.io/reviews/treeverse/lakefs/966)

### DIFF
--- a/webui/src/actions/objects.js
+++ b/webui/src/actions/objects.js
@@ -1,5 +1,7 @@
 import * as api from "./api";
 import {AsyncActionType} from "./request";
+import { Link,  Element, Events, scrollSpy, scroller, animateScroll as scroll } from "react-scroll";
+import * as Scroll from 'react-scroll';
 
 export const PAGINATION_AMOUNT = 300;
 export const OBJECTS_LIST_TREE = new AsyncActionType('OBJECTS_GET_TREE');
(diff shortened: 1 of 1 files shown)


arielshaqed:
@nopcoder could you please review?

@hi-18-K please note that JavaScript lint checks failed; I don't think the configured versions like that `constructor()` construct :-/

nopcoder (changes requested):
1. Manage the scroll position is handled by the component and not by our action state.  The component can compare the scroll position state and restore it when the action is done.
2. About the error you see - the code block is not a 'class' so you can't define 'constructor' as part of code execution.

hi-18-K:
okay, got why it happens. I will modify it.

hi-18-K:
sir, please look into the error once again. It is showing some installation requirements. How should we proceed for the same?

hi-18-K:
maybe you will have to update react too for using animation feature from "react-scroll" or we can change it according older version..but it willll not be looking good..

nopcoder:
Closing this one as the current development focus is on architectural changes of the core project.

Issue #955, opened 2020-11-24
UI - Add an arrow to indicate merge direction


itaiad200:
Current UI is confusing with the order of compare/merge between branches.

PR #968, opened 2020-11-27
Fix - merge direction #955

shamikakumar:
Implemented #955


itaiad200:
@shamikakumar Thanks for your effort!
Notice the direction of the arrow is opposite to the direction of the merge.
You can switch the arrow direction, however it makes more sense if you could simply switch the labels between the base branch and the target branch. e.g. merge to `master` should have `master` on the left, arrow pointing left, and the feature branch on the right.

itaiad200 (changes requested):
Should switch branches so that merge direction matches.

shamikakumar:
@itaiad200 oh, my bad. I see now how that's confusing, since we're used to the GitHub convention.

itaiad200:
Hey @shamikakumar 
You're absolutely right about keeping Github's convention! 

I still see the merge direction the other way around:
image

The scenario: created `feature-1` branch from `master`, added a single file to `feature-1` branch and committed. When I'm comparing I should get the above results when `master` is on the left (base branch) and `feature-1` branch is on the right. You can also use the description in the Merge button for the direction. 

Let me know if you have any further questions, I think we're almost there :)

shamikakumar:
@itaiad200 could you check now ? I think you may be viewing my previous commit

itaiad200:
LGTM
Thanks for your help @shamikakumar !

itaiad200 (approved):

PR #972, opened 2020-11-29
catalog interface | change Dereference to RevParse

guy-har:


catalog/rocks/catalog.go
```diff
@@ -9,6 +9,7 @@ import (
 
 // DiffType represents a changed state for a given entry (added, removed, changed, conflict)
 type DiffType uint8
+type ReferenceType uint8
```
nopcoder, line 12: Move to live 21 and add code doc

nopcoder (approved):

PR #974, opened 2020-11-29
check before building and pushing an existing image in benchmark flow

itaiad200:


.github/workflows/benchmark.yaml
```diff
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - master
+  pull_request:
```
itaiad200, line 6: TODO: remove once passed

nopcoder (approved):

PR #977, opened 2020-11-30
Upload multiple files in `lakectl fs upload`

arielshaqed:
Behaves much like `cp -t`: You can pass multiple files after `-s`.  That works (only) if the
destination lakeFS path ends with a slash.  That path is then treated as a directory.  Files
specified are uploaded "into" that directory, by their basenames.  An example should make it
clear:

```
$ lakectl fs upload lakefs://repo@master/a/b/ -s /tmp/baz -s /tmp/subdir/bar
Path: a/b/baz
Modified Time: 2020-11-30 14:31:03 +0200 IST
Size: 1686 bytes
Human Size: 1.7 kB
Checksum: 2637bdfc53a8a230d74287aef75cb257
Path: a/b/bar
Modified Time: 2020-11-30 14:31:03 +0200 IST
Size: 4 bytes
Human Size: 4 B
Checksum: 90580754da3ed1b3e4be38c9b277bc9b
```

The `/tmp/` prefix was stripped from the local filenames.  Each was uploaded to a file under
`lakefs://repo@master/a/b/`, keeping its basename (`baz` or `bar`).

This partially resolves [this query](https://lakefs.slack.com/archives/C016726JLJW/p1606735311131500) on our #help channel.

### DIFF
--- a/cmd/lakectl/cmd/fs.go
+++ b/cmd/lakectl/cmd/fs.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"io"
 	"os"
+	"path"
 	"strings"
 
 	"github.com/go-openapi/swag"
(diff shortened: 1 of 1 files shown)


arielshaqed:
We prefer #979.

PR #981, opened 2020-11-30
Feature/staging mgr postgres

johnnyaug:


catalog/rocks/staging/iterator.go
```diff
@@ -0,0 +1,68 @@
...
+	buffer      []*rocks.EntryRecord
+}
+
+func NewIterator(ctx context.Context, mgr *Manager, st rocks.StagingToken, nextFrom rocks.Path) *Iterator {
+	return &Iterator{nextFrom: nextFrom, mgr: mgr, ctx: ctx, st: st, dbHasNext: true}
+}
+
+func (s *Iterator) Next() bool {
+	s.idxInBuffer++
+	if s.idxInBuffer >= len(s.buffer) {
```
nopcoder, line 28: reverse the check so all the code will indent back left

catalog/rocks/staging/iterator.go
```diff
@@ -0,0 +1,68 @@
...
+	dbHasNext   bool
+	buffer      []*rocks.EntryRecord
+}
+
+func NewIterator(ctx context.Context, mgr *Manager, st rocks.StagingToken, nextFrom rocks.Path) *Iterator {
+	return &Iterator{nextFrom: nextFrom, mgr: mgr, ctx: ctx, st: st, dbHasNext: true}
+}
+
+func (s *Iterator) Next() bool {
+	s.idxInBuffer++
```
nopcoder, line 27: error check before change the state

catalog/rocks/staging/iterator.go
```diff
@@ -0,0 +1,68 @@
...
+	if s.idxInBuffer >= len(s.buffer) {
+		if !s.dbHasNext {
+			return false
+		}
+		res, err := s.mgr.listEntries(s.ctx, s.st, s.nextFrom, batchSize)
+		if err != nil {
+			s.err = err
+			return false
+		}
+		if len(res.entries) == 0 {
```
nopcoder, line 37: do we need to set has next to false there are no results so we will not request again list entries on empty result?

catalog/rocks/staging/iterator.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+		s.dbHasNext = res.hasNext
+		s.idxInBuffer = 0
+		s.buffer = res.entries
+		return true
+	}
+	return true
+}
+
+func (s *Iterator) SeekGE(path rocks.Path) bool {
```
nopcoder, line 51: do we support calling SeekGE after error?

catalog/rocks/staging/iterator.go
```diff
@@ -0,0 +1,68 @@
...
+	s.buffer = nil
+	s.idxInBuffer = 0
+	s.nextFrom = path
+	s.dbHasNext = true
+	return s.Next()
+}
+
+func (s *Iterator) Value() *rocks.EntryRecord {
+	return s.buffer[s.idxInBuffer]
```
nopcoder, line 60: len and error check

catalog/rocks/staging/manager.go
```diff
@@ -0,0 +1,111 @@
...
+		err := tx.Get(entry, "SELECT address, last_modified_date, size, checksum, metadata FROM staging_entries WHERE staging_token=$1 AND path=$2", st, path)
+		return entry, err
+	}, p.txOpts(ctx, db.ReadOnly())...)
+	if err != nil {
+		return nil, err
+	}
+	entry := res.(*rocks.Entry)
+	if entry.Address == "" {
+		return nil, nil
+	}
```
nopcoder, line 35: why is that?

catalog/rocks/staging/manager.go
```diff
@@ -0,0 +1,111 @@
...
+	if entry == nil {
+		entry = &rocks.Entry{}
+	}
+	_, err := p.db.Transact(func(tx db.Tx) (interface{}, error) {
+		return tx.Exec("INSERT INTO staging_entries"+
+			"(staging_token, path, address, last_modified_date, size, checksum, metadata) "+
+			"VALUES ($1, $2,$3, $4, $5, $6, $7)"+
+			"ON CONFLICT (staging_token, path)"+
+			"DO UPDATE SET (staging_token, path, addre"+
+			"ss, last_modified_date, size, checksum, metadata) ="+
```
nopcoder, line 49: very strange splits - use ` and format the query

ddl/000015_staging_mgr.down.sql
```diff
@@ -0,0 +1,3 @@
+BEGIN;
+DROP TABLE staging_entries;
```
nopcoder, line 2: 1. can you use the drop if exists syntax
2. do we need to drop the index?

ddl/000015_staging_mgr.up.sql
```diff
@@ -0,0 +1,25 @@
+BEGIN;
+CREATE TABLE IF NOT EXISTS staging_entries
```
nopcoder, line 2: need to apply the same "if not exists" to all the rest

catalog/rocks/staging/iterator.go
```diff
@@ -0,0 +1,68 @@
...
+	nextFrom    rocks.Path
+	idxInBuffer int
+	err         error
+	ctx         context.Context
+	st          rocks.StagingToken
+	dbHasNext   bool
+	buffer      []*rocks.EntryRecord
+}
+
+func NewIterator(ctx context.Context, mgr *Manager, st rocks.StagingToken, nextFrom rocks.Path) *Iterator {
```
nopcoder, line 22: just something I thought about - with this implementation we can simplify the creation and remove the 'nextFrom'.
it is just a state change we can perform by calling to SeekGE after NewIterator.

catalog/rocks/staging/manager.go
```diff
@@ -0,0 +1,111 @@
...
+	return NewIterator(ctx, p, st, from), nil
+}
+
+type listEntriesResult struct {
+	entries  []*rocks.EntryRecord
+	hasNext  bool
+	nextPath rocks.Path
+}
+
+func (p *Manager) listEntries(ctx context.Context, st rocks.StagingToken, from rocks.Path, limit int) (*listEntriesResult, error) {
```
nopcoder, line 73: I thought it will be the other way around - iterator will select the data, listing with pagination will use the iterator.
The layer above doesn't have access to list entries - it will re-implement this part using the iterator.

catalog/rocks/staging/manager.go
johnnyaug: That's a tombstone :(

catalog/rocks/staging/manager.go
johnnyaug: :laughing: a mistake

catalog/rocks/staging/iterator.go
johnnyaug: I don't see why not. I will clear the error in this case

catalog/rocks/staging.go
```diff
@@ -0,0 +1,78 @@
...
+		err := tx.Get(entry, "SELECT address, last_modified_date, size, e_tag, metadata FROM staging_entries WHERE staging_token=$1 AND path=$2", st, path)
+		return entry, err
+	}, p.txOpts(ctx, db.ReadOnly())...)
+	if err != nil {
+		return nil, err
+	}
+	entry := res.(*Entry)
+	if entry.Address == "" {
+		return nil, nil
+	}
```
nopcoder, line 34: so the part I was worried about here is the hidden assumption that I'm not using empty Address for non-tombstone

catalog/rocks/staging_test.go
```diff
@@ -0,0 +1,230 @@
...
+	"testing"
+	"time"
+
+	"github.com/treeverse/lakefs/db"
+	"github.com/treeverse/lakefs/testutil"
+)
+
+var entryMetadata = map[string]string{"metadata_field": "metadata_value"}
+
+func prepareMgr(t *testing.T) StagingManager {
```
nopcoder, line 16:
```suggestion
func testNewStatingManager(t *testing.T) StagingManager {
```

catalog/rocks/staging_test.go
```diff
@@ -0,0 +1,230 @@
...
+	t.Helper()
+	conn, _ := testutil.GetDB(t, databaseURI)
+	return NewStageManager(conn)
+}
+
+func TestSetGet(t *testing.T) {
```
nopcoder, line 22: see that you don't like table driven testing

catalog/rocks/staging_test.go
```diff
@@ -0,0 +1,230 @@
...
+	"errors"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/treeverse/lakefs/db"
+	"github.com/treeverse/lakefs/testutil"
+)
+
+var entryMetadata = map[string]string{"metadata_field": "metadata_value"}
```
nopcoder, line 14: can you move it to the relevant test block

catalog/rocks/staging_test.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		t.Fatalf("got unexpected error: %v", err)
+	}
+	_, err = s.GetEntry(context.Background(), "t1", "entry1")
+	if !errors.Is(err, db.ErrNotFound) {
+		t.Fatalf("error different than expected. expected=%v, got=%v", db.ErrNotFound, err)
+	}
+}
+
+func entry(addr string) *Entry {
```
nopcoder, line 222: newTestEntry

catalog/rocks/staging_test.go
```diff
@@ -0,0 +1,230 @@
...
+	return &Entry{
+		LastModified: time.Now(),
+		Address:      addr,
+		Metadata:     entryMetadata,
```
nopcoder, line 226: 1. can be nil
2. make a new one for each entry, so they will not share the same reference

catalog/rocks/staging_test.go
```diff
@@ -0,0 +1,230 @@
...
+		t.Fatalf("got unexpected error: %v", err)
+	}
+	if e.Address != "addr3" {
+		t.Fatalf("got unexpected entry address. expected=%s, got=%s", "addr3", e.Address)
+	}
+	err = s.DeleteEntry(context.Background(), "t1", "entry1")
+	if err != nil {
+		t.Fatalf("got unexpected error: %v", err)
+	}
+	_, err = s.GetEntry(context.Background(), "t1", "entry1")
```
nopcoder, line 216: add test for listing tomstone

catalog/rocks/staging_test.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+	if it.Err() != nil {
+		t.Fatalf("got unexpected error from list: %v", it.Err())
+	}
+	if len(res) != numOfEntries {
+		t.Errorf("got unexpected number of results. expected=%d, got=%d", numOfEntries, len(res))
+	}
+	for i, e := range res {
+		if e.Path != Path(fmt.Sprintf("entry%04d", i)) {
+			t.Fatalf("got unexpected entry from list at index %d: expected entry%04d, got: %s", i, i, e.Path)
```
nopcoder, line 143: set expectedPath = fmt.Sprintf("entry%04d", i) and use it twice

ddl/000015_staging_mgr.down.sql
```diff
@@ -0,0 +1,3 @@
+BEGIN;
```
nopcoder, line 1: do we need to drop the old catalog tablles?

nopcoder (approved):

ddl/000016_staging_mgr.up.sql
```diff
@@ -0,0 +1,22 @@
...
+CREATE TABLE IF NOT EXISTS kv_staging_p2 PARTITION OF kv_staging FOR VALUES WITH (MODULUS 10, REMAINDER 2);
+CREATE TABLE IF NOT EXISTS kv_staging_p3 PARTITION OF kv_staging FOR VALUES WITH (MODULUS 10, REMAINDER 3);
+CREATE TABLE IF NOT EXISTS kv_staging_p4 PARTITION OF kv_staging FOR VALUES WITH (MODULUS 10, REMAINDER 4);
+CREATE TABLE IF NOT EXISTS kv_staging_p5 PARTITION OF kv_staging FOR VALUES WITH (MODULUS 10, REMAINDER 5);
+CREATE TABLE IF NOT EXISTS kv_staging_p6 PARTITION OF kv_staging FOR VALUES WITH (MODULUS 10, REMAINDER 6);
+CREATE TABLE IF NOT EXISTS kv_staging_p7 PARTITION OF kv_staging FOR VALUES WITH (MODULUS 10, REMAINDER 7);
+CREATE TABLE IF NOT EXISTS kv_staging_p8 PARTITION OF kv_staging FOR VALUES WITH (MODULUS 10, REMAINDER 8);
+CREATE TABLE IF NOT EXISTS kv_staging_p9 PARTITION OF kv_staging FOR VALUES WITH (MODULUS 10, REMAINDER 9);
+
+CREATE UNIQUE index kv_staging_uidx
```
nopcoder, line 20: IF NOT EXISTS?

graveler/graveler.go
```diff
@@ -354,27 +354,23 @@ type CommittedManager interface {
...
-	//   Returns ErrNotFound if no value found on key
-	Get(ctx context.Context, repositoryID RepositoryID, branchID BranchID, st StagingToken, key Key) (*Value, error)
+	// Get returns the value for the provided staging token and key
+	// Returns ErrNotFound if no value found on key.
+	Get(ctx context.Context, st StagingToken, key Key) (*Value, error)
 
-	// Set writes an value (or nil value to represent a tombstone)
-	Set(ctx context.Context, repositoryID RepositoryID, branchID BranchID, key Key, value *Value) error
+	// Set writes a value under the given staging token and key.
+	Set(ctx context.Context, st StagingToken, key Key, value *Value) error
```
nopcoder, line 364:
```suggestion
	Set(ctx context.Context, st StagingToken, key Key, value Value) error
```

nopcoder (approved):
approved - minor comments + we will have to align the interface of the iterators after Oz's changes

PR #983, opened 2020-12-01
add installation id after init

### LINKED ISSUE #982: init event sent without installation id for init through API
When setting up lakeFS through the API (for example, using the UI on /setup) - the init event is sent without installation id.

johnnyaug:
Closes #982


ozkatz (approved):
thanks!

PR #985, opened 2020-12-01
WIP: pg refs manager for rocks impl

ozkatz:


catalog/rocks/catalog.go
```diff
@@ -104,17 +113,50 @@ type EntryRecord struct {
...
+	CreationDate time.Time        `db:"creation_date"`
+	Parents      CommitParents    `db:"parents"`
+	Metadata     catalog.Metadata `db:"metadata"`
+}
+
+func (c Commit) ID() CommitID {
+	// TODO(ozkatz): move the hashing logic and encoding to a module others can use
+	h := sha256.New()
+	h.Write([]byte(c.Committer))
+	h.Write([]byte{0})
```
arielshaqed, line 128: I prefer to use length-prefixed writes.  Otherwise we get another illegal rune in most fields.  Or... AFAIK `git` just hashes its file contents, you could decide on a single textual format, stream a commit into that format, and then pass it `h`.

catalog/rocks/catalog.go
```diff
@@ -104,17 +113,50 @@ type EntryRecord struct {
...
+	h.Write([]byte(c.Message))
+	h.Write([]byte{0})
+	h.Write([]byte(c.TreeID))
+	h.Write([]byte{0})
+	binary.Write(h, binary.BigEndian, c.CreationDate.Unix())
```
arielshaqed, line 133: Consider just writing ISO8601 format string: easier to track.

catalog/rocks/catalog.go
```diff
@@ -314,7 +356,9 @@ type RefManager interface {
 	// ListBranches lists branches
 	ListBranches(ctx context.Context, repositoryID RepositoryID, from BranchID) (BranchIterator, error)
 
-	// GetCommit returns the Commit metadata object for the given CommitID
+	// GetCommit returns the Commit metadata object for the given CommitID.
+	// The given CommitID may be arbitrarily truncated and should return a valid commit as long
+	// as results are not ambiguous
```
arielshaqed, line 361: Isn't that the job of a dereferencer, rather than of `GetCommit`?

catalog/rocks/pgiterator.go
```diff
@@ -0,0 +1,288 @@
...
+type PGRepositoryIterator struct {
+	db  db.Database
+	ctx context.Context
+
+	value *RepositoryRecord
+	buf   []*RepositoryRecord
+
+	offset           string
+	prefetchSize     int
+	prefetchRequired bool
```
arielshaqed, line 28: This field has a confusing name.  AFAICT it behaves as though it could be called `dontNeedToSelect`; if so, could you replace it with a field with a name more like `needToSelect`?

catalog/rocks/pgiterator.go
```diff
@@ -0,0 +1,288 @@
...
+		ri.prefetchRequired = false
+	}
+	return len(ri.buf) > 0
+}
+
+func (ri *PGRepositoryIterator) SeekGE(id RepositoryID) bool {
+	ri.offset = string(id)
+	ri.prefetchRequired = true
+	ri.buf = make([]*RepositoryRecord, 0)
+	ri.prefetch()
```
arielshaqed, line 90: `offset` is `id`, and `prefetch` says `id > $1`, so AFAICT this function is really `SeekGT`.  (Which admittedly makes more sense for most iteration scenarios, unless you want to call `SeekGE(after + "\0")`).

catalog/rocks/pgiterator.go
```diff
@@ -0,0 +1,288 @@
...
+
+	if len(ri.buf) == 0 {
+		return false
+	}
+
+	// stage a value and increment offset
+	ri.value = ri.buf[len(ri.buf)-1]
```
arielshaqed, line 97: I don't understand this function.  Why does it take the _last_ element here?

catalog/rocks/pgrefs.go
```diff
@@ -0,0 +1,222 @@
...
+	return repository.(*Repository), nil
+}
+
+func (m *PGRefManager) CreateRepository(ctx context.Context, repositoryID RepositoryID, repository Repository, branch Branch) error {
+	_, err := m.db.Transact(func(tx db.Tx) (interface{}, error) {
+		_, err := tx.Exec(
+			`INSERT INTO kv_repositories (id, storage_namespace, creation_date, default_branch) VALUES ($1, $2, $3, $4)`,
+			repositoryID, repository.StorageNamespace, repository.CreationDate, repository.DefaultBranchID)
+		if err != nil {
+			return nil, err
```
arielshaqed, line 75: This returns a DB error when a repo is created twice with the same name.  We had both ways of doing it in the cataloger, I think we should pick one and stick to it throughout.  Just announce that we've decided this way :-/

catalog/rocks/pgrefs.go
```diff
@@ -0,0 +1,222 @@
...
+	}, db.WithContext(ctx))
+	return err
+}
+
+func (m *PGRefManager) DeleteBranch(ctx context.Context, repositoryID RepositoryID, branchID BranchID) error {
+	_, err := m.db.Transact(func(tx db.Tx) (interface{}, error) {
+		_, err := tx.Exec(
+			`DELETE FROM kv_branches WHERE repository_id = $1 AND id = $2`,
+			repositoryID, branchID)
+		return nil, err
```
arielshaqed, line 152: Should it return an error if branch does not exist?

catalog/rocks/pgrefs.go
```diff
@@ -0,0 +1,222 @@
...
+		if errors.Is(err, db.ErrNotFound) {
+			return nil, ErrNotFound
+		}
+		if err != nil {
+			return nil, err
+		}
+		startWith := make([]*Commit, 0)
+		for _, c := range records {
+			if strings.HasPrefix(string(c.CommitID), string(commitID)) {
+				startWith = append(startWith, c.Commit)
```
arielshaqed, line 179: Nice trick (with the `LIMIT 2`), but subtle.  Add a comment?

catalog/rocks/pgrefs.go
```diff
@@ -0,0 +1,222 @@
...
+
+func (m *PGRefManager) AddCommit(ctx context.Context, repositoryID RepositoryID, commit Commit) (CommitID, error) {
+	_, err := m.db.Transact(func(tx db.Tx) (interface{}, error) {
+		// commits are written based on their content hash, if we insert the same ID again,
+		// it will necessarily have the same attributes as the existing one, so no need to overwrite it
+		_, err := tx.Exec(`
+				INSERT INTO kv_commits 
+				(repository_id, id, committer, message, creation_date, parents, tree_id, metadata)
+				VALUES ($1, $2, $3, $4, $5, $6, $7, $8)
+				ON CONFLICT DO NOTHING`,
```
arielshaqed, line 204: It would be easier to debug weirdness in calls to this layer if we _did_ fail the insert if any values change.

catalog/rocks/pgrefs.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+func (m *PGRefManager) FindMergeBase(ctx context.Context, repositoryID RepositoryID, commitIDs ...CommitID) (*Commit, error) {
+	// TODO(ozkatz): This actually has some logic to it!
+	panic("implement me")
+}
+
+func (m *PGRefManager) Log(ctx context.Context, repositoryID RepositoryID, from CommitID) (CommitIterator, error) {
+	return NewCommitIterator(ctx, m.db, repositoryID, from), nil
+}
```
arielshaqed, line 222: Reviewed only up to here :-/, will keep going later.

arielshaqed (commented):
Partial review, I think there are a few implicit decisions that we need to make explicit.

catalog/rocks/catalog.go
ozkatz: yea, this requires a redo. we need a canonical way of doing stuff like that that works regardless of entity type (same as we had before MVCC with an `Identifiable` interface that defines an `Identity() []byte` method

catalog/rocks/catalog.go
ozkatz: started with that, got bitten by Postgres that converted the UTC date I inserted into my local timezone when I selected the record, resulting in a different hash than the one I used when storing the object. epoch is always UTC and it seems like it's what Git is using (according to `git cat-file -p ...` anyway)

catalog/rocks/catalog.go
ozkatz: good question. I don't know. If you feel strongly about this, I'm ok with changing that.

catalog/rocks/pgiterator.go
ozkatz: not sure I understand. it indicates wheteher a call to prefetch is required to advance the cursor. to me that makes sense (but again, not opinionated on that, so I'm ok with whatever concrete change you suggest)

catalog/rocks/pgiterator.go
ozkatz: yep, that's a bug. I'll add a test. prefetch should do `id >= $1`

catalog/rocks/pgiterator.go
ozkatz: bug. will add test.

catalog/rocks/pgrefs.go
ozkatz: what would you like to see returned?

catalog/rocks/pgrefs.go
ozkatz: this makes deletes idempotent and allows for blind writes. but I'm fine either way.

catalog/rocks/pgrefs.go
ozkatz: open to code suggestions.

catalog/rocks/revparse.go
```diff
@@ -0,0 +1,105 @@
...
+		return p, nil
+	}
+
+	// was in a modifier
+	mod, err := parseMod(buf)
+	if err != nil {
+		return p, err
+	}
+	p.Modifiers = append(p.Modifiers, mod)
+	return p, nil
```
johnnyaug, line 104:
```suggestion
	regex := regexp.MustCompile("(^|[~^])[^^~]*")
	parts := regex.FindAllString(r, -1)
	p.BaseRev = parts[0]
	if p.BaseRev == "" {
		return p, ErrInvalidRef
	}
	for _, part := range parts[1:] {
		mod, err := parseMod(part)
		if err != nil {
			return p, err
		}
		p.Modifiers = append(p.Modifiers, mod)
	}
	return p, nil
```

I hope you like regular expressions. Less efficient, but more readable IMO

catalog/rocks/revparse.go
```diff
@@ -0,0 +1,76 @@
+package rocks
+
+import (
+	"regexp"
+	"strconv"
+)
+
+var (
+	hashRegexp      = regexp.MustCompile("^[a-fA-F0-9]{1,64}$")
+	modifiersRegexp = regexp.MustCompile("(^|[~^])[^^~]*")
```
arielshaqed, line 10: Not sure I understand.  I think you don't need capture here, so
```suggestion
	modifiersRegexp = regexp.MustCompile("(?:"^|[~^])[^^~]*")
```
would make it explicit?

catalog/rocks/revparse.go
```diff
@@ -0,0 +1,76 @@
...
+	Value int
+}
+
+func parseMod(buf string) (RevModifier, error) {
+	amount := 1
+	var err error
+	if len(buf) > 1 {
+		amount, err = strconv.Atoi(buf[1:])
+		if err != nil {
+			return RevModifier{}, ErrInvalidRef
```
arielshaqed, line 40: Consider including something from `err`.

catalog/rocks/revparse.go
```diff
@@ -0,0 +1,76 @@
...
+func RevParse(r Ref) (ParsedRev, error) {
+	p := ParsedRev{
+		Modifiers: make([]RevModifier, 0),
+	}
+	parts := modifiersRegexp.FindAllString(string(r), -1)
+	p.BaseRev = parts[0]
+	if p.BaseRev == "" {
+		return p, ErrInvalidRef
+	}
+	for _, part := range parts[1:] {
```
arielshaqed, line 68: This is really hard to understand.  You use the fact that all the matches must occur consecutively because of the structure of the regexp.  So now I don't understand why you allow the string to start without `^` or `~`.

I think this would really be more readable as a loop without regexps:
```go
for len(r) > 0 {
  typ := ParseType(r[0])
  if typ == "" { error... }
  r = r[0:]
  value, r := parseZeroOrMoreDigits(r)
  if value == 0 { value = 1 }
  p.Modifiers = append(p.Modifiers, RevModifier{type, value})
}
```
where `parseZeroOrMoreDigits` eats as many digits as it can off of `r` and returns the digits as an integer and the rest of r.

ident/ident.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func ContentAddress(entity Identifiable) string {
+	h := sha256.New()
+	_, _ = h.Write(entity.Identity())
+	return hex.EncodeToString(h.Sum(nil))
+}
+
+type Buffer struct {
+	buf []byte
```
arielshaqed, line 21: Not sure why we ever want an actual `Buffer` for this, it could be quite long and fragment our memory.  If I'm only going to call `ContentAddress` on it, why not stream the data into the hash in the first place?

ident/ident.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+func NewBuffer() *Buffer {
+	return &Buffer{
+		buf: make([]byte, 0),
+	}
+}
+
+func (b *Buffer) MarshalBytes(v []byte) {
```
arielshaqed, line 30: Here and in every function below, you should also stream a type byte into the hash/buffer.  Otherwise different object formats can end up with the same hash, which is a bit surprising.

arielshaqed (changes requested):
Thanks!

Very nice code, but a lot of it, so many comments.

catalog/rocks/pgrefs.go
ozkatz: I feel that making this action idempontent would be better. We should (and do) test elsewhere that a given commit hashes to the same value consistently, and that different commits hash to different values. We'd need to add a lot of code here to
a. figure out that an insert on that PK has already taken place
b. reimplementing the logic of the commit's Identity function (i.e. to compare the stored commit is the same as the one we're attempting to insert).

I don't think the value of easier debugging justifies the added code and complexity

catalog/rocks/pgrefs_test.go
```diff
@@ -0,0 +1,370 @@
...
+
+	t.Run("listing all repos", func(t *testing.T) {
+		iter, err := r.ListRepositories(context.Background(), "")
+		if err != nil {
+			t.Fatalf("unexpected error: %v", err)
+		}
+
+		repoIds := make([]rocks.RepositoryID, 0)
+		for iter.Next() {
+			repo := iter.Value()
```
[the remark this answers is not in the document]
ozkatz: yep, added.

catalog/rocks/pgrefs_test.go
```diff
@@ -0,0 +1,370 @@
...
+		Parents:      rocks.CommitParents{"deadbeef1", "deadbeef12"},
+		Metadata:     catalog.Metadata{"foo": "bar"},
+	}
+
+	cid, err := r.AddCommit(context.Background(), "repo1", c)
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	if cid != "cd3ed2b2ccaf6dae8a2f2e514899cd78b5f917b6687cc4425624dfd6cd1858b1" {
```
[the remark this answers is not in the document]
ozkatz: I think that protection is important.

There are places where I don't care about the values but rather the parenting/relationships between them (for example, `TestPGRefManager_DereferenceWithGraph` or `TestFindLowestCommonAncestor`) where I assign them a logical name instead.

catalog/rocks/pgrefs.go
```diff
@@ -0,0 +1,226 @@
...
+			return nil, err
+		}
+		_, err = tx.Exec(`DELETE FROM kv_repositories WHERE id = $1`, repositoryID)
+		return nil, err
+	}, db.WithContext(ctx))
+	return err
+}
+
+func (m *PGRefManager) RevParse(ctx context.Context, repositoryID RepositoryID, ref Ref) (Reference, error) {
+	return ResolveRef(ctx, m, repositoryID, ref)
```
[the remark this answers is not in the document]
ozkatz: I'm bad at naming. Will happily accept better names

catalog/rocks/refresolver.go
```diff
@@ -0,0 +1,132 @@
...
+		// check if it's a branch
+		branch, err := store.GetBranch(ctx, repositoryID, BranchID(parsed.BaseRev))
+		if err != nil && !errors.Is(err, ErrNotFound) {
+			return nil, err
+		}
+		if err == nil {
+			baseCommit = branch.CommitID
+		}
+
+		if err == nil && len(parsed.Modifiers) == 0 {
```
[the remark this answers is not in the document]
ozkatz: Ya, it hides the private field - staging token.
Only a branch, and only at its tip will have that - this is why we return a branch object (that contains it, albeit privately) instead of resolving it to the underlying commit.

catalog/rocks/refresolver.go
```diff
@@ -0,0 +1,132 @@
...
+			// skip mod.ValueNumeric iterations
+			iter, err := store.Log(ctx, repositoryID, baseCommit)
+			if err != nil {
+				return nil, err
+			}
+			i := 0
+			found := false
+			for iter.Next() {
+				i++ // adding 1 because we start at base commit
+				if i == mod.Value+1 {
```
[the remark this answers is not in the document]
ozkatz: Not sure that's more readable?

catalog/rocks/revparse.go
ozkatz: not sure what you meant to do here.. but I count 3 unescaped `"`'s, so I don't think it'll compile :)

catalog/rocks/revparse.go
ozkatz: good idea.

catalog/rocks/revparse.go
ozkatz: that's funny because I originally had pretty much that. I accepted @johnnyaug 's suggestion that changed it into a regexp to make it easier to understand :) Given how subjective "understandability" seems to be, in regards to regexp in particular, I'm reluctant to make another change unless we have very good reasoning to do so.

ident/ident.go
ozkatz: I guess you're right. changed.

ident/ident.go
ozkatz: good idea, changed.

graveler/refmergebase.go
```diff
@@ -0,0 +1,101 @@
...
+			return commit, err
+		}
+		if iterLeft.Value() == nil && iterRight.Value() == nil {
+			break
+		}
+	}
+	return nil, nil
+}
+
+func findLowerCommonAncestorNextIter(discoveredSet map[string]struct{}, iter *CommitWalker) (*Commit, error) {
```
arielshaqed, line 92:
```suggestion
func findLowestCommonAncestorNextIter(discoveredSet map[string]struct{}, iter *CommitWalker) (*Commit, error) {
```
?

arielshaqed (approved):
Thanks!

ddl/000016_rocks_cataloger.down.sql
```diff
@@ -0,0 +1 @@
+-- 000016 doesn't support down migrations.
```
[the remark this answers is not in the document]
ozkatz: at this point I've decided to drop them. When we introduce a data migration that reads MVCC and writes the graveler data, that one should fail on DOWN

graveler/pgiterator_test.go
```diff
@@ -36,7 +37,7 @@ func TestPGRepositoryIterator(t *testing.T) {
 		}
 		iter.Close()
 
-		if !reflect.DeepEqual(repoIds, []graveler.RepositoryID{"a", "aa", "b", "c", "d", "e"}) {
+		if diffs := deep.Equal(repoIds, []graveler.RepositoryID{"a", "aa", "b", "c", "d", "e"}); diffs != nil {
 			t.Fatalf("got wrong list of repo IDs")
```
arielshaqed, line 41:
```suggestion
			t.Fatalf("got wrong list of repo IDs", diffs)
```
`deep.Equal` doesn't help if you don't log it :shrug: 

Also below.

arielshaqed (approved):
Thanks!

StillLGTM, but please logs diffs from `deep.Equal`.

graveler/pgiterator_test.go
ozkatz: added, thanks!

Issue #987, opened 2020-12-02
"Content-Type" response header should be "application/xml"


keimoon:
The `Content-Type` header for S3 compatible API is currently set to `text/xml charset=utf-8`. While this is fine with awscli, some strict clients like https://github.com/akka/alpakka only accepts `application/xml`. The real S3 API also returns `application/xml` so I think we should stick to that.

arielshaqed:
Thanks!
This looks like a bug in the alpakka client.  Specifically, [RFC 7303](https://tools.ietf.org/html/rfc7303#page-21) registers both application/xml and text/xml and states that
>  The registration information for text/xml is in all respects the same
>  as that given for application/xml above (Section 9.1), except that
>  the "Type name" is "text".
Clearly this argument cuts both ways.  The incumbent `text/xml` has the advantage that it works with all our current clients; the proposed `application/xml` has the advantage that it works with another client.  I shall see what I can do.

arielshaqed:
Here's what I can do: perform content negotiation on the [`Accept:`` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept).  So _if_ the client explicitly asks for either of `text/xml` or `application/xml` lakefs will return that type.
In order for this to work, your S3 client will need to send an appropriate header.  Is there a compiled client (hopefully CLI) that I could use on my system to capture headers and see whether that would work?
The effect on boto should be nil, it does not appear to send an `Accept` header.

ozkatz:
Seems like S3 returns `Content-Type: application/xml` even if passing `Accept: text/xml` in the request.

ozkatz:
@arielshaqed I think this is fixed by tge linked PR, no?

arielshaqed:
> @arielshaqed I think this is fixed by tge linked PR, no?

No customer {dis,}approval, so I guess you're right. Closing... and can always reopen if not good.

PR #992, opened 2020-12-03
Generate {application,text}/xml headers on gateway by client's choice

arielshaqed:
If no header is specified, none is explicitly generated.  Then the Go http server generates
one by scanning content, it happens to be text/xml.  If either of these headers is
specified, generate that requested content type.  If (only) other headers are generated,
respond "Not Acceptable".

This does not break backwards compatibility with clients that do not generate such a
header (boto, mc).  This should improve compatibility with a client that explicitly asks for
application/xml (the Akka S3 client alpakka only accepts that Content-Type, so it had better
ask for it).  It *will* break a lying client that requests only anything-but-standard-XML.

### DIFF
--- a/gateway/handler.go
+++ b/gateway/handler.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"net/http"
 	"reflect"
+	"regexp"
 	"strconv"
 	"strings"
 	"time"
@@ -347,6 +348,32 @@ func notFound(w http.ResponseWriter, _ *http.Request) {
 	w.WriteHeader(http.StatusNotFound)
 }
 
+var commaSeparator = regexp.MustCompile(`,\s*`)
+
+func selectContentType(acceptable []string) *string {
+	for _, acceptableTypes := range acceptable {
+		acceptable := commaSeparator.Split(acceptableTypes, -1)
+		for _, a := range acceptable {
+			if a == "text/xml" || a == "application/xml" {
+				return &a
+			}
+		}
+	}
+	return nil
+}
+
+func setContentType(w http.ResponseWriter, r *http.Request) {
+	acceptable, ok := r.Header["Accept"]
+	if ok {
+		contentType := selectContentType(acceptable)
+		if contentType != nil {
+			w.Header().Set("Content-Type", *contentType)
+		} else {
+			w.WriteHeader(http.StatusNotAcceptable)
+		}
+	}
+}
+
 func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// pprof endpoints
 	handler := h.servePathBased(r)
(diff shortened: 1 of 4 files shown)


arielshaqed:
Tested by capturing traffic (from the minio/mc client) to our gateway and adding various `Accept:` header values.  (The AWS sigv4 security protocol allows replays, that is so very useful for testing, if horribly badly wrong for actual security...)

Should fix #987 , but ideally wait for customer confirmation that it does before pulling :-)

arielshaqed:
@keimoon could you please take a look, too, and see whether this PR would fix your issue?

keimoon:
Unfortunately, alpakka does not send `Accept` header so this fix does not work for us. Currently we are using an work around (proxy the response to replace the response header) to overcome this issue.

johnnyaug (approved):

arielshaqed:
Thanks!  Also need to set the default to `application/xml` because @keimoon points out the Alpakka client doesn't pass an `Accepts:` header.  Turns out that passing this header just doesn't happen in the world of S3 clients.  Since S3 and Minio both seem happy to return application/xml, go with that.

gateway/handler.go
```diff
@@ -358,6 +385,7 @@ func (h *handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 	start := time.Now()
 	mrw := httputil.NewMetricResponseWriter(w)
+	setContentType(mrw, r)
```
ozkatz, line 396: Won't this return `application/xml` for non-xml responses? (i.e. `GetObject`?)

arielshaqed: > Won't this return `application/xml` for non-xml responses? (i.e. `GetObject`?)

Ouch, good catch!  I'll need to figure out how to make go-swagger prefer application/xml :-(

arielshaqed:
Dismissed reviews because of Oz' issue with adding content types to files and heads from the backing store.  Will re-request soon(-ish).

arielshaqed:
@johnnyaug @ozkatz PTAL, I cleaned up behaviour of `Accepts:`, `get-object` is back to ignoring everything except the auto-detected content-type.

gateway/handler.go
```diff
@@ -347,8 +348,46 @@ func notFound(w http.ResponseWriter, _ *http.Request) {
...
+	}
+	return nil
+}
+
+func setDefaultContentType(w http.ResponseWriter, r *http.Request) {
+	acceptable, ok := r.Header["Accept"]
+	if ok {
+		defaultContentType := selectContentType(acceptable)
+		// No requested content type matched.  This is OK at least for a GET or
+		// HEAD, so set up to auto-detect.
```
johnnyaug, line 378: Rephrase comment, or put it somewhere else

gateway/handler.go
```diff
@@ -347,8 +348,46 @@ func notFound(w http.ResponseWriter, _ *http.Request) {
...
+		if defaultContentType != nil {
+			w.Header().Set("Content-Type", *defaultContentType)
+		}
+	} else {
+		// For all proxied content the type will be reset according to whatever headers
```
johnnyaug, line 383: Also rephrase to clarify how this comment is related to this code

johnnyaug (approved):

arielshaqed:
@ozkatz are we good to go with this?

ozkatz (approved):
looks good!

arielshaqed:
Thanks!

PR #1000, opened 2020-12-05
Committing an object deletion to master erases object from all commits to master - #997

tzahij:
fixed in regular and by prefix listing


catalog/mvcc/cataloger_delete_entry_test.go
```diff
@@ -120,3 +122,94 @@ func testDeleteEntryCommitAndExpectNotFound(t *testing.T, ctx context.Context, c
 		t.Fatalf("DeleteEntry() get entry err = %s, want = %s", err, wantErr)
 	}
 }
+
+func TestCataloger_DeleteEntryAndCheckItRemainsInCommis(t *testing.T) {
```
arielshaqed, line 126:
```suggestion
func TestCataloger_DeleteEntryAndCheckItRemainsInCommits(t *testing.T) {
```

catalog/mvcc/cataloger_delete_entry_test.go
```diff
@@ -5,6 +5,8 @@ import (
 	"errors"
```
arielshaqed, line 5: Not sure I understand the relationship between these tests and #999.  Can we rebase on top of that, so we have those tests?  Or delete test from one side if the other side is better?

catalog/mvcc/cataloger_list_entries.go
```diff
@@ -286,30 +286,24 @@ func findLowestResultInBranches(branchRanges map[int64][]entryPathPrefixInfo, br
...
 	for _, l := range lineage {
-		unionMap[l.BranchID] = selectSingleBranch(l.BranchID, false, branchEntryLimit, l.CommitID, prefixLen, endOfPrefixRange)
+		unionMap[l.BranchID] = selectSingleBranch(l.BranchID, branchEntryLimit, l.CommitID, prefixLen, endOfPrefixRange)
 	}
 	return unionMap
 }
 
-func selectSingleBranch(branchID int64, isBaseBranch bool, branchBatchSize int, topCommitID CommitID, prefixLen int, endOfPrefixRange string) sq.SelectBuilder {
-	rawSelect := sq.Select("branch_id", "min_commit").
+func selectSingleBranch(branchID int64, branchBatchSize int, topCommitID CommitID, prefixLen int, endOfPrefixRange string) sq.SelectBuilder {
```
arielshaqed, line 296: Please document the function with these parameters.  The action of this function is confusing.

arielshaqed, line 296: Also I don't understand the action of parameter `isBaseBranch`, why we needed it, and why we no longer do.  I don't want to expose some bug that caused that parameter, and it is getting hard for me to track.
I can see that `isBaseBranch` used to change the behaviour of `max_commit`.  Is an unchanged leftover from changing the `min_commit` and `max_commit` values?

arielshaqed (commented):
Please rebase your branch on `master`.  Right now there are 35 commits in this branch, which makes it hard to review each commit separately.

I do not understand this commit well enough to approve it.  OTOH, I am therefore unequally unqualified to say that it is safe and does not break other behaviours.

catalog/mvcc/cataloger_delete_entry_test.go
tzahij: I started before 999 was opened. I just copied the test that was added in 999. Since this is the only change of 999, I think 999 can be deleted

arielshaqed:
Also, can we retitle this?  AFAICT it does *not* "erase the object from all commits from master", see my comments on #997 where I reproduced.  The objects are deleted from the _listings_, but they still exist.  We should avoid misinformation.

catalog/mvcc/views.go
```diff
@@ -14,11 +14,17 @@ const (
...
+		actualCommit = MaxCommitID
+	} else {
+		actualCommit = requestedCommit
+	}
 	entriesQ := sq.Select("*",
 		fmt.Sprintf("min_commit != %d AS is_committed", MinCommitUncommittedIndicator),
 		"max_commit = 0 AS is_tombstone",
 		"ctid AS entry_ctid\n",
-		fmt.Sprintf("max_commit < %d AS is_deleted", MaxCommitID)).
+		fmt.Sprintf("max_commit < %d AS is_deleted", actualCommit)).
```
nopcoder, line 27: Use the Column syntax - ex: .Column("min_commit != ? AS is_committed", MinCommitUncommittedIndicator)

nopcoder (changes requested):
The code doesn't pass the test

nopcoder (approved):

arielshaqed:
This comment (missed in the commits) needs to be addressed immediately:

> Also, can we retitle this? AFAICT it does _not_ "erase the object from all commits from master", see my comments on #997 where I reproduced. The objects are deleted from the _listings_, but they still exist. We should avoid misinformation.

The title of this PR is simply the title of issue #997.  That issue was titled without a root-cause analysis, since it came from outside (thanks again, @tsykes-2020, for a really good catch!).   To limit further fallout I am changing the PR title unilaterally.  @tzahij  and others: please feel free to correct the title further as needed!

PR #1008, opened 2020-12-07
TierFS enhancements

itaiad200:
1. Add metrics to pyramid.TierFS
2. Single access to block.Adapter during concurrent reads.
3. Thread-safe directory creation and deletions.


pyramid/directory.go
```diff
@@ -0,0 +1,73 @@
...
+	"fmt"
+	"io"
+	"os"
+	"path"
+	"path/filepath"
+	"sync"
+)
+
+// directory allows synchronization between file operation that might
+// change (create/delete) directories
```
arielshaqed, line 13:
```suggestion
// directory synchronizes between file operations that might change (create/delete) directories
```

pyramid/directory.go
```diff
@@ -0,0 +1,73 @@
...
+type directory struct {
+	sync.Mutex
+}
+
+func (d *directory) deleteDirRecIfEmpty(dir string) error {
```
arielshaqed, line 18: Please document this and other functions.

pyramid/directory.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+		if !empty {
+			return nil
+		}
+
+		parentDir := path.Dir(dir)
+		if err := os.Remove(dir); err != nil {
+			return err
+		}
+		dir = parentDir
```
arielshaqed, line 35: This loop can continue _way_ up, above the planned location.  I think `directory` needs a concept of a `ceilingDir` or `firstAncestorDir` or something.

pyramid/directory.go
```diff
@@ -0,0 +1,73 @@
...
+
+func isDirEmpty(name string) (bool, error) {
+	f, err := os.Open(name)
+	if err != nil {
+		return false, err
+	}
+	defer f.Close()
+
+	_, err = f.Readdirnames(1)
+	if err == io.EOF {
```
arielshaqed, line 47:
```suggestion
	if errors.Is(err, io.EOF) {
```

pyramid/file.go
```diff
@@ -7,36 +7,16 @@ import (
...
-
-func (f *File) Stat() (os.FileInfo, error) {
-	return f.fh.Stat()
-}
-
-func (f *File) Sync() error {
-	return f.fh.Sync()
-}
-
 func (f *File) Close() error {
```
arielshaqed, line 17: This existence of this function makes me doubt that the change at line 10 is safe.  Specifically, it makes `pyramid.File` have a safe conversion to `File`.  So now I can
```go
func logAndClose(f *os.File) {
  log("closing!")
  f.Close()
}

func unsafe(f *pyramid.File) {
  // use f...
  logAndClose(f.File)
}

File: pyramid/metrics.go [hunk 6]
```diff
@@ -0,0 +1,41 @@
...
+	prometheus.CounterOpts{
+		Name: "tier_fs_cache_miss_total",
+		Help: "TierFS cache miss total count",
+	}, []string{fsNameLabel})
+
+var evictionHistograms = promauto.NewHistogramVec(
+	prometheus.HistogramOpts{
+		Name:    "tier_fs_eviction_bytes",
+		Help:    "TierFS evicted object size by bytes",
+		Buckets: []float64{0.5 * kb, 1 * kb, 16 * kb, 32 * kb, 128 * kb, 512 * kb, 1 * mb, 2 * mb, 4 * mb, 8 * mb, 16 * mb, 64 * mb},
```
arielshaqed, line 31: It will be hard to compare buckets of these sizes, they are not equally-spaced on a log scale.  The multipliers between successive elements should be constant, here they are 2, 16, 2, 4, 4, 2, 2, 2, 2, 2, 4.  I think [`ExponentialBuckets`](https://godoc.org/github.com/prometheus/client_golang/prometheus#ExponentialBuckets) does exactly that.

pyramid/tier_fs.go
```diff
@@ -113,46 +119,23 @@ func handleExistingFiles(eviction eviction, fsLocalBaseDir string) error {
...
-	}
-
-	dir := path.Dir(p)
-	empty, err := isDirEmpty(dir)
-	if err != nil {
-		logger.WithError(err).WithField("dir", dir).Error("Checking if dir empty failed")
-		return
-	}
-	if !empty {
+		tfs.logger.WithError(err).WithField("path", p).Error("Removing file failed")
```
arielshaqed, line 132: Errors like this should also be monitored!

pyramid/metrics.go
```diff
@@ -0,0 +1,41 @@
...
+	prometheus.CounterOpts{
+		Name: "tier_fs_cache_hits_total",
+		Help: "TierFS cache hits total count",
+	}, []string{fsNameLabel})
+
+var cacheMiss = promauto.NewCounterVec(
+	prometheus.CounterOpts{
+		Name: "tier_fs_cache_miss_total",
+		Help: "TierFS cache miss total count",
+	}, []string{fsNameLabel})
```
arielshaqed, line 25: I very much prefer to monitor caches with a single metric "access", that has a "status" label "hit" or "miss".  That makes displaying miss percentages much easier, because no need to join the streams to get a total.

pyramid/tier_fs_test.go
```diff
@@ -138,7 +148,6 @@ func TestStartup(t *testing.T) {
 	bytes, err := ioutil.ReadAll(f)
 	require.NoError(t, err)
 	require.Equal(t, content, string(bytes))
-
 }
 
 func testEviction(t *testing.T, namespaces ...string) {
```
arielshaqed, line 153: This is used from multiple places, so AFAIU it needs to have a `t.Run()` somewhere around it -- either inside or at the call site.  Otherwise it's just harder to tell what happened.

pyramid/tier_fs_test.go
```diff
@@ -172,6 +181,35 @@ func TestInvalidArgs(t *testing.T) {
...
+	readsSoFar := adapter.gets.Get()
+
+	// try to read that file - only a single access to block storage is expected
+	concurrencyLevel := 50
+	adapter.wait = make(chan struct{})
+	var wg sync.WaitGroup
+	for i := 0; i < concurrencyLevel; i++ {
+		wg.Add(1)
+		go func() {
+			checkContent(t, namespace, filename, content)
```
arielshaqed, line 202: I don't understand this test.  https://github.com/treeverse/lakeFS/pull/1008/files#diff-2c0831b4a8b319091f86b0ee82570cb8780ba086dce76558f7e96c1638028ca8R245 says to read this one filename only once, and wait on the others.  And that is be tested  using `Get()`.  So what does the `adapter.wait` do or test?

pyramid/tier_fs_test.go
```diff
@@ -17,7 +23,7 @@ import (
 
 var (
 	fs      FS
-	adapter block.Adapter
+	adapter *memAdapter
```
arielshaqed, line 26: Can we just avoid global state and create its test with its own adapter?  As is we cannot run tests in parallel, and we need weird checks on the `gets` counter.

arielshaqed (changes requested):
Thanks!  Looks really good.

Important points:
1. No test for concurrent deletes, or for concurrent adds and deletes.
2. Monitoring should put hits and misses in same metric.

pyramid/directory.go
itaiad200: I was planning for rooted `workspace` dir to back me up, but it's indeed flaky. Fixed.

pyramid/file.go
itaiad200: Good catch, fixed!

pyramid/tier_fs_test.go
itaiad200: Not sure I follow - this is a utility function. If it fails any test (which it could), the failing test name will appear in the log.
What information is missing in your opinion?

pyramid/tier_fs_test.go
itaiad200: Well, mem adapter is rather fast. If I won't wrap it and use the code here, I'll get inconsistent behaviour.
The first Get to the adapter will read the file and block. Subsequent Gets will have the file on the disk and read it from there.
However if later on I'll introduce a bug that makes the Get call non-blocking, I won't catch it here since the mem adapter is fast enough to return the first Get with the file before the next calls are reading from the disk.

pyramid/tier_fs_test.go
itaiad200: Yes - fixed.

pyramid/directory.go
```diff
@@ -0,0 +1,84 @@
...
+	"path/filepath"
+	"sync"
+)
+
+// directory synchronizes between file operations that might change (create/delete) directories
+type directory struct {
+	// ceilingDir is the root directory of the FS - shouldn't never be deleted
+	ceilingDir string
+	sync.Mutex
+}
```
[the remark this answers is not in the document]
arielshaqed: :arrow_up: :arrow_up: :arrow_up: :arrow_up: :arrow_up: :question: :arrow_up: :arrow_up: :arrow_up: :arrow_up: :arrow_up:

arielshaqed (commented):
Thanks!

Concurrency testing is still  incomplete, it seems hard to synchronize all threads of execution to reach the single decision of allowing just one to complete.  Perhaps see how @ozkatz did it in `ChanLocker`?

pyramid/tier_fs_test.go
itaiad200: Synced f2f.

pyramid/directory.go
itaiad200: Adding documentation for now, really can't see how the paths (other than the base directory) will be user-controlled in the future.

arielshaqed: Cool.  I don't particularly mind "user-controlled" paths when that user is the owner of lakeFS.  I'm worried about "user-controlled" paths when they come from a lakeFS user, who is supposed only to have permissions to act on files _inside_ lakeFS.
Thanks!

arielshaqed (approved):
Great, thanks!  Sorry for the delayed approval, I forgot about this one.

PR #1010, opened 2020-12-07
cache map ready for PR

tzahij:


forest/cache_map/cache_map_manger.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func (h heapItemsType) Len() int           { return len(h) }
+func (h heapItemsType) Less(i, j int) bool { return h[i].weight > h[j].weight }
+func (h heapItemsType) Swap(i, j int)      { h[i], h[j] = h[j], h[i] }
+func (h *heapItemsType) Pop() interface{} {
+	old := *h
+	n := len(old)
+	x := old[n-1]
+	*h = old[0 : n-1]
```
arielshaqed, line 107: I don't understand why we need `old`, if we never modify `*h` until the last line.
```suggestion
	n := len(*h)
	x := (*h)[n-1]
	*h = *h[0 : n-1]
```

forest/cache_map/cache_map_manger.go
```diff
@@ -0,0 +1,112 @@
+package cachemap
```
arielshaqed, line 1: I think the filename is misspelled.  I am also confused what "man[a]ger" here means, it looks like a cache.

arielshaqed (changes requested):
1. Why do we need our own cache?  (_This is the single most important comment; let's answer it *before* we do anything else!_)
2. This implementation is not thread-safe.
3. AFAICT this is a decaying LFU, maybe write that down somewhere so the next person to try to figure it out has an easier time reading about it?
4. This implementation has poor performance.  In particular it has an O(n) code path (which is even O(n) amortized) during get.  I don't understand why that code path is required, see comment there for an alternative.  It also has an O(n) code path during set, which once for every cache miss.
5. There is no facility for monitoring this cache.  It will be impossible to know how well a cache is doing when it is inobservable.

arielshaqed:
Also please rename the PR to state what it does.  I don't think we need to go full [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), but e.g. https://chris.beams.io/posts/git-commit/ is a good simple guide.

ozkatz:
Iterating on what @arielshaqed asked - why do we need another cache implementation in lakeFS?
Ideally we should standardize on *one* unless we have a concrete reason to do so.

forest/cache_map/cache_map_manger.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	for _, minItem := range heapItems {
+		delete(c.cache, minItem.key)
+	}
+}
+
+func (h heapItemsType) Len() int           { return len(h) }
+func (h heapItemsType) Less(i, j int) bool { return h[i].weight > h[j].weight }
+func (h heapItemsType) Swap(i, j int)      { h[i], h[j] = h[j], h[i] }
+func (h *heapItemsType) Pop() interface{} {
```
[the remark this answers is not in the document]
nopcoder: missed this one... which check? nil?

tzahij:
> 1. Why do we need our own cache?  (_This is the single most important comment; let's answer it _before_ we do anything else!_)
I am sorry to say you are totally right
> 2. This implementation is not thread-safe.
right. It is easy to make it thread safe, but as you said before, that would be an exercise in futility
> 3. AFAICT this is a decaying LFU, maybe write that down somewhere so the next person to try to figure it out has an easier time reading about it?
right
> 4. This implementation has poor performance.  In particular it has an O(n) code path (which is even O(n) amortized) during get.  I don't understand why that code path is required, see comment there for an alternative.  It also has an O(n) code path during set, which once for every cache miss.
I beg to differ. it is O(1) for get and set. the decay process is called only once in NumberOfCaheEntries calls and may be tuned.  Eviction is rarely called, and in is O(n) + O(h*log h) where h is the number of entries to be evicted,
> 5. There is no facility for monitoring this cache.  It will be impossible to know how well a cache is doing when it is inobservable.
right

ozkatz:
Since there doesn't seem to be a concensus on the necesaity of this, I'm closing it for now. If anyone has good reasoning to do so, please reopen :)

tzahij:
agreed

________________________________

PR #1014, opened 2020-12-08
Change graveler iterator interface

itaiad200:


graveler/pgiterator.go
```diff
@@ -109,10 +123,12 @@ func (ri *PGRepositoryIterator) SeekGE(id RepositoryID) bool {
 	} else {
 		ri.buf = make([]*RepositoryRecord, 0)
 	}
-	return true
 }
 
 func (ri *PGRepositoryIterator) Value() *RepositoryRecord {
+	if ri.postInit || ri.postSeekGE {
```
guy-har, line 129: It would be nice to also return nil in case of error

graveler/pgiterator.go
```diff
@@ -91,14 +104,15 @@ func (ri *PGRepositoryIterator) fetch(initial bool) {
...
 
-func (ri *PGRepositoryIterator) SeekGE(id RepositoryID) bool {
+func (ri *PGRepositoryIterator) SeekGE(id RepositoryID) {
 	ri.offset = string(id)
 	ri.shouldFetch = true
 	ri.buf = make([]*RepositoryRecord, 0)
 	ri.fetch(true) // do a new initial fetch
-
+	ri.postSeekGE = true
+	ri.postInit = false
```
guy-har, line 113: If SeekGE doesn't return any value, doesn't make more sense to move the fetch into next?

guy-har (approved):
Looks good, Thanks

graveler/pgiterator.go
itaiad200: It does, fixed.

graveler/pgiterator.go
itaiad200: Done.

graveler/pgiterator.go
```diff
@@ -251,6 +243,11 @@ type PGCommitIterator struct {
 	ctx          context.Context
 	repositoryID RepositoryID
 
+	// initPhase turns true when the iterator was created or `SeekGE()` was called.
+	// initPhase turns false when Next() is called.
+	// When initPhase is true, Value() should return nil.
+	initPhase bool
```
ozkatz, line 249: why not use the value? it's nil when you initialize, it makes sense to set it to nil when you `SeekGE`

ozkatz (approved):

PR #1016, opened 2020-12-08
add documentation for list by prefix

tzahij:


catalog/mvcc/cataloger_list_entries.go
```diff
@@ -241,6 +251,8 @@ func processSinglePrefix(response []entryPathPrefixInfo, delimiter string, branc
 }
 
 func getMoreRows(path string, branch int64, branchRanges map[int64][]entryPathPrefixInfo, readParams readParamsType) error {
+	// if the buffer initially read is exhausted (e.g. - all entries that were read were deleted) - more rows are read for that branch
+	// results are directly stored in the branchRanges map. (agreed - there are simpler and clearer ways to do it)
```
arielshaqed, line 255: This comment is a bit hard to understand, because it is in the passive voice.  E.g.
> getMoreRows reads more rows for a branch after all its read are exhausted from (_I don't know where, actually_)..., and stores those results directly into branchRanges.

catalog/mvcc/cataloger_list_entries.go
```diff
@@ -283,17 +298,28 @@ func findLowestResultInBranches(branchRanges map[int64][]entryPathPrefixInfo, br
...
 	}
 	return unionMap
 }
 
 func buildSingleBranchQuery(branchID int64, branchBatchSize int, topCommitID CommitID, prefixLen int, endOfPrefixRange string) sq.SelectBuilder {
+	// builds a query on a single branch of the lineage. called mainly from "buildBaseLevelQuery" above.
+	// the other function that calls it is "getMoreRows" that needs entries for a single branch.
+	// topCommitId contains the requested commit for that branch. its implications:
+	// 1. entries where the minCommitId is more than the requested commit id will be filtered out
+	// 2. entries that were deleted after this commit (maxCommitId > topCommitId) will be considered
```
arielshaqed, line 320: This reads a bit like describing how the feature was implemented more than what the parameter does.  I _think_ that an equivalent description (with fewer details of implementation, and instead much more focus on what it does when you pass it):
> Returns entries as they were at topCommitID.
(Of course if this is wrong then don't copy this wrong description :smiley_cat: )

arielshaqed (changes requested):
Thanks!

Most important is to phrase comments as required for godoc, see first comment.  Also some wording fixes.

arielshaqed (approved):
Thanks!
These are still not standard Go comments. There is a very precise format for them in the links that I sent. Following that format means tools in the Go ecosystem can work on our code, which is good.
Approving because pulling this would make our code better. But please rephrase (to match the standard format).
Thanks!

PR #1021, opened 2020-12-09
use default amount for ui list branches amount

nopcoder:


webui/src/components/Tree.js
```diff
@@ -327,7 +327,7 @@ const Tree = ({path, list, repo, refId, diffResults, onNavigate, onDelete, showA
     const refreshData = useCallback(() => {
         if (refId.type === 'branch') {
             if (refId.id === repo.default_branch) {
-                listBranches(repo.id, "", 2) // trigger list branches to
+                listBranches(repo.id, "") // trigger list branches to
```
itaiad200, line 330: Will the amount param be sent at all, or is it empty("&amount=")?

nopcoder: It will fallback to use this one - 
https://github.com/treeverse/lakeFS/blob/master/webui/src/actions/api.js#L4

itaiad200 (approved):

PR #1024, opened 2020-12-09
fix gateway branch name validation

nopcoder:


ozkatz (approved):
Assiming this section will be refactored soon, LGTM :)

PR #1036, opened 2020-12-10
staging manager: drop by prefix

johnnyaug:


nopcoder (changes requested):

guy-har (approved):
LGTM
Thanks

graveler/staging.go
```diff
@@ -64,6 +66,33 @@ func (p *stagingManager) Drop(ctx context.Context, st StagingToken) error {
...
+	copy(upperBound, prefix)
+	for i := len(prefix) - 1; i >= 0; i-- {
+		if upperBound[i] == math.MaxUint8 {
+			upperBound = upperBound[:i]
+		} else {
+			useUpperBound = true
+			upperBound[i] += 1
+			break
+		}
+	}
```
nopcoder, line 81: 1. can we extract this one to a different func
2. can we lookup the upper bound index and based on the value, return nil, return position, or make a copy and increment. (using bytes.LastIndexByte for MaxUInt8)

graveler/staging_test.go
```diff
@@ -118,6 +116,158 @@ func TestDrop(t *testing.T) {
 	}
 }
 
+func TestDropByPrefix(t *testing.T) {
```
nopcoder, line 119: can you extract value - context.Background() use the same one for all the func scope

graveler/graveler.go
```diff
@@ -394,14 +394,17 @@ type StagingManager interface {
...
 	List(ctx context.Context, st StagingToken) (ValueIterator, error)
 
+	// DropKey clears a value by staging token and key
+	DropKey(ctx context.Context, st StagingToken, key Key) error
+
 	// Drop clears the given staging area
 	Drop(ctx context.Context, st StagingToken) error
+
+	// DropByPrefix drops all keys starting with the given prefix, from the given staging area
+	DropByPrefix(ctx context.Context, st StagingToken, prefix Key) error
```
nopcoder, line 407: DropPrefix or change DropKey to DropByKey

graveler/staging_test.go
```diff
@@ -118,6 +116,158 @@ func TestDrop(t *testing.T) {
...
+	if err != nil {
+		// key0000 does not start with the deleted prefix - should be returned
+		t.Fatalf("got unexpected error: %v", err)
+	}
+	it, _ := s.List(context.Background(), "t1")
+	count := 0
+	for it.Next() {
+		count++
+	}
+	it.Close()
```
nopcoder, line 151: extract method? also can check the error or extract smaller scope and have count by iter that will use anything with Next to return length

graveler/staging_test.go
```diff
@@ -118,6 +116,158 @@ func TestDrop(t *testing.T) {
...
+	if count != numOfValues {
+		t.Errorf("got unexpected number of results. expected=%d, got=%d", numOfValues, count)
+	}
+}
+
+func TestDropPrefixBytes(t *testing.T) {
```
nopcoder, line 166: usually, when you use idea to create a test function it will also add 'name' as the test parameter which helps describe the use case we are testing. in this case, it looks like it will be t<number> which is less descriptive.

graveler/staging_test.go
```diff
@@ -118,6 +116,158 @@ func TestDrop(t *testing.T) {
...
+			t.Fatalf("got unexpected error: %v", err)
+		}
+		it, err := s.List(context.Background(), st)
+		if err != nil {
+			t.Fatalf("got unexpected error: %v", err)
+		}
+		count := 0
+		for it.Next() {
+			count++
+		}
```
nopcoder, line 261: missing call to it.Close()

nopcoder (changes requested):

graveler/staging_test.go
johnnyaug: use t.Run with name

graveler/staging.go
```diff
@@ -64,6 +66,39 @@ func (p *stagingManager) Drop(ctx context.Context, st StagingToken) error {
...
+	if incrementIdx == -1 {
+		return nil
+	}
+	upperBound := make(Key, incrementIdx+1)
+	copy(upperBound, prefix[:incrementIdx+1])
+	upperBound[incrementIdx] += 1
+	return upperBound
```
nopcoder, line 99:
```suggestion
	idx := len(prefix) - 1
	for idx >= 0 && prefix[idx] != math.MaxUint8 {
		idx--
	}
	if idx == -1 {
		return nil
	}
	upperBound := make(Key, idx+1)
	copy(upperBound, prefix[:idx+1])
	upperBound[idx]++
	return upperBound
```

nopcoder (approved):
approved, had just one minor how to reuse the index variable (optional)

PR #1040, opened 2020-12-10
Allow lakectl to run without a config file

itaiad200:


cmd/lakectl/cmd/root.go
```diff
@@ -41,7 +42,11 @@ lakectl is a CLI tool allowing exploration and manipulation of a lakeFS environm
 			return
 		}
 		if _, ok := cfgFileErr.(viper.ConfigFileNotFoundError); ok {
-			// specific message in case the file doesn't not found
+			if emptyCfgFile {
```
nopcoder, line 45: can you check `cfgFile` empty?

itaiad200: It's never empty, since we use the default value. 
But if that file isn't found, you can still try and run with the default values or one read from the env file, no?

cmd/lakectl/cmd/root.go
```diff
@@ -41,7 +41,11 @@ lakectl is a CLI tool allowing exploration and manipulation of a lakeFS environm
 			return
 		}
 		if _, ok := cfgFileErr.(viper.ConfigFileNotFoundError); ok {
```
nopcoder, line 43: if len(cfgFile) > 0 && errors.As(cfgFileErr, &viper.ConfigFileNotFoundError{}) {

itaiad200: This is also inaccurate as it will catch the second condition for `cfgFileErr != nil`.
Refactored again.

nopcoder (approved):

PR #1041, opened 2020-12-10
Implementation of sstable layer

itaiad200:
Including temp interface for the cache and serialization.

---
This change is [Reviewable](https://reviewable.io/reviews/treeverse/lakefs/1041)


graveler/committed/sstable/manager.go
```diff
@@ -0,0 +1,96 @@
...
+type PebbleSSTableManager struct {
+	cache  cache
+	fs     pyramid.FS
+	logger logging.Logger
+	hash   hash.Hash
+}
+
+const sstableTierFSNamespace = "sstables"
+
+func NewPebbleSSTableManager(cache cache, fs pyramid.FS) Manager {
```
ozkatz, line 25: cso this implements `sstable.Manager`?
I have to say, looking at the interface itself, it is *completely* decoupled from the fact that the underlying format is an sstable! that's a good thing. I think maybe it could even be defined one layer up, as part of `committed`. This way, trees work against a truely generic interface dealing with storage and are not bound to a specific format (as a matter of fact, this is alreay the case, it's only the package name of `sstable.Manager` that implies any sort of coupling..)

@itaiad200  wdyt?

graveler/committed/sstable/manager.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// SSTableIterator takes a given SSTable and returns an EntryIterator seeked to >= "from" path
+func (m *PebbleSSTableManager) NewSSTableIterator(ns Namespace, tid ID, from graveler.Key) (graveler.ValueIterator, error) {
+	reader, derefer, err := m.cache.GetOrOpen(string(ns), tid)
+	if err != nil {
+		return nil, err
+	}
+
+	iter, err := reader.NewIter(from, nil)
```
tzahij, line 76: It is hard to understand from Pebble documentation, but I think that after a call to NewIter, there is no specific location. Think yo have to call First() or SeekGE to position the reading point to the start.

tzahij (changes requested):

graveler/committed/sstable/iterator.go
```diff
@@ -0,0 +1,69 @@
...
+	if !iter.postSeek {
+		iter.currKey, iter.currValue = iter.it.Next()
+	}
+	iter.postSeek = false
+
+	if iter.currKey == nil && iter.currValue == nil {
+		return false
+	}
+
+	iter.valParsed, iter.err = deserializeValue(iter.currValue)
```
arielshaqed, line 37: Nice!  So IIUC if deserialization fails for one element, I can still call `Next()` again and succeed on the next element :-)

graveler/committed/sstable/iterator.go
```diff
@@ -0,0 +1,69 @@
...
+		Key:   iter.currKey.UserKey,
+		Value: iter.valParsed,
+	}
+}
+
+func (iter *Iterator) Err() error {
+	return iter.err
+}
+
+func (iter *Iterator) Close() {
```
arielshaqed, line 56: We cannot return `error` here because of the interface we chose for iterators?  That makes me sad :-(

graveler/committed/sstable/manager.go
```diff
@@ -0,0 +1,96 @@
...
+		return nil, fmt.Errorf("create iterator: %w", err)
+	}
+	defer m.execAndLog(it.Close, "Failed to close iterator")
+
+	// actual reading
+	key, val := it.Next()
+	if key == nil {
+		// checking if an error occurred or key simply not found
+		if it.Error() != nil {
+			return nil, fmt.Errorf("reading key: %w", it.Error())
```
arielshaqed, line 54: Include `lookup`, `tid` to make errors easier to debug?  (But perhaps not if we consider them user data...)

graveler/committed/sstable/writer.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	// updating stats
+	if dw.count == 0 {
+		dw.first = record.Key
+	}
+	dw.last = record.Identity
+	dw.count++
+
+	if _, err := dw.hash.Write(keyBytes); err != nil {
```
arielshaqed, line 59: I don't think this hashing works.  Specifically, it will hash `keyBytes, valueBytes = []byte{1,2,3}, []byte{4,5}` to the same value as `keyBytes, valueBytes = []byte{1,2},[]byte{3,4,5}`.  Hashing is like marshalling (serialization): you need to write things that you _could_ unmarshal, had the hash only not clobbered them.  So e.g. first write `len(keyBytes)` (in you favourite binary encoding) and then `keyBytes`.  Or choose a separator and escape it out of `keyBytes`.  Or pass `dw.hash` to some existing marshalling library.

arielshaqed (changes requested):
Thanks!  Lots of great stuff here.  But I think at least `BatchCloser` needs tests.

itaiad200 (commented):
> *[Reviewable](https://reviewable.io/reviews/treeverse/lakefs/1041)* status: 0 of 16 files reviewed, 13 unresolved discussions (waiting on @arielshaqed, @ozkatz, and @tzahij)

___
*[graveler/committed/sstable/batch.go, line 16 at r1](https://reviewable.io/reviews/treeverse/lakefs/1041#-MOVi8JcwIZADooPDwGk-r1-16:-MOVncys6AFXHUdSKmM9:b-896fix) ([raw file](https://github.com/treeverse/lakefs/blob/18e6a2e9aafd1b1e0415b5f2c88107ae855a0dca/graveler/committed/sstable/batch.go#L16)):*
<details><summary><i>Previously, itaiad200 wrote…</i></summary><blockquote>

Added.
</blockquote></details>

Done.
___
*[graveler/committed/sstable/batch.go, line 27 at r1](https://reviewable.io/reviews/treeverse/lakefs/1041#-MOVi8JcwIZADooPDwGk-r1-27:-MOVnehb9RwKVdz2UVbe:b-896fix) ([raw file](https://github.com/treeverse/lakefs/blob/18e6a2e9aafd1b1e0415b5f2c88107ae855a0dca/graveler/committed/sstable/batch.go#L27)):*
<details><summary><i>Previously, itaiad200 wrote…</i></summary><blockquote>

It's not synchronised. However synchronisation isn't needed, worst case is that you closed a writer you didn't have to, err won't be overridden with nil anyhow.
</blockquote></details>

Done.
___
*[graveler/committed/sstable/batch.go, line 46 at r1](https://reviewable.io/reviews/treeverse/lakefs/1041#-MOVi8JcwIZADooPDwGk-r1-46:-MOVng65BWEHPdmhLu5h:b-896fix) ([raw file](https://github.com/treeverse/lakefs/blob/18e6a2e9aafd1b1e0415b5f2c88107ae855a0dca/graveler/committed/sstable/batch.go#L46)):*
<details><summary><i>Previously, itaiad200 wrote…</i></summary><blockquote>

synced f2f, go ftw.
</blockquote></details>

Done.
___
*[graveler/committed/sstable/batch.go, line 60 at r1](https://reviewable.io/reviews/treeverse/lakefs/1041#-MOVi8JcwIZADooPDwGk-r1-60:-MOVnhjP7CabaBuUn00Z:b-896fix) ([raw file](https://github.com/treeverse/lakefs/blob/18e6a2e9aafd1b1e0415b5f2c88107ae855a0dca/graveler/committed/sstable/batch.go#L60)):*
<details><summary><i>Previously, itaiad200 wrote…</i></summary><blockquote>

Good one.
</blockquote></details>

Done.
___
*[graveler/committed/sstable/iterator.go, line 37 at r1](https://reviewable.io/reviews/treeverse/lakefs/1041#-MOVi8JcwIZADooPDwGm-r1-37:-MOVniR8BDtMnyeyjAaj:b-896fix) ([raw file](https://github.com/treeverse/lakefs/blob/18e6a2e9aafd1b1e0415b5f2c88107ae855a0dca/graveler/committed/sstable/iterator.go#L37)):*
<details><summary><i>Previously, itaiad200 wrote…</i></summary><blockquote>

Yap, the interface doesn't declare what's the expected behaviour in that case.
</blockquote></details>

Done.
___
*[graveler/committed/sstable/iterator.go, line 56 at r1](https://reviewable.io/reviews/treeverse/lakefs/1041#-MOVi8JcwIZADooPDwGm-r1-56:-MOVnjqFCSFwaP1MLEf_:b-896fix) ([raw file](https://github.com/treeverse/lakefs/blob/18e6a2e9aafd1b1e0415b5f2c88107ae855a0dca/graveler/committed/sstable/iterator.go#L56)):*
<details><summary><i>Previously, itaiad200 wrote…</i></summary><blockquote>

You and me both.
</blockquote></details>

Done.
___
*[graveler/committed/sstable/manager.go, line 25 at r1](https://reviewable.io/reviews/treeverse/lakefs/1041#-MOVi8JcwIZADooPDwGn-r1-25:-MOVnkVnEWDD6bZF_qEe:b-896fix) ([raw file](https://github.com/treeverse/lakefs/blob/18e6a2e9aafd1b1e0415b5f2c88107ae855a0dca/graveler/committed/sstable/manager.go#L25)):*
<details><summary><i>Previously, itaiad200 wrote…</i></summary><blockquote>

I approve this message, Done.
</blockquote></details>

Done.
___
*[graveler/committed/sstable/manager.go, line 54 at r1](https://reviewable.io/reviews/treeverse/lakefs/1041#-MOVi8JcwIZADooPDwGn-r1-54:-MOVnoSO1E0BJLDk6vi7:b-896fix) ([raw file](https://github.com/treeverse/lakefs/blob/18e6a2e9aafd1b1e0415b5f2c88107ae855a0dca/graveler/committed/sstable/manager.go#L54)):*
<details><summary><i>Previously, itaiad200 wrote…</i></summary><blockquote>

I guess lookup is user data, tid isn't.
</blockquote></details>

Done.
___
*[graveler/committed/sstable/serialization.go, line 7 at r1](https://reviewable.io/reviews/treeverse/lakefs/1041#-MOVi8JcwIZADooPDwGo-r1-7:-MOVnzAc07vw7OM3BSys:b-896fix) ([raw file](https://github.com/treeverse/lakefs/blob/18e6a2e9aafd1b1e0415b5f2c88107ae855a0dca/graveler/committed/sstable/serialization.go#L7)):*
<details><summary><i>Previously, itaiad200 wrote…</i></summary><blockquote>

Yap..
</blockquote></details>

Done.
___
*[graveler/committed/sstable/writer.go, line 59 at r1](https://reviewable.io/reviews/treeverse/lakefs/1041#-MOVi8JcwIZADooPDwGp-r1-59:-MOVnred6uYQFaH9hZhi:b-896fix) ([raw file](https://github.com/treeverse/lakefs/blob/18e6a2e9aafd1b1e0415b5f2c88107ae855a0dca/graveler/committed/sstable/writer.go#L59)):*
<details><summary><i>Previously, itaiad200 wrote…</i></summary><blockquote>

Nice, done.
</blockquote></details>

Done.

graveler/committed/sstable/batch.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+type BatchCloser struct {
+	results []committed.WriteResult
+	err     atomic.Error
+
+	wg sync.WaitGroup
+
+	// resultsLock locks any access to the accumulated results
+	resultsLock sync.Mutex
```
arielshaqed, line 19:
```suggestion
	// resultsLock locks any access to results.
	resultsLock sync.Mutex
```

graveler/committed/sstable/batch.go
```diff
@@ -0,0 +1,68 @@
...
+	"sync"
+
+	"github.com/treeverse/lakefs/graveler/committed"
+
+	"go.uber.org/atomic"
+)
+
+type BatchCloser struct {
+	results []committed.WriteResult
+	err     atomic.Error
```
arielshaqed, line 14: Not sure, the package seems to prefer [`*Error`](https://godoc.org/go.uber.org/atomic#NewError).  In particular I am not sure you want what would happen to `err` if I copied a `BatchCloser`.  `Error` holds a (`sync/atomic`) [`Value`](https://golang.org/pkg/sync/atomic/#Value), which "must not be copied after first use".

arielshaqed: I reckon we should get rid of atomics here, and possibly throughout the file.  I am not sure that they are worth the effort and the weird semantics that modern architectures + Go + Uber combine to put on them.

graveler/committed/sstable/batch.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func (bc *BatchCloser) closeWriter(w committed.Writer) {
+	defer bc.wg.Done()
+	res, err := w.Close()
+	if err != nil {
+		if bc.err.Load() == nil {
+			// keeping first error is enough
+			bc.err.Store(err)
+		}
```
arielshaqed, line 49: This is still a race, and does not keep the first error (but it does keep _an_ error).  You would need a `Swap` operation like [`Uint64.Swap](https://godoc.org/go.uber.org/atomic#Uint64.Swap), probably using `unsafe.Pointer`s and [`atomic.SwapPointer`](https://golang.org/pkg/sync/atomic/#SwapPointer).
But again, I doubt that it matters.  You could just change the semantics and
```suggestion
	if err != nil {
		// keeping last error is enough
		bc.err.Store(err)
	}
```

arielshaqed: Or _don't use atomics!  Why bother to speed up the error-handling case when the good path (below) has a lock?

graveler/committed/sstable/batch.go
```diff
@@ -0,0 +1,68 @@
...
+
+	bc.resultsLock.Lock()
+	defer bc.resultsLock.Unlock()
+	bc.results = append(bc.results, *res)
+}
+
+// Wait returns when all Writers finished.
+// Any failure to close a single Writer will return with a nil results slice and an error.
+func (bc *BatchCloser) Wait() ([]committed.WriteResult, error) {
+	err := bc.err.Load()
```
arielshaqed, line 61: Same race with error reporting here.

arielshaqed (changes requested):
1. I think we could just get rid of (all) the atomic operations.  There is not enough to speed up using them, they tend to appear next to a slower path.  The code would be easier for me to read in the mutex just protected everything.  The package `sync/atomic` has this to say:
   >  These functions require great care to be used correctly. Except for special, low-level applications, synchronization is better done with channels or the facilities of the sync package. Share memory by communicating; don't communicate by sharing memory.
2. Not sure that using the mocks is easier than just writing the fakes.  (But not a mocking lover; this comment is in any case *_not required_*.)

itaiad200 (commented):
1. Done.
2. We'll see..

> *[Reviewable](https://reviewable.io/reviews/treeverse/lakefs/1041#-:-MOWgNqc1RjvCcyGF8y5:b-6bz530)* status: 0 of 22 files reviewed, 17 unresolved discussions (waiting on @arielshaqed, @ozkatz, and @tzahij)

___
*[graveler/committed/sstable/batch.go, line 14 at r3](https://reviewable.io/reviews/treeverse/lakefs/1041#-MOVi8JcwIZADooPDwGk-r3-14:-MOWgW8kAL6vnImrx9yr:b-896fix) ([raw file](https://github.com/treeverse/lakefs/blob/976f2506b3c6ad0fb2b354464e382937079a3e1a/graveler/committed/sstable/batch.go#L14)):*
<details><summary><i>Previously, arielshaqed wrote…</i></summary><blockquote>

I reckon we should get rid of atomics here, and possibly throughout the file.  I am not sure that they are worth the effort and the weird semantics that modern architectures + Go + Uber combine to put on them.
</blockquote></details>

Done.
___
*[graveler/committed/sstable/batch.go, line 19 at r3](https://reviewable.io/reviews/treeverse/lakefs/1041#-MOVi8JcwIZADooPDwGk-r3-19:-MOWga4F-nCBAYau2RZg:b-896fix) ([raw file](https://github.com/treeverse/lakefs/blob/976f2506b3c6ad0fb2b354464e382937079a3e1a/graveler/committed/sstable/batch.go#L19)):*
<details><summary><i>Previously, arielshaqed wrote…</i></summary><blockquote>

```suggestion
	// resultsLock locks any access to results.
	resultsLock sync.Mutex
```
</blockquote></details>

Done.
___
*[graveler/committed/sstable/batch.go, line 49 at r3](https://reviewable.io/reviews/treeverse/lakefs/1041#-MOVi8JcwIZADooPDwGk-r3-49:-MOWgcpk65Htem2259BU:b-896fix) ([raw file](https://github.com/treeverse/lakefs/blob/976f2506b3c6ad0fb2b354464e382937079a3e1a/graveler/committed/sstable/batch.go#L49)):*
<details><summary><i>Previously, arielshaqed wrote…</i></summary><blockquote>

Or _don't use atomics!  Why bother to speed up the error-handling case when the good path (below) has a lock?
</blockquote></details>

Done.
___
*[graveler/committed/sstable/batch.go, line 61 at r3](https://reviewable.io/reviews/treeverse/lakefs/1041#-MOVi8JcwIZADooPDwGk-r3-61:-MOWgefo2muc62NNngx7:b-896fix) ([raw file](https://github.com/treeverse/lakefs/blob/976f2506b3c6ad0fb2b354464e382937079a3e1a/graveler/committed/sstable/batch.go#L61)):*
<details><summary><i>Previously, arielshaqed wrote…</i></summary><blockquote>

Same race with error reporting here.
</blockquote></details>

Done.

graveler/committed/sstable/batch.go
```diff
@@ -0,0 +1,93 @@
...
+
+// CloseWriterAsync adds Writer instance for the BatchWriterCloser to handle.
+// Any writes executed to the writer after this call are not guaranteed to succeed.
+// If Wait() has already been called, returns an error.
+func (bc *BatchCloser) CloseWriterAsync(w committed.Writer) error {
+	bc.lock.Lock()
+	defer bc.lock.Unlock()
+
+	if bc.err != nil {
+		// don't accept new writers if previous error occurred
```
arielshaqed, line 40:
```suggestion
		// don't accept new writers if previous error occurred.  In particular, if Wait has started then this is errMultipleWaitCalls.
```

graveler/committed/sstable/serialization.go
```diff
@@ -0,0 +1,10 @@
+package sstable
+
+import "github.com/treeverse/lakefs/graveler"
+
+// serializer is a placeholder interface for committed way of
+// serializing the graveler.Value
+type serializer interface {
+	SerializeValue(inVal graveler.Value) ([]byte, error)
```
arielshaqed, line 8: :spiral_notepad:  See #1067.  The plan is to move this _outside_ of this directory, `committed` doesn't care.  (No action required here and now, of course!)

arielshaqed (approved):
Great stuff, thanks!

graveler/committed/sstable/serialization.go
itaiad200: Yap - that's my next PR.

graveler/committed/sstable/manager.go
```diff
@@ -0,0 +1,96 @@
...
+	defer m.execAndLog(derefer, "Failed to dereference reader")
+
+	it, err := reader.NewIter(lookup, nil)
+	if err != nil {
+		return nil, fmt.Errorf("create iterator: %w", err)
+	}
+	defer m.execAndLog(it.Close, "Failed to close iterator")
+
+	// actual reading
+	key, val := it.Next()
```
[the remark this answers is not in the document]
itaiad200: I don't ;)

tzahij (approved):

PR #1044, opened 2020-12-10
Add CacheWithDisposal to cache resources that require disposal

arielshaqed:
Whenever an element is fetched from the cache, a value is returned along with a deref function
that must be called to allow the element to be returned.  Once evicted, the object can only be
disposed of after all its references have been returned to the cache.

The implementation is mostly just mixing a reference count onto each entry in the cache.

And now, a rant.

Caching resources that require disposal (such as anything with a file descriptor) is difficult
in garbage-collected languages, _because_ of garbage collection.  Only memory presure triggers
GC, so it is entirely inappropriate for handling other resources.  But the lack of destructors
leaves no way to force resource cleanups.  Instead we have to return an explicit function that
returns the object to the pool.  A language such as C++ calls destructors at clear places, and
would allow returning special object references that handle lifetimes.  In fact it already has
`shared_ptr` in the standard library.

---
This change is [Reviewable](https://reviewable.io/reviews/treeverse/lakefs/1044)

### DIFF
--- a/cache/refcnted.go
+++ b/cache/refcnted.go
@@ -26,6 +26,7 @@ type CacheWithDisposal interface {
 
 type ParamsWithDisposal struct {
 	Name   string
+	Logger logging.Logger
 	Size   int
 	Shards int
 	// OnDispose disposes of an entry.  It is called when the last reference to an entry
@@ -108,10 +109,7 @@ func NewSingleThreadedCacheWithDisposal(p ParamsWithDisposal) *SingleThreadedCac
 	onEvict := func(k interface{}, v interface{}) {
 		entry := v.(*cacheEntry)
 		err := entry.release(ret)
-		logging.Default().WithFields(logging.Fields{
-			"key":   k,
-			"value": v,
-		}).WithError(err).Error("[internal] failed to release during eviction")
+		p.Logger.WithField("key", k).WithError(err).Error("[internal] failed to release during eviction")
 	}
 	ret.p = &p
 	var err error
--- a/cache/refcnted.go
+++ b/cache/refcnted.go
@@ -1,0 +1,113 @@
+package cache
+
+import (
+	"errors"
+	"fmt"
+	"sync"
+	"sync/atomic"
+
+	"github.com/hnlq715/golang-lru/simplelru"
+
+	"github.com/treeverse/lakefs/logging"
+)
+
+// Derefer is the type of a function that returns an object to the cache, possibly causing
+// its eviction.
+type Derefer func() error
+
+// CacheWithDisposal is a cache that calls a disposal callback when an entry has been evicted
+// and is no longer in use.  Because of the structure of Go, it uses release callbacks and is
+// not a Cache.
+type CacheWithDisposal interface {
+	Name() string
+	GetOrSet(k interface{}, setFn SetFn) (v interface{}, disposer Derefer, err error)
+}
+
+type ParamsWithDisposal struct {
+	Name string
+	Size int
+	// OnDispose disposes of an entry.  It is called when the last reference to an entry
+	// is released.  If it fails not much can be done.
+	OnDispose func(value interface{}) error
+}
+
+// SingleThreadedCacheWithDisposal is a CacheWithDisposal that uses a single critical section
+// to prevent concurrent access to the cache and to reference counts on values.
+type SingleThreadedCacheWithDisposal struct {
+	p    *ParamsWithDisposal
+	name string
+	mu   sync.Mutex // protects lru
+	lru  *simplelru.LRU
+}
+
+// cacheEntry is a single entry in the cache.  It uses only atomic operations on refs and is
+// safe for concurrent use.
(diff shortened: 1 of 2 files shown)


cache/refcnted.go
```diff
@@ -0,0 +1,92 @@
...
+func (rcc *ReferenceCountedCache) GetOrSet(k interface{}, setFn SetFn) (interface{}, Derefer, error) {
+	e, err := rcc.c.GetOrSet(k, func() (interface{}, error) {
+		v, err := setFn()
+		var ret *cacheEntry
+		if v != nil {
+			ret = &cacheEntry{refs: 1, value: v}
+		}
+		return ret, err
+	})
+	entry := e.(*cacheEntry)
```
ozkatz, line 88: assuming the setfunc doesn't get called (i.e. this is cache hit, where does refs get incremented?

cache/refcnted.go
```diff
@@ -0,0 +1,92 @@
...
+func (e *cacheEntry) release(rcc *ReferenceCountedCache) error {
+	refs := atomic.AddInt32(&e.refs, -1)
+	if refs < 0 {
+		return fmt.Errorf("release from %s: %w %d; may leak or fail", rcc.Name(), ErrNegativeReferenceCount, refs)
+	}
+	if refs > 0 {
+		return nil
+	}
+
+	return rcc.p.OnDispose(e.value)
```
ozkatz, line 55: I'm not sure where refs gets incremented, but I'm guessing there's a chance it could happen between the atomic decrement above and the OnDispose function getting called, leading to it getting called when refs is actually > 0. Can't determine if this race exist since I don't understand how refs ever grows beyond 1

ozkatz (changes requested):
Looks good - I like the tests. There's one thing I think is missing - refs incrementation when there's a cache hit and the set function of GetOrSet doesn't get called

arielshaqed (commented):
> Looks good - I like the tests. There's one thing I think is missing - refs incrementation when there's a cache hit and the set function of GetOrSet doesn't get called

This is of course all true... except for the bit about liking the tests.  They were in fact incomplete and completely missed the bug that you uncovered.

Rewritten to be mostly a critical section around reference counting.

PTAL...

ozkatz (changes requested):
> *[Reviewable](https://reviewable.io/reviews/treeverse/lakefs/1044)* status: 0 of 2 files reviewed, 2 unresolved discussions (waiting on @arielshaqed and @ozkatz)

___
*[cache/refcnted.go, line 82 at r1](https://reviewable.io/reviews/treeverse/lakefs/1044#-MOWA5Co79mZ-lOGW21b:-MOWA5Co79mZ-lOGW21c:b-1i1pla) ([raw file](https://github.com/treeverse/lakefs/blob/c41d96b01880c5937b8e06f90863e29eb2f5525b/cache/refcnted.go#L82)):*
> ```Go
> 		entry := v.(*cacheEntry)
> 		err := entry.release(ret)
> 		logging.Default().WithFields(logging.Fields{
> ```

cached values are opaque and might contain sensitive information, not sure I'd spill them blindly to the log.
Additionally, might be better to receive a logger in the constructor ? 
___
*[cache/refcnted.go, line 104 at r1](https://reviewable.io/reviews/treeverse/lakefs/1044#-MOWAyYz0QrWgMIYwP68:-MOWAyYz0QrWgMIYwP69:b-7dvr1q) ([raw file](https://github.com/treeverse/lakefs/blob/c41d96b01880c5937b8e06f90863e29eb2f5525b/cache/refcnted.go#L104)):*
> ```Go
> 		entry.acquire()
> 	} else {
> 		v, err := setFn()
> ```

so we're serializing all `setFn()`s of all keys? That's an expensive operation.

arielshaqed:
Not responding on Reviewable because it asks for permissions on *my* (public) repos in order to work on *Treeverse* repos.  Not responding inline because Reviewable put comments from @ozkatz on a single mega-comment up top.
Sorry.

> cache/refcnted.go, line 82 at r1 (raw file):
>
>    		entry := v.(*cacheEntry)
>    		err := entry.release(ret)
>    		logging.Default().WithFields(logging.Fields{
>
> cached values are opaque and might contain sensitive information, not sure I'd spill them blindly to the log.
Additionally, might be better to receive a logger in the constructor ?

Right and righter.  96144d2c07b8aad22198d282929c5cdc672de475

> cache/refcnted.go, line 104 at r1 (raw file):

>    		entry.acquire()
>    	} else {
>    		v, err := setFn()
>
> so we're serializing all setFn()s of all keys? That's an expensive operation.

We actually have to do this because LRU doesn't give us enough atomicity in its operations to be able to add reference counts.  Instead of fixing this I just sharded everything in f8edf33c494433f327eb5c6a41fc98b5bcba416e.

Thanks; PTAL!

ozkatz (approved):
Nice work!

arielshaqed:
Thanks!  Pulling...

PR #1050, opened 2020-12-12
ResolvePath align check and add tests

nopcoder:
Testing extract branch and path from Gateway request path. Verify value from start.

---
This change is [Reviewable](https://reviewable.io/reviews/treeverse/lakefs/1050)


gateway/path/resolver_test.go
```diff
@@ -0,0 +1,70 @@
...
+	type args struct {
+		encodedPath string
+	}
+	tests := []struct {
+		name    string
+		args    args
+		want    ResolvedPath
+		wantErr bool
+	}{
+		{
```
itaiad200, line 18: Could you add cases that starts or ends with `/` ? I think the behaviour is inconsistent for these cases.

itaiad200 (changes requested):

itaiad200 (approved):
> *[Reviewable](https://reviewable.io/reviews/treeverse/lakefs/1050)* status: 0 of 2 files reviewed, all discussions resolved (waiting on @itaiad200)

PR #1053, opened 2020-12-13
Fix race during concurrent cache entry creation

arielshaqed:
Previously concurrent calls to GetSet on the same cache entry would fail all except one ("the first", for some ordering of the calls).  But given that a value returned from GetSet moments _after_ the creation would be good, and that a call moments _before_ creation would shift creation and also work, the lifetime of a value is anyway *not* determined by the cache.  So best to return the newly computed value.

---
This change is [Reviewable](https://reviewable.io/reviews/treeverse/lakefs/1053)


cache/cache_test.go
```diff
@@ -0,0 +1,46 @@
...
+				k := j % worldSize
+				kk, err := c.GetOrSet(k, func() (interface{}, error) {
+					return k * k, nil
+				})
+				if err != nil {
+					t.Error(err)
+					return
+				}
+				if kk.(int) != k*k {
+					t.Errorf("[%d] got %d^2=%d, expected %d", i, k, kk, k*k)
```
nopcoder, line 38: goroutine can't call Error or Fatal

arielshaqed: So I thought so too, then I forgot, so now I had to check.  https://golang.org/pkg/testing/#T says I am allowed to call Error -- but _not_ Fatal -- in a goroutine:

>  A test ends when its Test function returns or calls any of the methods FailNow, Fatal, Fatalf, SkipNow, Skip, or Skipf. Those methods, as well as the Parallel method, must be called only from the goroutine running the Test function.
>
> The other reporting methods, such as the variations of Log and Error, may be called simultaneously from multiple goroutines. 

So "citation required"?

arielshaqed (commented):
Thanks!

@ozkatz please re-review: singleflight adds nothing, it is not a cache, and prevents using non-string keys, and needs explicit calls to `Forget`, so not using it.
@nopcoder I found documentation that seems to allow calling `t.Error`; could you re-check?

Thanks!

cache/cache_test.go
nopcoder: You got it right, my bad - the call to Error will work

pyramid/tier_fs.go
```diff
@@ -242,35 +242,33 @@ func (tfs *TierFS) openFile(fileRef localFileRef, fh *os.File) (*ROFile, error)
 // and places it in the local FS for further reading.
 // It returns a file handle to the local file.
 func (tfs *TierFS) readFromBlockStorage(fileRef localFileRef) (*os.File, error) {
-	var e error
-	tfs.keyLock.Lock(fileRef.filename, func() {
+	_, err := tfs.keyLock.Compute(fileRef.filename, func() (interface{}, error) {
+		var err error
```
nopcoder, line 246: do we need this one?

nopcoder (approved):

pyramid/tier_fs.go
arielshaqed: Alas, yes.  The last error handler on l. 264 is nonfatal (not sure why; @itaiad200 can explain) and returned only after being reported to the histograms.  So we need to generate an actual `err` here.

arielshaqed (commented):
Thanks!

Pulling...

PR #1061, opened 2020-12-14
Enhanced multipart upload test by matching uploaded content

### LINKED ISSUE #888: Verify MultipartTest content match
We currently only upload the file and read it without asserting on content mismatch.

Sufiyan1997:
closes #888 

1. Concatenated all parts in a byte slice before uploading
2. After reading the uploaded file compared it with slice obtained in step 1

---
This change is [Reviewable](https://reviewable.io/reviews/treeverse/lakefs/1061)

### DIFF
--- a/nessie/multipart_test.go
+++ b/nessie/multipart_test.go
@@ -2,7 +2,6 @@ package nessie
 
 import (
 	"bytes"
-	"math/rand"
 	"sync"
 	"testing"
 
@@ -17,7 +16,7 @@ import (
 )
 
 const (
-	multipartNumberOfParts = 5
+	multipartNumberOfParts = 20
 	multipartPartSize      = 6 * 1024 * 1024
 )
 
@@ -37,7 +36,7 @@ func TestMultipartUpload(t *testing.T) {
 	parts := make([][]byte, multipartNumberOfParts)
 	var partsConcat []byte
 	for i := 0; i < multipartNumberOfParts; i++ {
-		parts[i] = randstr.Bytes(multipartPartSize + 1 + rand.Intn(19))
+		parts[i] = randstr.Bytes(multipartPartSize + i)
 		partsConcat = append(partsConcat, parts[i]...)
 	}
 
--- a/nessie/multipart_test.go
+++ b/nessie/multipart_test.go
@@ -54,7 +54,7 @@ func TestMultipartUpload(t *testing.T) {
 			WithRef(masterBranch).
 			WithPath(file), nil, &b)
 	require.NoError(t, err, "failed to get object")
-	require.True(t, bytes.Equal(b.Bytes(), partsConcat), "uploaded object did not match")
+	require.Equal(t, b.Bytes(), partsConcat, "uploaded object did not match")
 }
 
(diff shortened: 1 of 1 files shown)


nessie/multipart_test.go
```diff
@@ -47,9 +54,10 @@ func TestMultipartUpload(t *testing.T) {
 			WithRef(masterBranch).
 			WithPath(file), nil, &b)
 	require.NoError(t, err, "failed to get object")
+	require.True(t, bytes.Equal(b.Bytes(), partsConcat), "uploaded object did not match")
```
arielshaqed, line 57: Can you instead use testify's Equal method, it should give better error reporting (and also be shorter)?
```suggestion
	require.Equal(t, b.Bytes(), partsConcat, "uploaded object did not match")
```

nessie/multipart_test.go
```diff
@@ -33,7 +33,14 @@ func TestMultipartUpload(t *testing.T) {
 	require.NoError(t, err, "failed to create multipart upload")
 	logger.Info("Created multipart upload request")
 
-	completedParts := uploadMultipartParts(t, logger, resp)
+	parts := make([][]byte, multipartNumberOfParts)
+	var partsConcat []byte
+	for i := 0; i < multipartNumberOfParts; i++ {
+		parts[i] = randstr.Bytes(multipartPartSize)
```
arielshaqed, line 39: Let's vary the size a bit -- 6 MiB is a really round number, and we might miss some padding nastiness somewhere.

```suggestion
		parts[i] = randstr.Bytes(multipartPartSize + 1 + rand.Intn(19))
```

I know that the previous code _also_ used these round numbers; I just think it will be better, seeing as we're here anyway.

arielshaqed (approved):
This is great, thanks!

Please see two small comments and let me know if you agree they are improvements -- I'll approve any which way, as this is a better test.

nessie/multipart_test.go
Sufiyan1997: Initially, I thought `Equals` will not work on slices (reference types). But in docs I found it does.

nessie/multipart_test.go
Sufiyan1997: Anything special about the number 19 here? 
Also, Won't this introduce randomness in tests? i.e, sometimes it passes and sometimes it does not?

arielshaqed: You're right, 19 is nothing special, and it does introduce randomness.

Regarding randomness: there is indeed a chance of inducing flakiness into the test.  However this is an integration test, flakiness is somewhat expected.  A very reasonable alternative would be to have many more parts (around 20, perhaps 19...) of sizes `multiPartSize + 17 * i` (again, arbitrary numbers).

arielshaqed (approved):
Thanks!

I shall happily pull this code any way.  I would prefer to have different lengths, but that is just me asking for icing on the cake -- you already added the really important bits.  Let me know if you don't want to add the different lengths, and I shall pull.

arielshaqed (approved):
Really great.  Thanks, this is a *major* improvement to our integration test coverage!

PR #1067, opened 2020-12-15
Add simple marshalling for graveler.Value

arielshaqed:
As discussed f2f this allows separating graveler from the underlying trees wrapper, making each
unaware of the others.  Use our own trivial serialization format that allows high-speed
operation with no copying: in particular there is no need to copy data, or to deserialize more
than Identity.

---
This change is [Reviewable](https://reviewable.io/reviews/treeverse/lakefs/1067)

### DIFF
--- a/graveler/value.go
+++ b/graveler/value.go
@@ -1,0 +1,88 @@
+package graveler
+
+import (
+	"bytes"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"io"
+)
+
+/*
+ * Value is serialized in a trivial fixed-order format:
+ *
+ *    | len(Identity) | Identity | len(Value) | Value |
+ *
+ * where each length is serialized as a varint and additional bytes after Value are silently
+ * ignored.
+ */
+
+func writeVarint(w io.Writer, i int) error {
+	e := make([]byte, binary.MaxVarintLen64)
+	l := binary.PutVarint(e, int64(i))
+	_, err := w.Write(e[:l])
+	return err
+}
+
+func writeBytes(tag string, w io.Writer, b []byte) error {
+	if err := writeVarint(w, len(b)); err != nil {
+		return fmt.Errorf("write %s length: %w", tag, err)
+	}
+	if _, err := w.Write(b); err != nil {
+		return fmt.Errorf("write %s: %w", tag, err)
+	}
+	return nil
+}
+
+// MarshalValue returns bytes that uniquely unmarshal into a Value equal to v.
+func MarshalValue(v *Value) ([]byte, error) {
+	var w = &bytes.Buffer{}
+
+	if err := writeBytes("identity", w, v.Identity); err != nil {
+		return nil, err
+	}
+	if err := writeBytes("data", w, v.Data); err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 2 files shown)


graveler/value.go
```diff
@@ -0,0 +1,88 @@
...
+	"io"
+)
+
+/*
+ * Value is serialized in a trivial fixed-order format:
+ *
+ *    | len(Identity) | Identity | len(Value) | Value |
+ *
+ * where each length is serialized as a varint and additional bytes after Value are silently
+ * ignored.
```
ozkatz, line 17: 1. From the `binary` package docs, it seems that Varint is basically the Protobuf implementation of a Varint. This means any Graveler implementation (other than the Go one) will either depend on Protobuf to parse these, or will have to implement the Protobuf implementation of a Varint.

2. Why silently ignore? Why not just assume everything after `Identity` is the value?

graveler/value.go
```diff
@@ -0,0 +1,88 @@
...
+		return fmt.Errorf("write %s: %w", tag, err)
+	}
+	return nil
+}
+
+// MarshalValue returns bytes that uniquely unmarshal into a Value equal to v.
+func MarshalValue(v *Value) ([]byte, error) {
+	var w = &bytes.Buffer{}
+
+	if err := writeBytes("identity", w, v.Identity); err != nil {
```
ozkatz, line 41: Assuming we have exactly 2 tags that are constant, could be better to define them as such?

ozkatz (approved):
LGTM!

graveler/value.go
arielshaqed: Not sure I follow.  But the only purpose of the `tag` argument is to report errors -- obviously we can do it here outside, so removing that.

arielshaqed (commented):
Thanks!

Good points about serialization; I think it's a matter of different plans for how we shall do backwards compatibility.

I'm going to pull this as soon as tests pass (I think that's what you intended by approving :-) ).  We are not releasing any graveler right now that will start writing files that require backwards compatibility in future versions.  So if we continue to disagree about the serialization points we can still change it, at least in the coming week.

graveler/value.go
arielshaqed: 1. `Varint` is used also in Google protobufs.  I feel safe using it, given that the implementation is entirely self-contained and is [just a few short routines](https://golang.org/src/encoding/binary/varint.go) that bang bits together.  Would you feel safer if I used Uvarint, which is even shorter?
2. I insist on encoding `Value` in order to get extensibility.  Otherwise in future I will not be able to add some third field needed by graveler.  "But why not encode that putative future extra field inside `Value`, and have graveler return the inner `Value.ActualOldValue` field?" Well, then for backwards compatibility (new code reading old files) I would need either file-level metadata (yuck!) or to rely on some known property of the old values as actually specified by the code calling graveler.  And it still wouldn't work for old code reading new files (e.g. a dump routine that prints `"[key] -> [value]\n"`.  Doing it this way provides compatibility the protobuf way: it is always safe to add optional ignorable fields.

arielshaqed (commented):
(Missed sending this comment)

PR #1068, opened 2020-12-16
[post-pull CR] Marshal directly into buffer without using a Writer

arielshaqed:
This change is [Reviewable](https://reviewable.io/reviews/treeverse/lakefs/1068)

### DIFF
--- a/graveler/value.go
+++ b/graveler/value.go
@@ -1,11 +1,9 @@
 package graveler
 
 import (
-	"bytes"
 	"encoding/binary"
 	"errors"
 	"fmt"
-	"io"
 )
 
 /*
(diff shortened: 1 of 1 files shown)


graveler/value.go
```diff
@@ -17,34 +15,23 @@ import (
...
-	}
-	if err := writeBytes("data", w, v.Data); err != nil {
-		return nil, err
-	}
-	return w.Bytes(), nil
+	ret := make([]byte, 0, len(v.Identity)+len(v.Data)+2*binary.MaxVarintLen32)
+	putBytes(&ret, v.Identity)
+	putBytes(&ret, v.Data)
+	return ret, nil
 }
```
nopcoder, line 35: Suggesting the following, less pointers
```
// MarshalValue returns bytes that uniquely unmarshal into a Value equal to v.
func MarshalValue(v *Value) ([]byte, error) {
	ret := make([]byte, len(v.Identity)+len(v.Data)+2*binary.MaxVarintLen32)
	l := marshalVarintData(ret, v.Identity)
	l += marshalVarintData(ret[l:], v.Data)
	return ret[:l], nil
}

func marshalVarintData(out []byte, data []byte) int {
	dataLen := len(data)
	pos := binary.PutVarint(out, int64(dataLen))
	copy(out[pos:], data)
	pos += dataLen
	return pos
}```

Comment (arielshaqed) [reply graveler/value.go:35]: I'd rather not.  The code with fewer pointers looks much nicer but is brittle: the caller has to allocate a buffer large enough to hold the result.  Most of the ugliness of the existing code comes from having to call `append()`.  I think it is safer, hence worth the ugliness.  (It is surprisingly difficult to catch an error in the length calculation during `ret := make([]byte, ...)`: e.g. if you add a third value and say
```go
	ret := make([]byte, len(v.Identity)+len(v.Data)+len(v.Another)+2*binary.MaxVarintLen32)
```
then it will actually _work_ for all tests that use less than at least 1 GiB of data -- but it is still wrong.  I prefer the uglier code here :shrug: .

nopcoder (approved):

PR #1070, opened 2020-12-16
sstable layer to serve key/values without identity

itaiad200:
This change is [Reviewable](https://reviewable.io/reviews/treeverse/lakefs/1070)


graveler/committed/sstable/iterator.go
```diff
@@ -49,18 +45,17 @@ func (iter *Iterator) Next() bool {
...
 	}
 
-	iter.valParsed, iter.err = iter.ser.DeserializeValue(iter.currValue)
-	return iter.err == nil
+	return true
 }
 
-func (iter *Iterator) Value() *graveler.ValueRecord {
+func (iter *Iterator) Value() *committed.Record {
 	if iter.currKey == nil || iter.err != nil || iter.postSeek {
```
tzahij, line 52: if next() was not called after SeekGE - you return nil. It indicates a bug.  Maybe an error should be posted as well

tzahij (approved):
Looks good.
There is a general bothering issue about Pebble.  I suspect that when you get a key anv value from Next(),SeekGE etc, it is a pointer to internal structures in Pebble buffers. I am afraid it may change once you issue more commands on the iterator.
the reason for my suspicion is the behavior of [the Pebble Get opetation](https://godoc.org/github.com/cockroachdb/pebble#DB.Get). It indicates possibilities of memory leaks if you do not call the returned Closer() function.
Pebble is much more complex, and so may have different requirements.  On the other hans - sstables is a less public facing interface, and the documentation may be missing.

arielshaqed (approved):
Cool, thanks!

graveler/committed/sstable/iterator.go
itaiad200: I follow the interface as we decided upon it and documented. If we want to change the behaviour we need to do so for all iterators, not just here.

itaiad200:
> Looks good.
> There is a general bothering issue about Pebble. I suspect that when you get a key and value from Next(),SeekGE etc, it is a pointer to internal structures in Pebble\ sstables buffers. I am afraid it may change once you issue more commands on the iterator.
> the reason for my suspicion is the behavior of [the Pebble Get operation](https://godoc.org/github.com/cockroachdb/pebble#DB.Get). It indicates possibilities of memory leaks if you do not call the returned Closer() function.
> Pebble is much more complex, and so may have different requirements. On the other hand - sstables is a less public facing interface, and the documentation may be missing.

Good points - Reading on it now.

PR #1071, opened 2020-12-16
Generate the mocks using go generate

itaiad200:
This change is [Reviewable](https://reviewable.io/reviews/treeverse/lakefs/1071)


Makefile
```diff
@@ -71,6 +72,9 @@ gen-api: $(GOBINPATH)/swagger ## Run the go-swagger code generator
 	$(GOBINPATH)/swagger generate client -q -A lakefs -f ./swagger.yml -P models.User -t $(API_BUILD_DIR)
 	$(GOBINPATH)/swagger generate server -q -A lakefs -f ./swagger.yml -P models.User -t $(API_BUILD_DIR) --exclude-main
```
arielshaqed, line 73: I think this will now happen automagically when you `go generate` because of the top of `api/handler.go`.  Best not leave _both_, they will just drift apart and debugging that will hurt.

arielshaqed (changes requested):
1. Building Docker image fails.
2. Double Swagger generation worries me.

arielshaqed (approved):
Nice! Thanks, much better to auto generate everything possible.

PR #1085, opened 2020-12-19
Feature/import filter by prefix

johnnyaug:


block/s3/inventory.go
```diff
@@ -125,5 +127,43 @@ func sortManifest(m *Manifest, logger logging.Logger, reader s3inventory.IReader
...
 	}
+	filteredFiles := make([]inventoryFile, 0, len(m.Files))
+	if len(prefixes) == 0 {
+		return nil
+	}
+	// filter manifest files according to prefixes:
+	sort.Strings(prefixes)
+	currentPrefixIdx := 0
+	defer func() {
+		m.Files = filteredFiles
```
ozkatz, line 138: Mutating state in a `defer`ed function makes it harder to follow and debug IMO. it's basically a GOTO where the logical flow of the program jumps to a different location...

block/s3/inventory.go
```diff
@@ -125,5 +127,43 @@ func sortManifest(m *Manifest, logger logging.Logger, reader s3inventory.IReader
 			return ErrInventoryFilesRangesOverlap
 		}
 	}
+	filteredFiles := make([]inventoryFile, 0, len(m.Files))
```
ozkatz, line 130: I think the filtration should happen in a seperate function that takes a slice of inventory files (i.e. `m.Files`) and returns a filtered list. This way it's much easier to unit test (this, IMO, tends to be a common case for functions that contain the word `And` :) ).

cmd/lakefs/cmd/import.go
```diff
@@ -44,7 +46,9 @@ var importCmd = &cobra.Command{
 		dryRun, _ := flags.GetBool(DryRunFlagName)
 		manifestURL, _ := flags.GetString(ManifestURLFlagName)
 		withMerge, _ := flags.GetBool(WithMergeFlagName)
-		hideProgress, _ := flags.GetBool(HideProgress)
+		hideProgress, _ := flags.GetBool(HideProgressFlagName)
+		prefixFile, _ := flags.GetString(PrefixesFileFlagName)
```
ozkatz, line 50: I think maybe passing in multiple `--prefix` flags would be easier? It's possible using [pflag.StringArray](https://pkg.go.dev/github.com/spf13/pflag#StringArray)

onboard/import.go
```diff
@@ -62,7 +67,12 @@ func CreateImporter(ctx context.Context, logger logging.Logger, config *Config)
 		return nil, fmt.Errorf("failed to get previous commit: %w", err)
 	}
 	res.previousCommit = previousCommit
-	res.inventory, err = config.InventoryGenerator.GenerateInventory(ctx, logger, config.InventoryURL, res.previousCommit != nil)
+	shouldSort := res.previousCommit != nil
+	res.inventory, err = config.InventoryGenerator.GenerateInventory(ctx, logger, config.InventoryURL, shouldSort, config.KeyPrefixes)
+	res.prefixes = config.KeyPrefixes
+	if !res.validateSamePrefixes() {
```
ozkatz, line 73: why do we care?

ozkatz (changes requested):
Nice work - love the tests! The only thing I'm not crazy about is the use of `defer` to change state in the manifest sort and filter method

cmd/lakefs/cmd/import.go
johnnyaug: It's possible, just that it can potentially become a huge flag, in an already bloated command

onboard/import.go
johnnyaug: synced f2f: will validate that the new set of prefixes contains the previous one

ozkatz (approved):
Looks great!

PR #1086, opened 2020-12-19
pyramid main test

nopcoder:
- move TestMain to a main_test.go
- remove 3rd party package for sync counter
- delete temporary directory on test code
- ensure parent dir exists without ceilingDir


pyramid/directory.go
```diff
@@ -78,9 +74,20 @@ func (d *directory) renameFile(src, dst string) error {
...
+	if err := d.ensureParentDir(dst); err != nil {
+		return err
 	}
-
 	return os.Rename(src, dst)
 }
+
+func (d *directory) ensureParentDir(path string) error {
+	parentDir := filepath.Dir(path)
+	if parentDir == d.ceilingDir {
```
itaiad200, line 85: Why this check? MkdirAll returns nil if path exists.

pyramid/directory.go
```diff
@@ -27,23 +27,19 @@ func (d *directory) deleteDirRecIfEmpty(dir string) error {
...
-			if errors.Is(err, os.ErrNotExist) {
-				return nil
-			}
-			return err
-		}
-		if !empty {
+		if errors.Is(err, os.ErrNotExist) {
 			return nil
 		}
+		if err != nil || !empty {
```
itaiad200, line 33: It's not that it's not valid, but it's somewhat hard to understand. Maybe simplify with:
```suggestion
		if err != nil {
		   return err
		 }
		 if !empty {
		   return nil
		 }
```

pyramid/directory.go
nopcoder: Thought this check was required in order not to create the root folder, the following comment:
// ceilingDir is the root directory of the FS - shouldn't never be deleted

pyramid/tier_fs.go
```diff
@@ -62,17 +62,17 @@ const workspaceDir = "workspace"
 // It will traverse the existing local folders and will update
 // the local disk cache to reflect existing files.
 func NewFS(c *Config) (FS, error) {
-	fsLocalBaseDir := path.Join(c.localBaseDir, c.fsName)
-	if err := os.MkdirAll(fsLocalBaseDir, os.ModePerm); err != nil {
-		return nil, fmt.Errorf("creating base dir: %w", err)
+	baseDir := filepath.Clean(path.Join(c.localBaseDir, c.fsName))
```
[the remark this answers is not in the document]
nopcoder: A local variable which usually I keep names short and simple. Probably did it while reviewing the code - will revert.

pyramid/directory.go
nopcoder: that was the previous version I had :) thought the shorter version made mode sense

nopcoder:
need to discuss if we need to testify which comes with mocks and mockgen, and in general, if we need it at the first place

pyramid/directory.go
itaiad200: MkdirAll is a no-op for existing dirs. IIUC you're saying that you want to expose a possible bug of deleting the ceilingDir by making sure we're not ever creating it? ok then.

pyramid/eviction.go
```diff
@@ -59,5 +59,7 @@ func (re *ristrettoEviction) store(rPath relativePath, filesize int64) bool {
 }
 
 func (re *ristrettoEviction) onEvict(item *ristretto.Item) {
-	re.evictCallback(item.Value.(relativePath), item.Cost)
+	if item.Value != nil {
```
itaiad200, line 62: `ristrettoEviction` is coupled with the implementation of `TierFS`. I would much rather check for nil `onEvict` callback during creation, which isn't really allowed, than checking for nil here.

pyramid/pyramid.go
```diff
@@ -1,5 +1,7 @@
 package pyramid
 
+//go:generate mockgen -source=pyramid.go -destination=mock/pyramid.go -package=mock
```
itaiad200, line 3: Great!

itaiad200 (approved):

pyramid/directory.go
nopcoder: exactly

pyramid/pyramid.go
nopcoder: Thanks I had to change the FS interface in order to make the mockgen to work

PR #1087, opened 2020-12-20
Cache SSTables

arielshaqed:
Hints for review: The main changes are in `cache.go`.  The rest is mostly noise to move tests from `sstable` to `sstable_test` in order to avoid an import cycle when mocking.


graveler/committed/sstable/cache.go
```diff
@@ -1,14 +1,103 @@
...
+
+func (i *item) GetSSTable() *sstable.Reader {
+	return i.r
+}
+
+func (i *item) Close() error {
+	return i.r.Close()
+}
+
+func NewCache(p lru.ParamsWithDisposal, fs pyramid.FS, readerOptions sstable.ReaderOptions) cache {
```
itaiad200, line 46: Not sure why the `ParamsWithDisposal` are passed here. Is it a smart way of passing the underline cache configuration? 
Why do we need another dispose hook for the layer that calls the cache?

graveler/committed/sstable/cache.go
```diff
@@ -1,14 +1,103 @@
...
+
+		reader, err := sstable.NewReader(file, readerOptions)
+		if err != nil {
+			return nil, fmt.Errorf("open SSTable %s: %w", id, err)
+		}
+		return &item{reader}, nil
+	})
+}
+
+func NewCacheWithOpener(p lru.ParamsWithDisposal, open opener) cache {
```
itaiad200, line 61: If it's just for a test hook, maybe make it internal?

graveler/committed/sstable/cache.go
```diff
@@ -1,14 +1,103 @@
...
+	wrappedOnDispose := p.OnDispose
+	p.OnDispose = func(v interface{}) error {
```
itaiad200, line 63: I think this can be a lot simpler if we remove the option for the passed hook, which I don't think is needed.

itaiad200 (changes requested):
Looks really good!
Requested changes for possibly making things simpler if removing the hook.

cache/refcnted.go
```diff
@@ -38,6 +38,7 @@ type ParamsWithDisposal struct {
 // keyus have a String() method that is compatible with their equality, i.e. k1.GoString() ==
 // k2.GoString() implies k1 equals k2 in the sense of map comparison.
 type shardedCacheWithDisposal struct {
+	name   string
```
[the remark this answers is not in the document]
arielshaqed: Bug, I don't.  Adding it...

graveler/committed/sstable/cache.go
arielshaqed: Not sure what you mean, you're way ahead of me.  It's a *stupid* way of giving the configuration parameters.  Anything on ParamsWithDisposal is fair game, including 	Name, Logger, Size, Shards, which I think the caller will have to set (probably from configuration).

graveler/committed/sstable/cache.go
arielshaqed: We split up packages to separate test and non-test.  @nopcoder says to test at the caller level.

Also this avoids import cycles with mocks.  Otherwise, package mock depends on package x-for-code, and package x-for-test depends on package mock.  So if x-for-test is the same as x-for-code... we get an import cycle.

graveler/committed/sstable/cache.go
arielshaqed: I think we'll end up needing it and "it's just code" is the Go motto, but since we don't right now I'm removing it.

graveler/committed/sstable/mock/cache.go
```diff
@@ -0,0 +1,52 @@
+// Code generated by MockGen. DO NOT EDIT.
```
arielshaqed: Getting rid of this code; bad gitignore at some stage I guess.

arielshaqed (commented):
Thanks; PTAL...

johnnyaug (commented):
Can we rename refcnted.go to be clearer?

itaiad200 (approved):
LGTM

arielshaqed:

arielshaqed:
# Hints for improving performance...

If you're here, you've noticed that CockroachDB Pebble SSTables has disappointing performance.  [sst_bench](https://github.com/treeverse/sst_bench) is a useful tool for benchmarking write and especially read performance.

The *worst* scalability for multithreaded reading comes from what this cache implements: single reader shared between goroutines.  This ends up with all goroutines contending to use the same (read-only, immutable!) data.  You can improve that ~ x1.2 by giving each goroutine its own reader, and another ~ x1.2 by giving each goroutine its own reader with its own cache.  (You still need to cache no matter what you do, otherwise you pay the price of opening a file and deserializing metadata for each access, which is _really_ bad!)

At this point you'll be asking yourself why the pebbles SSTable implementation doesn't support the RocksDB "plain" file format, which removes the need for caching (files are anyway essentially mmap()ed in) and almost all locking (just read locking, to avoid releasing the backing memory).  So one way out of the hole is to use RocksDB sstables.

Hopefully we will never have performance issues on this :-)

PR #1088, opened 2020-12-20
Add Iterator to Tree

arielshaqed:
Iterates over all values in a tree, putting a "new part" header as an explicit iterated value.
Makes it easier to write Apply (and hopefully Diff) easier.


arielshaqed:
Please have a look whether you think this is The Right Way :tm: (or at least A Right Way) to write an abstraction to help us write Apply and Diff.

graveler/committed/tree/apply.go
```diff
@@ -0,0 +1,91 @@
...
+			if ok := source.Next(); !ok {
+				sourceDone = true
+			}
+		}
+	}
+	for !diffsDone {
+		diffValue := diffs.Value()
+		if err := writer.WriteRecord(*diffValue); err != nil {
+			return fmt.Errorf("write added record: %w", err)
+		}
```
guy-har, line 85: Maybe we should check for tombstone here  (even though it shouldn't happen)

graveler/committed/tree/tree.go
```diff
@@ -19,6 +19,22 @@ type Tree struct {
 	Parts []Part
 }
 
+// PartsAndValuesIterator iterates over all part headers and values of a tree, allowing
+// seeking by entire parts.
+type PartsAndValuesIterator interface {
+	// Next moves to iterator to look at the next value in the current part, or a header
+	// for the next part if the current part is over.
```
guy-har, line 26: moves the Iterator

graveler/committed/tree/parts_and_values.go
```diff
@@ -0,0 +1,78 @@
...
+		return false
+	}
+	pvi.parts = pvi.parts[1:]
+	pvi.it = nil
+	return true
+}
+
+func (pvi *partsAndValuesIterator) Next() bool {
+	if pvi.it != nil {
+		ok := pvi.it.Next()
```
johnnyaug, line 27: I like `if pvi.it.Next()` - but matter of personal taste, I guess

guy-har (commented):
Looks Good, I like the new iterator

johnnyaug (commented):
Looks very good

graveler/committed/tree/parts_and_values.go
```diff
@@ -0,0 +1,78 @@
+package tree
+
+import "github.com/treeverse/lakefs/graveler"
+
+type partsAndValuesIterator struct {
```
nopcoder, line 5: Suggestion shorter name - TreeValueIterator

graveler/committed/tree/parts_and_values.go
```diff
@@ -0,0 +1,78 @@
...
+	repo  Repo
+	parts []Part
+	it    graveler.ValueIterator
+	err   error
+}
```
nopcoder, line 10: This iterator also behaves a bit differently as the expected behavior from our current Iterators / Scanners are to call Next, on true you can access Value to get the current iterator value.

nopcoder (approved):
love it, think we need to align our iterators/scanners in the Graveler project and we will be fine 👍

arielshaqed (commented):
Removing code for `Apply` to pull just the iterator.  `Apply` will come in the next PR.  Meanwhile, this is no longer a draft, so please request changes if you don't want this iterator.

graveler/committed/tree/parts_and_values.go
johnnyaug: I suggest: TreeIterator for this one,
and the other iterator (which only returns values) will be TreeValueIterator

johnnyaug (approved):

guy-har (approved):

graveler/committed/tree/parts_and_values.go
johnnyaug: And since we're under the tree -pun not intended- package, how about just Iterator.

arielshaqed:
Thanks for reviews!

System test failed because of some DNS lookup failure flakiness (attn @ozkatz ...), so merging by admin force -- this code is not yet on any path exercised by the system test!

PR #1092, opened 2020-12-21
Move graveler listing prefix and delimiter to catalog

nopcoder:


catalog/rocks/entry_listing_iterator.go
```diff
@@ -1,67 +1,110 @@
...
+	nextFunc  func() bool
+	value     *EntryListing
 }
 
-func NewEntryListingIterator(it graveler.ListingIterator) EntryListingIterator {
-	return &entryListingIterator{
-		it: it,
+// getListingNextPath returns the following value (i.e will increase the last byte by 1)
+// in the following cases will return received value: empty value, the last byte is math.MaxUint8
+func getListingNextPath(value Path) Path {
```
guy-har, line 18: Check if we have same logic under diff

guy-har (approved):
LGTM

catalog/rocks/entry_listing_iterator.go
nopcoder: good idea

PR #1093, opened 2020-12-22
Implement (and test) tree Apply

arielshaqed:


graveler/committed/tree/apply.go
```diff
@@ -0,0 +1,88 @@
...
+		if diffValue.IsTombstone() {
+			// internal error but no data lost: deletion requested of a
+			// file that was not there.
+			logger.WithField("id", string(diffValue.Identity)).Warn("[I] unmatched delete")
+			continue
+		}
+		if err := writer.WriteRecord(*diffValue); err != nil {
+			return fmt.Errorf("write added record: %w", err)
+		}
+	}
```
nopcoder, line 86:
```suggestion
	for ; haveDiffs; haveDiffs = diffs.Next() {
		diffValue := diffs.Value()
		if diffValue.IsTombstone() {
			// internal error but no data lost: deletion requested of a
			// file that was not there.
			logger.WithField("id", string(diffValue.Identity)).Warn("[I] unmatched delete")
		} else if err := writer.WriteRecord(*diffValue); err != nil {
			return fmt.Errorf("write added record: %w", err)
		}
	}
```

arielshaqed: Thanks, this would be a great change on its own, and I know I was for it when we talked f2f.  But I'm _not_ going to do it, to keep the code more like the (similar but more complicated) `for haveSource` loop just above it.

guy-har (approved):
LGTM

arielshaqed:
Thanks!

PR #1097, opened 2020-12-22
Cache node_modules between runs

arielshaqed:
See
[the manual](https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows)
for details of how the GitHub actions cache works.

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -56,10 +56,13 @@ jobs:
         with:
           go-version: 1.15.2
         id: go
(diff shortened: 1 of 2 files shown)


arielshaqed:
Note that it _doesn't_ do much for nessie: that builds a new version inside a container, so cannot enjoy the existing cache.

@itaiad200 why does nessie build inside a container?

nopcoder (approved):

PR #1098, opened 2020-12-22
Enable rocks cataloger listing and remove 'from' argument when possible

nopcoder:


catalog/rocks/cataloger.go
```diff
@@ -83,7 +91,46 @@ func (c *cataloger) DeleteRepository(ctx context.Context, repository string) err
...
+
+	var repos []*catalog.Repository
+	for it.Next() {
+		record := it.Value()
+		repos = append(repos, &catalog.Repository{
+			Name:             record.RepositoryID.String(),
+			StorageNamespace: record.StorageNamespace.String(),
+			DefaultBranch:    record.DefaultBranchID.String(),
+			CreationDate:     record.CreationDate,
+		})
```
guy-har, line 118: Check that you don't return "after" repository

catalog/rocks/cataloger.go
```diff
@@ -132,7 +179,58 @@ func (c *cataloger) DeleteBranch(ctx context.Context, repository string, branch
...
+	}
+	it, err := c.EntryCatalog.ListBranches(ctx, repositoryID)
+	if err != nil {
+		return nil, false, err
+	}
+	if afterBranch == "" {
+		it.SeekGE(prefixBranch)
+	} else {
+		it.SeekGE(afterBranch)
+	}
```
guy-har, line 206: what if after < prefix ?

catalog/rocks/cataloger.go
```diff
@@ -378,8 +566,51 @@ func (c *cataloger) GetCommit(ctx context.Context, repository string, reference
 	return catalogCommitLog, nil
 }
 
-func (c *cataloger) ListCommits(ctx context.Context, repository string, branch string, fromReference string, limit int) ([]*catalog.CommitLog, bool, error) {
-	panic("not implemented") // TODO: Implement
+func (c *cataloger) ListCommits(ctx context.Context, repository string, _ string, fromReference string, limit int) ([]*catalog.CommitLog, bool, error) {
+	repositoryID, err := graveler.NewRepositoryID(repository)
+	if err != nil {
```
guy-har, line 571: We should have here two kind of "froms", one is the Reference the Log is built out of and the second is for the pagination.

guy-har (changes requested):
Looks good, 
I think we should add tests (at least for the paginators)

graveler/testutil/fakes.go
```diff
@@ -194,7 +194,7 @@ func (m *RefsFake) DeleteBranch(_ context.Context, _ graveler.RepositoryID, _ gr
 	return nil
 }
 
-func (m *RefsFake) ListBranches(_ context.Context, _ graveler.RepositoryID, _ graveler.BranchID) (graveler.BranchIterator, error) {
+func (m *RefsFake) ListBranches(ctx context.Context, repositoryID graveler.RepositoryID) (graveler.BranchIterator, error) {
 	return m.ListBranchesRes, nil
```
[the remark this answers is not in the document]
nopcoder: Remember you told me to use the refactor tool to update the signature from the interface - so this is it :)
It updates the implementation with the method signature - fixed.

catalog/rocks/cataloger.go
nopcoder: My implementation should consider it as a miss and return no items. I can check if after < from the prefix and use prefix as you suggest I agree it will have the more expected behavior.

catalog/rocks/cataloger.go
nopcoder: resolved the branch and iterated until from ref if passed any value

catalog/rocks/cataloger.go
nopcoder: skipped any value that equal to 'after' - let me know if you liked the implementation I have another, not sure which one is cleaner.

catalog/rocks/cataloger.go
```diff
@@ -179,50 +182,43 @@ func (c *cataloger) DeleteBranch(ctx context.Context, repository string, branch
...
 	}
-	it, err := c.EntryCatalog.ListBranches(ctx, repositoryID)
+	it, err := c.EntryCatalog.ListBranches(ctx, graveler.RepositoryID(repository))
 	if err != nil {
 		return nil, false, err
 	}
-	if afterBranch == "" {
+	afterBranch := graveler.BranchID(after)
+	prefixBranch := graveler.BranchID(prefix)
+	if afterBranch == "" || afterBranch < prefixBranch {
```
guy-har, line 195: afterBrunch < prefixBrunch is enough

guy-har (approved):
👍

PR #1101, opened 2020-12-24
Read ristretto cleanup during test that fails it

itaiad200:


nopcoder:
Suggestion - the default FS uses the ristreto cache which as I understand you want to abstract for testing.
You already have an evict interface which we already implemented by ristreto cache.
Instead of creating a function callback for factory of this type, make the interface public and accept it as part of the config.
Do not require it.
In case it is nil - create new restreto cache.
This way we have fewer callbacks, less code and the user of the FS doesn't need to care about it unless it is part of his/her agenda to override the default evict implementation.

nopcoder (changes requested):
let me know what you think about my suggestion to simplify the use of FS

itaiad200:
@nopcoder Agreed, changed the FS.

nopcoder (approved):

PR #1104, opened 2020-12-27
graveler ref package implementation

nopcoder:


graveler/ref/branch_iterator_test.go
```diff
@@ -0,0 +1,97 @@
...
+		DefaultBranchID:  "master",
+	}, graveler.Branch{}))
+
+	// prepare data
+	for _, b := range branches {
+		testutil.Must(t, r.SetBranch(ctx, "repo1", b, graveler.Branch{CommitID: "c1"}))
+	}
+
+	t.Run("listing all branches", func(t *testing.T) {
+		iter := ref.NewBranchIterator(ctx, db, "repo1", 3)
```
arielshaqed, line 30: 3 divides 6, so this doesn't test partial batching.  Can you add another branch on l.16, or change "3" to "4" or "5", please?

graveler/ref/branch_iterator_test.go
```diff
@@ -0,0 +1,97 @@
...
+		for iter.Next() {
+			b := iter.Value()
+			ids = append(ids, b.BranchID)
+		}
+		if iter.Err() != nil {
+			t.Fatalf("unexpected error: %v", iter.Err())
+		}
+		iter.Close()
+
+		if diffs := deep.Equal(ids, []graveler.BranchID{"a", "aa", "b", "c", "d", "e", "master"}); diffs != nil {
```
arielshaqed, line 41: Does the interface guarantee branches are returned in sorted order?

graveler/ref/branch_iterator_test.go
```diff
@@ -0,0 +1,97 @@
...
+			t.Fatalf("got wrong list of IDs: %v", diffs)
```
arielshaqed, line 42: nit:
```suggestion
			t.Errorf("got wrong list of IDs: %v", diffs)
```

graveler/ref/branch_iterator_test.go
```diff
@@ -0,0 +1,97 @@
...
+		}
+	})
+
+	t.Run("listing branches from prefix", func(t *testing.T) {
```
arielshaqed, line 46: Not sure I understand the name (but the test is good and worthwhile!)

graveler/ref/branch_iterator_test.go
```diff
@@ -0,0 +1,97 @@
...
+			t.Fatalf("unexpected error: %v", iter.Err())
+		}
+		iter.Close()
+
+		if diffs := deep.Equal(ids, []graveler.BranchID{"b", "c", "d", "e", "master"}); diffs != nil {
+			t.Fatalf("got wrong list of branch IDs: %v", diffs)
+		}
+
+		// now let's seek
+		iter.SeekGE("aa")
```
arielshaqed, line 82: This does not crash, which is a bug: future iterators _will_.  Let's make this iterator panic on user-after-Close?

graveler/ref/branch_iterator_test.go
```diff
@@ -0,0 +1,97 @@
...
+		iter.SeekGE("b")
+		ids := make([]graveler.BranchID, 0)
+		for iter.Next() {
+			b := iter.Value()
+			ids = append(ids, b.BranchID)
+		}
+		if iter.Err() != nil {
+			t.Fatalf("unexpected error: %v", iter.Err())
+		}
+		iter.Close()
```
arielshaqed, line 75:
```suggestion
```
(see also next comment)

graveler/ref/repository_iterator_test.go
```diff
@@ -0,0 +1,96 @@
...
+		iter.SeekGE("b")
+		repoIds := make([]graveler.RepositoryID, 0)
+		for iter.Next() {
+			repo := iter.Value()
+			repoIds = append(repoIds, repo.RepositoryID)
+		}
+		if iter.Err() != nil {
+			t.Fatalf("unexpected error: %v", iter.Err())
+		}
+		iter.Close()
```
arielshaqed, line 73: Again use-after-Close works, because the PG-based implementation doesn't care.

arielshaqed (changes requested):
Thanks!

Blocking is that use-after-Close works (and shouldn't), and that tests do that by mistake (which they shouldn't).  Everything else can be skipped or done later.

graveler/ref/branch_iterator.go
```diff
@@ -0,0 +1,122 @@
...
+	value        *graveler.BranchRecord
+	buf          []*graveler.BranchRecord
+
+	// blockValue is true when the iterator is created, or SeekGE() is called.
+	// A single Next() call turns it to false. When it's true, Value() returns nil.
+	blockValue bool
+
+	offset      string
+	fetchSize   int
+	shouldFetch bool
```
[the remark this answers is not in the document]
nopcoder: can document it - 'shouldFetch' is true as an indicator for the initial state. on new iterator or after Seek, we shouldFetch is true, until we reach a point the last query doesn't fill our buffer.

graveler/ref/commit_walkter.go
```diff
@@ -1,36 +1,35 @@
-package graveler
```
johnnyaug, line 1: typo in filename

graveler/ref/commit_walkter_test.go
```diff
@@ -1,10 +1,11 @@
-package graveler_test
```
johnnyaug, line 1: typo in filename

johnnyaug (approved):
Cool!

graveler/ref/branch_iterator.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+}
+
+func (ri *BranchIterator) Next() bool {
+	ri.blockValue = false
+
+	// no buffer is initialized
+	if ri.buf == nil {
+		ri.fetch(true) // initial fetch
+	} else if len(ri.buf) == 0 {
```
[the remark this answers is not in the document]
nopcoder: I see your point - I'll rewrite this part using two booleans. Fetch called and fetch ended.

graveler/ref/branch_iterator_test.go
nopcoder: For every iterator that we have SeekGE - the key is ordered

graveler/ref/branch_iterator_test.go
nopcoder: why? just wonder as in this test there no other check after this one

graveler/ref/branch_iterator_test.go
nopcoder: changed to "using prefix"

graveler/ref/branch_iterator.go
```diff
@@ -8,22 +8,17 @@ import (
...
-	fetchSize   int
-	shouldFetch bool
-
-	err error
+	offset       string
+	fetchSize    int
+	err          error
+	fetchCalled  bool
+	fetchEnded   bool
+	closed       bool
```
arielshaqed, line 21: Thanks, good names!  Now I see and understand: _it's a state machine_!  Not in this PR, but I would like to try refactor it to have one explicit state.  Think it's worth a shot?

graveler/ref/branch_iterator.go
```diff
@@ -38,43 +33,44 @@ func NewBranchIterator(ctx context.Context, db db.Database, repositoryID gravele
...
+	if len(ri.buf) == 0 {
+		return false
+	}
 	ri.value = ri.buf[0]
 	ri.offset = string(ri.value.BranchID)
 	if len(ri.buf) > 1 {
 		ri.buf = ri.buf[1:]
 	} else {
-		ri.buf = make([]*graveler.BranchRecord, 0)
+		ri.buf = ri.buf[:0]
```
arielshaqed, line 58: # :+1:

graveler/ref/iterator.go
```diff
@@ -1,10 +1,14 @@
 package ref
 
+import "errors"
+
 const (
 	iteratorOffsetGE = ">="
 	iteratorOffsetGT = ">"
 )
 
+var ErrIteratorClosed = errors.New("iterator already closed")
```
arielshaqed, line 10: All of these should be the same error.

Also: what is The Go Dan to keep these changes sync'ed?  I'm reviewing these by reading through them, but this is a *horrible* way to do it.  Surely we're not the first people to run into this...

graveler/ref/repository_iterator_test.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/treeverse/lakefs/testutil"
 )
 
-func TestPGRepositoryIterator(t *testing.T) {
+func TestRepositoryIterator(t *testing.T) {
 	r, db := testRefManagerWithDB(t)
 	repos := []graveler.RepositoryID{"a", "aa", "b", "c", "e", "d"}
```
arielshaqed, line 16: 6 is still divisible by 3, I'd rather we also tested a partial fetch.

graveler/ref/branch_iterator_test.go
```diff
@@ -11,7 +11,7 @@ import (
 	"github.com/treeverse/lakefs/testutil"
 )
 
-func TestPGBranchIterator(t *testing.T) {
+func TestBranchIterator(t *testing.T) {
 	r, db := testRefManagerWithDB(t)
 	branches := []graveler.BranchID{"a", "aa", "b", "c", "e", "d"}
```
arielshaqed, line 16: 6 divisible by 3; I think this is unchanged from the previous review where it means that we end up not testing partial fetches (we should do both).  Or, perhaps run the "listing all branches" test in a loop for _all_ batch sizes 1..7?

arielshaqed (approved):
Great, thanks!
Some comments inside, nothing blocking.

graveler/ref/branch_iterator_test.go
nopcoder: There are 6 branches here + master = so we have 7. I'll make sure tag and repo will have the same.

graveler/ref/branch_iterator_test.go
nopcoder: keeping branch iterator working on 7 items.

graveler/ref/commit_walkter.go
nopcoder: Thanks!

graveler/ref/repository_iterator_test.go
nopcoder: fixed

graveler/ref/iterator.go
nopcoder: https://github.com/googleapis/google-api-go-client/blob/v0.36.0/iterator/iterator.go

can modify the code not to panic - just report an error. can condition the SeekGE not to reset the error just for this case / state.

What do you think?

arielshaqed: Yeah.  Probably not in this PR, though :-)

graveler/ref/branch_iterator_test.go
arielshaqed: Ouch!  A rich corporation once said, "[math is hard](https://whyplay.mattel.com/the-barbie-vlog-empowers-girls-with-words/)" -- and they're still not sorry!

arielshaqed (approved):
Thanks!

graveler/ref/iterator.go
nopcoder: Don't stop - we will get it better next round

PR #1105, opened 2020-12-27
added validation for storage namespace to check for compatibility with configured block store type

### LINKED ISSUE #494: Raise validation error when trying to create a repository with incompatible storage type
lakeFS can be started with different storage types.

This is configured using the `blockstore.type` configuration (currently supported are s3, local, mem and transient).

When creating a repository, a *storage namespace* is specified. This determines where the data will be stored. The storage namespace should be compatible with the selected blockstore type.

For example "s3://my-bucket/" is a valid storage namespace for blockstore type "s3", but invalid for all other blockstore types. Similarly, "local:///home/user/dir/" is valid only for "local" blockstore type.

Need to add a validation for this rule when creating a repository.

Sufiyan1997:
closes #494 

 - While creating a repository, added validation to check if the provided storage namespace is compatible with configured block store type
 - Added same validation on backend

### DIFF
--- a/catalog/mvcc/validate.go
+++ b/catalog/mvcc/validate.go
@@ -3,7 +3,9 @@ package mvcc
 import (
 	"fmt"
 	"regexp"
+	"strings"
 
+	"github.com/spf13/viper"
 	"github.com/treeverse/lakefs/catalog"
 )
 
@@ -100,7 +102,11 @@ func ValidateCommitter(name string) ValidateFunc {
 
 func ValidateStorageNamespace(storageNamespace string) ValidateFunc {
 	return func() bool {
-		return IsNonEmptyString(storageNamespace)
+		if !IsNonEmptyString(storageNamespace) {
+			return false
+		}
+		configuredBlockStoreType := viper.GetString("blockstore.type")
+		return strings.HasPrefix(storageNamespace, configuredBlockStoreType+"://")
 	}
 }
 
--- a/catalog/mvcc/validate.go
+++ b/catalog/mvcc/validate.go
@@ -3,9 +3,7 @@ package mvcc
 import (
 	"fmt"
 	"regexp"
-	"strings"
 
-	"github.com/spf13/viper"
 	"github.com/treeverse/lakefs/catalog"
 )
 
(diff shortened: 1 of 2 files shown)


catalog/mvcc/validate.go
```diff
@@ -100,7 +102,11 @@ func ValidateCommitter(name string) ValidateFunc {
 
 func ValidateStorageNamespace(storageNamespace string) ValidateFunc {
 	return func() bool {
-		return IsNonEmptyString(storageNamespace)
+		if !IsNonEmptyString(storageNamespace) {
+			return false
+		}
+		configuredBlockStoreType := viper.GetString("blockstore.type")
```
arielshaqed, line 108: I think that this code should go in `api.Controller.CreateRepositoryHandler`.  There you have `deps`, and you could ask `deps.BlockAdapter` whether it likes this storage namespace.

Why?
We prefer not to read configuration from arbitrary locations, it creates tricky dependencies (e.g. if `blockstore.type` is renamed, or if we add another way to configure blockstores).  It is OK for catalog to validate storage namespace for generally being nonempty, but catalog has no idea of the blockstore and should not do this validation.

webui/src/components/RepositoryCreateForm.js
```diff
@@ -51,6 +50,7 @@ export const RepositoryCreateForm = connect(
     }, [getConfig]);
 
     const blockstoreType = config.payload ? config.payload['blockstore.type'] : DEFAULT_BLOCKSTORE_TYPE;
+    const storageNamespaceValidityRegex = new RegExp(`^(${blockstoreType}):/.*$`);
```
arielshaqed, line 53: I'm sure you're asking "so how come the UI is allowed to touch configuration directly?!" here.  And you're right, it shouldn't.  This is tech debt on the UI -- and there is no expectation to fix it in this PR (of course)!

arielshaqed (changes requested):
Thanks!
I think that we have to split repo storage namespace validation on the server into 2: the catalog can continue to do simple textual validation, but only the block adapter can decide what is a valid URL, so add validation on the API controller code.

catalog/mvcc/validate.go
Sufiyan1997: Ok. I will revert it back to original.
As far as `api_controller` is concerned, I think it is already done.
https://github.com/treeverse/lakeFS/blob/8731b60fb472e28ac714933aff0c8ee3bc5b14de/api/api_controller.go#L547

webui/src/components/RepositoryCreateForm.js
Sufiyan1997: Agreed. 
But just out of curiosity, how can that be fixed? By sending only a limited set of configuration options to UI?

catalog/mvcc/validate.go
arielshaqed: Thanks, I missed that!  It's a bit subtle, but since `block.Adapter.Put` actually verifies the type of the storage namespace with the adaptor everything works.

arielshaqed (approved):
Great, thanks!

PR #1107, opened 2020-12-27
Add abort option to pyramid and sstable writer

itaiad200:


.gitignore
```diff
@@ -73,5 +73,6 @@ terraform.rc
 
 # generated mock files
 /graveler/committed/**/mock/
+/pyramid/mock/*
```
nopcoder, line 76: need to remove the file pyramid/mock/pyramid.go first

pyramid/tier_fs.go
```diff
@@ -176,6 +176,11 @@ func (tfs *TierFS) store(namespace, originalPath, filename string) error {
 	}
 }
 
+// removeTempFile removes the closed non-stored file from the FS.
+func (tfs *TierFS) removeTempFile(originalPath string) error {
+	return os.Remove(originalPath)
+}
```
nopcoder, line 182: why do we need this method? we can just call the os.Remove

pyramid/wr_file.go
```diff
@@ -12,28 +12,56 @@ type WRFile struct {
...
-		f.persisted = true
+	if err := f.idempotentClose(); err != nil {
+		return err
+	}
+
+	return f.abort()
+}
+
+// idempotentClose is like Close(), but doesn't fail when the file is already closed.
+func (f *WRFile) idempotentClose() error {
```
nopcoder, line 61: why not implement Close and check 'closed' bool as we did for abort?

pyramid/wr_file.go
```diff
@@ -12,28 +12,56 @@ type WRFile struct {
...
+		return err
 	}
 
+	return f.store(filename)
+}
+
+// Abort delete the file and cleans all traces of it.
+// If file was already stored, returns an error.
+func (f *WRFile) Abort() error {
 	if f.persisted {
```
nopcoder, line 48: want to check c.aborted first and return

nopcoder (changes requested):

.gitignore
itaiad200: Done.

pyramid/tier_fs.go
itaiad200: Done.

pyramid/wr_file.go
itaiad200: Since `*os.File` is embedded. User will still be able to write `weFile.File.Close()` without passing thru pyramid close. I can however not embed the os.File and implement a pass through for everything, but I rather avoid it.

pyramid/wr_file.go
itaiad200: Why? I'm fine with letting it fail if the file isn't there. Like any other misplaced operation on an os file handler.

nopcoder: it just looks easier to protect the call using the bool you already have and not having the user try to understand if a resource already aborted

nopcoder (approved):

pyramid/wr_file.go
itaiad200: You have a bug anyhow if you call multiple aborts, so I think custom pyramid error is the same as io.ErrFileNotFound

PR #1112, opened 2020-12-27
Add tree writer

guy-har:


graveler/committed/tree/writer.go
```diff
@@ -0,0 +1,180 @@
...
+		Value: data,
+	}, nil
+}
+
+// writePartToTree writes the part to the treeWriter
+func (w *writer) writePartToTree(part Part) error {
+	if w.treeWriter == nil {
+		var err error
+		treePath := w.treePath
+		if w.treeWriter, err = w.manager.GetWriter(treePath); err != nil {
```
itaiad200, line 165: I would be really careful here. You're calling the part manager for storing trees, not parts.
You need to make sure that the manager is init with the right FS, logger, cache and hash function.

It's so easy to make a mistake here that I strongly suggest that you DON'T use the PartManager interface, but wrap the sstable manager with something else and use it here.
All sstable management will still use the same code, but we'll know that we don't end up mixing the different usages of them.

graveler/committed/tree/writer.go
```diff
@@ -0,0 +1,180 @@
...
+func (w *writer) breakPoint(key graveler.Key) (bool, error) {
+	h, err := hashToInt(key)
+	if ErrUnsorted != nil {
+		return false, err
+	}
+	return int(h)%w.breakRatio == 0, nil
+}
+
+// PartToTreeRecord returns a tree record created from
+func PartToTreeRecord(part Part) (*committed.Record, error) {
```
itaiad200, line 141: Why is it exported?

graveler/committed/tree/writer.go
```diff
@@ -0,0 +1,180 @@
...
+	data, err := MarshalPart(part)
+	if err != nil {
+		return nil, err
+	}
+	partValue := &graveler.Value{
+		Identity: []byte(part.ID),
+		Data:     data,
```
itaiad200, line 148: Matryoshka ftw!

graveler/committed/tree/writer.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	data, err = MarshalValue(partValue)
+	if err != nil {
+		return nil, err
+	}
+	return &committed.Record{
```
itaiad200, line 154: The big downside of using `committed.Record` for storing tree records is that it has the same type as the underline parts records which makes it very error prune.

graveler/committed/tree/writer.go
```diff
@@ -0,0 +1,180 @@
...
+		Key:   part.MinKey,
```
itaiad200, line 155: Why? Please document.

graveler/committed/tree/writer.go
itaiad200, line 141: I believe this will be called from many places, hence I suggest moving it to it's own file.

itaiad200: And maybe add the opposite function too, from the sstable tree to the Tree struct.

graveler/committed/tree/writer.go
```diff
@@ -0,0 +1,180 @@
...
+	if err != nil {
+		return nil, err
+	}
+	treeID := graveler.TreeID(tree.PartID)
+	return &treeID, nil
+}
+
+func hashToInt(s []byte) (uint32, error) {
+	h := fnv.New32a()
+	_, err := h.Write(s)
```
itaiad200, line 124: I don't think this can really fail. Best to panic.

graveler/committed/tree/writer.go
```diff
@@ -0,0 +1,180 @@
...
+
+func (w *writer) SaveTree() (*graveler.TreeID, error) {
+	if err := w.closeCurrentPart(false); err != nil {
+		return nil, err
+	}
+	tree, err := w.treeWriter.Close()
+	if err != nil {
+		return nil, err
+	}
+	treeID := graveler.TreeID(tree.PartID)
```
itaiad200, line 118: It's not a PartID, right? 
See my suggestion below regarding wrapping the sstable impl for trees handling.

graveler/committed/tree/writer.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+	}
+	return nil
+}
+
+func (w *writer) closeCurrentPart(reachedBrakePoint bool) error {
+	if w.partWriter == nil {
+		return nil
+	}
+	wr, err := w.partWriter.Close()
```
itaiad200, line 77: Notice this simple `Close()` call stores the file to s3. During the write of the tree you currently serially wait for every one of those inner Stores To end!
Look for `BatchWriterCloser` that was designed specifically to handle this problem.

itaiad200 (changes requested):

graveler/committed/tree/tree.go
```diff
@@ -1,21 +1,42 @@
 package tree
 
 //go:generate mockgen -source=tree.go -destination=mock/tree.go -package=mock
+//go:generate protoc --proto_path=. --go_out=. --go_opt=paths=source_relative tree.proto
```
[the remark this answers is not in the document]
nopcoder: Created PR to extract the protoc to a step in a Makefile

nopcoder (changes requested):

graveler/committed/tree/writer.go
```diff
@@ -0,0 +1,180 @@
...
+	tree, err := w.treeWriter.Close()
+	if err != nil {
+		return nil, err
+	}
+	treeID := graveler.TreeID(tree.PartID)
+	return &treeID, nil
+}
+
+func hashToInt(s []byte) (uint32, error) {
+	h := fnv.New32a()
```
arielshaqed, line 123: A 32-bit hash makes me a bit worried.  The 64-bit version should just about be as efficient, and should behave better.  With 32 bits, the birthday paradox will ensure nastiness with fewer than 10_000 elements, and 50% chance of a collision with 2^16==65536 elements.  Rather than analyze, add bits: I hear 64-bit microprocessors are already available!

graveler/committed/tree/writer.go
```diff
@@ -0,0 +1,180 @@
...
+	return h.Sum32(), nil
+}
+
+// breakPoint returns true if should brake part after the given key
+func (w *writer) breakPoint(key graveler.Key) (bool, error) {
+	h, err := hashToInt(key)
+	if ErrUnsorted != nil {
+		return false, err
+	}
+	return int(h)%w.breakRatio == 0, nil
```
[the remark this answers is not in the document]
arielshaqed: Please _don't_ use maphash here.  maphash is seeded and the seed changes in different processes.  The goal of hashing here is to keep breakpoints _consistent_, so that trees will tend to break at the same place.  Changing the hash at each restart destroys this property.
Everything else, however, is true.  In particular, please *do not* convert an unsigned int to a signed int in order to hash it, that changes some weights a bit.  Instead convert the break selection to a uint.

graveler/committed/tree/writer.go
arielshaqed: I don't think that I agree here.  Worst-case we should use 2 separate managers.

arielshaqed (commented):
Good going!
Most worried about hashing, it should be 64 bits and the "ratio" parameter needs documentation.  But it's a long PR, so plenty of room for comments and nit-picking!

graveler/committed/tree/writer.go
guy-har: It's part of the Design, written in the `lakeFS on the Rocks - Design Document`. not sure we should document it here

graveler/committed/tree/tree.go
```diff
@@ -1,21 +1,42 @@
...
+	EstimatedSize     int
+	ReachedBrakePoint bool
+}
+
+func MarshalPart(part Part) ([]byte, error) {
+	return proto.Marshal(&PartData{
+		MinKey:        part.MinKey,
+		MaxKey:        part.MaxKey,
+		EstimatedSize: int64(part.EstimatedSize),
+		BreakOnHash:   part.ReachedBrakePoint,
```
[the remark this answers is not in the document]
guy-har: @arielshaqed I started by using `PartData`, the problem is that the generated struct by Protoc contains a lock, which blocked me from passing it by value.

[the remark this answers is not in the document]
arielshaqed: Yuck, that's terrible!  But it gets worse, see https://stackoverflow.com/questions/64183794/why-do-the-go-generated-protobuf-files-contain-mutex-locks: it *doesn't* contain `mutex.Lock`, it contains a `[0]mutex.Lock` that relies on what They euphemistically call a "go vet feature" (more like a bug) to prevent copying.  And the reason you should prevent copying is that _it would be **immoral**_: of all the Go `struct`s in the world, you should not copy protobufs because that would only be a shallow copy.

But anyway: could we just get rid of `Part` and have _only_ `PartData`?  And hold it as a pointer because "_something something_ Go is better than K&R C and perl4"?

arielshaqed (commented):
:-)

graveler/committed/tree/writer.go
guy-har: In order to use in tests

guy-har: Not sure it will, leaving it like this for now.

graveler/committed/tree/tree.go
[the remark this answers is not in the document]
guy-har: Not sure we want to.
Part has ID as a member and part data does not,
and Doing that will force us not to pass Part as value in functions.

graveler/committed/tree/writer.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+	if record.Value == nil {
+		return ErrNilValue
+	}
+
+	var err error
+	if w.partWriter == nil {
+		w.partWriter, err = w.manager.GetWriter(w.partPath)
+		if err != nil {
+			return err
```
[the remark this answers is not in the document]
arielshaqed: This is *still* an issue.  E.g. on l.160 you do say it's a tree writer that failed, why not say here it's a part writer?

graveler/committed/tree/writer.go
```diff
@@ -0,0 +1,180 @@
...
+	part := Part{
+		ID:                wr.PartID,
+		MinKey:            wr.First,
+		MaxKey:            wr.Last,
+		EstimatedSize:     wr.Count,
+		ReachedBrakePoint: reachedBrakePoint,
+	}
+	err = w.writePartToTree(part)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
arielshaqed: Again, we failed to write a part, why not say it?

graveler/committed/tree/writer.go
```diff
@@ -55,90 +61,87 @@ func (w *writer) WriteRecord(record graveler.ValueRecord) error {
...
+	parts, err := w.getBatchedParts()
 	if err != nil {
 		return nil, err
 	}
-	treeID := graveler.TreeID(tree.PartID)
-	return &treeID, nil
+	parts = append(parts, w.parts...)
+	sort.Slice(parts, func(i, j int) bool {
+		return bytes.Compare(parts[i].MinKey, parts[j].MinKey) < 0
+	})
```
arielshaqed, line 127: @itaiad200 this is a bit unfortunate, we got the parts in order, sent them over to a batch closer, the batch closer kept information about each of them but lost that order. :-(

graveler/committed/tree/writer.go
```diff
@@ -5,45 +5,51 @@ import (
...
+	ErrNilValue     = errors.New("record value should not be nil")
+)
 
-func NewWriter(manager committed.PartManager, breakRatio int, treePath, partPath committed.Namespace) Writer {
-	return &writer{
-		manager:    manager,
-		breakRatio: breakRatio,
-		partPath:   partPath,
-		treePath:   treePath,
+func NewWriter(partManager, treeManager committed.PartManager, approximatePartSize uint64, namespace committed.Namespace) *GeneralWriter {
```
arielshaqed, line 31:
```suggestion
func NewWriter(partManager, treeManager committed.PartManager, approximatePartSizeBytes uint64, namespace committed.Namespace) *GeneralWriter {
```

arielshaqed (approved):
Cool, thanks!

graveler/committed/tree/tree.go
```diff
@@ -1,21 +1,42 @@
...
-	Name   committed.ID
-	MaxKey graveler.Key
+	ID                committed.ID
+	MinKey            committed.Key
+	MaxKey            committed.Key
+	EstimatedSize     int
+	ReachedBrakePoint bool
+}
+
+func MarshalPart(part Part) ([]byte, error) {
```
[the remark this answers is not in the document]
guy-har: Will do it next as part of Renaming PR

graveler/committed/tree/writer.go
```diff
@@ -55,90 +61,87 @@ func (w *writer) WriteRecord(record graveler.ValueRecord) error {
...
 	}
-	return int(h)%w.breakRatio == 0, nil
+	n := h.Sum64() % w.approximatePartSize
+	return n == 0, nil
 }
 
-// PartToTreeRecord returns a tree record created from
-func PartToTreeRecord(part Part) (*committed.Record, error) {
+// PartToValue returns a Value representing a Part in Tree
+func PartToValue(part Part) (committed.Value, error) {
```
itaiad200, line 145: unexport?

graveler/committed/tree/writer.go
itaiad200: Also, some parts were passed as is, so sorting here is mandatory.

graveler/committed/tree/writer.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	parts = append(parts, w.parts...)
+	sort.Slice(parts, func(i, j int) bool {
+		return bytes.Compare(parts[i].MinKey, parts[j].MinKey) < 0
+	})
+	w.parts = parts
+	return w.writePartsToTree()
+}
+
+// shouldBreakAtKey returns true if should brake part after the given key
```
johnnyaug, line 133:
```suggestion
// shouldBreakAtKey returns true if should break part after the given key
```

itaiad200 (approved):

graveler/committed/tree/writer.go
```diff
@@ -0,0 +1,185 @@
...
+	_, err := h.Write(key)
+	if err != nil {
+		return false, err
+	}
+	n := h.Sum64() % w.approximatePartSize
+	return n == 0, nil
+}
+
+// PartToValue returns a Value representing a Part in Tree
+func PartToValue(part Part) (committed.Value, error) {
```
johnnyaug, line 145: must be exported?

nopcoder (approved):

graveler/committed/tree/writer.go
```diff
@@ -0,0 +1,185 @@
...
+	wr, err := treeWriter.Close()
+	if err != nil {
+		return nil, fmt.Errorf("failed closing treeWriter: %w", err)
+	}
+	treeID := graveler.TreeID(wr.PartID)
+	return &treeID, nil
+}
+
+func (w *GeneralWriter) Abort() error {
+	return w.partWriter.Abort()
```
johnnyaug, line 184: nil check?

graveler/committed/tree/tree.proto
```diff
@@ -5,8 +5,8 @@ option go_package = "github.com/treevese/lakefs/graveler/committed/tree";
 package tree;
 
 message PartData {
+	// First key of part, must be >= key of part in SSTable.  If missing, == key of part in SSTable.
```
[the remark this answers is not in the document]
guy-har: I will change it as part of the Naming PR

PR #1113, opened 2020-12-27
use gaul/s3proxy

johnnyaug:


gateway/handler.go
```diff
@@ -269,7 +275,16 @@ func RepoOperationHandler(sc *ServerContext, repoID string, handler operations.R
 		repo, err := authOp.Cataloger.GetRepository(sc.ctx, repoID)
 		if errors.Is(err, db.ErrNotFound) {
 			authOp.Log().WithField("repository", repoID).Warn("the specified repo does not exist")
```
nopcoder, line 277: warning in both cases will be probably bad on the performance

config/config.go
```diff
@@ -300,6 +300,10 @@ func (c *Config) GetS3GatewayDomainName() string {
 	return viper.GetString("gateways.s3.domain_name")
 }
 
+func (c *Config) GetS3GatewayProxyURL() string {
+	return viper.GetString("gateways.s3.proxy_url")
```
ozkatz, line 304: I would probaby assume a `proxy_url` is a proxy we use to talk to S3 :)
I suggest calling it `gatways.s3.fallback_url`?

gateway/handler.go
ozkatz: I don't think we should warn on invalid requests or missing resources

gateway/handler.go
```diff
@@ -269,7 +275,16 @@ func RepoOperationHandler(sc *ServerContext, repoID string, handler operations.R
...
-			authOp.EncodeError(gatewayerrors.ErrNoSuchBucket.ToAPIErr())
+			if sc.proxyURL == "" {
+				authOp.EncodeError(gatewayerrors.ErrNoSuchBucket.ToAPIErr())
+				return
+			}
+			proxyURL, err := url.Parse(sc.proxyURL)
```
ozkatz, line 282: we can do this once on initialization instead of per request?

gateway/handler.go
```diff
@@ -269,7 +275,16 @@ func RepoOperationHandler(sc *ServerContext, repoID string, handler operations.R
...
+			if err != nil {
+				authOp.EncodeError(gatewayerrors.ErrInternalError.ToAPIErr())
+				return
+			}
+			gohttputil.NewSingleHostReverseProxy(proxyURL).ServeHTTP(writer, request)
```
ozkatz, line 287: so simple!

gateway/handler.go
```diff
@@ -306,14 +321,22 @@ func PathOperationHandler(sc *ServerContext, repoID, refID, path string, handler
...
```
ozkatz, line 333: maybe extract this to a function to avoid repetition?

ozkatz (changes requested):
Looks great!! elegant and simple solution.
I would strongly suggest changing the configuration setting name

gateway/handler.go
```diff
@@ -268,8 +278,12 @@ func RepoOperationHandler(sc *ServerContext, repoID string, handler operations.R
 		// validate repo exists
 		repo, err := authOp.Cataloger.GetRepository(sc.ctx, repoID)
 		if errors.Is(err, db.ErrNotFound) {
-			authOp.Log().WithField("repository", repoID).Warn("the specified repo does not exist")
-			authOp.EncodeError(gatewayerrors.ErrNoSuchBucket.ToAPIErr())
+			if sc.fallbackProxy == nil {
+				authOp.Log().WithField("repository", repoID).Warn("the specified repo does not exist")
```
ozkatz, line 282: For anything that is the user's fault I recommend using `Debug`

docs/reference/configuration.md
```diff
@@ -74,6 +74,7 @@ This reference uses `.` to denote the nesting of values.
   (`*.s3.local.lakefs.io` always resolves to 127.0.0.1, useful for
   local development
 * `gateways.s3.region` `(string : "us-east-1")` - AWS region we're pretending to be. Should match the region configuration used in AWS SDK clients
+* `gateways.s3.fallback_url` `(string)` - If specified, requests with a non-existing repository will be forwarded to this url.
```
ozkatz, line 77: Maybe add a note about s3 proxy and the use case?

ozkatz (approved):
Nice work! I'm fine with merging this, only a couple of small comments

PR #1114, opened 2020-12-27
Run protoc as part of a makefile step optional

nopcoder:
In order to use `go generate` when possible - running `go generate` as part of the build and describe how to build proto files using the same technic collides. The protoc is not required in the build process as a native binary, so in this change we move the build outside the `go generate` into our Makefile.

- move protoc compile as a makefile step (not run by default)
- validator will validate protoc generated output matched the checked-in source code

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -22,8 +22,6 @@ jobs:
         run: make gen
       - name: Install Protoc
         uses: arduino/setup-protoc@v1
-      - name: Install the protocol compiler plugins for Go
-        run: go get google.golang.org/protobuf/cmd/protoc-gen-go google.golang.org/grpc/cmd/protoc-gen-go-grpc
       - name: Checks validator
         run: make checks-validator
         env:
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -20,6 +20,8 @@ jobs:
         id: go
       - name: Generated code
         run: make gen
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v1
       - name: Checks validator
         run: make checks-validator
         env:
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -22,6 +22,8 @@ jobs:
         run: make gen
       - name: Install Protoc
         uses: arduino/setup-protoc@v1
+      - name: Install the protocol compiler plugins for Go
+        run: go get google.golang.org/protobuf/cmd/protoc-gen-go google.golang.org/grpc/cmd/protoc-gen-go-grpc
       - name: Checks validator
         run: make checks-validator
         env:
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 GOCMD=$(or $(shell which go), $(error "Missing dependency - no go in PATH"))
 DOCKER=$(or $(shell which docker), $(error "Missing dependency - no docker in PATH"))
 GOBINPATH=$(shell $(GOCMD) env GOPATH)/bin
+PROTOC=$(or $(shell which protoc), $(error "Missing protobuf compilter - no protoc on PATH"))
 NPM=$(or $(shell which npm), $(error "Missing dependency - no npm in PATH"))
 
 export PATH:= $(PATH):$(GOBINPATH)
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -65,36 +67,38 @@ docs-serve: ### Serve local docs
...
 	$(GOGENERATE) ./api/...
 
 del-gen-api:
 	@rm -rf $(API_BUILD_DIR)
 	@mkdir -p $(API_BUILD_DIR)
 
 .PHONY: gen-mockgen
-gen-mockgen: $(GOBINPATH)/mockgen ## Run the generator for inline commands
+gen-mockgen: go-install ## Run the generator for inline commands
 	$(GOGENERATE) ./graveler/committed/...
```
itaiad200, line 90: Can we run it on the entire tree now?

itaiad200 (approved):

.github/workflows/go.yaml
```diff
@@ -20,6 +20,8 @@ jobs:
         id: go
       - name: Generated code
         run: make gen
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v1
```
arielshaqed, line 24: arduino created this action?  No other action?? :unamused:

Makefile
```diff
@@ -1,6 +1,7 @@
 GOCMD=$(or $(shell which go), $(error "Missing dependency - no go in PATH"))
 DOCKER=$(or $(shell which docker), $(error "Missing dependency - no docker in PATH"))
 GOBINPATH=$(shell $(GOCMD) env GOPATH)/bin
+PROTOC=$(or $(shell which protoc), $(error "Missing protobuf compilter - no protoc on PATH"))
```
arielshaqed, line 4:
```suggestion
PROTOC=$(or $(shell which protoc), $(error "Missing protobuf compiler - no protoc on PATH"))
```

Makefile
```diff
@@ -14,6 +15,7 @@ GOTEST=$(GOCMD) test
 GOTESTRACE=$(GOTEST) -race
 GOGET=$(GOCMD) get
 GOFMT=$(GOCMD)fmt
+PROTOC=protoc
```
arielshaqed, line 18: Didn't you set it above?

Makefile
arielshaqed, line 4: Shouldn't all of these be `:=` and not `=`?  We don't need to run a shell to find which program to run before _each_ time we use one of these macros??!?

Makefile
```diff
@@ -65,36 +67,38 @@ docs-serve: ### Serve local docs
...
 	@mkdir -p $(API_BUILD_DIR)
 
 .PHONY: gen-mockgen
-gen-mockgen: $(GOBINPATH)/mockgen ## Run the generator for inline commands
+gen-mockgen: go-install ## Run the generator for inline commands
 	$(GOGENERATE) ./graveler/committed/...
 	$(GOGENERATE) ./pyramid
 
-validate-swagger: $(GOBINPATH)/swagger  ## Validate swagger.yaml
+swagger-validator: go-install ## Validate swagger.yaml
```
arielshaqed, line 93:
```suggestion
swagger-validate: go-install
```
Most of our actions are verbs in the imperative mode.  And the comment adds little.

Makefile
```diff
@@ -130,7 +134,11 @@ fmt-validator:  ## Validate go format
 		echo Your code formatting is according to gofmt standards; \
 	fi
 
-checks-validator: lint fmt-validator validate-swagger ## Run all validation/linting steps
+.PHONY: proto-validator
+proto-validator: proto ## build proto and check if diff found
+	@git diff --quiet -- catalog/rocks/catalog.pb.go
```
arielshaqed, line 139:
```suggestion
	git diff --quiet -- catalog/rocks/catalog.pb.go
```
# NOT ASHAMED TO BE RUNNING GIT!
OTOH, error messages from `git` will look strange if the last command run and displayed is `protoc`.

arielshaqed (approved):
Nice, thanks!

Makefile
ozkatz: FYI the comment is parsed and shown as part of `make help`.. not sure that's a compelling enough reason to add it though

Makefile
nopcoder: yes, will do it in another PR

.github/workflows/go.yaml
nopcoder: image
didn't find anything better in the github marketplace

Makefile
nopcoder: np, I'll replace all the XXX-validator to validate-XXX

Makefile
nopcoder: yes :)

Makefile
nopcoder: good point - I'll update

PR #1115, opened 2020-12-28
Add commit from existing tree

itaiad200:


graveler/graveler.go
```diff
@@ -841,6 +841,54 @@ func (g *graveler) Commit(ctx context.Context, repositoryID RepositoryID, branch
...
+
+func (g *graveler) CommitExistingTree(ctx context.Context, repositoryID RepositoryID, branchID BranchID, treeID TreeID, committer string, message string, metadata Metadata) (CommitID, error) {
+	cancel, err := g.branchLocker.AquireMetadataUpdate(repositoryID, branchID)
+	if err != nil {
+		return "", fmt.Errorf("acquire metadata update: %w", err)
+	}
+	defer cancel()
+	repo, err := g.RefManager.GetRepository(ctx, repositoryID)
+	if err != nil {
+		return "", fmt.Errorf("get repository: %w", err)
```
arielshaqed, line 857: Consider
```suggestion
		return "", fmt.Errorf("get repository %s: %w", repositoryID, err)
```
otherwise the error is much less actionable.  (Similarly below and throughout)

graveler/graveler.go
```diff
@@ -841,6 +841,54 @@ func (g *graveler) Commit(ctx context.Context, repositoryID RepositoryID, branch
...
+	}
+	branch, err := g.RefManager.GetBranch(ctx, repositoryID, branchID)
+	if err != nil {
+		return "", fmt.Errorf("get branch: %w", err)
+	}
+	changes, err := g.StagingManager.List(ctx, branch.stagingToken)
+	if err != nil {
+		return "", fmt.Errorf("staging list: %w", err)
+	}
+	if changes.Next() {
```
arielshaqed, line 867: This seems quite inefficient: we want to know whether there are any changes to the branch, so we try to _list_ them all (or, at least, a first page).  @nopcoder I think we need another API here.

graveler/graveler.go
```diff
@@ -841,6 +841,54 @@ func (g *graveler) Commit(ctx context.Context, repositoryID RepositoryID, branch
...
+		return "", ErrDirtyBranch
+	}
+
+	treeIt, err := g.CommittedManager.List(ctx, repo.StorageNamespace, treeID)
+	if err != nil {
+		return "", fmt.Errorf("treeID %s: %w", treeID, ErrTreeNotFound)
+	}
+	defer treeIt.Close()
```
arielshaqed, line 875: Why do we need this?

graveler/graveler.go
arielshaqed, line 867: I think you need to do this?
```suggestion
	defer changes.Close()
	if changes.Next() {
```

graveler/graveler_test.go
```diff
@@ -429,6 +429,142 @@ func TestGraveler_Commit(t *testing.T) {
...
+				repositoryID: "repo",
+				branchID:     "branch",
+				committer:    "committer",
+				message:      "a message",
+				metadata:     nil,
+			},
+			want:        expectedCommitID,
+			expectedErr: graveler.ErrTreeNotFound,
+		},
+		{
```
arielshaqed, line 509: do we also need "branch not found"?

graveler/graveler_test.go
```diff
@@ -429,6 +429,142 @@ func TestGraveler_Commit(t *testing.T) {
...
+			want:        expectedCommitID,
+			expectedErr: graveler.ErrConflictFound,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			expectedCommitID := graveler.CommitID("expectedCommitId")
+			expectedTreeID := graveler.TreeID("expectedTreeID")
+			expectedBranchID := graveler.BranchID("expectedBranchID")
+			expectedRepositoryID := graveler.RepositoryID("expectedTreeID")
```
arielshaqed, line 533: Could we just put all of these as function-level `const`s?

graveler/graveler_test.go
```diff
@@ -429,6 +429,142 @@ func TestGraveler_Commit(t *testing.T) {
...
+			got, err := g.CommitExistingTree(context.Background(), expectedRepositoryID, expectedBranchID, expectedTreeID, tt.args.committer, tt.args.message, tt.args.metadata)
+			if !errors.Is(err, tt.expectedErr) {
+				t.Fatalf("unexpected err got = %v, wanted = %v", err, tt.expectedErr)
+			}
+			if err != nil {
+				return
+			}
+
+			// checking apply wasn't called
+			if diff := deep.Equal(tt.fields.CommittedManager.AppliedData, testutil.AppliedData{}); diff != nil {
```
arielshaqed, line 544: This is strange (but understandable perhaps): it tests that something did _not_ happen.

graveler/graveler_test.go
```diff
@@ -429,6 +429,142 @@ func TestGraveler_Commit(t *testing.T) {
...
+				Committer: tt.args.committer,
+				Message:   tt.args.message,
+				TreeID:    expectedTreeID,
+				Parents:   graveler.CommitParents{expectedCommitID},
+				Metadata:  graveler.Metadata{},
+			}); diff != nil {
+				t.Errorf("unexpected added commit %s", diff)
+			}
+			if tt.fields.StagingManager.DropCalled {
+				t.Errorf("Staging manager drop shouldn't be called")
```
arielshaqed, line 558:
```suggestion
				t.Error("Staging manager drop shouldn't be called")
```

arielshaqed (changes requested):
Thanks!
Only required changes are "why do we need this?" (below) and an iterator `Close()` I think is missing.

graveler/graveler.go
itaiad200: Done.

graveler/graveler.go
itaiad200: Done.

itaiad200: Agree. Added.

graveler/graveler.go
itaiad200: Changed.

graveler/graveler_test.go
itaiad200: Harder with the current fakes and adds little coverage.
Deferring until we'll switch to generated mocks here too.

graveler/graveler_test.go
itaiad200: Done.

graveler/graveler_test.go
itaiad200: Exactly! I want to make sure commit isn't done using Apply, otherwise it's a mess.

graveler/graveler_test.go
itaiad200: Done.

graveler/graveler.go
```diff
@@ -395,6 +395,9 @@ type CommittedManager interface {
 	// Get returns the provided key, if exists, from the provided TreeID
 	Get(ctx context.Context, ns StorageNamespace, treeID TreeID, key Key) (*Value, error)
 
+	// IsTreeExist returns true iff the tree with treeID exists
+	IsTreeExist(ctx context.Context, ns StorageNamespace, treeID TreeID) (bool, error)
```
nopcoder, line 399: Think we should open GetTree and not IsTreeExist, expose an interface that can represent a Tree with ID() as a method will do better work.

graveler/graveler.go
```diff
@@ -425,6 +428,9 @@ type StagingManager interface {
 	// List returns a ValueIterator for the given staging token
 	List(ctx context.Context, st StagingToken) (ValueIterator, error)
 
+	// IsDirty returns true iff staging area for the token isn't empty
```
nopcoder, line 431: s/iff/if/

graveler/graveler.go
```diff
@@ -425,6 +428,9 @@ type StagingManager interface {
...
+	IsDirty(ctx context.Context, st StagingToken) (bool, error)
```
nopcoder, line 432: Why not use the List API?

graveler/graveler.go
```diff
@@ -841,6 +847,56 @@ func (g *graveler) Commit(ctx context.Context, repositoryID RepositoryID, branch
 	return newCommit, nil
 }
 
+var (
+	ErrDirtyBranch  = errors.New("can't apply tree on dirty branch")
+	ErrTreeNotFound = errors.New("tree not found")
```
nopcoder, line 852: move to errors.go

nopcoder (changes requested):

graveler/graveler.go
itaiad200: if-and-only-if

graveler/graveler.go
itaiad200: To avoid iterating and fetching unnecessary objects.

graveler/graveler.go
itaiad200: Done.

graveler/graveler.go
itaiad200: GetTree will read objects we don't want to read, this is the cheapest way of achieving the same.

graveler/graveler_test.go
arielshaqed: I think I now better understand why this is "strange": it is a non-functional test.  Say I somehow managed to implement `Commit` using `Apply` and it was really elegant and efficient, and passed all the functional tests (i.e. a client would not be able to distinguish the results).  Then IIUC this test would fail, enforcing matters of our current good taste?

Why would we want to do this?  E.g. if we compacted some large uncommitted state into a "staging" tree (`git` style...) and wanted to commit _that_?  Dunno, anyway not sure about this part of the test.

OTOH, we've probably wasted enough time about this one line.

graveler/graveler.go
nopcoder: @itaiad200 get tree needs to read just the main part

arielshaqed (approved):
Thanks!

graveler/graveler.go
nopcoder: if we will have an issue with the iterator we can optimize it

graveler/graveler_test.go
itaiad200: Removed it.

graveler/graveler.go
```diff
@@ -388,14 +388,21 @@ type RefManager interface {
 	Log(ctx context.Context, repositoryID RepositoryID, commitID CommitID) (CommitIterator, error)
 }
 
+// Tree abstracts the data structure of the committed data.
```
arielshaqed, line 391: Will this interface evolve?  Otherwise, how do I use `GetTree`?  Is it just to check whether it `!= nil`, and make `IsTreeExist` seem like it returns a tree?

graveler/testutil/fakes.go
```diff
@@ -20,10 +20,17 @@ type CommittedFake struct {
 	diffIterator  graveler.DiffIterator
 	Err           error
 	TreeID        graveler.TreeID
-	TreeExists    bool
 	AppliedData   AppliedData
 }
 
+type Tree struct {
```
arielshaqed, line 26:
```suggestion
type TreeFake struct {
```
otherwise `testutil.Tree` implements `graveler.Tree` and confuses me.

arielshaqed (approved):
Thanks!

All good, any which way.

nopcoder (approved):

graveler/graveler.go
itaiad200: The latter more or less for now. It will probably add the tree parts later on.

PR #1118, opened 2020-12-28
go mod update

nopcoder:
- switch from deprecated terminal package to term


go.mod
```diff
@@ -3,92 +3,84 @@ module github.com/treeverse/lakefs
 go 1.15
 
 require (
-	cloud.google.com/go v0.63.0
-	cloud.google.com/go/storage v1.10.0
-	github.com/Masterminds/squirrel v1.4.0
+	cloud.google.com/go v0.74.0
```
johnnyaug, line 6: why are there changes in so many other packages?

nopcoder: Used go-mod-upgrade to check which package can be upgraded, for each one checked the release notes, compiled and run the tests.
Some packages got good fixes we will need to add at some point.
Create PR to see full checks.

johnnyaug (approved):

PR #1119, opened 2020-12-29
Create TierFS pyramid from configuration

arielshaqed:
The main user-visible change (or rather will-be-user-visible-once-we-release-lakeFS-on-the-rocks) is to rename configuration parameters in `config.go`.

The main code change is to move the configuration parts of pyramid over to a subdirectory `pyramid/params`.  This allows configuration parameters to live in a separate package, breaking the dependency cycle.


config/config.go
```diff
@@ -320,6 +312,22 @@ func (c *Config) GetStatsFlushInterval() time.Duration {
...
 }
 
+// GetCommittedTierFSParams returns parameters for building a tierFS.  Caller must separately
+// build and populate Adaptor.
+func (c *Config) GetCommittedTierFSParams() *pyramidparams.Params {
+	logger := logging.Default().WithField("module", "pyramid")
+	return &pyramidparams.Params{
+		FSName:             "committed",
+		Logger:             logger,
+		Adaptor:            nil, // Built by factory during construction
```
itaiad200, line 322: When and how is this passed?

arielshaqed: Sorry, went back-and-forth on where this is generated and ended up with nowhere.

config/config.go
```diff
@@ -314,18 +315,22 @@ func (c *Config) GetStatsFlushInterval() time.Duration {
...
 		Logger:             logger,
-		Adaptor:            nil, // Built by factory during construction
+		Adaptor:            adaptor,
 		BlockStoragePrefix: viper.GetString("committed.block_storage_prefix"),
 		Local: pyramidparams.LocalDiskParams{
 			BaseDir:        viper.GetString("committed.local_cache.dir"),
 			AllocatedBytes: viper.GetInt64("committed.local_cache.size_bytes"),
 		},
-	}
+	}, err
```
itaiad200, line 333:
```suggestion
	}, nil
```

config/config.go
```diff
@@ -320,6 +313,26 @@ func (c *Config) GetStatsFlushInterval() time.Duration {
 	return viper.GetDuration("stats.flush_interval")
 }
 
+// GetCommittedTierFSParams returns parameters for building a tierFS.  Caller must separately
+// build and populate Adaptor.
+func (c *Config) GetCommittedTierFSParams() (*pyramidparams.Params, error) {
+	adaptor, err := factory.BuildBlockAdapter(c)
```
itaiad200, line 319: This is changing the current pattern of block adapter being a singleton, is it desired?
I can imagine a scenario where we want just a single instance to control throughputs, retries, etc ..

You can still use the factory and have a singleton if you save the first instance you created and return it for secondary requests.

itaiad200 (changes requested):
Requesting changes for clarification regarding the use of the adaptor.

config/config.go
arielshaqed: Done!

config/config.go
arielshaqed: Not done.  Singletons may or may not be needed here, the fact we currently have just the one block adapter seems just a historical trivium to me, with no rationale one way or the other.  I can imagine all of these use-cases, and cannot prefer one to another right now:
1. Everything entirely separate.
1. Some adapters separate, some shared, some separate with shared component.  (E.g. S3 would be a good candidate for this!)
1. One adapter of each type.
1. One adapter for everything.

I really hate infrastructure that forces singletons on me.  E.g. @johnnyaug had some problems last and this week because there can only be one S3a endpoint in a particular software version.

So until I know better, I want to go with the least effort -- which is what I implements (because I was lazy and didn't even think about it).

arielshaqed (commented):
Thanks!  Please also see the included c4184ef3a88b5f094902c8887f33f26b4a85be26 which fixes a dumb spelling error which I consistently made.

itaiad200 (approved):

arielshaqed:
Thanks!  Pulling...

PR #1120, opened 2020-12-29
disk writer abort after close does nothing

nopcoder:


itaiad200 (changes requested):
Add a simple test case that verifies it.

itaiad200 (approved):
LGTM, thanks!

PR #1126, opened 2020-12-30
Committed range iterator

nopcoder:
- committed.Iterator will use RangeManager to get range iterator to scan each range on meta range.
- committed.RangeIterator will get graveler.Iterator and unmarshal it to to the committed record.


graveler/committed/range_iterator.go
```diff
@@ -23,10 +23,12 @@ func (r *RangeIterator) Next() bool {
 		return false
 	}
 	val := r.it.Value()
+	// unmarshal value
 	var v *graveler.Value
 	if val.Value != nil {
 		v, r.err = UnmarshalValue(val.Value)
 		if r.err != nil {
+			r.value = nil
```
arielshaqed, line 31: Actually I don't particularly like this.  If `UnmarshalValue` decided to return a value, why erase it here?

graveler/testutil/fakes.go
```diff
@@ -329,6 +330,44 @@ func (r *valueIteratorFake) Err() error {
...
+	return &committedValueIteratorFake{records: records, current: -1}
+}
+
+func (r *committedValueIteratorFake) Next() bool {
+	r.current++
+	return r.current < len(r.records)
+}
+
+func (r *committedValueIteratorFake) SeekGE(id committed.Key) {
+	for i, record := range r.records {
```
arielshaqed, line 349: I think this should `sort.BSearch`.

arielshaqed (approved):
Thanks!

graveler/committed/range_iterator.go
nopcoder: so .Err() will return the error and .Value() will return nil

PR #1128, opened 2020-12-30
Fix minor lint issues

nopcoder:


cloud/aws/s3inventory/reader_test.go
```diff
@@ -72,6 +72,9 @@ func generateParquet(t *testing.T, objs <-chan *TestObject, fieldToRemove string
 		_ = fw.Close()
 	}()
 	jsonSchema, err := json.Marshal(parquetSchema(fieldToRemove))
+	if err != nil {
+		t.Fatalf("failed to json marshal: %s", err)
```
arielshaqed, line 76:
```suggestion
		t.Fatalf("failed to marshal to JSON %+v: %s", parquetSchema(fieldToRemove), err)
```

export/tasks_generator_test.go
```diff
@@ -153,15 +153,6 @@ func makeAllDependencies(dd directDependencies) allDependencies {
 	return &allDependenciesSet{dd: dd, s: make(map[parade.TaskID]map[parade.TaskID]struct{})}
 }
 
-func validateAllAfter(t *testing.T, d allDependencies, typ string, before parade.TaskID, afters []parade.TaskID) {
```
arielshaqed, line 156: This one's mine.  Note to self: verify that we don't really need to use it any longer.

graveler/ref/resolver_test.go
```diff
@@ -43,10 +43,11 @@ func TestRefManager_Dereference(t *testing.T) {
 		t.Fatalf("unexpected error: %v", err)
 	}
 
-	ids := make([]graveler.CommitID, 0)
 	for iter.Next() {
 		c := iter.Value()
-		ids = append(ids, c.CommitID)
```
arielshaqed, line 49: I don't understand.  Where do we validate that we actually received good values, or even the correct number of values  E.g. if the iterator never iterates and ends directly, this passes.  (Not introduced in this PR, of course, so fine to open an issue and assign to whoever should fix)

parade/parade_test.go
```diff
@@ -22,16 +22,10 @@ import (
...
-const (
-	dbContainerTimeoutSeconds = 10 * 60 // 10 min
-	dbName                    = "parade_db"
-)
-
 var (
 	pool        *dockertest.Pool
 	databaseURI string
 
-	postgresUrl = flag.String("postgres-url", "", "Postgres connection string.  If unset, run a Postgres in a Docker container.")
```
arielshaqed, line 34: Yikes!  Good catch...

arielshaqed (approved):
Cool!

What changed?  Why are these (finally...) failing to lint?

nopcoder:
> Cool!
> 
> What changed?  Why are these (finally...) failing to lint?

didn't fail, we are just graceful with the tests code.

export/tasks_generator_test.go
arielshaqed: Reply to self: D'uh, you removed it in favour of a better checked with full transitivity that lets you test dependencies directly, e.g. `expectedDeps` on l.358.

nopcoder:
the build fails after adding err check - so I still need to verify if ignoring it was a bug or there is a bug

catalog/mvcc/cataloger_delete_entry_test.go
```diff
@@ -150,8 +151,11 @@ func TestCataloger_DeleteEntryAndCheckItRemainsInCommits(t *testing.T) {
 	if err != nil {
 		t.Fatal("Failed to commit after delete:", err)
 	}
-	entry, err = c.GetEntry(ctx, repository, nextCommit.Reference, "/file2", catalog.GetEntryParams{})
-	entry, err = c.GetEntry(ctx, repository, prevCommit.Reference, "/file2", catalog.GetEntryParams{})
+	_, err = c.GetEntry(ctx, repository, nextCommit.Reference, "/file2", catalog.GetEntryParams{})
+	testutil.Must(t, err)
```
arielshaqed, line 155: Shouldn't the file be missing here?  Indeed, the test appears to fail.

arielshaqed (changes requested):
Requesting changes because test is failing

graveler/ref/resolver_test.go
nopcoder: Adding a test code to compare the values.
I think it was made just to check there was no error after scanning the commit log.

catalog/mvcc/cataloger_delete_entry_test.go
nopcoder: Will verify that the file is missing

graveler/ref/resolver_test.go
```diff
@@ -43,37 +44,44 @@ func TestRefManager_Dereference(t *testing.T) {
...
+		"bfa1e0382ff3c51905dc62ced0a67588b5219c1bba71a517ae7e7857f0c26afe",
+		"d2248dcc1a4de004e10e3bc6b820655e649b8d986d983b60ec98a357a0df194b",
+		"a2d98d820f6ff3f221223dbe6a22548f78549830d3b19286b101f13a0ee34085",
+		"4f13621ec00d4e44e8a0f0ad340224f9d51db9b6518ee7bef17f598aea9e0431",
+		"df87d5329f4438662d6ecb9b90ee17c0bdc9a78a884acc93c0c4fe9f0f79d059",
+		"29706d36de7219e0796c31b278f87201ef835e8cdafbcc3c907d292cd31f77d5",
+	}
+
+	if diff := deep.Equal(commitIDs, commitLog); diff != nil {
+		t.Fatal("Difference found on commit log", diff)
```
arielshaqed, line 83:
```suggestion
		t.Error("Difference found on commit log", diff)
```
(the test can continue regardless, and the additional error messages may help understand what went wrong -- e.g. if `commitIDs` is empty.)

arielshaqed (approved):
Nice, thanks!

nopcoder:
> Nice, thanks!

Thank you.

PR #1130, opened 2020-12-31
Merge Implementation

johnnyaug:


graveler/committed/merge_iterator.go
```diff
@@ -0,0 +1,108 @@
...
+	ns           graveler.StorageNamespace
+	err          error
+}
+
+// NewMergeIterator returns a ValueIterator with all changes to be performed on left in order to merge right into left, relative to base.
+// The iterator will return ErrConflictFound when it reaches a conflict.
+func NewMergeIterator(ctx context.Context, ns graveler.StorageNamespace, left, right, base graveler.MetaRangeID, committedMgr graveler.CommittedManager) (*mergeIterator, error) {
+	diffIt, err := committedMgr.Diff(ctx, ns, left, right)
+	if err != nil {
+		return nil, err
```
itaiad200, line 26: Add some information on the action that failed.

itaiad200, line 26: Add some information on the action the failed.

graveler/committed/merge_iterator.go
```diff
@@ -0,0 +1,108 @@
...
+func (d *mergeIterator) Value() *graveler.ValueRecord {
+	return d.val
+}
+
+func (d *mergeIterator) Err() error {
+	return d.err
+}
+
+func (d *mergeIterator) Close() {
+	d.diffIt.Close()
```
itaiad200, line 107: propagation of the error?

graveler/graveler.go
```diff
@@ -159,9 +159,10 @@ type TagRecord struct {
 
 // Diff represents a change in value based on key
 type Diff struct {
-	Type  DiffType
-	Key   Key
-	Value *Value
+	Type        DiffType
+	Key         Key
+	Value       *Value
+	OldIdentity []byte // the Identity of the value on the left side of the diff
```
ozkatz, line 165: Why is it "old"?

graveler/committed/merge_iterator.go
```diff
@@ -0,0 +1,109 @@
...
+// relative to base as the merge base.
+// The iterator will return ErrConflictFound when it reaches a conflict.
+func NewMergeIterator(diffTheirsToOurs graveler.DiffIterator, base Iterator) (*mergeIterator, error) {
+	return &mergeIterator{diffIt: diffTheirsToOurs, base: base}, nil
+}
+
+func (d *mergeIterator) valueFromBase(key graveler.Key) *graveler.ValueRecord {
+	d.base.SeekGE(key)
+	var val *graveler.ValueRecord
+	for d.base.Next() && val == nil {
```
ozkatz, line 27: `val == nil` will always be true here because you just declared it and haven't assigned to it yet.

graveler/committed/merge_iterator.go
```diff
@@ -0,0 +1,109 @@
...
+		key := val.Key
+		typ := val.Type
+		baseVal := d.valueFromBase(key)
+		switch typ {
+		case graveler.DiffTypeAdded:
+			if baseVal == nil {
+				d.setValue()
+				return true
+			}
+			if !bytes.Equal(baseVal.Identity, val.Value.Identity) {
```
ozkatz, line 48: if I read this correctly, this means we got to a record where their identity for a key is different from the base identity for a key. why is it a conflict?

ozkatz (changes requested):
sorry for blocking - I'm not 100% clear on the implmentation. I think the merge iterator's Next() implementation should be simplified or explained in more detail.

graveler/committed/merge_iterator.go
johnnyaug: but it's a loop

johnnyaug:
@ozkatz added some comments to Next. Hopefully it is clearer now

graveler/graveler.go
```diff
@@ -159,10 +159,10 @@ type TagRecord struct {
...
 // Diff represents a change in value based on key
 type Diff struct {
-	Type        DiffType
-	Key         Key
-	Value       *Value
-	OldIdentity []byte // the Identity of the value on the left side of the diff
+	Type         DiffType
+	Key          Key
+	Value        *Value
+	LeftIdentity []byte // the Identity of the value on the left side of the diff
```
ozkatz, line 165: Why is `Identity` different from a `Value`? why isn't the `Value`field called `LeftValue`? I'm fine with or without left, but I think it should be consistent between Value and Identity...

ozkatz (approved):
Approving, although I think the naming of Diff struct fields could be clearer..
Nice work!

graveler/graveler.go
johnnyaug: I'll try to explain, maybe you can help me find the right naming.
The "Value" field is typically the value on the right side of the diff (value includes both identity and data) - which is the value "after" the change.
I'm saying typically because in the case of deletion, there is no value on the right side, and in this case you will get the value before the deletion, which is left.

The left identity will give you the identity of the object before the change. It always comes for the left side, and in case of addition (where the left side is empty), you will get a nil array. This field is needed for merge (or any three way diff), where you want to know the identity before the change.

PR #1135, opened 2020-12-31
Handle tombstone in combined iterator and add tests

guy-har:


graveler/combined_iterator.go
```diff
@@ -46,7 +53,6 @@ func (c *combinedIterator) Next() bool {
 		// value of iterA < value of iterB
 		c.iterB.Next()
```
nopcoder, line 54: use nextA/nextB for all Next() calls

graveler/combined_iterator.go
```diff
@@ -18,23 +22,26 @@ func NewCombinedIterator(iterA, iterB ValueIterator) ValueIterator {
 	}
 }
 
-func (c *combinedIterator) Next() bool {
-	// call next with the relevant iterators
+// advanceInnerIterator advances the inner iterators and returns true if has more
+func (c *combinedIterator) advanceInnerIterator() bool {
```
nopcoder, line 26: optional: advanceInnerIterator -> advanceInnerIterators

graveler/combined_iterator.go
guy-har: changed, thanks

graveler/combined_iterator.go
guy-har: 👍

nopcoder (approved):

PR #1136, opened 2020-12-31
Change metarange key from minKey to MaxKey

guy-har:


nopcoder (approved):

arielshaqed (approved):
Neat!

Is there some way to test that we actually wrote the correct keys?  I mean, we changed the written key and didn't have to fix the test; this indicates maybe we're not testing the key properly.

PR #1138, opened 2021-01-01
Validate catalog arguments based on current mvcc cataloger

nopcoder:


catalog/rocks/entry_catalog.go
```diff
@@ -140,52 +200,138 @@ func (e *EntryCatalog) ListTags(ctx context.Context, repositoryID graveler.Repos
...
 }
 
 func (e *EntryCatalog) Merge(ctx context.Context, repositoryID graveler.RepositoryID, from graveler.Ref, to graveler.BranchID, committer string, message string, metadata graveler.Metadata) (graveler.CommitID, error) {
+	if err := Validate("repositoryID", repositoryID, ValidateRepositoryID); err != nil {
+		return "", err
+	}
+	if err := Validate("from", from, ValidateRef); err != nil {
+		return "", err
+	}
+	if err := Validate("to", to, ValidateRef); err != nil {
```
itaiad200, line 311:
```suggestion
	if err := Validate("to", to, ValidateBranchID); err != nil {
```

catalog/rocks/entry_catalog.go
```diff
@@ -140,52 +200,138 @@ func (e *EntryCatalog) ListTags(ctx context.Context, repositoryID graveler.Repos
...
+		return "", err
+	}
+	if err := Validate("committer", committer, ValidateNonEmptyString); err != nil {
+		return "", err
+	}
+	if err := Validate("message", message, ValidateNonEmptyString); err != nil {
+		return "", err
+	}
+	if err := Validate("message", message, ValidateNonEmptyString); err != nil {
```
itaiad200, line 320: Duplicated.

catalog/rocks/validate.go
```diff
@@ -0,0 +1,97 @@
...
+	"fmt"
+	"regexp"
+
+	"github.com/treeverse/lakefs/graveler"
+	"github.com/treeverse/lakefs/ident"
+)
+
+var (
+	validRefRegexp            = regexp.MustCompile(`^[^\s]+$`)
+	validBranchNameRegexp     = regexp.MustCompile(`^\w[-\w]*$`)
```
itaiad200, line 14: Let's use either `BranchName` or `BranchID` everywhere.

itaiad200: Also - isn't this a dup of `gravaler.reValidBranchID` & `graveler.reValidRepositoryID`?

catalog/rocks/entry_catalog.go
```diff
@@ -140,52 +200,138 @@ func (e *EntryCatalog) ListTags(ctx context.Context, repositoryID graveler.Repos
...
 }
 
 func (e *EntryCatalog) ResetKey(ctx context.Context, repositoryID graveler.RepositoryID, branchID graveler.BranchID, path Path) error {
+	if err := Validate("repositoryID", repositoryID, ValidateRepositoryID); err != nil {
+		return err
+	}
+	if err := Validate("branchID", branchID, ValidateBranchID); err != nil {
+		return err
+	}
+	if err := Validate("path", path.String(), ValidateNonEmptyString); err != nil {
```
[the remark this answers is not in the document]
nopcoder: added another check to validate 1024

catalog/rocks/validate.go
nopcoder: Moved them into this file

catalog/rocks/entry_catalog.go
nopcoder: Done

catalog/rocks/entry_catalog.go
```diff
@@ -140,52 +200,138 @@ func (e *EntryCatalog) ListTags(ctx context.Context, repositoryID graveler.Repos
...
+	if err := Validate("repositoryID", repositoryID, ValidateRepositoryID); err != nil {
+		return err
+	}
+	if err := Validate("branchID", branchID, ValidateBranchID); err != nil {
+		return err
+	}
+	if err := Validate("path", path, ValidatePath); err != nil {
+		return err
+	}
+	if err := Validate("path", path.String(), ValidateNonEmptyString); err != nil {
```
itaiad200, line 279: Why don't you use `ValidatePath` here?

catalog/rocks/validate.go
```diff
@@ -0,0 +1,113 @@
...
+
+var (
+	reValidRef          = regexp.MustCompile(`^[^\s]+$`)
+	reValidBranchID     = regexp.MustCompile(`^\w[-\w]*$`)
+	reValidRepositoryID = regexp.MustCompile(`^[a-z0-9][a-z0-9-]{2,62}$`)
+)
+
+var (
+	ErrInvalidValue = errors.New("invalid value")
+	ErrInvalidType  = errors.New("invalid type")
```
itaiad200, line 24: I think our errors are not very descriptive, especially if they propagate to the user.
e.g. Running `git commit` without an empty message ends up with:
`Aborting commit due to empty commit message.`
We'll write back:
`message: invalid value`

catalog/rocks/entry_catalog.go
```diff
@@ -202,6 +358,19 @@ func (e *EntryCatalog) Diff(ctx context.Context, repositoryID graveler.Repositor
...
+	if err := Validate("repositoryID", repositoryID, ValidateRepositoryID); err != nil {
+		return nil, err
+	}
+	if err := Validate("ref", ref, ValidateRef); err != nil {
+		return nil, err
+	}
+	if err := Validate("path", path, ValidatePath); err != nil {
+		return nil, err
+	}
+	if err := Validate("path", path.String(), ValidateNonEmptyString); err != nil {
```
itaiad200, line 370: Here and elsewhere: 
Calling `ValidateNonEmptyString` instead of `ValidatePath`.

itaiad200 (changes requested):
Requesting changes for the path validations, slightly because of the error descriptions.

catalog/rocks/entry_catalog.go
nopcoder: If you asked it means I need better naming.
ValidatePath validate that the path is valid - but not required, I will add another function that will call both.

catalog/rocks/validate.go
nopcoder: I'll return ErrRequiredValue in case value is required.
The InvalidType are bugs I can fail with panic if we think it will be a better way to handle it. WDYT?

itaiad200: Panics are better IMO.

itaiad200 (approved):

catalog/rocks/validate.go
```diff
@@ -20,35 +20,63 @@ var (
...
+			return fmt.Errorf("argument %s: %w", arg.Name, err)
+		}
 	}
 	return nil
 }
 
-func ValidateNonEmptyString(v interface{}) error {
-	s, ok := v.(string)
+func MakeValidateOptional(fn ValidateFunc) ValidateFunc {
+	return func(v interface{}) error {
```
itaiad200, line 47: nil `v` isn't an option?

itaiad200 (approved):
LGTM!

catalog/rocks/validate.go
nopcoder: for string or stringer no - but I'll add support

PR #1144, opened 2021-01-04
Simplify example Spark code

christophergrant:
This very small PR cleans up the example Spark code in the docs, with the following changes:

1. The header option for the DataframeReader with Parquet is ignored, and therefore unnecessary.
2. Running the toDF() method on an existing Dataframe is not necessary.


ozkatz (approved):
Makes sense. Thanks for contributing!

PR #1146, opened 2021-01-04
Pyramid disk allocation per FS instance

itaiad200:


catalog/rocks/entry_catalog.go
```diff
@@ -67,16 +69,37 @@ type EntryCatalog struct {
 	store graveler.Graveler
 }
 
+const (
+	rangeAllocationPercent     = 0.8
+	metaRangeAllocationPercent = 1 - rangeAllocationPercent
```
arielshaqed, line 74: These should be parameters in `config`.  Otherwise we need to release to change them.  Different values may be useful to different customers.  Also I think 0.8 is *much* too small -- and just having this discussion means it's hardly a constant.

arielshaqed: Also these are proportions not percents.

arielshaqed (changes requested):
Thanks!  Need to expose these in config.  Everything else is less important.

catalog/rocks/entry_catalog.go
itaiad200: Done.

config/config.go
```diff
@@ -349,20 +355,30 @@ func (c *Config) GetStatsFlushInterval() time.Duration {
...
-func (c *Config) GetCommittedTierFSParams() (*pyramidparams.Params, error) {
+func (c *Config) GetCommittedTierFSParams() (*pyramidparams.ExtParams, error) {
 	adapter, err := factory.BuildBlockAdapter(c)
 	if err != nil {
 		return nil, fmt.Errorf("build block adapter: %w", err)
 	}
+	rangePro := viper.GetFloat64(CommittedLocalCacheRangeProportion)
+	metaRangePro := viper.GetFloat64(CommittedLocalCacheMetaRangeProportion)
+
+	if rangePro+metaRangePro != 1 {
```
arielshaqed, line 366: This also makes life difficult, because 10 is divisble by 5 so nice ASCII floats don't sum exactly (e.g. https://play.golang.org/p/f3T-Ccruu7J is a bit scary).  Consider adding some slop, e.g.
```suggestion
	if something.Abs(rangePro+metaRangePro-1.0) > 1e-6 {
```
(I am not sure whether it is possible to create a nice example when summing just 2 numbers, though)

arielshaqed (approved):
Thanks!

PR #1151, opened 2021-01-05
Write downloaded file to temp location first for avoiding other readers inconsistencies

### LINKED ISSUE #1149: github.com/treeverse/lakefs/pyramid TestMultipleConcurrentReads flakes out
E.g.:
```sh
ariels@redqueen:~/dev/lakeFS$ go test --count=20 --race github.com/treeverse/lakefs/pyramid
--- FAIL: TestMultipleConcurrentReads (0.01s)
    tier_fs_test.go:175: 
                Error Trace:    tier_fs_test.go:206
                                                        tier_fs_test.go:175
                                                        asm_amd64.s:1374
                Error:          Not equal: 
                                expected: []byte{0x68, 0x65, 0x6c, 0x6c, 0x6f, 0x20, 0x77, 0x6f, 0x72, 0x6c, 0x64, 0x21}
                                actual  : []byte{}
                            
                                Diff:
                                --- Expected
                                +++ Actual
                                @@ -1,3 +1,2 @@
                                -([]uint8) (len=12) {
                                - 00000000  68 65 6c 6c 6f 20 77 6f  72 6c 64 21              |hello world!|
                                +([]uint8) {
                                 }
                Test:           TestMultipleConcurrentReads
FAIL
FAIL    github.com/treeverse/lakefs/pyramid     53.283s
FAIL
```

itaiad200:
closes #1149

### DIFF
--- a/pyramid/tier_fs.go
+++ b/pyramid/tier_fs.go
@@ -239,15 +239,14 @@ func (tfs *TierFS) openFile(fileRef localFileRef, fh *os.File) (*ROFile, error) 
 func (tfs *TierFS) openWithLock(fileRef localFileRef) (*os.File, error) {
 	_, err := tfs.keyLock.Compute(fileRef.filename, func() (interface{}, error) {
 		// check again file existence, now that we have the lock
-		fh, err := os.Open(fileRef.fullPath)
+		_, err := os.Stat(fileRef.fullPath)
 		if err == nil {
 			// file exists after all
 			cacheAccess.WithLabelValues(tfs.fsName, "Hit").Inc()
-			_ = fh.Close()
 			return nil, nil
 		}
 		if !os.IsNotExist(err) {
-			return nil, fmt.Errorf("open file: %w", err)
+			return nil, fmt.Errorf("stat file: %w", err)
 		}
 
 		// get the file from the block storage
@@ -266,7 +265,7 @@ func (tfs *TierFS) openWithLock(fileRef localFileRef) (*os.File, error) {
 
 		written, err := io.Copy(writer, reader)
 		if err != nil {
-			return nil, fmt.Errorf("copying date to file: %w", err)
+			return nil, fmt.Errorf("copying data to file: %w", err)
 		}
 		downloadHistograms.WithLabelValues(tfs.fsName).Observe(float64(written))
 
--- a/pyramid/tier_fs.go
+++ b/pyramid/tier_fs.go
@@ -190,6 +190,7 @@ func (tfs *TierFS) Open(namespace, filename string) (File, error) {
 		return nil, err
 	}
 
+	// check if file is there - without taking the lock
 	fileRef := tfs.newLocalFileRef(namespace, filename)
 	fh, err := os.Open(fileRef.fullPath)
 	if err == nil {
@@ -201,7 +202,7 @@ func (tfs *TierFS) Open(namespace, filename string) (File, error) {
 	}
 
 	cacheAccess.WithLabelValues(tfs.fsName, "Miss").Inc()
-	fh, err = tfs.readFromBlockStorage(fileRef)
+	fh, err = tfs.openWithLock(fileRef)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


pyramid/tier_fs.go
```diff
@@ -232,19 +233,33 @@ func (tfs *TierFS) openFile(fileRef localFileRef, fh *os.File) (*ROFile, error)
...
+func (tfs *TierFS) openWithLock(fileRef localFileRef) (*os.File, error) {
 	_, err := tfs.keyLock.Compute(fileRef.filename, func() (interface{}, error) {
-		var err error
+		// check again file existence, now that we have the lock
+		fh, err := os.Open(fileRef.fullPath)
+		if err == nil {
+			// file exists after all
+			cacheAccess.WithLabelValues(tfs.fsName, "Hit").Inc()
+			_ = fh.Close()
+			return nil, nil
```
arielshaqed, line 247: The file exists  and all is good; why not return it?
```suggestion
			return fh, nil
```

arielshaqed: In fact, I would expect this case to crash on the `tfs.openFile` call.

pyramid/tier_fs.go
```diff
@@ -253,12 +268,18 @@ func (tfs *TierFS) readFromBlockStorage(fileRef localFileRef) (*os.File, error)
 		if err != nil {
 			return nil, fmt.Errorf("copying date to file: %w", err)
```
arielshaqed, line 269: nit:
```suggestion
			return nil, fmt.Errorf("copying data to file: %w", err)
```

pyramid/tier_fs.go
```diff
@@ -253,12 +268,18 @@ func (tfs *TierFS) readFromBlockStorage(fileRef localFileRef) (*os.File, error)
...
 		}
-		downloadHistograms.WithLabelValues(tfs.fsName).Observe(float64(written))
-		return nil, err
+
+		// copy from temp path to actual path
+		if err = tfs.syncDir.renameFile(tmpFullPath, fileRef.fullPath); err != nil {
+			return nil, fmt.Errorf("rename temp file: %w", err)
+		}
+
+		return nil, nil
```
arielshaqed, line 282:
```suggestion
		return fh, nil
```

arielshaqed: Again, this is really a race.  On line 289 you *again* try to open the file.  But the file might already have been removed.  This is rare, but easy enough to fix: keep the return value from `keyLock.Compute` and return it here if possible.

arielshaqed (changes requested):
Small race remains between opening the file and re-opening it.  Also that's inefficient -- but I think we can fix it easily enough.

pyramid/tier_fs.go
itaiad200: Done.

pyramid/tier_fs.go
itaiad200: I agree that it's a better habit to return the file-handler and I'll fix it.
However notice it's not a race. The file wasn't stored in the eviction cache yet, it happens in the `openFile` func.
So file wasn't stored in the eviction -> can't be evicted -> file is always there.

pyramid/tier_fs.go
itaiad200: See discussion below.

pyramid/tier_fs.go
itaiad200: Actually, after applying the change, it is wrong.
`Compute` only calls the function once for concurrent calls, so returning the same file-handler isn't right.

arielshaqed: Yikes!  True...  So the way to do this would be to keep an external `var fh *os.File`, and set that here (and still `return nil, nil`), then use that `fh` if not `nil` after this `Compute` call.  Which is unpleasant.  (oh, and I forgot to rewind the file before returning, of course).
I still don't like this double-open (you open a file twice when you download it), but I do not see a race to cause trouble here - just some inefficiency.

arielshaqed (approved):
THANKS!  Good change.

PR #1152, opened 2021-01-05
Generate BatchWriterClosers as needed, not via MetaRangeManager

arielshaqed:
There is no dependency on RangeManager.  Removing it makes it easier to write a
MetaRangeManager and *improves* the current tests -- which can now use a proper
BatchWriterCloser rather than a mock.

Remove TestWriter_SortOnClose: it tests some correctness.  But it is quite hard
to test correctly, and ends up only verifying that one `sort.Slice` line exists
in the code.  Better test this during integration.

### DIFF
--- a/graveler/committed/batch.go
+++ b/graveler/committed/batch.go
@@ -1,14 +1,16 @@
-package sstable
+package committed
 
 import (
 	"errors"
 	"sync"
-
-	"github.com/treeverse/lakefs/graveler/committed"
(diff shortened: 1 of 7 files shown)


guy-har (approved):
Looks great
Thanks

itaiad200 (approved):

arielshaqed (commented):
Thanks!

PR #1157, opened 2021-01-06
Remove unused variable in test code

nopcoder:
- remove unused variable in test code
- use deep equal to compare diff keys


johnnyaug (approved):
Thanks!!

PR #1160, opened 2021-01-06
Graveler diagnostics

itaiad200:
Still missing:
- Graveler composition to test it.

### DIFF
--- a/block/gs/adapter.go
+++ b/block/gs/adapter.go
@@ -121,6 +121,29 @@ func (a *Adapter) Get(obj block.ObjectPointer, _ int64) (io.ReadCloser, error) {
 	return r, err
 }
 
+func (a *Adapter) List(storageNamespace, prefix string) ([]string, error) {
+	var err error
+	defer reportMetrics("List", time.Now(), nil, &err)
+
+	if err != nil {
+		return nil, err
+	}
+
+	var keys []string
+	iter := a.client.Bucket(storageNamespace).Objects(context.Background(), &storage.Query{Prefix: prefix})
+	for {
+		attrs, err := iter.Next()
+		if errors.Is(err, iterator.Done) {
+			break
+		}
+		if err != nil {
+			return nil, fmt.Errorf("bucket(%s).Objects(): %w", storageNamespace, err)
+		}
+		keys = append(keys, attrs.Name)
+	}
+	return keys, nil
+}
+
 func (a *Adapter) Exists(obj block.ObjectPointer) (bool, error) {
 	var err error
 	defer reportMetrics("Exists", time.Now(), nil, &err)
--- a/block/gs/adapter.go
+++ b/block/gs/adapter.go
@@ -88,6 +88,17 @@ func resolveNamespace(obj block.ObjectPointer) (block.QualifiedKey, error) {
 	return qualifiedKey, nil
 }
 
+func resolveNamespacePrefix(lsOpts block.ListOpts) (block.QualifiedPrefix, error) {
+	qualifiedPrefix, err := block.ResolveNamespacePrefix(lsOpts.StorageNamespace, lsOpts.Prefix)
+	if err != nil {
+		return qualifiedPrefix, err
+	}
+	if qualifiedPrefix.StorageType != block.StorageTypeGS {
+		return qualifiedPrefix, block.ErrInvalidNamespace
+	}
+	return qualifiedPrefix, nil
+}
+
 func (a *Adapter) Put(obj block.ObjectPointer, sizeBytes int64, reader io.Reader, _ block.PutOpts) error {
 	var err error
 	defer reportMetrics("Put", time.Now(), &sizeBytes, &err)
@@ -121,23 +132,23 @@ func (a *Adapter) Get(obj block.ObjectPointer, _ int64) (io.ReadCloser, error) {
 	return r, err
 }
 
-func (a *Adapter) List(storageNamespace, prefix string) ([]string, error) {
+func (a *Adapter) List(lsOpt block.ListOpts) ([]string, error) {
 	var err error
 	defer reportMetrics("List", time.Now(), nil, &err)
-
+	qualifiedPrefix, err := resolveNamespacePrefix(lsOpt)
 	if err != nil {
 		return nil, err
 	}
 
 	var keys []string
-	iter := a.client.Bucket(storageNamespace).Objects(context.Background(), &storage.Query{Prefix: prefix})
+	iter := a.client.Bucket(qualifiedPrefix.StorageNamespace).Objects(context.Background(), &storage.Query{Prefix: qualifiedPrefix.Prefix})
 	for {
 		attrs, err := iter.Next()
 		if errors.Is(err, iterator.Done) {
 			break
 		}
 		if err != nil {
-			return nil, fmt.Errorf("bucket(%s).Objects(): %w", storageNamespace, err)
+			return nil, fmt.Errorf("bucket(%s).Objects(): %w", qualifiedPrefix.StorageNamespace, err)
 		}
 		keys = append(keys, attrs.Name)
 	}
--- a/block/gs/adapter.go
+++ b/block/gs/adapter.go
@@ -88,7 +88,7 @@ func resolveNamespace(obj block.ObjectPointer) (block.QualifiedKey, error) {
 	return qualifiedKey, nil
 }
 
-func resolveNamespacePrefix(lsOpts block.ListOpts) (block.QualifiedPrefix, error) {
+func resolveNamespacePrefix(lsOpts block.WalkOpts) (block.QualifiedPrefix, error) {
 	qualifiedPrefix, err := block.ResolveNamespacePrefix(lsOpts.StorageNamespace, lsOpts.Prefix)
 	if err != nil {
 		return qualifiedPrefix, err
@@ -132,27 +132,31 @@ func (a *Adapter) Get(obj block.ObjectPointer, _ int64) (io.ReadCloser, error) {
 	return r, err
 }
 
-func (a *Adapter) List(lsOpt block.ListOpts) ([]string, error) {
+func (a *Adapter) Walk(walkOpt block.WalkOpts, walkFn block.WalkFunc) error {
 	var err error
-	defer reportMetrics("List", time.Now(), nil, &err)
-	qualifiedPrefix, err := resolveNamespacePrefix(lsOpt)
+	defer reportMetrics("Walk", time.Now(), nil, &err)
+	qualifiedPrefix, err := resolveNamespacePrefix(walkOpt)
 	if err != nil {
-		return nil, err
+		return err
 	}
 
-	var keys []string
 	iter := a.client.Bucket(qualifiedPrefix.StorageNamespace).Objects(context.Background(), &storage.Query{Prefix: qualifiedPrefix.Prefix})
+
 	for {
 		attrs, err := iter.Next()
+
 		if errors.Is(err, iterator.Done) {
 			break
 		}
 		if err != nil {
-			return nil, fmt.Errorf("bucket(%s).Objects(): %w", qualifiedPrefix.StorageNamespace, err)
+			return fmt.Errorf("bucket(%s).Objects(): %w", qualifiedPrefix.StorageNamespace, err)
 		}
-		keys = append(keys, attrs.Name)
+
+		if err := walkFn(attrs.Name); err != nil {
+			return err
+		}
 	}
-	return keys, nil
+	return nil
 }
 
 func (a *Adapter) Exists(obj block.ObjectPointer) (bool, error) {
--- a/block/local/adapter.go
+++ b/block/local/adapter.go
@@ -73,6 +73,7 @@ func NewAdapter(path string, opts ...func(a *Adapter)) (*Adapter, error) {
 	}
 	return adapter, nil
 }
+
 func resolveNamespace(obj block.ObjectPointer) (block.QualifiedKey, error) {
 	qualifiedKey, err := block.ResolveNamespace(obj.StorageNamespace, obj.Identifier)
 	if err != nil {
@@ -172,15 +173,15 @@ func (l *Adapter) Get(obj block.ObjectPointer, _ int64) (reader io.ReadCloser, e
 	return f, nil
 }
 
-func (l *Adapter) List(storageNamespace, prefix string) ([]string, error) {
-	p := path.Join(l.path, storageNamespace, prefix)
+func (l *Adapter) List(lsOpt block.ListOpts) ([]string, error) {
+	p := filepath.Clean(path.Join(l.path, lsOpt.StorageNamespace, lsOpt.Prefix))
 	files, err := ioutil.ReadDir(p)
 	if errors.Is(err, os.ErrNotExist) {
 		// no files under prefix
 		return nil, nil
 	}
 	if err != nil {
-		return nil, fmt.Errorf("stat on a dir path %s: %w", p, err)
+		return nil, fmt.Errorf("read dir path %s: %w", p, err)
 	}
 
 	var keys []string
--- a/block/local/adapter.go
+++ b/block/local/adapter.go
@@ -172,6 +172,24 @@ func (l *Adapter) Get(obj block.ObjectPointer, _ int64) (reader io.ReadCloser, e
 	return f, nil
 }
 
+func (l *Adapter) List(storageNamespace, prefix string) ([]string, error) {
+	p := path.Join(l.path, storageNamespace, prefix)
+	files, err := ioutil.ReadDir(p)
+	if errors.Is(err, os.ErrNotExist) {
+		// no files under prefix
+		return nil, nil
+	}
+	if err != nil {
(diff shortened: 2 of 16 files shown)


block/gs/adapter.go
```diff
@@ -121,6 +121,29 @@ func (a *Adapter) Get(obj block.ObjectPointer, _ int64) (io.ReadCloser, error) {
 	return r, err
 }
 
+func (a *Adapter) List(storageNamespace, prefix string) ([]string, error) {
+	var err error
+	defer reportMetrics("List", time.Now(), nil, &err)
+
+	if err != nil {
+		return nil, err
+	}
```
nopcoder, line 130: what do we check?

block/gs/adapter.go
```diff
@@ -121,6 +121,29 @@ func (a *Adapter) Get(obj block.ObjectPointer, _ int64) (io.ReadCloser, error) {
...
+
+	var keys []string
+	iter := a.client.Bucket(storageNamespace).Objects(context.Background(), &storage.Query{Prefix: prefix})
```
nopcoder, line 133: Missing namespace resolve.
Every adapter today (which I think we can consolidate) resolves the object path using the given namespace and turn it into the actual storage namespace (aka bucket) - namespace in overloaded.
If you use the repository namespace directly it will not work as it will include the bucket name with an optional path, which in this case will fail.

block/gs/adapter.go
```diff
@@ -121,6 +121,29 @@ func (a *Adapter) Get(obj block.ObjectPointer, _ int64) (io.ReadCloser, error) {
 	return r, err
 }
 
+func (a *Adapter) List(storageNamespace, prefix string) ([]string, error) {
```
nopcoder, line 124: If we supply a listing interface the does a full scan, we will get a memory issue at some point.
Think that if we will accept a callback (with option return value to stop processing) in this case it cover the count use-case future requirements without hitting any memory issues.

block/local/adapter.go
```diff
@@ -172,6 +172,24 @@ func (l *Adapter) Get(obj block.ObjectPointer, _ int64) (reader io.ReadCloser, e
 	return f, nil
 }
 
+func (l *Adapter) List(storageNamespace, prefix string) ([]string, error) {
+	p := path.Join(l.path, storageNamespace, prefix)
+	files, err := ioutil.ReadDir(p)
```
nopcoder, line 177: wrap p with filepath.Clean - or the individuals: storage namespace and prefix

block/local/adapter.go
```diff
@@ -172,6 +172,24 @@ func (l *Adapter) Get(obj block.ObjectPointer, _ int64) (reader io.ReadCloser, e
...
+	if errors.Is(err, os.ErrNotExist) {
+		// no files under prefix
+		return nil, nil
+	}
+	if err != nil {
+		return nil, fmt.Errorf("stat on a dir path %s: %w", p, err)
```
nopcoder, line 183: read dir instead of stat

block/transient/adapter.go
```diff
@@ -69,6 +69,10 @@ func (a *Adapter) Copy(_, _ block.ObjectPointer) error {
 	return nil
 }
 
+func (a *Adapter) List(storageNamespace, prefix string) ([]string, error) {
+	return nil, nil
```
nopcoder, line 73: It should be something like this - 
```go
defer a.mutex.RUnlock()
data, ok := a.data[getKey(obj)]
p := storageNamespace + ":" + prefix
var keys []strings 
for _, k := range a.data {
	if strings.HasPrefix(k, p) {
		keys = append(keys, k[len(storageNamespace)+1:])
	}
}
return keys, nil
```

catalog/rocks/entry_catalog.go
```diff
@@ -65,14 +65,19 @@ type EntryCatalog struct {
 	store graveler.Graveler
 }
 
+const (
+	RangeFSName     = "range"
+	MetaRangeFSName = "meta-range"
```
nopcoder, line 70: 1. We don't need these defaults in graveler?
2. Didn't discuss it - but we need to choose between metarange and meta-range?

cmd/lakefs/cmd/diagnostics.go
```diff
@@ -5,6 +5,9 @@ import (
 	"log"
 	"os"
 
+	"github.com/treeverse/lakefs/block/factory"
+	catalogfactory "github.com/treeverse/lakefs/catalog/factory"
+
 	"github.com/spf13/cobra"
```
nopcoder, line 11: remove space

cmd/lakefs/cmd/diagnostics.go
```diff
@@ -20,8 +23,16 @@ var diagnosticsCmd = &cobra.Command{
 
 		dbPool := db.BuildDatabaseConnection(cfg.GetDatabaseParams())
 		defer dbPool.Close()
+		adapter, err := factory.BuildBlockAdapter(cfg)
+		if err != nil {
+			log.Print("Failed to create block adapter")
```
nopcoder, line 28: add the 'err'

cmd/lakefs/cmd/diagnostics.go
```diff
@@ -20,8 +23,16 @@ var diagnosticsCmd = &cobra.Command{
...
+		}
+		cataloger, err := catalogfactory.BuildCataloger(dbPool, cfg)
+		if err != nil {
+			log.Print("Failed to create cataloger")
```
nopcoder, line 32: add the 'err'

diagnostics/block.go
```diff
@@ -0,0 +1,116 @@
...
+
+	"github.com/treeverse/lakefs/catalog/rocks"
+
+	"github.com/treeverse/lakefs/config"
+
+	"github.com/treeverse/lakefs/block"
+
+	"github.com/treeverse/lakefs/db"
+
+	"github.com/treeverse/lakefs/catalog"
```
nopcoder, line 21: remove blank lines

diagnostics/block.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+// DBCollector collects diagnostics information and write the collected content into a writer in a zip format
+type BlockCollector struct {
```
nopcoder, line 25: No correlation between file <-> code <-> code doc

diagnostics/block.go
```diff
@@ -0,0 +1,116 @@
...
+	repos, _, err := c.cataloger.ListRepositories(ctx, -1, "")
+	if err != nil {
+		// Cannot list repos, nothing to do..
+		errs = append(errs, fmt.Errorf("listing repositories: %w", err))
+		return errs
+	}
+
+	tierFSParams, err := c.cfg.GetCommittedTierFSParams()
+	if err != nil {
+		errs = append(errs, fmt.Errorf("listing repositories: %w", err))
```
nopcoder, line 69: get tierfs params

diagnostics/block.go
```diff
@@ -0,0 +1,116 @@
...
+		return errs
+	}
+
+	rangesFile, err := writer.Create("graveler-ranges")
+	if err != nil {
+		errs = append(errs, fmt.Errorf("creating ranges file: %w", err))
+		return errs
```
nopcoder, line 76: why do we return in this case?

diagnostics/block.go
```diff
@@ -0,0 +1,116 @@
...
+	dbCollector *DBCollector
+}
+
+// NewDBCollector accepts database to work with during collect
+func NewBlockCollector(adb db.Database, cataloger catalog.Cataloger, cfg *config.Config, adapter block.Adapter) *BlockCollector {
+	return &BlockCollector{
+		cataloger:   cataloger,
+		cfg:         cfg,
+		adapter:     adapter,
+		dbCollector: NewDBCollector(adb),
```
nopcoder, line 38: why one collector holds another?

diagnostics/factory.go
```diff
@@ -0,0 +1,22 @@
...
+
+type Collector interface {
+	Collect(ctx context.Context, w io.Writer) (err error)
+}
+
+func CreateCollector(adb db.Database, cataloger catalog.Cataloger, cfg *config.Config, adapter block.Adapter) Collector {
+	if cataloger == nil {
+		return NewDBCollector(adb)
+	}
+	return NewBlockCollector(adb, cataloger, cfg, adapter)
```
nopcoder, line 21: thought we are going to collect everything we will find

go.mod
```diff
@@ -28,6 +29,7 @@ require (
 	github.com/go-openapi/validate v0.20.0
 	github.com/go-swagger/go-swagger v0.25.0
 	github.com/go-test/deep v1.0.7
+	github.com/gobuffalo/logger v0.0.0-20190315122211-86e12af44bc2
```
nopcoder, line 32: do we need this one?

diagnostics/factory.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/treeverse/lakefs/catalog"
+	"github.com/treeverse/lakefs/config"
+	"github.com/treeverse/lakefs/db"
+)
+
+type Collector interface {
+	Collect(ctx context.Context, w io.Writer) (err error)
+}
+
+func CreateCollector(adb db.Database, cataloger catalog.Cataloger, cfg *config.Config, adapter block.Adapter) Collector {
```
nopcoder, line 17: Not sure we need factory - we can have one collect implementation and return it.
The 'if' that checks what needs to collect and be inside the 'Collect' implementation.
WDYT?

nopcoder (changes requested):

block/gs/adapter.go
itaiad200: I see, good catch. Fixed.

block/local/adapter.go
itaiad200: Done.

block/local/adapter.go
itaiad200: Done.

block/transient/adapter.go
itaiad200: Done.

catalog/rocks/entry_catalog.go
itaiad200: 1. Nope - You only need them to create the FS. Diagnostics needs them to read the blockstorage prefix.
2. Didn't get what you meant.

cmd/lakefs/cmd/diagnostics.go
itaiad200: Done.

cmd/lakefs/cmd/diagnostics.go
itaiad200: Done.

cmd/lakefs/cmd/diagnostics.go
itaiad200: Done.

diagnostics/block.go
itaiad200: Done.

diagnostics/block.go
itaiad200: Done.

diagnostics/block.go
itaiad200: So that the `DBCollector` be a standalone unit like today, and the `BlockCollector` wraps it and adds graveler statistics on top of it.

diagnostics/block.go
itaiad200: Done.

diagnostics/block.go
itaiad200: We failed to create the file that holds the data. There's no point in collecting that data if there's no where to write it to.

diagnostics/factory.go
itaiad200: My main reason for doing it is to change the existing functionality as little as I could.
One design pattern or the other is not a big deal IMO, I will change if you insist.

diagnostics/factory.go
itaiad200: From the DB yes, I see no value in scanning empty prefixes in the block-storage. Do you think otherwise?

go.mod
itaiad200: Nope, removed.

catalog/rocks/entry_catalog.go
nopcoder: 1. cool
2. there are places (not in your code) where we call it "metarange" and places "meta-range" :)

block/gs/adapter.go
nopcoder: I prefer to have something like Walk or an Iterator and will use it later if needed

diagnostics/factory.go
nopcoder: Cool

block/mem/adapter.go
```diff
@@ -230,8 +235,18 @@ func (a *Adapter) CompleteMultiPartUpload(obj block.ObjectPointer, uploadID stri
 	return &hexCode, int64(len(data)), nil
 }
 
-func (a *Adapter) List(_, _ string) ([]string, error) {
-	return nil, nil
+func (a *Adapter) List(lsOpts block.ListOpts) ([]string, error) {
+	a.mutex.Lock()
```
nopcoder, line 239: RLock

itaiad200: Done.

block/gs/adapter.go
```diff
@@ -132,27 +132,31 @@ func (a *Adapter) Get(obj block.ObjectPointer, _ int64) (io.ReadCloser, error) {
...
 		if errors.Is(err, iterator.Done) {
 			break
 		}
 		if err != nil {
-			return nil, fmt.Errorf("bucket(%s).Objects(): %w", qualifiedPrefix.StorageNamespace, err)
+			return fmt.Errorf("bucket(%s).Objects(): %w", qualifiedPrefix.StorageNamespace, err)
+		}
+
+		if err := walkFn(attrs.Name); err != nil {
+			return fmt.Errorf("walkFn failed: %w", err)
```
nopcoder, line 156: can we return the original error in this case? it will enable the user of the able to stop the processing by returning EOF - and I don't think we need to wrap in this case.

nopcoder: (relevant to all adapters)

block/local/adapter.go
```diff
@@ -173,22 +173,22 @@ func (l *Adapter) Get(obj block.ObjectPointer, _ int64) (reader io.ReadCloser, e
...
-	if errors.Is(err, os.ErrNotExist) {
-		// no files under prefix
-		return nil, nil
-	}
-	if err != nil {
-		return nil, fmt.Errorf("read dir path %s: %w", p, err)
-	}
+func (l *Adapter) Walk(walkOpt block.WalkOpts, walkFn block.WalkFunc) error {
+	p := filepath.Clean(path.Join(l.path, walkOpt.StorageNamespace, walkOpt.Prefix))
+	if err := filepath.Walk(p, func(path string, info os.FileInfo, err error) error {
```
nopcoder, line 178: no need to check the err - return the value by filepath.Walk

block/s3/adapter.go
```diff
@@ -319,37 +319,47 @@ func (a *Adapter) GetRange(obj block.ObjectPointer, startPosition int64, endPosi
 	return objectOutput.Body, nil
 }
 
-func (a *Adapter) List(lsOpt block.ListOpts) ([]string, error) {
+func (a *Adapter) Walk(walkOpt block.WalkOpts, walkFn block.WalkFunc) error {
 	log := a.log().WithField("operation", "ListObjects")
 	var err error
 	var lenRes int64
 	defer reportMetrics("GetRange", time.Now(), &lenRes, &err)
```
nopcoder, line 326: "Walk"

block/s3/adapter.go
```diff
@@ -319,37 +319,47 @@ func (a *Adapter) GetRange(obj block.ObjectPointer, startPosition int64, endPosi
 	return objectOutput.Body, nil
 }
 
-func (a *Adapter) List(lsOpt block.ListOpts) ([]string, error) {
+func (a *Adapter) Walk(walkOpt block.WalkOpts, walkFn block.WalkFunc) error {
 	log := a.log().WithField("operation", "ListObjects")
```
nopcoder, line 323: Walk

block/gs/adapter.go
itaiad200: Done.

nopcoder (commented):
Posted all my comments, not in bluck - I had to stop and didn't knew when I'll continue.

block/local/adapter.go
```diff
@@ -172,6 +173,21 @@ func (l *Adapter) Get(obj block.ObjectPointer, _ int64) (reader io.ReadCloser, e
...
+		if err != nil {
+			return err
+		}
+
+		return walkFn(p)
+	}); err != nil {
+		return err
+	}
+
+	return nil
```
nopcoder, line 188:
```suggestion
	return filepath.Walk(p, func(path string, info os.FileInfo, err error) error {
		return walkFn(p)
	})
```

nopcoder (approved):
minor comment

block/local/adapter.go
itaiad200: I still want to handle the inner error, the wrapped code is Done.

PR #1161, opened 2021-01-06
Fix disk writer should get a new hash for each file

### LINKED ISSUE #1159: Graveler's sstable manager share hash
Graveler's sstable manager holds hash which holds a state and share it each NewDiskWriter.
Because hash holds a state this can't be share.

nopcoder:
Fix #1159

### DIFF
--- a/graveler/sstable/manager.go
+++ b/graveler/sstable/manager.go
@@ -2,9 +2,9 @@ package sstable
 
 import (
 	"bytes"
+	"crypto"
 	"errors"
 	"fmt"
-	"hash"
 
 	"github.com/treeverse/lakefs/graveler/committed"
 	"github.com/treeverse/lakefs/logging"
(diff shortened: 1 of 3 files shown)


arielshaqed (approved):
Good catch!

A unit test cannot easily detect issues like this (loss of catalog dedupe).  Wondering how to automate testing this in an ongoing manner.  Best I can see is to run specific import scenarios and track disk usage.

graveler/sstable/manager.go
```diff
@@ -16,10 +16,10 @@ type Manager struct {
 	fs    pyramid.FS
 	// TODO(ariels): Replace with loggers constructed from context.
 	logger logging.Logger
-	hash   hash.Hash
+	hash   crypto.Hash
```
arielshaqed, line 19: Amusingly, neither of `{hash,crypto}.Hash` is the other: `crypto.Hash` might better be called `crypto.HashAlgorithm`, but that may be 10 years too late. ;-(

itaiad200 (approved):

graveler/sstable/manager.go
nopcoder: guess they like it simple - I liked the fact that my `int` can create an hash instance

nopcoder:
> Good catch!
> 
> A unit test cannot easily detect issues like this (loss of catalog dedupe). Wondering how to automate testing this in an ongoing manner. Best I can see is to run specific import scenarios and track disk usage.

Think I can write something that the race detector will catch.

PR #1166, opened 2021-01-07
Separate S3 gateway to various http handlers

johnnyaug:


nopcoder (changes requested):

gateway/handler.go
```diff
@@ -88,9 +87,13 @@ func NewHandler(
...
 ) http.Handler {
-	var fallbackProxy *gohttputil.ReverseProxy
+	var fallbackHandler http.Handler
 	if fallbackURL != nil {
-		fallbackProxy = gohttputil.NewSingleHostReverseProxy(fallbackURL)
+		fallbackProxy := gohttputil.NewSingleHostReverseProxy(fallbackURL)
+		fallbackHandler = http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
+			request.Host = strings.Replace(request.Host, bareDomain, fallbackURL.Host, 1)
+			fallbackProxy.ServeHTTP(writer, request)
+		})
```
nopcoder, line 96: We model this one as a handler after the repo enrich - in case there is no repo it will call the proxy (if configured)

nopcoder (approved):

PR #1167, opened 2021-01-08
Allow manual triggering of benchmark and nessie CI workflows

arielshaqed:
These are not hermetic, re-triggering makes sense.

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -3,6 +3,7 @@ on:
   push:
     branches:
       - master
+  workflow_dispatch:
 
 jobs:
   deploy-executor-image:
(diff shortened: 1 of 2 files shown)


itaiad200:
Not sure what you're aiming for. You can rerun a workflow today (Actions -> Nessie -> Picking a run -> "Re-run jobs".
IIUC this `workflow_dispatch` will allow you to trigger them for master branch, but you can still get that by re-triggering that job.

itaiad200 (changes requested):
Not sure what you're aiming for. You can rerun a workflow today (Actions -> Nessie -> Picking a run -> "Re-run jobs".
IIUC this workflow_dispatch will allow you to trigger them for master branch, but you can still get that by re-triggering that job.

arielshaqed:
Insufficient traction to do this.  Will try again when we have more CD.

PR #1169, opened 2021-01-10
Entry catalog enable first commit and listing using graveler

nopcoder:


graveler/graveler.go
```diff
@@ -960,15 +960,19 @@ func (g *graveler) DiffUncommitted(ctx context.Context, repositoryID RepositoryI
 	if err != nil {
 		return nil, err
 	}
+	var metaRangeID MetaRangeID
 	commit, err := g.RefManager.GetCommit(ctx, repositoryID, branch.CommitID)
```
arielshaqed, line 964: I think in this case `branch.CommitID == ""`.  We could check that, and then be able to give a meaningful internal-ish error message if `branch.CommitID != "" && errors.Is(err, ErrCommitNotFound)`.

graveler/uncommitted_diff_iterator.go
```diff
@@ -17,6 +17,7 @@ type uncommittedDiffIterator struct {
 	ctx              context.Context
 }
 
+// NewUncommittedDiffIterator list uncommitted changes as diff. The `metaRangeID` can be empty in case there is no commit
```
arielshaqed, line 20:
```suggestion
// NewUncommittedDiffIterator lists uncommitted changes as a diff. If `metaRangeID` is empty then there is no commit and it returns all objects as added.
```

arielshaqed (approved):
Cool, thanks!

graveler/graveler.go
nopcoder: Added branch's commit ID check before loading the commit information

cache/reference_counted.go
```diff
@@ -35,17 +34,25 @@ type ParamsWithDisposal struct {
...
+const (
+	DefaultCacheWithDisposalShards = 1
+	DefaultCacheWithDisposalSize   = 1000
+)
+
 func NewCacheWithDisposal(p ParamsWithDisposal) *shardedCacheWithDisposal {
 	if p.Shards <= 0 {
-		panic("need at least 1 shard")
+		p.Shards = DefaultCacheWithDisposalShards
+	}
```
arielshaqed, line 53: This is almost always wrong.  Instead, we create from configuration and force a minimum *there*.  (I made that change concurrently with your branch, so it wasn't there initially).
E.g. 0 might be better used as a magic value that says "don't cache".

graveler/committed/value_iterator.go
```diff
@@ -0,0 +1,45 @@
...
+	it Iterator
+}
+
+func (v *valueIterator) Next() bool {
+	for {
+		r := v.it.Next()
+		if !r {
+			return false
+		}
+		// skip ranges
```
arielshaqed, line 15:
```suggestion
		// skip range headers
```

graveler/committed/value_iterator.go
```diff
@@ -0,0 +1,45 @@
...
+		_, rng := v.it.Value()
+		if rng == nil {
+			break
+		}
+	}
```
arielshaqed, line 20: This doesn't match the interface:
```go
	// Value returns a nil ValueRecord and a Range before starting a Range, or a Value and
	// that Range when inside a Range.
```
It should look like this (but I haven't tested it):
```suggestion
	for {
		r := v.it.Next()
		if !r {
			return false
		}
		// skip ranges
		val := v.it.Value()
		if val != nil {
			break
		}
	}
```

arielshaqed: Or easier for me to read perhaps:
```go
for v.it.Next() {
	if val := v.it.Value(); val != nil {
		return true
	}
}
return false
```

graveler/graveler.go
```diff
@@ -55,7 +56,7 @@ type Ref string
 // TagID represents a named tag pointing at a commit
 type TagID string
 
-// CommitParents commit's parents slice
+// CommitParents commit parents slice
```
arielshaqed, line 59: Not sure this comment adds information :-(

arielshaqed (changes requested):
Thanks!

I'm worried that the "skip range headers" iterator seems wrong, yet tests passed.  Can we test it (directly or indirectly)?

cache/reference_counted.go
nopcoder: removed the default size and added no cache option

graveler/committed/value_iterator.go
nopcoder: Like this one!

graveler/graveler.go
nopcoder: Agree

config/config.go
```diff
@@ -37,6 +38,10 @@ const (
 	DefaultCommittedLocalCacheMetaRangePercent = 0.1
 	DefaultCommittedLocalCacheBytes            = 1 * 1024 * 1024 * 1024
 	DefaultCommittedLocalCacheDir              = "~/lakefs/local_tier"
```
arielshaqed: Can you please make sure we're not somehow getting rid of the tilde-expansion fix that I added?

gateway/handler.go
```diff
@@ -94,284 +105,168 @@ func NewHandler(
...
 	}
 
 	// setup routes
 	var h http.Handler
 	h = &handler{
 		BareDomain:         bareDomain,
 		sc:                 sc,
-		NotFoundHandler:    http.HandlerFunc(notFound),
 		ServerErrorHandler: nil,
+		operationHandlers: map[operations.OperationID]http.Handler{
```
arielshaqed: This is not part of this PR, why are we including it?  Merge vs. rebase issue?

gateway/handler_test.go
```diff
@@ -1,13 +1,14 @@
...
 import (
-	"github.com/stretchr/testify/assert"
-	"github.com/treeverse/lakefs/gateway/simulator"
 	"io"
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
```
arielshaqed: nit: blank line between 3rd-party and local imports.

graveler/sstable/cache.go
```diff
@@ -53,21 +53,26 @@ func (i *item) Close() error {
...
-			if err != nil {
-				return nil, fmt.Errorf("open SSTable %s: %w", id, err)
-			}
-			return &item{reader}, nil
-		},
-		fs.Exists,
-	)
+	opener := makePyramidOpener(fs, readerOptions)
+	if p.Size == 0 {
+		return &noCache{opener: opener}
```
arielshaqed, line 58: Do you still want to add `noCache`?

graveler/sstable/cache.go
```diff
@@ -108,3 +113,20 @@ func (c *lruCache) GetOrOpen(namespace string, id committed.ID) (*sstable.Reader
...
+	item, err := n.opener(namespace, string(id))
+	if err != nil {
+		return nil, nil, err
+	}
+	reader := item.GetSSTable()
+	return reader, reader.Close, nil
+}
+
+func (n *noCache) Exists(namespace string, id committed.ID) (bool, error) {
+	panic("implement me")
```
arielshaqed, line 131: This *will* be called, and then fail.

pyramid/tier_fs_test.go
```diff
@@ -161,18 +156,19 @@ func TestMultipleConcurrentReads(t *testing.T) {
...
 		return nil
-	}))
+	})
+	require.NoError(t, err)
 	// try to read that file - only a single access to block storage is expected
-	concurrencyLevel := 50
+	const concurrencyLevel = 50
 	adapter.wait = make(chan struct{})
 	var wg sync.WaitGroup
+	wg.Add(concurrencyLevel)
```
arielshaqed, line 170: nit: personally prefer multiple `wg.Add(1)` calls, they are (sometimes) less brittle during test maintenance.

arielshaqed (approved):
Great stuff, thanks!

I'm pretty sure `noCache.Exists` *will* get called, so you *must* implement it (unless you remove it, of course).

graveler/sstable/cache.go
nopcoder: fixed

graveler/sstable/cache.go
nopcoder: can remove it - needed something to work.  WDYT?

pyramid/tier_fs_test.go
nopcoder: good point. sometimes I just think about performance.

PR #1171, opened 2021-01-11
Fix unexpanded tildes in cache dirs

arielshaqed:
Makes defaults make sense, so you can run `lakefs` in rocks mode by setting just `cataloger.type: rocks`.


guy-har (approved):
looks good
Thanks

arielshaqed:
Thanks -- silly bug, caused me lots of grief.

PR #1176, opened 2021-01-11
Committed manager get by key

nopcoder:


graveler/committed/manager.go
```diff
@@ -23,7 +23,22 @@ func (c *committedManager) Exists(ns graveler.StorageNamespace, id graveler.Meta
...
+	if err != nil {
+		return nil, err
+	}
+	valIt := NewValueIterator(it)
+	valIt.SeekGE(key)
+	// return the next value
+	if valIt.Next() {
+		rec := valIt.Value()
+		return rec.Value, nil
+	}
```
guy-har, line 36: Need to check if it is equal

guy-har (changes requested):

graveler/committed/manager.go
nopcoder: true... fixing

guy-har (approved):

PR #1179, opened 2021-01-11
Share a single Pebble SSTable cache between all SSTable readers

### LINKED ISSUE #1177: pebble cache ref-count exit with an error
pebble: cache (0xc00017c820) has non-zero reference count: 1

arielshaqed:
Fixes #1177 (probably): the single cache is never finalized and should be OK.

Configure a pebble SSTable cache and use _that_.  This lets the admin control the amount of
RAM used to cache SSTable blocks.  Using a single cache makes *more* sense than giving each
reader its own cache: reading a block from an open reader costs the same no matter which it
is, so a single cache maximizes throughput.  There is no design goal to minimze latency per
reader (and this would *not* translate to minimizing latency per operation, rather it would
lower throughput on ranges that are heavily used).

This single cache has 2 references, one from the metarange readers cache and the other from
the range readers cache.  Because we never release those caches the Pebble SSTable cache is
never finalized until the program stops running.  (A panic is *still* possible then; fixing
that would require writing a finalizer.  This noncomposability of pebble SSTables indicates
just how hard that is to get right.)

There may well still be SSTable reader leaks which caused the original bug.  Here we merely
prevent those leaks from failing lakefs with a panic.

### DIFF
--- a/catalog/rocks/entry_catalog.go
+++ b/catalog/rocks/entry_catalog.go
@@ -16,6 +16,7 @@ import (
 	"github.com/treeverse/lakefs/pyramid"
 	"github.com/treeverse/lakefs/pyramid/params"
 
+	pebble "github.com/cockroachdb/pebble"
 	pebble_sst "github.com/cockroachdb/pebble/sstable"
 )
 
@@ -103,12 +104,14 @@ func NewEntryCatalog(cfg *config.Config, db db.Database) (*EntryCatalog, error) 
 		return nil, fmt.Errorf("create tiered FS for committed meta-range: %w", err)
 	}
 
+	pebbleSSTableCache := pebble.NewCache(tierFSParams.PebbleSSTableCacheSizeBytes)
 	metaRangeCache := sstable.NewCache(*cfg.GetCommittedRangeSSTableCacheParams(),
 		metaRangeFS,
-		pebble_sst.ReaderOptions{})
+		pebble_sst.ReaderOptions{Cache: pebbleSSTableCache})
+	pebbleSSTableCache.Ref() // rangeCache keeps a second reference.
 	rangeCache := sstable.NewCache(*cfg.GetCommittedMetaRangeSSTableCacheParams(),
 		rangeFS,
-		pebble_sst.ReaderOptions{})
+		pebble_sst.ReaderOptions{Cache: pebbleSSTableCache})
 
 	sstableManager := sstable.NewPebbleSSTableRangeManager(rangeCache, rangeFS, hashAlg)
 	sstableMetaManager := sstable.NewPebbleSSTableRangeManager(metaRangeCache, metaRangeFS, hashAlg)
--- a/catalog/rocks/entry_catalog.go
+++ b/catalog/rocks/entry_catalog.go
@@ -105,10 +105,10 @@ func NewEntryCatalog(cfg *config.Config, db db.Database) (*EntryCatalog, error) 
 	}
 
 	pebbleSSTableCache := pebble.NewCache(tierFSParams.PebbleSSTableCacheSizeBytes)
+	defer pebbleSSTableCache.Unref()
 	metaRangeCache := sstable.NewCache(*cfg.GetCommittedRangeSSTableCacheParams(),
 		metaRangeFS,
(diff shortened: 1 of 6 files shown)


catalog/rocks/entry_catalog.go
```diff
@@ -103,12 +104,14 @@ func NewEntryCatalog(cfg *config.Config, db db.Database) (*EntryCatalog, error)
 		return nil, fmt.Errorf("create tiered FS for committed meta-range: %w", err)
 	}
 
+	pebbleSSTableCache := pebble.NewCache(tierFSParams.PebbleSSTableCacheSizeBytes)
 	metaRangeCache := sstable.NewCache(*cfg.GetCommittedRangeSSTableCacheParams(),
 		metaRangeFS,
-		pebble_sst.ReaderOptions{})
+		pebble_sst.ReaderOptions{Cache: pebbleSSTableCache})
+	pebbleSSTableCache.Ref() // rangeCache keeps a second reference.
```
nopcoder, line 111: Don't think we should call Ref() here.
SST cache is created here ref=1.
Each service that will accept this cache should call Ref if it keeps the reference (and Unref if it releases it)
In this case when we pass the cache to metaRangeCache and rangeCache each one should call Ref and EntryCatalog should call Unref.

graveler/committed/manager.go
```diff
@@ -29,6 +29,7 @@ func (c *committedManager) Get(ctx context.Context, ns graveler.StorageNamespace
 		return nil, err
 	}
 	valIt := NewValueIterator(it)
+	defer valIt.Close()
```
nopcoder, line 32: 🙏

graveler/committed/manager.go
```diff
@@ -106,6 +107,7 @@ func (c *committedManager) Apply(ctx context.Context, ns graveler.StorageNamespa
 	if err != nil {
 		return "", fmt.Errorf("get metarange ns=%s id=%s: %w", ns, rangeID, err)
 	}
+	defer metaRangeIterator.Close()
```
nopcoder, line 110: 🙇

graveler/sstable/cache.go
```diff
@@ -53,6 +56,12 @@ func (i *item) Close() error {
 }
 
 func NewCache(p lru.ParamsWithDisposal, fs pyramid.FS, readerOptions sstable.ReaderOptions) Cache {
+	if readerOptions.Cache == nil {
+		// Shared cache between *all* readers, otherwise sstable.NewReader keeps its
+		// refcnt but forgets only reference.  *This* cache is held inside opener so
+		// can never be finalized.
+		readerOptions.Cache = pebble.NewCache(DefaultPebbleSSTableCacheSize)
```
nopcoder, line 63: Just checking - currently, there is no code that calls NewCache without sst cache in the readerOptions, right?
The shared cache in the comment is created in the EntryCatalog.

graveler/committed/manager.go
arielshaqed: Resource management is too hard for human beings.  If only we had something like finalizers, but with deterministic  execution, automatic resource management would be so easy...  We could call them something like "destroyers" or "anticonstructors".

graveler/sstable/cache.go
arielshaqed: Yes (we do inject openers in tests, calling `NewCacheWithOpener` directly... and not a pebble SSTable in sight).

catalog/rocks/entry_catalog.go
arielshaqed: Good idea.  Done!

arielshaqed (commented):
Thanks!

Did some reference counting dances, hope this matches what you requested.  I think it does make sense.

graveler/committed/manager.go
nopcoder: maybe we can take the http request Close linter and make it work on our iterators

nopcoder (approved):

ozkatz (approved):

graveler/committed/manager.go
arielshaqed: I shall have a look!

arielshaqed (commented):
Thanks; pulling!

PR #1180, opened 2021-01-11
Plumb context through graveler and pyramid

arielshaqed:
Allows cancellation, better logging (in future; mostly not done yet).


itaiad200 (approved):
LGTM!
Now we just need to use that context everywhere..

arielshaqed:
Thanks!

PR #1185, opened 2021-01-12
Branch locker based on postgres

### LINKED ISSUE #1165: Catalog branch locker should work cross process
The current branch locker uses an in-process locking mechanism - we should move this mechanism to a central location (Postgres) in order to support multiple lakeFS instances working together.

nopcoder:
Close #1165

Branch locker implementation based on Postgres advisory locks.
https://www.postgresql.org/docs/9.4/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS


graveler/branch_locker.go
```diff
@@ -1,96 +1,82 @@
...
 
-func formatBranchLockerKey(repositoryID RepositoryID, branchID BranchID) string {
-	return fmt.Sprintf("%s/%s", repositoryID, branchID)
+func calculateBranchLockerKeys(repositoryID RepositoryID, branchID BranchID) (int32, int32) {
+	h := fnv.New32()
+	_, _ = h.Write([]byte(repositoryID))
+	_, _ = h.Write([]byte{0})
+	_, _ = h.Write([]byte(branchID))
+	_, _ = h.Write([]byte{0})
+	key1 := int32(h.Sum32())
```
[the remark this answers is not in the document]
arielshaqed: Also, this allows bad users to block operations on other branches by merely selecting a good branch name.  With 32 bits it should be a one-time operation of <10seconds.

graveler/branch_locker.go
```diff
@@ -1,96 +1,82 @@
...
-		return nil, ErrBranchLocked
-	}
-	// wait until all writers will leave
-	data.metadataUpdate = true
-	for data.writers > 0 {
-		l.c.Wait()
-	}
-	return func() { l.releaseMetadataUpdate(key) }, nil
+// Writer try to lock as writer, using postgres advisory lock for the span of calling `lockedCB`.
+// returns ErrLockNotAcquired in case we fail to acquire the lock or commit is in progress
```
johnnyaug, line 29:
```suggestion
// Writer tries to acquire the lock as a writer. On success, it calls `lockedCB` and then releases the lock.
// Returns ErrLockNotAcquired in case we fail to acquire the lock.
```

graveler/branch_locker_test.go
```diff
@@ -10,73 +13,153 @@ import (
...
+		const writers = 5
+		var wgLocked sync.WaitGroup
+		var wgDone sync.WaitGroup
+		wgLocked.Add(writers)
+		wgDone.Add(writers)
+		for i := 0; i < writers; i++ {
+			go func() {
+				defer wgDone.Done()
+				_, err := bl.Writer(ctx, "a", testutil.DefaultBranchID, func() (interface{}, error) {
+					wgLocked.Done()
```
arielshaqed, line 31: Consider waiting for all the other writers to lock here (i.e. add _another_ WaitGroup, add 1 to it before calling `bl.Writer` and calling its Done and then its Wait in here.  Otherwise the scheduler could just run the writers in sequence :-(

graveler/branch_locker_test.go
```diff
@@ -10,73 +13,153 @@ import (
...
-	closeWrite()
+	ctx := context.Background()
+	t.Run("multiple_writers", func(t *testing.T) {
+		stopWritersCh := make(chan struct{})
+		const writers = 5
+		var wgLocked sync.WaitGroup
+		var wgDone sync.WaitGroup
+		wgLocked.Add(writers)
+		wgDone.Add(writers)
+		for i := 0; i < writers; i++ {
```
arielshaqed, line 27: Consider doing everything in here 100 times in a loop, to increase the chance of a race.

graveler/branch_locker.go
```diff
@@ -1,96 +1,82 @@
...
+
+	"github.com/treeverse/lakefs/db"
 )
 
-// branchLocker enforces the branch locking logic
+// BranchLocker enforces the branch locking logic
 // The logic is as follows:
 // allow concurrent writers to branch
 // metadata update commands will wait until all current running writers are done
 // while metadata update is running or waiting to run, writes and metadata update commands will be blocked
```
johnnyaug, line 17:
```suggestion
// BranchLocker enforces the branch locking logic
// The logic is as follows:
// - Allow concurrent writers to acquire the lock.
// - A Metadata update waits for all current writers to release the lock, and then gets the lock.
// - While a metadata update has the lock or is waiting for the lock, any other operation fails to acquire the lock.
```

graveler/committed/committed.proto
```diff
@@ -1,5 +1,5 @@
 syntax = "proto3";
-option go_package = "github.com/treevese/lakefs/graveler/committed";
+option go_package = "github.com/treeverse/lakefs/graveler/committed";
```
arielshaqed, line 2: Uggh.

arielshaqed (approved):
Nice test!

graveler/branch_locker.go
```diff
@@ -1,96 +1,82 @@
...
+			return nil, fmt.Errorf("%w (%d)", ErrLockNotAcquired, key1)
+		}
+		// lock writer key
+		_, err = tx.Exec(`SELECT pg_advisory_xact_lock($1);`, key1)
+		if err != nil {
+			return nil, fmt.Errorf("%w (%d): %s", ErrLockNotAcquired, key1, err)
+		}
+		if !locked {
+			return nil, fmt.Errorf("%w (%d)", ErrLockNotAcquired, key1)
+		}
```
johnnyaug, line 68: `locked` always true here

graveler/branch_locker.go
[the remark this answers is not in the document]
nopcoder: true, at first I thought I'll use the API that accepts both keys which require the 32bit.
switch to use 64bit.

graveler/branch_locker.go
```diff
@@ -1,96 +1,82 @@
...
+// Writer try to lock as writer, using postgres advisory lock for the span of calling `lockedCB`.
+// returns ErrLockNotAcquired in case we fail to acquire the lock or commit is in progress
+func (l *BranchLocker) Writer(ctx context.Context, repositoryID RepositoryID, branchID BranchID, lockedCB func() (interface{}, error)) (interface{}, error) {
+	key1, _ := calculateBranchLockerKeys(repositoryID, branchID)
+	return l.db.Transact(func(tx db.Tx) (interface{}, error) {
+		// try lock committer key
+		var locked bool
+		err := tx.GetPrimitive(&locked, `SELECT pg_try_advisory_xact_lock_shared($1)`, key1)
+		if err != nil {
+			return nil, fmt.Errorf("%w (%d): %s", ErrLockNotAcquired, key1, err)
```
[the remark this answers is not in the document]
nopcoder: I'll try to wrap both of them, the problem is that the current caller would like to know if the lock was acquired and doesn't handle specific database errors for the locking mechanism.
If a database error for the locking failed, we can handle the specific errors under the brach locker without exposing the actual DB layer outside the branch locker.
This means I want to expose the ErrLockNotAcquired at this point and not the DB implementation error outside - at least as an actual error you need to handle.

johnnyaug (changes requested):
Nice and clean!

graveler/branch_locker.go
nopcoder: good point

graveler/branch_locker.go
nopcoder: mixed your comment with ariels's comment

graveler/branch_locker.go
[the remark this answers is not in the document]
nopcoder: @arielshaqed can add a package like https://pkg.go.dev/go.uber.org/multierr to answer both errors.Is.
Prefer to keep it this way until the point we think the caller really needs to compare the locker's DB error.

nopcoder:
Still looking at why the test passes on my laptop and not on the CI - sorry about that.

graveler/graveler.go
```diff
@@ -474,17 +476,17 @@ type graveler struct {
 	CommittedManager CommittedManager
 	StagingManager   StagingManager
 	RefManager       RefManager
-	branchLocker     branchLocker
+	branchLocker     *BranchLocker
 	log              logging.Logger
 }
 
-func NewGraveler(committedManager CommittedManager, stagingManager StagingManager, refManager RefManager) *graveler {
+func NewGraveler(db db.Database, committedManager CommittedManager, stagingManager StagingManager, refManager RefManager) *graveler {
```
ozkatz, line 483: Why not create a branch locker interface and do dependency injection on that as well? to me it feels like a healthier seperation of concerns, the same way we treat all other parts that graveler is composed of.

ozkatz (approved):
Looks great to me! nice work. I learned about pg advisory locks because of this :)
One thing I'm not too excited about is Graveler instantiating a branchLocker which is coupled to pg.

My suggestion is to extract an interface from branchLocker and do dependency injection the same way we compose the rest of Graveler. Not a blocker and could easily be refactored later if you agree, so, approved!

graveler/graveler.go
nopcoder: 1. No problem - will do.
2. At first I thought graveler will require Postgres to generate unique identifiers per branch and DB will be used for that. But I liked the hash solution better.

nopcoder: Moved the implementation to `ref` package and NewGraveler consume BranchLocker interface

nopcoder:
> Nice and clean!

still need your blessing/approval on this one

graveler/graveler.go
```diff
@@ -448,6 +446,16 @@ type StagingManager interface {
 	DropByPrefix(ctx context.Context, st StagingToken, prefix Key) error
 }
 
+// BranchLocker enforces the branch locking logic
+// The logic is as follows:
+// - Allow concurrent writers to acquire the lock.
+// - A Metadata update waits for all current writers to release the lock, and then gets the lock.
+// - While a metadata update has the lock or is waiting for the lock, any other operation fails to acquire the lock.
+type BranchLocker interface {
+	Writer(ctx context.Context, repositoryID RepositoryID, branchID BranchID, lockedCB func() (interface{}, error)) (interface{}, error)
```
ozkatz, line 455: lockedCB could be at type alias, kind of like `context.CancelFunc`, to avoid repetition...

ozkatz (approved):

graveler/graveler.go
nopcoder: done - a little type coupling, but nice to read.

johnnyaug (approved):

PR #1186, opened 2021-01-13
[1158] Control metadata range file sizes

### LINKED ISSUE #1158: Split range files around ApproximateRangeSizeBytes
In GeneralMetaRangeWriter, the hash of the key determines if the range should break at the current entry.
Each entry represents more than a single byte, hence name is misleading and implementation has a bug.

arielshaqed:
See [issue comment](https://github.com/treeverse/lakeFS/1158#issuecomment-758640791) for full details.

Add 3 parameters:

* minimum range size (bytes)
* maximum range size (bytes)
* raggedness factor (entries)

Close (split) a range after adding an entry if any of these holds:

* Next need to add an entire range (TODO: collect data to see whether we want to buffer some range entries to avoid this creating very small ranges). OR...
* The current range size is >= maximum range size. (SO the maximum is not really a maximum, we go one whole entry above it). OR...
* The current range is >= minimum range size and hash(entry.Key) / maxHashValue > 1/raggedness.

(Initial) default min range size is 0 to avoid forcing the non-essential change.  We shall
change at least the parameters, and possibly the entire logic.  Making this change requires
data, and data requires a working system, so do this for now.

Fixes #1158.

### DIFF
--- a/graveler/committed/meta_range_manager.go
+++ b/graveler/committed/meta_range_manager.go
@@ -10,9 +10,16 @@ import (
 )
 
 type Params struct {
-	// ApproximateRangeSizeBytes is a target for sizes of range partitions.  It should be
-	// kept fixed to maximize re-use of range partitions.
-	ApproximateRangeSizeBytes uint64
+	// MinRangeSizeBytes is the smallest size for splitting a range partition as a result
+	// of adding a record.  Smaller ranges are still possible due to re-using an existing
+	MinRangeSizeBytes uint64
+	// MaxRangeSizeBytes is the largest size of a range partition.  In practice the range
+	// is split only after an additional record.
+	MaxRangeSizeBytes uint64
+	// RangeSizeRaggedness allows raggedness in splitting range partitions.  It is
+	// the expected number of records after MinRangeSizeBytes at which to split the range
+	// -- ranges are split at the first key with hash divisible by this raggedness.
+	RangeSizeRaggedness float64
 }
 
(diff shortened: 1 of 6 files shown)


arielshaqed:
Prefer to get approvals from *both* @itaiad200 and @guy-har.  Thanks!

graveler/committed/meta_range_manager.go
```diff
@@ -10,9 +10,16 @@ import (
...
+	// MinRangeSizeBytes is the smallest size for splitting a range partition as a result
+	// of adding a record.  Smaller ranges are still possible due to re-using an existing
+	MinRangeSizeBytes uint64
+	// MaxRangeSizeBytes is the largest size of a range partition.  In practice the range
+	// is split only after an additional record.
+	MaxRangeSizeBytes uint64
+	// RangeSizeRaggedness allows raggedness in splitting range partitions.  It is
+	// the expected number of records after MinRangeSizeBytes at which to split the range
+	// -- ranges are split at the first key with hash divisible by this raggedness.
+	RangeSizeRaggedness float64
```
itaiad200, line 22: Other two are in bytes so I would mention the difference in the name:
```suggestion
	RangeSizeEntriesRaggedness float64
```

itaiad200 (approved):

arielshaqed:
Thanks!

Made the requested change because it's clearly a better name, rebased to get MVCC _and_ Rocks Nessie, and will pull once tests pass.

PR #1187, opened 2021-01-13
Run Nessie system test also on rocks configuration

arielshaqed:


arielshaqed:
_Sort-of ready for review but not really..._ Want to let it run, make it non-blocking, etc., before requesting a review.

itaiad200 (approved):
LGTM!
Minor: Extract `${{ github.run_number }}-${{ matrix.cataloger }}` into a workflow-level var and reuse it.

PR #1191, opened 2021-01-13
add record count to metarange

### LINKED ISSUE #1188: Add Number of Records to range meta range metadata
In meta-range sstable,
We would like to add the number of records in the range for every range under metadata.
Each row should look like this:
```
key  	| Identity 	| value 
---------------------------------
max_key | range_id 	| metadata:
		|		   	| min_key
		|			| estimated_size
		|			| num_of_records
```

johnnyaug:
Closes #1188


graveler/committed/committed.proto
```diff
@@ -8,4 +8,5 @@ message RangeData {
 	bytes min_key = 1;
 	bytes max_key = 2;
 	uint64 estimated_size = 3;
+	uint32 count = 4;
```
arielshaqed, line 11:
```suggestion
	uint32 num_records = 4;
```
or
```suggestion
	uint32 records_count = 4;
```
or similar: "count" is units but not what is being counted.

graveler/committed/meta_range_writer.go
```diff
@@ -96,6 +96,7 @@ func (w *GeneralMetaRangeWriter) getBatchedRanges() ([]Range, error) {
 			MinKey:        r.First,
 			MaxKey:        r.Last,
 			EstimatedSize: r.EstimatedRangeSizeBytes,
+			Count:         uint32(r.Count),
```
arielshaqed, line 99: Consider erroring out if `r.Count < 0`.  (Might happen if someone decides to mark some sentinel value for counting).

graveler/committed/committed.proto
arielshaqed: Also consider
```suggestion
	int64 count = 4;
```
a. Lets you stay signed (so safe arithmetic is safer)
b. We can expect customers to be within 1000x of this once we have a big user; why be close to the limit?
c. Varint encoding makes it free in disk space.

graveler/committed/meta_range_writer_test.go
```diff
@@ -107,7 +107,7 @@ func TestWriter_RecordRangeAndClose(t *testing.T) {
 
 	namespace := committed.Namespace("ns")
 	record := graveler.ValueRecord{Key: nil, Value: &graveler.Value{}}
-	rng := committed.Range{MinKey: committed.Key("a"), MaxKey: committed.Key("g")}
+	rng := committed.Range{ID: "rng2-id", MinKey: committed.Key("a"), MaxKey: committed.Key("g"), Count: 4}
```
arielshaqed, line 110: I don't understand why "4".  A comment might help me and people like me.

arielshaqed (commented):
Thanks!

Not approving because I don't understand the "4" in the test (comment inline).  Not requesting changes because if I'm missing something and @guy-har approves then I'm happy to pull this.

graveler/committed/meta_range_writer_test.go
johnnyaug: No particular reason, much like the "a" there

graveler/committed/committed.proto
johnnyaug: Changing the uint32 to int64.
I think "count" is descriptive enough and in accordance with the structs in our code.

graveler/committed/meta_range_writer_test.go
guy-har: It's a bit confusing, but it's fine. 
It is the range that will be added and not the expected range (could have any value)

guy-har (approved):
LGTM
Thanks

graveler/committed/meta_range_writer_test.go
```diff
@@ -118,10 +118,33 @@ func TestWriter_RecordRangeAndClose(t *testing.T) {
...
+		Value: &graveler.Value{
+			Identity: []byte("rng-id"),
+			Data: mustMarshalRange(committed.Range{
+				ID:     "rng-id",
+				MinKey: []byte("a"),
+				MaxKey: []byte("a"),
+				Count:  1,
+			}),
+		},
+	})).After(call1)
```
arielshaqed: This is over-specified.  It would be correct to perform this write at any time.  If you want to sequence things, then sequence the `GetWriter` calls to return different mock writers, and verify that each mock gets the right writes.

arielshaqed (approved):
Thanks!

Approving, but please consider removing the over-specification of behaviour in the tests (the sequencing there verifies an implementation rather than correctness) or changing it to be more accurate.  (Or _both_: remove the over-specification now, add the exact required separate-writers behaviour in  a later PR...)

PR #1193, opened 2021-01-13
Import on top of existing commit

itaiad200:


graveler/graveler.go
```diff
@@ -842,24 +842,14 @@ func newStagingToken(repositoryID RepositoryID, branchID BranchID) StagingToken
...
-	branch, err := g.RefManager.GetBranch(ctx, repositoryID, branchID)
+	_, err = g.RefManager.GetCommit(ctx, repositoryID, parentCommitID)
 	if err != nil {
-		return "", fmt.Errorf("get branch %s: %w", branchID, err)
-	}
-	if empty, err := g.stagingEmpty(ctx, branch); err != nil {
-		return "", err
-	} else if !empty {
-		return "", ErrDirtyBranch
+		return "", fmt.Errorf("get commit %s: %w", parentCommitID, err)
```
nopcoder, line 852: get parent commit

graveler/graveler.go
```diff
@@ -842,24 +842,14 @@ func newStagingToken(repositoryID RepositoryID, branchID BranchID) StagingToken
...
-		return "", fmt.Errorf("acquire metadata update: %w", err)
-	}
-	defer cancel()
+func (g *graveler) CommitExistingMetaRange(ctx context.Context, repositoryID RepositoryID, parentCommitID CommitID, metaRangeID MetaRangeID, committer string, message string, metadata Metadata) (CommitID, error) {
 	repo, err := g.RefManager.GetRepository(ctx, repositoryID)
 	if err != nil {
 		return "", fmt.Errorf("get repository %s: %w", repositoryID, err)
 	}
-	branch, err := g.RefManager.GetBranch(ctx, repositoryID, branchID)
+	_, err = g.RefManager.GetCommit(ctx, repositoryID, parentCommitID)
```
nopcoder, line 850: Think we should consider import to an empty repository - there will be no parent commit, or you should consider empty parent commit as valid.
Probably need to ignore the error ErrCommitNotFound and when filling the commit information, do not fill Parents with slice with empty string.

onboard/rocks_catalog_actions.go
```diff
@@ -33,23 +34,27 @@ func (c *RocksCatalogRepoActions) Progress() []*cmdutils.Progress {
...
 		metaRanger: metaRanger,
 		repoID:     repository,
+		commitID:   commitID,
 		committer:  committer,
 		logger:     logger,
-		progress:   cmdutils.NewActiveProgress("Objects imported", cmdutils.Spinner),
-		commit:     cmdutils.NewActiveProgress("Commit progress", cmdutils.Spinner),
+
+		progress: cmdutils.NewActiveProgress("Objects imported", cmdutils.Spinner),
+		commit:   cmdutils.NewActiveProgress("Commit progress", cmdutils.Spinner),
```
nopcoder, line 51: do we need this one?

onboard/rocks_catalog_actions.go
```diff
@@ -84,10 +93,15 @@ func (c *RocksCatalogRepoActions) Commit(ctx context.Context, commitMsg string,
 	if c.createdMetaRangeID == nil {
 		return "", ErrNoMetaRange
 	}
-	commitID, err := c.metaRanger.CommitExistingMetaRange(ctx, c.repoID, catalog.DefaultImportBranchName, *c.createdMetaRangeID, c.committer, commitMsg, graveler.Metadata(metadata))
+	commitID, err := c.metaRanger.CommitExistingMetaRange(ctx, c.repoID, c.commitID, *c.createdMetaRangeID, c.committer, commitMsg, graveler.Metadata(metadata))
```
nopcoder, line 96: One thing related to this flow - in case we manage to complete everything until the point we try to update the branch and fail.
Retry the operation, will it add the same commit again? because the ref manager just AddCommit - maybe we need to ident.ContentAddress(commit) and skip this step in case the commit is already there, so retry will work.

nopcoder (commented):
Cool - just a single concern I had when I read the code

graveler/graveler.go
```diff
@@ -236,7 +236,7 @@ type VersionController interface {
 	// CommitExistingMetaRange creates a commit in the branch from the given pre-existing tree.
 	// Returns ErrDirtyBranch if the branch has uncommitted changes.
 	// Returns ErrTreeNotFound if the referenced treeID doesn't exist.
```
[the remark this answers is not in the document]
itaiad200: > This comment is really out of date:

> What does ErrDirtyBranch mean now that there is no branch?

Removed. 

> What's a "treeID" (there didn't use to be a treeID; there still is no treeID, but there didn't use to be one, either)?

Updated the comment to match the existing error.

> This interface no longer allows Graveler to advance the branch head (and, TBH, I couldn't find where the existing implementation does that). Is that intentional?

Right, branch isn't advanced here to support the API we defined. However existing functionality exists with `UpdateBranch`.

> Generally this sounds that once we have this method, either this or Commit becomes a function not a method (i.e., it could be composed of methods except maybe a line or 2 that could be added as small methods), so why force every interface implementer to implement it? It seems to have long and complex semantics, and as this PR shows it seems really focused on implementing one particular case of "import". Could we split it up maybe into a function (external to the interface) and 1, 2 or 3 added methods to do just small bits that don't already exist?

Rocks import operation is currently composed of the following:
1. ListEntries of the parent commit.
2. List the inventory (not part of the interface)
3. Write metarange with the merged iterator.
4. Commit the metarange without advancing the branch.
5. Advance the branch (if requested - plumbing will not advance).

I suggest we discuss all 5, instead of discussing of ways to split just 4.
1 & 5 are part of the existing interface, 2 isn't. So we're left with 3 & 4 which were written to suit the import flow as you mentioned. I can think of a way to merging them into one, but I'm pretty sure that breaking them further will cause more incoherent-why-do-we-need-this-oh-it's-for-import methods, not less.

graveler/graveler.go
```diff
@@ -864,60 +863,44 @@ func newStagingToken(repositoryID RepositoryID, branchID BranchID) StagingToken
...
-		branch, err := g.RefManager.GetBranch(ctx, repositoryID, branchID)
-		if err != nil {
-			return "", fmt.Errorf("get branch %s: %w", branchID, err)
-		}
-		if empty, err := g.stagingEmpty(ctx, branch); err != nil {
-			return "", err
-		} else if !empty {
-			return "", ErrDirtyBranch
-		}
+func (g *graveler) CommitExistingMetaRange(ctx context.Context, repositoryID RepositoryID, parentCommitID CommitID, metaRangeID MetaRangeID, committer string, message string, metadata Metadata) (CommitID, error) {
```
nopcoder, line 866: Just fyi we would probably need to update this method to accept Commit (struct) in order to control the time and parents (more than one) to migrate.

onboard/merge_iterator.go
```diff
@@ -0,0 +1,170 @@
...
+		_ = pmi.advanceIt(pmi.committedIt)
+		return true
+
+	default:
+		_ = pmi.advanceIt(pmi.invIt)
+		return true
+	}
+}
+
+func (pmi *prefixMergeIterator) advanceIt(it rocks.EntryIterator) bool {
```
nopcoder, line 58: `it` is copied by value

onboard/merge_iterator.go
```diff
@@ -0,0 +1,170 @@
...
+	hasNext := it.Next()
+	pmi.err = it.Err()
+	return hasNext
+}
+
+var ErrNotSeekable = errors.New("iterator isn't seekable")
```
nopcoder, line 64: can you push this one up after the 'types'

onboard/merge_iterator.go
```diff
@@ -0,0 +1,170 @@
...
+	return invVal
+}
+
+func (pmi *prefixMergeIterator) Err() error {
+	return pmi.err
+}
+
+func (pmi *prefixMergeIterator) Close() {
+	pmi.invIt.Close()
+	pmi.err = pmi.invIt.Err()
```
nopcoder, line 98: not sure you are allowed to call Err after Close

arielshaqed, line 98: Don't you also need to close `committedIt`?

onboard/merge_iterator.go
```diff
@@ -0,0 +1,170 @@
...
+	"errors"
+	"strings"
+
+	"github.com/treeverse/lakefs/catalog/rocks"
+)
+
+// prefixMergeIterator takes inventory iterator and commit iterator and merges them.
+// If an entry's prefix is in `prefixes`, it would be taken from invIt.
+// No entry from committedIt would be returned if its prefix is in prefixes.
+type prefixMergeIterator struct {
```
arielshaqed, line 13: I think this iterator needs a test.  Doesn't have to be in this PR, but should be before the upcoming release.

onboard/merge_iterator.go
```diff
@@ -0,0 +1,170 @@
...
+func (pmi *prefixMergeIterator) Err() error {
+	return pmi.err
+}
+
+func (pmi *prefixMergeIterator) Close() {
+	pmi.invIt.Close()
+	pmi.err = pmi.invIt.Err()
+}
+
+type ignorePrefixIterator struct {
```
arielshaqed, line 101: I think this iterator needs a test.  Doesn't have to be in this PR, but should be before the upcoming release.

arielshaqed (commented):
Not sure about the iterators; see comments.

onboard/merge_iterator.go
```diff
@@ -0,0 +1,170 @@
...
+	err   error
+	value *rocks.EntryRecord
+}
+
+func newIgnorePrefixIterator(it rocks.EntryListingIterator, prefixes []string) rocks.EntryIterator {
+	return &ignorePrefixIterator{
+		it:       it,
+		prefixes: prefixes,
+		err:      nil,
+		value:    nil,
```
nopcoder, line 114: 1. general note - we don't need to initialize defaults
2. err will be always it.Err() so we can drop it

onboard/merge_iterator.go
```diff
@@ -0,0 +1,170 @@
...
+	return ipi.value
+}
+
+func (ipi *ignorePrefixIterator) Err() error {
+	return ipi.err
+}
+
+func (ipi *ignorePrefixIterator) Close() {
+	ipi.it.Close()
+	ipi.err = ipi.it.Err()
```
nopcoder, line 169: don't use the iterator after Close

nopcoder (changes requested):

onboard/merge_iterator.go
```diff
@@ -0,0 +1,170 @@
...
+
+		// from now on - both iterator has values,
+		// so we'll always return true
+	case strings.Compare(pmi.committedIt.Value().Path.String(), pmi.invIt.Value().Path.String()) < 0:
+		_ = pmi.advanceIt(pmi.committedIt)
+		return true
+
+	default:
+		_ = pmi.advanceIt(pmi.invIt)
+		return true
```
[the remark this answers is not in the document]
itaiad200: By definition, both iterators look at different prefixes, hence they are distinct.
Added it to the prefixMergeIterator definition, so it will be clearer.

onboard/merge_iterator.go
itaiad200: You sure about that? `it` is an interface.. I guess tests will tell :)

onboard/merge_iterator.go
itaiad200: Sure.

onboard/merge_iterator.go
itaiad200: So how can one receive an error that derived from `Close()`?

itaiad200: Done.

onboard/merge_iterator.go
itaiad200: I will add as part of this PR.

onboard/merge_iterator.go
itaiad200: 1. Removed.
2. You have the `NotSeekable` error which isn't from the inner iterator..

onboard/merge_iterator.go
itaiad200: I don't understand how errors from `Close` are returned then..

onboard/rocks_catalog_actions.go
itaiad200: Added validation.

onboard/merge_iterator.go
itaiad200: I will add it as part of this PR.

graveler/graveler.go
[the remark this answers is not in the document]
arielshaqed: I think we're good now, especially with #1229 already appearing to re-use this interface.

onboard/merge_iterator.go
nopcoder: my bad - missed that part it is an interface.

onboard/merge_iterator.go
nopcoder: Maybe I missed this one in the Iterator protocol - thought that we wanted to ignore any error from Close as we cloud add `Close() error` to the interface.
Just note that most of the iterators will now hold an error when the iterator is closed.
Which doesn't mean that there is an error on Close.

But to the point - there is no need to hold `ipi.err` it will always be `ipi.it.Err()`, right?

nopcoder (changes requested):
Just one comment related to the iterator err member

onboard/merge_iterator.go
```diff
@@ -0,0 +1,170 @@
...
+
+	case pmi.invIt.Value() == nil:
+		return pmi.advanceIt(pmi.committedIt)
+
+	case pmi.committedIt.Value() == nil:
+		return pmi.advanceIt(pmi.invIt)
+
+		// from now on - both iterator has values,
+		// so we'll always return true
+	case strings.Compare(pmi.committedIt.Value().Path.String(), pmi.invIt.Value().Path.String()) < 0:
```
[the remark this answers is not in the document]
arielshaqed: ETOOMANYITERATORS -- I was thinking of the wrong iterator type, sorry.

onboard/merge_iterator.go
itaiad200: Almost, we still have the `ErrNotSeekable `

onboard/merge_iterator.go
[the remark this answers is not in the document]
arielshaqed: This is a very distant invariant between `committedIt` and `invIt`.  The closest thing that I can find in the `prefixMergeIterator` docs is:

> No entry from committedIt would be returned if its prefix is in prefixes

which I do not think applies here.  If this is some distant invariant from whatever code creates the PMI, then it is scary that we will never see if it fails.  I would prefer a suitably scary log in that case.

onboard/merge_iterator_test.go
```diff
@@ -0,0 +1,91 @@
...
+					{Path: "a/b/d"},
+					{Path: "a/b/c/d"},
+					{Path: "a/c/d"},
+					{Path: "b/c/d"},
+				}), "", ""),
+			Prefixes: []string{"a/"},
+			Results:  []string{"a/b/c", "a/d/c", "b/c/d"},
+		},
+		{
+			Name: "Empty prefix - All inventory",
```
arielshaqed, line 38:
```suggestion
			Name: "No prefixes - All inventory",
```
(_Every_ string has an empty prefix; AFAICS the point here is that there *no* prefix*es* are provided.  Sorry, I am very literal minded.)

onboard/merge_iterator_test.go
```diff
@@ -0,0 +1,91 @@
...
+			InvIt: testutils.NewFakeEntryIterator([]*rocks.EntryRecord{
+				{Path: "a/b/c"},
+				{Path: "e/f/s/a"},
+			}),
+			CommittedIt: rocks.NewEntryListingIterator(testutils.NewFakeEntryIterator(
+				[]*rocks.EntryRecord{
+					{Path: "some/other/prefix"},
+					{Path: "some/other/prefix/c"},
+					{Path: "some/other/prefix/e"},
+				}), "", "/"),
```
arielshaqed, line 48: Is delimiter non-empty here intentionally?  If so, please reflect that in the test name.

onboard/merge_iterator_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/treeverse/lakefs/catalog/rocks/testutils"
+
+	"github.com/treeverse/lakefs/catalog/rocks"
+)
+
+func TestMergeIterator(t *testing.T) {
+	cases := []struct {
```
arielshaqed, line 14: I would be much happier if some test here had the same path in both `InvIt` _and_ `CommittedIt`, and it was part of the processed prefixes.

arielshaqed (commented):
Thanks for the tests!  I'm still unsure what prevents PMI from receiving the same path in both inventory and committed; I think it will emit that path twice.

onboard/merge_iterator.go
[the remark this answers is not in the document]
itaiad200: I agree, but logging isn't enough.
Made some changes to actually enforce that in the code, although it would be a NoOp with the current implementation of the inner inventory iterator

onboard/merge_iterator_test.go
itaiad200: You got it.

onboard/merge_iterator_test.go
itaiad200: Done.

onboard/merge_iterator_test.go
itaiad200: Not intentional, nice catch!

onboard/merge_iterator_test.go
```diff
@@ -23,6 +23,8 @@ func TestMergeIterator(t *testing.T) {
 			InvIt: testutils.NewFakeEntryIterator([]*rocks.EntryRecord{
 				{Path: "a/b/c"},
 				{Path: "a/d/c"},
```
arielshaqed, line 25:
```suggestion
				{Path: "a/c/d"},
```
(I want a dupe in something included by the prefix!)

arielshaqed (approved):
Thanks!

graveler/graveler_test.go
```diff
@@ -490,30 +504,14 @@ func TestGraveler_CommitExistingRange(t *testing.T) {
 			want:        expectedCommitID,
 			expectedErr: nil,
 		},
-		{
-			name: "staging is dirty",
```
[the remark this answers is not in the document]
itaiad200: No sir, you can't. Mainly because you're committing an existing meta-range and simply link to some parent. We DON'T advance the branch ref here, so checking for dirty branch is irrelevant.

onboard/merge_iterator_test.go
itaiad200: You got it.

onboard/merge_iterator.go
nopcoder: But this error is relevant after calling SeekGE not 'Close'

nopcoder: Got it

nopcoder (approved):

PR #1197, opened 2021-01-14
Fix (committed and uncommitted) delete file

### LINKED ISSUE #1183: Pyramid | removing file fails
I am getting the following error when running lakeFS:

```
ERRO[0034] Removing file failed                          error="remove /tmp/experiment/ranges/tmp/experiment/ranges/repo/test/ba87edb6b02b7213af0f5fd928459d3b59ecf932d75b27d1280502e5c37a0349: no such file or directory" path=/tmp/experiment/ranges/tmp/experiment/ranges/repo/test/ba87edb6b02b7213af0f5fd928459d3b59ecf932d75b27d1280502e5c37a0349
```

arielshaqed:
Reviewer: consider reviewing by commits, they are fairly well separated.

Fixes #1183.

### DIFF
--- a/gateway/operations/deleteobject.go
+++ b/gateway/operations/deleteobject.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"net/http"
 
+	"github.com/treeverse/lakefs/graveler"
 	"github.com/treeverse/lakefs/logging"
 
 	"github.com/treeverse/lakefs/block"
@@ -51,7 +52,7 @@ func (controller *DeleteObject) Handle(w http.ResponseWriter, req *http.Request,
 	lg := o.Log(req).WithField("key", o.Path)
 	err := o.Cataloger.DeleteEntry(req.Context(), o.Repository.Name, o.Reference, o.Path)
 	switch {
-	case errors.Is(err, db.ErrNotFound):
+	case errors.Is(err, db.ErrNotFound) || errors.Is(err, graveler.ErrNotFound):
 		lg.WithError(err).Debug("could not delete object, it doesn't exist")
 	case err != nil:
 		lg.WithError(err).Error("could not delete object")
--- a/graveler/graveler.go
+++ b/graveler/graveler.go
@@ -731,19 +731,8 @@ func (g *graveler) Delete(ctx context.Context, repositoryID RepositoryID, branch
 			return nil, err
 		}
 
-		// is the key in staging?
-		entry, err := g.StagingManager.Get(ctx, branch.StagingToken, key)
-
-		if err != nil {
-			// Either err is ErrNotFound -- and should be returned -- or some other error
-			// -- which should also be returned.
-			return nil, err
-		}
-
-		// return not found if we already got tombstone
-		if err == nil && entry == nil {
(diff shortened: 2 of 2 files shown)


gateway/operations/deleteobject.go
```diff
@@ -51,7 +52,7 @@ func (controller *DeleteObject) Handle(w http.ResponseWriter, req *http.Request,
 	lg := o.Log(req).WithField("key", o.Path)
 	err := o.Cataloger.DeleteEntry(req.Context(), o.Repository.Name, o.Reference, o.Path)
 	switch {
-	case errors.Is(err, db.ErrNotFound):
```
guy-har, line 54: Graveler now won't return ErrNotFound on delete, so I guess you don't need this

guy-har (approved):
Looks good,
Two things to consider
1. The MVCC version throws an error when deleting a file that does not exist, given that S3 does not, I am OK with removing it.
2. the combined iterator logs an error in case it gets an unexpected tombstone, please remove it. It only logs the error and continues, so I am approving but I suggest you remove it.

gateway/operations/deleteobject.go
arielshaqed: Agreed.  It's just too close for comfort.

I look forward to deleting `db.ErrNotFound` when we remove Postgres.

arielshaqed (commented):
> Looks good,
> Thanks
> Two things to consider
> 
>     1. The MVCC version throws an error when deleting a file that does not exist, given that S3 does not, I am OK with removing it.
> 
>     2. the combined iterator logs an error in case it gets an unexpected tombstone, please remove it. It only logs the error and continues, so I am approving but I suggest you remove it.

Not sure about either one.

1. S3 does not return an error on re-deletion because it used to be eventually consistent -- it _had no idea_ whether the file still existed or not.  This is now gone in the S3 world, and was never true here on lakeFS.  So I prefer to keep the "good" behaviour, and return the S3 "deleted unless something broke" error value for mere compatibility.  Note that it _also_ allows for writing better tests, so it must be good.
2. If the existing behaviour is broken I'd like to fix it in a separate PR.  But if we (should) never get these errors then I am _very_ happy to have the combining iterator complain -- it will help us detect bad behaviour wh^Wif it happens.

arielshaqed:
Thanks for the rapid review!

Fixed test failures (wrong error message on double-deletes), will pull once retests pass.

ozkatz:
I see I didn't make it in time to review it but I'll give my 2 cents anyway:
1. Fix looks good, of course!
2. I'm in favor of the metadata layer returning an error when deleting something that doesn't exist, regardless of S3's behavior - especially because we also have our own API (and another one for metadata only which is planned). If the S3 gateway chooses not to return an error that's a fine choice but we shouldn't limit the metadata layer because of it.
3. I feel like there's an implicit behavior here which deserves at least a unit test (i.e. if we say that `DeleteEntry` should return a `graveler.ErrNotFound` the 2nd best thing we can do is write an explicit test that validates this behavior (1st best would be a type system that gives this guarantee at compile time, but I guess we'll have to settle on runtime instead).

PR #1201, opened 2021-01-17
Write metadata to SSTables

### LINKED ISSUE #1200: sstables files do not write metadata describing their type
each sstable file should write a user property describing the type of sstable file (i.e a graveler meta-range or graveler range. and for ranges, ideally the proto or at least name of the serialized record type)

arielshaqed:
Fixes #1200.


go.mod
```diff
@@ -11,7 +11,9 @@ require (
 	github.com/avast/retry-go v2.6.1+incompatible
 	github.com/aws/aws-sdk-go v1.36.15
 	github.com/cenkalti/backoff/v3 v3.2.2 // indirect
-	github.com/cockroachdb/pebble v0.0.0-20201130172119-f19faf8529d6
+	github.com/cockroachdb/errors v1.8.2 // indirect
+	github.com/cockroachdb/pebble v0.0.0-20210113182756-a53f229c8c9c
```
ozkatz, line 15: I'm assuming this is inentional?

go.mod
```diff
@@ -29,9 +31,11 @@ require (
 	github.com/go-swagger/go-swagger v0.25.0
 	github.com/go-test/deep v1.0.7
 	github.com/gofrs/uuid v3.3.0+incompatible // indirect
+	github.com/gogo/protobuf v1.3.2 // indirect
```
ozkatz, line 34: not directly related to the review, but that's an interesting library :)

graveler/committed/meta_range_writer.go
```diff
@@ -55,6 +55,7 @@ func (w *GeneralMetaRangeWriter) WriteRecord(record graveler.ValueRecord) error
 		if err != nil {
 			return fmt.Errorf("get range writer: %w", err)
 		}
+		w.rangeWriter.AddMetadata("type", "ranges")
```
ozkatz, line 58: Since we're going to be reading these back and comparing the value, I think the types should be exported constants

graveler/committed/range_manager.go
```diff
@@ -77,6 +77,10 @@ type RangeWriter interface {
 	// WriteRecord appends the given record to the Range.
 	WriteRecord(record Record) error
 
+	// AddMetadata associates metadata value (which will be stringified when the writer is
+	// Closed) with key.
+	AddMetadata(key string, value interface{})
```
ozkatz, line 82: if they end up getting stringified I suggest changing the signature to accept either a byte slice, or better yet, a string. for many types, the string representation will be either lossy or simply unparsable.

graveler/sstable/writer.go
```diff
@@ -106,6 +116,14 @@ func (dw *DiskWriter) Close() (*committed.WriteResult, error) {
 	tableHash := dw.hash.Sum(nil)
 	sstableID := hex.EncodeToString(tableHash)
 
+	// Prepare metadata properties for Close to write.  The map was already set in the
+	// sstable.Writer constructor and cannot be changed, but we can replace its values
+	// before writing it out.
+	dw.AddMetadata("first_key", dw.first)
```
ozkatz, line 122: perhaps we should align these values with our canonical names for these (i.e. `min_key`, `max_key` and whatever it is we're calling `num_records` which I believe is currently `count` (??)

ozkatz (changes requested):
Thanks! this is very helpful! the only thing I'm blocking approval on (but could be convinced otherwise) is the stringification of values passed to `AddMetadata`

go.mod
arielshaqed: Of course!  You didn't seriously expect to be able to write SSTable metadata with `v0.0.0-20201130172119-f19faf8529d6`?!

go.mod
arielshaqed: The great thing about standard libraries like protobuf is that there are so many to choose from :-(

graveler/committed/meta_range_writer.go
arielshaqed: Good idea, done!

graveler/committed/range_manager.go
arielshaqed: Done, although it makes writing numbers a bit nasty.

graveler/sstable/writer.go
arielshaqed: Done.  Although we _also_ call them "first" and "last" in `WriteResult`, just a few lines down.  Not cleaning that up right now, but metadata tags are forever.

arielshaqed (commented):
Thanks!  PTAL...

ozkatz (approved):
Looks great! thanks

arielshaqed:
Thanks!

Pulling...

PR #1203, opened 2021-01-17
call merge implementation from graveler

johnnyaug:


nopcoder (approved):

graveler/committed/manager.go
```diff
@@ -91,8 +91,20 @@ func (c *committedManager) Diff(ctx context.Context, ns graveler.StorageNamespac
 	return NewDiffIterator(leftIt, rightIt), nil
 }
 
-func (c *committedManager) Merge(ctx context.Context, ns graveler.StorageNamespace, left, right, base graveler.MetaRangeID, committer string, message string, metadata graveler.Metadata) (graveler.MetaRangeID, error) {
-	panic("implement me")
+func (c *committedManager) Merge(ctx context.Context, ns graveler.StorageNamespace, theirs, ours, base graveler.MetaRangeID) (graveler.MetaRangeID, error) {
+	diffIt, err := c.Diff(ctx, ns, theirs, ours)
+	if err != nil {
+		return "", err
```
nopcoder, line 97: can use fmt.Errorf for each return err here and add context information like "diff: %w", "merge iterator: %w"
it will help understand the flow when something goes wrong

Issue #1164, opened 2021-01-07
Objects physical path inference should be implemented by the adapter and called from the outside


nopcoder:
For each object operation, we resolve the repository's namespace and path to the object store bucket and path.
This operation can be done outside each implementation and pass the parsed parts to each request.

itaiad200:
Moreover - the adapter implementation should also format the underlining storage url.

PR #1204, opened 2021-01-17
Add physical address to the stat response

itaiad200:


api/api_controller.go
```diff
@@ -902,11 +902,12 @@ func (c *Controller) ObjectsStatObjectHandler() objects.StatObjectHandler {
...
 		obj := &models.ObjectStats{
-			Checksum:  entry.Checksum,
-			Mtime:     entry.CreationDate.Unix(),
-			Path:      params.Path,
-			PathType:  models.ObjectStatsPathTypeObject,
-			SizeBytes: entry.Size,
+			Checksum:        entry.Checksum,
+			Mtime:           entry.CreationDate.Unix(),
+			Path:            params.Path,
+			PhysicalAddress: entry.PhysicalAddress,
```
ozkatz, line 908: Our use of the `PhysicalAddress` internally is kind of weird: it might be a fully qualified URI (i.e. `s3://foo/bar/baz`) or it might be a non-qualified string (i.e. `bar/baz`). The later is what happens when you write an object through the API or S3 Gateway. The former is what happens when you do `lakefs import`.

The idea is that if it isn't a URI, we append it to the repository's storage namespace, otherwise we pass it to the block adapter as is. If we don't want API clients to have to reinvent this logic, we can simply always return the full URI. This means wrapping every returned physical address with `block.ResolveNamespace`. I suggest we do that.

ozkatz (changes requested):
I suggest wrapping all PhysicalAddress fields with `block.ResolveNamespace` to ensure we return a consistent usable address regardless of whether or not the object was born inside lakeFS or not.

Additionally, I think we should open an issue to support this field in the CLI as well (at least for `lakectl fs stat`

api/api_controller.go
itaiad200: Done.

itaiad200: And added an issue to achieve the same with lakectl.

api/api_controller.go
```diff
@@ -1138,25 +1148,37 @@ func (c *Controller) ObjectsListObjectsHandler() objects.ListObjectsHandler {
...
+			if err != nil {
+				return objects.NewStatObjectDefault(http.StatusInternalServerError).WithPayload(responseErrorFrom(err))
+			}
+
 			if entry.CommonLevel {
 				objList[i] = &models.ObjectStats{
-					Path:     entry.Path,
-					PathType: models.ObjectStatsPathTypeCommonPrefix,
+					Path:            entry.Path,
+					PhysicalAddress: qk.Format(),
```
ozkatz, line 1167: This adds a physical address to common prefixes as well, which is a little confusing (it simply returns the storage namespace of the repository). This could be misleading since that prefix might contain objects that don't belong in the storage namespace.

block/namespace.go
```diff
@@ -32,6 +33,24 @@ type QualifiedPrefix struct {
 	Prefix           string
 }
 
+func (qk QualifiedKey) Format() string {
+	scheme := ""
+	switch qk.StorageType {
+	case StorageTypeMem:
```
ozkatz, line 39: Probably out of scope for this PR, but ideally each block adapter should define its oun URI formatter that this function would then use (or better, a formatter that accepts a QualifiedKey and formats it). This way, adding a new block adapter type requires also updating the `QualifiedKey` struct methods - a weird coupling to make.

ozkatz (changes requested):
Looks great! immediately useful for visibility of the system too. The only thing I'd change is to remove the physical address field from common prefixes when listing objects in the API.

block/namespace.go
itaiad200: #1164

ozkatz (approved):
LGTM, Thanks!!

PR #1205, opened 2021-01-17
Correct slice body when reporting a response for trace

arielshaqed:
This actually crashes when the response is short-ish, need to do this.


guy-har (approved):

arielshaqed:
Thanks!  Sorry.

PR #1209, opened 2021-01-17
fix merge bugs

johnnyaug:


nopcoder (approved):
always miss the iterator close

nopcoder (approved):

PR #1214, opened 2021-01-17
Add a plumbing command to read graveler sstable files

ozkatz:


cmd/lakectl/cmd/sst.go
```diff
@@ -0,0 +1,180 @@
...
+		size: r.data.Size(),
+	}, nil
+}
+
+func getIterFromFile(filePath string) (committed.ValueIterator, map[string]string, error) {
+	var file sst.ReadableFile
+	var err error
+	// read from stdin (file has to be seekable/stat-able so we have this weird wrapper
+	if filePath == "-" {
+		data, err := ioutil.ReadAll(os.Stdin)
```
arielshaqed, line 53: This is scary because of memory consumption.  But there is no reason to do that.
1. Many times standard input *will* be seekable, just try.
2. When standard input isn't seekable, copy it to a temporary file and read _that_.

cmd/lakectl/cmd/sst.go
```diff
@@ -0,0 +1,180 @@
+package cmd
+
+import (
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"time"
+
+	sst "github.com/cockroachdb/pebble/sstable"
```
arielshaqed, line 10: We call it `pebblesst` in `entry_cataloger.go`.

cmd/lakectl/cmd/sst.go
```diff
@@ -0,0 +1,180 @@
...
+	iter, err := reader.NewIter(nil, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+	// wrap it in a Graveler iterator
+	dummyDeref := func() error { return nil }
+	return sstable.NewIterator(iter, dummyDeref), reader.Properties.UserProperties, nil
+}
+
+func writeRangeSSTable(iter committed.ValueIterator, amount int) *Table {
```
arielshaqed, line 81: This name is quite misleading, I was expecting it to write an SSTable.

cmd/lakectl/cmd/sst.go
```diff
@@ -0,0 +1,180 @@
...
+			panic(err)
+		}
+		ent, err := rocks.ValueToEntry(gv)
+		if err != nil {
+			panic(err)
+		}
+		rows = append(rows, []interface{}{
+			string(v.Key),
+			fmt.Sprintf("%x", gv.Identity),
+			fmt.Sprint(ent.GetSize()),
```
arielshaqed, line 99: Why convert a number to a string, if it's going to be an `interface{}` anyway?

cmd/lakectl/cmd/sst.go
```diff
@@ -0,0 +1,180 @@
...
+	return &Table{
+		Headers: []interface{}{"range_id", "min_key", "max_key", "count", "estimated_size"},
+		Rows:    rows,
+	}
+}
+
+var sstCatTemplate = `{{ .Table | table }}`
+
+var sstCmd = &cobra.Command{
+	Use:    "cat-sst <sst-file>",
```
arielshaqed, line 144: I'd like to have, at last eventually:
* `cat-sst --file <sst-file>`
* `cat-sst --metarange <metarange_id>`
* `cat-sst --range <range_id>`.

To allow that in future, could this be a flag arg rather than positional?

cmd/lakectl/cmd/sst.go
```diff
@@ -0,0 +1,180 @@
...
+		amount, _ := cmd.Flags().GetInt("amount")
+		iter, props, err := getIterFromFile(args[0])
+		if err != nil {
+			DieErr(err)
+		}
+		defer iter.Close()
+
+		// get props
+		typ, ok := props[committed.MetadataTypeKey]
+		if !ok || (typ != committed.MetadataRangesType && typ != committed.MetadataMetarangesType) {
```
arielshaqed, line 158: `switch` here instead.

arielshaqed (commented):
Thanks!  This is very cool.

Required changes are just:
1. use a flag arg for sst filename
2. don't read entire files into memory.

arielshaqed (changes requested):
I had one job, to request changes....

cmd/lakectl/cmd/sst.go
ozkatz: because `go-pretty/table` formats something very weird for numbers if I leave them as is.

cmd/lakectl/cmd/sst.go
```diff
@@ -0,0 +1,180 @@
...
+		rows = append(rows, []interface{}{
+			string(v.Key),
+			fmt.Sprintf("%x", gv.Identity),
+			fmt.Sprint(ent.GetSize()),
+			fmt.Sprint(ent.GetETag()),
+			ent.GetLastModified().AsTime().String(),
+			ent.GetAddress(),
+			fmt.Sprint(ent.GetMetadata()),
+		})
+	}
```
guy-har, line 105:
```suggestion
	}
	if err := iter.Err(); err != nil {
		panic(err)
	}
```

cmd/lakectl/cmd/sst.go
```diff
@@ -0,0 +1,180 @@
...
+		}
+		rows = append(rows, []interface{}{
+			committed.ID(gv.Identity),
+			string(r.MinKey),
+			string(r.MaxKey),
+			fmt.Sprint(r.Count),
+			fmt.Sprint(r.EstimatedSize),
+		})
+	}
+	return &Table{
```
guy-har, line 135:
```suggestion
	}
	if err := iter.Err(); err != nil {
		panic(err)
	}
	return &Table{
```

guy-har (commented):
Great tool, very helpful, Thanks!

arielshaqed:
> Thanks! This is very cool.

To expand: I wish I'd had this tool when we started integration, it is immediately useful for observing the system.

cmd/lakectl/cmd/sst.go
```diff
@@ -1,70 +1,44 @@
...
 	if filePath == "-" {
-		data, err := ioutil.ReadAll(os.Stdin)
+		fh, err := ioutil.TempFile("", "cat-sst-*")
 		if err != nil {
-			return nil, nil, err
+			return nil, nil, fmt.Errorf("error creating tempfile: %w", err)
 		}
-		file = &InMemReadableFile{
-			bytes.NewReader(data),
+		_, err = io.Copy(fh, os.Stdin)
```
arielshaqed, line 25: Cool.  Can you additionally add a TODO or open as issue to just keep `os.Stdin` if it is seekable?  (I think it will be enough to seek to 0 from the current position to determine whether an open file descriptor is seekable.)  It's something to make the tool a bit more convenient, but certainly not blocking.

arielshaqed (approved):
Thanks!  As mentioned above, this tool will be useful _even last week_...

guy-har (approved):

cmd/lakectl/cmd/sst.go
ozkatz: I assumed most people using stdin will read from a pipe but that's indeed not necessarily the case.
Implemeneted your suggestion.

arielshaqed (approved):
# :+1: :100:

PR #1226, opened 2021-01-18
Fix diff garbage result

guy-har:


johnnyaug (approved):

graveler/graveler.go
```diff
@@ -475,6 +475,12 @@ func (id Ref) String() string {
 	return string(id)
 }
 
+func (id Key) Copy() []byte {
```
ozkatz, line 478: shouldn't it return a `Key` instead?

ozkatz (changes requested):

graveler/graveler.go
guy-har: Yes it should, Thanks!

ozkatz (approved):
Thanks!

PR #1229, opened 2021-01-18
Migrate MVCC database information into new entry catalog format

### LINKED ISSUE #1163: Migrate data from mvcc catalog format to graveler based catalog

nopcoder:
Close #1163

### DIFF
--- a/catalog/migrate/iterator.go
+++ b/catalog/migrate/iterator.go
@@ -1,0 +1,148 @@
+package migrate
+
+import (
+	"context"
+	"errors"
+
+	"github.com/treeverse/lakefs/catalog"
+	"github.com/treeverse/lakefs/catalog/mvcc"
+	"github.com/treeverse/lakefs/catalog/rocks"
+	"github.com/treeverse/lakefs/db"
+)
+
+type iteratorState int
+
+const (
+	iteratorStateInit iteratorState = iota
+	iteratorStateQuerying
+	iteratorStateDone
+)
+
+type Iterator struct {
+	state     iteratorState
+	ctx       context.Context
+	db        db.Database
+	branchID  int64
+	commitID  int64
+	buf       []*rocks.EntryRecord
+	err       error
+	offset    string
+	fetchSize int
+	value     *rocks.EntryRecord
+}
+
+var ErrIteratorClosed = errors.New("iterator closed")
+
+// NewIterator returns an iterator over an mvcc branch/commit, giving entries as EntryCatalog.
+func NewIterator(ctx context.Context, db db.Database, branchID int64, commitID int64, fetchSize int) *Iterator {
+	return &Iterator{
+		ctx:       ctx,
+		db:        db,
+		branchID:  branchID,
+		commitID:  commitID,
+		buf:       make([]*rocks.EntryRecord, 0, fetchSize),
+		fetchSize: fetchSize,
+	}
+}
+
+func (it *Iterator) Next() bool {
+	if it.err != nil {
+		return false
+	}
+
+	it.maybeFetch()
+
+	// stage a value and increment offset
+	if len(it.buf) == 0 {
+		return false
+	}
+	it.value = it.buf[0]
+	it.buf = it.buf[1:]
+	it.offset = string(it.value.Path)
+	return true
+}
+
+func (it *Iterator) SeekGE(id rocks.Path) {
+	if errors.Is(it.err, ErrIteratorClosed) {
+		return
+	}
+	it.state = iteratorStateInit
+	it.offset = id.String()
+	it.buf = it.buf[:0]
+	it.value = nil
+	it.err = nil
+}
+
+func (it *Iterator) Value() *rocks.EntryRecord {
+	if it.err != nil {
+		return nil
+	}
+	return it.value
+}
+
+func (it *Iterator) Err() error {
+	return it.err
+}
+
+func (it *Iterator) Close() {
+	it.buf = nil
+	it.state = iteratorStateDone
+	it.err = ErrIteratorClosed
+}
+
+func (it *Iterator) maybeFetch() {
+	if it.state == iteratorStateDone {
+		return
+	}
+	if len(it.buf) > 0 {
+		return
+	}
+	if it.state == iteratorStateInit {
+		it.state = iteratorStateQuerying
+	}
+
+	var res interface{}
+	res, it.err = it.db.Transact(func(tx db.Tx) (interface{}, error) {
+		return mvcc.ListEntriesTx(tx, it.branchID, mvcc.CommitID(it.commitID), "", it.offset, it.fetchSize)
+	}, db.WithContext(it.ctx), db.ReadOnly())
+	if it.err != nil {
+		return
+	}
+	entries := res.([]*catalog.Entry)
+	for _, entry := range entries {
+		rec := &rocks.EntryRecord{
+			Path:  rocks.Path(entry.Path),
+			Entry: rocks.EntryFromCatalogEntry(*entry),
+		}
+		it.buf = append(it.buf, rec)
+	}
+	if len(it.buf) < it.fetchSize {
+		it.state = iteratorStateDone
+	} else if len(it.buf) > 0 {
+		it.offset = it.buf[len(it.buf)-1].Path.String()
+	}
+}
+
+type emptyIterator struct{}
+
+func (e *emptyIterator) Next() bool {
+	return false
+}
+
+func (e *emptyIterator) SeekGE(rocks.Path) {
+}
+
+func (e *emptyIterator) Value() *rocks.EntryRecord {
+	return nil
+}
+
+func (e *emptyIterator) Err() error {
+	return nil
+}
+
+func (e *emptyIterator) Close() {
+}
+
+func newEmptyIterator() rocks.EntryIterator {
+	return &emptyIterator{}
+}
--- a/catalog/migrate/migrate.go
+++ b/catalog/migrate/migrate.go
@@ -1,0 +1,363 @@
+package migrate
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"regexp"
+	"time"
+
+	"github.com/hashicorp/go-multierror"
+	"github.com/jackc/pgx/v4"
+	"github.com/treeverse/lakefs/catalog"
+	"github.com/treeverse/lakefs/catalog/mvcc"
+	"github.com/treeverse/lakefs/catalog/rocks"
+	"github.com/treeverse/lakefs/db"
+	"github.com/treeverse/lakefs/graveler"
+	"github.com/treeverse/lakefs/logging"
+)
+
+type Migrate struct {
+	db            db.Database
+	mvccCataloger catalog.Cataloger
+	entryCatalog  *rocks.EntryCatalog
+	repositoryRe  *regexp.Regexp
+	log           logging.Logger
+
+	// per repository state
+	lastCommit graveler.CommitID
+	branches   map[graveler.BranchID]graveler.CommitID
+	tags       map[graveler.TagID]graveler.CommitID
+}
+
+type commitRecord struct {
+	BranchID              int64
+	BranchName            string
+	CommitID              int64
+	PreviousCommitID      int64
+	Committer             string
+	Message               string
+	CreationDate          time.Time
+	Metadata              catalog.Metadata
+	MergeType             string
+	MergeSourceBranch     *int64
+	MergeSourceBranchName string
+	MergeSourceCommit     *int64
+}
+
+const (
+	migrateFetchSize = 1000
+
+	initialCommitter     = "migrate"
+	initialCommitMessage = "Create empty new branch for migrate"
+)
+
+func (m *Migrate) Close() error {
+	return m.mvccCataloger.Close()
+}
+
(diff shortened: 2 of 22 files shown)


catalog/migrate/iterator.go
```diff
@@ -0,0 +1,165 @@
...
+	buf       []*rocks.EntryRecord
+	err       error
+	offset    string
+	fetchSize int
+	value     *rocks.EntryRecord
+}
+
+var ErrIteratorClosed = errors.New("iterator closed")
+
+// NewIterator accepts mvcc branch/commit, read entries as EntryCatalog entries
```
arielshaqed, line 37:
```suggestion
// NewIterator returns an iterator over an mvcc branch/commit, giving entries as EntryCatalog.
```

catalog/migrate/iterator.go
```diff
@@ -0,0 +1,165 @@
...
+		db:        db,
+		branchID:  branchID,
+		commitID:  commitID,
+		buf:       make([]*rocks.EntryRecord, 0, fetchSize),
+		fetchSize: fetchSize,
+	}
+}
+
+func (it *Iterator) Next() bool {
+	if it.state == iteratorStateClosed {
```
arielshaqed, line 50: Could this just be a special error, stored in `it.err`?

catalog/migrate/iterator.go
```diff
@@ -0,0 +1,165 @@
...
+	it.state = iteratorStateClosed
+	it.err = ErrIteratorClosed
+}
+
+func (it *Iterator) fetch() {
+	if it.state == iteratorStateDone {
+		return
+	}
+	if len(it.buf) > 0 {
+		return
```
arielshaqed, line 115: Consider calling it `fetchIfEmpty` or `maybeFetch`; will make reading the call to it easier.

catalog/migrate/migrate.go
```diff
@@ -0,0 +1,333 @@
...
+	if err != nil {
+		return err
+	}
+	m.repositoryRe = repositoryRe
+	return nil
+}
+
+func (m *Migrate) Run() error {
+	ctx := context.Background()
+	// logger := logging.FromContext(ctx)
```
arielshaqed, line 86: Let's remove this.

catalog/migrate/migrate.go
```diff
@@ -0,0 +1,333 @@
...
+	fmt.Println("==> Listing repositories to import")
```
arielshaqed, line 87: I know this is a CLI, but can we log?  Alternatively, you could write to `fmt.Print*` under `cmd`, I guess (pass a "log it" callback to this function).

catalog/migrate/migrate.go
```diff
@@ -0,0 +1,333 @@
...
+}
+
+func (m *Migrate) getOrCreateTargetRepository(ctx context.Context, repository *catalog.Repository) (*graveler.RepositoryRecord, error) {
+	repoID := graveler.RepositoryID(repository.Name)
+	branchID := graveler.BranchID(repository.DefaultBranch)
+
+	// check and create repository if needed
+	repo, err := m.entryCatalog.GetRepository(ctx, repoID)
+	if errors.Is(err, graveler.ErrRepositoryNotFound) {
+		repo, err = m.entryCatalog.CreateRepository(ctx, repoID, graveler.StorageNamespace(repository.StorageNamespace), branchID)
```
arielshaqed, line 148: Consider just creating the repo and ignoring an "already exists" error.

catalog/migrate/migrate.go
```diff
@@ -0,0 +1,333 @@
...
+	if err != nil {
+		return nil, err
+	}
+	m.lastCommit = branch.CommitID
+	if m.lastCommit == "" {
+		metaRangeID, err := m.entryCatalog.WriteMetaRange(ctx, repoID, newEmptyIterator())
+		if err != nil {
+			return nil, err
+		}
+		commitID, err := m.entryCatalog.CommitExistingMetaRange(ctx, repoID, branchID, *metaRangeID, initialCommitter, initialCommitMessage, nil)
```
arielshaqed, line 168: Let's have a more informative commit message.  Perhaps "Create empty new branch for import" or something.

catalog/migrate/migrate.go
```diff
@@ -0,0 +1,333 @@
...
+
+	return &graveler.RepositoryRecord{
+		RepositoryID: graveler.RepositoryID(repository.Name),
+		Repository:   repo,
+	}, nil
+}
+
+// importCommit based on the commit record type
+//  - list entries, create meta range ID and commit it to our repository.
+//  - add a tag with the mvcc commit id to our new repository
```
arielshaqed, line 188: Can we godocize this comment?

catalog/migrate/migrate.go
```diff
@@ -0,0 +1,333 @@
...
+		ORDER BY commit_id`,
+		mvccRepoID)
+	if err != nil {
+		return nil, fmt.Errorf("select commits (%s): %w", repo.RepositoryID, err)
+	}
+	return rows, nil
+}
+
+func (c *commitRecord) Scan(rows pgx.Row) error {
+	return rows.Scan(c.BranchID, c.BranchName, c.CommitID,
```
arielshaqed, line 330: I don't understand Go, AFAICT this cannot fill in most of the fields on c: how can this scan into e.g. c.CommitID; didn't you just take its _value_?

catalog/rocks/entry_catalog.go
```diff
@@ -134,6 +134,15 @@ func (e *EntryCatalog) CommitExistingMetaRange(ctx context.Context, repositoryID
 	return e.store.CommitExistingMetaRange(ctx, repositoryID, branchID, metaRangeID, committer, message, metadata)
 }
 
+func (e *EntryCatalog) AddCommitNoLock(ctx context.Context, repositoryID graveler.RepositoryID, commit graveler.Commit) (graveler.CommitID, error) {
```
arielshaqed, line 137: Please document (any function that contains the words "NoLock").

ozkatz (commented):
A big PR - I'm a bit worried about my lack of ability to properly test this before going through the code.
I think a good test that needs to happen here is to import a big S3 inventory (we have one of those laying around), do a few commits on top, that change an arbitrary amount of keys (between 100s to 1E6s of keys per commit). Add a few weird branching/merging scenarios on top and see that we can a functionally similar system after migration.

catalog/rocks/validate.go
```diff
@@ -4,13 +4,16 @@ import (
...
+	"strings"
 
 	"github.com/treeverse/lakefs/graveler"
 	"github.com/treeverse/lakefs/ident"
 )
 
 const (
 	MaxPathLength = 1024
+
+	minControlCharCode = 40
```
arielshaqed, line 16: This is the char code for '8'.  The closest I can think of is `maxControlCharCode = 040-1`, assuming golang has octals.

catalog/rocks/validate.go
```diff
@@ -105,12 +108,26 @@ func ValidateTagID(v interface{}) error {
...
 		return ErrRequiredValue
 	}
-	if !reValidBranchID.MatchString(string(s)) {
+	if strings.HasSuffix(tag, ".") || strings.HasSuffix(tag, ".lock") || strings.HasSuffix(tag, "/") {
+		return ErrInvalidValue
+	}
+	if strings.Contains(tag, "..") || strings.Contains(tag, "//") || strings.Contains(tag, "@{") {
 		return ErrInvalidValue
 	}
+	// Not like git we do enable '~' to support migrate from our previous ref format where commit started with tilda
```
arielshaqed, line 122:
```suggestion
	// Unlike git, we do allow '~'.  That supports migration from our previous ref format where commits started with a tilde.
```

catalog/rocks/validate.go
```diff
@@ -105,12 +108,26 @@ func ValidateTagID(v interface{}) error {
...
 		panic(ErrInvalidType)
 	}
-	if len(s) == 0 {
+	// http://git-scm.com/docs/git-check-ref-format
+	tag := string(s)
+	if len(tag) == 0 || tag == "@" {
 		return ErrRequiredValue
 	}
-	if !reValidBranchID.MatchString(string(s)) {
+	if strings.HasSuffix(tag, ".") || strings.HasSuffix(tag, ".lock") || strings.HasSuffix(tag, "/") {
```
arielshaqed, line 116: I don't understand what a "tag" is.  If this is the user-visible name for a tag, then it seems like this function imposes arbitrary limitations on the vast majority of users, who will never migrate from mvcc.

arielshaqed (changes requested):
Didn't understand multiple things, sorry. :-(

catalog/migrate/migrate.go
nopcoder: will use logger and aggregate the errors for the cli

catalog/migrate/migrate.go
nopcoder: I can do it, but it will modify the behavior a little - when a repository exists I use the information to perform all graveler/entry catalog API - which means that you can create a repository point to a new storage namespace and import.
The import to the new storage namespace.
In this case, it is working before I get the information, if I switch to create - I will need to re-read the repository information after getting 'repository already exists'. WDYT?

catalog/migrate/migrate.go
nopcoder: You are right, I make this change in the last sec and I didn't have a test to verify the import.

catalog/rocks/validate.go
nopcoder: True - it should be 040 or octa40

catalog/rocks/validate.go
nopcoder: Took the same limitations as git in this case - we can lose some of them as we are git like.

catalog/rocks/entry_catalog.go
nopcoder: True, I documented just the interface 🤔

nopcoder:
> Didn't understand multiple things, sorry. :-(

Nothing to sorry - the more the merrier.

catalog/migrate/iterator.go
```diff
@@ -0,0 +1,152 @@
...
+	if len(it.buf) == 0 {
+		return false
+	}
+	it.value = it.buf[0]
+	it.offset = string(it.value.Path)
+	if len(it.buf) > 1 {
+		it.buf = it.buf[1:]
+	} else {
+		it.buf = it.buf[:0]
+	}
```
arielshaqed, line 65: We know `len(it.buf) >= 1`, so isn't this just
```suggestion
	it.buf = it.buf[1:]
```
:question: 

I can see no performance advantage to `it.buf = it.buf[:0]`, given that by the time `len(it.buf)==1` that slice is anyway a _subslice_ of the allocated slice.

catalog/migrate/migrate.go
arielshaqed: Yeah, you're right.  Your way is better, my way is SQL on ACID.

catalog/rocks/validate.go
arielshaqed: Oh, I see!

catalog/rocks/validate.go
```diff
@@ -105,12 +108,29 @@ func ValidateTagID(v interface{}) error {
...
+	}
+	if strings.Contains(tag, "..") || strings.Contains(tag, "//") || strings.Contains(tag, "@{") {
+		return ErrInvalidValue
+	}
+	// Unlike git, we do allow '~'.  That supports migration from our previous ref format where commits started with a tilde.
+	if strings.ContainsAny(tag, "^:?*[\\") {
 		return ErrInvalidValue
 	}
+	for _, r := range tag {
+		if r <= minControlCharCode {
```
arielshaqed, line 130: `minControlCode` is misnamed as space is not a control code.  Perhaps replace the constant with a named function,
```go
func isControlCodeOrSpace(r rune) {
	return r <= 0x20
}
```
(Octal hasn't made sense since PDP-8)

graveler/graveler.go
```diff
@@ -238,6 +247,10 @@ type VersionController interface {
 	// Returns ErrTreeNotFound if the referenced treeID doesn't exist.
 	CommitExistingMetaRange(ctx context.Context, repositoryID RepositoryID, branchID BranchID, metaRangeID MetaRangeID, committer string, message string, metadata Metadata) (CommitID, error)
 
+	// AddCommitNoLock creates a commit and associated it to the repository without locking metadata update.
```
arielshaqed, line 250:
```suggestion
	// AddCommitNoLock creates a commit and associates it with the repository without locking metadata for update.
```

arielshaqed (approved):
Cool, thanks!

Is there _any_ way that we can test this?  If not via a unit test, perhaps Nessie could do it?

I _know_ it will only run once -- but we need to provide some confidence that it will work.  Alternatively, we might guide users how to validate that migration worked.  Finally, we just might be able to _ask_ all our users whether anyone has MVCC in production, and maybe if none of them do it will be safe enough.

catalog/rocks/validate.go
nopcoder: cool - took it.

catalog/migrate/iterator.go
nopcoder: totally right, I'll align all the rest of the code that uses this way.

nopcoder:
Had to rebase with master.
- Addressed the noCache check Exists
- Created a basic test that uses import

PR #1230, opened 2021-01-18
show only 16 first hex chars for commits in the UI

ozkatz:
At least where it makes sense to do so (lists, dropdowns)

### DIFF
--- a/webui/src/components/ComparePage.js
+++ b/webui/src/components/ComparePage.js
@@ -99,7 +99,7 @@ const CompareToolbar = ({repo, refId, compare, refresh}) => {
                     params.set('branch', ref.id);
                     params.delete('commit'); // if we explicitly selected a branch, remove an existing commit if any
                 } else {
-                    params.set('commit', ref.id);
+                    params.set('commit', ref.id.slice(0, 16));
                     params.delete('branch'); // if we explicitly selected a commit, remove an existing branch if any
                 }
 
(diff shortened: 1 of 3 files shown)


ozkatz:
@nopcoder not sure the length of MVCC commit IDs - I hope this doesn't break them. Can you confirm?

webui/src/components/ComparePage.js
```diff
@@ -99,7 +99,7 @@ const CompareToolbar = ({repo, refId, compare, refresh}) => {
                     params.set('branch', ref.id);
                     params.delete('commit'); // if we explicitly selected a branch, remove an existing commit if any
                 } else {
-                    params.set('commit', ref.id);
+                    params.set('commit', ref.id.slice(0, 16));
```
nopcoder, line 102: This one is not the UI, right? it will pass it as a short version of the commit.
if this is the case. I think we should hold just with this one, so we will still work with mvcc.

ozkatz: it controls what gets displayed in the button that opens the ref selector after picking a commit. It might possibly control the URI of the resulting request to diff as well

ozkatz:
@nopcoder to clarify, if MVCC commit IDs are potentially longer, we should not pull this. Even if the UI is still functional but users can no longer safely copy-paste IDs from it, it's probably not a desirable change before MVCC depracation.

nopcoder:
> @nopcoder to clarify, if MVCC commit IDs are potentially longer, we should not pull this. Even if the UI is still functional but users can no longer safely copy-paste IDs from it, it's probably not a desirable change before MVCC depreciation.

MVCC commits are bigger, but I still think we can support it. Partial ID or adding it ...as an indicator + UI option to copy the full ID is a valid option as I see it.

nopcoder:
@ozkatz the following is the part I didn't want to commit yet. It will work only on non-mvcc:

image

ozkatz:
closing for now. Will redo this post release of lakeFS on the rocks deliverable #1

PR #1235, opened 2021-01-18
sort commits before finding commit by ref

johnnyaug:


graveler/ref/manager.go
```diff
@@ -214,6 +214,7 @@ func (m *Manager) GetCommitByPrefix(ctx context.Context, repositoryID graveler.R
 					SELECT id, committer, message, creation_date, parents, meta_range_id, metadata
 					FROM graveler_commits
 					WHERE repository_id = $1 AND id >= $2
+					ORDER BY id
```
arielshaqed, line 217: I don't understand this change (but, to the PR's defence, I had several rounds with @ozkatz around this very SQL query).  This query asks for at most 2 rows; if there is more than one row then l.235 returns `ErrRefAmbiguous`.  So... AFAIU either `GetCommitByPrefix` fetches and returns a single row or it returns an error.  As such order *cannot* matter.

arielshaqed (changes requested):
Not sure this PR has any effect, but it might make things slower.  @ozkatz perhaps the comment would be more noticeable if be placed *after* the query?

johnnyaug:
The ordering is performed before the limit. Without the order by clause,
the query may return any 2 rows with keys greater than the prefix

graveler/ref/manager.go
johnnyaug: The ordering is performed before the limit. Without the order by clause,
the query may return any 2 rows with keys greater than the prefix

graveler/ref/manager.go
```diff
@@ -213,7 +213,7 @@ func (m *Manager) GetCommitByPrefix(ctx context.Context, repositoryID graveler.R
 		err := tx.Select(&records, `
 					SELECT id, committer, message, creation_date, parents, meta_range_id, metadata
 					FROM graveler_commits
-					WHERE repository_id = $1 AND id >= $2
+					WHERE repository_id = $1 AND id LIKE $2 || '%'
```
arielshaqed, line 216: This is the important part of the fix, right?  LGTM!

arielshaqed (approved):
Nice!  As discussed f2f, the worry is that a database (let's call it PostgreSQeviL) might always return the last 2 commit digests, neither of which has the prefix, when there is precisely one prefix.
Maybe add a test?

Also, I think maybe "addressProvider" is from another change?  So I did *not* review it, I'm assuming you will undo before pushing.

THANKS!

johnnyaug:
@arielshaqed actually the addressProvider was added in order to be able to test this. I'm actually not sure it's worth it, but I wanted to add a test. WDYT?

arielshaqed (approved):
OK, now I see it!  That's actually really neat, and it allows us (you) to test that things work.  THANKS!

PR #1239, opened 2021-01-19
Distinction between two-dot and three-dot ("Compare") diffs

### LINKED ISSUE #1208: In Compare view, the diff is reversed, and is not a three way diff
1. The diff is reversed:
Reproduce: create a side branch, add a file to it, and compare it with master.
The file is presented as removed (see photo)
image

2. The diff is not a three way diff: reverse the compare direction, and you will still see the file

johnnyaug:
Closes #1208 

This change aligns lakeFS with git conventions regarding diff: it introduces the distinction between diff types:
1. two-dot / "commit-ish" / simple diff
2. three-dot / "tree-ish" / three-way diff.

For naming the methods under graveler, I've used "diff" to denote two-dot diff, and "compare" for the the three-dot diff.
The implementation of diff in the mvcc implementation is a three-way diff, and hence was changed to "compare".

In terms of behavior, everything stays the same after this change (except for fixing #1208 , which is a bug).


catalog/mvcc/cataloger_list_commits_test.go
```diff
@@ -373,7 +373,7 @@ func TestCataloger_ListCommits_LineageFromChild(t *testing.T) {
 		t.Fatal("Merge Summary", diff)
 	}
 	// TODO(barak): enable test after diff between commits is supported
```
ozkatz, line 375: so I think we should enable this test then? @nopcoder @johnnyaug

catalog/mvcc/cataloger_merge_test.go
```diff
@@ -82,7 +82,7 @@ func TestCataloger_Merge_FromParentNoChangesInChild(t *testing.T) {
 		t.Fatal("Merge() expected ErrNoDifferenceWasFound, got:", err)
 	}
 	// TODO(barak): enable test after diff between commits is supported
-	//differences, _, err := c.Diff(ctx, repository, commitLog.Parents[0], commitLog.Parents[1], -1, "")
```
ozkatz, line 85: enable all these tests?

export/export_handler.go
```diff
@@ -93,7 +93,7 @@ func (h *Handler) generateTasks(startData StartData, config catalog.ExportConfig
 			diffs, hasMore, err = getDiffFromBase(context.Background(), startData.Repo, startData.ToCommitRef, after, limit, h.cataloger)
 		} else {
 			// TODO(guys) change this to work with diff iterator once it is available outside of cataloger
```
ozkatz, line 95: this comment is outdated then?

graveler/committed/compare_iterator_test.go
```diff
@@ -0,0 +1 @@
+package committed_test
```
ozkatz, line 1: unless we have tests, I think this file can be removed

graveler/committed/merge_iterator.go
```diff
@@ -6,22 +6,31 @@ import (
...
 }
 
 // NewMergeIterator accepts an iterator describing a diff from theirs to ours.
 // It returns a ValueIterator with the changes to perform on theirs, in order to merge ours into it,
 // relative to base as the merge base.
 // The iterator will return ErrConflictFound when it reaches a conflict.
-func NewMergeIterator(diffTheirsToOurs graveler.DiffIterator, base Iterator) (*mergeIterator, error) {
-	return &mergeIterator{diffIt: diffTheirsToOurs, base: base}, nil
+func NewMergeIterator(diffTheirsToOurs graveler.DiffIterator, base Iterator) *compareValueIterator {
+	return &compareValueIterator{compareIterator: &compareIterator{diffIt: diffTheirsToOurs, base: base, errorOnConflict: true}}
```
ozkatz, line 26: so if `errOnConflict` is true, a merge operation will fail as soon as a single conflict is found. At that point, the calling user would probably want to resolve conflicts, and probably not do it one conflict at a time. How can a user get the full list of conflicting objects?

swagger.yml
```diff
@@ -1653,6 +1657,10 @@ paths:
         name: amount
         type: integer
         default: 100
+      - in: query
+        name: type
+        type: string
+        <<: *DIFF_TYPE
```
ozkatz, line 1663: I don't understand the `<<: *...` notation :(

swagger.yml
```diff
@@ -150,6 +150,10 @@ definitions:
         type: string
         enum: [ common_prefix, object ]
 
+  diff_type: &DIFF_TYPE
```
ozkatz, line 153: What is being dereferenced here? I am confused.

ozkatz (approved):
I _think_ this looks good! a little hesitant since there are a couple of parts I don't fully understand.

graveler/committed/merge_iterator.go
johnnyaug: Then you use the compare iterator.
I'll add better comments

swagger.yml
johnnyaug: This is a swagger 2 syntax that allows you to reuse enum types.
In OpenAPI 3 they are reusable in a more straightforward way.
 
For more information: https://swagger.io/docs/specification/2-0/enums/

catalog/mvcc/cataloger_list_commits_test.go
johnnyaug: These tests use the MVCC implementation. Although we may consider trying to get them to work with the new one.
@nopcoder WDYT

export/export_handler.go
guy-har: Not yet, 
we still need to paginate over cataloger results instead of using an iterator.
will be outdated once we could get an iterator from the cataloger (or once we decide we don't want to support it).

ozkatz (approved):
Nice work with this!

graveler/committed/merge_iterator.go
```diff
@@ -6,22 +6,35 @@ import (
...
-// The iterator will return ErrConflictFound when it reaches a conflict.
-func NewMergeIterator(diffTheirsToOurs graveler.DiffIterator, base Iterator) (*mergeIterator, error) {
-	return &mergeIterator{diffIt: diffTheirsToOurs, base: base}, nil
+// When reaching a conflict, the iterator will enter an error state with the graveler.ErrConflictFound error.
+func NewMergeIterator(diffTheirsToOurs graveler.DiffIterator, base Iterator) *compareValueIterator {
+	return &compareValueIterator{compareIterator: &compareIterator{diffIt: diffTheirsToOurs, base: base, errorOnConflict: true}}
 }
 
-func (d *mergeIterator) valueFromBase(key graveler.Key) (*graveler.ValueRecord, error) {
+// NewMergeIterator accepts an iterator describing a diff from theirs to ours.
```
arielshaqed, line 29:
```suggestion
// NewCompareIterator accepts an iterator describing a diff from theirs to ours.
```
?

graveler/committed/merge_iterator.go
```diff
@@ -6,22 +6,35 @@ import (
...
+// It returns a graveler.DiffIterator with the changes to perform on theirs, in order to merge ours into it,
+// relative to base as the merge base.
+// When reaching a conflict, the returned Diff will be of type graveler.DiffTypeConflict.
+func NewCompareIterator(diffTheirsToOurs graveler.DiffIterator, base Iterator) *compareIterator {
```
arielshaqed, line 33: Really like the parameter names here!

graveler/committed/merge_iterator_test.go
```diff
@@ -265,6 +260,231 @@ func TestMergeSeek(t *testing.T) {
...
+			expectedDiffTypes:  []graveler.DiffType{conflict},
+		},
+		{
+			seekTo:             "k99",
+			expectedKeys:       nil,
+			expectedIdentities: nil,
+		},
+	}
+	for i, tst := range tests {
+		t.Run(fmt.Sprintf("test%d", i), func(t *testing.T) {
```
arielshaqed, line 458: Can we add a name here?  It makes it easier to understand what each case is testing.

arielshaqed (approved):
Thanks, nice how you can add something like this.  I was expected a LOT more code.

Nothing on the application level, just a few nits on the low levels (feel free to ignore).

johnnyaug:
Will address doneness separately later today

Issue #1238, opened 2021-01-19
Diff after merge fails inconsistently 


johnnyaug:
After performing a merge, diffing the two branches sometimes fails.
The error is related to an SSTable ID somehow becoming grabage.

Steps to reproduce:
1. Create the repo:
```
REPO_NAME=myrepo
lakectl repo create lakefs://${REPO_NAME} local://new/${REPO_NAME}
```
2. Upload and commit file to master:
```
lakectl fs upload -s ./some/local/file lakefs://${REPO_NAME}@master/aa
lakectl commit lakefs://${REPO_NAME}@master -m "mastercommit1"
```

3. Create a branch, upload to it, and commit:
```
lakectl branch create lakefs://${REPO_NAME}@ftr --source lakefs://${REPO_NAME}@master
lakectl fs upload -s ./some/local/file lakefs://${REPO_NAME}@ftr/bb
lakectl commit lakefs://${REPO_NAME}@ftr -m "ftrcommit1"
```

4. Merge branch to master:
```
lakectl merge lakefs://${REPO_NAME}@ftr lakefs://${REPO_NAME}@master
```

5. Do a diff:
```
lakectl diff lakefs://${REPO_NAME}@master lakefs://${REPO_NAME}@ftr
```
This command gives unconsistent results: sometimes it gives an error.
Note that due to another unfortunate bug (I'm on it), this error is swallowed. To see the error put a breakpoint in range_manager.go:108 (return nil ,err)

Example error:
```
open SSTable :��`�2pF\V�����+���'��"�e������ after fetch from next tier: fetch :��`�2pF\V�����+���'��"�e������ from next tier: read from block storage: open /home/local/data/new/myrepo/_lakefs/range/:��`�2pF\V�����+���'��"�e������: no such file or directory
```

arielshaqed:
[as discussed vf2vf, this is a release blocker and all]

arielshaqed:
I can occasionally _also_ receive:
```
+ added aa
+ added bb                                                                                          
```
when running multiple times concurrently!

```
[...]

PR #1244, opened 2021-01-19
Configure SSTable reader cache by default

arielshaqed:
Hook configuration up for `committed.sstable.memory.cache_size_bytes`.  Because it was
missing, the "fail-safe" noCache option was used... which is apparently unsafe or just exposes
some *other* issue with calling too many SSTable readers.

Relevant for #1238.


ozkatz (approved):
LGTM

arielshaqed:
Thanks! Will pull once tests pass.

PR #1245, opened 2021-01-19
Close pebble SSTable iterators only once

### LINKED ISSUE #1238: Diff after merge fails inconsistently

arielshaqed:
Unlike our iterators, it is unsafe to close a pebble SSTable multiple times.  (CockroachDB is
the most sensitive roach EVER.)  So don't do that.

Fixes #1238.  Apparently -- it's a flaky situation.

### DIFF
--- a/graveler/sstable/iterator.go
+++ b/graveler/sstable/iterator.go
@@ -61,11 +61,16 @@ func (iter *Iterator) Err() error {
 }
 
 func (iter *Iterator) Close() {
+	if iter.it == nil {
+		return
+	}
 	err := iter.it.Close()
 	iter.updateOnNilErr(err)
 
 	err = iter.derefer()
 	iter.updateOnNilErr(err)
+
+	iter.it = nil
 }
 
 func (iter *Iterator) updateOnNilErr(err error) {


arielshaqed:
Don't pull before #1244.

ozkatz:
Nice catch! I think we need to add a test to avoid regressions (at least here but possibly in other places where we wrap Pebble's iterators).
Alternatively, wrap Pebble with something that validates and errs when a `Close()` happens multiple times.

arielshaqed:
> Nice catch! I think we need to add a test to avoid regressions (at least here but possibly in other places where we wrap Pebble's iterators).

This is the only place where we wrap Pebble iterators.  Unfortunately there is nothing to test: if I `Close` this iterator multiple times, _nothing bad will happen_ -- even with the previous version.  I need to overload a `sync.Pool` in the pebble implementation somehow (don't know how, not even sure which of 2 _different_ pools), and then perform an action, and then somehow detemine that something bad happened.

Maybe in a system load test.

> Alternatively, wrap Pebble with something that validates and errs when a `Close()` happens multiple times.

Again, I am not sure where I could put such a wrapper.  It *is* legal to `Close` one of our iterators more than once.  So adding this wrapper -- and remembering to examine the error -- does rather little.

I *could* write just a wrapper for Cockroach Pebble iterators to make them safe to close twice.  We would still need to remember to use it.  Maybe fork and send them a PR, and/or vendor?

arielshaqed:
Hmmm.... looking for more details, it's a bug in our implementation.  So pebble Iterators may be closed any number of times: https://pkg.go.dev/github.com/cockroachdb/pebble@v0.0.0-20210119143827-dbe38cf9cbdc/internal/base#InternalIterator says:

```go
	// Close closes the iterator and returns any accumulated error. Exhausting
	// all the key/value pairs in a table is not considered to be an error.
	// It is valid to call Close multiple times. Other methods should not be
	// called after the iterator has been closed.
```

But https://pkg.go.dev/github.com/cockroachdb/pebble/sstable#pkg-overview also says:

> A reader can be used concurrently. Multiple goroutines can call NewIter concurrently, and each iterator can run concurrently with other iterators. However, any particular iterator should not be used concurrently, and ***iterators should not be used once a reader is closed.***

(Emphasis mine).

So if...
1. The same reader is shared between 2 iterators (_which can happen during a `diff` when iterating over a shared range_), and
2. both iterators are closed multiple times concurrently, and
3. no other iterator is currently attached to that range,
then we will unsafely use an iterator.  Even worse, we may end up closing the reader multiple times, which is possible dangerous (or not).

Open to suggestions how to test any of this.

ozkatz:
I don't have a good suggestion re how to properly test for this. Approving in the meantime. Thanks!

ozkatz (approved):

arielshaqed:
Thanks!  Pulling...

PR #1247, opened 2021-01-20
Fix error handling

guy-har:


graveler/committed/merge_iterator.go
```diff
@@ -91,6 +98,7 @@ func (d *mergeIterator) Next() bool {
 			// added on theirs, but not on ours - continue
 		}
 	}
+	d.err = d.diffIt.Err()
```
johnnyaug, line 101: The test fails because your change reveals a bug - that I forgot to return false after line 96, when the conflict is found.
Can you please add it as well?

johnnyaug (approved):
Nice!

graveler/committed/merge_iterator.go
```diff
@@ -21,24 +21,31 @@ func NewMergeIterator(diffTheirsToOurs graveler.DiffIterator, base Iterator) (*m
...
+	return val, nil
 }
 
 func (d *mergeIterator) Next() bool {
 	for d.diffIt.Next() {
 		val := d.diffIt.Value()
 		key := val.Key
 		typ := val.Type
-		baseVal := d.valueFromBase(key)
+		baseVal, err := d.valueFromBase(key)
```
nopcoder, line 44: think we need to set d.value to nil in case of an error or condition Value() if err

nopcoder (changes requested):

graveler/committed/apply.go
```diff
@@ -71,6 +71,12 @@ func Apply(ctx context.Context, writer MetaRangeWriter, source Iterator, diffs g
 			haveSource = source.Next()
 		}
 	}
+	if err := source.Err(); err != nil {
+		return err
+	}
+	if err := diffs.Err(); err != nil {
+		return err
+	}
```
arielshaqed, line 79: Consider the shorter
```suggestion
	if source.Err() {
		return source.Err()
	}
	if diffs.Err() {
		return diffs.Err()
	}
```
(and similarly below).

arielshaqed (approved):
Thanks!  Great catch in the basic iterator...

graveler/committed/apply.go
johnnyaug: python style

nopcoder (approved):

graveler/committed/apply.go
nopcoder: the first option captures the Err value in case Err can't be inlined or it is more than just a simple return.

PR #1266, opened 2021-01-21
Test per-range skip/enter in diff iterator

arielshaqed:
By explicitly forbidding diff iterator from entering some ranges we can be more explicit what
should happen where.  Indeed, this hints at a bug (marked) that a range is needlessly entered
after SeekGE.


arielshaqed:
Closing -- not worth the extra complexity.

Issue #1264, opened 2021-01-20
rocks: when merging 2 commits, shared ranges are opened by tierFS


ozkatz:
when running a merge operation (i.e. Apply) when logging level set to TRACE, it appears all range files (including those that appear in both commits) are being opened by TierFS. See log:

```
time="2021-01-20T17:29:25Z" level=trace msg="opened locally" func="pyramid.(*TierFS).Open" file="pyramid/tier_fs.go:211" filename=83523eb6cff072663e86b6806b923d16ed00f881fc3eba11e4148afa2a7374af module=pyramid namespace="local://" ns_path=""
time="2021-01-20T17:29:25Z" level=trace msg="opened locally" func="pyramid.(*TierFS).Open" file="pyramid/tier_fs.go:211" filename=83523eb6cff072663e86b6806b923d16ed00f881fc3eba11e4148afa2a7374af module=pyramid namespace="local://" ns_path=""
time="2021-01-20T17:29:25Z" level=trace msg="opened locally" func="pyramid.(*TierFS).Open" file="pyramid/tier_fs.go:211" filename=01d25069f8efe656083ef4fa95c6514d5043d263f9641fc290415db9d6a08b29 module=pyramid namespace="local://" ns_path=""
time="2021-01-20T17:29:25Z" level=trace msg="opened locally" func="pyramid.(*TierFS).Open" file="pyramid/tier_fs.go:211" filename=01d25069f8efe656083ef4fa95c6514d5043d263f9641fc290415db9d6a08b29 module=pyramid namespace="local://" ns_path=""
time="2021-01-20T17:29:25Z" level=trace msg="opened locally" func="pyramid.(*TierFS).Open" file="pyramid/tier_fs.go:211" filename=3878dcff1f54bcbd50ef4d6eb7ad0dfb7e36f4ec1bbb1080a8da7973ad836bb1 module=pyramid namespace="local://" ns_path=""
time="2021-01-20T17:29:25Z" level=trace msg="opened locally" func="pyramid.(*TierFS).Open" file="pyramid/tier_fs.go:211" filename=3878dcff1f54bcbd50ef4d6eb7ad0dfb7e36f4ec1bbb1080a8da7973ad836bb1 module=pyramid namespace="local://" ns_path=""
[...]

Issue #1240, opened 2021-01-19
Graveler | Slow commands


guy-har:
When running lakeFS an EC2 instance of type `t2.xlarge` with a large dataset
I noticed that it takes too much time for some commands to run.

commit (after adding one file to the beginning) - ~3 minutes
Diff to the one file change - ~6 minutes 
Merge the change -  ~9 minutes

ozkatz:
@guy-har does this reproduce after the cache fixes maxe by @arielshaqed?

arielshaqed:
> @guy-har does this reproduce after the cache fixes made by @arielshaqed?

I would *not* expect these fixes to affect performance greatly.  #1244 _maybe_, but probably not by a lot.  #1245 not at all.

However I _would_ want to be sure that we are after #1234.  The numbers are about right for "no caching at all".  If we can still reproduce, can you attach logs at level TRACE?  For multi-minute operations the added slowdown should be minor.

arielshaqed:
Closed by #1267.

PR #1267, opened 2021-01-21
Open range SSTables only when we need to fetch their values

arielshaqed:
No unit test support in this PR: that is tricky.

I want to get it out there ASAP, and will add unit testing in the next PR.


arielshaqed:
Relevant for #1264,  #1240.

ozkatz (approved):
Nice catch!

arielshaqed:
Thanks!

Let's hope this improves matters.

PR #1271, opened 2021-01-21
Feature/import composition

itaiad200:


onboard/import.go
```diff
@@ -58,17 +61,20 @@ func CreateImporter(ctx context.Context, logger logging.Logger, config *Config)
...
+		res.CatalogActions = buildRepoActions(config, logger)
 	}
+
 	previousCommit, err := res.CatalogActions.GetPreviousCommit(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get previous commit: %w", err)
 	}
 	res.previousCommit = previousCommit
-	shouldSort := res.previousCommit != nil
+	shouldSort := res.previousCommit != nil || res.rocks
```
johnnyaug, line 77: why is this removed? don't we need the inventory to return sorted?

cmd/lakefs/cmd/import.go
```diff
@@ -131,6 +133,16 @@ var importCmd = &cobra.Command{
 			Cataloger:          cataloger,
 			KeyPrefixes:        prefixes,
 		}
```
johnnyaug, line 135:
```suggestion
			KeyPrefixes:        prefixes,
			Rocks:              isRocks,
		}
```

johnnyaug: Also, IMO - move the `if` to before the creation of the config, and put the entryCatalog in a varialbe - so that you'll already create the config with all the right values, without editing it later.

onboard/rocks_catalog_actions.go
```diff
@@ -20,43 +20,47 @@ type RocksCatalogRepoActions struct {
...
 	CommitExistingMetaRange(ctx context.Context, repositoryID graveler.RepositoryID, parentCommitID graveler.CommitID, metaRangeID graveler.MetaRangeID, committer string, message string, metadata graveler.Metadata) (graveler.CommitID, error)
 	ListEntries(ctx context.Context, repositoryID graveler.RepositoryID, ref graveler.Ref, prefix, delimiter rocks.Path) (rocks.EntryListingIterator, error)
 	UpdateBranch(ctx context.Context, repositoryID graveler.RepositoryID, branchID graveler.BranchID, ref graveler.Ref) (*graveler.Branch, error)
+	GetBranch(ctx context.Context, repositoryID graveler.RepositoryID, branchID graveler.BranchID) (*graveler.Branch, error)
+	CreateBranch(ctx context.Context, repositoryID graveler.RepositoryID, branchID graveler.BranchID, ref graveler.Ref) (*graveler.Branch, error)
 }
 
-func NewRocksCatalogRepoActions(metaRanger metaRangeManager, repository graveler.RepositoryID, committer string, logger logging.Logger) *RocksCatalogRepoActions {
+func NewRocksCatalogRepoActions(metaRanger entryCataloger, repository graveler.RepositoryID, committer string, logger logging.Logger, prefixes []string) *RocksCatalogRepoActions {
```
johnnyaug, line 47: parameter name >> cataloger (or something)

onboard/rocks_catalog_actions.go
```diff
@@ -95,12 +115,12 @@ func (c *RocksCatalogRepoActions) Commit(ctx context.Context, commitMsg string,
 	if c.createdMetaRangeID == nil {
 		return "", ErrNoMetaRange
 	}
-	commitID, err := c.metaRanger.CommitExistingMetaRange(ctx, c.repoID, c.commitID, *c.createdMetaRangeID, c.committer, commitMsg, graveler.Metadata(metadata))
+	commitID, err := c.entryCataloger.CommitExistingMetaRange(ctx, c.repoID, c.previousCommitID, *c.createdMetaRangeID, c.committer, commitMsg, graveler.Metadata(metadata))
```
johnnyaug, line 118: I think we have a concurrency problem here. It is not created by this PR, but becomes more problematic when using master.

Suppose during the import, someone performs commit A to master.
We started the import when there were no commits to master.
So we now create a commit B which has no parent, and we set master to point to it. We detached commit A from our repository. I don't have a real solution for that. Maybe the relevant branch should be locked throughout the entire process?

johnnyaug (changes requested):

onboard/import.go
itaiad200: What is removed? For rocks you always have to return the inventory in order.

cmd/lakefs/cmd/import.go
itaiad200: Done.

onboard/rocks_catalog_actions.go
itaiad200: Done.

onboard/rocks_catalog_actions.go
itaiad200: I agree, and I think we discussed that during the previous PR of import.
We don't provide any graveler-external locking for now, this PR is not really the place to introduce one. 

I'm opening an issue to track this, I think we probably have a similar issue with `migrate` #1313 .

The real problem is with 2 detached commits. Anything else is still a race, but one you can recover from easily. Up until now, import did not create the first commit to the repo.

johnnyaug (approved):

PR #1272, opened 2021-01-21
Check if tracing logs are emitted before constructing costly loggers

arielshaqed:
Rules for putting something inside `if log.Tracing() ...`:
* Calls to `.Tracef` or `.Trace` with parameters, especially if any parameters require
  computation (but we should not be using `log.*f` at all
* Calls to `.With*()`, they construct a new logger

* Plain calls to `.Trace` need not be wrapped, that's the only thing that they do.


arielshaqed:
Should fix (minor) performance issue in large applies and committed actions.

arielshaqed:
(Every C or C++ logging library _ever_ has used macros that do this automatically :sob: :cry: :crying_cat_face: )

ozkatz (approved):
Thanks!

PR #1274, opened 2021-01-21
Close base iterator after comparison

arielshaqed:


ozkatz (approved):

arielshaqed:
Thanks!  Still too many opens of meta-range, though.

PR #1281, opened 2021-01-21
Migrate iterator query using scan entries for better performance

### LINKED ISSUE #1273: Migrate DB from MVCC to EntryCatalog is slow
Listing entries from MVCC using paging API is very slow and while migrate does it for each commit, migrate can it a very long time.

nopcoder:
Close #1273

### DIFF
--- a/catalog/migrate/iterator.go
+++ b/catalog/migrate/iterator.go
@@ -3,74 +3,69 @@ package migrate
 import (
 	"context"
 	"errors"
+	"time"
 
-	"github.com/treeverse/lakefs/catalog"
+	"github.com/jackc/pgx/v4"
 	"github.com/treeverse/lakefs/catalog/mvcc"
 	"github.com/treeverse/lakefs/catalog/rocks"
 	"github.com/treeverse/lakefs/db"
-)
-
-type iteratorState int
-
-const (
-	iteratorStateInit iteratorState = iota
-	iteratorStateQuerying
-	iteratorStateDone
+	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
 type Iterator struct {
-	state     iteratorState
-	ctx       context.Context
-	db        db.Database
-	branchID  int64
-	commitID  int64
-	buf       []*rocks.EntryRecord
-	err       error
-	offset    string
-	fetchSize int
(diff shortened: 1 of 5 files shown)


catalog/migrate/iterator.go
```diff
@@ -3,74 +3,69 @@ package migrate
...
+	return &Iterator{
+		rows: rows,
+	}, nil
 }
 
 func (it *Iterator) Next() bool {
-	if it.err != nil {
+	if !it.rows.Next() {
+		it.err = it.Err()
+		it.value = nil
```
guy-har, line 43:
```suggestion
	if !it.rows.Next() {
		it.err = it.rows.Err()
		it.value = nil
```

catalog/migrate/iterator.go
```diff
@@ -81,46 +76,16 @@ func (it *Iterator) Value() *rocks.EntryRecord {
 }
 
 func (it *Iterator) Err() error {
-	return it.err
+	return it.rows.Err()
```
guy-har, line 79:
```suggestion
	return it.err
```

catalog/migrate/iterator.go
```diff
@@ -3,74 +3,69 @@ package migrate
...
+	if it.err != nil {
 		return
 	}
-	it.state = iteratorStateInit
-	it.offset = id.String()
-	it.buf = it.buf[:0]
-	it.value = nil
-	it.err = nil
+	it.rows.Close()
+	it.err = ErrNotSeekable
```
guy-har, line 68: You don't have an error check on Next, not sure what happens if you call next after close

guy-har (changes requested):
Looks good,
seems like there is a small problem with the error in the migrate iterator

catalog/migrate/iterator.go
nopcoder: added

catalog/migrate/iterator.go
nopcoder: fixed - part of moving from using iterator error my our own err.

guy-har (approved):
Cool, Thanks

nopcoder:
> Cool, Thanks

Thank you!

PR #1284, opened 2021-01-21
Adaptively increase page size in "lakectl diff"

arielshaqed:
If we're going to grab a diff containing hundreds of millions of files, taking it 100 at a time will be... slow.  Adaptively increase the page size.  With this PR `lakectl` increases to ~700M when grabbing on such a scale; this is probably worth it for working with huge files (but I could be convinced to decrease it 3x).

### DIFF
--- a/cmd/lakectl/cmd/diff.go
+++ b/cmd/lakectl/cmd/diff.go
@@ -14,9 +14,10 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


ozkatz (approved):
:+1:

arielshaqed:
Thanks!

Issue #1263, opened 2021-01-20
rocks: merge operation doesn't apply all changes 


ozkatz:
diffing two branches (say A and B), then merging B into A, results in most of the changes on B being applied to A, except for a single one. Doing a diff does show that there's a single object in B not in A. This is also consistent with listing the objects at both A and B (B showing 1 object that A doesn't have).

I can give access to an environment where this reproduces in DM

PR #1289, opened 2021-01-23
add metarange ID to commit log and show

ozkatz:
Should aid in debugging user issues

### DIFF
--- a/api/api_controller.go
+++ b/api/api_controller.go
@@ -411,6 +411,7 @@ func (c *Controller) GetCommitHandler() commits.GetCommitHandler {
 			Message:      commit.Message,
 			Metadata:     commit.Metadata,
 			Parents:      commit.Parents,
+			MetarangeID:  commit.MetaRangeID,
 		})
 	})
 }
@@ -484,6 +485,7 @@ func (c *Controller) CommitsGetBranchCommitLogHandler() commits.GetBranchCommitL
 				ID:           commit.Reference,
 				Message:      commit.Message,
 				Metadata:     commit.Metadata,
+				MetarangeID:  commit.MetaRangeID,
 				Parents:      commit.Parents,
 			}
 			lastID = commit.Reference
--- a/api/api_controller.go
+++ b/api/api_controller.go
@@ -411,7 +411,7 @@ func (c *Controller) GetCommitHandler() commits.GetCommitHandler {
 			Message:      commit.Message,
 			Metadata:     commit.Metadata,
 			Parents:      commit.Parents,
-			MetarangeID:  commit.MetaRangeID,
+			MetaRangeID:  commit.MetaRangeID,
 		})
 	})
 }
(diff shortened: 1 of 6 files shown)


ozkatz:
This will come in handy with issues like #1263 (or any other hard to explain behavior where we might want metadata from the user in order to reproduce)

catalog/model.go
```diff
@@ -37,6 +37,7 @@ type CommitLog struct {
 	Message      string    `db:"message"`
 	CreationDate time.Time `db:"creation_date"`
 	Metadata     Metadata  `db:"metadata"`
+	MetaRangeID  string    `db:"meta_range_id"`
```
itaiad200, line 40: DDL upgrade/downgrade needed?

api/api_controller.go
```diff
@@ -411,6 +411,7 @@ func (c *Controller) GetCommitHandler() commits.GetCommitHandler {
 			Message:      commit.Message,
 			Metadata:     commit.Metadata,
 			Parents:      commit.Parents,
+			MetarangeID:  commit.MetaRangeID,
```
itaiad200, line 414: nit: let's stick with `Metarange` or `MetaRange`.

itaiad200 (changes requested):
Blocking since I believe DB upgrade is needed. Other than that LGTM.

api/api_controller.go
ozkatz: Fixed

catalog/model.go
ozkatz: No, meta_range_id column introduced in [DDL 20](https://github.com/treeverse/lakeFS/blob/master/ddl/000020_tree_to_range.up.sql#L3).

arielshaqed (approved):
Approving because it helps, but not a fan.

Metarange ID is an implementation detail.  It should be visible to plumbing, but this PR adds it to porcelain.  E.g. in `git` I think you need to take an extra step _on plumbing_ to get tree ID from commit ID, e.g. use the `%T` format (`git show --format='%T' ...`) or [tree object](https://git-scm.com/book/en/v2/Git-Internals-Git-Objects#tree_objects) shows how to `git show` the tree of a commit using the `BRANCH^{tree}` syntax.

So I'd rather have it displayed only if a format option is specified.  We can start with individual format flags for now, and go for full formats... later.  WDYT?

itaiad200 (approved):

ozkatz:
@arielshaqed I've added a `--show-meta-range-id` flag to `lakectl show` and `lakectl log`. By default the meta range ID will be hidden unless this flag is set.

cmd/lakectl/cmd/log.go
```diff
@@ -44,14 +46,17 @@ var logCmd = &cobra.Command{
 		if err != nil {
 			DieErr(err)
 		}
+		showMetaRangeID, _ := cmd.Flags().GetBool("show-meta-range-id")
```
arielshaqed, line 49: Sure is better than `--format "%H"` (as the _only_ way I could find to convince git to show me a tree-ID)!

arielshaqed (approved):
Great, thanks!

PR #1290, opened 2021-01-23
fixed yet another slice pointer bug with iterator keys

### LINKED ISSUE #1263: rocks: merge operation doesn't apply all changes

ozkatz:
closes #1263

### DIFF
--- a/graveler/committed/unmarshal_iterator.go
+++ b/graveler/committed/unmarshal_iterator.go
@@ -34,7 +34,7 @@ func (r *UnmarshalIterator) Next() bool {
 		}
 	}
 	r.value = &graveler.ValueRecord{
-		Key:   graveler.Key(val.Key),
+		Key:   graveler.Key(val.Key.Copy()),
 		Value: v,
 	}
 	return true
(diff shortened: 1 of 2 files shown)


graveler/committed/unmarshal_iterator.go
```diff
@@ -34,7 +34,7 @@ func (r *UnmarshalIterator) Next() bool {
 		}
 	}
 	r.value = &graveler.ValueRecord{
-		Key:   graveler.Key(val.Key),
+		Key:   graveler.Key(val.Key.Copy()),
```
arielshaqed, line 37: This is something of an opportunity to get some benefit from what little type safety Go gives: if we wrote this as a function that copies and converts, it would be easier and lazier to do the right thing.

arielshaqed (approved):
If it works, what's not to like? Pull like the wind! 
That said, I'd like some tools to help human programmers avoid this and similar errors.

PR #1298, opened 2021-01-24
Add tests and fix race in refcounted cache

arielshaqed:
These tests pass, but if we've got them they're nice to have, so I want to keep them!

### DIFF
--- a/cache/only_one_test.go
+++ b/cache/only_one_test.go
@@ -2,6 +2,7 @@ package cache_test
 
 import (
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -85,5 +86,48 @@ func TestOnlyOne_ComputeConcurrentlyOnce(t *testing.T) {
 	}
 	if did5 {
 		t.Error("did not expect to run concurrent compute after 5ms")
+	}
+}
+
+func TestOnlyOne_ComputeConcurrentlyOnceKeyedByStruct(t *testing.T) {
+	c := cache.NewChanOnlyOne()
+
+	type Key struct {
+		a, b int
+	}
+	const (
+		size        = 3
+		repeats     = 50
+		parallelism = 9
+	)
+
+	for i := 0; i < repeats; i++ {
+		var wg sync.WaitGroup
+		wg.Add(size * size)
+		ch := make(chan struct{})
+		numExec := int32(0)
+		for j := 0; j < parallelism; j++ {
+			go func() {
+				<-ch
+				for k := 0; k < size; k++ {
+					for l := 0; l < size; l++ {
+						v, err := c.Compute(Key{k, l}, func() (interface{}, error) {
+							atomic.AddInt32(&numExec, 1)
+							return Key{k + 1, l + 1}, nil
+						})
+						if err != nil {
+							t.Error(err)
+							continue
+						}
+						got := v.(Key)
+						expected := Key{a: k + 1, b: l + 1}
+						if got != expected {
+							t.Errorf("got %v instead of [%d %d]", got, k+1, l+1)
+						}
+					}
+				}
+			}()
+		}
+		close(ch)
 	}
 }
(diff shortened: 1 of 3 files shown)


arielshaqed:
*Reviewer:* Easier to review commit-by-commit.

cache/only_one_test.go
```diff
@@ -87,3 +88,46 @@ func TestOnlyOne_ComputeConcurrentlyOnce(t *testing.T) {
...
+		a, b int
+	}
+	const (
+		size        = 3
+		repeats     = 50
+		parallelism = 9
+	)
+
+	for i := 0; i < repeats; i++ {
+		var wg sync.WaitGroup
```
itaiad200, line 105: What do you do with the `wg`?

cache/only_one_test.go
```diff
@@ -87,3 +88,46 @@ func TestOnlyOne_ComputeConcurrentlyOnce(t *testing.T) {
...
+		wg.Add(size * size)
+		ch := make(chan struct{})
+		numExec := int32(0)
+		for j := 0; j < parallelism; j++ {
+			go func() {
+				<-ch
```
itaiad200, line 111: Clever.

cache/only_one_test.go
```diff
@@ -87,3 +88,46 @@ func TestOnlyOne_ComputeConcurrentlyOnce(t *testing.T) {
...
+						expected := Key{a: k + 1, b: l + 1}
+						if got != expected {
+							t.Errorf("got %v instead of [%d %d]", got, k+1, l+1)
+						}
+					}
+				}
+			}()
+		}
+		close(ch)
+	}
```
itaiad200, line 132: You're not waiting on anything here..

itaiad200: Did you mean to use the `wg` for that?

cache/only_one_test.go
```diff
@@ -87,3 +88,46 @@ func TestOnlyOne_ComputeConcurrentlyOnce(t *testing.T) {
...
+		size        = 3
+		repeats     = 50
+		parallelism = 9
+	)
+
+	for i := 0; i < repeats; i++ {
+		var wg sync.WaitGroup
+		wg.Add(size * size)
+		ch := make(chan struct{})
+		numExec := int32(0)
```
itaiad200, line 108: You never check it, so what is it for?

cache/only_one_test.go
```diff
@@ -87,3 +88,46 @@ func TestOnlyOne_ComputeConcurrentlyOnce(t *testing.T) {
...
+						v, err := c.Compute(Key{k, l}, func() (interface{}, error) {
+							atomic.AddInt32(&numExec, 1)
+							return Key{k + 1, l + 1}, nil
+						})
+						if err != nil {
+							t.Error(err)
+							continue
+						}
+						got := v.(Key)
+						expected := Key{a: k + 1, b: l + 1}
```
itaiad200, line 123: Even if `Compute` is completely broken, that will always succeed, no?

You increase the incoming values by 1, even if they run in parallel for the same key, than you will still get the same output.

cache/reference_counted_test.go
```diff
@@ -13,88 +13,100 @@ import (
...
+		t.Run(tc.name, func(t *testing.T) {
+			var elements = make([]*record, tc.size)
+			ch := make(chan error, tc.parallelism*2)
+			allErrs := make(chan []error, tc.parallelism*2)
+			go func() {
+				errs := make([]error, 0)
+				for err := range ch {
+					errs = append(errs, err)
+				}
+				allErrs <- errs
```
itaiad200, line 42: Why do you need to error channels?

itaiad200 (changes requested):
I think the first test doesn't work as expected.
Seconds one looks good!

cache/reference_counted_test.go
arielshaqed: Leftover from when we thought goroutines were not allowed to issue errors.  SInce they're only prohibited from failing, removed!  Thanks!

arielshaqed (commented):
Thanks!  With the requested change to `TestOnlyOne_ComputeConcurrentlyOnceKeyedByStruct` the test actually tests something *and fails*.  So I fixed that.

PTAL...

arielshaqed:
No current need for review: added tests revealed that `OnlyOne` is currently used in 2 incompatible ways.

arielshaqed:
Closing -- this has been superseded by other work, notable #1332 .

Issue #1172, opened 2021-01-11
Empty commits work


arielshaqed:
```sh
ariels@redqueen:~/dev/lakeFS$ ./lakectl commit lakefs://foo@trunk --message "j/t"
Commit for branch "trunk" completed.

ID: 0fd798bfa5ea98151e0864a9382dbce96be604379312b7e327776800649c7158
Message: j/t
Timestamp: 2021-01-11 09:45:09 +0200 IST
Parents: 9a25c115fc9819322807cfbe247bb7292724a71350c972f6f6e18e01b257589d

ariels@redqueen:~/dev/lakeFS$ AWS_PROFILE=lake aws --endpoint-url=http://s3.local.lakefs.io:8000  s3 ls s3://foo/trunk/
ariels@redqueen:~/dev/lakeFS$ ./lakectl commit lakefs://foo@trunk --message "j/t"Commit for branch "trunk" completed.

ID: b79d628b7af0c79c7bc2bff985d5caf9b5a1028ce1914723e44bcf16c9104443
Message: j/t
Timestamp: 2021-01-11 09:45:39 +0200 IST
Parents: 0fd798bfa5ea98151e0864a9382dbce96be604379312b7e327776800649c7158

```

In `git` an empty commit fails.  Consider doing the same.

arielshaqed:
This is on my current (unpushed) work.

arielshaqed:
Also be even more of a `git` and fail on empty commit message.

itaiad200:
Dup of #1251 ?

arielshaqed:
Still happens, and like #1251 blocks.

PR #1310, opened 2021-01-25
Fail commits that change nothing

arielshaqed:


arielshaqed:
Fixes #1172.

nopcoder (changes requested):
Make it optional as there are still cases we would like to enable empty commits.

nopcoder (approved):

arielshaqed:
Thanks!

PR #1324, opened 2021-01-26
Welcome print out version and cataloger type if set

nopcoder:


cmd/lakefs/cmd/run.go
```diff
@@ -203,6 +203,11 @@ const runBanner = `
 
 func printWelcome(w io.Writer) {
 	_, _ = fmt.Fprint(w, runBanner)
+	_, _ = fmt.Fprintf(w, "Version %s\n", config.Version)
+	catalogerType := cfg.GetCatalogerType()
+	if len(catalogerType) > 0 {
```
arielshaqed, line 208: cataloger type is implemented to have a default, but that default is *not* returned by `GetCatalogerType` (I assume we didn't do the whole config dance because a temporary config flag).  I think we should print the default if there is no cataloger type -- otherwise it will be hard to tell what's *really* running, just when it matters most!

cmd/lakefs/cmd/run.go
```diff
@@ -203,6 +203,7 @@ const runBanner = `
 
 func printWelcome(w io.Writer) {
 	_, _ = fmt.Fprint(w, runBanner)
+	_, _ = fmt.Fprintf(w, "Version %s\nCataloger %s\n\n", config.Version, cfg.GetCatalogerType())
```
arielshaqed, line 206: Thanks!  It's still a bit confusing _for the user_, because the default will say `Cataloger ` (default is implemented at [point of use in the catalog builder](https://github.com/treeverse/lakeFS/blob/e2d239f461645758322c1de624a22ee1aef6992b/catalog/factory/build.go#L19) rather than in config.go), but we'll have enough information to debug.
Good enough given that The Other Cataloger Type is going away.

arielshaqed (approved):
Thanks!

cmd/lakefs/cmd/run.go
nopcoder: The idea was to have a peek at the user configuration while we start lakeFS - help understand what I just run. At least until the point where we drop the MVCC code.

PR #1330, opened 2021-01-26
Fix/commit compare error #1286

### LINKED ISSUE #1286: UI: comparing commit with itself returns  error Branch not found
Screen Shot 2021-01-21 at 21 17 24

shamikakumar:
Fixes #1286. Changed the api call to refs which handles both branches&commits. Also removed the if-else statement because the api handles the condition for comparing a branch/commit to itself.


johnnyaug (changes requested):
Hey @shamikakumar, thank you for your effort!

The `branches/diff` and `refs/diff` APIs are entirely different things: the first is used to show all _uncommitted_ changes (mainly for the Changes tab), and the second one compares two _commits_.
Your change removes the functionality of viewing uncommitted changes. You will need to make a less intrusive change.

Thanks again, and feel free to consult with me on Slack

shamikakumar:
Hi @johnnyaug thanks for clarifying! I misunderstood what the APIs were doing. I see now what consequences that had. I've modified my PR so as to not call the diff api when the sources are the same. Do review and let me know.

johnnyaug (approved):

johnnyaug:
Thanks @shamikakumar for your contribution!

PR #1332, opened 2021-01-26
Remove SSTable reader cache, use shared pebble in-memory cache

### LINKED ISSUE #1295: sstable reader cache appears to be a no-op
We still see the same metarange file being opened repeatedly.  This may be a cause of slowness.

E.g. on a test system we still have e3b0 opened 600+ times at times (and <100 at others):

```
[ariels@ip-10-20-1-237 ~]$ ls -l /proc/3406/fd 2>&1  | grep -c e3b0c4...digest...
602
[ariels@ip-10-20-1-237 ~]$ ls -l /proc/3406/fd 2>&1  | grep -c e3b0c4...digest...
194
[ariels@ip-10-20-1-237 ~]$ ls -l /proc/3406/fd 2>&1  | grep -c e3b0c4...digest...
181
[ariels@ip-10-20-1-237 ~]$ ls -l /proc/3406/fd 2>&1  | grep -c e3b0c4...digest...
724
```

arielshaqed:
Experimentation (see #1295 for some details) shows this cache _decreases_ throughput on diffs
by 10-20%.  It is not clear that it decreases latency of individual GetEntry calls by enough.
Given the complexity of both implementation and wiring, getting rid of it seems best.

Fixes #1295.


arielshaqed:
[Six hundred threescore and six](https://www.kingjamesbibleonline.org/Revelation-13-18/) deleted lines! :man: :japanese_ogre:

catalog/rocks/entry_catalog.go
```diff
@@ -111,15 +110,9 @@ func NewEntryCatalog(cfg *config.Config, db db.Database) (*EntryCatalog, error)
 
 	pebbleSSTableCache := pebble.NewCache(tierFSParams.PebbleSSTableCacheSizeBytes)
```
itaiad200, line 111: Why share the same cache instance between the 2?
I would the same proportions (not percents ;) ) of the disk space and create 2 matching pebble caches.
Like we don't want the 2 competing for disk space, we should also avoid competition for memory.

catalog/rocks/entry_catalog.go
```diff
@@ -111,15 +110,9 @@ func NewEntryCatalog(cfg *config.Config, db db.Database) (*EntryCatalog, error)
...
 	defer pebbleSSTableCache.Unref()
```
itaiad200, line 112: I'm confused, it looks like the `Unref` happens before the cache is actually being closed?

itaiad200 (changes requested):
Looks good! Requesting changes because I don't understand the Ref/Unref part.

catalog/rocks/entry_catalog.go
arielshaqed: I'm not sure it makes sense even for the local tier.  But separate caching _always_ gives *worse* throughput.  Unlike the disk local tier, every block in the pebble memory cache should cost exactly the same to fetch: it's just a read from a disk.  (Well, if you're using disks made of spinning rust you _might_ see a difference, but who does that?)

Given that this PR is mostly a rollback of an expensive premature optimization, I am hesitant to add more options to this.

*NOTE TO SELF:* Changes (removes) config parameters, so requires a doc change and _should not be pushed today_, let rocks go... :-)

catalog/rocks/entry_catalog.go
arielshaqed: Yes.
Cache docs say to do:
```go
//   c := pebble.NewCache(...)
//   defer c.Unref()
//   d, err := pebble.Open(pebble.Options{Cache: c})
```
and here `NewPebbleSSTableRangeManager` will Ref the cache.

arielshaqed (commented):
Thanks!

catalog/rocks/entry_catalog.go
itaiad200: And where does PebbleSSTableRangeManager Unref it?

catalog/rocks/entry_catalog.go
itaiad200: Not sure that's true in this case since you will always read the the metarange entry before reading the range entry, so cleaning the metarange entry has potentially a higher cost.

However I'm far from having numbers that support that and indeed it's more of an optimization -> resolving.

catalog/rocks/entry_catalog.go
arielshaqed: In the destructor?  Since we don't have  `PebbleSSTableRangeManager.Close`, we don't unref it.  I do not want to add code that is never needed to run - it will never really be tested.
I *do* need to ref the cache, though: if the same cache is passed to any other Pebble object, *that* might ref/unref it.

arielshaqed (commented):
Thanks!

catalog/rocks/entry_catalog.go
itaiad200: I was pretty sure that `pebble.Cache` panics when you finish using it and `cache.Ref >= 1` and we use to get it in some unit test.
But I don't find evidence to it in the version that we're using - so we're good to go.

itaiad200 (approved):

arielshaqed:
Thanks!
Pulling...

PR #1333, opened 2021-01-26
Default to use rocks

nopcoder:
verify on lakefs run that migration is required. cataloger type is not
set, there are repositories on mvcc tables and migration didn't run

### DIFF
--- a/cmd/lakefs/cmd/run.go
+++ b/cmd/lakefs/cmd/run.go
@@ -192,7 +192,7 @@ var runCmd = &cobra.Command{
 	},
 }
 
-const migrateRequiredMsg = `Data migration is required.
+const migrateRequiredMsg = `Data migration is required - https://docs.lakefs.io/deploying/upgrade.html.
 Starting version 0.30.0, lakeFS handles your committed metadata in a new way,
 which is more robust and has better performance.
 To move your existing data, you will need to run the following upgrade command:
@@ -204,6 +204,7 @@ To do so, add the following to your configuration:
 
 cataloger:
   type: rocks
+
 `
 
 const runBanner = `
--- a/cmd/lakefs/cmd/run.go
+++ b/cmd/lakefs/cmd/run.go
@@ -89,7 +89,8 @@ var runCmd = &cobra.Command{
 		if cfg.GetCatalogerType() == "" {
 			migrationRequired := catalogmigrate.CheckMigrationRequired(dbPool)
 			if migrationRequired {
-				logger.Fatal(migrateRequiredMsg)
+				logger.Fatal("Data migration is required")
+				fmt.Println(migrateRequiredMsg)
 			}
 		}
 
--- a/cmd/lakefs/cmd/run.go
+++ b/cmd/lakefs/cmd/run.go
@@ -205,6 +205,10 @@ To do so, add the following to your configuration:
 cataloger:
   type: rocks
 
+And run:
+
+  $ lakefs migrate up
+
 `
 
 const runBanner = `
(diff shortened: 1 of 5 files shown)


cmd/lakefs/cmd/run.go
```diff
@@ -83,6 +84,15 @@ var runCmd = &cobra.Command{
 			logger.WithError(err).Fatal("Failed to create block adapter")
 		}
 
+		//  Migrate old MVCC - if cataloger type is not set,
+		//  warn the user in case migrate didn't run and there are MVCC repositories
+		if cfg.GetCatalogerType() == "" {
+			migrationRequired := catalogmigrate.CheckMigrationRequired(dbPool)
+			if migrationRequired {
+				logger.Fatal("Migration of old data is required! (lakefs migrate db)")
```
ozkatz, line 92: please link to the the future release notes that explain the migration steps

ozkatz (changes requested):

cmd/lakefs/cmd/run.go
johnnyaug, line 92: We need to print out a very nice message here:

_Starting version 0.30.0, lakeFS handles your committed metadata in a new way, which is more robust and has better performance. To move your existing data, you will need to run the following upgrade command:
...
If you want to start over, ignoring your existing data, yoou need to explicitly state this in your lakeFS configuration file. To do so, add the following to your configuration:
```
cataloger:
  type: rocks
```_

I think we also need to add this to the documentation

**Review by johnnyaug — CHANGES_REQUESTED** [review]

File: cmd/lakefs/cmd/run.go (hunk 1 above)
Comment (nopcoder) [reply cmd/lakefs/cmd/run.go:92]: Update https://docs.lakefs.io/deploying/upgrade.html?
or
Link https://github.com/treeverse/lakeFS/releases/tag/v0.30.0

Comment (nopcoder) [reply cmd/lakefs/cmd/run.go]: Added link to the docs in the code and detailed explanation by Yoni.
The release notes will include the same information.

File: docs/deploying/upgrade.md [hunk 2]
```diff
@@ -24,3 +25,22 @@ lakefs migrate up
...
+lakefs migrate db
+```
+
+If you want to start over, discarding your existing data, you need to explicitly state this in your lakeFS configuration file.
+To do so, add the following to your configuration:
+
+```yaml
+cataloger:
+  type: rocks
+```
```
guy-har, line 45: In this case they should still run 
```bash
lakefs migrate up
```

docs/deploying/upgrade.md
```diff
@@ -24,3 +25,22 @@ lakefs migrate up
 Deploy (or run) the new version of lakeFS.
 
 Note that an older version of lakeFS cannot run on a migrated database.
+
+
+# Migrating from lakeFS < 0.30.0
+
+Starting version 0.30.0, lakeFS handles your committed metadata in a new way, which is more robust and has better performance.
```
ozkatz, line 32:
```suggestion
Starting version 0.30.0, lakeFS handles your committed metadata in a [new way](https://docs.google.com/document/d/1jzD7-jun-tdU5BGapmnMBe9ovSzBvTNjXCcVztV07A4/edit?usp=sharing){: target="_blank" }, which is more robust and has better performance.
```

docs/deploying/upgrade.md
```diff
@@ -24,3 +25,22 @@ lakefs migrate up
...
+To move your existing data, you will need to run the following upgrade command with the updated lakefs binary:
```
ozkatz, line 33: perhaps verify first that you're running a binary that is == 0.30.x with `lakefs --version`

ozkatz (changes requested):
link to the new architecture

johnnyaug (approved):

ozkatz (approved):

guy-har (approved):

PR #1334, opened 2021-01-26
delete from stage on repository without commit

nopcoder:

PR #1337, opened 2021-01-26
Pass cataloger type env var

itaiad200:


nopcoder (approved):

nopcoder:
although it will be the default soon...

itaiad200:
Validated - committed data is stored in s3 for rocks cataloger and not for mvcc.

This cannot pass until #1334  is merged.

itaiad200:
> although it will be the default soon...

Yap but then mvcc will run as rocks.

nopcoder:
Applied part of PR #1338

PR #1342, opened 2021-01-27
Cancel file Store during Abort

### LINKED ISSUE #1320: Terminate context when writer is aborted
When sstable writer is being aborted, it closes the local file but does nothing with the ongoing upload to s3.
This keeps the failing flow hanging until the request completes, but the result is meaningless.

itaiad200:
closes #1320

### DIFF
--- a/pyramid/file_test.go
+++ b/pyramid/file_test.go
@@ -28,10 +28,12 @@ func TestPyramidWriteFile(t *testing.T) {
 
 	storeCalled := false
 	abortCalled := false
+	var storeCtx context.Context
 	sut := WRFile{
 		File: fh,
-		store: func(context.Context, string) error {
+		store: func(innerCtx context.Context, _ string) error {
 			storeCalled = true
+			storeCtx = innerCtx
 			return nil
 		},
 		abort: func(context.Context) error {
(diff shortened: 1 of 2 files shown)


pyramid/file_test.go
```diff
@@ -51,11 +51,13 @@ func TestPyramidWriteFile(t *testing.T) {
 
 	require.NoError(t, sut.Close())
 	require.NoError(t, sut.Store(ctx, filename))
+	require.NotNil(t, sut.cancelStore)
```
arielshaqed, line 54: I don't understand what this tests.  AFAICS it only tests _that_ a cancel function was returned, not that it has anything to do with actually cancelling the context (e.g. correctly wired up).

pyramid/wr_file.go
```diff
@@ -35,12 +36,21 @@ func (f *WRFile) Store(ctx context.Context, filename string) error {
 		return err
 	}
 
+	// keep the cancel function for the Store's context,
+	// so that the long operation is cancellable during a call to Abort.
+	ctx, f.cancelStore = context.WithCancel(ctx)
 	return f.store(ctx, filename)
 }
 
 // Abort delete the file and cleans all traces of it.
```
arielshaqed, line 45:
```suggestion
// Abort deletes the file and cleans all traces of it.
```

arielshaqed (approved):
Cool!
How can we test this (i.e., that cancellation is actually wired through & "works")?  It's OK to do it manually, even once, or in a unit test with a fake slow store, or even to open an issue.  Just let's make sure...

pyramid/file_test.go
itaiad200: Right, fixed.

PR #1343, opened 2021-01-27
auto confirm commands with -y instead of -f, and only for commands that require it

ozkatz:


cmd/lakectl/cmd/input.go
```diff
@@ -5,16 +5,30 @@ import (
...
+
+func Confirm(flags *pflag.FlagSet, question string) (bool, error) {
+	yes, err := flags.GetBool(AutoConfirmFlagName)
+	if err != nil {
+		// no auto confirm flag
+		yes = false
+	}
+	if yes {
 		return true, nil
 	}
```
johnnyaug, line 26:
```suggestion
	if err == nil {
		// got auto confirm flag
		return true, nil
	}
```

johnnyaug (approved):
left small suggestion, not critical

cmd/lakectl/cmd/input.go
```diff
@@ -5,16 +5,30 @@ import (
...
+func AssignAutoConfirmFlag(flags *pflag.FlagSet) {
+	flags.BoolP(AutoConfirmFlagName, AutoConfigFlagShortName, false, AutoConfirmFlagHelp)
+}
+
+func Confirm(flags *pflag.FlagSet, question string) (bool, error) {
+	yes, err := flags.GetBool(AutoConfirmFlagName)
+	if err != nil {
+		// no auto confirm flag
+		yes = false
+	}
```
nopcoder, line 23: this is not the default?

cmd/lakectl/cmd/input.go
```diff
@@ -5,16 +5,30 @@ import (
...
-	force, _ := flags.GetBool("force")
-	if force {
+const (
+	AutoConfirmFlagName     = "yes"
+	AutoConfigFlagShortName = "y"
+	AutoConfirmFlagHelp     = "Automatically say yes to all confirmations"
+)
+
+func AssignAutoConfirmFlag(flags *pflag.FlagSet) {
+	flags.BoolP(AutoConfirmFlagName, AutoConfigFlagShortName, false, AutoConfirmFlagHelp)
```
nopcoder, line 15: why not use global persist flag for "yes" it is very common to have it available for all commands

nopcoder (changes requested):

cmd/lakectl/cmd/input.go
ozkatz: That's what we had before. I don't want to do that because:
1. it will show up in the help text for all commands eventhough it's a no-op for most, so confusing
2. it's less flexible since `-y` might take a different meaning in another command but can't be used anymore

nopcoder: 1. The fact that the global flag always appears the chance that somebody will run it over is lower.
2. The user get used to the consistent interface
3. Any new command that we will add or change the current flow of existing comment doesn't need to worry about how auto-confirmation is done (fewer bugs)

cmd/lakectl/cmd/input.go
ozkatz: resolved, plz review

nopcoder (approved):
Approved - don't forget to update the docs https://docs.lakefs.io/reference/commands.html

ozkatz:
> Approved - don't forget to update the docs https://docs.lakefs.io/reference/commands.html

yea this document is outdated anyway. I'll open a seperate issue to update it.

PR #1346, opened 2021-01-27
Remove MVCC import

itaiad200:


onboard/import.go
```diff
@@ -37,45 +36,37 @@ type Config struct {
...
+	if err := res.CatalogActions.InitBranch(ctx); err != nil {
 		return nil, fmt.Errorf("failed to get previous commit: %w", err)
 	}
-	res.previousCommit = previousCommit
-	shouldSort := res.previousCommit != nil || res.rocks
-	res.inventory, err = config.InventoryGenerator.GenerateInventory(ctx, logger, config.InventoryURL, shouldSort, config.KeyPrefixes)
+
+	res.inventory, err = config.InventoryGenerator.GenerateInventory(ctx, logger, config.InventoryURL, true, config.KeyPrefixes)
 	res.prefixes = config.KeyPrefixes
 	if !res.validatePrefixes() {
```
johnnyaug, line 71: Looks like this wouldn't work anymore (not sure if due to this PR or a previous one). Under the Importer struct, we still have a previousCommit member which is used in this method, but it is never assigned.

johnnyaug (changes requested):

onboard/import.go
itaiad200: This one. MVCC used that feature.

johnnyaug: But we still want to validate the prefixes, no?

itaiad200: function name was misleading, it validated that the old and new prefixes match.
That doesn't matter anymore.

johnnyaug (approved):

PR #1347, opened 2021-01-27
remove MVCC from code base, along with retention that isnt supported yet

ozkatz:
This PR also fixes API integration tests that were broken when running with Rocks


ozkatz:
Still TODO:
1. remove retention from UI
2. decide what to do with continuous export which *could* work but currently doesn't
3. remove cataloger factory, expose a single EntryCatalog from the catalog package

cmd/lakefs/cmd/run.go
```diff
@@ -21,7 +21,6 @@ import (
 	"github.com/treeverse/lakefs/auth/crypt"
```
nopcoder, line 21: can you remove the Catalog print in this file - had no line to report this one.

nopcoder (approved):

nopcoder:
image

catalog/factory/build.go
```diff
@@ -2,19 +2,11 @@ package factory
 
 import (
 	"github.com/treeverse/lakefs/catalog"
-	"github.com/treeverse/lakefs/catalog/mvcc"
 	"github.com/treeverse/lakefs/catalog/rocks"
 	"github.com/treeverse/lakefs/config"
 	"github.com/treeverse/lakefs/db"
 )
 
 func BuildCataloger(db db.Database, c *config.Config) (catalog.Cataloger, error) {
```
itaiad200, line 10: You can remove this file entirely.

ozkatz: Will do :) see TODO above

itaiad200 (approved):
Bye Bye.

cmd/lakefs/cmd/import.go
```diff
@@ -91,7 +90,7 @@ var importCmd = &cobra.Command{
 			os.Exit(1)
 		}
 
-		repo, err := getRepository(ctx, cataloger, repoName, dryRun, isRocks)
```
ozkatz, line 94: @itaiad200 would be happy if you give these changes a close look :)

ozkatz:
@itaiad200 import tests are failing for me. Probably has to do with the changes I made in e32bae31f - but from double checking, it should be doing all the same stuff it done before when there's a rocks catalog. Can you help?

api/api_controller_test.go
```diff
@@ -486,7 +485,7 @@ func TestHandler_ListBranchesHandler(t *testing.T) {
 		if err != nil {
 			t.Fatalf("unexpected error listing branches: %s", err)
 		}
-		const expectedBranchesLen = 2 // branch creation and import branch
+		const expectedBranchesLen = 1
```
arielshaqed, line 488: This difference from the change at l.235 (-1 here, -2 there) makes me wonder: either the old way was strange or the new way is strange.  I hope we're sure that the old way is strange :-)

api/handler_test.go
```diff
@@ -88,15 +89,21 @@ func getHandler(t *testing.T, blockstoreType string, opts ...testutil.GetDBOptio
...
 		blockstoreType, _ = os.LookupEnv(testutil.EnvKeyUseBlockAdapter)
 	}
+	if blockstoreType == "" {
+		blockstoreType = "mem"
+	}
 	blockAdapter = testutil.NewBlockAdapterByType(t, &block.NoOpTranslator{}, blockstoreType)
-	cataloger, err := catalogfactory.BuildCataloger(conn, nil)
+	cfg := config.NewConfig()
+	cfg.Override(func(configurator config.Configurator) {
+		configurator.SetDefault(config.BlockstoreTypeKey, "mem")
```
arielshaqed, line 98:
```suggestion
 		configurator.SetDefault(config.BlockstoreTypeKey, blockstoreType)
```
?

But if so, why does `blockstoreType` come from the environment?  Anyway current code seems inconsistent.

cmd/lakectl/cmd/repo.go
```diff
@@ -129,66 +126,12 @@ var repoDeleteCmd = &cobra.Command{
 	},
 }
 
-var retentionCmd = &cobra.Command{
-	Use:    "retention [sub-command]",
-	Short:  "manage repository retention policies",
-	Hidden: true,
-}
-
-var getPolicyCmd = &cobra.Command{
```
arielshaqed, line 138: Do we want a short note somewhere, possibly on Slack so we don't keep a scary notice that only interests 1 or 0 people on documentation forever?

cmd/lakefs/cmd/run.go
arielshaqed: Please keep it for a while -- so users (and us) know what's running.

arielshaqed (commented):
I missed yesterday, so I'm probably missing the plan here.  How does an existing user migrate up from mvcc, if we delete migration?  If it's a 2-step plan then we may need to document it somewhere.

config/config.go
```diff
@@ -167,6 +166,21 @@ func setDefaults() {
...
+}
+
+type viperConfigurator struct {
+}
+
+func (v viperConfigurator) SetDefault(key string, value interface{}) {
+	viper.SetDefault(key, value)
+}
+
+func (c *Config) Override(fn func(Configurator)) {
```
arielshaqed, line 180: Not a fan.  The viper configuration object seems global.  Certainly the call inside `viperConfigurator.SetDefault` is a global mutator.  This facility lets us hide a global mutation behind 3 layers of abstraction; I think it makes code that uses viper less safe.
I would rather just give a global `config.SetDefault` call by saying e.g.
```go
var SetDefault = viper.SetDefault
```

graveler/ref/manager_test.go
```diff
@@ -124,9 +124,8 @@ func TestManager_DeleteRepository(t *testing.T) {
 	})
 
 	t.Run("repo_does_not_exists", func(t *testing.T) {
```
arielshaqed, line 126:
```suggestion
	t.Run("repo_does_not_exist", func(t *testing.T) {
```

arielshaqed (commented):
Cool!

api/api_controller_test.go
ozkatz: The old way is definitely strange.

ozkatz:
> I missed yesterday, so I'm probably missing the plan here. How does an existing user migrate up from mvcc, if we delete migration? If it's a 2-step plan then we may need to document it somewhere.

@arielshaqed Going to add a note in the upgrade guide: users with version < 0.30.0 must *first* upgrade to 0.30.0, and run the migration to rocks before upgrading to any higher version

config/config.go
ozkatz: it was a step towards instantiating a viper instance instead of using a global one. in general, current `config.NewConfig()` isn't really creating anything new, as config a wrapper around a global. I don't mind going with your direction and basically removing the config struct. All its methods can be functions anyway (but I don't want to tack that onto this PR).

cmd/lakefs/cmd/run.go
nopcoder: But for this version, there is no point in printing it, right?

nopcoder:
> > I missed yesterday, so I'm probably missing the plan here. How does an existing user migrate up from mvcc, if we delete migration? If it's a 2-step plan then we may need to document it somewhere.
> 
> @arielshaqed Going to add a note in the upgrade guide: users with version < 0.30.0 must _first_ upgrade to 0.30.0, and run the migration to rocks before upgrading to any higher version

We can extract just the listing query and not remove the migrate - it will be easier to keep the upgrade until the next milestone.

nopcoder (approved):

PR #1348, opened 2021-01-27
Local block adapter remove empty folders when possible

### LINKED ISSUE #998: export doesn't remove empty directories on POSIXy block backends
[Meaningless for production when exporting to S3 adapters.  *ONLY RELEVANT* for the `local` block adapter type, so important for kicking the tyres]

This breaks export in scenarios like  the following:  With continuous export enabled for a branch (here `master`) to the _local_ block adapter:
1. Create `foo@master/a/b/c`.
2. Commit.  `a/b/c` is exported.
3. Remove `foo@master/a/b/c`.
4. Commit.  `a/b/c` is removed; directories `a/b/` and `a/` are _not_ removed, because POSIX filesystems are not like S3.
5. Create `foo@master/a/b`.
6. Export fails: POSIX still has that directory.

nopcoder:
Closes #998

### DIFF
--- a/block/local/adapter.go
+++ b/block/local/adapter.go
@@ -152,7 +152,7 @@ func (l *Adapter) Remove(obj block.ObjectPointer) error {
 }
 
 func removeEmptyDirUntil(dir string, stopAt string) {
-	for stopAt != "" && !strings.HasPrefix(stopAt, dir) {
+	for stopAt != "" && strings.HasPrefix(dir, stopAt) && dir != stopAt {
 		err := os.Remove(dir)
 		if err != nil {
 			break
--- a/block/local/adapter.go
+++ b/block/local/adapter.go
@@ -152,13 +152,13 @@ func (l *Adapter) Remove(obj block.ObjectPointer) error {
 }
 
 func removeEmptyDirUntil(dir string, stopAt string) {
-	for {
+	for stopAt == "" || !strings.HasPrefix(stopAt, dir) {
 		err := os.Remove(dir)
 		if err != nil {
 			break
 		}
 		dir = filepath.Dir(dir)
-		if dir == "/" || strings.HasPrefix(stopAt, dir) {
+		if dir == "/" {
 			break
 		}
 	}
--- a/block/local/adapter.go
+++ b/block/local/adapter.go
@@ -71,6 +71,7 @@ func NewAdapter(path string, opts ...func(a *Adapter)) (*Adapter, error) {
 		path:               path,
 		ctx:                context.Background(),
 		uploadIDTranslator: &block.NoOpTranslator{},
+		removeEmptyDir:     true,
 	}
 	for _, opt := range opts {
 		opt(adapter)
@@ -141,21 +142,9 @@ func (l *Adapter) Remove(obj block.ObjectPointer) error {
 	}
 	if l.removeEmptyDir {
 		dir := filepath.Dir(p)
-		removeEmptyDir(dir)
-	}
-	return nil
-}
-
-func removeEmptyDir(dir string) {
-	d, err := os.Open(dir)
-	if err != nil {
-		return
-	}
-	_, err = d.Readdir(1)
-	_ = d.Close()
-	if err == io.EOF {
 		_ = os.Remove(dir)
 	}
+	return nil
 }
 
 func (l *Adapter) Copy(sourceObj, destinationObj block.ObjectPointer) error {
--- a/block/local/adapter.go
+++ b/block/local/adapter.go
@@ -28,6 +28,7 @@ type Adapter struct {
 	path               string
 	ctx                context.Context
 	uploadIDTranslator block.UploadIDTranslator
+	removeEmptyDir     bool
 }
 
 var (
@@ -41,12 +42,19 @@ func (l *Adapter) WithContext(ctx context.Context) block.Adapter {
 		path:               l.path,
 		ctx:                ctx,
 		uploadIDTranslator: l.uploadIDTranslator,
+		removeEmptyDir:     l.removeEmptyDir,
 	}
 }
 
 func WithTranslator(t block.UploadIDTranslator) func(a *Adapter) {
 	return func(a *Adapter) {
 		a.uploadIDTranslator = t
+	}
+}
+
+func WithRemoveEmptyDir(b bool) func(a *Adapter) {
+	return func(a *Adapter) {
+		a.removeEmptyDir = b
 	}
 }
 
@@ -127,7 +135,27 @@ func (l *Adapter) Remove(obj block.ObjectPointer) error {
 		return err
 	}
 	p = filepath.Clean(p)
-	return os.Remove(p)
+	err = os.Remove(p)
+	if err != nil {
+		return err
+	}
+	if l.removeEmptyDir {
+		dir := filepath.Dir(p)
+		removeEmptyDir(dir)
+	}
+	return nil
+}
+
+func removeEmptyDir(dir string) {
+	d, err := os.Open(dir)
+	if err != nil {
+		return
+	}
+	_, err = d.Readdir(1)
+	_ = d.Close()
+	if err == io.EOF {
+		_ = os.Remove(dir)
+	}
 }
 
 func (l *Adapter) Copy(sourceObj, destinationObj block.ObjectPointer) error {
--- a/block/local/adapter.go
+++ b/block/local/adapter.go
@@ -113,6 +113,10 @@ func maybeMkdir(path string, f func(p string) (*os.File, error)) (*os.File, erro
 	return f(path)
 }
 
+func (l *Adapter) Path() string {
+	return l.path
+}
+
 func (l *Adapter) Put(obj block.ObjectPointer, _ int64, reader io.Reader, _ block.PutOpts) error {
 	p, err := l.getPath(obj)
 	if err != nil {
@@ -142,9 +146,22 @@ func (l *Adapter) Remove(obj block.ObjectPointer) error {
 	}
 	if l.removeEmptyDir {
 		dir := filepath.Dir(p)
-		_ = os.Remove(dir)
+		removeEmptyDirUntil(dir, l.path)
 	}
 	return nil
+}
+
+func removeEmptyDirUntil(dir string, stopAt string) {
+	for {
+		err := os.Remove(dir)
+		if err != nil {
+			break
+		}
+		dir = filepath.Dir(dir)
+		if dir == "/" || strings.HasPrefix(stopAt, dir) {
+			break
+		}
+	}
 }
 
 func (l *Adapter) Copy(sourceObj, destinationObj block.ObjectPointer) error {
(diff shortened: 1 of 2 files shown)


nopcoder:
@arielshaqed, I assume this is part of the solution to the problem we are keeping empty folders.
do we like to enable it just for export, or the issue is the fact that there is no "Delete" when we continuous export?

block/local/adapter.go
```diff
@@ -127,7 +135,27 @@ func (l *Adapter) Remove(obj block.ObjectPointer) error {
...
+func removeEmptyDir(dir string) {
+	d, err := os.Open(dir)
+	if err != nil {
+		return
+	}
+	_, err = d.Readdir(1)
+	_ = d.Close()
+	if err == io.EOF {
+		_ = os.Remove(dir)
+	}
```
arielshaqed, line 158: It's usually best to try something on a filesystem rather than to try to reproduce the OS internal logic.  In this case, I think the current code fails when the user has execute but not read permissions on directory.
Perhaps something like this (untested)?
```suggestion
	err := os.Remove(dir)
	if err != nil && !errors.Is(err, syscall.ENOTEMPTY)) {
		return err
	}
	return nil
```

arielshaqed (approved):
Great, thanks!
No reason not to delete empty directories, AFAIR it was just an oversight because S3 doesn't have directories.

block/local/adapter.go
nopcoder: cool, so if we also do not report errors related to our empty folder cleanup I just call

```go
_ = os.Remove(dir)
```

nopcoder:
> @arielshaqed, I assume this is part of the solution to the problem we are keeping empty folders.
> do we like to enable it just for export, or the issue is the fact that there is no "Delete" when we continuous export?

bug related question - if I'll enable the above by default, do I missing something related to fixing the associated bug?

block/local/adapter.go
arielshaqed: Yup!  But I'd rather still log the error before ignoring it, information always helps debug issues.  Of course if you did that you'd still need the `if` above :-/

arielshaqed:
> > @arielshaqed, I assume this is part of the solution to the problem we are keeping empty folders.
> > do we like to enable it just for export, or the issue is the fact that there is no "Delete" when we continuous export?
> 
> bug related question - if I'll enable the above by default, do I missing something related to fixing the associated bug?

The only thing you're missing is that hopefully export will also go away soon, it can be an external tool once we have a thick Spark (or similar) client.

arielshaqed (approved):

nopcoder:
Great, +1 about the export reimplementation - just wanted to crush all the open bugs

arielshaqed (changes requested):
Actually doesn't fix the bug.  Need to go *up* along the tree and continue to remove directories, up to (but not including) the adapter root directory :-(

block/local/adapter.go
```diff
@@ -127,7 +140,28 @@ func (l *Adapter) Remove(obj block.ObjectPointer) error {
...
+}
+
+func removeEmptyDirUntil(dir string, stopAt string) {
+	for {
+		err := os.Remove(dir)
+		if err != nil {
+			break
+		}
+		dir = filepath.Dir(dir)
+		if dir == "/" || strings.HasPrefix(stopAt, dir) {
```
arielshaqed: This tests whether `stopAt` begins with `dir` (https://golang.org/pkg/strings/#HasPrefix).  Given that this is supposed to protect files that are outside the adapter's area of responsibility it might be safer to say at the top of the loop
```go
for dir != "" && strings.HasPrefix(dir, stopAt)
```

The differences are in what happens when the initial condition `strings.HasPrefix(dir, stopAt)` doesn't hold when the function is called (we don't want the loop to go all the way to the root).

block/local/adapter_test.go
```diff
@@ -141,3 +142,56 @@ func TestLocalCopy(t *testing.T) {
...
+	}
+}
+
+func dumpPathTree(t testing.TB, root string) []string {
+	tree := make([]string, 0)
+	err := filepath.Walk(root, func(path string, _ os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		p := strings.TrimPrefix(path, root)
```
arielshaqed, line 201: I think this is `FileInfo.Name()` (apart from the first step, at the root... sigh...).

block/local/adapter_test.go
```diff
@@ -141,3 +142,56 @@ func TestLocalCopy(t *testing.T) {
...
+	testutil.MustDo(t, "Put", err)
+	err = a.Remove(obj3)
+	testutil.MustDo(t, "Remove", err)
+	tree3 := dumpPathTree(t, root)
+	if diff := deep.Equal(tree3, []string{""}); diff != nil {
+		t.Fatalf("Found diff after remove: %s", diff)
+	}
+}
+
+func dumpPathTree(t testing.TB, root string) []string {
```
arielshaqed, line 195: Not sure what this function does.  I think it works well (only) when the passed-in directory contains only zero or one elements, recursively all the way down.  Otherwise `Walk` does not guarantee the order in which files are passed to the callback.
Doesn't matter in *this* test, I think, but it's worthy of at least documentation.  Or if you want to fix, consider collecting entire `TrimPrefix(path, root)`s and then sorting.

block/local/adapter_test.go
```diff
@@ -141,3 +142,56 @@ func TestLocalCopy(t *testing.T) {
...
+
+	obj3 := makePointer("a/b/c/d.txt")
+	err = a.Put(obj3, 0, strings.NewReader(contents), block.PutOpts{})
+	testutil.MustDo(t, "Put", err)
+	err = a.Remove(obj3)
+	testutil.MustDo(t, "Remove", err)
+	tree3 := dumpPathTree(t, root)
+	if diff := deep.Equal(tree3, []string{""}); diff != nil {
+		t.Fatalf("Found diff after remove: %s", diff)
+	}
```
arielshaqed: Loop over a slice of test cases?

block/local/adapter_test.go
```diff
@@ -141,3 +142,56 @@ func TestLocalCopy(t *testing.T) {
 		t.Errorf("expected to read \"%s\" as written, got \"%s\"", contents, string(got))
 	}
 }
+
+func TestRemove(t *testing.T) {
```
arielshaqed, line 148: This doesn't test what happens when an object is removed and *not* all directories are deleted.  E.g. add `a/b/c/d`, `/a/b/e/f` and remove `/a/b/e/f`.

arielshaqed (approved):
Thanks!

Comments about the test.

block/local/adapter.go
nopcoder: Agree, at first, I wanted to support the case there is no stop and I just need to remove the folder that holds the file.

block/local/adapter_test.go
nopcoder: Because it is a temporary name I remote the full prefix - remove just the name is problematic

block/local/adapter_test.go
nopcoder: I will keep the result sorted - but I must remove the prefix as it is a temporary folder.

block/local/adapter_test.go
nopcoder: added such test

block/local/adapter.go
nopcoder: I have to test using `!strings.HasPrefix(stopAt, dir)` - don't want to test the case of the adapter root folder

block/local/adapter.go
```diff
@@ -127,7 +140,28 @@ func (l *Adapter) Remove(obj block.ObjectPointer) error {
...
+	}
+	if l.removeEmptyDir {
+		dir := filepath.Dir(p)
+		removeEmptyDirUntil(dir, l.path)
+	}
+	return nil
+}
+
+func removeEmptyDirUntil(dir string, stopAt string) {
+	for stopAt != "" && !strings.HasPrefix(stopAt, dir) {
```
arielshaqed, line 155: Sorry, still not sure why the reversal is the safer choice.
```suggestion
	for stopAt != "" && strings.HasPrefix(dir, stopAt) {
```
has the advantage (to my mind) that it stops sooner rather than later when `stopAt` is not a prefix of `dir`.  (You might want to ensure `stopAt` ends in a slash, but that is still safer).  Unless I'm very confused, this code does the wrong thing if an external user can convince a local adapter to remove a file that is not under the control of the adapter.  This is currently *NOT* possible, but I still prefer the safer code.

arielshaqed (approved):
Thanks!

block/local/adapter.go
nopcoder: To make `stopAt != "" && strings.HasPrefix(dir, stopAt)` work I also need the add the == case.
In case we use "local://" I don't want to delete the adapter directory.

nopcoder: Updated the code

block/local/adapter_test.go
```diff
@@ -141,3 +144,79 @@ func TestLocalCopy(t *testing.T) {
...
+				obj := makePointer(o)
+				testutil.MustDo(t, "Put", adp.Put(obj, 0, strings.NewReader(content), block.PutOpts{}))
+			}
+			// test Remove with remove empty folders
+			obj := makePointer(tt.path)
+			if err := adp.Remove(obj); (err != nil) != tt.wantErr {
+				t.Errorf("Remove() error = %v, wantErr %v", err, tt.wantErr)
+			}
+			tree := dumpPathTree(t, adp.Path())
+			if diff := deep.Equal(tt.wantTree, tree); diff != nil {
```
arielshaqed, line 217: This comparison will be flaky AFAICT, depending even on properties of the underlying *filesystem*!  `filepath.Walk` doesn't guarantee order of returned objects.  Best to sort `tree` first.

block/local/adapter.go
```diff
@@ -127,7 +140,28 @@ func (l *Adapter) Remove(obj block.ObjectPointer) error {
...
+	}
+	if l.removeEmptyDir {
+		dir := filepath.Dir(p)
+		removeEmptyDirUntil(dir, l.path)
+	}
+	return nil
+}
+
+func removeEmptyDirUntil(dir string, stopAt string) {
+	for stopAt != "" && strings.HasPrefix(dir, stopAt) && dir != stopAt {
```
arielshaqed, line 155: Yikes... if `stopAt` doesn't end in `/` this can delete a file _next to it_.  E.g. say the block adapter is rooted at `/home/ariels/data`.  Then this will happily delete a file `/home/ariels/datadog`.
So (_also_...!) need to append a `/` to `stopAt` if it does not already end in `/`'

arielshaqed (approved):
Great, thanks!

block/local/adapter_test.go
nopcoder: I forgot the sort!

block/local/adapter.go
nopcoder: added / to stopAt if needed.

arielshaqed (commented):
Hurrah!  Sorry for all the nit-picking.  POSIX directories are _the worst_.

nopcoder:
> Hurrah! Sorry for all the nit-picking. POSIX directories are _the worst_.

Very good feedback - thank you!

PR #1350, opened 2021-01-28
Control number of concurrent range writer uploaders

### LINKED ISSUE #1319: Control the amount of concurrent closed sstable writers
A flux of ongoing uploads on a slow internet connection is bound to starve some requests and fail the entire metarange write operation

arielshaqed:
Fixes #1319.

Requires replacing use of mock range writers with fake range writers!  Goroutines, go testing
and go mocks cannot work together: if a mock receives an unexpected parameter it cannot throw
an exception (because Go...), so it calls some `testing.t.Fa*`.  Unfortunately testing cannot
support failing in goroutines.


arielshaqed:
Branch renamed.  On the third minute look to GitHub for a new PR!

PR #1352, opened 2021-01-28
Unref cache from RangeManager at finalization

### LINKED ISSUE #1349: pebble: cache (0xc0004e0190) has non-zero reference count: 2
@ozkatz reports about running integration tests with full rocks:
> Part of that also moved the API's integration tests to use a rocks cataloger (apparently it wasn't using it until now). Ever since, running the API's tests with the race detector turned on passes. Using it without, fails on `pebble: cache (0xc0004e0190) has non-zero reference count: 2`.

arielshaqed:
Required (*only!*) to prevent pebble crashing the program when it finalizes due to a cache
having a positive reference count.

Why go further into the finalization hole?  There is no real choice.  An alternative would
be to add a Close() method to RangeManager, then add such a method *everywhere* that holds
a RangeManager, all the way up the stack.

Fixes #1349: with this patched in, `go test -v --count 1 github.com/treeverse/lakefs/api/...` passes (_without_ `--race`...).

### DIFF
--- a/graveler/sstable/range_manager.go
+++ b/graveler/sstable/range_manager.go
@@ -6,6 +6,7 @@ import (
 	"crypto"
 	"errors"
 	"fmt"
+	"runtime"
 
 	"github.com/cockroachdb/pebble"
 	"github.com/cockroachdb/pebble/sstable"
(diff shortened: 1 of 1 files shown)


itaiad200 (approved):
It's my bad - we've experienced it before. When we discussed that in the previous PR I didn't find evidence in Pebble's code to the panic scenario and assumed it was fixed in the current version - sorry.

Small suggestion (probably not fixable in this PR): Move cache init up and propagate it all the way down, then you can defer the `Unref` and avoid finalizers..

arielshaqed:
> It's my bad - we've experienced it before. When we discussed that in the previous PR I didn't find evidence in Pebble's code to the panic scenario and assumed it was fixed in the current version - sorry.

I _did_ brush it off as a bug.  What I _should_ have said is "this is a bug in the pebble SSTable implementation when used outside of CockroachDB".

> Small suggestion (probably not fixable in this PR): Move cache init up and propagate it all the way down, then you can defer the `Unref` and avoid finalizers..

This is the "destructors" solution, only we make it the caller's responsibility.  To make this work you would need to construct the pebble Cache very high up and _not_ Ref the cache inside the RangeManager.  But there are already currently 4 calls to create this manager, would need to push each one up.  Now e.g. `NewEntryCatalog` is called from 3 places, 2 commands (which can deref immediately) and the other in `NewCataloger`.  Now it seems we're good, because `NewCataloger` is going away.  Unfortunately we shall still need to `BuildCataloger`, even if not in a factory... and there are 10 calls to _that_.  So explicit destruction would have to happen in 14 places, all of them far away from `RangeManager` construction.  This is hard and nonscalable.

Destructors rule.

arielshaqed:
Thanks!

@ozkatz I'm pulling so you can rebase onto master and see that it helps.  I *do* want your review, so please do review and I shall reply and change after-the-fact in a separate PR.

ozkatz:
Thanks @arielshaqed! It indeed solves my issue (rebased mvcc removal on master).

PR #1353, opened 2021-01-28
allow revert merge commit

johnnyaug:


catalog/cataloger.go
```diff
@@ -35,6 +35,12 @@ type DiffParams struct {
 	AdditionalFields []string // db fields names that will be load in additional to Path on Difference's Entry
 }
 
+type RevertParams struct {
+	Reference    string // the commit to revert
+	ParentNumber int    // if reverting a merge commit, the change will be reversed relative to this parent number (1-based).
```
nopcoder, line 40: can we make at least the api use a default - so 0 will be like 1?

catalog/rocks/entry_catalog.go
```diff
@@ -346,33 +346,33 @@ func (e *EntryCatalog) ResetPrefix(ctx context.Context, repositoryID graveler.Re
 	return e.Store.ResetPrefix(ctx, repositoryID, branchID, keyPrefix)
 }
 
-func (e *EntryCatalog) Revert(ctx context.Context, repositoryID graveler.RepositoryID, branchID graveler.BranchID, ref graveler.Ref, committer string, message string, metadata graveler.Metadata) (graveler.CommitID, error) {
+func (e *EntryCatalog) Revert(ctx context.Context, repositoryID graveler.RepositoryID, branchID graveler.BranchID, ref graveler.Ref, parentNumber int, commitParams graveler.CommitParams) (graveler.CommitID, error) {
```
nopcoder, line 349: validate parentNumber

cmd/lakectl/cmd/branch.go
```diff
@@ -252,6 +260,8 @@ func init() {
 	branchResetCmd.Flags().String("prefix", "", "prefix of the objects to be reset")
 	branchResetCmd.Flags().String("object", "", "path to object to be reset")
 
+	branchRevertCmd.Flags().IntP(ParentNumberFlagName, "m", 0, "the parent number (starting from 1) of the mainline. The revert will reverse the change relative to the specified parent.")
```
nopcoder, line 263: 1. why IntP()?
2. want the default to be 1?

graveler/graveler.go
```diff
@@ -1043,15 +1049,22 @@ func (g *Graveler) ResetPrefix(ctx context.Context, repositoryID RepositoryID, b
...
+
+		if len(commitRecord.Parents) > 1 && parentNumber <= 0 {
+			return "", ErrRevertMergeNoParent
+		}
+		if parentNumber > 0 {
+			if parentNumber > len(commitRecord.Parents) { // parent number is 1-based
+				return "", fmt.Errorf("%w: parent %d", ErrRevertParentOutOfRange, parentNumber)
+			}
+			parentNumber--
 		}
```
nopcoder, line 1068: 1. parameter verification can be done outside the lock
2. Think we can consider 0 as no value was passed and assume the default (which is 1)

swagger.yml
```diff
@@ -1627,6 +1627,9 @@ paths:
               ref:
                 type: string
                 description: the commit to revert, given by a ref
+              parent_number:
```
nopcoder, line 1630: For ease use of the API we can set the default also here to be 1

nopcoder (changes requested):

catalog/rocks/validate.go
```diff
@@ -191,5 +191,16 @@ func ValidateRequiredString(v interface{}) error {
 	return nil
 }
 
+func ValidatePositiveInt(v interface{}) error {
+	i, ok := v.(int)
+	if !ok {
+		panic(ErrInvalidType)
+	}
+	if i <= 0 {
```
nopcoder, line 199: < 0? Thought this parameter can be 0 in case the client didn't pass any decision in the request.

cmd/lakectl/cmd/branch.go
```diff
@@ -133,11 +136,16 @@ var branchRevertCmd = &cobra.Command{
 		u := uri.Must(uri.Parse(args[0]))
 		commitRef := args[1]
 		clt := getClient()
+		hasParentNumber := cmd.Flags().Changed(ParentNumberFlagName)
+		parentNumber, _ := cmd.Flags().GetInt(ParentNumberFlagName)
+		if hasParentNumber && parentNumber <= 0 {
+			Die("parent number must be non-negative, if specified", 1)
+		}
```
nopcoder, line 143: Why do you need Changed? we only validate it is non-negative value and the default is zero, which tells the server no selection was made.

graveler/graveler.go
```diff
@@ -1043,16 +1049,24 @@ func (g *Graveler) ResetPrefix(ctx context.Context, repositoryID RepositoryID, b
...
+	if err != nil {
+		return "", fmt.Errorf("get commit from ref %s: %w", ref, err)
+	}
+	if len(commitRecord.Parents) > 1 && parentNumber <= 0 {
+		// if commit has more than one parent, must explictly specify parent number
+		return "", ErrRevertMergeNoParent
+	}
+	if parentNumber > 0 {
+		// validate parent is in range:
+		if parentNumber > len(commitRecord.Parents) { // parent number is 1-based
```
nopcoder, line 1064: >= or do -- before

nopcoder (changes requested):

cmd/lakectl/cmd/branch.go
johnnyaug: @nopcoder "changed" means that you explicitly specified it. If you do so, zero is not valid

graveler/graveler.go
johnnyaug: @nopcoder didn't want to `--` before, because I use the number for the error
`>` is valid here, I think.

catalog/rocks/validate.go
johnnyaug: Oops! fixed

graveler/graveler.go
nopcoder: My bad

nopcoder (approved):

PR #1359, opened 2021-01-28
Compute applied diff summary and pass it out on cataloger merge

### LINKED ISSUE #1300: Merge operation should return a summary of the merge
In the MVCC implementation, the cataloger would return a summary of the merge. The CLI would then print this summary. The summary includes the number of added, deleted and changed files.

Decide whether this feature is needed, and if so, implement it for the Rocks catalog.

arielshaqed:
The summary is still thrown away for commits and for reverts, but _could_ be used there too (or to supply a status summary by passing a non-writing writer).

Fixes #1300.

### DIFF
--- a/catalog/rocks/cataloger.go
+++ b/catalog/rocks/cataloger.go
@@ -556,7 +556,7 @@ func (c *cataloger) Revert(ctx context.Context, repository string, branch string
 	repositoryID := graveler.RepositoryID(repository)
 	branchID := graveler.BranchID(branch)
 	ref := graveler.Ref(reference)
-	_, err := c.EntryCatalog.Revert(ctx, repositoryID, branchID, ref, committer, fmt.Sprintf("Revert %s", reference), nil)
+	_, _, err := c.EntryCatalog.Revert(ctx, repositoryID, branchID, ref, committer, fmt.Sprintf("Revert %s", reference), nil)
 	return err
 }
 
@@ -606,7 +606,10 @@ func listDiffHelper(it EntryDiffIterator, limit int, after string) (catalog.Diff
 		if v.Path == afterPath {
 			continue
 		}
-		diff := newDifferenceFromEntryDiff(v)
+		diff, err := newDifferenceFromEntryDiff(v)
+		if err != nil {
+			return nil, false, fmt.Errorf("[I] %w", err)
+		}
 		diffs = append(diffs, diff)
 		if len(diffs) >= limit+1 {
 			break
@@ -628,7 +631,7 @@ func (c *cataloger) Merge(ctx context.Context, repository string, leftBranch str
 	leftRef := graveler.Ref(leftBranch)
 	rightBranchID := graveler.BranchID(rightBranch)
 	meta := graveler.Metadata(metadata)
-	commitID, err := c.EntryCatalog.Merge(ctx, repositoryID, leftRef, rightBranchID, committer, message, meta)
+	commitID, summary, err := c.EntryCatalog.Merge(ctx, repositoryID, leftRef, rightBranchID, committer, message, meta)
 	if errors.Is(err, graveler.ErrConflictFound) {
 		// for compatibility with old cataloger
 		return &catalog.MergeResult{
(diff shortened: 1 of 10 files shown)


catalog/rocks/cataloger.go
```diff
@@ -638,9 +641,18 @@ func (c *cataloger) Merge(ctx context.Context, repository string, leftBranch str
 	if err != nil {
 		return nil, err
 	}
+	count := make(map[catalog.DifferenceType]int)
+	if summary.Count != nil {
```
nopcoder, line 645: Go's range works on nil, no need to check

catalog/rocks/cataloger.go
```diff
@@ -691,18 +703,29 @@ func newCatalogEntryFromEntry(commonPrefix bool, path string, ent *Entry) catalo
 	return catEnt
 }
 
-func newDifferenceFromEntryDiff(v *EntryDiff) catalog.Difference {
-	var diff catalog.Difference
-	switch v.Type {
+var ErrUnknownDiffType = errors.New("unknown graveler difference type")
```
nopcoder, line 706: can you move this line to the start of the file - it will help to find globals and keep a file structure

cmd/lakectl/cmd/merge.go
```diff
@@ -37,12 +37,13 @@ var mergeCmd = &cobra.Command{
...
 		if errors.Is(err, catalog.ErrConflictFound) {
-			DieFmt("%d conflict(s) found\n", result.Summary.Conflict)
+			_, _ = fmt.Printf("Conflicts: %d\n", result.Summary.Conflict)
+			return
 		}
 		if err != nil {
 			DieErr(err)
 		}
-		fmt.Printf("Merged %s\n", result.Reference)
+		_, _ = fmt.Printf("new: %d modified: %d removed: %d\n", result.Summary.Added, result.Summary.Changed, result.Summary.Removed)
```
nopcoder, line 46: Can we also have the commit id (Reference)?

graveler/committed/apply.go
```diff
@@ -14,6 +14,18 @@ type ApplyOptions struct {
...
 
+func add(d *graveler.DiffSummary, typ graveler.DiffType, n int) {
+	if d.Count != nil {
+		d.Count[typ] += n
+	}
+}
+
+func inc(d *graveler.DiffSummary, typ graveler.DiffType) {
+	add(d, typ, 1)
+}
```
nopcoder, line 25: Consider move them into methods of `DiffSummary` or add diffSummary to the function name - package level func named 'add' is a bit general

graveler/graveler.go
```diff
@@ -1105,15 +1115,16 @@ func (g *Graveler) Revert(ctx context.Context, repositoryID RepositoryID, branch
...
-		return commitID, nil
+		return &CommitIDAndSummary{commitID, summary}, nil
 	})
+	c := res.(*CommitIDAndSummary)
 	if err != nil {
-		return "", err
+		return "", DiffSummary{}, err
 	}
-	return commitID.(CommitID), nil
+	return c.ID, c.Summary, nil
```
nopcoder, line 1124: Usually, I call the cast after the error check in case somebody will return just an error it will not panic.

graveler/graveler.go
```diff
@@ -1158,12 +1169,13 @@ func (g *Graveler) Merge(ctx context.Context, repositoryID RepositoryID, from Re
 		if err != nil {
 			return "", fmt.Errorf("update branch %s: %w", to, err)
 		}
-		return commitID, nil
+		return &CommitIDAndSummary{commitID, summary}, nil
 	})
+	c := res.(*CommitIDAndSummary)
```
nopcoder, line 1174: Same as before, if possible do this after the error check

nopcoder (changes requested):
Cool - minor comments

catalog/rocks/cataloger.go
arielshaqed: Cool, thanks!

catalog/rocks/cataloger.go
arielshaqed: Sure.

cmd/lakectl/cmd/merge.go
arielshaqed: Good idea.  Added it, and moved to a proper template and all, and even added some more fields to be (a bit) more like `lakectl commit`.

graveler/committed/apply.go
arielshaqed: Done.  Using a longer name rather than a method, because these are not really methods and the summary is POD.

graveler/graveler.go
arielshaqed: :-) Thanks!

graveler/graveler.go
arielshaqed: Yup!

arielshaqed (commented):
Thanks!

PTAL...

nopcoder (approved):
🤩

arielshaqed:
Thanks!  Pu{sh,ll}ing...

PR #1360, opened 2021-01-29
Add testutil env to reuse DB container while running tests

nopcoder:
Test connect DB support "reuse-db" flag - it will locate Docker container names "postgres" and use it for testing.

```sh
TESTUTIL_DB_CONTAINER=postgres go test
```

Test connect DB will lookup "postgres" Docker container name and extract connection information parameters without starting or stopping the container.

### DIFF
--- a/testutil/db.go
+++ b/testutil/db.go
@@ -27,8 +27,6 @@ import (
 )
 
 const (
-	TimeFormat = "Jan 2 15:04:05 2006 -0700"
-
 	DBContainerTimeoutSeconds = 60 * 30 // 30 minutes
 
 	EnvKeyUseBlockAdapter = "USE_BLOCK_ADAPTER"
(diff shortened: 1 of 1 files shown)


ozkatz:
What happens if the discovered container gets killed by the module that spawned it?

nopcoder:
> What happens if the discovered container gets killed by the module that spawned it?

The idea is having a running container and running the all the tests to use the running instance.
It means that each package, if needed, will connect to the running instance and not run a new Postgres instance.

Thought about the above as part of how to help reduce the build/test time as we don't have global run once for `go test`.
The same can be used by passing `-db` and the address of the database. This one is a bit easier to use for a developer, as it will extract all the connection information from the container's environment.
All you need to make sure you have a container names 'postgres'.

testutil/db.go
```diff
@@ -37,10 +37,21 @@ const (
...
-var addrDB = flag.String("db", "", "DB address to use")
+var (
+	keepDB  = flag.Bool("keep-db", false, "keep test DB instance running")
+	addrDB  = flag.String("db", "", "DB address to use")
+	reuseDB = flag.Bool("reuse-db", false, "re-use running DB Docker container")
+)
 
 func GetDBInstance(pool *dockertest.Pool) (string, func()) {
+	if *reuseDB {
+		resource, ok := pool.ContainerByName("postgres")
```
ozkatz, line 48: Perhaps read name from an env var? Makes it a little more flexible

nopcoder: Done

ozkatz: Sorry for being annoying: why not accept a PG_TEST_URI and use that? Could be a docker container or simply a locally running pg instance, which might aid in local debugging. On a local dev machine, one could set it and forget it..

nopcoder: Agree, I can use `-db` the above code just make it 'auto-discover' as it extracts the connection information from the container.

nopcoder:
The PR made connect to a running DB while using docker easier. But we can leverage `-db` to have the same effect without the auto-configure DB URI capability.

nopcoder:
Will keep the new functionality and use `PG_TEST_URI` for connecting the existing database, `PG_TEST_CONTAINER` for connecting the existing container.

nopcoder:
Rebased over current master and apply the changes suggested.

ozkatz (approved):
Thanks!

PR #1361, opened 2021-01-30
API package cleanup, cataloger dedup and expire

nopcoder:
- api_controller -> controller
- turn api's Handler into a `func Serve() http.Handler`
- remove cataloger dedup and mark for expiration


api/auth_handlers_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	t.Run("invalid Auth secret", func(t *testing.T) {
+		invalidAuth := client.BasicAuth(creds.AccessKeyID, "foobarbaz")
+		_, err := clt.Repositories.ListRepositories(repositories.NewListRepositoriesParams().WithTimeout(timeout), invalidAuth)
+		var unauthorized *repositories.ListRepositoriesUnauthorized
+		if !errors.As(err, &unauthorized) {
+			t.Fatalf("got %s not unauthorized error", err)
+		}
+	})
+}
```
arielshaqed, line 31: Can we also test JWT wiring?

api/cookie_api_handler.go
```diff
@@ -0,0 +1,17 @@
...
+
+const JWTAuthorizationHeaderName = "X-JWT-Authorization"
+
+// CookieAPIHandler set JWT header with JWT cookie value
+func CookieAPIHandler(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// read cookie (no need to validate, this will be done in the API
+		cookie, err := r.Cookie(JWTCookieName)
+		if err == nil {
+			r.Header.Set(JWTAuthorizationHeaderName, cookie.Value)
```
arielshaqed, line 13: I realize this is existing behaviour.  But per https://jwt.io/introduction this is the wrong header to use:
> Whenever the user wants to access a protected route or resource, the user agent should send the JWT, typically in the Authorization header using the Bearer schema. The content of the header should look like the following:
>
> ```
> Authorization: Bearer <token>
> ```

We can (and I think should) change it here, and eventually on the client side too.  This should be in a separate PR, so I opened #1362.

arielshaqed (approved):
Great, thanks!

Let's open a bug to decide about dedupe and maybe revive it, so all know it's missing right now.

api/api.go
```diff
@@ -0,0 +1,74 @@
...
+	_ "github.com/treeverse/lakefs/statik"
+	"github.com/treeverse/lakefs/stats"
+)
+
+const (
+	RequestIDHeaderName = "X-Request-ID"
+	LoggerServiceName   = "rest_api"
+)
+
+func Serve(
```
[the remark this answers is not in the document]
nopcoder: I'll name it `serve.go` so it will be easy to locate Serve func

api/api.go
```diff
@@ -0,0 +1,74 @@
...
+	cataloger catalog.Cataloger,
+	blockStore block.Adapter,
+	authService auth.Service,
+	metadataManager auth.MetadataManager,
+	stats stats.Collector,
+	migrator db.Migrator,
+	parade parade.Parade,
+	logger logging.Logger,
+) http.Handler {
+	logger.Info("initialized OpenAPI server")
```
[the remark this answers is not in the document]
nopcoder: true :)

api/api.go
```diff
@@ -0,0 +1,74 @@
...
+		return httputil.LoggingMiddleware(
+			RequestIDHeaderName,
+			logging.Fields{"service_name": LoggerServiceName},
+			promhttp.InstrumentHandlerCounter(requestCounter,
+				MetricsHandler(api.Context(),
+					CookieAPIHandler(handler))))
+	})
+	uiHandler := UIHandler(authService)
+
+	mux := http.NewServeMux()
```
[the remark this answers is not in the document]
nopcoder: Agree - https://github.com/treeverse/lakeFS/issues/1363

api/cookie_api_handler.go
nopcoder: Agree, I just wrote a test today for the JWT auth, then I realized I need to set a non-standard header.

api/auth_handlers_test.go
nopcoder: Had race condition - https://github.com/treeverse/lakeFS/blob/master/api/handler_test.go#L177

api/api.go
```diff
@@ -0,0 +1,74 @@
...
+
+	api := operations.NewLakefsAPI(swaggerSpec)
+	api.Logger = func(msg string, ctx ...interface{}) {
+		logging.Default().WithField("logger", "swagger").Debugf(msg, ctx)
+	}
+	api.BasicAuthAuth = BasicAuthHandler(authService)
+	api.JwtTokenAuth = JwtTokenAuthHandler(authService)
+
+	// bind our handlers to the server
+	controller := NewController(cataloger, authService, blockStore, stats, parade, metadataManager, migrator, stats, logger)
```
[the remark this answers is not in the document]
nopcoder: Passing Controller will require also expose some of the services - like `auth.Service`.
Checking if I can pass a Services struct instead.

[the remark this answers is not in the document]
nopcoder: Used Dependencies struct - also found dup field there :)

api/controller.go
```diff
@@ -59,12 +59,11 @@ type Dependencies struct {
 	Cataloger       catalog.Cataloger
 	Auth            auth.Service
 	BlockAdapter    block.Adapter
-	Stats           stats.Collector
```
arielshaqed, line 63: [stats or collector?  why can't we have both?!](https://imgflip.com/gif/4w2ztr)

arielshaqed: OMG gifs werk in github.  The internet is _so_ over.

arielshaqed (approved):
Even greaterer!

nopcoder:
@arielshaqed just a small change I've added during the merge with master - when authentication fails I do not log in with Warn, just Debug so production login tries will not fill the log.
Let me know if you still think it is worth it.

api/controller.go
nopcoder: wonder how much it is easy to see that we are passing the same service twice :)

arielshaqed (commented):
Thanks!

StillLGTM.  In future please consider rebasing rather than merging.  Merging is harder (for me) to review, since I need to replay changes in files like `apply_test.go` and `wr_file.go` (_most_ of the changes in the latest commit!) only to see that the only changes are those from the merge.

PR #1364, opened 2021-01-31
Cleanup diagnostics

itaiad200:


diagnostics/collector.go
```diff
@@ -24,29 +43,24 @@ const (
...
 	writer := zip.NewWriter(w)
 	defer func() { err = writer.Close() }()
 
-	errs := c.collectWithZip(ctx, writer)
+	errs := c.collectDB(ctx, writer)
 
-	// write all errors into log
-	if err := writeErrors(writer, errs); err != nil {
-		return fmt.Errorf("write errors: %w", err)
+	rangeErrs := c.rangesStats(ctx, writer)
```
nopcoder, line 53: Check github.com/hashicorp/go-multierror we already using it in the project
I think it will be better than named return parameters with the last error found

diagnostics/collector.go
```diff
@@ -24,29 +43,24 @@ const (
...
+		err = fmt.Errorf("write errors: %w", err)
 	}
 
 	return nil
 }
 
-// Collect query information from the database into csv files and write everything to io writer
-func (c *DBCollector) collectWithZip(ctx context.Context, writer *zip.Writer) []error {
+// collectDB query information from the database into csv files and write everything to io writer
+func (c *Collector) collectDB(ctx context.Context, writer *zip.Writer) []error {
```
nopcoder, line 63: We can stop collecting information from `catalog_*` tables

diagnostics/collector.go
```diff
@@ -165,16 +184,53 @@ func (c *DBCollector) writeRawQueryContent(ctx context.Context, writer *zip.Writ
...
-	return c.writeQueryContent(ctx, writer, name, q)
-}
-
-func (c *DBCollector) writeTableCount(ctx context.Context, writer *zip.Writer, name string) error {
+func (c *Collector) writeTableCount(ctx context.Context, writer *zip.Writer, name string) error {
 	q := sq.Select("COUNT(*)").From(name)
 	return c.writeQueryContent(ctx, writer, name+"_count", q)
 }
 
+func (c *Collector) rangesStats(ctx context.Context, writer *zip.Writer) []error {
```
nopcoder, line 192: same as before I think the multierror will be a better solution than the slice of errors

diagnostics/collector.go
```diff
@@ -165,16 +184,53 @@ func (c *DBCollector) writeRawQueryContent(ctx context.Context, writer *zip.Writ
...
+			return nil
+		}
+
+		if err := c.adapter.Walk(block.WalkOpts{
+			StorageNamespace: repo.StorageNamespace,
+			Prefix:           c.pyramidCfg.BlockStoragePrefix,
+		}, counter); err != nil {
+			errs = append(errs, fmt.Errorf("listing ranges and meta-ranges: %w", err))
+		}
+		if err := csvWriter.Write([]string{repo.Name, "range", strconv.Itoa(count)}); err != nil {
```
nopcoder, line 226: You still try to write the count to the CSV even if the previous step failed, it could be misleading.

nopcoder (changes requested):

diagnostics/collector.go
itaiad200: Done.

diagnostics/collector.go
itaiad200: Done.

diagnostics/collector.go
itaiad200: Done.

diagnostics/collector.go
itaiad200: I agree but think it's valuable. 
Failing on the first iteration isn't like failing on the n'th so writing the number of ranges seen is good debugging information.

diagnostics/collector.go
itaiad200: Done.

nopcoder (approved):

diagnostics/collector.go
```diff
@@ -0,0 +1,250 @@
...
+}
+
+func writeErrors(writer *zip.Writer, err error) error {
+	if err == nil {
+		return nil
+	}
+	combinedErr, ok := err.(*multierror.Error)
+	if !ok {
+		logging.Default().Error("error not of type multierror.Error: %w", err)
+	}
```
nopcoder, line 239: Think you can remove the cast and just print the error - the multierror got a multiline format so the single print should work for any error

nopcoder (changes requested):
dump the multierror without cast

nopcoder (approved):

PR #1365, opened 2021-01-31
Implement committed.Iterator.NextRange() without recursive call

### LINKED ISSUE #1178: Avoid recursive calls on committed iterator
Currently, graveler's committed iterator implements NextRange() by calling NextRange.
It is done for a specific case where part ends, but still, I think it will be a better option not to include a recursive call while iterating.

arielshaqed:
Better because golang does not support tail recursion.

Fixes #1178.


guy-har (approved):
Cool
Looks great

arielshaqed:
Thanks!

PR #1376, opened 2021-01-31
Test tags list open API

nopcoder:


api/controller_test.go
```diff
@@ -545,6 +546,98 @@ func TestController_ListBranchesHandler(t *testing.T) {
...
+				bauth)
+			testutil.Must(t, err)
+			payload := resp.GetPayload()
+			results = append(results, payload.Results...)
+			if !swag.BoolValue(payload.Pagination.HasMore) {
+				break
+			}
+			after = payload.Pagination.NextOffset
+		}
+		expectedCalls := createTagLen/pageSize + 1
```
itaiad200, line 621: If `createTagLen` was 8, wouldn't you expect just 4 calls instead of 5?
It's ok to make it const IMO.

api/controller_test.go
```diff
@@ -545,6 +546,98 @@ func TestController_ListBranchesHandler(t *testing.T) {
...
+			t.Fatal("ListTags results diff:", diff)
+		}
+	})
+
+	t.Run("no repository", func(t *testing.T) {
+		_, err := clt.Tags.ListTags(
+			tags.NewListTagsParamsWithTimeout(timeout).
+				WithRepository("repoX"),
+			bauth)
+		if err == nil {
```
itaiad200, line 635: Add validation for the error type.

itaiad200 (approved):
LGTM - just 2 minor comments.

api/controller_test.go
nopcoder: true - not good code, fixing.

Issue #1373, opened 2021-01-31
BranchLocker test code fail sometimes


nopcoder:
branch_locker_test fails sometimes with time out: `panic: test timed out after 10m0s`

```
/home/runner/work/lakeFS/lakeFS/graveler/ref/branch_locker_test.go:122 +0x246
```

[logs_12078.zip](https://github.com/treeverse/lakeFS/files/5899834/logs_12078.zip)

arielshaqed:
I think I've seen this, too.  E.g. https://github.com/treeverse/lakeFS/pull/1351#issuecomment-770358448 is _precisely_ this test breakage, and went away on a rebase... +rerun.

arielshaqed:
E.g. this failure: https://github.com/treeverse/lakeFS/actions/runs/518485340

arielshaqed:
Here are the *relevant* stacktraces from the above failing test:
```

goroutine 147 [semacquire, 9 minutes]:
sync.runtime_Semacquire(0xc000043f28)
	/opt/hostedtoolcache/go/1.15.2/x64/src/runtime/sema.go:56 +0x45
sync.(*WaitGroup).Wait(0xc000043f20)
	/opt/hostedtoolcache/go/1.15.2/x64/src/sync/waitgroup.go:130 +0xe5
github.com/jackc/puddle.(*Pool).Close(0xc0001fe880)
	/home/runner/go/pkg/mod/github.com/jackc/puddle@v1.1.3/pool.go:163 +0x31d
github.com/jackc/pgx/v4/pgxpool.(*Pool).Close(...)
	/home/runner/go/pkg/mod/github.com/jackc/pgx/v4@v4.10.1/pgxpool/pool.go:315
github.com/treeverse/lakefs/testutil.GetDB.func1()
	/home/runner/work/lakeFS/lakeFS/testutil/db.go:143 +0x85
testing.(*common).Cleanup.func1()
	/opt/hostedtoolcache/go/1.15.2/x64/src/testing/testing.go:872 +0x104
testing.(*common).runCleanup(0xc0004dcd80, 0x0, 0x0, 0x0)
	/opt/hostedtoolcache/go/1.15.2/x64/src/testing/testing.go:949 +0xcb
testing.tRunner.func2(0xc0004dcd80)
	/opt/hostedtoolcache/go/1.15.2/x64/src/testing/testing.go:1121 +0x75
testing.tRunner(0xc0004dcd80, 0x19508d0)
[...]

PR #1379, opened 2021-02-01
Allow goroutines to run through end of branch locker test

arielshaqed:
All `testing.Fa*` routines are dangerous because they leave goroutines running but call
`t.Finish()` routines.  That might deadlock the finish routine against a goroutine.  So
instead error rather than fatal out of the `committer_wait_for_writers` test.  Once the
test flakes out again, we may have more information.

Also emit `ErrAlreadyLocked` (which wraps `ErrLockNotAcquired`) if locking branch fails
because of existing lock.

Relevant to #1373 but does not fix it.

### DIFF
--- a/graveler/errors.go
+++ b/graveler/errors.go
@@ -37,6 +37,7 @@ var (
 	ErrDirtyBranch             = errors.New("can't apply meta-range on dirty branch")
 	ErrMetaRangeNotFound       = errors.New("metarange not found")
 	ErrLockNotAcquired         = errors.New("lock not acquired")
+	ErrAlreadyLocked           = wrapError(ErrLockNotAcquired, "already locked")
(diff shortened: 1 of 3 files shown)


nopcoder (approved):

arielshaqed:
Thanks!

Re-ran the Go tests (maybe I'll get lucky and get a failure...), pulling when that's done.

nopcoder:
wonder if adding some t.Log to get a better understanding where it fails will effects the test

arielshaqed:
> wonder if adding some t.Log to get a better understanding where it fails will effects the test

You're waaaay ahead of me here.  The first thing I hope to get from this is _the actual error message_ for failing.  Which will include the failing line.

nopcoder:
also we can call defer on the channel we create so if Fatal is called the other goroutine will be free

PR #1381, opened 2021-02-01
Try to generate some coverage for real with Nessie

arielshaqed:
`go test` args should be sooner rather than later.  And might as well upload and merge _all_
Nessie tests -- interesting to see block adapter coverage, too!

### DIFF
--- a/.github/workflows/nessie.yaml
+++ b/.github/workflows/nessie.yaml
@@ -86,13 +86,13 @@ jobs:
           ECR_REPOSITORY_LAKEFS: lakefs
         run: |
           set +e
-          describe_image="$( aws ecr describe-images --repository-name $ECR_REPOSITORY_LAKEFS --image-ids imageTag=${{ steps.version.outputs.tag }})"
+          describe_image="$( aws ecr describe-images --repository-name $ECR_REPOSITORY_LAKEFS --image-ids imageTag=${{ steps.version.outputs.tag }}-cover)"
           if [ $? -eq 0 ]; then
             echo "Image exists"
           else
             echo "Image doesn't exist"
-            docker build -t $ECR_REGISTRY/$ECR_REPOSITORY_LAKEFS:${{ steps.version.outputs.tag }} --build-arg VERSION=${{ steps.version.outputs.tag }} .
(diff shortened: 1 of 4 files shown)


nopcoder (approved):

.github/workflows/nessie.yaml
```diff
@@ -145,7 +145,7 @@ jobs:
           NESSIE_STORAGE_NAMESPACE: s3://nessie-system-testing/${{ github.run_number }}-${{ matrix.cataloger }}
           NESSIE_AWS_ACCESS_KEY_ID: ${{ secrets.NESSIE_AWS_ACCESS_KEY_ID }}
           NESSIE_AWS_SECRET_ACCESS_KEY: ${{ secrets.NESSIE_AWS_SECRET_ACCESS_KEY }}
-        run: go test -v ./nessie --system-tests
+        run: go test -v --coverprofile=nessie-s3-cover.out --cover ./nessie --system-tests
```
itaiad200, line 148: I am really confused about this.
You run Nessie with coverprofile, but you're really interested in the coverage of the lakeFS app running inside a container. What would covering Nessie's code give you?

itaiad200 (changes requested):
Not sure about the coverage subject.

.github/workflows/nessie.yaml
arielshaqed: Yikes, true!

Dockerfile
```diff
@@ -45,7 +48,7 @@ RUN apk add netcat-openbsd
 WORKDIR /app
 COPY ./wait-for ./
 ENV PATH /app:$PATH
-COPY --from=build /build/lakefs /build/lakectl ./
+COPY --from=build /build/lakefs /build/lakefs.cover /build/lakectl ./
```
itaiad200, line 51: Are we sure that we want our official docker image to contain another binary which sole purpose is to be run in Nessie?
I would much rather we build a dedicated image for Nessie that will contain just the cover instead of adding it to our published image. @nopcoder WDYT?

itaiad200:
Closing due to inactivity. @arielshaqed reopen if it's needed.

PR #1391, opened 2021-02-01
Free the lock before adding to channel

### LINKED ISSUE #1390: Deadlock in batch-writer-closer
New closer gets the mutex and writes to channel.

When this channel gets full, it still holds the mutex. When all other threads are trying to get the mutex here, they can’t because the first go routine has the lock.

itaiad200:
closes #1390


arielshaqed (approved):
Ouch, nasty deadlock; good catch!
Can you also create a test which would illustrate the problem this addresses?  Not needed in this PR, can open an issue.

itaiad200:
Thanks! Opened #1393 for hardening the tests.
Pulling.

PR #1392, opened 2021-02-01
Batch closer can wait merely on the closer goroutines

arielshaqed:
Simplifies using a waitgroup, which is now a regular blocker usage (Add only called at
startup) rather than dynamically increasing and decreasing.  This way does not have to
synchronize as closely across the waitgroup.


itaiad200 (approved):
Now I see - Yes it is better!

arielshaqed:
Cool, thanks!
Rebasing, fixing conflicts, and pushing.

PR #1396, opened 2021-02-02
Test batch closer with parallelism

### LINKED ISSUE #1393: Stretch-test batch closer for deadlocks
Bug #1390 identified a deadlock we have in batch close code.
We do have some concurrency testing for that code, but obviously needs hardening.

itaiad200:
closes #1393

### DIFF
--- a/graveler/committed/batch_test.go
+++ b/graveler/committed/batch_test.go
@@ -92,7 +92,7 @@ func NewFakeRangeWriter(result *committed.WriteResult, err error) *FakeRangeWrit
 }
 
 func TestBatchCloserSuccess(t *testing.T) {
-	runSuccessScenario(t)
+	runSuccessScenario(t, 502, 5)
 }
 
 func TestBatchWriterFailed(t *testing.T) {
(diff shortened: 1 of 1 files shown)


graveler/committed/batch_test.go
```diff
@@ -126,18 +126,19 @@ func TestBatchCloserMultipleWaitCalls(t *testing.T) {
...
 	res, err := sut.Wait()
 	require.Nil(t, res)
 	require.Error(t, err, committed.ErrMultipleWaitCalls)
 }
 
-func runSuccessScenario(t *testing.T) *committed.BatchCloser {
-	const writersCount = 10
-	writers := make([]*FakeRangeWriter, writersCount)
-	for i := 0; i < writersCount; i++ {
+func runSuccessScenario(t *testing.T, writersAmount, closersAmount int) *committed.BatchCloser {
```
arielshaqed, line 137:
```suggestion
func runSuccessScenario(t *testing.T, numWriters, numClosers int) *committed.BatchCloser {
```
Rationale: "amount" is for uncountable quantities, these parameters specifically count.

graveler/committed/batch_test.go
```diff
@@ -92,7 +92,7 @@ func NewFakeRangeWriter(result *committed.WriteResult, err error) *FakeRangeWrit
 }
 
 func TestBatchCloserSuccess(t *testing.T) {
-	runSuccessScenario(t)
+	runSuccessScenario(t, 500, 5)
```
arielshaqed, line 95:
```suggestion
	runSuccessScenario(t, 502, 5)
```
Not a fan of even divisors when I can avoid them.  (I see no harm in having an even divisor, I just prefer not to have to think about whether it weakens the test.)

arielshaqed (approved):
Great, thanks!

All comments minor and non-blocking whatever you do decide.

Issue #1356, opened 2021-01-28
UI basic auth triggers browser login


nopcoder:
Web-users that browse to a non-login with an unauthorized key should be directed to the login page and non be prompt by the default browser user/password.

![image](https://user-images.githubusercontent.com/8110/106165675-01e87e00-6194-11eb-9778-323692372065.png)

nopcoder:
The dialog will pop up but after #1397 it will redirect to the login page.

PR #1397, opened 2021-02-02
Revert login local storage cache with minor improvement

nopcoder:
Related to #1356

- revert local storage change
- browser basic auth dialog will pop, but after cancel it will redirect to the application login

### DIFF
--- a/webui/src/actions/api.js
+++ b/webui/src/actions/api.js
@@ -30,15 +30,23 @@ export const extractError = async (response) => {
     return body;
 };
 
-const apiRequest = async (uri, requestData = {}, additionalHeaders = {}, defaultHeaders ={"Accept": "application/json",
-    "Content-Type": "application/json",}) => {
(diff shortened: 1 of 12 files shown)


webui/src/actions/api.js
```diff
@@ -30,15 +30,23 @@ export const extractError = async (response) => {
...
         headers: new Headers({
             ...defaultHeaders,
             ...additionalHeaders,
         }),
         ...requestData,
     });
+    if (response.status === 401) {
+        // unauthorized status code will throw the current user and reload
+        localStorage.removeItem('user');
+        window.location.reload(true);
```
johnnyaug, line 47: will it work if we call auth.logout instead of removing localStorage and reloading?

johnnyaug (approved):

webui/src/actions/api.js
nopcoder: didn't want to import store/ at this point.

johnnyaug (approved):

PR #1403, opened 2021-02-04
Send facts when installation occurs during Run

itaiad200:


api/controller.go
```diff
@@ -241,6 +241,7 @@ func (c *Controller) SetupLakeFSHandler() setupop.SetupLakeFSHandler {
 			c.deps.Logger.Error("failed to write metadata after setup")
 		} else {
 			c.deps.Collector.SetInstallationID(metadata[auth.InstallationIDKeyName])
+			c.deps.Collector.CollectMetadata(convertMetadata(metadata))
```
johnnyaug, line 244: I'd like to suggest something else, instead of creating the convertMetadata function: call `stats.NewMetadata` to generate the metadata. cloudMetadataProvider can be nil.
One caveat - it currently accepts a `config.Config` object which you don't have. The config is only used to get the blockstore type for the metadata. Can be solved in different ways, e.g. by changing the parameter in `NewMetadata` from being the config to just the blockstoreType, and here passing nil.

Feel free to reject this comment.

johnnyaug (approved):

api/controller.go
itaiad200: I feel the stats need more work than that. There really shouldn't be more than 1 function call when you want to collect metadata, and there's no reason why we only collect some of the information (for example here, no blockstore info). I'm opening an issue so we can track this.

I didn't follow your suggestion since I'm not really sure what would be the consequences of sending `blockStorageType: ""` here, or not sending it when it should appear.

PR #1407, opened 2021-02-04
Graveler pre-commit/merge hooks

### LINKED ISSUE #1431: Graveler to trigger hooks during commit/merge

nopcoder:
closes #1431 
- Extend Graveler pre-commit and pre-merge hooks - calling a list of callbacks that on error will stop the commit/merge
- Model 'Action' document. An Action includes event and branch spec on which a list of hooks will be executed. A successful run of all hooks will enable the flow of commit/merge.


graveler/graveler.go
```diff
@@ -300,6 +311,10 @@ type VersionController interface {
 	// Compare returns the difference between the commit where 'to' was last synced into 'from', and the most recent commit of `from`.
 	// This is similar to a three-dot (from...to) diff in git.
 	Compare(ctx context.Context, repositoryID RepositoryID, from, to Ref) (DiffIterator, error)
+
+	// Hooks are used to register callbacks on specific actions (pre commit/merge)
+	// When a callback returns an error, the action will fail
+	Hooks() *Hooks
```
ozkatz, line 317: I'd accept a context here for logging and cancellation as well as key this by a repository ID as hooks belong to a specific repo

graveler/graveler.go
```diff
@@ -866,6 +882,23 @@ func (g *Graveler) Commit(ctx context.Context, repositoryID RepositoryID, branch
...
+			CreationDate: time.Now(),
+			Metadata:     params.Metadata,
+		}
+		if branch.CommitID != "" {
+			commit.Parents = CommitParents{branch.CommitID}
+		}
+
+		err = g.callPreCommitHooks(ctx, repositoryID, branchID, commit)
+		if err != nil {
+			return "", fmt.Errorf("pre-commit hooks: %w", err)
```
ozkatz, line 899: I think we should have an `ErrAbortedByHook` (bad name, but you get the idea) that will be propagated when a commit or merge operation fail. This will allow the API to return an informative error (possibly with a link to the failing hook for details, etc)

graveler/graveler.go
```diff
@@ -1327,3 +1358,23 @@ func (g *Graveler) getCommitsForMerge(ctx context.Context, repositoryID Reposito
 	}
 	return fromCommit, toCommit, baseCommit, nil
 }
+
+func (g *Graveler) callPreCommitHooks(ctx context.Context, repositoryID RepositoryID, branchID BranchID, commit Commit) error {
+	for _, hook := range g.hooks.PreCommit {
```
ozkatz, line 1363: assuming each hook is an independent unit that doesn't depend on previous hooks, not sure we want to run them serially.. I'm usually opposed to doing concurrency where it's not required, but I'm not sure it's not required :) we already provide a pretty narrow timeframe for hooks to complete in - this makes the situation worse by having all hooks compete for a slice of that time..

ozkatz (changes requested):
I like how this fits pretty cleanly into Graveler! A testiment to the design. Overall looks good, proper error propagation and parallel execution are required imo. 

Also, not sure if a part of the PR - auditing of hook events (including them waiting to execute or are currently executing): will this be external to Graveler so in a different PR?

graveler/graveler.go
nopcoder: Graveler supplies the basic callback functionality we had in the cataloger.
There is no execution here, just getter.

graveler/graveler.go
nopcoder: True, as EntryCatalog will implement the actual load/parse/evaluate/execute of the actions it will use a specific error code with the execute/run ID to identify the specific call.

catalog/webhooks.go
```diff
@@ -0,0 +1,23 @@
+package catalog
```
itaiad200: Maybe pass it to a dedicated package and avoid 'Hook' prefix everywhere?

graveler/graveler.go
nopcoder: At graveler level, I don't have a problem running the hook one by one and keep it simple at this point.
Currently, we have only one consumer - entry catalog, and the execution of the webhooks will be in parallel.

graveler/graveler.go
```diff
@@ -1207,21 +1230,25 @@ func (g *Graveler) Merge(ctx context.Context, repositoryID RepositoryID, destina
...
 			CreationDate: time.Now(),
 			Parents:      []CommitID{fromCommit.CommitID, toCommit.CommitID},
 			Metadata:     commitParams.Metadata,
 		}
+		err = g.callPreMergeHooks(ctx, repositoryID, destination, fromCommit.CommitID.Ref(), commit)
+		if err != nil {
+			return "", fmt.Errorf("pre-merge hooks: %w", err)
+		}
+		var summary DiffSummary
+		commit.MetaRangeID, summary, err = g.CommittedManager.Merge(ctx, repo.StorageNamespace, toCommit.MetaRangeID, fromCommit.MetaRangeID, baseCommit.MetaRangeID)
```
itaiad200, line 1245: Sometimes merge fails due to conflict.

IMO it's a bad ux to run (what could be) a long running hook, let it pass, and then fail it immediately due to conflicts.

Is there a way to identify conflicts before calling the hook?

catalog/webhooks.go
nopcoder: Will consolidate into 3 types - I don't think we need all of them - just generated them.

graveler/graveler.go
nopcoder: sorry no - identify conflict will require going over the changes, which is a merge functionality.
if we hook implementation is consider small/fast at this point - I think it should stay between input parameter verification and the actual work.

itaiad200: Not blocking on this on this PR, but this is something we need to take care of.

Think about the experience here: A user can try to make the hooks pass over and over again, pushing new code to his branch, committing and trying to merge. Finally, after several iterations the hook passed, but then he finds out immediately that all his work was for nothing - there are conflicts with the destination branch, he should of been fixing them first but lakeFS returned errors in the wrong order.

nopcoder: UX point of view I agree, I would like any UI to tell me it is going to fail before I do any work.
It is something we need to discuss if moving the pre-merge step after the actual merge, but keep in mind that the pre-hook is there to verify the work that was done on the source ref, committed data - in the UI case can we can easily trigger 'diff' operation to identify conflict before submitting any merge.

graveler/graveler.go
nopcoder: Moving the handling to EntryLevel and keeping simple callback at graveler level.

graveler/graveler.go
nopcoder: Used multi error package to format the final error.
It can be used with errors.Is and ErrAbortedByHook or the hook's error if needed.

graveler/graveler.go
nopcoder: Moving the hook after the merge operation.

nopcoder:
> Also, not sure if a part of the PR - auditing of hook events (including them waiting to execute or are currently executing): will this be external to Graveler so in a different PR?

Different PRs

catalog/action.go
```diff
@@ -0,0 +1,49 @@
...
+	Properties  map[string]string `yaml:"properties"`
+}
+
+var reHookID = regexp.MustCompile(`^[_a-zA-Z][_a-zA-Z0-9]{1,255}$`)
+
+func (a *Action) Validate() error {
+	if a.On.PreMerge == nil && a.On.PreCommit == nil {
+		return fmt.Errorf("%w 'on' is required", ErrInvalidAction)
+	}
+	ids := make(map[string]struct{})
```
itaiad200, line 35:
```suggestion
	ids := make(map[string]bool)
```
will give you a nicer usage:
```go
		if ids[hook.ID] {
			return fmt.Errorf("%w hook[%d] duplicate ID", ErrInvalidAction, i)
		}
```

catalog/action.go
```diff
@@ -0,0 +1,49 @@
...
+	for i, hook := range a.Hooks {
+		if !reHookID.MatchString(hook.ID) {
+			return fmt.Errorf("%w hook[%d] missing ID", ErrInvalidAction, i)
```
itaiad200, line 38:
```suggestion
			return fmt.Errorf("hook[%d] missing ID: %w", i, ErrInvalidAction)
```

catalog/action.go
```diff
@@ -0,0 +1,49 @@
...
+		}
+		if _, found := ids[hook.ID]; found {
+			return fmt.Errorf("%w hook[%d] duplicate ID", ErrInvalidAction, i)
```
itaiad200, line 41: It's nicer to present the hook.ID (which is ok at this point):
```suggestion
			return fmt.Errorf("hook[%s] duplicate ID: %w", hook.ID, ErrInvalidAction)
```

catalog/action.go
```diff
@@ -0,0 +1,49 @@
...
+		}
+		ids[hook.ID] = struct{}{}
+		if hook.Type != "webhook" {
+			return fmt.Errorf("%w hook[%d] unknown type", ErrInvalidAction, i)
```
itaiad200, line 45: Same here.

graveler/graveler.go
```diff
@@ -1327,3 +1372,36 @@ func (g *Graveler) getCommitsForMerge(ctx context.Context, repositoryID Reposito
...
 	return fromCommit, toCommit, baseCommit, nil
 }
+
+func newErrAbortedByHook(err error) error {
+	if err == nil {
+		return ErrAbortedByHook
+	}
+	merr := multierror.Append(ErrAbortedByHook, err)
+	merr.ErrorFormat = func(errs []error) string {
+		return fmt.Sprintf("%s: %s", errs[0], errs[1])
```
itaiad200, line 1382: This works for almost everything, unless original error was of type `*multierror.Error`

graveler/graveler.go
```diff
@@ -1309,6 +1338,22 @@ func (g *Graveler) Compare(ctx context.Context, repositoryID RepositoryID, from,
 	return g.CommittedManager.Compare(ctx, repo.StorageNamespace, toCommit.MetaRangeID, fromCommit.MetaRangeID, baseCommit.MetaRangeID)
 }
 
+func (g *Graveler) PreCommitHook() PreCommitFunc {
```
itaiad200, line 1341: Why do we need the getters?

graveler/graveler_test.go
```diff
@@ -457,6 +457,205 @@ func TestGraveler_Commit(t *testing.T) {
...
+				Committer: commitCommitter,
+				Message:   commitMessage,
+				Metadata:  commitMetadata,
+			})
+			// check err composition
+			if !errors.Is(err, tt.err) {
+				t.Fatalf("Commit err=%v, expected=%v", err, tt.err)
+			}
+			if err != nil && !errors.Is(err, graveler.ErrAbortedByHook) {
+				t.Fatalf("Commit err=%v, expected ErrAbortedByHookv", err)
```
itaiad200, line 533:
```suggestion
				t.Fatalf("Commit err=%v, expected ErrAbortedByHook", err)
```

itaiad200 (changes requested):

catalog/action.go
```diff
@@ -0,0 +1,49 @@
...
+	ID          string            `yaml:"id"`
+	Type        string            `yaml:"type"`
+	Description string            `yaml:"description"`
+	Properties  map[string]string `yaml:"properties"`
+}
+
+var reHookID = regexp.MustCompile(`^[_a-zA-Z][_a-zA-Z0-9]{1,255}$`)
+
+func (a *Action) Validate() error {
+	if a.On.PreMerge == nil && a.On.PreCommit == nil {
```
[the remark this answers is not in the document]
nopcoder: 1. The verification here is for the existence of the event type
2. Not part of the verification - but empty will allow you to trigger the action on all branches.

catalog/action.go
nopcoder: I prefer the zero size value

graveler/graveler.go
nopcoder: added just for the symmetry

itaiad200 (approved):
LGTM!

catalog/action.go
```diff
@@ -0,0 +1,49 @@
...
+}
+
+type ActionHook struct {
+	ID          string            `yaml:"id"`
+	Type        string            `yaml:"type"`
+	Description string            `yaml:"description"`
+	Properties  map[string]string `yaml:"properties"`
+}
+
+var reHookID = regexp.MustCompile(`^[_a-zA-Z][_a-zA-Z0-9]{1,255}$`)
```
ozkatz, line 29: why not `-`?

ozkatz (approved):
Looks great!

Issue #1372, opened 2021-01-31
UI delete repository operation


nopcoder:
Enable delete repository action from the UI

shamikakumar:
Hey, can I work on this issue ?

itaiad200:
All yours @shamikakumar , let us know if you need anything from us.

PR #1410, opened 2021-02-06
Fix/delete repo #1372

shamikakumar:
Implemented #1372


webui/src/components/RepositoryListPage.js
```diff
@@ -29,7 +30,28 @@ const CreateRepositoryModal = ({show, error, onSubmit, onCancel}) => {
...
 
-const RepositoryList = ({ list, paginate }) => {
+const RepositoryList = connect(
+    ({ repositories }) => ({ deleteStatus: repositories.delete }),
+    ({ deleteRepository, listRepositories })
+)(({ list, paginate, deleteStatus, deleteRepository, listRepositories}) => {
+
+    const [showingDeleteModal, setShowDeleteModal] = useState(false);
+    const closeDeleteModal = () => setShowDeleteModal(false);
+    const showDeleteModal = () => setShowDeleteModal(true);
```
nopcoder, line 40: can you call `setShowDeleteModal` with true/false when needed, I think it will be clear enough and don't want to get the function population going up because of defaults.

webui/src/components/RepositoryListPage.js
```diff
@@ -29,7 +30,28 @@ const CreateRepositoryModal = ({show, error, onSubmit, onCancel}) => {
...
+    const [selectedRepo, setSelectedRepo] = useState("");
+
+    const deleteRepo = () => {
+        if(deleteStatus.inProgress)
+            return;
```
nopcoder, line 45:
```suggestion
        if (deleteStatus.inProgress) {
            return;
        }
```

nopcoder (approved):
Looks good, thanks! just minor code style changes.

webui/src/components/RepositoryListPage.js
```diff
@@ -49,29 +70,37 @@ const RepositoryList = ({ list, paginate }) => {
...
+                                        </small>
+                                    </p>
+                                </Card.Body>
+                            </Card>
+                        </Col>
+                    </Row>
+                ))}
+                {paginator}
+            </div>
+            <ConfirmationModal show={showingDeleteModal} onHide={() => {setShowDeleteModal(false)}} msg={`are you sure you wish to delete repository ${selectedRepo}?`} onConfirm={deleteRepo}/>
```
ozkatz, line 100: Since deleting a repository is probably the most destructive behavior we have in the system (and the one hardest to recover from), perhaps noting *what actually gets deleted* is required here. Example text:

```
Are you sure you wish to delete repository ${selectedRepo}?
This action cannot be undone. This will delete the following:

- All commits
- All branches
- All tags
- All repository configuration

Data in the underlying object store will not be deleted by this action. 

Please type ${selectedRepo} to confirm:
+------------------------------------------+
|                                          |
+------------------------------------------+

```

ozkatz (changes requested):

webui/src/components/RepositoryListPage.js
ozkatz: Screenshot of what deleteing a repo on GitHub looks like for further reference:

image

shamikakumar:
Hey @ozkatz, I've made the changes you requested, do review and let me know

ozkatz (changes requested):
It works! and it's very nice!

image

However, looking at the user experience, I feel that this is a very uncommon action that is given too much emphasis in the UI as it stands. 

How do you feel about moving the button (with same confirmation modal) from the repository list view to the settings tab of the repository itself? This is what GitHub/BitBucket do as well and I think it makes sense.

shamikakumar:
I agree @ozkatz , working on the changes

shamikakumar:
hey! I'm not sure how to redirect to the /repositories page after the repo delete action, I could use some help here please and thank you

nopcoder:
> hey! I'm not sure how to redirect to the /repositories page after the repo delete action, I could use some help here please and thank you

Sure, it should be similar to create a repository, it moves to the specific repository after the operation completes.

So delete repository will probably look something like this:

```go
    useEffect(()=> {
        if (deleteStatus.done) {
            history.push('/repositories');
        }
        listRepositories();
    }, [listRepositories, deleteStatus.done, history]);
```

Thanks!

shamikakumar:
@nopcoder I'm getting this error after I submit delete repository. Looks like it isn't redirecting. Am I missing something? ->

image

nopcoder:
The issue is related a reset of the state we need to perform, which is missing here and in the create repository page.

In `actions/repositories.js` you will need to add `deleteRepositoryDone` (or other name) that will reset the `REPOSITORY_DELETE`.
Import it and call when you push the `/repositories` into the the `history`.
Additionaly we need to add a call to `createRepositoryDone` in `components/RepositoryListPage.js` when create.done.

The problem was that after the action completed rendering the list repositories kept the previous state and viewed the repository you deleted as if it is just created.
For example if you had full refresh and than go and delete your repository - it would have seen that all is working.
Now when the state changes to `done` we reset the action so for delete and create repository the state reset after successful operation.

nopcoder (approved):
Great, thanks @shamikakumar

ozkatz (approved):
Thank you @shamikakumar !

PR #1412, opened 2021-02-07
Wip/scala client

johnnyaug:


clients/src/main/scala/io/treeverse/clients/s3/S3MetaRangeReader.scala
```diff
@@ -0,0 +1,22 @@
+package io.treeverse.clients.s3
+
+import com.amazonaws.services.s3.AmazonS3
+import com.amazonaws.services.s3.model.GetObjectRequest
+import io.treeverse.clients
+import io.treeverse.clients.{MetaRangeReader, SSTableReader}
+
+import java.io.File
+
+class S3MetaRangeReader(var s3: AmazonS3, var bucket: String, var repositoryPrefix: String) extends MetaRangeReader {
```
ozkatz, line 10: I don't know what's the final purpose of this client, but it *probably* shouldn't be calling S3 directly (i.e. if it's a Hadoop embeddable library, you should be coordinating IO through the HadoopFileSystem

clients/src/main/scala/io/treeverse/clients/VarInt.scala
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * from https://github.com/addthis/stream-lib/blob/master/src/main/java/com/clearspring/analytics/util/Varint.java
+ */
+object VarInt {
+  @throws[IOException]
+  def readSignedVarLong(in: DataInput): Long = {
+    val raw = readUnsignedVarLong(in)
+    // This undoes the trick in writeSignedVarLong()
+    val temp = (((raw << 63) >> 63) ^ raw) >> 1
```
ozkatz, line 15: love it.

clients/.gitignore
```diff
@@ -0,0 +1 @@
+target
```
ozkatz, line 1: I feel strongly that the Scala client should live in another repository and not in the main lakeFS one.

ozkatz (changes requested):

clients/src/main/scala/io/treeverse/clients/VarInt.scala
arielshaqed: A classic.  Only needed in JVMs and similar "unsigned is too difficult".  Zigzag usually looks something like this (e.g. in Go):
```go
	ux := uint64(x) << 1
	if x < 0 {
		ux = ^ux
	}
```
which has the advantage of not having to pick between `>>` and `>>>` while remaining branch-free...

clients/.gitignore
arielshaqed: I somewhat disagree: I have found it difficult to see the advantages of multi-repo.  But in any case a prerequisite for going multi-repo on our clients is to be able to publish our protobuf files to _somewhere_.  Saying "git submodules" does not tend to make friends...

arielshaqed (changes requested):
Thanks!

A mixture of nits and important comments.  These are the important ones:
1. In JVM, strings are not the same as byte sequences.
2. Possible late dynamic loading; possibly unnecessary restriction of SSTable format to Snappy compression.
3. Exceptions should be typed.
4. Avoid mixing operations on 2 overloaded streams (if possible).

clients/project/protoc.sbt
```diff
@@ -0,0 +1,3 @@
+addSbtPlugin("com.thesamet" % "sbt-protoc" % "1.0.0")
+
+libraryDependencies += "com.thesamet.scalapb" %% "compilerplugin" % "0.10.10"
```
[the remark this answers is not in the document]
johnnyaug: No, here is a dependency for the compile time, while the other one if for runtime

johnnyaug:
Moved to a separate repository

PR #1418, opened 2021-02-08
Add zero commit to create repo by default

### LINKED ISSUE #1417: Commit zero when repository created
lakeFS by default should create an empty commit when repository is created.

itaiad200:
close #1417

### DIFF
--- a/graveler/ref/manager.go
+++ b/graveler/ref/manager.go
@@ -3,6 +3,7 @@ package ref
 import (
 	"context"
 	"errors"
+	"time"
 
 	"github.com/treeverse/lakefs/db"
 	"github.com/treeverse/lakefs/graveler"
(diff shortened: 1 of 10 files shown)


graveler/ref/manager.go
```diff
@@ -41,7 +42,17 @@ func (m *Manager) GetRepository(ctx context.Context, repositoryID graveler.Repos
 	return repository.(*graveler.Repository), nil
 }
 
-func (m *Manager) CreateRepository(ctx context.Context, repositoryID graveler.RepositoryID, repository graveler.Repository, branch graveler.Branch) error {
+const (
+	firstCommitMsg = "lakeFS repository created!"
```
nopcoder, line 46: 1. Public under graveler so we can verify when testing.
2. Consider "Initial repository created" or something without lakeFS.

graveler/ref/manager.go
```diff
@@ -41,7 +42,17 @@ func (m *Manager) GetRepository(ctx context.Context, repositoryID graveler.Repos
...
+)
+
+func (m *Manager) CreateRepository(ctx context.Context, repositoryID graveler.RepositoryID, repository graveler.Repository, token graveler.StagingToken) error {
+	firstCommit := graveler.Commit{
+		Message:      firstCommitMsg,
+		CreationDate: time.Now(),
```
nopcoder, line 52: verity if you need to call UTC or the pg lib does that for you

graveler/ref/manager.go
itaiad200: Done. "Initial" for repository is kinda weird, so kept it simple.

graveler/ref/manager.go
itaiad200: It is done by the `addCommit` method below.

nopcoder (approved):

PR #1422, opened 2021-02-08
Import should use repo default branch

### LINKED ISSUE #1411: import doesn't work when default branch isn't master
When it's the first import, the import branch is created from "master" instead if the user-chosen branch.

itaiad200:
close #1411

### DIFF
--- a/onboard/catalog_actions.go
+++ b/onboard/catalog_actions.go
@@ -57,8 +57,8 @@ type entryCataloger interface {
 func NewCatalogRepoActions(config *Config, logger logging.Logger) *CatalogRepoActions {
 	return &CatalogRepoActions{
 		entryCataloger:  config.EntryCatalog,
-		repoID:          config.RepoID,
-		defaultBranchID: graveler.BranchID(config.Repo.DefaultBranch),
+		repoID:          config.RepositoryID,
+		defaultBranchID: config.DefaultBranchID,
 		committer:       config.CommitUsername,
 		logger:          logger,
 		prefixes:        config.KeyPrefixes,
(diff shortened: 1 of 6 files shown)


onboard/import.go
```diff
@@ -26,12 +29,13 @@ type Importer struct {
 type Config struct {
 	CommitUsername     string
 	InventoryURL       string
-	Repository         string
+	RepoID             graveler.RepositoryID
```
nopcoder, line 32:
```suggestion
	RepositoryID             graveler.RepositoryID
```

onboard/import.go
```diff
@@ -26,12 +29,13 @@ type Importer struct {
...
+	Repo               *catalog.Repository
```
nopcoder, line 33: Don't think this one should be part of the configuration if we already passing the RepositoryID.
We need the full Repository information because we do not have access to the entry catalog - ask just Reposoty without passing the repository ID.

onboard/catalog_actions.go
```diff
@@ -112,18 +114,18 @@ func (c *CatalogRepoActions) Init(ctx context.Context, baseCommit graveler.Commi
 }
 
 func (c *CatalogRepoActions) initBranch(ctx context.Context) error {
-	c.branchID = catalog.DefaultImportBranchName
-	branch, err := c.entryCataloger.GetBranch(ctx, c.repoID, catalog.DefaultImportBranchName)
+	c.branchID = DefaultImportBranchName
+	branch, err := c.entryCataloger.GetBranch(ctx, c.repoID, DefaultImportBranchName)
```
nopcoder, line 118: When we get the branch and we fail, not with 'not found' we need to return the error - like we are doing when we try to create the branch.

nopcoder (changes requested):

onboard/catalog_actions.go
itaiad200: Right, done.

onboard/import.go
itaiad200: Done.

onboard/import.go
itaiad200: Passed just what's mandatory - default branch id

nopcoder (approved):

PR #1430, opened 2021-02-09
Collect block storage type during setup (when started using run)

### LINKED ISSUE #1429: Some installations stats are sent without block storage metadata
Installations metadata that are collected during `lakefs run` and not `lakefs init` don't have the block storage stats.

itaiad200:
closes #1429

### DIFF
--- a/api/controller.go
+++ b/api/controller.go
@@ -32,6 +32,7 @@ import (
 	"github.com/treeverse/lakefs/auth/model"
 	"github.com/treeverse/lakefs/block"
 	"github.com/treeverse/lakefs/catalog"
+	"github.com/treeverse/lakefs/cloud"
 	"github.com/treeverse/lakefs/db"
 	"github.com/treeverse/lakefs/graveler"
 	"github.com/treeverse/lakefs/httputil"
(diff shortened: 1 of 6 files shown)


api/controller.go
```diff
@@ -10,6 +10,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/treeverse/lakefs/cloud"
+
```
johnnyaug, line 14: as per usual

api/controller.go
```diff
@@ -51,14 +53,15 @@ const (
...
-	Logger          logging.Logger
+	ctx                   context.Context
+	Cataloger             catalog.Cataloger
+	Auth                  auth.Service
+	BlockAdapter          block.Adapter
+	MetadataManager       auth.MetadataManager
+	Migrator              db.Migrator
+	Collector             stats.Collector
+	Logger                logging.Logger
+	CloudMetadataProvider cloud.MetadataProvider
```
johnnyaug, line 64: I'd keep the logger last

api/controller_test.go
```diff
@@ -11,6 +11,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/treeverse/lakefs/stats"
+
```
johnnyaug, line 15: dd

johnnyaug (approved):

PR #1436, opened 2021-02-09
upgrade immer >=8.0.1

nopcoder:
Address CVE-2020-28477

### DIFF
--- a/webui/package.json
+++ b/webui/package.json
@@ -21,7 +21,8 @@
     "react-router-dom": "^5.2.0",
     "react-scripts": "^4.0.1",
     "redux": "^4.0.5",
-    "redux-thunk": "^2.3.0"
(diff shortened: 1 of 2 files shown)


webui/package.json
```diff
@@ -21,7 +21,8 @@
     "react-router-dom": "^5.2.0",
     "react-scripts": "^4.0.1",
     "redux": "^4.0.5",
-    "redux-thunk": "^2.3.0"
+    "redux-thunk": "^2.3.0",
+    "immer": ">=8.0.1"
```
ozkatz, line 25: don't we need react-scripts that depends on this to do this?

nopcoder: followed the recommendation, but I also prefer to upgrade react and not dependencies - closing this one. thanks.

PR #1437, opened 2021-02-09
Delete benchmarks workflow since it's not working

itaiad200:


arielshaqed (approved):
Thanks!  Please open an issue to fix it, though.

Pulling to save $$$.

PR #1441, opened 2021-02-09
Feature/refs restore

ozkatz:
Now that we can `refs-dump`, we can also `refs-restore` into a bare repo!
Additionally, we can now create bare repos.


api/client.go
```diff
@@ -89,6 +90,7 @@ type RepositoryClient interface {
 	Symlink(ctx context.Context, repoID, ref, path string) (string, error)
 
 	RefsDump(ctx context.Context, repository string) (*models.RefsDump, error)
+	RefsRestore(ctx context.Context, repository string, manifest *models.RefsDump) error
```
arielshaqed, line 93: Getting ever more worried about the size of `manifest`.  We should consider (*not in this PR!*) exporting it and passing around a reference.

api/controller.go
```diff
@@ -167,6 +167,7 @@ func (c *Controller) Configure(api *operations.LakefsAPI) {
 	api.ConfigGetConfigHandler = c.ConfigGetConfigHandler()
 
 	api.RefsRefsDumpHandler = c.RefsRefsDumpHandler()
+	api.RefsRefsRestoreHandler = c.RefsRefsRestoreHandler()
```
arielshaqed, line 170: These names are somewhat funny: by default I would expect "refs dump" to dump refs, not sure why it needs to be "refs refs dump".

api/controller.go
```diff
@@ -2372,6 +2396,68 @@ func (c *Controller) DetachPolicyFromGroupHandler() authop.DetachPolicyFromGroup
...
+				Resource: permissions.RepoArn(params.Repository),
+			},
+			{
+				Action:   permissions.ListBranchesAction,
+				Resource: permissions.RepoArn(params.Repository),
+			},
+			{
+				Action:   permissions.ListCommitsAction,
+				Resource: permissions.RepoArn(params.Repository),
+			},
```
arielshaqed, line 2413: I think permission actions are wrong here: this is a modifying operation, I think we should test some action that says "Write".

api/controller.go
```diff
@@ -2372,6 +2396,68 @@ func (c *Controller) DetachPolicyFromGroupHandler() authop.DetachPolicyFromGroup
...
+				WithPayload(responseErrorFrom(err))
+		} else if err != nil {
+			return refs.NewRefsRestoreDefault(http.StatusInternalServerError).
+				WithPayload(responseErrorFrom(err))
+		}
+
+		// ensure no refs currently found
+		_, _, err = deps.Cataloger.ListCommits(deps.ctx, repo.Name, repo.DefaultBranch, "", 1)
+		if !errors.Is(err, graveler.ErrNotFound) {
+			return refs.NewRefsRestoreDefault(http.StatusInternalServerError).
```
arielshaqed, line 2433: Consider reporting more specific (non-internal) errors for business logic failures.

api/controller.go
```diff
@@ -2372,6 +2396,68 @@ func (c *Controller) DetachPolicyFromGroupHandler() authop.DetachPolicyFromGroup
...
+				WithPayload(responseErrorFrom(err))
+		}
+
+		err = deps.Cataloger.LoadTags(deps.ctx, repo.Name, params.Manifest.TagsMetaRangeID)
+		if err != nil {
+			return refs.NewRefsRestoreDefault(http.StatusInternalServerError).
+				WithPayload(responseErrorFrom(err))
+		}
+
```
arielshaqed, line 2455: If we failed before this time, we are in an inconsistent state.  I understand why, but I am not sure how the user could clean up the mess we leave in order to be able to try again.

catalog/cataloger.go
```diff
@@ -49,6 +49,9 @@ type Cataloger interface {
 	// CreateRepository create a new repository pointing to 'storageNamespace' (ex: s3://bucket1/repo) with default branch name 'branch'
 	CreateRepository(ctx context.Context, repository string, storageNamespace string, branch string) (*Repository, error)
 
+	// CreateBareRepository create a new repository pointing to 'storageNamespace' (ex: s3://bucket1/repo) with no initial branch or commit
+	CreateBareRepository(ctx context.Context, repository string, storageNamespace string, branch string) (*Repository, error)
```
arielshaqed, line 53: Not sure what parameter `branch` does, the godoc makes me think it shouldn't even be there.

catalog/entry_catalog.go
```diff
@@ -173,6 +174,16 @@ func (e *EntryCatalog) CreateRepository(ctx context.Context, repositoryID gravel
...
 }
 
+func (e *EntryCatalog) CreateBareRepository(ctx context.Context, repositoryID graveler.RepositoryID, storageNamespace graveler.StorageNamespace, branchID graveler.BranchID) (*graveler.Repository, error) {
+	if err := Validate([]ValidateArg{
+		{"repositoryID", repositoryID, ValidateRepositoryID},
+		{"storageNamespace", storageNamespace, ValidateStorageNamespace},
+	}); err != nil {
+		return nil, err
+	}
+	return e.Store.CreateBareRepository(ctx, repositoryID, storageNamespace, branchID)
```
arielshaqed, line 184: This is one of multiple copies of `CreateRepository` => `CreateBareRepository`.  It seems like a lot more work than passing another boolean parameter, and/or splitting `CreateRepository` into `CreateBareRepository` + `SomethingRepositoryDefaultBranchAndCommit` higher up the call tree.

catalog/rocks_cataloger.go
```diff
@@ -57,6 +57,24 @@ func (c *cataloger) CreateRepository(ctx context.Context, repository string, sto
 	return catalogRepo, nil
 }
 
+// CreateBareRepository create a new repository pointing to 'storageNamespace' (ex: s3://bucket1/repo) with no initial branch or commit
```
arielshaqed, line 60:
```suggestion
// CreateBareRepository creates a new repository pointing to 'storageNamespace' (ex: s3://bucket1/repo) with no initial branch or commit
```

graveler/graveler.go
```diff
@@ -1397,6 +1427,119 @@ func (g *Graveler) getCommitsForMerge(ctx context.Context, repositoryID Reposito
...
+	}
+	iter, err := g.CommittedManager.List(ctx, repo.StorageNamespace, metaRangeID)
+	if err != nil {
+		return err
+	}
+	defer iter.Close()
+	for iter.Next() {
+		rawValue := iter.Value()
+		if rawValue == nil {
+			break // better check for error
```
arielshaqed, line 1489: Not sure whether this is a TODO or an instruction; consider making it a clearer imperative, e.g. (if the comment I wrote is true...)
```suggestion
			break // report deserialization error
```

But then... what if the iterator returns `nil` but no error?  Need to test that this iterator cannot do this, otherwise we might successfully "import" a repo incompletely.

graveler/ref/manager.go
```diff
@@ -72,6 +72,19 @@ func (m *Manager) CreateRepository(ctx context.Context, repositoryID graveler.Re
 	return err
 }
 
+func (m *Manager) CreateBareRepository(ctx context.Context, repositoryID graveler.RepositoryID, repository graveler.Repository) error {
+	_, err := m.db.Transact(func(tx db.Tx) (interface{}, error) {
+		_, err := tx.Exec(
```
arielshaqed, line 77: This repeats code from `CreateRepository`.  That makes me want to keep it as a separate function even less.

arielshaqed (changes requested):
Great, thanks!

How was this tested?

Most missing is some way to get confidence that the entire repo was indeed processed: if some iterator stopped too soon we're in trouble.  Even trivial "checksums" such as reporting the number of commits (and/or other counts) on backup and restore should give users a lot more clarity and safety.

Most desired changes are the above, permissions, better error messaging, and recovering from inconsistency if restore fails.

api/controller.go
ozkatz: fixed

api/controller.go
ozkatz: fixed

api/controller.go
ozkatz: Since ref-loading only works on an initial bare repo, simplest thing to do would be to drop the repo and start over on failure.
All actions should in theory be idempotent, so it could be possible to allow the user to retry but I'm not sure it's worth it

api/controller.go
ozkatz: Actually you caught a bug, it was meant to be a `NewRestoreBadRequest` instead.

catalog/cataloger.go
ozkatz: gave it a better name.

catalog/entry_catalog.go
ozkatz: Yea, still not sure how to handle "plumbing" calls in our internal APIs. Open to concrete suggestions (i'm fine with all options you suggested above)

graveler/ref/manager.go
ozkatz: I guess it should be a composition.

api/client.go
ozkatz: not sure I understand. the manifest is a tiny json file externally, and a small 3 field struct when deserialized. I'm probably missing your meaning though.

ozkatz:
> Great, thanks!
> 
> How was this tested?
> 
> Most missing is some way to get confidence that the entire repo was indeed processed: if some iterator stopped too soon we're in trouble. Even trivial "checksums" such as reporting the number of commits (and/or other counts) on backup and restore should give users a lot more clarity and safety.
> 
> Most desired changes are the above, permissions, better error messaging, and recovering from inconsistency if restore fails.

Thanks Ariel.

I tested this manually with a couple of sizable repos. Probably not good enough and I think this warrants system tests - I'm not sure how cleanly I could fit this into Nessie's model though.

Regarding checksums: since ref-dumps returns meta range IDs, we already have checksums! Since I'm reading each metarange from start to end, I could theoretically compute each item's identity and construct a hash that should match the meta range ID I was passed, but the `committedManager` doesn't expose anything that allows doing that at the moment (i.e. `List` is not enough since I want to compare every iterated Range ID and eventually build a hash of all range IDs that yields the meta range hash, but I'd rather not completely break the committed manager's API just to do this). WDYT?

arielshaqed:
[...]
> I tested this manually with a couple of sizable repos. Probably not good enough and I think this warrants system tests - I'm not sure how cleanly I could fit this into Nessie's model though.

Some sizeable repos are enough, at least for now.  Could we (in a future PR!) test this in Nessie by creating a repo, dumping it, then creating a brand-new repo and importing into *that*?  Does Nessie insist on just the one repo?

> 
> Regarding checksums: since ref-dumps returns meta range IDs, we already have checksums! Since I'm reading each metarange from start to end, I could theoretically compute each item's identity and construct a hash that should match the meta range ID I was passed, but the `committedManager` doesn't expose anything that allows doing that at the moment (i.e. `List` is not enough since I want to compare every iterated Range ID and eventually build a hash of all range IDs that yields the meta range hash, but I'd rather not completely break the committed manager's API just to do this). WDYT?

I think we don't understand one another.  See [this comment](https://github.com/treeverse/lakeFS/pull/1441/files#r573727933) for what looks like a way for an incomplete import somehow to succeed -- the code there is not entirely consistent with our general iterator semantics, and seems to rely on some additional property to determine end of iteration.

api/controller.go
arielshaqed: Cool.  Maybe just make the error message explain this here, then.

arielshaqed (commented):
You seem to have merged master into this branch, which makes "changes from my last commit" harder to grok.  Will take me a bit longer to parse them, sorry.

docs/reference/commands.md
```diff
@@ -1402,6 +1402,32 @@ lakectl refs-dump <repository uri> [flags]
...
+aws s3 cp s3://bucket/_lakefs/refs_manifest.json - | lakectl refs-load lakefs://my-bare-repository --manifest -
+```
+
+#### Options
+
+```
+  -h, --help                 help for refs-restore
+      --manifest refs-dump   path to a refs manifest json file (as generated by refs-dump). Alternatively, use "-" to read from stdin
+```
+
```
arielshaqed, line 1428: Maybe document how (not) to recover from a bad restore here?

api/controller.go
```diff
@@ -2430,7 +2430,7 @@ func (c *Controller) RefsRestoreHandler() refs.RestoreHandler {
 		// ensure no refs currently found
 		_, _, err = deps.Cataloger.ListCommits(deps.ctx, repo.Name, repo.DefaultBranch, "", 1)
 		if !errors.Is(err, graveler.ErrNotFound) {
-			return refs.NewRestoreDefault(http.StatusInternalServerError).
+			return refs.NewRestoreBadRequest().
 				WithPayload(responseError("restoring refs is supported only for bare repositories"))
```
arielshaqed, line 2434:
```suggestion
 				WithPayload(responseError("can only restore into a bare repository"))
```

arielshaqed (approved):
*Neat!*  Even before we start talking about replication, the ability to restore is critical in an application for managing data.  Now we have the ability to restore *cleanly*, and we can add tests.

api/client.go
ozkatz: Yet another overloaded word. If you have a concrete suggestion to rename it, i'll be OK with that :)

arielshaqed (approved):
LEBTM!

PR #1444, opened 2021-02-10
Add Azure adapter

guy-har:


block/azure/chunkwriting.go
```diff
@@ -0,0 +1,273 @@
...
+	base64Etag := base64.StdEncoding.EncodeToString([]byte(etag))
+
+	// write to blockIDs
+	pd := strings.Join(cp.id.issued(), "\n") + "\n"
+	_, err = cp.toIDs.StageBlock(cp.ctx, base64Etag, strings.NewReader(pd), cp.o.AccessConditions.LeaseAccessConditions, nil, cp.o.ClientProvidedKeyOptions)
+	if err != nil {
+		return "", fmt.Errorf("failed staging part data: %w", err)
+	}
+	// write block sizes
+	sd := strconv.Itoa(int(cp.reader.CopiedSize)) + "\n"
```
johnnyaug, line 116: IIUC, the size file contains only a single line, that has the total size of the part.
If so, I would use the same files for IDs and size - e.g. with the first line representing the size.

block/azure/chunkwriting.go
```diff
@@ -0,0 +1,273 @@
...
+	_, err = cp.toSizes.StageBlock(cp.ctx, base64Etag, strings.NewReader(sd), cp.o.AccessConditions.LeaseAccessConditions, nil, cp.o.ClientProvidedKeyOptions)
```
johnnyaug, line 117: Suggestion:
Use a struct:
```go
type PartInfo struct {
  blockIDs []string `json:"block_ids"`
  size int `json:"size"`
}
```

Then, have just one file for all the blockInfos, and read it with one method and not two.

nopcoder (changes requested):
A partial review looking into the adapter.go now

block/azure/adapter.go
```diff
@@ -0,0 +1,505 @@
...
+	return containerURL.NewBlockBlobURL(fileName)
+}
+
+func (a *Adapter) getIDURL(containerName, fileName string) azblob.BlockBlobURL {
+	return a.getBlockBlobURL(containerName, fileName+idSuffix)
+}
+
+func (a *Adapter) getSizeURL(containerName, fileName string) azblob.BlockBlobURL {
+	return a.getBlockBlobURL(containerName, fileName+sizeSuffix)
+}
```
nopcoder, line 127: Think we just need one function that builds the container URL object and not a lot of helpers that will continue to call the base function that will parse the URL again and again

block/azure/adapter.go
```diff
@@ -0,0 +1,505 @@
...
+	_, err := blobURL.CommitBlockList(a.ctx, base64BlockIDs, azblob.BlobHTTPHeaders{}, azblob.Metadata{}, azblob.BlobAccessConditions{}, azblob.AccessTierNone, azblob.BlobTagsMap{}, azblob.ClientProvidedKeyOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	downloadResponse, err := blobURL.Download(a.ctx, 0, azblob.CountToEnd, azblob.BlobAccessConditions{}, false, azblob.ClientProvidedKeyOptions{})
+	if err != nil {
+		return nil, err
+	}
+	bodyStream := downloadResponse.Body(a.configurations.retryReaderOptions)
```
nopcoder, line 460: Close

nopcoder (changes requested):

block/azure/adapter.go
```diff
@@ -0,0 +1,505 @@
...
+	}
+
+	blobURL := a.getBlobURL(qualifiedKey.StorageNamespace, qualifiedKey.Key)
+	_, err = blobURL.GetProperties(a.ctx, azblob.BlobAccessConditions{}, azblob.ClientProvidedKeyOptions{})
+	if err != nil {
+		if err.(azblob.StorageError).ServiceCode() == azblob.ServiceCodeBlobNotFound {
+			return false, nil
+		}
+		return false, err
+	}
```
nopcoder, line 226: This one should probably do the same using Go's error package
```suggestion
	var storageErr azblob.StorageError
	if errors.As(err, &storageErr)  && storageErr.ServiceCode() == azblob.ServiceCodeBlobNotFound {
		return false, nil
	} else if err != nil {
		return false, err
	}
```

block/azure/adapter.go
```diff
@@ -0,0 +1,505 @@
...
+
+func (a *Adapter) CompleteMultiPartUpload(obj block.ObjectPointer, uploadID string, multipartList *block.MultipartUploadCompletion) (*string, int64, error) {
+	var err error
+	defer reportMetrics("CompleteMultiPartUpload", time.Now(), nil, &err)
+	qualifiedKey, err := resolveNamespace(obj)
+	if err != nil {
+		return nil, 0, err
+	}
+	containerName := qualifiedKey.StorageNamespace
+	_ = a.log().WithFields(logging.Fields{
```
nopcoder, line 414: need to store or print something

block/azure/adapter.go
```diff
@@ -0,0 +1,504 @@
...
+	"github.com/treeverse/lakefs/block"
+	"github.com/treeverse/lakefs/logging"
+)
+
+var (
+	ErrNotImplemented = errors.New("not implemented")
+)
+
+const (
+	BlockstoreType          = "wasb"
```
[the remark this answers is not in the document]
guy-har: Windows Azure Storage Blob

block/azure/adapter.go
```diff
@@ -0,0 +1,505 @@
...
+		"qualified_key": qualifiedKey.Key,
+		"key":           obj.Identifier,
+	})
+
+	parts := multipartList.Part
+	sort.Slice(parts, func(i, j int) bool {
+		return *parts[i].PartNumber < *parts[j].PartNumber
+	})
+	// extract staging blockIDs
+	metaBlockIDs := make([]string, 0)
```
nopcoder, line 426:
```suggestion
	metaBlockIDs := make([]string, 0, len(parts))
```

block/azure/adapter.go
```diff
@@ -0,0 +1,505 @@
...
+	for _, part := range multipartList.Part {
+		base64Etag := base64.StdEncoding.EncodeToString([]byte(*part.ETag))
+		metaBlockIDs = append(metaBlockIDs, base64Etag)
+	}
+
+	fileName := qualifiedKey.Key
+	stageBlockIDs, err := a.getMultipartIDs(containerName, fileName, metaBlockIDs)
```
nopcoder, line 433: check err or ignore

block/factory/build.go
```diff
@@ -108,3 +117,20 @@ func buildGSAdapter(params params.GS) (*gs.Adapter, error) {
...
+		accountName, accountKey = os.Getenv("AZURE_STORAGE_ACCOUNT"), os.Getenv("AZURE_STORAGE_ACCESS_KEY")
+	}
+	// Create a default request pipeline using your storage account name and account key.
+	credential, err := azblob.NewSharedKeyCredential(accountName, accountKey)
+	if err != nil {
+		return nil, fmt.Errorf("invalid credentials : %w", err)
+	}
+	p := azblob.NewPipeline(credential, azblob.PipelineOptions{Retry: azblob.RetryOptions{TryTimeout: params.TryTimeout}})
+
+	return azure.NewAdapter(p, accountName), nil
```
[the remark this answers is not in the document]
guy-har: pipeline is the azure client

credentials are used to initiate the pipeline, The adapter does need to hold the credentials, but does need the account name

block/azure/adapter.go
```diff
@@ -0,0 +1,505 @@
...
+	size, err := a.getMultipartSize(containerName, fileName, metaBlockIDs)
+
+	blobURL := a.getBlockBlobURL(containerName, fileName)
+
+	res, err := blobURL.CommitBlockList(a.ctx, stageBlockIDs, azblob.BlobHTTPHeaders{}, azblob.Metadata{}, azblob.BlobAccessConditions{}, azblob.AccessTierNone, azblob.BlobTagsMap{}, azblob.ClientProvidedKeyOptions{})
+	if err != nil {
+		return nil, 0, err
+	}
+	etag := string(res.ETag())
+	etag = etag[1 : len(etag)-2]
```
nopcoder, line 444:
```suggestion
	etag = strings.Trim(etag, "\"")
```

block/azure/adapter.go
nopcoder, line 426:
```suggestion
	metaBlockIDs := make([]string, 0, len(parts))
```

block/azure/adapter.go
```diff
@@ -0,0 +1,505 @@
...
+	ids := make([]string, 0)
+	for scanner.Scan() {
+		id := scanner.Text()
+		ids = append(ids, id)
+	}
+
+	// remove
+	_, err = blobURL.Delete(a.ctx, azblob.DeleteSnapshotsOptionNone, azblob.BlobAccessConditions{})
+	if err != nil {
+		a.log().WithError(err).Warn("Failed to delete multipart ids data file")
```
nopcoder, line 471: without containerName, fileName we have no idea which one failed

block/azure/adapter.go
```diff
@@ -0,0 +1,505 @@
...
+	_, err := sizeURL.CommitBlockList(a.ctx, base64BlockIDs, azblob.BlobHTTPHeaders{}, azblob.Metadata{}, azblob.BlobAccessConditions{}, "", azblob.BlobTagsMap{}, azblob.ClientProvidedKeyOptions{})
+	if err != nil {
+		return 0, err
+	}
+
+	downloadResponse, err := sizeURL.Download(a.ctx, 0, azblob.CountToEnd, azblob.BlobAccessConditions{}, false, azblob.ClientProvidedKeyOptions{})
+	if err != nil {
+		return 0, err
+	}
+	bodyStream := downloadResponse.Body(a.configurations.retryReaderOptions)
```
nopcoder, line 487: Close

block/azure/adapter.go
```diff
@@ -0,0 +1,505 @@
...
+		if err != nil {
+			return 0, err
+		}
+		size += stageSize
+	}
+
+	// remove
+	_, err = sizeURL.Delete(a.ctx, azblob.DeleteSnapshotsOptionNone, azblob.BlobAccessConditions{})
+	if err != nil {
+		a.log().WithError(err).Warn("Failed to delete multipart size data file")
```
nopcoder, line 502: adding containerName and fileName will help identify the call

block/azure/chunkwriting.go
```diff
@@ -0,0 +1,247 @@
...
+// And once complete multipart is called we commit all the blockIDs
+
+// blockWriter provides methods to upload blocks that represent a file to a server and commit them.
+// This allows us to provide a local implementation that fakes the server for hermetic testing.
+type blockWriter interface {
+	StageBlock(context.Context, string, io.ReadSeeker, azblob.LeaseAccessConditions, []byte, azblob.ClientProvidedKeyOptions) (*azblob.BlockBlobStageBlockResponse, error)
+	CommitBlockList(context.Context, []string, azblob.BlobHTTPHeaders, azblob.Metadata, azblob.BlobAccessConditions, azblob.AccessTierType, azblob.BlobTagsMap, azblob.ClientProvidedKeyOptions) (*azblob.BlockBlobCommitBlockListResponse, error)
+}
+
+func defaults(u *azblob.UploadStreamToBlockBlobOptions) error {
```
nopcoder, line 32: think we can remove this part and just initialize NewStaticBuffer

block/azure/chunkwriting.go
```diff
@@ -0,0 +1,247 @@
...
+	var err error
+	u.TransferManager, err = azblob.NewStaticBuffer(u.BufferSize, u.MaxBuffers)
+	if err != nil {
+		return fmt.Errorf("bug: default transfer manager could not be created: %w", err)
+	}
+	return nil
+}
+
+// copyFromReader copies a source io.Reader to blob storage using concurrent uploads.
+// TODO(someone): The existing model provides a buffer size and buffer limit as limiting factors.  The buffer size is probably
```
nopcoder, line 54: consider removing their TODOs

nopcoder (changes requested):

nopcoder (approved):

PR #1445, opened 2021-02-10
Load hooks

### LINKED ISSUE #1438: Load actions on pre-commit/merge
List and parse all existing yaml file in repo.

itaiad200:
close #1438


actions/action.go
```diff
@@ -119,7 +124,46 @@ type Source interface {
 }
 
 func LoadActions(source Source) ([]*Action, error) {
-	return nil, nil
+	hooksPath := source.List()
```
nopcoder, line 127: Something like

```go
func LoadActions(source Source) ([]*Action, error) {
	actionLocations := source.List()
	if len(actionLocations) == 0 {
		return nil, nil
	}
	actions := make([]*Action, len(actionLocations))
	var g multierror.Group
	for i, location := range actionLocations {
		g.Go(func() error {
			bytes, err := source.Load(p)
			if err != nil {
				return fmt.Errorf("loading action %s: %w", p, err)
			}
			action, err := ParseAction(bytes)
			if err != nil {
				return fmt.Errorf("parsing action %s: %w", p, err)
			}
			actions[i] = action
		})
	}
	err := g.Wait()
	if err != nil {
		return nil, err
	}
	return actions, nil
}
```

actions/action.go
```diff
@@ -113,13 +114,38 @@ func ParseAction(data []byte) (*Action, error) {
...
-type Source interface {
-	List() []string
-	Load(name string) ([]byte, error)
-}
-
 func LoadActions(source Source) ([]*Action, error) {
-	return nil, nil
+	hooksAddresses, err := source.List()
+	if err != nil {
+		return nil, fmt.Errorf("reading actions fir commit: %w", err)
```
nopcoder, line 120:
```suggestion
		return nil, fmt.Errorf("list actions from source: %w", err)
```

actions/action.go
```diff
@@ -113,13 +114,38 @@ func ParseAction(data []byte) (*Action, error) {
...
+	}
+
+	actions := make([]*Action, len(hooksAddresses))
+	var errGroup multierror.Group
+	for i := range hooksAddresses {
+		// localization of loop vars
+		ii := i
```
nopcoder, line 127: update comment - maybe // pin

actions/hook.go
```diff
@@ -29,10 +39,10 @@ var hooks = map[HookType]NewHookFunc{
 
 var ErrUnknownHookType = errors.New("unknown hook type")
 
-func NewHook(h HookType, a *Action, ah ActionHook) (Hook, error) {
-	f := hooks[h]
+func NewHook(a *Action, ah ActionHook) (Hook, error) {
+	f := hooks[HookType(ah.Type)]
```
nopcoder, line 43: I've also updated the type to match HookType

actions/hook.go
```diff
@@ -29,10 +39,10 @@ var hooks = map[HookType]NewHookFunc{
...
 	if f == nil {
-		return nil, fmt.Errorf("%w (%s)", ErrUnknownHookType, h)
+		return nil, fmt.Errorf("%w (%s)", ErrUnknownHookType, ah)
```
nopcoder, line 45: ah.Type ?

.gitignore
```diff
@@ -76,5 +76,6 @@ terraform.rc
 /graveler/sstable/mock/
 /pyramid/mock/
 /onboard/mock/
+/actions/mock/
```
nopcoder, line 79: if we check in the generated code we do not need to ignore it

Makefile
```diff
@@ -43,7 +43,7 @@ all: build
 
 clean:
 	@rm -rf $(API_BUILD_DIR) $(UI_BUILD_DIR) ddl/statik.go statik $(LAKEFS_BINARY_NAME) $(LAKECTL_BINARY_NAME) \
-	    graveler/committed/mock graveler/sstable/mock
```
nopcoder, line 46: not sure we require to clean the folder of a generated code if we check it in

catalog/actions.go
```diff
@@ -0,0 +1,56 @@
...
+	adapter block.Adapter
+
+	repositoryID graveler.RepositoryID
+	repository   graveler.Repository
+	ref          graveler.Ref
+}
+
+const actionsPrefix = "_lakefs_actions/"
+
+func (as *actionsSource) List(ctx context.Context) ([]string, error) {
```
nopcoder, line 24: This implementation can be an issue - as we do not return the real name here.

catalog/actions.go
```diff
@@ -0,0 +1,56 @@
...
+		addresses = append(addresses, it.Value().Entry.Address)
+	}
+	if it.Err() != nil {
+		return nil, fmt.Errorf("entries iterator: %w", it.Err())
+	}
+
+	return addresses, nil
+}
+
+func (as *actionsSource) Load(address string) ([]byte, error) {
```
nopcoder, line 42: loading by the physical address can be an issue - when we will have an error we will not know how to address which file had the problem

go.sum
```diff
@@ -308,6 +308,7 @@ github.com/gavv/httpexpect v2.0.0+incompatible/go.mod h1:x+9tiU1YnrOvnB725RkpoLv
 github.com/georgysavva/scany v0.2.7 h1:SBEuurTvWOUp7FnGBOjeSF9XWaWmVzc91h9baPo6y2s=
 github.com/georgysavva/scany v0.2.7/go.mod h1:bcxPhzeQFQqAUmjlZVwTGlu6AnWFSOiHpalfBe0xQ6U=
 github.com/ghemawat/stream v0.0.0-20171120220530-696b145b53b9/go.mod h1:106OIgooyS7OzLDOpUGgm9fA3bQENb/cFSyyBmMoJDs=
+github.com/ghodss/yaml v1.0.0 h1:wQHKEahhL6wmXdzwWG11gIVCkOv05bNOh+Rxn0yngAk=
```
nopcoder, line 311: ?

actions/hook.go
itaiad200: Saw that. Still not in master, right?

actions/hook.go
itaiad200: Done.

go.sum
itaiad200: reverted.

catalog/actions.go
itaiad200: Right. Fixed.

catalog/actions.go
itaiad200: Done.

.gitignore
nopcoder: sure, just remove the mock/hook.go

actions/hook.go
nopcoder: right :(

nopcoder (approved):

PR #1447, opened 2021-02-10
Different dbpool for branch-locker

### LINKED ISSUE #1446: BranchLocker deadlocks lakeFS
When using branch locker 2 DB connections are needed, one for the lock and the other for the actual action.
In some cases, when the system is under high load, some thread acquires the lock, but fails to create another connection to the DB since they were exhausted (by other requests. Default is 25 connections)

itaiad200:
closes #1446

### DIFF
--- a/catalog/rocks_cataloger.go
+++ b/catalog/rocks_cataloger.go
@@ -6,8 +6,6 @@ import (
 	"fmt"
 	"strings"
 
-	"github.com/treeverse/lakefs/config"
-	"github.com/treeverse/lakefs/db"
 	"github.com/treeverse/lakefs/graveler"
 	"github.com/treeverse/lakefs/logging"
 	"google.golang.org/protobuf/types/known/timestamppb"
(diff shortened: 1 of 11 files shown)


catalog/rocks_cataloger.go
```diff
@@ -28,8 +28,12 @@ const (
 
 var ErrUnknownDiffType = errors.New("unknown graveler difference type")
 
-func NewCataloger(db db.Database, cfg *config.Config) (Cataloger, error) {
-	entryCatalog, err := NewEntryCatalog(cfg, db)
+func NewCatalogerSingleWriter(db db.Database, cfg *config.Config) (Cataloger, error) {
```
ozkatz, line 31: Why do we need to special case this? why not always pass 2 pools?

ozkatz (approved):
That was fast!

catalog/rocks_cataloger.go
itaiad200: Discuss f2f, switched to config struct.

PR #1449, opened 2021-02-12
Change protobuf readers to a single generic(-ish) reader

arielshaqed:
All graveler readers are the same, they only take a different protobuf message type.  So
create a single generic parser and reader.

Use the standard technique: pass an empty "prototype" message.  Protobufs allow for creating
sort-of message clones, bypassing the need for understanding much of how to use generics.


arielshaqed:
@johnnyaug please have a look.  I think this technique for generic protobufs allows us roughly to *halve* the amount of plumbing code required to support 2 message types.  (It also makes it easier to support future method types, and maybe even make some testing happier.)

johnnyaug (approved):
Nice

arielshaqed:
Thanks!
Protobuf lint failure is unrelated. Pulling...

PR #1452, opened 2021-02-14
db rollback in case of panic

### LINKED ISSUE #1454: Panic database transaction should be rollbacked
Database transaction uses a callback to perform a scope operation during the database transaction.
In case of this function panic, we currently do not rollback the transaction which can cause resources such as locked to be kept (at least until database timeout).

nopcoder:
Fix #1454

### DIFF
--- a/db/database.go
+++ b/db/database.go
@@ -162,6 +162,13 @@ func (d *PgxDatabase) Transact(fn TxFunc, opts ...TxOpt) (interface{}, error) {
 	}
 	var attempt int
 	var ret interface{}
+	var tx pgx.Tx
+	defer func() {
+		if p := recover(); p != nil && tx != nil {
+			_ = tx.Rollback(options.ctx)
+			panic(p)
+		}
+	}()
 	for attempt < SerializationRetryMaxAttempts {
 		if attempt > 0 {
 			duration := SerializationRetryStartInterval * time.Duration(attempt)
--- a/db/database.go
+++ b/db/database.go
@@ -162,6 +162,7 @@ func (d *PgxDatabase) Transact(fn TxFunc, opts ...TxOpt) (interface{}, error) {
 	}
 	var attempt int
 	var ret interface{}
+	var err error
 	var tx pgx.Tx
 	defer func() {
 		if p := recover(); p != nil && tx != nil {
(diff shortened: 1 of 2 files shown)


ozkatz:
@nopcoder if a panic happens while a tx is open, it will not get committed, so why explicitly roll it back?

nopcoder:
> @nopcoder if a panic happens while a tx is open, it will not get committed, so why explicitly roll it back?

We don't need to have panic in the code, but we also need to abort open transactions.
In the case of the above, a test caused the panic and the out metadata locked transaction was left open, the next test that asked for out application-level locking was waiting until the test timed out.

nopcoder:
> @nopcoder if a panic happens while a tx is open, it will not get committed, so why explicitly roll it back?

I'll add a test code

db/database.go
```diff
@@ -162,6 +162,13 @@ func (d *PgxDatabase) Transact(fn TxFunc, opts ...TxOpt) (interface{}, error) {
 	}
 	var attempt int
 	var ret interface{}
+	var tx pgx.Tx
+	defer func() {
+		if p := recover(); p != nil && tx != nil {
+			_ = tx.Rollback(options.ctx)
```
arielshaqed, line 168: Consider reporting this error, e.g.
```suggestion
			err = tx.Rollback(options.ctx)
			if err != nil {
				p = fmt.Errorf("%w, and %s while rolling back", p, err)
			}
```
If something *does* go wrong rolling back during a panic, more information is better.

db/database.go
```diff
@@ -162,6 +162,13 @@ func (d *PgxDatabase) Transact(fn TxFunc, opts ...TxOpt) (interface{}, error) {
 	}
 	var attempt int
 	var ret interface{}
+	var tx pgx.Tx
+	defer func() {
```
arielshaqed, line 166: Can you just move this to after we open the transaction, below?  That way we can just use tx, and I think the code becomes more clear.

db/database.go
```diff
@@ -162,6 +162,13 @@ func (d *PgxDatabase) Transact(fn TxFunc, opts ...TxOpt) (interface{}, error) {
...
+		if p := recover(); p != nil && tx != nil {
+			_ = tx.Rollback(options.ctx)
+			panic(p)
```
arielshaqed, line 169: Does this lose the panic stacktrace (or is it stored on `p`?)?  If so, can we keep it somehow?

arielshaqed (approved):
Nice!
Makes sense, and should help our tests fail more cleanly.

db/database.go
nopcoder: using defer in a loop can be a problem - we would like to have this one once or scope the operation into a different function and make sure it is committed there

db/database.go
nopcoder: Verified that the stack trace captures the original panic stack

db/database.go
nopcoder: Currently, we do no log rollback in the Transaction function. We can capture the information and the stack trace in the top-level handler if needed.
But what will be the added value of the rollback error as we assume everything is aborted in this flow.

PR #1455, opened 2021-02-14
basic client config to pass default max conns per host.

ozkatz:
By default the Swagger client doesn't configure this, which falls back to Go's default http.DefaultMaxIdleConnsPerHost which is 2, which will exhaust TCP connections if calling the API with many concurrent goroutines.


api/client.go
```diff
@@ -728,16 +730,31 @@ func (c *client) RefsDump(ctx context.Context, repository string) (*models.RefsD
 	return resp.GetPayload(), nil
 }
 
-func NewClient(endpointURL, accessKeyID, secretAccessKey string) (Client, error) {
+type ClientOption func(*client)
+
+func MaxIdleConnsPerHost(maxIdelConnsPerHost int) ClientOption {
```
arielshaqed, line 735:
```suggestion
func MaxIdleConnsPerHost(maxIdleConnsPerHost int) ClientOption {
```

arielshaqed (changes requested):
I don't understand why we are _increasing_ `MaxIdleConnsPerHost`; how does that _reduce_ the number of connections used?  I **think** we want to set MaxConnsPerHost.

api/client.go
```diff
@@ -728,16 +730,31 @@ func (c *client) RefsDump(ctx context.Context, repository string) (*models.RefsD
...
+	transport := httptransport.New(parsedURL.Host, parsedURL.Path, []string{parsedURL.Scheme})
+	transport.Transport.(*http.Transport).MaxIdleConnsPerHost = 100 // make this configurable?
+	c := &client{
+		transport: transport,
+		remote:    genclient.New(transport, strfmt.Default),
+		auth:      httptransport.BasicAuth(accessKeyID, secretAccessKey),
+	}
+	for _, opt := range opts {
+		opt(c)
+	}
```
nopcoder, line 758: OpenAPI uses the defaults from http.DefaultTransport and gives you the option to replace the client.
I'm worried that if we keep this path and try to configure every aspect in the client using options it will not end well or break in the case for example we will want to use some of the options the generated code enables us.

```go
	clientTransport := httptransport.NewWithClient(parsedURL.Host, parsedURL.Path, []string{parsedURL.Scheme}, client)
	clientTransport.EnableConnectionReuse()
	clientTransport.DefaultAuthentication = httptransport.BasicAuth(accessKeyID, secretAccessKey)
```
if we `EnableConnectionReuse` for example the implementation replaces the underlying Transport and the cast we introduced will break.

I suggest we configure the client we want and the default authentication.
We also need to consider using the generated client directly as for now it doesn't look like the only value is the option to have a flat interface and not a fluent API like swagger generates.

nopcoder (changes requested):

api/client.go
ozkatz: not having to pass authentication info on every call but rather once on initialization is the main reason the client wrapper was introduced, along with encapsulating the domain knowledge of how to assemble a Swagger generated client

It is encapsulated by desgin - the constructor wraps the logic of how to assemble a swagger generated client, so we can rely on casting because we know the type of transport is only controlled by us.

ozkatz:
> I don't understand why we are _increasing_ `MaxIdleConnsPerHost`; how does that _reduce_ the number of connections used? I **think** we want to set MaxConnsPerHost.

See https://stackoverflow.com/questions/39813587/go-client-program-generates-a-lot-a-sockets-in-time-wait-state - when running the `abuse` tests, I quikly generated more "live" sockets than the OS allowed me. Almost all of them were in TIME_WAIT state. By increasing `MaxIdleConnsPerHost` we avoid closing and then immediately repopening a new connection (which will be opened before the server closes the connection too, thus we accumulate many connections in `TIME_WAIT` status)

api/client.go
ozkatz: Fixed

nopcoder (approved):

arielshaqed:
> > I don't understand why we are _increasing_ `MaxIdleConnsPerHost`; how does that _reduce_ the number of connections used? I **think** we want to set MaxConnsPerHost.
> 
> @arielshaqed See https://stackoverflow.com/questions/39813587/go-client-program-generates-a-lot-a-sockets-in-time-wait-state - when running the `abuse` tests, I quikly generated more "live" sockets than the OS allowed me. Almost all of them were in TIME_WAIT state. By increasing `MaxIdleConnsPerHost` we avoid closing and then immediately repopening a new connection (which will be opened before the server closes the connection too, thus we accumulate many connections in `TIME_WAIT` status)

Thanks for the ref!  **tl;dr:** Not a fan of this solution, but the most rapid unblocking.  And:
1. Complaints about tuning are reduced by not needing to tune it _right now_.
2. Complaints about possible bad performance will be made stronger by showing a better (higher throughput) implementation.

It is mysterious a knob I don't understand how to tweak.

I still don't understand this SO "solution"; it will be hard to understand just how large the magic value needs to be.  This solution appears to cover up an issue with how we use the Go HTTP client.  I believe that for performance testing a correct solution is to use a separate client connection pool for each goroutine.

arielshaqed (approved):

PR #1465, opened 2021-02-16
Add "get metarange", "get range" API endpoints

arielshaqed:
Returns data (currently just the address of the object) by ID.

### DIFF
--- a/api/controller.go
+++ b/api/controller.go
@@ -1281,6 +1281,56 @@ func writeSymlinkToS3(params metadata.CreateSymlinkParams, repo *catalog.Reposit
 	return err
 }
 
+func (c *Controller) MetadataGetMetarangeHandler() metadata.GetMetaRangeHandler {
+	return metadata.GetMetaRangeHandlerFunc(func(params metadata.GetMetaRangeParams, user *models.User) middleware.Responder {
+		deps, err := c.setupRequest(user, params.HTTPRequest, []permissions.Permission{
+			{
+				// TODO(oz): May want tighter permissions here.
+				Action:   permissions.ReadRepositoryAction,
+				Resource: permissions.RepoArn(params.Repository),
+			},
+		})
+		if err != nil {
+			return metadata.NewGetMetaRangeUnauthorized().WithPayload(responseErrorFrom(err))
+		}
+		deps.LogAction("metadata_get_metarange")
+		cataloger := deps.Cataloger
+
+		metarange, err := cataloger.GetMetaRange(deps.ctx, params.Repository, params.MetaRange)
+		if err != nil {
+			return metadata.NewGetMetaRangeDefault(http.StatusInternalServerError).WithPayload(responseErrorFrom(err))
+		}
+		ret := metadata.NewGetMetaRangeOK()
+		ret.Location = metarange.Address
+		return ret
+	})
+}
+
+func (c *Controller) MetadataGetRangeHandler() metadata.GetRangeHandler {
+	return metadata.GetRangeHandlerFunc(func(params metadata.GetRangeParams, user *models.User) middleware.Responder {
+		deps, err := c.setupRequest(user, params.HTTPRequest, []permissions.Permission{
+			{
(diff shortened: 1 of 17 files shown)


graveler/graveler.go
```diff
@@ -321,6 +334,15 @@ type VersionController interface {
 	SetHooksHandler(handler HooksHandler)
 }
 
+// Plumbing includes commands for fiddling more directly with graveler implementation
+// internals.
+type Plumbing interface {
+	// GetMetarange returns information where metarangeID is stored.
+	GetMetaRange(ctx context.Context, repositoryID RepositoryID, metaRangeID MetaRangeID) (MetaRangeData, error)
+	// GetRange returns information where rangeID is stored.
```
nopcoder, line 342: Consider calling it xxxStats returning xxxInfo
ex: MetaRangeStat -> MetaRangeInfo

api/controller.go
```diff
@@ -1265,6 +1265,56 @@ func writeSymlinkToS3(params metadata.CreateSymlinkParams, repo *catalog.Reposit
 	return err
 }
 
+func (c *Controller) MetadataGetMetarangeHandler() metadata.GetMetaRangeHandler {
+	return metadata.GetMetaRangeHandlerFunc(func(params metadata.GetMetaRangeParams, user *models.User) middleware.Responder {
+		deps, err := c.setupRequest(user, params.HTTPRequest, []permissions.Permission{
+			{
+				// TODO(oz): May want tighter permissions here.
```
nopcoder, line 1272: Remember to remove the todo after verifying

api/controller.go
```diff
@@ -1265,6 +1265,56 @@ func writeSymlinkToS3(params metadata.CreateSymlinkParams, repo *catalog.Reposit
...
+		ret.Location = metarange.Address
+		return ret
+	})
+}
+
+func (c *Controller) MetadataGetRangeHandler() metadata.GetRangeHandler {
+	return metadata.GetRangeHandlerFunc(func(params metadata.GetRangeParams, user *models.User) middleware.Responder {
+		deps, err := c.setupRequest(user, params.HTTPRequest, []permissions.Permission{
+			{
+				// TODO(oz): May want tighter permissions here.
```
nopcoder, line 1297: Remember to remove the todo after verifying

swagger.yml
```diff
@@ -2154,6 +2154,64 @@ paths:
...
+        - metadata
+      operationId: getMetaRange
+      summary: return URI to a meta-range file
+      # TODO(ariels): Actually redirect to S3.
+      responses:
+        200:
+          description: meta-range URI
+          headers:
+            Location:
+              type: string
```
nopcoder, line 2178: Why did we prefer to return the response just as a header for this case?
All of our API usually return the response as JSON body.
We can have `location` (bool) parameter to control if we like to get the response just as a location header for a client that doesn't want to parse the response body.

nopcoder (approved):
Looks great, with minor comments/suggestions and a question down at the end before you commit.

swagger.yml
arielshaqed: This is *almost* a redirect action.  In fact, if we go ahead with signed URLs then it _will_ be a redirect.  I prefer to abuse HTTP semantics here, at least a bit.

arielshaqed (commented):
Thanks!

Keeping the Location header, changing the names from "...Data" to "...Info".

arielshaqed:
+ @ozkatz to look at permissions for these actions, please.

api/controller.go
arielshaqed: After f2fing, changed to require _both_ list objects _and_ read repo: it gives a cross between the 2.  E.g. with (just) read permission on underlying storage it gives list objects (and without it gives size of branch), which is list-objects-like.  But it _also_ gives the storage namespace for the repo, which is read-repo-like.  So best to require both.

arielshaqed (commented):
Thanks!
Pulling once the tests settle.

PR #1473, opened 2021-02-17
Add Azure configuration to docs

guy-har:


nopcoder (changes requested):
endpoint is missing

nopcoder (approved):
consider align the service_url name to the same one we have for s3 - endpoint

PR #1478, opened 2021-02-17
Add nessie test for hooks

itaiad200:


nessie/hooks_test.go
```diff
@@ -0,0 +1,239 @@
...
+	fmt.Println("Using port:", port)
+	s := &http.Server{
+		Handler: mux,
+	}
+
+	go func() {
+		if err = s.Serve(listener); err != nil && err != http.ErrServerClosed {
+			log.Fatalf("listen:%s\n", err)
+		}
+	}()
```
nopcoder, line 189: Think we better off with just
```go
s := &http.Server{ Handler: mux }
go s.ListenAndServe(":0", nil)
```

nopcoder (approved):

api/controller.go
```diff
@@ -250,6 +250,10 @@ func (c *Controller) SetupLakeFSHandler() setupop.SetupLakeFSHandler {
 				})
 		}
 
+		c.deps.Logger.
+			WithField("RemoteAddr", setupReq.HTTPRequest.RemoteAddr).
+			Info("Received setup request")
```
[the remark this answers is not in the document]
itaiad200: Done.

nessie/hooks_test.go
itaiad200: I wish, need the port..

PR #1470, opened 2021-02-16
Hooks results db persist

nopcoder:

PR #1486, opened 2021-02-20
Feature/hook db alt impl where actions know catalog

nopcoder:
This is an alternative for pr #1470


itaiad200 (commented):
WIP

actions/action.go
```diff
@@ -73,9 +78,9 @@ func (a *Action) Match(spec MatchSpec) (bool, error) {
 	// at least one matched event definition
 	var actionOn *ActionOn
 	switch spec.EventType {
-	case EventTypePreCommit:
+	case graveler.EventTypePreCommit:
 		actionOn = a.On.PreCommit
-	case EventTypePreMerge:
+	case graveler.EventTypePreMerge:
 		actionOn = a.On.PreMerge
 	default:
```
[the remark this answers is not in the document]
nopcoder: Not exactly - the switch case just selects the right entry based on the event. The entry itself can be nil if not set in the yaml.

graveler/graveler.go
```diff
@@ -1017,7 +1032,24 @@ func (g *Graveler) Commit(ctx context.Context, repositoryID RepositoryID, branch
 	if err != nil {
 		return "", err
 	}
-	return res.(CommitID), nil
+	newCommitID := res.(CommitID)
+	runID := NewRunID()
```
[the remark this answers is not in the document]
nopcoder: I think it is confusing - the record describes the current event and the run id is of the current event.
If we will name it preRunID, we need to do 'if' on the event type to identify which run ID we should use and limit the hook trigger just to pre and post-event.

graveler/graveler.go
```diff
@@ -1017,7 +1032,24 @@ func (g *Graveler) Commit(ctx context.Context, repositoryID RepositoryID, branch
...
+		RepositoryID:     repositoryID,
+		StorageNamespace: storageNamespace,
+		BranchID:         branchID,
+		Commit:           commit,
+		CommitID:         newCommitID,
+		PreRunID:         preRunID,
+	})
+	if err != nil {
+		g.log.WithError(err).
+			WithField("run_id", runID).
```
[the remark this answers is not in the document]
nopcoder: Adding the pre_run_id to more information - but I still think that the associated run id should keep the name - it doesn't matter if the event is pre or post

go.sum
```diff
@@ -1,5 +1,6 @@
 4d63.com/gochecknoglobals v0.0.0-20201008074935-acfc0b28355a h1:wFEQiK85fRsEVF0CRrPAos5LoAryUsIX1kPW/WrIqFw=
 4d63.com/gochecknoglobals v0.0.0-20201008074935-acfc0b28355a/go.mod h1:wfdC5ZjKSPr7CybKEcgJhUOgeAQW1+7WcyK8OvUilfo=
+bazil.org/fuse v0.0.0-20160811212531-371fbbdaa898 h1:SC+c6A1qTFstO9qmB86mPV2IpYme/2ZoEQ0hrP+wo+Q=
```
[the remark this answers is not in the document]
nopcoder: looks like it - updated with go mod tidy

ddl/000023_actions.up.sql
```diff
@@ -0,0 +1,35 @@
...
+    run_id              text NOT NULL,
+    hook_run_id         text NOT NULL,
+
+    action_name         text NOT NULL,
+    hook_id             text NOT NULL,
+    start_time          timestamptz NOT NULL,
+    end_time            timestamptz NOT NULL,
+    passed              BOOLEAN DEFAULT false NOT NULL,
+
+    PRIMARY KEY (repository_id, run_id, hook_run_id)
```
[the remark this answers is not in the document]
nopcoder: from my point of view, the hook run is part of the actions run so it is part of the key - when we select the hooks we run as part of a specific run_id we pass where repository_id and run_id which is part of the index.

catalog/entry_catalog.go
```diff
@@ -84,7 +81,8 @@ type Store interface {
 }
 
 type Actions interface {
```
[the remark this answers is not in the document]
nopcoder: no, right - removing

graveler/graveler.go
```diff
@@ -944,18 +943,23 @@ func (g *Graveler) List(ctx context.Context, repositoryID RepositoryID, ref Ref)
 }
```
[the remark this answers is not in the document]
nopcoder: We use hookNoOp when `graveler` initialized, but in order not to have null check, we set the hookNoOp if case the caller do not want to register any handler.

actions/service.go
```diff
@@ -102,40 +121,199 @@ func (s *Service) loadMatchedActions(ctx context.Context, source Source, spec Ma
...
 		})
 	}
 	return g.Wait().ErrorOrNil()
 }
 
-func NewRunID(t time.Time) string {
-	uid := strings.ReplaceAll(uuid.New().String(), "-", "")
-	runID := t.UTC().Format(time.RFC3339) + "_" + uid
-	return runID
+func (s *Service) insertRunInformation(ctx context.Context, record graveler.HookRecord, tasks []*Task) error {
```
itaiad200: Name is a little bit misleading since it will only work for Runs which are `pre_*` (since commitID isn't set for them).
Maybe just mention that in the func comment?

actions/service.go
```diff
@@ -5,94 +5,113 @@ package actions
...
+
+type RunResult struct {
+	RunID     string    `db:"run_id"`
+	BranchID  string    `db:"branch_id"`
+	SourceRef string    `db:"source_ref"`
+	EventType string    `db:"event_type"`
+	StartTime time.Time `db:"start_time"`
+	EndTime   time.Time `db:"end_time"`
+	Passed    bool      `db:"passed"`
+	CommitID  string    `db:"commit_id"`
```
itaiad200, line 41: Nullable, no?

actions/service.go
```diff
@@ -102,40 +121,199 @@ func (s *Service) loadMatchedActions(ctx context.Context, source Source, spec Ma
...
+	return s.Run(ctx, record)
+}
+
+func (s *Service) PostCommitHook(ctx context.Context, record graveler.HookRecord) error {
+	// update pre-commit with commit ID if needed
+	err := s.UpdateCommitID(ctx, record.RepositoryID.String(), record.PreRunID, record.CommitID.String())
+	if err != nil {
+		return err
+	}
+	return nil
```
itaiad200, line 316:
```suggestion
	return s.UpdateCommitID(ctx, record.RepositoryID.String(), record.PreRunID, record.CommitID.String())
```

actions/service.go
```diff
@@ -102,40 +121,199 @@ func (s *Service) loadMatchedActions(ctx context.Context, source Source, spec Ma
...
+}
+
+func (s *Service) PostMergeHook(ctx context.Context, record graveler.HookRecord) error {
+	// update pre-merge with commit ID if needed
+	err := s.UpdateCommitID(ctx, record.RepositoryID.String(), record.PreRunID, record.CommitID.String())
+	if err != nil {
+		return err
+	}
+	// TODO(barak): invoke post merge actions
+	return nil
```
itaiad200, line 330:
```suggestion
	return s.UpdateCommitID(ctx, record.RepositoryID.String(), record.PreRunID, record.CommitID.String())
```

api/controller.go
```diff
@@ -69,9 +69,10 @@ type Dependencies struct {
 }
 
 type actionsHandler interface {
-	GetRun(repository, runID string) (actions.RunResult, error)
-	ListHooks(repository, runID string, before time.Time) (actions.TaskResultIter, error)
-	ListRuns(repository string, before time.Time, branch *string) (actions.RunResultIter, error)
+	GetRunResult(ctx context.Context, repositoryID string, runID string) (*actions.RunResult, error)
+	GetTaskResult(ctx context.Context, repositoryID string, runID string, hookRunID string) (*actions.TaskResult, error)
+	ListRuns(ctx context.Context, repositoryID string, branchID *string, after string) (actions.RunResultIterator, error)
```
itaiad200, line 74: For consistency:
```suggestion
	ListRunResults(ctx context.Context, repositoryID string, branchID *string, after string) (actions.RunResultIterator, error)
```

api/controller.go
```diff
@@ -69,9 +69,10 @@ type Dependencies struct {
...
+	ListRunTasks(ctx context.Context, repositoryID string, runID string, after string) (actions.TaskResultIterator, error)
```
itaiad200, line 75:
```suggestion
	ListRunTaskResults(ctx context.Context, repositoryID string, runID string, after string) (actions.TaskResultIterator, error)
```

api/controller.go
```diff
@@ -427,6 +428,10 @@ func (c *Controller) CommitHandler() commits.CommitHandler {
 		commitMessage := swag.StringValue(params.Commit.Message)
 		commit, err := deps.Cataloger.Commit(deps.ctx, params.Repository,
 			params.Branch, commitMessage, committer, params.Commit.Metadata)
+		var hookAbortErr *graveler.HookAbortError
+		if errors.As(err, &hookAbortErr) {
+			return commits.NewCommitPreconditionFailed().WithPayload(responseError("aborted by hooks, run id: " + hookAbortErr.RunID))
```
itaiad200, line 433: Actual error information isn't returned?

actions/service.go
nopcoder: I prefer to have an empty string as no commit_id instead of playing with nil

api/controller.go
```diff
@@ -878,17 +883,21 @@ func (c *Controller) MergeMergeIntoBranchHandler() refs.MergeIntoBranchHandler {
...
 		res, err := deps.Cataloger.Merge(deps.ctx,
 			params.Repository, params.DestinationBranch, params.SourceRef,
 			userModel.Username,
 			message,
-			metadata)
+			meta)
 
+		var hookAbortErr *graveler.HookAbortError
+		if errors.As(err, &hookAbortErr) {
+			return refs.NewMergeIntoBranchPreconditionFailed().WithPayload(responseError("aborted by hooks, run id: " + hookAbortErr.RunID))
```
itaiad200, line 899: Same concern about error info.

api/controller.go
```diff
@@ -2738,58 +2735,51 @@ func (c *Controller) ActionsListRunHooksHandler() actionsop.ListRunHooksHandler
...
-			val := hooksIter.Value()
-			status := models.HookRunStatusCompleted
-			if !val.Passed {
-				status = models.HookRunStatusFailed
-			}
+		defer tasksIter.Close()
 
-			res.Results = append(res.Results, &models.HookRun{
+		payload := &actionsop.ListRunHooksOKBody{
+			Results: make([]*models.HookRun, 0),
```
itaiad200, line 2746:
```suggestion
			Results: make([]*models.HookRun),
```

actions/service.go
nopcoder: leaving a comment to post merge/commit actions

actions/service.go
nopcoder: leaving a comment to post merge/commit actions

actions/service.go
nopcoder: Updated the code to update 'commit_id' - currently, this code updates information just for preXXX operations, but it supports update information for postXXX events too

graveler/graveler.go
```diff
@@ -1337,6 +1386,24 @@ func (g *Graveler) Merge(ctx context.Context, repositoryID RepositoryID, destina
 		return "", DiffSummary{}, err
 	}
 	c := res.(*CommitIDAndSummary)
+	runID := NewRunID()
```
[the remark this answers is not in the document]
nopcoder: named the local variable postXXX but keeping the record and logging as RunID

api/controller.go
```diff
@@ -2801,64 +2791,50 @@ func (c *Controller) ActionsListRunsHandler() actionsop.ListRunsHandler {
...
 			}
-
 			return actionsop.NewListRunsDefault(http.StatusInternalServerError).
 				WithPayload(responseErrorFrom(err))
 		}
 
-		res := &actionsop.ListRunsOKBody{Pagination: &models.Pagination{HasMore: swag.Bool(true)}}
-		for runsIter.Next() {
+		payload := &actionsop.ListRunsOKBody{
+			Results: make([]*models.ActionRun, 0),
```
itaiad200, line 2806:
```suggestion
			Results: make([]*models.ActionRun),
```

api/controller.go
nopcoder: This is on purpose so the caller doesn't need to handle nil - will get empty array

api/controller.go
nopcoder: This is on purpose so the caller doesn't need to handle nil - will get empty array

itaiad200 (approved):
LGTM

PR #1492, opened 2021-02-21
fixed diff throwing 500 instead of 404 when ref/repo not found

### LINKED ISSUE #1476: OpenAPI diff returns 500 when repository/ref not found instead of 404
when calling the `diffRefs` operation, the swagger.yaml spec defines a 404 `reference not found` response type, but when passing a repository, leftRef or rightRef that are not found, I get back:

`500 Internal Server Error: {"message":"could not diff references: get commit by ref somebranch: not found"}`

ozkatz:
closes #1476


nopcoder (approved):
nice backward compatibility

PR #1495, opened 2021-02-21
make swagger.yml pass jsonschema validation

ozkatz:


nopcoder (approved):
nice, there are more validations related to swagger that fixed part of the hooks

PR #1507, opened 2021-02-23
set webhook default timeout to one minute

nopcoder:
webhook: set http client default timeout to one minute

### DIFF
--- a/actions/webhook.go
+++ b/actions/webhook.go
@@ -32,7 +32,7 @@ type WebhookEventInfo struct {
 	Metadata      map[string]string `json:"metadata,omitempty"`
 }
 
-const webhookClientTimeout = 5 * time.Minute
+const webhookClientTimeout = 1 * time.Minute
 
 var (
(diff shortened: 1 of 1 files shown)


actions/webhook.go
```diff
@@ -32,7 +32,10 @@ type WebhookEventInfo struct {
 	Metadata      map[string]string `json:"metadata,omitempty"`
 }
 
-const webhookClientTimeout = 5 * time.Minute
+const (
+	webhookClientTimeout      = 1 * time.Minute
+	webhookTimeoutPropertyKey = "timeout"
```
itaiad200, line 37: So the user can supply the value for this?

itaiad200: Maybe it should start with some prefix to differentiate between lakeFS properties and custom user properties.

nopcoder: All the properties are associated with the hook implementation. if you want custom properties per webhook it should be property under properties.

itaiad200 (approved):

PR #1508, opened 2021-02-23
Actions run output under hook run id

### LINKED ISSUE #1505: Actions: capture runs information into the repository storage
Actions run information is currently kept inside Postgres. In order to associate the log files with the action/hook that produced them without database access, we should render a file with that information.
A JSON file under the run_id folder with the above information can supply the relevant information and enable better integrations with lakeFS.

nopcoder:
Close #1505

- generate and store run manifest under _lakefs/actions/log/<run_id>/run.manifest

Manifest is a JSON file that captures the results - it can help map the log files to the actual action/hook_id

 

- [ ] add more test code


actions/service.go
```diff
@@ -174,66 +183,101 @@ func (s *Service) runTasks(ctx context.Context, record graveler.HookRecord, task
...
+				manifest.Run.Passed = false
 			}
 		}
 	}
-	return
+
+	return manifest
 }
 
 func (s *Service) UpdateCommitID(ctx context.Context, repositoryID string, runID string, commitID string) error {
```
itaiad200, line 283: This needs to update OS too.

actions/service.go
```diff
@@ -174,66 +183,101 @@ func (s *Service) runTasks(ctx context.Context, record graveler.HookRecord, task
...
-					break
-				}
-				taskPassed := task.Err == nil
-				_, err = tx.Exec(`INSERT INTO actions_run_hooks(repository_id, run_id, hook_run_id, action_name, hook_id, start_time, end_time, passed)
-			VALUES ($1,$2,$3,$4,$5,$6,$7,$8)`,
-					record.RepositoryID, task.RunID, task.HookRunID, task.Action.Name, task.HookID, task.StartTime, task.EndTime, taskPassed)
-				if err != nil {
-					return nil, fmt.Errorf("insert run hook information (%s %s): %w", task.Action.Name, task.HookID, err)
-				}
+		for _, hookRun := range manifest.HooksRun {
```
itaiad200, line 225: Just a thought (@ozkatz thought really): Can we avoid writing to the db and read directly from the os?
I think we can, the only thing missing is the ability to filter by branches or specific hook runs.

actions/service.go
```diff
@@ -174,66 +183,101 @@ func (s *Service) runTasks(ctx context.Context, record graveler.HookRecord, task
...
+		return fmt.Errorf("marshal run manifest: %w", err)
+	}
+	runManifestPath := FormatRunManifestOutputPath(record.RunID)
+	manifestReader := bytes.NewReader(manifestJSON)
+	manifestSize := int64(len(manifestJSON))
+	storageNamespace := record.StorageNamespace.String()
+	return s.Writer.OutputWrite(ctx, storageNamespace, runManifestPath, manifestReader, manifestSize)
+}
+
+func (s *Service) insertRunInformation(ctx context.Context, repositoryID graveler.RepositoryID, manifest runManifest) error {
```
itaiad200, line 213:
```suggestion
func (s *Service) saveRunInformationDB(ctx context.Context, repositoryID graveler.RepositoryID, manifest runManifest) error {
```

actions/service.go
```diff
@@ -174,66 +183,101 @@ func (s *Service) runTasks(ctx context.Context, record graveler.HookRecord, task
...
+
+	err := s.insertRunInformation(ctx, record.RepositoryID, manifest)
+	if err != nil {
+		return fmt.Errorf("insert run information: %w", err)
+	}
+
+	return s.saveRunManifest(ctx, record, manifest)
+}
+
+func (s *Service) saveRunManifest(ctx context.Context, record graveler.HookRecord, manifest runManifest) error {
```
itaiad200, line 201:
```suggestion
func (s *Service) saveRunManifestOS(ctx context.Context, record graveler.HookRecord, manifest runManifest) error {
```

actions/service.go
```diff
@@ -174,66 +183,101 @@ func (s *Service) runTasks(ctx context.Context, record graveler.HookRecord, task
...
+}
+
+func (s *Service) insertRunInformation(ctx context.Context, repositoryID graveler.RepositoryID, manifest runManifest) error {
 	_, err := s.DB.Transact(func(tx db.Tx) (interface{}, error) {
 		// insert run information
+		run := manifest.Run
 		_, err := tx.Exec(`INSERT INTO actions_runs(repository_id, run_id, event_type, start_time, end_time, branch_id, source_ref, commit_id, passed)
 			VALUES ($1,$2,$3,$4,$5,$6,$7,$8,$9)`,
-			record.RepositoryID, record.RunID, record.EventType, runStartTime, runEndTime, record.BranchID, record.SourceRef, record.CommitID, runPassed)
+			repositoryID, run.RunID, run.EventType, run.StartTime, run.EndTime, run.BranchID, run.SourceRef, run.CommitID, run.Passed)
```
itaiad200, line 219: `run.CommitID` is always empty in this stage, no?

itaiad200 (changes requested):
Must update OS too on `UpdateCommitID`

actions/service.go
nopcoder: I'll read the file and update the commit ID - not sure I like to re-write but it will reflect the same information as you said.

actions/service.go
nopcoder: will need need to update our adapters to support more options in the listing and filter, but possible.
opened https://github.com/treeverse/lakeFS/issues/1511

actions/service.go
nopcoder: yes, but when you call Run for post events - it will include the commit ID

actions/service.go
nopcoder: Updated

actions/service.go
```diff
@@ -280,31 +279,56 @@ func buildRunManifest(record graveler.HookRecord, tasks [][]*Task) runManifest {
 	return manifest
 }
 
-func (s *Service) UpdateCommitID(ctx context.Context, repositoryID string, runID string, commitID string) error {
+// UpdateCommitID assume record is a post event, we use the PreRunID to update the commit_id and save the run manifest again
+func (s *Service) UpdateCommitID(ctx context.Context, repositoryID string, storageNamespace string, runID string, commitID string) error {
+	if runID == "" || commitID == "" {
+		return nil
```
itaiad200, line 285: Why isn't this an error?

nopcoder: true, I'll verify just the run ID and let update commit ID to none if needed

nopcoder: I assume the pre event completed in order to post event to trigger, so we must of run ID. if this will change in the future. Prefer to skip if no run ID was sent to UpdateCommitID

itaiad200 (approved):

Issue #1509, opened 2021-02-24
diffBranch returns 500 on branch/repo not found instead of 404


ozkatz:

ozkatz:
closed by #1510

PR #1510, opened 2021-02-24
fixed diffBranch returning 500 instead of 404 when branch not found

ozkatz:


ozkatz:
closes #1509

api/controller.go
```diff
@@ -969,6 +969,10 @@ func (c *Controller) BranchesDiffBranchHandler() branches.DiffBranchHandler {
 		limit := int(swag.Int64Value(params.Amount))
 		after := swag.StringValue(params.After)
 		diff, hasMore, err := cataloger.DiffUncommitted(deps.ctx, params.Repository, params.Branch, limit, after)
+		if errors.Is(err, catalog.ErrNotFound) {
```
arielshaqed, line 972: We have TWO different ways not to find something: `catalog.ErrNotFound` and `graveler.ErrNotFound`.  Well, actually 3 because we also have one in `db`.  Elsewhere in this file we test for both alternatives; could you do so here, too?
One day we'll finish getting rid of package `catalog` and then we'll only have one way not to find something.

arielshaqed (approved):
Thanks!
In view of the comment (I still want to test both, no matter what...), how did we test this?

api/controller.go
ozkatz: Actually, it's safe to use `catalog.ErrNotFound` everywhere (`catalog.ErrNotFound` is actually `db.ErrNotFound`, while `graveler.ErrNotFound` is a wrapper around `db.ErrNotFound`, or in other words, `errors.Is(err, catalog.ErrNotFound)` will catch any of the 3). It's not pretty but it's temporary until we reduce the amount of layers.

ozkatz:
> Thanks!
> In view of the comment (I still want to test both, no matter what...), how did we test this?

Please rereview, this was a good opportunity to add a test case.

api/controller.go
arielshaqed: Cool.  But a yucky shade of cool, because it is clearly wrong.  Luckily `db` is going away soon, so we can keep this.

api/controller_test.go
```diff
@@ -698,6 +698,59 @@ func TestController_GetBranchHandler(t *testing.T) {
...
+		if diff.Payload.Results[0].Path != "a/b" {
+			t.Fatalf("got wrong diff object, expected a/b, got %s", diff.Payload.Results[0].Path)
+		}
+	})
+
+	t.Run("diff branch that doesn't exist", func(t *testing.T) {
+		_, err := clt.Branches.DiffBranch(branches.NewDiffBranchParams().
+			WithRepository("repo1").
+			WithBranch("some-other-missing-branch"), bauth)
+		if _, ok := err.(*branches.DiffBranchNotFound); !ok {
```
arielshaqed, line 747: Please use errors.Is.

arielshaqed (approved):
Thanks!

Please errors.Is, but good.

api/controller_test.go
ozkatz: if only I could. it's generated swagger code, so this is just a regular struct and not a wrapped error.

PR #1516, opened 2021-02-24
Add query params to webhook action

### LINKED ISSUE #1514: Add webhook property for query params
Dynamic way of adding info to the request

itaiad200:
closes #1514

### DIFF
--- a/actions/webhook.go
+++ b/actions/webhook.go
@@ -13,10 +13,11 @@ import (
 )
 
 type Webhook struct {
-	ID         string
-	ActionName string
-	URL        string
-	Timeout    time.Duration
+	ID          string
+	ActionName  string
+	URL         string
+	Timeout     time.Duration
+	QueryParams map[string][]string
 }
 
 type WebhookEventInfo struct {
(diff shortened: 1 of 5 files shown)


actions/webhook.go
```diff
@@ -33,32 +34,39 @@ type WebhookEventInfo struct {
...
+	if err != nil {
+		return nil, fmt.Errorf("extracting query params: %w", err)
 	}
 
 	requestTimeout := webhookClientTimeout
-	timeoutDuration := h.Properties["timeout"]
-	if len(timeoutDuration) > 0 {
-		d, err := time.ParseDuration(timeoutDuration)
+	timeoutDuration, ok := h.Properties["timeout"]
+	if ok && len(timeoutDuration.(string)) > 0 {
```
nopcoder, line 57: You will need to do this one first:
```go
timeoutString, ok := timeoutDuration.(string)
```
to verify the property is a string - without the explicit capture you will get panic

nopcoder:
I don't think adding query parameters as part of the action structure is relevant at this point.
This is why we didn't add custom parameters to the request body in the first place.

actions/webhook.go
```diff
@@ -35,33 +36,40 @@ type WebhookEventInfo struct {
...
 
 var (
 	ErrWebhookRequestFailed = errors.New("webhook request failed")
-	ErrWebhookMissingURL    = errors.New("webhook missing url")
+	ErrWebhookWrongFormat   = errors.New("webhook wrong format")
 )
 
 func NewWebhook(h ActionHook, action *Action) (Hook, error) {
-	webhookURL := h.Properties["url"]
+	webhookURL := h.Properties["url"].(string)
```
nopcoder, line 48: need to capture if the "string" type matched

actions/webhook.go
```diff
@@ -119,3 +135,28 @@ func (w *Webhook) marshalEventInformation(record graveler.HookRecord) ([]byte, e
...
+	paramsMap, ok := params.(map[string]interface{})
+	if !ok {
+		return nil, fmt.Errorf("unsupported query params: %w", ErrWebhookWrongFormat)
+	}
+
+	res := map[string][]string{}
+	for k, v := range paramsMap {
+		if ar, ok := v.([]interface{}); ok {
+			for _, v := range ar {
+				res[k] = append(res[k], v.(string))
```
nopcoder, line 154: can panic in case the value is not string

actions/webhook.go
```diff
@@ -119,3 +135,28 @@ func (w *Webhook) marshalEventInformation(record graveler.HookRecord) ([]byte, e
...
+			}
+			continue
+		}
+		res[k] = []string{v.(string)}
```
nopcoder, line 158: string check - can panic

nopcoder (approved):

PR #1519, opened 2021-02-24
Enrich webhook output file

### LINKED ISSUE #1517: Enrich webhook output log with more than the resp.body

itaiad200:
closes #1517

### DIFF
--- a/actions/service_test.go
+++ b/actions/service_test.go
@@ -1,6 +1,7 @@
 package actions_test
 
 import (
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
@@ -9,7 +10,6 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"strconv"
-	"strings"
 	"sync/atomic"
 	"testing"
 	"time"
@@ -116,13 +116,13 @@ hooks:
 	testOutputWriter := mock.NewMockOutputWriter(ctrl)
 	expectedHookRunID := "1"
 	var lastManifest *actions.RunManifest
-	var data []byte
+	var writerBytes []byte
 	testOutputWriter.EXPECT().
 		OutputWrite(ctx, record.StorageNamespace.String(), actions.FormatHookOutputPath(record.RunID, expectedHookRunID), gomock.Any(), gomock.Any()).
 		Return(nil).
 		DoAndReturn(func(ctx context.Context, storageNamespace, name string, reader io.Reader, size int64) error {
 			var err error
-			data, err = ioutil.ReadAll(reader)
+			writerBytes, err = ioutil.ReadAll(reader)
 			return err
 		})
 	testOutputWriter.EXPECT().
@@ -233,9 +233,5 @@ hooks:
 		t.Errorf("GetRunResult() result=%v, expected nil", runResult)
 	}
 
-	bytes, err := ioutil.ReadAll(outputReader)
-	require.NoError(t, err)
-	str := string(bytes)
-	rows := strings.Split(str, "\n")
-	require.Greater(t, len(rows), 10)
+	require.Greater(t, bytes.Count(writerBytes, []byte("\n")), 10)
 }
--- a/actions/service_test.go
+++ b/actions/service_test.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"strconv"
+	"strings"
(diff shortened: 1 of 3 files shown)


actions/webhook.go
```diff
@@ -103,6 +104,36 @@ func (w *Webhook) Run(ctx context.Context, record graveler.HookRecord, writer *H
 	return nil
 }
 
+func writeOutput(ctx context.Context, writer *HookOutputWriter, req *http.Request, resp *http.Response, reqBody []byte) error {
```
nopcoder, line 107: https://golang.org/pkg/net/http/httputil
I was thinking we should use dump request and response

actions/service_test.go
```diff
@@ -8,9 +8,12 @@ import (
 	"io/ioutil"
 	"net/http"
 	"net/http/httptest"
+	"strings"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
```
nopcoder, line 15: remove empty line

nopcoder (changes requested):

actions/service_test.go
```diff
@@ -155,4 +161,10 @@ hooks:
 	if runResult != nil {
 		t.Errorf("GetRunResult() result=%v, expected nil", runResult)
 	}
+
+	bytes, err := ioutil.ReadAll(outputReader)
```
nopcoder, line 165: need to capture this one in the DoAndReturn - can't assume the reader will be valid at this point

actions/service_test.go
```diff
@@ -155,4 +161,10 @@ hooks:
...
+	require.NoError(t, err)
+	str := string(bytes)
+	rows := strings.Split(str, "\n")
+	require.Greater(t, len(rows), 10)
```
nopcoder, line 169: You can use bytes.Split if you just want to count rows, and prevent the string()

actions/webhook.go
```diff
@@ -103,6 +112,26 @@ func (w *Webhook) Run(ctx context.Context, record graveler.HookRecord, writer *H
...
+	buf.WriteString("\n\nWebhook response:\n")
+	dumpResp, err := httputil.DumpResponse(resp, true)
+	if err != nil {
+		return err
+	}
+	buf.Write(dumpResp)
+
+	buf.WriteString(fmt.Sprintf("\nRequest duration: %s\n", elapsed))
+
+	return writer.OutputWrite(ctx, buf, int64(buf.Len()))
```
nopcoder, line 132:
```suggestion
	var buf bytes.Buffer
	dumpReq, err := httputil.DumpRequestOut(req, true)
	if err != nil {
		buf.WriteString("Error during request dump: " + err.Error() + "\n")
		return err
	}
	_, err = buf.WriteString("Request:\n")
	if err != nil {
		return err
	}
	_, err = buf.Write(dumpReq)
	if err != nil {
		return err
	}

	dumpResp, err := httputil.DumpResponse(resp, true)
	if err != nil {
		return err
	}
	_, err = buf.WriteString("\n\nResponse:\n")
	if err != nil {
		buf.WriteString("Error during response dump: " + err.Error() + "\n")
		return err
	}
	_, err = buf.Write(dumpResp)
	if err != nil {
		return err
	}

	_, err = buf.WriteString("\nRequest duration: " + elapsed.String() + "\n")
	if err != nil {
		return err
	}
	return writer.OutputWrite(ctx, &buf, int64(buf.Len()))
```

actions/webhook.go
```diff
@@ -80,20 +83,26 @@ func (w *Webhook) Run(ctx context.Context, record graveler.HookRecord, writer *H
...
+	elapsed := time.Since(start)
 	if err != nil {
 		return err
 	}
 	defer func() {
 		_ = resp.Body.Close()
 	}()
 
+	// seeking for logging the request
+	if _, err := reqReader.Seek(0, 0); err != nil {
```
nopcoder, line 99: why?

nopcoder (changes requested):

actions/webhook.go
itaiad200: Did something similar, have a look.

actions/webhook.go
itaiad200: Otherwise `httputil.DumpRequestOut` returns an error.

nopcoder (approved):

PR #1529, opened 2021-02-25
added a stage object operation to the API to allow adding arbitrary object metadata

ozkatz:
Adds the ability to register an external location to a lakefs branch+path.
Expected input is a physical address, size and checksum.
If the physical address is not in the configured block adapter, operation will fail with a 400.

Additinoally, adds a plumbing command `lakectl fs stage` that wraps this API operation.

### DIFF
--- a/api/controller.go
+++ b/api/controller.go
@@ -174,6 +174,7 @@ func (c *Controller) Configure(api *operations.LakefsAPI) {
 	api.ObjectsListObjectsHandler = c.ObjectsListObjectsHandler()
 	api.ObjectsGetObjectHandler = c.ObjectsGetObjectHandler()
 	api.ObjectsUploadObjectHandler = c.ObjectsUploadObjectHandler()
+	api.ObjectsStageObjectHandler = c.ObjectsStageObjectHandler()
 	api.ObjectsDeleteObjectHandler = c.ObjectsDeleteObjectHandler()
 
 	api.MetadataCreateSymlinkHandler = c.MetadataCreateSymlinkHandler()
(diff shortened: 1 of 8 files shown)


ozkatz:
Closes #1530

cmd/lakectl/cmd/fs.go
```diff
@@ -159,6 +159,36 @@ var fsUploadCmd = &cobra.Command{
...
+	Args: cmdutils.ValidationChain(
+		cobra.ExactArgs(1),
+		cmdutils.FuncValidator(0, uri.ValidatePathURI),
+	),
+	Run: func(cmd *cobra.Command, args []string) {
+		client := getClient()
+		pathURI := uri.Must(uri.Parse(args[0]))
+		size, _ := cmd.Flags().GetInt64("size")
+		location, _ := cmd.Flags().GetString("location")
+		cksum, _ := cmd.Flags().GetString("checksum")
```
nopcoder, line 175: wish our adapter can fetch the size and checksum from the location at some point in time - but for basic plumbing, this is enough for now.

docs/reference/commands.md
```diff
@@ -60,6 +60,205 @@ lakectl is a CLI tool allowing exploration and manipulation of a lakeFS environm
...
+
+#### Options
+
+```
+  -h, --help   help for help
+```
+
+
+
+### lakectl actions runs list
```
nopcoder, line 232: Thanks - can add this one in a different PR if needed.

nopcoder (approved):

api/controller_test.go
```diff
@@ -1201,6 +1201,87 @@ func TestController_ObjectsUploadObjectHandler(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		if got.Payload.PhysicalAddress != "s3://another-bucket/some/location" {
+			t.Fatalf("unexpected physical address: %s", got.Payload.PhysicalAddress)
+		}
+	})
+
+	t.Run("upload object missing branch", func(t *testing.T) {
+
```
arielshaqed, line 1252: nit: no blank lines.
```suggestion
```

api/controller_test.go
```diff
@@ -1201,6 +1201,87 @@ func TestController_ObjectsUploadObjectHandler(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if got.Payload.PhysicalAddress != "s3://another-bucket/some/location" {
+			t.Fatalf("unexpected physical address: %s", got.Payload.PhysicalAddress)
+		}
+	})
+
+	t.Run("upload object missing branch", func(t *testing.T) {
```
arielshaqed, line 1251: Can we also test for no such repo?  We've had bugs for being precise on that error, recently #1476.

docs/reference/commands.md
```diff
@@ -60,6 +60,205 @@ lakectl is a CLI tool allowing exploration and manipulation of a lakeFS environm
...
+
+```
+  -h, --help   help for abuse
+```
+
+
+
+### lakectl abuse create-branches
+
+Create a lot of branches very quickly.
```
arielshaqed, line 80:
```suggestion
Create many branches very quickly.
```

swagger.yml
```diff
@@ -2010,6 +2010,36 @@ paths:
         name: path
```
arielshaqed, line 2010: Can we also do the JavaScript version nonsense in this PR?  (`make build` and commit)

swagger.yml
```diff
@@ -2010,6 +2010,36 @@ paths:
...
         required: true
         type: string
+
+    put:
+      tags:
+        - objects
+      operationId: stageObject
+      summary: stage an object's metadata for the given branch
+      parameters:
+        - in: body
```
arielshaqed, line 2020: This body is weirdly-shaped: it has a single field `stats`.

swagger.yml
```diff
@@ -2010,6 +2010,36 @@ paths:
...
+          name: stats
+          schema:
+            $ref: "#/definitions/object_stats"
```
arielshaqed, line 2023: This is too broad an object: it includes `mtime` and `path_type`, which AFAICT will (and should!) be ignored.  And it can keep on growing, meaning that all clients will need to keep on growing.

arielshaqed (approved):
Unhappy with the Swagger API.

I'm also a bit worried about the potential for uploading the wrong ETag.  We may want to verify it.  But since that _anyway_ will have to be optional, it's not required in this PR.

Generally we might want to document a reminder that any staged objects should be treated as immutable.

api/controller.go
```diff
@@ -1502,6 +1503,70 @@ func (c *Controller) ObjectsListObjectsHandler() objects.ListObjectsHandler {
...
+		if qk.StorageType.String() != deps.BlockAdapter.BlockstoreType() {
+			return objects.NewStageObjectBadRequest().WithPayload(
+				responseError("invalid storage type: %s: current block adapter is %s",
+					qk.StorageType.String(),
+					deps.BlockAdapter.BlockstoreType(),
+				))
+		}
+
+		writeTime := time.Now()
+		entry := catalog.DBEntry{
```
itaiad200, line 1542: What about `metadata`? I think it's valuable for these kinds of advance-use commands.

cmd/lakectl/cmd/fs.go
```diff
@@ -159,6 +159,36 @@ var fsUploadCmd = &cobra.Command{
 	},
 }
 
+var fsStageCmd = &cobra.Command{
+	Use:    "stage <path uri>",
+	Short:  "stage an object at the specified URI",
```
johnnyaug, line 164: Please explain better what this command does: adds a reference / imports an existing object, and allows it be managed in lakeFS.

api/controller.go
ozkatz: Metadata apparently gets dropped when writing to staging since that field is never serialized.Will open a separate issue.

johnnyaug (approved):

api/controller.go
ozkatz: it is not. Added metadata support to API and CLI

PR #1533, opened 2021-02-27
Action hook run results fix listing hooks order

### LINKED ISSUE #1532: Actions list hook run results order
Hook results are sorted by hook run ID.  The ID is composed of a timestamp and a random part.  Because we allocate the running hooks information at the same time - all same the same prefix and the sort order is based on the random part.
Using the order the allocate the tasks for execution or the actual time will produce a better sort order for listing these results.

nopcoder:
Fix #1532

Remove the random part from the hook run ID, in order to get a consistent and more meaningful order of the list of hook results.
Kept the task allocation and ID assignment at the same place - used a sequence number to assign IDs.

### DIFF
--- a/actions/service.go
+++ b/actions/service.go
@@ -18,10 +18,10 @@ import (
 )
 
 type Service struct {
-	DB             db.Database
-	Source         Source
-	Writer         OutputWriter
-	RunIDGenerator func() string
+	DB                 db.Database
+	Source             Source
+	Writer             OutputWriter
+	HookRunIDGenerator func(id int) string
 }
 
 type Task struct {
@@ -85,10 +85,10 @@ var ErrNotFound = errors.New("not found")
 
 func NewService(db db.Database, source Source, writer OutputWriter) *Service {
 	return &Service{
-		DB:             db,
-		Source:         source,
-		Writer:         writer,
-		RunIDGenerator: graveler.NewRunID,
+		DB:                 db,
+		Source:             source,
+		Writer:             writer,
+		HookRunIDGenerator: NewHookRunID,
 	}
 }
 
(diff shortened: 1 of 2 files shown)


actions/service.go
```diff
@@ -421,3 +423,8 @@ func (s *Service) PostMergeHook(ctx context.Context, record graveler.HookRecord)
 	// post merge actions will be enabled here
 	return nil
 }
+
+func NewHookRunID(id int) string {
+	tm := time.Now().UTC().Format("02150405")
+	return fmt.Sprintf("H%s%06d", tm, id)
```
itaiad200, line 429: What does the `H` stand for?

actions/service.go
```diff
@@ -421,3 +423,8 @@ func (s *Service) PostMergeHook(ctx context.Context, record graveler.HookRecord)
 	// post merge actions will be enabled here
 	return nil
 }
+
+func NewHookRunID(id int) string {
+	tm := time.Now().UTC().Format("02150405")
```
itaiad200, line 428: You don't use the same time for the tasks, so we already have prefix which is sorted and unique, no?

itaiad200 (changes requested):

actions/service.go
nopcoder: Will remove it - used it first to identify the hook run id

actions/service.go
nopcoder: right, f2f chat - will use action and hook idx to identify the specific task

nopcoder: went with 4 digits for each range - just in case.

itaiad200 (approved):

PR #1542, opened 2021-02-28
Merge error rendered as alert with download hook output

### LINKED ISSUE #1524: Merge UI error should not use alert
Render proper error UI message in case of a merge error and not use browser's alert().
The common way to produce this one is by failed pre-merge hook.

### LINKED ISSUE #1540: Actions: lakectl describe output improvments
1. Information should be rendered using tables
2. Field names should be highlighted
3. Status value should be red/green
4. Remove "Output:" title before hook output

nopcoder:
Close #1524


itaiad200 (approved):
LGTM
Nice work!

Issue #1369, opened 2021-01-31
Export gone with Rocks


itaiad200:
Export no longer works with rocks implementation.

itaiad200:
Dup of #1567

PR #1543, opened 2021-02-28
remove old treeverse-distcp from docs and its extraction manifest

### LINKED ISSUE #1497: Remove treeverse-distcp
treeverse-distcp does not work with rocks, we need to remove it from the lakeFS documentation and remove the repository.

ozkatz:
closes #1497


arielshaqed (approved):
Note that we still do not have #1369.  Soon...

arielshaqed:
Also note that this is *not* continuous export, but rather the "get me out of here!"  card.  To replace it the "export" Spark(?) job will need to have a way to specify a metarange ID manually.

guy-har (approved):
Thanks,
What about https://github.com/treeverse/treeverse-distcp?
Should we make unavailable?

PR #1545, opened 2021-03-01
Actions listing pagination fix result amout

### LINKED ISSUE #1546: Actions: listing results pagination only if there are two items left
Listing reports have more when there is additional information that didn't cut to the amount requested.
Actions runs result and hook results requires at least two items to report "has more"

nopcoder:
Fix #1546

The limit check was done after getting the next element while iterating.


itaiad200 (approved):
Really nice catch!
Is there an easy way to add a UT to catch this for next time?

PR #1555, opened 2021-03-02
Convert merge errors to bad requests

### LINKED ISSUE #1547: DirtyBranch renders as internal server error
Returning ErrDirtyBranch, currently part of Merge and Revert rendered as HTTP status code 500 with an internal server error.
It is a UX that the end-user will have a hard time understanding the root cause.

itaiad200:
closes #1547

### DIFF
--- a/api/controller.go
+++ b/api/controller.go
@@ -959,19 +959,19 @@ func (c *Controller) MergeMergeIntoBranchHandler() refs.MergeIntoBranchHandler {
 		case err == nil:
 			payload := newMergeResultFromCatalog(res)
 			return refs.NewMergeIntoBranchOK().WithPayload(payload)
-		case errors.Is(err, catalog.ErrUnsupportedRelation):
-			return refs.NewMergeIntoBranchDefault(http.StatusInternalServerError).WithPayload(responseError("branches have no common base"))
 		case errors.Is(err, catalog.ErrBranchNotFound) || errors.Is(err, graveler.ErrBranchNotFound):
-			return refs.NewMergeIntoBranchDefault(http.StatusInternalServerError).WithPayload(responseError("a branch does not exist "))
+			return refs.NewMergeIntoBranchBadRequest().WithPayload(responseError("a branch does not exist"))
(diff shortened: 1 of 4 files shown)


api/controller.go
```diff
@@ -959,19 +959,19 @@ func (c *Controller) MergeMergeIntoBranchHandler() refs.MergeIntoBranchHandler {
...
 		case errors.Is(err, catalog.ErrConflictFound) || errors.Is(err, graveler.ErrConflictFound):
 			payload := newMergeResultFromCatalog(res)
 			return refs.NewMergeIntoBranchConflict().WithPayload(payload)
 		case errors.Is(err, catalog.ErrNoDifferenceWasFound) || errors.Is(err, graveler.ErrNoChanges):
-			return refs.NewMergeIntoBranchDefault(http.StatusInternalServerError).WithPayload(responseError("no difference was found"))
+			return refs.NewMergeIntoBranchBadRequest().WithPayload(responseError("no difference was found"))
 		case errors.Is(err, graveler.ErrLockNotAcquired):
 			return refs.NewMergeIntoBranchDefault(http.StatusInternalServerError).WithPayload(responseError("branch is currently locked, try again later"))
+		case errors.Is(err, graveler.ErrDirtyBranch):
+			return refs.NewMergeIntoBranchBadRequest().WithPayload(responseError("can't merge into a dirty branch"))
```
guy-har, line 972: I think this one should return `409 conflict`

guy-har (approved):

api/controller.go
itaiad200: I think it's confusing to overload the term `conflict`. Nothing really "conflicts" here, it's more of a limitation we introduced and might remove later.

guy-har: I understand, The thing is that Bad request is used for client error. I don't think that this is a client error. It seems like it fits the documentation of `409 conflict`

itaiad200: But it returns a `merge_result` with conflicts. Should we return an empty object in that case?
Or an empty body? 2 response codes with different payloads is error prune IMO.

PR #1562, opened 2021-03-03
Killed `WithContext` in API, DB and all `block.Adapter`s

ozkatz:


auth/service_test.go
```diff
@@ -103,10 +105,11 @@ func userWithPolicies(t testing.TB, s auth.Service, policies []*model.Policy) st
 }
 
 func TestDBAuthService_ListPaged(t *testing.T) {
+	ctx := context.Background()
 	const chars = "abcdefghijklmnopqrstuvwxyz"
 	adb, _ := testutil.GetDB(t, databaseURI)
 	type row struct{ A string }
-	if _, err := adb.Exec(`CREATE TABLE test_pages (a text PRIMARY KEY)`); err != nil {
+	if _, err := adb.Exec(context.Background(), `CREATE TABLE test_pages (a text PRIMARY KEY)`); err != nil {
```
nopcoder, line 112:
```suggestion
	if _, err := adb.Exec(ctx, `CREATE TABLE test_pages (a text PRIMARY KEY)`); err != nil {
```

nopcoder (approved):
Great, refactor!
Two minor comments

guy-har (approved):
Amazing!
Noticed that we still have usage `background.context` in `database.go` and in `build.go`. Think we should remove them or comment why we are using them over there. It doesn't have to be part of this PR.

nopcoder (approved):

Issue #1480, opened 2021-02-18
Hooks: UI to list and output actions and hooks


ozkatz:
- [ ] Actions runs tab - listing all, branch or commit
- [ ] Each run in the Actions tab cloud be clicked and render the same information found in our ctl describe
- [ ] Commit completion - success or failure should disable the action information associated and link to the run in the Actions tab
- [ ] Merge completion - success or failure should disable the action information associated and link to the run in the Actions tab

itaiad200:
completed with #1584

PR #1564, opened 2021-03-03
Actions tab UI (hook runs)

nopcoder:
Part of #1480

Actions tab - lists actions runs with log of each hook execution.

Doesn't include filter on commit or branch


itaiad200 (approved):
Tested locally, looks really good!

PR #1576, opened 2021-03-06
Remove entry catalog

ozkatz:
This seems to work OK but tests aren't passing: seems that I've been able to reincarnate #1349 as I'm seeing the exact same errors happening in all the same places.

For reference: the only change I made that I can think about having anything to do with Pebble's cache, was moving its initialization from the now non-existent EntryCatalog to RocksCatalog, both doing the exact same thing and both `defer`ing an `Unref()` for it.


catalog/catalog.go
```diff
@@ -0,0 +1,1099 @@
...
+)
+
+var ErrUnknownDiffType = errors.New("unknown graveler difference type")
+
+func New(ctx context.Context, cfg Config) (*Catalog, error) {
+	if cfg.LockDB == nil {
+		cfg.LockDB = cfg.DB
+	}
+
+	tierFSParams, err := cfg.Config.GetCommittedTierFSParams(ctx)
```
nopcoder, line 125: not related to this pr - think we need to extract this one out.  The Config should hold everything needed in order to instantiate a new catalog.

catalog/catalog.go
```diff
@@ -0,0 +1,1099 @@
...
+}
+
+func (c *Catalog) Close() error {
+	for _, manager := range c.managers {
+		err := manager.Close()
+		if err != nil {
+			return err
+		}
+	}
+	return nil
```
nopcoder, line 1057: go-multierror - we like to close them all

graveler/sstable/range_manager.go
```diff
@@ -110,8 +101,8 @@ func (m *RangeManager) GetValueGE(ctx context.Context, ns committed.Namespace, i
 	}, nil
 }
 
-// GetEntry returns the entry matching the path in the SSTable referenced by the id.
-// If path not found, (nil, ErrPathNotFound) is returned.
+// GetValue returns the Record matching the key in the SSTable referenced by the id.
+// If key is not found, (nil, ErrKeyNotFound) is returned.
```
nopcoder, line 105: nice

nopcoder (changes requested):

ozkatz:
@nopcoder fixed all comments, PTAL

nopcoder (approved):

PR #1585, opened 2021-03-08
Adding arm64 to our release binaries

nopcoder:


johnnyaug (approved):

arielshaqed:
We should probably test on `arm64`.  Otherwise... surprises.

nopcoder:
The drive was to have something working on M1 -- so I just run some basic commands to see that it is working natively.
For using lakeFS on arm64 we can use AWS Linux for testing.

PR #1603, opened 2021-03-10
Check context during apply, diff and merge loops

### LINKED ISSUE #1598: Graveler - Handle context cancel in Merge and Diff
Merge and Diff keep on running in case context is cancelled.

We would like them to exit on context cancel.

nopcoder:
Fix #1598


graveler/committed/apply.go
```diff
@@ -98,6 +110,12 @@ func Apply(ctx context.Context, writer MetaRangeWriter, source Iterator, diffs g
 	haveSource, haveDiffs := source.Next(), diffs.Next()
 	changed := false
 	for haveSource && haveDiffs {
+		select {
+		case <-ctx.Done():
+			return ret, ctx.Err()
```
ozkatz, line 115: How was this tested? (question applies to all other places where this type of check has been introduced..)

graveler/committed/apply.go
```diff
@@ -64,9 +70,15 @@ func applyFromSource(logger logging.Logger, writer MetaRangeWriter, source Itera
 
 // applyFromDiffs applies all changes from diffs to writer and returns the number elements it
 // added.
-func applyFromDiffs(logger logging.Logger, writer MetaRangeWriter, diffs graveler.ValueIterator) (int, error) {
+func applyFromDiffs(ctx context.Context, logger logging.Logger, writer MetaRangeWriter, diffs graveler.ValueIterator) (int, error) {
 	numAdded := 0
 	for {
+		select {
+		case <-ctx.Done():
+			return 0, ctx.Err()
```
ozkatz, line 78: How was this tested?

ozkatz (changes requested):
Looks great overall, but these changes come with no test coverage so kind of hard to validate correctness (now and in future changes).

graveler/committed/apply.go
nopcoder: added test code

ozkatz (approved):
Looks great!

arielshaqed:
As part of reviewing #1652, I discovered that I don't understand this PR.  It seems like polling for context cancellation throughout work that is not CPU intensive.  I would *expect* the iterators to receive context and pipe it through to some RocksDB SSTable iterator, which would handle it at some stage.

If the SSTable RocksDB iterators _don't_ handle cancellation (maybe because they cannot pipe it through to IO, because of the whole Go IO / os.File no-context fiasco), then I would expect to handle this at the iterator level.

Handling it here in `Apply` means that we cannot make use of some hypothetical better underlying iterator in the future, and also that every operation that does _not_ use `Apply` cannot be cancelled.

nopcoder:
It does polling the context for cancelation in order to support cancellation regardless of the iterator's implementation.

Why the extra check prevents better underlying interator? the iterator can cancel by context if needed. The concern, for example, is if you wrap the implementation with a cache that will iterator the data from memory without calling any underlying API that will check the content.

We can assign the responsibility of context cancellation on the underlying but need to verify in some way that all iterators either check the context or call an API that does that.

PR #1609, opened 2021-03-10
lakectl fs ls fix default delimiter and support recursive

### LINKED ISSUE #1590: lakectl fs ls ignores delimiters
currently the command `lakectl fs ls` returns all keys, and does not support common-prefixes.
for example if we are using the delimiter `/` and have the keys 
`/foo/bar`
`/foo/bar-2`
`/bar`

running `lakectl fs ls` on root should return 
`/foo/`
`/bar`

nopcoder:
Fix #1590


guy-har (approved):
Looks good

PR #1618, opened 2021-03-12
batch hot paths for a very short duration

ozkatz:
Looking at the access pattern for critical path operations, we mostly call PostgreSQL with the same exact queries many times.

For GetObject, StatObject, ListObjects (which probably make up the majority of data lake calls), we ALWAYS start by doing the same set of roundtrips to PG:

- Get the repository (to extract the storage namespace)
- Resolve the ref (i.e. figure out if this is a commit/commit prefix/branch/tag)
- Resolve the underlying commit ID (if a branch, prefix or tag)

Our access pattern is such that many requests at a given point in time are extremely likely to not only share the same repository details, but also the same branch/commit/tag, as big data systems tend to be bursty in nature.

Since caching is not an option since it sacrifices consistency (in the sense that reading after a successful write returns - might return a stale value), instead of keeping the result around for a while, we can keep the requests around for a while.

This is what this PR does: for a given type of request (i.e. to a specific branch/repo/tag, etc), wait a couple of milliseconds: if other identical requests arive in that time, do a single roundtrip and return the results once for all those requests.

Testing this on the same environment used for the sizing guide (2 x c5ad.xlarge AWS instances), I now get the following results:

- lakectl abuse random-read on a commit ID: throughput goes up from **8-10k requests/second** to **45k requests/second**
- Amount of DB queries drops from ~20k/s to less than 1k/s.

I'm OK with not accepting this due to it being a premature optimization (it is!), but I feel the added complexity is relatively small and the gain is pretty big (if only to show better numbers per core as possible).

### DIFF
--- a/pkg/batch/executor.go
+++ b/pkg/batch/executor.go
@@ -1,0 +1,87 @@
+package batch
+
+import (
+	"time"
+
+	"github.com/treeverse/lakefs/pkg/logging"
+)
+
+type BatchFn func() (interface{}, error)
+
+type Batcher interface {
+	BatchFor(key string, dur time.Duration, fn BatchFn) (interface{}, error)
+}
+
+type response struct {
+	v   interface{}
+	err error
+}
+
+type request struct {
+	key              string
+	dur              time.Duration
+	fn               BatchFn
+	responseCallback chan *response
+}
+
+type Executor struct {
+	requests chan *request
+	execs    chan string
+	keys     map[string][]*request
+	logger   logging.Logger
+}
+
+func NewExecutor(logger logging.Logger) *Executor {
+	e := &Executor{
+		requests: make(chan *request),
+		execs:    make(chan string),
+		keys:     make(map[string][]*request),
+		logger:   logger,
+	}
+	go e.Run() // TODO(ozkatz): should probably be managed by the user (also, allow stopping it)
+	return e
+}
+
+func (e *Executor) BatchFor(key string, dur time.Duration, fn BatchFn) (interface{}, error) {
+	cb := make(chan *response)
+	e.requests <- &request{
+		key:              key,
+		dur:              dur,
+		fn:               fn,
+		responseCallback: cb,
+	}
+	response := <-cb
+	return response.v, response.err
+}
+
+func (e *Executor) Run() {
+	for {
+		select {
+		case req := <-e.requests:
+			// see if we have it scheduled already
+			if _, exists := e.keys[req.key]; !exists {
+				// this is a new key, let's fire a timer for it
+				go func(req *request) {
+					time.Sleep(req.dur)
+					e.execs <- req.key
+				}(req)
+			}
+			e.keys[req.key] = append(e.keys[req.key], req)
+		case execKey := <-e.execs:
+			// let's take all callbacks
+			waiters := e.keys[execKey]
+			delete(e.keys, execKey)
+			go func(key string) {
+				// execute and call all mapped callbacks
+				v, err := waiters[0].fn()
+				e.logger.WithFields(logging.Fields{
+					"waiters": len(waiters),
+					"key":     key,
+				}).Trace("dispatched BatchFn")
+				for _, waiter := range waiters {
+					waiter.responseCallback <- &response{v, err}
+				}
+			}(execKey)
+		}
+	}
+}
--- a/pkg/batch/executor.go
+++ b/pkg/batch/executor.go
@@ -1,6 +1,7 @@
 package batch
 
 import (
+	"context"
 	"time"
 
 	"github.com/treeverse/lakefs/pkg/logging"
@@ -8,8 +9,17 @@ import (
 
 type BatchFn func() (interface{}, error)
 
+type DelayFn func(dur time.Duration)
+
 type Batcher interface {
 	BatchFor(key string, dur time.Duration, fn BatchFn) (interface{}, error)
+}
+
+type nonBatchingExecutor struct {
+}
+
+func (n *nonBatchingExecutor) BatchFor(key string, dur time.Duration, fn BatchFn) (interface{}, error) {
+	return fn()
 }
 
 type response struct {
@@ -28,18 +38,22 @@ type Executor struct {
 	requests chan *request
 	execs    chan string
 	keys     map[string][]*request
-	logger   logging.Logger
+	Logger   logging.Logger
+	Delayer  DelayFn
+}
+
+func NopExecutor() *nonBatchingExecutor {
+	return &nonBatchingExecutor{}
 }
 
 func NewExecutor(logger logging.Logger) *Executor {
-	e := &Executor{
+	return &Executor{
 		requests: make(chan *request),
 		execs:    make(chan string),
 		keys:     make(map[string][]*request),
-		logger:   logger,
+		Logger:   logger,
+		Delayer:  time.Sleep,
 	}
-	go e.Run() // TODO(ozkatz): should probably be managed by the user (also, allow stopping it)
-	return e
 }
 
 func (e *Executor) BatchFor(key string, dur time.Duration, fn BatchFn) (interface{}, error) {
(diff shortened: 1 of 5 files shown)


nopcoder:
Think we can we can simplify it a bit more while using `singleflight` package + wrapping the function call with the same delay before calling the actual function that fetches the data.

ozkatz:
> Think we can we can simplify it a bit more while using `singleflight` package + wrapping the function call with the same delay before calling the actual function that fetches the data.

@nopcoder sounds nice! Feel free to give that a go 🙂

nopcoder:
> > Think we can we can simplify it a bit more while using `singleflight` package + wrapping the function call with the same delay before calling the actual function that fetches the data.
> 
> @nopcoder sounds nice! Feel free to give that a go 🙂

https://github.com/treeverse/lakeFS/pull/1620

Need to set up a test env for testing the above.

ozkatz:
> > > Think we can we can simplify it a bit more while using `singleflight` package + wrapping the function call with the same delay before calling the actual function that fetches the data.
> > 
> > 
> > @nopcoder sounds nice! Feel free to give that a go slightly_smiling_face
> 
> #1620
> 
> Need to set up a test env for testing the above.

Used the same env to test your branch - behaves just the same (~45k requests/second). I agree your implementation is simpler.

pkg/batch/executor.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func NewExecutor(logger logging.Logger) *Executor {
+	e := &Executor{
+		requests: make(chan *request),
+		execs:    make(chan string),
+		keys:     make(map[string][]*request),
+		logger:   logger,
+	}
+	go e.Run() // TODO(ozkatz): should probably be managed by the user (also, allow stopping it)
```
nopcoder, line 41: You can move this one into Run() with defer close, so we will not require to add Stop() and/or Close() methods to handle this resource.

pkg/graveler/ref/manager.go
```diff
@@ -3,8 +3,13 @@ package ref
...
 	"context"
 	"errors"
+	"fmt"
 	"time"
 
+	"github.com/treeverse/lakefs/pkg/logging"
+
+	"github.com/treeverse/lakefs/pkg/batch"
+
 	"github.com/treeverse/lakefs/pkg/db"
```
nopcoder, line 13: remove blank lines

pkg/batch/executor.go
```diff
@@ -0,0 +1,87 @@
...
+}
+
+func (e *Executor) Run() {
+	for {
+		select {
+		case req := <-e.requests:
+			// see if we have it scheduled already
+			if _, exists := e.keys[req.key]; !exists {
+				// this is a new key, let's fire a timer for it
+				go func(req *request) {
```
nopcoder, line 64: 1. Add WaitGroup that will count ongoing goroutines.
2. Add Close() method to Executor to wait for the wait group done.
3. Close() should also train execs and call the responseCallback(s)

pkg/batch/executor.go
```diff
@@ -0,0 +1,87 @@
...
+					time.Sleep(req.dur)
+					e.execs <- req.key
+				}(req)
+			}
+			e.keys[req.key] = append(e.keys[req.key], req)
+		case execKey := <-e.execs:
+			// let's take all callbacks
+			waiters := e.keys[execKey]
+			delete(e.keys, execKey)
+			go func(key string) {
```
nopcoder, line 74: pass waiters as you pass the key - just to be symmetric

pkg/batch/executor.go
```diff
@@ -0,0 +1,87 @@
...
+				// execute and call all mapped callbacks
+				v, err := waiters[0].fn()
```
nopcoder, line 76: Suggestion: will probably will like to capture this one inside a func with recover that will return an error

nopcoder (changes requested):

ozkatz:
@itaiad200 please see the tests I added: I attempted to prove this method does not violate read-after-write consistency

nopcoder (approved):

pkg/graveler/ref/manager.go
```diff
@@ -13,26 +15,36 @@ import (
 // IteratorPrefetchSize is the amount of records to maybeFetch from PG
 const IteratorPrefetchSize = 1000
 
+const MaxBatchDelay = time.Millisecond * 3
```
itaiad200, line 18: How did we reach this number?
Maybe just add a comment describing it's a guesstimate.

pkg/batch/executor.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+func NopExecutor() *nonBatchingExecutor {
+	return &nonBatchingExecutor{}
+}
+
+func NewExecutor(logger logging.Logger) *Executor {
+	return &Executor{
+		requests: make(chan *request),
+		execs:    make(chan string),
```
itaiad200, line 52: If you'll create a buffered channel (i.e. size more than 1), logic will still be consistent but you'll have less open go routines.

pkg/catalog/catalog.go
```diff
@@ -154,20 +170,25 @@ func New(ctx context.Context, cfg Config) (*Catalog, error) {
 		sstableManager,
 	)
 	if err != nil {
+		cancelFn()
```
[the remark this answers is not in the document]
ozkatz: true, but the cancelFn must be called - it's a leak otherwise: from https://golang.org/pkg/context/#WithCancel -

> Canceling this context releases resources associated with it, so code should call cancel as soon as the operations running in this Context complete.

pkg/batch/executor_test.go
```diff
@@ -0,0 +1,85 @@
...
+	"testing"
+	"time"
+
+	"github.com/treeverse/lakefs/pkg/batch"
+	"github.com/treeverse/lakefs/pkg/logging"
+)
+
+func TestExecutor_BatchFor(t *testing.T) {
+	// Setup executor
+	exec := batch.NewExecutor(logging.Dummy())
```
arielshaqed, line 16: I'd actually _like_ to emit the trace logs during the test.

pkg/batch/executor_test.go
```diff
@@ -0,0 +1,85 @@
...
+	go exec.Run(context.Background())
+	// Prove the executor does not violate read-after-write consistency.
+	// First, let's define read-after-write consistency:
+	// 	Any read that started after a successful write has returned, must return the updated value.
+	// to test this, let's simulate the following scenario:
```
arielshaqed, line 21: mega-nit:
```suggestion
	// To test this, simulate the following scenario:
```

pkg/batch/executor_test.go
```diff
@@ -0,0 +1,85 @@
...
+	read2Done := make(chan bool)
+
+	// we pass a custom delay func that ensures we make the write only after
+	//  reader1 started
+	waitWrite := make(chan bool)
+	delays := int32(0)
+	delayFn := func(dur time.Duration) {
+		delaysDone := atomic.AddInt32(&delays, 1)
+		if delaysDone == 1 {
+			close(waitWrite)
```
arielshaqed, line 41: Note that the write can occur before https://github.com/treeverse/lakeFS/pull/1618/files#diff-c9e7aae146c0798d32ade9be6fa5013612246e323b7a2796dbcdc83a0151c607R82 ever happens (because the scheduler is evil).  I think you may need to wait on *another* channel here, that writer1 will close after it _does_ write.

pkg/catalog/catalog.go
```diff
@@ -115,13 +117,25 @@ const (
 
 var ErrUnknownDiffType = errors.New("unknown graveler difference type")
 
+type ctxCloser struct {
+	fn context.CancelFunc
```
arielshaqed, line 121: I may prefer:
```suggestion
	close context.CancelFunc
```
describes what it does, not its type.

pkg/graveler/ref/manager.go
arielshaqed: I'd prefer it to be a configurable parameter, if we believe it can be tuned to affect performance.

arielshaqed (commented):
Thanks!

Most requested is the change to the test, to ensure reader1 actually starts waiting after writer1 writes.

pkg/batch/executor.go
```diff
@@ -0,0 +1,103 @@
...
+}
+
+type request struct {
+	key              string
+	dur              time.Duration
+	fn               BatchFn
+	responseCallback chan *response
+}
+
+type Executor struct {
```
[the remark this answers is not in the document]
ozkatz: I'm bad at naming, I'll admit to that. Suggestions are welcome :)

itaiad200 (approved):

PR #1620, opened 2021-03-12
Using singleflight for batching graveler db reads

nopcoder:


pkg/graveler/ref/manager.go
```diff
@@ -3,36 +3,59 @@ package ref
...
 	"github.com/treeverse/lakefs/pkg/db"
 	"github.com/treeverse/lakefs/pkg/graveler"
 	"github.com/treeverse/lakefs/pkg/ident"
+	"golang.org/x/sync/singleflight"
 )
 
 // IteratorPrefetchSize is the amount of records to maybeFetch from PG
 const IteratorPrefetchSize = 1000
 
+// MaxBatchDelay wait time before perform database read operation, it enables batching requests at that period of time
```
arielshaqed, line 18: nit:
```suggestion
// MaxBatchDelay is the time to wait before performing a database read operation, batching requests during that time
```

pkg/graveler/ref/manager.go
```diff
@@ -3,36 +3,59 @@ package ref
...
+const MaxBatchDelay = 3 * time.Millisecond
```
arielshaqed, line 19: Consider
```suggestion
const BatchDelay = 3 * time.Millisecond
```
The "max" is a bit confusing.

pkg/graveler/ref/manager.go
```diff
@@ -3,36 +3,59 @@ package ref
...
-		err := tx.Get(repository,
-			`SELECT storage_namespace, creation_date, default_branch FROM graveler_repositories WHERE id = $1`,
-			repositoryID)
-		if err != nil {
-			return nil, err
+func batchDelayCall(ctx context.Context, fn func() (interface{}, error)) func() (interface{}, error) {
+	return func() (interface{}, error) {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
```
arielshaqed, line 38: Neat!

arielshaqed (commented):
Very nice implementation.  I particularly like that batching is separate from the query.  So it should be easy to switch it over to the upcoming Ref Manager X (which will also benefit from this batching!).

pkg/graveler/ref/manager.go
```diff
@@ -277,17 +309,20 @@ func (m *Manager) GetCommitByPrefix(ctx context.Context, repositoryID graveler.R
 }
 
 func (m *Manager) GetCommit(ctx context.Context, repositoryID graveler.RepositoryID, commitID graveler.CommitID) (*graveler.Commit, error) {
-	commit, err := m.db.Transact(ctx, func(tx db.Tx) (interface{}, error) {
-		var rec commitRecord
-		err := tx.Get(&rec, `
+	key := fmt.Sprintf("GetCommitByPrefix:%s:%s", repositoryID, commitID)
```
itaiad200, line 312:
```suggestion
	key := fmt.Sprintf("GetCommit:%s:%s", repositoryID, commitID)
```

nopcoder: Fixed, thanks!

itaiad200:
@nopcoder @ozkatz I think that batching requests will break consistency, at least at some level.

Request GET-1 reads value v0 from the DB and lingers on the way back.
Request WRITE-1 writes value v1 to the DB and returns to the client.
Request GET-2 tries to read the value, but sees an ongoing GET-1 request, so it wait for it to return.
When GET-1 finally returns, both GET-1 and GET-2 will return v0 to the client. While the result is valid for GET-1 (request started before the WRITE-1, so both values are ok here), it's invalid for GET-2 as the request started after WRITE-1 succeeded.

nopcoder:
@itaiad200 is right and the consistency is broken - close this one. thanks all for your time and insights!

PR #1631, opened 2021-03-15
merge wip - draft

guy-har:


pkg/graveler/committed/merge.go
```diff
@@ -0,0 +1,308 @@
...
+		switch state {
+		case mergeStateRangeAddedBySource:
+			ret.Count[graveler.DiffTypeAdded] += int(sourceRange.Count)
+			sourceNext = sourceIter.NextRange()
+			err = writer.WriteRange(*sourceRange)
+		case mergeStateRangeAddedByDest:
+			destNext = destIter.NextRange()
+			err = writer.WriteRange(*destRange)
+		case mergeStateRangeNoChange:
+			destNext = destIter.NextRange()
```
arielshaqed, line 68: Shouldn't this also skip source?
```suggestion
			sourceNext = sourceIter.NextRange()
			destNext = destIter.NextRange()			
```

pkg/graveler/committed/merge.go
```diff
@@ -0,0 +1,308 @@
...
+		case record != nil && bytes.Compare(key, record.Key) <= 0:
+			return nil
+		default:
+			hasMore = iter.Next()
+		}
+
+	}
+	return iter.Err()
+}
+func getState(baseIter Iterator, sourceValue *graveler.ValueRecord, sourceRange *Range, destValue *graveler.ValueRecord, destRange *Range) (mergeState, error) {
```
arielshaqed, line 228: This function does change state, the name "get" is misleading.  Perhaps it should be called something like `compareAndMoveBase`.

pkg/graveler/committed/merge.go
```diff
@@ -0,0 +1,308 @@
...
+		return compareResultRangeDestBeforeValue
+	case destValue != nil && bytes.Compare(destValue.Key, sourceRange.MinKey) < 0:
+		return compareResultValueDestBefore
+	case destValue != nil && bytes.Compare(sourceRange.MaxKey, destValue.Key) < 0:
+		return compareResultRangeSourceBefore
+	case destValue != nil:
+		return compareResultRangeSourceBeforeValue
+	case bytes.Compare(sourceRange.MaxKey, destRange.MinKey) < 0:
+		return compareResultRangeSourceBefore
+	case bytes.Compare(destRange.MaxKey, sourceRange.MinKey) < 0:
```
arielshaqed, line 181: This repeats the previous comparison; please break the single switch, perform the comparison once into a variable and switch on that variable.

pkg/graveler/committed/merge.go
```diff
@@ -0,0 +1,308 @@
...
+		return compareValues(sourceValue, destValue)
+	case sourceValue != nil && bytes.Compare(sourceValue.Key, destRange.MinKey) < 0:
+		return compareResultValueSourceBefore
+	case sourceValue != nil && bytes.Compare(destRange.MaxKey, sourceValue.Key) < 0:
+		return compareResultRangeDestBefore
+	case sourceValue != nil:
+		return compareResultRangeDestBeforeValue
+	case destValue != nil && bytes.Compare(destValue.Key, sourceRange.MinKey) < 0:
+		return compareResultValueDestBefore
+	case destValue != nil && bytes.Compare(sourceRange.MaxKey, destValue.Key) < 0:
```
arielshaqed, line 175: This repeats the previous comparison; perform the comparison once into a variable and switch on that variable.  (Might be easier to split it off into a function).

pkg/graveler/committed/merge.go
```diff
@@ -0,0 +1,308 @@
...
+		return compareResultRangeConflict
+		// ranges are overlapping with different IDs
+	}
+}
+
+func proceedToRange(iter Iterator, key Key) error {
+	hasMore := true
+	for hasMore {
+		_, rng := iter.Value()
+		if rng == nil {
```
arielshaqed, line 195: Not sure I understand.  The range should never be nil (unless the iterator has ended), https://github.com/treeverse/lakeFS/blob/84d1164f70afebe3249872793462323959b0e824/pkg/graveler/committed/meta_range.go#L19.  Perhaps here (and elsewhere?) should be
```suggestion
		value, rng := iter.Value()
		if value == nil {
```
?

pkg/graveler/committed/merge.go
```diff
@@ -0,0 +1,308 @@
...
+			hasMore = iter.Next() // TODO(Guys): change
+		}
+		if rng == nil || bytes.Compare(key, rng.MaxKey) <= 0 {
+			return nil
+		}
+		hasMore = iter.NextRange()
```
arielshaqed, line 201: Doing it this way (compared to seeking) means that we always deserialize every range of base.  That may be expensive.

If you don't want to change the iterator interface, you might instead `iter.SeekGE(key)` to get the range, then seek the iterator to a bit before that range (i.e. subtract 1 from the returned range.MaxKey and pad with lots of 0xFF bytes), and advance to the end of _that_ range.  Still gonna be expensive, but I reckon it'll be a bit faster than churning and deserializing a few megabytes.

pkg/graveler/committed/merge.go
```diff
@@ -0,0 +1,308 @@
...
+	}
+	return iter.Err()
+}
+
+func proceedToValue(iter Iterator, key Key) error {
```
arielshaqed, line 206: AFAICT this is a linear-time version of SeekGE.

arielshaqed (commented):
Some comments from going over the code.

pkg/graveler/committed/merge.go
```diff
@@ -0,0 +1,308 @@
...
+			err = writer.WriteRange(*sourceRange)
+		case mergeStateRangeAddedByDest:
+			destNext = destIter.NextRange()
+			err = writer.WriteRange(*destRange)
+		case mergeStateRangeNoChange:
+			destNext = destIter.NextRange()
+			err = writer.WriteRange(*destRange)
+		case mergeStateRangeRemovedBySource:
+			ret.Count[graveler.DiffTypeRemoved] += int(sourceRange.Count)
+			destNext = destIter.NextRange()
```
arielshaqed, line 72: Don't understand this, either: don't we also need to
```go
sourceNext = sourceIter.NextIter()
```
in order actually to skip past this?

arielshaqed (commented):
I'm confused about the code, so please feel free to dismiss all comments.

ozkatz:
@guy-har I think this can be closed?

PR #1639, opened 2021-03-16
Add OpenAPI endpoints

arielshaqed:
Allow splitting "upload" support for (thick) clients:

1. Client requests a physical location and an opaque token to upload a key to a branch on a
   repo.
2. Client uploads object to physical location.
3. Client requests _linking_ the physical location to the key and completes the upload.  If
   the requested token does not match the existing token then a commit occurred, and client
   should retry, possibly with a different physical location.

### DIFF
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -122,7 +122,7 @@ definitions:
 
   staging_location:
     type: object
-    description: location for placing an object before staging it
+    description: location for placing an object when staging it
     properties:
       physical_address:
         type: string
@@ -131,6 +131,28 @@ definitions:
         description: opaque staging token to use to link uploaded object
     required:
       - token
+
+  staging_metadata:
+    type: object
+    description: information about uploaded object
+    properties:
+      staging:
+        type: object
+        $ref: "#/definitions/staging_location"
+      checksum:
+        type: string
+        description: unique identifier of object content on backing store (typically ETag)
+      mtime:
+        type: integer
+        format: int64
+      size_bytes:
+        type: integer
+        format: int64
+    required:
+      - staging
+      - checksum
+      - mtime
+      - size_bytes
 
   object_stage_creation:
     type: object
@@ -2042,12 +2064,10 @@ paths:
         physical address with the supplied path.  Otherwise return a conflict and hint where
         to place the object to try again.
       parameters:
-        # TODO(ariels): Don't need a sub-object to use a schema if we switch to OpenAPI 3.0
-        #     before release.
         - in: body
-          name: staging
+          name: metadata
           schema:
-            $ref: "#/definitions/staging_location"
+            $ref: "#/definitions/staging_metadata"
       responses:
         200:
           # This actually violates HTTP, which requires returning 201 if a new object was
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -120,6 +120,18 @@ definitions:
         type: string
         enum: [ common_prefix, object ]
 
+  staging_location:
+    type: object
+    description: location for placing an object before staging it
+    properties:
+      physical_address:
+        type: string
+      token:
+        type: string
+        description: opaque staging token to use to link uploaded object
+    required:
+      - token
+
   object_stage_creation:
     type: object
     required:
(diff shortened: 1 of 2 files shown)


api/swagger.yml
```diff
@@ -1986,6 +1998,78 @@ paths:
...
+      summary: get a physical address and a return token to write object to underlying storage
+      responses:
+        200:
+          description: physical address for staging area
+          schema:
+            $ref: "#/definitions/staging_location"
+        401:
+          $ref: "#/responses/Unauthorized"
+        404:
+          description: branch not found
```
nopcoder, line 2028: repository, branch or path not found

api/swagger.yml
```diff
@@ -120,6 +120,18 @@ definitions:
...
+    type: object
+    description: location for placing an object before staging it
+    properties:
+      physical_address:
+        type: string
+      token:
+        type: string
+        description: opaque staging token to use to link uploaded object
+    required:
+      - token
```
nopcoder, line 133: can we make the physical_address required too?

api/swagger.yml
```diff
@@ -1986,6 +1998,78 @@ paths:
...
+        If the supplied token matches the current staging token, associate the object as the
+        physical address with the supplied path.  Otherwise return a conflict and hint where
+        to place the object to try again.
+      parameters:
+        # TODO(ariels): Don't need a sub-object to use a schema if we switch to OpenAPI 3.0
+        #     before release.
+        - in: body
+          name: staging
+          schema:
+            $ref: "#/definitions/staging_location"
```
nopcoder, line 2050: staging location links just the physical address - the plan is to access the object and extract the ETag, size, and timestamp from the underlying storage?

arielshaqed (commented):
Thanks; PTAL!

nopcoder (approved):

arielshaqed:
Thanks!

api/swagger.yml
```diff
@@ -1986,6 +2020,76 @@ paths:
...
+        physical address with the supplied path.  Otherwise return a conflict and hint where
+        to place the object to try again.
+      parameters:
+        - in: body
+          name: metadata
+          schema:
+            $ref: "#/definitions/staging_metadata"
+      responses:
+        200:
+          # This actually violates HTTP, which requires returning 201 if a new object was
```
ozkatz, line 2073: so why not a 201?

api/swagger.yml
```diff
@@ -1986,6 +2020,76 @@ paths:
...
+        401:
+          $ref: "#/responses/Unauthorized"
+        404:
+          description: branch not found
+          schema:
+            $ref: "#/definitions/error"
+        409:
+          description: conflict with a commit, try here
+          schema:
+            $ref: "#/definitions/staging_location"
```
ozkatz, line 2086: This implies there's a certain flow of requests that should take place between client and server. The happy path is quite clear, but assuming I'm writing a client, how should I handle this response? What should I do with the data I've already uploaded? Should I clean that up? will the lakeFS server do it for me? How often should I retry? should I employ some sort of backoff? At the very least we should document these.

api/swagger.yml
```diff
@@ -120,6 +120,40 @@ definitions:
...
+    properties:
+      staging:
+        type: object
+        $ref: "#/definitions/staging_location"
+      checksum:
+        type: string
+        description: unique identifier of object content on backing store (typically ETag)
+      mtime:
+        type: integer
+        format: int64
```
ozkatz, line 147: we probably make this mistake in other places too, but we should at least provide in a description what unit this int represents (i.e. seconds? ms? ns?)

ozkatz (commented):
Looks good IMO! My comments are mostly around documenting the flow (especially the unhappy one)

api/swagger.yml
arielshaqed: Sure, will document.
In brief:
- It says "try here", you should try there.  Or give up -- I'm an API response, not a cop.
- Cleanup is deferred to when we decide where to upload, and how to handle lifecycle in general.  I will _recommend_ that clients copy the file to the recommended location (should be fast if intra-region), delete the original, and call again.
- The lakeFS won't do anything for you if it asks you to try again, so no.  In general I don't think the server should do data operations for the client if the client is in "let me do my own data operations" mode.
- There is no need to backoff on retries: this will not be an idempotent retry, and it's not due to load.  The server is asking the client to try again with _different_ parameters.

api/swagger.yml
arielshaqed: So far it has always appeared with this name, not sure whether it is more confusing to introduce a _new_ name (and then users can never figure out the name of the field) or re-use the _old_ (bad) name.
Will change.

PR #1641, opened 2021-03-16
proposal: impl. milestone #3 using user-controlled workspaces

ozkatz:
An initial proposal for implementing milestone #3 by having a user-controlled workspace, so some of the concurrency burden falls on the client. This is still WIP and requires fleshing out the user experience for SDK, CLI and UI to better understand the trade-offs vs current design (or the possible Raft implementation).


design/user_controller_workspaces.md
```diff
@@ -0,0 +1,143 @@
...
+    --from-path "s3://bucket/kafka/events/2021/" \
+    --to-path "lakefs://my-repo@ingest-branch/kafka/events/2021/" \
+    -m "ingest events: $(date -u)"
+
+# Or just see what changes would be introduced
+$ lakectl ingest diff \
+    --from-path "s3://bucket/kafka/events/2021/" \
+    --to-path "lakefs://my-repo@ingest-branch/kafka/events/2021/" \
+```
+Here, instead of treating the given object store location as a diffIterator, we look at it as a subtree, to be diffed with the underlying commit.
```
arielshaqed, line 120: This ingest can be very expensive, because it involves a listing.  It makes ops tricky: it is never clear _which_ commit received a file.  Indeed, _commits receive files out of order with a unboundedly higher skew_ if the ingester is multi-threaded.  _Any_ skew, no matter how small, gets amplified to the _time between ingests_.

## Example

We ingest files every 2 minutes.

1. Commit starts for all files starting `PREFIX-12:34`.
2. Commit lists first page of files, up to `PREFIX-12:34:56.002`.
3. Another ingester thread finally manages to write `PREFIX-12:34:56.001`.
4. Committed continues listing, up to `PREFIX-12:35:59.999`.
5. File `PREFIX-12:34:56.001` is missing.  It arrived a few milliseconds too late, and shows up in a commit two minutes later.

design/user_controller_workspaces.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+We can provide tooling to generate or configure those on behalf of the user.
+
+## Standalone mode
+
+Running on a single instance we don't *need* a CAS operation, we can serialize access to the ref keys in-process, so a ref could sit in any object store or local directory. `lakefs -dev` now works without a database (I'd still show a big red warning in the UI when running like this - also true for the local:// adapter in general).
+
+
+## Uploading/diff staged in the UI
```
arielshaqed, line 155: Please let's have something about the CLI experience.

design/user_controller_workspaces.md
```diff
@@ -0,0 +1,143 @@
...
+
+When uploading objects we can generate a workspace ID and stick it in `localStorage`.
```
arielshaqed, line 157: However there is no list of "current staging" (`git stash list`), so hard to use the UI to finagle about with existing workspaces.  This will make debugging broken pipelines harder.

design/user_controller_workspaces.md
```diff
@@ -28,37 +28,58 @@ I believe this new responsibility could be mostly abstracted away from the user.
...
+$ lakectl fs upload ./another.parquet lakefs://my-repo/main
+Using workspace ID from ~./lakectl/workspaces/main: ozkatz-main-2021031522000
+Upload complete. You can see your uncommitted changes by doing `lakectl diff lakefs://my-repo/main+ozkatz-main-2021031522000`
+$ lakectl fs commit lakefs://my-repo/main
+Using workspace ID from ~./lakectl/workspaces/main: ozkatz-main-2021031522000
+Commit successful, wrote 2 files, deleted 0. Removing workspace ID: ~./lakectl/workspaces/main
+$ # alternatively, pass an explicit workspace ID
+$ lakectl fs upload ./another.parquet lakefs://my-repo/main+ozkatz-main-2021031522000
+$ # see workspaces per branch
+$ lakectl branch list-workspaces lakefs://my-repo/main
```
arielshaqed, line 63: Where does lakefs know what workspaces there are?  Another file...?  That would require CAS.  Perhaps a "directory", and then we can pretend atomicity matters less.

arielshaqed (commented):
Most comments from _before_ the second commit, sorry.

design/user_controller_workspaces.md
arielshaqed: # On the bright side...

It's not _really_ "unboundedly higher skew", it's "skewed by time between ingests".

# On the cloud side...

Such scenarios can break semantics.  E.g. `_success` and similar files can easily be written into a commit that doesn't include all their files, or they might be named `_finished_writing_up_to_12:36` (and lie).

design/user_controller_workspaces.md
```diff
@@ -0,0 +1,143 @@
...
+
+// commit!
+ws.all().commit("commit everything!")
+
+// or add stuff
+ws.add("**/*.parquet", "**/*.json", "another/location/_SUCCESS")
+    .commit("a commit of only some of the stuff!")
+
+// or revert it
+ws.revert()
```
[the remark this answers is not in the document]
ozkatz: could be a bulk delete, which s3 does support (being slow is ok for things outside the critical path, maybe we need to decide on an upper bound though)

alternatively, mark it somewhere and have whatever retention mechansim we end up having, do the cleanup later

design/user_controller_workspaces.md
itaiad200: Possible mitigation of the `_success` files scenario, going over all files under the prefix twice:
1. Pick up *only* `_success` files.
2. Pick up everything which isn't a `_success` file.

design/user_controller_workspaces.md
```diff
@@ -0,0 +1,164 @@
...
+$ lakectl ingest diff \
+    --from-path "s3://bucket/kafka/events/2021/" \
+    --to-path "lakefs://my-repo@ingest-branch/kafka/events/2021/"
+```
+Here, instead of treating the given object store location as a diffIterator, we look at it as a subtree, to be diffed with the underlying commit.
+
+
+## IAM permissions
+
+set them normally on `s3://storagenamespace/*/path/a`, `s3://storagenamespace/*/path/b`, etc.
```
itaiad200, line 125: If the workspace token is generated with the branch in mind, we might be able to do something with branch-based permission too.

design/user_controller_workspaces.md
```diff
@@ -0,0 +1,143 @@
...
+It is the user's responsibility to ensure they commit at the end of their job/pipeline/script,
+but it's possible to pass around that `workspaceID` between jobs and processes, even if reading/writing concurrently.
+
+Once a commit is done for a given workspace, it should no longer be used. if using the lakeFS SDK (as shown above),
+calling any operation on a `commit()`ed `Workspace` object will raise an error. We can go further and place a "lockfile" (`_lakefs_committed/${workspaceID}`) and check it before writing if we think this is neccesary (on a best effort basis).
+
+## Using the S3 gateway
+
+just append a workspace ID: `s3://my-repo/master+abc123/my/collection/` to read/write.
+otherwise, just pass a ref as you normally would, to read committed data.
```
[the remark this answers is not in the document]
arielshaqed: In exactly the same stateless way that it works with the token generated on the server.  If the server generates `BRANCH-UUID`, then clients generate `BRANCH-CLIENTLY-UNIQUE-NAME`.  If the first part isn't a valid branch, do the same thing you would do on the server -- e.g. return "branch not found".

ozkatz:
Thanks @itaiad200 and @arielshaqed - I'm going to rework this and I think I have a better proposal. Closing for now :)

Issue #1650, opened 2021-03-18
Entry's address to hold full path


itaiad200:
Entry's address has 2 possible formats:
1. Full block store address, if the item was imported.
2. Relative path to the storage namespace, in case it was created by the cataloger.

Support for the 2 formats must stay, since sstables are already written with the 2 options.
But moving forward, it's much easier to always store the full address path, so that clients can read it directly from the object store, without applying more logic to it.
Any relative address path, will simply be concatenated to the repo's storage namespace, since that was the logic when the addresses were relative.

arielshaqed:
I think there are two distinct issues here:

* stored (DB/Rocks) format: what is actually stored inside graveler (or other DBs that are used internally).
* wire ("client") format: what clients receive over the wire, and what is presented on the API.

I believe that the second is far more valuable than the first.  Indeed you write that:

> Support for the 2 formats must stay, since sstables are already written with the 2 options.

^^ applies _only_ to the stored format.

I can see few (if any) advantages to having both options on the wire, and even fewer to presenting both to the client.  So mega-:+1: from me on that.  This is also an enormously cheaper change.

I _do_ see advantages to storing relative names internally.  E.g., consider some future "git remote"-based replication: pushing between two lakeFS instances, _where data are also replicated along_.  Allowing both pathname formats (absolute and relative) means that we preserve computed IDs; it also makes the actual push (much) simpler with many more usable tools.  So I am quite wary of _storing_ full pathnames.  We need to explore more before we decide to do *that*.

arielshaqed:
#1651 changes "stored" format; I like it less than something that would change what the client sees.  In particular, note that it doesn't help write a simpler client unless we backfill all existing installations or tell them to throw away all of their data.

itaiad200:
What the client "sees" depends on where it gets its information from.
If it reads sstables and tries to do something with the underlining data (ahm `export`), then presenting the full path from the API does no good. It must change on the Entry level for supporting that.

Regarding the client logic, I agree that we'll always have to handle the relative path option since sstables were already written. But the logic is fairly simple IMHO: `fullpath -> use it`, `relativePath -> do storageNS + relativepath`. 

Anything else, like objects address patterns per repo is much more expressive and it will require the client to know more as it keeps evolving, not less.

arielshaqed:
> What the client "sees" depends on where it gets its information from.
[...]

PR #1651, opened 2021-03-18
Always write full path to entry address

itaiad200:


itaiad200:
Following the discussion on #1650, closing this PR.

PR #1652, opened 2021-03-18
Fix merge large changes performance

guy-har:


pkg/graveler/committed/apply.go
```diff
@@ -26,99 +26,129 @@ func incrementDiffSummary(d *graveler.DiffSummary, typ graveler.DiffType) {
 
 // ReferenceType represents the type of the reference
 
-// applyFromSource applies all changes from source to writer.
-func applyFromSource(ctx context.Context, logger logging.Logger, writer MetaRangeWriter, source Iterator) error {
+// applyFromIter applies all changes from Iterator to writer.
+// returns the number of writes
```
arielshaqed, line 30:
```suggestion
// applyFromIter applies all changes from iter to writer and returns the number of writes
```

(not sure what "number of writes" means)

pkg/graveler/committed/apply.go
```diff
@@ -26,99 +26,129 @@ func incrementDiffSummary(d *graveler.DiffSummary, typ graveler.DiffType) {
...
-				}).Trace("copy entire source range at end")
-			}
-			if err := writer.WriteRange(*sourceRange); err != nil {
-				return fmt.Errorf("copy source range %s: %w", sourceRange.ID, err)
+		iterValue, iterRange := iter.Value()
+		if iterValue == nil {
+			if iterRange.IsTombstone() {
+				// internal error but no data lost: deletion requested of a
+				// file that was not there.
+				logger.WithField("id", string(iterRange.ID)).Warn("[I] unmatched delete")
```
arielshaqed, line 44: This is a range, let's state as much!

```suggestion
				logger.WithFields(logging.Fields{
					"from": string(sourceRange.MinKey),
					"to":   string(sourceRange.MaxKey),
					"ID", string(iterRange.ID),
				}).Warn("[I] unmatched delete range")
```
(although, TBH, not sure why we don't use lowercase throughout for fields).

pkg/graveler/committed/apply.go
```diff
@@ -26,99 +26,129 @@ func incrementDiffSummary(d *graveler.DiffSummary, typ graveler.DiffType) {
...
-					"to":   string(sourceRange.MaxKey),
-					"ID":   sourceRange.ID,
-				}).Trace("copy entire source range at end")
-			}
-			if err := writer.WriteRange(*sourceRange); err != nil {
-				return fmt.Errorf("copy source range %s: %w", sourceRange.ID, err)
+		iterValue, iterRange := iter.Value()
+		if iterValue == nil {
+			if iterRange.IsTombstone() {
+				// internal error but no data lost: deletion requested of a
```
arielshaqed, line 42: This is not counted, I assume on purpose because it's a probably bug?  Maybe still count unjustified deletes just to be sure...

pkg/graveler/committed/apply.go
```diff
@@ -26,99 +26,129 @@ func incrementDiffSummary(d *graveler.DiffSummary, typ graveler.DiffType) {
...
+			return true
 		}
 	}
-	return numAdded, nil
+	return false
 }
 
-func Apply(ctx context.Context, writer MetaRangeWriter, source Iterator, diffs graveler.ValueIterator, opts *ApplyOptions) (graveler.DiffSummary, error) {
+func Apply(ctx context.Context, writer MetaRangeWriter, source Iterator, diffs Iterator, opts *ApplyOptions) (graveler.DiffSummary, error) {
 	logger := logging.FromContext(ctx)
```
arielshaqed, line 96: Note that this "cyclomatic complexity" nonsense already bit us, which is why I split off the 2 functions at the top.  We should probably bite off a bit more here, maybe stuff like keeping `haveSource`, `haveDiffs` inside a state object with the iterators and performing actions on _that_; or just passing that state object into a separate function that holds each of the `select`s.

pkg/graveler/committed/apply.go
```diff
@@ -26,99 +26,129 @@ func incrementDiffSummary(d *graveler.DiffSummary, typ graveler.DiffType) {
...
 		sourceValue, sourceRange := source.Value()
-		diffValue := diffs.Value()
-		if sourceValue == nil {
+		diffValue, diffRange := diffs.Value()
+		switch {
+		case diffValue == nil && sourceValue == nil:
+			switch {
+			case bytes.Compare(diffRange.MaxKey, sourceRange.MinKey) < 0:
+				// insert diff
+				writer.WriteRange(*diffRange)
```
arielshaqed, line 112: Need to check all `Write*` values :-/

pkg/graveler/committed/apply.go
```diff
@@ -26,99 +26,129 @@ func incrementDiffSummary(d *graveler.DiffSummary, typ graveler.DiffType) {
...
+				haveDiffs = diffs.Next()
+			case bytes.Compare(sourceRange.MaxKey, diffRange.MinKey) < 0:
+				//insert source
+				writer.WriteRange(*sourceRange)
+				haveSource = source.Next()
+			case diffRange.ID == sourceRange.ID && diffRange.IsTombstone():
+				addIntoDiffSummary(&ret, graveler.DiffTypeRemoved, int(diffRange.Count))
+				haveSource = source.Next()
+				haveDiffs = diffs.Next()
+			default:
```
arielshaqed, line 123: :point_right: I don't understand: are these all the options?  What about when `sourceRange` and `diffRange` have overlapping keys?

```
source               [min].....[max]
        ------------------------------------------------------------------------------------------------
diffs       [min...........max]

File: pkg/graveler/committed/apply.go [hunk 7]
```diff
@@ -26,99 +26,129 @@ func incrementDiffSummary(d *graveler.DiffSummary, typ graveler.DiffType) {
...
+						logger.WithFields(logging.Fields{
+							"from": string(diffRange.MinKey),
+							"to":   string(diffRange.MaxKey),
+							"ID":   diffRange.ID,
+						}).Trace("copy entire diff range")
+					}
+					if err := writer.WriteRange(*diffRange); err != nil {
+						return ret, fmt.Errorf("copy diff range %s: %w", sourceRange.ID, err)
+					}
+					haveSource = diffs.NextRange()
```
arielshaqed, line 145:
```suggestion
					haveDiffs = diffs.NextRange()
```
# :question:

arielshaqed (changes requested):
Reviewed `apply` and `apply_test` (only).  Not sure how it passes tests; I think there's a bug, which would mean we need more test coverage.  OR (even more likely) I don't understand the code.  Either is a good reason to pause reviewing and give you a chance to respond!
Sorry...

pkg/graveler/committed/apply.go
guy-har: In case of overlapping keys, we call next on both ranges In order to compare keys.

guy-har:
@arielshaqed Thanks,
I made some changes and added some tests.
I am planning on adding some more tests.

pkg/graveler/committed/merge_iterator.go
```diff
@@ -68,75 +81,135 @@ func (d *compareIterator) handleConflict() bool {
...
-				// removed on dest, but changed on source
-				return d.handleConflict()
-			}
-			continue
-		case graveler.DiffTypeChanged:
-			if baseVal == nil {
-				// added on dest and source, with different identities
-				return d.handleConflict()
+		val, rngDiff := d.diffIt.Value()
+		if rngDiff != nil {
```
arielshaqed, line 100: Not sure this (allowing a call to `.Value()` on an iterator that has ended) is part of the Go iterator contract; paging @nopcoder...

pkg/graveler/committed/merge_iterator.go
```diff
@@ -68,75 +81,135 @@ func (d *compareIterator) handleConflict() bool {
...
+			typ := rngDiff.Type
+			rng := rngDiff.Range
+			baseRange, err := d.rangeFromBase(graveler.Key(rng.MinKey))
```
arielshaqed, line 103: Doesn't this continually seek the base iterator, once on every value?

pkg/graveler/committed/merge_iterator.go
```diff
@@ -68,75 +81,135 @@ func (d *compareIterator) handleConflict() bool {
...
+			if err != nil {
+				d.err = err
+				return false
 			}
-			if bytes.Equal(baseVal.Identity, val.Value.Identity) {
-				// changed on dest, but not on source
-				continue
+			switch typ {
+			case graveler.DiffTypeAdded:
+				// exists on source, but not on dest
```
arielshaqed, line 110: Can you document `DiffTypeAdded` better?  I would expect this to be "exists on dest, but not on source" -- here it's the reverse?!  (Maybe add pretty pictures...?)

pkg/graveler/committed/merge_iterator.go
```diff
@@ -68,75 +81,135 @@ func (d *compareIterator) handleConflict() bool {
...
+				if baseRange == nil || bytes.Compare(rng.MaxKey, baseRange.MinKey) < 0 {
+					// added only on source
+					d.rng = rngDiff
+					d.val = nil
+					return true
+				}
+
+				hasNext = d.diffIt.Next()
```
arielshaqed, line 118: This case is a bit harder to follow than the next one, comments would help see the *difference* between them.  In particular, why do we `continue` on the very similar line 130 but keep going here?

pkg/graveler/committed/merge_iterator_test.go
```diff
@@ -246,18 +248,20 @@ func TestMergeSeek(t *testing.T) {
 	for _, tst := range tests {
 		t.Run(fmt.Sprintf("seek to %s", tst.seekTo), func(t *testing.T) {
 			it.SeekGE([]byte(tst.seekTo))
-			if it.Value() != nil {
-				t.Fatalf("value expected to be nil after SeekGE. got=%v", it.Value())
+			val, _ := it.Value()
```
arielshaqed, line 251: Shouldn't we also test the returned ranges?

pkg/graveler/committed/meta_range.go
```diff
@@ -24,6 +24,29 @@ type Iterator interface {
...
 }
 
+// DiffIterator iterates over all Range headers and values of a Diff, allowing seeking by entire
+// ranges.
+type DiffIterator interface {
+	// Next moves to look at the next value in the current Range, or a header for the next Range if the current Range is over and a next range exists.
+	Next() bool
+	// NextRange skips the current range
+	// possible only if we are currently inside a range
+	// the next value could be Range or Value
```
arielshaqed, line 34: Why?

pkg/graveler/committed/meta_range.go
```diff
@@ -24,6 +24,29 @@ type Iterator interface {
...
+	NextRange() bool
+	// Value returns a nil ValueRecord and a Range before starting a Range, or a Value and that Range when inside a Range.
+	// In contrast to Iterator, the DiffIterator might not have a current range
```
arielshaqed, line 37: Can you document _when_ there is no current range?  The code above seems to rely on that.

arielshaqed (changes requested):
Thanks!  Nice work, but many details in a large :elephant:  PR.

A bit hard to review because some comments appear to have no responses and also no changes.  Specifically I am unsure why we poll on cancellation.

I have some (ok, many) places I did not understand, and would prefer to go over those before I read the tests more deeply.

pkg/graveler/committed/meta_range.go
guy-har: We could only get a range when one of the ranges is completely before the other range,
In case we have overlapping ranges we could not return range.

pkg/graveler/committed/apply_test.go
```diff
@@ -59,19 +58,19 @@ func TestApplyReplace(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
 
-	range2 := &committed.Range{ID: "two", MaxKey: committed.Key("dz")}
+	range2 := &committed.Range{ID: "two", MaxKey: committed.Key("dz"), Count: 1, EstimatedSize: 1020}
 	source := testutil.NewFakeIterator()
 	source.
-		AddRange(&committed.Range{ID: "one", MaxKey: committed.Key("cz")}).
+		AddRange(&committed.Range{ID: "one", MaxKey: committed.Key("cz"), Count: 1, EstimatedSize: 1020}).
 		AddValueRecords(makeV("a", "source:a"), makeV("b", "source:b"), makeV("c", "source:c")).
```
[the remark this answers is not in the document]
guy-har: When adding a range, I just check that the right range is added ( there is no need to go in to the keys ), that's the reason it is working... It is confusing, I will change it! thanks

pkg/graveler/committed/apply_test.go
```diff
@@ -59,19 +58,19 @@ func TestApplyReplace(t *testing.T) {
 	ctrl := gomock.NewController(t)
 	defer ctrl.Finish()
 
-	range2 := &committed.Range{ID: "two", MaxKey: committed.Key("dz")}
+	range2 := &committed.Range{ID: "two", MaxKey: committed.Key("dz"), Count: 1, EstimatedSize: 1020}
```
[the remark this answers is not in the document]
guy-har: Because this range is added as a range, on l. 37 it's part of overlapping ranges which causes the keys written and not the range. I am adding metadata to all the ranges in this file in order to make it less confusing. Thanks

pkg/graveler/committed/diff.go
```diff
@@ -81,16 +128,27 @@ func (d *diffIterator) compareDiffKeys() int {
...
+	if leftRange != nil && rightRange != nil && leftRange.ID == rightRange.ID {
 		return diffItCompareResultSameRanges
 	}
-	leftStartRange := d.leftValue.rng != nil && d.leftValue.record == nil
-	rightStartRange := d.rightValue.rng != nil && d.rightValue.record == nil
+	leftStartRange := leftRange != nil && d.leftValue.record == nil
+	rightStartRange := rightRange != nil && d.rightValue.record == nil
+	rightBeforeLeft, leftBeforeRight := false, false
+	if leftStartRange && rightStartRange {
+		rightBeforeLeft = bytes.Compare(leftRange.MaxKey, rightRange.MinKey) < 0
```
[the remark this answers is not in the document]
guy-har: guess I needed the coffee, accidentally swapped the names. fixing it.

pkg/graveler/committed/diff.go
```diff
@@ -81,16 +128,27 @@ func (d *diffIterator) compareDiffKeys() int {
...
+		leftBeforeRight = bytes.Compare(rightRange.MaxKey, leftRange.MinKey) < 0
+	}
 	comp := d.compareDiffKeys()
 	switch {
+	case leftStartRange && rightStartRange && rightBeforeLeft:
```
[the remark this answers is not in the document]
guy-har: we don't! fixing

pkg/graveler/committed/diff.go
```diff
@@ -112,61 +170,91 @@ func (d *diffIterator) Next() bool {
...
+		d.currentRange.value.record, d.currentRange.value.rng, d.currentRange.value.err = diffIteratorNextValue(d.currentRange.iter)
+		if d.currentRange.value.record != nil {
+			var leftIdentity []byte
+			if d.currentRange.isLeft {
+				leftIdentity = d.currentRange.value.record.Identity
+			}
+			d.currentDiff = &graveler.Diff{Type: d.currentRange.currentRangeDiff.Type, Key: d.currentRange.value.record.Key.Copy(), Value: d.currentRange.value.record.Value, LeftIdentity: leftIdentity}
+			return true
+		}
+		d.emptyCurrentRange()
```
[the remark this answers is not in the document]
guy-har: Once we get a nil record on current range, we know that the current range (which we know is entirely before the other range/value) is over. At this point we need to empty the current range data we have and continue to the next step.

pkg/graveler/committed/diff.go
```diff
@@ -112,61 +170,91 @@ func (d *diffIterator) Next() bool {
...
-			d.currentDiff = &graveler.Diff{Type: graveler.DiffTypeAdded, Key: d.rightValue.record.Key.Copy(), Value: d.rightValue.record.Value}
-			d.rightValue.record, d.rightValue.rng, d.rightValue.err = diffIteratorNextValue(d.right)
-			return true
 		}
+		for {
+			if d.rightValue.err != nil {
+				d.err = d.rightValue.err
+			}
+			if d.leftValue.err != nil {
+				d.err = d.leftValue.err
```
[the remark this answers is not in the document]
guy-har: It is OK, Thanks

pkg/graveler/committed/merge_iterator.go
```diff
@@ -68,75 +81,135 @@ func (d *compareIterator) handleConflict() bool {
...
 }
 
-func (d *compareIterator) Next() bool {
-	for {
-		// get next value or return with context/iterator error (if any)
+func (d *compareIterator) Step() bool {
+	hasNext := true
+	for hasNext {
 		select {
 		case <-d.ctx.Done():
```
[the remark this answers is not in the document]
guy-har: The context cancellation was added as part of https://github.com/treeverse/lakeFS/pull/1603

pkg/graveler/committed/merge_iterator.go
guy-har: At this point the iterator shouldn't end ( If it could, there is a bug here).
`Step` is called after the  `Next` or `NextRange` has returned true and runs in a loop that updates hasNext.
Obviously it's not clear... changing it.

pkg/graveler/committed/diff.go
```diff
@@ -112,61 +170,91 @@ func (d *diffIterator) Next() bool {
...
+			if d.currentRange.isLeft {
+				leftIdentity = d.currentRange.value.record.Identity
+			}
+			d.currentDiff = &graveler.Diff{Type: d.currentRange.currentRangeDiff.Type, Key: d.currentRange.value.record.Key.Copy(), Value: d.currentRange.value.record.Value, LeftIdentity: leftIdentity}
+			return true
+		}
+		d.emptyCurrentRange()
+	}
 	select {
 	case <-d.ctx.Done():
```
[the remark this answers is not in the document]
guy-har: The context cancellation was added as part of #1603

pkg/graveler/committed/diff.go
```diff
@@ -112,61 +170,91 @@ func (d *diffIterator) Next() bool {
...
+		for {
+			if d.rightValue.err != nil {
+				d.err = d.rightValue.err
+			}
+			if d.leftValue.err != nil {
+				d.err = d.leftValue.err
+			}
+			if d.err != nil {
+				d.currentDiff = nil
+				d.emptyCurrentRange()
```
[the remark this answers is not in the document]
guy-har: True, but we return nil in values in case of errors 
```// In case `Next()` returns false, `Value()` returns nil and `Err()` should be checked.```
At least this is what we wrote in the documentation in graveler

File: pkg/graveler/committed/diff.go [hunk 24]
```diff
@@ -112,61 +170,91 @@ func (d *diffIterator) Next() bool {
...
+				return false
+			}
+			compareResult := d.compareDiffIterators()
+			switch compareResult {
+			case diffItCompareResultDone:
+				d.currentDiff = nil
+				return false
+			case diffItCompareResultSameRanges:
+				d.leftValue.rng, d.leftValue.err = diffIteratorNextRange(d.left)
+				d.rightValue.rng, d.rightValue.err = diffIteratorNextRange(d.right)
```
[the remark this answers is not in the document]
guy-har: In range cases we don't look at value so it is not used. I could change it if it makes more sense.

pkg/graveler/committed/diff_flat.go
```diff
@@ -0,0 +1,48 @@
+package committed
+
+import (
+	"context"
+
+	"github.com/treeverse/lakefs/pkg/graveler"
+)
+
+// flatDiffIterator wraps a diffIterator in order to return only values
```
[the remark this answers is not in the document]
guy-har: Changing, Thanks!

pkg/graveler/committed/apply_test.go
```diff
@@ -121,31 +115,28 @@ func TestApplyDelete(t *testing.T) {
 	}, summary)
 }
 
-func TestApplyCopiesLeftoverDiffs(t *testing.T) {
+func TestAppCopiesLeftoverDiffs(t *testing.T) {
```
johnnyaug, line 118:
```suggestion
func TestApplyCopiesLeftoverDiffs(t *testing.T) {
```

johnnyaug (changes requested):

johnnyaug:
Great optimization!!

pkg/graveler/committed/apply_test.go
```diff
@@ -93,24 +89,22 @@ func TestApplyDelete(t *testing.T) {
...
-
+	diffs := testutil.NewFakeIterator()
+	diffs.
+		AddRange(&committed.Range{ID: "diffs:one", MinKey: committed.Key("b"), MaxKey: committed.Key("e"), Count: 2, EstimatedSize: 2012}).
+		AddValueRecords(makeTombstoneV("b"), makeTombstoneV("e"))
 	writer := mock.NewMockMetaRangeWriter(ctrl)
 	writer.EXPECT().WriteRecord(gomock.Eq(*makeV("a", "source:a")))
 	writer.EXPECT().WriteRecord(gomock.Eq(*makeV("c", "source:c")))
-	writer.EXPECT().WriteRange(gomock.Eq(*range2))
+	writer.EXPECT().WriteRange(gomock.Eq(committed.Range{ID: "two", MinKey: committed.Key("d"), MaxKey: committed.Key("dz"), Count: 1, EstimatedSize: 1011}))
```
johnnyaug, line 107: Also add a test for a deleted range here? (i.e. EstimatedSize==0)

In general, this test file can be expanded.

pkg/graveler/committed/merge_iterator.go
```diff
@@ -68,75 +78,135 @@ func (d *compareIterator) handleConflict() bool {
...
+					// changed on dest, removed on source
+					return d.handleConflict()
+				}
+				// added on dest, but not on source - next value
 			}
-			// added on dest, but not on source - continue
+			hasNext = d.diffIt.Next()
 		}
 	}
+	if d.err != nil {
```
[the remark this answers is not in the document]
guy-har: It should be d.err, if we have an error leave the error, otherwise use the diff iterator error

pkg/graveler/committed/apply.go
```diff
@@ -14,205 +14,285 @@ type ApplyOptions struct {
...
+				a.logger.WithFields(logging.Fields{
+					"from": string(diffRange.MinKey),
+					"to":   string(diffRange.MaxKey),
+					"ID":   diffRange.ID,
+				}).Trace("copy entire diff range")
+			}
+			if err := a.writer.WriteRange(*diffRange); err != nil {
+				return fmt.Errorf("copy diff range %s: %w", diffRange.ID, err)
+			}
+			a.haveSource = a.diffs.NextRange()
```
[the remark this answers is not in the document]
guy-har: have diffs... Adding test, thanks!

guy-har:
I have made changes according to the comments and added tests.
@arielshaqed @johnnyaug PTAL.

johnnyaug (changes requested):

pkg/graveler/committed/merge_iterator.go
```diff
@@ -192,11 +125,113 @@ func (d *compareIterator) Step() bool {
...
+		d.err = err
+		return false, true
+	}
+	switch typ {
+	case graveler.DiffTypeAdded:
+		// exists on source, but not on dest
+		if baseVal == nil {
+			// added only on source
+			d.val = val.Copy()
+			d.setRangeDiff(rngDiff)
```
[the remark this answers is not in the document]
guy-har: The iterators return range when they can, so that if you are Iterating through a range even if you get a value you could get the current range and understand what to expect on calling nextRange. 
The reason they do not always return range is that they don't always have a range (e.g diff of two overlapping ranges)

pkg/graveler/committed/merge_iterator_test.go
```diff
@@ -158,6 +158,359 @@ func TestMerge(t *testing.T) {
...
+}
+
+func getRangeID(rng *committed.Range) string {
+	if rng == nil {
+		return ""
+	}
+	return string(rng.ID)
+}
+
+func TestMergeRangTemp(t *testing.T) {
```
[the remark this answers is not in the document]
guy-har: thanks

pkg/graveler/committed/diff_test.go
```diff
@@ -192,20 +221,28 @@ func TestDiff(t *testing.T) {
...
+			it := committed.NewDiffIterator(ctx, fakeLeft, fakeRight)
 			defer it.Close()
 			var diffs []*graveler.Diff
 			actualDiffKeys := make([]string, 0)
+			actualDiffRanges := make([]*expectedRange, 0)
 			for it.Next() {
-				actualDiffKeys = append(actualDiffKeys, string(it.Value().Key))
-				diffs = append(diffs, it.Value())
+				diff, rng := it.Value()
+				actualDiffRanges = append(actualDiffRanges, newExpectedRange(rng))
```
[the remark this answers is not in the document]
guy-har: both are already in use, changing to diffTestRange

pkg/graveler/committed/merge_iterator.go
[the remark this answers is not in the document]
guy-har: It's the same behavior as `committed.iterator`, with the exception that we don't always have a range

pkg/graveler/committed/apply.go
arielshaqed: Still need to check all `Write*` values.

pkg/graveler/committed/apply_test.go
```diff
@@ -32,11 +32,10 @@ func TestApplyAdd(t *testing.T) {
...
 		AddValueRecords(makeV("d", "source:d"))
-	diffs := testutil.NewValueIteratorFake([]graveler.ValueRecord{
-		*makeV("b", "dest:b"),
-		*makeV("e", "dest:e"),
-		*makeV("f", "dest:f"),
-	})
+	diffs := testutil.NewFakeIterator()
+	diffs.
+		AddRange(&committed.Range{ID: "b", MaxKey: committed.Key("e")}).
+		AddValueRecords(makeV("b", "dest:b"), makeV("e", "dest:e"), makeV("f", "dest:f"))
```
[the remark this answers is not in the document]
arielshaqed: Not sure how this is "resolved".

arielshaqed:
@guy-har , it is very hard for me to review a monolithic PR such as this one.  I wrote in the previous round that:

> A bit hard to review because some comments appear to have no responses and also no changes. Specifically I am unsure why we poll on cancellation.

Please note that I am trying to be very explicit about both these requests.  The second remains open -- we poll on cancellation throughout the code, I am not sure why.

About the second: You marked _most_ changes as resolved, but not all -- some just vanish into an "outdated" section.  This makes it harder for me to track what came of each one.  Sometimes the comment is marked "resolved" and but I do not understand the resolution.  For instance I did not understand how you resolved [this comment](https://github.com/treeverse/lakeFS/pull/1652/files/30b932e01644e482c3af95d006bf96a12e3291ae#r598243084).

A short comment might help, particularly when the fix is nontrivial.  

In any case, apologies for an expected long round-trip time.

pkg/graveler/committed/apply.go
guy-har: `applyFromIter` only returns addition because it runs at when there is no iterator to compare with, so it should have only additions. I could add "deletions" to it, but I don't think It should be counted. @arielshaqed  what do you think ?

pkg/graveler/committed/merge_iterator.go
guy-har: Changed and made both cases  (added, removed) similar

johnnyaug (approved):

pkg/graveler/committed/apply_test.go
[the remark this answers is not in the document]
guy-har: by changing MaxKey to f

guy-har:
@arielshaqed I marked the resolved after changing according to the comment, I will go over them again.
according to polling on cancelation. I answered in both comments and marked as resolved, which made it not visible (sorry about that). My answer to the polling was that it was added as part of #1603, I basically just moved it around. We could decide we wan't to remove it, but I don't think it should be part of this PR.

pkg/graveler/committed/diff.go
```diff
@@ -112,61 +170,91 @@ func (d *diffIterator) Next() bool {
...
-			d.leftValue.record, d.leftValue.rng, d.leftValue.err = diffIteratorNextValue(d.left)
-			d.rightValue.record, d.rightValue.rng, d.rightValue.err = diffIteratorNextValue(d.right)
-			return true
-		case diffItCompareResultSameIdentities, diffItCompareResultNeedStartRangeBoth:
-			d.leftValue.record, d.leftValue.rng, d.leftValue.err = diffIteratorNextValue(d.left)
-			d.rightValue.record, d.rightValue.rng, d.rightValue.err = diffIteratorNextValue(d.right)
-		case diffItCompareResultNeedStartRangeLeft:
+		if d.state == diffIteratorStatePreInit {
+			d.state = diffIteratorStateOpen
 			d.leftValue.record, d.leftValue.rng, d.leftValue.err = diffIteratorNextValue(d.left)
```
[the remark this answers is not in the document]
guy-har: Not changed as part of this PR, prefer to leave it as is

pkg/graveler/committed/merge_iterator.go
guy-har: This also wasn't added as part of this PR, the reason is that we are holding a diff iterator that does a 2 way diff between destination and source. I could also change it if it makes more sense. prefer not as part of this PR. @arielshaqed what do you think?

pkg/graveler/committed/merge_iterator_test.go
guy-har: Added another test for merge seek with ranges

pkg/graveler/committed/apply.go
```diff
@@ -14,205 +14,296 @@ type ApplyOptions struct {
...
-func addIntoDiffSummary(d *graveler.DiffSummary, typ graveler.DiffType, n int) {
-	if d.Count != nil {
-		d.Count[typ] += n
-	}
-}
+// ReferenceType represents the type of the reference
 
-func incrementDiffSummary(d *graveler.DiffSummary, typ graveler.DiffType) {
-	addIntoDiffSummary(d, typ, 1)
+type applier struct {
```
arielshaqed, line 19: I don't understand why this is an object.  The only state it holds is a diff summary, which is anyway implemented as a function, and the two `have*` members.  It is created once, then its `apply` method is called -- with no parameters.  AFAICT it is just a function that has been represented as an object.  Having it as an object makes me have to worry about whether it does actually have important state.  Calls to functions that modify `have*` could be modeled as re-assigning some function-local state variables.  Alternatively, if there is a useful abstraction for "iterator-and-the-last-value-from-next", then let's implement that and put the `X` `Iterator` and `haveX` into a single class named something like `IteratorWithEnd`.

If the goal is to have a nice struct with all the used parameters in it then just have a struct, no need to attach methods to it.  But I would really question why passing a single parameter named `a` to a function is clearer than passing the named parameters to that function.  E.g. as it stands, the type signature for `addIntoDiffSumary` doesn't indicate that it actually changes only `a.summary`, but does not advance any iterators or write ranges.  Sure, the call `a.addIntoDiffSummary(typ, n)` is shorter than `addIntoDiffSummary(summary, type, n)`, but I claim that is just because it gives less information, and forces me to go read the code.

Also nit: `applier` sounds like Go naming for an interface.

pkg/graveler/committed/apply.go
arielshaqed: Dunno.  I'd count it as a separate category maybe, that way if it starts showing up we have a chance of someone complaining.

pkg/graveler/committed/apply.go
```diff
@@ -14,205 +14,296 @@ type ApplyOptions struct {
...
+	summary               graveler.DiffSummary
+	haveDiffs, haveSource bool
 }
 
-// ReferenceType represents the type of the reference
-
-// applyFromSource applies all changes from source to writer.
-func applyFromSource(ctx context.Context, logger logging.Logger, writer MetaRangeWriter, source Iterator) error {
+// applyAll applies all changes from Iterator to writer and returns the number of writes
+func (a *applier) applyAll(iter Iterator) (int, error) {
```
arielshaqed, line 31: E.g. because of the iterator/have separation above, I think this function is unsafe to call on iterator `X` unless `haveX`.

pkg/graveler/committed/apply.go
```diff
@@ -14,205 +14,296 @@ type ApplyOptions struct {
...
-// applyFromDiffs applies all changes from diffs to writer and returns the number elements it
-// added.
-func applyFromDiffs(ctx context.Context, logger logging.Logger, writer MetaRangeWriter, diffs graveler.ValueIterator) (int, error) {
-	numAdded := 0
-	for {
-		select {
-		case <-ctx.Done():
-			return 0, ctx.Err()
-		default:
+func (a *applier) hasChanges(summary graveler.DiffSummary) bool {
```
arielshaqed, line 90: This is a pure function not a method.

pkg/graveler/committed/apply.go
```diff
@@ -14,205 +14,296 @@ type ApplyOptions struct {
...
 			}
-			if !source.Next() {
+			count++
+			if !iter.Next() {
 				break
 			}
 		}
 	}
-	return source.Err()
+	return count, iter.Err()
```
arielshaqed, line 87: E.g. after calling this method the caller has to increment the correct statistics field -- even though it sort-of has access to the statistics (it just doesn't know what type to increment).

pkg/graveler/committed/apply_test.go
```diff
@@ -55,30 +52,94 @@ func TestApplyAdd(t *testing.T) {
...
+	diffs.
+		AddRange(&committed.Range{ID: "inner-range-1", MinKey: committed.Key("k2"), MaxKey: committed.Key("k3"), Count: 2}).
+		AddValueRecords(makeV("k2", "dest:k2"), makeV("k3", "dest:k3")).
+		AddRange(&committed.Range{ID: "inner-range-2", MinKey: committed.Key("k4"), MaxKey: committed.Key("k5"), Count: 2}).
+		AddValueRecords(makeV("k4", "dest:k4"), makeV("k5", "dest:k5"))
+	writer := mock.NewMockMetaRangeWriter(ctrl)
+	writer.EXPECT().WriteRecord(gomock.Eq(*makeV("k1", "source:k1")))
+	writer.EXPECT().WriteRecord(gomock.Eq(*makeV("k2", "dest:k2")))
+	writer.EXPECT().WriteRecord(gomock.Eq(*makeV("k3", "dest:k3")))
+	writer.EXPECT().WriteRange(gomock.Eq(*&committed.Range{ID: "inner-range-2", MinKey: committed.Key("k4"), MaxKey: committed.Key("k5"), Count: 2}))
```
arielshaqed, line 75: Not sure what `*&` does here, don't believe it is necessary.

```suggestion
	writer.EXPECT().WriteRange(gomock.Eq(committed.Range{ID: "inner-range-2", MinKey: committed.Key("k4"), MaxKey: committed.Key("k5"), Count: 2}))
```

Also below.

pkg/graveler/committed/diff.go
```diff
@@ -112,61 +174,95 @@ func (d *diffIterator) Next() bool {
...
+				// nothing on left, or right before left
+				d.currentDiff = &graveler.Diff{Type: graveler.DiffTypeAdded, Key: d.rightValue.record.Key.Copy(), Value: d.rightValue.record.Value}
+				d.rightValue.record, d.rightValue.rng, d.rightValue.err = diffIteratorNextValue(d.right)
+				return true
+			}
+		}
+	}
+}
+
+func (d *diffIterator) NextRange() bool {
```
arielshaqed, line 258: Given the comment on `ErrNoRange`, I believe that now _this_ function needs to be documented.  Specifically, AFAICT it cannot be called twice in a row (unlike a regular iterator).  That's OK, it's just different -- and requires documentation.

pkg/graveler/committed/diff_test.go
```diff
@@ -14,6 +14,19 @@ import (
 	"github.com/treeverse/lakefs/pkg/graveler/testutil"
 )
 
+type diffTestRange struct {
+	minKey string
+	maxKey string
+	count  int64
+}
```
arielshaqed, line 21: This is examined using `deep.Equal`.  That's a no-op for lowercase fields :-(
https://play.golang.org/p/c7RVJ6_lslg

pkg/graveler/committed/diff_test.go
```diff
@@ -238,6 +275,98 @@ func TestDiffCancelContext(t *testing.T) {
...
+			t.Fatal("expected record to be nil")
+		}
+		if !it.Next() { // move to k1
+			t.Fatal("expected iterator to have value")
+		}
+		record, rng = it.Value()
+		if record == nil || string(record.Key) != "k1" {
+			t.Fatal("expected record key to be k1")
+		}
+		if !it.NextRange() { // move to k3
```
arielshaqed, line 299: `NextRange` enters the range?  I think that's different from the single iterator (see [documentation](https://github.com/treeverse/lakeFS/blob/303b949346864bf84ce3f8e6675487b5d96e06a8/pkg/graveler/committed/meta_range.go#L17)).  So at the very least we must document this function.  Ideally we should also know why they are different.

pkg/graveler/committed/diff_test.go
```diff
@@ -238,6 +275,98 @@ func TestDiffCancelContext(t *testing.T) {
...
+		newFakeMetaRangeIterator([][]string{{"k1", "k2"}, {"k3", "k4"}}, [][]string{{"i1", "i2"}, {"i3", "i4"}}),
+		newFakeMetaRangeIterator([][]string{{"k3", "k4"}, {"k5", "k6"}}, [][]string{{"i3a", "i4a"}, {"i5", "i6"}}))
+
+	t.Run("next range - in middle of range", func(t *testing.T) {
+		if !it.Next() { // move to range k1-k2
+			t.Fatal("expected iterator to have value")
+		}
+		record, rng := it.Value()
+		if record != nil {
+			t.Fatal("expected record to be nil")
```
arielshaqed, line 290: Why not an error here?  Continuing the test would give more error messages, possibly useful.  Also for _many_ (but not _all_) of the fatals downstream.

pkg/graveler/committed/manager.go
```diff
@@ -105,10 +117,10 @@ func (c *committedManager) Merge(ctx context.Context, ns graveler.StorageNamespa
 	defer baseIt.Close()
 	patchIterator := NewMergeIterator(ctx, diffIt, baseIt)
 	defer patchIterator.Close()
-	return c.Apply(ctx, ns, destination, patchIterator)
+	return c.applyOnDiffWithRanges(ctx, ns, destination, patchIterator)
 }
 
-func (c *committedManager) Apply(ctx context.Context, ns graveler.StorageNamespace, rangeID graveler.MetaRangeID, diffs graveler.ValueIterator) (graveler.MetaRangeID, graveler.DiffSummary, error) {
+func (c *committedManager) applyOnDiffWithRanges(ctx context.Context, ns graveler.StorageNamespace, rangeID graveler.MetaRangeID, diffs Iterator) (graveler.MetaRangeID, graveler.DiffSummary, error) {
```
arielshaqed, line 123: Why the change of names?  Is it no longer possible to call this `apply` from the outside?

pkg/graveler/committed/merge_iterator_test.go
```diff
@@ -156,6 +157,359 @@ func TestMerge(t *testing.T) {
...
+}
+
+func getRangeID(rng *committed.Range) string {
+	if rng == nil {
+		return ""
+	}
+	return string(rng.ID)
+}
+
+func TestMergeRange(t *testing.T) {
```
arielshaqed, line 206: Nice test! :sunglasses:

pkg/graveler/committed/merge_iterator_test.go
```diff
@@ -156,6 +157,359 @@ func TestMerge(t *testing.T) {
...
+	}
+	diffIt.AddRange(rangeDiff2)
+
+	diffIt.AddValueRecords(makeDV(added, "k4", "i4", ""), makeDV(added, "k5", "i5", ""))
+
+	ctx := context.Background()
+	it := committed.NewMergeIterator(ctx, diffIt, baseIt)
+
+	if !it.Next() {
+		t.Fatalf("expected to have next - didn't get get next with error:%v", it.Err())
```
arielshaqed, line 474: 2 gets, but also could just be shorter:
```suggestion
		t.Fatalf("expected to have next (err %v)", it.Err())
```

pkg/graveler/committed/meta_range.go
```diff
@@ -24,6 +24,39 @@ type Iterator interface {
...
+// ranges.
+type DiffIterator interface {
+	// Next moves to look at the next value in the current Range, or a header for the next Range if the current Range is over and a next range exists.
+	Next() bool
+	// NextRange skips the current range
+	// possible only if we are currently inside a range
+	// the next value could be Range or Value
+	NextRange() bool
+	// Value returns a nil ValueRecord and a Range before starting a Range, or a Value and that Range when inside a Range.
+	// In contrast to Iterator, the DiffIterator might not have a current range -  this could happen if the current value exists in two different ranges
```
arielshaqed, line 37: Can we explain what `NextRange` does in this case?

pkg/graveler/committed/range.go
```diff
@@ -10,6 +10,25 @@ type Range struct {
...
+	Tombstone     bool
+}
+
+func (r Range) Copy() *Range {
+	return &Range{
+		ID:            r.ID,
+		MinKey:        r.MinKey.Copy(),
+		MaxKey:        r.MaxKey.Copy(),
+		EstimatedSize: r.EstimatedSize,
+		Count:         r.Count,
```
arielshaqed, line 22: Why not copy the tombstone?

pkg/graveler/committed/range.go
```diff
@@ -10,6 +10,25 @@ type Range struct {
...
+	}
+}
+
+func (r Range) IsTombstone() bool {
+	return r.Tombstone
+}
+
+func (r *Range) SetTombstone() {
+	r.Tombstone = true
```
arielshaqed, line 31: No other field has a mutator.  Why does tombstone need one?

arielshaqed (approved):
Thanks!

Nothing major (I hope), but must do these before pulling:

1. Document merge iterator `NextRange` and explain diff iterator `NextRange` when there is no range.
2. AFAICT one test uses `deep.Equal` on a struct with lowercase field names, so it does very little.

pkg/graveler/committed/diff_test.go
guy-har: It will not enter the range if there is a range, in this case k3 is part of two different ranges.
Basically the next range is a "headerless" range. 
Adding a comment

pkg/graveler/committed/manager.go
guy-har: The diff might be confusing because `Apply` logic moved into `applyOnDiffWithRanges`
`Apply` still exists, it now calls `applyOnDiffWithRanges`.
`Apply` gets a diff with no ranges, because graveler is not aware of ranges.

pkg/graveler/committed/merge_iterator_test.go
guy-har: Thanks 😄

pkg/graveler/committed/meta_range.go
guy-har: @arielshaqed , I changed the documentation of the `DiffIterator`, PTAL

pkg/graveler/committed/meta_range.go
```diff
@@ -26,15 +26,26 @@ type Iterator interface {
...
+//
+// left        [min].R1.[max]                     [min].R3.[max]        [min]...............R5..............[max]
+//        ------------------------------------------------------------------------------------------------
+// right                        [min].R2.[max]     [min.....R4....max]  [min].R6.[max] [min].R7.[max]
+//
+// R1 -  will return as diff with header
+// R2 - will return as diff with header
+// R3 and R4 - could not return a header because we must enter the ranges in order to get some header values (such as count)
+// R5 and R6 - same as R3 and R4
+// R7 - in case R5 has no values in the R7 range, R7 would return as a diff with header
```
arielshaqed, line 40: Neat explanation, great drawing!  I would maybe rephrase the tense to make it a bit shorter.

```suggestion
// DiffIterator may contain ranges without headers. For instance, consider the diff of
// these ranges:
//
// left        [min].R1.[max]                     [min].R3.[max]        [min]...............R5..............[max]
// ------------------------------------------------------------------------------------------------
// right                        [min].R2.[max]     [min.....R4....max]  [min].R6.[max] [min].R7.[max]
//
// R1 - returned as a diff inside range R1.
// R2 - returned as a diff inside range R2.
// R3 and R4 - not in any existing range (and creating a new range would require consuming them).
// R5 and R6 - as above, not in any existing range.
// R7 - returned as a diff inside range R7 if R5 has no keys inside the range of R7.
```

arielshaqed (approved):
Great pic!

LEBTM (*looks* _even better_ to me), thanks.

PR #1662, opened 2021-03-21
Stabilize TestExecutor__BatchFor

### LINKED ISSUE #1647: TestExecutor_BatchFor is flaky
--- FAIL: TestExecutor_BatchFor (0.15s)
    executor_test.go:73: this should not be called, only r1's

talSofer:
fix #1647

**Background**
To reduce db access frequency, the executor intends to batch identical requests that are made in a pre-defined delay, and TestExecutor_BatchFor is to prove that read after write consistency is preserved with the batching logic. 
In this unit test, we simulate the following scenario: 
1. reader (r1) starts (Current version: v0)
2. writer (w1) writes v1
3. writer (w1) returns (Current version: v1)
4. reader (r2) starts
5. both readers (r1,r2) return with v1 as their response.

The root cause for the bug was that the test didn't guarantee that r1 and r2 are part of the same batch, causing r2 to trigger an additional call to the db. 

**Change list** 
1. Stabilize TestExecutor_BatchFor by forcing reader1 and reader2 to be part of the same batch. 
2. Fix the test failures in https://github.com/treeverse/lakeFS/runs/2119092066?check_suite_focus=true by eliminating the usage of t.Fatalf
3. Add two more unit tests for the batcher functionality.

### DIFF
--- a/pkg/batch/executor.go
+++ b/pkg/batch/executor.go
@@ -28,7 +28,7 @@ func (b BatchFn) Execute() (interface{}, error) {
 type DelayFn func(dur time.Duration)
 
 type Batcher interface {
-	BatchFor(key string, dur time.Duration, fn BatchFn) (interface{}, error)
+	BatchFor(key string, dur time.Duration, exec Executable) (interface{}, error)
 }
 
 type nonBatchingExecutor struct {
@@ -47,7 +47,6 @@ type request struct {
 	key        string
 	timeout    time.Duration
 	exec       Executable
-	onBatched  chan *response
 	onResponse chan *response
 }
 
--- a/pkg/batch/executor.go
+++ b/pkg/batch/executor.go
@@ -16,7 +16,7 @@ type Executer interface {
 }
 
 type BatchTracker interface {
-	// This method should be called on adding a request to an existing batch
+	// Batched is called when a request is added to an existing batch.
 	Batched()
 }
 
@@ -35,8 +35,8 @@ type Batcher interface {
 type nonBatchingExecutor struct {
 }
 
-func (n *nonBatchingExecutor) BatchFor(_ string, _ time.Duration, ft Executer) (interface{}, error) {
-	return ft.Execute()
+func (n *nonBatchingExecutor) BatchFor(_ string, _ time.Duration, exec Executer) (interface{}, error) {
+	return exec.Execute()
 }
 
 type response struct {
--- a/pkg/batch/executor.go
+++ b/pkg/batch/executor.go
@@ -11,11 +11,12 @@ import (
 // dispatching new ones would start blocking.
 const RequestBufferSize = 1 << 17
 
-type Executable interface {
+type Executer interface {
 	Execute() (interface{}, error)
 }
 
-type Batchable interface {
+type BatchTracker interface {
+	// This method should be called on adding a request to an existing batch
 	Batched()
 }
 
@@ -28,14 +29,14 @@ func (b BatchFn) Execute() (interface{}, error) {
 type DelayFn func(dur time.Duration)
 
 type Batcher interface {
-	BatchFor(key string, dur time.Duration, exec Executable) (interface{}, error)
+	BatchFor(key string, dur time.Duration, exec Executer) (interface{}, error)
 }
 
 type nonBatchingExecutor struct {
 }
 
-func (n *nonBatchingExecutor) BatchFor(_ string, _ time.Duration, exec Executable) (interface{}, error) {
-	return exec.Execute()
+func (n *nonBatchingExecutor) BatchFor(_ string, _ time.Duration, ft Executer) (interface{}, error) {
+	return ft.Execute()
 }
 
 type response struct {
@@ -46,7 +47,7 @@ type response struct {
 type request struct {
 	key        string
 	timeout    time.Duration
-	exec       Executable
+	exec       Executer
 	onResponse chan *response
 }
 
@@ -75,7 +76,7 @@ func NewExecutor(logger logging.Logger) *Executor {
 	}
 }
 
-func (e *Executor) BatchFor(key string, timeout time.Duration, exec Executable) (interface{}, error) {
+func (e *Executor) BatchFor(key string, timeout time.Duration, exec Executer) (interface{}, error) {
 	cb := make(chan *response)
 	e.requests <- &request{
 		key:        key,
@@ -102,7 +103,7 @@ func (e *Executor) Run(ctx context.Context) {
 					e.execs <- req.key
 				}(req)
 			} else {
-				if b, ok := req.exec.(Batchable); ok {
+				if b, ok := req.exec.(BatchTracker); ok {
 					b.Batched()
 				}
 				e.waitingOnKey[req.key] = append(e.waitingOnKey[req.key], req)
--- a/pkg/batch/executor.go
+++ b/pkg/batch/executor.go
@@ -11,7 +11,19 @@ import (
 // dispatching new ones would start blocking.
 const RequestBufferSize = 1 << 17
 
+type Executable interface {
+	Execute() (interface{}, error)
+}
+
+type Batchable interface {
+	Batched()
+}
+
 type BatchFn func() (interface{}, error)
+
+func (b BatchFn) Execute() (interface{}, error) {
+	return b()
+}
 
 type DelayFn func(dur time.Duration)
 
@@ -22,8 +34,8 @@ type Batcher interface {
 type nonBatchingExecutor struct {
 }
 
-func (n *nonBatchingExecutor) BatchFor(key string, dur time.Duration, fn BatchFn) (interface{}, error) {
-	return fn()
+func (n *nonBatchingExecutor) BatchFor(_ string, _ time.Duration, exec Executable) (interface{}, error) {
+	return exec.Execute()
 }
 
 type response struct {
@@ -34,7 +46,8 @@ type response struct {
 type request struct {
 	key        string
 	timeout    time.Duration
-	fn         BatchFn
+	exec       Executable
+	onBatched  chan *response
 	onResponse chan *response
 }
 
@@ -63,12 +76,12 @@ func NewExecutor(logger logging.Logger) *Executor {
 	}
 }
 
-func (e *Executor) BatchFor(key string, timeout time.Duration, fn BatchFn) (interface{}, error) {
+func (e *Executor) BatchFor(key string, timeout time.Duration, exec Executable) (interface{}, error) {
 	cb := make(chan *response)
 	e.requests <- &request{
 		key:        key,
 		timeout:    timeout,
-		fn:         fn,
+		exec:       exec,
 		onResponse: cb,
 	}
 	response := <-cb
@@ -83,20 +96,25 @@ func (e *Executor) Run(ctx context.Context) {
 		case req := <-e.requests:
 			// see if we have it scheduled already
 			if _, exists := e.waitingOnKey[req.key]; !exists {
+				e.waitingOnKey[req.key] = []*request{req}
 				// this is a new key, let's fire a timer for it
 				go func(req *request) {
 					e.Delay(req.timeout)
 					e.execs <- req.key
 				}(req)
+			} else {
+				if b, ok := req.exec.(Batchable); ok {
+					b.Batched()
+				}
+				e.waitingOnKey[req.key] = append(e.waitingOnKey[req.key], req)
 			}
-			e.waitingOnKey[req.key] = append(e.waitingOnKey[req.key], req)
 		case execKey := <-e.execs:
 			// let's take all callbacks
 			waiters := e.waitingOnKey[execKey]
 			delete(e.waitingOnKey, execKey)
 			go func(key string) {
 				// execute and call all mapped callbacks
-				v, err := waiters[0].fn()
+				v, err := waiters[0].exec.Execute()
 				if e.Logger.IsTracing() {
 					e.Logger.WithFields(logging.Fields{
 						"waiters": len(waiters),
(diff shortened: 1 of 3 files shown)


pkg/batch/executor.go
```diff
@@ -11,19 +11,31 @@ import (
 // dispatching new ones would start blocking.
 const RequestBufferSize = 1 << 17
 
+type Executable interface {
+	Execute() (interface{}, error)
+}
+
+type Batchable interface {
+	Batched()
```
arielshaqed, line 19: Can you document this method?  I _think_ it's what the JavaScript world would call `OnBatched`.

pkg/batch/executor.go
```diff
@@ -11,19 +11,31 @@ import (
 // dispatching new ones would start blocking.
 const RequestBufferSize = 1 << 17
 
+type Executable interface {
```
arielshaqed, line 14: I think this should be

```suggestion
type Executer interface {
```

according to https://golang.org/doc/effective_go#interface-names.

pkg/batch/executor.go
```diff
@@ -11,19 +11,31 @@ import (
...
+	Execute() (interface{}, error)
+}
+
+type Batchable interface {
```
arielshaqed, line 18: Similarly, this should also be a `-er` if possible.  In this case the name also misled me, because in fact we do allow non-`Batchable` operations to be batched.

pkg/batch/executor_test.go
```diff
@@ -11,6 +11,50 @@ import (
 	"github.com/treeverse/lakefs/pkg/logging"
 )
 
+const accessedOnce = 1
+const accessedTwice = 2
+
+type reader struct {
+	Ch         chan bool
```
arielshaqed, line 18: Can you name this to refer to what it does, or alternatively add a comment?

arielshaqed, line 18: Also, AFAICT this could even be
```suggestion
	Ch         chan struct{}
```
since we only ever close it.

pkg/batch/executor_test.go
```diff
@@ -11,6 +11,50 @@ import (
...
+	Get(key string) string
+	GetCount() int
+}
+
+func (d *db) Insert(key string, val string) {
+	d.KvStore.Store(key, val)
+}
+
+func (d *db) Get(key string) (interface{}, bool) {
+	d.AccessCount++
```
arielshaqed, line 50: Given that `KvStore` is a `sync.Map`, this is a race.  You might need to perform an atomic action here, or just lock the whole `db` and do away with the `sync.Map`.

pkg/batch/executor_test.go
```diff
@@ -24,12 +68,13 @@ func testReadAfterWrite(t *testing.T) {
 	// 3. writer (w1) returns (Current version: v1)
 	// 4. reader (r2) starts
 	// 5. both readers (r1,r2) return with v1 as their response.
-	var db = sync.Map{}
-	db.Store("v", "v0")
+	var db = &db{sync.Map{}, 0}
```
arielshaqed, line 71: I think these are the defaults; you could just use

```suggestion
	var db = &db{}
```

pkg/batch/executor_test.go
```diff
@@ -11,6 +11,50 @@ import (
 	"github.com/treeverse/lakefs/pkg/logging"
 )
 
+const accessedOnce = 1
+const accessedTwice = 2
```
arielshaqed, line 15: Not a fan of explicitly naming "1" and "2": they appear in the names, and cannot be changed without changing the name and all of its occurrences (and checking at each occurrence whether it needs to be kept or changed).  E.g., if for some reason the second number had to change to 17, the change would be
```go
const accessedSeventeenTimes = 17
```
rather than
```go
const accessedTwice = 17
```
and we would _still_ need to go over each `accessedTwice` to determine how many times we now want it to be accessed.

The same holds for `accessedOnce`: wherever we use it, it's easier for me just to read the number 1.

arielshaqed (changes requested):
Cool, thanks!

Only _required_ change is the race on the counter in the test code.

pkg/batch/executor_test.go
```diff
@@ -11,6 +11,50 @@ import (
...
+func (r *reader) Batched() {
+	close(r.Ch)
+}
+
+type db struct {
+	KvStore     sync.Map
+	AccessCount int
+}
+
+type Dao interface {
```
nopcoder, line 39: No need to Dao interface

pkg/batch/executor_test.go
```diff
@@ -40,49 +85,154 @@ func testReadAfterWrite(t *testing.T) {
...
+	}(&r2)
+
 	<-read1Done
 	<-read2Done
+
+	r1Succeeded := r1.IsExecuted
+	r2Succeeded := r2.IsExecuted
+	if !(r1Succeeded && r2Succeeded) {
+		t.Error("both r1 and r2's exec functions should be executed but r1Succeeded=#{r1Succeeded} " +
+			"and r2Succeeded=#{r1Succeeded}")
```
itaiad200, line 234:
```suggestion
			"and r2Succeeded=#{r2Succeeded}")
```

itaiad200 (changes requested):
Great work on solving the flakiness and adding an important coverage to the executor! 
Requesting changes mainly due to a few nits, but otherwise looks good!

pkg/batch/executor_test.go
```diff
@@ -11,6 +11,50 @@ import (
 	"github.com/treeverse/lakefs/pkg/logging"
 )
 
+const accessedOnce = 1
+const accessedTwice = 2
+
+type reader struct {
+	Ch         chan bool
+	IsExecuted bool
+	Fn         ExecFn
```
nopcoder, line 20: Think that adding `Fn` callback as part of the reader which is a callback can make the code a bit complex.
Based on the use of this function I suggest just adding another channel, one that the reader will close when it is executed.

nopcoder (changes requested):

pkg/batch/executor_test.go
```diff
@@ -11,6 +11,50 @@ import (
 	"github.com/treeverse/lakefs/pkg/logging"
 )
 
+const accessedOnce = 1
+const accessedTwice = 2
+
+type reader struct {
```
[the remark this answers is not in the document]
talSofer: Yes, thanks :)

pkg/batch/executor_test.go
talSofer: > cannot be changed without changing the name and all of its occurrences
Good point, I removed the constants because they are no longer used by multiple unit tests but I agree.

pkg/batch/executor_test.go
```diff
@@ -40,49 +85,154 @@ func testReadAfterWrite(t *testing.T) {
...
+	go func() {
+		<-read1Done // ensure r2 starts after r1 has returned
+		exec.BatchFor("k", time.Millisecond*50, batch.BatchFn(func() (interface{}, error) {
+			version, _ := db.Get("v")
+			return version, nil
+		}))
+		ac := db.GetAccessCount()
+		if ac != accessedTwice {
+			t.Errorf("r2 should have triggered the second db call, but access count= #{ac}")
 		}
```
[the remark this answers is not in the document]
talSofer: I changed the test logic so this part of the test no longer exists

pkg/batch/executor.go
```diff
@@ -11,8 +11,20 @@ import (
...
```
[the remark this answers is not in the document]
talSofer: This is what I changed it to  would you add additional documentation to it?   
```
type BatchTracker interface {
	// This method should be called on adding a request to an existing batch
	Batched()
}

File: pkg/batch/executor.go [hunk 14]
```diff
@@ -83,20 +95,25 @@ func (e *Executor) Run(ctx context.Context) {
...
+				e.waitingOnKey[req.key] = []*request{req}
 				// this is a new key, let's fire a timer for it
 				go func(req *request) {
 					e.Delay(req.timeout)
 					e.execs <- req.key
 				}(req)
+			} else {
+				if b, ok := req.exec.(Batchable); ok {
+					b.Batched()
+				}
```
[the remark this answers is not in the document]
talSofer: You are right, there is is no chance that `req.exec` is not batchable, that was a misleading interface name. In the updated version of this code it is called "batchTracker". and there could be `req.exe` that are not trackable.

itaiad200 (approved):
More tests and less flakiness, good job!

pkg/batch/executor.go
```diff
@@ -11,11 +11,12 @@ import (
...
 const RequestBufferSize = 1 << 17
 
-type Executable interface {
+type Executer interface {
 	Execute() (interface{}, error)
 }
 
-type Batchable interface {
+type BatchTracker interface {
+	// This method should be called on adding a request to an existing batch
```
arielshaqed, line 19: Go has weird documentation rules.  Here it should probably be
```suggestion
	// Batched is called when a request is added to an existing batch.
```

pkg/batch/executor.go
```diff
@@ -28,14 +29,14 @@ func (b BatchFn) Execute() (interface{}, error) {
...
-	BatchFor(key string, dur time.Duration, exec Executable) (interface{}, error)
+	BatchFor(key string, dur time.Duration, exec Executer) (interface{}, error)
 }
 
 type nonBatchingExecutor struct {
 }
 
-func (n *nonBatchingExecutor) BatchFor(_ string, _ time.Duration, exec Executable) (interface{}, error) {
-	return exec.Execute()
+func (n *nonBatchingExecutor) BatchFor(_ string, _ time.Duration, ft Executer) (interface{}, error) {
```
arielshaqed, line 38: Not a fan of this renaming: it is not clear what "ft" means.  (I would be OK with a single-letter name because the scope is so short, and the single letter clearly means nothing.  Just "`ft`" makes me try to glean a meaning, which is unimportant here.)

arielshaqed (approved):
Thanks, nice work!

As discussed F2F, I find the channel-based reporting trickier to understand than an atomic variable would have been.  However, given that the standard Go API for atomics is so easy to skip, I can understand why you implement nonblocking signalling by channel.

No comments are required (let alone blocking).

pkg/batch/executor.go
talSofer: nice catch thanks! this are the leftovers from the experimental rename I did to fetcher. I changed it back to exec

pkg/batch/executor_test.go
```diff
@@ -115,12 +117,15 @@ func testReadAfterWrite(t *testing.T) {
...
-		if r.IsExecuted || db.GetAccessCount() != accessedOnce {
+		te := trackableExecuter{batchTracker: read2Batched, execTracker: make(chan struct{})}
+		r2, _ := exec.BatchFor("k", 50*time.Millisecond, &te)
+		r2v := r2.(string)
+		if r2v != "v1" {
+			t.Errorf("expected r2 to get v1, got %s instead", r2v)
+		}
+
+		// We expect r2's exec function to not execute because it should join r1's batch
+		if te.WasExecuted() || db.GetAccessCount() != 1 {
```
nopcoder, line 128: can you split the check into 2 if(s) so when it will fail we will know which one failed or capture the result of the check and print it

pkg/batch/executor_test.go
```diff
@@ -198,40 +199,34 @@ func testBatchByKey(t *testing.T) {
...
 
-	r1Succeeded := r1.IsExecuted
-	r2Succeeded := r2.IsExecuted
+	r1Succeeded := te1.WasExecuted()
+	r2Succeeded := te2.WasExecuted()
 	if !(r1Succeeded && r2Succeeded) {
-		t.Error("both r1 and r2's exec functions should be executed but r1Succeeded=#{r1Succeeded} " +
-			"and r2Succeeded=#{r1Succeeded}")
+		t.Error("both r1's and r2's exec functions should be executed but r1Succeeded=#{r1Succeeded} " +
+			"and r2Succeeded=#{r2Succeeded}")
```
nopcoder, line 229:
```suggestion
		t.Errorf("both r1's and r2's exec functions should be executed but r1Succeeded=%t and r2Succeeded=%t", r1Succeeded, r2Succeeded)
```

pkg/batch/executor_test.go
```diff
@@ -133,42 +138,37 @@ func testReadAfterWrite(t *testing.T) {
...
-			version, _ := db.Get("v")
-			return version, nil
+		r1, _ := exec.BatchFor("k", time.Millisecond*50, batch.BatchFn(func() (interface{}, error) {
+			return "v1", nil
 		}))
-		ac := db.GetAccessCount()
-		if ac != accessedOnce {
-			t.Errorf("r1 should have triggered a single db call, but access count= #{ac}")
+		if r1 != "v1" {
+			t.Errorf("expected r1 to get v1 but got #{r1} instead")
```
nopcoder, line 160:
```suggestion
			t.Errorf("expected r1 to get v1 but got '%s' instead", r1)
```

pkg/batch/executor_test.go
```diff
@@ -133,42 +138,37 @@ func testReadAfterWrite(t *testing.T) {
...
-			version, _ := db.Get("v")
-			return version, nil
+		r2, _ := exec.BatchFor("k", time.Millisecond*50, batch.BatchFn(func() (interface{}, error) {
+			return "v2", nil
 		}))
-		ac := db.GetAccessCount()
-		if ac != accessedTwice {
-			t.Errorf("r2 should have triggered the second db call, but access count= #{ac}")
+		if r2 != "v2" {
+			t.Errorf("expected r2 to get v2 but got #{r2} instead")
```
nopcoder, line 171:
```suggestion
			t.Errorf("expected r2 to get v2 but got '%s' instead", r2)
```

nopcoder (changes requested):

nopcoder (approved):

PR #1666, opened 2021-03-22
Improve export (sql query, file ordering, errors log)

### LINKED ISSUE #1659: [Export] Improve diff SQL query
Current query uses 2 joins. We should strive for a single join for better performance.

### LINKED ISSUE #1660: [Export] Write '_success' files last
For supporting applications like Spark, it would be helpful of '_success' files are written last, so that consumers know it's safe to read the data.

itaiad200:
close #1659 
close #1660

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
@@ -6,12 +6,15 @@ import org.apache.spark.SerializableWritable
 import org.apache.spark.sql.{DataFrame, Row, SparkSession}
 import java.io.IOException
 import java.net.URL
-import java.time.LocalDateTime
-import java.time.format.DateTimeFormatter
 import scala.util.Random
 
 
-class Exporter(spark : SparkSession, apiClient: ApiClient, repoName: String, dstRoot: String) {
+class Exporter(spark : SparkSession, apiClient: ApiClient, filter: KeyFilter, repoName: String, dstRoot: String) {
+
+  def this(spark : SparkSession, apiClient: ApiClient, repoName: String, dstRoot: String) {
+    this(spark, apiClient, new SparkFilter(), repoName, dstRoot)
+  }
+
   def exportAllFromBranch(branch: String): Unit = {
     val commitID = apiClient.getBranchHEADCommit(repoName, branch)
     exportAllFromCommit(commitID)
@@ -31,22 +34,23 @@ class Exporter(spark : SparkSession, apiClient: ApiClient, repoName: String, dst
     actOnActions(ns, dst, commitID, actionsDF)
   }
 
-  private def export(ns: String, rel: String, commitID: String, actionsDF: DataFrame, successFiles: Boolean) : Boolean =  {
+  final private val maxLoggedErrors = 10000
+  private def export(round: Int, ns: String, rel: String, commitID: String, actionsDF: DataFrame) : Boolean =  {
     import spark.implicits._
 
     val hadoopConf = spark.sparkContext.hadoopConfiguration
     val serializedConf = new SerializableWritable(hadoopConf)
-
+    val f = filter
     val errs = actionsDF.map(
       row =>
-        Exporter.handleRow(ns, rel, serializedConf, row, successFiles)
+        Exporter.handleRow(f, round, ns, rel, serializedConf, row)
     ).filter(err => err != null)
 
-    if (errs.count() == 0){
+    if (errs.isEmpty){
       return true
     }
 
-    writeSummaryFile(false, commitID, errs.reduce(_+"\n"+_))
+    writeSummaryFile(false, commitID, errs.take(maxLoggedErrors).reduce(_+"\n"+_))
     false
   }
 
@@ -83,20 +87,17 @@ class Exporter(spark : SparkSession, apiClient: ApiClient, repoName: String, dst
     actOnActions(ns, dst, commitID, actionsDF)
   }
 
-  final val successMsg = "Export completed successfully!"
+  final private val successMsg = "Export completed successfully!"
   private def actOnActions(ns: String, dst: String, commitID: String, actionsDF: DataFrame): Unit = {
-    if (!export(ns, dst, commitID, actionsDF, false)) {
-      // failed to export files, don't export _success files
-      return
-    }
-    if (!export(ns, dst, commitID, actionsDF, true)){
-      return
+    for( i <- 1 to filter.rounds()){
+      if (!export(i, ns, dst, commitID, actionsDF)) {
+        // failed to export files, don't export next round
+        return
+      }
     }
 
     writeSummaryFile(true, commitID, successMsg)
   }
-
-
 
   private def writeSummaryFile(success: Boolean, commitID: String, content : String) = {
     val suffix = if(success) "SUCCESS" else "FAILURE"
@@ -108,7 +109,7 @@ class Exporter(spark : SparkSession, apiClient: ApiClient, repoName: String, dst
     stream.close()
   }
 
-  final val prefixLen = 8
+  final private val prefixLen = 8
   private def randPrefix() : String = {
     val gen = Random.alphanumeric.dropWhile(_.isDigit)
     gen.take(prefixLen).mkString("")
@@ -116,26 +117,13 @@ class Exporter(spark : SparkSession, apiClient: ApiClient, repoName: String, dst
 }
 
 object Exporter {
-  // spark success files must end with "/_SUCCESS", unless they are in the root
-  // dir, then they should be equal to "_SUCCESS".
-  final val sparkSuccessFileRegex = "((.*\\/_SUCCESS)|(^_SUCCESS))$".r
-
-  private def handleRow(ns: String, rootDst: String, serializedConf: SerializableWritable[Configuration], row: Row, successFiles :Boolean): String =  {
+  private def handleRow(filter: KeyFilter, round:Int, ns: String, rootDst: String, serializedConf: SerializableWritable[Configuration], row: Row) : String =  {
     val action = row(0)
     val key = row(1).toString()
     val address = row(2).toString()
 
-    key match {
-      case sparkSuccessFileRegex() =>
-        if(!successFiles) {
-          // key match, end gracefully
-          return null
-        }
-      case default =>
-        if (successFiles) {
-          // key doesn't match, end gracefully
-          return null
-        }
+    if (!filter.shouldHandleKey(key, round)){
+      return null
     }
 
     val conf = serializedConf.value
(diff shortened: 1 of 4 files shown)


clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
```diff
@@ -42,15 +43,18 @@ def exportFrom(branch: String, prevCommitID: String): Unit = {
...
-      "LEFT JOIN prev_commit " +
-      "ON (new_commit.etag = prev_commit.etag AND new_commit.key = prev_commit.key) " +
-      "WHERE prev_commit.key is NULL " +
-      "UNION " +
-      "SELECT 'delete' as action, prev_commit.* FROM prev_commit " +
-      "LEFT OUTER JOIN new_commit " +
-      "ON (new_commit.key = prev_commit.key) " +
-      "WHERE new_commit.key is NULL")
+
+    val actionsDF = spark.sql("WITH raw_rows AS (SELECT n.key as nkey, n.address as naddress, n.etag as netag, " +
```
arielshaqed, line 47: I think I would find this easier to read as a subselect than as a CTE -- in particular the fact that `raw_rows` is elided in the query confuses me.

I.e. it could like something like

```sql
SELECT CASE ... END as action, CASE ... AS key, ... FROM (
    SELECT n.key AS nkey, ... FROM new_commit n FULL OUTER JOIN prev_commit p ...
)
```

arielshaqed, line 47: Can we use a multi-line literal?  (`"""..."""`)

arielshaqed (changes requested):
Looks neat.  Did you benchmark?

itaiad200:
> Looks neat. Did you benchmark?

Not since the initial benchmarking. Making some more improvements and then will benchmark again.

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
```diff
@@ -38,38 +45,51 @@ class Exporter(spark : SparkSession, apiClient: ApiClient, repoName: String, dst
...
+    """)
+
+    export(ns, dst, actionsDF, false)
+    export(ns, dst, actionsDF, true)
   }
 }
 
 object Exporter {
-  private def handleRow(ns: String, rel: String, serializedConf: SerializableWritable[org.apache.hadoop.conf.Configuration], row: Row): Unit =  {
+  val sparkSuccessFileSuffix = "_SUCCESS"
```
arielshaqed, line 78: This probably needs to be a configuration parameter.  `_SUCCESS` is a great default, but I'm sure other values will be useful.  Also I expect a regexp (or other rule engine) to be a better choice -- e.g. I can export success files last _in some directories_

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
```diff
@@ -38,38 +45,51 @@ class Exporter(spark : SparkSession, apiClient: ApiClient, repoName: String, dst
...
+  private def handleRow(ns: String, rootDst: String, serializedConf: SerializableWritable[Configuration], row: Row, successFiles: Boolean): Unit =  {
     val action = row(0)
-    val key = row(1)
+    val key = row(1).toString()
     val address = row(2).toString()
-    val conf = serializedConf.value
 
-    val srcPath = new org.apache.hadoop.fs.Path( if (address.contains("://")) address else ns + "/" + address)
-    val dstPath = new org.apache.hadoop.fs.Path(rel + "/" + key)
+    val isSuccessFile = key.endsWith(sparkSuccessFileSuffix)
```
arielshaqed, line 85: This incorrectly identifies `s3://a/b/c/NOT_A_SUCCESS` as a success file.

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
```diff
@@ -38,38 +45,51 @@ class Exporter(spark : SparkSession, apiClient: ApiClient, repoName: String, dst
...
+    if (isSuccessFile != successFiles){
```
arielshaqed, line 86: nit:
```suggestion
    if (isSuccessFile != successFiles) {
```

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
```diff
@@ -12,22 +16,25 @@ class Exporter(spark : SparkSession, apiClient: ApiClient, repoName: String, dst
...
 
     // pin Exporter field to avoid serialization
     val dst = dstRoot
-    val actionsDF = spark.sql("SELECT 'copy' as action, * FROM commit")
+    val actionsDF = spark.sql(s"SELECT 'copy' as action, * FROM ${tableName}")
 
-    export(ns, dst, actionsDF)
-    spark.sparkContext.stop()
+    export(ns, dst, actionsDF, false)
+    export(ns, dst, actionsDF, true)
```
arielshaqed, line 28: :warning: Must verify that if the first export fails, the second does not go through.

arielshaqed (changes requested):
Neat!

Only change required in this PR is not to try to export success files if any other files fail to export.

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
itaiad200: Deferred for now, makes the code cumbersome and I need to think of a way to pass this in an orderly manner.

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
```diff
@@ -69,22 +80,62 @@ class Exporter(spark : SparkSession, apiClient: ApiClient, repoName: String, dst
...
+    val gen = Random.alphanumeric.dropWhile(_.isDigit)
+    gen.take(prefixLen).mkString("")
   }
 }
 
 object Exporter {
-  val sparkSuccessFileSuffix = "_SUCCESS"
+  // spark success files must end with "/_SUCCESS", unless they are in the root
+  // dir, then they should be equal to "_SUCCESS".
+  final val sparkSuccessFileRegex = "((.*\\/_SUCCESS)|(^_SUCCESS))$".r
```
arielshaqed, line 121:
```suggestion
  final val sparkSuccessFileRegex = "((.*/_SUCCESS)|(^_SUCCESS))$".r
```
or
```suggestion
  final val sparkSuccessFileRegex = raw"((.*/_SUCCESS)|(^_SUCCESS))$".r
```
(see https://www.scala-lang.org/api/current/scala/util/matching/Regex.html)
or improve the regex,
```suggestion
  final val sparkSuccessFileRegex = "(^|/)_SUCCESS$".r
```

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
```diff
@@ -69,22 +80,62 @@ class Exporter(spark : SparkSession, apiClient: ApiClient, repoName: String, dst
...
 
-  private def handleRow(ns: String, rootDst: String, serializedConf: SerializableWritable[Configuration], row: Row, successFiles: Boolean): Unit =  {
+  private def handleRow(ns: String, rootDst: String, serializedConf: SerializableWritable[Configuration], row: Row, successFiles :Boolean): String =  {
```
arielshaqed, line 123: nit: spacing around ":Boolean".

arielshaqed (commented):
Neat, thanks!

Worried about huge failure files being held in memory on the driver.

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
```diff
@@ -12,79 +20,162 @@ class Exporter(spark : SparkSession, apiClient: ApiClient, repoName: String, dst
...
-    action match {
-      case "delete" => {
-        dstFS.delete(dstPath,false) : Unit
+      action match {
+        case "delete" => {
+          try {
+            dstFS.delete(dstPath, false)
+            null
+          } catch {
+            case e : (IOException) =>  s"Unable to delete file ${dstPath.toString}: ${e.toString}"
```
arielshaqed, line 153: Why not catch _any_ `Exception`?  It failed in any such case, no?  (Or is the intent to fail by throwing an exception, and let Spark handle that?)

itaiad200: Yes, anything else is unexpected and for all we know should fail Spark entirely.

clients/spark/core/src/main/scala/io/treeverse/clients/KeyFilter.scala
```diff
@@ -0,0 +1,12 @@
...
+
+trait KeyFilter extends Serializable {
+  // Number of rounds for this matcher.
+  // Matcher logic could be changed based on the number of rounds.
+  // For example, export all non-success file on the first round,
+  // and all success files on the seconds round.
+  def rounds(): Int
+
+  // Given the key and the round, decide whether the key should be handled.
+  def shouldHandleKey(key: String, round: Int) : Boolean
```
arielshaqed, line 11: This is odd: a `KeyFilter` can decide to handle the same key multiple times?  Otherwise, consider
```suggestion
  def shouldHandleKey(key: String): Int
```

clients/spark/core/src/main/scala/io/treeverse/clients/SparkFilter.scala
```diff
@@ -0,0 +1,20 @@
...
+  // spark success files must end with "/_SUCCESS", unless they are in the root
+  // dir, then they should be equal to "_SUCCESS".
+  final private val sparkSuccessFileRegex = "(/|^)_SUCCESS$".r
+
+  override def rounds(): Int = 2
+
+  override def shouldHandleKey(key: String, round:  Int): Boolean = {
+    val isMatch = sparkSuccessFileRegex.findFirstIn(key) != None
+
+    round match{
```
arielshaqed, line 14: nit:
```suggestion
    round match {
```

clients/spark/core/src/main/scala/io/treeverse/clients/KeyFilter.scala
```diff
@@ -0,0 +1,12 @@
+package io.treeverse.clients
+
+trait KeyFilter extends Serializable {
+  // Number of rounds for this matcher.
+  // Matcher logic could be changed based on the number of rounds.
```
arielshaqed, line 5:
```suggestion
  // Rounds are counted from 1, allowing matcher logic could be changed based on the number of rounds.
```
(the important bit is to count from 1)

arielshaqed (approved):
Very neat, thanks!

Just asking for some additional clarity around `KeyFilter`, nothing required.

PR #1667, opened 2021-03-22
Feature/openapi3

nopcoder:
Using OpenAPI3 using https://github.com/deepmap/oapi-codegen

Modified the server template code to parse incoming body in case of one body in JSON format.
Removed the python code which verifies the schema, it is not compatible with v3.

TODOs

- [x] align the object put/get API to be compatible with the previous version
- [x] test must of lakectl commands with the previous version


api/swagger.yml
```diff
@@ -1,507 +1,795 @@
...
+
+servers:
+  - url: '/api/v1'
+    description: lakeFS server endpoint
+    variables:
+      protocol:
+        enum:
+          - http
+          - https
+        default: http
```
arielshaqed, line 17: I don't understand, when is this used?  (If I did understand, I would want to avoid `http` in favour of `https`, to have the client encrypt by default.)

Reading [server URL format](https://swagger.io/docs/specification/api-host-and-base-path/#url-format), I would expect these variables to have no effect, as they do not appear in the `url`.  Reading further, I would expect
```yaml
servers:
  - url: http://localhost:8000/api/v1
```

to give a single useful default.  Or perhaps

```yaml
    servers:
      - url: http://localhost:8000/api/v1
      - url: '{protocol}://{endpoint}/api/v1'
        variables:
          protocol:
            enum:
              - http
              - https
            default: https
          endpoint:
            default: localhost:8000
```

api/swagger.yml
```diff
@@ -1,507 +1,795 @@
...
-        pattern: '^[a-z0-9][a-z0-9-]{2,62}$'
-      storage_namespace:
-        type: string
-        description: "Filesystem URI to store the underlying data in (e.g. 's3://my-bucket/some/path/')"
-        example: "s3://example-bucket/"
-        pattern: '^(s3|gs|https?|mem|local|transient)://.*$'
-      default_branch:
-        example: "master"
-        type: string
+        minimum: -1
```
arielshaqed, line 49: Can we cap it at minimum 0 and have "missing" to mean "use server default"?

api/swagger.yml
```diff
@@ -1,507 +1,795 @@
...
+          schema:
+            $ref: "#/components/schemas/Error"
+    ValidationError:
+      description: Validation Error
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/Error"
+
+
```
arielshaqed, line 79: Nit: I prefer a single space here.  Double spaces waste space and are hard to maintain.

api/swagger.yml
```diff
@@ -1,507 +1,795 @@
...
+  schemas:
+
```
arielshaqed, line 81: ninit: I prefer no space here.

api/swagger.yml
```diff
@@ -1,507 +1,795 @@
...
+    Pagination:
+      type: object
+      required:
+        - has_more
+        - max_per_page
```
arielshaqed, line 86: Can you document this response type?  Specifically, why this parameter?

api/swagger.yml
```diff
@@ -1,507 +1,795 @@
...
+      type: object
+      required:
+        - creation_date
+        - id
+        - default_branch
+        - storage_namespace
+      properties:
+        id:
+          type: string
+        creation_date:
```
arielshaqed, line 111: Let's document what kind of `int64` this is! (And also other "times").

api/swagger.yml
```diff
@@ -1,507 +1,795 @@
...
+          type: integer
+          format: int64
+        default_branch:
+          example: "master"
```
arielshaqed, line 115: Good chance to go for `example: "main"` or `example: "trunk"` (your call).  We know we should, we know we shall, we're touching it now...

api/swagger.yml
```diff
@@ -1,507 +1,795 @@
...
+            $ref: "#/components/schemas/Error"
+    ValidationError:
+      description: Validation Error
+      content:
+        application/json:
+          schema:
+            $ref: "#/components/schemas/Error"
+
+
+  schemas:
```
arielshaqed, line 80:
```suggestion
  schemata:
```
# (j/k) :stuck_out_tongue_closed_eyes:

api/swagger.yml
```diff
@@ -1,507 +1,795 @@
...
+        - storage_namespace
+      properties:
+        name:
+          type: string
+          pattern: '^[a-z0-9][a-z0-9-]{2,62}$'
+        storage_namespace:
+          type: string
+          description: "Filesystem URI to store the underlying data in (e.g. 's3://my-bucket/some/path/')"
+          example: "s3://example-bucket/"
+          pattern: '^(s3|gs|https?|mem|local|transient)://.*$'
```
arielshaqed, line 172: Possibly silly question: Can we tell our code generator to validate this only on the server??  Otherwise we add support for random backing protocol `abc`, and every client has to change.

api/swagger.yml
```diff
@@ -1,507 +1,795 @@
...
-    properties:
-      id:
-        type: string
-      commit_id:
-        type: string
+    UnderlyingObjectProperties:
+      type: object
+      properties:
+        storage_class:
+          x-nullable: true
```
arielshaqed, line 238: I think this was a Swaggerism.  Is this still TWTDI in OpenAPI 3?  I _think_ I added this because Swagger used OpenAPI 2 but go-swagger couldn't distinguish between a missing and an empty value without it.  Here I believe you might be able to say:
```suggestion
          nullable: true
```

Or just (or additionally) allow it to drop outright, and let our New Swagger Code Generator DTRT and distinguish between missing and empty.

api/swagger.yml
```diff
@@ -1,507 +1,795 @@
...
       security: [ ]
-      parameters:
-        - in: body
-          name: user
-          schema:
-            $ref: "#/definitions/setup"
+      requestBody:
+        required: true
+        content:
+          application/json:
```
arielshaqed, line 769: This will get annoying fairly quickly.  Can we somehow default this away, like we could with Swagger?

api/swagger.yml
```diff
@@ -1231,85 +1420,71 @@ paths:
...
-        401:
-          $ref: "#/responses/Unauthorized"
-        default:
-          description: generic error response
-          schema:
-            $ref: "#/definitions/error"
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/RepositoryList'
```
arielshaqed, line 1444: The mix of single and double quotes is confusing.  The difference (https://yaml.org/spec/1.2/spec.html#style/flow/single-quoted, https://yaml.org/spec/1.2/spec.html#style/flow/double-quoted is in how they handle escaping.  But here we don't need _any_ quoting AFAICT; using both is strange.

api/swagger.yml
```diff
@@ -1339,152 +1512,138 @@ paths:
...
         name: repository
         required: true
-        type: string
+        schema:
+          type: string
     put:
       tags:
         - refs
-      operationId: dump
+      operationId: dumpRefs
```
arielshaqed, line 1531: Previously we relied on the first tag to give a prefix.  If we now need to repeat it, consider putting the tag at the beginning for better scoping.

api/swagger.yml
```diff
@@ -2207,38 +2332,37 @@ paths:
...
+              schema:
+                $ref: "#/components/schemas/ObjectStats"
         401:
-          $ref: "#/responses/Unauthorized"
+          $ref: "#/components/responses/Unauthorized"
         404:
-          description: path or branch not found
-          schema:
-            $ref: "#/definitions/error"
+          $ref: "#/components/responses/NotFound"
```
arielshaqed, line 2342: Here we have a less informative description of the response.  Consider having multiple versions of `NotFound`, each with different documentation "path or branch [or repo] not found", "branch [or repo] not found", "repo not found".

api/swagger.yml
```diff
@@ -2485,90 +2573,94 @@ paths:
...
       tags:
         - metadata
       operationId: getRange
       summary: return URI to a range file
-      # TODO(ariels): Actually redirect to S3.
       responses:
         200:
           description: range URI
-          headers:
-            Location:
```
arielshaqed, line 2564: This moves us further away from REST.  Ideally this could be a redirect, making it more HTTPish and also making it easier to implement some clients -- e.g. if/once we decide to use presigned URLs.

arielshaqed (commented):
Thanks!

Trying to grow this mega-PR.  Just skimmed swagger.yml, mostly to show my ignorance.  Happy for comments that will tell me what is most urgent for me to read first, the better to review this PR.

api/swagger.yml
nopcoder: Updated the swagger to include just our base URL and removed the schema information.
Why? - the `kin-openapi` package uses middleware to validate the swagger to match the request, the previous. It tried to verify the full URL match for each request and the endpoint value.
Leaving just the patch match for each request just the api/v1 part.

api/swagger.yml
nopcoder: converted all to single space

api/swagger.yml
nopcoder: 😄

api/swagger.yml
nopcoder: I guess this is an implementation detail - in our case the verification is done on the server side by middleware that validates the swagger spec.

api/swagger.yml
nopcoder: guess the openapi2 to 3 didn't catch it - updated to nullable.

api/swagger.yml
nopcoder: sorry, didn't found a way in the spec to default this one.

api/swagger.yml
nopcoder: Converted everything to double-quotes

api/swagger.yml
nopcoder: We only have one 404 - the error description is a resource not found as a general way to describe any of the resources in the way of the request.
A single request can return errors of repository, branch, and reference not found. So the current solution is a more general way to describe it.

api/swagger.yml
nopcoder: Used Oz's conversion - it looks like we wanted to be more like the operation itself in the code than a prefix.
Using the tag as prefix in the code can be modified by the template - but for now if you check the generated code it much cleaner.

api/swagger.yml
nopcoder: missed this one - thanks

api/swagger.yml
nopcoder: Today zero means - I don't want results - just has more indication.
We can change it and have 0 as the server's default, but I think we can do it in a different issue if we all agree that we will lose the capability.

pkg/api/serve.go
```diff
@@ -40,26 +41,22 @@ func Serve(
...
-
-	api.JSONConsumer = runtime.JSONConsumer()
-	api.MultipartformConsumer = runtime.DiscardConsumer
-
-	api.BinProducer = runtime.ByteStreamProducer()
-	api.JSONProducer = runtime.JSONProducer()
+	r := chi.NewRouter()
+	apiRouter := r.With(
+		OapiRequestValidatorWithOptions(swagger, &openapi3filter.Options{
+			AuthenticationFunc: openapi3filter.NoopAuthenticationFunc,
```
itaiad200, line 51: Why not ride the authentication hook provided here instead of adding a custom middleware?
At least document the reason..

pkg/api/serve.go
```diff
@@ -71,25 +68,70 @@ func Serve(
...
+		o.ExcludeRequestBody = true
+		requestValidationInput.Options = &o
+	}
+
+	if err := openapi3filter.ValidateRequest(r.Context(), requestValidationInput); err != nil {
+		var reqErr *openapi3filter.RequestError
+		if errors.As(err, &reqErr) {
+			return http.StatusBadRequest, err
+		}
+		var seqErr *openapi3filter.SecurityRequirementsError
```
itaiad200, line 127: Who returns this error? I thought we're in charge of auth and we don't return this type.

pkg/api/serve.go
```diff
@@ -71,25 +68,70 @@ func Serve(
...
+		Request:    r,
+		PathParams: pathParams,
+		Route:      route,
+		Options:    options,
+	}
+
+	if _, ok := route.Operation.Extensions["x-validation-exclude-body"]; ok {
+		o := *options
+		o.ExcludeRequestBody = true
+		requestValidationInput.Options = &o
```
itaiad200, line 119: Are these options used more than once? i.e. why the special pointers handling?

pkg/api/serve.go
```diff
@@ -71,25 +68,70 @@ func Serve(
...
+			next.ServeHTTP(w, r)
+		})
+	}
+}
+
+func validateRequest(r *http.Request, router routers.Router, options *openapi3filter.Options) (int, error) {
+	// Find route
+	route, pathParams, err := router.FindRoute(r)
+	if err != nil {
+		return http.StatusBadRequest, err // We failed to find a matching route for the request.
```
itaiad200, line 105: Isn't `NotFound` a better error here?

pkg/api/serve.go
```diff
@@ -71,25 +68,70 @@ func Serve(
...
+		if errors.As(err, &reqErr) {
+			return http.StatusBadRequest, err
+		}
+		var seqErr *openapi3filter.SecurityRequirementsError
+		if errors.As(err, &seqErr) {
+			return http.StatusUnauthorized, err
+		}
+		// This should never happen today, but if our upstream code changes,
+		// we don't want to crash the server, so handle the unexpected error.
+		return http.StatusInternalServerError, err
```
itaiad200, line 133: A log here is really helpful for debugging unintended routes.

pkg/api/serve.go
```diff
@@ -71,25 +68,70 @@ func Serve(
...
+	}
 
-	mux := http.NewServeMux()
-	mux.Handle("/_health", httputil.ServeHealth())
-	mux.Handle("/metrics", promhttp.Handler())
-	mux.Handle("/_pprof/", httputil.ServePPROF("/_pprof/"))
-	mux.Handle("/api/", apiHandler)
-	mux.Handle("/swagger.json", apiHandler)
```
itaiad200, line 92: I didn't find where we serve the actual swagger file.

pkg/api/tmpl/chi-interface.tmpl
```diff
@@ -0,0 +1,7 @@
+// ServerInterface represents all server handlers.
```
itaiad200, line 1: Add a github link to the file you replaced here, so we can keep track of changes as the original template changes.

pkg/api/tmpl/chi-interface.tmpl
```diff
@@ -0,0 +1,7 @@
...
+type ServerInterface interface {
+{{range .}}{{.SummaryAsComment }}{{$opid := .OperationId}}
+// ({{.Method}} {{.Path}}) 
+{{.OperationId}}(w http.ResponseWriter, r *http.Request{{if .HasBody}}{{range .Bodies}}{{if and .Default (eq .NameTag "JSON")}}, body {{$opid}}{{.NameTag}}RequestBody {{end}}{{end}}{{end}}{{genParamArgs .PathParams}}{{if .RequiresParamObject}}, params {{.OperationId}}Params{{end}})
```
itaiad200, line 5: Dumb question: Is there a way to break this line and add indentation so it's somewhat readable?

itaiad200: Oh and also - Great job with the templates!!

pkg/api/tmpl/chi-middleware.tmpl
```diff
@@ -0,0 +1,190 @@
+// ServerInterfaceWrapper converts contexts to parameters.
```
itaiad200, line 1: Add a link to the original template and also an explanation of what changed, so it's easier to keep track of changes.

pkg/api/tmpl/chi-middleware.tmpl
```diff
@@ -0,0 +1,190 @@
...
+{{range .}}{{$opid := .OperationId}}
+
+// {{$opid}} operation middleware
+func (siw *ServerInterfaceWrapper) {{$opid}}(w http.ResponseWriter, r *http.Request) {
+  ctx := r.Context()
+  {{if or .RequiresParamObject (gt (len .PathParams) 0) }}
+  var err error
+  {{end}}
+
+  {{if .HasBody}}{{range .Bodies}}{{if and .Default (eq .NameTag "JSON")}}// ------------- Body parse -------------
```
itaiad200, line 18: Add a comment describing what you did here.

pkg/api/controller.go
```diff
@@ -74,2864 +61,2535 @@ type Controller struct {
...
+	if !c.authorize(w, r, []permissions.Permission{
+		{
+			Action:   permissions.CreateGroupAction,
+			Resource: permissions.GroupArn(body.Id),
+		},
+	}) {
+		return
+	}
+	ctx := r.Context()
+	c.LogAction(ctx, "create_group")
```
itaiad200, line 119: It's soooooo much better if you could push this repeatable error prune operation to a generic middleware.

pkg/api/controller.go
```diff
@@ -74,2864 +61,2535 @@ type Controller struct {
...
+			Action:   permissions.ReadGroupAction,
+			Resource: permissions.GroupArn(groupID),
+		},
+	}) {
+		return
+	}
 
-		return returnValue
+	ctx := r.Context()
+	c.LogAction(ctx, "list_user_policies")
```
itaiad200, line 269: For example, an action name error here.

itaiad200 (changes requested):
Haven't gone thru it all yet.. looking great so far

pkg/api/auth_middleware.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, fmt.Errorf("basic auth could not get access key: %w", err))
+		return nil
+	}
+	if secretKey != credentials.AccessSecretKey {
+		logger.WithField("access_key", accessKey).Debug("access key secret does not match")
+		writeError(w, http.StatusUnauthorized, ErrAuthenticationFailed)
+		return nil
+	}
+	user, err := authService.GetUserByID(ctx, credentials.UserID)
```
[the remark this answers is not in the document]
nopcoder: if it was more than one func call I would have consolidated - but we log different things.
adding missing debug log.

pkg/api/auth_middleware.go
```diff
@@ -0,0 +1,118 @@
...
+		writeError(w, http.StatusUnauthorized, ErrAuthenticationFailed)
+		return nil
+	}
+	userData, err := authService.GetUserByID(ctx, cred.UserID)
+	if err != nil {
+		logger.WithFields(logging.Fields{
+			"user_id": cred.UserID,
+			"subject": claims.Subject,
+		}).Debug("could not find user id by credentials")
+		writeError(w, http.StatusUnauthorized, ErrAuthenticationFailed)
```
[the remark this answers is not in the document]
nopcoder: I'll return errors from these functions

pkg/api/auth_middleware.go
```diff
@@ -0,0 +1,118 @@
...
+				if ok {
+					user = userByAuth(ctx, w, logger, authService, accessKey, secretKey)
+					if user == nil {
+						return
+					}
+				}
+			}
+
+			// keep user on the request context
+			r = r.WithContext(context.WithValue(r.Context(), UserContextKey, user))
```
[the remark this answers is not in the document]
nopcoder: adding check

pkg/api/auth_middleware_test.go
```diff
@@ -0,0 +1,53 @@
+package api_test
+
+//
```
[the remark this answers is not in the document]
nopcoder: 😅 on it

pkg/api/serve.go
```diff
@@ -40,26 +41,22 @@ func Serve(
...
+	r := chi.NewRouter()
+	apiRouter := r.With(
+		OapiRequestValidatorWithOptions(swagger, &openapi3filter.Options{
+			AuthenticationFunc: openapi3filter.NoopAuthenticationFunc,
+			ExcludeRequestBody: false,
+		}),
+		RequestCounterMiddleware(),
+		AuthMiddleware(logger, authService),
+		httputil.LoggingMiddleware(RequestIDHeaderName, logging.Fields{"service_name": LoggerServiceName}),
+		MetricsMiddleware(swagger),
```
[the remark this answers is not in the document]
nopcoder: InstrumentHandlerCounter was wrapped into RequestCounterMiddleware.
Can merge the functionality into one - unless we like to count also invalid requests, can move it up the chain.

pkg/api/serve.go
nopcoder: It is based on the middleware implementation, when the request is invalid - doesn't match the swagger schema vs not found which is when the api is valid and the resource is not found (branch, repository, tag and etc...)

pkg/api/serve.go
nopcoder: 1. I'll reverse the order of the Extensions check just to update the options
2. Still I'll have to copy the original options as I like to keep the code generic - use the options the call accepts and just modify the specific flag in the RequestValidationInput.

pkg/api/serve.go
nopcoder: added handler - thanks!

pkg/api/tmpl/chi-interface.tmpl
nopcoder: added some new-lines, let me know if it is better

pkg/api/tmpl/chi-middleware.tmpl
nopcoder: Body parse? if I'll describe the template it will get into the code

pkg/api/serve.go
nopcoder: We use NotFound when a resource in the request is not found like branch, repo or tag and BadRequest (400) when the request is not valid.

pkg/api/serve.go
nopcoder: Debug and trace logging middleware does it for us.

pkg/api/serve.go
nopcoder: added

pkg/api/tmpl/chi-interface.tmpl
nopcoder: added some new lines.

cmd/lakectl/cmd/auth.go
```diff
@@ -12,17 +12,17 @@ import (
 )
 
 var userCreatedTemplate = `{{ "User created successfully." | green }}
-ID: {{ .ID | bold }}
+ID: {{ .Id | bold }}
```
arielshaqed, line 15: This is unfortunate, but can we tweak the codegen to generate ID?  The strict-ish [standard for initialisms](https://github.com/golang/go/wiki/CodeReviewComments#initialisms) says:

> This rule also applies to "ID" when it is short for "identifier" (which is pretty much all cases when it's not the "id" as in "ego", "superego"), so write "appID" instead of "appId".
>
> Code generated by the protocol buffer compiler is exempt from this rule. Human-written code is held to a higher standard than machine-written code.

So:
1. It _is_ required, but...
2. Other robots can claim an exemption so as not to be discriminated vs. Googley tools.

But it still looks odd in Go code, and it *will* confuse us.  I'd be much happier with tweaking our input openAPI spec to generate an `ID` in Go -- if possible.  And preferably before we have other client code that depends on this style.

docs/assets/js/swagger.yml
```diff
@@ -5,25 +5,15 @@ info:
   title: lakeFS API
   version: 0.1.0
 
-
 servers:
-  - url: '/api/v1'
+  - url: "/api/v1"
```
arielshaqed, line 9: Don't think we even need quotes here.

docs/assets/js/swagger.yml
```diff
@@ -89,14 +77,18 @@ components:
...
           type: string
+          description: Token used to retrieve the next page
         results:
           type: integer
           minimum: 0
+          description: The number of values found in the results
         max_per_page:
           type: integer
           minimum: 0
+          description: The maximum value per page that can be returned
```
arielshaqed, line 91:
```suggestion
          description: Maximal number of entries per page
```

docs/assets/js/swagger.yml
```diff
@@ -89,14 +77,18 @@ components:
...
         has_more:
           type: boolean
+          description: Next page is available
         next_offset:
           type: string
+          description: Token used to retrieve the next page
         results:
           type: integer
           minimum: 0
+          description: The number of values found in the results
```
arielshaqed, line 87: nit:
```suggestion
          description: Number of values found in the results
```

But I think this is *not* the total number of values, but somehow the number of entries returned on this page?

docs/assets/js/swagger.yml
```diff
@@ -111,12 +103,12 @@ components:
...
           type: integer
           format: int64
+          description: Unix Epoch in seconds
         default_branch:
-          example: "master"
           type: string
         storage_namespace:
           type: string
-          description: "Filesystem URI to store the underlying data in (e.g. 's3://my-bucket/some/path/')"
+          description: 'Filesystem URI to store the underlying data in (e.g. "s3://my-bucket/some/path/")'
```
arielshaqed, line 111: I don't think we need quotes here.

pkg/api/controller.go
nopcoder: If possible to do the same without losing the BI, prefer to do it in a different PR.

pkg/api/controller.go
nopcoder: agree - if it was aligned with the request OperationID it would have done automatically.
Can be aligned after adding some legacy to the BI.

arielshaqed (commented):
Thanks!  Looks *very* nice.

Partial comments; will do the rest after rebase is pushed.

benchmarks/main.go
```diff
@@ -157,14 +159,26 @@ func uploader(ctx context.Context, ch chan string, repoName, contentPrefix strin
...
+				logger.WithError(err).Error("CreateFormFile failed")
+				return failed
+			}
+			_, err = contentWriter.Write([]byte(content))
+			if err != nil {
+				logger.WithError(err).Error("CreateFormFile write content failed")
+				return failed
+			}
+			w.Close()
+			contentType := w.FormDataContentType()
```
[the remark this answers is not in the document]
nopcoder: Will create a new client package that will include the above - different PR.

cmd/lakectl/cmd/auth.go
nopcoder: Agree, but it will require a fork and PR for oapi-codegen

pkg/api/auth_middleware.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		return nil, ErrAuthenticationFailed
+	}
+	claims, ok := token.Claims.(*jwt.StandardClaims)
+	if !ok || !token.Valid {
+		return nil, ErrAuthenticationFailed
+	}
+	cred, err := authService.GetCredentials(ctx, claims.Subject)
+	if err != nil {
+		logger.WithField("subject", claims.Subject).Debug("could not find credentials for token")
```
arielshaqed, line 89: I think we can log at INFO here:
1. The user already has a valid JWT -- is authenticated.
2. In particular the JWT times match, so it's not some old token.
3. "It should not normally happen."
Let's help the admin debug?

pkg/api/serve.go
arielshaqed: REST says there is no such difference; I have to agree with @itaiad200 here.  `400` is always appropriate (because it says so little), however accessing a URL that is not mounted is a classic 404.

pkg/api/transform.go
```diff
@@ -1,36 +1,20 @@
...
-
-	"github.com/treeverse/lakefs/pkg/api/gen/models"
 	"github.com/treeverse/lakefs/pkg/catalog"
 )
 
 func transformDifferenceTypeToString(d catalog.DifferenceType) string {
 	switch d {
 	case catalog.DifferenceTypeAdded:
-		return models.DiffTypeAdded
+		return "added"
```
arielshaqed, line 10: An advantage of doing it the old way is that you get compile-time checking.  If oapi gives named enum constants for the allowed values, it would be better to continue to use them here.

(Also below, e.g. "allow"/"deny")

arielshaqed (approved):
Neat, very nice!

Required changes:
* Upload errors are handled unsafely in `lakectl` -- succeed on unknown error.

Additional (meta-) comments:
* The "WithResponses" client interface is quite nasty.  Maybe just not use it?  (Not necessary in this PR)

benchmarks/main.go
```diff
@@ -236,3 +240,29 @@ func reader(ctx context.Context, ch chan string, repoName, _ string) int {
...
+
+func (a *APIError) Error() string {
+	return fmt.Sprintf("%s failed: %s", a.Action, a.Message)
+}
+
+func responseAsError(action string, response interface{}) error {
+	r := reflect.ValueOf(response)
+	f := reflect.Indirect(r).FieldByName("HTTPResponse")
+	resp := f.Interface().(*http.Response)
+	if resp.StatusCode >= http.StatusOK && resp.StatusCode < http.StatusMultipleChoices {
```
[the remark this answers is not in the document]
nopcoder: adding api.IsStatusCodeOK func and use it.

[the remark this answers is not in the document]
nopcoder: My hope is if our http client supports redirect it should handle the redirect and return 2xx at the end

benchmarks/main.go
```diff
@@ -72,27 +68,29 @@ func testBenchmarkLakeFS() error {
...
 
-	_, err = client.Branches.CreateBranch(&branches.CreateBranchParams{
-		Branch: &models.BranchCreation{
-			Source: swag.String("master"),
-			Name:   swag.String(branchName),
-		},
-		Repository: repo.Payload.ID,
-		Context:    ctx,
-	}, nil)
+	repo := createRepoResp.JSON201
```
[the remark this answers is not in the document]
nopcoder: will have to update the client code template and support OK and Err which make it easier to use.

cmd/lakectl/cmd/common_helpers.go
```diff
@@ -167,11 +170,45 @@ func DieErr(err error) {
...
+
+func DieOnResponseError(response interface{}, err error) {
+	if err != nil {
+		DieErr(err)
+	}
+	// check http response code
+	var statusCode int
+	if stat, ok := response.(StatusCoder); ok {
+		statusCode = stat.StatusCode()
+		if statusCode >= http.StatusOK && statusCode < http.StatusMultipleChoices {
```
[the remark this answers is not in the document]
nopcoder: api.IsStatusCodeOK done - thanks!

cmd/lakectl/cmd/fs.go
```diff
@@ -103,25 +129,148 @@ var fsCatCmd = &cobra.Command{
...
+	if err != nil {
+		return nil, err
+	}
+	if resp.StatusCode() != http.StatusCreated {
+		apiErrors := []*api.Error{resp.JSON400, resp.JSON401, resp.JSON404, resp.JSONDefault}
+		for _, apiErr := range apiErrors {
+			if apiErr != nil {
+				return nil, fmt.Errorf("%w: %s", ErrRequestFailed, apiErr.Message)
+			}
+		}
```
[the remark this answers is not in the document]
nopcoder: extracted from default and added fallback to the status code message.

pkg/api/auth_middleware.go
```diff
@@ -0,0 +1,119 @@
...
+			if user != nil {
+				r = r.WithContext(context.WithValue(r.Context(), UserContextKey, user))
+			}
+			next.ServeHTTP(w, r)
+		})
+	}
+}
+
+func userByToken(ctx context.Context, logger logging.Logger, authService auth.Service, tokenString string) (*model.User, error) {
+	claims := &jwt.StandardClaims{}
```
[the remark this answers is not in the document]
nopcoder: 1. The idea here what I understand from the code - passing our claims struct, we can make sure that the token holds the same claims type. So after calling parse we can cast the interface we get back and access the members. We can do the same using the generic map as we expect the token to be formatted by the spec.
2. The callback function here is just to get the secret - the parse verifies and returns the token and if it is valid (which I think is true if there is no error, and we can drop the Valid? check) we use the information for user lookup. Valid or non-error parse is saying that the JWT token is valid.

pkg/api/transform.go
nopcoder: Agree I think this is another reason to fork or suggest a PR to have the code generation introduce these constants.
From what I saw it generates constants just for top-level enums.

cmd/lakectl/cmd/auth.go
```diff
@@ -81,11 +84,11 @@ var authUsersCreate = &cobra.Command{
...
 
-		user, err := clt.CreateUser(cmd.Context(), id)
-		if err != nil {
-			DieErr(err)
-		}
-
+		resp, err := clt.CreateUserWithResponse(cmd.Context(), api.CreateUserJSONRequestBody{
+			Id: id,
+		})
+		DieOnResponseError(resp, err)
```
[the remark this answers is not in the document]
nopcoder: Wanted a typed response so I'll keep it this way until I'll have a better client-generated code.

benchmarks/main.go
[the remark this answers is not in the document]
nopcoder: Hope not to get there - one option is to let the underlying HTTP client handle the redirect for example or not use this code for the specific code.

benchmarks/main.go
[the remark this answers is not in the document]
nopcoder: Used WithResponses because I wanted the parsed responses.
Think we need to improve the client template next and make it as easy as possible to access the default OK/success response.

api/swagger.yml
```diff
@@ -1777,6 +1777,28 @@ paths:
         required: true
         schema:
           type: string
+    get:
+      deprecated: true
```
arielshaqed, line 1781: What's the replacement?  Can we document the replacement -- even as a `description`?

cmd/lakectl/cmd/fs.go
```diff
@@ -175,16 +173,28 @@ func uploadObject(ctx context.Context, client api.ClientWithResponsesInterface,
...
+		if resp.JSONDefault != nil {
+			return nil, fmt.Errorf("%w: %s", ErrRequestFailed, resp.JSONDefault.Message)
 		}
+		return nil, fmt.Errorf("%w: %s", ErrRequestFailed, http.StatusText(resp.StatusCode()))
 	}
 	return resp.JSON201, nil
 }
 
+// readSize returns the size of r.
+func readSize(r io.Seeker) (int64, error) {
```
arielshaqed, line 185: I think root.go _also_ performs this very thing, and on trunk I've already refactored (https://github.com/treeverse/lakeFS/blob/933ec252c1ed0176581d9887db615f4abf3354e8/pkg/api/client.go#L746).  Let's add a TODO(ariels) to move it to helpers?

arielshaqed (approved):
Great, thanks!

Please review whether we wnat to return 404 for an unknown path in `validateRequest`.  Any which way, I think this is good (great..) to go!

api/swagger.yml
nopcoder: Added description part of the summary

ozkatz:
🎉

einato:
🎉🎉

PR #1669, opened 2021-03-23
1668 sync quickstart docs sidebar with headers

peacing:
Think it's nice to alignt he quickstart sidebar with headers. Small improvement, but something I noticed.


ozkatz (approved):
Thanks!

itaiad200 (approved):
Looks great!

PR #1675, opened 2021-03-24
publish spark client to s3

johnnyaug:
* Publish artifacts to S3 bucket `treeverse-clients-us-east`.
* Simplify Scala cross build, at the price of coupling the Scala version with the Spark version. This is suitable for all relevant Databricks and EMR versions, so IMO it's enough for now. What we gain is:
  * Better `build.sbt` readability.
  * Simpler artifact naming (the "247"/"301" are ommitted)
  * Ability to import into IntelliJ.
  * By default, sbt only compiles / builds / publishes for the Scala version used to run sbt. Can use `sbt +compile` /  `sbt +publish` / etc. to compile for all (both) versions.

### DIFF
--- a/clients/spark/build.sbt
+++ b/clients/spark/build.sbt
@@ -1,9 +1,7 @@
-import build.BuildType
-
 lazy val baseName = "lakefs-spark"
 
-lazy val projectVersion = "0.1.0-SNAPSHOT.4"
-isSnapshot := true
+lazy val projectVersion = "0.1.0-SNAPSHOT"
+ThisBuild / isSnapshot := true
 
 // Spark versions 2.4.7 and 3.0.1 use different Scala versions.  Changing this is a deep
 // change, so key the Spark distinction by the Scala distinction.  sbt doesn't appear to
@@ -24,21 +22,21 @@ def settingsToCompileIn(dir: String) = {
   )
 }
 
-def generateCoreProject(buildType: BuildType) =
-  Project(s"${baseName}-client-${buildType.name}", file(s"target/core-${buildType.name}"))
+lazy val core =
+  (project in file(s"core"))
     .settings(
+      name := "lakefs-spark-client",
       sharedSettings,
       settingsToCompileIn("core"),
-      scalaVersion := buildType.scalaVersion,
       PB.targets := Seq(
         scalapb.gen() -> (Compile / sourceManaged).value / "scalapb"
       ),
-      libraryDependencies ++= Seq("org.rocksdb" % "rocksdbjni" % "6.6.4",
+      libraryDependencies ++= Seq(
+        "org.rocksdb" % "rocksdbjni" % "6.6.4",
         "commons-codec" % "commons-codec" % "1.15",
-        "org.apache.spark" %% "spark-sql" % buildType.sparkVersion % "provided",
         "com.thesamet.scalapb" %% "scalapb-runtime" % scalapb.compiler.Version.scalapbVersion % "protobuf",
-        "org.apache.hadoop" % "hadoop-aws" % buildType.hadoopVersion,
-        "org.apache.hadoop" % "hadoop-common" % buildType.hadoopVersion,
+        "org.apache.hadoop" % "hadoop-aws" % "2.7.7",
+        "org.apache.hadoop" % "hadoop-common" % "2.7.7",
         "org.scalaj" %% "scalaj-http" % "2.4.2",
         "org.json4s" %% "json4s-native" % "3.7.0-M8",
         "com.google.guava" % "guava" % "16.0.1",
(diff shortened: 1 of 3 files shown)


arielshaqed:
(Commenting separately as that's where "Quote Reply" is easiest.  Code review to follow!)

Not sure about this part of the PR.  I think it impacts our users negatively in order to improve the developer experience on a specific toolchain (IntelliJ).

>     * Simplify Scala cross build, at the price of coupling the Scala version with the Spark version. This is suitable for all relevant Databricks and EMR versions, so IMO it's enough for now. What we gain is:
>       
>       * Better `build.sbt` readability.
>       * Simpler artifact naming (the "247"/"301" are ommitted)

Not sure that's a feature.  The name is `s3://treeverse-clients-us-east/io/treeverse/lakefs-spark-client_2.12/0.1.0-SNAPSHOT/lakefs-spark-client_2.12-0.1.0-SNAPSHOT.jar` and should be used for Spark 3.x, the other will be called `s3://treeverse-clients-us-east/io/treeverse/lakefs-spark-client_2.12/0.1.0-SNAPSHOT/lakefs-spark-client_2.11-0.1.0-SNAPSHOT.jar` and should be used for Spark 2.x.  So it feels to me like the user has to know _more_ in order to pick the right client.

>       * Ability to import into IntelliJ.
>       * By default, sbt only compiles / builds / publishes for the Scala version used to run sbt. Can use `sbt +compile` /  `sbt +publish` / etc. to compile for all (both) versions.

Why is that a feature?

clients/spark/build.sbt
```diff
@@ -24,47 +22,38 @@ def settingsToCompileIn(dir: String) = {
...
-def generateExamplesProject(buildType: BuildType) =
-  Project(s"${baseName}-examples-${buildType.name}", file(s"target/examples-${buildType.name}"))
+lazy val examples =
+  (project in file(s"examples"))
     .settings(
+      name := "lakefs-spark-example",
       sharedSettings,
       settingsToCompileIn("examples"),
-      scalaVersion := buildType.scalaVersion,
-      libraryDependencies += "org.apache.spark" %% "spark-sql" % buildType.sparkVersion % "provided",
```
arielshaqed, line 55: Cool, we don't need this dependency?!

clients/spark/build.sbt
```diff
@@ -24,47 +22,38 @@ def settingsToCompileIn(dir: String) = {
...
-lazy val spark3Type = new BuildType("301", scala212Version, "3.0.1", "0.10.11", "2.7.7")
-
-lazy val core2 = generateCoreProject(spark2Type)
-lazy val core3 = generateCoreProject(spark3Type)
-lazy val examples2 = generateExamplesProject(spark2Type).dependsOn(core2)
-lazy val examples3 = generateExamplesProject(spark3Type).dependsOn(core3)
+    ).dependsOn(core)
 
-lazy val root = (project in file(".")).aggregate(core2, core3, examples2, examples3)
+lazy val root = (project in file(".")).aggregate(core, examples).disablePlugins(PublishMore)
```
arielshaqed, line 56: Why do we disable `PublishMore`?  Isn't `root / publish / skip := true` enough, and/or can't we do something similar with config to sbt-publish-more?

arielshaqed (changes requested):
Great S3 publish (and other fixes)!

I don't think tying the Spark version to the Scala version is a net positive.  As soon as we want to publish another Spark version, we have to revert much of this PR.  As soon as we want to support Spark 2.4.7 with Scala 2.12, we have to revert much of this PR.

This is not academic.  E.g. [Spark download](https://spark.apache.org/docs/2.4.7/#downloading) says:
> Spark runs on Java 8, Python 2.7+/3.4+ and R 3.5+. For the Scala API, Spark 2.4.7 uses Scala 2.12. You will need to use a compatible Scala version (2.12.x).
>
> Note that support for Java 7, Python 2.6 and old Hadoop versions before 2.6.5 were removed as of Spark 2.2.0. Support for Scala 2.10 was removed as of 2.3.0. Support for Scala 2.11 is deprecated as of Spark 2.4.1 and will be removed in Spark 3.0.

Indeed, "2.11" is part of the 2.4.7 artifact name on Maven Central, see e.g. https://search.maven.org/artifact/org.apache.spark/spark-sql_2.11/2.4.7/jar.

I believe this part of the change is a mistake -- but will bow to pressure if others (or you) insist.

clients/spark/build.sbt
johnnyaug: Unfortunately the plugin it doesn't work well with sub-projects, so I had to disable it for the root project. Otherwise it also tries to publish it.

clients/spark/build.sbt
johnnyaug: It was moved to `commonSettings`. But seeing your comments, I will revise this PR.

Issue #1622, opened 2021-03-14
When viewing files by commit ref, Delete option should be disabled


itaiad200:

shamikakumar:
Can I work on this issue ?

itaiad200:
All yours @shamikakumar .
Adding some information: In the below image, when viewing the objects by commit ref, Delete button should not be displayed.
![image](https://user-images.githubusercontent.com/9861474/111639533-03cdd580-8804-11eb-9e5f-92f8cff03390.png)

PR #1677, opened 2021-03-25
Fix/disable delete #1622

shamikakumar:
Implemented #1622


johnnyaug (changes requested):
Hey @shamikakumar, thank you for working on this issue!
I've tested it and it works great. I would like to suggest a solution with less code duplication: since the only component to be hidden is the `Dropdown.Item`, could you apply the condition only to it, without copying the `Dropdown.Menu`?

shamikakumar:
sure, let me fix that

johnnyaug (approved):
Thank you very much @shamikakumar!

PR #1679, opened 2021-03-25
Avoid logging v2 sigs on failure

arielshaqed:
Auth headers are inherently unsafe: if Eve can read (just) headers and can guess message
contents, then she can fix a broken message and replay it slightly later.

Easiest just to avoid logging bad headers.  If there is a problem record the actual HTTP
requests and responses on the client and send them out of band.


arielshaqed:
Found by LGTM.com.

ozkatz (approved):

PR #1682, opened 2021-03-25
Limit length of graveler serialization

arielshaqed:
Found by LGTM: a user with access to lakeFS and its backing object storage could...

1. Forge a range or metarange graveler file with too-long (multi GiB) values;
2. Convince a lakeFS server to read that range or metarange server;
3. Commit or merge, causing that lakeFS server to overflow when computing the desired length
   of a serialization buffer;
4. Crash that lakeFS server goroutine, possibly denying service to that branch;
5. Sorry, no clear path ends in "???; PROFIT!".

Easiest to fix it so we don't need to explain ourselves to new users and/or know we can safely
ignore this on LGTM.com.


ozkatz (approved):
Thanks!

arielshaqed:
Thanks!

PR #1685, opened 2021-03-28
proposal doc for staging with a queue

arielshaqed:
Very preliminary "let's do Raft (or Kinesis)" design doc.  Happy to hear opinions.

It is actually quite similar to Oz' "S3 + kv store" proposal: batching to improve performance, staging tokens, a degree of optimism when writing.

Assuming no gaping holes, the next stage is to analyze expected performance, cost of queue (if Kinesis), and differences.


nopcoder (approved):

arielshaqed:
This might become the basis for an in-memory Raft kv store, but otherwise it is entirely subsumed by #2466 combined with #3091 .

Closing -- thanks @ozkatz for considering this for #2466!

PR #1688, opened 2021-03-29
proposal doc for object store based LSM

ozkatz:


ozkatz:
Documenting discovery tasks that we decided to follow with:

- [ ] Database compatibility with `RefStore` requirements (on Azure, GCS, AWS, Raft: ZK, Consul, ETCd, custom impl)?
- [ ] Database performance tests for list appends
- [ ] Find ways to model compactions happening concurrently with commits
- [ ] Understand the write profile using existing S3 inventories we have access to
- [ ] Understand pricing model for e.g. DynamoDB and comparable database-as-a-service offerings
- [ ] Find correct hueristic for flushing to L1 (Size? Time? Something else? combination?)

nopcoder (approved):

ozkatz:
Closing this proposal in favor of #2466 which I believe is better.

PR #1690, opened 2021-03-29
Better error handling for file not find exception during export

### LINKED ISSUE #1687: [Export] Don't treat FileNotFoundException as error

itaiad200:
closes #1687

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/ExportStatus.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/ExportStatus.scala
@@ -1,0 +1,4 @@
+package io.treeverse.clients
+
+// ExportStatus is the export status of a single file
+case class ExportStatus(path: String, success: Boolean, msg: String) extends Serializable {}
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/ExportStatus.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/ExportStatus.scala
@@ -1,4 +1,9 @@
 package io.treeverse.clients
 
-// ExportStatus is the export status of a single file
+/** Export status of a single file
+ *
+ * @param path of the file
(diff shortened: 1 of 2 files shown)


clients/spark/core/src/main/scala/io/treeverse/clients/ExportStatus.scala
```diff
@@ -0,0 +1,4 @@
+package io.treeverse.clients
+
+// ExportStatus is the export status of a single file
+case class ExportStatus(path: String, success: Boolean, msg: String) extends Serializable {}
```
arielshaqed, line 4: Ewww... That's a Go comment... 
Consider /** ... */. And it needs the imperative mode, without repeating the function name. 
🤷

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
```diff
@@ -133,31 +134,32 @@ object Exporter {
...
+            srcPath.getFileSystem(conf),
+            srcPath,
+            dstFS,
+            dstPath,
+            false,
+            conf)
+          null
+        } catch  {
+          case e : (FileNotFoundException) =>  ExportStatus(dstPath.toString, success = true, s"Unable to copy file ${dstPath.toString} from source ${srcPath.toString} since source file is missing: ${e.toString}")
+          case e : (IOException) =>  ExportStatus(dstPath.toString, success = false, s"Unable to copy file ${dstPath.toString} from source ${srcPath.toString}: ${e.toString}")
```
arielshaqed, line 159: Do you really need all those toString calls when interpolating strings? (If so, possibly a reason to use % formatting)

arielshaqed (approved):
Cool!

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala

PR #1692, opened 2021-03-30
jars: publish to s3

johnnyaug:


clients/spark/build.sbt
```diff
@@ -89,9 +89,15 @@ lazy val commonSettings = Seq(
...
 lazy val publishSettings = Seq(
+  publishResolvers := Seq(
+    if (isSnapshot.value)
+      MavenRepository("lakefs-snapshots", s"$nexus/content/repositories/snapshots")
+    else MavenRepository("lakefs-releases", s"$nexus/service/local/staging/deploy/maven2"),
+    MavenRepository("lakefs-s3-repo", "s3://treeverse-clients-us-east/"),
+  ),
   publishTo := {
-    val nexus = "https://s01.oss.sonatype.org/"
     if (isSnapshot.value) Some("snapshots" at nexus + "content/repositories/snapshots")
```
arielshaqed, line 101: It is unfortunate that we need to say "snapshots" here and in the resolver settings.

clients/spark/project/plugins.sbt
```diff
@@ -1,2 +1,4 @@
 addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.15.0")
 addSbtPlugin("com.github.sbt" % "sbt-pgp" % "2.1.2")
+addSbtPlugin("laughedelic" % "sbt-publish-more" % "0.1.0")
```
arielshaqed, line 3: Do we still need this?

arielshaqed (commented):
Cool, thanks!

clients/spark/build.sbt
johnnyaug: Yes, although the `sbt-publish-more` should set the publishTo by default, which would remove the need to explicitly specify it - it doesn't work well with subprojects.

clients/spark/project/plugins.sbt
johnnyaug: Yes - this is what allows us to publish to multiple repos (sonatype + s3)

arielshaqed (approved):
Cool, thanks!

PR #1696, opened 2021-03-31
Replace Java's URL with URI

### LINKED ISSUE #1695: Use URI instead of URL in exporter

itaiad200:
closes #1695

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
@@ -162,7 +162,7 @@ object Exporter {
     }
   }
 
-  private def resolveURL(uri :URI, extraPath: String): Path = {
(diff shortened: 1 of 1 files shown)


clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
```diff
@@ -162,10 +162,8 @@ object Exporter {
     }
   }
 
-  private def resolveURL(baseUrl :URL, extraPath: String): Path = {
-    val uri = baseUrl.toURI
+  private def resolveURL(uri :URI, extraPath: String): Path = {
```
arielshaqed, line 165:
```suggestion
  private def resolveURL(uri: URI, extraPath: String): Path = {
```

arielshaqed (approved):
At last!  Empirical proof that a URI is not the same as a URL!

PR #1705, opened 2021-04-01
Refactor client response-to-error to API helper and add Nessie tests for direct client upload

arielshaqed:
It should be The One True Way To Do It.


pkg/api/helpers/upload.go
```diff
@@ -0,0 +1,108 @@
...
+)
+
+// ClientUpload uploads contents as a file using client-side ("direct") access to underlying
+// storage.  It requires credentials both to lakeFS and to underlying storage, but
+// considerably reduces the load on the lakeFS server.
+func ClientUpload(ctx context.Context, client api.ClientWithResponsesInterface, repoID, branchID, filePath string, metadata map[string]string, contents io.ReadSeeker) (*api.ObjectStats, error) {
+	resp, err := client.GetPhysicalAddressWithResponse(ctx, repoID, branchID, &api.GetPhysicalAddressParams{
+		Path: filePath,
+	})
+	if err != nil {
```
itaiad200, line 25: Not part of this PR, but is this the recommended way to check for errors from the client?
3 Different checks just to make sure nothing failed..

pkg/api/helpers/upload.go
```diff
@@ -0,0 +1,108 @@
...
+				// BUG(ariels): Unavailable on S3, remove this field entirely
+				//     OR add it to the server staging manager API.
+				Mtime:           time.Now().Unix(),
+				Path:            filePath,
+				PathType:        "object",
+				PhysicalAddress: physicalAddress,
+				SizeBytes:       &size,
+			}, nil
+		}
+		if resp.JSON409 == nil {
```
itaiad200, line 102: This is confusing, but maybe it's just me.
First if checks for the status code, second if checks for the value that comes with the specific status code, without checking the status code first?

nessie/system_test.go
```diff
@@ -75,22 +89,35 @@ func createRepository(ctx context.Context, t *testing.T, name string, repoStorag
...
-		return "", "", err
-	}
-	if resp.StatusCode() != http.StatusCreated {
-		return "", "", fmt.Errorf("%w: %s (%d)", errUploadFailed, resp.Status(), resp.StatusCode())
+	if direct {
+		stats, err := helpers.ClientUpload(ctx, client, repo, branch, objPath, nil, strings.NewReader(objContent))
+		if errors.Is(err, helpers.ErrUnsupportedProtocol) {
+			t.Logf("direct upload not supported yet (%s), switch to use gateway", err)
+			directUploadSupported = false
+			return uploadFileRandomDataAndReport(ctx, t, repo, branch, objPath, false)
```
itaiad200, line 105: This is an indirect way to understand if the current settings support direct upload or not.
Since we're trying to test the actual system, it might ignore real errors. For instance, s3 returns `ErrUnsupportedProtocol`. 

Any chance we can pass it as a test argument and expects all actions to succeed (or fail) according to the prior expectation, and not one inferred from the first call?

nessie/system_test.go
```diff
@@ -110,8 +137,8 @@ func uploadContent(ctx context.Context, repo string, branch string, objPath stri
 	}, w.FormDataContentType(), &b)
 }
 
-func uploadFileRandomData(ctx context.Context, t *testing.T, repo, branch, objPath string) (checksum, content string) {
-	checksum, content, err := uploadFileRandomDataAndReport(ctx, repo, branch, objPath)
+func uploadFileRandomData(ctx context.Context, t *testing.T, repo, branch, objPath string, direct bool) (checksum, content string) {
+	checksum, content, err := uploadFileRandomDataAndReport(ctx, t, repo, branch, objPath, direct && directUploadSupported)
```
itaiad200, line 141: Maybe check the value of `directUploadSupported` in the inner method instead of maybe missing it occasionally from other callers of `uploadFileRandomDataAndReport`.

itaiad200 (changes requested):
Really nice work with enabling client upload and testing it.

A little bit worried about the reflection part and deciding when to use the direct upload from Nessie.

nessie/system_test.go
arielshaqed: This no longer exists, and `direct` is piped right through.

pkg/api/helpers/upload.go
arielshaqed: No idea.  @nopcoder WDYT?

pkg/api/helpers/upload.go
arielshaqed: Yes, AFAIU.  The code gen creates a response type with a separate field for each possible status code.  That field will be `nil` if the status code is different.  The converse need not be true, so above we _do_ test the status code because we ignore all response fields.
@nopcoder to confirm.

pkg/api/helpers/errors.go
```diff
@@ -0,0 +1,66 @@
...
+// ResponseAsError returns a ErrRequestFailed wrapping a response from the server.  It
+// searches for a non-nil unsuccessful HTTPResponse field and uses its message, along with a
+// Body that it assumes is an api.Error.
+func ResponseAsError(response interface{}) error {
+	r := reflect.Indirect(reflect.ValueOf(response))
+	if !r.IsValid() {
+		return fmt.Errorf("%w: bad type: cannot indirect", ErrRequestFailed)
+	}
+
+	f := r.FieldByName("HTTPResponse")
```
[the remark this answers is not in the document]
arielshaqed: Done, sort-of.  The docs mean that _`r`_ has to have a struct type.  It does, so that's OK.

Modified anyway to test for the kind of `r` -- it will be an internal (logic) error reported as an error, but probably less scary than a `panic`.

pkg/api/helpers/errors.go
```diff
@@ -0,0 +1,66 @@
...
+	f = r.FieldByName("Body")
+	if f.IsValid() && f.Type().Kind() == reflect.Slice && f.Type().Elem().Kind() == reflect.Uint8 {
+		body := f.Bytes()
+		var apiError api.Error
+		if json.Unmarshal(body, &apiError) == nil && apiError.Message != "" {
+			message = apiError.Message
+		}
+	}
+
+	return fmt.Errorf("%w: %s", ErrRequestFailed, message)
```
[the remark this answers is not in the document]
arielshaqed: Done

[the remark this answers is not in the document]
arielshaqed: No.  I want to get "request failed: invalid storage type s3: current block adapter is gcp".  With the above change I get "invalid storage type s3: current block adapter is gcp: request failed" which is not as nice.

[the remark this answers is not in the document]
arielshaqed: Including it.

pkg/api/helpers/errors_test.go
```diff
@@ -0,0 +1,65 @@
...
+}
+
+func TestResponseAsError(t *testing.T) {
+	cases := []struct {
+		name     string
+		response interface{}
+		message  string // non-empty to match return error; empty to signal no error
+	}{
+		{"no HTTPResponse field", &struct{ A int }{17}, "no HTTPResponse"},
+		{"OK", &Response{&http.Response{StatusCode: 234}}, ""},
```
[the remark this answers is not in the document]
arielshaqed: Discussed f2f; `reflect.Value.Field` has unexpected behaviour... and this is OK.

arielshaqed (commented):
Thanks!

PTAL...

itaiad200 (approved):

arielshaqed:

PR #1708, opened 2021-04-04
Handle no path for delete objects in gateway

### LINKED ISSUE #1706: lakeFS incompatibility issues with Spark (2.4.7) DeleteObjects request
time="2021-04-02T12:54:22Z" level=debug msg="object set for deletion" func="pkg/gateway/operations.(*DeleteObjects).Handle" file="pkg/gateway/operations/deleteobjects.go:75" h
ost="medium.s3.itai-local-test.lakefs.dev:8000" key=main/dummy-data-32/ method=POST path="/?delete" repository=medium request_id=fcc3cd03-c0b6-487b-83f6-78b136b65b44 service_n
ame=s3_gateway user=Itai

time="2021-04-02T12:54:22Z" level=error msg="failed deleting object" func="pkg/gateway/operations.(*DeleteObjects).Handle" file="pkg/gateway/operations/deleteobjects.go:67" error="argument path: required value" host="medium.s3.itai-local-test.lakefs.dev:8000" key=main/ method=POST path="/?delete" repository=medium request_id=fcc3cd03-c0b6-487b-83f6-78b136b65b44 service_name=s3_gateway user=Itai

itaiad200:
closes #1706

### DIFF
--- a/pkg/gateway/operations/deleteobjects.go
+++ b/pkg/gateway/operations/deleteobjects.go
@@ -63,7 +63,7 @@ func (controller *DeleteObjects) Handle(w http.ResponseWriter, req *http.Request
 		switch {
 		case errors.Is(err, catalog.ErrNotFound):
 			lg.Debug("tried to delete a non-existent object")
-		case errors.Is(err, catalog.ErrRequiredValue) && resolvedPath.Path == "":
+		case errors.Is(err, catalog.ErrPathRequiredValue):
 			// issue #1706 - https://github.com/treeverse/lakeFS/issues/1706
 			// Spark trying to delete the path "main/", which we map to branch "main" with an empty path.
 			// Spark expects it to succeed (not deleting anything is a success), instead of returning an error.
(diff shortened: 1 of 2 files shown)


pkg/gateway/operations/deleteobjects.go
```diff
@@ -63,6 +63,11 @@ func (controller *DeleteObjects) Handle(w http.ResponseWriter, req *http.Request
 		switch {
 		case errors.Is(err, catalog.ErrNotFound):
 			lg.Debug("tried to delete a non-existent object")
+		case errors.Is(err, catalog.ErrRequiredValue) && resolvedPath.Path == "":
```
johnnyaug, line 66: This is risky: `DeleteEntry` will also return `ErrRequiredValue` if the repository or branch names are empty, and here you are possibly suppressing this error. Consider defining a new error type for an empty path.

johnnyaug (changes requested):

pkg/gateway/operations/deleteobjects.go
nopcoder: Alt, you can just check if the `path` we pass we parsed is empty and return success after successful get repo/branch.

itaiad200: Done.

johnnyaug (approved):

pkg/gateway/operations/deleteobjects.go
itaiad200: @nopcoder you mean get repo/branch here or in the cataloger?

If it's here, then it's a wasteful repeatable operation I would like to avoid.
If in the underlying cataloger then I'll need to rely or even modify behaviour for the gateway, which I think is suboptimal.

PR #1710, opened 2021-04-04
fix: API merge message is optional

nopcoder:
Fix backward compatibility for merge request

### DIFF
--- a/docs/assets/js/swagger.yml
+++ b/docs/assets/js/swagger.yml
@@ -367,8 +367,6 @@ components:
 
     Merge:
       type: object
-      required:
-        - message
       properties:
         message:
           type: string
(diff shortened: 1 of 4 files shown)


docs/assets/js/swagger.yml
```diff
@@ -1777,6 +1775,30 @@ paths:
         required: true
         schema:
           type: string
+    get:
```
arielshaqed, line 1778: This seems to be off by one, it incorporates a _previous_ change into this swaggacopy but not _this_ change.  Can you regenerate it before pulling?

arielshaqed (approved):
Cool, but please keep the js copy of swagger.yml aligned with this PR.

THANKS!

docs/assets/js/swagger.yml
nopcoder: sure - I was sure the makefile validates they are the same.

nopcoder:
```shell
% sha256sum api/swagger.yml docs/assets/js/swagger.yml
9674aad548414c911e7c7d21f22bae22632e9c6e9f7784875549eb74917c34f4  api/swagger.yml
9674aad548414c911e7c7d21f22bae22632e9c6e9f7784875549eb74917c34f4  docs/assets/js/swagger.yml
```

PR #1712, opened 2021-04-04
[bug] Link correct physical address in client direct upload

arielshaqed:
2nd phase `LinkPhysicalAddress` was incorrectly repeating the lakeFS path.  Use the requested
upload location instead.

Probably broken during the OpenAPI 2 -> OpenAPI3 change.  "Why isn't this tested?", you may
well ask.  Quite right: in fact, this was breaking the current effort to add a test for direct
upload.

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -167,7 +167,7 @@ func (c *Controller) LinkPhysicalAddress(w http.ResponseWriter, r *http.Request,
 	entry := catalog.DBEntry{
 		CommonLevel:     false,
 		Path:            params.Path,
-		PhysicalAddress: params.Path,
+		PhysicalAddress: *body.Staging.PhysicalAddress,
 		CreationDate:    writeTime,
 		Size:            body.SizeBytes,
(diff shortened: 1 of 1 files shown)


pkg/api/controller.go
```diff
@@ -167,7 +167,7 @@ func (c *Controller) LinkPhysicalAddress(w http.ResponseWriter, r *http.Request,
 	entry := catalog.DBEntry{
 		CommonLevel:     false,
 		Path:            params.Path,
-		PhysicalAddress: params.Path,
+		PhysicalAddress: *body.Staging.PhysicalAddress,
```
nopcoder, line 170: my bad :( - thanks!

nopcoder (approved):

pkg/api/controller.go
arielshaqed: A wise singer once told me,

If you liked it then you should have put a test on it.  (Doing that now...)

arielshaqed (commented):
Thanks for rapid review!

Pulling once tests pass.

PR #1713, opened 2021-04-04
readLines make sure we close the reader on scanner error

nopcoder:


cmd/lakectl/cmd/abuse.go
```diff
@@ -23,21 +23,21 @@ var abuseCmd = &cobra.Command{
 	Hidden: true,
 }
 
-func readLines(path string) ([]string, error) {
+func readLines(path string) (lines []string, err error) {
 	reader := OpenByPath(path)
+	defer func() {
+		err = reader.Close()
```
arielshaqed, line 29: This is subtle and I believe when I discussed it with @itaiad200 he strongly asked that we  *not* do this sort of thing.  (You can put the body of the function after opening `reader` in a sub-function, get an error from there, and then close inline -- if you want to keep some elegance).

Also, while elegant, it can hide a *first* error (e.g. flagged by `scanner.Err()`) by returning only the *second* error (if `reader.Close()` _also_ fails).  I definitely prefer returning the first or ideally both:
```suggestion
		if closeErr = reader.Close(); closeErr != nil {
			if err == nil {
				err = closeErr
			} else {
				err = fmt.Errorf("%w, and while closing %s", err, closeErr)
			}				
```

arielshaqed (approved):
Thanks, but see comment about assigning to named return values in a `defer` -- not popular here AFAIR, and also this implementation can mask an important error behind a less important one.

cmd/lakectl/cmd/abuse.go
nopcoder: Agree, I'll align the above to use this practice.

nopcoder: Updated the code - had to keep the named return values, without it the err value was back to nil after the return.
If it is too much I can just add DieErr as OpenByPath and the return function when using this helper does it.
what do you think?

nopcoder:
> Thanks, but see comment about assigning to named return values in a `defer` -- not popular here AFAIR, and also this implementation can mask an important error behind a less important one.

Let me know if my answer and change are not enough. I have tested the option to use non-named return values - cloud not change the `err` value after the func return (as expected)
https://play.golang.org/p/SCDHkHXPybU

PR #1716, opened 2021-04-05
Support direct download using statObject

arielshaqed:


cmd/lakectl/cmd/fs.go
```diff
@@ -121,11 +122,26 @@ var fsCatCmd = &cobra.Command{
...
+			if err != nil {
+				DieErr(err)
+			}
+		} else {
+			resp, err := client.GetObjectWithResponse(cmd.Context(), pathURI.Repository, pathURI.Ref, &api.GetObjectParams{
+				Path: *pathURI.Path,
+			})
+			DieOnResponseError(resp, err)
+			contents = resp.Body
+		}
```
itaiad200, line 143: I think a helper that chooses the right method to download is appropriate here as this will probably be repeated elsewhere.

pkg/api/helpers/download.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	physicalAddress:= resp.JSON200.PhysicalAddress
+	parsedAddress, err := url.Parse(physicalAddress)
+	if err != nil {
+		return nil, nil, fmt.Errorf("parse physical address URL %s: %w", physicalAddress, err)
+	}
+
+	// TODO(ariels): plug-in support for other protocols
+	if parsedAddress.Scheme != "s3" {
```
itaiad200, line 36: I've seen this elsewhere too, I think the supported scheme knowledge shouldn't spread in more than 1 location.

pkg/api/helpers/download.go
```diff
@@ -0,0 +1,60 @@
...
+	"context"
+	"fmt"
+	"io"
+	"net/url"
+)
+
+// ClientDownload downloads a file using client-side ("direct") access to underlying storage.
+// It requires credentials both to lakeFS and to underlying storage, but considerably reduces
+// the load on the lakeFS server.
+func ClientDownload(ctx context.Context, client api.ClientWithResponsesInterface, repoID, ref, filePath string) (*api.ObjectStats, io.ReadCloser, error) {
```
itaiad200, line 18: Suggestion: accept a block adapter and avoid a specific s3 implementation.

cmd/lakectl/cmd/fs.go
arielshaqed: Yes, but not yet: don't want to break any workflows (yet).  Or (even better) all adapters will support it.  We will still need indirect uploads (when no credentials exist), but it should be less common.
The thing is, right now it takes at least one extra RPC to each of lakeFS and S3 to determine this automatically.
Opened #1726 to track.

pkg/api/helpers/download.go
arielshaqed: Tempting but currently not really possible.  Adapters are server-side, I probably don't want to bring them into each client.

pkg/api/helpers/download.go
arielshaqed: My plan:
1. Pull this.
2. Write additional support for other backing stores.
3. As part of :arrow_up: :arrow_up: :arrow_up: :arrow_up:  write some factory with `get{Up,Down}Loader` methods.
4. Automate on _that_, plus an explicit additional "auth failed" error (to indicate an indirect retry may work).

Again, this is too early to write the factory I think.

arielshaqed (commented):
Great requests, but postponing all of them till after this PR.  See inline and PTAL.

arielshaqed:
Wrote the factory anyway :-)

pkg/api/helpers/adapters.go
```diff
@@ -0,0 +1,98 @@
...
+		SharedConfigState: session.SharedConfigEnable,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("connect to S3 session: %w", err)
+	}
+	sess.ClientConfig(s3.ServiceName)
+	return &s3Adapter{svc: s3.New(sess)}, nil
+}
+
+func (s *s3Adapter) Upload(ctx context.Context, physicalAddress *url.URL, contents io.ReadSeeker) (ObjectStats, error) {
```
itaiad200, line 67: Validate matching scheme

pkg/api/helpers/adapters.go
```diff
@@ -0,0 +1,98 @@
...
+		return ObjectStats{}, fmt.Errorf("read stream size: %w", err)
+	}
+	return ObjectStats{
+		Size: size,
+		ETag: api.StringValue(putObjectResponse.ETag),
+		// S3 PutObject does not return creation time.
+	}, nil
+}
+
+func (s *s3Adapter) Download(ctx context.Context, physicalAddress *url.URL) (io.ReadCloser, error) {
```
itaiad200, line 88: Validate matching scheme

pkg/api/helpers/upload.go
```diff
@@ -89,20 +62,23 @@ func ClientUpload(ctx context.Context, client api.ClientWithResponsesInterface,
...
 				PathType:        "object",
 				PhysicalAddress: physicalAddress,
-				SizeBytes:       &size,
+				SizeBytes:       &stats.Size,
 			}, nil
 		}
 		if resp.JSON409 == nil {
 			return nil, fmt.Errorf("link object to backing store: %w (status code %d)", ErrRequestFailed, resp.StatusCode())
 		}
 		// Try again!
```
itaiad200, line 78: Forever? consider limiting the tries.

itaiad200 (approved):
Looks great! see minor suggestions

arielshaqed (commented):
Thanks!

Pulling once it passes.

PR #1717, opened 2021-04-05
lakeFS endpoint as URI with support of base URI by flag and env var

### LINKED ISSUE #1487: Breaking: change URI scheme to be more standard, replacing '@' with '/'
so instead of doing `lakefs://repo@branch/path`, we'll do a more standard `lakefs://repo/branch/path`.

### LINKED ISSUE #1488: CLI: Support a standard `LAKECTL_BASE_URI` environment variable and `--base-uri` flag
to set a prefix for all other relative URIs in the command

### LINKED ISSUE #1489: CLI: Show fully qualified URIs for all command parameters on output

nopcoder:
Close #1489
Close #1488
Close #1487

1. Use '/' as a separator between repository and branch
2. Parse endpoint as URI
3. In case the endpoint doesn't start with lakefs:// and LAKECTL_BASE_URI or flag --base-uri is set. Use the base URI as a prefix for the parsed URI

### DIFF
--- a/cmd/lakectl/cmd/refs.go
+++ b/cmd/lakectl/cmd/refs.go
@@ -32,7 +32,7 @@ Since a bare repo is expected, in case of transient failure, delete the reposito
 		cmdutils.FuncValidator(0, uri.ValidateRepoURI),
 	),
 	Run: func(cmd *cobra.Command, args []string) {
-		repoURI := uri.Must(uri.Parse(args[0]))
+		repoURI := MustParseURI(args[0])
 		manifestFileName, _ := cmd.Flags().GetString("manifest")
 		fp := OpenByPath(manifestFileName)
 		defer func() {
@@ -66,7 +66,7 @@ var refsDumpCmd = &cobra.Command{
 		cmdutils.FuncValidator(0, uri.ValidateRepoURI),
 	),
 	Run: func(cmd *cobra.Command, args []string) {
-		repoURI := uri.Must(uri.Parse(args[0]))
+		repoURI := MustParseURI(args[0])
 
 		client := getClient()
 		resp, err := client.DumpRefsWithResponse(cmd.Context(), repoURI.Repository)
--- a/cmd/lakectl/cmd/refs.go
+++ b/cmd/lakectl/cmd/refs.go
@@ -6,8 +6,6 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/treeverse/lakefs/pkg/api"
-	"github.com/treeverse/lakefs/pkg/cmdutils"
-	"github.com/treeverse/lakefs/pkg/uri"
 )
 
 var metadataDumpTemplate = `
@@ -27,12 +25,9 @@ This command is expected to run on a bare repository (i.e. one created with 'lak
 Since a bare repo is expected, in case of transient failure, delete the repository and recreate it as bare and retry.`,
 	Example: "aws s3 cp s3://bucket/_lakefs/refs_manifest.json - | lakectl refs-load lakefs://my-bare-repository --manifest -",
 	Hidden:  true,
-	Args: cmdutils.ValidationChain(
-		cobra.ExactArgs(1),
-		cmdutils.FuncValidator(0, uri.ValidateRepoURI),
-	),
+	Args:    cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		repoURI := MustParseURI(args[0])
+		repoURI := MustParseRepoURI("repository", args[0])
 		manifestFileName, _ := cmd.Flags().GetString("manifest")
 		fp := OpenByPath(manifestFileName)
 		defer func() {
@@ -61,13 +56,9 @@ var refsDumpCmd = &cobra.Command{
 	Use:    "refs-dump <repository uri>",
 	Short:  "dumps refs (branches, commits, tags) to the underlying object store",
 	Hidden: true,
-	Args: cmdutils.ValidationChain(
-		cobra.ExactArgs(1),
-		cmdutils.FuncValidator(0, uri.ValidateRepoURI),
-	),
+	Args:   cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		repoURI := MustParseURI(args[0])
-
+		repoURI := MustParseRepoURI("repository", args[0])
 		client := getClient()
 		resp, err := client.DumpRefsWithResponse(cmd.Context(), repoURI.Repository)
 		DieOnResponseError(resp, err)
--- a/cmd/lakectl/cmd/refs.go
+++ b/cmd/lakectl/cmd/refs.go
@@ -28,6 +28,7 @@ Since a bare repo is expected, in case of transient failure, delete the reposito
 	Args:    cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		repoURI := MustParseRepoURI("repository", args[0])
+		Fmt("Repository: %s\n", repoURI.String())
 		manifestFileName, _ := cmd.Flags().GetString("manifest")
 		fp := OpenByPath(manifestFileName)
 		defer func() {
@@ -59,6 +60,7 @@ var refsDumpCmd = &cobra.Command{
 	Args:   cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		repoURI := MustParseRepoURI("repository", args[0])
+		Fmt("Repository: %s\n", repoURI.String())
 		client := getClient()
 		resp, err := client.DumpRefsWithResponse(cmd.Context(), repoURI.Repository)
 		DieOnResponseError(resp, err)
--- a/cmd/lakectl/cmd/root.go
+++ b/cmd/lakectl/cmd/root.go
@@ -124,8 +124,7 @@ func init() {
 	cobra.OnInitialize(initConfig)
 	rootCmd.PersistentFlags().StringVarP(&cfgFile, "config", "c", "", "config file (default is $HOME/.lakectl.yaml)")
 	rootCmd.PersistentFlags().BoolVar(&noColorRequested, "no-color", false, "don't use fancy output colors (default when not attached to an interactive terminal)")
-	baseURIDefault := os.Getenv("LAKECTL_BASE_URI")
-	rootCmd.PersistentFlags().StringVarP(&baseURI, "base-uri", "", baseURIDefault, "base URI used for lakeFS address parse")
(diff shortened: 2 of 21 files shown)


pkg/uri/parser.go
```diff
@@ -51,118 +46,87 @@ func (u *URI) IsFullyQualified() bool {
...
-			path = buf.String()
-			uri.Path = &path
+		const refAndPathParts = 2
+		levels := strings.SplitN(u.Path[1:], "/", refAndPathParts)
+		if len(levels) == refAndPathParts {
+			ref = levels[0]
+			path = &levels[1]
+		} else if len(levels) == 1 {
+			ref = levels[0]
 		}
```
itaiad200, line 108: How is an empty path different from a nil path?
What's the difference between:
`lakefs://example-repo/main/` and `lakefs://example-repo/main`?

pkg/uri/parser.go
```diff
@@ -51,118 +46,87 @@ func (u *URI) IsFullyQualified() bool {
...
-		case stateInRef:
-			uri.Ref = buf.String()
-		case stateInPath:
-			path = buf.String()
-			uri.Path = &path
+		const refAndPathParts = 2
+		levels := strings.SplitN(u.Path[1:], "/", refAndPathParts)
+		if len(levels) == refAndPathParts {
+			ref = levels[0]
+			path = &levels[1]
```
itaiad200, line 105: I don't understand why path is nullable and ref not.
IIUC they both could be missing when referencing a repo.

pkg/uri/parser.go
```diff
@@ -51,118 +46,87 @@ func (u *URI) IsFullyQualified() bool {
...
-			uri.Path = &path
-			buf.Reset()
-		default:
-			buf.WriteRune(ch)
+	var ref string
+	var path *string
+	if len(u.Path) > 0 {
+		if !strings.HasPrefix(u.Path, "/") {
+			return nil, ErrMalformedURI
 		}
```
itaiad200, line 100: Is this even possible?
Skip this check if it isn't.

pkg/uri/parser.go
```diff
@@ -51,118 +46,87 @@ func (u *URI) IsFullyQualified() bool {
...
-			uri.Ref = buf.String()
-			state = stateInPath
-			uri.Path = &path
-			buf.Reset()
-		default:
-			buf.WriteRune(ch)
+	var ref string
+	var path *string
+	if len(u.Path) > 0 {
+		if !strings.HasPrefix(u.Path, "/") {
```
itaiad200, line 98: Use `PathSeparator`.

pkg/uri/parser.go
```diff
@@ -51,118 +46,87 @@ func (u *URI) IsFullyQualified() bool {
...
-		switch state {
-		case stateInRepo:
-			uri.Repository = buf.String()
-		case stateInRef:
-			uri.Ref = buf.String()
-		case stateInPath:
-			path = buf.String()
-			uri.Path = &path
+		const refAndPathParts = 2
+		levels := strings.SplitN(u.Path[1:], "/", refAndPathParts)
```
itaiad200, line 102: Use PathSeparator.

pkg/uri/parser_test.go
```diff
@@ -18,61 +18,60 @@ func TestParse(t *testing.T) {
 		Err      error
 		Expected *uri.URI
 	}{
-		{"lakefs://foo@bar/baz", nil, &uri.URI{
```
itaiad200, line 21: Add one for referencing just the repo.

pkg/uri/parser_test.go
```diff
@@ -18,61 +18,60 @@ func TestParse(t *testing.T) {
...
 			Repository: "foo",
 			Ref:        "bar",
 		}},
-		{"lakefssss://foo@bar/baz", uri.ErrMalformedURI, nil},
-		{"lakefs:/foo@bar/baz", uri.ErrMalformedURI, nil},
-		{"lakefs//foo@bar/baz", uri.ErrMalformedURI, nil},
+		{"lakefssss://foo/bar/baz", uri.ErrMalformedURI, nil},
+		{"lakefs:/foo/bar/baz", uri.ErrMalformedURI, nil},
+		{"lakefs//foo/bar/baz", uri.ErrMalformedURI, nil},
 	}
```
itaiad200, line 63: I would leave some of the `lakefs://foo@bar/goo` tests as negative ones.

cmd/lakectl/cmd/root.go
```diff
@@ -148,6 +124,8 @@ func init() {
 	cobra.OnInitialize(initConfig)
 	rootCmd.PersistentFlags().StringVarP(&cfgFile, "config", "c", "", "config file (default is $HOME/.lakectl.yaml)")
 	rootCmd.PersistentFlags().BoolVar(&noColorRequested, "no-color", false, "don't use fancy output colors (default when not attached to an interactive terminal)")
+	baseURIDefault := os.Getenv("LAKECTL_BASE_URI")
```
itaiad200, line 127: That's not the default.

cmd/lakefs-loadtest/cmd/entry.go
```diff
@@ -26,12 +25,13 @@ const createEntryPathLength = 110
...
-		cobra.ExactArgs(1),
-		cmdutils.FuncValidator(0, uri.ValidateRefURI),
-	),
+	Args:  cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
 		u := uri.Must(uri.Parse(args[0]))
+		if !u.IsRef() {
+			fmt.Printf("Invalid 'ref': %s", uri.ErrInvalidRefURI)
+			os.Exit(1)
+		}
```
itaiad200, line 34: `MustRef...`

cmd/lakectl/cmd/root.go
```diff
@@ -148,6 +124,8 @@ func init() {
...
+	rootCmd.PersistentFlags().StringVarP(&baseURI, "base-uri", "", baseURIDefault, "base URI used for lakeFS address parse")
```
itaiad200, line 128: Can `base-uri` include anything? Just the repo? repo+ref? path too?

cmd/lakectl/cmd/refs.go
```diff
@@ -27,12 +25,10 @@ This command is expected to run on a bare repository (i.e. one created with 'lak
...
 	Example: "aws s3 cp s3://bucket/_lakefs/refs_manifest.json - | lakectl refs-load lakefs://my-bare-repository --manifest -",
 	Hidden:  true,
-	Args: cmdutils.ValidationChain(
-		cobra.ExactArgs(1),
-		cmdutils.FuncValidator(0, uri.ValidateRepoURI),
-	),
+	Args:    cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		repoURI := uri.Must(uri.Parse(args[0]))
+		repoURI := MustParseRepoURI("repository", args[0])
```
itaiad200, line 30: Why do you need to pass the "repository" arg? Isn't it clear from the func call?

itaiad200 (changes requested):
Great improvements, however I'm not clear on what the `base-uri` may include and our way of enforcing that..

nopcoder:
> Great improvements, however I'm not clear on what the `base-uri` may include and our way of enforcing that..

At first base-uri or repo was a valid endpoint, not it is no requirement or validation. We use it as a prefix in case you are not specifying lakefs:// protocol in our endpoint.
This makes the end result very expected, as we do not add any manipulation or "merge" between them - just concatenation.

pkg/uri/parser.go
```diff
@@ -51,118 +46,87 @@ func (u *URI) IsFullyQualified() bool {
...
+		if len(levels) == refAndPathParts {
+			ref = levels[0]
+			path = &levels[1]
+		} else if len(levels) == 1 {
+			ref = levels[0]
 		}
 	}
-	return &uri, nil
+	return &URI{
+		Protocol:   u.Scheme,
```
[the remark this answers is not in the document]
nopcoder: true, I guess you can create URI with your own Protocol - but I don't see the use case for us.

pkg/uri/parser.go
nopcoder: `lakefs://example-repo/main` is valid ref. (no path)
`lakefs://example-repo/main/` is valid path. ('/' path)

pkg/uri/parser.go
nopcoder: `lakefs://a`

pkg/uri/parser.go
nopcoder: There is a difference between no path and empty path

pkg/uri/parser_test.go
nopcoder: good idea - in our case foo@bar will be a repository name

cmd/lakefs-loadtest/cmd/entry.go
nopcoder: The MustXXX functions are based on the parse with default where we use the lakectl specific environment variable.
In this case, we do not use it - this is part of `lakefs-loadtest` command line.
If we will have multiple command-line tools that use the same technic, I'll extract this functions out to `uri` package and pass the base URI value on each call.

cmd/lakectl/cmd/root.go
nopcoder: yes, it can even just include the schema - `--base-uril lakefs://` which is cool

cmd/lakectl/cmd/refs.go
nopcoder: What thinking about it as it is possible to use for copy repository or other future commands where you parse more than one repository name.
Also aligned with the other parse functions.

nopcoder:
Rebased and apply code changes for the above comments

itaiad200:
> > Great improvements, however I'm not clear on what the `base-uri` may include and our way of enforcing that..
> 
> At first base-uri or repo was a valid endpoint, not it is no requirement or validation. We use it as a prefix in case you are not specifying lakefs:// protocol in our endpoint.
> This makes the end result very expected, as we do not add any manipulation or "merge" between them - just concatenation.

I see. Please document it in the code as well.

itaiad200 (approved):

nopcoder:
> > > Great improvements, however I'm not clear on what the `base-uri` may include and our way of enforcing that..
> > 
> > 
> > At first base-uri or repo was a valid endpoint, not it is no requirement or validation. We use it as a prefix in case you are not specifying lakefs:// protocol in our endpoint.
> > This makes the end result very expected, as we do not add any manipulation or "merge" between them - just concatenation.
> 
> I see. Please document it in the code as well.

Documented `baseURI` in the code

PR #1720, opened 2021-04-05
Support multiple gateway domain names

### LINKED ISSUE #1719: Support multiple S3 gateway domain names
[Requested [on Slack](https://lakefs.slack.com/archives/C016726JLJW/p1617649557038900)]

The use case is when there are _two_ ways to access lakeFS as an S3 gateway:
1. via external domain name  https://lakefs.xxx.xxx.xx  though load balancer .
2. direct access from inside k8s,  e.g. http://lakefs

arielshaqed:
Remaining issues:
1. Middleware EnrichWithOperation always takes first domain.  Can we take req.Host?  (If not, at
   make sure to allow *zero* domains)
2. sigV2 implementation uses the *configured* "bare domain" to sign.  This is sort-of quite a
   lot like using req.Host (but not entirely).

Fixes #1719.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"math"
 	"reflect"
+	"sort"
 	"strings"
 	"time"
 
@@ -66,6 +67,7 @@ var (
 	ErrBadConfiguration  = errors.New("bad configuration")
 	ErrMissingSecretKey  = fmt.Errorf("%w: auth.encrypt.secret_key cannot be empty", ErrBadConfiguration)
 	ErrInvalidProportion = fmt.Errorf("%w: total proportion isn't 1.0", ErrBadConfiguration)
+	ErrBadDomainNames = fmt.Errorf("%w: domain names are prefixes", ErrBadConfiguration)
 )
 
 type LogrusAWSAdapter struct {
@@ -94,6 +96,14 @@ func NewConfig() (*Config, error) {
 	setupLogger()
 
 	err := viper.UnmarshalExact(&c.values)
+	if err != nil {
+		return nil, err
+	}
+
+	err = c.validateDomainNames()
+	if err != nil {
+		return nil, err
+	}
 	return c, err
 }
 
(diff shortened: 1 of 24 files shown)


arielshaqed:
Tested host-style addressing *manually* using Postman:
1. Recorded a request.
2. Configured lakeFS to accept *both* domains.
2. Added DNS records for `puddle.lakefs.io`, `*.puddle.lakefs.io` to point at 127.0.0.1 (just like `local.lakefs.io` and `*.local.lakefs.io` do).
2. Set "AWS Signature" in Authorization with my AccessKey [ID...] and Secret [Access] Key, the region, and service name "s3".
3. Changed "Host" header to other domain name.
4. Changed DNS in GET (not really needed)
5. Removed bucket from GET URL.

It worked, proving host-based addressing works.

arielshaqed:
Opened #1769, #1770 for missing test automation.

pkg/config/config.go
```diff
@@ -133,8 +133,8 @@ const (
 
 	CommittedPebbleSSTableCacheSizeBytesKey = "committed.sstable.memory.cache_size_bytes"
 
-	GatewaysS3DomainNameKey = "gateways.s3.domain_name"
-	GatewaysS3RegionKey     = "gateways.s3.region"
+	GatewaysS3DomainNamesKey = "gateways.s3.domain_name"
```
guy-har, line 136:
```suggestion
	GatewaysS3DomainNamesKey = "gateways.s3.domain_names"
```

pkg/config/template.go
```diff
@@ -95,7 +121,7 @@ type configuration struct {
 	}
 	Gateways struct {
 		S3 struct {
-			DomainName  string `mapstructure:"domain_name"`
+			DomainNames Strings `mapstructure:"domain_name"`
```
guy-har, line 124:
```suggestion
			DomainNames Strings `mapstructure:"domain_names"`
```

pkg/config/template.go
```diff
@@ -1,6 +1,32 @@
...
 
-import "time"
+import (
+	"reflect"
+	"time"
+)
+
+// Strings is a []string that mapstructure can deserialize from a single string or from a list
+// of strings.
+type Strings []string
```
guy-har, line 10: Nice 😎

guy-har (changes requested):
Looks Great, 
some bugs might occur in case one bare domain is a prefix of another.
I think we should extract the relevant bare domain for each case, that should also help out with `sigV2` and `FQDN`

pkg/gateway/middleware.go
```diff
@@ -80,14 +80,21 @@ func EnrichWithParts(bareDomains []string, next http.Handler) http.Handler {
 	})
 }
 
+func getBareDomain(hostname string, bareDomains []string) string {
+	for _, bd := range bareDomains {
+		if strings.HasSuffix(hostname, "." + bd) {
```
guy-har, line 85: what about `bd == hostname`?

guy-har (approved):
LGTM
one small comment.

guy-har:
Adding here documentation related to sigV2 and FQDN
`https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html`
`https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html`

pkg/gateway/middleware.go
```diff
@@ -195,11 +208,12 @@ func Parts(host string, urlPath string, bareDomain string) (repo string, ref str
...
-		if !strings.HasSuffix(host, ourHost) {
-			repo = ""
-		} else {
-			repo = strings.TrimSuffix(host, "."+ourHost)
+		repo = ""
+		for _, ourHost := range ourHosts {
+			if strings.HasSuffix(host, ourHost) {
+				repo = strings.TrimSuffix(host, "."+ourHost)
+				break
+			}
```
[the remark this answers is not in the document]
guy-har: 👍

arielshaqed (commented):
Thanks; handled all things written down and/or discussed.  Changes should be nicely organized by commit, so easiest to review by individual commit.

arielshaqed:
Thanks!

Pulling once tests pass...

PR #1721, opened 2021-04-06
[Export] Parallelising within a partition

### LINKED ISSUE #1715: [Export] Improve runtime
Exporting an entire commit takes much longer than importing just the diff for the same amount of files.

itaiad200:
close #1715

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
@@ -6,12 +6,9 @@ import org.apache.spark.sql.{DataFrame, Row, SparkSession}
 import java.io.{FileNotFoundException, IOException, Serializable}
 import java.net.URI
 import java.time.format.DateTimeFormatter
-import java.time.ZonedDateTime
-import java.util.Calendar
-import java.util
 import java.util.concurrent.{Callable, ExecutorService, Executors}
 import scala.util.Random
-import collection.JavaConverters._
+import scala.collection.JavaConverters._
 
 
 class Exporter(spark : SparkSession, apiClient: ApiClient, filter: KeyFilter, repoName: String, dstRoot: String, parallelism: Int) {
@@ -45,17 +42,14 @@ class Exporter(spark : SparkSession, apiClient: ApiClient, filter: KeyFilter, re
     val hadoopConf = spark.sparkContext.hadoopConfiguration
     val serializedConf = new SerializableWritable(hadoopConf)
     val f = filter
-    var par = parallelism
+    val par = parallelism
     val errs = actionsDF.mapPartitions( part => {
       val pool: ExecutorService = Executors.newFixedThreadPool(par)
 
-      val callableTasks = new util.ArrayList[Callable[ExportStatus]]
-      while (part.hasNext) {
-        val row = part.next()
-        callableTasks.add(new Handler(f, round, ns, rel, serializedConf, row))
-      }
-
-      val res = pool.invokeAll(callableTasks).asScala.map(f => f.get()).iterator
+      val res = pool.invokeAll(
+        part.map(row =>
+          new Handler(f, round, ns, rel, serializedConf, row)
+        ).toList.asJava).asScala.map(fut => fut.get()).iterator
       pool.shutdown()
       res
     }
@@ -64,8 +58,8 @@ class Exporter(spark : SparkSession, apiClient: ApiClient, filter: KeyFilter, re
     if (errs.isEmpty){
       return true
     }
-
-    writeSummaryFile(false, commitID, errs.take(maxLoggedErrors).map(s => s.msg).reduce(_+"\n"+_))
+    val count = errs.count()
+    writeSummaryFile(false, commitID, errs.sample(math.min(1.0, maxLoggedErrors.toDouble/count)).map(s => s.msg).reduce(_+"\n"+_))
     false
   }
 
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
@@ -15,10 +15,8 @@ import collection.JavaConverters._
 
 
 class Exporter(spark : SparkSession, apiClient: ApiClient, filter: KeyFilter, repoName: String, dstRoot: String, parallelism: Int) {
-  final private val defaultParallelism = 10
-
   def this(spark : SparkSession, apiClient: ApiClient, repoName: String, dstRoot: String) {
-    this(spark, apiClient, new SparkFilter(), repoName, dstRoot, defaultParallelism)
+    this(spark, apiClient, new SparkFilter(), repoName, dstRoot, Exporter.defaultParallelism)
   }
 
   def exportAllFromBranch(branch: String): Unit = {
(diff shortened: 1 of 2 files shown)


johnnyaug (approved):

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
```diff
@@ -42,16 +42,27 @@ class Exporter(spark : SparkSession, apiClient: ApiClient, filter: KeyFilter, re
...
+    val errs = actionsDF.mapPartitions( part => {
+      val pool: ExecutorService = Executors.newFixedThreadPool(par)
+
+      val callableTasks = new util.ArrayList[Callable[ExportStatus]]
+      while (part.hasNext) {
+        val row = part.next()
+        callableTasks.add(new Handler(f, round, ns, rel, serializedConf, row))
+      }
+
+      val res = pool.invokeAll(callableTasks).asScala.map(f => f.get()).iterator
```
arielshaqed, line 55: Why so procedural?  I think this should be enough (you may need to get an iterator from `res`, but that's about all).
```suggestion
      val res = pool.invokeAll(part.map(row => callableTasks.add(new Handler(f, round, ns, rel, serializedConf, row))))
```

Joker asks: Why so procedural?

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
```diff
@@ -42,16 +42,27 @@ class Exporter(spark : SparkSession, apiClient: ApiClient, filter: KeyFilter, re
...
+      res
+    }
     ).filter(status => !status.success)
 
     if (errs.isEmpty){
       return true
     }
 
-    writeSummaryFile(false, commitID, errs.sample(maxLoggedErrors).map(s => s.msg).reduce(_+"\n"+_))
+    writeSummaryFile(false, commitID, errs.take(maxLoggedErrors).map(s => s.msg).reduce(_+"\n"+_))
```
arielshaqed, line 65: Can we leave it as a `sample`?

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
```diff
@@ -102,7 +113,7 @@ class Exporter(spark : SparkSession, apiClient: ApiClient, filter: KeyFilter, re
 
   private def writeSummaryFile(success: Boolean, commitID: String, content : String) = {
     val suffix = if(success) "SUCCESS" else "FAILURE"
-    val dstPath = resolveURL(new URI(dstRoot), s"EXPORT_${commitID}_${suffix}")
+    val dstPath = URLResolver.resolveURL(new URI(dstRoot), s"EXPORT_${commitID}_${suffix}")
```
arielshaqed, line 116: Why do we need this change?

clients/spark/core/src/main/scala/io/treeverse/clients/URLResolver.scala
```diff
@@ -0,0 +1,11 @@
+package io.treeverse.clients
+
+import java.net.URI
+import org.apache.hadoop.fs.Path
+
+object URLResolver extends Serializable{
+  def resolveURL(uri: URI, extraPath: String): Path = {
+    val newPath: String = uri.getPath + '/' + extraPath
```
arielshaqed, line 8: I don't understand this function.  It seems to me it either does what `uri.resolve(extraPath)` does when extraPath is relative, or some _else_ (strange but in a different manner) when `extraPath` is absolute.  And it does something *really* odd when `extraPath` starts with a protocol.

arielshaqed (changes requested):
Thanks!

Most comments pertain to changes that appear unrelated to the main topic of the PR.

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
itaiad200: HAHA Done.

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
itaiad200: `Handler` uses it and calling `resolveURL`. Since `resolveURL` was part of `Exporter` it had to be `Serializable`. It's easier to use another struct than making our `APIClient` and other members `Serializable` too.

clients/spark/core/src/main/scala/io/treeverse/clients/URLResolver.scala
itaiad200: This function was just moved to another file (see explanation in previous comment):

> it either does what uri.resolve(extraPath) does when extraPath is relative

Nope, for `uri=s3://bucket/with/prefix`, `extraPath=/foo/bar`, `uri.resolve(extraPath)` will return (surprise surprise) `s3://bucket/foo/bar`.

> when extraPath is absolute

Didn't think about this one. I'll document against such usage

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
```diff
@@ -42,16 +42,27 @@ class Exporter(spark : SparkSession, apiClient: ApiClient, filter: KeyFilter, re
...
     val serializedConf = new SerializableWritable(hadoopConf)
     val f = filter
-    val errs = actionsDF.map(
-      row =>
-        Exporter.handleRow(f, round, ns, rel, serializedConf, row)
+    var par = parallelism
+    val errs = actionsDF.mapPartitions( part => {
+      val pool: ExecutorService = Executors.newFixedThreadPool(par)
+
+      val callableTasks = new util.ArrayList[Callable[ExportStatus]]
```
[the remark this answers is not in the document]
itaiad200: synced f2f - cannot be done.

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
arielshaqed: Not sure I follow.  `resolveURL` is a method, Spark tries to serialize a method?!

clients/spark/core/src/main/scala/io/treeverse/clients/URLResolver.scala
arielshaqed: > This function was just moved to another file (see explanation in previous comment):
> 
> > it either does what uri.resolve(extraPath) does when extraPath is relative
> 
> Nope, for `uri=s3://bucket/with/prefix`, `extraPath=/foo/bar`, `uri.resolve(extraPath)` will return (surprise surprise) `s3://bucket/foo/bar`.

Sorry, bad usage of the word "relative".  Should have said "URI with a relative path", which is a special case of a relative URL.  `extraPath` above is a relative URI with an absolute path.

To be clear(-er): I claim that a URI with a trailing slash resolves URIs with relative paths as resolveURL does.

The function above returns `s3://bucket/fwith/prefix//foo/bar`; note the double slashes / "empty" "directory".

> 
> > when extraPath is absolute
> 
> Didn't think about this one. I'll document against such usage

To summarize: This function does _the same thing_ as `URI.resolve` in the common case, except that it *forbids* a slash at the end of `uri`, whereas using `URI` like this *requires* it.  In all the uncommon cases (`extraPath` starts with a slash, uri ends/doesn't end in a slash, etc.), both do something "unexpected" in the sense that the code probably doesn't work.  I would argue that `URI.resolve` is _well-defined_ and _standard_.  We might be able to achieve parity on definitions, but not on standardization.  So `URI.resolve` wins to my understanding.

arielshaqed (approved):
Thanks!

I would _still_ very much prefer to use a known-good URI resolver.  However I cannot block once you are aware of this.  Please reconsider (and hopefully change).

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
itaiad200: Nope, if it is called inside a spark task (`exporter.resolveURL(...)`), the entire `exporter` class will be passed to the task (hence needs to be `Serializable`) to use its method. Even if `resolveURL` uses none of the `exporter` properties. 
That's why it made much more sense to extract that single method to its own simple class.

PR #1725, opened 2021-04-06
OpenAPI Python Client

nopcoder:
Part of #1725 
Changes made to the project Makefile to generate the python client code.
Added:
- `clients/python/lakefs/client.py` to enable a single client with access to all APIs
- `clients/python/build-package.sh` to build pypi package


.github/workflows/goreleaser.yaml
```diff
@@ -32,4 +32,4 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
```
itaiad200, line 35: Add newline.

.github/workflows/pypi.yaml
```diff
@@ -0,0 +1,30 @@
...
+    types: [published]
+
+jobs:
+  pypi:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: '1'
```
itaiad200, line 14: This is the default behaviour.

.github/workflows/pypi.yaml
```diff
@@ -0,0 +1,30 @@
...
+      - name: Extract version
+        shell: bash
+        run: echo "::set-output name=tag::$(echo ${GITHUB_REF##*/} | sed s/^v//g)"
+        id: version
+      - name: Python re-generate client
+        run: make client-python PACKAGE_VERSION=${{ steps.version.outputs.tag }}
+      - name: Python build package
+        run: make package-python 
+      - name: Python publish package
+        if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags')
```
itaiad200, line 24: How can it be a `push` event when you're only triggered by `published`?

.goreleaser.yml
```diff
@@ -2,7 +2,7 @@ project_name: lakeFS
 before:
   hooks:
     - go mod download
-    - make gen
+    - make gen PACKAGE_VERSION={{ .Version }}
```
itaiad200, line 5: You use `${{ steps.version.outputs.tag }}` in `pypi.yaml`

Makefile
```diff
@@ -7,6 +7,15 @@ NPM=$(or $(shell which npm), $(error "Missing dependency - no npm in PATH"))
 PROTOC_IMAGE="treeverse/protoc:3.14.0"
 PROTOC=$(DOCKER) run --rm -v $(shell pwd):/mnt $(PROTOC_IMAGE)
 
+# https://openapi-generator.tech
+OPENAPI_GENERATOR_IMAGE=openapitools/openapi-generator-cli:v5.1.0
+OPENAPI_GENERATOR=$(DOCKER) run --rm -v $(shell pwd):/mnt $(OPENAPI_GENERATOR_IMAGE)
+ifndef PACKAGE_VERSION
+	PACKAGE_VERSION=0.1.0.dev
```
itaiad200, line 14: Use git's SHA instead?

Makefile
```diff
@@ -91,6 +100,21 @@ go-install: go-mod-download ## Install dependencies
...
+	$(OPENAPI_GENERATOR) generate \
+		-i /mnt/$< \
+		-g python \
+		--package-name lakefs \
+		--additional-properties=infoName=Treeverse,infoEmail=services@treeverse.io,packageName=lakefs,packageVersion=$(PACKAGE_VERSION),projectName=lakefs,packageUrl=https://github.com/treeverse/lakeFS/tree/feature/python-client/clients/python \
+		-o /mnt/clients/python
+
+clients: client-python
+
+package-python:
```
itaiad200, line 113: Shouldn't it depend on `client-python`?

clients/python/.gitignore
```diff
@@ -0,0 +1,66 @@
+# Byte-compiled / optimized / DLL files
```
itaiad200, line 1: Do we really need everything in this file?

clients/python/.gitlab-ci.yml
```diff
@@ -0,0 +1,24 @@
+# ref: https://docs.gitlab.com/ee/ci/README.html
```
itaiad200, line 1: Can you add a comment about this file?
I have 0 understanding of why we need it.

.github/workflows/pypi.yaml
nopcoder: true, I was triggering it manually - I'll remove the first part and just keep the tags check

.goreleaser.yml
nopcoder: Right, this was part of a test to use gorelease to publish the package - will remove it

Makefile
nopcoder: Needed a constat. The build process runs `gen` and `checks-validator` which generate the client and check if there is a diff the client's API folder for changes.
This is to make sure in the PR step that you build the python client after updating the swagger API.

clients/python/.gitignore
nopcoder: it is auto-generated by the openapi-generator

clients/python/.gitlab-ci.yml
nopcoder: I'll remove it and add it to the ignore list so the generator will not re-create it.

itaiad200 (approved):
Great work!

PR #1733, opened 2021-04-07
Specify key structure type (relative/full) in entry model

### LINKED ISSUE #1646: block.ResolveNamespace treats anything unparsable as a relative URL
These lines https://github.com/treeverse/lakeFS/blob/10c7fcf914d423d4b590ed1f9773b7ef68f4c092/pkg/block/namespace.go#L116-L118 in block.ResolveNamespace are called for practically _any_ URL on the underlying block store.  So *any* parse failure means that key is treated as a relative URL and ends up appended (with some leading/trailing slashes stripped) to the relative namespace URL by `formatPathWithNamespace`.

While directory traversal is impossible in production (because blockstores don't know about `..`), our local block adapter can probably be fooled on this.  Additionally, this may easily create very weird results when some other error condition occurred (e.g. invalid characters in text).

This one will be somewhat tricky to fix, because `url` -- being a standard Go library -- doesn't report errors correctly: there are no base error `var`s that one can use to determine which error was returned.

itaiad200:
closes #1646


pkg/block/namespace.go
```diff
@@ -146,6 +152,20 @@ func ResolveNamespace(defaultNamespace, key string) (QualifiedKey, error) {
...
 }
 
+func parseKey(key string) (parsedKey *url.URL, relative bool, err error) {
+	parsedKey, err = url.ParseRequestURI(key)
+	if err == nil {
+		return parsedKey, false, nil
+	}
+	if strings.Contains(key, "://") {
+		// No better way to understand if the key failed to parse since it's relative
+		// or due to bad formatting. See https://github.com/treeverse/lakeFS/issues/1646
```
arielshaqed, line 162: Not sure I understand.  `s3://a/path/with/a/prefix/ending/in/://and/an/empty/component` is a perfectly valid path. We should be able to do better, these are URLs after all, and have a unique parse.

arielshaqed (changes requested):
Not sure why this solves the linked issue.  AFAICT it introduces many limitations on otherwise-valid paths.

clients/spark/core/src/main/resources/catalog.proto
```diff
@@ -12,4 +12,10 @@ message Entry {
 	int64 size = 3;
 	string e_tag = 4;
 	map<string,string> metadata = 5;
+	enum AddressType {
+		UNKNOWN = 0;
```
arielshaqed, line 16: As an alternative, we might:
1. Rename the field `address` -> `deprecated_by_prefix_address` (or otherwise discourage its use).
1. Add a new field `string address = 7`.
1. Have just `RELATIVE` (0, default) and `FULL` address types.
1. Document that `address_type` applies _only_ to `address`.

This ensures we can distinguish missing type fields to see whether they are old (and intentional) or new (and therefore an error).

pkg/block/adapter.go
```diff
@@ -15,6 +15,10 @@ type MultipartUploadCompletion struct{ Part []*s3.CompletedPart }
 type ObjectPointer struct {
 	StorageNamespace string
 	Identifier       string
+
+	// Indicates whether the Identifier is relative to the StorageNamespace.
+	// nil value means it's unknown.
```
arielshaqed, line 20: Easier to read _and_ write calls if we use a tristate bool here.

pkg/catalog/model.go
```diff
@@ -56,6 +58,38 @@ type Tag struct {
...
+// AddressType is the type of an entry address
+type AddressType int32
+
+const (
+	// AddressTypeUnknown indicates that the address might be relative or full.
+	// Used only for backward compatibility and should not be used for creating entries.
+	AddressTypeUnknown AddressType = 0
+
+	// AddressTypeRelative indicates that the address is relative to the storage namespace.
+	// For example: "/foo/bar"
```
arielshaqed, line 70: I don't understand.  `namespace_test.go` has _no leading slashes_ on its relative addresses; doesn't this refer to the same thing?

pkg/catalog/model.go
```diff
@@ -56,6 +58,38 @@ type Tag struct {
...
+	AddressTypeRelative AddressType = 1
+
+	// AddressTypeFull indicates that the address is the full address of the object in the object store.
+	// For example: "s3://bucket/foo/bar"
+	AddressTypeFull AddressType = 2
+)
+
+func (at AddressType) IsRelative() *bool {
```
arielshaqed, line 78: The name `Is*` is confusing for a function returning a tristate.

arielshaqed (changes requested):
Thanks!

Most important changes are around tristates (not overloading them!) and some documentation.

pkg/catalog/model.go
itaiad200: Done.

pkg/catalog/model.go
itaiad200: Done.

pkg/block/adapter.go
itaiad200: Done.

pkg/block/namespace_test.go
```diff
@@ -64,35 +70,61 @@ func TestResolveNamespace(t *testing.T) {
...
+			switch r {
+			case block.IdentifierTypeUnknownDeprecated:
+				relativeName = "unknown"
+			case block.IdentifierTypeRelative:
+				relativeName = "relative"
+			case block.IdentifierTypeFull:
+				relativeName = "full"
 			}
-		})
+			t.Run(fmt.Sprintf("%s_%s", cas.Name, relativeName), func(t *testing.T) {
```
arielshaqed, line 117:
```suggestion
			t.Run(fmt.Sprintf("%s/%s", cas.Name, relativeName), func(t *testing.T) {
```
will make it more like the regular naming of nested `t.Run`s.  Ideally you would have `t.Run(case.Name)` and nest `t.Run(relativeName)` inside it.  But my 80-column brain can understand why you might wish to avoid that much indentation.

arielshaqed (approved):
Cool, thanks!

Issue #501, opened 2020-08-17
Align config name "access_secret_key" to "secret_access_key"


itaiad200:
Currently, the AWS secret key is given under the key `blockstore.s3.credentials.access_secret_key` in the lakeFS configuration file.

However, in other places (for example, when logging in from the UI) this configuration is called "secret_access_key". This one is also the convention in the AWS ecosystem.

Align all `access_secret_key` appearances in our code base to `secret_access_key`.

aaronmbos:
Hey 👋 open source newbie here! I'd like to pick this issue up for the upcoming Hacktoberfest as it seems relatively straightforward. I'll follow all the instructions in the contribution guide and reach out on the Slack channel for any questions.

YaelRiv:
Hello Aaron! 

It's yours. Thanks for giving us a hand 💪
See you on the Slack channel. 

Yael

iamRishabh07:
Hi, I would like to work on this issue.

YaelRiv:
@iamRishabh07 Thanks for your offer! This is already assigned.

Issue #1729, opened 2021-04-07
Validate configuration file


arielshaqed:
Configuration uses `viper`, which is super-easy to use.  Unfortunately it _performs no validation_ of the configuration file.  So e.g. in #1728 the user and I have no way of knowing whether the configuration is valid or not (hint: it contains the invalid configuration key `blockstore.s3.credentials.credentials_file`).

The `viper` API has way of knowing the configuration schema (because everything accessed is optional, and everything _not_ accessed might be accessed later).  Our users probably deserve something that validates.

arielshaqed:
We *could* add separate config file validation on top of `viper`.  I argue that that is a tech debt generator: there is no way to verify that the schema is consistent with the code.

I *really* want to use another library, one that can validate.

arielshaqed:
https://github.com/spf13/viper/issues/702 :-(

arielshaqed:
One way to validate is to use https://pkg.go.dev/github.com/spf13/viper#UnmarshalExact, which will error out (see oauth2-proxy/oauth2-proxy#500).

WIll try that, see how we like it.

PR #1734, opened 2021-04-07
align "access_secret_key" to "secret_access_key"

johnnyaug:
Solves #501


arielshaqed (commented):
Oh yeah!

The thing is, this breaks every single config file out there.  You're not going to believe this, I'm a
# :flags:  Bearer of Good News :unicorn: 
here...

In #1729 we're switching to deeper use of viper, mostly to get validation.  *But* I discovered that viper also has a way to alias one key onto the other -- "[Registering and Using Aliases](https://pkg.go.dev/github.com/spf13/viper#readme-registering-and-using-aliases)".  So we can keep the old name and just stop documenting it.  (And of course add that to config_test)

WDYT?

arielshaqed:
May fix #501.

pkg/config/config.go
```diff
@@ -99,12 +99,15 @@ const (
...
-	BlockstoreS3MaxRetriesKey            = "blockstore.s3.max_retries"
+	BlockstoreTypeKey                         = "blockstore.type"
+	BlockstoreLocalPathKey                    = "blockstore.local.path"
+	BlockstoreS3RegionKey                     = "blockstore.s3.region"
+	BlockstoreS3StreamingChunkSizeKey         = "blockstore.s3.streaming_chunk_size"
+	BlockstoreS3StreamingChunkTimeoutKey      = "blockstore.s3.streaming_chunk_timeout"
+	BlockstoreS3MaxRetriesKey                 = "blockstore.s3.max_retries"
+	BlockstoreS3CredentialsAccessKeyIdKey     = "blockstore.s3.credentials.access_key_id"
+	BlockstoreS3CredentialsSecretAccessKeyKey = "blockstore.s3.credentials.secret_access_key"
+	BlockstoreS3CredentialsSessionTokenKey    = "blockstore.s3.credentials.session_token"
```
arielshaqed, line 110: I'm in the middle of a mega-cleanup of config.go, #1729.  These string keys will soon be gone (all of them, entirely), so please avoid fixing them for now.  I promise I shall eradicate them.
Sorry for not being specific about this when I said I was "doing something something".

pkg/config/config.go
```diff
@@ -174,6 +177,8 @@ func setDefaults() {
 
 	viper.SetDefault(BlockstoreAzureTryTimeoutKey, DefaultAzureTryTimeout)
 	viper.SetDefault(BlockstoreAzureAuthMethod, DefaultAzureAuthMethod)
+	// allow using deprecated key for AWS secret key:
+	viper.RegisterAlias(BlockstoreS3CredentialsSecretAccessKeyKey, "blockstore.s3.credentials.access_secret_key")
```
arielshaqed, line 181:
```suggestion
	viper.RegisterAlias("blockstore.s3.credentials.access_secret_key", BlockstoreS3CredentialsSecretAccessKeyKey)
```
I think the old name should be the alias, the new name should be the new key.

arielshaqed (commented):
Great, thanks!

Please _don't_ give names to any other keys, I am eliminating all of these string names and it will only make my conflicts worse.

Approving as soon as those named `*Key`s are gone.

pkg/config/config.go
```diff
@@ -174,6 +174,8 @@ func setDefaults() {
 
 	viper.SetDefault(BlockstoreAzureTryTimeoutKey, DefaultAzureTryTimeout)
 	viper.SetDefault(BlockstoreAzureAuthMethod, DefaultAzureAuthMethod)
+	// allow using deprecated key for AWS secret key:
+	viper.RegisterAlias("blockstore.s3.credentials.secret_access_key", "blockstore.s3.credentials.access_secret_key")
```
arielshaqed, line 178: Let's test this alias in at least one place. You know,  for BC.

arielshaqed (approved):
Very cool, thanks! We might want to test that the old version still works,  say in auth_test.

arielshaqed (approved):
Thanks!  The no-aliasing hack is unavoidable but truly horrible.  Can you add a comment to this PR to link to the Viper bug?

johnnyaug:
This is a workaround to https://github.com/spf13/viper/issues/689

arielshaqed:
> This is a workaround to bypass [spf13/viper#689](https://github.com/spf13/viper/issues/689)

That bug has a PR to fix it. And that PR will be 3 tomorrow!

@johnnyaug @nopcoder istio have [a fork of viper](https://github.com/istio/istio/pull/13594/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R9), which may have fixed it (the istio PR links to 689). Perhaps time to investigate whether the istio viper is better for us than the spf13 viper?

nopcoder:
Thanks @arielshaqed, will take a look tomorrow the difference.

PR #1737, opened 2021-04-08
docs: lakeFS endpoint as URI documentation changes

nopcoder:
Part of #1489. Should be merged after #1717 is released.

### DIFF
--- a/docs/architecture/sizing-guide.md
+++ b/docs/architecture/sizing-guide.md
@@ -139,7 +139,7 @@ lakectl abuse random-read \
     --from-file randomly_selected_paths.txt \
     --amount 500000 \
     --parallelism 128 \
-    lakefs://example-repo@<commit hash>
+    lakefs://example-repo/<commit hash>
 ```
 
 **Result Histogram (raw):**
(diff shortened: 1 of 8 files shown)


itaiad200 (changes requested):
This is where not having versioning for our docs bites us in the a**..
It becomes unusable for any prior versions.

I strongly recommend adding examples for older versions instead of replacing existing ones.

nopcoder:
> This is where not having versioning for our docs bites us in the a**..
> It becomes unusable for any prior versions.
> 
> I strongly recommend adding examples for older versions instead of replacing existing ones.

agree, need versioned documentation - about older examples, for all the docs or specific section.
adding note and example in the lakectl page will be enough? wdyt?

itaiad200:
> > This is where not having versioning for our docs bites us in the a**..
> > It becomes unusable for any prior versions.
> > I strongly recommend adding examples for older versions instead of replacing existing ones.
> 
> agree, need versioned documentation - about older examples, for all the docs or specific section.
> adding note and example in the lakectl page will be enough? wdyt?

I think that we need at least one example per with the old format.
Or instead a Note that explains how to reference an object in older versions (again - the note should exists in all applicable pages)

PR #1738, opened 2021-04-08
Error on Azure async copy

### LINKED ISSUE #1474: Azure adapter - Copy command should be synchronous
The Azure adapter copy command uses the [Azure API copy command](https://docs.microsoft.com/en-us/rest/api/storageservices/copy-blob) which is asynchronous,
We should change The adapter copy command to be synchronous by Polling the destination blob until we get a result.

guy-har:
closes #1474


nopcoder (approved):
LGTM!

PR #1740, opened 2021-04-11
fix response parse for create branch

### LINKED ISSUE #1741: UI: create branch succeed with error
Create branch from the UI succeed with error fail to parse JSON response.

nopcoder:
Close #1741


itaiad200 (approved):
Approving but link to the bug pls. Thanks!!

PR #1743, opened 2021-04-11
List repository actions should not check branch existence

### LINKED ISSUE #1727: Bug: list actions by branch after branch delete should return old runs
Listing actions by branch checks if a branch exists when it should not.
The information is kept after the branch is deleted and should be available in the API.

nopcoder:
Fix #1727

Removed the check for branch exist for list actions runs.
To fix the test code was required to pass the same block adapter from our catalog to all of other services used by our controller test code.

### DIFF
--- a/pkg/api/controller_test.go
+++ b/pkg/api/controller_test.go
@@ -9,6 +9,7 @@ import (
 	"math"
 	"mime/multipart"
 	"net/http"
+	"net/http/httptest"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -26,7 +27,6 @@ import (
 )
 
 const (
-	//timeout       = 10 * time.Second
 	DefaultUserID = "example_user"
 )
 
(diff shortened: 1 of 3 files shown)


pkg/api/controller_test.go
```diff
@@ -1285,3 +1285,95 @@ func TestController_SetupLakeFSHandler(t *testing.T) {
...
+	defer httpServer.Close()
+	// create repository
+	resp, err := clt.CreateRepositoryWithResponse(ctx, &api.CreateRepositoryParams{}, api.CreateRepositoryJSONRequestBody{
+		DefaultBranch:    api.StringPtr("master"),
+		Name:             "repo9",
+		StorageNamespace: "mem://repo9",
+	})
+	verifyResponseOK(t, resp, err)
+	// upload action for pre-commit
+	actionContent := `---
```
arielshaqed, line 1304: This is quite hard to read.  Can we e.g. indent?

I tried with a random Python YAML parser and with The YPaste Official YAML cgi-bin, and I think that this works (http://ben-kiki.org/ypaste/data/2571/index.html):
```yaml
---
   a:
     - 2
     - 3
   b: yes
   c: no
```

So you might be able to get by with indenting all of the lines.

pkg/api/controller_test.go
```diff
@@ -1285,3 +1285,95 @@ func TestController_SetupLakeFSHandler(t *testing.T) {
...
+	t.Run("total", func(t *testing.T) {
+		respList, err := clt.ListRepositoryRunsWithResponse(ctx, "repo9", &api.ListRepositoryRunsParams{
+			Amount: api.PaginationAmountPtr(100),
+		})
+		verifyResponseOK(t, respList, err)
+		runsCount := len(respList.JSON200.Results)
+		expectedRunsCount := contentCount + 1 // additional one for the actions commit
+		if runsCount != expectedRunsCount {
+			t.Fatalf("ListRepositoryRuns() got %d results, expected %d", runsCount, expectedRunsCount)
+		}
```
arielshaqed, line 1349: I think this error message will be easier to read.  Also the code is shorter.
```suggestion
		if runsCount != contentCount + 1 { // additional one for the actions commit
			t.Fatalf("ListRepositoryRuns() got %d results, expected %d+1", runsCount, contentCount+1)
		}
```

arielshaqed (approved):
Neat!

pkg/api/controller_test.go
nopcoder: checking if it is possible or an alternative

nopcoder: cloud not indent so I've extract it into a template variable - let me know if it is more readable now.

arielshaqed (approved):
Great!

Your solution for indentation looking horrible is much simpler & better than my proposal :-)

Issue #1665, opened 2021-03-22
Collect Azure CloudID


itaiad200:
- Collect Azure cloud id
- Update the view in https://github.com/treeverse/callhome/blob/master/reporting_views_ddl.sql to exclude our Azure cloud id
- Revisit the existing Redash widgets to confirm that our Azure cloud id is excluded from queries.

PR #1750, opened 2021-04-11
feature/collect azure stats

johnnyaug:
Solves #1665

### DIFF
--- a/pkg/cloud/azure/metadata.go
+++ b/pkg/cloud/azure/metadata.go
@@ -20,7 +20,9 @@ func NewMetadataProvider(logger logging.Logger) *MetadataProvider {
 }
 
 type instanceMetadataResponse struct {
-	SubscriptionId string `json:"subscriptionId"`
+	Compute struct {
+		SubscriptionId string `json:"subscriptionId"`
+	} `json:"compute"`
 }
 
 func (m *MetadataProvider) GetMetadata() map[string]string {
@@ -45,8 +47,12 @@ func (m *MetadataProvider) GetMetadata() map[string]string {
 		m.logger.Warnf("%v: failed to get Azure subscription ID from instance metadata", err)
 		return nil
 	}
+	if responseObj.Compute.SubscriptionId == "" {
+		m.logger.Warn("got empty subscription id from azure")
+		return nil
+	}
 	return map[string]string{
-		cloud.IDKey:     fmt.Sprintf("%x", md5.Sum([]byte(responseObj.SubscriptionId))), //nolint:gosec
+		cloud.IDKey:     fmt.Sprintf("%x", md5.Sum([]byte(responseObj.Compute.SubscriptionId))), //nolint:gosec
 		cloud.IDTypeKey: "azure_subscription_id",
 	}
 }
--- a/pkg/cloud/azure/metadata.go
+++ b/pkg/cloud/azure/metadata.go
@@ -1,0 +1,52 @@
+package azure
+
+import (
+	"crypto/md5" //nolint:gosec
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+
+	"github.com/treeverse/lakefs/pkg/cloud"
+	"github.com/treeverse/lakefs/pkg/logging"
+)
+
+type MetadataProvider struct {
+	logger logging.Logger
+}
+
+func NewMetadataProvider(logger logging.Logger) *MetadataProvider {
+	return &MetadataProvider{logger: logger}
+}
+
+type instanceMetadataResponse struct {
+	SubscriptionId string `json:"subscriptionId"`
+}
+
+func (m *MetadataProvider) GetMetadata() map[string]string {
+	var PTransport = &http.Transport{Proxy: nil}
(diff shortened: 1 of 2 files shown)


pkg/cloud/azure/metadata.go
```diff
@@ -0,0 +1,58 @@
...
+type instanceMetadataResponse struct {
+	Compute struct {
+		SubscriptionId string `json:"subscriptionId"`
+	} `json:"compute"`
+}
+
+func (m *MetadataProvider) GetMetadata() map[string]string {
+	var PTransport = &http.Transport{Proxy: nil}
+	client := http.Client{Transport: PTransport}
+	req, _ := http.NewRequest("GET", "http://169.254.169.254/metadata/instance", nil)
```
nopcoder, line 31: 1. http.MethodGet
2. check and log error - in case of an error req will be nil and it will trigger panic on the next line
3. Use NewRequestWithContext+timeout or set Timeout on the HTTP client for this request as the default will hang forever.

pkg/cloud/azure/metadata.go
```diff
@@ -0,0 +1,58 @@
...
+
+type instanceMetadataResponse struct {
+	Compute struct {
+		SubscriptionId string `json:"subscriptionId"`
+	} `json:"compute"`
+}
+
+func (m *MetadataProvider) GetMetadata() map[string]string {
+	var PTransport = &http.Transport{Proxy: nil}
+	client := http.Client{Transport: PTransport}
```
nopcoder, line 30: Why do we need to set no proxy? - the default uses the environment to capture this information which is recommended.

pkg/cloud/azure/metadata.go
```diff
@@ -0,0 +1,58 @@
...
+	req, _ := http.NewRequest("GET", "http://169.254.169.254/metadata/instance", nil)
+	req.Header.Add("Metadata", "True")
+	q := req.URL.Query()
```
nopcoder, line 33: No need to parse the url query

```suggestion
	q := url.Values{}
```

pkg/cloud/azure/metadata.go
```diff
@@ -0,0 +1,58 @@
...
+	q.Add("format", "json")
+	q.Add("api-version", "2019-03-11")
+	req.URL.RawQuery = q.Encode()
+	resp, err := client.Do(req)
+	if err != nil {
+		m.logger.Warnf("%v: failed to get Azure subscription ID from instance metadata", err)
```
nopcoder, line 39:
```suggestion
		m.logger.WithError(err).Warn("Failed to get Azure subscription ID from instance metadata")
```
structured logging

pkg/cloud/azure/metadata.go
```diff
@@ -0,0 +1,58 @@
...
+		return nil
+	}
+	defer resp.Body.Close()
```
nopcoder, line 42:
```suggestion
	defer func() {
		_ = resp.Body.Close()
	}()
```

pkg/cloud/azure/metadata.go
```diff
@@ -0,0 +1,58 @@
...
+	responseBody, _ := ioutil.ReadAll(resp.Body)
+	responseObj := &instanceMetadataResponse{}
+	err = json.Unmarshal(responseBody, responseObj)
+	if err != nil {
+		m.logger.Warnf("%v: failed to get Azure subscription ID from instance metadata", err)
```
nopcoder, line 51:
```suggestion
		m.logger.Error(err).Warn("Failed to get Azure subscription ID from instance metadata")
```

pkg/cloud/azure/metadata.go
```diff
@@ -0,0 +1,58 @@
...
+		return nil
+	}
+	if responseObj.Compute.SubscriptionId == "" {
+		m.logger.Warn("got empty subscription id from azure")
```
nopcoder, line 55: Why not consider empty value as valid and have a hash for all instances without subscription Id?
Also if it is something that can be an option for lakeFS - running without an option to extract this information - this should be logged as info and not a warning.

nopcoder (approved):

nopcoder:
Minor questions and suggestions

pkg/cloud/azure/metadata.go
```diff
@@ -0,0 +1,62 @@
...
+}
+
+type instanceMetadataResponse struct {
+	Compute struct {
+		SubscriptionID string `json:"subscriptionId"`
+	} `json:"compute"`
+}
+
+func (m *MetadataProvider) GetMetadata() map[string]string {
+	var PTransport = &http.Transport{Proxy: nil}
```
nopcoder, line 29: Not a valid variable name

pkg/cloud/azure/metadata.go
johnnyaug: An empty value is an empty value: we want to know that it was empty if it was reported as such.
It's harder to work with the hash of the empty string in all SQL queries.

guy-har (approved):
LGTM! 

Doesn't Azure have any SDK for getting the subscriptionID ?

johnnyaug:
@guy-har Unfortunately no

PR #1754, opened 2021-04-12
Switch parents order in commit to match git ordering

### LINKED ISSUE #1730: Bug: parent order for merge commits is reversed
Background: when merging a branch into another, the resulting merge commit will have 2 parents:
- The commit the source branch is currently pointing to
- The latest commit on the destination branch at the time of merge

When Git creates a merge commit, their order is always `[destination branch, source branch]`.
This is useful for diffing: for every commit, whether it's a merge commit or not, comparing the commit with its first parent will result in a diff of all the changes introduced from parent to child.

In lakeFS, the order is currently `[source branch, destination branch]`.
This means that when doing a diff with the first parent, we're asking what changed between a set of changes and the results of merging them to another branch. The answer is an empty diff. To get the "correct" result when diffing, we need to special-case merge commits and compare the commit with its second parent. This not only breaks compatibility with Git, it makes the developer experience cumbersome.

This is, of course, a breaking change - diffing merge commits genereted prior to making this change will incorrectly show an empty diff. I suggest doing it *now*, since it'll always be mroe painful to do it later.

itaiad200:
close #1730


arielshaqed (changes requested):
Can you please add the plan to this PR (or to the linked issue 1730)?  I don't get it.  Specifically, AFAICT this will change the order of parents without changing anything on the DB schema or backfilling old commits.  So (I _think_) it means that we can never do any backfill.  Are we committing to having good merge histories only past the release of this PR?

itaiad200:
> Can you please add the plan to this PR (or to the linked issue 1730)? I don't get it. Specifically, AFAICT this will change the order of parents without changing anything on the DB schema or backfilling old commits. So (I _think_) it means that we can never do any backfill. Are we committing to having good merge histories only past the release of this PR?

Changed this PR entirely to migrate the DB with correct ordering of the parents. Let me know if something is still unclear.

itaiad200:
Still adding more verifications to the test and testing on a live installation.

itaiad200:
Updating on the status of this PR.
As agreed, we will only fix the bug of the parents swap, and mark versions for commits.
Commit existing prior to this fix will keep the version `commit:v1`, and all new commits will have the version `commit:v2`.
Since the commit version was hardcoded, the hash id of the old commits isn't expected to change.
In order to support the versioning, `version` column was added to the `graveler_commits` table and the table was backfilled.

The change was tested by first creating a repo in version v0.33.1, committing and merging, then upgrading lakeFS and testing again after the migration. No issues were encountered.

pkg/ddl/000026_add_commit_version.up.sql
```diff
@@ -0,0 +1,4 @@
+BEGIN;
+ALTER TABLE graveler_commits
+    ADD COLUMN IF NOT EXISTS version text DEFAULT 'commit:v1' NOT NULL;
```
johnnyaug, line 3: make numerical.
find new name for column

pkg/graveler/ref/manager.go
```diff
@@ -71,6 +71,7 @@ func createBareRepository(tx db.Tx, repositoryID graveler.RepositoryID, reposito
 
 func (m *Manager) CreateRepository(ctx context.Context, repositoryID graveler.RepositoryID, repository graveler.Repository, token graveler.StagingToken) error {
 	firstCommit := graveler.Commit{
+		Version:      graveler.CommitVersion,
```
johnnyaug, line 74: set this inside graveler, not from the outside

pkg/ddl/000026_add_commit_version.up.sql
itaiad200: If I make it numerical we'll lose compatibility with what we had so far which was 'commit:v1'

itaiad200: And what's wrong with `version`?

pkg/graveler/ref/manager.go
itaiad200: Agree and disagree with you on this one.
This function has access to all of graveler internals (like `FirstCommitMsg` and even the time is debatable) and the version is not an exception to that. Without setting it here we'll need to add an API (or a const) to graveler named `GetFirstCommit`. Do you like that better?

pkg/ddl/000026_add_commit_version.up.sql
arielshaqed: If it's a string. please document it clearly as "must have sortable keys" or something.  Otherwise how do we know whether we're before or after a given version?
Also, this is a pretty large field on disk (doesn't matter that much for commits).
Personally I'd go for named (boolean) feature toggles, or ints.

pkg/graveler/graveler_test.go
```diff
@@ -601,17 +601,21 @@ func TestGraveler_PreMergeHook(t *testing.T) {
 	branchLocker := ref.NewBranchLocker(conn)
 	const expectedRangeID = graveler.MetaRangeID("expectedRangeID")
 	const expectedCommitID = graveler.CommitID("expectedCommitId")
+	const expectedCommitID2 = graveler.CommitID("expectedCommitId2")
```
arielshaqed, line 604: Don't understand what's a "commit ID 2"?  Please document or rename.

pkg/graveler/graveler_test.go
```diff
@@ -659,6 +663,16 @@ func TestGraveler_PreMergeHook(t *testing.T) {
 			if err != nil && !errors.As(err, &hookErr) {
 				t.Fatalf("Merge err=%v, pre-merge error expected HookAbortError", err)
 			}
+			if refManager.AddedCommit.MetaRangeID == "" {
+				t.Fatalf("Empty MetaRangeID, commit was successful")
```
arielshaqed, line 667: Can you log the entire commit?  It may have clues.  (And if not, just `t.Fatal`).

pkg/graveler/graveler_test.go
```diff
@@ -659,6 +663,16 @@ func TestGraveler_PreMergeHook(t *testing.T) {
...
+			}
+			parents := refManager.AddedCommit.Parents
+			if len(parents) != 2 {
+				t.Fatal("Merge commit should have 2 parents")
```
arielshaqed, line 671: Can you log some clues here, too?

pkg/graveler/ref/resolve_ref_test.go
```diff
@@ -26,6 +26,7 @@ func TestResolveRef_Dereference(t *testing.T) {
 	var previous graveler.CommitID
 	for i := 0; i < 20; i++ {
 		c := graveler.Commit{
+			Version:      "commit:v1",
```
arielshaqed, line 29: The behaviour of `~` (and possibly `^`) matters a great deal with this change, so let's test it!

arielshaqed (changes requested):
Thanks, I think we should be done soon!

Mostly changes related to naming and documenting.

pkg/ddl/000026_add_commit_version.up.sql
nopcoder: Combine multiple inputs from the above and @talSofer - `version` will be numeric where zero is the initial version and each change will have a new number.
In order to identify it in the code, we will have something like an enum where each value will be a change we made.
Version will be part of the commit identify will be savedd/loaded

pkg/graveler/ref/manager.go
nopcoder: Will add `NewCommit` func that will return the struct initialized so the use of the constant of the current version will be in just one place.

pkg/graveler/ref/resolve_ref_test.go
nopcoder: 1. Added test
2. Kept "commit:v1" as a marker - so no new identity will be created

johnnyaug (approved):

ozkatz (approved):
LGTM!

arielshaqed (approved):
Thanks!
sorry for delay responding.

pkg/ddl/000026_add_commit_version.up.sql
itaiad200: Commits hash for version 0 were calculated with 'commit:v1'. Not sure we need to keep it compatible, as I can't think of a scenario where we recalculate it.

PR #1758, opened 2021-04-12
python client class ensure server endpoint incldue default path api/v1

nopcoder:
LakeFSClient client class should ensure that configuration's host property will include the API base path.
Example:
```py
import lakefs_client
from lakefs_client.client import LakefsClient
configuration = lakefs_client.Configuration(
  host = "http://localhost:8000",
  username = "key",
  password = "secret"
)
client = LakeFSClient(configuration=configuration)
```
The LakeFSClient will ensure that the API base URL will include "/api/v1" (DEFAULT_BASE_URL_PATH constant) in the endpoint so the end-user will not be required to align the API endpoint with the package version.

### DIFF
--- a/clients/python/lakefs_client/client.py
+++ b/clients/python/lakefs_client/client.py
@@ -4,13 +4,10 @@ import lakefs_client
 from lakefs_client.apis import ActionsApi, AuthApi, BranchesApi, CommitsApi, ConfigApi, HealthCheckApi, MetadataApi, \
     ObjectsApi, RefsApi, RepositoriesApi, StagingApi, TagsApi
 
-DEFAULT_BASE_URL_PATH = "/api/v1"
-
-
 class LakeFSClient:
     def __init__(self, configuration=None, header_name=None, header_value=None, cookie=None, pool_threads=1):
         if configuration:
-            configuration.host = LakeFSClient._ensure_endpoint(configuration.host)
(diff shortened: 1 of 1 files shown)


clients/python/lakefs_client/client.py
```diff
@@ -1,9 +1,18 @@
+from urllib.parse import urlparse, urlunparse
+
 import lakefs_client
-from lakefs_client.apis import ActionsApi, AuthApi, BranchesApi, CommitsApi, ConfigApi, HealthCheckApi, MetadataApi, ObjectsApi, RefsApi, RepositoriesApi, StagingApi, TagsApi
+from lakefs_client.apis import ActionsApi, AuthApi, BranchesApi, CommitsApi, ConfigApi, HealthCheckApi, MetadataApi, \
+    ObjectsApi, RefsApi, RepositoriesApi, StagingApi, TagsApi
+
+DEFAULT_BASE_URL_PATH = "/api/v1"
```
itaiad200, line 7: Doesn't exist in the generated code?
If it does use it instead of copying it here.

itaiad200 (approved):

clients/python/lakefs_client/client.py
nopcoder: added code to extract the path from the first host configuration

itaiad200 (approved):

PR #1759, opened 2021-04-13
Feature/lakectl internal paging

johnnyaug:


cmd/lakectl/cmd/repo.go
```diff
@@ -160,7 +137,7 @@ func init() {
 	repoCmd.AddCommand(repoCreateBareCmd)
 	repoCmd.AddCommand(repoDeleteCmd)
 
-	repoListCmd.Flags().Int("amount", -1, "how many results to return, or '-1' for default (used for pagination)")
+	repoListCmd.Flags().Int("amount", 100, "number of results to return")
```
nopcoder, line 140: constant. relevant to all the places in this change.

pkg/actions/db_task_result_iterator.go
```diff
@@ -61,7 +61,7 @@ func (it *DBTaskResultIterator) maybeFetch() {
 		Select("run_id", "hook_run_id", "hook_id", "action_name", "start_time", "end_time", "passed").
 		From("actions_run_hooks").
 		Where(sq.Eq{"repository_id": it.repositoryID, "run_id": it.runID}).
-		Where(sq.Gt{"action_name": it.offset}).
+		Where(sq.Gt{"hook_run_id": it.offset}).
```
nopcoder, line 64: good catch - I need to a test for this one

pkg/auth/service.go
```diff
@@ -561,7 +561,7 @@ func (s *DBAuthService) ListGroupUsers(ctx context.Context, groupDisplayName str
 			WHERE
 				auth_groups.display_name = $1
 				AND auth_users.display_name > $2
-			ORDER BY auth_groups.display_name
+			ORDER BY auth_users.display_name
```
nopcoder, line 564: how this one worked so far

nopcoder (changes requested):

pkg/auth/service.go
johnnyaug: it didn't :)

pkg/auth/service.go
```diff
@@ -582,9 +582,10 @@ func (s *DBAuthService) ListGroupUsers(ctx context.Context, groupDisplayName str
 			WHERE
 				auth_groups.display_name = $1
 				AND auth_users.display_name > $2
-			ORDER BY auth_users.display_name
-			LIMIT $3`,
-			groupDisplayName, params.After, params.Amount+1)
+				AND auth_users.display_name like $3
```
nopcoder: why listing needs to filter by prefix? we use this one as `after` which means you should get something that doesn't match the value's prefix.

nopcoder: All good - didn't saw it is prefix.

nopcoder (approved):

PR #1763, opened 2021-04-17
Remove codecov call and write env

arielshaqed:
https://about.codecov.io/security-update/

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -91,8 +91,8 @@ jobs:
       - name: Run tests
         run: |
           make test
-      - name: Publish coverage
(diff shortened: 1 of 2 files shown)


nopcoder (approved):

ozkatz (approved):
Thanks!

arielshaqed:
Pulling *untested*.  GH actions seem down right now (possibly related to the CodeCov breakdown).  I believe that this affects only GH actions, so it is safe.  And I further believe that this change is good.

Feel *very* free to roll back if either of the above is incorrect, or if you disagree, or _especially_ if anything breaks!

PR #1773, opened 2021-04-18
Make mandatory tests non conditional

ozkatz:
Currently, if a PR introduces a change that doesn't trigger a mandatory tests,
That PR will never be marked as passing, since mandatroy tests did not complete.

Probably better to have tests run even if they don't add any additional coverage,
than have unmergable PRs. This is not a good solution, but it doesn't seem like
GitHub actions supports anything better.

### DIFF
--- a/.github/workflows/go.yaml
+++ b/.github/workflows/go.yaml
@@ -1,17 +1,17 @@
 name: Go
 on:
   push:
-    paths-ignore:
-      - "docs/**"
-      - "webui/**"
-      - "design/**"
+#    paths-ignore:
+#      - "docs/**"
+#      - "webui/**"
+#      - "design/**"
     branches:
       - master
   pull_request:
-    paths-ignore:
-      - "docs/**"
-      - "webui/**"
-      - "design/**"
+#    paths-ignore:
+#      - "docs/**"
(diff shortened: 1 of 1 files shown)


guy-har (approved):

.github/workflows/go.yaml
```diff
@@ -1,17 +1,17 @@
 name: Go
 on:
   push:
-    paths-ignore:
-      - "docs/**"
-      - "webui/**"
-      - "design/**"
+#    paths-ignore:
```
itaiad200, line 4: What are mandatory tests? Do all 3 contain these?

arielshaqed, line 4: No code in comments, please :-)

arielshaqed (changes requested):
I pick nits.

.github/workflows/go.yaml
ozkatz: our go tests are required to pass before merging to master
paths-ignore means that if I only changed files under `docs/` or `webui/`, the go tests will not run.
This is a deadlock - I'll never be able to merge my changes because mandatory tests are conditional.

arielshaqed (approved):

PR #1795, opened 2021-04-20
bug fix: login check and auth middleware check required security providers

### LINKED ISSUE #1794: setup_lakefs request checks credentials
Setup lakefs API checks user credentials. This causes an issue when the user already has an invalid cookie.
Expected not to check the user credentials in order to succeed in setup a new lakeFS instance and/or clear user credentials when going into `setup` page.

nopcoder:
Close #1794 

1. bug fix: login handler fail on auth check
2. auth middleware will match required security requirements based on swagger definition per route.

### DIFF
--- a/pkg/api/auth_middleware.go
+++ b/pkg/api/auth_middleware.go
@@ -53,48 +53,45 @@ func AuthMiddleware(logger logging.Logger, swagger *openapi3.Swagger, authServic
 				writeError(w, http.StatusBadRequest, err)
 				return
 			}
-			if len(securityRequirements) == 0 {
-				// nothing to check
-				return
-			}
-
-			// validate jwt token from cookie
-			jwtCookie, _ := r.Cookie(JWTCookieName)
-			if jwtCookie != nil {
-				user, err = userByToken(ctx, logger, authService, jwtCookie.Value)
-				if err != nil {
-					writeError(w, http.StatusUnauthorized, err)
-					return
-				}
-			}
-
-			// validate jwt token from header
-			if user == nil {
-				token := r.Header.Get(JWTAuthorizationHeaderName)
-				if token != "" {
-					user, err = userByToken(ctx, logger, authService, token)
+			if len(securityRequirements) > 0 {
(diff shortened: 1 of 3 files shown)


pkg/api/auth_middleware.go
```diff
@@ -20,54 +23,88 @@ var (
 	ErrAuthenticationFailed    = errors.New("error authenticating request")
 )
 
-func AuthMiddleware(logger logging.Logger, authService auth.Service) func(next http.Handler) http.Handler {
+// extractSecurityRequirements using swagger return an array of security requirements set for the request
```
arielshaqed, line 26:
```suggestion
// extractSecurityRequirements using Swagger returns an array of security requirements set for the request.
```
or
```suggestion
// extractSecurityRequirements uses Swagger to return an array of security requirements set for the request.
```

pkg/api/auth_middleware.go
```diff
@@ -20,54 +23,88 @@ var (
...
-			jwtCookie, _ := r.Cookie(JWTCookieName)
-			if jwtCookie != nil {
-				user, err = userByToken(ctx, logger, authService, jwtCookie.Value)
-				if err != nil {
-					writeError(w, http.StatusUnauthorized, err)
-					return
-				}
+			securityRequirements, err := extractSecurityRequirements(router, r)
+			if err != nil {
+				writeError(w, http.StatusBadRequest, err)
```
arielshaqed, line 48: If this goes to users, is the return from `router.FindRoute` safe to show users (and friendly)?

arielshaqed (approved):
:100: :+1:

pkg/api/auth_middleware.go
nopcoder: 1. The possibilities: "no matching operation was found" or "method not allowed"
2. At this point this call should not fail - The first middleware that responsible to verify the request does the route lookup first. We have 3 places and this is the second one where we are required to do the same lookup again to get the relevant information related to this call.

PR #1796, opened 2021-04-21
Add log4j.properties for some debug logging of examples

arielshaqed:
Question for reviewer: should all fields be commented-out, keeping the file only to show what is possible and how to do it?

### DIFF
--- a/clients/spark/examples/src/main/resources/log4j.properties
+++ b/clients/spark/examples/src/main/resources/log4j.properties
@@ -1,0 +1,15 @@
+# Define the root logger with appender file
+log = /tmp/log4j
+log4j.rootLogger = console, FILE
+
+# Define the file appender
+log4j.appender.FILE=org.apache.log4j.FileAppender
+log4j.appender.FILE.File=${log}/log.out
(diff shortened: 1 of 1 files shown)


clients/spark/examples/src/main/resources/log4j.properties
```diff
@@ -0,0 +1,15 @@
+# Define the root logger with appender file
+log = /tmp/log4j
```
johnnyaug, line 2: change to /var/log

johnnyaug (approved):

arielshaqed:
Thanks!

Closing this: having this would include (the _same_) configuration of log4j in _all_ example Spark clients, as part of the assembly.  That would make configuring logging (_inside *Spark*_, which also configures log4j) *very* interesting.

Best simply to avoid doing this.  Will instead put up a Gist with instructions how to log during a debugging section.

arielshaqed:
Find it [as a gist](https://gist.github.com/arielshaqed/03ee3997f438ccb80797cedb3c0333c8).

PR #1805, opened 2021-04-22
Entry's identity to not contain physical address path

### LINKED ISSUE #1762: Diff outputs objects with no changes
Steps to reproduce:

1. To a lakeFS repository, upload a file to the `main` branch, and commit.
2. Create a new branch `br1` from the main branch.
3. Upload the same file with the same name to the new branch, and commit.
4. Run `lakectl diff lakefs://repo@br1 lakefs://repo@master`.

Expected: no results.
Actual: file appears as modified.

Related Slack question: https://lakefs.slack.com/archives/C016726JLJW/p1618528555058400?thread_ts=1618528530.058300&cid=C016726JLJW

itaiad200:
closes #1762

### DIFF
--- a/nessie/identity_test.go
+++ b/nessie/identity_test.go
@@ -1,0 +1,52 @@
+package nessie
+
+import (
+	"testing"
+
+	"github.com/treeverse/lakefs/pkg/api"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestIdentity(t *testing.T) {
+	for _, direct := range testDirectDataAccess {
+		name := "indirect"
+		if direct {
+			name = "direct"
+		}
+		t.Run(name, func(t *testing.T) {
+			ctx, _, repo := setupTest(t)
+			objPath := "1.txt"
+			branch1 := "feature-1"
+			branch2 := "feature-2"
+
+			_, err := client.CreateBranchWithResponse(ctx, repo, api.CreateBranchJSONRequestBody{
+				Name:   branch1,
+				Source: mainBranch,
+			})
+			require.NoError(t, err, "failed creating branch1")
+
+			checksum, objContent, err := uploadFileRandomDataAndReport(ctx, repo, branch1, objPath, direct)
+			require.NoError(t, err, "failed uploading file")
+			commitResp, err := client.CommitWithResponse(ctx, repo, branch1, api.CommitJSONRequestBody{
+				Message: "commit on branch1",
+			})
+			require.NoError(t, err, "failed to commit changes")
+			require.NoErrorf(t, verifyResponse(commitResp.HTTPResponse, commitResp.Body),
+				"failed to commit changes repo %s branch %s", repo, mainBranch)
+
+			// upload the same content again to a different branch
+			_, err = client.CreateBranchWithResponse(ctx, repo, api.CreateBranchJSONRequestBody{
+				Name:   branch2,
+				Source: mainBranch,
+			})
+			require.NoError(t, err, "failed creating branch2")
+			checksumNew, err := uploadFileAndReport(ctx, repo, mainBranch, objPath, objContent, direct)
+			require.Equal(t, checksum, checksumNew, "Same file uploaded to committed branch, expected no checksum difference")
+
+			diff, err := client.DiffRefsWithResponse(ctx, repo, branch1, branch2, &api.DiffRefsParams{})
+			require.NoError(t, err, "Diff refs failed")
+			require.Empty(t, diff.JSON200.Results, "Expected no diff files")
+		})
+	}
+}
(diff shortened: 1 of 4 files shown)


nessie/identity_test.go
```diff
@@ -0,0 +1,56 @@
+package nessie
+
+import (
+	"testing"
+
+	"github.com/treeverse/lakefs/pkg/api"
+
```
johnnyaug, line 7: 😠

nessie/identity_test.go
```diff
@@ -0,0 +1,56 @@
...
+			require.NoError(t, err, "failed creating branch2")
+			checksumNew, err := uploadFileAndReport(ctx, repo, branch2, objPath, objContent, direct)
+			require.Equal(t, checksum, checksumNew, "Same file uploaded to committed branch, expected no checksum difference")
+			commitResp, err = client.CommitWithResponse(ctx, repo, branch2, api.CommitJSONRequestBody{
+				Message: "commit on branch2",
+			})
+			require.NoError(t, err, "failed to commit changes")
+
+			diff, err := client.DiffRefsWithResponse(ctx, repo, branch1, branch2, &api.DiffRefsParams{})
+			require.NoError(t, err, "Diff refs failed")
```
johnnyaug, line 52: maybe also merge and verify that it succeeds?

johnnyaug (approved):
Nice. added suggestion for test and complaint about linebreak

arielshaqed (commented):
This is a very desirable change!  But I also want to think about it for a few more hours, especially in conjunction with direct upload ("thick" clients).

The issue here is that it makes the identity depend on the object ETag.  *That* can fail in multiple ways, e.g.:
* A client can mistakenly upload incorrect ETags (e.g. Go makes uploading a zero ETag really easy).
* A client can intentionally upload an incorrect ETag.  That -- even without dedupe, only merge -- provides essentially a  hash reversal service!  Depending on other moving parts in the system, I can bypass IAM and discover whether another file with the same ETag exists (answering questions such as "am I getting fired or a raise?") or decode an ETag for a file that I cannot access.

I do understand that some of these capabilities have to exist in a system such as we are building,  I would however like to think about it in a wider forum.

arielshaqed:
> This is a very desirable change! But I also want to think about it for a few more hours, especially in conjunction with direct upload ("thick" clients).
> 
> The issue here is that it makes the identity depend on the object ETag. _That_ can fail in multiple ways, e.g.:
> 
>     * A client can mistakenly upload incorrect ETags (e.g. Go makes uploading a zero ETag really easy).
> 
>     * A client can intentionally upload an incorrect ETag.  That -- even without dedupe, only merge -- provides essentially a  hash reversal service!  Depending on other moving parts in the system, I can bypass IAM and discover whether another file with the same ETag exists (answering questions such as "am I getting fired or a raise?") or decode an ETag for a file that I cannot access.
> 
> 
> I do understand that some of these capabilities have to exist in a system such as we are building, I would however like to think about it in a wider forum.

Thinking about it some more, I do see that these problem are inherent in the requirements, so essentially exists in _all_ implementations.  As such, I withdraw any objections to this PR.

Sorry for delaying (the inevitable).

nessie/identity_test.go
itaiad200: Done.

itaiad200: Done.

nessie/identity_test.go
itaiad200: Adding merge check, but it should fail since there are no changes between the 2 branches ;)

itaiad200:
> > This is a very desirable change! But I also want to think about it for a few more hours, especially in conjunction with direct upload ("thick" clients).
> > The issue here is that it makes the identity depend on the object ETag. _That_ can fail in multiple ways, e.g.:
> > ```
> > * A client can mistakenly upload incorrect ETags (e.g. Go makes uploading a zero ETag really easy).
> > 
> > * A client can intentionally upload an incorrect ETag.  That -- even without dedupe, only merge -- provides essentially a  hash reversal service!  Depending on other moving parts in the system, I can bypass IAM and discover whether another file with the same ETag exists (answering questions such as "am I getting fired or a raise?") or decode an ETag for a file that I cannot access.
> > ```
> > 
> > 
> > I do understand that some of these capabilities have to exist in a system such as we are building, I would however like to think about it in a wider forum.
> 
> Thinking about it some more, I do see that these problem are inherent in the requirements, so essentially exists in _all_ implementations. As such, I withdraw any objections to this PR.
> 
> Sorry for delaying (the inevitable).

Thanks Ariel, overall I agree that we didn't look at our physical address hashes from an IAM perspective. I still struggle to find an attack surface that allows you to read a file that you don't have access to, even if you acquire its hash.
However, as you said, this PR didn't expose a new vulnerability.

PR #1820, opened 2021-04-26
Add some structure to "using with", bump the first few pages into it

arielshaqed:


arielshaqed:
The "using with.,.." page now looks like this (but a bit nicer, because it gets more CSS).
[Using lakeFS with.zip](https://github.com/treeverse/lakeFS/files/6382309/Using.lakeFS.with.lakeFS.zip)

arielshaqed:
With this change:
- The "using with" page gets categorized.
- When creating a new "using..." page you *must* add one of these `tags` headers: `using/client-apps`, `using/sdks`, `using/saas-apps`.
- If you don't add the tag, it doesn't go anywhere on the page.
- Pages are still ordered by `nav_order`.
- The sidebar remain uncategorized, and sorted by `nav_order`.

arielshaqed:
Part of fixing #1822 .

YaelRiv:
Looks great! Can we arrange the side bar according to the new listing (e.g. start with Spark, Hive, Presto)

ozkatz (approved):
From what I gather, this doesn't break the URLs so exsiting links won't break. If that is indeed the case, happy to approve! :)

ozkatz:
@arielshaqed Should we resolve the conflicts and merge? or was this superseeded by the recent structural changes to the docs?

arielshaqed:
Going another (and *better*) way; thanks, @YaelRiv + @johnnyaug!

PR #1823, opened 2021-04-27
added support for put-if-absent operations

ozkatz:


clients/python/lakefs_client/api/objects_api.py
```diff
@@ -935,6 +935,7 @@ def __upload_object(
 
             Keyword Args:
                 storage_class (str): [optional]
+                if_none_match (str): Currently supports only \"*\" to allow uploading an object only if one doesn't exist yet.. [optional]
```
nopcoder, line 938: note that we can avoid the '..' at the end if the description value will not end with '.'

api/swagger.yml
```diff
@@ -2383,6 +2389,14 @@ paths:
...
           schema:
             type: string
+        - in: header
+          name: If-None-Match
+          description: Currently supports only "*" to allow uploading an object only if one doesn't exist yet.
+          example: "*"
+          required: false
+          schema:
+            type: string
+            pattern: '\*'  # Currently, only "*" is supported
```
nopcoder, line 2399: I think this pattern will match anything with * - we would like to limit the pattern to '^*$'

docs/assets/js/swagger.yml
```diff
@@ -70,6 +70,12 @@ components:
         application/json:
           schema:
             $ref: "#/components/schemas/Error"
+    PreconditionFailed:
```
nopcoder, line 73: note these are the docs

pkg/graveler/graveler.go
```diff
@@ -55,6 +55,14 @@ type RangeInfo struct {
 	Address string
 }
 
+type WriteCondition struct {
+	ifAbsent bool
+}
```
nopcoder, line 60: The member `ifAbsent' is private and we like to control the value - I see that the only way to enable the value to true is using the `IfAbsent` function call. But it means that if I have a boolean which I would like to use when calling `Set`.  It will require an optional call to IfAbsent to get the right value for the call.

pkg/graveler/graveler.go
```diff
@@ -871,13 +879,32 @@ func (g *Graveler) Get(ctx context.Context, repositoryID RepositoryID, ref Ref,
...
+		allowOverwrite := true
+		for _, cond := range writeConditions {
+			if cond.ifAbsent {
+				allowOverwrite = false
+			}
+		}
+		if !allowOverwrite {
+			// ensure the given key doesn't exist in the underlying commit first
+			// Since we're being protected by the branch locker, we're guaranteed the commit
+			// won't change before we finish the operation
```
nopcoder, line 897: Ensure or since

pkg/graveler/graveler.go
```diff
@@ -871,13 +879,32 @@ func (g *Graveler) Get(ctx context.Context, repositoryID RepositoryID, ref Ref,
...
+			_, err := g.Get(ctx, repositoryID, Ref(branch.CommitID), key)
+			if err == nil {
+				// we got a key here already!
+				return nil, ErrPreconditionFailed
+			} else if !errors.Is(err, ErrNotFound) {
```
nopcoder, line 902:
```suggestion
			} 
			if !errors.Is(err, ErrNotFound) {
```

pkg/graveler/graveler.go
```diff
@@ -871,13 +879,32 @@ func (g *Graveler) Get(ctx context.Context, repositoryID RepositoryID, ref Ref,
...
 			return nil, err
 		}
-		err = g.StagingManager.Set(ctx, branch.StagingToken, key, &value)
+		allowOverwrite := true
+		for _, cond := range writeConditions {
+			if cond.ifAbsent {
+				allowOverwrite = false
+			}
+		}
+		if !allowOverwrite {
```
nopcoder, line 894: This is optimization as the ref manager will be passed override=false and it should fail in this case.

pkg/api/controller.go
```diff
@@ -1686,6 +1686,28 @@ func (c *Controller) UploadObject(w http.ResponseWriter, r *http.Request, reposi
 		return
 	}
 
+	// before writing body, ensure preconditions - this means we essentially check for object existence twice:
+	// once before uploading the body to save resources and time,
+	//	and then graveler will check again when passed a WriteCondition.
+	allowOverwrite := true
```
nopcoder, line 1692: 1. Swagger validator should catch invalid value for `IfNoneMatch` - part of the JSON schema validation (if not ignore this comment)
2. Why do we need to write code at this level to verify the existence and not just pass the boolean value to the underlying catalog? if WriteCondition members were public this will be easier to pass in this case.

pkg/graveler/staging/manager_test.go
```diff
@@ -25,7 +25,7 @@ func TestSetGet(t *testing.T) {
 		t.Fatalf("error different than expected. expected=%v, got=%v", graveler.ErrNotFound, err)
 	}
 	value := newTestValue("identity1", "value1")
-	err = s.Set(ctx, "t1", []byte("a/b/c/"), value)
+	err = s.Set(ctx, "t1", []byte("a/b/c/"), value, true)
```
johnnyaug, line 28: Can we add a case to this test with overwrite=false?

johnnyaug (changes requested):
Nice!

pkg/api/controller.go
ozkatz: The idea is to test this specific condition outside of Graveler - why wait for the user to upload the body only to fail at CreateEntry if I already know this write is going to fail?

pkg/graveler/staging/manager_test.go
ozkatz: Done

pkg/api/controller_test.go
```diff
@@ -690,6 +690,104 @@ func uploadObjectHelper(t testing.TB, ctx context.Context, clt api.ClientWithRes
...
+			t.Fatalf("got 500 while uploading: %v", b.JSONDefault)
+		}
+		if b.StatusCode() != 201 {
+			t.Fatalf("expected 201 for UploadObject, got %d", b.StatusCode())
+		}
+	})
+
+	t.Run("overwrite", func(t *testing.T) {
+		// write
+		contentType, buf := writeMultipart("content", "bar", "hello world!")
```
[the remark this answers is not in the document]
ozkatz: Done

nopcoder (approved):

johnnyaug (approved):

api/swagger.yml
```diff
@@ -2383,6 +2389,14 @@ paths:
...
           schema:
             type: string
+        - in: header
+          name: If-None-Match
+          description: Currently supports only "*" to allow uploading an object only if one doesn't exist yet
+          example: "*"
+          required: false
+          schema:
+            type: string
+            pattern: '^\*$'  # Currently, only "*" is supported
```
arielshaqed, line 2399: @nopcoder I hope our generated clients don't validate.  Otherwise when we add other values, an old client library won't work with an application that wants to use a non-`*` value.

pkg/api/controller.go
```diff
@@ -1714,7 +1736,12 @@ func (c *Controller) UploadObject(w http.ResponseWriter, r *http.Request, reposi
 		Size:            blob.Size,
 		Checksum:        blob.Checksum,
 	}
-	err = c.Catalog.CreateEntry(ctx, repo.Name, branch, entry)
+
+	err = c.Catalog.CreateEntry(ctx, repo.Name, branch, entry, graveler.IfAbsent(!allowOverwrite))
```
arielshaqed, line 1740: Cool, double-checked locking actually works here!  (The previous test is only an optimization, but still nice to know.)

pkg/catalog/catalog.go
```diff
@@ -671,7 +671,7 @@ func (c *Catalog) CreateEntry(ctx context.Context, repository string, branch str
 	if err != nil {
 		return err
 	}
-	return c.Store.Set(ctx, repositoryID, branchID, key, *value)
+	return c.Store.Set(ctx, repositoryID, branchID, key, *value, writeConditions...)
```
arielshaqed, line 674: Style nit: no real reason to have varargs here, this API only ever gets generated slices (here on the server side).  varargs now means we cannot varargs later, when we might really want it.

arielshaqed (commented):
Sorry for late review.  Nice code, and a _great_ indicator of where we can go with this!

We might want to tighten conditions on this: a write-only user can use conditional uploads to determine whether a file exists.

PR #1824, opened 2021-04-27
add uploadIfAbsent to spark client

johnnyaug:


clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -80,4 +81,18 @@ class ApiClient(apiUrl: String, accessKey: String, secretKey: String) {
         throw new RuntimeException(s"expected string commit_id in ${resp.body}")
     }
   }
+
+  def uploadIfAbsent(repoName: String, branch: String, path: String, stream: InputStream, byteSize: Long): Unit = {
+    val uploadObjectURI = URI.create("%s/repositories/%s/branches/%s/objects?path=%s".format(apiUrl, repoName, branch, path)).normalize()
```
ozkatz, line 86: I don't think `URI.create` does any escaping.
Since we don't reject unknown query strings, if the path contains a `&`, we'll silently upload to an incorrect path.

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -80,4 +81,18 @@ class ApiClient(apiUrl: String, accessKey: String, secretKey: String) {
...
+    val resp = Http(uploadObjectURI.toString)
+      .auth(accessKey, secretKey)
+      .header("If-None-Match", "*")
+      .postMulti(new MultiPart("content", path, "application/octet-stream", stream, byteSize, _ => {})).asString
+    if (resp.code == HttpStatus.SC_PRECONDITION_FAILED) {
+      throw new AlreadyExistsException(s"put if absent failed: file already exists")
```
ozkatz, line 92: Might want to encode the path into the error here, for a useful stacktrace.

ozkatz (changes requested):
Nice - a couple of small requests

ozkatz (approved):
nice!

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -80,4 +81,18 @@ class ApiClient(apiUrl: String, accessKey: String, secretKey: String) {
         throw new RuntimeException(s"expected string commit_id in ${resp.body}")
     }
   }
+
+  def uploadIfAbsent(repoName: String, branch: String, path: String, stream: InputStream, byteSize: Long): Unit = {
+    val uploadObjectURI = URI.create("%s/repositories/%s/branches/%s/objects?path=%s".format(apiUrl, repoName, branch, URLEncoder.encode(path, "UTF-8"))).normalize()
```
nopcoder, line 86: you need to encode each value and not the total result

nopcoder (changes requested):

johnnyaug:
Not needed anymore

PR #1838, opened 2021-04-29
Allow Metastore copy from glue to hive

guy-har:
This is Currently a draft PR
Still has some TODOs 
The test file needs to be changed it. is currently the same Test that was in the previous version with relevant changes to test this version


cmd/lakectl/cmd/metastore.go
```diff
@@ -249,13 +243,15 @@ func init() {
...
+	_ = metastoreDiffCmd.Flags().String("to-client-type", "", "metastore type [hive, glue]")
 	_ = metastoreDiffCmd.Flags().String("to-schema", "", "destination schema name ")
 	_ = metastoreDiffCmd.Flags().String("to-table", "", "destination table name [default is from-table]")
 
 	metastoreCmd.AddCommand(metastoreCopyAllCmd)
-
+	_ = metastoreCopyAllCmd.Flags().String("from-client-type", "", "metastore type [hive, glue]")
 	_ = metastoreCopyAllCmd.Flags().String("from-address", "", "source metastore address")
-	_ = metastoreCopyAllCmd.MarkFlagRequired("from-address")
+	//_ = metastoreCopyAllCmd.MarkFlagRequired("from-address")
```
nopcoder, line 253: remove this one too

pkg/metastore/column_collection.go
```diff
@@ -0,0 +1,61 @@
...
+
+func compareColumns(columnA, columnB *FieldSchema) CompareResult {
+	nameA, nameB := columnA.Name, columnB.Name
+	if nameA < nameB {
+		return ItemLess
+	}
+	if nameA > nameB {
+		return ItemGreater
+	}
+	if ColumnEqual(columnA, columnB) {
```
nopcoder, line 47: can compare just 
```go
column.Type == otherColumn.Type && column.Comment == otherColumn.Comment
```

pkg/metastore/glue/converter.go
```diff
@@ -0,0 +1,269 @@
...
+		Parameters:             aws.StringMap(sd.Parameters),
+		SchemaReference:        schemaRef,
+		SerdeInfo:              serDeLocalToGlue(sd.SerdeInfo),
+		SkewedInfo:             skewedLocalToGlue(sd.SkewedInfo),
+		SortColumns:            sortColumnsLocalToGlue(sd.SortCols),
+		StoredAsSubDirectories: sd.StoredAsSubDirectories,
+	}
+}
+
+func SDGlueToLocal(sd *glue.StorageDescriptor) *metastore.StorageDescriptor {
```
nopcoder, line 246: do we need to address `SchemaReference` also in this func as we do in SDLocalToGlue?

pkg/metastore/hive/metastore_client.go
```diff
@@ -0,0 +1,221 @@
...
+import (
+	"context"
+	"crypto/tls"
+	"errors"
+	"strings"
+
+	"github.com/apache/thrift/lib/go/thrift"
+	"github.com/treeverse/lakefs/pkg/metastore/hive/gen-go/hive_metastore"
+
+	"github.com/treeverse/lakefs/pkg/metastore"
```
nopcoder, line 12: space

pkg/metastore/hive/metastore_client.go
```diff
@@ -0,0 +1,221 @@
...
+	AlterPartitions(ctx context.Context, dbName string, tableName string, newPartitions []*hive_metastore.Partition) (err error)
+	AlterPartition(ctx context.Context, dbName string, tableName string, values *hive_metastore.Partition) (err error)
+	AddPartition(ctx context.Context, newPartition *hive_metastore.Partition) (r *hive_metastore.Partition, err error)
+	DropPartition(ctx context.Context, dbName string, tableName string, values []string, deleteData bool) (r bool, err error)
+	GetDatabase(ctx context.Context, name string) (r *hive_metastore.Database, err error)
+	GetDatabases(ctx context.Context, pattern string) (r []string, err error)
+	GetAllDatabases(ctx context.Context) (r []string, err error)
+	CreateDatabase(ctx context.Context, database *hive_metastore.Database) (err error)
+	GetTables(ctx context.Context, dbName string, pattern string) (r []string, err error)
+}
```
nopcoder, line 31: remove the returned named parameters from the interface

pkg/metastore/ms_client.go
```diff
@@ -0,0 +1,275 @@
...
+
+func CopyOrMergeAll(ctx context.Context, fromClient, toClient Client, schemaFilter, tableFilter, toBranch string) error {
+	databases, err := fromClient.GetDatabases(ctx, schemaFilter)
+	if err != nil {
+		return err
+	}
+	for _, database := range databases {
+		fromDBName := database.Name
+		toDBName := toClient.NormalizeDBName(database.Name)
+		database.Name = toClient.NormalizeDBName(toDBName)
```
nopcoder, line 99: we are passing `database.Name` to normalize twice?

pkg/metastore/ms_client.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+func CopyOrMerge(ctx context.Context, fromClient Client, toClient Client, fromDB, fromTable, toDB, toTable, toBranch, serde string, partition []string) error {
+	transformLocation := func(location string) (string, error) {
+		return ReplaceBranchName(location, toBranch)
+	}
+	return copyOrMerge(ctx, fromClient, toClient, fromDB, fromTable, toDB, toTable, serde, partition, transformLocation)
+}
+
+func copyOrMerge(ctx context.Context, fromClient, toClient Client, fromDB, fromTable, toDB, toTable, serde string, partition []string, transformLocation func(location string) (string, error)) error {
```
nopcoder, line 44: suggest using a different name as we already have CopyOrMerge

pkg/metastore/ms_client.go
```diff
@@ -0,0 +1,275 @@
...
+	}
+	partitions, err := fromClient.GetPartitions(ctx, fromDB, fromTable)
+	if err != nil {
+		return err
+	}
+	partitionCollection := NewPartitionCollection(partitions)
+	return Merge(ctx, table, partitionCollection, toDB, toTable, serde, toClient, transformLocation)
+}
+
+func CopyOrMergeFromValues(ctx context.Context, fromClient Client, fTable *Table, toClient Client, fromDB, fromTable, toDB, toTable, serde string, transformLocation func(location string) (string, error)) error {
```
nopcoder, line 75: general note - passing a func for transformation which we do not use for anything else to configure the behavior makes the implementation hard to read.
looks like the operation should hold a state/struct that would have to hold the relevant information for the copy operation it would remove the need for this function.

pkg/metastore/ms_client.go
```diff
@@ -0,0 +1,275 @@
...
+	hasTable, err := toClient.HasTable(ctx, toDB, toTable)
+	if err != nil {
+		return err
+	}
+	partitions, err := fromClient.GetPartitions(ctx, fromDB, fromTable)
+	if !hasTable {
+		if err != nil {
+			return err
+		}
+		return Copy(ctx, fTable, partitions, toDB, toTable, serde, toClient, transformLocation)
```
nopcoder, line 85: This is strange to see that we call an operation and test the error after we verify another condition and later use the return value.
Usually when the call returns an err we should not use the return value. tldr looks like you need to check errors first or ignore specific errors.

pkg/metastore/ms_client.go
```diff
@@ -0,0 +1,275 @@
...
+		return err
+	}
+	p2, err := toClient.GetPartition(ctx, toDB, toTable, partition)
+	if err != nil {
+		return err
+	}
+	err = p1.Update(toDB, toTable, serde, transformLocation)
+	if err != nil {
+		return err
+	}
```
nopcoder, line 215: strange that copy operation perform update call on the source

pkg/metastore/partition_collection.go
```diff
@@ -0,0 +1,59 @@
...
+
+func (p *PartitionCollection) CompareWith(i int, v interface{}, j int) CompareResult {
+	if otherIter, ok := v.(*PartitionCollection); ok {
+		return comparePartitions(p.partitions[i], otherIter.partitions[j])
+	}
+	err := fmt.Errorf("%w *PartitionCollection, got %T", ErrExpectedType, v)
+	panic(err)
+}
+
+func partitionValueEqual(partitionA, partitionB *Partition) bool {
```
nopcoder, line 43: when this helper function is used inside `comparePartitions` it looks ok, but it is also used in other functions where we do not compare the partition name and I think it is missing there.

nopcoder (approved):
wrote some comments and approved as I don't want to hold this deliverable.
we can go over the details later...

pkg/metastore/hive/converter.go
```diff
@@ -0,0 +1,251 @@
...
+func PartitionsHiveToLocal(gluePartitions []*hive_metastore.Partition) []*metastore.Partition {
+	partitions := make([]*metastore.Partition, len(gluePartitions))
+	for i, partition := range gluePartitions {
+		partitions[i] = PartitionHiveToLocal(partition)
+	}
+	return partitions
+}
+
+func DatabaseLocalToHive(db *metastore.Database) *hive_metastore.Database {
+	path := "file:/user/hive/warehouse/"               // TODO(Guys): configure
```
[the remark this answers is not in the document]
guy-har: 1. moving it to be configurable
2. weird... but it is. 
Thanks

pkg/metastore/ms_client.go
guy-har: It's strange to see because it's a bug.
missed it when refactoring, thanks

pkg/metastore/ms_client.go
guy-har: @nopcoder 
It updates the partition returned from the source and saves it to the destination.
Would changing the name of the function make it more clear?
Or maybe returning a new partition instead of changing it?

cmd/lakectl/cmd/config/config.go
```diff
@@ -90,3 +105,10 @@ func (c *Config) GetMetastoreGlueCatalogID() string {
 func (c *Config) GetMetastoreType() string {
 	return c.Metastore.Type
 }
+
+func (c *Config) GetHiveDBLocationURI() string {
+	return c.Metastore.Hive.DBLocationtURI
+}
+func (c *Config) GetGlueDBLocationURI() string {
+	return c.Metastore.Glue.DBLocationtURI
+}
```
nopcoder, line 114: add space between two funcs

pkg/metastore/glue/metastore_client.go
```diff
@@ -16,25 +16,31 @@ import (
...
+func NewMSClient(cfg *aws.Config, catalogID, baselLocationURI string) (metastore.Client, error) {
 	sess := session.Must(session.NewSession(cfg))
 	sess.ClientConfig("glue")
 	gl := glue.New(sess)
 	return &MSClient{
-		client:    gl,
-		catalogID: catalogID,
+		client:          gl,
+		catalogID:       catalogID,
+		baseLocationURI: baselLocationURI,
```
nopcoder, line 39: I think this is the point where we need to trim right '/' or make sure we have '/' on the right (same for the other client).
Check GetDBLocation - in one case we trim and the other we don't. if we know the state of the member - we don't need to assume on each get

nopcoder (approved):
Minor comment

Issue #1840, opened 2021-04-30
Diff outputs objects for files with no changes (by copy)


xubofei1983:
Steps to reproduce:  
1. create a repo with a folder and some files in folder.  us aws s3 sync to local.
~/temp   aws s3  sync s3://test/master/folder folder                          
download: s3://test/master/folder/GCA_010094145.1 to folder/GCA_010094145.1
download: s3://test/master/folder/README.md to folder/README.md

~/temp   aws s3 sync folder s3://test/master/folder    (no files uploaded as expected.)

2. rename folder to folder_cp and copy same content back to folder. ( To simulate generation of the same data content.)
~/temp   mv folder folder_cp                                                       
~/temp   cp -r folder_cp folder 

3. try aws s3 sync, it now triggers reupload.
~/temp   aws s3 sync folder s3://test/master/folder                              
upload: folder/GCA_010094145.1 to s3://test/master/folder/GCA_010094145.1
upload: folder/README.md to s3://test/master/folder/README.md

4. try lakectl diff, it shows all files changed.
~/temp lakectl diff lakefs://test@master                                    
~ modified folder/GCA_010094145.1
~ modified folder/README.md 

Expected:  No files should be reuploaded.  No files in diff.
Actual: Files uploaded and diffs outputs all files.

talSofer:
Hi @xubofei1983, 
Regarding the concern that you raised about files being uploaded with `aws s3 sync`, this logic belongs to aws s3 and therefore we support it. We reproduced this in s3 as follows: 
```bash
test % aws s3 sync s3://bucket/folder folder         
download: s3://bucket/folder/test-file.txt to folder/test-file.txt    

test % mv folder folder_cp
test % ls 
folder_cp
test % cp -r folder_cp folder

ls -lsa folder
3 May  2 10:26 test-file.txt
test % ls -lsa folder_cp 
3 May  2 10:25 test-file.txt

test % aws s3 sync folder s3://bucket/folder 
upload: folder/test-file.txt to s3://bucket/folder/test-file.txt      
```

The second issue that you brought up (files are in diff) exposed inconsistency between uncommitted diff to committed diff (uncommitted diff = staging<>branch, committed diff = branch1<>branch2), https://github.com/treeverse/lakeFS/pull/1841 fixes it. With this pr, running `~/temp lakectl diff lakefs://test@master` will show no changes. 

[...]

PR #1841, opened 2021-05-01
Handle unchanged files correctly in diff with staging

guy-har:


arielshaqed (changes requested):
Thanks!  Algorithm looks good and clearly required, and should fix a bug.

Requested changes:
1. Rename the commit and the PR to make history more readable and searchable.  That usually involves saying what the PR does rather than why you needed it.  In this case something like "Handle unchanged files correctly in diff" might make for a better title.
2. Please try to simplify the diff iterator interface.  This PR adds an _impossible value_ to `DiffType`, which is exported.  So every client has to be aware that this value will never actually be returned.  And it impedes linting for switch-case coverage (e.g. golangci lint `exhaustive`).

Given possible customer urgency it may make sense to fix only (1), dismiss this review and pull this PR, and open an issue committing to fixing (2).

guy-har:
@arielshaqed Thanks for the review
changed according to requests, PTAL

guy-har:
closes #1840

pkg/graveler/uncommitted_diff_iterator.go
```diff
@@ -28,59 +29,69 @@ func NewUncommittedDiffIterator(ctx context.Context, manager CommittedManager, l
...
+		return nil, err
 	}
-	return true, nil
+	return value, nil
 }
 
-func (d *uncommittedDiffIterator) getDiffType(val ValueRecord) (DiffType, error) {
-	existsInCommitted, err := d.valueExistsInCommitted(val)
+// getDiffType returns the diffType between value with committed
+// will return skip == true in case of no diff
```
arielshaqed, line 47:
```suggestion
// Returns skip == true in case of no diff
```

arielshaqed (approved):
Thanks!

This implements what we agreed in today's sync: We continue to ignore mtime in object differences.  And we may revisit this decision.  (Also it addresses all of my concerns regarding implementation -- thanks again for *that*!)

pkg/graveler/uncommitted_diff_iterator.go
```diff
@@ -28,59 +29,69 @@ func NewUncommittedDiffIterator(ctx context.Context, manager CommittedManager, l
 	}
 }
 
-func (d *uncommittedDiffIterator) valueExistsInCommitted(val ValueRecord) (bool, error) {
+func (d *uncommittedDiffIterator) getValueFromCommittedIfExists(val ValueRecord) (*Value, error) {
```
itaiad200, line 32: @guy-har @arielshaqed A little late to the party, but I still want to raise a little concern:
For each value that you read from the committed data, you go thru the cycle of fetching metarange, reading it, going thru and finding the matching range, fetching the range and searching it. This could happen multiple times for a single range, if some uncommitted values fell into the same range.
Instead you could list the commit, and `SeekGE` to the values. It will save you some overhead of re-handling the same metarange and ranges over and over again since the committed iterator has a state.

Having said all that, I know that we cache metaranges/ranges and I assume that most installations under a normal workload and a decent disk size, will probably read these files from cache, therefore saving 99.9% of the above overhead (i.e. reading from s3).

guy-har: Thanks, Great Input.
Don't think it should be part of this PR anyway.
Created a new another PR for it.
#1865

PR #1837, opened 2021-04-29
Add Java (and Scala...) API client

arielshaqed:
None of the Scala generators appear to manage to generate code for an OpenAPI file that uses words such as "path", so going for a Java-based solution instead.

Works with Scala!!1!  E.g. in the REPL (with appropriately-set classpath; `mvn dependency:build-classpath` helped):
```scala
scala> val c = new ApiClient
val c: io.treeverse.lakefs.clients.api.ApiClient = io.treeverse.lakefs.clients.api.ApiClient@160a5e87

scala> c.setUsername("...")
scala> c.setPassword("...")

scala> val ra = new RepositoriesApi(c)
val ra: io.treeverse.lakefs.clients.api.RepositoriesApi = io.treeverse.lakefs.clients.api.RepositoriesApi@6c540ca5

scala> c.setBasePath("http://localhost:8000/api/v1")
val res25: io.treeverse.lakefs.clients.api.ApiClient = io.treeverse.lakefs.clients.api.ApiClient@160a5e87

scala> ra.getRepository("moo")
val res28: io.treeverse.lakefs.clients.api.model.Repository =
class Repository {
    id: moo
    creationDate: 1618297711
    defaultBranch: main
    storageNamespace: s3://treeverse-ariels-test/repos/direct/moo
}
```

PR #1842, opened 2021-05-02
Use generated Java client in Spark API client

arielshaqed:
Tested by running List client on a repo.  There are _many_ keys involved:

```sh
spark-submit --master local  --class io.treeverse.clients.examples.List \
    --conf=spark.hadoop.lakefs.api.url=http://localhost:8000/api/v1 \
    --conf=spark.hadoop.lakefs.api.access_key=AKIASOMEKEY/FOR/LAKEFS/EXAMPLE \
    --conf=spark.hadoop.lakefs.api.secret_key=TOPSECRET/FOR/LAKEFS/EXAMPLEKEY \
    --conf=spark.hadoop.fs.s3a.access.key=AKIASOMEKEY/FOR/S3/EXAMPLE \
    --conf=spark.hadoop.fs.s3a.secret.key=TOPSECRET/FOR/S3/EXAMPLEKEY \
    --conf=spark.hadoop.fs.s3a.endpoint=s3.eu-central-1.amazonaws.com  \
    target/examples-301/target/scala-2.12/lakefs-spark-examples-301-assembly-0.1.0-SNAPSHOT.jar \
    moo main s3a://path/to/something
```

Based on #1837  (in review).

DO NOT MERGE before publishing the Spark client to Maven central and to S3!

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
@@ -12,7 +12,10 @@ import java.util.concurrent.Callable
 
 private object ApiClient {
   def translateS3(uri: URI): URI =
-    if (uri.getScheme == "s3") new URI("s3a", uri.getHost, uri.getPath, uri.getFragment) else uri
+    if (uri.getScheme == "s3") new URI(
+      "s3a", uri.getUserInfo, uri.getHost, uri.getPort, uri.getPath, uri.getQuery, uri.getFragment)
+    else
+      uri
 }
 
 class ApiClient(apiUrl: String, accessKey: String, secretKey: String) {
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
@@ -3,14 +3,28 @@ package io.treeverse.clients
 import com.google.common.cache.CacheBuilder
 import org.json4s._
 import org.json4s.native.JsonMethods._
-import scalaj.http.Http
+import io.treeverse.lakefs.clients.api
 
 import java.net.URI
 import java.time.Duration
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.Callable
 
+private object ApiClient {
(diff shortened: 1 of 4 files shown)


clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -3,14 +3,28 @@ package io.treeverse.clients
...
+import io.treeverse.lakefs.clients.api
 
 import java.net.URI
 import java.time.Duration
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.Callable
 
+private object ApiClient {
+  def translateS3(uri: URI): URI =
+    if (uri.getScheme == "s3") new URI("s3a", uri.getHost, uri.getPath, uri.getFragment) else uri
```
nopcoder, line 15: worried that we do not include all the URI parts (ex: getPort getQuery) for cases where the endpoint is not amazon aws.

nopcoder (approved):
except that looks good

arielshaqed:
Note: to run in e.g. `eu-central-1`, which supports only sigV4, you'll have to add to `spark-submit`:
```
--conf=spark.driver.extraJavaOptions='-Dcom.amazonaws.services.s3.enableV4' --conf=spark.executor.extraJavaOptions='-Dcom.amazonaws.services.s3.enableV4'
```

See https://stackoverflow.com/questions/24924808/how-to-specify-aws-access-key-id-and-secret-access-key-as-part-of-a-amazon-s3n-u.

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
arielshaqed: Good idea.  Changed to use the MAX-ARGS [CONSTRUCTOR](https://docs.oracle.com/javase/7/docs/api/java/net/URI.html#URI(java.lang.String,%20java.lang.String,%20java.lang.String,%20int,%20java.lang.String,%20java.lang.String,%20java.lang.String))

arielshaqed (commented):
Thanks!

PTAL...

arielshaqed:
Thanks!  Resolving conflicts (should be minor post #1837) and pulling.

PR #1863, opened 2021-05-02
hadoopfs java package

nopcoder:


nopcoder:
created the hadoopfs under <project>/clients

PR #1864, opened 2021-05-02
Feature/ingest cmd

### LINKED ISSUE #1848: Allow mapping object store paths without going through inventory
Currently there are 2 ways of "importing" data into lakeFS without actually copying it:
1. using `lakectl fs stage` or the equivilent [stageObject API endpont](https://github.com/treeverse/lakeFS/blob/bd9febe/api/swagger.yml#L2344).
2. Using `lakefs import` that utilizes the S3 inventory to read an entire bucket (and potentially only load a subset into lakeFS)

While (1) provides a reasonable solution for importing a single object, reading a directory or common prefix requires scripting that may or may not be trivial for the user. On the other hand (2) is great for loading a big (>1M objects) bucket into lakeFS - the ops overhead is substantial because:

1. it's a lakeFS command, not a lakectl one - so the machine running it needs access to the PostgreSQL database, to the underlying storage, to the lakeFS config values as well as a lakeFS binary
2. It requires a functioning inventory (and a good understanding of s3 inventory) configured and accessible to the caller
3. It requres S3 - lakeFS doesn't support this with other object stores.

We're missing a middle ground - the ability to ingest data directly from the object store into lakeFS, by using native object listing. This provides a relatively easy way to load a common prefix, a small table or a set of partitions (<1M objects) in a way that is more accessible to a data engineer.

ozkatz:
Closes #1848


cmd/lakectl/cmd/store/factory.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+var (
+	ErrNoStorageAdapter = errors.New("no storage adapter found")
+)
+
+type ObjectStoreEntry struct {
+	FullKey     string
+	RelativeKey string
+	Address     string
```
itaiad200, line 18: Please add some documentation with examples on what you expect these to contain.

nopcoder (changes requested):

cmd/lakectl/cmd/ingest.go
```diff
@@ -0,0 +1,93 @@
...
+				lakefsURI.Repository,
+				lakefsURI.Ref, &api.StageObjectParams{
+					Path: key,
+				},
+				api.StageObjectJSONRequestBody{
+					Checksum:        e.ETag,
+					PhysicalAddress: e.Address,
+					SizeBytes:       e.Size,
+				},
+			)
```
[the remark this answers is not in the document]
ozkatz: As discussed, let's open a seperate issue

cmd/lakectl/cmd/ingest.go
```diff
@@ -0,0 +1,93 @@
...
+		verbose := MustBool(cmd.Flags().GetBool("verbose"))
+		dryRun := MustBool(cmd.Flags().GetBool("dry-run"))
+		from := MustString(cmd.Flags().GetString("from"))
+		to := MustString(cmd.Flags().GetString("to"))
+		lakefsURI := MustParsePathURI("to", to)
+
+		var staged int64
+		var stagedBytes int64
+		client := getClient()
+		err := store.Walk(ctx, from, func(e store.ObjectStoreEntry) error {
```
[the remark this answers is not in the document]
ozkatz: currently possible with --verbose which will print each individual iobject being staged. Progress bars are nice for things that are done interactively, I feel the use case here is more "do this before running my job in airflow" in which case the output will appear in a log file, rendering the progress bar unreadable.. Happy to hear objections or suggestions though..

pkg/api/controller.go
```diff
@@ -1791,11 +1792,11 @@ func (c *Controller) StageObject(w http.ResponseWriter, r *http.Request, body St
 		return
 	}
 
-	blockStoreType := c.BlockAdapter.BlockstoreType()
-	if qk.StorageType.String() != blockStoreType {
-		writeError(w, http.StatusBadRequest, fmt.Sprintf("invalid storage type: %s: current block adapter is %s",
-			qk.StorageType.String(),
-			blockStoreType,
+	// see what storage type this is and whether it fits our configuration
+	uriRegex := c.BlockAdapter.GetStorageNamespaceInfo().ValidityRegex
```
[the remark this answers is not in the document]
ozkatz: Agree, but yes - another PR

cmd/lakectl/cmd/fs.go
```diff
@@ -304,6 +311,7 @@ func init() {
 	fsStageCmd.Flags().String("location", "", "fully qualified storage location (i.e. \"s3://bucket/path/to/object\")")
 	fsStageCmd.Flags().Int64("size", 0, "Object size in bytes")
 	fsStageCmd.Flags().String("checksum", "", "Object MD5 checksum as a hexadecimal string")
+	fsStageCmd.Flags().Int64("mtime", 0, "Object modified time (Unix Epoch in seconds). Defaults to current time.")
```
nopcoder, line 314:
```suggestion
	fsStageCmd.Flags().Int64("mtime", 0, "Object modified time (Unix Epoch in seconds). Defaults to current time")
```

docs/reference/import.md
```diff
@@ -28,8 +29,58 @@ Unfortunately, copying data is not always feasible for the following reasons:
...
+```shell
+export AZURE_STORAGE_ACCOUNT="storageAccountName"
+export AZURE_STORAGE_ACCESS_KEY="EXAMPLEroozoo2gaec9fooTieWah6Oshai5Sheofievohthapob0aidee5Shaekahw7loo1aishoonuuquahr3=="
+lakectl ingest \
+   --from https://storageAccountName.blob.core.windows.net/container/optional/prefix/ \
+   --to lakefs://my-repo/ingest-branch/optional/path/
+```
+
+The `lakectl ingest` command currently supports storage accounts configured through environment variables as shown above.
+
```
guy-har, line 66: We might want to note that we only support `http` and `https` (not `wasb` `abfs` `adls`)

guy-har (approved):
LGTM

docs/reference/import.md
ozkatz: Thanks, added a note about that.

guy-har (approved):

nopcoder (approved):

PR #1865, opened 2021-05-03
Improve uncommitted diff

guy-har:


pkg/graveler/uncommitted_diff_iterator.go
```diff
@@ -109,5 +110,5 @@ func (d *uncommittedDiffIterator) Err() error {
 }
 
 func (d *uncommittedDiffIterator) Close() {
```
itaiad200, line 112: Need to close the committedList as well.

itaiad200 (approved):
LGTM! Really simple and elegant solution. Approving but please fix the single comment (about the unclosed iterator)

itaiad200 (approved):
LGTM! Really simple and elegant solution. Approving but please fix the single comment (about the unclosed iterator)

pkg/graveler/uncommitted_diff_iterator.go
guy-har: Thanks

PR #1868, opened 2021-05-03
Remove generation timestamp from generated Java files

arielshaqed:
We us a version control system (git) to handle versions.  Generation timestamps only hinder that, by making every generated timestamped file change whenever the client is regenerated.


ozkatz (approved):
Thanks!

talSofer (approved):

PR #1869, opened 2021-05-03
Publish Java API client to Sonatype repository

arielshaqed:
The OpenAPI generator writes its own `pom.xml` file, and `pom.xml` files are non-composable.
Instead request that it use a *parent* `pom.xml`, configure signing and deployment there (in
an ungenerated file which we can control entirely).

To deploy, run `mvn deploy` inside `clients/java` (*not* `clients`!).


arielshaqed:
(Already published 0.1.0 with this, sorry-not-sorry)

clients/hadoopfs/pom.xml
```diff
@@ -30,7 +30,7 @@
         <dependency>
             <groupId>io.lakefs</groupId>
             <artifactId>api-client</artifactId>
-            <version>0.1.0</version>
+            <version>0.40.3-RC.0</version>
```
talSofer, line 33: This should be 0.1.0 right?

arielshaqed: Nope, but I bumped it to be the release version.  (We may revisit this decision later).

arielshaqed (commented):
Thanks!  PTAL...

clients/hadoopfs/pom.xml
```diff
@@ -62,7 +62,7 @@
         <dependency>
             <groupId>io.lakefs</groupId>
             <artifactId>api-client</artifactId>
-            <version>0.1.0</version>
+            <version>0.40.3</version>
```
talSofer, line 65: version is incorrect

clients/pom.xml
```diff
@@ -0,0 +1,84 @@
...
+    </developer>
+  </developers>
+
+  <scm>
+    <url>https://github.com/treeverse/lakefs</url>
+    <connection>scm:git:git://github.com/treeverse/lakefs.git</connection>
+    <developerConnection>scm:git:ssh://git@github.com/treeverse/lakefs.git</developerConnection>
+  </scm>
+
+  <profiles>			<!-- to deploy (signed), must run "gpg -Prelease deploy"-->
```
talSofer, line 36: I would expect the deploy command to use the profile id you defined below. i.e. "gpg -Ptreeverse-signing deploy"

talSofer (approved):

clients/pom.xml
arielshaqed: Indeed you can do that. Not sure where this profile selection is missing, so will add it when I discover.

Thanks!

arielshaqed (commented):
Thanks!
Pulling. . .

arielshaqed (commented):
Thanks!

PR #1873, opened 2021-05-03
lakefs hadoopfs initial code

nopcoder:
lakefs hadoop filesystem code scafold with github action to compile the code


.github/workflows/hadoopfs.yaml
```diff
@@ -0,0 +1,25 @@
...
+      - master
+  pull_request:
+jobs:
+  lakefs-hadoopfs:
+    name: lakeFS hadoopFS
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check-out code
+        uses: actions/checkout@v2
+      - uses: actions/setup-java@v2
```
talSofer, line 16: you can add a readable step name (Setup java 8)

clients/hadoopfs/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+    </properties>
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.8.1</version>
+                <configuration>
+                    <source>8</source>
+                    <target>8</target>
```
talSofer, line 23: It looks like we should use 
    <configuration>
        <source>1.8</source>
        <target>1.8</target>
      </configuration>

see for a reference http://tutorials.jenkov.com/maven/java-compiler.html

talSofer (approved):

PR #1884, opened 2021-05-04
Add workflow to deploy Java client to Nexus Sonatype (Maven Central)

arielshaqed:
[This gist](https://gist.github.com/sualeh/ae78dc16123899d7942bc38baba5203c) describes
how we import GPG signing secrets and OSSRH logins (not to worry, it's a single stanza in `java-api-client.yaml`, not some random GitHub action that can change each time...).  `OSSRH_GPG_SECRET_KEY` comes from the "Treeverse signing" environment -- a dev team member needs to approve each run.


.github/workflows/java-api-client.yaml
```diff
@@ -0,0 +1,46 @@
...
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: Tag/version to publish
+
+jobs:
+  pypi:
```
talSofer, line 12: Can you rename the job to comply with the java client? maybe to "maven"

clients/pom.xml
```diff
@@ -44,10 +44,19 @@
 	  <plugin>
 	    <groupId>org.apache.maven.plugins</groupId>
             <artifactId>maven-gpg-plugin</artifactId>
-	    <version>1.5</version> <!-- must path java/pom.xml -->
+	    <version>1.5</version> <!-- must match java/pom.xml -->
 	    <configuration>
 	      <keyname>A670BB5ACA82B971772B6DEFDFFC0764F5A9A357</keyname>
 	    </configuration>
```
talSofer, line 50: given that you created the github workflow to publish to maven central, is the keyname still needed in this pom? (I was not sure about the answer by just reading the docs)

.github/workflows/java-api-client.yaml
arielshaqed: Thanks!  Picked "`deploy`", because that's what it is... and should be understandable.  Preferred to avoid "maven" which is an implementation detail.

clients/pom.xml
arielshaqed: Yes.

The secrets `OSSRH_USERNAME` and `OSSRH_TOKEN` are to authenticate to OSSRH Sonatype Nexus, which is the repository for Maven Central.  But Maven Central itself _also_ requires that all artifacts be signed, using a GPG key.  That GPG secret key is inserted through secret `OSSRH_GPG_SECRET_KEY`; I inserted it into the GitHub secret environment "Treeverse signing".  This keyname is the ID of the signing key; this configuration means that (by default) we sign with our key.  (You can deploy from a developer laptop by installing and accessing this GPG secret key, of course).

One confusing thing is that  two identities are involved:
1. User on OSSRH Sonatype Nexus;
2. JAR signer.
Indeed AFAICT this is a security bug in Maven Central: AFAICT there is no binding between the two identities.  So JAR signing is required, but all JARs might as well be signed by Bugs Bunny.
JARs might as well be signed by Bugs Bunny: "What's up, doc?"

arielshaqed (commented):
Thanks!  PTAL>..

clients/pom.xml
talSofer: Thanks for the thorough explanation!

talSofer (approved):

arielshaqed:
Thanks!

PR #1885, opened 2021-05-04
Don't upload parts smaller than 8192 bytes

### LINKED ISSUE #1883: [gateway] Avoid sending chunks smaller than 8KB
When part is being uploaded to lakeFS, lakeFS uses streaming in order to upload that part to the object store. Streaming has some limitations when the connection is slow. It must sent a chunk every minute or so, in order to keep the connections alive. However chunks (except for the last one) minimum size is 8KB. Streaming will always fail when chunks smaller than that are uploaded.

Instead of sending small chunks after the timeout passed, we should optimistically wait until it reaches the min size. If the connection was closed by the block storage, we should return a meaningful error and not simply the error of the underline storage.

itaiad200:
closes #1883


guy-har (approved):
Looks Great

PR #1889, opened 2021-05-04
update client FILES

nopcoder:


arielshaqed:
They went away? Strange... 
Can you explain what this file does?

nopcoder:
@arielshaqed ignore this one - it looks like we updated the package but didn't generated the sources

PR #1890, opened 2021-05-04
update generated files to use io.lakefs package

nopcoder:


arielshaqed (approved):
This is just `make client-java`,  right?

nopcoder:
yes, but I had to do it twice - it looks like the second run of the generator remove a list of test files from this list

nopcoder:
@arielshaqed Looks like we have an issue with the code gen and the test file - running the step again, it finds the test source files and skip them `(Test files never overwrite an existing file of the same name.)` - but doesn't report them to FILES like it does with the source code.

nopcoder:
Ignoring the FILES will solve this issue for now

PR #1891, opened 2021-05-05
Unshade RocksDB in Spark clients; rename api client to match new name

arielshaqed:
Removes this error when starting spark example app "list", on local spark-submit (3.0.1):
```
Exception in thread "main" java.lang.UnsatisfiedLinkError: shadedrocksdb.Options.newOptions()J
```


johnnyaug (approved):

arielshaqed (commented):

PR #1894, opened 2021-05-05
hadoopfs assembly, handle workdir and lakefs client

nopcoder:
Tests will be addressed in a differnt PR

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -1,6 +1,10 @@
 package io.lakefs;
 
+import io.lakefs.clients.api.ApiException;
+import io.lakefs.clients.api.RepositoriesApi;
 import io.lakefs.clients.api.auth.HttpBasicAuth;
+import io.lakefs.clients.api.model.Repository;
+
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.*;
 import org.apache.hadoop.fs.permission.FsPermission;
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -162,7 +162,6 @@ public class LakeFSFileSystem extends FileSystem {
                 modificationTime = mtime * 1000; // convert to ms
             }
             Path filePath = path.makeQualified(this.uri, this.workingDirectory);
-            LOG.debug("[BABA] filePath=" + filePath);
             return new FileStatus(length, false, 0, 0, modificationTime, filePath);
         } catch (ApiException e) {
             throw new IOException("statObject", e);
@@ -267,6 +266,10 @@ public class LakeFSFileSystem extends FileSystem {
     }
 
     private static class Location {
+        private String repository;
+        private String ref;
+        private String path;
+        
         public String getRepository() {
             return repository;
         }
@@ -290,9 +293,5 @@ public class LakeFSFileSystem extends FileSystem {
         public void setPath(String path) {
             this.path = path;
         }
-
-        private String repository;
-        private String ref;
-        private String path;
     }
 }
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -1,5 +1,6 @@
 package io.lakefs;
 
+import io.lakefs.clients.api.ApiClient;
 import io.lakefs.clients.api.ApiException;
 import io.lakefs.clients.api.ObjectsApi;
 import io.lakefs.clients.api.auth.HttpBasicAuth;
@@ -14,7 +15,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.StringBufferInputStream;
 import java.net.URI;
-import java.util.Objects;
 import java.util.concurrent.TimeUnit;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -140,9 +144,29 @@ public boolean mkdirs(Path path, FsPermission fsPermission) throws IOException {
...
+            ObjectStats objectStat = objectsApi.statObject(loc.getRepository(), loc.getRef(), loc.getPath());
+            long length = 0;
+            Long sizeBytes = objectStat.getSizeBytes();
+            if (sizeBytes != null) {
+                length = sizeBytes;
+            }
+            long modificationTime = 0;
+            Long mtime = objectStat.getMtime();
+            if (mtime != null) {
+                modificationTime = mtime * 1000; // convert to ms
```
talSofer, line 162: you can use TimeUnit.tomillis here (e.g. https://www.geeksforgeeks.org/timeunit-tomillis-method-in-java-with-examples/)

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -218,4 +242,57 @@ public boolean seekToNewSource(long l) throws IOException {
...
+        public String getPath() {
+            return path;
+        }
+
+        public void setPath(String path) {
+            this.path = path;
+        }
+
+        private String repository;
+        private String ref;
```
talSofer, line 295: To implement the delete operation I'm going to need a branch. I saw that ref could be either a commit id or a branch. In the context of a file system, can't we always expect to get a branch? if so, it will clearer calling it a branch

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -218,4 +242,57 @@ public boolean seekToNewSource(long l) throws IOException {
...
+        }
+
+        URI uri = path.toUri();
+        if (uri.getScheme() != null && uri.getPath().isEmpty()) {
+            return null;
+        }
+
+        Location loc = new Location();
+        loc.setRepository(uri.getHost());
+        String s = uri.getPath().substring(1);
```
talSofer, line 258: What char are you cleaning here? can you document the uri format with an example uri? you can add the example in the method docs and it will help to understand the path parsing.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -218,4 +242,57 @@ public boolean seekToNewSource(long l) throws IOException {
...
+
+        public String getPath() {
+            return path;
+        }
+
+        public void setPath(String path) {
+            this.path = path;
+        }
+
+        private String repository;
```
talSofer, line 294: please move the members to the top of the class definition

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -140,9 +144,29 @@ public boolean mkdirs(Path path, FsPermission fsPermission) throws IOException {
...
+            if (sizeBytes != null) {
+                length = sizeBytes;
+            }
+            long modificationTime = 0;
+            Long mtime = objectStat.getMtime();
+            if (mtime != null) {
+                modificationTime = mtime * 1000; // convert to ms
+            }
+            Path filePath = path.makeQualified(this.uri, this.workingDirectory);
+            LOG.debug("[BABA] filePath=" + filePath);
```
talSofer, line 165: remove debugging string

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
nopcoder: cool

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -140,9 +146,28 @@ public boolean mkdirs(Path path, FsPermission fsPermission) throws IOException {
 
     @Override
     public FileStatus getFileStatus(Path path) throws IOException {
-        LOG.debug("$$$$$$$$$$$$$$$$$$$$$$$$$$$$ getFileStatus, path: {} $$$$$$$$$$$$$$$$$$$$$$$$$$$$", path.toString());
-        FileStatus fStatus = new FileStatus(0, false, 1, 20, 1, path);
-        return fStatus;
+        Location loc = pathToLocation(path);
```
talSofer, line 149: What do you think of adding debug logging at the beginning in case we need to track the filesystem behavior in new environments? (in a different format than I used)

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -218,4 +243,66 @@ public boolean seekToNewSource(long l) throws IOException {
             return false;
         }
     }
+
+    /**
+     * Returns Location with repository, ref and path used by lakeFS based on filesystem path
```
talSofer, line 248: nit; add . at the end of the function description.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -218,4 +243,66 @@ public boolean seekToNewSource(long l) throws IOException {
...
+     * @param path
+     * @return
```
talSofer, line 250: pls fill in the param and return value descriptions

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -218,4 +243,66 @@ public boolean seekToNewSource(long l) throws IOException {
...
+     */
+    private Location pathToLocation(Path path) {
+        if (!path.isAbsolute()) {
+            path = new Path(this.workingDirectory, path);
+        }
+
+        URI uri = path.toUri();
+        if (uri.getScheme() != null && uri.getPath().isEmpty()) {
+            return null;
```
talSofer, line 259: Can you log a debug message saying why we are returning a null location?

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -218,4 +243,66 @@ public boolean seekToNewSource(long l) throws IOException {
...
+        if (i == -1) {
+            loc.setRef(s);
+        } else {
+            loc.setRef(s.substring(0, i));
+            loc.setPath(s.substring(i+1));
+        }
+        return loc;
+    }
+
+    private static class Location {
```
talSofer, line 279: Can you give it a name that will make it clear that the location is a lakefs location? maybe LakefsAddress / LakefsLocation
what do you think?

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
nopcoder: Here I don't think that logging is required - not an error and logging everything is too much.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
nopcoder: Can rename to something else, but I think prefix Lakefs for everything in this package can be wrong.
Have other suggestions?

talSofer: yes, I hear you.. as discussed let's call it ObjectLocation for now

talSofer (approved):

Issue #1858, opened 2021-05-02
[filesystem] Create API


nopcoder:
Implement lakeFS filesystem API

PR #1895, opened 2021-05-05
Add "open" method

### LINKED ISSUE #1856: [filesystem] Open API - return stream to read with close

arielshaqed:
Untested -- need an integration test (vs. e.g. minIO and/or true S3).

Closes #1856.

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -52,28 +52,6 @@ public class LakeFSFileSystem extends FileSystem {
     private Path workingDirectory = new Path(SEPARATOR);
     private ApiClient apiClient;
 
-    private static class ParsedPath {
-	ParsedPath(String repo, String ref, String path) {
-	    this.repo = repo;
-	    this.ref = ref;
-	    this.path = path;
-	}
-	public String repo;
-	public String ref;
-	public String path;
-    }
-
-    private ParsedPath parsePath(Path path) {
-	if (!path.isAbsolute()) path = new Path(workingDirectory, path);
-
-	URI pathUri = path.toUri();
-	String[] parts = pathUri.getPath().split("/", 3);
-	if (!parts[0].equals("")) {
-	    throw new RuntimeException("path does not start with / (" + String.join(" [/] ", parts) + ")");
-	}
-	return new ParsedPath(pathUri.getAuthority(), parts[1], parts[2]);
-    }
-
     private URI translateUri(URI uri) throws java.net.URISyntaxException {
 	switch (uri.getScheme()) {
 	case "s3":
@@ -128,8 +106,8 @@ public class LakeFSFileSystem extends FileSystem {
 	    LOG.debug("$$$$$$$$$$$$$$$$$$$$$$$$$$$$ open(" + path.getName() + ") $$$$$$$$$$$$$$$$$$$$$$$$$$$$");
 
 	    ObjectsApi objects = new ObjectsApi(apiClient);
-	    ParsedPath pp = parsePath(path);
-	    ObjectStats stats = objects.statObject(pp.repo, pp.ref, pp.path);
+	    ObjectLocation objLoc = pathToObjectLocation(path);
+	    ObjectStats stats = objects.statObject(objLoc.getRepository(), objLoc.getRef(), objLoc.getPath());
 	    URI physicalUri = translateUri(new URI(stats.getPhysicalAddress()));
 
 	    Path physicalPath = new Path(physicalUri.toString());
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -114,7 +114,7 @@ public class LakeFSFileSystem extends FileSystem {
 	    FileSystem physicalFs = physicalPath.getFileSystem(conf);
 	    return physicalFs.open(physicalPath, bufSize);
 	} catch (io.lakefs.clients.api.ApiException e) {
-	    throw new RuntimeException("API exception: " + e.getResponseBody());
+	    throw new RuntimeException("lakeFS API exception", e);
 	} catch (java.net.URISyntaxException e) {
 	    throw new RuntimeException(e);
 	}
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -10,6 +10,10 @@ import org.apache.hadoop.fs.*;
 import org.apache.hadoop.fs.permission.FsPermission;
 import org.apache.hadoop.util.Progressable;
 
+import io.lakefs.clients.api.ApiClient;
+import io.lakefs.clients.api.ObjectsApi;
+import io.lakefs.clients.api.model.ObjectStats;
+
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


arielshaqed:
Manual testing: Start a Spark shell configured to run everything:
```sh
spark-shell  --packages org.apache.hadoop:hadoop-aws:2.7.7  --jars ./target/hadoopfs-0.1.0.jar --master local  --conf=spark.hadoop.fs.lakefs.impl=io.lakefs.LakeFSFileSystem  --conf=spark.driver.extraJavaOptions='-Dcom.amazonaws.services.s3.enableV4'      --conf=spark.hadoop.lakefs.api.url=http://localhost:8000/api/v1  --conf=spark.hadoop.fs.lakefs.access.key=AKIAlakeFS --conf=spark.hadoop.fs.lakefs.secret.key=lakeFShhhh --conf=spark.hadoop.fs.s3a.access.key=AKIApleaseDONTstealMYawsKEY --conf=spark.hadoop.fs.s3a.secret.key=Shhh...  --conf=spark.hadoop.fs.s3a.endpoint=s3.eu-central-1.amazonaws.com
```

Now run this in the shell:
```scala
scala> import org.apache.hadoop.fs.Path; val p = new Path("lakefs://moo/main/yes"); val fs = p.getFileSystem(sc.hadoopConfiguration);      val c = fs.open(p)
import org.apache.hadoop.fs.Path
p: org.apache.hadoop.fs.Path = lakefs://moo/main/yes
fs: org.apache.hadoop.fs.FileSystem = io.lakefs.LakeFSFileSystem@3b59d246
c: org.apache.hadoop.fs.FSDataInputStream = org.apache.hadoop.fs.FSDataInputStream@3fbf4614

scala> val buf = new Array[Byte](5555)
buf: Array[Byte] = Array(0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0,...

scala> val len = c.read(buf)
len: Int = 5

scala> new String(buf.slice(0, len))
res0: String =
"huh!
"
```

arielshaqed:
[+ @nopcoder for general layout, try to share `translateUri`, etc.]

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -37,9 +41,42 @@
...
+
+    private ParsedPath parsePath(Path path) {
+	if (!path.isAbsolute()) path = new Path(workingDirectory, path);
+
+	URI pathUri = path.toUri();
+	String[] parts = pathUri.getPath().split("/", 3);
+	if (!parts[0].equals("")) {
+	    throw new RuntimeException("path does not start with / (" + String.join(" [/] ", parts) + ")");
+	}
+	return new ParsedPath(pathUri.getAuthority(), parts[1], parts[2]);
```
nopcoder, line 68: used the getHost for this one as we don't use the port for the repository - not sure which one is the better

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -81,11 +119,23 @@ public void initialize(URI name, Configuration conf) throws IOException {
...
+	try {
+	    LOG.debug("$$$$$$$$$$$$$$$$$$$$$$$$$$$$ open(" + path.getName() + ") $$$$$$$$$$$$$$$$$$$$$$$$$$$$");
+
+	    ObjectsApi objects = new ObjectsApi(apiClient);
+	    ParsedPath pp = parsePath(path);
+	    ObjectStats stats = objects.statObject(pp.repo, pp.ref, pp.path);
+	    URI physicalUri = translateUri(new URI(stats.getPhysicalAddress()));
+
+	    Path physicalPath = new Path(physicalUri.toString());
+	    FileSystem physicalFs = physicalPath.getFileSystem(conf);
```
nopcoder, line 132: This is the part where I thought we need to initialize the underlying filesystem once in the `initialize` and use it here and just call `open`.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -81,11 +119,23 @@ public void initialize(URI name, Configuration conf) throws IOException {
...
+	    return physicalFs.open(physicalPath, bufSize);
+	} catch (io.lakefs.clients.api.ApiException e) {
+	    throw new RuntimeException("API exception: " + e.getResponseBody());
```
nopcoder, line 135: think you can pass `e` as inner exception here

talSofer (commented):
@arielshaqed I reviewed this, and currently don't have comments to add to Barak's comments

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
arielshaqed: Yeah, I was hoping to do that too.  Disappointed that this is actually surprisingly difficult to do correctly.  Now according to [`S3AFileSystem.initialize`](https://github.com/grahamar/hadoop-aws/blob/c127dfbd70e6d166d6e6517680cf46c5e15738e9/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L167) (please look at the code), the repository is guaranteed to exist on the URL.  (TBH, I don't understand _why_, but we might assume as much).  So we _could_ generate the underlying file system at that point.

Let's circle back to this point once we have a working FS?  Because what this would do is (only) speed up multiple calls to `open` (and `create`, after #1858 ) when they occur on the same FileSystem.  I.e. it speeds up the already-fast part of multi-part file read/write operations -- if it works.

Opened #1906.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
arielshaqed: Yup, forgot about exception causes (aka "Go `%w` done right"...).  Thanks!

arielshaqed (commented):
Thanks!

PTAL...

talSofer (approved):

arielshaqed:
Thanks!

Pulling...

PR #1899, opened 2021-05-06
lower hadoop common to 2.7.7

nopcoder:


arielshaqed (approved):
Thanks!

Best to do it sooner rather than later.

Important to document why we do this: The default prebuilt binaries for apache-spark all use Hadoop 2.7.  This is also what you get on EMR.  It is important to have the same version of Hadoop -- we are both _users_ of Hadoop (the underlying FileSystems) and _implementors_ of Hadoop FileSystems.

arielshaqed:
@talSofer FYI

arielshaqed:
Thanks!

PR #1909, opened 2021-05-06
allow ancestry in s3 gateway

johnnyaug:


pkg/gateway/path/resolver.go
```diff
@@ -15,12 +15,12 @@ const (
 	Separator = "/"
 
 	rePath      = "(?P<path>.*)"
-	reReference = `(?P<ref>\w[-\w]*)`
+	reReference = `(?P<ref>\w[-\w]*[~^0-9]*)`
```
arielshaqed, line 18: Not sure this is a good way to parse here.  TBH I don't understand why we need to enforce morality of references: why not just take *everything* up to the next slash and try to use that?  IIRC this was originally due to using the parsed branch name in a DB lookup, with the (misguided but common) fear that maybe Bobbie Tables can overcome DB parametrized queries and perform SQL injection.  I say: just allow all characters here, and fail when you revparse them!

pkg/gateway/path/resolver_test.go
```diff
@@ -74,6 +74,36 @@ func TestResolvePath(t *testing.T) {
 			args:    args{encodedPath: "-invalid"},
 			wantErr: true,
 		},
+		{
```
arielshaqed, line 77: FTR, the _real_ tests are in pkg/graveler/ref/rev_parse_test.go.  I see no reason to try to replicate them or the regexp here.

pkg/gateway/path/resolver.go
```diff
@@ -15,12 +15,12 @@ const (
...
 )
 
 var (
 	EncodedPathRe          = regexp.MustCompile(fmt.Sprintf("^/?%s/%s", reReference, rePath))
-	EncodedPathReferenceRe = regexp.MustCompile(fmt.Sprintf("^/?%s", reReference))
+	EncodedPathReferenceRe = regexp.MustCompile(fmt.Sprintf("^/?%s$", reReference))
```
arielshaqed, line 23: To be clear: I am asking why we cannot just write
```suggestion
	EncodedPathReferenceRe = regexp.MustCompile("^/?[^/]?*$")
```
and be done with it: RevParse will do the rest.

arielshaqed (changes requested):
Not a fan of validation here.  I claim that enforcing something before parsing it only makes life harder.  So I say we should just trust RevParse to parse revs.

pkg/gateway/path/resolver.go
arielshaqed: Cool, with the new `reReference` I understand what this does!  BTW, I am not sure that we need `$` here, I'm pretty sure it doesn't _change_ any behaviour for a legal path... but it *is* easier to understand what it does.

arielshaqed (approved):
Thanks, cool!

This should probably have a bit of manual QA.

Also maybe consult @ozkatz about the test changes.

PR #1920, opened 2021-05-09
Implement lakefsFS delete method

### LINKED ISSUE #1854: [filesystem] Delete API - metadata

talSofer:
Closes #1854

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -215,9 +215,23 @@ public class LakeFSFileSystem extends FileSystem {
     }
 
     @Override
-    public boolean delete(Path path, boolean b) throws IOException {
-        LOG.debug("$$$$$$$$$$$$$$$$$$$$$$$$$$$$ delete $$$$$$$$$$$$$$$$$$$$$$$$$$$$ ");
-        return false;
+    public boolean delete(Path path, boolean recursive) throws IOException {
+        LOG.debug("$$$$$$$$$$$$$$$$$$$$$$$$$$$$ Delete path {} - recursive {} $$$$$$$$$$$$$$$$$$$$$$$$$$$$",
+                path, recursive);
+
+        ObjectLocation loc = pathToObjectLocation(path);
+        if (loc == null) {
+            throw new FileNotFoundException(path.toString());
+        }
+
+        ObjectsApi objectsApi = new ObjectsApi(this.apiClient);
+        try {
+            objectsApi.deleteObject(loc.getRepository(), loc.getRef(), loc.getPath());
+        } catch (ApiException e) {
+            throw new IOException("deleteObject", e);
+        }
+        LOG.debug("Successfully deleted {}", path.toString());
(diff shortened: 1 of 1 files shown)


clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -143,9 +143,23 @@ public boolean rename(Path path, Path path1) throws IOException {
...
 
     @Override
-    public boolean delete(Path path, boolean b) throws IOException {
-        LOG.debug("$$$$$$$$$$$$$$$$$$$$$$$$$$$$ delete $$$$$$$$$$$$$$$$$$$$$$$$$$$$ ");
-        return false;
+    public boolean delete(Path path, boolean recursive) throws IOException {
+        LOG.debug("$$$$$$$$$$$$$$$$$$$$$$$$$$$$ Delete path {} - recursive {} $$$$$$$$$$$$$$$$$$$$$$$$$$$$",
+                path, recursive);
+
+        ObjectLocation loc = pathToObjectLocation(path);
```
arielshaqed, line 150:
```suggestion
        ObjectLocation objectLoc = pathToObjectLocation(path);
```
(now that we have `stagingLoc`...)

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -143,9 +143,23 @@ public boolean rename(Path path, Path path1) throws IOException {
...
+        if (loc == null) {
+            throw new FileNotFoundException(path.toString());
```
nopcoder, line 152: This is the case we wanted `false` in return

nopcoder (changes requested):

arielshaqed (approved):
Thanks!

Did you manage to test this (manually)?

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
arielshaqed: I'm not sure.  I think this is analogous to S3A receiving an error from AWS when deleting, https://github.com/apache/hadoop/blob/7f93349ee74da5f35276b7535781714501ab2457/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L2746.

talSofer:
> Thanks!
> 
> Did you manage to test this (manually)?

@arielshaqed, yes I tested the following scenarios: 
- deletion of non-existing file, outcome - the delete action has no impact in this case. 
- deletion of a committed file, outcome - tombstone is created for the file. 
- deletion of existing uncommitted file, outcome - file is deleted from staging.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
talSofer: I removed this null check with my latest commit. Digging into the implementation of pathToObjectLocation I found out that it can't return null values.

nopcoder (approved):

PR #1936, opened 2021-05-10
Throw FileNotFoundException from getFileStatus

arielshaqed:
S3AFileSystem also throws such an exception, e.g. at
https://github.com/apache/hadoop/blob/359c0c1d1edb5ef3901e95f09364be6bd742866e/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L3098

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -283,7 +283,7 @@ public class LakeFSFileSystem extends FileSystem {
             Path filePath = path.makeQualified(this.uri, this.workingDirectory);
             return new FileStatus(length, false, 0, 0, modificationTime, filePath);
         } catch (ApiException e) {
-            if (e.getCode() == 404) {
+            if (e.getCode() == HttpStatus.SC_NOT_FOUND) {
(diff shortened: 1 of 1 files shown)


clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -283,6 +283,9 @@ public FileStatus getFileStatus(Path path) throws IOException {
             Path filePath = path.makeQualified(this.uri, this.workingDirectory);
             return new FileStatus(length, false, 0, 0, modificationTime, filePath);
         } catch (ApiException e) {
+            if (e.getCode() == 404) {
```
nopcoder, line 286:
```suggestion
            if (e.getCode() == HttpStatus.SC_NOT_FOUND) {
```

nopcoder (approved):

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
arielshaqed: Gotta love that "`SC_`" prefix on a (scoped!) enum values.   Brings out the worst features of Java and C...

nopcoder (approved):

PR #1937, opened 2021-05-11
Move FileStatus in line with Spark expectations

arielshaqed:
- `exists` must return more than just `false`.  Implement a minimal exists that checks for file existence on lakeFS.  Note that it _will_ have to change if we ever do directories.
- `FileStatus` needs a block size entry.  Use value from underlying storage.  Also return some "reasonable"-ish _default_ block size (not used by readers, some writers or other code might use that).


arielshaqed (commented):
Thanks!

arielshaqed (commented):
Thanks!

I cleaned up some exceptions to IOException, rephrased some messages, rebased, and am pulling.

PR #1950, opened 2021-05-12
working dir set to java client

nopcoder:


arielshaqed (approved):
Should do the job I hope.

Thanks -- nice catch and fix.

Issue #1948, opened 2021-05-12
Unexpected merge conflicts


johnnyaug:
Getting merge conflicts due to wrong merge-base being computed.

Steps to reproduce:
1. Create branch from main.
2. Upload a file to this branch and commit.
3. Merge to main.
4. Upload another file to the same path and commit it.
5. Try to merge to main - get a conflict.

johnnyaug:
## Our current BFS solution can't work for finding the merge-base

Consider the following unrealistic but possible DAG:

![image](https://user-images.githubusercontent.com/3149885/118391017-e1f1a300-b63a-11eb-9975-81128fb641d1.png)

The merge-base for _x_ and _y_ is _c2_. _c1_ is also a common ancestor but it is not the "best" one since _c2_ descends from it.
However, to get to _c2_ from _x_ one would need to go 3 steps, while _c1_ can be reached by no more than 2 steps from both _x_ and _y_. Hence, our algorithm would find _c1_ before _c2_.

PR #1958, opened 2021-05-13
Fix unexpected merge conflicts bug

johnnyaug:
Solves #1948


johnnyaug:
## Our current BFS solution can't work for finding the merge-base

Consider the following unrealistic but possible DAG:

image

The merge-base for _x_ and _y_ is _c2_. _c1_ is also a common ancestor but it is not the "best" one since _c2_ descends from it.
However, to get to _c2_ from _x_ one would need to go 3 steps, while _c1_ can be reached by no more than 2 steps from both _x_ and _y_. Hence, our algorithm would find _c1_ before _c2_.

ozkatz:
Good analysis @johnnyaug!

pkg/graveler/ref/commit_walker.go
```diff
@@ -68,34 +68,38 @@ func (w *CommitWalker) Err() error {
 	return w.err
 }
 
-func FindLowestCommonAncestor(ctx context.Context, getter CommitGetter, addressProvider ident.AddressProvider, repositoryID graveler.RepositoryID, left, right graveler.CommitID) (*graveler.Commit, error) {
-	discoveredSet := make(map[string]struct{})
+func FindLowestCommonAncestor(ctx context.Context, getter CommitGetter, addressProvider ident.AddressProvider, repositoryID graveler.RepositoryID, left, right graveler.CommitID) *graveler.Commit {
 	iterLeft := NewCommitWalker(ctx, getter, repositoryID, left)
+	// find the set of all ancestors of the left commit:
+	leftAncestors := make(map[string]bool)
```
ozkatz, line 74: so we're keeping in memory a set of all of left's ancesstors? this is unbounded and potentially very large..

pkg/graveler/ref/commit_walker.go
```diff
@@ -68,34 +68,38 @@ func (w *CommitWalker) Err() error {
...
+	commonAncestors := make(map[string]*graveler.Commit)
+	for iterRight.Next() {
+		commit := iterRight.Value()
+		addr := addressProvider.ContentAddress(commit)
+		if leftAncestors[addr] {
+			commonAncestors[addr] = commit
 		}
-		if iterLeft.Value() == nil && iterRight.Value() == nil {
-			break
+	}
```
ozkatz, line 88: Just to make sure I understand the algorithm:

1. scan all left ancestors
2. scan all right ancestors, comparing to the left set to derive the common ones
3. for each common ancesstor, increment a counter for its parents
4. find and return the first arbitrarily found common ancestor that has no child node that is also a common ancestor

so time: O(N+M) where N is left ancestors and M is right ancestors
space: O(N) because we keep a set of all leftAncestors

My worries:
1. whether this will always result in a correct lowest common ancestor
2. the high cost involved with the solution.

Not claiming I have a simpler one in mind, I think it is a rather complex problem to solve efficiently - but I feel this is a problem that has already received quite a lot of attention as it is solved in Git. I would recommend either adopting their solution, or sticking with this one - with a clear explanation of why it's better for lakeFS.

ozkatz (changes requested):
I'd insist on:
1. Error checking for iterators
2. A clear explanation of why we're not doing *exactly* what Git does (or, doing exactly what Git does)

pkg/graveler/ref/commit_walker.go
johnnyaug: You understand correctly. 
This problem didn't get as much attention as you would expect. For trees yes, but not for DAGs.

Your worry (1) can be solved using documentation.

This is absolutely not documented in git, but if I understand the code correctly, they solve this using a priority-queue, where the priority function is the generation of the commit (generation is the distance from the root). I will look into it.

ozkatz: A good post I found: https://devblogs.microsoft.com/devops/supercharging-the-git-commit-graph-iii-generations/

pkg/graveler/ref/commit_priority_queue.go
```diff
@@ -0,0 +1,58 @@
+package ref
+
+import "github.com/treeverse/lakefs/pkg/graveler"
+
+// CommitsPriorityQueue implements heap.Interface such that the commit with the oldest creation date is at the root
+// of the heap.
+type CommitsPriorityQueue struct {
+	queue []*graveler.CommitRecord
+}
+
```
nopcoder, line 10: Two suggestions - it will look like less reuse but I think of separation.
1. There are two types here - so split the implementation into two
2. Type the queue without another `struct` level - `type CommitsPriorityQueue []*graveler.CommitRecord`

pkg/graveler/ref/merge_base_finder.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"container/heap"
+	"context"
+
+	"github.com/treeverse/lakefs/pkg/graveler"
+)
+
+// taken from history:
+// https://github.com/treeverse/lakeFS/blob/606bf07969c14a569a60efe9c92831f424fa7f36/index/dag/commit_iterator.go
+type CommitGetter interface {
```
nopcoder, line 12: Just make it valid codedoc
```suggestion
// CommitGetter - taken from history:
// https://github.com/treeverse/lakeFS/blob/606bf07969c14a569a60efe9c92831f424fa7f36/index/dag/commit_iterator.go
type CommitGetter interface {
```

pkg/graveler/ref/merge_base_finder.go
```diff
@@ -0,0 +1,60 @@
...
+type reachedFlags struct {
+	fromLeft  bool
+	fromRight bool
+}
+
+func (d reachedFlags) or(fromLeft bool, fromRight bool) reachedFlags {
+	d.fromLeft = d.fromLeft || fromLeft
+	d.fromRight = d.fromRight || fromRight
+	return d
+}
```
nopcoder, line 25: If you use bitmask flags it can simplify some of the code
```suggestion
type ReachedFlags uint8

const (
	ReachedLeft ReachedFlags = 1 << iota
	ReachedRight
)

func (r ReachedFlags) ReachedBoth() bool {
	return r&ReachedRight != 0 && r&ReachedLeft != 0
}
```

pkg/graveler/ref/merge_base_finder.go
```diff
@@ -0,0 +1,60 @@
...
+
+// FindMergeBase finds the best common ancestor according to its definition it the git-merge-base documentation: https://git-scm.com/docs/git-merge-base
+// One common ancestor is better than another common ancestor if the latter is an ancestor of the former.
+func FindMergeBase(ctx context.Context, getter CommitGetter, repositoryID graveler.RepositoryID, leftID, rightID graveler.CommitID) (*graveler.Commit, error) {
+	var commitRecord *graveler.CommitRecord
+	var commitFlags reachedFlags
```
nopcoder, line 31: Don't need this one at this scope

pkg/graveler/ref/merge_base_finder.go
```diff
@@ -0,0 +1,60 @@
...
+	queue := NewCommitsGenerationPriorityQueue()
+	reached := make(map[graveler.CommitID]reachedFlags)
+	reached[rightID] = reached[rightID].or(false, true)
+	reached[leftID] = reached[leftID].or(true, false)
```
nopcoder, line 35: In case of bitmasks
```suggestion
	reached[rightID] |= ReachedRight
	reached[leftID] |= ReachedLeft
```

pkg/graveler/ref/merge_base_finder.go
```diff
@@ -0,0 +1,60 @@
...
+			return commitRecord.Commit, nil
+		}
+		for _, parent := range commitRecord.Parents {
+			parentCommit, err := getter.GetCommit(ctx, repositoryID, parent)
+			if err != nil {
+				return nil, err
+			}
+			heap.Push(queue, &graveler.CommitRecord{CommitID: parent, Commit: parentCommit})
+			// mark the parent with the flag values from its descendents:
+			reached[parent] = reached[parent].or(commitFlags.fromLeft, commitFlags.fromRight)
```
nopcoder, line 57: In case of bitmasks
```suggestion
			reached[parent] |= commitFlags
```

pkg/graveler/graveler.go
```diff
@@ -1412,6 +1421,10 @@ func (g *Graveler) Merge(ctx context.Context, repositoryID RepositoryID, destina
 		commit.Message = commitParams.Message
 		commit.MetaRangeID = metaRangeID
 		commit.Parents = []CommitID{toCommit.CommitID, fromCommit.CommitID}
+		commit.Generation = fromCommit.Generation + 1
+		if toCommit.Generation > fromCommit.Generation {
+			commit.Generation = toCommit.Generation + 1
+		}
```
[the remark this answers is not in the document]
johnnyaug: Done

nessie/merge_test.go
```diff
@@ -64,10 +79,6 @@ func TestMergeAndList(t *testing.T) {
 	require.False(t, pagin.HasMore, "pagination shouldn't have more items")
 	require.Len(t, objs, totalFiles)
 	require.Equal(t, totalFiles, pagin.Results)
-	logger.WithField("objs", objs).WithField("pagin", pagin).Info("Listed successfully")
-
-	for _, obj := range objs {
-		_, ok := checksums[obj.Checksum]
-		require.True(t, ok, "file exists in main but shouldn't, obj: %s", obj)
-	}
+	logger.WithField("iteration", iteration).WithField("objs", objs).WithField("pagin", pagin).Info("Listed successfully")
```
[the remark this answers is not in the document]
johnnyaug: Done

pkg/graveler/ref/merge_base_finder.go
johnnyaug: Done

pkg/graveler/ref/merge_base_finder.go
johnnyaug: Done

nopcoder (approved):

pkg/ddl/000027_commit_generation.up.sql
```diff
@@ -0,0 +1,14 @@
+BEGIN;
+    ALTER TABLE graveler_commits
+        ADD COLUMN generation INT;
+    WITH RECURSIVE cte AS
```
ozkatz, line 4: How did we test this?

pkg/graveler/graveler.go
```diff
@@ -1591,6 +1606,9 @@ func (g *Graveler) LoadCommits(ctx context.Context, repositoryID RepositoryID, m
 		for i, p := range commit.GetParents() {
 			parents[i] = CommitID(p)
 		}
+		if commit.GetGeneration() == 0 {
```
ozkatz, line 1609: Use a named constant?

pkg/graveler/graveler.go
```diff
@@ -1611,6 +1629,10 @@ func (g *Graveler) LoadCommits(ctx context.Context, repositoryID RepositoryID, m
 	if iter.Err() != nil {
 		return iter.Err()
 	}
+	if missingGenerations {
+		g.log.Debug("computing the generation field for loaded commits")
```
ozkatz, line 1633: add fields for repo, from and to to be able to debug this

pkg/graveler/ref/commit_generation_priority_queue.go
```diff
@@ -0,0 +1,41 @@
...
+func (c CommitsGenerationPriorityQueue) Len() int {
+	return len(c)
+}
+
+func (c CommitsGenerationPriorityQueue) Swap(i, j int) {
+	c[i], c[j] = c[j], c[i]
+}
+
+func (c *CommitsGenerationPriorityQueue) Push(x interface{}) {
+	rec := x.(*graveler.CommitRecord)
```
ozkatz, line 24: If you're implementing your own structure, why accept an `interface{}` instead of `*CommitRecord`?

pkg/graveler/ref/manager.go
```diff
@@ -391,3 +392,21 @@ func (m *Manager) ListCommits(ctx context.Context, repositoryID graveler.Reposit
 	}
 	return NewOrderedCommitIterator(ctx, m.db, repositoryID, IteratorPrefetchSize)
 }
+
+func (m *Manager) FillGenerations(ctx context.Context, repositoryID graveler.RepositoryID) error {
+	_, err := m.db.Transact(ctx, func(tx db.Tx) (interface{}, error) {
+		return tx.Exec(`WITH RECURSIVE cte AS
+								(
```
ozkatz, line 399: indentation seems a bit.. off?

pkg/graveler/ref/manager_test.go
```diff
@@ -702,3 +702,52 @@ func TestManager_GetCommitByPrefix(t *testing.T) {
...
+	}
+	c1 := addNextCommit()
+	c2 := addNextCommit(c1)
+	c3 := addNextCommit(c1)
+	c4 := addNextCommit(c2)
+	c5 := addNextCommit(c3)
+	c6 := addNextCommit(c5)
+	c7 := addNextCommit(c4)
+	c8 := addNextCommit(c6, c1)
+	/*
```
ozkatz, line 734: nice test

pkg/graveler/ref/merge_base_finder_test.go
```diff
@@ -0,0 +1,305 @@
...
+				t.Fatalf("unexpected error %v", err)
+			}
+			verifyResult(t, base, cas.Expected)
+
+		})
+	}
+}
+
+func TestGrid(t *testing.T) {
+	// Construct the following grid, taken from https://github.com/git/git/blob/master/t/t6600-test-reach.sh
```
ozkatz, line 244: nice

ozkatz (approved):
A bit worried about the data migration and would highly recommend not using `interface{}` for the priority queue.

pkg/graveler/ref/commit_generation_priority_queue.go
johnnyaug: Have to implement `heap.Interface` :(

pkg/graveler/graveler.go
johnnyaug: Constant for what? 0 is just the default value for int in this context

pkg/ddl/000027_commit_generation.up.sql
johnnyaug: Just manual testing. I understand the concern, will try to create a better test

guy-har (approved):
Great solution, clear and easy to follow, Thanks!

PR #1964, opened 2021-05-18
use a worker pool to ingest entries faster

ozkatz:


cmd/lakectl/cmd/ingest.go
```diff
@@ -12,6 +14,32 @@ const ingestSummaryTemplate = `
...
+}
+
+type stageResponse struct {
+	resp  *api.StageObjectResponse
+	error error
+}
+
+func stageWorker(wg *sync.WaitGroup, requests <-chan *stageRequest, responses chan<- *stageResponse) {
+	defer wg.Done()
+	client := getClient()
```
arielshaqed, line 32: Can we share clients?  I think HTTP connection pooling/reuse anyway happens at a level below, so it might manage to _reduce_ connection re-use which could _increase_ performance (or _decrease_ performance...).

cmd/lakectl/cmd/ingest.go
```diff
@@ -21,51 +49,77 @@ var ingestCmd = &cobra.Command{
...
+		}()
+
+		for response := range responses {
+			DieOnResponseError(response.resp, response.error)
+			summary.Objects += 1
+			summary.Bytes += api.Int64Value(response.resp.JSON201.SizeBytes)
+
+			// update every 10k records
+			if summary.Objects%10000 == 0 {
+				Write("Staged {{ .Objects | green }} objects so far...\n", summary)
```
arielshaqed, line 117:
```suggestion
				Write("Staged {{ .Objects | green }} objects so far...\r", summary)
```
might give a cute single-line "animation".  (But you'd have to write that `\n` later, when done)

arielshaqed (approved):
Neat!

I'm not sure how much speedup you can get from something like this -- each client goroutine appears to do _less_ work than the server, and the server can load multiple DB threads but they all contend for the same resource.  Still, at least now the client does the most it can do.

cmd/lakectl/cmd/ingest.go
```diff
@@ -12,6 +14,32 @@ const ingestSummaryTemplate = `
 Staged {{ .Objects | yellow }} external objects (total of {{ .Bytes | human_bytes | yellow }})
 `
 
+type stageRequest struct {
+	ctx        context.Context
```
nopcoder, line 18: ctx can be passed to the workers once on creation.

cmd/lakectl/cmd/ingest.go
```diff
@@ -81,5 +135,6 @@ func init() {
 	_ = ingestCmd.MarkFlagRequired("to")
 	ingestCmd.Flags().Bool("dry-run", false, "only print the paths to be ingested")
 	ingestCmd.Flags().BoolP("verbose", "v", false, "print stats for each individual object staged")
+	ingestCmd.Flags().IntP("concurrency", "C", 64, "max concurrent API calls to make to the lakeFS server")
```
itaiad200, line 138: Why capital 'C'?
```suggestion
	ingestCmd.Flags().IntP("concurrency", "c", 64, "max concurrent API calls to make to the lakeFS server")
```

cmd/lakectl/cmd/ingest.go
```diff
@@ -21,51 +49,77 @@ var ingestCmd = &cobra.Command{
...
+		go func() {
+			err := store.Walk(ctx, from, func(e store.ObjectStoreEntry) error {
+				if dryRun {
+					Fmt("%s\n", e)
+					return nil
+				}
+				key := e.RelativeKey
+				if lakefsURI.Path != nil && *lakefsURI.Path != "" {
+					path := *lakefsURI.Path
+					if strings.HasSuffix(*lakefsURI.Path, "/") {
```
nopcoder, line 79: This can be done once outside the loop - just set `prefix` with the right value

cmd/lakectl/cmd/ingest.go
```diff
@@ -21,51 +49,77 @@ var ingestCmd = &cobra.Command{
...
-				} else {
-					key = path + "/" + key
-				}
+			close(requests)  // we're done feeding work!
+			wg.Wait()        // until all responses have been written
+			close(responses) // so we're also done with responses
+		}()
+
+		for response := range responses {
+			DieOnResponseError(response.resp, response.error)
```
nopcoder, line 111: we just die - so no point in passing errors from the worker - can reduce all the specific response channel and just process responses

cmd/lakectl/cmd/ingest.go
```diff
@@ -21,51 +49,77 @@ var ingestCmd = &cobra.Command{
...
+			summary.Objects += 1
+			summary.Bytes += api.Int64Value(response.resp.JSON201.SizeBytes)
+
+			// update every 10k records
+			if summary.Objects%10000 == 0 {
```
nopcoder, line 116: Consider using time since instead of round numbers or optional progress bar

itaiad200 (approved):
Tested it, works like a charm!

cmd/lakectl/cmd/ingest.go
nopcoder: From a different angle - getClient should be called once, it can fail and we do not want to dump multiple errors.
They all share the same transport - but we don't need to perform the same operation multiple times to construct it.

nopcoder (changes requested):

cmd/lakectl/cmd/ingest.go
ozkatz: Done

cmd/lakectl/cmd/ingest.go
ozkatz: Nice idea. Done.

cmd/lakectl/cmd/ingest.go
ozkatz: because we use `-c` on the global `lakectl` command as short for `--config`

cmd/lakectl/cmd/ingest.go
ozkatz: Done

cmd/lakectl/cmd/ingest.go
ozkatz: Done

cmd/lakectl/cmd/ingest.go
ozkatz: Done

cmd/lakectl/cmd/ingest.go
ozkatz: Done

ozkatz:
> Neat!
> 
> I'm not sure how much speedup you can get from something like this -- each client goroutine appears to do _less_ work than the server, and the server can load multiple DB threads but they all contend for the same resource. Still, at least now the client does the most it can do.

@arielshaqed the main speedup comes from the fact that this is a `lakectl` command - there's good chance a user will run it locally from their machine. In that case, the latency between user -> lakeFS server could be big, so doing it serially one object at a time really adds up.

arielshaqed (approved):
Neat.  Love the progress-by-time idea & implementation.

nopcoder (approved):
👍

PR #1970, opened 2021-05-19
[filesystem] directory support for delete status and exists operations

### LINKED ISSUE #1855: [filesystem] Stat API - metadata

nopcoder:
- delete API - recursive use listing and delete each file. non-recursive performs object delete.
- exists API - fallback to listing one element to identify a path
- file status API - fallback to listing one element to return directory file status
Fix listing should return the full path to lakeFS object URI

Close #1855

### DIFF
--- a/clients/hadoopfs/pom.xml
+++ b/clients/hadoopfs/pom.xml
@@ -46,20 +46,20 @@
                                     </excludes>
                                 </filter>
                             </filters>
-			    <relocations>
-			      <relocation>
-				  <pattern>org.apache.httpcomponents</pattern>
-				  <shadedPattern>io.lakefs.shaded.apache.httpcomponents</shadedPattern>
-			      </relocation>
-			      <relocation>
-				<pattern>okio</pattern>
-				<shadedPattern>io.lakefs.shaded.okio</shadedPattern>
-			      </relocation>
-			      <relocation>
-				<pattern>okhttp3</pattern>
-				<shadedPattern>io.lakefs.shaded.okhttp3</shadedPattern>
-			      </relocation>
-			    </relocations>
+                            <relocations>
+                                <relocation>
+                                    <pattern>org.apache.httpcomponents</pattern>
+                                    <shadedPattern>io.lakefs.shaded.apache.httpcomponents</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>okio</pattern>
+                                    <shadedPattern>io.lakefs.shaded.okio</shadedPattern>
+                                </relocation>
+                                <relocation>
+                                    <pattern>okhttp3</pattern>
+                                    <shadedPattern>io.lakefs.shaded.okhttp3</shadedPattern>
+                                </relocation>
+                            </relocations>
                         </configuration>
                     </execution>
                 </executions>
@@ -85,16 +85,16 @@
             <version>2.7.7</version>
         </dependency>
 
-	<dependency>
-	  <groupId>com.amazonaws</groupId>
-	  <artifactId>aws-java-sdk</artifactId>
-	  <version>1.7.4</version><!-- should match hadoop-aws version(!) -->
-	</dependency>
-	<dependency>		<!-- needed for aws-java-sdk in tests(!) -->
-	  <groupId>org.apache.httpcomponents</groupId>
-	  <artifactId>httpclient</artifactId>
-	  <version>4.3.6</version>
-	</dependency>
+        <dependency>
+            <groupId>com.amazonaws</groupId>
+            <artifactId>aws-java-sdk</artifactId>
+            <version>1.7.4</version><!-- should match hadoop-aws version(!) -->
+        </dependency>
+        <dependency>        <!-- needed for aws-java-sdk in tests(!) -->
+            <groupId>org.apache.httpcomponents</groupId>
+            <artifactId>httpclient</artifactId>
+            <version>4.3.6</version>
+        </dependency>
 
         <dependency>
             <groupId>junit</groupId>
@@ -107,7 +107,7 @@
             <groupId>org.mockito</groupId>
             <artifactId>mockito-all</artifactId>
             <version>1.10.19</version>
-	    <scope>test</scope>
+            <scope>test</scope>
         </dependency>
     </dependencies>
 </project>
--- a/clients/hadoopfs/src/main/java/io/lakefs/Constants.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/Constants.java
@@ -2,7 +2,7 @@ package io.lakefs;
 
 import org.apache.commons.io.FileUtils;
 
-public class Constants {
+class Constants {
     public static final String URI_SCHEME = "lakefs";
     public static final String FS_LAKEFS_ENDPOINT_KEY = "fs.lakefs.endpoint";
     public static final String FS_LAKEFS_ACCESS_KEY = "fs.lakefs.access.key";
--- a/clients/hadoopfs/src/main/java/io/lakefs/Constants.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/Constants.java
@@ -2,9 +2,8 @@ package io.lakefs;
 
 import org.apache.commons.io.FileUtils;
 
-class Constants {
+public class Constants {
     public static final String URI_SCHEME = "lakefs";
-    public static final String URI_SEPARATOR = "/";
     public static final String FS_LAKEFS_ENDPOINT_KEY = "fs.lakefs.endpoint";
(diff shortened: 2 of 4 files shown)


arielshaqed:
Could you look at https://github.com/treeverse/lakeFS/pull/1970/checks?check_run_id=2619514247?  "Test lakeFS S3 with Spark client" seems to fail all the time with "bad request" but we ignore failures.

arielshaqed:
> Could you look at https://github.com/treeverse/lakeFS/pull/1970/checks?check_run_id=2619514247? "Test lakeFS S3 with Spark client" seems to fail all the time with "bad request" but we ignore failures.

E.g. https://github.com/treeverse/lakeFS/runs/2619834094?check_suite_focus=true actually appeared to fail there, but in a Spark-related way.

clients/hadoopfs/pom.xml
```diff
@@ -46,20 +46,20 @@
                                     </excludes>
                                 </filter>
                             </filters>
-			    <relocations>
```
arielshaqed, line 49: I think we agreed no tabs,evidently I only configured it for Java not XML.  Sorry ;-(.

clients/hadoopfs/src/main/java/io/lakefs/Constants.java
```diff
@@ -2,9 +2,8 @@
 
 import org.apache.commons.io.FileUtils;
 
-class Constants {
+public class Constants {
```
arielshaqed, line 5: Why do we need a public class?

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -275,17 +269,32 @@ public boolean delete(Path path, boolean recursive) throws IOException {
...
+                ObjectLocation objectLoc = pathToObjectLocation(path);
+                objectsApi.deleteObject(objectLoc.getRepository(), objectLoc.getRef(), objectLoc.getPath());
+            } catch (ApiException e) {
+                // This condition mimics s3a behaviour in https://github.com/apache/hadoop/blob/7f93349ee74da5f35276b7535781714501ab2457/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L2741
+                if (e.getCode() == HttpStatus.SC_NOT_FOUND) {
+                    LOG.error("Could not delete: {}, reason: {}", path, e.getResponseBody());
+                    return false;
+                }
+                throw new IOException("deleteObject", e);
             }
```
arielshaqed, line 297: Can we share this code with the loop?

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -317,32 +326,22 @@ public boolean mkdirs(Path path, FsPermission fsPermission) throws IOException {
...
-            if (e.getCode() == HttpStatus.SC_NOT_FOUND) {
-                throw new FileNotFoundException(path + " not found");
+            if (e.getCode() != HttpStatus.SC_NOT_FOUND) {
+                throw new IOException("statObject", e);
             }
-            throw new IOException("statObject", e);
-        } catch (java.net.URISyntaxException e) {
-            throw new IOException("underlying storage uri", e);
         }
+        // check if path is a directory
```
arielshaqed, line 338: This logic is obviously correct but subtle.  Maybe aid the reader with some more documentation (but open to any further ideas, of course)!
```suggestion
        // not found as a file; check if path is a "directory", i.e. a prefix.
```

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -317,32 +326,22 @@ public boolean mkdirs(Path path, FsPermission fsPermission) throws IOException {
...
+        ListingIterator iterator = new ListingIterator(path, true, 1);
```
arielshaqed, line 339: Don't you need to add a slash?

arielshaqed (changes requested):
Neat and elegant!

Only blocker is that I don't understand why the test said "bad request".

clients/hadoopfs/pom.xml
nopcoder: right, fixing - I gave the idea a chance in reformat

clients/hadoopfs/src/main/java/io/lakefs/Constants.java
nopcoder: think it me while writing a test code and want to use these constants - removed

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
nopcoder: The current implementation listing iterator works only on directories - it will because path doesn't pass the leading '/' can't tell the difference between listing lakefs://repo/branch/name and lakefs://repo/branch/name/.
Can't use it as a prefix, alternatively can make an extra call to understand if there is a file. which is what I do while before calling the listing.

arielshaqed:
Thanks!  Still getting this failure when running Nessie: https://github.com/treeverse/lakeFS/runs/2621208812?check_suite_focus=true#step:12:477 .  It doesn't fail the test because we ignore errors in running the thick Spark client, but it does mean something is wrong here.

nopcoder:
> Thanks! Still getting this failure when running Nessie: [treeverse/lakeFS/runs/2621208812?check_suite_focus=true#step:12:477](https://github.com/treeverse/lakeFS/runs/2621208812?check_suite_focus=true#step:12:477) . It doesn't fail the test because we ignore errors in running the thick Spark client, but it does mean something is wrong here.

yes... on it now - thanks

arielshaqed (changes requested):
Thanks!

One response or change to approve, about working dir (see comment).

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -358,13 +356,12 @@ private FileStatus convertObjectStatsToFileStatus(ObjectStats objectStat) throws
             if (mtime != null) {
                 modificationTime = TimeUnit.SECONDS.toMillis(mtime);
             }
-            Path filePath = new Path(objectStat.getPath()).makeQualified(this.uri, this.workingDirectory);
-
-            URI physicalUri = translateUri(new URI(objectStat.getPhysicalAddress()));
-            Path physicalPath =new Path(physicalUri.toString());
-            FileSystem physicalFS = physicalPath.getFileSystem(conf);
-            long blockSize = physicalFS.getDefaultBlockSize(physicalPath);
+            Path filePath = new Path(ObjectLocation.formatPath(repository, ref, objectStat.getPath()));
```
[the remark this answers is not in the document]
nopcoder: Assumed that `objectStat` we accept in `convertObjectStatsToFileStatus` is lakeFS's path - which is full path from the repository root. The use of workingDirectory and uri should be when we parse/build ObjectLocation in the case the path we accept as input is relative.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -406,14 +406,13 @@ private FileStatus convertObjectStatsToFileStatus(ObjectStats objectStat) throws
...
-            Path filePath = new Path(objectStat.getPath()).makeQualified(this.uri, this.workingDirectory);
-
-            URI physicalUri = translateUri(new URI(objectStat.getPhysicalAddress()));
-            Path physicalPath =new Path(physicalUri.toString());
-            FileSystem physicalFS = physicalPath.getFileSystem(conf);
-            long blockSize = physicalFS.getDefaultBlockSize(physicalPath);
-            boolean isdir = objectStat.getPathType() == ObjectStats.PathTypeEnum.COMMON_PREFIX;
-            return new FileStatus(length, isdir, 0, blockSize, modificationTime, filePath);
+            Path filePath = new Path(ObjectLocation.formatPath(repository, ref, objectStat.getPath()));
+            boolean isDir = objectStat.getPathType() == ObjectStats.PathTypeEnum.COMMON_PREFIX;
```
talSofer, line 410: I think that it can add clarity if you encapsulate this logic into a private isDir(ObjectStats/pathType) method. it will make life easier for the next person that tries to figure out what we consider as dir, and you can reuse it in line 548.

talSofer (approved):

arielshaqed (approved):
Thanks!  Sorry for unnecessary delay.

PR #1972, opened 2021-05-19
Filesystem: non-atomic files rename 

### LINKED ISSUE #1965: Support non-atomic rename
- Support non-atomic rename as the initial rename implementation 
- Rename will be implemented as a metadata operation

talSofer:
closes #1965

### Testing done 
1. rename an existing src file to an unused destination path. rename succeeded.
2. rename a non-existing src file. result: src not found and rename failed.
3. rename an existing src file to an existing destination path. rename succeeded, src was renamed and the destination file is overridden with the renamed file.

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -16,8 +16,7 @@ import com.amazonaws.auth.InstanceProfileCredentialsProvider;
 import com.amazonaws.services.s3.AmazonS3;
 import com.amazonaws.services.s3.AmazonS3Client;
 
-import io.lakefs.clients.api.model.ObjectStatsList;
-import io.lakefs.clients.api.model.Pagination;
+import io.lakefs.clients.api.model.*;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.*;
 import org.apache.hadoop.fs.permission.FsPermission;
@@ -32,8 +31,6 @@ import org.slf4j.LoggerFactory;
 import io.lakefs.clients.api.ApiException;
 import io.lakefs.clients.api.ObjectsApi;
 import io.lakefs.clients.api.StagingApi;
-import io.lakefs.clients.api.model.ObjectStats;
-import io.lakefs.clients.api.model.StagingLocation;
 
 import javax.annotation.Nonnull;
 
@@ -243,10 +240,83 @@ public class LakeFSFileSystem extends FileSystem {
         throw new UnsupportedOperationException("Append is not supported by LakeFSFileSystem");
     }
 
+    /**
+     * This method is implemented under the following assumptions:
+     * 1. rename is only supported for uncommitted data on the same branch.
+     * 2. file rename operation is supported, directories rename is unsupported.
+     * 3. the rename dst  path can be an uncommitted file, that will be overridden as a result of the rename operation.
+     * 4. On rename operation a new mtime is generated, therefore we don't preserve the mtime of the src object.
+     *
+     * @throws IOException
+     */
     @Override
-    public boolean rename(Path path, Path path1) throws IOException {
-        LOG.debug("$$$$$$$$$$$$$$$$$$$$$$$$$$$$ rename $$$$$$$$$$$$$$$$$$$$$$$$$$$$ ");
-        return false;
+    public boolean rename(Path src, Path dst) throws IOException {
+        LOG.debug("$$$$$$$$$$$$$$$$$$$$$$$$$$$$ Rename path {} to {} $$$$$$$$$$$$$$$$$$$$$$$$$$$$", src, dst);
+
+        ObjectLocation srcObjectLoc = pathToObjectLocation(src);
+        ObjectLocation dstObjectLoc = pathToObjectLocation(dst);
+        if (srcObjectLoc.equals(dstObjectLoc)) {
+            LOG.debug("rename: src and dst refer to the same lakefs object location: {}", dst);
+            return true;
+        }
+
+        if (!srcObjectLoc.onSameBranch(dstObjectLoc)) {
+            LOG.error("rename: src {} and dst {} are not on the same branch. rename outside this scope is unsupported "
+                    + "by lakefs.", src, dst);
+            return false;
+        }
+
+        ObjectStats srcStat;
+        ObjectsApi objects = lfsClient.getObjects();
+        try {
+            // Stat src file to get its metadata
+            srcStat = objects.statObject(srcObjectLoc.getRepository(), srcObjectLoc.getRef(),
+                    srcObjectLoc.getPath());
+        } catch (ApiException e) {
+            LOG.error("rename: Could not get src Object stats. src:{}", src, e);
+            return false;
+        }
+
+        return renameObject(srcStat, srcObjectLoc, dstObjectLoc);
+    }
+
+    /**
+     * Non-atomic rename operation.
+     * @return true if rename succeeded, false otherwise
+     */
+    private boolean renameObject(ObjectStats srcStat, ObjectLocation srcObjectLoc, ObjectLocation dstObjectLoc)
+            throws IOException {
+
+        ObjectsApi objects = lfsClient.getObjects();
+
+        //TODO (Tals): Can we add metadata? we currently don't have an API to get the metadata of an object.
+        ObjectStageCreation creationReq = new ObjectStageCreation()
+                .checksum(srcStat.getChecksum())
+                .sizeBytes(srcStat.getSizeBytes())
+                .physicalAddress(srcStat.getPhysicalAddress());
+
+        try {
+            objects.stageObject(dstObjectLoc.getRepository(), dstObjectLoc.getRef(), dstObjectLoc.getPath(),
+                    creationReq);
+        } catch (ApiException e) {
+            LOG.error("rename: Could not stage object on dst:{}", dstObjectLoc.getPath(), e);
+            return false;
+        }
+
+        // delete src path
+        try {
+            objects.deleteObject(srcObjectLoc.getRepository(), srcObjectLoc.getRef(), srcObjectLoc.getPath());
+        } catch (ApiException e) {
+            // This condition mimics s3a behaviour in https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L2741
+            if (e.getCode() == HttpStatus.SC_NOT_FOUND) {
+                LOG.error("Could not delete: {}, reason: {}", srcObjectLoc.getPath(), e.getResponseBody());
+                return false;
+            }
+            throw new IOException("deleteObject", e);
+        }
+
+        LOG.debug("rename: successfully renamed {} to {}", srcObjectLoc.getPath(), dstObjectLoc.getPath());
+        return true;
     }
 
     @Override
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -273,7 +273,7 @@ public class LakeFSFileSystem extends FileSystem {
             srcStat = objects.statObject(srcObjectLoc.getRepository(), srcObjectLoc.getRef(),
                     srcObjectLoc.getPath());
         } catch (ApiException e) {
-            LOG.error("rename: Could not get src Object stats. src:{}", src, e);
+            LOG.error("rename: could not get src object stats. src:{}", src, e);
             return false;
         }
 
@@ -286,7 +286,6 @@ public class LakeFSFileSystem extends FileSystem {
      */
     private boolean renameObject(ObjectStats srcStat, ObjectLocation srcObjectLoc, ObjectLocation dstObjectLoc)
             throws IOException {
-
         ObjectsApi objects = lfsClient.getObjects();
 
         //TODO (Tals): Can we add metadata? we currently don't have an API to get the metadata of an object.
--- a/clients/hadoopfs/src/main/java/io/lakefs/ObjectLocation.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/ObjectLocation.java
@@ -35,4 +35,28 @@ class ObjectLocation {
         }
         return s;
     }
+
+    @Override
+    public boolean equals(Object obj) {
+        if (obj == this) {
+            return true;
+        }
+        if (!(obj instanceof ObjectLocation)) {
+            return false;
+        }
+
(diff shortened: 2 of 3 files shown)


clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -265,9 +259,68 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
...
+            srcStatus = getFileStatus(src);
+        } catch (FileNotFoundException e) {
+            LOG.error("rename: src not found {}", src);
+            return false;
+        }
+
+        ObjectLocation srcObjectLoc = pathToObjectLocation(src);
+        ObjectLocation dstObjectLoc = pathToObjectLocation(dst);
+        if (srcObjectLoc.equals(dstObjectLoc)) {
+            LOG.debug("rename: src and dst refer to the same lakefs object location: {}", dst);
```
nopcoder, line 278: general note - "rename:" lower or title case - just align the rest

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -265,9 +259,68 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
...
+        StagingApi staging = new StagingApi(apiClient);
+        StagingLocation srcStagingLoc;
+        try {
+            srcStagingLoc = staging.getPhysicalAddress(srcObjectLoc.getRepository(),
+                    srcObjectLoc.getRef(), srcObjectLoc.getPath());
+        } catch (ApiException e) {
+            LOG.error("Rename: Could not get srcStagingLoc for src:{}", src, e);
+            return false;
+        }
+        StagingMetadata srcMetadata = new StagingMetadata().staging(srcStagingLoc)
```
nopcoder, line 300: user_metadata?

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -265,9 +259,68 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
...
+        try {
+            staging.linkPhysicalAddress(dstObjectLoc.getRepository(), dstObjectLoc.getRef(), dstObjectLoc.getPath(),
+                    srcMetadata);
+        } catch (ApiException e) {
+            LOG.error("Rename: Could not link lakefs dst location to src physical address. src:{}, dst:{}", src, dst, e);
+            return false;
+        }
+
+        // delete src path
+        delete(src, false); // delete is a metadata operation. TODO (Tals): with directories support, send recursive true
```
nopcoder, line 314: consider calling the api's deleteObject and not the public interface or extract the implementation just in case we will add statistics. extract the main functionality and call from both places also great.

nopcoder (changes requested):

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -265,9 +259,68 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
...
+        ObjectLocation dstObjectLoc = pathToObjectLocation(dst);
+        if (srcObjectLoc.equals(dstObjectLoc)) {
+            LOG.debug("rename: src and dst refer to the same lakefs object location: {}", dst);
+            return srcStatus.isFile();
+        }
+
+        ObjectStats stats;
+        try {
+            // Stat src file to get its metadata
+            stats = getObjectStats(src);
```
arielshaqed, line 285: I think we just stat'ed this file.  Doing it twice is usually worse than doing it once.  Maybe it existed here and got deleted by the time we opened?  Unless we can offer more information using this, I see no advantage in having *two* paths to return the same error, especially given that we now add a second API call.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -265,9 +259,68 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
...
+        } catch (ApiException e) {
+            LOG.error("Rename: Could not get src Object stats. src:{}", src, e);
+            return false;
+        }
+
+        StagingApi staging = new StagingApi(apiClient);
+        StagingLocation srcStagingLoc;
+        try {
+            srcStagingLoc = staging.getPhysicalAddress(srcObjectLoc.getRepository(),
```
arielshaqed, line 294: I don't understand: this requests a physical path _where we could write a new file_.  I don't think we need it here if we're doing just data operations.  The existing src file is in any case specified on `srcStatus`.

arielshaqed (changes requested):
Can you please test this, even manually (once) from spark-shell or pyspark?

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -265,9 +259,68 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
...
-        LOG.debug("$$$$$$$$$$$$$$$$$$$$$$$$$$$$ rename $$$$$$$$$$$$$$$$$$$$$$$$$$$$ ");
-        return false;
+    public boolean rename(Path src, Path dst) throws IOException {
+        LOG.debug("$$$$$$$$$$$$$$$$$$$$$$$$$$$$ Rename path {} to {} $$$$$$$$$$$$$$$$$$$$$$$$$$$$", src, dst);
+
+        // This implementation supports rename when src and dst are files. directories are not supported. TODO (Tals): support dirs
+
+        FileStatus srcStatus;
+        try {
+            srcStatus = getFileStatus(src);
```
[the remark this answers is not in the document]
talSofer: Thanks, this is indeed redaundant.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
talSofer: I removed the first call to getFileStatus following the comment in https://github.com/treeverse/lakeFS/pull/1972/files#r635169771

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
talSofer: as discussed, we currently don't have an API to get the metadata of a staged object. I added a TODO with the details.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
talSofer: You are right, I don't need this here. I chose to use the StageObject endpoint and removed this code.

talSofer:
> Can you please test this, even manually (once) from spark-shell or pyspark?

Tested with pyspark and spark-submit. 
sample output from pyspark - 
```
>>> spark._jsc.hadoopConfiguration().set("fs.lakefs.access.key", "<lakefs access key>")
>>> spark._jsc.hadoopConfiguration().set("fs.lakefs.secret.key", "<lakefs secret key>")
>>> spark._jsc.hadoopConfiguration().set("fs.defaultFS", "lakefs://aws-repo/main/initFS.txt")
>>> spark._jsc.hadoopConfiguration().set("fs.lakefs.impl", "io.lakefs.LakeFSFileSystem")
>>> path = lambda p: spark._jvm.org.apache.hadoop.fs.Path(p)
>>> fs = spark._jvm.org.apache.hadoop.fs.FileSystem.get(spark._jsc.hadoopConfiguration())
>>> print("****", fs.getScheme())
**** lakefs
>>> fs.rename(path('lakefs://aws-repo/main/peopleLakefs.parquet/brand-new.parquet'), path('lakefs://aws-repo/main/peopleLakefs.parquet/testedWithPyspark.parquet'));
True
```
Screen Shot 2021-05-20 at 13 31 37

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -264,23 +259,96 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
...
+
+        ObjectLocation srcObjectLoc = pathToObjectLocation(src);
+        ObjectLocation dstObjectLoc = pathToObjectLocation(dst);
+        if (srcObjectLoc.equals(dstObjectLoc)) {
+            LOG.debug("rename: src and dst refer to the same lakefs object location: {}", dst);
+            return true;
+        }
+
+        if (!srcObjectLoc.onSameBranch(dstObjectLoc)) {
+            LOG.error("rename: src and dst are not on the same branch. rename outside this scope is unsupported "
```
nopcoder, line 283: log the src and dst

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -264,23 +259,96 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
...
+        }
+
+        return renameObject(srcStat, srcObjectLoc, dstObjectLoc);
+    }
+
+    /**
+     * Non-atomic rename operation.
+     * @return true if rename succeeded, false otherwise
+     */
+    private boolean renameObject(ObjectStats srcStat, ObjectLocation srcObjectLoc, ObjectLocation dstObjectLoc)
```
nopcoder, line 306: Consider passing the source repository and branch instead of `srcObjectLoc`.
Later when we will iterator over a directory we will have ObjectStats for each entry and we can reuse this helper for each entry.

nopcoder (changes requested):

nopcoder (approved):
minor comments

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
talSofer: I'm keeping the current implementation, for now, will revisit while adding directories support.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -243,10 +240,83 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
...
+        ObjectLocation dstObjectLoc = pathToObjectLocation(dst);
+        if (srcObjectLoc.equals(dstObjectLoc)) {
+            LOG.debug("rename: src and dst refer to the same lakefs object location: {}", dst);
+            return true;
+        }
+
+        if (!srcObjectLoc.onSameBranch(dstObjectLoc)) {
+            LOG.error("rename: src {} and dst {} are not on the same branch. rename outside this scope is unsupported "
+                    + "by lakefs.", src, dst);
+            return false;
```
arielshaqed, line 266: Nice!

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -243,10 +240,83 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
...
+        }
+
+        ObjectStats srcStat;
+        ObjectsApi objects = lfsClient.getObjects();
+        try {
+            // Stat src file to get its metadata
+            srcStat = objects.statObject(srcObjectLoc.getRepository(), srcObjectLoc.getRef(),
+                    srcObjectLoc.getPath());
+        } catch (ApiException e) {
+            LOG.error("rename: Could not get src Object stats. src:{}", src, e);
```
arielshaqed, line 276: nit: capitalization is weird.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -243,10 +240,83 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
...
+        return renameObject(srcStat, srcObjectLoc, dstObjectLoc);
+    }
+
+    /**
+     * Non-atomic rename operation.
+     * @return true if rename succeeded, false otherwise
+     */
+    private boolean renameObject(ObjectStats srcStat, ObjectLocation srcObjectLoc, ObjectLocation dstObjectLoc)
+            throws IOException {
+
```
arielshaqed, line 289: nit: drop this blank line, distraction.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -243,10 +240,83 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
...
+        ObjectsApi objects = lfsClient.getObjects();
+
+        //TODO (Tals): Can we add metadata? we currently don't have an API to get the metadata of an object.
```
arielshaqed, line 292: This looks like a missing API: either statObject or a new API call should return it.  Let's open an issue!

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -243,10 +240,83 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
...
+        } catch (ApiException e) {
+            LOG.error("rename: Could not stage object on dst:{}", dstObjectLoc.getPath(), e);
+            return false;
+        }
+
+        // delete src path
+        try {
+            objects.deleteObject(srcObjectLoc.getRepository(), srcObjectLoc.getRef(), srcObjectLoc.getPath());
+        } catch (ApiException e) {
+            // This condition mimics s3a behaviour in https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L2741
```
arielshaqed, line 310: Nice!  Does S3A also log an error?  (Weird, because I'd expect the Java way to be throwing an Exception here.  But obviously I prefer to copy S3A here than be right.)

arielshaqed (approved):
Nice.  Thanks!

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
talSofer: Thanks! Opened https://github.com/treeverse/lakeFS/issues/1981

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
talSofer: yes, what s3a does in this case is -
```
    } catch (FileNotFoundException e) {      
LOG.debug("Couldn't delete {} - does not exist: {}", f, e.toString());
instrumentation.errorIgnored();      
return false;
```

My guess is that not throwing an exception is according to the Spark way of doing things. i.e. spark expects to get renamed=false in case that the file does not exist and can recover from it gracefully.

PR #1976, opened 2021-05-20
Extract the lakefs client with api to a dedicated client

nopcoder:


clients/hadoopfs/src/main/java/io/lakefs/LakeFSClient.java
```diff
@@ -0,0 +1,46 @@
...
+
+import io.lakefs.clients.api.ApiClient;
+import io.lakefs.clients.api.ObjectsApi;
+import io.lakefs.clients.api.StagingApi;
+import io.lakefs.clients.api.auth.HttpBasicAuth;
+import org.apache.hadoop.conf.Configuration;
+
+import java.io.IOException;
+
+public class LakeFSClient {
```
talSofer, line 11: please add class documentation

talSofer (approved):
Thanks for implementing this change!

PR #1983, opened 2021-05-23
updated faq

YaelRiv:


docs/faq.md
```diff
@@ -8,26 +8,17 @@ has_children: false
 
 # FAQ
 
-### 1. What open source license are you using?
-lakeFS is completely free and open source and licensed under the [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0) License.
+### 1. Is lakeFS open source?
+lakeFS is completely free and open source and licensed under the [Apache 2.0](https://www.apache.org/licenses/LICENSE-2.0) License. We maintain a public [product roadmap](https://docs.lakefs.io/understand/roadmap.html) and [Slack channel](https://join.slack.com/t/lakefs/shared_invite/zt-ks1fwp0w-bgD9PIekW86WF25nE_8_tw) for open discussions.
```
johnnyaug, line 12: I think you should use https://lakefs.io/slack for the Slack link, which is automatically updated with the invite token.

docs/faq.md
```diff
@@ -8,26 +8,17 @@ has_children: false
...
 
-### 2. The project is backed by a commercial company. What is your commitment to open source?
-Since it is important to us to align expectations, we have gathered our thoughts on the subject [here](https://docs.lakefs.io/understand/licensing.html).
+### 2. How does lakeFS data versioning work?
+lakeFS uses a copy-on-write mechanism. An object that did not change will invoke a metadata operation in lakeFS and an object that changed will be copied-on-wrote creating another version of the data and its corresponding metadata.
```
johnnyaug, line 15:
```suggestion
lakeFS uses a copy-on-write mechanism to avoid data duplication. For example, creating a new branch is a metadata-only operation: no objects are actually copied. Only when an object changes does lakeFS create another version of the data in the storage. For more information, see [Data Model](https://docs.lakefs.io/understand/data-model.html).
```

johnnyaug (changes requested):

johnnyaug (approved):

PR #1985, opened 2021-05-23
[filesystem] initialize using repository storage namespace

nopcoder:
Get the underlying filesystem by repository storage namespace instead of trying to address an existing file path.

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -1,14 +1,23 @@
 package io.lakefs;
 
+import java.io.FileNotFoundException;
+import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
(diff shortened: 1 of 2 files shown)


clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -93,15 +89,16 @@ public void initialize(URI name, Configuration conf) throws IOException {
         Path path = new Path(name);
 
         // TODO(ariels): Retrieve base filesystem configuration for URI from new API.  Needed
-        //     when this fs is contructed in order to create a new file, which cannot be Stat'ed
```
arielshaqed, line 96: This comment is no longer valid -- I believe YOUDID this TODO.

arielshaqed (approved):
Neat, thanks!  I hadn't realized we could do this :-)

BTW, *now* we might well decide to cache the repo->FS mapping.

talSofer (approved):

PR #1990, opened 2021-05-24
gateway on get without version should return after error

### LINKED ISSUE #1989: Gateway get object without version encode two errors in response
The lakeFS s3 gateway 'get object' encodes two errors when there is no underlying object (expiry)
Expect one error to be encoded into the response.

nopcoder:
Fix #1989


arielshaqed (approved):
Nice catch.  THANKS!

PR #1992, opened 2021-05-24
bug-fix/fix readNextChunk 

talSofer:
- Reverse the output returned from isDirectory 
- Modify readNextChunk to continue listing objects until the chunk is populated or there are no items left for listing.


nopcoder (approved):
Thanks!

PR #1994, opened 2021-05-24
fix adapter qualified key format to not use path.Join

### LINKED ISSUE #1861: Objects with physical address containing dots or slashes return an incorrect physical address
When using a GetObject/StatObject call, the API and S3 gateway use QualifiedKey.Format to format the address. This method calls the native `path.Join` method, which calls `path.Clean` - While this will work for the majority of paths, `Clean` also does (From docs):

```
1. Replace multiple slashes with a single slash.
2. Eliminate each . path name element (the current directory).
3. Eliminate each inner .. path name element (the parent directory)
   along with the non-.. element that precedes it.
4. Eliminate .. elements that begin a rooted path:
   that is, replace "/.." by "/" at the beginning of a path.
```

ll these are prefectly valid in a POSIX filesystem but completely broken on an object store where none of these notations carry any special meaning - it's just characters. This means that if we remove or modify them in any way, it'll result in broken physicall addresses returned to the user.

nopcoder:
Fix #1861

### DIFF
--- a/pkg/block/namespace.go
+++ b/pkg/block/namespace.go
@@ -4,7 +4,6 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
-	"path"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


arielshaqed (approved):
(Much) better!

PR #1995, opened 2021-05-24
Add lakefsfs component test

arielshaqed:


clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -446,6 +449,8 @@ public boolean exists(Path path) throws IOException {
         ObjectsApi objects = lfsClient.getObjects();
         ObjectLocation objectLoc = pathToObjectLocation(path);
         try {
+            LOG.debug(
+                    "[DEBUG] loc " + objectLoc.getRepository() + " " + objectLoc.getRef() + " " + objectLoc.getPath());
```
nopcoder, line 453:
```suggestion
            LOG.debug("loc {} {} {}", objectLoc.getRepository(), objectLoc.getRef(), objectLoc.getPath());
```

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -73,6 +73,10 @@ public URI getUri() {
 
     @Override
     public void initialize(URI name, Configuration conf) throws IOException {
+        initialize(name, conf, new LakeFSClient(conf));
+    }
+
+    void initialize(URI name, Configuration conf, LakeFSClient lfsClient) throws IOException {
```
nopcoder, line 79: Suggestion: different signature - mixing inheritance and overloading is never a good thing.

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
```diff
@@ -1,33 +1,166 @@
...
 
-    private LakeFSFileSystem fs;
+    protected String s3Base;
+    protected String s3Bucket;
+
+    private static final DockerImageName MINIO = DockerImageName.parse("minio/minio:RELEASE.2021-05-16T05-32-34Z");
+    protected static final String S3_ACCESS_KEY_ID = "AKIArootkey";
+    protected static final String S3_SECRET_ACCESS_KEY = "secret/minio/key=";
+
+    protected static final Regions REGION = Regions.US_EAST_1;
```
nopcoder, line 81: Can be removed

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
```diff
@@ -1,33 +1,166 @@
...
+
+    @Rule
+    public final GenericContainer s3 = new GenericContainer(MINIO.toString()).
+        withCommand("minio", "server", "/data").
+        withEnv("MINIO_ROOT_USER", S3_ACCESS_KEY_ID).
+        withEnv("MINIO_ROOT_PASSWORD", S3_SECRET_ACCESS_KEY).
+        //        withEnv("MINIO_REGION_NAME", REGION.getName()).
```
nopcoder, line 88: Can be removed

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
```diff
@@ -1,33 +1,166 @@
...
+        withEnv("MINIO_DOMAIN", "s3.local.lakefs.io").
+        withExposedPorts(9000);
+
+    protected static String makeS3BucketName() {
+        String slug = NanoIdUtils.randomNanoId(NanoIdUtils.DEFAULT_NUMBER_GENERATOR,
+                                               "abcdefghijklmnopqrstuvwxyz-0123456789".toCharArray(), 14);
```
nopcoder, line 94: Suggestion: remove the '-' from the possible characters. based on the name bucket-x it looks like we use it as a separator.

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
```diff
@@ -1,33 +1,166 @@
...
+        s3Client = new AmazonS3Client(creds, clientConfiguration);
+
+        S3ClientOptions s3ClientOptions = new S3ClientOptions()
+            .withPathStyleAccess(true);
+        s3Client.setS3ClientOptions(s3ClientOptions);
+        s3Client.setEndpoint(s3Endpoint);
+
+        // S3ClientOptions opts = new S3ClientOptions();
+        // opts.setPathStyleAccess(true);
+        // s3Client.setS3ClientOptions(opts);
```
nopcoder, line 129: remove comment

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
```diff
@@ -1,33 +1,166 @@
...
+
+        s3Bucket = makeS3BucketName();
+        s3Base = String.format("s3://%s", s3Bucket);
```
nopcoder, line 132: bucket doesn't end with '/' - so we use path with '/' prefix or we need to add one here.

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
```diff
@@ -36,6 +169,57 @@ public void getUri() throws URISyntaxException, IOException {
...
+        ByteArrayOutputStream actualBytes = new ByteArrayOutputStream();
+        {
+            int len;
+            byte[] buf = new byte[1024];
+            while ((len = in.read(buf)) > 0) {
+                actualBytes.write(buf, 0, len);
+            }
+        }
+        actualBytes.close();
+        String actual = actualBytes.toString();
```
nopcoder, line 216:
```suggestion
        String actual = IOUtils.toString(in);
```

nopcoder (changes requested):
Looks good, some minor comments

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
arielshaqed: I don't follow, sorry.  Are you suggesting changing the _name_ in addition to the signature?  Assuming "yes" and renaming this one.  Note that it is *not* public or protected, so I'm not sure overloading could ever break anything here.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
arielshaqed: True.  Actually not useful in the longer term, so removing entirely.

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
arielshaqed: Done.

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
arielshaqed: Done.

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
arielshaqed: Not done.  The (only) reason to add "-x" is to avoid ending the bucket name with a dash, which is "of course" not allowed for some buckets sometimes.

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
arielshaqed: Yes, this is done throughout the test.  Not changing.

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
arielshaqed: Done.

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
arielshaqed: Thanks!  Your way also works and might be _just a bit_ more readable :-).

arielshaqed (commented):
Thanks!  Did almost everything, said why I don't want to do some (minor I hope) cleanups.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
nopcoder: yes having two methods with the same name that one signature is set by inheritance and the other one (the protected) is not can be hard to read/debug  at runtime.

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
nopcoder: so why use '-' at the first place with nano?

nopcoder (approved):

arielshaqed:
Thanks; pulling!

PR #1997, opened 2021-05-25
Support Spark SQL tables in metastore copy

guy-har:


pkg/metastore/metastore.go
```diff
@@ -18,13 +24,28 @@ func (m *Table) Update(db, table, serde string, transformLocation func(location
...
+	if err != nil {
+		return err
+	}
+
+	if isSparkSQLTable {
+		// Table was created by Spark SQL, we should change the internal stored Spark SQL location
+		if l, ok := m.Sd.SerdeInfo.Parameters[sparkSQLProviderLocationKey]; ok {
+			m.Sd.SerdeInfo.Parameters[sparkSQLProviderLocationKey], err = transformLocation(l)
+		}
+	}
```
nopcoder, line 39: I think if we build a `func UpdateStorageDescriptor(sd *StorageDescriptor) (*StorageDescriptor, error)` and pass it to the table and partition, it will move this logic into one place where we select the way we like to address the change and how it will be applied.

pkg/metastore/metastore.go
```diff
@@ -35,9 +56,19 @@ func (m *Partition) Update(db, table, serde string, transformLocation func(locat
...
+	if m.Sd.Location != "" && !(isSparkSQLTable && strings.HasSuffix(m.Sd.Location, sparkSQLWorkaroundSuffix)) {
 		m.Sd.Location, err = transformLocation(m.Sd.Location)
 	}
+	if err != nil {
+		return err
+	}
+
+	if isSparkSQLTable {
+		// Table was created by Spark SQL, we should change the internal stored Spark SQL location
+		if l, ok := m.Sd.SerdeInfo.Parameters[sparkSQLProviderLocationKey]; ok {
```
nopcoder, line 68: note - the way we check and the way we update are the same in this case.  it is like we don't need to ask if `isSparkSQLTable` because if it was the next part will ask the same question and update.

nopcoder (changes requested):

pkg/metastore/metastore.go
guy-har: It's not the same, `isSparkSQLTable` checks for a spark provider
here we change the `path`, which can be used by other clients

nopcoder (approved):

PR #2001, opened 2021-05-25
spark test app write multiple formats

nopcoder:


test/spark/app/src/main/scala/Sonnets.scala
```diff
@@ -7,15 +7,19 @@ object Sonnets {
...
     sc.setLogLevel("INFO")
     import spark.implicits._
     val sonnets = sc.textFile(input)
     val counts = sonnets.flatMap(line => line.split(" ")).map(word => (word, 1)).reduceByKey(_ + _).toDF
-    counts.write.format("csv").save(output)
 
-    val wc = spark.read.format("csv").load(output)
+    for (fmt <- Seq("csv", "parquet", "json", "orc")) {
+      val fname = "%s.%s".format(output, fmt)
+      counts.write.format(fmt).save(fname)
```
arielshaqed, line 19: I think we should increase # of partitions here to be sure we write >1.

test/spark/app/src/main/scala/Sonnets.scala
```diff
@@ -7,15 +7,19 @@ object Sonnets {
...
+    }
+
+    val wc = spark.read.format("csv").load(output + ".csv")
```
arielshaqed, line 22: This should also be in the loop: we should be able to _read_ (not just _write) multiple formats too.

arielshaqed (changes requested):
Neat, modular! :+1: 

Requested change to approve is to read all formats not just write.  Increasing # partitions is also important, of course, but can happen in a later PR.

test/spark/app/src/main/scala/Sonnets.scala
```diff
@@ -12,18 +15,38 @@ object Sonnets {
     sc.setLogLevel("INFO")
     import spark.implicits._
     val sonnets = sc.textFile(input)
-    val counts = sonnets.flatMap(line => line.split(" ")).map(word => (word, 1)).reduceByKey(_ + _).toDF
+    val partitions = 7
+    val counts = sonnets
+      .flatMap(line => line.split(" "))
+      .map(word => (word, 1))
+      .reduceByKey(_ + _)
+      .repartition(10)
```
arielshaqed, line 23:
```suggestion
      .repartition(partitions)
```
:-)

arielshaqed (approved):
Neat, thanks!

Issue #1998, opened 2021-05-25
Warn about using "local" adapter in production


arielshaqed:
The "local" adapter is great for testing but entirely unsuitable for use in production.  When configured warn against production usage:
1. In logs.
2. On startup.
3. In the UI.
4.

PR #2003, opened 2021-05-26
Verify paths used in local block adapter are under its base dir

arielshaqed:
No current failures, but LGTM repeatedly complains that it cannot see some parameter (in this case, `uploadID`) is properly validated to avoid path traversal when using the local block adapter.

This adapter is anyway useful only for tests (#1998), but:
1. Why take a risk?
2. Try to please LGTM about this adapter once and for all.

### DIFF
--- a/pkg/block/local/adapter.go
+++ b/pkg/block/local/adapter.go
@@ -34,6 +34,7 @@ var (
 	ErrPathNotWritable       = errors.New("path provided is not writable")
 	ErrInventoryNotSupported = errors.New("inventory feature not implemented for local storage adapter")
 	ErrInvalidUploadIDFormat = errors.New("invalid upload id format")
+	ErrBadPath               = errors.New("bad path traversal blocked")
 )
 
 func WithTranslator(t block.UploadIDTranslator) func(a *Adapter) {
@@ -49,6 +50,7 @@ func WithRemoveEmptyDir(b bool) func(a *Adapter) {
 }
 
 func NewAdapter(path string, opts ...func(a *Adapter)) (*Adapter, error) {
+	// Clean() the path so that misconfiguration does not allow path traversal.
 	path = filepath.Clean(path)
 	err := os.MkdirAll(path, 0700)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


arielshaqed:
Code was safe, still is safe, but doesn't actually make LGTM any happier :-(.  I still want to pull it, of course, new code depends less on separate components to stay safe, so it is easier to reason about and see that it's safe.

arielshaqed:
@itaiad200 for tierFS, which AFAIR can _also_ use the local block adapter.

pkg/block/local/adapter.go
```diff
@@ -79,18 +81,33 @@ func resolveNamespace(obj block.ObjectPointer) (block.QualifiedKey, error) {
 	return qualifiedKey, nil
 }
 
+// verifyPath ensures that p is under the directory controlled by this adapter.  It does not
+// examine the filesystem and can mistakenly error out when symbolic links are involved.
+func (l *Adapter) verifyPath(p string) error {
```
itaiad200, line 86: This new check breaks existing local repos, no?

pkg/block/local/adapter.go
```diff
@@ -79,18 +81,33 @@ func resolveNamespace(obj block.ObjectPointer) (block.QualifiedKey, error) {
...
+	if !strings.HasPrefix(filepath.Clean(p), l.path) {
```
itaiad200, line 87: Isn't better to disallow any path traversal attempts, like
```suggestion
	if filepath.Clean(p) != p {
```

itaiad200 (changes requested):
I think we should avoid serving path traversal attempts at all and kind of wondering what happens to existing repos.

pkg/block/local/adapter.go
arielshaqed: I can see it breaking an existing repo *if*:
1. It uses the local block adapter.  (See #1998 for some more work against anyone doing this for real...).
2. One of the "create from staging" (direct upload) APIs was used to access files *not* under the root of local storage.

I *could* add a startup check, it would go over the entire DB and see that everything is OK.  Not sure it is warranted.  @ozkatz do you have a producty opinion?

pkg/block/local/adapter.go
arielshaqed: I don't think so.  E.g. the prefix itself might not be clean, or tierFS might be able to use it, etc.

Also doing this means that all callers must ensure they stitch paths onto the prefix.  I'd rather not trust callers in this way.

arielshaqed:
@ozkatz to weigh in on whether it's OK to break bad old repos that use a `local` block adapter and have files written outside its root.

ozkatz:
> @ozkatz to weigh in on whether it's OK to break bad old repos that use a `local` block adapter and have files written outside its root.

@arielshaqed edge case which I believe is unlikely to exist in reality. I'm ok with breaking (while properly communicating that breakage on release)

arielshaqed:
> > @ozkatz to weigh in on whether it's OK to break bad old repos that use a `local` block adapter and have files written outside its root.
> 
> @arielshaqed edge case which I believe is unlikely to exist in reality. I'm ok with breaking (while properly communicating that breakage on release)

OK, will write it up when releasing.

@itaiad200 PTAL.

itaiad200 (approved):

arielshaqed (commented):
Thanks!

Will release with this note:

> This release changes the behaviour of the local block adapter **not to allow** any access to files outside its root directory.  If you are using a system with the local block adapter *and* have used the staging APIs to create files whose physical path is outside its root directory, you will lose access to those files.  This limitation applies *only* to the local block adapter; it **does not apply** to the block adapters used in production (all oibject store protocols S3, Azure, GCS).

PR #2005, opened 2021-05-26
Test LakeFSFilesystem.open

### LINKED ISSUE #1966: Add lakeFS Hadoop client component test
Write a test that runs lakeFSFS versus an S3 bucket ([minIO](https://min.io) is fine).  Use it to test functionality and error handling.

## Why not a unit test?

A unit test would have to fake S3 (as well as lakeFS).  This is bad because we do not control S3, so our responses are less similar to real S3.  Using minIO is still not *quite* S3, but close enough -- plus a supported target anyway.

We shall probably still fake lakeFS, to make error testing easier.

## Why not an integration test (Nessie)?

Nessie takes a long time and is difficult to run on developer machines.  We *do* run a Spark client test as part of Nessie, and will continue to do it.  That tests a full Spark flow; the intent of this test is to test _only_ the client.

arielshaqed:
Closes #1966 because if it's good enough to test another call, it's good enough for now.

### DIFF
--- a/clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
+++ b/clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
@@ -4,6 +4,7 @@ import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.when;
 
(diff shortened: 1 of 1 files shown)


nopcoder (approved):
LGTM

arielshaqed:
Thanks!

PR #2006, opened 2021-05-26
filesystem: Add directories rename support 

### LINKED ISSUE #1982: [filesystem]: support non-atomic directories rename
Make lakefsFS support directories (common prefixes in lakefs) rename. This is required to successfully write Parquet files into lakefs. 

With this change rename of the following format should work: 
```
src = lakefs://repo/main/a.parquet/_temporary/0/_temporary/attempt_202105131342205343125378474902589_0001_m_000000_1 

dest = lakefs://repo/main/a.parquet/_temporary/0/task_202105131342205343125378474902589_0001_m_000000
```

talSofer:
Closes #1982 

### Tested scenarios
1.  rename existing src file to non-existing dst. result: rename(src.txt, non-existing-dst) -> non-existing-dst, nonexisting-dst is a file.
2. rename non-existing src file. result: src not found, rename failed. 
3. rename existing src file to existing dst. result: rename is successful, dst file is overridden with the renamed file
4. rename existing src file to existing directory. result: src is renamed to dir/src
5. rename existing dir to non-existing dir name. result: directory -> non-existing directory: rename(srcDir(containing srcDir/a.txt), non-existing-dstdir) -> non-existing-dstdir/a.txt
6. recursive dir rename: rename existing dir that contains directories under it to non-existing dir name. result: behavior is identical to dir rename (cases 4,5). 
7. rename existing dir to an existing dir name.  result:  rename(srcDir(containing srcDir/a.txt), existing-dstdir) -> existing-dstdir/srcDir/a.txt

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileStatus.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileStatus.java
@@ -1,0 +1,86 @@
+package io.lakefs;
+
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.Path;
+
+/**
+ * A class that represents the Lakefs status of a path. A path can be
+ * {@link io.lakefs.clients.api.model.ObjectStats.PathTypeEnum#OBJECT} or a
+ * {@link io.lakefs.clients.api.model.ObjectStats.PathTypeEnum#COMMON_PREFIX}.
+ * For lakefs objects, an instance of this class encapsulates its {@link io.lakefs.clients.api.model.ObjectStats}, and
+ * for common-prefix it acts as a markup that a common-prefix considered a directory.
+ */
+public class LakeFSFileStatus extends FileStatus {
+
+    private String checksum;
+    private String physicalAddress;
+
+    public LakeFSFileStatus(Builder builder) {
+        super(builder.length, builder.isdir, builder.blockReplication, builder.blockSize, builder.mTime, builder.path);
+        this.checksum = builder.checksum;
+        this.physicalAddress = builder.physicalAddress;
+    }
+
+    public String getChecksum() {
+        return checksum;
+    }
+
+    public String getPhysicalAddress() {
+        return physicalAddress;
+    }
+
+    public static class Builder {
+        private Path path;
+        private long length;
+        private boolean isdir;
+        private short blockReplication;
+        private long blockSize;
+        private long mTime;
+        private String checksum;
+        private String physicalAddress;
+
+        public Builder(Path path) {
+            this.path = path;
+        }
+
+        public Builder length(long length) {
+            this.length = length;
+            return this;
+        }
+
+        public Builder isdir(boolean isdir) {
+            this.isdir = isdir;
+            return this;
+        }
+
+        public Builder blockReplication(short blockReplication) {
+            this.blockReplication = blockReplication;
+            return this;
+        }
+
+        public Builder blocksize(long blocksize) {
+            this.blockSize = blocksize;
+            return this;
+        }
+
+        public Builder mTime(long mTime) {
+            this.mTime = mTime;
+            return this;
+        }
+
+        public Builder checksum(String checksum) {
+            this.checksum = checksum;
+            return this;
+        }
+
+        public Builder physicalAddress(String physicalAddress) {
+            this.physicalAddress = physicalAddress;
+            return this;
+        }
+
+        public LakeFSFileStatus build() {
+            LakeFSFileStatus lakeFSFileStatus =  new LakeFSFileStatus(this);
+            return lakeFSFileStatus;
+        }
+    }
+}
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -356,7 +356,7 @@ public class LakeFSFileSystem extends FileSystem {
      */
     private Path buildObjPathOnNonExistingDestinationDir(Path renamedObj, Path srcDir, Path dstDir) {
         String renamedObjName = renamedObj.toUri().getPath().substring(srcDir.toUri().getPath().length() + 1);
-        String newObjPath = dstDir.toUri() + renamedObjName;
+        String newObjPath = dstDir.toUri() + SEPARATOR + renamedObjName;
         return new Path(newObjPath);
     }
 
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -215,8 +215,7 @@ public class LakeFSFileSystem extends FileSystem {
     @Override
     public RemoteIterator<LocatedFileStatus> listFiles(Path f, boolean recursive) throws FileNotFoundException, IOException {
         LOG.debug("$$$$$$$$$$$$$$$$$$$$$$$$$$$$ listFiles path: {}, recursive {} $$$$$$$$$$$$$$$$$$$$$$$$$$$$ ", f.toString(), recursive);
-
-        return new ListingIterator(f, recursive, listAmount);
+        return toLocatedFileStatusIterator(new ListingIterator(f, recursive, listAmount));
     }
 
     /**
@@ -259,11 +258,19 @@ public class LakeFSFileSystem extends FileSystem {
     }
 
     /**
-     * This method is implemented under the following assumptions:
-     * 1. rename is only supported for uncommitted data on the same branch.
-     * 2. file rename operation is supported, directories rename is unsupported.
-     * 3. the rename dst  path can be an uncommitted file, that will be overridden as a result of the rename operation.
-     * 4. On rename operation a new mtime is generated, therefore we don't preserve the mtime of the src object.
+     * Rename, behaving similarly to the POSIX "mv" command, but non-atomically.
+     * 1. Rename is only supported for uncommitted data on the same branch.
+     * 2. The following rename scenarios are supported:
+     * * file -> existing-file-name: rename(src.txt, existing-dst.txt) -> existing-dst.txt, existing-dst.txt is overriden
+     * * file -> existing-directory-name: rename(src.txt, existing-dstdir) -> existing-dstdir/src.txt
+     * * file -> non-existing dst: in case of non-existing rename target, the src file is renamed to a file with the
+     * destination name. rename(src.txt, non-existing-dst) -> non-existing-dst, nonexisting-dst is a file.
+     * * directory -> non-existing directory:
+     * rename(srcDir(containing srcDir/a.txt), non-existing-dstdir) -> non-existing-dstdir/a.txt
+     * * directory -> existing directory:
+     * rename(srcDir(containing srcDir/a.txt), existing-dstdir) -> existing-dstdir/srcDir/a.txt
+     * 3. The rename dst  path can be an uncommitted file, that will be overridden as a result of the rename operation.
+     * 4. The mtime of the src object is not preserved.
      *
      * @throws IOException
      */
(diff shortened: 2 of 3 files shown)


clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileStatus.java
```diff
@@ -0,0 +1,34 @@
...
+import javax.annotation.Nullable;
+import java.util.Objects;
+
+public class LakeFSFileStatus extends FileStatus {
+
+    private String checksum;
+    private String physicalAddress;
+
+    public LakeFSFileStatus(long length, boolean isdir, int block_replication,
+                            long blocksize, long modification_time, Path path,
```
arielshaqed, line 15: nit: Mixed argument styles: snake_case `modification_time` and `block_replication`, uncased `isdir` (and maybe `blocksize`), whackyCapped `physicalAddress`.  Prefer to have just one.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileStatus.java
```diff
@@ -0,0 +1,34 @@
...
+                            @Nullable String checksum, @Nullable String physicalAddress) {
+        super(length, isdir, block_replication, blocksize, modification_time, path);
+        if (!isdir) {
+            // directories are not objects in lakefs, therefore we only expect to have these details for non-directories
+            Objects.requireNonNull(checksum);
+            Objects.requireNonNull(physicalAddress);
```
arielshaqed, line 21: I don't understand why we need the `@Nullable` annotation _at all_, if we're going to check at runtime.  Maybe construct directories from some separate (probably named static) method?

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileStatus.java
```diff
@@ -0,0 +1,34 @@
+package io.lakefs;
+
+import org.apache.hadoop.fs.FileStatus;
+import org.apache.hadoop.fs.Path;
+
+import javax.annotation.Nullable;
+import java.util.Objects;
+
```
arielshaqed, line 8: Please document when we return this status.  I think our users should be allowed to depend on it; if you agree then we should document it.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -241,7 +240,15 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
...
-     * 2. file rename operation is supported, directories rename is unsupported.
+     * 2. The following rename scenarios are supported:
+     *      * file -> existing-file-name: rename(src.txt, existing-dst.txt) -> existing-dst.txt, existing-dst.txt is overriden
+     *      * file -> existing-directory-name: rename(src.txt, existing-dstdir) -> existing-dstdir/src.txt
+     *      * file -> non-existing dst: in case of non-existing rename target, the src file is renamed to a file with the
+     *        destination name. rename(src.txt, non-existing-dst) -> non-existing-dst, nonexisting-dst is a file.
+     *      * directory -> non-existing directory:
+     *        rename(srcDir(containing srcDir/a.txt), non-existing-dstdir) -> non-existing-dstdir/a.txt
+     *      * directory -> existing directory:
+     *        rename(srcDir(containing srcDir/a.txt), existing-dstdir) -> existing-dstdir/srcDir/a.txt
```
arielshaqed, line 251: Great summary!

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -241,7 +240,15 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
     /**
      * This method is implemented under the following assumptions:
      * 1. rename is only supported for uncommitted data on the same branch.
-     * 2. file rename operation is supported, directories rename is unsupported.
+     * 2. The following rename scenarios are supported:
+     *      * file -> existing-file-name: rename(src.txt, existing-dst.txt) -> existing-dst.txt, existing-dst.txt is overriden
+     *      * file -> existing-directory-name: rename(src.txt, existing-dstdir) -> existing-dstdir/src.txt
```
arielshaqed, line 245: [nit, not in this PR] Maybe explain here (or somewhere...) what's a "directory"?

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -241,7 +240,15 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
...
+     *      * file -> non-existing dst: in case of non-existing rename target, the src file is renamed to a file with the
+     *        destination name. rename(src.txt, non-existing-dst) -> non-existing-dst, nonexisting-dst is a file.
```
arielshaqed, line 247:
```suggestion
     *      * file -> non-existing dst: rename(src.txt, non-existing-dst) -> non-existing-dst, nonexisting-dst is a new file.
```

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -241,7 +240,15 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
     /**
      * This method is implemented under the following assumptions:
```
arielshaqed, line 241:
```suggestion
     * Rename, behaving similarly to the POSIX "mv" command, but non-atomically.
```
(assuming that is true...).

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -241,7 +240,15 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
...
+     * 2. The following rename scenarios are supported:
+     *      * file -> existing-file-name: rename(src.txt, existing-dst.txt) -> existing-dst.txt, existing-dst.txt is overriden
+     *      * file -> existing-directory-name: rename(src.txt, existing-dstdir) -> existing-dstdir/src.txt
+     *      * file -> non-existing dst: in case of non-existing rename target, the src file is renamed to a file with the
+     *        destination name. rename(src.txt, non-existing-dst) -> non-existing-dst, nonexisting-dst is a file.
+     *      * directory -> non-existing directory:
+     *        rename(srcDir(containing srcDir/a.txt), non-existing-dstdir) -> non-existing-dstdir/a.txt
+     *      * directory -> existing directory:
+     *        rename(srcDir(containing srcDir/a.txt), existing-dstdir) -> existing-dstdir/srcDir/a.txt
      * 3. the rename dst  path can be an uncommitted file, that will be overridden as a result of the rename operation.
```
arielshaqed, line 252:
```suggestion
     * 3. the rename dst  path can be an uncommitted file, that will be overwritten by the rename operation.
```

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -241,7 +240,15 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
...
      * 4. On rename operation a new mtime is generated, therefore we don't preserve the mtime of the src object.
```
arielshaqed, line 253:
```suggestion
     * 4. The mtime of the src object is not preserved.
```

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -264,40 +271,133 @@ public boolean rename(Path src, Path dst) throws IOException {
...
-        ObjectStats srcStat;
-        ObjectsApi objects = lfsClient.getObjects();
+        LakeFSFileStatus srcStatus;
         try {
-            // Stat src file to get its metadata
-            srcStat = objects.statObject(srcObjectLoc.getRepository(), srcObjectLoc.getRef(),
-                    srcObjectLoc.getPath());
-        } catch (ApiException e) {
-            LOG.error("rename: could not get src object stats. src:{}", src, e);
+            srcStatus = getFileStatus(src);
```
arielshaqed, line 276: Nice -- I like that we do a seemingly excessive "getFileStatus" rather than test for src being a directory... but then pass the result for a file.  So either it's a directory "move" and one more action doesn't matter (and atomicity went out the window), or it's a file move and we needed to get a status anyway.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -264,40 +271,133 @@ public boolean rename(Path src, Path dst) throws IOException {
...
+    private boolean renameDirectory(Path src, Path dst) throws IOException {
+        boolean dstExists = false;
+        LakeFSFileStatus dstFileStatus;
+        try {
+            // May be unnecessary with https://github.com/treeverse/lakeFS/issues/1691
+            dstFileStatus = getFileStatus(dst);
+            dstExists = true;
+            if (!dstFileStatus.isDirectory()) {
+                LOG.error("renameDirectory: dst {} is not a directory, rename of directory to a file " +
+                        "is unsupported", dst);
```
arielshaqed, line 300: Do we need to repeat the function name?  Doesn't the logger know it?  In any case, if we have to repeat the function name here then either don't say "directory" in the message or lie about the function name and call it "`rename`" (and I can lie to myself that that's the name of the _operation_).

```suggestion
                LOG.error("renameDirectory: dst {} is not a directory", dst);
```
or
```suggestion
                LOG.error("renameDirectory: destination {} is not a directory but source {} is", dst, src);
```
or even copy the POSIX error message from `mv`
```suggestion
                LOG.error("renameDirectory: cannot overwrite non-directory '{}' with directory '{}'", dst, src);
```

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -264,40 +271,133 @@ public boolean rename(Path src, Path dst) throws IOException {
...
+                return false;
+            }
+        } catch (FileNotFoundException e) {
+            LOG.debug("renameDirectory: dst {} does not exist", dst);
+        }
+
+        ListingIterator iterator = new ListingIterator(src, true, listAmount);
```
arielshaqed, line 307: Sequentially copying files is really slow.  Add a TODO to parallelize this.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -264,40 +271,133 @@ public boolean rename(Path src, Path dst) throws IOException {
...
+     * srcDirPath: lakefs://repo/main/dir1
+     * dstDirPath: lakefs://repo/main/dir2
+     * output:
+     * lakefs://repo/main/dir2/file1.txt
+     */
+    private Path buildObjPathOnNonExistingDestinationDir(Path renamedObj, Path srcDir, Path dstDir) {
+        ObjectLocation renamedObjLoc = pathToObjectLocation(renamedObj);
+        ObjectLocation srcDirLoc = pathToObjectLocation(srcDir);
+        ObjectLocation dstDirLoc = pathToObjectLocation(dstDir);
+        String formattedPath = renamedObjLoc.getPath().replace(srcDirLoc.getPath(), dstDirLoc.getPath());
```
arielshaqed, line 340: This (and any unanchored string search+replace) is unsafe when paths repeat themselves.

arielshaqed (changes requested):
Neat, thanks!

Only required change is not to use `replace` to edit paths -- I don't understand how to be sure it will replace the right portion of the path.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileStatus.java
talSofer: sure, thanks!

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
talSofer: The logger does not add the method name unless we configure it to do so. I checked this and it seems that it impacts performance so I don't think we should do it. I will change the wording per your suggestions :)

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileStatus.java
```diff
@@ -0,0 +1,34 @@
...
+public class LakeFSFileStatus extends FileStatus {
+
+    private String checksum;
+    private String physicalAddress;
+
+    public LakeFSFileStatus(long length, boolean isdir, int block_replication,
```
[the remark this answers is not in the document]
talSofer: As discussed, I'm dropping the comment because the parent FileStatus does not have a fluent API and Object creation will remain complex.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -306,6 +306,7 @@ private boolean renameDirectory(Path src, Path dst) throws IOException {
 
         ListingIterator iterator = new ListingIterator(src, true, listAmount);
         while (iterator.hasNext()) {
+            //TODO (Tals): parallelize objects rename process.
```
arielshaqed, line 309: nit:

```suggestion
            // TODO (Tals): parallelize objects rename process.
```

(I follow https://google.github.io/styleguide/cppguide.html#TODO_Comments mostly)

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileStatus.java
```diff
@@ -31,4 +28,59 @@ public String getChecksum() {
     public String getPhysicalAddress() {
         return physicalAddress;
     }
+
+    public static class LakeFSFileStatusBuilder {
```
arielshaqed, line 32: I've also seen this style:
```suggestion
    public static class Builder {
```
to avoid having the repeat "LakeFSFileStatus" (it's there in the outer classname).

arielshaqed (approved):
Great, thanks!

PR #2010, opened 2021-05-27
Cache Docker build image on GitHub Docker Registry

arielshaqed:
Loosely based on [Caching Docker builds in GitHub Actions: Which approach is the fastest? 🤔 A
research](https://dev.to/dtinth/caching-docker-builds-in-github-actions-which-approach-is-the-fastest-a-research-18ei),
which indicates it should be the fastest.

### DIFF
--- a/.github/workflows/nessie.yaml
+++ b/.github/workflows/nessie.yaml
@@ -108,19 +108,21 @@ jobs:
           true
       - name: Build Docker image with cache and push
         run: |
-          set +e
+          set -ex
           ghr=docker.pkg.github.com/${GITHUB_REPOSITORY,,}/build-cache
-          declare -a ghr_tags=(${{ steps.version.outputs.tag }})
-          if [[ -n $GITHUB_REF ]]; then ghr_tags=(${ghr_tags[@]} ${GITHUB_REF##*/}); fi
-          ghr_repo_tags=$(printf "%s " "${ghr_tags[@]/#/-t ${ghr}:}")
+          ghr_build_tag="-t ${ghr}:${{ steps.version.outputs.tag }}"
+          if [[ -n $GITHUB_REF ]]; then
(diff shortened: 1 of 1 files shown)


arielshaqed:
Shaves ~100s off Nessie time when everything seems ready.  Compare [this on `master` ](https://github.com/treeverse/lakeFS/runs/2703634805?check_suite_focus=true) without the patch (5m6s) with [this](https://github.com/treeverse/lakeFS/runs/2704258565?check_suite_focus=true) with the patch (3m24s).

If we can move more setup time into the Docker build we should be able to get more caching action, so that's probably next.

.github/workflows/nessie.yaml
```diff
@@ -90,18 +90,47 @@ jobs:
...
+          if [[ -n $GITHUB_REF ]]; then ghr_tags=(${ghr_tags[@]} ${GITHUB_REF##*/}); fi
+          ghr_repo_tags=$(printf "%s " "${ghr_tags[@]/#/-t ${ghr}:}")
+          docker build . \
+              $ghr_repo_tags \
+              -t $ECR_REGISTRY/$ECR_REPOSITORY_LAKEFS:${{ steps.version.outputs.tag }} \
+              --build-arg VERSION=${{ steps.version.outputs.tag }} \
+              --cache-from=${ghr}
+          for ghr_tag in ${ghr_tags[0]}; do
+              docker push ${ghr}:${ghr_tag}
+          done
```
itaiad200, line 123: Drop the array and make our life simpler!

itaiad200 (approved):
Single comment, approving but do fix it :)

arielshaqed:
Thanks!  Still shaves ~50s off of the Spark client run.  Nessie itself remains slow, will next focus attention/caching there

PR #2015, opened 2021-05-27
filesystem listStatus implementation

nopcoder:
- listStatus api for file/directory uses get file status and listing api
- share gson package: the lakefs java client generated code uses gson to deserialize the response from the server. the version found with spark use different version that failed to work with string as enum type (path_type in this case)

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -629,7 +629,7 @@ public class LakeFSFileSystem extends FileSystem {
     }
 
     class ListingIterator implements RemoteIterator<LakeFSLocatedFileStatus> {
-        private final boolean recursive;
+        private final boolean removeDirectory;
         private final ObjectLocation objectLocation;
         private final String delimiter;
         private final int amount;
@@ -648,7 +648,7 @@ public class LakeFSFileSystem extends FileSystem {
          * @param amount    buffer size to fetch listing
          */
         public ListingIterator(Path path, boolean recursive, int amount) {
-            this.recursive = recursive;
+            this.removeDirectory = recursive;
             this.chunk = Collections.emptyList();
             this.objectLocation = pathToObjectLocation(path);
             String locationPath = this.objectLocation.getPath();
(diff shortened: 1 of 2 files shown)


arielshaqed:
Note: I think the [current Spark test failure](https://github.com/treeverse/lakeFS/pull/2015/checks?check_run_id=2686158268#step:12:251) is due to reading an empty "parts" file (possibly not for CSV format).

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -473,7 +470,24 @@ private boolean deleteHelper(Path path) throws IOException {
...
+            if (e.getCode() != HttpStatus.SC_NOT_FOUND) {
+                throw new IOException("statObject", e);
+            }
+        }
+        List<FileStatus> fileStatuses = new ArrayList<>();
+        ListingIterator iterator = new ListingIterator(path, false, listAmount);
+        while (iterator.hasNext()) {
+            LocatedFileStatus fileStatus = iterator.next();
+            fileStatuses.add(fileStatus);
+        }
```
arielshaqed, line 489: Can you write some Java streams+collect magic to do this?

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -615,6 +629,7 @@ public ObjectLocation pathToObjectLocation(Path path) {
     }
 
     class ListingIterator implements RemoteIterator<LakeFSLocatedFileStatus> {
+        private final boolean recursive;
```
arielshaqed, line 632: Please rename this parameter for what it does on the iterator.  LFSFS _uses_ it to implement "recursive" listings. But AFAICT what it actually _does_ in the iterator -- which is created after the LakeFS API call has already been made with the "recursive" flag -- is remove "directories" from the list.  Maybe it should be called "removeDirectories"?

arielshaqed (approved):
Nice advance, thanks!

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
nopcoder: tried a little - it is not pretty - this iterator implements RemoteIterator interface and not Iterator. Working with streams will require converting it to Iterator and make a stream out of it - if you can show me an easy way to do it I'll apply in the next PR.

PR #2019, opened 2021-05-31
Fill block locations for LocatedFileStatus

nopcoder:
Listing returns LocatedFileStatus which includes file block locations. So far assumed it is not used, but dive into it, spark iterate the block locations for files, and expect an array - null will throw an exception.

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileStatus.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileStatus.java
@@ -12,8 +12,8 @@ import org.apache.hadoop.fs.Path;
  */
 public class LakeFSFileStatus extends FileStatus {
 
-    private final String checksum;
-    private final String physicalAddress;
+    private String checksum;
+    private String physicalAddress;
 
     public LakeFSFileStatus(Builder builder) {
         super(builder.length, builder.isdir, builder.blockReplication, builder.blockSize, builder.mTime, builder.path);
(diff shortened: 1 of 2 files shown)


clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileStatus.java
```diff
@@ -12,8 +12,8 @@
  */
 public class LakeFSFileStatus extends FileStatus {
 
-    private String checksum;
-    private String physicalAddress;
+    private final String checksum;
+    private final String physicalAddress;
```
arielshaqed, line 16: Not sure Java value types use final fields. Whatever Hadoop FileStatus does we should copy. (afk...)

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileStatus.java
```diff
@@ -30,7 +30,7 @@ public String getPhysicalAddress() {
     }
 
     public static class Builder {
-        private Path path;
+        private final Path path;
```
arielshaqed, line 33: Why just this one final field?

arielshaqed (approved):
Great catch, thanks!

See nonbinding comments about finals.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileStatus.java
nopcoder: took the ide recommendations - will revert this file.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileStatus.java
nopcoder: guess this one is the one that the builder can't change.

PR #2022, opened 2021-05-31
Require Spark thick client Nessie test to pass

arielshaqed:
Working since @nopcoder's #2019 (see [post merge
test](https://github.com/treeverse/lakeFS/runs/2708293022?check_suite_focus=true)), let's not slide back!


nopcoder (approved):
🧪

PR #2023, opened 2021-05-31
Run Nessie in parallel on each cloud

arielshaqed:
Generate code _once_, then run Nessie on each cloud separately.  Should be
faster.

### DIFF
--- a/.github/workflows/nessie.yaml
+++ b/.github/workflows/nessie.yaml
@@ -214,14 +214,10 @@ jobs:
         working-directory: test/spark
         run: docker-compose logs --tail=1000 lakefs
 
-  run-system:
-    name: Run latest lakeFS app
+  gen-code:
+    name: Generate code from latest lakeFS app
     needs: deploy-image
     runs-on: ubuntu-20.04
-    env:
-      TAG: ${{ needs.deploy-image.outputs.tag }}
-      # Setting Account_ID as a secret as a way to avoid specifying it here
-      REPO: ${{ secrets.AWS_ACCOUNT_ID }}.dkr.ecr.us-east-1.amazonaws.com
     steps:
       - name: Check-out code
         uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


arielshaqed:
This reduces time to run Nessie by 3-5 minutes, I see 12m14s (single run) instead of 14m32s (once) and generally 16-20 minutes on the previous version.

arielshaqed:
Removing "Run lakeFS S3 to use with local API key" would reduce runtime by another 2m.  Or I could parallelize that _too_, making those 2m run in parallel, and reducing runtime by a bit less while keeping the same coverage.

@itaiad200 WDYT?

.github/workflows/nessie.yaml
```diff
@@ -287,6 +306,32 @@ jobs:
...
+        run: tar -xf /tmp/generated.tar.gz
+      - name: Configure AWS Credentials
+        uses: aws-actions/configure-aws-credentials@v1
+        with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: us-east-1
+      - name: Login to Amazon ECR
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v1
```
itaiad200, line 334: Why do we need aws creds for gs?

.github/workflows/nessie.yaml
```diff
@@ -304,6 +349,33 @@ jobs:
...
```
itaiad200, line 377: Why do we need aws creds for Azure?

itaiad200 (approved):

.github/workflows/nessie.yaml
arielshaqed: Fetch from ECR.  AFAIK that repo is not open for public read.  E.g. if I don't do the `aws ecr get-login ...; docker login` dance then Docker cannot fetch images:
```sh
ariels@redqueen:~/dev/lakeFS$ docker pull 977611293394.dkr.ecr.us-east-1.amazonaws.com/lakefs:latest
Error response from daemon: Head https://977611293394.dkr.ecr.us-east-1.amazonaws.com/v2/lakefs/manifests/latest: no basic auth credentials
```

We might ditch ECR (for this, at least) and just use the GitHub Docker registry.  But that would be in a separate PR.

.github/workflows/nessie.yaml
arielshaqed: Same same.  >:-)

arielshaqed (commented):
Thanks!

PR #2026, opened 2021-05-31
filesystem listing should assume pagination information is set

nopcoder:
remove conditional check of listing api pagination information


arielshaqed (approved):
Cool!

PR #2034, opened 2021-05-31
filesystem test delete api

nopcoder:


clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
```diff
@@ -44,6 +45,7 @@
 import org.junit.Test;
 import org.mockito.Answers;
 import org.mockito.ArgumentCaptor;
+import org.mockito.stubbing.OngoingStubbing;
```
arielshaqed, line 48:
```suggestion
```
Please tell IntelliJ organize imports that _it had **one job**_.

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
```diff
@@ -198,6 +196,73 @@ public void testExists_NotExists() throws ApiException, IOException {
 
         Assert.assertFalse(fs.exists(p));
     }
+    
+    @Test
+    public void testDelete_FileExists() throws ApiException, IOException {
+        when(objectsApi.statObject(eq("repo"), eq("main"), eq("delete/sample/file.txt")))
```
arielshaqed, line 202:
```suggestion
        when(objectsApi.statObject("repo", "main", "delete/sample/file.txt"))
```
(and similarly below)

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
```diff
@@ -198,6 +196,73 @@ public void testExists_NotExists() throws ApiException, IOException {
...
+                .thenReturn(new ObjectStats().
+                        path("lakefs://repo/main/delete/sample/file.txt").
+                        pathType(PathTypeEnum.OBJECT).
+                        physicalAddress(s3Url("delete/sample/file.txt")).
+                        checksum("unused").
+                        mtime(0L).
+                        sizeBytes(42L));
```
arielshaqed, line 209: @talSofer names these constants in #2014.  Can we use those here, too?  (Alternatively, I don't mind ditching the constants in favour of these values, as long as we do _that_ consistently.)

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
```diff
@@ -198,6 +196,73 @@ public void testExists_NotExists() throws ApiException, IOException {
...
+        // return true because file found
+        Assert.assertTrue(
+                fs.delete(new Path("lakefs://repo/main/no/place/file.txt"), false));
```
arielshaqed, line 212: This tests metadata mutation but not data mutation.  If you created a file and passed it in the physical address, you could verify it was actually deleted from minio.

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
```diff
@@ -198,6 +196,73 @@ public void testExists_NotExists() throws ApiException, IOException {
...
+    }
+
+    @Test
+    public void testDelete_FileNotExists() throws ApiException, IOException {
+        doThrow(new ApiException(HttpStatus.SC_NOT_FOUND, "not found"))
+                .when(objectsApi).deleteObject(eq("repo"), eq("main"), eq("no/place/file.txt"));
```
arielshaqed, line 218: We use the opposite style here, mostly because that's what the Mockito tutorial shows.  I don't mind either order, but let's pick one set of verbs and use it.
```suggestion
        when(objectsApi).deleteObject("repo", "main", "no/place/file.txt")
            .thenThrow(new ApiException(HttpStatus.SC_NOT_FOUND, "not found"));
```

arielshaqed (changes requested):
Requested change: create files and verify deletion on backing store (minio).  See the `create` and `open` tests for examples.

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
nopcoder: :) not like goimports - optimize imports

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
nopcoder: didn't work for me - had no choice on how to handle this case: mock a method that returns void.

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
nopcoder: Discussed this idea with @talSofer - we test the side-effect of the filesystem and not testing lakefs.
Currently delete is a request to lakeFS metadata only. We do not delete the data file. This is why we also do not check if the physical file was deleted.
Checking if the file is deleted means we check lakeFS.

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
nopcoder: reused the constants

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
arielshaqed: Oh right, I forgot.  You are (of course) 100% correct (and I am the corresponding percentage _less_ than 100% correct...).

arielshaqed (commented):
Thanks!  sorry for the data confusion.

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
arielshaqed: Right... I forgot that you have to do this because Java.

arielshaqed (approved):
Speaking of having _one job_...

nopcoder:
Thanks @arielshaqed!

PR #2036, opened 2021-05-31
Export using spark-submit

### LINKED ISSUE #2035: Export should be available to use with spark-submit
Example usage:
```
spark-submit --packages io.treeverse.... --class io.treverse....export param1 param2
```

Instead of writing custom code.

itaiad200:
closes #2035 

Still needs manual testing


clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
```diff
@@ -130,6 +132,46 @@ object Exporter{
   final val defaultParallelism = 10
 }
 
+object Main {
+  def main(args: Array[String]) {
+    val conf = new Conf(args)
+    println("repo is: " + conf.repo()+", rootLocation is: " + conf.rootLocation())
```
arielshaqed, line 138: nit: spacing.  Or use `s"..."` strings with interpolation.

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
```diff
@@ -130,6 +132,46 @@ object Exporter{
...
+    if (conf.prev_commit_id.isSupplied) {
+      exporter.exportFrom(conf.branch(), conf.prev_commit_id())
+      return
+    }
+
+    exporter.exportAllFromBranch(conf.branch())
+  }
+}
+
+class Conf(arguments: Seq[String]) extends ScallopConf(arguments) {
```
arielshaqed, line 164: It is highly non-trivial how to use this, so documentation of CLI args is probably required here. And thinking of it, given that there are 4 or 5 arguments, mostly optional -- consider making them flags.

arielshaqed (changes requested):
Thanks!

Only required change is to make it possible for a user to understand how to run this program.  Right now that's really hard.

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
itaiad200: Sorry, stupid debug log, removing.

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
itaiad200: `repo` and `rootLocation` are mandatory and therefore args.
Everything else is optional, hence the `opt`.

Adding a link to their wiki which is pretty good.

arielshaqed: Oh gee... the flag name is the variable name.  That's quite surprising.

So much for Scala being a function language (it fails referential transparency, because not only does `opt[String](required=false)` do something different on each call, it even does something different if the name of the variable is changed...).

Anyway, how does the user who just wants to run the program know what args?  Do we get a usage message from scallop if I run with no parameters?

arielshaqed (approved):
Thanks!

Approving because it's better than we had before, but **please** find some way of documenting usage for a user who might not be aware of Scallop.

clients/spark/core/src/main/scala/io/treeverse/clients/Exporter.scala
```diff
@@ -130,6 +132,49 @@ object Exporter{
...
+}
+
+
+/** Conf options and arguments usage is documented in https://github.com/scallop/scallop/wiki
+ *
+ */
+class Conf(arguments: Seq[String]) extends ScallopConf(arguments) {
+  val branch = opt[String](required=false)
+  val commit_id = opt[String](required=false)
+  val prev_commit_id = opt[String](required=false)
```
johnnyaug, line 170: According to https://github.com/scallop/scallop/issues/150#issuecomment-336894579, scallop converts camelCase to dash-case. So you can use `commitID`, `prevCommitID` etc.

johnnyaug (approved):
Looking good. Missing a detailed help text

PR #2039, opened 2021-06-01
Add test for file not existing on open()

arielshaqed:


clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
```diff
@@ -316,6 +316,18 @@ public void testOpen() throws ApiException, IOException {
         Assert.assertEquals(contents, actual);
     }
 
+    @Test
+    public void testOpen_NotExists() throws ApiException {
```
talSofer, line 320: - Consider using the new translateException to throw a FileNotFoundException in case of an APIException with code=HttpStatus.SC_NOT_FOUND. 
- Then you can change the test to use this annotation @Test(expected = FileNotFoundException.class), and replace lines 325 - 328 with simply calling fs.open(p).

arielshaqed: Neat, thanks!

arielshaqed:
Thanks @talSofer, PTAL!

talSofer (approved):
Looks good!

arielshaqed:
Thanks!

PR #2040, opened 2021-06-01
fix dump refs bug

johnnyaug:
Fixes a regression where the refs-dump is not created because of the OpenAPI client, which deserilizes the response and puts it under the `JSON201` property of the response object. The output of this command is a JSON returned from the server (see RefsDump in swagger.yml).


arielshaqed (approved):
How was this tested?  Can you just include the output from running it?  I mean, _of **course**_ it works, the goal of this is to explain to someone who searches and finds this PR how they might do something similar.

Thanks!

johnnyaug:
@arielshaqed , added a description, hope that clarifies the matter.

Issue #2029, opened 2021-05-31
UI - Uninformative Changes / Compere views


LiorItzhak:
We need better views to give us better insights on our workspace.

we ask ourselves:
which collections have been changed? / which path prefixes will be modified? 
in what why those collections changed (new partitions? removed partitions? modified partitions?)
how many different files under a path prefix?

We need better navigation between the affected collections/paths using a more sophisticated view with path hierarchy.

We visualize it as more similar to the GitLab/GitHub changes view.

It would help if you considered making it possible to compare an uncommitted branch (Wokespace) with a different branch (similar to "compare with branch" functionality in IDEAs like IntelliJ)

You may also consider adding annotation for the objects under the 'Objects' view - specify the last commit/user that modified the object (like git blame...)  

TNX - Lior Itzhak Similarweb

PR #2051, opened 2021-06-02
Support common prefix listing of diffs (uncommitted, commit view, compare view)

ozkatz:
Addressing some of #2029 - When working with large diffs, paging across multiple data objects is not optimal.
This PR adds the ability (a toggle in the UI) to view diffs with a directory structure. Should make answering "which tables/partitions were modifed" much easier.

### DIFF
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -2094,6 +2094,7 @@ paths:
     parameters:
       - $ref: "#/components/parameters/PaginationAfter"
       - $ref: "#/components/parameters/PaginationAmount"
+      - $ref: "#/components/parameters/PaginationPrefix"
       - in: path
         name: repository
         required: true
@@ -2104,6 +2105,11 @@ paths:
         required: true
         schema:
           type: string
+      - in: query
+        name: delimiter
+        schema:
+          type: string
+          default: "/"
     get:
       tags:
         - branches
@@ -2144,6 +2150,7 @@ paths:
         description: a reference (could be either a branch or a commit ID) to compare against
       - $ref: "#/components/parameters/PaginationAfter"
       - $ref: "#/components/parameters/PaginationAmount"
+      - $ref: "#/components/parameters/PaginationPrefix"
       - in: query
         name: type
         schema:
@@ -2154,6 +2161,11 @@ paths:
           type: string
           enum: [two_dot, three_dot]
           default: three_dot
+      - in: query
+        name: delimiter
+        schema:
+          type: string
+          default: "/"
     get:
       tags:
         - refs
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -307,6 +307,7 @@ components:
           enum: [ common_prefix, object ]
         size_bytes:
           type: integer
+          description: represents the size of the added/changed/deleted entry
           format: int64
 
     DiffList:
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -35,12 +35,14 @@ components:
       description: return items prefixed with this value
       schema:
         type: string
+
     PaginationAfter:
       in: query
       name: after
       description: return items after this value
       schema:
         type: string
+
     PaginationAmount:
       in: query
       name: amount
@@ -50,6 +52,13 @@ components:
         minimum: -1
         maximum: 1000
         default: 100
+
+    PaginationDelimiter:
+      in: query
+      name: delimiter
+      description: delimiter used to group common prefixes by
+      schema:
+        type: string
 
   responses:
     Unauthorized:
@@ -296,6 +305,9 @@ components:
         path_type:
           type: string
           enum: [ common_prefix, object ]
+        size_bytes:
+          type: integer
+          format: int64
 
     DiffList:
       type: object
@@ -2095,6 +2107,7 @@ paths:
       - $ref: "#/components/parameters/PaginationAfter"
       - $ref: "#/components/parameters/PaginationAmount"
       - $ref: "#/components/parameters/PaginationPrefix"
+      - $ref: "#/components/parameters/PaginationDelimiter"
       - in: path
         name: repository
         required: true
@@ -2105,11 +2118,7 @@ paths:
         required: true
         schema:
           type: string
-      - in: query
-        name: delimiter
-        schema:
-          type: string
-          default: "/"
+
     get:
       tags:
         - branches
@@ -2151,6 +2160,7 @@ paths:
       - $ref: "#/components/parameters/PaginationAfter"
       - $ref: "#/components/parameters/PaginationAmount"
       - $ref: "#/components/parameters/PaginationPrefix"
+      - $ref: "#/components/parameters/PaginationDelimiter"
       - in: query
         name: type
         schema:
@@ -2161,11 +2171,7 @@ paths:
           type: string
           enum: [two_dot, three_dot]
           default: three_dot
-      - in: query
-        name: delimiter
-        schema:
-          type: string
-          default: "/"
+
     get:
       tags:
         - refs
@@ -2544,18 +2550,11 @@ paths:
         schema:
           type: string
         description: a reference (could be either a branch or a commit ID)
-      - in: query
-        name: prefix
-        required: false
-        schema:
-          type: string
       - $ref: "#/components/parameters/PaginationAfter"
       - $ref: "#/components/parameters/PaginationAmount"
-      - in: query
-        name: delimiter
-        schema:
-          type: string
-          default: "/"
+      - $ref: "#/components/parameters/PaginationDelimiter"
+      - $ref: "#/components/parameters/PaginationPrefix"
+
     get:
       tags:
         - objects
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -42,6 +42,9 @@ const (
 
 	actionStatusCompleted = "completed"
 	actionStatusFailed    = "failed"
+
+	entryTypeObject       = "object"
+	entryTypeCommonPrefix = "common_prefix"
 )
 
 type actionsHandler interface {
@@ -1536,11 +1539,11 @@ func (c *Controller) ResetBranch(w http.ResponseWriter, r *http.Request, body Re
 
 	var err error
 	switch body.Type {
-	case "common_prefix":
+	case entryTypeCommonPrefix:
 		err = c.Catalog.ResetEntries(ctx, repository, branch, StringValue(body.Path))
 	case "reset":
 		err = c.Catalog.ResetBranch(ctx, repository, branch)
-	case "object":
+	case entryTypeObject:
 		err = c.Catalog.ResetEntry(ctx, repository, branch, StringValue(body.Path))
 	default:
 		writeError(w, http.StatusNotFound, "reset type not found")
@@ -1629,9 +1632,9 @@ func (c *Controller) DiffBranch(w http.ResponseWriter, r *http.Request, reposito
 
 	results := make([]Diff, 0, len(diff))
 	for _, d := range diff {
-		pathType := "object"
+		pathType := entryTypeObject
 		if d.Type == catalog.DifferenceTypeCommonPrefix {
-			pathType = "common_prefix"
+			pathType = entryTypeCommonPrefix
 		}
 		results = append(results, Diff{
 			Path:     d.Path,
@@ -1767,7 +1770,7 @@ func (c *Controller) UploadObject(w http.ResponseWriter, r *http.Request, reposi
 		Checksum:        blob.Checksum,
 		Mtime:           writeTime.Unix(),
 		Path:            params.Path,
-		PathType:        "object",
+		PathType:        entryTypeObject,
 		PhysicalAddress: qk.Format(),
 		SizeBytes:       Int64Ptr(blob.Size),
 	}
@@ -1833,7 +1836,7 @@ func (c *Controller) StageObject(w http.ResponseWriter, r *http.Request, body St
 		Checksum:        entry.Checksum,
 		Mtime:           entry.CreationDate.Unix(),
 		Path:            entry.Path,
-		PathType:        "object",
+		PathType:        entryTypeObject,
 		PhysicalAddress: qk.Format(),
 		SizeBytes:       Int64Ptr(entry.Size),
 	}
@@ -2196,9 +2199,9 @@ func (c *Controller) DiffRefs(w http.ResponseWriter, r *http.Request, repository
 	}
 	results := make([]Diff, 0, len(diff))
 	for _, d := range diff {
-		pathType := "object"
+		pathType := entryTypeObject
 		if d.Type == catalog.DifferenceTypeCommonPrefix {
-			pathType = "common_prefix"
+			pathType = entryTypeCommonPrefix
 		}
 		results = append(results, Diff{
 			Path:     d.Path,
@@ -2371,7 +2374,7 @@ func (c *Controller) ListObjects(w http.ResponseWriter, r *http.Request, reposit
 		if entry.CommonLevel {
 			objList = append(objList, ObjectStats{
 				Path:     entry.Path,
-				PathType: "common_prefix",
+				PathType: entryTypeCommonPrefix,
 			})
 		} else {
 			var mtime int64
@@ -2383,7 +2386,7 @@ func (c *Controller) ListObjects(w http.ResponseWriter, r *http.Request, reposit
 				Mtime:           mtime,
 				Path:            entry.Path,
 				PhysicalAddress: qk.Format(),
-				PathType:        "object",
+				PathType:        entryTypeObject,
 				SizeBytes:       Int64Ptr(entry.Size),
 			})
 		}
@@ -2434,7 +2437,7 @@ func (c *Controller) StatObject(w http.ResponseWriter, r *http.Request, reposito
 		Checksum:        entry.Checksum,
 		Mtime:           entry.CreationDate.Unix(),
 		Path:            params.Path,
-		PathType:        "object",
+		PathType:        entryTypeObject,
 		PhysicalAddress: qk.Format(),
 		SizeBytes:       Int64Ptr(entry.Size),
 	}
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -1611,17 +1611,32 @@ func (c *Controller) DiffBranch(w http.ResponseWriter, r *http.Request, reposito
 	}
 	ctx := r.Context()
 	c.LogAction(ctx, "diff_workspace")
-	diff, hasMore, err := c.Catalog.DiffUncommitted(ctx, repository, branch, paginationAmount(params.Amount), paginationAfter(params.After))
+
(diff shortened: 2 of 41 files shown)


pkg/catalog/catalog.go
```diff
@@ -991,23 +990,75 @@ func (c *Catalog) DiffUncommitted(ctx context.Context, repository string, branch
...
+		if !strings.HasPrefix(path, prefix) {
+			break // we only want things that start with prefix, apparently there are none left
 		}
+
+		if delimiter != "" {
+			// common prefix logic goes here.
+			// for every path, after trimming "prefix", take the string upto-and-including the delimiter
+			// if the received path == the entire remainder, add that object as is
+			// if it's just a part of the name, add it as a "common prefix" entry -
+			//   and skip to next record following all those starting with this prefix
```
nopcoder, line 1036: Consider reduce the code block size inside this loop by splitting it to extractCommonPrefix, checking if there is a common prefix, and fall to the part where we append and check the limit)

webui/src/pages/repositories/repository/changes.jsx
```diff
@@ -198,57 +201,121 @@ const ChangesBrowser = ({ repo, reference, after, onSelectRef, onPaginate }) =>
...
             reference={reference}
             onPaginate={after => router.push({
                 pathname: `/repositories/:repoId/changes`,
                 params: {repoId: repo.id},
-                query: {ref: reference.id, after}
+                query: {
+                    ref: reference.id,
+                    after: (!!after) ? after : "",
+                    prefix: (!!prefix) ? prefix : "",
+                    delimiter: (!!delimiter) ? delimiter : "",
```
nopcoder, line 309: no need to convert to boolean when we evaluate an expression already

```suggestion
                    after: after ? after : "",
                    prefix: prefix ? prefix : "",
                    delimiter: delimiter ? delimiter : "",
```

webui/src/pages/repositories/repository/changes.jsx
```diff
@@ -198,57 +201,121 @@ const ChangesBrowser = ({ repo, reference, after, onSelectRef, onPaginate }) =>
...
+    const { after, prefix, delimiter } = router.query
 
     if (loading) return <Loading/>
     if (!!error) return <Error error={error}/>
 
     return (
         <ChangesBrowser
             after={(!!after) ? after : ""}
+            prefix={(!!prefix) ? prefix : ""}
+            delimiter={(!!delimiter) ? delimiter : ""}
```
nopcoder, line 299:
```suggestion
            after={after ? after : ""}
            prefix={prefix ? prefix : ""}
            delimiter={delimiter ? delimiter : ""}
```

webui/src/lib/components/repository/tree.jsx
```diff
@@ -220,23 +220,38 @@ function pathParts(path, rootName = "root") {
     return resolved;
 }
 
-const URINavigator = ({ repo, reference, path }) => {
+const buildPathURL = (params, query) => {
+    return {pathname: '/repositories/:repoId/objects', params, query};
+};
+
+export const URINavigator = ({ repo, reference, path, relativeTo = "", pathURLBuilder = buildPathURL }) => {
```
nopcoder, line 227:
```suggestion
export const URINavigator = ({ repo, reference, path, relativeTo="", pathURLBuilder=buildPathURL }) => {
```

nopcoder (approved):
Great! minor style comments

pkg/catalog/catalog.go
```diff
@@ -991,23 +990,79 @@ func (c *Catalog) DiffUncommitted(ctx context.Context, repository string, branch
 	}
 	it := NewEntryDiffIterator(iter)
 	defer it.Close()
-	return listDiffHelper(it, limit, after)
+	return listDiffHelper(it, prefix, delimiter, limit, after)
+}
+
+// GetStartPos finds the correct place to start iterating from based on delimiter, prefix & after
```
arielshaqed, line 996: Can you improve the documentation?  E.g. something like
```go
// GetStartPos returns the correct place to start iterating on all elements in the keys that start with `prefix' after
// `after' and taking `delimiter' into account.
```

However I still don't understand what the function does.  Specifically, for non-empty `delimiter` and `after` it does not actually return a reasonable place to start iterating, AFAICT it returns a place _after which_ to start iterating.  So perhaps
```go
// GetStartPos returns a key that SeekGE will transform to a place start iterating on all elements in the keys that start with `prefix' after
// `after' and taking `delimiter' into account
```
which is still too long.

Thing is, it's not a "correct place to start iterating" if you're going to advance past it.  So AFAIU somewhere in the godoc should say when you need to call SeekGE on the returned key.

pkg/catalog/catalog.go
```diff
@@ -991,23 +990,79 @@ func (c *Catalog) DiffUncommitted(ctx context.Context, repository string, branch
...
+func GetStartPos(prefix, after, delimiter string) string {
+	switch {
+	case prefix == "" && after == "":
+		// if no prefix given and no after given, start at the beginning
+		return ""
```
arielshaqed, line 1001: Given that the other 2 non-default cases are `after != ""`, this case seems superfluous and folds into the default.  The suggestion below incorporates this fact and may be more readable.

pkg/catalog/catalog.go
```diff
@@ -991,23 +990,79 @@ func (c *Catalog) DiffUncommitted(ctx context.Context, repository string, branch
...
+	case delimiter != "" && after != "":
+		// if we have a delimiter and after is not empty, start at the next common prefix after "after"
+		return string(graveler.UpperBoundForPrefix([]byte(after)))
+	case after != "":
+		// otherwise if after is set and no delimiter, continue from after
+		return after
+	default:
+		// whether we have a delimiter or not, if after is not set, start at prefix
+		return prefix
+	}
```
arielshaqed, line 1011: This looks like it might be an easier `if`-chain if we reverse main comparison in the `switch`:
```suggestion
	if after == "" {
		// whether we have a delimiter or not, if after is not set, start at prefix
		return prefix
	}
	if delimiter == "" {
		// no delimiter, continue from after
		return after
	}
	// there is a delimiter and after is not empty, start at the next common prefix after "after"
		return string(graveler.UpperBoundForPrefix([]byte(after)))
```

pkg/catalog/catalog_test.go
```diff
@@ -10,6 +10,46 @@ import (
...
+		t.Run(cas.Name, func(t *testing.T) {
+			got := GetStartPos(cas.Prefix, cas.After, cas.Delimiter)
+			if got != cas.Expected {
+				t.Fatalf("expected %s got %s", cas.Expected, got)
+			}
+		})
+	}
+}
+
 func TestCatalog_ListRepositories(t *testing.T) {
```
arielshaqed, line 53: Can/should we also add tests for listings with _empty_ prefixes, i.e. paths such as `a/b///c`, `a/b//c`, `a/b/c`?

pkg/catalog/diff.go
```diff
@@ -7,6 +7,7 @@ const (
 	DifferenceTypeRemoved
 	DifferenceTypeChanged
 	DifferenceTypeConflict
+	DifferenceTypeCommonPrefix
```
arielshaqed, line 10: We should probable be able to count these separately, unlike what we do above.  Not having to add a stringification for the new type is something of a smell.

webui/src/lib/components/repository/changes.jsx
```diff
@@ -109,7 +114,17 @@ export const ChangeEntryRow = ({ entry, showActions, onRevert }) => {
                     {diffIndicator}
                 </td>
                 <td className="tree-path">
-                    <span>{pathText}</span>
+                    {(entry.path_type === "common_prefix") ? (
```
arielshaqed, line 117: funny JSX indentation here at the end of ` {...}` block.

arielshaqed (changes requested):
Please note LGTM.com comments -- they are sometimes worthwhile, and in any case I'd rather keep the number of open comments down.

Thanks, it looks great and my comments are minor.

webui/src/pages/repositories/repository/compare.jsx
```diff
@@ -47,13 +50,80 @@ const CompareList = ({ repo, reference, compareReference, after, onSelectRef, on
...
-                            ))}
-                            </tbody>
-                        </Table>
+                        <Card.Header>
+                        <span className="float-left">
+                            {(delimiter !== "") && (
+                                <URINavigator
+                                    path={prefix}
+                                    reference={reference}
+                                    relativeTo={`${reference.id}...${compareReference.id}`}
```
johnnyaug, line 59: optional: trim if reference is a commit hash, to conform with a single commit view

webui/src/pages/repositories/repository/commits/commit/index.jsx
```diff
@@ -18,26 +18,79 @@ import ButtonGroup from "react-bootstrap/ButtonGroup";
...
+                                <URINavigator
+                                    path={prefix}
+                                    reference={{type: 'commit', id: commit.id.substr(0, 12)}}
+                                    relativeTo={commit.id.substr(0, 12)}
+                                    repo={repo}
+                                    pathURLBuilder={(params, query) => {
+                                        return {
+                                            pathname: '/repositories/:repoId/commits/:commitId',
+                                            params: {repoId: repo.id, commitId: commit.id},
+                                            query: {delimiter: "/", prefix: query.path}
```
johnnyaug, line 52: is this the right place to hardcode a `/`?

webui/src/pages/repositories/repository/commits/commit/index.jsx
```diff
@@ -18,26 +18,79 @@ import ButtonGroup from "react-bootstrap/ButtonGroup";
...
+                                        }
+                                    }}
+                                />
+                            )}
+                        </span>
+                        <span className="float-right">
+                            <Form>
+                                <Form.Switch
```
johnnyaug, line 60: Optional: extract the toggle switch to a component

johnnyaug (changes requested):
That was fast

pkg/catalog/catalog.go
```diff
@@ -991,23 +990,79 @@ func (c *Catalog) DiffUncommitted(ctx context.Context, repository string, branch
...
+
+// GetStartPos finds the correct place to start iterating from based on delimiter, prefix & after
+func GetStartPos(prefix, after, delimiter string) string {
+	switch {
+	case prefix == "" && after == "":
+		// if no prefix given and no after given, start at the beginning
+		return ""
+	case delimiter != "" && after != "":
+		// if we have a delimiter and after is not empty, start at the next common prefix after "after"
+		return string(graveler.UpperBoundForPrefix([]byte(after)))
```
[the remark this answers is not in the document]
ozkatz: with a normal `after`, you'd want to start the page with the next lexicographically sorted entry after `after`.
Once a delimiter is present, you'd want the next entry (or prefix) that doesn't start with `after`.

Example: 

after = `a/b/`
without a delimiter, `a/b/c` would be a valid first entry
with a delimiter `/`, `a/b/c` is wrong, as it is already folded into `a/b/` - the first valid entry would be the next entry not starting with `a/b/`

webui/src/pages/repositories/repository/changes.jsx
```diff
@@ -198,57 +201,121 @@ const ChangesBrowser = ({ repo, reference, after, onSelectRef, onPaginate }) =>
...
-
             <div className="tree-container">
-                    {(results.length === 0) ? <Alert variant="info">No changes</Alert> : (
-                        <Card>
+                {(results.length === 0) ? <Alert variant="info">No changes</Alert> : (
+                <Card>
+                    <Card.Header>
+                        <span className="float-left">
+                            {(delimiter !== "") && (
+                                <URINavigator path={prefix} reference={reference} repo={repo} relativeTo={`${reference.id} workspace`} pathURLBuilder={(params, query) => {
```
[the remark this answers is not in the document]
ozkatz: We do use it publicly..

api/swagger.yml
```diff
@@ -2104,6 +2105,11 @@ paths:
         required: true
         schema:
           type: string
+      - in: query
```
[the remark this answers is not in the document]
ozkatz: Unfortunately, $ref can only refer to a single component. And there's no way (that I could find) to group multiple components together

pkg/api/transform.go
```diff
@@ -10,7 +10,7 @@ func transformDifferenceTypeToString(d catalog.DifferenceType) string {
 		return "added"
 	case catalog.DifferenceTypeRemoved:
 		return "removed"
-	case catalog.DifferenceTypeChanged:
+	case catalog.DifferenceTypeChanged, catalog.DifferenceTypeCommonPrefix:
```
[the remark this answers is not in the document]
ozkatz: Added comment

pkg/catalog/catalog.go
```diff
@@ -991,23 +990,79 @@ func (c *Catalog) DiffUncommitted(ctx context.Context, repository string, branch
...
+	default:
+		// whether we have a delimiter or not, if after is not set, start at prefix
+		return prefix
+	}
 }
 
-func listDiffHelper(it EntryDiffIterator, limit int, after string) (Differences, bool, error) {
+const commonPrefixSplitParts = 2
+
+func listDiffHelper(it EntryDiffIterator, prefix, delimiter string, limit int, after string) (Differences, bool, error) {
```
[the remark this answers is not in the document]
ozkatz: Good point - but this is logic that should be in `GetStartPos` - as in this case the start pos should be "c" and not "a". I also added a test case for it.

pkg/catalog/catalog_test.go
ozkatz: Yep, added

pkg/catalog/diff.go
ozkatz: Good point.

webui/src/lib/components/repository/changes.jsx
ozkatz: Fixed

webui/src/pages/repositories/repository/compare.jsx
ozkatz: Will do

pkg/api/controller.go
```diff
@@ -2158,19 +2176,37 @@ func (c *Controller) DiffRefs(w http.ResponseWriter, r *http.Request, repository
 	if params.Type != nil && *params.Type == "two_dot" {
 		diffFunc = c.Catalog.Diff
 	}
+
+	// discern between an empty delimiter and no delimiter being passed at all
+	// by default, go-swagger will use the default value ("/") even if we pass
+	// a delimiter param that is explicitly empty. This overrides this (wrong) behavior.
```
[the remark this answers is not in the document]
ozkatz: Well, this uncovered a weird behavior in our system. Once upon a time, there was no delimiter in the object listing API.
It worked as if there's always a delimiter that is `"/"`. When we added a configuratble one, we tried being backwards compatible. IMO, this was probably a mistake. It worked like this:

* if no delimiter is passed, assume delimiter = `/`
* if delimiter is passed but is empty (i.e. `?delimiter=`), assume no delimiter
* if delimiter is passed and is any other value, use that.

while it retained the previous behavior it's a weird choice. it means that if I add a delimiter parameter to diff endpoints I can choose between:

* doing the same thing and now assume that no delimiter actually means `/` - this would now break existing diff clients.
* make the API inconsistent by having different defaults: for objects, no delimiter means `/`, for diff it means no delimiter.
* make the API consistent by fixing object listing (and breaking it): no delimiter means no delimiter.

I went with the painful option #3.

pkg/catalog/diff.go
ozkatz: Removed it.

pkg/catalog/catalog.go
ozkatz: Done

webui/src/lib/components/repository/changes.jsx
```diff
@@ -50,7 +52,10 @@ export const ChangeEntryRow = ({ entry, showActions, onRevert }) => {
             break;
     }
 
-    const pathText = entry.path;
+    let pathText = entry.path;
+    if (pathText.indexOf(relativeTo) === 0) {
```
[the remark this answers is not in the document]
ozkatz: Done

pkg/catalog/catalog.go
```diff
@@ -991,23 +990,75 @@ func (c *Catalog) DiffUncommitted(ctx context.Context, repository string, branch
 	}
 	it := NewEntryDiffIterator(iter)
 	defer it.Close()
-	return listDiffHelper(it, limit, after)
+	return listDiffHelper(it, prefix, delimiter, limit, after)
+}
+
+func GetStartPos(prefix, after, delimiter string) string {
+	// find the correct place to start iterating from based on delimiter, prefix, after
```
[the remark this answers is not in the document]
ozkatz: Done

pkg/api/controller.go
```diff
@@ -2158,19 +2176,37 @@ func (c *Controller) DiffRefs(w http.ResponseWriter, r *http.Request, repository
...
+	var delimiter string
+	if params.Delimiter == nil {
+		delimiter = ""
+	} else {
+		delimiter = *params.Delimiter
+	}
```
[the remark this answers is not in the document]
ozkatz: removed

pkg/api/controller.go
```diff
@@ -1611,17 +1614,32 @@ func (c *Controller) DiffBranch(w http.ResponseWriter, r *http.Request, reposito
...
```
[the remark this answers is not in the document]
ozkatz: removed

ozkatz:
@nopcoder @johnnyaug @arielshaqed - please notice that I broke backwards compatibility for the object listing API. If you have a better idea I'd love to not do it.

api/swagger.yml
```diff
@@ -296,6 +305,9 @@ components:
         path_type:
           type: string
           enum: [ common_prefix, object ]
+        size_bytes:
+          type: integer
+          format: int64
```
arielshaqed, line 310: Unclear what is the "size" of a diff.  Can you add a description please?

pkg/api/controller_test.go
```diff
@@ -935,8 +935,9 @@ func TestController_ObjectsListObjectsHandler(t *testing.T) {
 	}
 
 	t.Run("get object list", func(t *testing.T) {
+		pfx := api.PaginationPrefix("foo/")
```
arielshaqed, line 938:
```suggestion
		prefix := api.PaginationPrefix("foo/")
```
Why abbreviate?

arielshaqed (approved):
1. About b/c: as long as existing deployed clients are OK (I think we have just the `lakectl` command) I see no issues with breaking b/c on some weird-ish edge cases.
2. Note new LGTM.com complaint.
3. Two minor comments.
4. ???
5. THANKS (& PROFIT)!!!

johnnyaug (approved):
Nice. 
In the UI, I still suggest dropping the word "workspace" from the path.

Issue #2061, opened 2021-06-05
rename: modify lakefsFS to comply with s3a handling of rename to an existing destination


talSofer:
The goal of this issue is to modify the LakefsFS rename operation to support the exact same rename scenarios supported by s3a. 

s3a does not support rename operation in cases that the destination is an existing file or a non-empty directory. s3a fails when: 
- [file -> existing file path](https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1539 ) (file -> existing ampty and non-empty dir is supported)
- [dir -> existing file name](https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1526) 
- [dir -> existing non-empty dir](https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1530) 

**In addition to finding the code references this was manually tested against s3.

PR #2062, opened 2021-06-05
handle rename failures similarly to s3a

### LINKED ISSUE #2059: rename: handle rename failures similarly to s3a
The goal of this issue is to modify LakefsFS rename operation to handle rename failure in the same manner s3a handles each case. 

s3a rename split into two parts: 
1. rename initialization: includes [input validation](https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1609)
2. [rename execution](https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1620) 

s3a handles failure in phase 1 by returning false, while failures in phase 2 trigger IOException.

talSofer:
Closes #2059

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -263,6 +263,15 @@ public class LakeFSFileSystem extends FileSystem {
     public boolean rename(Path src, Path dst) throws IOException {
         ObjectLocation srcObjectLoc = pathToObjectLocation(src);
         ObjectLocation dstObjectLoc = pathToObjectLocation(dst);
+        if (srcObjectLoc.getPath().isEmpty()) {
+            LOG.error("rename: src {} is root directory", src);
+            return false;
+        }
+        if (dstObjectLoc.getPath().isEmpty()) {
+            LOG.error("rename: dst {} is root directory", dst);
+            return false;
+        }
+
         if (srcObjectLoc.equals(dstObjectLoc)) {
             LOG.debug("rename: src and dst refer to the same lakefs object location: {}", dst);
             return true;
@@ -278,6 +287,7 @@ public class LakeFSFileSystem extends FileSystem {
         try {
             srcStatus = getFileStatus(src);
         } catch (FileNotFoundException e) {
+            //TODO: throw FNFE, as s3a does in https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1505
             LOG.error("rename: src {} does not exist, rename failed.", src, e);
             return false;
         }
@@ -398,6 +408,7 @@ public class LakeFSFileSystem extends FileSystem {
             objects.stageObject(dstObjectLoc.getRepository(), dstObjectLoc.getRef(), dstObjectLoc.getPath(),
                     creationReq);
         } catch (ApiException e) {
+            //TODO: translate exception to FNFE and Forbidden when access denied AccessDeniedException
             LOG.error("renameObject: Could not stage object on dst:{}", dstObjectLoc.getPath(), e);
             return false;
         }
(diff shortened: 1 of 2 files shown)


clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -263,6 +264,15 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
     public boolean rename(Path src, Path dst) throws IOException {
         ObjectLocation srcObjectLoc = pathToObjectLocation(src);
         ObjectLocation dstObjectLoc = pathToObjectLocation(dst);
+        if (srcObjectLoc.getPath().isEmpty()) {
+            LOG.error("rename: src {} is root directory", src);
```
arielshaqed, line 268: S3A does this?!  Weird, it looks as legitimate a rename as any.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -263,6 +264,15 @@ public FSDataOutputStream append(Path path, int i, Progressable progressable) th
...
+            return false;
+        }
+        if (dstObjectLoc.getPath().isEmpty()) {
+            LOG.error("rename: dst {} is root directory", dst);
```
arielshaqed, line 272: And the same here: S3A doesn't allow me to move all files under `s3a://bucket/oops` to be under `s3a://bucket/`?

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -398,25 +406,37 @@ private boolean renameObject(LakeFSFileStatus srcStatus, Path dst) throws IOExce
...
+        switch (code){
+            case HttpStatus.SC_NOT_FOUND:
+                return new FileNotFoundException(msg);
+            case HttpStatus.SC_FORBIDDEN:
+                return new AccessDeniedException(msg);
+            default:
+                return new IOException(msg);
+        }
+    }
+
```
arielshaqed, line 439: Great stuff!

arielshaqed (commented):
Thanks!  On the fence about approving because of the strange root directory behaviour.

arielshaqed (approved):
After verifying f2f that the weirdnesses are indeed S3A compatible... please just add links to this matching S3A behaviour.  I'm obviously (very) happy with the code.

Thanks!

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
talSofer: yes, https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1498

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
talSofer: As discussed, s3a does not allow renaming a directory when the destination is a non-empty directory. I'm planning to handle this general case at https://github.com/treeverse/lakeFS/issues/2061. 

This line mimics what s3a does in https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1501

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
talSofer: Thanks!

nopcoder (approved):

PR #2065, opened 2021-06-05
Docs: rotate keys example

peacing:


docs/reference/authorization.md
```diff
@@ -275,4 +275,17 @@ Policies: `["FSReadWriteAll", "AuthManageOwnCredentials"]`
...
+
+
+### Rotating Access Credentials
+
+It is best practice to periodically rotate programmatic access credentials. This can be done using CLI commands or the lakeFS UI.
+
+The steps to rotate credentials in the UI are:
+
+1. Go to the Administation tab and click "Create Access Key" to generate a new key.
+1. Next to the old Access Key, click "Revoke".
```
arielshaqed, line 288: Not sure it matters, but this is replacement. To rotate, you would first deploy the new secret key,  then replace it on clients and verify that it works, and only then revoke the old ones. 
Rotation limits your downtime, possibly to 0.

arielshaqed (changes requested):

docs/reference/authorization.md
nopcoder: Updated the PR - changed the order in order to have 0 downtime while rotation.

johnnyaug:
No activity, closing.

PR #2072, opened 2021-06-06
lakefs fs spark test app

nopcoder:
Spark application for testing lakeFS filesystem.
The app process around ~10GB from [Amazon Reviews PDS](https://s3.amazonaws.com/amazon-reviews-pds/readme.html), partition the data by year, write to lakeFS, load the result and dump top 10.

### DIFF
--- a/test/lakefsfs/build.sbt
+++ b/test/lakefsfs/build.sbt
@@ -1,12 +1,74 @@
-name := "lakefsfs"
+import build.BuildType
 
-version := "0.1"
+val baseName = "lakefsfs"
 
-scalaVersion := "2.12.10"
+val projectVersion = "0.1.0"
 
-libraryDependencies ++= Seq(
-    "org.apache.spark" %% "spark-sql" % "3.1.1" % "provided",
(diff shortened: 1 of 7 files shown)


test/lakefsfs/build.sbt
```diff
@@ -0,0 +1,12 @@
+name := "lakefsfs"
+
+version := "0.1"
+
+scalaVersion := "2.12.10"
+
+libraryDependencies ++= Seq(
+    "org.apache.spark" %% "spark-sql" % "3.1.1" % "provided",
```
arielshaqed, line 8: Do we want to go Spark 3.1.1?  I guess we'll see when we ask customers to run it :-)

arielshaqed (approved):
Neat!

test/lakefsfs/build.sbt
nopcoder: Had to run it locally
But I'll check if there is a common way to generate a multi-target solution.

nopcoder: @arielshaqed updated the build to enable the above.

PR #2082, opened 2021-06-08
readme + version

johnnyaug:


clients/spark/README.md
```diff
@@ -0,0 +1,44 @@
...
+      host=treeverse-clients-us-east.s3.amazonaws.com
+      user=<AWS access key>
+      password=<AWS secret key>
+      ```
+1. Increment the version in the build.sbt file.
+
+1. From the lakeFS project root, run:
+   ```bash
+   make publish-scala
+   ```
```
itaiad200, line 44: Please add an example of importing the uber-jar vs the jar.

clients/spark/build.sbt
```diff
@@ -2,7 +2,7 @@ import build.BuildType
 
 lazy val baseName = "lakefs-spark"
 
-lazy val projectVersion = "0.1.0-SNAPSHOT"
+lazy val projectVersion = "0.1.4-SNAPSHOT"
```
itaiad200, line 5: Is there a way to avoid overrides? It's such a common error to publish without advancing this value..

itaiad200 (changes requested):

itaiad200 (approved):

PR #2083, opened 2021-06-08
Generate java client code

nopcoder:
Should fix https://github.com/treeverse/lakeFS/runs/2774417462?check_suite_focus=true


arielshaqed (approved):
# :-( :+1:

PR #2085, opened 2021-06-09
Run all playback sub-tests in parallel

arielshaqed:
On My Machine it reduces time to test gateway from 90s to 60s.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -152,8 +152,8 @@ nessie: ## run nessie (system testing)
 
 test: test-go test-hadoopfs  ## Run tests for the project
 
-test-go: gen
-	$(GOTEST) -count=1 -coverprofile=cover.out -race -cover -failfast $(GO_TEST_MODULES)
+test-go: gen			# Run parallism 4: most of our slow tests are *not* CPU-bound.
+	$(GOTEST) -count=1 -coverprofile=cover.out -race -cover -failfast --parallel 4 $(GO_TEST_MODULES)
 
 test-hadoopfs:
 	cd clients/hadoopfs && mvn test
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,8 @@ GOTESTRACE=$(GOTEST) -race
 GOGET=$(GOCMD) get
 GOFMT=$(GOCMD)fmt
 
+GOTEST_PARALLELISM=4
+
 GO_TEST_MODULES=$(shell $(GOCMD) list ./... | grep -v 'lakefs/pkg/api/gen/')
 
 LAKEFS_BINARY_NAME=lakefs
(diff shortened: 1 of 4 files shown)


arielshaqed:
(Asking for 2 reviews but will happily pull after getting the first, so if you review and  do *not* intend me to pull please be clear :-)

nopcoder (approved):
This is great! amazing that we had unused parameters too

arielshaqed:
Thanks!  Not a *great* speedup, but faster is better.

PTAL, I'm now force-parallelising to 4 because we're not CPU-bound and I suspect GH Actions containers will auto-detect just one core.

pkg/gateway/simulation_playback_test.go
```diff
@@ -44,7 +44,7 @@ type PlaybackParams struct {
 func makePlaybackParams(testDir string) *PlaybackParams {
 	return &PlaybackParams{
 		RecordingDir: filepath.Join(os.TempDir(), "lakeFS", "sourceRecordings", testDir),
-		PlaybackDir: filepath.Join(os.TempDir(), "lakeFS", "gatewayRecordings", time.Now().Format("01-02-15-04-05.000")),
+		PlaybackDir:  filepath.Join(os.TempDir(), "lakeFS", "gatewayRecordings", time.Now().Format("01-02-15-04-05.000")),
```
nopcoder, line 47: extra space

Makefile
```diff
@@ -152,8 +152,8 @@ nessie: ## run nessie (system testing)
 
 test: test-go test-hadoopfs  ## Run tests for the project
 
-test-go: gen
-	$(GOTEST) -count=1 -coverprofile=cover.out -race -cover -failfast $(GO_TEST_MODULES)
+test-go: gen			# Run parallism 4: most of our slow tests are *not* CPU-bound.
+	$(GOTEST) -count=1 -coverprofile=cover.out -race -cover -failfast --parallel 4 $(GO_TEST_MODULES)
```
nopcoder, line 156: can we control this one outside the makefile so in case you run it locally it will use all of our cores?

nopcoder (changes requested):

pkg/gateway/simulation_playback_test.go
arielshaqed: gofmt is happy :shrug:

Makefile
arielshaqed: Yes, but I don't want to.  Our slow tests are not CPU-bound.  There is little to no point in trying to run 1 test on every core.  (Actually there is little to no point in trying to run 1 test on every core even for CPU-bound tests, because some tests are multithreaded!)
If we could attach a "weight" or a set of measures to each test, then it would be possible to schedule them based on the size of the machine.  As it stands, it probably doesn't make much difference what we do here :-/

arielshaqed (commented):
Thanks!

Will roll back the parallelism in `go test` if you don't like it.

Makefile
nopcoder: can you at least set the value based on make variable we can override it

arielshaqed: Done!

arielshaqed:
@nopcoder PTAL...

nopcoder (approved):

nopcoder:
Great!

arielshaqed:
Thanks...

PR #2086, opened 2021-06-09
Handle foreign repos on lakeFS startup

### LINKED ISSUE #2076: Handle foreign repo in lakefs startup
Currently the following example scenario is possible. Any other combination of adapters would behave the same.
1. Starting lakeFS with a `local` block adapter settings. Creating a local repo.
2. Stopping lakeFS and starting it again with an `s3` block adapter settings.

Creating another repo is possible and all operations on it succeed (No surprise here as it was created on S3). Operations on the first repo are allowed but sometimes fail. Listing/Deleting (both metadata) work. Uploading/Downloading are allowed, but return a 500 response with the block adapter (wrong one) error.

We should either allow or disallow repos from different object stores, but decide on a way to block operations instead of returning a 500 response.

itaiad200:
closes #2076

### DIFF
--- a/cmd/lakefs/cmd/run.go
+++ b/cmd/lakefs/cmd/run.go
@@ -20,6 +20,7 @@ import (
 	"github.com/treeverse/lakefs/pkg/api"
 	"github.com/treeverse/lakefs/pkg/auth"
 	"github.com/treeverse/lakefs/pkg/auth/crypt"
+	"github.com/treeverse/lakefs/pkg/block"
 	"github.com/treeverse/lakefs/pkg/block/factory"
 	"github.com/treeverse/lakefs/pkg/catalog"
 	"github.com/treeverse/lakefs/pkg/db"
@@ -110,6 +111,8 @@ var runCmd = &cobra.Command{
 		// send metadata
 		bufferedCollector.CollectMetadata(metadata)
 
+		checkForeignRepos(ctx, logger, authMetadataManager, blockStore, c)
+
 		// update health info with installation ID
 		httputil.SetHealthHandlerInfo(metadata.InstallationID)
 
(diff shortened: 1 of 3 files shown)


cmd/lakefs/cmd/run.go
```diff
@@ -181,6 +184,48 @@ var runCmd = &cobra.Command{
...
 
+// A foreign repo is a repository which namespace doesn't match the current block adapter.
+// A foreign repo might exists if the lakeFS instance configuration changed after a repository was
+// already created. The behaviour of lakeFS for foreign repos is undefined and should be blocked.
+func checkForeignRepos(ctx context.Context, logger logging.Logger, authMetadataManager *auth.DBMetadataManager, blockStore block.Adapter, c *catalog.Catalog) {
+	initialized, err := authMetadataManager.IsInitialized(ctx)
+	if err != nil {
+		logger.WithError(err).Fatal("Failed to read setup time")
+	}
+	if initialized {
```
arielshaqed, line 195: Can you add some logs?  E.g. emit one of these:
* `failed to check if initialized`
* `not initialized, not checking repos`
* `initialized, checking to see repos match block adapter xyzzy`

Also, why won't it work before setup?  I'd expect `listRepositories` to exit with 0 read, meaning the loop ends.  No?

arielshaqed (approved):
This is great, thanks!

Not sure about the term "foreign".  Perhaps "mismatched adapter"?

cmd/lakefs/cmd/run.go
itaiad200: Added the logs.
It wouldn't work because the tables don't exist just yet, so it's a different type of error than ErrNoRows which returns on 0 results.

itaiad200:
> This is great, thanks!
> 
> Not sure about the term "foreign". Perhaps "mismatched adapter"?

Thanks! Changed the logs to use `mismatched adapter` instead of `foreign`.

PR #2088, opened 2021-06-09
add lakefs version to ui

johnnyaug:
Solves #2077


webui/src/lib/components/navbar.jsx
```diff
@@ -38,6 +42,10 @@ const NavUserInfo = () => {
                     }}>
                     Logout
                 </NavDropdown.Item>
+                {!versionLoading && !versionError && <><NavDropdown.Divider/>
```
arielshaqed, line 45: Consider displaying text even for errors or "loading", it can help.

arielshaqed (changes requested):
Thanks!

Requested change is to use objects (not strings) in responses.

api/swagger.yml
```diff
@@ -2831,7 +2831,21 @@ paths:
...
+        - config
+      operationId: getLakeFSVersion
+      description: get version of lakeFS server
+      responses:
+        200:
+          description: lakeFS storage configuration
+          content:
+            application/json:
+              schema:
+                type: string
```
arielshaqed, line 2846: This should be an object.  That way we can add more details, e.g. API versions, plugin versions, etc.

Passing a primitive or an array as the entire JSON response is somewhat similar to "primitive obsession", and can be particularly nasty in APIs.

api/swagger.yml
```diff
@@ -694,7 +694,11 @@ components:
           type: string
         blockstore_namespace_ValidityRegex:
           type: string
-
+    VersionConfig:
```
arielshaqed, line 697: nit:
```suggestion

    VersionConfig:
```

webui/src/lib/components/navbar.jsx
arielshaqed: After f2f discussion, we don't want to do this -- and certainly not piecemeal.

arielshaqed (approved):
Great, thanks!

PR #2090, opened 2021-06-09
Tell GH not to expand generated Python client files by default

arielshaqed:
This is controlled by telling "linguist" that they are generated.


nopcoder (approved):
This is great - thanks!

PR #2099, opened 2021-06-10
Fix spark lakefsfs test app work on EMR

### LINKED ISSUE #2095: Run lakefsfs spark test app on EMR
Explore what it takes to run `test/lakefsfs` under EMR cluster

nopcoder:
Close #2095

- Fix dependencies by building using `hadoop-lakefs-assembly` and not the lakefs java client.
- Create a repository that fails will not stop the test.
- Print the fail result into the log.

### DIFF
--- a/test/lakefsfs/build.sbt
+++ b/test/lakefsfs/build.sbt
@@ -4,6 +4,9 @@ val baseName = "lakefsfs"
 
 val projectVersion = "0.1.0"
 
+ThisBuild / resolvers +=
+  "Sonatype OSS Snapshots" at "https://s01.oss.sonatype.org/content/repositories/snapshots/"
+
 // Spark versions 2.4.7 and 3.0.1 use different Scala versions.  Changing this is a deep
 // change, so key the Spark distinction by the Scala distinction.  sbt doesn't appear to
 // support other ways of changing emitted Scala binary versions using the same compiler.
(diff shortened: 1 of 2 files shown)


test/lakefsfs/src/main/scala/io/lakefs/fs/LakeFSFS.scala
```diff
@@ -36,11 +37,16 @@ object LakeFSFS {
...
-    repositoriesApi.createRepository(repositoryCreation, false)
+    try {
+      val repositoryCreation = new RepositoryCreation()
+        .name(repository)
+        .defaultBranch(branch)
+        .storageNamespace(storageNamespace)
+      repositoriesApi.createRepository(repositoryCreation, false)
+      LOG.info("Repository created repository={}, branch={}, storageNamespace={}", repository, branch, storageNamespace)
+    } catch {
+      case e: ApiException => LOG.error("Create repository failed (code " + e.getCode() + ")", e)
```
arielshaqed, line 48: Please log the actual message, not just the code.

arielshaqed (approved):
Nice work!

test/lakefsfs/build.sbt
```diff
@@ -26,11 +29,10 @@ def generateProject(buildType: BuildType) =
       settingsToCompileIn(),
       scalaVersion := buildType.scalaVersion,
       libraryDependencies ++= Seq(
-        "io.lakefs" % "api-client" % "0.40.3",
+        "io.lakefs" % "hadoop-lakefs-assembly" % "0.1.0-RC.0-SNAPSHOT",
```
arielshaqed, line 32: Can you document in the PR (say in a response here) why this works?  Just in case we run into this again ;-(

test/lakefsfs/src/main/scala/io/lakefs/fs/LakeFSFS.scala
nopcoder: The message is part of Java's Exception, passed it as part of the last argument in the log call.

test/lakefsfs/build.sbt
nopcoder: Updated the build.sbt with a comment on why we used the lakefs-assembly

PR #2100, opened 2021-06-10
Test spark app with spark 2.x

### LINKED ISSUE #2087: Test hadoop-lakefs with Spark 2.4

nopcoder:
Fix #2087

- Compile Sonnets app to 2.11 and 2.12
- Github workflow setup Spark v2.x and Spark v3.x to run the app


arielshaqed (changes requested):
This way of cross-compiling keys the *Spark* dependencies on the *Scala* version.  So it can work for Spark 2.4.7 vs. 3.0.1 -- they switched Scala versions.  But AFAICT it will not work for Spark 3.1.  I.e. in order to fulfill our known goals (support Spark 3.1) I think that we will need to go back to the version already used in the Spark metadata client.

So let's not do this work twice...

test/spark/app/build.sbt
```diff
@@ -1,11 +1,69 @@
...
+      scalaVersion := buildType.scalaVersion,
+      libraryDependencies ++= Seq(
+        "org.apache.spark" %% "spark-sql" % buildType.sparkVersion % "provided",
+        "org.apache.hadoop" % "hadoop-aws" % buildType.hadoopVersion,
+        "org.apache.hadoop" % "hadoop-common" % buildType.hadoopVersion,
+      ),
+      target := { baseDirectory.value / "target" / s"${baseName}-${buildType.name}" }
+    )
+
+lazy val proj24 = generateProject(new BuildType("246", scala211Version, "2.4.6", "2.7.7"))
```
arielshaqed, line 36: This should probably be 2.4.7, because that's newer and [EMR 5.33.0 uses it](https://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-spark.html#w205aac53c25).

test/spark/app/build.sbt
```diff
@@ -1,11 +1,69 @@
...
+      compile / skip := true,
+      assembly / skip := true,
+      publish / skip := true,
+  )
+
+
+lazy val assemblySettings = Seq(
+  assembly / assemblyMergeStrategy := (_ => MergeStrategy.first),
+  assembly / assemblyShadeRules := Seq(
+    ShadeRule.rename("scala.collection.compat.**" -> "shadecompat.@1").inAll,
```
arielshaqed, line 51: Strange that we need to shade _this_: I would expect it to be a part of the Scala runtime.

.github/workflows/nessie.yaml
```diff
@@ -182,7 +183,7 @@ jobs:
 
       - name: Package Spark App
         working-directory: test/spark/app
-        run: sbt package
+        run: sbt ++2.11.12 package
```
arielshaqed, line 186: Does this work?
```suggestion
        run: sbt proj24/package
```

arielshaqed (approved):
Cool, thanks!

It's good because it works, but I would prefer some minor changes (particularly if it still works with those changes...).

test/spark/app/build.sbt
nopcoder: aligned with the docker compose we use. this is the sonnet app.

test/spark/app/build.sbt
nopcoder: right. I'll remove all shading

.github/workflows/nessie.yaml
nopcoder: ++2.11.12 is ignored.
going with `sbt sonnets-246/package`

test/spark/app/build.sbt
nopcoder: and assembly

PR #2101, opened 2021-06-11
Auth System Tests (#657)

### LINKED ISSUE #657: Auth system test
The current system tests use the default admin user for all requests.

The auth test should create users with less privileges and check actions are passing or blocked according to the permissions.

adunuthulan:
Fixes #657 

Created system tests using different clients, each with the policies of the user groups Developers, SuperUsers, and Viewers. Each test checks to see if the client correctly responds to the user's assigned policies. I plan on adding merging to the set of tests.

### DIFF
--- a/nessie/auth_test.go
+++ b/nessie/auth_test.go
@@ -1,0 +1,167 @@
+package nessie
+
+import (
+	"context"
+	"net/http"
+	"net/url"
+	"strings"
+	"testing"
+
+	"github.com/deepmap/oapi-codegen/pkg/securityprovider"
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/require"
+	"github.com/treeverse/lakefs/pkg/api"
+)
+
+// Test Developer Policies: AuthManageOwnCredentials, FSFullAccess, RepoManagementReadAll
+func TestDeveloperPolicies(t *testing.T) {
+
+	ctx, _, repo := setupTest(t)
+
+	gid := "Developers"
+	endpointURL := viper.GetString("endpoint_url") // defined in setup.go
+
+	// generate the Developer client
+	developerClient, err := NewClientFromGroup(t, ctx, gid, endpointURL)
+	require.NoError(t, err, "failed to initialize client with group")
+
+	// listing the available braches should succeed
+	resBch, err := developerClient.ListBranchesWithResponse(ctx, repo, &api.ListBranchesParams{})
+	require.NoError(t, err, "Developer unexpectedly failed to list branches of repository")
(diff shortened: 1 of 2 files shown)


nessie/auth_test.go
```diff
@@ -0,0 +1,167 @@
...
+		Name:   branch1,
+		Source: mainBranch,
+	})
+	require.NoError(t, err, "SuperUser failed while testing create branch")
+	require.Equal(t, http.StatusCreated, resAddBch.StatusCode(), "SuperUser unexpectedly failed to create branch")
+
+	// attempting to delete the repository should be authorized and result in no content response
+	resDelRep, err := superUserClient.DeleteRepositoryWithResponse(ctx, repo)
+	require.NoError(t, err, "SuperUser failed while testing delete repository")
+	require.Equal(t, http.StatusNoContent, resDelRep.StatusCode(), "SuperUser unexpectedly did not receive no content response while deleting repo")
```
itaiad200, line 90: The admin can do so much more, like managing groups. Let's add more checks for those as well.

itaiad200 (changes requested):
Really enjoyed reviewing this code, it is clean, concise and well-documented. The PR description is also very helpful in understanding where this stands.

Feel free to undraft for next iteration after fixing the comments and adding the tests that you wrote.

itaiad200 (approved):
Thanks for fixing the comments. Approved

pkg/testutil/setup.go
```diff
@@ -58,15 +58,7 @@ func SetupTestingEnv(params *SetupTestingEnvParams) (logging.Logger, api.ClientW
...
-	endpointURL := viper.GetString("endpoint_url")
-	u, err := url.Parse(endpointURL)
-	if err != nil {
-		logger.WithError(err).Fatal("Failed to parse endpoint URL", endpointURL)
-	}
-
-	if u.Path == "" || u.Path == "/" {
-		endpointURL = strings.TrimRight(endpointURL, "/") + api.BaseURL
-	}
+	endpointURL := ParseEndpointURL(logger, viper.GetString("endpoint_url"))
```
arielshaqed, line 61: Hurrah! :confetti_ball:

arielshaqed (commented):
Thanks for testing a really important part of the system!  This is a really good test.

I am asking for some refactoring it to a more data-driven style.  That is the Go Way, but more importantly I believe that it has significant advantages here in making it easier to review and maintain the code.

# There is :snail:  *NO URGENCY* :snail: in doing this!  The repo is better with this test than without it!

If you like I'd be very happy to go over how we might improve maintainability here.  Let me know :-)

Thanks!

PR #2102, opened 2021-06-13
Modify lakefsFS to comply with s3a handling of rename to an existing destination

### LINKED ISSUE #2061: rename: modify lakefsFS to comply with s3a handling of rename to an existing destination

talSofer:
closes #2061

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -311,9 +311,9 @@ public class LakeFSFileSystem extends FileSystem {
             if (!dstFileStatus.isDirectory()) {
                 // Same as https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1527
                 throw new FileAlreadyExistsException("Failed rename " + src + " to " + dst
-                        + ". destination file already exists.");
+                        + "; source is a directory and dest is a files");
             }
-            // lakefsFs only has non-empty directories. therefore, is the destination is an existing directory we consider
+            // lakefsFs only has non-empty directories. Therefore, if the destination is an existing directory we consider
             // it to be non-empty. The behaviour is same as https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1530
             LOG.error("renameDirectory: failed to rename src {} to dst {}. dst is a non-empty directory.", src, dst);
             return false;
@@ -386,7 +386,7 @@ public class LakeFSFileSystem extends FileSystem {
             } else {
                 // Same as https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1539
                 throw new FileAlreadyExistsException("Failed rename " + srcStatus.getPath() + " to " + dst
-                        + ". destination file already exists.");
+                        + "; destination file already exists.");
             }
         } catch (FileNotFoundException e) {
             LOG.debug("renameFile: dst does not exist, renaming src {} to a file called dst {}",
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -309,10 +309,9 @@ public class LakeFSFileSystem extends FileSystem {
             dstFileStatus = getFileStatus(dst);
             dstExists = true;
             if (!dstFileStatus.isDirectory()) {
-                //TODO (Tals): move all of this to the input validation part of the rename functionality with https://github.com/treeverse/lakeFS/issues/2061
-                // Mimics the behaviour in https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1527
-                LOG.error("renameDirectory: src {} is a directory and dst {} is a file", src, dst);
-                return false;
+                // Same as https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1527
+                throw new FileAlreadyExistsException("Failed rename " + src + " to " + dst
+                        + ". destination file already exists.");
             }
         } catch (FileNotFoundException e) {
             LOG.debug("renameDirectory: dst {} does not exist", dst);
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -313,6 +313,10 @@ public class LakeFSFileSystem extends FileSystem {
                 throw new FileAlreadyExistsException("Failed rename " + src + " to " + dst
                         + ". destination file already exists.");
             }
+            // lakefsFs only has non-empty directories. therefore, is the destination is an existing directory we consider
+            // it to be non-empty. The behaviour is same as https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1530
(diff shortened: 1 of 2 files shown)


clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -309,11 +309,14 @@ private boolean renameDirectory(Path src, Path dst) throws IOException {
             dstFileStatus = getFileStatus(dst);
             dstExists = true;
             if (!dstFileStatus.isDirectory()) {
-                //TODO (Tals): move all of this to the input validation part of the rename functionality with https://github.com/treeverse/lakeFS/issues/2061
-                // Mimics the behaviour in https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1527
-                LOG.error("renameDirectory: src {} is a directory and dst {} is a file", src, dst);
-                return false;
+                // Same as https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1527
+                throw new FileAlreadyExistsException("Failed rename " + src + " to " + dst
```
arielshaqed, line 313: Why not copy the error message from there?

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -309,11 +309,14 @@ private boolean renameDirectory(Path src, Path dst) throws IOException {
...
+                        + ". destination file already exists.");
             }
```
arielshaqed, line 315: Failing that copy, The S3A Way seems to be to use a semicolon (and along the way avoid the strange noncapitalization of "destination").

```suggestion
                // Same as https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1527
                throw new FileAlreadyExistsException("Failed rename " + src + " to " + dst
                        + "; destination file already exists.");
            }
```

Also below.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -309,11 +309,14 @@ private boolean renameDirectory(Path src, Path dst) throws IOException {
...
+            // lakefsFs only has non-empty directories. therefore, is the destination is an existing directory we consider
+            // it to be non-empty. The behaviour is same as https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1530
```
arielshaqed, line 317: I read that code and I don't understand: https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1531 throws a `RenamedFailedException`, here we report an error and fail the operation by returning false.

arielshaqed (changes requested):
Thanks!  This is really makes me feel safer about supporting everything that S3A does.

Not sure why one of the behaviours is correct; also want our exception messages to be even closer to the ones in S3A.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
talSofer: done

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
talSofer: If you look into https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1471 you can see how s3a rename operation handles RenameFailedExceptions. It returns its exit code which is false in case of dir -> non-empty dir rename.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -309,11 +309,14 @@ private boolean renameDirectory(Path src, Path dst) throws IOException {
             dstFileStatus = getFileStatus(dst);
             dstExists = true;
             if (!dstFileStatus.isDirectory()) {
-                //TODO (Tals): move all of this to the input validation part of the rename functionality with https://github.com/treeverse/lakeFS/issues/2061
-                // Mimics the behaviour in https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1527
-                LOG.error("renameDirectory: src {} is a directory and dst {} is a file", src, dst);
-                return false;
+                // Same as https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1527
+                throw new FileAlreadyExistsException("Failed rename " + src + " to " + dst
+                        + "; source is a directory and dest is a files");
```
arielshaqed, line 314: nit:
```suggestion
                        + "; source is a directory and dest is a file");
```

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
arielshaqed: Ouch!

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -309,11 +309,14 @@ private boolean renameDirectory(Path src, Path dst) throws IOException {
...
             }
+            // lakefsFs only has non-empty directories. Therefore, if the destination is an existing directory we consider
+            // it to be non-empty. The behaviour is same as https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1530
+            LOG.error("renameDirectory: failed to rename src {} to dst {}. dst is a non-empty directory.", src, dst);
```
arielshaqed, line 318: This is still with really weird punctuation and rather unlike the error message that I would expect from https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/PathIOException.java#L90.  Think that this should be something like
```suggestion
            LOG.error("renameDirectory: rename src {} to dst {}: non-empty directory.", src, dst);
```

Generally, "failed to..." is an uncommon phrasing for error messages, it's more common to say "WHAT: REASON".

arielshaqed (approved):
Thanks!  But please do go over the comments about error messages again...

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
talSofer: missed this I will address this is the logging review I'm planning, sorry about that.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
talSofer: missed this I will address this is the logging review I'm planning, sorry about that.

talSofer: fixed in https://github.com/treeverse/lakeFS/pull/2110/files

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
talSofer: fixed in https://github.com/treeverse/lakeFS/pull/2110/files

PR #2122, opened 2021-06-14
Reduce the output level of mvn on lakefsfs build

nopcoder:


clients/hadoopfs/src/test/resources/log4j.properties
```diff
@@ -1,5 +1,5 @@
-# Set root logger level to DEBUG and its only appender to A1.
-log4j.rootLogger=DEBUG, A1
+# Set root logger level to INFO and its only appender to A1.
+log4j.rootLogger=INFO, A1
```
arielshaqed, line 2: Not a fan of this change.  It leaves _only_ the boring (and many...) log lines.  But our code actually emits a lot of interesting messages at DEBUG level.

arielshaqed (changes requested):
But I *like* the added lines at DEBUG level.  Why remove them?  I claim that either you ignore them (and then you don't care), or you're looking for something -- and then the extra information does help (it helped me while writing the tests).

clients/hadoopfs/src/test/resources/log4j.properties
nopcoder: Added DEBUG level to io.lakefs

arielshaqed (approved):
Thanks!

PR #2126, opened 2021-06-15
Allow mismatch repos advance hidden flag

### LINKED ISSUE #2119: Allow mismatched adapter usage as advanced flag
Advanced lakeFS users would want to use multiple repos from multiple object stores.
Should have some sort of silent-flag that will allow it.

itaiad200:
close #2119


arielshaqed (approved):
Neat and useful.  Thanks!

PR #2129, opened 2021-06-16
Add GitHub action for Spark

guy-har:
This PR adds a GitHub action that validates:

The Spark projects compile correctly
The Spark projects are formatted correctly
The spark projects do not have unused imports
The plugins `scalafmt` and `scalafix` were added.
In order to remove unused imports run:
```
sbt scalafix
```
In order to format your code run:
```
sbt scalafmt 
```
Currently there aren't much formatting configurations (as you could see in `.scalafmt.conf`), and only one fix rule which is RemoveUnused (as you could see in `.scalafix.conf`)


guy-har:
This PR adds a GitHub action that validates:
 1. The Spark projects compile correctly 
 2. The Spark projects are formatted correctly 
 3. The spark projects do not have unused imports

The plugins `scalafmt` and `scalafix` were added.
In order to remove unused imports run:
```
sbt scalafix
```
In order to format your code run:
```
sbt scalafmt 
```

Currently there aren't much formatting configurations (as you could see in `.scalafmt.conf`), and only one fix rule which is RemoveUnused (as you could see in `scalafix.conf`)

clients/spark/project/plugins.sbt
```diff
@@ -2,3 +2,5 @@ addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.15.0")
 addSbtPlugin("com.github.sbt" % "sbt-pgp" % "2.1.2")
 addSbtPlugin("com.frugalmechanic" % "fm-sbt-s3-resolver" % "0.19.0")
 addSbtPlugin("cf.janga" % "sbts3" % "0.10.5")
+addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.29")
+addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.0")
```
johnnyaug, line 6: newline

clients/spark/.scalafmt.conf
```diff
@@ -1,2 +1,2 @@
 version = 2.5.0
-trailingCommas = always
\ No newline at end of file
+maxColumn = 100
```
johnnyaug, line 2: newline

johnnyaug (approved):
Lovely.

.github/workflows/spark.yaml
```diff
@@ -0,0 +1,28 @@
+name: spark
```
johnnyaug, line 1: naming: "spark" is too general, I would call this "spark metadata client checks"

johnnyaug: not necessarily in this field, I just want it to be easier to understand where it appears on github:
image

arielshaqed (approved):
Thanks!  This is obviously an improvement.

Personally I find many parts of the new style *less* readable.  I'd rather if we could change the formatter to allow it -- specifics inside.

.github/workflows/spark.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+      - name: package
+        working-directory: clients/spark
+        run: sbt package
+      - name: validate format
+        working-directory: clients/spark
+        run: sbt scalafmtCheck
+      - name: validate unused
+        working-directory: clients/spark
+        run: sbt "scalafix --check"
```
nopcoder, line 27: usually, we run the lint steps before the compile/test step

nopcoder (approved):
Cool! Just a minor comment about the step location.

.github/workflows/spark.yaml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check-out code
+        uses: actions/checkout@v2
+      - name: Setup Scala
+        uses: olafurpg/setup-scala@v10
+
+      - name: package
+        working-directory: clients/spark
+        run: sbt package
```
arielshaqed, line 21: Can we compile twice, once with 2.11 and once with 2.12?  Because we *do* need to support both.  (I think all Scala compilers take a language version command-line flag, so you will *not* have to install a separate Scala compiler.)

clients/spark/build.sbt
```diff
@@ -31,6 +31,9 @@ def generateCoreProject(buildType: BuildType) =
       s3UploadSettings,
       settingsToCompileIn("core"),
       scalaVersion := buildType.scalaVersion,
+      semanticdbEnabled := true, // enable SemanticDB
+      semanticdbVersion := scalafixSemanticdb.revision,
```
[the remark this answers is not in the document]
guy-har: scalafix uses semanticDB, AFAICT its just extra information.

```
SemanticDB decouples producers and consumers of semantic information about programs and establishes a rigorous specification of the interchange format.

Thanks to that, SemanticDB-based tools like Scalafix, Metadoc and Metals don't need to know about compiler internals and can work with any compiler that supports SemanticDB.
```
Taken from the semanticDB documentation

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -28,18 +33,22 @@ class ApiClient(apiUrl: String, accessKey: String, secretKey: String) {
   private val metadataApi = new api.MetadataApi(client)
   private val branchesApi = new api.BranchesApi(client)
 
-  private val storageNamespaceCache = CacheBuilder.newBuilder().expireAfterWrite(2, TimeUnit.MINUTES).build[String, String]()
+  private val storageNamespaceCache =
+    CacheBuilder.newBuilder().expireAfterWrite(2, TimeUnit.MINUTES).build[String, String]()
```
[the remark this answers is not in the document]
guy-har: It would've done it in case `CacheBuilder.newBuilder().expireAfterWrite(2, TimeUnit.MINUTES).build[String, String]()` was longer than 100 (including the spaces before)

PR #2131, opened 2021-06-16
Retry DB Connection on Postgres Error

### LINKED ISSUE #2017: Retry DB connection on startup
Got [this failure](https://github.com/treeverse/lakeFS/pull/2007/checks?check_run_id=2703565284#step:7:88):
```
lakefs_1    | panic: could not open DB: failed to connect to `host=postgres user=lakefs database=postgres`: server error (FATAL: the database system is starting up (SQLSTATE 57P03))
```
This looks like a race during startup, so need to retry _at least on this error_. This will also be useful for simplifying system automation.

adunuthulan:
Fixes #2017 

Retries pgxpool.ConnectConfig when encountering any error connecting to the DB. I use exponential backoff with a default of 6 tries. Should catch `FATAL: the database system is starting up` error.

### DIFF
--- a/pkg/db/connect.go
+++ b/pkg/db/connect.go
@@ -3,6 +3,7 @@ package db
 import (
 	"context"
 	"fmt"
+	"math"
 	"time"
 
 	"github.com/jackc/pgx/v4/pgxpool"
@@ -15,6 +16,8 @@ const (
 	DefaultMaxIdleConnections    = 25
 	DefaultConnectionMaxLifetime = 5 * time.Minute
 	DatabaseDriver               = "pgx"
+	DefaultMaxDatabaseRetries    = 6
+	RetryBase                    = 2
 )
 
 // BuildDatabaseConnection returns a database connection based on a pool for the configuration
@@ -60,9 +63,21 @@ func ConnectDBPool(ctx context.Context, p params.Database) (*pgxpool.Pool, error
 	config.MaxConnLifetime = p.ConnectionMaxLifetime
 
 	pool, err := pgxpool.ConnectConfig(ctx, config)
-	if err != nil {
-		return nil, fmt.Errorf("could not open DB: %w", err)
+	for i := 0; i < DefaultMaxDatabaseRetries; i++ {
+		if err != nil {
+			if i != DefaultMaxDatabaseRetries-1 {
+				waitTime := math.Pow(RetryBase, float64(i))
+				log.Info(fmt.Sprintf("could not open DB: trying again in %ds", int(waitTime)))
+				time.Sleep(time.Duration(waitTime) * time.Second)
+				pool, err = pgxpool.ConnectConfig(ctx, config)
+
+				continue
+			} else {
+				return nil, fmt.Errorf("could not open DB: %w", err)
+			}
+		}
 	}
+
 	err = Ping(ctx, pool)
 	if err != nil {
 		pool.Close()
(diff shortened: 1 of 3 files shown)


pkg/db/connect.go
```diff
@@ -60,9 +63,21 @@ func ConnectDBPool(ctx context.Context, p params.Database) (*pgxpool.Pool, error
...
+				log.Info(fmt.Sprintf("could not open DB: trying again in %ds", int(waitTime)))
+				time.Sleep(time.Duration(waitTime) * time.Second)
+				pool, err = pgxpool.ConnectConfig(ctx, config)
+
+				continue
+			} else {
+				return nil, fmt.Errorf("could not open DB: %w", err)
+			}
+		}
 	}
```
arielshaqed, line 79: This block is a bit hard to understand, possibly because you try to connect in one iteration but test the result in the next.  That also manifests in having to call `ConnectConfig` on two lines of code.  Also it means that you have the strange return just before the loop exits on l.76.

You might write the loop more directly:
```suggestion
    	for i := 0; i < DefaultMaxDatabaseRetries; i++ {
    	   	pool, err := pgxpool.ConnectConfig(ctx, config)
    	   	if err == nil {
    	   	    	return pool, nil
    	    	}
    	    	if i < DefaultMaxDatabaseRetries-1 {
    	    	    	wait := time.Duration(math.Pow(RetryBase, float64(i))) * time.Second
    	    	    	log.Info("open DB: %s.  Trying again in %s", wait)
    			time.Sleep(wait)
    		}
    	}
    	return nil, fmt.Errorf("open DB: %w", err)
```

And I would probably change the wait to
```go
wait := firstWait * pow(waitGrowth, i)
```
for some `const firstWait time.Duration` and `const waitGrowth float64`.  I like to retry aggressively at first (wait a WHOLE SECOND?), so e.g.
```go
const (
    	firstWait = 50 * time.MilliSecond
    	waitGrowth = 1.2
    	numRetries = /* whatever makes it fail within 3 seconds */
)
```
*Now* we can go all the way, and replace `numRetries` with a `totalWait` parameter so that we can say "just repeat this, with exponential backoff, I'm willing to wait up to `totalWait`.

Alternatively, what I like to do is have a "backoff iterator" that tells me whether to go again and how long to wait before I do.  E.g. I found this "retry" package has https://pkg.go.dev/gopkg.in/retry.v1.

arielshaqed (approved):
Cool, this is good and will help us by making our tests flake less often!

Unfortunately this is the perfect PR for me to air my strongly-held views both about backing off _and_ about iteration. :-(  Not required for pulling, although I believe that at least the first suggested loop is easier (for me) to read.

pkg/db/connect.go
adunuthulan: Thank you for the comments Ariel! I really appreciate the advice and the chance to improve my programming. If it's okay, I'll commit these changes and then ask for another review!

arielshaqed (approved):
Thanks, this is really great stuff!

You write:
> Thank you for the comments Ariel! I really appreciate the advice and the chance to improve my programming. If it's okay, I'll commit these changes and then ask for another review!

This PR was of course approved (and remains approved).  Sorry for not having said it earlier: marking "approved" means just that -- as a reviewer I believe that this code belongs in the repo.  So obviously any comments on an approved PR are requests that I explicitly say are optional.  You are welcome to another review round if you like, but you are also welcome to pull the PR.

PR #2141, opened 2021-06-20
Restore Azure system test

guy-har:


arielshaqed (approved):
Thanks!

If it passes then it's good :-)

PR #2154, opened 2021-06-22
Add Spark Job for generating garbage collection addresses

guy-har:


clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -0,0 +1,233 @@
...
+    val activeRangesDF = ranges.where("!expired")
+
+    // ranges existing in expired and not in active
+    val uniqueExpiredRangesDF = expiredRangesDF.join(
+      activeRangesDF,
+      expiredRangesDF("rangeID") === activeRangesDF("rangeID"),
+      "leftanti"
+    )
+
+    // for every expired range - get all active ranges with Intersecting range
```
johnnyaug, line 155:
```suggestion
    // for every expired range - get all intersecting active ranges
```

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -0,0 +1,233 @@
...
+          ((col("a.min_key") <= col("u.max_key")) && (col("a.min_Key") >= col("u.min_key"))),
+        "left"
+      )
+      .select(
+        col("u.rangeID").as("unique_range"),
+        col("a.rangeID").as("active_range"),
+        functions.concat(col("u.min_key"), col("u.max_key")).as("min-max")
+      )
+
+    val groupByMinMax = joinActiveByRange
```
johnnyaug, line 171: Add comment: what this group by does

johnnyaug (changes requested):
Wow! I actually enjoyed reading this.

Most of my comments are about naming - in many places where you use "address" it actually means entry. Also there are some one-letter variables which are a little hard to

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -0,0 +1,233 @@
...
+
+  private def getRangeTuples(
+      commitID: String,
+      repo: String,
+      conf: APIConfigurations
+  ): Set[(String, String, String)] = {
+    val location =
+      new ApiClient(conf.apiURL, conf.accessKey, conf.secretKey).getMetaRangeURL(repo, commitID)
+    SSTableReader
+      .forRange(new Configuration(), location)
```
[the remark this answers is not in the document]
guy-har: @johnnyaug Unfortunately thats the only the way to get Configuration that works when calling from UDF.

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -0,0 +1,233 @@
...
+object GarbageCollector {
+
+  case class APIConfigurations(apiURL: String, accessKey: String, secretKey: String)
+
+  def getCommitsDF(runID: String, commitDFLocation: String, spark: SparkSession): Dataset[Row] = {
+    spark.read
+      .option("header", value = true)
+      .option("inferSchema", value = true)
+      .csv(commitDFLocation)
+      .where(col("runID") === runID)
```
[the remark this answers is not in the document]
guy-har: Actually, I think it's better.
changing all columns to be in that convention

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -0,0 +1,233 @@
...
+    val commitsDF = getCommitsDF(runID, commitDFLocation, spark)
+    val rangesDF = getRangesDFFromCommits(commitsDF, repo, conf)
+    getAddressesDFFromRanges(rangesDF, conf)
+  }
+
+  def main(args: Array[String]) {
+    val spark = SparkSession.builder().getOrCreate()
+    if (args.length != 4) {
+      Console.err.println(
+        "Usage: ... <repo_name> <runID> s3://storageNamespace/addressesDFLocation s3://storageNamespace/addressesDFLocation s3://path/to/output/du"
```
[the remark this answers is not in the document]
guy-har: @johnnyaug I agree, though I think it could be in a different PR

pkg/pyramid/tier_fs.go
```diff
@@ -158,7 +159,8 @@ func (tfs *TierFS) store(ctx context.Context, namespace, originalPath, nsPath, f
 		return fmt.Errorf("adapter put %s %s: %w", namespace, filename, err)
 	}
 
-	if err := f.Close(); err != nil {
+	var target = os.ErrClosed
```
johnnyaug, line 162: oopsy

guy-har: oops

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -0,0 +1,244 @@
...
+      repo: String,
+      conf: APIConfigurations
+  ): Set[(String, String, String)] = {
+    val location =
+      new ApiClient(conf.apiURL, conf.accessKey, conf.secretKey).getMetaRangeURL(repo, commitID)
+    SSTableReader
+      .forMetaRange(new Configuration(), location)
+      .newIterator()
+      .map(range =>
+        (new String(range.id), range.message.minKey.toStringUtf8, range.message.maxKey.toStringUtf8)
```
arielshaqed, line 38: Thinking about it, this fails for invalid UTF-8 keys.  Per [this Oracle blog](https://blogs.oracle.com/corejavatechtips/overhauling-the-java-utf-8-charset) not all byte sequences are parsed as legal JVM Strings, so this may be an issue.  (It *is* possible to load invalid UTF-8 keys into lakeFS AFAIK, because by contrast the [Go ref](https://golang.org/ref/spec#String_types) defines strings as "_a (possibly empty) sequence of bytes_", without enforcing morality.

Can we keep these as bytes?

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -0,0 +1,244 @@
...
+    )
+
+    // for every expired range - get all intersecting active ranges
+    //  expired_range | active_ranges  | min-max (of expired_range)
+    val joinActiveByRange = uniqueExpiredRangesDF
+      .as("u")
+      .join(
+        activeRangesDF.as("a"),
+        (col("a.max_key") <= col("u.max_key")) && (col("a.max_key") >= col("u.min_key")) ||
+          ((col("a.min_key") <= col("u.max_key")) && (col("a.min_Key") >= col("u.min_key"))),
```
arielshaqed, line 170:
```suggestion
          ((col("a.min_key") <= col("u.max_key")) && (col("a.min_key") >= col("u.min_key"))),
```

arielshaqed, line 170: I find it easier to read "between" checks that look more like the mathematical formulation "a <= b <= c", and also many fewer brackets, e.g.
```suggestion
        col("u.min_key") <= col("a.max_key") && col("a.max_key") <= col("u.max_key") ||
        col("u.min_key") <= col("a.min_key") && col("a.min_key") <= col("u.max_key")),
```

If you do need those extra brackets for clarity, I would place them one level up, i.e. replacing "A && B || C && D" above with "(A && B) || (C && D)" (the original code is "(A) && (B) || (C) && (D)", which I feel only adds noise).

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -0,0 +1,244 @@
...
+
+    // ranges existing in expired and not in active
+    val uniqueExpiredRangesDF = expiredRangesDF.join(
+      activeRangesDF,
+      expiredRangesDF("range_id") === activeRangesDF("range_id"),
+      "leftanti"
+    )
+
+    // for every expired range - get all intersecting active ranges
+    //  expired_range | active_ranges  | min-max (of expired_range)
```
arielshaqed, line 164: This does _not_ match what I think of when I read "intersecting".  E.g., if u is entirely contained in a then `u.max_key < a.max_key`, so the first comparison on l. 169 is false, and `a.min_key < u.min_key`, so the second comparion on l. 170 is also false.  Is the intent here some kind of "overlapping but not contained"?

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -0,0 +1,244 @@
...
+      .join(
+        activeRangesDF.as("a"),
+        (col("a.max_key") <= col("u.max_key")) && (col("a.max_key") >= col("u.min_key")) ||
+          ((col("a.min_key") <= col("u.max_key")) && (col("a.min_Key") >= col("u.min_key"))),
+        "left"
+      )
+      .select(
+        col("u.rangeID").as("unique_range"),
+        col("a.rangeID").as("active_range"),
+        functions.concat(col("u.min_key"), col("u.max_key")).as("min-max")
```
arielshaqed, line 176: This is unsafe because it is not possible to invert the transformation.  E.g. when `min_max == "abcabd` we could have `u.min_key == "abc" && u.max_key == "abd"`, but we might alternatively have that `u.min_key == "ab" && u.max_key == "cabd"` or that `u.min_key == "abca" && u.max_key == "bd"`.

In general it is really tricky safely to transform 2 strings into a single string.  Can you write the _use_ of `min_max` to avoid this by just using two fields?

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -0,0 +1,244 @@
...
+    val apiURL = hc.get(LAKEFS_CONF_API_URL_KEY)
+    val accessKey = hc.get(LAKEFS_CONF_API_ACCESS_KEY_KEY)
+    val secretKey = hc.get(LAKEFS_CONF_API_SECRET_KEY_KEY)
+    val expiredAddresses = getExpiredAddresses(repo,
+                                               runID,
+                                               commitDFLocation,
+                                               spark,
+                                               APIConfigurations(apiURL, accessKey, secretKey)
+                                              ).withColumn("run_id", lit(runID))
+    expiredAddresses.write.partitionBy("run_id").mode(SaveMode.Append).parquet(addressesDFLocation)
```
arielshaqed, line 242: I don't understand the `partitionBy` clause.  AFAICS column `"run_id"` is the constant value `runID`.  So what is this partition?  Why do we even need it?

arielshaqed (changes requested):
Thanks!

Sorry, much lack of understanding of the code on my part.

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -0,0 +1,233 @@
...
+}
+import org.apache.hadoop.conf.Configuration
+import org.apache.spark.sql.functions._
+import org.apache.spark.sql.{SparkSession, _}
+
+import scala.annotation.tailrec
+
+object GarbageCollector {
+
+  case class APIConfigurations(apiURL: String, accessKey: String, secretKey: String)
```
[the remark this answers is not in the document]
guy-har: @arielshaqed, its configurations for our clientAPI, should I change the name?
Also, I assume that passing a AWS auth class would throw `Not Serializable`

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -0,0 +1,233 @@
...
+  /** receives a dataframe containing active and expired ranges and returns addresses contained only in expired ranges
+   *  @param ranges dataframe of type   rangeID:String | expired: Boolean
+   *  @return dataframe of type  key:String | address:String | relative:Boolean | last_modified:Long
+   */
+  def getAddressesDFFromRanges(ranges: Dataset[Row], conf: APIConfigurations): Dataset[Row] = {
+    val left_anti_join_addresses = udf((x: Seq[String], y: Seq[String]) => {
+      leftAntiJoinAddresses(x.toSet, y.toSet, conf).toSeq
+    })
+    val expiredRangesDF = ranges.where("expired")
+    val activeRangesDF = ranges.where("!expired")
```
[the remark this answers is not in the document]
guy-har: No, It's equivalent to `select where`

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
guy-har: In order to write the data partitioned in The object store (i.e under `s3://path/run_id=1234/...`)

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -0,0 +1,233 @@
...
+   *  @param rightRangeIDs
+   *  @param conf
+   *  @return tuples of type (key, address, isRelative, lastModified) for every address existing in leftRanges and not in rightRanges
+   */
+  def leftAntiJoinAddresses(
+      leftRangeIDs: Set[String],
+      rightRangeIDs: Set[String],
+      conf: APIConfigurations
+  ): Set[(String, String, Boolean, Long)] = {
+    @tailrec
```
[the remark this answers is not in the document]
guy-har: Changed, Thanks!

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
guy-har: Thanks, fixing it!

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
guy-har: 👍

johnnyaug (approved):

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
arielshaqed: So IIUC this is "just" a shortcut for naming the files as though they belonged to a partition, even though there's no partition here?  Not enough of a Spark person to know whether users will expect this -- so let's try and see!

arielshaqed (changes requested):
Thanks!  Looks cool except for the invalid UTF-8 issue.  I'd rather fix that before releasing: it's OK 99% of the time, until someone has such keys, and then it's very bad 100% of the time :-(

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
guy-har: Yes the parquet table in `addressesDFLocation` is partitioned by `run_id` the results of this run should be under the corresponding partition.

arielshaqed (approved):
Yay!  Thanks...

Issue #2068, opened 2021-06-06
[GC] Handle dangling commits


johnnyaug:
Our [algorithm](https://github.com/treeverse/lakeFS/issues/1932) doesn't handle commits which are not reachable from any branch HEAD. Decide what is the retention rule for dangling commits, and make sure they are also expired.

Issue #1932, opened 2021-05-10
Deleted Object Retention - Committed


johnnyaug:
## User story
Users should be able specify the period of time objects will be available after deleted. The user will be able to define a default period, and a period per branch.

## Requirements
* A cleanup job can be manually/periodically triggered.
* After the cleanup job has run:
  * An object will exist in the storage *iff* it existed in the HEAD of some branch within the retention period of that branch.
  * In particular, data existing only on branches that were deleted prior to the process should be deleted.
* Objects outside of storage namespace (e.g ingested data) should not be deleted.
* Staging data (of all branches) should not be deleted.

### Simple example: cleanup on the main branch
Consider the following history on the main branch:
![image](https://user-images.githubusercontent.com/3149885/117674250-2ea92a00-b1b4-11eb-83b2-cb976321bba9.png)

Suppose the retention period is 7 days, and we are running the cleanup job now. We note that 7 days ago, the branch's HEAD pointed to commit B. Therefore, after the cleanup job:

1. Object `example1` still exists in the storage, since it is accessible from commit B.
2. Object `example2` still exists in the storage, since it is accessible from commit B.
3. Object `example3` does not exist in the storage, since it is not accessible from commit B, or any of its descendants.

### Complex example: cleanup on multiple branches
Consider the following example, where a `feature1` branch is created from the `main` branch.
![image](https://user-images.githubusercontent.com/3149885/117683166-77fd7780-b1bc-11eb-828a-0f90444d6bc8.png)

On each of the branches we create a new file, and delete it in a subsequent commit. We also delete the file `example1` in the main branch. Suppose the retention period is 7 days for the main branch, and 3 days for the feature branch. 

**Note:**
* 7 days ago, the HEAD for the main branch was commit B.
* 3 days ago, the HEAD for the feature branch was commit D.

After the cleanup job:
1. Object `example2` still exists in the storage, since it is accessible from commit B.
2. Object `example3` does not exist in the storage, since it is not accessible from B, D, or any of their descendants. 
3. Object `example1` still exists in the storage, since it is accessible from commit D. Note that it is maintained even though it was deleted before the retention period of both branches.

### Configuring the retention period.
The user will have the option to define a default retention time (in days) and one for each desirable branch.
For example:
```bash
lakectl gc set-default 4 
lakectl gc set main 20
lakectl gc set other-branch 2
```
#### Definitions
[...]

PR #2169, opened 2021-06-27
Handle dangling commits

johnnyaug:
Solves #2068 according to the suggestion in #1932.


pkg/graveler/retention/garbage_collection_manager.go
```diff
@@ -108,15 +112,16 @@ func (m *GarbageCollectionManager) GetRunExpiredCommits(ctx context.Context, sto
...
-		return "", fmt.Errorf("list repository branches: %w", err)
-	}
 	commitGetter := &RepositoryCommitGetter{
 		refManager:   m.refManager,
 		repositoryID: repositoryID,
 	}
-	gcCommits, err := GetGarbageCollectionCommits(ctx, branchIterator, commitGetter, rules, previouslyExpiredCommits)
+	branchIterator := ref.NewBranchIterator(ctx, m.db, repositoryID, 1000, ref.WithOrderBy("commit_id"))
+	// get all commits that are not the first parent of any commit:
+	mainAncestryLeaves := "NOT EXISTS (SELECT * FROM graveler_commits c2 WHERE c2.repository_id=graveler_commits.repository_id AND c2.parents[1]=graveler_commits.id"
```
arielshaqed, line 121: AFAICS this codifies that a commit can have only 2 parents.  While this is true in the current implementation, we explicitly try _not_ to make this assumption anywhere.  It may have surprising results when we *do* allow multi-merges.  Given that we are using SQL, can we avoid this?

arielshaqed, line 121: This looks like something that would be quite hard for the query planner!  Perhaps we should reduce the isolation level, and/or fix it as  a proper anti-join.

pkg/graveler/retention/starting_point_iterator.go
```diff
@@ -0,0 +1,84 @@
+package retention
+
+import "github.com/treeverse/lakefs/pkg/graveler"
+
+// A GCStartingPoint represents a commit from which the GC algorithm should start scanning.
+// It could be either a branch HEAD, or a dangling commit.
```
arielshaqed, line 6: This confuses me about what the fields hold.  Specifically, does this mean that _either_ `BranchID` _or_ `CommitID` will be `nil`?  Or do I always get a `CommitID`?

arielshaqed (changes requested):
Thanks!

Looks neat.  Most comments are about understanding what the various fields hold.

I *am* very worried that we are going back into deep SQL-land here: this adds a JOIN that could be very expensive, an index that may be expensive to maintain or else is unused, and a function parameter that could in future cause an SQL injection.  It will also make our shift away from Postgres harder.

pkg/graveler/retention/expired_commits.go
```diff
@@ -22,24 +30,31 @@ func GetGarbageCollectionCommits(ctx context.Context, branchIterator graveler.Br
 	}
 	activeMap := make(map[graveler.CommitID]struct{})
 	expiredMap := make(map[graveler.CommitID]struct{})
-	for branchIterator.Next() {
-		branchRecord := branchIterator.Value()
+	for startingPointIterator.Next() {
+		var err error
```
nopcoder, line 34: can delete this one

pkg/graveler/retention/starting_point_iterator.go
```diff
@@ -0,0 +1,84 @@
...
+func (sp *GCStartingPointIterator) Next() bool {
+	prepareBranchValue := func() bool {
+		sp.value = &GCStartingPoint{
+			BranchID: sp.branchValue.BranchID,
+			CommitID: sp.branchValue.CommitID,
+		}
+		sp.branchValue = nil
+		return true
+	}
+	prepareCommitValue := func() bool {
```
nopcoder, line 37: used once

pkg/graveler/retention/starting_point_iterator.go
```diff
@@ -0,0 +1,84 @@
...
+	branchValue *graveler.BranchRecord
+	value       *GCStartingPoint
+}
+
+func NewGCStartingPointIterator(commitIterator graveler.CommitIterator, branchIterator graveler.BranchIterator) *GCStartingPointIterator {
+	return &GCStartingPointIterator{commitIterator: commitIterator, branchIterator: branchIterator}
+}
+
+func (sp *GCStartingPointIterator) Next() bool {
+	prepareBranchValue := func() bool {
```
nopcoder, line 29: think if you will change the order of the checks - we can remove this func too

pkg/graveler/retention/garbage_collection_manager.go
johnnyaug: This code does not make assumptions about about the number of parents - the first parent is always the relevant one.

However, added handling of the parent-switch bug we used to have

johnnyaug: This is the postgres way to anti-join :|

pkg/graveler/retention/starting_point_iterator.go
johnnyaug: Added comment

pkg/graveler/retention/starting_point_iterator.go
johnnyaug: done

pkg/graveler/retention/starting_point_iterator.go
johnnyaug: removed

pkg/graveler/ref/commit_ordered_iterator.go
```diff
@@ -65,11 +78,15 @@ func (iter *OrderedCommitIterator) maybeFetch() {
 	}
 
 	var buf []*commitRecord
+	additionalConditionExpression := ""
+	if iter.additionalCondition != "" {
+		additionalConditionExpression = " AND " + iter.additionalCondition + " "
+	}
```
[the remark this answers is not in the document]
johnnyaug: Done

pkg/graveler/retention/expired_commits.go
```diff
@@ -13,7 +13,15 @@ type GarbageCollectionCommits struct {
 }
 
 // GetGarbageCollectionCommits returns the sets of expired and active commits, according to the repository's garbage collection rules.
-func GetGarbageCollectionCommits(ctx context.Context, branchIterator graveler.BranchIterator, commitGetter *RepositoryCommitGetter, rules *graveler.GarbageCollectionRules, previouslyExpired []graveler.CommitID) (*GarbageCollectionCommits, error) {
+// From each starting point in the given startingPointIterator, it iterates through its main ancestry (see more below).
```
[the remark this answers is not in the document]
johnnyaug: Redid this comment

pkg/graveler/retention/expired_commits.go
```diff
@@ -13,7 +13,15 @@ type GarbageCollectionCommits struct {
...
+// While iterating, it marks all commits as active, until and including the first commit performed before the start of the retention period.
+// All further commits in the ancestry will be marked as expired, stopping when reaching a commit in the previouslyExpired set, or the DAG root.
+//
+// The main ancestry of a commit is acquired by recursively taking the first parent.
+// The commit from which to start depends on the starting point:
+// - if the starting point is a branch's HEAD, simply start from it.
+// - otherwise, start from a hypothetical HEAD which is the child of the starting point commit.
```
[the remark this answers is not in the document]
johnnyaug: Done + rephrased

pkg/graveler/retention/expired_commits.go
johnnyaug: done

pkg/graveler/retention/expired_commits.go
```diff
@@ -64,9 +79,10 @@ func GetGarbageCollectionCommits(ctx context.Context, branchIterator graveler.Br
 			processed[nextCommitID] = branchExpirationThreshold
 		}
 	}
-	if branchIterator.Err() != nil {
-		return nil, branchIterator.Err()
+	if startingPointIterator.Err() != nil {
+		return nil, startingPointIterator.Err()
 	}
+	startingPointIterator.Close()
 	return &GarbageCollectionCommits{active: commitSetToArray(activeMap), expired: commitSetToArray(expiredMap)}, nil
```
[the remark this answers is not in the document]
johnnyaug: FIxed

pkg/ddl/000030_commits_parents_index.up.sql
```diff
@@ -0,0 +1,4 @@
+BEGIN;
+CREATE INDEX graveler_commits_idx_parents ON graveler_commits(parents);
```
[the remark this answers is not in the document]
arielshaqed: I re-read https://stackoverflow.com/q/4058731/192263 and I am still not convinced that this index can ever be used.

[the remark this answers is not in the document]
arielshaqed: I tried it out.  According to this I can disable sequential scans, _except_ when they are the only way to perform a query:

> When indexes are not used, it can be useful for testing to force their use. There are run-time parameters that can turn off various plan types (see Section 19.7.1). For instance, turning off sequential scans (enable_seqscan) and nested-loop joins (enable_nestloop), which are the most basic plans, will force the system to use a different plan. If the system still chooses a sequential scan or nested-loop join then there is probably a more fundamental reason why the index is not being used; for example, the query condition does not match the index. (What kind of query can use what kind of index is explained in the previous sections.)

(https://www.postgresql.org/docs/13/indexes-examine.html)

And it turns out that an index on an array value is _not_ used for matching individual array elements:

```sql
ariels=# CREATE TABLE family (name TEXT, parents TEXT[]);
CREATE TABLE
ariels=# CREATE INDEX idx_parents ON family(parents);
CREATE INDEX
ariels=# SET enable_seqscan TO off;
SET
ariels=# INSERT INTO family VALUES('a', '{a,b,c}');
INSERT 0 1
ariels=# INSERT INTO family VALUES('b', '{c,b,f}');      
INSERT 0 1
ariels=# INSERT INTO family VALUES('c', '{c,x,f}');  
INSERT 0 1

ariels=# SELECT * FROM family WHERE parents[2] = 'b';
 name | parents 
------+---------
 a    | {a,b,c}
 b    | {c,b,f}
(2 rows)

ariels=# EXPLAIN ANALYZE SELECT * FROM family WHERE parents[2] = 'b';  -- this will seq scan (no index)
                                                       QUERY PLAN                                                       
-----------------------------------------------------------------------------------------------------------------------
 Seq Scan on family  (cost=10000000000.00..10000000001.04 rows=1 width=64) (actual time=40.776..40.780 rows=2 loops=1)
   Filter: (parents[2] = 'b'::text)
   Rows Removed by Filter: 1
 Planning Time: 0.171 ms
 JIT:
   Functions: 2
   Options: Inlining true, Optimization true, Expressions true, Deforming true
   Timing: Generation 0.792 ms, Inlining 15.325 ms, Optimization 17.722 ms, Emission 7.617 ms, Total 41.455 ms
 Execution Time: 41.629 ms
(9 rows)


ariels=# CREATE INDEX idx_parents_2 ON family((parents[2]));
CREATE INDEX
ariels=# EXPLAIN ANALYZE SELECT * FROM family WHERE parents[2] = 'b';  -- This will index scan
                                                      QUERY PLAN                                                       
-----------------------------------------------------------------------------------------------------------------------
 Index Scan using idx_parents_2 on family  (cost=0.13..8.15 rows=1 width=64) (actual time=0.020..0.023 rows=2 loops=1)
   Index Cond: (parents[2] = 'b'::text)
 Planning Time: 0.103 ms
 Execution Time: 0.049 ms
(4 rows)
```

pkg/graveler/ref/commit_ordered_iterator.go
```diff
@@ -55,23 +72,43 @@ func (iter *OrderedCommitIterator) maybeFetch() {
...
+			From("graveler_commits c2").
+			Where("c2.repository_id=graveler_commits.repository_id").
+			Where(sq.Or{
+				sq.And{
+					sq.Lt{"c2.version": graveler.CommitVersionParentSwitch},
+					sq.Expr("c2.parents[array_length(c2.parents, 1)]=graveler_commits.id")},
+				sq.And{
+					sq.GtOrEq{"c2.version": graveler.CommitVersionParentSwitch},
+					sq.Expr("c2.parents[1]=graveler_commits.id")},
+			})
```
arielshaqed, line 101: It _might_ be possible to create indices on `graveler_commits` for second and last parents:

```sql
CREATE INDEX graveler_commits_second_parent_idx ((parents[2]));
CREATE INDEX graveler_commits_last_parent_idx ((parents[array_length(parents, 1)]));
```
Note however that, especially for the second index, you will have to be _very_ careful to use an expression that the query planner can recognise as being equal to the index.

arielshaqed (changes requested):
Thanks!

Requesting a change on the parents index, as I do not believe it can be used by the Postgres query planner.

pkg/graveler/ref/commit_ordered_iterator.go
johnnyaug: Thanks Ariel! I'm looking into it

guy-har (approved):
Looks Great!
Loved the tests and the documentation, makes it easy to follow.
Thanks

pkg/graveler/retention/expired_commits.go
```diff
@@ -22,24 +26,30 @@ func GetGarbageCollectionCommits(ctx context.Context, branchIterator graveler.Br
...
+		}
+		branchExpirationThreshold := now.AddDate(0, 0, -retentionDays)
+		if startingPoint.BranchID != "" {
+			processed[startingPoint.CommitID] = now.AddDate(0, 0, -retentionDays)
 		}
-		processed[commitID] = branchExpirationThreshold
-		activeMap[commitID] = struct{}{}
 		for len(commit.Parents) > 0 {
 			// every branch retains only its main ancestry, acquired by recursively taking the first parent:
 			nextCommitID := commit.Parents[0]
```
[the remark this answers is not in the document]
johnnyaug: Exactly! done.

pkg/graveler/retention/expired_commits.go
```diff
@@ -64,13 +74,14 @@ func GetGarbageCollectionCommits(ctx context.Context, branchIterator graveler.Br
 			processed[nextCommitID] = branchExpirationThreshold
 		}
 	}
-	if branchIterator.Err() != nil {
-		return nil, branchIterator.Err()
+	if startingPointIterator.Err() != nil {
+		return nil, startingPointIterator.Err()
 	}
-	return &GarbageCollectionCommits{active: commitSetToArray(activeMap), expired: commitSetToArray(expiredMap)}, nil
+	startingPointIterator.Close()
```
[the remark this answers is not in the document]
johnnyaug: Done

pkg/graveler/ref/commit_ordered_iterator.go
```diff
@@ -4,15 +4,19 @@ import (
 	"context"
 	"errors"
 
+	sq "github.com/Masterminds/squirrel"
+
```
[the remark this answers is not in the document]
johnnyaug: done

pkg/ddl/000031_commits_parents_index.up.sql
```diff
@@ -1,4 +1,22 @@
...
+
+-- function to return the first parent:
+-- this takes into consideration commits created before fixing the parent switch bug (https://github.com/treeverse/lakeFS/issues/1730)
+CREATE OR REPLACE FUNCTION first_parent(parents text[], version int)
+RETURNS text AS
+    $$
+    DECLARE
+        parent_switch_commit_version int;
+    BEGIN
+        parent_switch_commit_version := 1;
```
arielshaqed, line 13: Does this work?  I think it will be more readable.  (You might also drop the `CONSTANT` if that's from too advanced a Postgres version, it will still be easier for me to read.)  https://www.postgresql.org/docs/13/plpgsql-declarations.html says it should work, even in version 11, so I reckon it might.

```suggestion
    DECLARE
        parent_switch_commit_version CONSTANT INT = 1;
    BEGIN
```

pkg/graveler/ref/commit_ordered_iterator.go
```diff
@@ -91,18 +90,11 @@ func (iter *OrderedCommitIterator) maybeFetch() {
...
 			Where("c2.repository_id=graveler_commits.repository_id").
-			Where(sq.Or{
-				sq.And{
-					sq.Lt{"c2.version": graveler.CommitVersionParentSwitch},
-					sq.Expr("c2.parents[array_length(c2.parents, 1)]=graveler_commits.id")},
-				sq.And{
-					sq.GtOrEq{"c2.version": graveler.CommitVersionParentSwitch},
-					sq.Expr("c2.parents[1]=graveler_commits.id")},
-			})
+			Where("first_parent(c2.parents, c2.version)=graveler_commits.id")
```
arielshaqed, line 93: Oooo.... nice!

pkg/graveler/ref/commit_ordered_iterator.go
```diff
@@ -91,18 +90,11 @@ func (iter *OrderedCommitIterator) maybeFetch() {
...
 		q = q.Where(notExistsCondition)
 	}
-	q.OrderBy("id ASC")
-	q.Limit(uint64(iter.prefetchSize))
+	q = q.OrderBy("id ASC")
+	q = q.Limit(uint64(iter.prefetchSize))
```
arielshaqed, line 97: I am puzzled by this style.  The Squirrel documentation _and_ [examples](https://github.com/Masterminds/squirrel/blob/cd1fe0a38ba74020c727421ca469fc4a4d7f0258/select_test.go#L15) all seem to use the assigned version.  Is this style supported?

AFAICT _this should not work_: it ends up using https://github.com/lann/ps/blob/master/map.go, which is an "immutable map", and every Set returns a *new* builder without disturbing the old one.  Please let's be very sure to test this.

arielshaqed (commented):
Thanks!  The change in Squirrel style to chain modifiers bothers me, because that's *not* how I remember the SQL builder working.  So not approving until you say it's intentional and does actually work, sorry.

pkg/graveler/ref/commit_ordered_iterator.go
arielshaqed: D'uh, this *adds* the chaining that I wanted.  So I just explained in a long comment why this change is good.  Oops :stuck_out_tongue_winking_eye: sorry

arielshaqed (approved):

nopcoder (approved):

PR #2172, opened 2021-06-28
GC S3 bulk deletes

guy-har:


johnnyaug (approved):

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -227,6 +231,66 @@ object GarbageCollector {
...
+  // collectByPartitionAndSize, for given column of type X return Lists[X] of size up to maxSize for rows within partition
+  def collectByPartitionAndSize(df: DataFrame, maxSize: Int, column: String) = {
+    val nRows = df.count()
+    val maxNRowsPartition = maxSize //make sure its a multiple of desired array length
+    val nPartitions = math.max(1, math.floor(nRows / maxNRowsPartition)).toInt
+    val repartitioned = df
+      .repartitionByRange(nPartitions, col(column).desc)
+      .withColumn("partitionId", spark_partition_id())
+
+    val w = Window.partitionBy(col("partitionId")).orderBy(column)
```
johnnyaug, line 249: Synced f2f: consider removing this part and using the partitions received by calling `repartitionByRange`. Those partitions may have more records than the target size, so this would need to be handled.

johnnyaug (changes requested):

johnnyaug (approved):

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -227,6 +236,103 @@ object GarbageCollector {
...
+      .partitionBy("run_id")
+      .mode(SaveMode.Append)
+      .parquet(addressesDFLocation) // TODO(Guys): consider changing to overwrite
+  }
+}
+
+object S3BulkDeleter {
+  def repartitionBySize(df: DataFrame, maxSize: Int, column: String): DataFrame = {
+    val nRows = df.count()
+    val nPartitions = math.max(1, math.floor(nRows / maxSize)).toInt
```
arielshaqed, line 249: This computes a strange number of partitions, the special case for 1 is strange.  Consider
```suggestion
    val nPartitions = math.ceil(nRows / maxSize).toInt
```
or even just
```suggestion
    val nPartitions = (nRows + maxSize - 1) / maxSize
```

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -227,6 +236,103 @@ object GarbageCollector {
...
+  def bulkRemoveFromIter(
+      keys: Iterator[String],
+      bucket: String,
+      region: String,
+      bulkSize: Int,
+      numRetries: Int
+  ): Iterator[mutable.Buffer[String]] = {
+    var nextBatch = keys.take(bulkSize)
+    var res = Seq[mutable.Buffer[String]]()
+    val s3Client = getS3Client(region, numRetries)
```
arielshaqed, line 281: Could we create the client once and pass that instead of passing `region` and `numRetries` parameters that are constant for the entire bulk remove operation?

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -227,6 +236,103 @@ object GarbageCollector {
...
+    S3Client.builder.region(Region.of(region)).overrideConfiguration(configuration).build
+  }
+
+  def bulkRemoveFromIter(
+      keys: Iterator[String],
+      bucket: String,
+      region: String,
+      bulkSize: Int,
+      numRetries: Int
+  ): Iterator[mutable.Buffer[String]] = {
```
arielshaqed, line 278: Why does this need to be a mutable buffer?

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -227,6 +236,103 @@ object GarbageCollector {
...
+    var nextBatch = keys.take(bulkSize)
+    var res = Seq[mutable.Buffer[String]]()
+    val s3Client = getS3Client(region, numRetries)
+    while (!nextBatch.isEmpty) {
+      res = res ++ delObjIteration(bucket, nextBatch.toSeq, s3Client).map(x =>
+        mutable.Buffer[String](x)
+      )
+      nextBatch = keys.take(bulkSize)
+    }
```
arielshaqed, line 287: A more functional (Scala-ish) way would be to use higher-order operators on a list, e.g. something like
```scala
val s3Client = getS3Client(region, numRetries)
(keys grouped bulkSize).
    flatMap(delObjOperation(bucket, _, s3Client))
```
which it might even be possible to write as
```scala
val s3Client = getS3Client(region, numRetries)
keys grouped bulkSize flatMap delObjOperation(bucket, _, s3Client)
```

Explanation: The operators `grouped` and `flatMap` are defined in https://docs.scala-lang.org/overviews/collections/iterators.html.  Take `keys` in groups of size `bulkSize` (list of batches of keys), run `delObjOperation` on each batch (each batch gives a batch of results), and flatten the list of those batches.

arielshaqed (commented):
Sorry for reviewing a pull_ed_ request...

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
guy-har: No, because it's not serializable

guy-har:
@arielshaqed, Thanks for the comments! fixed them in this PR #2188

PR #2178, opened 2021-06-29
url parse and resolve split

nopcoder:
- Split ref resolve to parse and resolve steps
- Parse supports '@' for branch last commit (ex: main@)
- Parse supports '$' for staging (ex: main$)


pkg/graveler/graveler.go
```diff
@@ -28,6 +30,37 @@ const (
...
+	RefModTypeAt     RefModType = '@'
+	RefModTypeDollar RefModType = '$'
+)
+
+type RefModifier struct {
+	Type  RefModType
+	Value int
+}
+
+type ParsedRef struct {
```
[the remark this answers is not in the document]
itaiad200: And maybe even a rename, `ParsedRef` -> `RawRef` or `UnresolvedRef`.

pkg/graveler/graveler.go
```diff
@@ -28,6 +30,37 @@ const (
...
+	BaseRef      string
+	RefModifiers []RefModifier
```
itaiad200, line 49: The order is important? It's worth mentioning it if that's the case.

pkg/graveler/graveler.go
```diff
@@ -28,6 +30,37 @@ const (
...
+}
+
+func (p *ParsedRef) Ref() Ref {
```
itaiad200, line 52: Missing documentation

pkg/catalog/catalog.go
```diff
@@ -815,23 +817,25 @@ func (c *Catalog) Commit(ctx context.Context, repository string, branch string,
...
 		return nil, err
 	}
-	commitID, err := c.Store.Dereference(ctx, repositoryID, ref)
+	resolvedRef, err := c.Store.Dereference(ctx, repositoryID, graveler.Ref(reference))
 	if err != nil {
 		return nil, err
 	}
-
+	if resolvedRef.CommitID == "" {
+		return nil, fmt.Errorf("%w: missing commit", ErrInvalidRef)
```
itaiad200, line 830:
```suggestion
		return nil, fmt.Errorf("missing commit: %w", ErrInvalidRef)
```

pkg/catalog/catalog.go
```diff
@@ -853,27 +857,27 @@ func (c *Catalog) ListCommits(ctx context.Context, repository string, branch str
...
-	branchCommitID, err := c.Store.Dereference(ctx, repositoryID, graveler.Ref(branchRef))
+	resolvedRef, err := c.Store.Dereference(ctx, repositoryID, graveler.Ref(branchRef))
 	if err != nil {
 		return nil, false, fmt.Errorf("branch ref: %w", err)
 	}
-	if branchCommitID == "" {
+	if resolvedRef.CommitID == "" {
 		// return empty log if there is no commit on branch yet
 		return make([]*CommitLog, 0), false, nil
 	}
```
itaiad200, line 867: I see this pattern returns several times, maybe have a utility function to handle missing deref + checking for commit.

nopcoder: true, the dereference was just to a commit

pkg/graveler/graveler.go
nopcoder: will add it part of the struct comment

pkg/graveler/graveler.go
nopcoder: will remove this one - thought it will be required for a test

talSofer (approved):

pkg/graveler/graveler.go
```diff
@@ -1044,8 +1101,8 @@ func (g *Graveler) List(ctx context.Context, repositoryID RepositoryID, ref Ref)
 	if err != nil {
 		return nil, err
 	}
-	if reference.Type() == ReferenceTypeBranch {
-		stagingList, err := g.StagingManager.List(ctx, reference.Branch().StagingToken)
+	if reference.Type == ReferenceTypeBranch {
+		stagingList, err := g.StagingManager.List(ctx, reference.StagingToken)
```
johnnyaug, line 1105: So if we explicitly use "@" to get only a commit - we get empty staging token here. And then you rely on the query from the DB to return empty?
If so, I would explicitly check for an empty staging token here, and not call the staging manager at all.

This comment may be relevant at other places in this file

pkg/catalog/catalog.go
```diff
@@ -917,7 +913,7 @@ func (c *Catalog) ListCommits(ctx context.Context, repository string, branch str
 func (c *Catalog) Revert(ctx context.Context, repository string, branch string, params RevertParams) error {
 	repositoryID := graveler.RepositoryID(repository)
 	branchID := graveler.BranchID(branch)
-	ref := graveler.Ref(params.Reference)
+	revertRef := graveler.Ref(params.Reference)
```
johnnyaug, line 916: confusing name, sounds like a function. why not let it be ref? (same for other functions)

johnnyaug (changes requested):

pkg/graveler/graveler.go
```diff
@@ -41,10 +63,29 @@ const (
...
+	ResolvedBranchModifierCommitted
+	ResolvedBranchModifierStaging
+)
+
+// ResolvedRef include resolved information of Ref/RawRef:
+//   Type: Branch / Tag / Commit
+//   BranchID: for type ReferenceTypeBranch will hold the branch ID
+//   ResolvedBranchModifier: branch indicator if resolved to a branch latest commit, staging or none was specified.
+//   CommitID: set for branch (none or committed), tag or commit
+//   StagingToken: set for branch (none or staging)
```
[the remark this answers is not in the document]
nopcoder: took 50% - the commit id is always set.

pkg/graveler/graveler.go
nopcoder: true, I'll switch the check to match ResolvedBranchModifierStaging and not the type of the reference we resolved

pkg/graveler/ref/resolve_ref_test.go
```diff
@@ -104,40 +109,77 @@ func TestResolveRef_Dereference(t *testing.T) {
...
+		ExpectedErr            error
 	}{
 		{
-			Name:     "branch_exist",
-			Ref:      graveler.Ref("branch1"),
-			Expected: branch1CommitID,
+			Name:                   "branch_exist",
+			Ref:                    graveler.Ref("branch1"),
+			ExpectedBranchModifier: graveler.ResolvedBranchModifierNone,
+			ExpectedCommitID:       branch1CommitID,
```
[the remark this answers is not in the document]
nopcoder: true - I'll fix it.

pkg/catalog/catalog.go
nopcoder: renamed to `reference`. `ref` conflicts with the package name.

pkg/graveler/graveler.go
```diff
@@ -28,6 +28,28 @@ const (
...
+)
+
+type RefModifier struct {
+	Type  RefModType
+	Value int
+}
+
+// RawRef is a parsed Ref that includes 'BaseRef' that holds the branch/tag/hash and a list of
+//   ordered modifiers that applied to the reference.
+// Example: master~2 will be parsed into {BaseRef:"master", Modifiers:[{Type:RefModTypeTilde, Value:2}]}
```
itaiad200, line 47: 👍

itaiad200 (approved):
Looks great. Thanks for applying the fixes!

pkg/catalog/catalog.go
```diff
@@ -597,24 +596,23 @@ func (c *Catalog) GetTag(ctx context.Context, repository string, tagID string) (
 // the entry with ExpiredError if it has expired from underlying storage.
 func (c *Catalog) GetEntry(ctx context.Context, repository string, reference string, path string, _ GetEntryParams) (*DBEntry, error) {
 	repositoryID := graveler.RepositoryID(repository)
-	ref := graveler.Ref(reference)
-	p := Path(path)
+	getRef := graveler.Ref(reference)
```
johnnyaug, line 599:
```suggestion
	reference := graveler.Ref(reference)
```

johnnyaug (approved):
Nice!

pkg/catalog/catalog.go
```diff
@@ -708,16 +706,16 @@ func (c *Catalog) ListEntries(ctx context.Context, repository string, reference
 	afterPath := Path(after)
 	delimiterPath := Path(delimiter)
 	repositoryID := graveler.RepositoryID(repository)
-	ref := graveler.Ref(reference)
+	listRef := graveler.Ref(reference)
```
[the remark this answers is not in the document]
nopcoder: cant use 'reference' twice

pkg/catalog/catalog.go
nopcoder: The 'reference' here is already a `string` type

PR #2193, opened 2021-07-04
Release Metadata Client

johnnyaug:
- Created a version 0.1.0 of the metadata client: not a snapshot.
- Organized Maven group names to be `io.lakefs` instead of `io.treeverse`
- Bumped java API client version to 0.44.0


clients/spark/build.sbt
```diff
@@ -2,8 +2,8 @@ import build.BuildType
 
 lazy val baseName = "lakefs-spark"
 
-lazy val projectVersion = "0.1.4-SNAPSHOT"
-ThisBuild / isSnapshot := true
+lazy val projectVersion = "0.1.0"
+ThisBuild / isSnapshot := false
```
johnnyaug, line 6: Not sure this change is desired here. Waitiing for your comments

clients/spark/build.sbt
```diff
@@ -174,7 +174,7 @@ ThisBuild / developers := List(
 )
 
 ThisBuild / versionScheme := Some("early-semver")
-ThisBuild / organization := "io.treeverse"
```
itaiad200, line 177: Should we clean up old releases with the previous organization?

itaiad200 (approved):

clients/spark/build.sbt
johnnyaug: Will do it in a month or so, after people switch to the new one

PR #2196, opened 2021-07-05
GC combine to one command and handle deduplications

guy-har:


clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -209,48 +229,76 @@ object GarbageCollector {
...
+    )
   }
 
   def main(args: Array[String]) {
     val spark = SparkSession.builder().getOrCreate()
-    if (args.length != 4) {
+    if (args.length != 2) {
       Console.err.println(
-        "Usage: ... <repo_name> <runID> s3://storageNamespace/prepared_commits_table s3://storageNamespace/output_destination_table"
+        "Usage: ... <repo_name> <region>"
```
johnnyaug, line 261: About the region - can we use [withForceGlobalBucketAccessEnabled](https://stackoverflow.com/a/46111903/2696779) to start the client without the region?

johnnyaug (changes requested):

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -209,48 +229,76 @@ object GarbageCollector {
...
       System.exit(1)
     }
 
     val repo = args(0)
-    val runID = args(1)
-    val commitDFLocation = args(2)
-    val addressesDFLocation = args(3)
-
+    val region = args(1)
+    val previousRunID = "" //args(2) // TODO(Guys): get previous runID from arguments or from storage
```
johnnyaug, line 268: So we're letting go of the previous run feature for now?

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -102,6 +102,26 @@ object GarbageCollector {
...
+                      repo: String,
+                      set:Set[String]
+                    ): Set[String] = {
+
+    val location =
+      new ApiClient(conf.apiURL, conf.accessKey, conf.secretKey).getRangeURL(repo, rangeID)
+    SSTableReader
+      .forRange(new Configuration(), location)
+      .newIterator()
+      .filter(!_.message.addressType.isRelative)
```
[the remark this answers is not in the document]
guy-har: Bug, fixing

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
guy-har: Thats for the previous SDK version. seems like they didn't solve it yet for the current version.
https://github.com/aws/aws-sdk-java-v2/issues/2229
https://github.com/aws/aws-sdk-java-v2/issues/52

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
guy-har: No, It will be the first thing to add after the release

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -326,16 +362,27 @@ object S3BulkDeleter {
...
-      .partitionBy("run_id")
-      .mode(SaveMode.Append)
-      .parquet(deletedAddressesDFLocation)
+      bulkRemove(df, MaxBulkSize, spark, bucket, region, awsRetries, snPrefix)
+        .toDF("addresses")
+        .collect()
+  }
+
+  def main(args: Array[String]): Unit = {
+    if (args.length != 5) {
```
johnnyaug, line 374:
```suggestion
    if (args.length != 4) {
```

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -326,16 +361,27 @@ object S3BulkDeleter {
       if (addSuffixSlash.startsWith("/")) addSuffixSlash.substring(1) else addSuffixSlash
 
     val df = spark.read
-      .parquet(addressesDFLocation)
+      .parquet(location)
```
johnnyaug, line 364: why rename?

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -209,48 +228,76 @@ object GarbageCollector {
...
-    getExpiredEntriesFromRanges(rangesDF, conf, repo)
+    val expired = getExpiredEntriesFromRanges(rangesDF, conf, repo)
+
+    // validate deduplications
+    val activeRangesDF = rangesDF.where("!expired")
+    subtractDeduplications(expired, activeRangesDF, conf, repo, spark)
+  }
+
+  private def subtractDeduplications(expired: Dataset[Row], activeRangesDF: Dataset[Row], conf: APIConfigurations, repo: String, spark: SparkSession) : Dataset[Row]= {
+    val expiredAddr: Set[String] = expired.select("address").collect().map(_.getString(0)).toSet
```
johnnyaug, line 239:
```suggestion
    val deleteCandidateAddresses: Set[String] = expired.select("address").collect().map(_.getString(0)).toSet
```

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -209,48 +228,76 @@ object GarbageCollector {
...
+    val ranges : Seq[String] = activeRangesDF.select("range_id").collect().map(_.getString(0)).toSeq.distinct
+    val rangesRDD = spark.sparkContext.parallelize(ranges)
```
johnnyaug, line 241:
```suggestion
    val activeRangesRDD = spark.sparkContext.parallelize(ranges)
```

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -209,48 +228,76 @@ object GarbageCollector {
...
+
+    val activeAddresses =rangesRDD.flatMap(range=> {
```
johnnyaug, line 243:
```suggestion
    val activeDeleteCandidateAddresses = activeRangeRDD.flatMap(range => {
```

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -209,48 +228,76 @@ object GarbageCollector {
...
+      getEntryAddressesIfInSet(range, conf, repo, expiredAddr)
+    }).map(x => Row(x))
+
+    val schema = new StructType().add(StructField("address",StringType, true))
+    val activeDF = spark.createDataFrame(activeAddresses, schema)
+    expired.join(
```
johnnyaug, line 249:
```suggestion
    // remove active addresses from delete candidates
    expired.join(
```

johnnyaug (changes requested):

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -102,6 +102,25 @@ object GarbageCollector {
...
+                      rangeID: String,
+                      conf: APIConfigurations,
+                      repo: String,
+                      set:Set[String]
+                    ): Set[String] = {
+
+    val location =
+      new ApiClient(conf.apiURL, conf.accessKey, conf.secretKey).getRangeURL(repo, rangeID)
+    SSTableReader
+      .forRange(new Configuration(), location)
```
arielshaqed, line 115: Is this truly the configuration we want here?

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -102,6 +102,25 @@ object GarbageCollector {
...
+      .newIterator()
+      .filter(x =>set.contains(x.message.address))
+      .map(a =>
+          new String(a.message.address),
```
arielshaqed, line 119: Why copy a `String`?  AFAIK they are anyway immutable.

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -209,48 +228,76 @@ object GarbageCollector {
...
+    val expired = getExpiredEntriesFromRanges(rangesDF, conf, repo)
+
+    // validate deduplications
+    val activeRangesDF = rangesDF.where("!expired")
+    subtractDeduplications(expired, activeRangesDF, conf, repo, spark)
+  }
+
+  private def subtractDeduplications(expired: Dataset[Row], activeRangesDF: Dataset[Row], conf: APIConfigurations, repo: String, spark: SparkSession) : Dataset[Row]= {
+    val expiredAddr: Set[String] = expired.select("address").collect().map(_.getString(0)).toSet
+    val ranges : Seq[String] = activeRangesDF.select("range_id").collect().map(_.getString(0)).toSeq.distinct
```
arielshaqed, line 240: Not sure I understand: doesn't `collect` directly move everything to the driver program and destroy parallelism?  At the very least, I would run `distinct` or its equivalent on the cluster.

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -209,48 +228,76 @@ object GarbageCollector {
...
+    val rangesRDD = spark.sparkContext.parallelize(ranges)
+
+    val activeAddresses =rangesRDD.flatMap(range=> {
+      getEntryAddressesIfInSet(range, conf, repo, expiredAddr)
+    }).map(x => Row(x))
```
arielshaqed, line 245: This looks like a flatMap with side data.  I *think* that it will be somewhat tricky, because `expiredAddr` needs to be broadcast to all the workers.  It looks like it is some kind of JOIN, so I would prefer if you could write `ranges` as an RDD (or any other Spark table, I don't really know the difference...) and join with that here.

arielshaqed (changes requested):
I am not sure about the code that selects things, IIUC it will not let Spark do its thing and parallelize the run properly :-(

johnnyaug (approved):

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -50,13 +56,24 @@ class ApiClient(apiUrl: String, accessKey: String, secretKey: String) {
...
     val commit = commitsApi.getCommit(repoName, commitID)
     val metaRangeID = commit.getMetaRangeId
-
-    val metaRange = metadataApi.getMetaRange(repoName, metaRangeID)
-    val location = metaRange.getLocation
-    URI.create(getStorageNamespace(repoName) + "/" + location).normalize().toString
+    if (metaRangeID != "") {
+      val metaRange = metadataApi.getMetaRange(repoName, metaRangeID)
+      val location = metaRange.getLocation
+      URI.create(getStorageNamespace(repoName) + "/" + location).normalize().toString
```
arielshaqed, line 75: As you probably know, I am not a fan of of constructing URIs by hand and prefer `resolve`: that is specified by a standard.  The issue here is sensitivity to trailing and leading slashes.
(Not a blocker)

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -38,13 +39,15 @@ object GarbageCollector {
...
     val location =
       new ApiClient(conf.apiURL, conf.accessKey, conf.secretKey).getMetaRangeURL(repo, commitID)
-    SSTableReader
-      .forMetaRange(new Configuration(), location)
-      .newIterator()
-      .map(range =>
-        (new String(range.id), range.message.minKey.toByteArray, range.message.maxKey.toByteArray)
-      )
-      .toSet
+    if (location == "") Set[(String, Array[Byte], Array[Byte])]()
```
arielshaqed, line 42: Why do you need the type signature here?  I would expect it to derive through the from the required return type.  E.g. this does type-check:
```scala
def f(x: Boolean): Set[(Int, Int)] = if (x) Set() else Set((2,3), (3,2))
```

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -209,48 +217,84 @@ object GarbageCollector {
   ): Dataset[Row] = {
     val commitsDF = getCommitsDF(runID, commitDFLocation, spark)
     val rangesDF = getRangesDFFromCommits(commitsDF, repo, conf)
-    getExpiredEntriesFromRanges(rangesDF, conf, repo)
+    val expired = getExpiredEntriesFromRanges(rangesDF, conf, repo)
+
+    // validate deduplications
```
arielshaqed, line 222: I don't understand this comment: when it find dedupes, it actually changes the return, no?

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
arielshaqed, line 42: I don't understand: when is `location == ""` a valid return?  If it has special meaning, please add a comment to document that meaning.

arielshaqed (commented):
Thanks!

Please add a comment on l. 42 about what an empty location means.

guy-har:
@arielshaqed, Thanks for the review, PTAL

arielshaqed (approved):
Cool!  Thanks...

PR #2198, opened 2021-07-05
hadoop-lakefs without dedicated s3 client

### LINKED ISSUE #2192: hadoop-lakefs write Parquet fails on databricks using instance profile
Settings just lakeFS fs hadoop configuration. Databricks access S3 using `Instance Profile`.
Write to lakeFS (lakefs://) using hadoop-lakefs fails without feedback. While trying to reproduce the following code the fs failed to access the underlying storage. 

Sample code:

```scala
val masterPath = s"lakefs://test31/master/parquet/"
case class Test(story: String)
val masterStory = Seq(Test("Say Yes More")).toDS
println(masterPath)
masterStory.write.mode("overwrite").parquet(masterPath)
```

Write fails with the following error:

```
Caused by: AmazonS3Exception: Access Denied; request: GET https://barak-bucket1.s3.amazonaws.com  {key=[2], key=[1], key=[test31/data/test31-master:b866c7ec-974a-49ff-8e9c-dfa87abe22c9/EIMM29s60WGIUtspwEZyR], key=[false]} Hadoop 2.7.4, aws-sdk-java/1.11.655 Linux/5.4.0-1049-aws OpenJDK_64-Bit_Server_VM/25.282-b08 java/1.8.0_282 scala/2.12.10 vendor/Azul_Systems,_Inc. com.amazonaws.services.s3.model.ListObjectsV2Request; Request ID: RKRB2HGS490PHEEN, Extended Request ID: ggxbplgjPdQnmk+Dml09wUctenUK4D1YG+unNZq+4EghUs1cKJfON1BAs8GhaOm0LQTC5BeAgks=, Cloud Provider: AWS, Instance ID: i-039277cb10ca37c85 (Service: Amazon S3; Status Code: 403; Error Code: AccessDenied; Request ID: RKRB2HGS490PHEEN; S3 Extended Request ID: ggxbplgjPdQnmk+Dml09wUctenUK4D1YG+unNZq+4EghUs1cKJfON1BAs8GhaOm0LQTC5BeAgks=)
```

nopcoder:
The Problem:

The `hadoop-lakefs` filesystem instantiates the s3 client in order to get a newly created object ETag.
Credentials used by the s3 client should be used to access the data the cluster created.
One of the issues was in Databricks instance profile credentials were not used by our client and we failed to access the data we just wrote using the underlying file system.

The Solution:

FileStatus in S3 is a specific type S3AFileStatus that holds the Etag, this is true for the client found in spark 3x.
We don't want to bind to a specific interface, so we use reflection to call the getter function to extract the ETag value from FileStatus.

Spark 2.x comes with an older version of the S3A file system which doesn't hold the Etag as part of the FileStatus. In order to address this version, we will have a fallback where ask the S3 client from the underlying filesystem.
The underlying file system has multiple implementations, we can't cast to a specific interface. Using reflection we accessed the underlying client and request the object metadata in order to extract the ETag.

Because we want to keep using the filesystem and not directly access the s3 client, we kept the s3 as a fallback although it will require the extra call.

Changes:

- Removed the s3 client instance
- Used reflection to access file status's ETag, and fallback to using AWS client using reflection after object creates completely.
- The above functionality is encapsulated into MetadataClient

Close #2192

### DIFF
--- a/clients/hadoopfs/README.md
+++ b/clients/hadoopfs/README.md
@@ -1,7 +1,0 @@
-# hadoop-lakefs file system
-
-Build assembly without signing
-
-```bash
-mvn -U install -Passembly -P'!treeverse-signing'
-```
--- a/clients/hadoopfs/README.md
+++ b/clients/hadoopfs/README.md
@@ -1,0 +1,7 @@
+# hadoop-lakefs file system
+
+Build assembly without signing
+
+```bash
+mvn -U install -Passembly -P'!treeverse-signing'
+```
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -382,22 +382,15 @@ public class LakeFSFileSystem extends FileSystem {
      * @return an IOException that corresponds to the translated API exception
      */
     private IOException translateException(String msg, ApiException e) {
-        IOException ex;
         int code = e.getCode();
         switch (code) {
             case HttpStatus.SC_NOT_FOUND:
-                ex = new FileNotFoundException(msg);
-                ex.initCause(e);
-                break;
+                return (FileNotFoundException) new FileNotFoundException(msg).initCause(e);
             case HttpStatus.SC_FORBIDDEN:
-                ex = new AccessDeniedException(msg);
-                ex.initCause(e);
-                break;
+                return (AccessDeniedException) new AccessDeniedException(msg).initCause(e);
             default:
-                ex = new IOException(msg, e);
-                break;
+                return new IOException(msg, e);
         }
-        return ex;
     }
 
     @Override
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -48,7 +48,6 @@ public class LakeFSFileSystem extends FileSystem {
     private LakeFSClient lfsClient;
     private int listAmount;
     private FileSystem fsForConfig;
-    private MetadataClient metadataClient;
 
     private URI translateUri(URI uri) throws java.net.URISyntaxException {
         switch (uri.getScheme()) {
@@ -97,8 +96,6 @@ public class LakeFSFileSystem extends FileSystem {
         } catch (ApiException | URISyntaxException e) {
             LOG.warn("get underlying filesystem for {}: {} (use default values)", path, e);
         }
-
-        this.metadataClient = new MetadataClient(fsForConfig);
     }
 
     @FunctionalInterface
@@ -177,7 +174,7 @@ public class LakeFSFileSystem extends FileSystem {
             // TODO(ariels): add fs.FileSystem.Statistics here to keep track.
             return new FSDataOutputStream(new LinkOnCloseOutputStream(staging, stagingLoc, objectLoc,
                     physicalUri,
-                    this.metadataClient,
+                    new MetadataClient(physicalFs),
                     // FSDataOutputStream is a kind of OutputStream(!)
                     physicalFs.create(physicalPath, false, bufferSize, replication, blockSize, progress)),
                     null);
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -8,6 +8,7 @@ import io.lakefs.clients.api.model.*;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.*;
 import org.apache.hadoop.fs.permission.FsPermission;
+import org.apache.hadoop.fs.s3a.S3AFileSystem;
 import org.apache.hadoop.util.Progressable;
 import org.apache.http.HttpStatus;
 import org.slf4j.Logger;
@@ -384,15 +385,22 @@ public class LakeFSFileSystem extends FileSystem {
      * @return an IOException that corresponds to the translated API exception
      */
     private IOException translateException(String msg, ApiException e) {
+        IOException ex;
         int code = e.getCode();
         switch (code) {
             case HttpStatus.SC_NOT_FOUND:
-                return (IOException) new FileNotFoundException(msg).initCause(e);
+                ex = new FileNotFoundException(msg);
+                ex.initCause(e);
+                break;
             case HttpStatus.SC_FORBIDDEN:
-                return (IOException) new AccessDeniedException(msg).initCause(e);
+                ex = new AccessDeniedException(msg);
+                ex.initCause(e);
+                break;
             default:
-                return new IOException(msg, e);
+                ex = new IOException(msg, e);
+                break;
         }
+        return ex;
     }
 
     @Override
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -63,7 +63,6 @@ public class LakeFSFileSystem extends FileSystem {
     private URI uri;
     private Path workingDirectory = new Path(Constants.SEPARATOR);
     private LakeFSClient lfsClient;
-    private AmazonS3 s3Client;
     private int listAmount;
     private FileSystem fsForConfig;
 
@@ -97,7 +96,6 @@ public class LakeFSFileSystem extends FileSystem {
         setConf(conf);
         this.uri = name;
 
-        s3Client = createS3ClientFromConf(conf);
         this.lfsClient = lfsClient;
 
         listAmount = conf.getInt(FS_LAKEFS_LIST_AMOUNT_KEY, DEFAULT_LIST_AMOUNT);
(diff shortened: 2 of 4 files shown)


clients/hadoopfs/README.md
```diff
@@ -0,0 +1,7 @@
+# hadoop-lakefs file system
```
arielshaqed, line 1: Not sure this is helpful without explaining some more targets and options.  Also not needed on this PR?

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -418,32 +381,33 @@ private boolean renameObject(LakeFSFileStatus srcStatus, Path dst) throws IOExce
...
-     * @param e the exception to translate
+     * @param e   the exception to translate
      * @return an IOException that corresponds to the translated API exception
      */
     private IOException translateException(String msg, ApiException e) {
         int code = e.getCode();
         switch (code) {
             case HttpStatus.SC_NOT_FOUND:
-                return (IOException)new FileNotFoundException(msg).initCause(e);
+                return (IOException) new FileNotFoundException(msg).initCause(e);
```
arielshaqed, line 408: Why do we even need the casts?  Just to be sure it's not some random RuntimeException we're throwing?

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -4,8 +4,14 @@
 import static io.lakefs.Constants.FS_LAKEFS_LIST_AMOUNT_KEY;
 import static io.lakefs.Constants.SEPARATOR;
 
+import java.beans.IntrospectionException;
```
arielshaqed, line 7: AFAICS we need to organize imports here.

clients/hadoopfs/src/main/java/io/lakefs/MetadataClient.java
```diff
@@ -0,0 +1,89 @@
...
+    /**
+     * get object metadata by physical address
+     * @param physicalUri physical uri of object
+     * @return ObjectMetadata filled with Etag and content length
+     * @throws IOException
+     */
+    ObjectMetadata getObjectMetadata(URI physicalUri) throws IOException {
+        String bucket = physicalUri.getHost();
+        String key = physicalUri.getPath().substring(1);
+        if (this.s3client != null && this.getMetadata != null) {
```
arielshaqed, line 49: I don't understand.  When is this method better than stealing the file status from S3A?  This will not be more efficient than `getFileStatus` on the network... but it might be less one day if we can tell S3A to cache things.  And it preserves a dependency on AWS code, which anyway changed version significantly since the Hadoop-AWS 2.7.7 days.  And we anyway need the stolen ETag from S3AFIleStatus method.  Can we drop this version and keep only that one?

clients/hadoopfs/src/main/java/io/lakefs/MetadataClient.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    private static String getEtagFromFileStatus(FileStatus fileStatus) throws IntrospectionException, InvocationTargetException, IllegalAccessException {
+        for (PropertyDescriptor pd : Introspector.getBeanInfo(fileStatus.getClass()).getPropertyDescriptors()) {
+            if (pd.getReadMethod() != null && pd.getName().equals("etag")) {
+                String etag = (String) pd.getReadMethod().invoke(fileStatus);
+                if (etag != null && !etag.isEmpty()) {
+                    return etag;
+                }
+            }
```
arielshaqed, line 84: This is quite an elaborate loop using introspection.  I think if you just used `Class` directly you would get a loop of the same length, but without many of the scary names.  It looks something like this should work:
```java
Class<FileStatus> cls = fileStatus.getClass();
Field etagField = cls.getDeclaredField("etag");
if (etagField == null) throw new HorridException();
String etag = etagField.getString(fileStatus);
```
(plus handle 1-2 exceptions).

arielshaqed (changes requested):
Thanks!

I think we can (and probably should) tighten this up a bit.  Also get rid of introspection, as the name "bean" implies it's often quite heavyweight.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
nopcoder: Will check - I think this code was updated because I triggered "code format".

nopcoder: The method `initCause` returns `Throwable` and the original exception type. fixing.

clients/hadoopfs/src/main/java/io/lakefs/MetadataClient.java
nopcoder: Can drop the beans code - thought it will be faster as should be the standard way to access properties and it should be using cache.

clients/hadoopfs/src/main/java/io/lakefs/MetadataClient.java
nopcoder: At first, there was just 'getFileStatus' and extract 'etag' - everything worked, except on old Spark cluster 2.4.x where the S3AFileStatus didn't contain that value.

Switched to extracting the s3 client from the underlying file system, as a fallback. But, it caused two calls one for the FileStatus and one for getObjectMetadata on the s3 client in case you are on the older version.
I understand the issue that we have when we perform s3 call on-call version.
Just to be clear that the older version will work a little harder in order for us to not use the s3 client in the newer version.

clients/hadoopfs/README.md
nopcoder: Will remove

nopcoder:
@arielshaqed do you think I should use the same 'fs' as the 'create' call uses, or the current cached 'fs' from `initialize' will be good enough?

nopcoder:
@arielshaqed I prefer the fs instance used by the `create` call.
so updated the PR - let me know what you think.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -402,15 +382,22 @@ private boolean renameObject(LakeFSFileStatus srcStatus, Path dst) throws IOExce
      * @return an IOException that corresponds to the translated API exception
      */
     private IOException translateException(String msg, ApiException e) {
+        IOException ex;
         int code = e.getCode();
         switch (code) {
             case HttpStatus.SC_NOT_FOUND:
-                return (IOException) new FileNotFoundException(msg).initCause(e);
+                ex = new FileNotFoundException(msg);
+                ex.initCause(e);
```
arielshaqed, line 390: Cool!  Can we also _not_ break it up into 2 lines, matching the [documentation on oracle.com](https://docs.oracle.com/javase/7/docs/api/java/lang/Throwable.html#initCause(java.lang.Throwable))?

clients/hadoopfs/src/main/java/io/lakefs/MetadataClient.java
```diff
@@ -0,0 +1,68 @@
...
+        try {
+            Method getETagMethod = fileStatus.getClass().getMethod("getETag");
+            String etag = (String) getETagMethod.invoke(fileStatus);
+            // return the two properties over object metadata for easy fallback
+            ObjectMetadata o = new ObjectMetadata();
+            o.setContentLength(fileStatus.getLen());
+            o.setHeader("ETag", etag);
+            return o;
+        } catch (InvocationTargetException | IllegalAccessException | NoSuchMethodException e) {
+            LOG.trace("failed to get etag from file status", e);
```
arielshaqed, line 51: Do we _ever_ expect this to occur?  Basically I don't see this occurring, so I'd like a warning here.  But if I'm missing a case, then sure...

clients/hadoopfs/src/main/java/io/lakefs/MetadataClient.java
```diff
@@ -19,71 +17,52 @@
...
-                String etag = (String) pd.getReadMethod().invoke(fileStatus);
-                if (etag != null && !etag.isEmpty()) {
-                    return etag;
-                }
-            }
+        // fallback - use the underlying s3 client, get object metadata
+        try {
+            // cache s3 client and get object metadata method
+            Method amazonS3ClientGetter = fs.getClass().getDeclaredMethod("getAmazonS3Client");
+            amazonS3ClientGetter.setAccessible(true);
```
arielshaqed, line 58: Of course.  How silly of me to forget :-)

arielshaqed (approved):
Thanks!

Still a few questions, but I like it!

arielshaqed:
> @arielshaqed do you think I should use the same 'fs' as the 'create' call use, or the current cached 'fs' from `initialize' will be good enough? I can turn `MetadataClient` to be just a single method in LakeFSFileSystem.

Using the same `fs` in `create` is always better. :-)

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
nopcoder: sure - I'll revert the code back and match the exception type

clients/hadoopfs/src/main/java/io/lakefs/MetadataClient.java
nopcoder: sure, this will happen all the time in case if old spark = NoSuchMethodException.
I can put the specific one in ignoring and leave the trace in warn level, it will be a better way to handle this one.

clients/hadoopfs/src/main/java/io/lakefs/MetadataClient.java
nopcoder: just fyi - Databricks got a method called 's3()' in public that does give access to the underlying s3 client.

talSofer (commented):
Nice! 
Are you planning to add unit tests?

clients/hadoopfs/src/main/java/io/lakefs/MetadataClient.java
nopcoder: @talSofer getAmazonS3Client was part of the S3A filesystem implementation

clients/hadoopfs/src/main/java/io/lakefs/MetadataClient.java
```diff
@@ -0,0 +1,69 @@
...
+
+        // use underlying filesystem to get the file status and extract
+        // content length and etag (using reflection)
+        Path physicalPath = new Path(physicalUri.getPath());
+        FileStatus fileStatus = this.fs.getFileStatus(physicalPath);
+        try {
+            Method getETagMethod = fileStatus.getClass().getMethod("getETag");
+            String etag = (String) getETagMethod.invoke(fileStatus);
+            // return the two properties over object metadata for easy fallback
+            ObjectMetadata o = new ObjectMetadata();
```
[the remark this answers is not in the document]
nopcoder: Go habits - fixing.

clients/hadoopfs/src/main/java/io/lakefs/MetadataClient.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.hadoop.fs.Path;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.net.URI;
+
+public class MetadataClient {
```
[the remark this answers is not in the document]
nopcoder: updated

PR #2203, opened 2021-07-06
Add design document for loading TWO rocksdbjni JARs into ONE Spark

arielshaqed:


arielshaqed:
(This is what I'm writing, not what I've written...)

itaiad200 (approved):
Great docs!

arielshaqed:
> Great docs!

Thanks! Now I just have to implement something like it... 🤷

PR #2212, opened 2021-07-08
Add support for fixing spark location placeholder on copy

guy-har:


cmd/lakectl/cmd/config/config.go
```diff
@@ -40,6 +40,7 @@ type configuration struct {
 			Region    string `mapstructure:"region"`
 			CatalogID string `mapstructure:"catalog_id"`
 		}
+		FixSparkPlaceholder bool `mapstructure:"fix_spark_placeholder"`
```
arielshaqed, line 43: Please document "Spark location placeholder", it might not be clear to readers.  Also I suspect it might not be unique to Spark, so maybe we can document what it is/does.

arielshaqed (commented):
Just a drive-by "please document xyzzy" because I don't understand the meaning of the added OpenAPI parameter.

nopcoder (approved):

cmd/lakectl/cmd/config/config.go
guy-har: Will do, I considered and eventually decided not to.
Thanks!

PR #2221, opened 2021-07-11
docs: update commands with base-uri additional information

nopcoder:


docs/reference/commands.md
```diff
@@ -63,6 +63,27 @@ lakectl is a CLI tool allowing exploration and manipulation of a lakeFS environm
 ```
 
 
+### lakectl base-uri
+
+Base URI value is a lakefs reference or part of a reference that starts lakectl uses for each argument that accept reference and doesn't start with `lakefs://` prefix.
+The value for base URI is set by a flag `--base-uri` and the default value when not set is the set by an environment variable LAKECTL_BASE_URI.
+
+Having the environemnt varilabe sets to a repository, for example, can shorten your command by passing relative path:
```
guy-har, line 71:
```suggestion
Setting the environment variable to a repository will allow you to use only the relative path, enabling you to shorten your commands, for example:
```

docs/reference/commands.md
```diff
@@ -63,6 +63,27 @@ lakectl is a CLI tool allowing exploration and manipulation of a lakeFS environm
 ```
 
 
+### lakectl base-uri
+
+Base URI value is a lakefs reference or part of a reference that starts lakectl uses for each argument that accept reference and doesn't start with `lakefs://` prefix.
```
guy-har, line 68:
```suggestion
Base URI allows you to use relative paths. When setting the Base URI, lakectl will add the Base URI as a prefix to all relative paths (paths not starting with lakefs://).
```

docs/reference/commands.md
```diff
@@ -63,6 +63,27 @@ lakectl is a CLI tool allowing exploration and manipulation of a lakeFS environm
...
+The value for base URI is set by a flag `--base-uri` and the default value when not set is the set by an environment variable LAKECTL_BASE_URI.
```
guy-har, line 69:
```suggestion
The value for base URI is set by the flag `--base-uri`. the default value could be set by the environment variable LAKECTL_BASE_URI.
```

guy-har (changes requested):
The documentation is not clear enough, gave some suggestions. Not sure they are much better

guy-har (approved):

PR #2222, opened 2021-07-11
TEST_WITH_SERVER_TIMEOUT env var boolean to control test server timeout

nopcoder:
Environment variable to control pkg/api test server timeout.
By default, the test server of the API controller uses a timeout handler that stops the client call.
This feature helps simulate real-world proxy timeout but makes life harder on a developer who likes to test client call or test code under the debugger.

using TEST_WITH_SERVER_TIMOUE=0 we can turn off the use of this handler.

### DIFF
--- a/pkg/api/serve_test.go
+++ b/pkg/api/serve_test.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"strconv"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/api/serve_test.go
```diff
@@ -147,7 +148,15 @@ func setupClientByEndpoint(t testing.TB, endpointURL string, accessKeyID, secret
 
 func setupServer(t testing.TB, handler http.Handler) *httptest.Server {
 	t.Helper()
-	server := httptest.NewServer(http.TimeoutHandler(handler, ServerTimeout, `{"error": "timeout"}`))
+	withServerTimeout, err := strconv.ParseBool(os.Getenv("TEST_WITH_SERVER_TIMEOUT"))
+	if err != nil {
+		// by default we test with server timeout
+		withServerTimeout = true
+	}
```
itaiad200, line 155: I would panic here instead of allowing `TEST_WITH_SERVER_TIMEOUT=NO` to be parsed as wanting server timeout. Or at least log the error. It's just a test so non-blocking.

nopcoder: can panic on invalid value - I was thinking that the main use case is the default - empty

itaiad200 (approved):
One cannot simply reject a PR with four repeating digits.

itaiad200 (approved):

PR #2223, opened 2021-07-12
Authorization docs fix example policies

### LINKED ISSUE #2218: policy schema enforces small letters
I think there is a bug in the API. Here you see that the required values need to be small letters for the  `effect` statement. whereas the documentation states and convention is  to use `"Allow"` , `"Deny"`. This should be the only accepted format in the schema: 

https://github.com/treeverse/lakeFS/blob/2c620a5db984ce96eb7b87a32209507486c57861/api/swagger.yml#L647

```
Error executing command: request failed: [400 Bad Request] request body has an error: doesn't match the schema: Error at "/statement/0/effect": value is not one of the allowed values
Schema:
  {
    "enum": [
      "allow",
      "deny"
    ],
    "type": "string"
  }

Value:
  "Allow"
```

itaiad200:
closes #2218

### DIFF
--- a/docs/reference/authorization.md
+++ b/docs/reference/authorization.md
@@ -56,7 +56,7 @@ There are 4 basic components to the system:
 
 **Resources** - A unique identifier representing a specific resource in the system - a repository, an object, a user, etc.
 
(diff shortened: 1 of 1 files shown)


talSofer (approved):
Looks good!

Issue #1847, opened 2021-05-02
Metadata client doesn't work on Databricks


johnnyaug:
Metadata client doesn't work on Databricks, probably due to a conflict in RocksJava version.

Steps to reproduce:
1. Have a public lakeFS installation ready (you can use Terraform), with a repository and a commit.
2. Start a Databricks cluster.
3. In the cluster configuration > _Libraries_ tab, click Install new.
4. Put in the Maven coordinates for the metadata client: `io.treeverse:lakefs-spark-client-301_2.12:0.1.0-SNAPSHOT`
5. Put in the repository: `https://s01.oss.sonatype.org/content/repositories/snapshots`
6. In the cluster configuration > _Configuration_ tab > _Advanced options_ > _Spark_ tab, configure the lakeFS client:
   ```
   spark.hadoop.lakefs.api.url http://<lakefs api endpoint>/api/v1/
   spark.hadoop.lakefs.api.access_key <lakefs access key>
   spark.hadoop.lakefs.api.secret_key <lakefs secret key>
   ```
7. In a Scala notebook attached to the cluster, run the following code:
   ```scala
   import io.treeverse.clients.LakeFSContext
   var df = LakeFSContext.newDF(spark, "example-repo", "your-lakefs-commit-id")
   df.show
   ```

The following error appears:
```
IllegalAccessError: tried to access field org.rocksdb.RocksObject.nativeHandle_ from class org.rocksdb.SstFileReader
```

The reason may be that Databricks has an old `rocksdb-jni` jar in the classpath.
A possible solution could be to shade the rocksdb jni in the lakeFS uber-jar, and install the uber-jar on Databricks instead of using the Maven coordinates. However, this will also be challenging since the underlying native library needs to somehow be wired to the shaded jars.

arielshaqed:
This smells like a shading issue, but with a JNI thrown in to complicate matters.

Per https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/invocation.html#library_version we may need to load different JNI versions into different ClassLoaders (the Spark hate of ClassLoaders might be striking us again).  I.e. we'd need to load parts of the RocksDB wrapper from a separate ClassLoader.

arielshaqed:
Shading really hard here, as is dynamic loading.

https://github.com/JohnSnowLabs/spark-nlp/issues/105 might be relevant (but _really_ old).

arielshaqed:
# Update :-(

[...]

PR #2225, opened 2021-07-12
Add a forwarding ClassLoader to help load 2 copies of rocksdbjni

arielshaqed:
Part of fixing #1847.

This implements most of the ClassLoader described in [design/double-rocksdbjni](https://github.com/treeverse/lakeFS/blob/master/design/double-rocksdbjni.md).  But it does not hook anything up yet, so does not affect anything (except a slight increase in client size).

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/utils/AsInterface.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/utils/AsInterface.scala
@@ -1,10 +1,52 @@
 package io.treeverse.utils
 
 import org.objectweb.asm.{ClassVisitor, MethodVisitor}
-import org.objectweb.asm.Opcodes.ASM8
+import org.objectweb.asm.Opcodes._
 import org.objectweb.asm.Type
 
+private object AsInterface {
+  def getName(typ: Type): String =
+    typ.getSort match {
+      case Type.OBJECT => typ.getInternalName
+      case Type.ARRAY => s"Array[${typ.getInternalName}]"
+      case _ => typ.getClassName
+    }
+
+  def loadFor(sort: Int) = {
+    import Type._
+    sort match {
+      case BOOLEAN | CHAR | SHORT | INT => ILOAD
+      case LONG => LLOAD
+      case FLOAT => FLOAD
+      case DOUBLE => DLOAD
+      case ARRAY | OBJECT => ALOAD
+      case _ => throw new IllegalArgumentException(s"cannot load type of sort ${sort}")
+    }
+  }
+
+  def returnFor(sort: Int) = {
+    import Type._
+    sort match {
+      case BOOLEAN | CHAR | SHORT | INT => IRETURN
+      case LONG => LRETURN
+      case FLOAT => FRETURN
+      case DOUBLE => DRETURN
+      case ARRAY | OBJECT => ARETURN
+      case VOID => RETURN
+      case _ => throw new IllegalArgumentException(s"cannot load type of sort ${sort}")
+    }
+  }
+
+  // Scala array equality is too hard: the best (only-ish) way is to use
+  // java.util.Arrays.equals, but that doesn't work because the variance
+  // between Array[A] and Array[Object] is not defined there.
+  def equals[T](a: Array[T], b: Array[T]): Boolean =
+    (a zip b).foldLeft(true){ case (e, (x, y)) => e && x == y }
+}
+
 private class AsInterface(cv: ClassVisitor, val ifaces: Map[String, Class[_]]) extends ClassVisitor(ASM8, cv) {
+  var className: String = ""
+
   override def visit(
     version: Int,
     access: Int,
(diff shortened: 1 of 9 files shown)


clients/spark/core/src/test/scala/io/treeverse/utils/AsInterfaceTest.scala
```diff
@@ -0,0 +1,71 @@
...
+  test("AsInterface: Boo as Foo") {
+    val booStream = this.getClass.getClassLoader.getResourceAsStream("io/treeverse/utils/Boo.class")
+    val booBytes = new ByteArrayOutputStream
+    ByteStreams.copy(booStream, booBytes)
+
+    val cr = new ClassReader(booBytes.toByteArray)
+    val cw = new ClassWriter(cr, 0)
+    val cc = new CheckClassAdapter(cw);
+    val ai = new AsInterface(cc, Map(("Foo", classOf[Foo])))
+    cr.accept(ai, 0)
```
itaiad200, line 27: Why are you repeating the logic of `AsInterfaceClassLoader` in a test?

clients/spark/core/src/test/scala/io/treeverse/utils/AsInterfaceTest.scala
```diff
@@ -0,0 +1,71 @@
...
+  test("Boo as Foo can use method taking and returning a Boo") {
+    val aicl = new AsInterfaceClassLoader(
+      getClass.getClassLoader,
+      Map(("io/treeverse/utils/Boo", classOf[Foo])))
+    val f: Foo = aicl.newInstance[Foo]("io.treeverse.utils.Boo", 9)
+    val f2: Foo = aicl.newInstance[Foo]("io.treeverse.utils.Boo", 3)
+    assert(f.addFoo(f2).foo(4) === (9 + 3) * 4)
+  }
+
+  test("Goo as Foo fails to call missing method at call time :-(") {
```
itaiad200, line 47: Why at call time? Can't we recognise the missing method during mapping or instance creation?

clients/spark/core/src/main/scala/io/treeverse/utils/AsInterface.scala
```diff
@@ -0,0 +1,167 @@
...
+  override def visit(
+      version: Int,
+      access: Int,
+      name: String,
+      signature: String,
+      superName: String,
+      interfaces: Array[String]
+  ) = {
+    val (maybeIface, newInterfaces) = ifaces get name match {
+      case Some(iface) => (Some(iface), interfaces :+ Type.getType(iface).getInternalName)
```
itaiad200, line 64: Need some explanation on this line.

clients/spark/core/src/main/scala/io/treeverse/utils/AsInterface.scala
```diff
@@ -0,0 +1,167 @@
...
+      case Some(iface) => {
+        // Generate wrappers for all methods of iface in the class.
+        iface.getMethods.foreach(method =>
+          generateForwardingMethod(ACC_PUBLIC | ACC_SYNTHETIC,
+                                   method.getName,
+                                   Type.getMethodDescriptor(method),
+                                   method.getExceptionTypes.map(e => Type.getType(e).getDescriptor)
+                                  )
+        )
+        // Generate wrapper for all constructors for the class.
```
itaiad200, line 80: Including private ones?

clients/spark/core/src/main/scala/io/treeverse/utils/AsInterface.scala
```diff
@@ -0,0 +1,167 @@
...
+    val (maybeIface, newInterfaces) = ifaces get name match {
+      case Some(iface) => (Some(iface), interfaces :+ Type.getType(iface).getInternalName)
+      case None        => (None, interfaces)
+    }
+    className = name
+    cv.visit(version, access, name, signature, superName, newInterfaces)
+
+    maybeIface match {
+      case Some(iface) => {
+        // Generate wrappers for all methods of iface in the class.
```
itaiad200, line 72:
```suggestion
        // Generate wrappers for all public methods of iface in the class.
```

clients/spark/core/src/main/scala/io/treeverse/utils/AsInterface.scala
```diff
@@ -0,0 +1,167 @@
...
+
+  def translateType(typ: Type) =
+    ifaces get AsInterface.undot(typ.getInternalName) match {
+      case Some(iface) => Type.getType(iface)
+      case None        => typ
+    }
+
+  def revTranslateType(typ: Type) =
+    revIfaces get typ.getInternalName match {
+      case Some(klassName) => Type.getObjectType(klassName)
```
itaiad200, line 101:
```suggestion
      case Some(className) => Type.getObjectType(className)
```

clients/spark/core/src/main/scala/io/treeverse/utils/AsInterface.scala
```diff
@@ -0,0 +1,167 @@
...
+    mv.visitCode()
+    // BUG(ariels): assumes non-static.
+    mv.visitVarInsn(ALOAD, 0) // load "this"
+
+    for (((argType, translatedArgType), index) <- (argTypes zip translatedArgTypes).zipWithIndex) {
+      mv.visitVarInsn(AsInterface.loadFor(argType.getSort), index + 1) // load arg
+      if (!argType.equals(translatedRetType)) {
+        mv.visitTypeInsn(CHECKCAST, argType.getInternalName)
+      }
+    }
```
itaiad200, line 158: I don't get this. What happens if the `CHECKCAST` fail?

clients/spark/core/src/main/scala/io/treeverse/utils/AsInterface.scala
```diff
@@ -0,0 +1,167 @@
...
+
+    mv.visitMethodInsn(INVOKESPECIAL, className, name, desc)
+    mv.visitInsn(AsInterface.returnFor(translatedRetType.getSort))
+
+    mv.visitMaxs(argTypes.length + 1, argTypes.length + 1)
```
itaiad200, line 163: What is this special handling?

itaiad200 (changes requested):

clients/spark/core/src/main/scala/io/treeverse/utils/AsInterface.scala
```diff
@@ -0,0 +1,167 @@
...
+      case VOID                         => RETURN
+      case _ => throw new IllegalArgumentException(s"cannot load type of sort ${sort}")
+    }
+  }
+
+  def undot(dotted: String) = dotted.replaceAll("\\.", "/")
+
+  // Scala array equality is too hard: the best (only-ish) way is to use
+  // java.util.Arrays.equals, but that doesn't work because the variance
+  // between Array[A] and Array[Object] is not defined there.
```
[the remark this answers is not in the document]
arielshaqed: Yup, I *used* to be stuck on Java-ish arrays and that code is from way back then.  But turns out I can even use just `sameElements`, which actually does what I want (and doesn't even care about the sequence type).  Fixed!

clients/spark/core/src/main/scala/io/treeverse/utils/AsInterface.scala
```diff
@@ -0,0 +1,167 @@
...
+  def equals[T](a: Seq[T], b: Seq[T]): Boolean =
+    (a zip b).foldLeft(true) { case (e, (x, y)) => e && x == y }
+}
+
+private class AsInterface(cv: ClassVisitor, val ifaces: Map[String, Class[_]])
```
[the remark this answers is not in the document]
arielshaqed: Added several hundred lines :shrug:.

clients/spark/core/src/main/scala/io/treeverse/utils/AsInterface.scala
arielshaqed: Nope.  Commented.

clients/spark/core/src/main/scala/io/treeverse/utils/AsInterface.scala
arielshaqed: Done, good catch!

clients/spark/core/src/main/scala/io/treeverse/utils/AsInterface.scala
arielshaqed: Done!

clients/spark/core/src/main/scala/io/treeverse/utils/AsInterface.scala
arielshaqed: An exception.  Which is exactly the exception that we want to happen here :-)

clients/spark/core/src/main/scala/io/treeverse/utils/AsInterface.scala
arielshaqed: Just noise.  Every JVM function needs to declare how many local variables and stack entries it uses.

clients/spark/core/src/main/scala/io/treeverse/utils/AsInterfaceClassLoader.scala
```diff
@@ -0,0 +1,77 @@
...
+import java.io.{ByteArrayOutputStream, PrintWriter}
+import com.google.common.io.ByteStreams
+
+class AsInterfaceClassLoader(
+    parent: ClassLoader,
+    ifaces: Map[String, Class[_]],
+    val pw: PrintWriter = null
+) extends ClassLoader(parent) {
+
+  val skip = ifaces.values.map(_.getName).toSet
```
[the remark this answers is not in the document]
arielshaqed: Added a comment.  These are the interfaces into which we shall translate.  It won't work to translate them when/if they appear.  And appear they will, because this ClassLoader will try to load a class that uses them (it translates to use them), and then try to load them.

clients/spark/core/src/test/scala/io/treeverse/utils/AsInterfaceTest.scala
arielshaqed: This tests `AsInterface` on its own, and is also a "simple" example of how `AsInterface` works.  Dunno, if you don't like it I shall remove it.

clients/spark/core/src/test/scala/io/treeverse/utils/AsInterfaceTest.scala
arielshaqed: Unfortunately not: it may be defined in a superclass, and there is no way to verify it at translation time.

In order to do it at construction time we would need to store all the required methods somewhere and test (that would work, but might be surprising).  In order to do it at translation time (which would give the best developer experience) we would need to load the entire class into *another* ClassLoader, and then go over its public methods.

We decided F2F to delay doing this.  If you now think it's too bad an experience (and I can certainly see why), let me know & I shall add something.

arielshaqed (commented):
Thanks!

PTAL... Do you also think we need similar comments, but more user-facing than developer-facing, to `AsInterfaceClassLoader`?

itaiad200:
> Thanks!
> 
> PTAL... Do you also think we need similar comments, but more user-facing than developer-facing, to `AsInterfaceClassLoader`?

Actually the usage, as the tests show, is really convenient and straight forward (given the circumstances). I was more worried about our ability to maintain this really important yet very advanced implementation, which is better now with the docs.

itaiad200 (approved):
Thanks for the adding the docs, I think it's much clearer now. LGTM! Really AMAZING job with this.

arielshaqed (commented):
Thanks!  Especially for insisting on some more internal documentation -- I feel that it improves the code.
Pulling once tests pass (they failed because `scalafmt` enforces a *strange* Scaladoc style).

Issue #2214, opened 2021-07-08
LakeFS API client is sensitive to trailing slashes


talSofer:
### Steps to reproduce
- Run spark with lakeFSFileSystem, and configure its lakefs endpoint to end with "/", e.g. spark.hadoop.fs.lakefs.endpoint=https://example.lakefs/api/v1/ 
- Try to use the file system to write a file
- The write will fail silently (as a result of #2190), and no objects are written to lakeFS. 

### Potential solutions
1. Handle trailing slashes on the JavaOpenAPI client by adding a custom layer as we did with the Python OpenAPI client. 
2. Make the LakeFSFileSystem sanitize the value configured for spark.hadoop.fs.lakefs.endpoint

johnnyaug:
Went with solution #2.
To do #1, we _could_ extend the ApiClient class of the Java API client with one of our own. However, this means that all users will need to remember to use it rather than the standard ApiClient - which is essentially the same as remembering to trim the "/".

PR #2230, opened 2021-07-13
strip "/" from api base url

johnnyaug:
Solves #2214.
Was hoping for a solution on the generated code side, but I don't see a fitting place for it.


clients/hadoopfs/src/main/java/io/lakefs/LakeFSClient.java
```diff
@@ -31,6 +31,9 @@ public LakeFSClient(Configuration conf) throws IOException {
 
         ApiClient apiClient = io.lakefs.clients.api.Configuration.getDefaultApiClient();
         String endpoint = conf.get(Constants.FS_LAKEFS_ENDPOINT_KEY, "http://localhost:8000/api/v1");
+         if (endpoint.endsWith("/")) {
```
itaiad200, line 34: indentation.

itaiad200 (approved):

itaiad200 (approved):

PR #2237, opened 2021-07-14
Fix NPE while reading partitioned parquet on Databricks

### LINKED ISSUE #2111: hadoop-lakefs test app read parquet fails on databricks
Using `test/lakefsfs` under Databricks Spark v3 the application reaches the point where it reads the 
 data from lakeFS and fails with the following exception:

[stderr](https://github.com/treeverse/lakeFS/files/6644038/stack.txt)
[log4j-active](https://github.com/treeverse/lakeFS/files/6644043/log4j-active.3.txt)

talSofer:
Fixes #2111

### Problem 
Partition parquet reads with LakeFSFileSystem on Databricks fail due to a NullPointerException. 

### Diagnosis
- The NPE occurred when a `LakeFSFileSystem.listStatus` operation is invoked on a directory that has directories in it. and therefore, fail partitioned parquet reads (partitions are represented as directories for each partition value). 
- For directories, `LakeFSFileSystem.listStatus` returns an array of LocatedFileStatus objects, as opposed to an array of FileStatus objects returned by `S3AFileSystem.listStatus`. 
- When a LocatedFileStatus is returned, spark tries to get its BlockLocation, but the LakeFSFileSystem sets it to be null for directories, causing the NPE. 

### Solution 
Make `LakeFSFileSystem.listStatus` return an array of FileStatus objects instead of LocatedFileStatus. This solution brings the LakeFSFileSystem implementation closer to the S3AFileSystem implementation.

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -444,7 +444,7 @@ public class LakeFSFileSystem extends FileSystem {
         ListingIterator iterator = new ListingIterator(path, false, listAmount);
         while (iterator.hasNext()) {
             LocatedFileStatus fileStatus = iterator.next();
-            fileStatuses.add(fileStatus);
+            fileStatuses.add(((LakeFSLocatedFileStatus)fileStatus).toLakeFSFileStatus());
         }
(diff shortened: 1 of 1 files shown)


nopcoder (approved):
like it when a small change fixes such a big problem ;)

arielshaqed (approved):
Almost _too_ easy...

PR #2240, opened 2021-07-14
Support post-merge and post-commit events

### LINKED ISSUE #2182: Support post-merge and post-commit hooks
This is useful in order to notify downstream systems.
In particular, this will allow systems that integrate with lakeFS using `symlink.txt` files (sometimes known as `SymlinkInputFormat`) to automatically generate these files on commit/merge, and to register them into an existing metastore such as Glue Data Catalog or a self-managed HSM.

itaiad200:
closes #2182 
Adds 2 more hooks triggering events: `post-commit` and `post-merge`. Both of them are non-blocking. For example, if a commit succeeded, the appropriate action will be triggered and the commit action will return without waiting for the `post-commit` run. To support it, the `ActionService` now has a draining mechanism for any ongoing running hook when the system shuts down.

### DIFF
--- a/nessie/hooks_test.go
+++ b/nessie/hooks_test.go
@@ -127,7 +127,7 @@ func TestHooksSuccess(t *testing.T) {
 
 	commitRecord := commitResp.JSON201
 	require.Equal(t, "pre-commit", preCommitEvent.EventType)
-	require.Equal(t, "Test Commit", preCommitEvent.ActionName)
+	require.Equal(t, "Test Pre Commit", preCommitEvent.ActionName)
 	require.Equal(t, "test_webhook", preCommitEvent.HookID)
 	require.Equal(t, repo, preCommitEvent.RepositoryID)
 	require.Equal(t, branch, preCommitEvent.BranchID)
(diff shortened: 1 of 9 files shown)


nessie/hooks_test.go
```diff
@@ -83,6 +98,17 @@ func TestHooksSuccess(t *testing.T) {
 	uploadResp, err := uploadContent(ctx, repo, branch, "_lakefs_actions/testing_pre_commit", preCommitAction)
 	require.NoError(t, err)
 	require.Equal(t, http.StatusCreated, uploadResp.StatusCode())
+
+	actionPostCommitTmpl := template.Must(template.New("action-post-commit").Parse(actionPostCommitYaml))
```
nopcoder, line 102: now that we got to the point where we copy this one for the third time, we can start reuse and extract to a func.

nopcoder (changes requested):

nessie/hooks_test.go
```diff
@@ -145,6 +186,26 @@ func TestHooksSuccess(t *testing.T) {
 	require.Equal(t, "main", run.Branch)
 }
 
+func parseAndUpload(t *testing.T, ctx context.Context, repo, branch, yaml, templateName, actionPath string) {
```
nopcoder, line 189: call `t.Helper()` at start so we will have the error report the caller as the test that failed

nopcoder (approved):
lgmt - minor comment

PR #2242, opened 2021-07-14
Invoke Terraform Workflow on push to Cryptolake

adunuthulan:
When there is a push to the cryptolake branch, the nessie workflow invokes the deploy-lakefs-on-ecs.yaml workflow in the terraform repository, which creates/updates a long lived lakeFS environment for Cryptolake using the image generated in the nessie workflow. The name of the environment is cryptolake-prod and it can be accessed from https://cryptolake-prod.lakefs.dev/

This workflow is run here within nessie so the deployed image is certain to have passed all the tests before being deployed.

When the deployed version of lakeFS should be upgraded is still an open question. It can be updated on release, on a schedule, or manually by pushing to this branch.


.github/workflows/nessie.yaml
```diff
@@ -492,3 +493,17 @@ jobs:
...
         #   files: ./nessie-cover.out
         #   fail_ci_if_error: false
+
+  invoke-terraform-workflow:
+    name: Invoke Terraform Workflow to deploy Cryptolake ECS container
+    needs: [deploy-image, spark2, spark3, run-system-aws-s3, run-system-gcp-gs, run-system-azure-abfs]  
+    if: github.ref == 'refs/heads/cryptolake'
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Invoke Terraform Workflow
```
itaiad200, line 503: Missing some information about that it is done using a `workflow_dispatch` event to the Treeverse's `terraform` repository.

itaiad200 (approved):
Nicely done!
Approving with a single comment.

arielshaqed (changes requested):
Neat!  It's a bit confusing that the branch is called `cryptolake` and the env is called `cryptolake-prod`, but fair enough.

I am missing some history here, there's this decision here that I would like to see documented on this PR for others to: why does running *nessie* (the system test) trigger a deployment of cryptolake?

adunuthulan:
@arielshaqed I wanted some way to ensure that the version of lakeFS merged to the cryptolake branch and deployed was a valid version that passed all the system tests. The way I decided on was to make the deployment dependent on if the system tests were passed. Is there a better way of ensuring the validity of a lakeFS version?

arielshaqed:
> @arielshaqed I wanted some way to ensure that the version of lakeFS merged to the cryptolake branch and deployed was a valid version that passed all the system tests. The way I decided on was to make the deployment dependent on if the system tests were passed. Is there a better way of ensuring the validity of a lakeFS version?

OK, I see! The usual tool that CI /CD systems provide for something like this is the confusingly named "dependency". It's called that because it is named for some old concept in Makefile, regardless of the way it's been used for the past 20 years.
Consider a separate workflow that depends on Nessie succeeding. https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run might be your friend here.

adunuthulan:
> > @arielshaqed I wanted some way to ensure that the version of lakeFS merged to the cryptolake branch and deployed was a valid version that passed all the system tests. The way I decided on was to make the deployment dependent on if the system tests were passed. Is there a better way of ensuring the validity of a lakeFS version?
> 
> OK, I see! The usual tool that CI /CD systems provide for something like this is the confusingly named "dependency". It's called that because it is named for some old concept in Makefile, regardless of the way it's been used for the past 20 years.
> Consider a separate workflow that depends on Nessie succeeding. https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run might be your friend here.

@arielshaqed I took a look at the event trigger! If I used this method, I would have to find some way of sending the sha-tag of the docker image built in nessie to this new workflow within lakeFS, or build a second image from within that workflow, and run this `curl` command using that sha-tag. 

This event trigger doesn't support passing in inputs (possibly because this is triggering on the completion of an outside workflow and it's probably not the best idea to make the internals of the nessie workflow open to external workflows). I can do this by using the same `curl` method used here to pass the sha-tag to a new workflow within the lakeFS repo using a `workflow_dispatch` trigger, which would then just run the same `curl` command to run the workflow in Terraform repo – that seems redundant!

One way I could do it using the `workflow_run` trigger is by saving the sha as an artifact and retrieving it in a different workflow, and then using the `curl` command (which can be done using this user made action but is not supported by any github provided action: https://github.com/dawidd6/action-download-artifact)

Another way using this trigger would be to build the image again, push it to ECR, and then send that tag all from within a new workflow, which seems a little wasteful.

I think the way it is currently makes sense, but I do see the benefits of separation. I'm not sure about the impacts of either of these trade-offs, though intuition tells me I should be careful of the user action breaking, and that the added cost of building the image again is not very large.

adunuthulan:
Aha, talking with Itai made me understand that the sha-tag can be obtained by running `git rev-parse --short HEAD` and doesn't need to be passed as an input, so this separation using a `workflow_run` trigger is possible! I'll try this out

.github/workflows/nessie.yaml
```diff
@@ -492,18 +492,4 @@ jobs:
...
-    runs-on: ubuntu-20.04
-    steps:
-      - name: Invoke Terraform Workflow
-        run: |
-          curl -u ":${{ secrets.PERSONAL_TOKEN }}" \
-          -X POST \
-          -H "Accept: application/vnd.github.v3+json" \
-          https://api.github.com/repos/treeverse/terraform/actions/workflows/deploy-lakefs-on-ecs.yaml/dispatches \
-          -d '{"ref":"master", "inputs": {"lakefs-version": "${{ needs.deploy-image.outputs.tag }}", "environment": "cryptolake-prod" }}'
+        #   fail_ci_if_error: false
```
itaiad200, line 495: Add new line.

.github/workflows/cryptolake.yaml
```diff
@@ -0,0 +1,31 @@
...
+    runs-on: ubuntu-20.04
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Extract version
+        shell: bash
+        run: echo "::set-output name=tag::sha-$(git rev-parse --short HEAD | sed s/^v//g)"
+        id: version
```
itaiad200, line 23: Check that the tag exists in ECR.

.github/workflows/cryptolake.yaml
```diff
@@ -0,0 +1,31 @@
+name: Cryptolake
+on:
+  workflow_run:
+    workflows:
+      - "Nessie"
+    branches:
+      - cryptolake
+    types:
+      - completed
```
itaiad200, line 9: Does `completed` means that the workflow ended successfully?

adunuthulan: Nope, we have to check later on to make sure that the workflow_run.conclusion == "success"

.github/workflows/nessie.yaml
arielshaqed: And why do we even have this line here?

arielshaqed (approved):
Cool, this looks like a really neat and useful automation!  I like how it's gitops (or maybe GitHubops) from the user perspective: I trigger a deploy by merging to `cryptolake` _and do nothing else_.  So as I user when I asked myself "but how do I roll back a deploy?", I realized `git branch cryptolake; git revert; git push origin` is a nice answer!

One last request (*_NOT_ on this PR...*): @itaiad200 do you think it would make sense to add a badge to our README about what version was deployed to our cryptolake, and whether that deployment succeeded?

adunuthulan:
> Cool, this looks like a really neat and useful automation! I like how it's gitops (or maybe GitHubops) from the user perspective: I trigger a deploy by merging to `cryptolake` _and do nothing else_. So as I user when I asked myself "but how do I roll back a deploy?", I realized `git branch cryptolake; git revert; git push origin` is a nice answer!
> 
> One last request (__NOT_ on this PR..._): @itaiad200 do you think it would make sense to add a badge to our README about what version was deployed to our cryptolake, and whether that deployment succeeded?

There's a small problem at the moment with the roll back functionality, being the migration of the postgres database. At the moment since I am getting the image using the sha-tag, there's not a way of knowing whether a `migrate up` or `migrate down` is needed, so if you're migrating down, you need to do that manually. I can check to see if it's possible!

itaiad200:
> One last request (__NOT_ on this PR..._): @itaiad200 do you think it would make sense to add a badge to our README about what version was deployed to our cryptolake, and whether that deployment succeeded?

Agree that a badge is a nice add-on. But not part of this PR though ;)

PR #2244, opened 2021-07-15
Wrap Python ApiClient to allow uploading most IOBase objects

### LINKED ISSUE #2241: Python Client unable to upload objects from a stream
I was putting together a demo pipeline and noticed that even though content in `upload_object()` can be a file_type, it needs to be an actual file on disk in order to be uploaded to lakefs. When generating data in a stream this requires a work around to write the data to a temp file, then reread that temp file to send in the upload. 

Here is what I was expecting would work and the error as well as what the workaround for this example is. Note that the `StringIO` object is already an open file, the same error happened when trying to use a `ByteIO` streamer as well

```
from faker import Faker
import logging
import os
import io
from random import randrange
from lakefs_client import models
import lakefs_client
from lakefs_client.client import LakeFSClient

# lakeFS credentials and endpoint
configuration = lakefs_client.Configuration()
configuration.username = os.getenv("ACCESS_KEY_ID")
configuration.password = os.getenv("SECRET_ACCESS_KEY")
configuration.host = 'https://lakefs.<redacted>.lmco.com/'
configuration.ssl_ca_cert = '/etc/ssl/certs/ca-certificates.crt'

client = LakeFSClient(configuration)

fake = Faker()
path = fake.pystr()

output = io.StringIO()

for line in range(randrange(1000)):
    output.write(f"{fake.name()},{fake.zipcode()},{fake.city()},{fake.state()},{fake.pyint()}\n")
client.objects.upload_object(repository='ai-summit', branch='pipeline-demo', path=f"{path}/00001.csv", content=output)
``` 

Error: 
```
~/.local/lib/python3.6/site-packages/lakefs_client/api_client.py in files_parameters(self, files)
    549                         "for %s must be open." % param_name
    550                     )
--> 551                 filename = os.path.basename(file_instance.name)
    552                 filedata = self.get_file_data_and_close_file(file_instance)
    553                 mimetype = (mimetypes.guess_type(filename)[0] or

AttributeError: '_io.StringIO' object has no attribute 'name'
```

Workaround: 
```
with open(f"tmp.txt", 'w+') as f:
    for line in range(randrange(1000)): # fake 100 lines per file
        f.write(f"{fake.name()},{fake.zipcode()},{fake.city()},{fake.state()},{fake.pyint()}\n")
with open(f"tmp.txt", "r+") as f:
    client.objects.upload_object(repository='ai-summit', branch='pipeline-demo', path=f"{path}/00001.csv", content=f)
```

arielshaqed:
Generated Python OpenAPI code needs upload contents to be File's... but that
is merely to try to guess a MIME type from them.  Wrap ApiClient (that makes
this decision) to patch other IOBase objects with an arbitrary name (with no
extension, so safe no guessable MIME type).

This is easier than writing and using alternative Python templates that will
avoid placing filenames on uploaded objects.  Also the RFC isn't clear to me
whether a filename should go there; see #2241 for some discussion of this.

Fixes #2241.

### DIFF
--- a/clients/python/lakefs_client/client.py
+++ b/clients/python/lakefs_client/client.py
@@ -4,12 +4,29 @@ import lakefs_client
 from lakefs_client.apis import ActionsApi, AuthApi, BranchesApi, CommitsApi, ConfigApi, HealthCheckApi, MetadataApi, \
     ObjectsApi, RefsApi, RepositoriesApi, StagingApi, TagsApi
 
+class _WrappedApiClient(lakefs_client.ApiClient):
+    """ApiClient that fixes some weirdnesses."""
+
+    # Wrap files_parameters to work with unnamed "files" (e.g. MemIOs).
+    def files_parameters(self, files=None):
+        if files is not None:
+            for (param_name, file_instances) in files.items():
+                i = 0
+                if file_instances is None:
+                    continue
+                for file_instance in file_instances:
+                    if file_instance is not None and not hasattr(file_instance, 'name'):
+                        # Generate a fake name.
+                        i += 1
+                        file_instance.name = f'{param_name}{i}'
+        return super().files_parameters(files)
+
 class LakeFSClient:
     def __init__(self, configuration=None, header_name=None, header_value=None, cookie=None, pool_threads=1):
         if configuration:
             configuration = LakeFSClient._ensure_endpoint(configuration)
-        self._api = lakefs_client.ApiClient(configuration=configuration, header_name=header_name,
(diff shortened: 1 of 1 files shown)


arielshaqed:
@tvanderwal-lmco could you please also review?  For some reason GitHub refuses to place you as a reviewer :-(

nopcoder (approved):
lgtm

arielshaqed:
Thanks!

nopcoder:
as you mention the http spec say this parameter I'd optional for multipart upload - so an issue to the code generator is possible. also open ali supports upload directly, but it will probably break the

PR #2247, opened 2021-07-15
preserve scheme for hadoop fs filesystem

nopcoder:
@arielshaqed found the following issue and a way to improve the use of lakeFS hadoop filesystem - Thanks!

### Problem

The lakefs filesystem uses the constant 'lakefs' as a scheme when formatting objects locations. 
This prevents instantiate multiple lakeFS implementations based on different schemes.

### Solution

Using the scheme passed to the implementation we will enable initiate multiple/different lakefs fs implementations. Each instance will use the scheme to format the object locations.

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -8,7 +8,6 @@ import io.lakefs.clients.api.model.*;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.*;
 import org.apache.hadoop.fs.permission.FsPermission;
-import org.apache.hadoop.fs.s3a.S3AFileSystem;
 import org.apache.hadoop.util.Progressable;
 import org.apache.http.HttpStatus;
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -48,6 +47,7 @@
     private LakeFSClient lfsClient;
     private int listAmount;
     private FileSystem fsForConfig;
+    private String scheme;
```
arielshaqed, line 50: Is this just a part of `uri`?

arielshaqed (approved):
Note that this affects #2238 -- and will *not* conflict with it in the places that matter most.

Cool, neat, thanks!

This should allow using lakeFSFS on *two* different lakeFS servers merely by configuring them on different names, maybe `lakefs1://` and `lakefs2://`.  That could be *very* useful, e.g. to distcp a bunch of files between repos on different servers.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
nopcoder: true, reused from the instance we already got. thanks!

PR #2248, opened 2021-07-15
Metastore tool | Add flag to support dbfs location

guy-har:


nopcoder (approved):
Looks good - think we need to refactor the copy and merge method to accept a struct with all the parameters required for the operation so the next required parameter will not refactor all call tree signatures.

PR #2249, opened 2021-07-15
everything lakefs docker compose file 

nopcoder:
A docker-compose file with everything we can throw into work with lakeFS.
The idea behind it is to have a quick way to demo different lakeFS integrations.


deployments/compose/docker-compose.yml
```diff
@@ -0,0 +1,142 @@
+version: "3"
+services:
+
```
guy-har, line 3: It's not everything without spark :)

guy-har (changes requested):
Looks good,
Thanks.
Please add the spark containers to the docker-compose as well.

deployments/compose/docker-compose.yml
nopcoder: Spark!

nopcoder: Added spark with 3 workers - note I didn't configure it to make use of lakeFS as a gateway or the underlying storage to be MINIO.

deployments/compose/docker-compose.yml
```diff
@@ -137,6 +137,52 @@ services:
...
+    deploy:
+      replicas: 3
+    links:
+      - lakefs:s3.docker.lakefs.io
+
+  spark-submit:
+    image: docker.io/bitnami/spark:3
+    profiles: ["client"]
+    volumes:
+      - ./:/local
```
guy-har, line 180:
```suggestion
    volumes:
      - ./targets:/targets
```

guy-har: Believe this is to support spark-submit from a local target, Not sure targets is the right name, but It should be in some subdirectory.

guy-har (approved):

deployments/compose/hive/wait-for
```diff
@@ -0,0 +1,79 @@
+#!/bin/sh
```
itaiad200, line 1: We have a `wait-for` script in this repo, can we reuse it?

nopcoder: I'll find an alternative and remove this copy of wait-for script

deployments/compose/docker-compose.yml
nopcoder: for general use the `local` mount will use to access any file in your env.

PR #2250, opened 2021-07-16
Fix LakeFS startup with Azure storage type

evelkey:
LakeFS is not able to start after a restart if using Azure Blob storage as it returns `https` as storage class and lakeFS uses `azure` as storage class. This is fixed by an extended check for exactly this case by this PR.

### DIFF
--- a/cmd/lakefs/cmd/run.go
+++ b/cmd/lakefs/cmd/run.go
@@ -231,7 +231,7 @@ func checkForeignRepos(ctx context.Context, logger logging.Logger, authMetadataM
 					logger.WithError(err).Fatalf("Failed to parse to parse storage type '%s'", nsURL)
 				}
 
-				if adapterStorageType != repoStorageType.String() {
+				if (adapterStorageType != repoStorageType.String() && adapterStorageType != "azure") || ("https" != repoStorageType.String() && adapterStorageType == "azure") {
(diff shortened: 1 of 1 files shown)


arielshaqed:
Hi @evelkey,
Thanks for the PR! I've gone ahead and assigned either of the people who know most about Azure urls to review. Please bear in mind that it's already weekend here so we may manage to review it only on Sunday.
 In the meantime I've kicked off our ci process on this PR.

nopcoder:
You can pass the flag -m that will skip this check at startup, until this issue will be resolved.

guy-har (changes requested):
Thanks for contributing, 
I prefer not to handle this as a special case, Separating the `StorageType.String()` method into two methods, one for returning the
storage type and one for returning the default schema could do the job.
I also noticed There is another validation in `LinkPhysicalAddress` in `controller.go`
that will also fail now.

guy-har (approved):
@evelkey Thanks for finding and resolving this issue!
Due to the importance of this issue, in order to publish a fix ASAP, I Approve.
Opening an issue so the change request will be resolved in an additional PR.

evelkey:
Thank you @guy-har, we tested it in our AKS deployment with Azure blob as the backend storage (needed to migrate from version 30 to 31) and all used features are now functional again.

PR #2255, opened 2021-07-18
Split StorageType.String

### LINKED ISSUE #2253: Fix LakeFS startup with Azure storage type
LakeFS is not able to start after a restart if using Azure Blob storage as it returns https as storage class and lakeFS uses azure as storage class.

itaiad200:
closes #2253 

Azure has a unique behaviour where the storage type (`azure`) doesn't match the scheme (`https` and `http`). Up until now the StorageType struct simply returned the Scheme as the String, which caused a mismatch with the adapter when checking for foreign repos (check #2253). 
This PR removes the `String` function which was used for both logic of `Scheme` and `MatchingAdapterType`, and creates those functions instead.

### DIFF
--- a/pkg/block/namespace.go
+++ b/pkg/block/namespace.go
@@ -21,7 +21,7 @@ var (
 	ErrInvalidNamespace = errors.New("invalid namespace")
 )
 
-func (s StorageType) MatchingAdapterPrefix() string {
+func (s StorageType) BlockstoreType() string {
 	switch s {
 	case StorageTypeAzure:
 		return "azure"
--- a/pkg/block/namespace.go
+++ b/pkg/block/namespace.go
@@ -21,7 +21,16 @@ var (
 	ErrInvalidNamespace = errors.New("invalid namespace")
 )
 
-func (s StorageType) String() string {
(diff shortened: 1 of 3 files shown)


itaiad200:
@evelkey Can you check if this solves your issue?

pkg/block/namespace.go
```diff
@@ -21,7 +21,16 @@ var (
 	ErrInvalidNamespace = errors.New("invalid namespace")
 )
 
-func (s StorageType) String() string {
+func (s StorageType) MatchingAdapterPrefix() string {
```
guy-har, line 24: MatchingAdapterPrefix is a bit confusing for me, it basically returns the storage type (BlockstoreType according to the adapter).

guy-har (approved):
Looks good, 
Thanks!

evelkey:
@itaiad200 Yes, it solves my issue and leads to a cleaner solution. Thanks for the quick fix.

pkg/block/namespace.go
itaiad200: Changed to `BlockstoreType`

PR #2257, opened 2021-07-18
Fix S3 exists

guy-har:


johnnyaug (approved):

pkg/block/s3/adapter.go
```diff
@@ -274,8 +274,8 @@ func (a *Adapter) Exists(ctx context.Context, obj block.ObjectPointer) (bool, er
 	}
 	_, err = a.s3.HeadObjectWithContext(ctx, &input)
 	if err != nil {
-		if aerr, ok := err.(awserr.Error); ok {
-			if aerr.Code() == s3.ErrCodeNoSuchKey {
+		if aerr, ok := err.(awserr.RequestFailure); ok {
+			if aerr.StatusCode() == http.StatusNotFound {
```
itaiad200, line 278: `errors.Is` and `errors.as` are cleaner and also return what you need here in case some error wrapping happened.

arielshaqed (approved):
:-)

Lovely interface you got from AWS :-/

PR #2258, opened 2021-07-19
Improve hooks docs and add an experimentation section

### LINKED ISSUE #2254: Add documentation with tools recommendations for experimenting with webhooks
We should have some sort of experimentation guide for devs developing lakeFS hooks.
One big downside of current experimentation with the hooks is the absence of alternatives to run a server to receive the requests, so that the actions wouldn't fail.
We should leverage tools like https://play.svix.com that manifests all events sent by lakeFS, while always returning `201` thus never blocking the actions.

itaiad200:
closes #2254


docs/setup/hooks.md
```diff
@@ -177,3 +199,35 @@ Example:
...
+
+## Experimentation
+
+It's sometimes easier to start experimenting with lakeFS webhooks, even before you have a running server to receive the calls.
+There are a couple of online tools that can intercept and display the webhook requests, one of them is Svix.
+
+1. Go to [play.svix.com](https://play.svix.com) and copy the URL address supplied by Svix.
+It should look like `https://api.relay.svix.com/api/v1/play/receive/<Random_Gen_String>/`
+   
+1. Upload the following action file to lakeFS under the path `_lakefs_actions\test.yaml` in the default branch:
```
johnnyaug, line 211:
```suggestion
1. Upload the following action file to lakeFS under the path `_lakefs_actions/test.yaml` in the default branch:
```

docs/setup/hooks.md
```diff
@@ -177,3 +199,35 @@ Example:
...
+         branches:
+      post-merge:
+         branches:
+   hooks:
+      - id: svix
+        type: webhook
+        properties:
+           url: "https://api.relay.svix.com/api/v1/play/receive/<Random_Gen_String>/"
+   ```
+1. Commit that file to the branch.
```
johnnyaug, line 231: Nice to have: add a coypable lakectl command for every step.
Nice to have 2: A screenshot from Svix after the hook has run

johnnyaug (approved):
Much better like this.
See minor/optional comments.

PR #2259, opened 2021-07-19
lakefs client configuration key lookup using specific scheme with fallback

nopcoder:
As part of supporting multiple schemes working with lakeFS file system, the configuration parameters should also look up the keys based on the same scheme.
In this PR the key, secret, and endpoint will try to lookup the specific key with fallback to the default 'lakefs'.

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/FSConfiguration.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/FSConfiguration.java
@@ -1,0 +1,48 @@
+package io.lakefs;
+
+import org.apache.hadoop.conf.Configuration;
+
+public final class FSConfiguration {
+    public static final String DEFAULT_SCHEME = "lakefs";
+
+    private static String formatFSConfigurationKey(String scheme, String key) {
+        return "fs." + scheme + "." + key;
+    }
+
+    /**
+     * lookup value from configuration based on scheme and key suffix.
+     * first try to get "fs.\<scheme\>.\<key suffix\>", if value not found use the default scheme
+     * to build a key for lookup.
+     * @param conf configuration object to get the value from
+     * @param scheme used to format the key for lookup
+     * @param keySuffix key suffix to lookup
+     * @return key value or null in case no value found
+     */
+    public static String get(Configuration conf, String scheme, String keySuffix) {
+        String key = formatFSConfigurationKey(scheme, keySuffix);
+        String value = conf.get(key);
+        if (value == null && !scheme.equals(DEFAULT_SCHEME)) {
+            key = formatFSConfigurationKey(DEFAULT_SCHEME, keySuffix);
+            value = conf.get(key);
+        }
+        return value;
+    }
+
+    /**
+     * lookup value from configuration based on scheme and key suffix, returns default in case of null value.
+     * @param conf configuration object to get the value from
+     * @param scheme used to format key for lookup
+     * @param keySuffix key suffix to lookup
+     * @param defaultValue default value returned in case of null
+     * @return value found or default value
+     */
+    public static String get(Configuration conf, String scheme, String keySuffix, String defaultValue) {
+        String value = get(conf, scheme, keySuffix);
+        return (value == null) ? defaultValue : value;
+    }
+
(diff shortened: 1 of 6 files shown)


clients/hadoopfs/src/main/java/io/lakefs/LakeFSClient.java
```diff
@@ -14,24 +14,30 @@
...
+
+        String secretKey = getFSConfigurationValue(conf, scheme, "secret.key");
         if (secretKey == null) {
             throw new IOException("Missing lakeFS secret key");
         }
 
         ApiClient apiClient = io.lakefs.clients.api.Configuration.getDefaultApiClient();
-        String endpoint = conf.get(Constants.FS_LAKEFS_ENDPOINT_KEY, "http://localhost:8000/api/v1");
-        if (endpoint.endsWith("/")) {
+        String endpoint = getFSConfigurationValue(conf, scheme, "endpoint");
```
arielshaqed, line 37: Can we add an optional 3rd parameter `defaultValue`?  Then we can even handle an error if `defaultValue == null` and the variable was not set.

arielshaqed (approved):
Sorry, missed that we're ready for review.

Neat, thanks!  There are some extra cleanups you might add (but I'm cool either way).

clients/hadoopfs/src/main/java/io/lakefs/FSConfiguration.java
```diff
@@ -0,0 +1,48 @@
...
+    public static String get(Configuration conf, String scheme, String keySuffix, String defaultValue) {
+        String value = get(conf, scheme, keySuffix);
+        return (value == null) ? defaultValue : value;
+    }
+
+    public static int getInt(Configuration conf, String scheme, String keySuffix, int defaultValue) {
+        String valueString = get(conf, scheme, keySuffix);
+        return (valueString == null) ? defaultValue : Integer.parseInt(valueString);
+    }
+}
```
arielshaqed, line 48: Nit: newline.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSClient.java
```diff
@@ -14,23 +14,29 @@
  * This class uses the configuration to initialize API client and instance per API interface we expose.
  */
 public class LakeFSClient {
+    public static final String DEFAULT_ENDPOINT = "http://localhost:8000/api/v1";
     private static final String BASIC_AUTH = "basic_auth";
+    public static final String ACCESS_KEY_KEY_SUFFIX = "access.key";
+    public static final String SECRET_KEY_KEY_SUFFIX = "secret.key";
+    public static final String ENDPOINT_KEY_SUFFIX = "endpoint";
```
arielshaqed, line 21: These could still be in `Constants`.  Why move them now?

clients/hadoopfs/src/test/java/io/lakefs/FSConfigurationTest.java
```diff
@@ -0,0 +1,47 @@
+package io.lakefs;
+
+import junit.framework.TestCase;
+import org.apache.hadoop.conf.Configuration;
+import org.junit.Assert;
+
+public class FSConfigurationTest extends TestCase {
+
+    public void testGet() {
```
arielshaqed, line 9: Wooah.... `extends TestCase`?!  No `@Test`!?  Looks like [you're writing JUnit3 tests](https://stackoverflow.com/questions/2635839/junit-confusion-use-extends-testcase-or-test).  Apart from falling afoul of my innate nitpickiness, this will (probably?) prevent any tests in this class from using `@Stuff` annotations.

Can you add `@Test` and remove the inheritance?

arielshaqed (approved):
Thanks!

Please change the test to be more JUnit4-ish and less JUnit3-ish.  (JUnit is already on version 5, of course...).

clients/hadoopfs/src/test/java/io/lakefs/FSConfigurationTest.java
nopcoder: Sure 🥺 I gave the IDE create this one - big mistake

PR #2262, opened 2021-07-19
Remove Benchmarks environment

itaiad200:
Been long time coming... Destroying the Benchmark environment since it wasn't working from April and no one noticed.
We can always revert it, or use the infrastructure to create an environment that is more useful. No point in wasting resources and time in maintaining it now.


arielshaqed (commented):
Not a fan.  Why not just turn it down, and leave the code?

itaiad200:
> Not a fan. Why not just turn it down, and leave the code?

We keep unused code in the repo? We can always revert it if we want to.
image

johnnyaug (approved):

arielshaqed (approved):
Lost this argument to a meme.

PR #2269, opened 2021-07-21
Bugfix/java api client publish action

johnnyaug:


arielshaqed (changes requested):
Thanks!

That *is* a bug, yuck.  IIUC the loopback thing is to skip blocking for the password to unlock the private key?

This is great, but I am unhappy about giving a 3rd party potentially untrusted action indefinite access to our keys.  I would be happier to fork the action repo and use our own copy.  If you do the fork and fix, I volunteer to go over the code.

Thanks!

.github/workflows/java-api-client.yaml
```diff
@@ -29,19 +29,19 @@ jobs:
         run: make client-java PACKAGE_VERSION=${{ steps.version.outputs.tag }}
       - name: Install secret key for signing
         run: |
-          cat <(echo -e '${{ secrets.OSSRH_GPG_SECRET_KEY }}') | gpg --batch --import
+          cat <(echo -e '${{ secrets.GPG_PRIVATE_KEY }}') | gpg --batch --import
```
arielshaqed, line 32: I would prefer to keep the old name.  Otherwise it gets *really* hard to know which secret is which from only their names.  E.g. suppose we later decide to add another action that needs to sign something else.

Also the old name puts it in line with other 2 `OSSRH_*` secrets below.

arielshaqed (approved):
Great fix, thanks!

.github/workflows/java-api-client.yaml
johnnyaug: Will do. Just to clarify, I had to change the secret value in order for this to work.
I don't know the problem with the previous value, since I can't extract it

PR #2279, opened 2021-07-23
Update to github.css

JamilAhmad:
Jamil Dev

### DIFF
--- a/docs/assets/css/pygments/github.css
+++ b/docs/assets/css/pygments/github.css
@@ -356,3 +356,7 @@ code{
 body .note-info code{
   color: #000!important;
 }
+.main ul li,.main ol li{
+  font-family: 'Rene-Bieder-Galano-Grotesque-Medium'!important;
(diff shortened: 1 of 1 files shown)


arielshaqed (changes requested):
Thanks for your submission! In accordance with many css style guides we do not allow overrides without a comment explaining why for future maintainers.

nopcoder:
Hi @JamilAhmad I've created https://github.com/treeverse/lakeFS/pull/2284 it is just a draft and not a final work on how the custom style can be applied without modifying github.css.
I don't have enough knowledge to sync everything you applied into it - there are missing parts in the 'toc' and 'code' which I will need more time to apply, but I'm sure for you it will go smoother.

Also, there is more information on how to customize the theme we use here https://github.com/pmarsceill/just-the-docs/blob/master/docs/customization.md

nopcoder:
Please include this change with the new pull request that handles the consumerization using custom.scss

PR #2290, opened 2021-07-26
Increase4 Nessie Spark jobs timeout

### LINKED ISSUE #2288: Nessie's Spark 3 tests are timing out
Few examples:
https://github.com/treeverse/lakeFS/runs/3161207911
https://github.com/treeverse/lakeFS/actions/runs/1067283196
https://github.com/treeverse/lakeFS/runs/3160892947

arielshaqed:
2 minutes is not enough time :-(

AFAICT we originally set it because the client could block indefinitely on
the lakeFS server and trying to resolve there.  But this no longer happens
so there is little point in failing fast when this does happen.

Fixes #2288.

### DIFF
--- a/.github/workflows/nessie.yaml
+++ b/.github/workflows/nessie.yaml
@@ -230,7 +230,7 @@ jobs:
         run: mvn -Passembly -Djar.finalName=client --batch-mode --update-snapshots package
 
       - name: Test lakeFS S3 with Spark 2.x thick client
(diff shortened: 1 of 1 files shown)


arielshaqed:
[Office Space Tom: I take the step which is timing out, and I increase its timeout](https://imgflip.com/i/5hni9a)

itaiad200 (approved):
Thanks!

PR #2292, opened 2021-07-26
publish hadoop lakefs using github workflow

nopcoder:


.github/workflows/publish-hadoop-lakefs.yaml
```diff
@@ -0,0 +1,54 @@
...
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: Tag/version to publish
+  push:
+    branches:
+      - chore/publish-hadoopfs
```
johnnyaug, line 12: TODO remove before merge

.github/workflows/publish-hadoop-lakefs.yaml
```diff
@@ -0,0 +1,54 @@
+name: publish hadoop lakefs
+
+on:
+  release:
```
johnnyaug, line 4: I think this workflow should be triggered only using "workflow_dispatch".
Also the version number (I think) should not be coupled with the lakeFS version.

When the developer wants to publish a new version, they change the version in pom manually, and run this workflow.

WDYT?

johnnyaug (changes requested):

.github/workflows/publish-hadoop-lakefs.yaml
nopcoder: will remove

johnnyaug (approved):

PR #2293, opened 2021-07-27
Retry DB connection on Migration

### LINKED ISSUE #2017: Retry DB connection on startup

adunuthulan:
Fixes #2017 

Retries the database connection after encountering any error while connecting for a migration task, specifically within the `getMigrate(params params.Database)` function. Both `lakefs run` and `lakefs setup` connect to the db to perform migration tasks, and will retry the connection if they encounter an error.

### DIFF
--- a/pkg/db/connect.go
+++ b/pkg/db/connect.go
@@ -2,9 +2,7 @@ package db
 
 import (
 	"context"
-	"errors"
 	"fmt"
-	"net"
 	"time"
 
 	"github.com/jackc/pgx/v4/pgxpool"
@@ -113,14 +111,13 @@ func tryConnectConfig(ctx context.Context, config *pgxpool.Config, log logging.L
 		if err == nil {
 			return pool, nil
 		}
-		netError := &net.OpError{}
-		if !(errors.As(err, &netError) && netError.Op == "dial") {
-			return nil, err
+		if !IsDialError(err) {
+			return nil, fmt.Errorf("error while connecting to DB: %w", err)
 		}
 		if a.More() {
-			log.WithError(err).Info("could not open DB: Trying again")
+			log.WithError(err).Info("could not connect to DB: Trying again")
 		}
 	}
 
-	return nil, fmt.Errorf("could not open DB: %w", err)
+	return nil, fmt.Errorf("could not connect to DB: %w", err)
 }
--- a/pkg/db/connect.go
+++ b/pkg/db/connect.go
@@ -115,9 +115,9 @@ func tryConnectConfig(ctx context.Context, config *pgxpool.Config, log logging.L
 			return nil, fmt.Errorf("error while connecting to DB: %w", err)
 		}
 		if a.More() {
-			log.WithError(err).Info("could not connect to DB: Trying again")
+			log.WithError(err).Info("Could not connect to DB: Trying again")
 		}
 	}
 
-	return nil, fmt.Errorf("could not connect to DB: %w", err)
+	return nil, fmt.Errorf("retries exhausted, could not connect to DB: %w", err)
 }
--- a/pkg/db/connect.go
+++ b/pkg/db/connect.go
@@ -111,7 +111,7 @@ func tryConnectConfig(ctx context.Context, config *pgxpool.Config, log logging.L
 		if err == nil {
 			return pool, nil
 		}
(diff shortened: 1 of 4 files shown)


pkg/db/migration.go
```diff
@@ -109,13 +110,28 @@ func getMigrate(params params.Database) (*migrate.Migrate, error) {
...
 	}
 	return m, nil
 }
 
+func tryNewWithSourceInstance(sourceName string, sourceInstance source.Driver, databaseURL string) (*migrate.Migrate, error) {
+	strategy := params.DatabaseRetryStrategy
+	var err error
+	for a := retry.Start(strategy, nil); a.Next(); {
+		m, err := migrate.NewWithSourceInstance(sourceName, sourceInstance, databaseURL)
+		if err == nil {
```
itaiad200, line 125: Without checking the actual error for the cause of the failure, how can we know that the failure was because of a connection and not a failure to perform the actual action?

pkg/db/connect.go
```diff
@@ -106,13 +103,7 @@ func normalizeDBParams(p *params.Database) {
...
-		retry.Exponential{
-			Initial: firstWait,
-			Factor:  waitGrowth,
-			Jitter:  true,
-		},
-	)
+	strategy := params.DatabaseRetryStrategy
 	var pool *pgxpool.Pool
 	var err error
 	for a := retry.Start(strategy, nil); a.Next(); {
```
itaiad200, line 109: Here too, the actual error isn't being checked. Does the package expose the error types?

pkg/db/migration.go
itaiad200: Also not sure if it's safe to try to migrate twice, it might failed half way through..

itaiad200 (changes requested):
Looks good, but I think we need to check for the returned errors and not retry on every possible error.

pkg/db/migration.go
adunuthulan: `migrate.NewWithSourceInstance` doesn't actually perform the migration, it only gets the Migrate instance

pkg/db/migration.go
```diff
@@ -125,8 +126,13 @@ func tryNewWithSourceInstance(sourceName string, sourceInstance source.Driver, d
 		if err == nil {
 			return m, nil
 		}
-
-		logging.Default().WithError(err).Error("could not open DB: Trying again")
+		netError := &net.OpError{}
+		if !(errors.As(err, &netError) && netError.Op == "dial") {
+			return nil, err
+		}
```
itaiad200, line 132: Extract this repeated logic to a common place.

itaiad200 (changes requested):
Single minor comment. Also - how was this tested?

nopcoder (approved):

pkg/db/migration.go
adunuthulan: I moved the logic to the errors.go file in the db package. I haven't seen code like this within other errors.go files, but an errors file seems like an appropriate place for code checking for errors. Is this okay?

The function is exported since it could conceivably be used elsewhere.

adunuthulan:
> Single minor comment. Also - how was this tested?

@itaiad200 This was tested by running the lakeFS image through VSCode. I put break points within both the 'lakefs run' code (where the first connection is made through `BuildDatabaseConnection()`) and 'lakefs setup' code (where the first connection is through `migrate.newWithSourceInstance()`), and I was able to confirm that running either command will perform the db connection retries before exiting execution, if postgres is not running yet.

pkg/db/migration.go
itaiad200: This file is the right place. As for exporting it, I would usually make it unexported until the first external use-case (which sometimes forces you to move it to an entirely different package anyhow). Not blocking anyways..

itaiad200 (approved):
Great work! see minor non-blocking comment..

PR #2295, opened 2021-07-27
Use the right bind flag name for minio

arielshaqed:
Otherwise `cd deployments/compose; docker-compose up` never brings up minio.


arielshaqed:
Bad one, sorry.

arielshaqed:
Was due to an old installed version.  Need to figure out how to force a new-enough version.

PR #2297, opened 2021-07-27
publish assemly and s3 upload

nopcoder:


johnnyaug (approved):
TODO remove push trigger

PR #2299, opened 2021-07-27
Test/lakefs fs contract tests

johnnyaug:
Implements Hadoop FS contract tests. Those are basically copied from the S3A contract tests and modified a little to match lakeFS.

We use docker-compose to spin up lakeFS over minio, and run the tests against this setup.

Some tests still do not pass and were commented out, #2309 was opened to address that. 

Links:
1. [Information about HadoopFS contracts](https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/filesystem/testing.html).
1. [The S3A tests source in Hadoop](https://github.com/apache/hadoop/tree/trunk/hadoop-tools/hadoop-aws/src/test/java/org/apache/hadoop/fs/contract/s3a)


clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -253,7 +253,12 @@ public boolean rename(Path src, Path dst) throws IOException {
         // Throws FileNotFoundException when src does not exist. mimics s3a's behaviour in
         // https://github.com/apache/hadoop/blob/2960d83c255a00a549f8809882cd3b73a6266b6d/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java#L1505
         LakeFSFileStatus srcStatus;
-        srcStatus = getFileStatus(src);
+        try {
+            srcStatus = getFileStatus(src);
+        }
+        catch (FileNotFoundException e) {
+            return false;
+        }
```
arielshaqed, line 261: This explicitly changes the behaviour documented on 253-254.  Is this intentional?  (IIRC `git blame` will show that the old behaviour is actually required by something S3A-ish, so we should see what S3A does for its contract).

clients/hadoopfs/src/test/java/io/lakefs/contract/LakeFSContract.java
```diff
@@ -0,0 +1,46 @@
+/*
```
arielshaqed, line 1: @ozkatz do we place licenses on our source files?

clients/hadoopfs/src/test/java/io/lakefs/contract/TestLakeFSContractDelete.java
```diff
@@ -0,0 +1,31 @@
...
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.contract.AbstractContractDeleteTest;
+import org.apache.hadoop.fs.contract.AbstractFSContract;
+
+public class TestLakeFSContractDelete extends AbstractContractDeleteTest {
+
+  @Override
+  protected AbstractFSContract createContract(Configuration conf) {
+    return new LakeFSContract(conf);
+  }
```
arielshaqed, line 30: This method is repeated in every contract class.  Is there a way to implement it _once_, as a mixin of some sort?  (I just read about putting methods inside Java interfaces, so obviously I want to implement mixins wherever possible... and this is [one way to use Java interface methods](https://en.wikipedia.org/wiki/Mixin#Interfaces_and_traits).

arielshaqed (commented):
Wow, thanks... This looks really interesting, and a *seriously* good idea.  Is it testing [this contract](https://hadoop.apache.org/docs/r2.7.0/hadoop-project-dist/hadoop-common/filesystem/testing.html)?  If so, please add a link...

A few extra questions:
1. It looks like JUnit3!  Is that required by the contract tester?
2. If we delete the hadoopfs workflow, then where do we explicitly run the _other_ unit tests?

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
johnnyaug: Reverted

clients/hadoopfs/src/test/java/io/lakefs/contract/TestLakeFSContractDelete.java
johnnyaug: I agree it's not the nicest thing, but this is the way suggested by Hadoop's [documentation](https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/filesystem/testing.html#Testing_a_new_filesystem).

johnnyaug:
@arielshaqed, regarding your questions:
1. Some of it is indeed JUnit3 - it's because the base class `FileSystemContractBaseTest` is JUnit3.
2. I've changed the run command to be `mvn ... verify`. It runs all the tests, including the previously existing unit tests.

test/lakefsfs_contract/setup-test.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash -ex
+
+cp auth-keys.xml ../../clients/hadoopfs/src/test/resources/auth-keys.xml
+cp core-site.xml ../../clients/hadoopfs/src/test/resources/core-site.xml
```
nopcoder, line 4: is it possible not to copy these files and have them referenced?

test/lakefsfs_contract/setup-test.sh
```diff
@@ -0,0 +1,9 @@
...
+
+docker-compose exec -T lakefs /app/wait-for localhost:8000
+docker-compose exec -T lakefs sh -c 'lakefs setup --user-name tester --access-key-id ${TESTER_ACCESS_KEY_ID} --secret-access-key ${TESTER_SECRET_ACCESS_KEY}'
+aws s3api create-bucket --endpoint-url http://s3.local.lakefs.io:9000 --bucket test-bucket
+docker-compose exec -T lakefs lakectl repo create "lakefs://${REPOSITORY}" ${STORAGE_NAMESPACE} -d main
```
nopcoder, line 9:
```suggestion
docker-compose exec -T lakefs lakectl repo create "lakefs://${REPOSITORY}" "${STORAGE_NAMESPACE}" -d main
```

nopcoder (approved):
minor comments, looks good

clients/hadoopfs/src/test/java/io/lakefs/contract/LakeFSTestUtils.java
```diff
@@ -0,0 +1,38 @@
...
+    String fsname = conf.getTrimmed(TestLakeFSFileSystemContract.TEST_FS_LAKEFS_NAME, "");
+
+
+    boolean liveTest = !StringUtils.isEmpty(fsname);
+    URI testURI = null;
+    if (liveTest) {
+      testURI = URI.create(fsname);
+      liveTest = testURI.getScheme().equals("lakefs");
+    }
+    if (!liveTest) {
```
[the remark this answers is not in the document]
johnnyaug: Note that liveTest is potentially also changed inside the first block

clients/hadoopfs/src/test/java/io/lakefs/contract/TestLakeFSContractCreate.java
```diff
@@ -0,0 +1,24 @@
+package io.lakefs.contract;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.contract.AbstractContractCreateTest;
+import org.apache.hadoop.fs.contract.AbstractFSContract;
+import org.apache.hadoop.fs.contract.ContractTestUtils;
+
+public class TestLakeFSContractCreate extends AbstractContractCreateTest {
```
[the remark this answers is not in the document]
johnnyaug: These tests are not ignored, rather, extending the base class causes the tests to run

clients/hadoopfs/src/test/java/io/lakefs/contract/LakeFSTestUtils.java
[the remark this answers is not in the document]
nopcoder: my bad, looks good

clients/hadoopfs/pom.xml
```diff
@@ -142,11 +155,22 @@ To export to S3:
...
+                <plugin>
+                    <groupId>org.apache.maven.plugins</groupId>
+                    <artifactId>maven-surefire-plugin</artifactId>
+                    <version>2.12.4</version>
+                    <configuration>
+                        <excludes>
+                            <exclude>${exclude.tests}</exclude>
+                        </excludes>
+                    </configuration>
+                </plugin>
```
nopcoder, line 173: indent 4 left

nopcoder (approved):
:+1:

talSofer (approved):
Thanks, this is a really great addition!

PR #2300, opened 2021-07-27
Fix publish hadoop-lakefs assembly

nopcoder:


clients/hadoopfs/pom.xml
```diff
@@ -20,7 +20,7 @@ To export to S3:
     <artifactId>hadoop-lakefs${name-suffix}</artifactId>
     <packaging>jar</packaging>
     <name>hadoop-lakefs${name-suffix}</name>
-    <version>0.1.0</version><!--MARKER: keep this comment here, we update the version with the lakeFS version when on deploy-->
+    <version>0.1.1</version>
```
arielshaqed, line 23: PR description does not support removing this MARKER.  Is this intentional?

nopcoder: no this is the previous way I tested it

johnnyaug (approved):

PR #2301, opened 2021-07-28
Change config file references in docs from config.yaml to .lakefs.yaml

itaiad200:
Following a question in the help channel, it is best if we use the same terminology everywhere when referencing the configuration file.

### DIFF
--- a/docs/deploy/aws.md
+++ b/docs/deploy/aws.md
@@ -40,7 +40,7 @@ If you already have a database, take note of the connection string and skip to t
 ## Installation Options
 
 ### On EC2
-1. Save the following configuration file as `config.yaml`:
+1. Save the following configuration file as `.lakefs.yaml`:
 
    ```yaml
    ---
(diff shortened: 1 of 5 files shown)


nopcoder (commented):
if we guide the user to create `.lakefs.yaml` which is usually the default name we load from the home directory - we skip the use of `--config` flag later and specify the file and just specify that the file should be located under home.

itaiad200:
> if we guide the user to create `.lakefs.yaml` which is usually the default name we load from the home directory - we skip the use of `--config` flag later and specify the file and just specify that the file should be located under home.

Thought about it. Adding a note everywhere that the config file is being used from the home directory felt like too much. However I do want to keep that file naming as `.lakefs.yaml` so users would always know that we reference the same configuration. I think that an always-working-but-longer-example is better than a sometimes-working-shorter-example, WDYT?

nopcoder:
So two things - the previous example of using the full path to a config.yaml is right and works.
Creating a dot file and reference it with a full path looks off. It is usually used to configures something to work on the specific folder or some default usually found in the user home folder.
The mix of the two looks strange.
One way is to show how we would like our end users to configure the system and add notes about the custom way it can specify the configuration.
Alternatively, show the custom way and add notes about how the default will work.

johnnyaug (approved):

itaiad200:
Closing since it's not as confusing as we initially though. If this reoccur we'll find the middle ground and improve the wording of referencing the configuration file.

PR #2305, opened 2021-07-28
Add a lakeFSFileSystem Tracer

talSofer:
This tracer can be used to compare the flow of file system operations, and the output of file system operations done by LakeFSFileSystem and S3AFileSystem. 

Example usage - 
```shell
% grep --color '\[RESULTS_COMPARISON\]\[getFileStatus\]' tracer_logs.txt  
21/07/28 14:15:43 TRACE FileSystemTracer: [RESULTS_COMPARISON][getFileStatus] 
lakefs: LakeFSFileStatus{path=lakefs://test7/master/amazon-reviews-pds/parquet/year=2009; isDirectory=true; modification_time=0; access_time=0; owner=; group=; permission=rwxrwxrwx; isSymlink=false}, 
s3a: S3AFileStatus{path=s3://example-bucket/amazon-reviews-pds/parquet/year=2009; isDirectory=true; modification_time=0; access_time=0; owner=root; group=root; permission=rwxrwxrwx; isSymlink=false} isEmptyDirectory=FALSE
```

Looking at the output above we can immediately see the difference between what's returned from LakeFSFileSystem.getFileStatus and S3AFileSystem.getFileStatus:
- LakeFSFileSystem returns a LakeFSFileStatus while S3AFileSystem returns S3AFileStatus
- S3AFileStatus has and extra isEmptyDirectory field

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
@@ -23,36 +23,35 @@ import java.net.URI;
  * This configures the FileSystemTracer to be the file system that handles paths with the lakefs scheme. i.e. paths with
  * lakefs:// prefix.
  *
- * What does the FileSystemTracer do:
+ * How does the FileSystemTracer work:
  * The FileSystemTracer holds instances of {@link LakeFSFileSystem} and {@link S3AFileSystem}. On a file system operation,
  * the FileSystemTracer invokes the operation on both file systems, logs the output of both calls, and returns the result
- * of one of the file systems according to a configuration (S3AFileSystem output by default).
+ * of one of the file systems based on a configuration (S3AFileSystem output by default).
 
- * Additional configurations:
- * Mandatory -
- * - fs.lakefs.tracer.working.dir - the s3 location in which the tracer can operate. This should be an S3 bucket name or
+ * Configuration:
+ * - fs.lakefs.tracer.working_dir - the s3 location in which the tracer can operate. This should be an S3 bucket name or
  *   an absolute path of a directory on a bucket.
  * Optional -
- * - fs.lakefs.tracer.use.lakefs - tells the tracer whether it should return the response coming from the lakefs file
+ * - fs.lakefs.tracer.use_lakefs_output - tells the tracer whether it should return the response coming from the lakefs file
  *   system or return s3a's response. by default it is set to false and returns s3a's response.
  *
  * Assumptions:
- * - listing results of lakefs://repository/branch/$PATH and s3a://${fs.lakefs.tracer.working.dir}/$PATH are identical.
+ * - The content of lakefs://repository/branch/ and s3a://${fs.lakefs.tracer.working.dir}/ should be identical.
  * - The s3 credentials available for Spark allow access to fs.lakefs.tracer.working.dir on s3.
  */
 public class FileSystemTracer extends FileSystem {
 
-    public static final Logger LOG = LoggerFactory.getLogger(FileSystemTracer.class);
-    private static final String TRACER_WORKING_DIR = "fs.lakefs.tracer.working.dir";
-    private static final String USE_LAKEFS_RESPONSE = "fs.lakefs.tracer.use.lakefs";
-    private static final Object S3_URI_PREFIX = "s3";
+    private static final Logger LOG = LoggerFactory.getLogger(FileSystemTracer.class);
+    private static final String TRACER_WORKING_DIR = "fs.lakefs.tracer.working_dir";
+    private static final String USE_LAKEFS_OUTPUT = "fs.lakefs.tracer.use_lakefs_output";
+    private static final Object S3_URI_SCHEME = "s3";
     private static final Object RESULTS_COMPARISON = "[RESULTS_COMPARISON]";
 
 
     /*
      A property that determines which file system's response the FileSystemTracer returns.
      */
-    private boolean useLakeFSFileSystemRes;
+    private boolean useLakeFSFileSystemResults;
     private LakeFSFileSystem lfsFileSystem;
     private FileSystem s3AFileSystem;
     private String s3aPathPrefix;
@@ -62,18 +61,10 @@ public class FileSystemTracer extends FileSystem {
      * Transforms a lakefs path into an s3 path.
      * Example:
      *   in: lakefs://repository/branch/key=1/a.parquet
-     *   out: s3://${fs.lakefs.tracer.working.dir}/key=1/a.parquet
+     *   out: s3://${fs.lakefs.tracer.working_dir}/key=1/a.parquet
      */
     private Path translateLakeFSPathToS3APath(Path path) {
-        ObjectLocation objectLoc = lfsFileSystem.pathToObjectLocation(path);
-        String lakefsPath = path.toString();
-        String lakefsPrefix = String.format("%s://%s/%s", objectLoc.getScheme(), objectLoc.getRepository(),
-                objectLoc.getRef());
-
-        String s3aPath = lakefsPath.replace(lakefsPrefix, s3aPathPrefix);
-        LOG.trace("Converted {} to {}", path, s3aPath);
-
-        return new Path(s3aPath);
+        return replacePathPrefix(path, lfsPathPrefix, s3aPathPrefix);
     }
 
     /**
@@ -81,16 +72,25 @@ public class FileSystemTracer extends FileSystem {
      * This method is used to transform the output of S3AFileSystem operations that include paths, so that Spark
      * continues to direct it's requests to the FileSystemTracer that handles paths with the "lakefs://" prefix.
      * Example:
-     *   in: s3://${fs.lakefs.tracer.working.dir}/key=1/a.parquet
+     *   in: s3://${fs.lakefs.tracer.working_dir}/key=1/a.parquet
      *   out: lakefs://repository/branch/key=1/a.parquet
      */
     private Path translateS3APathToLakeFSPath(Path path) {
-        String p = path.toString();
-        String lfsPath = p.replace(s3aPathPrefix, lfsPathPrefix);
-        LOG.trace("Converted {} to {}", path, lfsPath);
-        return new Path(lfsPath);
+        return replacePathPrefix(path, s3aPathPrefix, lfsPathPrefix);
     }
 
+    private Path replacePathPrefix(Path path, String curPrefix, String newPrefix) {
+        String p = path.toString();
+        boolean isValidPath = p.startsWith(curPrefix);
+        if (isValidPath) {
+            String objRelativePath = ObjectLocation.trimLeadingSlash(p.substring(curPrefix.length()));
+            String newPath = String.format("%s/%s", newPrefix, objRelativePath);
+            LOG.trace("Converted {} to {}", path, newPath);
+            return new Path(newPath);
+        }
+        LOG.error("Invalid path {}", path);
+        return null;
+    }
 
     @Override
     public void initialize(URI name, Configuration conf) throws IOException {
@@ -100,12 +100,15 @@ public class FileSystemTracer extends FileSystem {
         lfsPathPrefix = String.format("%s://%s/%s", loc.getScheme(), loc.getRepository(), loc.getRef());
 
         String tracerWorkingDir = conf.get(TRACER_WORKING_DIR);
-        s3aPathPrefix = String.format("%s://%s", S3_URI_PREFIX, tracerWorkingDir);
+        if (tracerWorkingDir == null) {
+            throw new IOException("tracerWorkingDir is null");
+        }
+        s3aPathPrefix = String.format("%s://%s", S3_URI_SCHEME, tracerWorkingDir);
         Path s3aPath = new Path(s3aPathPrefix);
         s3AFileSystem = s3aPath.getFileSystem(conf);
 
-        useLakeFSFileSystemRes = conf.getBoolean(USE_LAKEFS_RESPONSE, false);
-        LOG.trace("Initialized FileSystemTracer, fs.lakefs.tracer.use.lakefs: {}", useLakeFSFileSystemRes);
+        useLakeFSFileSystemResults = conf.getBoolean(USE_LAKEFS_OUTPUT, false);
+        LOG.trace("Initialization finished, fs.lakefs.tracer.use_lakefs_output: {}", useLakeFSFileSystemResults);
     }
 
     @Override
@@ -116,7 +119,7 @@ public class FileSystemTracer extends FileSystem {
         URI s3aRes = s3AFileSystem.getUri();
         LOG.trace("{}[getUri] lakefs: {}, s3a: {}", RESULTS_COMPARISON, lakefsRes, s3aRes);
 
-        if (useLakeFSFileSystemRes) {
+        if (useLakeFSFileSystemResults) {
             return lakefsRes;
         }
         return s3aRes;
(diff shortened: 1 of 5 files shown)


clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
```diff
@@ -0,0 +1,353 @@
...
+    /**
+     * Transforms a lakefs path into an s3 path.
+     * Example:
+     *   in: lakefs://repository/branch/key=1/a.parquet
+     *   out: s3://${fs.lakefs.tracer.working.dir}/key=1/a.parquet
+     */
+    private Path translateLakeFSPathToS3APath(Path path) {
+        ObjectLocation objectLoc = lfsFileSystem.pathToObjectLocation(path);
+        String lakefsPath = path.toString();
+        String lakefsPrefix = String.format("%s://%s/%s", objectLoc.getScheme(), objectLoc.getRepository(),
```
nopcoder, line 70: relevant for multiple places - we have a static method to format 3 and 4 parts addresses at ObjectLocation.formatPath

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
```diff
@@ -0,0 +1,353 @@
...
+
+
+    @Override
+    public void initialize(URI name, Configuration conf) throws IOException {
+        lfsFileSystem = new LakeFSFileSystem();
+        lfsFileSystem.initialize(name, conf);
+        ObjectLocation loc = lfsFileSystem.pathToObjectLocation(new Path(name));
+        lfsPathPrefix = String.format("%s://%s/%s", loc.getScheme(), loc.getRepository(), loc.getRef());
+
+        String tracerWorkingDir = conf.get(TRACER_WORKING_DIR);
```
nopcoder, line 102: throw an exception in case value is not found (null check)

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
```diff
@@ -0,0 +1,353 @@
...
+ * Assumptions:
+ * - listing results of lakefs://repository/branch/$PATH and s3a://${fs.lakefs.tracer.working.dir}/$PATH are identical.
+ * - The s3 credentials available for Spark allow access to fs.lakefs.tracer.working.dir on s3.
+ */
+public class FileSystemTracer extends FileSystem {
+
+    public static final Logger LOG = LoggerFactory.getLogger(FileSystemTracer.class);
+    private static final String TRACER_WORKING_DIR = "fs.lakefs.tracer.working.dir";
+    private static final String USE_LAKEFS_RESPONSE = "fs.lakefs.tracer.use.lakefs";
+    private static final Object S3_URI_PREFIX = "s3";
```
nopcoder, line 48: schema?

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
```diff
@@ -0,0 +1,353 @@
...
+        ObjectLocation loc = lfsFileSystem.pathToObjectLocation(new Path(name));
+        lfsPathPrefix = String.format("%s://%s/%s", loc.getScheme(), loc.getRepository(), loc.getRef());
+
+        String tracerWorkingDir = conf.get(TRACER_WORKING_DIR);
+        s3aPathPrefix = String.format("%s://%s", S3_URI_PREFIX, tracerWorkingDir);
+        Path s3aPath = new Path(s3aPathPrefix);
+        s3AFileSystem = s3aPath.getFileSystem(conf);
+
+        useLakeFSFileSystemRes = conf.getBoolean(USE_LAKEFS_RESPONSE, false);
+        LOG.trace("Initialized FileSystemTracer, fs.lakefs.tracer.use.lakefs: {}", useLakeFSFileSystemRes);
```
nopcoder, line 108: LOG log everything related to FileSystemTracer so no need to specify it also in the log messages

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
```diff
@@ -0,0 +1,353 @@
...
+        return s3aRes;
+    }
+
+    @Override
+    public Path makeQualified(Path path) {
+        LOG.trace("makeQualified");
+
+        Path lakefsRes = lfsFileSystem.makeQualified(path);
+        Path s3aRes = s3AFileSystem.makeQualified(translateLakeFSPathToS3APath(path));
+        LOG.trace("{}[makeQualified] lakefs: {}, s3a: {}",RESULTS_COMPARISON, lakefsRes, s3aRes);
```
nopcoder, line 131:
```suggestion
        LOG.trace("{}[makeQualified] lakefs: {}, s3a: {}", RESULTS_COMPARISON, lakefsRes, s3aRes);
```

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
```diff
@@ -0,0 +1,353 @@
...
+        LOG.trace("append(f {}, bufferSize {}, progress {})", f, bufferSize, progress);
+
+        FSDataOutputStream lakefsRes = lfsFileSystem.append(f, bufferSize, progress);
+        FSDataOutputStream s3aRes = s3AFileSystem.append(translateLakeFSPathToS3APath(f), bufferSize, progress);
+        LOG.trace("{}[append] lakefs: {}, s3a: {}", RESULTS_COMPARISON, lakefsRes, s3aRes);
+
+        if (useLakeFSFileSystemRes) {
+            return lakefsRes;
+        }
+        return s3aRes;//unused
```
nopcoder, line 174: depends on the configuration - or old comment

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
```diff
@@ -0,0 +1,353 @@
...
+        for (FileStatus stat : s3aRes) {
+            Path s3aPath = stat.getPath();
+            Path lfsPath = translateS3APathToLakeFSPath(s3aPath);
+            stat.setPath(lfsPath);
+        }
+        return s3aRes;
+    }
+
+    /**
+     * Set the current working directory for the given file system. All relative
```
nopcoder, line 226: note - think we can skip the code doc for the known method as we just trace them and each underlying file system will document the right behaviour better.

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
```diff
@@ -0,0 +1,353 @@
...
+     * @param new_dir
+     */
+    @Override
+    public void setWorkingDirectory(Path new_dir) {
+        LOG.trace("setWorkingDirectory(new_dir {})", new_dir);
+
+        if (useLakeFSFileSystemRes) {
+            lfsFileSystem.setWorkingDirectory(new_dir);
+        }
+        s3AFileSystem.setWorkingDirectory(translateLakeFSPathToS3APath(new_dir));
```
nopcoder, line 238: don't need this check here - this is not a return value

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
```diff
@@ -0,0 +1,353 @@
...
+        }
+        return s3aRes;
+    }
+
+    @Override
+    public FileStatus getFileStatus(Path f) throws IOException {
+        LOG.trace("getFileStatus(f {})", f);
+
+        FileStatus lakefsRes = new FileStatus();
+        FileStatus s3aRes = new FileStatus();
```
nopcoder, line 274: No need to allocate default values.
We need to capture the value returned and the exception.
if we have an exception we don't need to address the value, so we don't need the extra boolean too.

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
```diff
@@ -0,0 +1,353 @@
...
+            LOG.error("[getFileStatus] Can't get {} file status with S3AFileSystem", s3aPath);
+            fileNotFoundOnLakeS3A = true;
+        }
+
+        LOG.trace("{}[getFileStatus] lakefs: {}, s3a: {}", RESULTS_COMPARISON, lakefsRes, s3aRes);
+
+        if (useLakeFSFileSystemRes) {
+            if (fileNotFoundOnLakeFS) {
+                throw new FileNotFoundException("getFileStatus, not found on lakeFS");
+            }
```
nopcoder, line 296: we can capture the exception and throw it here

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
```diff
@@ -0,0 +1,353 @@
...
+            fileNotFoundOnLakeFS = true;
+        }
+        try {
+            s3aRes = s3AFileSystem.getFileStatus(s3aPath);
+        } catch (Exception e) {
+            LOG.error("[getFileStatus] Can't get {} file status with S3AFileSystem", s3aPath);
+            fileNotFoundOnLakeS3A = true;
+        }
+
+        LOG.trace("{}[getFileStatus] lakefs: {}, s3a: {}", RESULTS_COMPARISON, lakefsRes, s3aRes);
```
nopcoder, line 291: need to trace the exception information too

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
```diff
@@ -0,0 +1,353 @@
...
+
+        private FSDataOutputStream lakeFSStream;
+        private FSDataOutputStream s3aStream;
+
+        public TracerOutputTStream(Path f, FsPermission permission, boolean overwrite, int bufferSize,
+                                   short replication, long blockSize, Progressable progress) throws IOException {
+            lakeFSStream = lfsFileSystem.create(f, permission, overwrite, bufferSize, replication, blockSize, progress);
+            s3aStream = s3AFileSystem.create(translateLakeFSPathToS3APath(f), permission, overwrite, bufferSize,
+                    replication, blockSize, progress);
+        }
```
nopcoder, line 321: The call to `create` should be part of the `create` method.
The stream needs to capture the streams and reflect the operations done on the streams and not how they were created.

nopcoder (changes requested):

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
talSofer: yeah, changed to "scheme". thanks!

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
talSofer: old comment, removed it

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
talSofer: agreed, I removed the docs from the other methods and missed this one

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
talSofer: let me know if what I have now is good. I logged the message of the exception above while catching it.

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
```diff
@@ -0,0 +1,353 @@
...
+ * The FileSystemTracer holds instances of {@link LakeFSFileSystem} and {@link S3AFileSystem}. On a file system operation,
+ * the FileSystemTracer invokes the operation on both file systems, logs the output of both calls, and returns the result
+ * of one of the file systems according to a configuration (S3AFileSystem output by default).
+
+ * Additional configurations:
+ * Mandatory -
+ * - fs.lakefs.tracer.working.dir - the s3 location in which the tracer can operate. This should be an S3 bucket name or
+ *   an absolute path of a directory on a bucket.
+ * Optional -
+ * - fs.lakefs.tracer.use.lakefs - tells the tracer whether it should return the response coming from the lakefs file
```
johnnyaug, line 36:
```suggestion
 * - fs.lakefs.tracer.use_lakefs_output - tells the tracer whether it should return the response coming from the lakefs file
```

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
```diff
@@ -0,0 +1,353 @@
...
+        }
+        Path lfsPath = translateS3APathToLakeFSPath(s3aPath);
+        s3aRes.setPath(lfsPath);
+        return s3aRes;
+    }
+
+    /**
+     * An output stream encapsulating two output streams, one for each file system the tracer uses.
+     */
+    private class TracerOutputTStream extends OutputStream {
```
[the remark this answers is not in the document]
talSofer: Cool! I was unaware of it :) 
I looked at using it but decided not to because it was easy to implement and using the commons bundle may cause shading problems.

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
```diff
@@ -0,0 +1,489 @@
...
+        LOG.error("Invalid path {}", path);
+        return null;
+    }
+
+    @Override
+    public void initialize(URI name, Configuration conf) throws IOException {
+        lfsFileSystem = new LakeFSFileSystem();
+        lfsFileSystem.initialize(name, conf);
+        ObjectLocation loc = lfsFileSystem.pathToObjectLocation(new Path(name));
+        lfsPathPrefix = String.format("%s://%s/%s", loc.getScheme(), loc.getRepository(), loc.getRef());
```
nopcoder, line 100: can use ObjectLocation format

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
```diff
@@ -0,0 +1,489 @@
...
+        if (useLakeFSFileSystemResults && lakeFSException != null) {
+            LOG.trace("[create] exception by lakeFSFileSystem");
+            s3aStream.close();
+            throw lakeFSException;
+        }
+        if (!useLakeFSFileSystemResults && s3aException != null) {
+            LOG.trace("[create] exception by S3AFileSystem");
+            lakeFSStream.close();
+            throw s3aException;
+        }
```
nopcoder, line 212: my concern here is when we throw and choose to close - it doesn't mean that we managed to open the other stream and we probably get null pointer exception when we try to close it.

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
```diff
@@ -0,0 +1,489 @@
...
+            lakefsRes = lfsFileSystem.rename(src, dst);
+        } catch (IOException e) {
+            lakeFSException = e;
+            LOG.error("[rename] Can't rename {} to {} with lakeFSFileSystem, exception {}", src, dst, e.getMessage());
+        }
+        try {
+            s3aRes = s3AFileSystem.rename(translateLakeFSPathToS3APath(src), translateLakeFSPathToS3APath(dst));
+        } catch (IOException e) {
+            s3aException = e;
+            LOG.error("[rename] Can't  rename {} to {} with S3AFileSystem, exception {}", src, dst, e.getMessage());
```
nopcoder, line 250: two spaces between can't and rename

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
```diff
@@ -0,0 +1,489 @@
...
+        for (FileStatus stat : s3aRes) {
+            Path filePath = stat.getPath();
+            Path lfsPath = translateS3APathToLakeFSPath(filePath);
+            stat.setPath(lfsPath);
+        }
+        return s3aRes;
+    }
+
+    @Override
+    public void setWorkingDirectory(Path new_dir) {
```
nopcoder, line 349: naming: new_dir should be in camel case

clients/hadoopfs/src/main/java/io/lakefs/FileSystemTracer.java
```diff
@@ -0,0 +1,489 @@
...
+
+        public TracerOutputTStream(FSDataOutputStream lakeFSStream, FSDataOutputStream s3aStream) throws IOException {
+            this.lakeFSStream = lakeFSStream;
+            this.s3aStream = s3aStream;
+        }
+
+        @Override
+        public void write(int b) throws IOException {
+            lakeFSStream.write(b);
+            s3aStream.write(b);
```
nopcoder, line 462: For both streams need to check for null before write - same for other operations.
Stream null is a valid option.

nopcoder (changes requested):

nopcoder (approved):

PR #2311, opened 2021-07-30
Fix cryptolake workflow input name

adunuthulan:
`lakefs-version` was updated to `lakefs_version` in the Terraform repository


itaiad200 (approved):

itaiad200:
Approved but I'm kind of worried of the use of the `cryptolake` branch for PRs. It's better to create a designated branch for every PR, especially when pushing to this branch might trigger a deployment.

adunuthulan:
> Approved but I'm kind of worried of the use of the `cryptolake` branch for PRs. It's better to create a designated branch for every PR, especially when pushing to this branch might trigger a deployment.

I'll make new branches from now on!

Issue #1908, opened 2021-05-06
Blank page when rapidly switching between tabs


johnnyaug:
Steps to reproduce:
1. Go to a repository in the UI.
2. Click the "Changes" tab, and immediately click the "Commits" tab.
3. Get a blank page with a strange URL.

![image](https://user-images.githubusercontent.com/3149885/117325789-b8e24d00-ae99-11eb-9cc2-22283fbdd351.png)

PR #2317, opened 2021-08-03
bugfix/blank screen in ui

johnnyaug:
Solves #1908.

This fixes a bug in the repository page, in which rapidly switching between tabs would cause a blank screen. 
Cause: repoId variable wasn't populated when the page was in loading state.
Solution: take repoId from router params.


nopcoder (approved):
Nice!

Issue #2326, opened 2021-08-04
Enable using underlying S3 buckets in multiple regions 


talSofer:
Use the storage namespace -> region mapping to invoke the storage operation on the S3 client tied to the storage namespace's region. 

Connect the functionality implemented in https://github.com/treeverse/lakeFS/issues/2325 with https://github.com/treeverse/lakeFS/issues/2324.

#### Acceptance criteria
- S3 Adapter can handle repositories on any AWS region.
- The region associated with a repository is automatically detected and does not depend on user input. 
- All the storage actions on any region should work.
- The MinIO integration should continue working.

Issue #2325, opened 2021-08-04
Region detection by storage namespace


talSofer:
Given a storage namespace, the S3 storage adapter should be able to detect the region hosting the S3 bucket that serves as the storage namespace. This can be done using the `GetBucketLocation in service/s3/api.go`. 

#### Acceptance criteria
- The storage namespace -> region/s3Client mapping should be cached. The cache should be invalidated upon repository deletion. 
- The MinIO integration should continue working.

Issue #2245, opened 2021-07-15
Support using underlying S3 buckets in multiple regions


ozkatz:
Currently, the S3 block adapter is tied to a specific S3 region.
In some cases, it is desirable for a single lakeFS installation to host repositories backed by buckets in different regions.

This might be possible without adding additional configuration by probing S3 to get a bucket's region (either by directly calling some API endpoint, or by attempting to use a default region and capturing the correct region from the resulting error message if any).

johnnyaug:
@talSofer, a thought about multi-region lakeFS: unless you're using the lakeFS HadoopFS for all heavy operations, it's probably not the best practice - cost and performance wise - to deploy lakeFS on a different region than its buckets. Our documentation should note the implication of doing that.

talSofer:
The tasks that implement this feature: 
- https://github.com/treeverse/lakeFS/issues/2324
- https://github.com/treeverse/lakeFS/issues/2325
- https://github.com/treeverse/lakeFS/issues/2326

FYI @johnnyaug

PR #2327, opened 2021-08-04
s3 adapter multi region

johnnyaug:
Solves #2245, through #2325 + #2326.

The new solution (lazily) creates an S3 client for every AWS region.
The s3 block adapter will use the suitable S3 client according to the region of the underlying bucket.

When the underlying storage is an S3-compatible service, with no region support (e.g. MinIO), a default S3 client is created and no change in behavior is expected


.github/workflows/nessie.yaml
```diff
@@ -129,18 +129,18 @@ jobs:
...
         with:
-          registry: docker.pkg.github.com
+          registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Pull last built docker image cache
         run: |
-          docker pull docker.pkg.github.com/${GITHUB_REPOSITORY,,}/build-cache:${{ steps.version.outputs.tag }} || \
-          docker pull docker.pkg.github.com/${GITHUB_REPOSITORY,,}/build-cache || \
+          docker pull ghcr.io/${GITHUB_REPOSITORY,,}/build-cache:${{ steps.version.outputs.tag }} || \
```
johnnyaug, line 137: reviewers - please ignore the ghcr.io change, it will be introduced in another PR.

pkg/block/s3/inventory.go
```diff
@@ -23,35 +22,35 @@ type Manifest struct {
...
-type inventoryFile struct {
+type InventoryFile struct {
 	Key      string `json:"key"` // an s3 key for an inventory list file
 	firstKey string
 	lastKey  string
 }
 
 func (a *Adapter) GenerateInventory(ctx context.Context, logger logging.Logger, manifestURL string, shouldSort bool, prefixes []string) (block.Inventory, error) {
-	return GenerateInventory(logger, manifestURL, a.s3, s3inventory.NewReader(ctx, a.s3, logger), shouldSort, prefixes)
+	m, err := a.loadManifest(ctx, manifestURL)
```
talSofer, line 38: As discussed, we shall open an issue to adding import integration tests.

pkg/block/s3/adapter.go
```diff
@@ -109,6 +116,29 @@ func NewAdapter(s3 s3iface.S3API, opts ...func(a *Adapter)) *Adapter {
...
+	region, err := s3manager.GetBucketRegion(ctx, a.awsSession, bucket, "")
+	if err != nil {
+		a.log(ctx).WithError(err).Error("failed to get region for bucket, falling back to default region")
+		region = *a.awsSession.Config.Region
+	}
+	a.regionByBucket[bucket] = region
+	return region
+}
+
+func (a *Adapter) s3Client(ctx context.Context, bucket string) s3iface.S3API {
```
talSofer, line 133: are you planning to add unit tests for this piece? I mean to test that clients are created lazily.

pkg/block/s3/adapter.go
```diff
@@ -62,7 +65,9 @@ func resolveNamespacePrefix(opts block.WalkOpts) (block.QualifiedPrefix, error)
 }
 
 type Adapter struct {
-	s3                    s3iface.S3API
+	awsSession            *session.Session
+	s3ClientByRegion      map[string]s3iface.S3API
+	regionByBucket        map[string]string
```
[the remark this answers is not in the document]
johnnyaug: These are bucket names, not storage namespaces. The confusion stems from bad naming we have on the `QualifiedKey` struct. In the context of this struct, "StorageNamespace" is actually the bucket name.

pkg/block/s3/adapter.go
```diff
@@ -62,7 +65,9 @@ func resolveNamespacePrefix(opts block.WalkOpts) (block.QualifiedPrefix, error)
 }
 
 type Adapter struct {
-	s3                    s3iface.S3API
+	awsSession            *session.Session
+	s3ClientByRegion      map[string]s3iface.S3API
```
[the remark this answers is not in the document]
johnnyaug: Done

pkg/block/s3/inventory.go
johnnyaug: Opened #2332

pkg/block/s3/adapter.go
johnnyaug: On it

pkg/block/s3/adapter.go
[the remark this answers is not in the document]
talSofer: Can you rename the field in QualifiedKey to bucket name or open an issue to keep track of that?

pkg/block/s3/adapter.go
```diff
@@ -109,6 +116,29 @@ func NewAdapter(s3 s3iface.S3API, opts ...func(a *Adapter)) *Adapter {
 	return a
 }
 
+func (a *Adapter) getBucketRegion(ctx context.Context, bucket string) string {
```
itaiad200, line 119: This isn't thread safe, is it possible to reach here from different concurrent paths on startup? Also, the locking will also save you from the need to call s3 for the same request over and over again.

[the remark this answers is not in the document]
itaiad200: And then what happens if there are two storage namespaces within the same bucket? 🤯

pkg/block/s3/adapter.go
```diff
@@ -109,6 +116,29 @@ func NewAdapter(s3 s3iface.S3API, opts ...func(a *Adapter)) *Adapter {
...
+		a.log(ctx).WithError(err).Error("failed to get region for bucket, falling back to default region")
+		region = *a.awsSession.Config.Region
+	}
+	a.bucketToRegion[bucket] = region
+	return region
+}
+
+func (a *Adapter) s3Client(ctx context.Context, bucket string) s3iface.S3API {
+	region := a.getBucketRegion(ctx, bucket)
+	if _, hasClient := a.regionToS3Client[region]; !hasClient {
```
itaiad200, line 135: Thread safety?

pkg/block/s3/inventory.go
```diff
@@ -23,35 +22,35 @@ type Manifest struct {
...
+	if err != nil {
+		return nil, err
+	}
+	svc := a.s3Client(ctx, m.inventoryBucket)
+	return GenerateInventory(logger, m, s3inventory.NewReader(ctx, svc, logger), shouldSort, prefixes)
 }
 
-func GenerateInventory(logger logging.Logger, manifestURL string, s3 s3iface.S3API, inventoryReader s3inventory.IReader, shouldSort bool, prefixes []string) (block.Inventory, error) {
+func GenerateInventory(logger logging.Logger, m *Manifest, inventoryReader s3inventory.IReader, shouldSort bool, prefixes []string) (block.Inventory, error) {
```
itaiad200, line 46: Should still be exported?

pkg/block/s3/adapter.go
```diff
@@ -126,7 +156,7 @@ func (a *Adapter) Put(ctx context.Context, obj block.ObjectPointer, sizeBytes in
 		Key:          aws.String(qualifiedKey.Key),
 		StorageClass: opts.StorageClass,
 	}
-	sdkRequest, _ := a.s3.PutObjectRequest(&putObject)
+	sdkRequest, _ := a.s3Client(ctx, qualifiedKey.StorageNamespace).PutObjectRequest(&putObject)
```
itaiad200, line 159: `qualifiedKey.StorageNamespace` is sometimes different than the bucket.

itaiad200 (changes requested):
Overall looks good. I think there are some issues related to thread safety.

Also - how will this change the collected metrics? I think we should at least add another key for the number of regions per installation.

itaiad200 (changes requested):
Overall looks good. I think there are some issues related to thread safety.

Also - how will this change the collected metrics? I think we should at least add another key for the number of regions per installation.

pkg/block/s3/adapter.go
johnnyaug: Is it? It used as the bucket for the request 4 lines above

pkg/block/s3/inventory.go
johnnyaug: Yes, used in the test

pkg/block/s3/adapter.go
johnnyaug: Changed to sync.Map. Thanks!

pkg/block/s3/inventory.go
talSofer: Thanks!

talSofer (approved):
Looks good! 
I think it's worth mentioning in the issue description what's the expected behavior for s3-compatible objects storage (e.g., Minio)

PR #2331, opened 2021-08-08
github - use container registry

johnnyaug:
Migrate to using github container registry, which replaced docker registry.
Lately there have been problems with the Nessie build related to the Docker registry. Let's see if the migration solves them.

See: https://docs.github.com/en/packages/working-with-a-github-packages-registry/migrating-to-the-container-registry-from-the-docker-registry

### DIFF
--- a/.github/workflows/nessie.yaml
+++ b/.github/workflows/nessie.yaml
@@ -129,18 +129,18 @@ jobs:
       - name: Login to GitHub Docker Registry
         uses: docker/login-action@v1
         with:
-          registry: docker.pkg.github.com
+          registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Pull last built docker image cache
         run: |
-          docker pull docker.pkg.github.com/${GITHUB_REPOSITORY,,}/build-cache:${{ steps.version.outputs.tag }} || \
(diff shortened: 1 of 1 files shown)


.github/workflows/nessie.yaml
```diff
@@ -129,18 +129,18 @@ jobs:
...
         with:
-          registry: docker.pkg.github.com
+          registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
       - name: Pull last built docker image cache
         run: |
-          docker pull docker.pkg.github.com/${GITHUB_REPOSITORY,,}/build-cache:${{ steps.version.outputs.tag }} || \
-          docker pull docker.pkg.github.com/${GITHUB_REPOSITORY,,}/build-cache || \
+          docker pull ghcr.io/${GITHUB_REPOSITORY,,}/build-cache:${{ steps.version.outputs.tag }} || \
```
nopcoder, line 137: everything looks good. just question regarding the ${GITHUB_REPOSITORY,,} - what does ,, is for?

nopcoder (approved):

.github/workflows/nessie.yaml
johnnyaug: Apparently it converts to lower case (thanks @arielshaqed!)

PR #2333, opened 2021-08-08
Read action's secrets from env vars and avoid logging them

itaiad200:
This implementation assumes that secrets are to be used only in specific locations and are not available in all places. For example, query-params & headers are good place for secrets, but the url itself isn't.

While this might limit the users a bit when trying to pass variables to the execution itself, it will allow us to log all other parameters of the actions, without worrying about leakage of secrets. Env var references in other places in the action yaml file are simply ignored.


.github/workflows/nessie.yaml
```diff
@@ -375,6 +375,7 @@ jobs:
       TAG: ${{ needs.deploy-image.outputs.tag }}
       # Setting Account_ID as a secret as a way to avoid specifying it here
       REPO: ${{ secrets.AWS_ACCOUNT_ID }}.dkr.ecr.us-east-1.amazonaws.com
+      ACTIONS_VAR: "this_is_actions_var"
```
johnnyaug, line 378: If possible, inject this env var from the test code and not from here - otherwise when running locally, one has to make sure the var is set in order for the test to work.

pkg/actions/secure_string.go
```diff
@@ -0,0 +1,39 @@
...
+	if s.secret {
+		return "*******"
+	}
+	return s.val
+}
+
+var envVarRegex = regexp.MustCompile(`{{ ENV\.(.*) }}`)
+
+// Creates a new SecureString, reading env var if needed.
+func NewFromString(s string) (SecureString, error) {
```
johnnyaug, line 27:
```suggestion
func NewSecureString(s string) (SecureString, error) {
```

pkg/actions/secure_string.go
```diff
@@ -0,0 +1,39 @@
...
+	matches := envVarRegex.FindStringSubmatch(s)
+	if len(matches) == 0 {
+		return SecureString{val: s, secret: false}, nil
+	}
+	envVarName := s[len("{{ ENV.") : len(s)-len(" }}")]
```
johnnyaug, line 32: This looks a bit too strict: `{{ENV.MY_VAR}}` and others should also work, IMO

pkg/actions/testdata/action_secrets.yaml
```diff
@@ -0,0 +1,24 @@
...
+    properties:
+      url: "https://api.lakefs.io/webhook1?t=1za2PbkZK1bd4prMuTDr6BeEQwWYcX2R"
+      query_params:
+        disallow: ["user_", "{{ ENV.BLA_BLA }}" ]
+        prefix: "{{ ENV.BLA }}"
+  - id: no_freeze
+    type: webhook
+    description: check production is not in dev freeze
+    properties:
+      url: "https://api.lakefs.io/webhook2?t=1za2PbkZK1bd4prMuTDr6BeEQwWYcX2R"
```
johnnyaug, line 24: missing newline

.github/workflows/nessie.yaml
itaiad200: lakeFS is running separately from the test code. I can't control it from the test code, I can set it always in lakeFS but that seems like too much. WDYT?

pkg/actions/secure_string.go
```diff
@@ -0,0 +1,39 @@
+package actions
+
+import (
+	"os"
+	"regexp"
+)
+
+// SecureString is a string tha may or may not contain a secret.
+// Secret is considered any value that was read from an environment variable, in the following syntax:
+// "{{ ENV.VAR_KEY }}"
```
johnnyaug, line 10:
```suggestion
// SecureString is a string that may be populated from an environment variable.
// If constructed with a string of the form {{ ENV.EXAMPLE_VARIABLE }}, the value is populated from EXAMPLE_VARIABLE and is considered a secret. Otherwise the value is taken from the string as-is, and is not considered a secret.
```

johnnyaug (changes requested):
Looking good.
Missing: documentation, CHANGELOG entry.

pkg/actions/secure_string.go
```diff
@@ -21,19 +24,37 @@ func (s SecureString) String() string {
...
+	var err error
+	ret := envVarRegex.ReplaceAllStringFunc(s, func(origin string) string {
+		if err != nil {
+			return ""
+		}
+		matches++
+		raw := strings.Trim(origin, "{} ")
+		parts := strings.SplitN(raw, ".", 2)
+		if len(parts) != 2 || parts[0] != "ENV" {
+			return s
```
johnnyaug, line 41: I think you should return origin here? Add a test case which covers this if.

PR #2340, opened 2021-08-09
fix dbTx time counting error

holajiawei:


nopcoder:
@holajiawei good catch! thanks for the contribution.
I'll associate an issue with the PR before addressing it.
It will be faster to look up this fix later.

nopcoder (approved):
LGTM

arielshaqed (commented):
Good catch, thanks!

PR #2360, opened 2021-08-11
Add api endpoint to retrieve user-defined metadata

adunuthulan:
Adds the endpoint `/repositories/{repository}/refs/{ref}/objects/metadata` to retrieve user-defined metadata

### DIFF
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -216,13 +216,13 @@ components:
           type: string
         checksum:
           type: string
+        size_bytes:
+          type: integer
+          format: int64
         mtime:
           type: integer
           format: int64
           description: Unix Epoch in seconds
-        size_bytes:
-          type: integer
-          format: int64
 
     ObjectStatsList:
       type: object
@@ -256,9 +256,12 @@ components:
           format: int64
           description: Unix Epoch in seconds
         metadata:
-          type: object
-          additionalProperties:
-            type: string
+          $ref: "#/components/schemas/ObjectUserMetadata"
+
+    ObjectUserMetadata:
+      type: object
+      additionalProperties:
+        type: string
 
     UnderlyingObjectProperties:
       type: object
@@ -2439,7 +2442,7 @@ paths:
       tags:
         - objects
       operationId: stageObject
-      summary: stage an object"s metadata for the given branch
+      summary: stage an object's metadata for the given branch
       requestBody:
         required: true
         content:
@@ -2553,6 +2556,45 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/ObjectStats"
+        401:
+          $ref: "#/components/responses/Unauthorized"
+        404:
+          $ref: "#/components/responses/NotFound"
+        default:
+          $ref: "#/components/responses/ServerError"
+        410:
+          description: object gone (but partial metadata may be available)
+
+  /repositories/{repository}/refs/{ref}/objects/metadata:
+    parameters:
+      - in: path
+        name: repository
+        required: true
+        schema:
+          type: string
+      - in: path
+        name: ref
+        required: true
+        schema:
+          type: string
+        description: a reference (could be either a branch or a commit ID)
+      - in: query
+        name: path
+        required: true
+        schema:
+          type: string
+    get:
+      tags:
+        - objects
+      operationId: getObjectUserMetadata
+      summary: get user-defined object metadata
+      responses:
+        200:
+          description: user-defined object metadata
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ObjectUserMetadata"
         401:
           $ref: "#/components/responses/Unauthorized"
         404:
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -2508,6 +2508,46 @@ func (c *Controller) StatObject(w http.ResponseWriter, r *http.Request, reposito
 	writeResponse(w, code, objStat)
 }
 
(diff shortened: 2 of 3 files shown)


api/swagger.yml
```diff
@@ -2562,6 +2565,45 @@ paths:
         410:
           description: object gone (but partial metadata may be available)
 
+  /repositories/{repository}/refs/{ref}/objects/metadata:
```
itaiad200, line 2568: So we have:
1. `.../objects/stat` - that returns lakeFS metadata without the user metadata.
2. `.../objects/underlying_properties` - that returns the metadata on the physical object.
3. `.../objects/metadata` - that returns custom user defined metadata.

@talSofer @arielshaqed 
Do we really need all 3? Isn't it easier to consolidate some of these?

pkg/api/controller.go
```diff
@@ -2508,6 +2508,46 @@ func (c *Controller) StatObject(w http.ResponseWriter, r *http.Request, reposito
...
+	if !c.authorize(w, r, []permissions.Permission{
+		{
+			Action:   permissions.ReadObjectAction,
+			Resource: permissions.ObjectArn(repository, params.Path),
+		},
+	}) {
+		return
+	}
+	ctx := r.Context()
+	c.LogAction(ctx, "object_user_metadata")
```
talSofer, line 2521:
```suggestion
	c.LogAction(ctx, "get_object_user_metadata")
```

api/swagger.yml
talSofer: @itaiad200 good point, this endpoint is unused `.../objects/underlying_properties.` 

How about only keeping `.../objects/stat` endpoint and add parameters to specify if the response should contain **underlying properties** and **user metadata**? the default value for these params should be false, and they will be optional. 

@itaiad200 @arielshaqed @adunuthulan WDYT?

talSofer (commented):
Thanks for working on this @adunuthulan !
Two general comments:
- Please edit the PR description to include the motivation for your design. Why you decided to add another endpoint and the problem that you were trying to overcome by doing that. (you explained it when we spoke :))
- Please add the issue number this PR closes so that the issue I closed automatically.

api/swagger.yml
adunuthulan: I've removed the third endpoint and made the default `/repositories/{repository}/refs/{ref}/objects/stat` endpoint and `/repositories/{repository}/refs/{ref}/objects/ls` endpoint return the metadata. Each has a query param that allows the user to not retrieve the user defined metadata

adunuthulan:
Hey all! Would anyone be able to help me understand why the `make checks-validator` is failing? Should I be committing the changes I have made to the python and java clients?

api/swagger.yml
```diff
@@ -2541,6 +2546,12 @@ paths:
         required: true
         schema:
           type: string
+      - in: query
+        name: user_metadata
+        required: false
+        schema: 
+          type: boolean
+          default: false
```
nopcoder, line 2554: The default here is false, the next API (ls) the default is set to true - is it on purpose?

clients/.project
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
nopcoder, line 1: looks like this file was added in the wrong place - opened by eclipse or moved from the generated files.

pkg/api/controller.go
```diff
@@ -2441,14 +2441,18 @@ func (c *Controller) ListObjects(w http.ResponseWriter, r *http.Request, reposit
...
+			objStat := ObjectStats{
 				Checksum:        entry.Checksum,
 				Mtime:           mtime,
 				Path:            entry.Path,
 				PhysicalAddress: qk.Format(),
 				PathType:        entryTypeObject,
 				SizeBytes:       Int64Ptr(entry.Size),
-			})
+			}
+			if (params.UserMetadata == nil || *params.UserMetadata) && entry.Metadata != nil {
```
nopcoder, line 2452: make sure that for 'ls' if the default is passed by the generated code - as far as I remember the oapi middleware just verify that the request and response are valid.
The default behaviour when there is no value is handled by the code - which means that if 'params.UserMetadata' is nil it is also true.

nopcoder (changes requested):

talSofer (commented):
@adunuthulan the new approach looks good. Please address Barak's comments and we will approve the pr. Thanks!

itaiad200:
I like the idea of using the query-param!
Please add/change existing tests for the new functionality.

pkg/api/controller.go
adunuthulan: OAPI doesn't handle the default values, you're right, so setting the defaults here doesn't actually do anything, which I find strange. 

While I was testing the call, if the `user_metadata` query param wasn't set in the request, this function would dereference the nil pointer, and the code would throw a runtime error, so I believe this check is needed. With this check, the default behavior I described (returning the user-defined metadata) occurs. 

Since OAPI defaults don't actually set a default, but the behavior is coded here instead, would it be okay to leave the defaults in the Swagger file to act as documentation for these api calls?

api/swagger.yml
adunuthulan: This should be set to true, thanks for the catch!

clients/.project
adunuthulan: Deleted

itaiad200 (approved):

pkg/api/controller.go
nopcoder: Got it. Leave the default in the swagger as it can affect the client-side code generated.

nopcoder (approved):

arielshaqed (approved):
Thanks! :-)

PR #2367, opened 2021-08-15
docs: toc style update

JamilAhmad:
Table content Style changes

### DIFF
--- a/docs/_sass/custom/custom.scss
+++ b/docs/_sass/custom/custom.scss
@@ -109,11 +109,12 @@ body {
 		font-weight: 600;
 		font-family: GalanoGrotesque-Bold;
 	}
-	h4,
-	.text-delta {
-		text-transform: none;
-		width: 100%;
-	    max-width: 264px;
+	.text-delta{
+		display: none;
+	}
+	h4{
+	    text-transform: none;
+	    width: 100%;
 	    display: block;
 	    background-color: $teal-green;
 	    color: $white;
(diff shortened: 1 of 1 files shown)


docs/_sass/custom/custom.scss
```diff
@@ -183,7 +216,7 @@ body {
 
 	pre code {
 		background-color: $darkblack;
-    	border: 0px solid $lightgreen;
+    		border: 0px solid $lightgreen;
```
nopcoder, line 219: indent

docs/_sass/custom/custom.scss
```diff
@@ -109,11 +109,12 @@ body {
...
 		font-family: GalanoGrotesque-Bold;
 	}
-	h4,
-	.text-delta {
-		text-transform: none;
-		width: 100%;
-	    max-width: 264px;
+	.text-delta{
+		display: none;
+	}
```
nopcoder, line 114: You hide the real table of content and inject your own using 'before'?
The CSS should content the style and not the content.

docs/_sass/custom/custom.scss
```diff
@@ -887,3 +920,15 @@ div.highlighter-rouge{
...
 }
+@media (max-width:767px){
+	.text-delta + ul, #markdown-toc{
+		li, li a{
+			margin-bottom: 12px;
+		    font-size: 12px !important;
+		    display: block;
+		}
+		li a span{
+			font-size: 8px!important;
```
nopcoder, line 931: do we need to use '!important' for all the places we change the font-size?

nopcoder (changes requested):

johnnyaug:
@JamilAhmad, thank you for this contirubtion! Could you please add to the PR description the reason for this change and what it does?

docs/_sass/custom/custom.scss
JamilAhmad: Yes I hide it because we can't make table auto width based on content because the DOM structure for the table is not correct(table header and table content list are not in any container) so that's why I used before  for table list so it will auto width based on table content, hope that make sense

docs/_sass/custom/custom.scss
JamilAhmad: Yes I only used important where needed because override things not working here so that's why important needed

docs/_sass/custom/custom.scss
JamilAhmad: hmm, can you do that at your end? just remove extra space etc

nopcoder: no, you will need to push the changes to this pull request.
also, this like is just an example there are a couple of places here where you used tab vs spaces for example. Please make sure you have all the styles aligned.

docs/_sass/custom/custom.scss
nopcoder: The page layout is - https://github.com/treeverse/lakeFS/blob/master/docs/_layouts/default.html
You can change the structure of the toc if needed.

nopcoder (changes requested):

nopcoder:
Fixed by https://github.com/treeverse/lakeFS/pull/2385

PR #2368, opened 2021-08-15
move blockstore type consts to block package

johnnyaug:
These constants caused some packages to point at the concrete blockstore implementations (s3, gs, etc).
For example, the `stats` package pointed at all of the blockstore implementation packages.
This caused a dependency cycle when trying to use the `stats` package from the S3 blockstore.


itaiad200 (approved):

nopcoder (approved):
cool, next we can also register our adapters and turn the BlockType into a type.

Issue #2181, opened 2021-06-29
Add "protected" property to branches


ozkatz:
Allow some branches to be configured as "protected". When configured, these branches cannot be directly written to (i.e. technically, will not have a staging area). These branches can only receive changes by being merged into.

This is useful when designing a CI/CD pipeline: assuming data consumers read from "main", we can configure pre-merge hooks to run and validate any new change being applied. This gives a strong guarantee to readers about the validity and quality of anything they see on main.

PR #2369, opened 2021-08-15
Protected Branches - Design

johnnyaug:
Design suggestions for #2181

### DIFF
--- a/design/protected-branches.md
+++ b/design/protected-branches.md
@@ -27,7 +27,7 @@ Cons:
 
 The implementation depends on the constraint type we want to enforce.
 
-### Constarint type: block commits
+### Constraint type: block commits
 
 * When performing a commit, the protection rules of the repository will be fetched from the storage.
 * Right before executing the pre-commit hooks, check whether the branch name matches any of the protection rules.
--- a/design/protected-branches.md
+++ b/design/protected-branches.md
@@ -36,7 +36,8 @@ Pros:
 Cons:
 - Changing protection rules is a commit visible in the log
 - As a repository owner, I need to manage rules in multiple branches. For example, if I want to add rules to existing branches, I need to merge these rules to all of my developer's branches.
-
+- Can currently only be used to block commits, and not the staging area.
+- 
 Pro/con:
 - Changing protection rules is a commit visible in the log
 ## Implementation - Suggestion 1
--- a/design/protected-branches.md
+++ b/design/protected-branches.md
@@ -22,6 +22,7 @@ Pros:
 - no overhead: branch is anyway fetched from the database before a commit.
 
 Cons:
+- a boolean is probably not enough for the long run: we will want other constraints, not just limiting commits.
 - yet another dependency on postgres
 - no pattern rules - every branch needs to marked as protected separately.
 
@@ -33,6 +34,6 @@ Pros:
 - rules can be patterns and not just concrete branches, similar to GitHub's protected branches.
 - native to the storage
 
-cons:
+Cons:
 - UI will probably be a multiline textbox where you can edit the json, otherwise consistency problems.
 - CLI will also need to get the full JSON, or we need to do some locking on the server-side to add/remove protection rules.
--- a/design/protected-branches.md
+++ b/design/protected-branches.md
@@ -40,7 +40,7 @@ Cons:
 Pro/con:
 - Changing protection rules is a commit visible in the log
 
-### Descision
+### Decision
 
 After a meeting discussing the matter, we decided to go with suggestion #1, but to add a repository-level settings feature to be used in order to store the branch protection rules. See the [issue](https://github.com/treeverse/lakeFS/issues/2406).
 
--- a/design/protected-branches.md
+++ b/design/protected-branches.md
@@ -1,0 +1,35 @@
+# Protected Branches
+
+## Requirements
+
+1. As a lakeFS user, I should be able to mark branches as "protected" - from the UI, CLI and API.
+1. Protected branches cannot be directly committed to. Only merges are allowed into protected branches.
+1. Commits to protected branches should fail with a meaningful error message.
+
+## Where to save the model
+
+The data regarding which branch is protected needs to be saved somewhere. It needs to be fetched at runtime, probably before a commit is performed.
+
+### Option 1: branches table in postgres
+
+Add a boolean column to the branches table, denoting whether the branch is protected.
+
+Pros:
+- no consistency problems
+- no overhead: branch is anyway fetched from the database before a commit.
+
+Cons:
+- yet another dependency on postgres
+- no pattern rules - every branch needs to marked as protected separately.
+
+Option 2: save it as a JSON array in the storage
+
+Add it as a file under the repository's _lakefs directory.
+
+Pros:
+- rules can be patterns and not just concrete branches, similar to GitHub's protected branches.
+- native to the storage
+
+cons:
+- UI will probably be a multiline textbox where you can edit the json, otherwise consistency problems.
+- CLI will also need to get the full JSON, or we need to do some locking on the server-side to add/remove protection rules.
--- a/design/protected-branches.md
+++ b/design/protected-branches.md
@@ -30,7 +30,7 @@ Cons:
 ```json
 [
 	{
-		"branch_name_pattern": "stable/*",
+		"branch_name_pattern": "main",
 		"constraints": ["merge_only"]
 	},
 	{
--- a/design/protected-branches.md
+++ b/design/protected-branches.md
@@ -11,7 +11,10 @@
 1. Force options - blocked operations should have a reasonable way to be forced. In the future, the extent to which they can be forced may also be configurable (similarly to GitHub).
 
 ## Where to save the model
-Suggestion: save protection rules as a JSON object under the repository's _lakefs_. This is similar to what we do with retention rules.
+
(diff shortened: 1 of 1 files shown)


design/protected-branches.md
```diff
@@ -0,0 +1,38 @@
+# Protected Branches
+
+## Requirements
+
+1. As a lakeFS user, I should be able to mark branches as "protected" - from the UI, CLI and API.
```
nopcoder, line 5: My thoughts about "branch protection":
Branch protection is a repository-level set of rules that we check for specific operations.
The user can define a rule that matches zero to many branches, patterns like: "dev-*" that apply a specific set of constraints.
One of the constraints is 'merge only' - which allows only merge operations on the branch.
As repository wide operations we should keep them at the same level as we keep retention (think it is option two here).

design/protected-branches.md
```diff
@@ -0,0 +1,38 @@
...
+
+Add it as a file under the repository's _lakefs directory.
+
+Pros:
+- rules can be patterns and not just concrete branches, similar to GitHub's protected branches.
+- native to the storage
+
+cons:
+- UI will probably be a multiline textbox where you can edit the json, otherwise consistency problems.
+- CLI will also need to get the full JSON, or we need to do some locking on the server-side to add/remove protection rules.
```
itaiad200, line 38: You need locking no matter what... you can't have race conditions when changing protected branches.

itaiad200 (changes requested):

design/protected-branches.md
```diff
@@ -0,0 +1,38 @@
...
+### Option 1: branches table in postgres
+
+Add a boolean column to the branches table, denoting whether the branch is protected.
+
+Pros:
+- no consistency problems
+- no overhead: branch is anyway fetched from the database before a commit.
+
+Cons:
+- yet another dependency on postgres
```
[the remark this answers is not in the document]
johnnyaug: Yeah, it's always the argument when adding things to postgres.

design/protected-branches.md
johnnyaug: This locking is for when changing protection rules. If you're on postgres, you get this locking for free.

design/protected-branches.md
```diff
@@ -0,0 +1,38 @@
+# Protected Branches
+
+## Requirements
+
+1. As a lakeFS user, I should be able to mark branches as "protected" - from the UI, CLI and API.
+1. Protected branches cannot be directly committed to. Only merges are allowed into protected branches.
```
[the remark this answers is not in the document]
johnnyaug: Very interesting idea, adding it

arielshaqed (commented):
As discussed F2F, I believe that the way to add this feature is _not_ to add more metadata but rather to use what we already have.  (So not like GitHub branch protection in this sense but rather like their (and our!) actions.)  Adding metadata means that you must take care of, among other things:

* Authorization
* Audit
* Branch matching
* Ordering with respect to actions
* Historical behaviour (e.g. committing becomes non-hermetic, the fact that you could commit at ref X yesterday does not means that you can commit there today!)

All this because it adds a new concept, and now the users and we need to understand how that concept interacts with other concepts in the system.  And users cannot modify the branch protection feature, so we will have to do it for them.  (For instance, almost every branch protection feature on GitHub today at https://github.com/treeverse/lakeFS/settings/branch_protection_rules/17304120 would require a lakeFS user to add an action; arguably already having an action makes adding these features _easier_ not harder.)  So for any of these features AFAICS we would just send them a new action to install (or they would).  And now you have *two* audit logs: one for "lakefs branch protection", one for "extra branch protections that got added as actions".  And the user needs to understand how everything interacts.

I would like to suggest a different underlying implementation, that in turn influences the behaviour with respect to most of the above.  Add a pre-commit action that prevents some commits.  E.g. it looks at its branch name, and if the branch name matches one of a bunch of regexps / wildcards / ... then it fails the commit.

* Users get authorization through existing lakefs IAM permissions to edit these files.
* Users get their audit using `lakectl log ...`.  As a bonus, the log "Add branch protection" appears in the same log, users can see when it happened.
* Branch matching is entirely flexible and written in a language the user might know (bash, Python, ...).
* Ordering with respect to actions *is* the ordering of actions.
* Behaviour is hermetic and reproducible.

This implementation can still happen from the GUI!  E.g. press a button, it adds a commit to the default branch with an activity that does the configured matching or modifies that commit.

johnnyaug:
@arielshaqed, I understand the desire to create less places where configuration is done. I also completely see the elegance of using an existing solution like Hooks.
However, my concern is that it will be harder for users to grasp and understand. I tried to sum it up to a short paragraph in the doc -  please let me know if I got the gist of it.

guy-har:
I believe that staging must be blocked, otherwise the branch isn't really protected. If we choose to block staging on protected branches using hooks might complicate things more than making them simple.

design/protected-branches.md
```diff
@@ -0,0 +1,79 @@
...
+* When performing a commit, the protection rules of the repository will be fetched from the storage.
+* Right before executing the pre-commit hooks, check whether the branch name matches any of the protection rules.
+* If so, the commit will fail with a dedicated error type.
+* A force flag will be optionally included, allowing the commit even if the branch is protected.
+
+#### Constraint type: block staging area
+
+* In Graveler, before every Staging Manager write operation (Set, Drop, DropKey) - fetch the rules.
+* If the branch name matches a rule, fail the operation with a dedicated error type.
+* Each of these operations should allow a force flag.
```
arielshaqed, line 79: This is obviously necessary, so it also has to work with all the ways to write to staging.  Otherwise for most use-cases there is no way to force writes -- almost nothing uses the lakeFS API.

1. How does the user make this work with the S3 gateway?  We should consider e.g. whacky-coding something into the branch name on the S3 path (or on the repo?), so that I can change my Spark job to force-write.
2. How does the user configure this with the Spark HadoopFS?  We should consider e.g. a similar coding on the lakefs path.

design/protected-branches.md
```diff
@@ -0,0 +1,79 @@
...
+Cons:
+- As a repository owner, I need to manage rules in multiple branches. For example, if I want to add rules to existing branches, I need to merge these rules to all of my developer's branches.
+- Can currently only be used to block commits, and not the staging area.
+ 
+Pro/con:
+- Changing protection rules is a commit visible in the log
+
+### Descision
+
+After a meeting discussing the matter, we decided to go with suggestion #1, but to add a repository-level settings feature to be used in order to store the branch protection rules. See the [issue](https://github.com/treeverse/lakeFS/issues/2406).
```
arielshaqed, line 45: How do we audit changes?  (Do we?)

design/protected-branches.md
```diff
@@ -0,0 +1,78 @@
...
+### Runtime
+
+The enforcement of the protection depends on the constraint type.
+In the future we may want to add an option to force these operations.
+
+#### Constraint type: block commits
+
+* When performing a commit, the protection rules of the repository will be fetched from the storage.
+* Right before executing the pre-commit hooks, check whether the branch name matches any of the protection rules.
+* If so, the commit will fail with a dedicated error type.
```
arielshaqed, line 73: Is this the `merge_only` constraint type?  If so, perhaps it should be stated more like
```json
{
  "...": "...",
  "commit": false
}
```

design/protected-branches.md
```diff
@@ -0,0 +1,78 @@
...
+
+## Implementation - Suggestion 1
+
+### Example Branch Protection JSON
+
+```json
+[
+	{
+		"branch_name_pattern": "main",
+		"constraints": ["merge_only"]
```
arielshaqed, line 55: The "[Runtime](https://github.com/treeverse/lakeFS/pull/2369/files#diff-629b7ff0be70bac13faacb18dfd2c7aa8b0e726e739a0f0fa8a8b96a5e432acdR64)" design below seems to talk in terms of _capabilities_, and uses different "constraints".

I suspect that there are really capabilities hiding here too, maybe this should be an object denying some of them?

design/protected-branches.md
```diff
@@ -0,0 +1,78 @@
...
+Cons:
+- As a repository owner, I need to manage rules in multiple branches. For example, if I want to add rules to existing branches, I need to merge these rules to all of my developer's branches.
+- Can currently only be used to block commits, and not the staging area.
+ 
+Pro/con:
+- Changing protection rules is a commit visible in the log
+
+### Decision
+
+After a meeting discussing the matter, we decided to go with suggestion #1, but to add a repository-level settings feature to be used in order to store the branch protection rules. See the [issue](https://github.com/treeverse/lakeFS/issues/2406).
```
arielshaqed, line 45: Given the recurrence of questions around this (most recently #2491), we need to add some more definitive explanation of this decision.

I believe that this is a dependency on #2406.

arielshaqed (changes requested):
1. AFAICT this design depends on #2406.  If we do not want to hold its approval subject to approving _that_ one, then we should advance on #2406 at least as far as giving the semantics that that design will provide (e.g. consistency, performance, audit, authorization, etc.).
2. The policy syntax seems arbitrary and inconsistent with other policy types in the system.

design/protected-branches.md
johnnyaug: As I mentioned on Slack, #2491 is a design choice for implementing #2406.

I don't think there's a dependency here on any of them

itaiad200 (approved):

PR #2370, opened 2021-08-15
Add metrics for multi region

johnnyaug:
Following #2327, adding a metric to the BI counting the AWS clients created for each region.

~In order to access the stats collector from the relevant place, I put it on the go-context passed to the block adapter (don't like it - open to suggestions).~


pkg/config/config.go
```diff
@@ -356,11 +357,7 @@ const floatSumTolerance = 1e-6
 
 // GetCommittedTierFSParams returns parameters for building a tierFS.  Caller must separately
 // build and populate Adapter.
-func (c *Config) GetCommittedTierFSParams(ctx context.Context) (*pyramidparams.ExtParams, error) {
-	adapter, err := factory.BuildBlockAdapter(ctx, c)
-	if err != nil {
-		return nil, fmt.Errorf("build block adapter: %w", err)
-	}
+func (c *Config) GetCommittedTierFSParams(ctx context.Context, adapter block.Adapter) (*pyramidparams.ExtParams, error) {
```
itaiad200, line 360: Is the context still needed?

pkg/config/config.go
```diff
@@ -10,14 +10,15 @@ import (
 	"strings"
 	"time"
 
+	"github.com/treeverse/lakefs/pkg/block"
```
itaiad200, line 13: arrange deps

itaiad200 (approved):

arielshaqed:
This will probably conflict with #2377, make sure to rebase to `master` HEAD before merging.

cmd/lakefs/cmd/run.go
```diff
@@ -113,8 +107,13 @@ var runCmd = &cobra.Command{
 				Error("Block adapter NOT SUPPORTED for production use")
 		}
 		metadata := stats.NewMetadata(ctx, logger, blockstoreType, authMetadataManager, cloudMetadataProvider)
-		bufferedCollector := stats.NewBufferedCollector(metadata.InstallationID, blockStore.RuntimeStats, cfg)
-
+		bufferedCollector := stats.NewBufferedCollector(metadata.InstallationID, cfg)
+		// init block store
+		blockStore, err := factory.BuildBlockAdapter(context.WithValue(ctx, block.ContextKeyStatsCollector, bufferedCollector), cfg)
```
nopcoder, line 112: Talked f2f over passing the stats collector as a parameter to the factory.
The factory will pass the collector using the specific adapter options.
The adapter can define the interface of how to send stats and not directly use the stats package.

pkg/block/s3/adapter.go
```diff
@@ -93,7 +94,13 @@ func WithTranslator(t block.UploadIDTranslator) func(a *Adapter) {
...
 }
 
-func NewAdapter(awsSession *session.Session, opts ...func(a *Adapter)) *Adapter {
+func WithStatsCollector(s stats.Collector) func(a *Adapter) {
+	return func(a *Adapter) {
+		a.clients = a.clients.WithStatsCollector(s)
+	}
+}
+
+func NewAdapter(_ context.Context, awsSession *session.Session, opts ...func(a *Adapter)) *Adapter {
```
nopcoder, line 103: now we can remove the ctx we added in this pr

nopcoder (changes requested):

pkg/block/s3/adapter.go
johnnyaug: oops

nopcoder (approved):

PR #2371, opened 2021-08-15
reduce db connect calls by reduce calls from migrate

nopcoder:
In order to reduce the number of times, we create connections pools on startup, connect to the database first and reuse the pool for migrating version functionality.


johnnyaug (approved):
Yay! 
Consider adding to CHANGELOG as an optimization.

PR #2372, opened 2021-08-15
lakefs config.yaml align name and search order

### LINKED ISSUE #2355: lakeFS configuration should load from a standard service location
lakeFS configuration should first load from a known location. Common service location is machine level path: /etc/lakefs.yaml
Need to support the current home dot/hidden configuration and document the change in the integration documentation.

nopcoder:
lakefs service configuration - align the filename and search locations to other services, with fallback support:

- configuration file named lakefs.yaml
- search for /etc/lakefs/config.yaml
- search for $HOME/.lakefs/config.yaml
- search for ./config.yaml
- (fallback) search for $HOME/.lakefs.yaml

Close #2355

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 
 ## Unreleased - XXXX-XX-XX
 
- - lakeFS starts configuration load non-hidden config files first (#2355)
+ - Add search locations to load lakeFS configuration. More information on https://docs.lakefs.io/reference/configuration (#2355)
 
 ## v0.48.0 - 2021-08-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 
 ## Unreleased - XXXX-XX-XX
 
+ - lakeFS starts configuration load non-hidden config files first (#2355)
 
 ## v0.48.0 - 2021-08-22
 
--- a/cmd/lakefs/cmd/root.go
+++ b/cmd/lakefs/cmd/root.go
@@ -1,8 +1,10 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"os"
+	"path"
 	"strings"
 	"sync"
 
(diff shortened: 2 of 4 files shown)


cmd/lakefs/cmd/root.go
```diff
@@ -50,42 +52,58 @@ func init() {
...
-		logger.Info("loaded configuration from file")
-	} else if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
-		logger.WithError(err).Fatal("failed to read config file")
+	logger = logger.WithField("file", viper.ConfigFileUsed()) // should be called after SetConfigFile
+	var errFileNotFound viper.ConfigFileNotFoundError
+	if err != nil && !errors.As(err, &errFileNotFound) {
+		logger.WithError(err).Fatal("Failed to find a config file")
+	}
+	// fallback - try to load the previous supported $HOME/.lakefs.yaml
+	//   if err is set it will be file-not-found based on previous check
```
arielshaqed, line 79: I don't understand this.  Didn't line 62 _already_ cause Viper to search for this file?

arielshaqed (changes requested):
Great!  I don't understand why we search explicitly for the config file in the home directory after configuring Viper to search for it anyway; see comment inline.

The above is not necessarily blocking.  Requested changes are all in documentation:
1. Please document the location of the configuration file.
1. Please add a row to `CHANGELOG.md`, this change is (very) user-visible.

cmd/lakefs/cmd/root.go
nopcoder: The package supports an additional search path - this case is to support an additional filename.
The current search lookup for `.lakefs.yaml` (starts with dot) - in order to support it, we try to find a configuration find (the new standard) first and need to try to read the specific one - unless the user specified it already.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
  - Actions secrets support with env vars (#2333)
  - Reduce the number of database connections used on startup
  - Validate required configuration keys blockstore.type, auth.encrypt.secret_key.  This *breaks* existing configurations that assume a default blockstore.type of "local".  But no such configuration may be production.
+ - lakeFS configuration load non-hidden config files first (#2355)
```
arielshaqed, line 10: Yes, but I also want *more*!  Can you give the list of locations where lakeFS searches for configuration, or alternatively link to the docs that gives the location?
```suggestion
 - Add search path to load lakeFS, see https://lakefs.io/docs/blah/blue (#2355)
```

docs/reference/configuration.md
```diff
@@ -12,7 +12,16 @@ has_children: false
 
 {% include toc.html %}
 
-Configuring lakeFS is done using a yaml configuration file.
+Configuring lakeFS is done using a yaml configuration file and/or environment variable.
+The configuration file location can be set with the '--config' flag. It will lookup, load, and use the first file found in the following order:
+1. ./config.yaml
+1. `<home directory>`/lakefs/config.yaml
+1. /etc/lakefs/config.yaml
+1. `<home directory>`/.lakefs.yaml
```
arielshaqed, line 20: This confused me because `--config` override the existing path.  Consider
```suggestion
The configuration file location can be set with the '--config' flag. If not specified, the the first file found in the following order will be used:
1. ./config.yaml
1. `$HOME`/lakefs/config.yaml
1. /etc/lakefs/config.yaml
1. `$HOME`/.lakefs.yaml
```

arielshaqed (approved):
Thanks!

PR #2377, opened 2021-08-16
Define and detect missing required config keys

### LINKED ISSUE #2363: Do not start if a required configuration parameter is missing
Ideally find >1 parameter, e.g. block adapter type.  Auth keys are _less useful_ for this: in some deployments they will be loaded from environment variables, so the auth key may be present despite not having a config file.

Candidates:
- Block adapter configuration (`blockstore.type`)
- Auth storage secret (`auth.encrypt.secret_key`)

arielshaqed:
Fixes #2363.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -24,7 +24,6 @@ import (
 )
 
 const (
-	DefaultBlockStoreType                    = "local"
 	DefaultBlockStoreLocalPath               = "~/data/lakefs/block"
 	DefaultBlockStoreS3Region                = "us-east-1"
 	DefaultBlockStoreS3StreamingChunkSize    = 2 << 19         // 1MiB by default per chunk
@@ -62,10 +61,11 @@ const (
 )
 
 var (
-	ErrBadConfiguration  = errors.New("bad configuration")
-	ErrMissingSecretKey  = fmt.Errorf("%w: auth.encrypt.secret_key cannot be empty", ErrBadConfiguration)
-	ErrInvalidProportion = fmt.Errorf("%w: total proportion isn't 1.0", ErrBadConfiguration)
-	ErrBadDomainNames    = fmt.Errorf("%w: domain names are prefixes", ErrBadConfiguration)
+	ErrBadConfiguration    = errors.New("bad configuration")
+	ErrMissingSecretKey    = fmt.Errorf("%w: auth.encrypt.secret_key cannot be empty", ErrBadConfiguration)
+	ErrInvalidProportion   = fmt.Errorf("%w: total proportion isn't 1.0", ErrBadConfiguration)
+	ErrBadDomainNames      = fmt.Errorf("%w: domain names are prefixes", ErrBadConfiguration)
+	ErrMissingRequiredKeys = fmt.Errorf("%w: missing required keys", ErrBadConfiguration)
 )
 
 type Config struct {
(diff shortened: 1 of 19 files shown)


arielshaqed:
This will probably conflict with #2370, make sure to rebase to `master` HEAD before merging.

arielshaqed:
Ready for review! :-)

1. Passes all its tests.
2. Tested manually with and without required fields.
3. Scanned docs and reference to see that we have no config files without the required fields.
4. https://compose.lakefs.io also includes the required fields.
5. treeverse/charts#86 makes it work with our Helm chart, and I tested that that loads by pushing my own Docker image and running the test.  But please review that one too!

nopcoder (approved):
Did you consider a method IsValid to Config that will check if the required field is not empty or just return the list of missing required fields?
Unless you think we will require more fields in the future - think it is a bit complex or too generic (if there is such thing)

arielshaqed:
Thanks!

>  Unless you think we will require more fields in the future - think it is a bit complex or too generic (if there is such thing)

I think that we will have more fields in future.  The actual checking code is surprisingly fragile to write by hand, particularly when pointers are present.  That's when you begin to feel the missing `.?` operator of ECMAScript.  And I think that we should use it in `lakectl` config as well.

I did write some fairly extensive tests, in order to increase confidence.

Also I want `template.go` to be the single source of truth for configuration.  Consider e.g. wanting to generate the list of configuration keys automatically.

Still waiting on treeverse/charts#86.

pkg/config/config.go
```diff
@@ -96,7 +96,24 @@ func NewConfig() (*Config, error) {
...
+
+	return c, nil
+}
+
+func NewConfig() (*Config, error) {
+	c, err := newConfigInternal()
+	if err != nil {
+		return nil, err
+	}
+	missingKeys := GetMissingRequiredKeys(c.values, "mapstructure", "squash", "required")
```
nopcoder, line 108: consider using "verify" or other tag names that will hold "required" as value, later we can add more constraints.

pkg/config/struct_keys.go
```diff
@@ -25,26 +25,108 @@ func appendStructKeys(typ reflect.Type, tag, squashValue string, prefix []string
...
+	case reflect.Chan:
+	case reflect.Map:
+	case reflect.Ptr:
+	case reflect.Slice:
+		return false
+	}
+	return true
+}
+
+func appendStructKeysIfZero(value reflect.Value, typ reflect.Type, tag, squashValue, requiredTag string, prefix []string, keys []string) []string {
```
johnnyaug, line 76: This function is very hard to read. Can you please add some comments?

johnnyaug (changes requested):

pkg/config/config.go
```diff
@@ -96,7 +96,24 @@ func NewConfig() (*Config, error) {
...
+	if len(missingKeys) > 0 {
+		return nil, fmt.Errorf("%w: %v", ErrMissingRequiredKeys, missingKeys)
+	}
+	return c, nil
+}
+
+func NewConfigIncompleteForTesting() (*Config, error) {
```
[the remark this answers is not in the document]
arielshaqed: Did (1)!

arielshaqed (commented):
Thanks; PTAL!

johnnyaug (approved):

arielshaqed:
Thanks!  Fixed merge conflicts, waiting for tests and then pushing...

Issue #2190, opened 2021-07-01
API with unknown path should return an error


nopcoder:
Currently lakeFS register openapi handlers and handle all specific routes.
In case of a call to `/api/v1/test`, the unknown path under the API prefix, the mux will serve the request by the UI handler and return a valid HTML (UI) page.

The expected behaviour is to return a non-2xx status code with JSON error - prefered the internal error format, so the developer will handle an error and not fail to parse the response in case of a bad API call.

nopcoder:
The above bug also affects `hadoop-lakefs` when the lakefs api endpoint is configured currently.
Example: https://prod.lakefs.dev/api/v/
No handler registered to handle this call and it falls to the default, which is UI handler that returns html with 2xx.

DataDavD:
Hey @nopcoder I haven't dug into the code specifics yet but I am hoping to take this up if that is fine, ok? Thanks in advance!

Best,
David D

nopcoder:
sure, thanks!

nopcoder:
@DataDavD let me know if you have any questions or need help in this one.

DataDavD:
Ok great, thank you @nopcoder. I'll take a close look this weekend and return with any questions I'll probably have. 

Thanks again!!!!
David

DataDavD:
hey @nopcoder just wanted to give you an update. I was able to take a look at the code and believe I understand the specific issue at hand. I intend to look much more closely tomorrow or the following day now that I pinpointed the code at issue; I will most likely send you some clarifying questions then. 

Thank you,
dd

nopcoder:
thanks for the updates, here for anything that will help you get this one moving.

DataDavD:
hey @nopcoder I am assuming the issue is centered around this line of code. I've been having a little trouble reading through the code because a number of objects/types like `HandlerFromMuxWithBaseURL` show as unresolved refs and I believe are causing my `make build` to fail with error `error generating code: error generating Go handlers for Paths: error generating inlined spec: template: inline.tmpl:5:8: executing "inline.tmpl" at <.>: range can't iterate over` . Do you know what the issue is or does this even prevent correcting it, or is just an annoyance for now?

Also to clarify, for any route past /api/v1 return json 401 error instead of UIHandler, correct?

nopcoder:
Hi @DataDavD, make build should run `go generate` and fill the missing parts - the specific one is lakefs.gen.go which includes the openapi generated code.
[...]

PR #2382, opened 2021-08-17
Fix/api unknown path return error 2190

DataDavD:
This PR fixes Issue #2190 by creating a handler to return invalid error for routes downstream of the `BaseURL`.

This PR fixes the issue by creating a handler that ensures that only routes on the mounted route (i.e. `/api/v1`) will be successful. Other routes are returned with `ErrInvalidAPIEndpoint` with the error string being `invalid API endpoint`.

This handler can be used with future/other routes by chi Mount'ing the handler to the specific route you want so that users are returned with the invalid endpoint error when trying to access any downstream route from the route/pattern the handler is chi Mount'ed to. 

Tests were successful for the tested for this fix (test created by @nopcoder - thank you!), but I still had issues with getting `TestLocalLoad` to pass successfully. I wanted to put this PR up for review and then we can talk about any issues it has AND how to resolve the testing issue for `TestLocalLoad`. 

Thank you.

### DIFF
--- a/pkg/api/serve.go
+++ b/pkg/api/serve.go
@@ -33,6 +33,8 @@ const (
 	extensionValidationExcludeBody = "x-validation-exclude-body"
 )
 
+var ErrInvalidAPIEndpoint = errors.New("invalid API endpoint")
+
 type responseError struct {
 	Message string `json:"message"`
 }
@@ -85,6 +87,7 @@ func Serve(
 	r.Mount("/_pprof/", httputil.ServePPROF("/_pprof/"))
 	r.Mount("/swagger.json", http.HandlerFunc(swaggerSpecHandler))
 	r.Mount("/", uiHandler)
+	r.Mount(BaseURL, http.HandlerFunc(InvalidAPIEndpointHandler))
 	return r
 }
 
@@ -161,4 +164,11 @@ func validateRequest(r *http.Request, router routers.Router, options *openapi3fi
 		return http.StatusInternalServerError, err
 	}
 	return http.StatusOK, nil
+}
+
+// InvalidAPIEndpointHandler returns ErrInvalidAPIEndpoint, and is currently being used to ensure
+// that routes under the pattern it is used with in chi.Router.Mount (i.e. /api/v1) are
+// not accessible.
+func InvalidAPIEndpointHandler(w http.ResponseWriter, _ *http.Request) {
+	writeError(w, http.StatusInternalServerError, ErrInvalidAPIEndpoint)
 }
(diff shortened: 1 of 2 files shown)


pkg/api/ui_handler.go
```diff
@@ -61,8 +61,29 @@ func NoCacheHandler(handler http.Handler) http.Handler {
 	})
 }
 
+// BaseURLHandler ensures that routes under BaseURL (i.e. /api/v1) are returned
+// with ErrInvalidAPIEndpoint.
+func BaseURLHandler(handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		fmt.Print("printing the url.path now", r.URL.Path)
```
nopcoder, line 68: remove Print from final code

pkg/api/ui_handler.go
```diff
@@ -61,8 +61,29 @@ func NoCacheHandler(handler http.Handler) http.Handler {
...
+		if r.URL.Path != "/" {
```
nopcoder, line 69: The server uses mux to handle the requested route, if we register the invalid API error handler on the base URL route - it will catch all the calls that don't match the current API handlers.
This way we will not require any 'if' to check the 'r.URL.Path' because the router will do this job for us.

```go
r.Mount(BaseURL, http.HandleFunc(InvalidAPIEndpointHandler))
```

This handler will be just response with an error:

```go
func InvalidAPIEndpointHandler(w http.ResponseWriter, _ *http.Request) {
	writeError(w, http.StatusInternalServerError, ErrInvalidAPIEndpoint)
}
```

With the above, we can move this code to `serve.go` - no longer related to the UI handler.

nopcoder (changes requested):

pkg/api/ui_handler.go
DataDavD: Ahh yes, thank you for catching this, completely forgot/didn't see to remove this leftover debug print ha

DataDavD: Done

pkg/api/ui_handler.go
DataDavD: Done

DataDavD:
Hey @nopcoder thank you so much for the great suggestions/info (i.e. basically doing the work for me 🤣). Anyways, I made the changes needed. I tried to rebase my history so I don't have multiple commits just adding and removing stuff, but since my rebase experience is lacking I had a hard time. 

FYI I think I got it right, but please let me know if its an issue. To be clear, so long as I rebase my forked commits, rebasing should be fine before I merge my PR branch, correct?

Thanks again for all your help.

best,
dd

DataDavD:
Hey @nopcoder FYI I forgot to mention that the tests were still successful after the change, but I still had the `TestLocalLoad` failing :(

nopcoder:
@DataDavD all you - will squash and merge it after the tests will complete green on github.
If you like me to help with debug why the test doesn't run locally - you are welcome to jump on our slack and post the error you see, me and my team would be happy to help.

nopcoder (approved):

DataDavD:
Ok yes, I will definitely do that, especially since I want to help contribute more. 

Just to confirm, the PR is fine even though that local test is failing on my side?

Issue #1779, opened 2021-04-19
Complete multipart response - wrong location field  


guy-har:
The response from `Complete multipart upload` contains a Location Field.
The value returned seems invalid in two aspects:
1. `o.Repository` is used where `o.Repository.Name` should be used
2. The `FQDN` value will not always be correct (e.g when the endpoint is configured without port but the request contains port)

We should check if Location returned in the result is valid and if not, fix it so it would be as expected.
https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html

ozkatz:
This is partially solved by #2390 (the right repository name field is used).
The FQDN populated on `o.FQDN` will not contain the proper port if a non-default one is used.

nopcoder:
#2456 algin the value of 'Location' in response to include: host, port, and consider path-style based on the request format.

PR #2390, opened 2021-08-22
fix possible sigv2 bug when no s3 gw domain is set

ozkatz:


guy-har (approved):
LGTM

arielshaqed:
Possibly relevant to #1779 (in spirit if not actually fixing it...).

ozkatz:
> Possibly relevant to #1779 (in spirit if not actually fixing it...).

@arielshaqed I think it solves the wrong stringification (`%s`-ing on `o.Repository` instead of using `o.Repository.Name`)

It doesn't seem like the FQDN field will include a port if one was specified in the request, as it is stripped from `req.Host` before populating the FQDN field. So the bug remains.

PR #2394, opened 2021-08-23
Use pgx driver defaults in migration

### LINKED ISSUE #2302: Remove default value for db_connection_string
It is hardly usable, and it masks away errors when the config isn't loaded properly.
Better to exit if that value is missing, instead of returning a connection error which isn't always very indicative to the fact that there's no db running in the other end.

arielshaqed:
golang-migrate has a subtly different default: it relies on a database name
inside the DB connection URL.  Explicitly create migration database.Driver,
to allow using the empty default value.

Fixes #2302.

### DIFF
--- a/pkg/db/migration.go
+++ b/pkg/db/migration.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/golang-migrate/migrate/v4"
+	"github.com/golang-migrate/migrate/v4/database"
 	"github.com/golang-migrate/migrate/v4/database/postgres"
 	"github.com/golang-migrate/migrate/v4/source"
 	"github.com/golang-migrate/migrate/v4/source/httpfs"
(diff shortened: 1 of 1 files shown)


nopcoder (dismissed):
Simple - like!

arielshaqed:
Doesn't actually migrate with an empty connection string, sorry.  Dismissing review, will explicitly request another review when it appears actually to work :shrug:.

arielshaqed:
Actually sorry, it _does_ work.  Tested by removing the connection_string from my config, and then running with
```ini
    PGUSER=lakefs
    PGPASSWORD=lakefs
```

Works for both `migrate` and `run`.

Please see that the above test convinces you, and if so -- please reapprove.

nopcoder (approved):
Still thinks its doing good

arielshaqed:
Thanks -- and sorry for the runaround!

Issue #2309, opened 2021-07-29
Fix failing HadoopFS tests


johnnyaug:
Some of the Hadoop FileSystem [contract tests](https://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/filesystem/testing.html) are not passing and have been commented out.

Analyze the reason for the failures, and decide whether or not a fix in lakeFS HadoopFS is needed.
It's possible that some tests are not applicable to lakeFS, or that the tests needs to be adapted to lakeFS.

The list of failing tests:
- In TestLakeFSFileSystemContract:
  - [ ] testWorkingDirectory
  - [x] testRenameFileMoveToNonExistentDirectory
  - [x] testListStatus
  - [x] testRenameDirectoryMoveToNonExistentDirectory
  - [x] testRenameFileAsExistingDirectory
  - [x] testListStatusThrowsExceptionForNonExistentFile
  - [x] testOverwrite
  - [x] testMkdirs
  - [x] testRenameFileAsExistingFile
  - [x] testRenameDirectoryAsExistingFile
  - [x] testRenameNonExistentPath
- In TestLakeFSContractRename:
  - [x] testRenameNonexistentFile
  - [x] testRenameFileNonexistentDir
- In TestLakeFSContractDelete:
  - [x] testDeleteDeepEmptyDir
- In TestLakeFSContractCreate:
  - [x] testCreateFileOverExistingFileNoOverwrite

nopcoder:
Fixing the contract was done by aligning the lakeFS file system with s3a v2.7 implementation for the above tests.

PR #2395, opened 2021-08-23
test: fix hadoop-fs 'delete deep empty directory' contract test

nopcoder:
Part of #2309 - fix delete deep empty directory


clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -749,6 +749,7 @@ public boolean exists(Path path) throws IOException {
         }
         // use listing to check if directory exists
         ListingIterator iterator = new ListingIterator(path, true, 1);
+        iterator.setRemoveDirectory(false);
```
arielshaqed, line 752: Good catch!

This is *obviously* a bad move for a readonly operation such as `exists`.  Do you know why we had this in the first place?

guy-har (approved):
LGTM, Nice catch

PR #2396, opened 2021-08-24
test: fix hadoop fs 'rename' contract test

nopcoder:
Part of #2309 

Fix hadoop-lakefs contracts for TestLakeFSContractRename:
 - testRenameNonexistentFile
 - testRenameFileNonexistentDir

The contract was aligned to s3a on the following:
```
fs.contract.rename-returns-false-if-source-missing=false
fs.contract.rename-creates-dest-dirs=true
```

Based on s3a contract: https://github.com/apache/hadoop/blob/rel/release-3.3.1/hadoop-tools/hadoop-aws/src/test/resources/contract/s3a.xml

### DIFF
--- a/clients/hadoopfs/src/test/java/io/lakefs/contract/TestLakeFSContractRename.java
+++ b/clients/hadoopfs/src/test/java/io/lakefs/contract/TestLakeFSContractRename.java
@@ -1,12 +1,10 @@
 package io.lakefs.contract;
 
 import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.fs.contract.AbstractContractRenameTest;
-import org.apache.hadoop.fs.contract.AbstractFSContract;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
-import org.apache.hadoop.fs.contract.ContractTestUtils;
-import org.junit.Test;
+import org.apache.hadoop.fs.contract.AbstractContractRenameTest;
+import org.apache.hadoop.fs.contract.AbstractFSContract;
 
 import static org.apache.hadoop.fs.contract.ContractTestUtils.dataset;
 import static org.apache.hadoop.fs.contract.ContractTestUtils.writeDataset;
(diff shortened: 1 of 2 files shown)


clients/hadoopfs/src/test/java/io/lakefs/contract/TestLakeFSContractRename.java
```diff
@@ -1,12 +1,10 @@
...
 
 import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.fs.contract.AbstractContractRenameTest;
-import org.apache.hadoop.fs.contract.AbstractFSContract;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
-import org.apache.hadoop.fs.contract.ContractTestUtils;
-import org.junit.Test;
+import org.apache.hadoop.fs.contract.AbstractContractRenameTest;
+import org.apache.hadoop.fs.contract.AbstractFSContract;
```
arielshaqed, line 7: Why is this import now needed?

arielshaqed (changes requested):
Not sure I understand the title.  AFAICS this PR does not change the SUT, it only moves disabling the contract from Java code to the XML properties file?

That may well be the intention ("not sure I understand..." above).  So if intended, could you change the title and add an explanation of why we want it disabled (probably as a comment above the disabling property)?

clients/hadoopfs/src/test/java/io/lakefs/contract/TestLakeFSContractRename.java
nopcoder: It wasn't added - just got to the right place after sort.

nopcoder:
> Not sure I understand the title. AFAICS this PR does not change the SUT, it only moves disabling the contract from Java code to the XML properties file?
> 
> That may well be the intention ("not sure I understand..." above). So if intended, could you change the title and add an explanation of why we want it disabled (probably as a comment above the disabling property)?

The test suite code that checks each fs operation extract the behaviour from an xml file - there is one for s3a and one for lakefs now.
Setting a feature to true/false changes the test code that verify the behaviour.
What I did here is aligned the expectation from lakefs filesystem to s3a.

arielshaqed (approved):
Great, thanks for explaining!

PR #2401, opened 2021-08-24
fix hadoopfs-lakefs throw exception on listStatus of non-exist directory

nopcoder:


talSofer (approved):
Please link the pr to the issue you are addressing and write a short description.

PR #2404, opened 2021-08-25
add blockstore type to compose

johnnyaug:
The blockstore type has become a mandatory field in a recent version of lakeFS.


arielshaqed (approved):
# Yikes!

I introduced this bug in #2377 .  And I even _knew_ about this file and forgot about it.  :-(

guy-har (commented):
Nice!

Issue #2407, opened 2021-08-25
Docs: Expand on the the docker-compose.yml [Everything Bagel]


peacing:
Right now the README.md is woefully written. Let's make sure someone wanting to run the bagel can understand how to do it from the README!

PR #2408, opened 2021-08-25
#2407 update the readme bagel

peacing:
Improve the docker compose bagel environment readme

### DIFF
--- a/deployments/compose/README.md
+++ b/deployments/compose/README.md
@@ -1,14 +1,24 @@
 # lakeFS `docker-compose`
 
 Compose file for lakeFS using variety of services lakeFS can work with.
+Automatically creates a minIO bucket and lakeFS repository as part of the start up.
 
+To spin up the environment, run:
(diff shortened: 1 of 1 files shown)


deployments/compose/README.md
```diff
@@ -1,21 +1,31 @@
...
+docker compose ps
+```
 
 ### Hive Client
 
 Can access Hive Server using client, run under 'client' profile:
 
 ```sh
-docker-compose --profile client run --rm hive-client
+docker compose --profile client run --rm hive-client
```
guy-har, line 21: I remember Barak had some bad experience with `docker compose`
Not sure this was the case. @nopcoder is it OK to use `docker compose` here?

guy-har (approved):
Thanks!! 
This was missing, it makes the documentation more clear now.

I approve, but please wait to see if @nopcoder has any reason not to use `docker compose`

PR #2422, opened 2021-08-31
hadoop-lakefs remove the use of LakeFSLocatedFileStatus

### LINKED ISSUE #2239: Make ListingIterator include LakeFSFileStatus instead of LakeFSLocatedFileStatus
This issue is a follow-up task to #2237. 

ListingIterator was implemented to make `LakeFSFileSystem.listFiles` return RemoteIterator<LocatedFileStatus>. therefore, ListingIterator has to do with LakeFSLocatedFileStatus. It also used by multiple LakeFSFileSystem operations to identify whether a path represents a directory (lakeFS's common prefix). 

To avoid bugs similar to #2237, we would like to make the default listing capability return LakeFSFileStatus by default and convert it LakeFSLocatedFileStatus when there is a need.

nopcoder:
Close #2239

Reduce the use of LakeFSLocatedFileStatus in order to prevent errors in cases the caller use the type for different functionality.
The only case we need the located file status is foe listFiles API - in this PR we removed all the extra code and kept just the convertion at the file listing iterator level.


clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -149,7 +149,24 @@ public FSDataInputStream open(Path path, int bufSize) throws IOException {
...
+                return iterator.hasNext();
+            }
+
+            @Override
+            public LocatedFileStatus next() throws IOException {
+                LakeFSFileStatus status = iterator.next();
+                BlockLocation[] locations = new BlockLocation[0];
+                if (status.isFile()) {
+                    locations = getFileBlockLocations(status, 0, status.getLen());
+                }
```
arielshaqed, line 166: This always allocates an empty array.  In Go that's almost fine, but it is a sin against the Java garbage collector.
```suggestion
                BlockLocation[] locations = status.isFile() ? getFileBlockLocations(status, 0, status.getLen()) : new BlockLocation[0];
```

arielshaqed (approved):
Thanks!  But please be sure not to allocate that unused size=0 array, it tends to be surprisingly nasty.

talSofer (approved):
LGTM!

PR #2426, opened 2021-08-31
Better exception message for unsupported FS

### LINKED ISSUE #2348: lakeFS Spark client does not validate storage configuration is supported
Currently the HadoopFS implementation used to delegate data access for Spark is only supported on AWS S3 using the s3a driver. 

While we will probably extend support for other object stores and FS implementations in the future, the client should probably fail with an explicit, easy to understand error when trying to use an unsupported storage backend.

itaiad200:
closes #2348

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -57,7 +57,7 @@ public class LakeFSFileSystem extends FileSystem {
                 return new URI("s3a", uri.getUserInfo(), uri.getHost(), uri.getPort(), uri.getPath(), uri.getQuery(),
                         uri.getFragment());
             default:
-                throw new RuntimeException(String.format("unsupported URI scheme %s, lakeFS file-system currently supports s3a only", uri.getScheme()));
+                throw new RuntimeException(String.format("unsupported URI scheme %s, lakeFS FileSystem currently supports translating s3 => s3a only", uri.getScheme()));
         }
     }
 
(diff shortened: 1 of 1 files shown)


clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -57,7 +57,7 @@ private URI translateUri(URI uri) throws java.net.URISyntaxException {
                 return new URI("s3a", uri.getUserInfo(), uri.getHost(), uri.getPort(), uri.getPath(), uri.getQuery(),
                         uri.getFragment());
             default:
-                throw new RuntimeException(String.format("unsupported URI scheme %s", uri.getScheme()));
+                throw new RuntimeException(String.format("unsupported URI scheme %s, lakeFS file-system currently supports s3a only", uri.getScheme()));
```
arielshaqed, line 60: This message may be a bit confusing, as the URI scheme returned from lakeFS will be `s3`.  (E.g. consider if you managed to convince the lakeFS server to return URI scheme `s3a`.  This code would throw a message "unsupported URI scheme s3a, lakeFS file-system currently supports s3a only"...).

Maybe something like this?
```suggestion
                throw new RuntimeException(String.format("cannot translate URI scheme %s from lakeFS server, lakeFS FileSystem currently supports translating s3 => s3a only", uri.getScheme()));
```

arielshaqed (approved):
Neat; thanks!

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
itaiad200: Translation part might confuse the users, adding the second part and keeping the `Unsupported` part explicit.

Issue #2246, opened 2021-07-15
Allow restricting users to create repositories only on specific namespaces


ozkatz:
Currently users can either create repo or not (determined by whether or not they have `fs:CreateRepository` privileges).
In some cases, admins might want to restrict creation to only allow certain users to create repositories on a pre-determined bucket or prefix. 

Suggestion: on repo creation, validaet the `fs:CreateRepository` permission as well as a new `fs:AttachRepositoryNamespace` permission where the `resource` is an ARN containing the storage namespace passed.

In conjunction with patterns/wildcards, it would allow a pretty flexible language to allow subsets of users to create repos on a certain convention of namespaces (bucket-per-team, prefix-of-bucket-name-per-team, paths within a bucket per user, etc)

ozkatz:
dup of #2220

PR #2430, opened 2021-09-01
Add "attach storage namespace [to repository]" IAM action type

### LINKED ISSUE #2220: Support policies to limit lakefs users to repo operations for specific buckets/blobs
According to the documentation [Authorization](https://docs.lakefs.io/reference/authorization.html) It seems like we are not able to limit the scope for lakefs users to a specific bucket/blob. 

## Use case 

We require to set up a multi-tenant data development environment such that teams are limited to managing repositories backed by a specific blob/bucket. This includes the create, delete repo operations only for repos that reside in a given bucket x. This would enable better encapsulation for data teams and make it easier to manage the stores from the cloud providers perspective as well. 

Using AWS as an example, I believe that being able to `lakectl auth users policies attach`  a functional policy like below to a lakefs user would suffice: 

```json 
{
  "Action": [
    "fs:*"
  ],
  "Effect": "Allow",
  "Resource": ["arn:aws:s3:::<BUCKET_NAME>", "arn:aws:s3:::<BUCKET_NAME_WITH_PATH_PREFIX>/*"]
}
```

Maybe, this could be generalised by providing an `arn` for the storage namespace: `arn:lakefs:fs:::storage/mystorage/*`.  If it's so that the mapping between storage namespace and repo is not 1-1, rather One To Many.

arielshaqed:
Fixes #2220 (and see discussion there).

### DIFF
--- a/pkg/ddl/000032_iam_add_attach_namespace.up.sql
+++ b/pkg/ddl/000032_iam_add_attach_namespace.up.sql
@@ -1,0 +1,45 @@
+BEGIN;
+
+-- wild_stars returns T if pattern matches hay using IAM-style wildcards: *
+-- is any string, ? is a single char.  It fails if pattern contains % or _
+-- chars.
+CREATE OR REPLACE FUNCTION pg_temp.wild_stars(pattern VARCHAR, hay VARCHAR)
+RETURNS BOOLEAN LANGUAGE plpgsql IMMUTABLE AS $$
+    DECLARE
+        unsafe BOOLEAN;
+	match BOOLEAN;
+
+    BEGIN
+        SELECT pattern LIKE '%\%%' OR pattern LIKE '%\_%' INTO STRICT unsafe;
+	IF unsafe THEN
+	    RAISE EXCEPTION 'unsafe pattern % contains "%%" or "_"', pattern;
+	END IF;
+        SELECT hay LIKE replace(replace(pattern, '*', '%'), '?', '_') INTO STRICT match;
+	RETURN match;
+    END;
+$$;
+
+-- jsonb_string translates a JSONB string object to SQL TEXT.  See
+-- https://stackoverflow.com/a/58755595.
+CREATE OR REPLACE FUNCTION pg_temp.jsonb_string(j JSONB) RETURNS TEXT LANGUAGE sql IMMUTABLE AS $$
+    SELECT j ->> 0
+$$;
+
+UPDATE auth_policies
+SET statement = statement || '[{"Action": ["fs:AttachStorageNamespace"], "Effect": "allow", "Resource": "*"}]'::jsonb
+WHERE id IN (
+    SELECT DISTINCT id FROM (
+        SELECT id, pg_temp.jsonb_string(s->'Effect') AS effect, jsonb_array_elements(s->'Action') AS action
+	FROM (
+	    SELECT id, jsonb_array_elements(statement) s FROM auth_policies
+	    -- Update only statements that never mention
+	    -- AttachStorageNamespace.  So downgrade can do nothing, and
+	    -- re-upgrading will not re-add an existing statement or harm
+	    -- pre-existing policies.
+	    WHERE NOT jsonb_path_exists(statement, 'strict $[*].Action[*] ? (@ == "fs:AttachStorageNamespace")')
+	) y
+    ) x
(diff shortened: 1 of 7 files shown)


arielshaqed:
@ozkatz,  following our chats here's a simple implementation which I believe implements the solution that you outlined in #2246.  Is this what you intended?

ozkatz (changes requested):
LGTM! I'm missing 2 important parts though:
1. Builtin policies that allow creating a repository should now include this new permission with a wildcard to preserve existing behavior
2. a data migration such that any policy in the user's database that included createRepository should also get this new permission with a wildcard.

arielshaqed:
> LGTM! I'm missing 2 important parts though:
> 
>     1. Builtin policies that allow creating a repository should now include this new permission with a wildcard to preserve existing behavior

AFAIU this is _not_ required, we already have policies for actions `fs:*` that include ARNs `*`, and no specific policy applies to `CreateRepository`.

>     2. a data migration such that any policy in the user's database that included createRepository should also get this new permission with a wildcard.

True.  I believe that we can do this by adding a new rule allowing action `fs:AttachRepositoryNamespace` on `*` for any policy that allows `fs:CreateRepository`.  I think that I should add this with a check that it's not already been added, and then _do nothing on downgrade_: it is safer to keep the old permissions than to lose them silently.

arielshaqed:
Working on upgrade script.  There's a split between possibilities here, complexity of SQL balanced against code complexity for a script that should run only once.
I would like the script to "add[...] a new rule allowing action fs:AttachRepositoryNamespace on * for any policy that allows fs:CreateRepository ".  Three possibilities that I can see:
Golang all the way!  Pros: probably more readable, easier to get right. uses a good programming language.  Cons: (very) poor ops experience: not part of the migration library that we use (or, indeed, any reasonable migration library). 
PostgreSQL jsonb SQL whackiness.  Write 100% safe pure-SQL code that updates the table.  Pros: Perfect ops experience.  Cons: The function itself will be tricky to write - it needs to handle IAM-style wildcards (*, ?) entirely correctly.
PostgreSQL jsonb SQL whackiness but less whacky.  Write SQL code that works on every reasonable IAM policy, but fails if the policy uses special (unused!) characters in action names.  Fail migration if such a policy is found (unlikely; such a policy includes an action that is unused!).  Specifically, fail for actions that include SQL-style wildcards (_. %); there are no such actions in lakeFS (or anywhere in IAM, really).

Unless I hear strenuous objections -- hopefully accompanied by suggestions how to do it better -- I am going with the third option.

arielshaqed:
@ozkatz I added data migration; PTAL!  The strategy is non-trivial, so please read also the e8f26a536 commit description.

pkg/ddl/000032_iam_add_attach_namespace.up.sql
```diff
@@ -0,0 +1,45 @@
...
+    END;
+$$;
+
+-- jsonb_string translates a JSONB string object to SQL TEXT.  See
+-- https://stackoverflow.com/a/58755595.
+CREATE OR REPLACE FUNCTION pg_temp.jsonb_string(j JSONB) RETURNS TEXT LANGUAGE sql IMMUTABLE AS $$
+    SELECT j ->> 0
+$$;
+
+UPDATE auth_policies
```
ozkatz, line 28: We don't do integration tests for migrations, but I'm assuming this was somehow tested nonetheless?

ozkatz (approved):
LGTM!

ozkatz (changes requested):
Actually, not LGTM :)
`jsonb_path_exists` is supported on PostgreSQL >= 12
As per [our docs](https://docs.lakefs.io/deploy/aws.html#creating-the-database-on-aws-rds) we support >= 11 (I believe this is also why your tests are failing)

pkg/ddl/000032_iam_add_attach_namespace.up.sql
arielshaqed: Tested manually: I added manual policies and upgraded and downgraded.  Also upgraded multiple times to ensure idempotency.

You're correct, of course: we have no Postgres testing facility to run unit tests on these migration scripts.

ozkatz (approved):
Thanks! *Now* it LGTM :)

arielshaqed:
> Thanks! _Now_ it LGTM :)

Thanks!  Also a shout-out to our CI, which came through again :-)

Issue #2353, opened 2021-08-11
Test our own RocksDB reader


talSofer:
Verify that the RocksDB reader implemented in https://github.com/treeverse/lakeFS/pull/2335 can read sstable files written by the pebble sst writer used by the latest lakeFS implementation. 

What the test should do:
- Use the same pebble version **used by latest lakeFS** to write large sstable files of different sizes. 
- Use the Reader from https://github.com/treeverse/lakeFS/pull/2335 to read the files.

PR #2432, opened 2021-09-01
Test sst parser

talSofer:
Part of #2353 

### Change list
* Plumbing to generate customized sstables that can be consumed by the BlockPraserSpec scala tests. 
* Initial test cases for the java parser. 

The flow of the tests in this PR is as follows: 
* A docker container with a golang image is created once for all the added tests
* On startup the script generate-sst.sh runs and executes the go app that generates customized sstables, and jsond that include the output expected from the parser. 
* The Scala tests then copy the files from within the container to temp files the parser tests can operate on. 
* Scala tests use the sstable and json files copied from the container as their input. 

#### Note
The go app that generates the test input does this in a deterministic way. that is, in case a test fails it should be possible to reproduce the failure.

### DIFF
--- a/clients/spark/core/src/test/resources/parser-test/generate-sst.sh
+++ b/clients/spark/core/src/test/resources/parser-test/generate-sst.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+# Generate sst files that will serve as an input jpebble parser tests
+export CGO_ENABLED=0
+cd /local
+go run sst_files_generator.go
+echo "done"
--- a/clients/spark/core/src/test/resources/parser-test/generate-sst.sh
+++ b/clients/spark/core/src/test/resources/parser-test/generate-sst.sh
@@ -1,7 +1,0 @@
-#!/bin/sh
-
-# Generate sst files that will serve as an input jpebble parser tests
-export CGO_ENABLED=0
-cd /local
-go run sst_files_generator.go
-echo "done"
--- a/clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
+++ b/clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
@@ -3,32 +3,31 @@ package main
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"math/rand"
 	"os"
 	"sort"
 
-	fuzz "github.com/google/gofuzz"
-
 	"github.com/cockroachdb/pebble/sstable"
+	fuzz "github.com/google/gofuzz"
 	nanoid "github.com/matoous/go-nanoid/v2"
 )
 
 const (
 	DefaultKeySizeBytes   = 100
 	DefaultValueSizeBytes = 5
-	MbToBytes             = 1024 * 1024
-	KbToBytes             = 1024
+	KiBToBytes            = 1 << 10
+	MiBToBytes            = KiBToBytes << 10
 	FuzzerSeed            = 50
-	// DefaultCommittedPermanentMaxRangeSizeMb - the max file size that can be written by graveler (pkg/config/config.go)
-	DefaultCommittedPermanentMaxRangeSizeMb = 20
+	// DefaultCommittedPermanentMaxRangeSizeBytes - the max file size that can be written by graveler (pkg/config/config.go)
+	DefaultCommittedPermanentMaxRangeSizeBytes = 20 * MiBToBytes
+	DefaultSstSizeBytes                        = 50 * KiBToBytes
+	DefaultTwoLevelSstSizeBytes                = 10 * KiBToBytes
+	DefaultMaxSizeKb                           = 100
 )
 
-func getDefaultUserProperties() map[string]string {
-	return map[string]string{
-		"user_prop_1": "val1",
-		"user_prop_2": "val2",
-	}
+var DefaultUserProperties = map[string]string{
+	"user_prop_1": "val1",
+	"user_prop_2": "val2",
 }
 
 type Entry struct {
@@ -37,16 +36,29 @@ type Entry struct {
 }
 
 func main() {
-	generateTwoLevelIdxSst()
-	fuzzSstContents()
-	fuzzWriterOptions()
-	sstsWithUnsupportedWriterOptions()
-	generateLargeSsts()
+	writeTwoLevelIdxSst()
+	writeMultiSizedSstsWithContentsFuzzing()
+	writeSstsWithWriterOptionsFuzzing()
+	writeSstsWithUnsupportedWriterOptions()
+	writeLargeSsts()
 }
 
-func sstsWithUnsupportedWriterOptions() {
-	sizeBytes := 10 * KbToBytes
-	keys := generateSortedSlice(sizeBytes)
+func generateNanoid() (string, error) {
+	return nanoid.New(DefaultValueSizeBytes)
+}
+
+func newGenerateFuzz() func() (string, error) {
+	f := fuzz.NewWithSeed(FuzzerSeed)
+	return func() (string, error) {
+		var val string
+		f.Fuzz(&val)
+		return val, nil
+	}
+}
+
+func writeSstsWithUnsupportedWriterOptions() {
+	sizeBytes := DefaultSstSizeBytes
+	keys := prepareSortedSliceWithNanoid(sizeBytes)
 
 	// Checksum specifies which checksum to use. lakeFS supports parsing sstables with crc checksum.
 	// ChecksumTypeXXHash and ChecksumTypeNone are unsupported and therefore not tested.
@@ -54,7 +66,7 @@ func sstsWithUnsupportedWriterOptions() {
 		Compression: sstable.SnappyCompression,
 		Checksum:    sstable.ChecksumTypeXXHash64,
 	}
-	writeTestInputFiles(keys, func() (string, error) { return nanoid.New(DefaultValueSizeBytes) }, sizeBytes,
+	createTestInputFiles(keys, generateNanoid, sizeBytes,
 		"checksum.type.xxHash64", writerOptions)
 
 	// TableFormat specifies the format version for writing sstables. The default
@@ -63,22 +75,21 @@ func sstsWithUnsupportedWriterOptions() {
 		Compression: sstable.SnappyCompression,
 		TableFormat: sstable.TableFormatLevelDB,
 	}
-	writeTestInputFiles(keys, func() (string, error) { return nanoid.New(DefaultValueSizeBytes) }, sizeBytes,
+	createTestInputFiles(keys, generateNanoid, sizeBytes,
 		"table.format.leveldb", writerOptions)
 }
 
-func fuzzWriterOptions() {
-	sizeBytes := 50 * KbToBytes
-	keys := generateSortedSlice(sizeBytes)
+func writeSstsWithWriterOptionsFuzzing() {
+	sizeBytes := DefaultSstSizeBytes
+	keys := prepareSortedSliceWithNanoid(sizeBytes)
 
 	// Use fuzzing to define sstable user properties.
-	f := fuzz.NewWithSeed(FuzzerSeed)
+	generateFuzz := newGenerateFuzz()
 	userProps := map[string]string{}
-	for i := 1; i <= 20; i++ {
-		var propKey string
-		var propVal string
-		f.Fuzz(&propKey)
-		f.Fuzz(&propVal)
+	const numOfUserDefinedProps = 20
+	for i := 1; i <= numOfUserDefinedProps; i++ {
+		propKey, _ := generateFuzz()
+		propVal, _ := generateFuzz()
 		userProps[propKey] = propVal
 		fmt.Printf("user property %d, key=%s, val=%s\n", i, propKey, propVal)
 	}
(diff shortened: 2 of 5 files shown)


clients/spark/core/src/test/resources/parser-test/generate-sst.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+# Generate sst files that will serve as an input jpebble parser tests
+export CGO_ENABLED=0
```
arielshaqed, line 4: Why do we need this?

clients/spark/core/src/test/resources/parser-test/generate-sst.sh
```diff
@@ -0,0 +1,7 @@
...
+cd /local
+go run sst_files_generator.go
+echo "done"
```
arielshaqed, line 7: This line means that the shell script always succeeds, even when it fails.  Please use e.g. `#!/bin/sh -e`, and maybe even `#!/bin/sh -eu`.  Run `help set` in `bash` for more about some useful shell options (you can also say `set -e; set -u` and similar to set these once the shell is already running.

clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
```diff
@@ -0,0 +1,286 @@
...
+	sstsWithUnsupportedWriterOptions()
+	generateLargeSsts()
+}
+
+func sstsWithUnsupportedWriterOptions() {
+	sizeBytes := 10 * KbToBytes
+	keys := generateSortedSlice(sizeBytes)
+
+	// Checksum specifies which checksum to use. lakeFS supports parsing sstables with crc checksum.
+	// ChecksumTypeXXHash and ChecksumTypeNone are unsupported and therefore not tested.
```
arielshaqed, line 52: Do we explicitly write our SSTables with these options?  If not, should we test (and fix...) these other options?

clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+func fuzzWriterOptions() {
+	sizeBytes := 50 * KbToBytes
+	keys := generateSortedSlice(sizeBytes)
+
+	// Use fuzzing to define sstable user properties.
+	f := fuzz.NewWithSeed(FuzzerSeed)
+	userProps := map[string]string{}
+	for i := 1; i <= 20; i++ {
```
arielshaqed, line 77: What is "20"?  Consider defining it as a const, then you get to give it a name.  (Matters because this creates test files that will be consumed by a separate test program.)

clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
```diff
@@ -0,0 +1,286 @@
...
+		Compression:             sstable.SnappyCompression,
+		TablePropertyCollectors: []func() sstable.TablePropertyCollector{NewStaticCollector(userProps)},
+	}
+	writeTestInputFiles(keys, func() (string, error) { return nanoid.New(DefaultValueSizeBytes) }, sizeBytes,
+		"fuzz.table.properties", writerOptions)
+
+	// BlockRestartInterval is the number of keys between restart points
+	// for delta encoding of keys.
+	// The default value is 16.
+	blockRestartInterval := rand.Intn(100)
```
arielshaqed, line 95: This is not seeded AFAICT: you set the `fuzz` seed but not the `rand` seed.

clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
```diff
@@ -167,7 +172,7 @@ class BlockParserSpec extends AnyFunSpec with Matchers {
         case histRe(count, word) => (word, count.toInt)
         case _ => throw new RuntimeException(s"Bad format h.txt line ${line}")
       }
-    ).toMap
+    ).toSeq
```
arielshaqed, line 175: Thanks!  :facepalm:

clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
```diff
@@ -261,3 +266,175 @@ class CountedIteratorSpec extends AnyFunSpec with Matchers {
...
+
+class GolangContainerSpec extends AnyFunSpec with ForAllTestContainer {
+
+  override val container: GenericContainer = GenericContainer("golang:1.16.2-alpine",
+    classpathResourceMapping = Seq(
+      ("parser-test/generate-sst.sh", "/local/generate-sst.sh", BindMode.READ_WRITE),
+      ("parser-test/sst_files_generator.go", "/local/sst_files_generator.go", BindMode.READ_WRITE),
+      ("parser-test/go.mod", "/local/go.mod", BindMode.READ_WRITE),
+      ("parser-test/go.sum", "/local/go.sum", BindMode.READ_WRITE)),
+    command = Seq("/local/generate-sst.sh"),
```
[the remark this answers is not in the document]
arielshaqed: Yes, please!  I believe that will work in any case, even with a shell script.  (But you do need to `set -e`).

clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
```diff
@@ -261,3 +266,175 @@ class CountedIteratorSpec extends AnyFunSpec with Matchers {
...
+    }
+
+    val testFiles = Seq(
+      "fuzz.contents.0",
+      "fuzz.contents.1",
+      "fuzz.contents.2",
+      "fuzz.contents.3",
+      "fuzz.contents.4",
+      "fuzz.contents.5"
+    )
```
arielshaqed, line 303: I don't understand.  WHy are there 6 files?

In any case, maybe iterate over all the files in a directory if we can, since we have an application that generates them.

clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
```diff
@@ -261,3 +266,175 @@ class CountedIteratorSpec extends AnyFunSpec with Matchers {
...
+    val tmpSstFile = copyTestFile(filename, ".sst")
+    val tmpJsonFile = copyTestFile(filename, ".json")
+
+    val in = new BlockReadableFileChannel(new java.io.FileInputStream(tmpSstFile).getChannel)
+    val jsonString = os.read(os.Path(tmpJsonFile.getAbsolutePath))
+    val data = ujson.read(jsonString)
+    val expected = data.arr.map(e => (e("Key").str, e("Value").str))
+    try {
+      test(in, expected, tmpSstFile.length())
+    } finally {
```
arielshaqed, line 436: Best pull all of these files inside the `try`, I guess, otherwise we can leave a file open for no good reason.

arielshaqed (changes requested):
Looks very nice; thanks!

johnnyaug (changes requested):

nopcoder (changes requested):

clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
```diff
@@ -0,0 +1,291 @@
...
+	DefaultKeySizeBytes   = 100
+	DefaultValueSizeBytes = 5
+	KiBToBytes            = 1 << 10
+	MiBToBytes            = KiBToBytes << 10
+	FuzzerSeed            = 50
+	// DefaultCommittedPermanentMaxRangeSizeBytes - the max file size that can be written by graveler (pkg/config/config.go)
+	DefaultCommittedPermanentMaxRangeSizeBytes = 20 * MiBToBytes
+	DefaultSstSizeBytes                        = 50 * KiBToBytes
+	DefaultTwoLevelSstSizeBytes                = 10 * KiBToBytes
+	DefaultMaxSizeKb                           = 100
```
arielshaqed, line 25:
```suggestion
	DefaultMaxSizeKiB                          = 100
```

clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
```diff
@@ -0,0 +1,291 @@
...
+		"checksum.type.xxHash64", writerOptions)
+
+	// TableFormat specifies the format version for writing sstables. The default
+	// is TableFormatRocksDBv2 which creates RocksDB compatible sstables. this is the only format supported by lakeFS.
+	writerOptions = sstable.WriterOptions{
+		Compression: sstable.SnappyCompression,
+		TableFormat: sstable.TableFormatLevelDB,
+	}
+	createTestInputFiles(keys, generateNanoid, sizeBytes,
+		"table.format.leveldb", writerOptions)
```
arielshaqed, line 79:
```suggestion
	createTestInputFiles(keys, generateNanoid, sizeBytes, "table.format.leveldb", writerOptions)
```

clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
```diff
@@ -0,0 +1,291 @@
...
+	writerOptions := sstable.WriterOptions{
+		Compression:             sstable.SnappyCompression,
+		TablePropertyCollectors: []func() sstable.TablePropertyCollector{NewStaticCollector(userProps)},
+	}
+	createTestInputFiles(keys, generateNanoid, sizeBytes, "fuzz.table.properties", writerOptions)
+
+	// BlockRestartInterval is the number of keys between restart points
+	// for delta encoding of keys.
+	// The default value is 16.
+	blockRestartInterval := 1 + rand.Intn(100)
```
arielshaqed, line 105: `rand` is still not seeded AFAICT; https://github.com/treeverse/lakeFS/pull/2432#discussion_r700965505 vanished due to a push, but I believe that it remains unresolved.

johnnyaug (approved):
Great!

clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
```diff
@@ -0,0 +1,291 @@
...
+	// would cause the block to be larger than the target block size.
+	//
+	// The default value is 90
+	blockSizeThreshold := 1 + rand.Intn(100)
+	writerOptions = sstable.WriterOptions{
+		Compression:             sstable.SnappyCompression,
+		TablePropertyCollectors: getDefaultTablePropertyCollectors(DefaultUserProperties),
+		BlockSizeThreshold:      blockSizeThreshold,
+	}
+	fmt.Printf("BlockSizeThreshold %d...\n", blockSizeThreshold)
```
arielshaqed, line 136: I believe that it may make more sense to mix all of the WriterOptions together.  So create X different files, with random values for each.

clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
```diff
@@ -0,0 +1,291 @@
...
+	generateFuzz := newGenerateFuzz()
+	// Keep generating keys until we reach the desired slice size. This guarantees that the sstable writer will have
+	// enough data to write an sstable of the desired size.
+	for keySumBytes < size {
+		key, _ := generateFuzz()
+		if key == "" { // TODO(Tals): remove after resolving https://github.com/treeverse/lakeFS/issues/2419
+			continue
+		}
+		keySumBytes += len(key)
+		slice = append(slice, key)
```
arielshaqed, line 183: This will barely go through the prefix decompression code, because keys will have only short shared prefixes.  Consider randomizing a prefix length at each key, then adding fuzz to _approximately_ a constant length.

clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
```diff
@@ -261,3 +266,175 @@ class CountedIteratorSpec extends AnyFunSpec with Matchers {
...
+  /**
+   * Lightweight fixture for running particular tests with SSTables and Json fules generated by a go application on
+   * test startup. The fixture uses sstables as the parser input, and the json as the source of the expected output of
+   * the parsing operation.
+   */
+  def withGeneratedTestFiles(filename: String, test: (BlockReadable, Seq[(String, String)], Long) => Any) = {
+    val tmpSstFile = copyTestFile(filename, ".sst")
+    val tmpJsonFile = copyTestFile(filename, ".json")
+
+    val in = new BlockReadableFileChannel(new java.io.FileInputStream(tmpSstFile).getChannel)
```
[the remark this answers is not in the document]
arielshaqed: A Java "SeekableInputStream" is actually called a FileChannel.  I think.

clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
```diff
@@ -261,3 +266,134 @@ class CountedIteratorSpec extends AnyFunSpec with Matchers {
...
+
+class GolangContainerSpec extends AnyFunSpec with ForAllTestContainer {
+
+  override val container: GenericContainer = GenericContainer("golang:1.16.2-alpine",
+    classpathResourceMapping = Seq(
+      ("parser-test/sst_files_generator.go", "/local/sst_files_generator.go", BindMode.READ_WRITE),
+      ("parser-test/go.mod", "/local/go.mod", BindMode.READ_WRITE),
+      ("parser-test/go.sum", "/local/go.sum", BindMode.READ_WRITE)),
+    command = Seq("/bin/sh", "-c", "cd /local && CGO_ENABLED=0 go run sst_files_generator.go && echo \"done\""),
+    waitStrategy = new LogMessageWaitStrategy().withRegEx("done\\n")
```
arielshaqed, line 278: I would _really_ like the comments @johnnyaug and I placed on this wait strategy to be resolved, or the reason to avoid them logged on this PR.  Why not use the [indefinite on-shot strategy](https://www.testcontainers.org/features/startup_and_waits/#indefinite-one-shot-startup-strategy-example)?  It is intended precisely for this option, and would simplify `command`, and is probably _safer_.

arielshaqed (changes requested):
Thanks!

clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
```diff
@@ -22,7 +22,7 @@ const (
 	DefaultCommittedPermanentMaxRangeSizeBytes = 20 * MiBToBytes
 	DefaultSstSizeBytes                        = 50 * KiBToBytes
 	DefaultTwoLevelSstSizeBytes                = 10 * KiBToBytes
-	DefaultMaxSizeKb                           = 100
+	DefaultMaxSizeKib                          = 100
```
arielshaqed, line 25:
```suggestion
	DefaultMaxSizeKiB                          = 100
```
or just use bytes:
```suggestion
	DefaultMaxSizeBytes                        = 100 * KibToBytes
```

clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
arielshaqed: Let's add a TODO and/or an issue to use the right strategy here!

clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
```diff
@@ -3,15 +3,17 @@ package io.treeverse.jpebble
 import org.scalatest._
 import matchers.should._
 import funspec._
-import com.dimafeng.testcontainers.{ForAllTestContainer, GenericContainer}
+import com.dimafeng.testcontainers.{DockerComposeContainer, ForAllTestContainer, GenericContainer}
```
arielshaqed, line 6: This is (probably) unused.

arielshaqed (approved):
Thanks!

clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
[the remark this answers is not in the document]
nopcoder: cool, so FileChannel.open should do the work

nopcoder (approved):

PR #2433, opened 2021-09-01
log lakefs configuration on startup

nopcoder:
Close #2416

Traverse and log our configuration on startup.

- Do not print out values for secrets
- Support the same field conventions that we used while loading. The dotted field name and squash levels

Issues to discuss:

- [x] which level do we like to use when we dump the information
- [x] implementation - can we reuse the same functionality found inside struct_keys

### DIFF
--- a/pkg/config/struct_fields.go
+++ b/pkg/config/struct_fields.go
@@ -1,0 +1,92 @@
+package config
+
+import (
+	"reflect"
+	"strings"
+
+	"github.com/treeverse/lakefs/pkg/logging"
+)
+
+const (
+	FieldMaskedValue   = "******"
+	FieldMaskedNoValue = "------"
+)
+
+// MapLoggingFields returns all logging.Fields withe formatted based on our configuration struct 'dot.name.key' with
+// the associated value. Support squash and secret to skip print out of secrets.
+func MapLoggingFields(value interface{}) logging.Fields {
+	fields := make(logging.Fields)
+	structFieldsFunc(reflect.ValueOf(value), "mapstructure", ",squash", "validate", "secret", nil, func(key string, value interface{}) {
+		fields[key] = value
+	})
+	return fields
+}
+
+func structFieldsFunc(value reflect.Value, tag, squashValue, validateTag, secretValue string, prefix []string, cb func(key string, value interface{})) {
+	// finite loop: Go types are well-founded.
+	for value.Kind() == reflect.Ptr {
+		if value.IsZero() {
+			// If required, would already have errored out.
+			return
+		}
+		value = value.Elem()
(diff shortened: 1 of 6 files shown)


pkg/config/struct_fields.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/treeverse/lakefs/pkg/logging"
+)
+
+const (
+	FieldMaskedValue   = "******"
+	FieldMaskedNoValue = "------"
+)
+
+// MapLoggingFields returns all logging.Fields withe formatted based on our configuration struct 'dot.name.key' with
+// the associated value. Support squash and secret to skip print out of secrets.
```
arielshaqed, line 16: This comment seems to have collected some typos.  Maybe something like
```suggestion
// MapLoggingFields returns all logging.Fields formatted based on our configuration keys 'dot.name.key' with
// associated values. Supports squash, and secret to skip printing out secrets.
```

pkg/config/struct_fields.go
```diff
@@ -0,0 +1,92 @@
...
+				}
+			}
+		}
+		// Update prefix to recurse into this field.
+		if !squash {
+			prefix = append(prefix, fieldName)
+		}
+		fieldValue := value.Field(i)
+		// Do not print or dive into secrets
+		if secret {
```
arielshaqed, line 77: I am not sure that we need a *tag* to specify that a value is _secret_.  (Also it is strange to put it on a tag "`validate`").  Consider implementing it as a _value type_.  So you have an interface `Secreter` and a type `SecretString`:
```go
type Secreter interface {
	SecretString() string
}
type SecretString string // implements Secreter
```
Now just call something like:
```go
valueString := fieldValue.Value.String()
if secreter, ok := fieldValue.Value.Interface().(Secreter); ok {
	valueString = secretr.SecretString()
}
```
and you have a good value.  This lets you also hide larger types, and seems somewhat safer (harder to forget, clearer at all points) than a tag.

Perhaps go one step further.  There is some sequence of calls to `.GoString` and `.String` etc. to convert an `interface{}` to a string.  You could just code
```go
type SecretString string
func (s SecretString) /*GoString?*/ String() string {
	return "****"
}
```
and I think you should be OK, with no need for any reflection.

arielshaqed (approved):
Thanks!

This is (very) good, but I'd like to suggest a non-reflective (diffusive?) alternative.

pkg/config/struct_fields.go
nopcoder: Good idea - and it can serve better at runtime. It will not handle cases of other types, mainly structs. But I don't see it as an issue at this point.

Issue #2307, opened 2021-07-29
Non-descriptive error message while trying to create an Admin user more than once 


talSofer:
### Description 
When a user tries to create an Admin user on a lakeFS setup that already has an Admin user, a non-descriptive error message is shown in the UI ("Conflict"). 
We should change the message to report the actual error, the message could be "Admin user already exists".

![Screen Shot 2021-07-29 at 9 24 04](https://user-images.githubusercontent.com/14786381/127442219-cb2401a2-9841-4d31-8a5c-921149fc4f5d.png)

### Steps to reproduce
1. Setup lakeFS or use an already existing setup
2. go to https://lakefs.domain/setup
3. Type in a username and click setup

DataDavD:
I'm not sure if anyone is currently working on this, but if not I would be happy to take it up.

talSofer:
Hi @DataDavD, Thanks for being an active community member!

@lynnro314 is currently working on this issue. You are welcome to choose another issue you would like to work on, if you pick an issue that doesn't have asignees, the chances that no one started working on it are higher. Let us know the issue you chose and we will confirm that no one started working on it :)

DataDavD:
Thank you for the quick, detailed response @talSofer. sounds good!

I asked on a couple other low activity/empty `good first issue ` issues actually so I'll look there too.

Thanks again.

Best, 
dd

DataDavD:
hey @talSofer FYI. I may get to this later than I originally anticipated as I'm only really just now getting started with https://github.com/treeverse/lakeFS/issues/2415#issuecomment-930441069 (had some work/life issues that took up a large part of time last couple weeks).  I'm happy to still work on this once I'm done with the other issue, but I just wanted to give you an update, just in case some knows/wants to work on this now/ASAP.

In the meantime, I'm still going to give this a more thorough look tonight since the last time I looked at the code for this issue a couple weeks ago I couldn't directly make how where the change should be centered. If I have more question I'll post those here as well as the lakeFS slack server. 

Thanks again for all y'alls help

Best,
David

talSofer:
Hi @DataDavD ! thanks for keeping me posted! 
This issue was already resolved by @lynnro314 so no further action is needed here. Thanks again.

DataDavD:
hey @talSofer thanks for update. I mistakenly thought I was making a comment on this issue https://github.com/treeverse/lakeFS/issues/2316#issuecomment-910048659 🤦 . Sorry about that.

PR #2435, opened 2021-09-01
Bugfix/2307 change error message in lakefs setup

lynnro314:
Close: #2307

When trying to set up lakeFS more than once, the error message was not descriptive.
I made two changes:
1. In the setup phase I added a verification of whether the setup already happened. In this case, a descriptive error message will be printed.
2. I changed the error message in the UI to be descriptive.

### DIFF
--- a/cmd/lakefs/cmd/setup.go
+++ b/cmd/lakefs/cmd/setup.go
@@ -63,7 +63,7 @@ var setupCmd = &cobra.Command{
 			os.Exit(1)
 		}
 		if initialized {
-			fmt.Printf("lakeFS already initialized.")
+			fmt.Printf("Setup is already complete.")
 			os.Exit(1)
 		}
 
--- a/cmd/lakefs/cmd/setup.go
+++ b/cmd/lakefs/cmd/setup.go
@@ -63,7 +63,7 @@ var setupCmd = &cobra.Command{
 			os.Exit(1)
 		}
 		if initialized {
-			fmt.Printf("Setup is already complete.")
+			fmt.Printf("Setup is already complete.\n")
 			os.Exit(1)
 		}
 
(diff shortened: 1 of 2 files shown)


nopcoder (changes requested):
Looks good, just think we need to use the same terminology - setup and not initialized and drop the lakeFS.

talSofer (commented):
Nice!

Please edit the following to the pr description:
1. Fixes #2307 (this automatically closes the issue when you merge the pr)
2. fix description. What was the problem, and what solution you implemented.

talSofer:
> Looks good, just think we need to use the same terminology - setup and not initialized and drop the lakeFS.

Agree with @nopcoder . To understand if setup already happened you are checking whether the metadata manager is already initialized which is an implementation detail. as a user I would want the software to speak the language of the actions I'm taking.

cmd/lakefs/cmd/setup.go
```diff
@@ -57,6 +57,16 @@ var setupCmd = &cobra.Command{
 		cloudMetadataProvider := stats.BuildMetadataProvider(logging.Default(), cfg)
 		metadata := stats.NewMetadata(ctx, logging.Default(), cfg.GetBlockstoreType(), metadataManager, cloudMetadataProvider)
 
+		initialized, err := metadataManager.IsInitialized(ctx)
+		if err != nil {
+			fmt.Printf("Setup failed: %s\n", err)
+			os.Exit(1)
+		}
+		if initialized {
+			fmt.Printf("Setup is already complete.")
```
nopcoder, line 66: Add new-line or use fmt.Println

nopcoder (changes requested):

nopcoder (approved):

PR #2436, opened 2021-09-02
Fix the double slash namespace bug

### LINKED ISSUE #2397: BlockAdapter namespace resolution is wrong when StorageNamespace is a prefix under the bucket
**Example**
Namespace: `gs://foo/bla`, Key: `bar/baz`

QualifiedKey:
Expected: StorageNamespace: `foo`, Key: `bla/bar/baz`
Actual: StorageNamespace: `foo`, Key: `/bla/bar/baz`

In some block storages like gs, this will result in the following object: `gs://foo//bla/bar/baz`. In S3, it doesn't make a difference and the object path is `gs://foo/bla/bar/baz`.

Since the adapter is consistent in the way it is referencing the object, we did not see an impact on usability when communicating with the lakeFS server solely.
BUT when trying to use other clients like the metadata client, which uses the physical address returned by the API (`gs://foo/bla/bar/baz`), operations fail with `NoSuchObject`.

Any fix to this issue MUST consider backward compatibility of existing repos which already stored according to the double-slash pattern.

itaiad200:
closes #2397 

### How was this tested?
1. Added unit-tests to verify that the fix works for new paths.
2. Created all 4 types of repositories in gcp & azure with the previous version of lakeFS. For some of the ones that have this bug, I stopped lakeFS and moved the data (`gsutil` for gcp, `azcopy` for azure). Then started the lakeFS version which has the fix, and in all cases I was able to read old data.

Is there any other test you think is missing?

### DIFF
--- a/cmd/lakefs/cmd/run.go
+++ b/cmd/lakefs/cmd/run.go
@@ -204,9 +204,7 @@ var runCmd = &cobra.Command{
 	},
 }
 
-// A foreign repo is a repository which namespace doesn't match the current block adapter.
-// A foreign repo might exists if the lakeFS instance configuration changed after a repository was
-// already created. The behaviour of lakeFS for foreign repos is undefined and should be blocked.
+// checkRepos iterates on all repos and validates that their settings is correct.
 func checkRepos(ctx context.Context, logger logging.Logger, authMetadataManager *auth.DBMetadataManager, blockStore block.Adapter, c *catalog.Catalog) {
 	initialized, err := authMetadataManager.IsInitialized(ctx)
 	if err != nil {
@@ -249,22 +247,30 @@ func checkRepos(ctx context.Context, logger logging.Logger, authMetadataManager 
 	}
 }
 
+// checkMetadataPrefix checks for non-migrated repos of issue #2397 (https://github.com/treeverse/lakeFS/issues/2397)
 func checkMetadataPrefix(ctx context.Context, repo *catalog.Repository, logger logging.Logger, adapter block.Adapter, repoStorageType block.StorageType) {
 	if repoStorageType != block.StorageTypeGS &&
 		repoStorageType != block.StorageTypeAzure {
 		return
 	}
 
-	dummyFile := "dummy"
+	const dummyFile = "dummy"
 	if _, err := adapter.Get(ctx, block.ObjectPointer{
 		StorageNamespace: repo.StorageNamespace,
 		Identifier:       dummyFile,
 	}, -1); err != nil {
-		logger.Fatalf("Can't find dummy (path: %s) file in storage namespace (%s), did you run the migration? "+
-			"(http://docs.lakefs.io/reference/upgrade.html#data-migration-for-version-v0490)", dummyFile, repo.StorageNamespace)
+		logger.WithFields(logging.Fields{
+			"path":              dummyFile,
+			"storage_namespace": repo.StorageNamespace,
+		}).Fatal("Can't find dummy file in storage namespace, did you run the migration? " +
+			"(http://docs.lakefs.io/reference/upgrade.html#data-migration-for-version-v0490)")
 	}
 }
 
+// checkForeignRepo checks whether a repo storage namespace matches the block adapter.
+// A foreign repo is a repository which namespace doesn't match the current block adapter.
+// A foreign repo might exists if the lakeFS instance configuration changed after a repository was
+// already created. The behaviour of lakeFS for foreign repos is undefined and should be blocked.
 func checkForeignRepo(repoStorageType block.StorageType, logger logging.Logger, adapterStorageType, repoName string) {
 	if adapterStorageType != repoStorageType.BlockstoreType() {
 		logger.Fatalf("Mismatched adapter detected. lakeFS started with adapter of type '%s', but repository '%s' is of type '%s'",
--- a/cmd/lakefs/cmd/run.go
+++ b/cmd/lakefs/cmd/run.go
@@ -123,7 +123,7 @@ var runCmd = &cobra.Command{
 			os.Exit(1)
 		}
 		if !allowForeign {
-			checkForeignRepos(ctx, logger, authMetadataManager, blockStore, c)
+			checkRepos(ctx, logger, authMetadataManager, blockStore, c)
 		}
 
 		// update health info with installation ID
@@ -207,7 +207,7 @@ var runCmd = &cobra.Command{
 // A foreign repo is a repository which namespace doesn't match the current block adapter.
 // A foreign repo might exists if the lakeFS instance configuration changed after a repository was
 // already created. The behaviour of lakeFS for foreign repos is undefined and should be blocked.
-func checkForeignRepos(ctx context.Context, logger logging.Logger, authMetadataManager *auth.DBMetadataManager, blockStore block.Adapter, c *catalog.Catalog) {
+func checkRepos(ctx context.Context, logger logging.Logger, authMetadataManager *auth.DBMetadataManager, blockStore block.Adapter, c *catalog.Catalog) {
 	initialized, err := authMetadataManager.IsInitialized(ctx)
 	if err != nil {
 		logger.WithError(err).Fatal("Failed to check if lakeFS is initialized")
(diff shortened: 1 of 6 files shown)


pkg/block/namespace.go
```diff
@@ -89,10 +89,9 @@ func GetStorageType(namespaceURL *url.URL) (StorageType, error) {
...
 
 func formatPathWithNamespace(namespacePath, keyPath string) string {
-	namespacePath = strings.TrimSuffix(namespacePath, "/")
-	keyPath = strings.TrimPrefix(keyPath, "/")
+	namespacePath = strings.Trim(namespacePath, "/")
 	if len(namespacePath) == 0 {
-		return keyPath
+		return strings.TrimPrefix(keyPath, "/")
 	}
 	return namespacePath + "/" + keyPath
```
nopcoder, line 96: Where is the // bug in the previous implementation? it makes sure that there will be no '/' before that path and just one '/' when concatenating the storage namespace.

In the new implementation if keyPath will have a '/' prefix it will contact '//' in case the namespace is not empty.

itaiad200: > In the new implementation if keyPath will have a '/' prefix it will contact '//' in case the namespace is not empty.

Which is right. If a key starts with a slash, you expect to see it as part of the path. However, if it doesn't, you don't expect an additional slash to be added from the `namespacePath`.

Look at the tests, I think the examples there explains it the best.

itaiad200: Oh and the `namespacePath` starts with a slash when it's not empty, therefore you need to trim both prefix and suffix.

cmd/lakefs/cmd/run.go
```diff
@@ -204,10 +204,8 @@ var runCmd = &cobra.Command{
 	},
 }
 
-// A foreign repo is a repository which namespace doesn't match the current block adapter.
-// A foreign repo might exists if the lakeFS instance configuration changed after a repository was
-// already created. The behaviour of lakeFS for foreign repos is undefined and should be blocked.
-func checkForeignRepos(ctx context.Context, logger logging.Logger, authMetadataManager *auth.DBMetadataManager, blockStore block.Adapter, c *catalog.Catalog) {
+// checkRepos iterates on all repos and validates that their settings is correct.
```
arielshaqed, line 207:
```suggestion
// checkRepos iterates on all repos and validates that their settings are correct.
```
or
```suggestion
// checkRepos iterates on all repos and validates their settings.
```

cmd/lakefs/cmd/run.go
```diff
@@ -240,17 +238,46 @@ func checkForeignRepos(ctx context.Context, logger logging.Logger, authMetadataM
...
+	const dummyFile = "dummy"
+	if _, err := adapter.Get(ctx, block.ObjectPointer{
+		StorageNamespace: repo.StorageNamespace,
+		Identifier:       dummyFile,
+	}, -1); err != nil {
+		logger.WithFields(logging.Fields{
+			"path":              dummyFile,
+			"storage_namespace": repo.StorageNamespace,
+		}).Fatal("Can't find dummy file in storage namespace, did you run the migration? " +
+			"(http://docs.lakefs.io/reference/upgrade.html#data-migration-for-version-v0490)")
```
arielshaqed, line 266: This path should change.

docs/reference/upgrade.md
```diff
@@ -66,3 +66,62 @@ To do so, add the following to your configuration (relevant **only** for 0.30.0)
 cataloger:
   type: rocks
 ```
+
+## Data Migration for Version v0.49.0
+
+We discovered a bug in the way lakeFS is storing objects in the underlying object store.
+It affects azure and gcp repositories and only some of those.
```
arielshaqed, line 73:
```suggestion
It affects  only repositories on Azure and GCP, and not all of these.
```

docs/reference/upgrade.md
```diff
@@ -66,3 +66,58 @@ To do so, add the following to your configuration (relevant **only** for 0.30.0)
...
+
+First, you need to acquire an [Account SAS](https://docs.microsoft.com/en-us/azure/storage/common/storage-sas-overview#account-sas).
+Using the Azure CLI:
+```shell
+az storage container generate-sas \
+    --account-name <ACCOUNT> \
+    --name <CONTAINER> \
+    --permissions cdrw \
+    --auth-mode key \
+    --expiry 2021-09-05
```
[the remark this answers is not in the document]
arielshaqed: Nonetheless, this expiry has elapsed.  Options:
1. Point it out and ask people to fill in a good date.
1. Set a date a month into the future.
1. Use GNU date (so bad news on Windows) and suggest ``--expiry `date -I --date '+1 day'` `` (_should_ also work on MacOS date, it tries to be GNUish, but check that first...).

docs/reference/upgrade.md
```diff
@@ -66,3 +66,62 @@ To do so, add the following to your configuration (relevant **only** for 0.30.0)
 cataloger:
   type: rocks
 ```
+
+## Data Migration for Version v0.49.0
```
arielshaqed, line 70: Reminder: as discussed F2F I had to use up this minor version on Thursday.  You'll have to bump in

arielshaqed (approved):
Thanks!

Please make sure also to add a `CHANGELOG.md` entry, some people may start from there.

docs/reference/upgrade.md
```diff
@@ -66,3 +66,62 @@ To do so, add the following to your configuration (relevant **only** for 0.30.0)
...
 ```
+
+## Data Migration for Version v0.50.0
+
+We discovered a bug in the way lakeFS is storing objects in the underlying object store.
+It affects only repositories on Azure and GCP, and not all of these.
+[Issue #2397](https://github.com/treeverse/lakeFS/issues/2397#issuecomment-908397229) describes the repository storage namespaces patterns 
+which are affected by this bug.
+
+In order to upgrade to any version which higher (or equal) than v0.50.0, you must follow these steps:
```
johnnyaug, line 77:
```suggestion
When first updating to a version greater or equal to v0.50.0, you must follow these steps:
```

docs/reference/upgrade.md
```diff
@@ -66,3 +66,62 @@ To do so, add the following to your configuration (relevant **only** for 0.30.0)
 cataloger:
   type: rocks
 ```
+
+## Data Migration for Version v0.50.0
+
+We discovered a bug in the way lakeFS is storing objects in the underlying object store.
+It affects only repositories on Azure and GCP, and not all of these.
```
johnnyaug, line 73:
```suggestion
It affects only some repositories on Azure and GCP. Repositories in S3 were not affected.
```

docs/reference/upgrade.md
```diff
@@ -66,3 +66,62 @@ To do so, add the following to your configuration (relevant **only** for 0.30.0)
...
+
+In order to upgrade to any version which higher (or equal) than v0.50.0, you must follow these steps:
+1. Stop lakeFS.
+1. Perform a data-migration (details below)
+1. Start lakeFS with the new version.
+1. After a successful run of the new version, and after validating the objects are accessible, you can delete the old data prefix.
+
+### Data migration
+
+The following patterns have been impacted by the bug:
```
johnnyaug, line 89:
```suggestion
To determine if your repositories were affected by this bug, take a look at their _storage namespace_. You can find it in the repository settings page in the UI, or using `lakectl repo list`.

The following storage namespace patterns have been impacted by the bug:
```

docs/reference/upgrade.md
```diff
@@ -66,3 +66,62 @@ To do so, add the following to your configuration (relevant **only** for 0.30.0)
...
+| azure | https://account.blob.core.windows.net/containerid         | https://account.blob.core.windows.net/containerid//*       | https://account.blob.core.windows.net/containerid/*        |
+| azure | https://account.blob.core.windows.net/containerid/        | https://account.blob.core.windows.net/containerid//*       | https://account.blob.core.windows.net/containerid/*        |
+| azure | https://account.blob.core.windows.net/containerid/prefix/ | https://account.blob.core.windows.net/containerid/prefix// | https://account.blob.core.windows.net/containerid/prefix/* |
+
+#### Migrating Google Storage data with gsutil
+
+[gsutil](https://cloud.google.com/storage/docs/gsutil) is a Python application that lets you access Cloud Storage from the command line.
+We can use it for copying the data between the prefixes in the Google bucket, and later on removing it.
+
+Copy the data with:
```
johnnyaug, line 110:
```suggestion
For every affected repository, copy its data with:
```

docs/reference/upgrade.md
```diff
@@ -66,3 +66,62 @@ To do so, add the following to your configuration (relevant **only** for 0.30.0)
...
+```shell
+gsutil -m cp -r gs://<BUCKET>//<PREFIX>/ gs://<BUCKET>/
+```
+
```
johnnyaug, line 114:
```suggestion
Note the double slash after the bucket name.
```

docs/reference/upgrade.md
```diff
@@ -66,3 +66,76 @@ To do so, add the following to your configuration (relevant **only** for 0.30.0)
...
+    --account-name <ACCOUNT> \
+    --name <CONTAINER> \
+    --permissions cdrw \
+    --auth-mode key \
+    --expiry 2021-12-31
+```
+
+With the resulted SAS, use AzCopy to copy the files:
+```shell
+azcopy copy \
```
johnnyaug, line 137: What is the command for when there is no prefix?

guy-har (approved):

PR #2438, opened 2021-09-02
Change refs-load to refs-restore in refs-restore example

guy-har:


pkg/api/controller.go
```diff
@@ -2127,8 +2127,11 @@ func (c *Controller) RestoreRefs(w http.ResponseWriter, r *http.Request, body Re
 	}
 
 	// ensure no refs currently found
-	_, _, err = c.Catalog.ListCommits(ctx, repo.Name, repo.DefaultBranch, "", 1)
-	if !errors.Is(err, graveler.ErrNotFound) {
+	commits, _, err := c.Catalog.ListCommits(ctx, repo.Name, repo.DefaultBranch, "", 2)
+	if handleAPIError(w, err) {
+		return
+	}
+	if len(commits) > 1 {
```
johnnyaug, line 2134: Why would a bare repository have a commit?

guy-har: @johnnyaug, The `repository created` commit

johnnyaug: Maybe it's not documented well enough, but refs-restore should be used with _bare_ repositories, which are create using `lakectl repo create-bare`. This way you restore the repository to the exact state it was, including commit IDs.

guy-har: I see.
It is documented, I missed that, closing this PR. 
Thanks

PR #2439, opened 2021-09-02
Fix ARNs parsing confusion when the account-ID field contained a slash or the resource-ID a colon.

arielshaqed:


pkg/auth/arn_test.go
```diff
@@ -38,12 +32,18 @@ func TestParseARN(t *testing.T) {
 			Region:     "",
 			AccountID:  "b",
 			ResourceID: "myrepo"}},
-		{Input: "arn:lakefs:repos::/myrepo", Arn: auth.Arn{
+		{Input: "arn:lakefs:repos:::myrepo", Arn: auth.Arn{
 			Partition:  "lakefs",
 			Service:    "repos",
 			Region:     "",
 			AccountID:  "",
 			ResourceID: "myrepo"}},
```
itaiad200, line 40: Shouldn't this be:
```suggestion
			ResourceID: ":myrepo"}},
```

ozkatz (approved):
Thanks, LGTM!

pkg/auth/arn_test.go
```diff
@@ -82,8 +82,7 @@ func TestArnMatch(t *testing.T) {
 		{"arn:lakefs:repos::b:myrepo", "arn:lakefs:repos::b:myrepo", true},
 		{"arn:lakefs:repos::b:*", "arn:lakefs:repos::b:myrepo", true},
 		{"arn:lakefs:repos::b:myrepo", "arn:lakefs:repos::b:*", false},
-		{"arn:lakefs:repos::b/myrepo", "arn:lakefs:repos::b/*", false},
-		{"arn:lakefs:repos::b/*", "arn:lakefs:repos::b/myrepo", true},
```
itaiad200, line 86: Can you fix these examples instead of deleting them?

itaiad200 (approved):
LGTM, 2 minor comments

pkg/auth/arn_test.go
arielshaqed: Sorry, but again no: these test the previous behaviour of allowing a `/` to separate account-ID from resource-ID.  This is simply no longer allowed.

arielshaqed (commented):
Thanks for a very speedy turnaround!

I shall rephrase the CHANGELOG entry to allow users who might rely on the existing behaviour to understand that it is now gone -- and pull.

arielshaqed:
> Thanks for a very speedy turnaround!
> 
> I shall rephrase the CHANGELOG entry to allow users who might rely on the existing behaviour to understand that it is now gone -- and pull.

Actually it's about as clear as can be.  And I doubt anyone is using it, given that there are currently no useful account IDs in lakeFS ARNs.

Issue #2406, opened 2021-08-25
[discovery] Repository-level Settings in Storage


johnnyaug:
Design and implement a layer to manage JSON configurations at the repository level.
The settings will be saved in the underlying storage.

Will be used by GC, object lifecycle (#2365) and branch protection rules (#2369).

PR #2442, opened 2021-09-02
Feature/repo level settings

johnnyaug:
Implements #2406.

Utilize the underlying storage to save repository-level settings.
Settings are saved under keys and they can be any `proto.Message`.

The settings are cached with an expiry time of 1 second, so when requesting a setting, the result is only eventually consistent.


arielshaqed (commented):
Good stuff!

I would be happier with some more clarity about the following points.  It might be possible to resolve them in package-level or class-level godoc, about how application code should use  settings from this library.

* **Efficiency**.  Document that reading is slow.
* **Consistency guarantees**.  I believe that with this implementation, callers must read repository setting _only once_ per operation.  A second read for the same operation may return different results.  (E.g. in a Spark job that needs repo configuration, the driver should read relevant configuration once and broadcast it to workers performing steps.  It is unsafe to read the configuration separately from each worker.)
* **Observability.** Operations using repo configuration should dump the configuration file (unparsed?) and/or the relevant extract parameters on start.  As an operator, particularly when trying to debug why something happened, I need to know what parameters were used on that particular operation.

pkg/graveler/settings/manager.go
```diff
@@ -0,0 +1,89 @@
...
+	repo, err := m.refManager.GetRepository(ctx, repositoryID)
+	if err != nil {
+		return err
+	}
+	objectPointer := block.ObjectPointer{
+		StorageNamespace: string(repo.StorageNamespace),
+		Identifier:       fmt.Sprintf(settingsSuffixTemplate, m.committedBlockStoragePrefix, key),
+		IdentifierType:   block.IdentifierTypeRelative,
+	}
+	reader, err := m.blockAdapter.Get(ctx, objectPointer, -1)
```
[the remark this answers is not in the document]
itaiad200: It's important to note the pyramid is skipped because objects are mutable in this implementation.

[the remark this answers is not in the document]
itaiad200: Which also makes me wonder - if branch staging locking is done using repo-level settings, wouldn't that mean that every upload to lakeFS requires an additional query to the object store?

pkg/graveler/settings/manager.go
```diff
@@ -0,0 +1,125 @@
...
+	messageBytes, err := io.ReadAll(reader)
+	if err != nil {
+		return err
+	}
+	err = proto.Unmarshal(messageBytes, message)
+	if err != nil {
+		return err
+	}
+	logger := logging.FromContext(ctx)
+	if logger.IsTracing() {
```
itaiad200, line 93: Why this check?

pkg/graveler/settings/manager.go
```diff
@@ -0,0 +1,125 @@
...
+	err = m.blockAdapter.Put(ctx, block.ObjectPointer{
+		StorageNamespace: string(repo.StorageNamespace),
+		Identifier:       fmt.Sprintf(settingsSuffixTemplate, m.committedBlockStoragePrefix, key),
+		IdentifierType:   block.IdentifierTypeRelative,
+	}, int64(len(messageBytes)), bytes.NewReader(messageBytes), block.PutOpts{})
+	if err != nil {
+		return err
+	}
+	logger := logging.FromContext(ctx)
+	if logger.IsTracing() {
```
itaiad200, line 52: Why checking? Isn't this supported out of the box when you use TRACE?

pkg/graveler/settings/manager.go
```diff
@@ -0,0 +1,125 @@
...
+	if err != nil {
+		return err
+	}
+	messageBytes, err := proto.Marshal(message)
+	if err != nil {
+		return err
+	}
+	err = m.blockAdapter.Put(ctx, block.ObjectPointer{
+		StorageNamespace: string(repo.StorageNamespace),
+		Identifier:       fmt.Sprintf(settingsSuffixTemplate, m.committedBlockStoragePrefix, key),
```
itaiad200, line 45: Identifier starts with a slash? Not again :scream:

pkg/graveler/settings/manager.go
```diff
@@ -0,0 +1,125 @@
...
+	repo, err := m.refManager.GetRepository(ctx, repositoryID)
+	if err != nil {
+		return err
+	}
+	_, err = m.branchLock.MetadataUpdater(ctx, repositoryID, repo.DefaultBranchID, func() (interface{}, error) {
+		err = m.Get(ctx, repositoryID, key, message)
+		if err != nil {
+			return nil, err
+		}
+		update()
```
itaiad200, line 117: Shouldn't update get the read value?

pkg/graveler/settings/manager.go
```diff
@@ -0,0 +1,125 @@
...
+			Trace("got repository-level setting")
+	}
+	return nil
+}
+
+// UpdateWithLock atomically performs the following sequence:
+// 1. Fetch the configuration under the given repository and key, and place it in message.
+// 2. Call the given update function (which should presumably update the message).
+// 3. Persist message to the store.
+func (m *Manager) UpdateWithLock(ctx context.Context, repositoryID graveler.RepositoryID, key string, message proto.Message, update func()) error {
```
itaiad200, line 107: Getting the settings while locking is not very useful if you act upon the settings after releasing the lock (settings might already changed..).

Using the `update` for doing the operation you wanted is one way of solving it, but it leads to some limitations, like not using the branch locker. Aside from the abstraction leak, Not sure we want to switch our entire branching technique to use this settings manager instead.

itaiad200 (commented):
Neat - using graveler building blocks look like the way to go.
Commenting and not blocking because I don't want to block forever, but I do have some concerns.

pkg/graveler/settings/manager.go
arielshaqed: Unfortunately everything up to the `Trace` call is called no matter what.  So here you would text-format `message` every time, for the sake of being able to trace it.  You end up paying a huge price just to be able to trace in very rare circumstances.

This is why C++ loggers hide the `if` in a macro, which allows them not to call any of the formatting functions (or even make a jump...).  But oh no macros are too difficult...

arielshaqed (commented):
Explain why `IsTracing` is needed.  TBH not sure why GH thought it was my turn on this review, so responding to make it Someone Else's Turn.

pkg/graveler/settings/manager.go
johnnyaug: The value is loaded into the given `message` variable, and `update` should use it.

pkg/graveler/settings/manager.go
johnnyaug: Fixed

pkg/graveler/settings/manager.go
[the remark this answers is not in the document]
johnnyaug: Added a cache

pkg/graveler/settings/manager.go
johnnyaug: The lock here is for doing the update atomically, not to use it for the settings.
Not sure I understand the comment

pkg/graveler/settings/manager.go
```diff
@@ -0,0 +1,183 @@
...
+// The result is eventually consistent: it is not guaranteed to be the most up-to-date setting. The cache expiry period is 1 second.
+func (m *Manager) Get(ctx context.Context, repositoryID graveler.RepositoryID, key string, message proto.Message) error {
+	qualifiedKey, err := json.Marshal(cacheKey{
+		RepositoryID: repositoryID,
+		Key:          key,
+	})
+	if err != nil {
+		return err
+	}
+	messageBytes, err := m.cache.GetOrSet(string(qualifiedKey), func() (v interface{}, err error) {
```
arielshaqed, line 107: Not sure I understand.  `GetOrSet` returns a value of type `interface{}`; why do we need to serialize to put something in and deserialize when we get it out?  E.g. [auth `LRUCache.GetCredential`](https://github.com/treeverse/lakeFS/blob/56ccefd3081af88450475d0992089aa6df39bf48/pkg/auth/cache.go#`L37) doesn't serialize.

pkg/graveler/settings/manager.go
arielshaqed: I also do not understand how `update` can affect the message.  We get `message` from m, then we call `update` _which cannot see `message`_, then we save the exact `message` that we just loaded.

arielshaqed (commented):
Thanks!  I am still missing some understanding of the intended use.

Also, we need to document for users of the settings library the expected behaviour of settings with regard to caching.

pkg/graveler/settings/manager.go
```diff
@@ -0,0 +1,183 @@
...
+			Trace("saving repository-level setting")
+	}
+	return nil
+}
+
+// Get fetches the setting under the given repository and key.
+// The result is placed in message.
+// The result is eventually consistent: it is not guaranteed to be the most up-to-date setting. The cache expiry period is 1 second.
+func (m *Manager) Get(ctx context.Context, repositoryID graveler.RepositoryID, key string, message proto.Message) error {
+	qualifiedKey, err := json.Marshal(cacheKey{
```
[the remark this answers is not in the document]
johnnyaug: Change to use the struct as the key

pkg/graveler/settings/manager.go
johnnyaug: Change to use the struct as the key

pkg/graveler/settings/manager.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+}
+
+func NewManager(refManager graveler.RefManager, branchLock graveler.BranchLocker, blockAdapter block.Adapter, committedBlockStoragePrefix string, options ...ManagerOption) *Manager {
+	m := &Manager{
+		refManager:                  refManager,
+		branchLock:                  branchLock,
+		blockAdapter:                blockAdapter,
+		committedBlockStoragePrefix: committedBlockStoragePrefix,
+		cache: cache.NewCache(cacheSize, defaultCacheExpiry, func() time.Duration {
```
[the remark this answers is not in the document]
johnnyaug: Done

pkg/cache/cache.go
```diff
@@ -21,6 +22,10 @@ type GetSetCache struct {
 	baseExpiry   time.Duration
 }
 
+func (c *GetSetCache) SetExpiry(baseExpiry time.Duration) {
```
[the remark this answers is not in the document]
johnnyaug: Done

pkg/graveler/settings/manager.go
```diff
@@ -59,49 +89,106 @@ func (m *Manager) Save(ctx context.Context, repositoryID graveler.RepositoryID,
...
+				entries: make(map[string]*entry),
+			}
+			m.cache.caches[repositoryID] = rCache
+		}
+		m.cache.lock.Unlock()
+	}
+	return rCache
+}
+
+func logSetting(logger logging.Logger, repositoryID graveler.RepositoryID, key string, message proto.Message) {
```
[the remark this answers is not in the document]
johnnyaug: Renamed to logFetchedSetting

pkg/graveler/settings/manager.go
```diff
@@ -59,49 +89,106 @@ func (m *Manager) Save(ctx context.Context, repositoryID graveler.RepositoryID,
...
+				return nil, err
+			}
+			err = proto.Unmarshal(messageBytes, message)
+			if err != nil {
+				return nil, err
+			}
+			logSetting(logging.FromContext(ctx), repositoryID, key, message)
+			e = &entry{
+				message: message,
+				expiry:  m.timeGetter().Add(1 * time.Second),
```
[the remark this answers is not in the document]
johnnyaug: Removed

pkg/graveler/settings/manager.go
talSofer: Can use the logSetting function you defined below

pkg/graveler/settings/manager.go
```diff
@@ -0,0 +1,179 @@
...
+		messageBytes, err := m.getFromStore(ctx, repositoryID, key)
+		if err == nil {
+			err = proto.Unmarshal(messageBytes, message)
+			if err != nil {
+				return nil, err
+			}
+		} else if !errors.Is(err, graveler.ErrNotFound) {
+			return nil, err
+		}
+		logFetchedSetting(logging.FromContext(ctx), repositoryID, key, message)
```
talSofer, line 160: Instead of renaming the logSetting method to logFetchedSetting, consider adding an action verb parameter to the method. It would be clearer if the logs report that setting X was updated, and setting Y fetched, added etc.

pkg/logging/logger.go
```diff
@@ -235,8 +235,8 @@ func (l *logrusEntryWrapper) Panicf(format string, args ...interface{}) {
 	l.e.Panicf(format, args...)
 }
 
-func (*logrusEntryWrapper) IsTracing() bool {
-	return logrus.IsLevelEnabled(logrus.TraceLevel)
+func (l *logrusEntryWrapper) IsTracing() bool {
+	return l.e.Logger.Level >= logrus.TraceLevel
```
arielshaqed, line 239: This is really weird, I know.  But d29d82e02a21f7f3ce38b016fd84a0822902e1ac changed it to have this global style.  So the existing mostly sorta kinda works, but completely relies on the _global_ `logrus` level.

Maybe do this in a separate PR, so we can have a look *just* at this?

pkg/graveler/settings/manager.go
```diff
@@ -0,0 +1,179 @@
...
+		return nil, err
+	}
+	defer func() {
+		_ = reader.Close()
+	}()
+	return io.ReadAll(reader)
+}
+
+// UpdateWithLock atomically gets a setting, performs the update function, and persists the setting to the store.
+func (m *Manager) UpdateWithLock(ctx context.Context, repositoryID graveler.RepositoryID, key string, emptyMessage proto.Message, update func(message proto.Message)) error {
```
arielshaqed, line 144:
```suggestion
func (m *Manager) UpdateWithLock(ctx context.Context, repositoryID graveler.RepositoryID, key string, messageTemplate proto.Message, update func(message proto.Message)) error {
```
might be a bit better, to explain how the default message is used.  Or document the parameter (and then the name matters less).  The thing is, the way I see it, the most important thing is _not_ that the message is empty, it's the _type_ of the message.

pkg/graveler/settings/manager.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/treeverse/lakefs/pkg/cache"
+	"github.com/treeverse/lakefs/pkg/graveler"
+	"github.com/treeverse/lakefs/pkg/logging"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+)
+
+const (
+	settingsSuffixTemplate = "%s/settings/%s.json"
+	cacheSize              = 100_000
```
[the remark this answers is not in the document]
johnnyaug: I didn't want to limit the cache by size, but it's required by the cache library that we're using. So I just gave an arbitrary, very big number (relative to the expected number of entries).

pkg/graveler/settings/manager.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// Get fetches the setting under the given repository and key, and returns the result.
+// The result is eventually consistent: it is not guaranteed to be the most up-to-date setting. The cache expiry period is 1 second.
+func (m *Manager) Get(ctx context.Context, repositoryID graveler.RepositoryID, key string, emptyMessage proto.Message) (proto.Message, error) {
+	k := cacheKey{
+		RepositoryID: repositoryID,
+		Key:          key,
+	}
+	message, err := m.cache.GetOrSet(k, func() (v interface{}, err error) {
```
arielshaqed, line 101: Using a struct as a cache key -- this is neat!  I was curious about whether this is guaranteed to work (so no action item requested, just explaining why it works for future reference by future self).  Here's the section in the spec that says it all works: https://golang.org/ref/spec#Map_types. 

>  The comparison operators `==` and `!=` must be fully defined for operands of the key type; thus the key type must not be a function, map, or slice. If the key type is an interface type, these comparison operators must be defined for the dynamic key values; failure will cause a run-time panic. 

Note that this also hints that our `cache.Cache` is not type-safe: it validates that key types are comparable _at runtime_ by panicking.  _This_ usage is, in any case, safe.

arielshaqed (approved):
Thanks for being patient with me over all the rounds!

pkg/graveler/settings/manager.go
johnnyaug: Done: added a logMessage argument

pkg/graveler/settings/manager.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/treeverse/lakefs/pkg/block/adapter"
+	"github.com/treeverse/lakefs/pkg/cache"
+	"github.com/treeverse/lakefs/pkg/graveler"
+	"github.com/treeverse/lakefs/pkg/logging"
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+)
+
+const (
+	settingsSuffixTemplate = "%s/settings/%s.json"
```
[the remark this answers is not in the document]
johnnyaug: Renamed and added comment

talSofer (commented):
Very cool!

pkg/graveler/settings/manager_test.go
```diff
@@ -0,0 +1,156 @@
...
+	testutil.Must(t, err)
+
+	// the result should be cached and we should get the first settings:
+	gotSettings, err = m.Get(ctx, "example-repo", "settingKey", emptySettings)
+	testutil.Must(t, err)
+	if diff := deep.Equal(firstSettings, gotSettings); diff != nil {
+		t.Fatal("got unexpected settings:", diff)
+	}
+	// after sleeping, we should get the new settings:
+	time.Sleep(100 * time.Millisecond)
```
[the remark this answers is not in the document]
johnnyaug: Don't want the test to take 1 second, so changed the default expiry to 50ms

pkg/graveler/settings/manager_test.go
```diff
@@ -0,0 +1,156 @@
...
+	bytes, err := ioutil.ReadAll(reader)
+	testutil.Must(t, err)
+	gotSettings := &ExampleSettings{}
+	testutil.Must(t, proto.Unmarshal(bytes, gotSettings))
+	if diff := deep.Equal(expectedSettings, gotSettings); diff != nil {
+		t.Fatal("got unexpected settings:", diff)
+	}
+}
+
+func TestEmpty(t *testing.T) {
```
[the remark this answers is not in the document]
johnnyaug: Added comments

pkg/graveler/settings/manager_test.go
```diff
@@ -0,0 +1,156 @@
...
+	gotSettings, err := m.Get(ctx, "example-repo", "settingKey", emptySettings)
+	testutil.Must(t, err)
+	if diff := deep.Equal(firstSettings, gotSettings); diff != nil {
+		t.Fatal("got unexpected settings:", diff)
+	}
+	secondSettings := &ExampleSettings{ExampleInt: 15, ExampleStr: "hi", ExampleMap: map[string]int32{"boo": 16}}
+	err = m.Save(ctx, "example-repo", "settingKey", secondSettings)
+	testutil.Must(t, err)
+
+	// the result should be cached and we should get the first settings:
```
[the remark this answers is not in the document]
johnnyaug: Great point. Done.

talSofer (approved):
Looks good! thanks!

PR #2443, opened 2021-09-02
Fix generated Java API client

arielshaqed:
Had to delete the old one and re(de?)generate a new one: a new
_incompatible_ API was generated to support the added _optional_ argument
`user_metadata`, but our Makefile failed to update the matching test.

### DIFF
--- a/clients/java/.project
+++ b/clients/java/.project
@@ -22,7 +22,7 @@
 	</natures>
 	<filteredResources>
 		<filter>
-			<id>1626737428619</id>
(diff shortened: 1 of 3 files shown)


ozkatz (approved):

arielshaqed:
Thanks!

Pulling.  We shall also have to bump a minor version number, it turns out that an innocuous added optional field in swagger.yml makes OpenAPI create an *incompatible* generated client API.

PR #2448, opened 2021-09-05
Remove MatchSimple entirely

arielshaqed:
Apparently inherited as ASL-2 code from minio.  The behaviour of `?` in
`MatchSimple` is not documented, and unclear.  Luckily for us we do not
use it anywhere.

Just remove it.


nopcoder (approved):
less code, less bugs

arielshaqed:
Thanks!

Issue #2447, opened 2021-09-05
RClone requires S3 setting no-check-bucket


arielshaqed:
Pointed out by @tvanderwal-lmco as part of #2445.

When rclone does pretty much anything, if the destination does not exist then it tries to create it.  Unfortunately it uses what it considers the S3 equivalent of `mkdir -p`.  Even worse, it (always!) tries to create the bucket.  Document that user needs to set `no_check_bucket = true` (or set the `--s3-no-check-bucket` flag) in order to work around this.

The underlying issue is that S3 will return a 409 (Conflict) when trying to recreate an already-existing bucket.  We return 405 (Method Not Allowed), and rclone gives up.

PR #2450, opened 2021-09-05
Rclone: Document setting `no_check_bucket`

arielshaqed:
Otherwise Rclone tries to create a repo when the _destination_ does not
exist... but the S3 gateway does not return the correct failure value for
creating an already-existing repo.


arielshaqed:
Relevant to #2447.  Not a complete fix -- while setting this is a great idea on _any_ S3-style store, actually creating an existing bucket/repo should fail with 409 and allow Rclone to proceed.

itaiad200 (approved):

arielshaqed:
Thanks!

PR #2451, opened 2021-09-05
[S3 gateway] Return HTTP 409 (Conflict) when creating existing repo

### LINKED ISSUE #2447: RClone requires S3 setting no-check-bucket

arielshaqed:
When copying to a new key, Rclone likes to try to create _all_ the S3 path
down (it seems to think it's on a POSIXy filesystem so it `mkdir -p`s).  A
repository cannot be created by the gateway -- there is not enough info on
the request -- but if we return a 409 (conflict) then Rclone continues and
succeeds.

There is no S3 gateway test, so tested manually using Rclone.

Fixes #2447.

### DIFF
--- a/pkg/gateway/operations/putbucket.go
+++ b/pkg/gateway/operations/putbucket.go
@@ -15,7 +15,10 @@ type PutBucket struct{}
 func (controller *PutBucket) RequiredPermissions(_ *http.Request, repoID string) ([]permissions.Permission, error) {
 	return []permissions.Permission{
 		{
-			Action:   permissions.ReadRepositoryAction,
+			// Mimic S3, which requires s3:CreateBucket to call
+			// create-bucket, even if we only want to receive
+			// 409.
+			Action:   permissions.CreateRepositoryAction,
 			Resource: permissions.RepoArn(repoID),
 		},
 	}, nil
--- a/pkg/gateway/operations/putbucket.go
+++ b/pkg/gateway/operations/putbucket.go
@@ -1,0 +1,32 @@
+package operations
+
+import (
+	"net/http"
+
+	gatewayerrors "github.com/treeverse/lakefs/pkg/gateway/errors"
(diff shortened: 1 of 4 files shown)


pkg/gateway/operations/putbucket.go
```diff
@@ -0,0 +1,32 @@
...
+
+// PutBucket handles S3 Create Bucket operations.  It does *not* actually
+// create new repos (there is not enough information in the S3 request to
+// create a new repo), but *does* detect whether the repo already exists.
+type PutBucket struct{}
+
+func (controller *PutBucket) RequiredPermissions(_ *http.Request, repoID string) ([]permissions.Permission, error) {
+	return []permissions.Permission{
+		{
+			Action:   permissions.ReadRepositoryAction,
```
guy-har, line 18: use `CreateRepositoryAction`

guy-har (approved):
Looks good
Thanks!

pkg/gateway/operations/putbucket.go
arielshaqed: Yikes!  You're right, S3 only accepts a single permission.  So if I don't have `CreateBucket` permission on S3 and I try to `create-bucket`, I get 403 (AccessDenied) rather than 409 (BucketAlreadyOwnedByYou).

Switching to match. :shrug:

arielshaqed:
Thanks!

PR #2456, opened 2021-09-06
fix multipart complete location url

nopcoder:
Based on how we extract the repository, format Location URL on complete multi-part response.
The docs: "Location - identifies the newly created object."
But it looks like s3 enables an http(s) interface to access the object which in this case can be valid only though our openapi endpoint.
Alternative implementaion for the above usecase can be found here - #2457


pkg/gateway/handler.go
```diff
@@ -181,9 +182,11 @@ func RepoOperationHandler(sc *ServerContext, handler operations.RepoOperationHan
 		repoOperation := &operations.RepoOperation{
 			AuthorizedOperation: authOp,
 			Repository:          repo,
+			MatchedHost:         matchedHost,
 		}
 		req = req.WithContext(logging.AddFields(ctx, logging.Fields{
-			"repository": repo.Name,
+			"repository":   repo.Name,
+			"matched_host": matchedHost,
```
arielshaqed, line 189: (Not in this PR, but... we really need to document the added context fields somehow and somewhere.Maybe we should make the key a named string type,and use constants.

arielshaqed (approved):
Neat!

A more architectural question: could we use a relative Location header to avoid having to know the name the client used?

pkg/gateway/handler.go
nopcoder: true, adding I'll describe it in an issue - at least the one we pass through the context, the keys can be part of our logging package is it is a wrapper around logrus

nopcoder:
> Neat!
> 
> A more architectural question: could we use a relative Location header to avoid having to know the name the client used?

All the docs and tests I made return a URI of the object - not a relative one. Tested on MinIO too - the same.
Passing the right headers to the request should be able to perform an s3 get with this address.
The only difference I found is that the default error for minio will be missing signature and for lakefs, because of the new addressing and required headers for s3 gateway, we will return the main web UI page.

PR #2457, opened 2021-09-07
multipart upload location url using our api endpoint - alt

nopcoder:
Alternative fix to s3 multipart upload complete location URI point to openapi endpoint - will be working http endpoint

### DIFF
--- a/pkg/gateway/operations/postobject.go
+++ b/pkg/gateway/operations/postobject.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/gateway/operations/postobject.go
```diff
@@ -110,7 +111,8 @@ func (controller *PostObject) HandleCompleteMultipartUpload(w http.ResponseWrite
 	}
 
 	scheme := httputil.RequestScheme(req)
-	location := fmt.Sprintf("%s://%s/%s/%s/%s", scheme, o.FQDN, o.Repository.Name, o.Reference, o.Path)
+	location := fmt.Sprintf("%s://%s/api/v1/repositories/%s/refs/%s/objects?path=%s",
+		scheme, req.Host, url.PathEscape(o.Repository.Name), url.PathEscape(o.Reference), url.QueryEscape(o.Path))
```
itaiad200, line 115: Isn't this a breaking change?

itaiad200 (changes requested):
A possible breaking change in the location of the completed object. Doesn't seem like a show stopper but we should at least add it to the CHANGELOG.

arielshaqed (approved):
This implementation might be cleaner than #2456 -- it is certainly shorter!  The same question remains, however: are we allowed to return a relative URL in `Location`?  This would (always!) simplify every part of the forwarding.

E.g. the existing code has another bug: if using TLS offloading, the `scheme` will be HTTP but the client still needs `https`.  Or really, proxying/forwarding headers are really great and everything, but there are _a lot_ of them, and relative URLs allow us to ignore their complexities.

pkg/gateway/operations/postobject.go
nopcoder: true, adding a bigger comment at the end of the pr why this alt. solution is not good.

nopcoder:
yes, this implementation is a shorter alternative, but introduce a new problem as the return URL will be valid as HTTP download through our open-api and not work as we don't know the published name of our server just the s3 hostname used to access the multi-part request.

nopcoder:
> This implementation might be cleaner than #2456 -- it is certainly shorter! The same question remains, however: are we allowed to return a relative URL in `Location`? This would (always!) simplify every part of the forwarding.
> 
> E.g. the existing code has another bug: if using TLS offloading, the `scheme` will be HTTP but the client still needs `https`. Or really, proxying/forwarding headers are really great and everything, but there are _a lot_ of them, and relative URLs allow us to ignore their complexities.

The docs, examples and the return value from AWS returns URI.  The s3 client itself doesn't use the location of the new object and I didn't manage to access the object URI as it will require credentials (it may be a valid case if the bucket has public access).
So I'll be keeping the full URI and the first implementation.

nopcoder:
keeping the first implementation pr #2456 as it doesn't have the domain name issue.

PR #2466, opened 2021-09-13
Proposal for running lakeFS on top of a key/value store

ozkatz:


design/metadata_kv/index.md
```diff
@@ -0,0 +1,187 @@
...
+- Azure Cosmos
+- Azure Blob Store
+- Google BigTable
+- Google Spanner
+- Google Cloud Storage
+- HBase
+- Cassandra (or compatible systems such as ScyllaDB)
+- Raft (embedded, or server implementations such as Consul/ETCd)
+- Persistent Redis (Redis Enterprise, AWS MemoryDB)
+- Simple in-memory tree (for simplifying and speeding up tests?)
```
arielshaqed, line 69: Can we include a list (or multiple lists) of required primitives for the KV store?

design/metadata_kv/index.md
```diff
@@ -0,0 +1,187 @@
...
+1. Reliance on write [idempotency](https://en.wikipedia.org/wiki/Idempotence) provided by Graveler (i.e., Writing the same exact entry, with the same identity - will not appear as a change)
+
+This is what the proposed implementation will look like:
+
+#### Committer flow
+
+We add an additional field to each `Branch` object: In addition to the existing `staging_token`, we add an array of strings named `sealed_tokens`.
+
+1. get branch, find current `staging_token`
+1. use `SetIf()` to update the branch (if not modified by another process): push existing `staging_token` into `sealed_tokens`, set new uuid as `staging_token`.
```
arielshaqed, line 147: IIUC this is `SetIf` the (existing, pushed onto `staging_token`) `staging_token` is unchanged?

design/metadata_kv/index.md
```diff
@@ -0,0 +1,187 @@
...
+1. take the list of sealed tokens, and using the [`CombinedIterator()`](https://github.com/treeverse/lakeFS/blob/master/pkg/graveler/combined_iterator.go#L11), turn them into a single iterator to be applied on top of the existing commit
+1. Once the commit has been persisted (metaranges and ranges stored in object store, commit itself stored to KV using `Set()`), perform another `SetIf()` that updates the branch again: replacing its commit ID with the new value, and clearing `sealed_tokens`, as these have materialized into the new commit.
+1. If `SetIf()` fails, this means another commit is happenning/has happened on the same branch. Can either retry or let the caller know.
```
arielshaqed, line 150:
```suggestion
1. If `SetIf()` fails, this means another commit is happening/has happened on the same branch. Can either retry or let the caller know.
```

arielshaqed, line 150: What is this `SetIf` on?  AFAICT just `staging_token` or just `commit_id` are not enough to provide ordering in the face of concurrent commits.  It might be if the entire value is unchanged, or possibly `staging_token` and `sealed_tokens`.

design/metadata_kv/index.md
```diff
@@ -0,0 +1,187 @@
...
+
+An important property here is delegating safety vs liveness to a single optimistic `SetIf()` operation: if a commit fails somewhere along the process, a subsequent commit would simply pick up where the failed one left off,
+adding the current staging_token into the set of changes to apply. In an environment where there aren't many concurrent commits on the same branch, and that commits mostly succeed - the size of `sealed_tokens` would be relatively small. As an optimization, compaction strategies could be added to merge tokens together, but this *might not be necessary*, at least not for use cases we're familiar with.
```
arielshaqed, line 153: I am less optimistic (sorry) here.  Suppose that we have an automated commit, once every X seconds (X ~ 1s).  Typically everything works really well and we're happy.  I would claim that this is a good way to implement streaming input: write to branch `input`, every X seconds commit and on success merge to `main`.  (It is actually hard to do something else!  E.g. writing to branch `input-NNNNNN` with a running commit counter `NNNNNN` pushes this _huge_ problem onto the writers, who now need essentially to resolve the same issue that we are resolving here.)

Now suppose that our database (or maybe that part of our object store) has a hiccup and a commit takes >X seconds.  Then `sealed_tokens` starts growing.  AFAIU this requires every concurrent commit to fail (there is no point in retrying if `sealed_tokens` has grown since the last call, it will never go back to what we expect.  So typically _only the last concurrent commit_ will succeed.  So until some time after recovery, not only do we have no commits, but also _reads get progressively slower_.

It may be OK to leave this open in the first version(s), but I do want to document this limitation while we have it.

design/metadata_kv/index.md
```diff
@@ -0,0 +1,187 @@
...
+
+![Committer Flow](./committer_flow.png)
+
+#### Writer flow
+
+1. Read the branch's existing staging token: currently, we batch requests for a short duration and amortize the cost across requests.
+1. Write to the staging token received
```
arielshaqed, line 160: Talking F2F I now understand that we store the path not so much in the object store as in the KV store.  Can we specify this here, too?

design/metadata_kv/index.md
```diff
@@ -0,0 +1,187 @@
...
+1. We now use the existing `CombinedIterator` to read through all staging tokens and underlying commit.
+
+![Reader Flow](./reader_flow.png)
+
+#### Important Note - exclusion duration
+
+It is important to understand that the current pessimistic approach locks the branch for the entire duration of the commit. 
+This takes time proportional to the amount of changes to be committted and is unbounded. All writes to a branch are blocked for that period of time.
+
+With the optimistic version, writes are retried for a bounded period of time: the time between committer's initial branch read, and the first SetIf() that replaces the `staging_token`.
```
arielshaqed, line 179:
```suggestion
With the optimistic version, writes are retried for a bounded period of time: the time between committer's initial branch read, and the first `SetIf()` that replaces the `staging_token`. 
```

arielshaqed (commented):
Neat!

design/metadata_kv/index.md
```diff
@@ -0,0 +1,187 @@
...
+    Get(key []byte) (value []byte, err error)
+    // Scan returns an iterator that scans keys in byte order, starting at or after the `start` position
+    Scan(start []byte) (iter KeyValueIterator, err error)
+    // Set stores the given value, overwriting an existing value if one exists
+    Set(key, value []byte) error
+    // Delete will delete the key/value at key, if any
+    Delete(key []byte) error
+    // SetIf returns an ErrPredicateFailed error if the valuePredicate passed
+    //  doesn't match the currently stored value.
+    SetIf(key, value, valuePredicate []byte) error
```
arielshaqed, line 46: This might allow a very powerful but complex predicate -- not clear because we don't document it anywhere.  Or is this a CAS?  Because CAS is weak, but hard to use efficiently.

Ideally we would break the predicate down into a much weaker language, allowing more efficient implementation on many KVs.  Or support CAS and document

arielshaqed (commented):
Asking for more details about `SetIf`.

design/metadata_kv/index.md
```diff
@@ -0,0 +1,187 @@
...
+1. If `SetIf()` fails, this means another commit is happenning/has happened on the same branch. Can either retry or let the caller know.
+
+An important property here is delegating safety vs liveness to a single optimistic `SetIf()` operation: if a commit fails somewhere along the process, a subsequent commit would simply pick up where the failed one left off,
+adding the current staging_token into the set of changes to apply. In an environment where there aren't many concurrent commits on the same branch, and that commits mostly succeed - the size of `sealed_tokens` would be relatively small. As an optimization, compaction strategies could be added to merge tokens together, but this *might not be necessary*, at least not for use cases we're familiar with.
+
+![Committer Flow](./committer_flow.png)
+
+#### Writer flow
+
+1. Read the branch's existing staging token: currently, we batch requests for a short duration and amortize the cost across requests.
```
johnnyaug, line 159: As far as I understand, this flow can cause a single write to appear in more than one commit. In particular, it can't be described as an idempotent operation in (3) below. Not saying this is necessarily bad but I'm finding it hard to describe it in terms of a consistency model.

In the following example, write 001 appears on commit `a`, reverted by commit `b` and then appears again as a new write on commit `c`.

Write 001 start – to staging token A
Commit `a` for token A
Write 002 start – to staging token B
Commit `b` for token B
Write 002 finalize: staging token was changed – write again to token C
Write 001 finalize: staging token was changed – write again to token C
Commit `c` for token C

johnnyaug: Update: talked f2f with @arielshaqed, who explained that this "limitation" was discussed by the team. @ozkatz can you please add an explanation about it?

arielshaqed: #2480 (created after F2F'ing with @ozkatz ) is my attempt to describe what guarantees we _do_ give.

design/metadata_kv/index.md
ozkatz: Added! thanks.

design/metadata_kv/index.md
ozkatz: Made it clear that `SetIf()` is actually a CAS operation on the entire branch (`staging_token`, `sealed_tokens` and `commit_id`).

design/metadata_kv/index.md
ozkatz: Fixed, thanks!

ozkatz: This was indeed unclear - made it explicit that `SetIf()` is actually a CAS operation on the entire branch (`staging_token`, `sealed_tokens` and `commit_id`).

design/metadata_kv/index.md
ozkatz: Added a suggestion on how to address it. If you have other strategies that could be used here, let's add them!

design/metadata_kv/index.md
ozkatz: Thanks, Fixed.

design/metadata_kv/index.md
ozkatz: This is a simple CAS. documented.

design/metadata_kv/index.md
```diff
@@ -0,0 +1,187 @@
...
+This is what the proposed implementation will look like:
+
+#### Committer flow
+
+We add an additional field to each `Branch` object: In addition to the existing `staging_token`, we add an array of strings named `sealed_tokens`.
+
+1. get branch, find current `staging_token`
+1. use `SetIf()` to update the branch (if not modified by another process): push existing `staging_token` into `sealed_tokens`, set new uuid as `staging_token`.
+1. take the list of sealed tokens, and using the [`CombinedIterator()`](https://github.com/treeverse/lakeFS/blob/master/pkg/graveler/combined_iterator.go#L11), turn them into a single iterator to be applied on top of the existing commit
+1. Once the commit has been persisted (metaranges and ranges stored in object store, commit itself stored to KV using `Set()`), perform another `SetIf()` that updates the branch again: replacing its commit ID with the new value, and clearing `sealed_tokens`, as these have materialized into the new commit.
```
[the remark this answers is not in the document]
ozkatz: That's a very good question - the answer to "when is it safe" would be "never" :)
I revised the reader flow to address this, also using optimistic concurrency control. Interested in hearing your opinion on the suggested solution (also, I'm not sure I did a good job explaining it).

design/metadata_kv/index.md
ozkatz: Great minds *do* think alike I guess. Thanks @arielshaqed for the description (should we maybe add it to this PR?)!

design/metadata_kv/index.md
arielshaqed: Thing is, we run into trouble _even if we **never** retry commits_ (so exponential backoff doesn't solve the issue)!  Every commit tries and fails, growing `sealed_tokens`.  As long as there's a nice fugue (commit starts by growing `sealed_tokens`, then runs until after the next commit starts) then no commit succeeds, and `sealed_tokens` grows, and reads take longer and longer.

I believe the only way around this will be to trim prefixes of `sealed_tokens`.  E.g. accept a commit if the end of its `sealed_tokens` overlaps that start of the current `sealed_tokens`, and trim the shared bit.

## Example

`sealed_tokens`: 1,2,3; `staging_token`: 4

Process A starts to commit tokens 1,2,3,4.

`sealed_tokens`: 1,2,3,4; `staging_token`: 5

Process B starts to commit tokens 1,2,3,4,5.

`sealed_tokens`: 1,2,3,4,5; `staging_token`: 6

Process C starts to commit tokens 1,2,3,4,5,6.

`sealed_tokens`: 1,2,3,4,5,6; `staging_token`: 7

Process A ends committing: it tries for a "fast-path commit", but since `staging_token` is no longer 5 that fails.  Instead it reads the branch (or uses the value return by Set-If, if we want to require such an interface).  Now 1,2,3,4 is a prefix of 1,2,3,4,5,6, so A tries again to commit the remaining `sealed_tokens` 5,6 and a `staging_token` of 7.  *That* works, so A succeeded!

`sealed_tokens`: 5,6; `staging_token`: 7

Process C ends committing: its fast-path commit works (`staging_token` is 7), so it drops all `sealed_tokens`.

`sealed_tokens`: (); `staging_token`: 7

Process B ends committing: its fast-path commit doesn't work, and it has an empty prefix, so it knows _it has already succeeded_.

arielshaqed (commented):
Neat!  I think you addressed a _different_ issue from the one that I had with periodic commits, so I blathered on a bit about the difference and how we could solve that one.

I do *NOT* think that this blocks an initial revision.  (However, having written it down, I now think that it is simple enough to do immediately.)

ozkatz:
> Neat! I think you addressed a _different_ issue from the one that I had with periodic commits, so I blathered on a bit about the difference and how we could solve that one.
> 
> I do _NOT_ think that this blocks an initial revision. (However, having written it down, I now think that it is simple enough to do immediately.)

@arielshaqed I like this idea a lot - and it is simple. 
Can you contribute it to this PR? Preferably as another sub-section so its independent (just thinking ahead to how we might break down this change into smaller bits if/when we decide we want to go for it)

design/metadata_kv/index.md
```diff
@@ -0,0 +1,247 @@
...
+* If a write succeeds, successive reads and listings will return the contents of either that
+  write or of some other write that did not happen before it.
+
+* A succesful read returns the contents of some write that did not happen after it.
+
+* A listing returns the contents of some writes that did not happen after it.
+
+* A successful commit holds the contents of some listing of the entire keyspace.
+
+* Mutating operations (commits and writes) may succeed or fail.  When they fail, their contents
```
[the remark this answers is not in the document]
ozkatz: I feel this item is mostly redundant since it's true for pretty much all object stores and databases that I can think of. perhaps @arielshaqed can clarify?

design/metadata_kv/index.md
```diff
@@ -0,0 +1,247 @@
...
+```
+
+It is possible to create a 2-way index for many-to-many relationships, but this is not generally required - it would be simpler to simply do a scan when needed and filter only the relevant values.
+This is because the IAM keyspace is relatively very small - it would be surprising if the biggest lakeFS installation to ever exist would contain more than, say, 50k users.
+
+Some care does need to be applied when managing these secondary indices - for example, when deleting an entity, secondary indices need to be pruned first, to avoid inconsistencies. 
+
+### Graveler Metadata - Commits, Tags, Repositories
+
+These are simpler entities - commits and tags are immutable and could potentially be stored on the object store and not in the key/value store (they are cheap either way, so it doesn't make much of a difference).
```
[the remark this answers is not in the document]
ozkatz: Fair enough. I do think they could be treated as immutable when discussing concurrency (i.e. it's ok to assume low/non contention when writing a single tag entry) - so less prone to read-modify-write issues.

design/metadata_kv/index.md
```diff
@@ -0,0 +1,247 @@
...
+    ...
+    return groupIds
+}
+
+```
+
+It is possible to create a 2-way index for many-to-many relationships, but this is not generally required - it would be simpler to simply do a scan when needed and filter only the relevant values.
+This is because the IAM keyspace is relatively very small - it would be surprising if the biggest lakeFS installation to ever exist would contain more than, say, 50k users.
+
+Some care does need to be applied when managing these secondary indices - for example, when deleting an entity, secondary indices need to be pruned first, to avoid inconsistencies.
```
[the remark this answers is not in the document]
ozkatz: I don't understand how this affects consistency though? If deleting A returned *successfully* it should appear as deleted to all observing lakeFS instances, making it safe to then delete B (unless they are configured to read from an async replica which would violate our consistency requirements anyway?)

design/metadata_kv/index.md
```diff
@@ -0,0 +1,247 @@
...
+
+To do this, we introduce a small in-memory cache (can utilize the same caching mechanism that already exists for IAM).
+Please note: this *does not violate consistency*, see the [Read flow](#reader-flow) and [Writer flow](writer-flow) below to understand how.
+
+#### Writer flow
+
+1. Read the branch's existing staging token: if branch exists in the cache, use it! Otherwise, do an amortized read (see [above](#caching-branch-pointers-and-amortized-reads)) and cache the result for a very short duration.
+1. Write to the staging token received - this is another key/value record (e.g. `"graveler/staging/${repoId}/${stagingToken}/${path}"`)
+1. Read the branch's existing staging token **again**. This is always an amortized read, not a cache read. If we get the same `staging_token` - great, no commit has *started while writing* the record, return success to the user. For a system with low contention between writes and commits, this will be the usual case.
+1. If the `staging_token` *has* changed - **retry the operation**. If the previous write made it in time to be included in the commit, we'll end up writing a record with the same identity - an idempotent operation.
```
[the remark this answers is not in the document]
itaiad200: I do think that we have a `B-A-N-A` problem here. If I try write `A` & `N` to the same path and try committing twice while doing it. The value could very well be:
1. `B` - before anything was done.
2. `A` - in the first commit.
3. `N` - in the second commit.
4. `A` - staging area after the second commit.

So from the `A` was "written twice" for a single write operation.

I think that the design assumes that no other value is being written to the same path concurrently, hence the mistake.

nopcoder (approved):

Issue #2400, opened 2021-08-24
Non-descriptive error message when trying to create a repository with a new storage type


lynnro314:
When a user runs lakeFS for the first time and creates his first repository he selects a specific storage type. 
Once lakeFS is installed, only that specific storage type is supported. 
Then, if the user attempts to create a repository with a different storage type, 
a non-descriptive error message appears:
![image](https://user-images.githubusercontent.com/64288262/130624098-0f382d54-fbcb-4c0e-897e-4ea182559b83.png)
This message should report the actual error, for example, "All repositories should have the same storage".

Steps to reproduce:
  1. Setup lakeFS or use an already existing setup
  2. Create your first repository.
  3. Try to create a new repository with a different storage namespace.

PR #2469, opened 2021-09-13
Bugfix/2400 change error message repo with new storage type

lynnro314:
Close: #2400 

When trying to create a repository with a different storage type than the one defined in the configuration, the error message was not descriptive.
I made two changes:

1. In create repo I added a verification of whether the storage type is the same as the one defined in the configuration. In this case, a descriptive error message will be printed.
2. I changed the error message in the UI to be descriptive.

### DIFF
--- a/pkg/api/auth_middleware.go
+++ b/pkg/api/auth_middleware.go
@@ -148,7 +148,7 @@ func userByAuth(ctx context.Context, logger logging.Logger, authService auth.Ser
 	}
 	user, err := authService.GetUserByID(ctx, cred.UserID)
 	if err != nil {
-		logger.WithFields(logging.Fields{"user_id": cred.UserID}).Debug("could not find user id by credentials")
+		logger.WithError(err).WithFields(logging.Fields{"user_id": cred.UserID}).Debug("could not find user id by credentials")
 		return nil, ErrAuthenticatingRequest
 	}
 	return user, nil
--- a/pkg/api/auth_middleware.go
+++ b/pkg/api/auth_middleware.go
@@ -3,7 +3,6 @@ package api
 import (
 	"context"
 	"crypto/subtle"
-	"errors"
 	"fmt"
 	"net/http"
 	"strings"
@@ -15,11 +14,6 @@ import (
 	"github.com/treeverse/lakefs/pkg/auth/model"
 	"github.com/treeverse/lakefs/pkg/logging"
 	"gopkg.in/dgrijalva/jwt-go.v3"
-)
-
-var (
-	ErrUnexpectedSigningMethod = errors.New("unexpected signing method")
-	ErrAuthenticationFailed    = errors.New("error authenticating request")
 )
 
 // extractSecurityRequirements using Swagger returns an array of security requirements set for the request.
--- a/pkg/api/auth_middleware.go
+++ b/pkg/api/auth_middleware.go
@@ -92,7 +92,7 @@ func checkSecurityRequirements(r *http.Request, securityRequirements openapi3.Se
 			default:
 				// unknown security requirement to check
 				logger.WithField("provider", provider).Error("Authentication middleware unknown security requirement provider")
-				return nil, ErrAuthenticationFailed
+				return nil, ErrAuthenticatingRequest
 			}
 			if err != nil {
 				return nil, err
@@ -114,16 +114,16 @@ func userByToken(ctx context.Context, logger logging.Logger, authService auth.Se
 		return authService.SecretStore().SharedSecret(), nil
 	})
 	if err != nil {
-		return nil, ErrAuthenticationFailed
+		return nil, ErrAuthenticatingRequest
 	}
 	claims, ok := token.Claims.(*jwt.StandardClaims)
 	if !ok || !token.Valid {
-		return nil, ErrAuthenticationFailed
+		return nil, ErrAuthenticatingRequest
 	}
 	cred, err := authService.GetCredentials(ctx, claims.Subject)
 	if err != nil {
 		logger.WithField("subject", claims.Subject).Info("could not find credentials for token")
-		return nil, ErrAuthenticationFailed
+		return nil, ErrAuthenticatingRequest
 	}
 	userData, err := authService.GetUserByID(ctx, cred.UserID)
 	if err != nil {
@@ -131,7 +131,7 @@ func userByToken(ctx context.Context, logger logging.Logger, authService auth.Se
 			"user_id": cred.UserID,
 			"subject": claims.Subject,
 		}).Debug("could not find user id by credentials")
-		return nil, ErrAuthenticationFailed
+		return nil, ErrAuthenticatingRequest
 	}
 	return userData, nil
 }
@@ -140,16 +140,16 @@ func userByAuth(ctx context.Context, logger logging.Logger, authService auth.Ser
 	cred, err := authService.GetCredentials(ctx, accessKey)
 	if err != nil {
 		logger.WithError(err).Error("failed getting credentials for key")
-		return nil, ErrAuthenticationFailed
+		return nil, ErrAuthenticatingRequest
 	}
 	if subtle.ConstantTimeCompare([]byte(secretKey), []byte(cred.SecretAccessKey)) != 1 {
 		logger.Debug("access key secret does not match")
-		return nil, ErrAuthenticationFailed
+		return nil, ErrAuthenticatingRequest
 	}
 	user, err := authService.GetUserByID(ctx, cred.UserID)
 	if err != nil {
 		logger.WithFields(logging.Fields{"user_id": cred.UserID}).Debug("could not find user id by credentials")
-		return nil, ErrAuthenticationFailed
+		return nil, ErrAuthenticatingRequest
 	}
 	return user, nil
 }
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -1109,18 +1109,18 @@ func (c *Controller) CreateRepository(w http.ResponseWriter, r *http.Request, bo
 		return
 	}
 
-	parsedNs, err := url.ParseRequestURI(body.StorageNamespace)
+	parsedNamespace, err := url.ParseRequestURI(body.StorageNamespace)
 	if err != nil {
-		writeError(w, http.StatusBadRequest, "error parsing storage namespace. Please enter valid repository url.")
+		writeError(w, http.StatusBadRequest, "error parsing storage namespace. Please enter a valid repository url.")
 		return
 	}
-	storageType, err := block.GetStorageType(parsedNs)
+	namespaceStorageType, err := block.GetStorageType(parsedNamespace)
 	if err != nil {
-		writeError(w, http.StatusBadRequest, "error creating repository: storage type is not valid.")
+		writeError(w, http.StatusBadRequest, "error creating repository: invalid storage type.")
 		return
 	}
-	if relStorageType := c.BlockAdapter.BlockstoreType(); relStorageType != storageType.BlockstoreType() {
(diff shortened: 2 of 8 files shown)


pkg/api/controller.go
```diff
@@ -1103,7 +1104,22 @@ func (c *Controller) CreateRepository(w http.ResponseWriter, r *http.Request, bo
 		return
 	}
 
-	err := ensureStorageNamespaceRW(ctx, c.BlockAdapter, body.StorageNamespace)
+	parsedNs, err := url.ParseRequestURI(body.StorageNamespace)
```
arielshaqed, line 1107: Not sure why this is not part of the ensure* call. I assume we'd want the same error detection if we added another "create repo" operation. 
If the issue is better error reporting from that call then we might want to do it there. E.g. Return an internal error code and an external message.

webui/src/lib/components/repositoryCreateForm.jsx
```diff
@@ -77,7 +78,7 @@ export const RepositoryCreateForm = ({ config, onSubmit, onCancel, error = null,
                     <Form.Control type="text" ref={storageNamespaceField} placeholder={storageNamespaceExample} onChange={checkStorageNamespaceValidity}/>
                     {!storageNamespaceValid &&
                     <Form.Text className="text-danger">
-                        Invalid Storage Namespace.
+                        {"Can only create repository with storage type: " + storageType}
```
arielshaqed, line 81: Can we use an error message from the server here?

arielshaqed (approved):
Neat, and *very* helpful to our users! 

I  think we might improve the structure of where we do the test in the API server and how much the client needs to know in order to report the error.

Also can we treat this somehow in `api_test.go`?

Thanks!

pkg/api/controller.go
talSofer: Agree with Ariel, this logic belongs to the ensureStorageNamespaceRW method because it's part of ensuring the validity of the storage namespace. Can you please move it there?

pkg/api/controller.go
```diff
@@ -1103,7 +1104,22 @@ func (c *Controller) CreateRepository(w http.ResponseWriter, r *http.Request, bo
...
+	if err != nil {
+		writeError(w, http.StatusBadRequest, "error parsing storage namespace. Please enter valid repository url.")
```
talSofer, line 1109:
```suggestion
		writeError(w, http.StatusBadRequest, "error parsing storage namespace. Please enter a valid repository url.")
```

pkg/api/controller.go
talSofer: nit; it's up to you but I prefer the name parsedNamespace for readability reasons

pkg/api/controller.go
```diff
@@ -1103,7 +1104,22 @@ func (c *Controller) CreateRepository(w http.ResponseWriter, r *http.Request, bo
...
+		return
+	}
+	storageType, err := block.GetStorageType(parsedNs)
+	if err != nil {
+		writeError(w, http.StatusBadRequest, "error creating repository: storage type is not valid.")
```
talSofer, line 1114:
```suggestion
		writeError(w, http.StatusBadRequest, "error creating repository: invalid storage type.")
```

pkg/api/controller.go
```diff
@@ -1103,7 +1104,22 @@ func (c *Controller) CreateRepository(w http.ResponseWriter, r *http.Request, bo
...
+		return
+	}
+	if relStorageType := c.BlockAdapter.BlockstoreType(); relStorageType != storageType.BlockstoreType() {
```
talSofer, line 1117: It is not clear what "rel" stands for.  can you please rename this variable? 
A suggestion: 
Use `adapterStorageType` instead of `relStorageType`, and `namespaceStorageType` instead of `storageType`

talSofer: Also, it is probably most readable to get rid of the variable and do: 
`if c.BlockAdapter.BlockstoreType() != namespaceStorageType.BlockstoreType()`

talSofer (changes requested):
A great improvement to our error reporting, thanks!

pkg/api/controller.go
lynnro314: No problem, will rename it. Also, I added the variable since I'm using it later when I'm writing the error.

pkg/api/controller.go
```diff
@@ -1108,13 +1108,13 @@ func (c *Controller) CreateRepository(w http.ResponseWriter, r *http.Request, bo
...
 
-	err := ensureStorageNamespaceRW(ctx, c.BlockAdapter, body.StorageNamespace)
+	err := ensureStorageNamespace(ctx, c.BlockAdapter, body.StorageNamespace)
 	if err != nil {
 		c.Logger.
 			WithError(err).
 			WithField("storage_namespace", body.StorageNamespace).
 			Warn("Could not access storage namespace")
-		writeError(w, http.StatusBadRequest, "error creating repository: could not access storage namespace")
+		writeError(w, http.StatusBadRequest, "error creating repository: "+err.Error())
```
arielshaqed, line 1117: Yikes... this can leak potentially sensitive information.  A good way to handle something like this is to return _two_ values (and document them!) from `ensureStorageNamespace`: the first is an `error`, the second is a `string` which will be shown to the user.  Then it is easier to ensure that nobody ever returns sensitive information from that function.

pkg/block/namespace.go
```diff
@@ -126,7 +126,7 @@ func ResolveNamespace(defaultNamespace, key string, identifierType IdentifierTyp
 func resolveFull(key string) (QualifiedKey, error) {
 	parsedKey, err := url.ParseRequestURI(key)
 	if err != nil {
-		return QualifiedKey{}, fmt.Errorf("key isn't a valid address: %w", err)
+		return QualifiedKey{}, fmt.Errorf("Could not parse URI: %w", err)
```
arielshaqed, line 129: Generally the negative is already in the base error message, so it may make sense merely to describe the failure:
```suggestion
		return QualifiedKey{}, fmt.Errorf("parse URI: %w", err)
```

webui/src/lib/components/repositoryCreateForm.jsx
arielshaqed: I believe that this is still open.

arielshaqed (changes requested):
Thanks!

Must-have change is something to make it less likely to leak sensitive information on error messages written out to clients.

arielshaqed (approved):
After talking f2f with @lynnro314 I agree that my comments about a _potential_ leak from an error message does not exist today and is anyway possible through _every other_ block adapter method.  So no justification to block on this.
Approving and we shall open a **separate** issue to return separate messages to users from _all_ API and S3 gateway calls.

pkg/api/controller_test.go
```diff
@@ -320,6 +320,25 @@ func TestController_CreateRepositoryHandler(t *testing.T) {
 			t.Fatalf("expected error creating duplicate repo")
 		}
 	})
+
+	t.Run("create repo with wrong storage type", func(t *testing.T) {
+		_, err := deps.catalog.CreateRepository(ctx, "repo3", onBlock(deps, "foo1"), "main")
+		if err != nil {
+			t.Fatal(err)
+		}
```
talSofer, line 328: This part looks redundant. You only want to create a single repo with a conflicting storage namespace (as you did in line 329)

pkg/api/controller_test.go
```diff
@@ -320,6 +320,25 @@ func TestController_CreateRepositoryHandler(t *testing.T) {
 			t.Fatalf("expected error creating duplicate repo")
 		}
 	})
+
+	t.Run("create repo with wrong storage type", func(t *testing.T) {
```
talSofer, line 324: nit; you can use the word "conflicting" instead of "wrong"?  wrong is less descriptive

talSofer (approved):
Looks good! please address the comments before you merge the code

pkg/api/controller_test.go
lynnro314: I want first to create a repo with some storage type, and then create another repo with different storage type. Sounds Ok?

talSofer: Why do you need to create the first repo? Do you mean that you are trying to test both the valid and invalid cases, or you meant something else? 

The sub-test description says "create repo with conflicting storage type", so what I expect to see is an attempt to create a repository with an invalid storage namespace that fails. 

If you are trying to test both cases I would separate their test. that is, you can define another test case with the description "create repo with a storage type" but this is already tested in line 289.

pkg/api/errors.go
```diff
@@ -0,0 +1,17 @@
+package api
+
+import (
+	"errors"
+)
+
+var (
+	// ErrEnsureStorageNamespace is an error returned when trying to create a new repository
+	// and the attempt to access the storage fails
+	ErrEnsureStorageNamespace = errors.New("failed to ensure access to the storage")
```
nopcoder, line 10: I think that we didn't fail to ensure - we failed to find a matched storage namespace.
Error like - mismatch storage namespace or something like it describes the issue.

nopcoder (changes requested):

pkg/api/controller_test.go
lynnro314: You are right, this test is not necessary so I took it down.

pkg/api/auth_middleware.go
```diff
@@ -114,24 +114,24 @@ func userByToken(ctx context.Context, logger logging.Logger, authService auth.Se
...
+		return nil, ErrAuthenticatingRequest
 	}
 	claims, ok := token.Claims.(*jwt.StandardClaims)
 	if !ok || !token.Valid {
-		return nil, ErrAuthenticationFailed
+		return nil, ErrAuthenticatingRequest
 	}
 	cred, err := authService.GetCredentials(ctx, claims.Subject)
 	if err != nil {
 		logger.WithField("subject", claims.Subject).Info("could not find credentials for token")
```
nopcoder, line 125: not your code - but can you add WithError to log the information

pkg/api/controller.go
```diff
@@ -1111,11 +1111,19 @@ func (c *Controller) CreateRepository(w http.ResponseWriter, r *http.Request, bo
...
 	err := ensureStorageNamespace(ctx, c.BlockAdapter, body.StorageNamespace)
 	if err != nil {
+		var errorMessage = "failed to access the storage"
+		if errors.Is(err, block.ErrInvalidNamespace) {
+			errorMessage = "Can only create repository with storage type: " + c.BlockAdapter.BlockstoreType()
+		}
+		var e *url.Error
+		if errors.As(err, &e) && e.Op == "parse" {
+			errorMessage = err.Error()
+		}
```
nopcoder, line 1121:
```go
		var ErrFailedToAccessStorage = " failed to access storage"
```

```suggestion
		var retErr error
		var e *url.Error
		if errors.As(err, &e) && e.Op == "parse" {
			retErr = err
		} else if errors.Is(err, block.ErrInvalidNamespace) {
			retErr = fmt.Errorf("repository storage type must match: %s", c.BlockAdapter.BlockstoreType())
		} else {
			retErr = ErrFailedToAccessStorage
		}
```

pkg/api/controller.go
```diff
@@ -1111,11 +1111,19 @@ func (c *Controller) CreateRepository(w http.ResponseWriter, r *http.Request, bo
...
 		c.Logger.
 			WithError(err).
 			WithField("storage_namespace", body.StorageNamespace).
 			Warn("Could not access storage namespace")
-		writeError(w, http.StatusBadRequest, err.Error())
+		writeError(w, http.StatusBadRequest, "error creating repository: "+errorMessage)
```
nopcoder, line 1126:
```suggestion
		writeError(w, http.StatusBadRequest, fmt.Errorf("failed to create repository: %w", retErr))
```

nopcoder (approved):
Looks good

PR #2470, opened 2021-09-14
Fix IsTracing method

johnnyaug:
Currently, `IsTracing` returns false even if lakeFS log level is set to `TRACE`.

### DIFF
--- a/pkg/logging/logger.go
+++ b/pkg/logging/logger.go
@@ -235,8 +235,8 @@ func (l *logrusEntryWrapper) Panicf(format string, args ...interface{}) {
 	l.e.Panicf(format, args...)
 }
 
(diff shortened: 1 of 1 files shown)


arielshaqed (approved):
Thanks!
Assuming this was tested.

arielshaqed:
This breaks the gateway playback test on GH actions due to creating a data race (during the initial download?!) -- see e.g. https://github.com/treeverse/lakeFS/runs/3597721059?check_suite_focus=true.  I'm rolling this back, because I've no idea how this could possibly break something like that... yet it does.

Quite possibly the bug is in the playback test, which tries to call `IsTracing` before it sets up a config, which is what sets up the logger, which is what changed.

PR #2473, opened 2021-09-14
Revert "Fix IsTracing method"

arielshaqed:
Reverts treeverse/lakeFS#2470

This breaks the gateway playback test on GH actions due to creating a data race (during the initial download?!) -- see e.g. https://github.com/treeverse/lakeFS/runs/3597721059?check_suite_focus=true.  I'm rolling this back, because I've no idea how this could possibly break something like that... yet it does.

Quite possibly the bug is in the playback test, which tries to call `IsTracing` before it sets up a config, which is what sets up the logger, which is what changed.

### DIFF
--- a/pkg/logging/logger.go
+++ b/pkg/logging/logger.go
@@ -235,8 +235,8 @@ func (l *logrusEntryWrapper) Panicf(format string, args ...interface{}) {
 	l.e.Panicf(format, args...)
 }
 
-func (l *logrusEntryWrapper) IsTracing() bool {
-	return l.e.Logger.Level >= logrus.TraceLevel
+func (*logrusEntryWrapper) IsTracing() bool {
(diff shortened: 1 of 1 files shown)


ozkatz (approved):

arielshaqed:
I'm _guessing_ that the issue may be related to the fact that logrus writes this value atomically, and we read it with a regular read.  However there may be something deeper (and worse).

arielshaqed:
@johnnyaug sorry for doing this so quickly, it's breaking CI right now and does not seem sufficiently important to keep in place.  (Worst case, we should be getting rid of the playback test, my PR #2464 even adds some direct S3 gateway testing so we're on the right track!  Once that test is gone, this PR can probably go in **as-is**...)

johnnyaug:
@arielshaqed, no worries. It's just that right now - most TRACE logs do not appear in lakeFS, so we need to address this somehow.

PR #2492, opened 2021-09-26
use constants for context logging field keys

### LINKED ISSUE #2489: Code documenting the logger context fields (log_fields)
There are a couple of cases where we pass logging fields using "Context" value called "log_fields".
In order to prevent misuse of these keys, which pass as strings, it would be better to have a constant and document each one.

nopcoder:
Close #2489


itaiad200 (approved):
LGTM

Issue #2058, opened 2021-06-04
Support ldap authentication (was: Dex (OpenID) integration) 


talSofer:
This feature was requested over slack on: 
https://lakefs.slack.com/archives/C016726JLJW/p1622700653019000?thread_ts=1622669223.018000&cid=C016726JLJW

Requirements and feature justifications mentioned in the thread:
> We use it in our clusters to allow authentication of users via our AD. Dex can then provide signed JWT tokens. https://dexidp.io/docs/using-dex/.
> So what would be nice is if LakeFS can consume these tokens so that we may have a single point of sign on.

> Dex integration is maybe one the top thing we care about and is a bit of a blocker for full adoption.

arielshaqed:
There are usually 2 parts to integrating a new auth system:
1. Integrate *authentication*. Make the login and relogin flows go through the API callback hoops and get a token. This is "just a small matter of programming"... 
2. Integrate *authorization*. Once a user is authenticates, we need to know what access permissions to grant them. That typically requires mapping whatever authorization is available on the provider (OpenID here) to required authorization. This is a product call, particularly when mapping dissimilar systems like OpenAPI and IAM.

arielshaqed:
# Proposal for phase 1 system: SSO

We shall provide SSO (authentication) via Dex but continue to provide authorization using lakeFS mechanisms.  Authorizations after being authenticated by Dex:
* If the user is (additionally) defined on lakeFS then lakeFS handles authorization in the same current way.
* If the user is unknown to lakeFS but authenticated then lakeFS attaches that user to a default group.

arielshaqed:
Taking this as part of wider discovery.

PR #2493, opened 2021-09-27
Redesign identity and AAA

arielshaqed:
* [x] Add definition of current state and problem
* [x] External authentication providers
* [x] ~External authorization providers~ - later
* [x] ~External audit providers~ - later

### DIFF
--- a/design/aaa.md
+++ b/design/aaa.md
@@ -134,7 +134,7 @@ application.
   The advantage here is that other OAuth2 providers can be used: GitHub,
   Google Suites, and many others.  LDAP users can still use bridges such
   as Dex to authenticate over OAuth2.  The [Microsoft Connector][dex-ms]
-  connector gives an example.
+  gives an example.
 
 #### Decision
 
@@ -171,6 +171,5 @@ This business logic is quite simple.
 OAuth2 has no audit mechanism and currently there is no request to allow
 such an audit.  Accordingly we shall not currently add audit logs.
 
-# References
-
+<!-- references -->
 [dex-ms]: https://dexidp.io/docs/connectors/microsoft/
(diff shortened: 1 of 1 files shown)


arielshaqed:
Relevant to #2058.

design/aaa.md
```diff
@@ -0,0 +1,175 @@
...
+have a groups mechanism in place which we might re-use.
+
+When do we attach policies do these groups?  Configuration changes apply
+at different times depending on this decision.  Authenticated users will
+see only changes in mappings that are attached before the change.
+
+#### Decision
+
+lakeFS shall receive groups from OAuth2 and encode them as claims on the
+JWT.  The remaining logic can proceed as today.
```
itaiad200, line 165: Would the groups be added to our IAM groups automatically/manually by the admin? I'm missing the full picture of setting up the oauth app, configuring the groups, attaching to external groups, etc..

arielshaqed:
@itaiad200 , thanks for the great review!

Closing this in favour of an upcoming new design to support LDAP directly-ish.  We only have current user requirements to do _that_.  We shall probably support (only) LDAP simple auth out of the box, and figure out an add-on interface for bigger configurations.

Meanwhile OAuth2 brings its own set of difficulties: most notably:
* OAuth2 doesn't really provide user identities; typically you'd want to add OpenID Connect.  But even there major providers offer vastly different query capabilities (e.g. Google _does not appear to provide_ group memberships on its OpenID Connect endpoint).
* lakeFS users will typically want to create AWS API keys for S3 and lakeFS API robots.  There is no clear path for these robots, because OAuth2 is not a directory server.
* OAuth2 servers typically authenticate users for short periods of time (hours) and often require exercising a "refresh token" even to get that far.  This is fine for a user logged into a browser client, but would require lakeFS to provide some other robot for refreshing tokens to keep AWS-style API keys current.

Stay tuned for LDAP.

Issue #2486, opened 2021-09-22
Store some per-file metadata (rclone multipart upload fails on unexpected ETag)


arielshaqed:
(Originally reported in #2445).

When uploading a file using rclone and multipart upload, rclone computes an MD5 checksum of the file contents, then runs HeadObject on the assembled file and expects to receive _that_ as the ETag.  Since the two are different, it fails.

Example (with 10M of random data in `/tmp/x`):
```
$ rclone copyto --dump headers,auth --s3-upload-cutoff 1M /tmp/x local-lake:/moo/main/2445/xyzzy
[... lots and lots of headers and retries and stuff...]
2021/09/22 15:19:23 DEBUG : >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
2021/09/22 15:19:23 DEBUG : HTTP REQUEST (req 0xc0000ffc00)
2021/09/22 15:19:23 DEBUG : HEAD /main/2445/xyzzy HTTP/1.1
Host: moo.s3.local.lakefs.io:8000
User-Agent: rclone/v1.56.0
Authorization: AWS4-HMAC-SHA256 Credential=AKIAIOSFODNN7EXAMPLE/20210922/eu-central-1/s3/aws4_request, SignedHeaders=host;x-amz-content-sha256;x-amz-date, Signature=47187765273b205a2bac70a19e3660f625e9f63e87fe821db6955bf9ba9fb098
X-Amz-Content-Sha256: e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855
X-Amz-Date: 20210922T121923Z

2021/09/22 15:19:23 DEBUG : >>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>>
2021/09/22 15:19:23 DEBUG : <<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<<
2021/09/22 15:19:23 DEBUG : HTTP RESPONSE (req 0xc0000ffc00)
2021/09/22 15:19:23 DEBUG : HTTP/1.1 200 OK
Content-Length: 31457280
Accept-Ranges: bytes
Date: Wed, 22 Sep 2021 12:19:23 GMT
Etag: "fc27755c7b393a94b8728fa8a959f837"
[...]

Issue #2429, opened 2021-09-01
Support Trino AVRO format access via S3 gateway


arielshaqed:
Currently #2296 causes this to fail -- it expects a particular Content-Type to be returned _when detecting an AVRO subtree_.

itaiad200:
@nopcoder make sure to test this one as well.

nopcoder:
move back to in-progress until we verify this one is fix and there are not followup issues we need to open.

nopcoder:
table creation, insert and select using AVRO format using Trino works.

Issue #2296, opened 2021-07-27
Support Content-Type


arielshaqed:
S3 supports Content-Type, and lakeFS should do the same.  E.g. Trino expects to see this (possibly as part of its S3A integration), when [checking for a directory](https://github.com/trinodb/trino/blob/78bbf3b5bc2aa4759174ccade6064f809e0b8d93/plugin/trino-hive/src/main/java/io/trino/plugin/hive/s3/TrinoS3FileSystem.java#L425) -- this is used when scanning to read Avro-format files (but not, it seems, other formats).

PR #2496, opened 2021-09-29
docs: design document on how to handle s3 metadata as part of our catalog

nopcoder:


design/object-metadata-s3.md
```diff
@@ -0,0 +1,49 @@
...
+1. Address type (relative / full)
+
+### What is missing?
+
+Mapped two issues that we currently do not address when we work with our S3 interface:
+
+1. Content-Type - Not keeping the posted object content-type and do not set it back when we get the object.
+  [Support Content-Type](https://github.com/treeverse/lakeFS/issues/2296)
+  We saw this issue when one of the clients used the value of the content type to identify a directory marker.
+  Assuming there are more use-cases where the content-type is set and/or used.
```
arielshaqed, line 21: More generally, because this exists in S3, it is reasonable for a user of the S3 gateway to expect that it will work.

design/object-metadata-s3.md
```diff
@@ -0,0 +1,49 @@
...
+
+1. Maintain a specific field for Content-Type as part of the object metadata. Will be use to set the content-type header when we return the object or HEAD information, example:
+
+   ```json
+   {
+       "AcceptRanges": "bytes",
+       "LastModified": "Tue, 28 Sep 2021 22:34:44 GMT",
+       "ContentLength": 10485760,
+       "ETag": "\"f962bf40d19fed2e80bcbaa33bd1dfe7\"",
+       "ContentType": "stam/data",
```
arielshaqed, line 43:
```suggestion
       "ContentType": "example/data",
```
(This ~MIME~Content type is [registered with IANA](https://www.iana.org/assignments/media-types/media-types.xhtml#examples) for use only in examples, preventing confusion...).

design/object-metadata-s3.md
```diff
@@ -0,0 +1,49 @@
...
+       "Metadata": {}
+   }
+   ```
+
+1. Use specific entry 'Content-Metadata' in the object's metadata map. The same map used in our open api. This header will have specific treatment, as we need to map it in a different way on get object. We will not enable the user to delete it and always set default value in case it is not set or set to empty string.
```
arielshaqed, line 48: I prefer this option.  E.g. [AWS S3](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html) calls all of these "metadata".

It also contains a list of some additional fields.  I believe that we should support all of them _where we can_.   E.g. `Content-MD5` is not really possible, and TBH I doubt that S3 can always return it).  And all or most of the `x-amz-*` fields (apart from `x-amz-meta-*`...) are inapplicable to lakeFS, and we should probably avoid setting them.

arielshaqed, line 48: `content-metadata`, or otherwise define how we canonicalize HTTP headers (which are generally as cAsEiNsEnSiTiVe as a Macbook).

arielshaqed (changes requested):
Thanks, this will be great for the applications that our users need to run!

My only request: Can you add a section about what we shall do with existing repositories?  I would like to be very clear about how this works; can be as simple as explicitly specifying behaviour for missing `ContentType` keys.

design/object-metadata-s3.md
nopcoder: Setting these as headers using the standard Go http will set them as standard TitleCase.

nopcoder: The part that wasn't clear or the reason why - mapping the S3 request metadata to lakefs entry metadata. Don't think we should do it or we should just for specific S3 headers?

design/object-metadata-s3.md
```diff
@@ -0,0 +1,48 @@
...
+1. Metadata
+1. Address type (relative / full)
+
+### What is missing?
+
+Mapped two issues that we currently do not address when we work with our S3 interface:
+
+1. Content-Type - Not keeping the posted object content-type and do not set it back when we get the object.
+    [Support Content-Type](https://github.com/treeverse/lakeFS/issues/2296)
+    We saw this issue when one of the clients used the value of the content type to identify a directory marker. But this is one of the basic properties that evey object store manages.
```
arielshaqed, line 20: Consider linking to the [S3 documentation on object metadata](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html#SysMetadata) to show that this is (currently) the only piece of user-modifiable system metadata.

design/object-metadata-s3.md
```diff
@@ -0,0 +1,48 @@
...
+Example of how head request on object with content-type will look like:
+
+```json
+{
+   "AcceptRanges": "bytes",
+   "LastModified": "Tue, 28 Sep 2021 22:34:44 GMT",
+   "ContentLength": 10485760,
+   "ETag": "\"f962bf40d19fed2e80bcbaa33bd1dfe7\"",
+   "ContentType": "example/data",
+   "Metadata": {}
```
arielshaqed, line 45: Do I (additionally) get the metadata here, or only in the `x-amz-meta-*` headers?  If not here, why do we have this field?

design/object-metadata-s3.md
```diff
@@ -0,0 +1,48 @@
...
+1. Additional metadata - when posting an object aws enables additional metadata to set. Currently we do not process these attributes. The object level metadata we currenly have on an object can be get/set only by our open api.
+    [Store some per-file metadata (rclone multipart upload fails on unexpected ETag) #2486](https://github.com/treeverse/lakeFS/issues/2486)
+
+
+### Solution
+
+AWS metadata posted with an object will be added to the object's metadata. We will map "x-amz-meta-<name>" to name/value in our metadata. The equivalent to create entry metadata using our open api.
+The s3 gateway get object will map the metadata key/value back to "x-amz-meta-<name>" headers.
+
+The content-type and the new metadata posted in multipart upload will be stored in our tracker. This information should be kept until the multipart complete is called and then the metadata is stored.
```
arielshaqed, line 30: How do I set content type using the API?

arielshaqed (commented):
Great, I think that this will really unblock 2 user issues.  Thanks!

arielshaqed:
Relevant issues so I can copy-paste into conversations:
* #2296 (Content-Type)
* #2429 (Trino Avro through S3 gateway)
* #2486 (RClone multipart upload)

PR #2497, opened 2021-09-29
Improve docs of  using lakeFS during deployment 

talSofer:
The title "version control" does not accurately describe the example it is followed with.

### DIFF
--- a/docs/usecases/ci.md
+++ b/docs/usecases/ci.md
@@ -18,7 +18,7 @@ Every day we introduce new data to the lake. And even if the code and infra does
 
 lakeFS enable CI/CD-inspired workflows to help validate expectations and assumptions about the data before it goes live in production or lands in the data environment.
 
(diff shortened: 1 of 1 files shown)


peacing (commented):
I understand the gist of the phrasing change. My suggestion is instead of "data change" to use "data update". Most of the time people aren't thinking "I have a dataset and I need to change it". It's more "I have a dataset that's updating daily/hourly/etc." and lakeFS helps make those updates safer

talSofer:
> I understand the gist of the phrasing change. My suggestion is instead of "data change" to use "data update". Most of the time people aren't thinking "I have a dataset and I need to change it". It's more "I have a dataset that's updating daily/hourly/etc." and lakeFS helps make those updates safer

Liked the suggestion, thanks.

peacing (approved):

Issue #2501, opened 2021-09-30
Bug: creating a user with invalid key format


nopcoder:
lakefs key is used for the s3 interface, which enforce the key to be a aws key format and it is not enforced.
Usually the user key/secret is generated - except the case we enable the user to setup or add superuser with specific credentials.

PR #2503, opened 2021-09-30
Validate user access key id on add credentials

nopcoder:
Fix: #2501


guy-har (changes requested):
Looks good,
Please add validation for secret key as well, it's a weird behaviour to enforce only a valid access key, also, having an invalid secret key might cause some bugs.

nopcoder:
A fixed found at #2505

PR #2504, opened 2021-09-30
hide delete button when viewing a non-HEAD

### LINKED ISSUE #1622: When viewing files by commit ref, Delete option should be disabled

johnnyaug:
Closes #1622


itaiad200 (approved):

arielshaqed (approved):
Cool,  thanks! Not sure how/why I dropped this during some previous PR.
Arguably this PR shows how much we really need to add semantics to viewing staging (+HEAD) vs viewing history (even at HEAD): mutation only makes sense for the first. That's a product language issue, *not* a limitation on this PR. The implementation here just shows how artificial this limitation is.

itaiad200:
Thanks for the quick fix @johnnyaug ! And many thanks and apologies to @shamikakumar who implemented and completed it months ago (before it was reverted accidentally).

PR #2507, opened 2021-10-03
Increase the code coverage of lakeFS SST parser

### LINKED ISSUE #2353: Test our own RocksDB reader

talSofer:
Closes  #2353
1. Add more unit tests for lakeFS's SST parser 
2. Add Jacoco plugin to track code coverage 

As part of the process of increasing the parser's code coverage I found the following two bugs: 

- https://github.com/treeverse/lakeFS/issues/2508
- https://github.com/treeverse/lakeFS/issues/2511


clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
```diff
@@ -57,11 +90,15 @@ func newGenerateNanoid(size int) func() (string, error) {
...
+
+func newGenerateFuzzWithSharedPrefixes(sharedPrefix string) func() (string, error) {
 	f := fuzz.NewWithSeed(FuzzerSeed)
 	return func() (string, error) {
-		var val string
-		f.Fuzz(&val)
-		return val, nil
+		var suffix string
+		f.Fuzz(&suffix)
+		return sharedPrefix + suffix, nil
```
arielshaqed, line 101: Can we vary the length of the shared prefix between adjacent elements?  E.g. maybe along these lines:

* Start with an empty prefix, sharedPrefixLen = 1 (say)
* Loop:
  * New key = prefix + random(10 bytes)
  * Randomly perform one of these actions (say each with probability 1/4, but it doesn't really matter):
    - Keep sharedPrefixLen.
    - Decrement sharedPrefixLen if it's >0
    - Increment sharedPrefixLen if it's <10
    - Set sharedPrefixLen = 0
  * Set prefix = substring(key, 0, sharedPrefixLen)

Now sharedPrefixLen will fluctuate between entries, but it _will_ change it it **will** occasionally become 0.  Might be a bit noisier.

WDYT?

clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
```diff
@@ -84,6 +121,12 @@ func writeSstsWithUnsupportedWriterOptions() {
 	writerOptions.TableFormat = sstable.TableFormatLevelDB
 	createTestInputFiles(keys, generateNanoidValue, sizeBytes, "table.format.leveldb",
 		writerOptions)
+
+	// Compression defines the per-block compression to use. lakeFS support Snappy or no compression.
+	writerOptions = newDefaultWriterOptions()
+	writerOptions.Compression = sstable.ZstdCompression
```
arielshaqed, line 127: So is it always with Snappy compression?

clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
```diff
@@ -287,71 +300,101 @@ class GolangContainerSpec extends AnyFunSpec with ForAllTestContainer {
...
     }
 
     describe("with random blocksize threshold") {
       it("should parse successfully") {
-        withGeneratedTestFiles("fuzz.block.size.threshold", verifyBlockParserOutput)
+        withGeneratedSstTestFiles("fuzz.block.size.threshold", verifyBlockParserOutput)
+      }
+    }
+
+    describe("with non-sst file of size 0") {
```
arielshaqed, line 379:
```suggestion
    describe("with empty file") {
```
(also rename the generator ?)

clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
```diff
@@ -48,6 +50,37 @@ func main() {
...
+	// after creating a valid Pebble sstable that includes a correct magic, remove part of the bytes that compose the
+	// magic to simulate an sstable with a bad magic.
+	fi, err := os.Stat(testFileName + ".sst")
+	if err != nil {
+		panic(err)
+	}
+	os.Truncate(testFileName+".sst", fi.Size()-MagicLengthBytes/2)
+}
+
+func writeNonSstFile() {
```
arielshaqed, line 71:
```suggestion
func writeEmptyFile() {
```

clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
```diff
@@ -287,71 +300,101 @@ class GolangContainerSpec extends AnyFunSpec with ForAllTestContainer {
...
+      ignore("should fail parsing") {
```
arielshaqed, line 380: Currently this would work?!?!

clients/spark/project/plugins.sbt
```diff
@@ -4,3 +4,4 @@ addSbtPlugin("com.frugalmechanic" % "fm-sbt-s3-resolver" % "0.19.0")
 addSbtPlugin("cf.janga" % "sbts3" % "0.10.5")
 addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.29")
 addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.0")
+addSbtPlugin("com.github.sbt" % "sbt-jacoco" % "3.3.0")
```
arielshaqed, line 7: Why is this in the project plugins?  AFAICT it is never used.  Perhaps put it in your personal plugins in `~/.sbt/1.0/plugins/plugins.sbt`?

arielshaqed (commented):
Neat... thanks!

clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
talSofer: I agree with the idea but took another approach to achieve this goal that worked better with the implementation. let me know if this addresses your comment.

clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
```diff
@@ -90,15 +90,30 @@ func newGenerateNanoid(size int) func() (string, error) {
...
+	}
 }
 
-func newGenerateFuzzWithSharedPrefixes(sharedPrefix string) func() (string, error) {
+func newGenerateFuzzWithSharedPrefixes() func() (string, error) {
 	f := fuzz.NewWithSeed(FuzzerSeed)
+	src := rand.NewSource(041021)
+	r := rand.New(src)
+	prefixBaseLen := 1 + r.Intn(100)
+	sharedPrefixBase, err := nanoid.New(prefixBaseLen)
```
arielshaqed, line 106: This is pure-ASCII.  Previously I think we had actually binary junk; that could be better because it might detect places where we performed bad binary -> String conversions.

clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
```diff
@@ -90,15 +90,30 @@ func newGenerateNanoid(size int) func() (string, error) {
...
+		var val string
+		f.Fuzz(&val)
+		return val, nil
+	}
 }
 
-func newGenerateFuzzWithSharedPrefixes(sharedPrefix string) func() (string, error) {
+func newGenerateFuzzWithSharedPrefixes() func() (string, error) {
 	f := fuzz.NewWithSeed(FuzzerSeed)
+	src := rand.NewSource(041021)
```
arielshaqed, line 103: This is an [octal integer literal](https://golang.org/ref/spec#Integer_literals), very useful on a [PDP-10](https://en.wikipedia.org/wiki/PDP-10) and great for encoding text in [SIXBIT](https://en.wikipedia.org/wiki/Six-bit_character_code), but possibly not what you intended. :eight: 

```suggestion
	src := rand.NewSource(20211004)
```

clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
```diff
@@ -90,15 +90,30 @@ func newGenerateNanoid(size int) func() (string, error) {
...
+	r := rand.New(src)
+	prefixBaseLen := 1 + r.Intn(100)
```
arielshaqed, line 105: Let's just take a fixed value here -- I worry that large values may behave a bit too differently (e.g. their random walk on line 115 may never go with length zero).

I'd go with something like
```suggestion
	prefixBaseLen := 1 + r.Intn(10)
```

Then line 115 goes to length 0 with probability 1/25, which is often enough.

clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
arielshaqed: Oh right, I forgot.  Can you please link the issue number in a comment?

arielshaqed (approved):
Thanks!

clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
talSofer: good point, thanks!

clients/spark/project/plugins.sbt
arielshaqed: I disagree.  It is not necessary to build or to test.  I searched for "sbt coverage" and I discovered [sbt-coverage](https://github.com/scoverage/sbt-scoverage).  Now we can have an argument which coverage plugin should we use?  E.g. think of participating in Hacktoberfest with such a line.

Basically I do not believe that we care to recommend a coverage plugin.  I regard this as part of preferred tooling.  We should avoid being opinionated here because:
a. We _can_; and
b. it's not our part of our core value.

arielshaqed (commented):
Sorry, missed response about sbt coverage.  tl;dr: let's not force our users to install tooling that is not necessary for development.

clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
```diff
@@ -100,13 +100,12 @@ func newGenerateFuzz() func() (string, error) {
...
+	src := rand.NewSource(20211004)
 	r := rand.New(src)
-	prefixBaseLen := 1 + r.Intn(100)
-	sharedPrefixBase, err := nanoid.New(prefixBaseLen)
-	if err != nil {
-		panic(err)
-	}
+	prefixBaseLen := 1 + r.Intn(10)
+	sharedPrefixBaseBytes := make([]byte, prefixBaseLen)
+	f.Fuzz(&sharedPrefixBaseBytes)
```
arielshaqed, line 107: I think that https://play.golang.org/p/ahKo2Y4YjNW works but this actually does not.

This actually does _not_ work: slices are not the same as arrays, and in particular the fuzz library takes only OUT parameters, so _it doesn't look at their currently allocated capacity or length_.  E.g. https://play.golang.org/p/HNZiBea8y_B does not manage to get the intended size.

clients/spark/core/src/test/resources/parser-test/sst_files_generator.go
```diff
@@ -100,13 +100,12 @@ func newGenerateFuzz() func() (string, error) {
...
+	sharedPrefixBase := string(sharedPrefixBaseBytes[:])
```
arielshaqed, line 108: (Slicing here is only required when you `sharedPrefixBaseBytes` is an array.)

clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
```diff
@@ -378,7 +378,7 @@ class GolangContainerSpec extends AnyFunSpec with ForAllTestContainer {
 
     describe("with empty file") {
       ignore("should fail parsing") {
-        assertThrows[BadFileFormatException]{ //TODO: make this test pass, and enable it
+        assertThrows[BadFileFormatException]{ //TODO (Tals): enable the test after closing https://github.com/treeverse/lakeFS/issues/2508
```
arielshaqed, line 381:
```suggestion
        assertThrows[BadFileFormatException]{ // TODO (tals): enable the test after closing https://github.com/treeverse/lakeFS/issues/2508
```

(I like the [Google TODO style](https://google.github.io/styleguide/cppguide.html#TODO_Comments).)

clients/spark/project/plugins.sbt
```diff
@@ -3,5 +3,4 @@ addSbtPlugin("com.github.sbt" % "sbt-pgp" % "2.1.2")
 addSbtPlugin("com.frugalmechanic" % "fm-sbt-s3-resolver" % "0.19.0")
 addSbtPlugin("cf.janga" % "sbts3" % "0.10.5")
 addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.29")
-addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.0")
-addSbtPlugin("com.github.sbt" % "sbt-jacoco" % "3.3.0")
\ No newline at end of file
+addSbtPlugin("org.scalameta" % "sbt-scalafmt" % "2.4.0")
```
arielshaqed, line 6: Can we please end files with a newline?  Most style guides recommend this; it sometimes matters when you want to run some script that does e.g.
```sh
echo 'new last line for file' >> clients/spark/project/plugins.sbt
```
because the expectation is that appending will be on a new line.

arielshaqed (commented):
Thanks!  See comment about confusing code for using fuzzer; AFAICT your way does not do what I expect it to do.

PR #2521, opened 2021-10-06
UI - keep path when changing reference in object view

### LINKED ISSUE #2441: [UI] Changing branch (Combobox) on the 'Objects' view resets the path
If the current path exists on the selected branch, you should keep it instead of reset into the root.

image

Taken from #2030

guy-har:
fixes #2441

### DIFF
--- a/webui/src/pages/repositories/repository/objects.jsx
+++ b/webui/src/pages/repositories/repository/objects.jsx
@@ -182,8 +182,8 @@ const ObjectsBrowser = ({ config, configError }) => {
                         withWorkspace={true}
                         selectRef={ref => router.push({
                             pathname: `/repositories/:repoId/objects`,
-                            params: {repoId: repo.id},
-                            query: {ref: ref.id}
+                            params: {repoId: repo.id, path:path},
+                            query: {ref: ref.id, path: path}
                         })}
                     />
                 </ActionGroup>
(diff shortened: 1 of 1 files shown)


webui/src/pages/repositories/repository/objects.jsx
```diff
@@ -182,8 +182,8 @@ const ObjectsBrowser = ({ config, configError }) => {
                         withWorkspace={true}
                         selectRef={ref => router.push({
                             pathname: `/repositories/:repoId/objects`,
-                            params: {repoId: repo.id},
-                            query: {ref: ref.id}
+                            params: {repoId: repo.id, path:path},
+                            query: {ref: ref.id, path: path}
```
itaiad200, line 186: I get `undefined` as path if I'm switching between branches without any path:
image

itaiad200 (changes requested):
See my comment around missing path.
Also - Do you think we should have the same behaviour for `Uncommitted Changes` and `Compare` tabs?

webui/src/pages/repositories/repository/objects.jsx
guy-har: Oops, fixed it, thanks!

guy-har:
@itaiad200 I am not sure it makes so much sense when it comes to `Uncommitted Changes` and `Compare`.
Maybe we should check with the user that requested this issue, WDYT?

itaiad200:
> @itaiad200 I am not sure it makes so much sense when it comes to `Uncommitted Changes` and `Compare`. Maybe we should check with the user that requested this issue, WDYT?

`Compare` involves 2 branches and it's a different experience, I agree. `Uncommitted` is basically the same view but with only uncommitted objects presented. I think that the expectation would be that they would behave the same.

itaiad200 (approved):
LGTM

PR #2526, opened 2021-10-06
Fix cloud meta logging message

nopcoder:


pkg/cloud/aws/metadata.go
```diff
@@ -44,7 +44,7 @@ func (m *MetadataProvider) GetMetadata() map[string]string {
 	stsClient := sts.New(sess)
 	identity, err := stsClient.GetCallerIdentity(&sts.GetCallerIdentityInput{})
 	if err != nil {
-		m.logger.WithError(err).Warn("%v: failed to get AWS account ID for BI")
+		m.logger.WithError(err).Warn("Failed to get AWS account ID for BI")
```
arielshaqed, line 47: I don't understand how this compiles. I thought Go can enforce Printf formats?!

arielshaqed (approved):
Good catch, thanks! ... I'm only worried why the compiler cannot catch this kind of thing.

PR #2528, opened 2021-10-06
s3 adapter disable chunked encoding

### LINKED ISSUE #2527: Oracle Object Storage S3 compatibility fail to put object
Oracle Object Storage enables Amazon S3 Compatibility API - https://docs.oracle.com/en-us/iaas/Content/Object/Tasks/s3compatibleapi.htm

We identified an issue while creating a repository - dummy file put failed with: `STREAMING-AWS4-HMAC-SHA256-PAYLOAD is not supported`.
Based on https://docs.oracle.com/en-us/iaas/Content/Object/Tasks/s3compatibleapi.htm - the API compatibility doesn't support chunked request that we use by default to upload object or part.

nopcoder:
Close #2527


nopcoder:
Thanks for the review, at this point will not go forwarding in enabling such an option as it will reduce the scalability of lakeFS's gateway.
Opened a specific feature request for Oracle Object Storage support https://github.com/treeverse/lakeFS/issues/2532.

PR #2535, opened 2021-10-09
Bug fix: pyramid store file fallback when destination already exist in cache

### LINKED ISSUE #2531: Bug: failed to merge branch running on Windows locally - access is denied
The following merge scenario failed running on windows:
1. create repository - local://my-bucket
2. upload object and commit to main branch
3. create branch 'branch1' based on 'main'
4. upload a different object to 'branch1' and commit
5. merge branch1 back to main

lakefs fail to merge, example of such error:

```
merge in CommitManager: close writer ns=local:// id=a2c182e06064ebeb386e63a4f839c09848a793da722640451d83f935939df45b: failed closing metarange writer: sstable store (653a7279d9d16a2dbf5d89158916557785e6c238a47c5ee51158fac14cbd23a7): rename C:\Users\dikla\data\lakefs\cache\meta-range/workspace/2d2f7ceb-81ad-4f21-a250-af557466d747 C:\Users\dikla\data\lakefs\cache\meta-range/653a7279d9d16a2dbf5d89158916557785e6c238a47c5ee51158fac14cbd23a7: Access is denied.
```
Attached process monitor with relevant records:
image (1)

nopcoder:
Fix #2531 

A common case when we merge changes from a branch to a destination branch is that the meta-range of the destination will be exactly as the source branch. While we read the meta-range to apply the merge and apply the changes to a new meta-range.  The last operation will try to move the generated output to a cache (at pyramid level). On Windows, this file is locked because an iterator still using it we fail to move the temporary data (new meta-range) to the cache folder.
The suggested change will skip moving the file if it already exists in the case and just removes the source. 
We assume that all files are downloaded and written to a temporary name, then moved to the cache, so if a file matches that name, it can be processed.

### DIFF
--- a/pkg/pyramid/directory.go
+++ b/pkg/pyramid/directory.go
@@ -75,7 +75,7 @@ func (d *directory) createFile(path string) (*os.File, error) {
 }
 
 // renameFile will move the src file to dst location and creates all parent dirs if missing.
-//   In we fail, we check if the destination exists, as it is content addressable, and remove the source.
+//   If we fail, we check if the destination exists, as it is content addressable, and remove the source.
 func (d *directory) renameFile(src, dst string) error {
 	d.mu.Lock()
 	defer d.mu.Unlock()


itaiad200:
@nopcoder what about the changelog?

PR #2556, opened 2021-10-13
Add multipart upload test to Spark integration tests

arielshaqed:


arielshaqed:
Closing: running through Spark tends to create huge in-memory or on-disk temporary files, and GitHub actions die off.  Will replace instead with a program that runs (only) hadoopfs S3A on lakeFS, configuring it for multipart upload.

PR #2568, opened 2021-10-17
Reduce verbosity of lakectl config file location

itaiad200:
This line appears on every use of lakectl which is redundant when you're probably always using the same file after you first configured it.


itaiad200:
Tested manually with/without config file and with different logging levels.

guy-har (approved):
Looks good! 
Thanks!

Issue #2345, opened 2021-08-10
Support reverting a range of commits


johnnyaug:
`lakectl branch revert <commit id>` adds a new commit which reverses the given commit. This is inspired by [git revert](https://git-scm.com/docs/git-revert#_examples).

If the user wants to revert multiple commits, they need to run the revert command multiple times. This will result in a lot of commits created for a simple revert. 

The git command has the ability to revert a range of commits, using a two-dot syntax:
```
git revert master~5..master~2 
```

Support a similar syntax on `lakectl branch revert`.

PR #2569, opened 2021-10-17
Support multiple commits in lakectl revert

eden-ohana:
Currently supports only individual commits.
Git revert is used to record some new commits to reverse the effect of some earlier commits.

#2345 We decided to not implement revert to a range of commits at the moment.

Open questions:
Git revert to multiple commits creates commit for every revert. Stick to git?
To implement the functionality on the server or on the client? When implemented in the client allow you to regret in the middle of the operation (revert commit 1 but not commit 2) like in git.

### DIFF
--- a/cmd/lakectl/cmd/branch.go
+++ b/cmd/lakectl/cmd/branch.go
@@ -2,7 +2,7 @@ package cmd
 
 import (
 	"fmt"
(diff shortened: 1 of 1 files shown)


johnnyaug:
Nice work! Please also add to the changelog

PR #2571, opened 2021-10-17
Test S3A hadoopfs multipart upload to lakeFS

### LINKED ISSUE #1927: Add integration tests for multipart upload
Multipart upload behaves differently depending on the client and the underlying storage.  
We would like to add integration tests for each of the following `client`/`adapter` combinations:

Clients:

* [x] AWS cli
* [ ] Spark 

Adapters:
- S3
- GCS
- Azure

arielshaqed:
Fixes #1927 by adding a multipart-from-S3A test.  (#2556 attempted and failed to do this through Spark, so use a program that directly runs S3A.)


arielshaqed:
*Reviewer:* Please review!  The test is currently failing _only_ because it cannot get logs.  It might be related to docker/compose#2067.  I will take a closer look and try to reproduce on my machine, but this should not interfere with the review.

More generally, if this is a recurring rough spot, we might want to find another way to verify the test actually used MPU.  Or we could skip that part -- it only verifies internal correctness of the test, which should remain...

PR #2615, opened 2021-10-26
Feature - present commit history for list of objects or/and prefixes

### LINKED ISSUE #2251: Present commit history for a specific file or prefix
We would like to support presenting commit history for a specific file of prefix (like in git log ).
e.g 
`lakectl log lakefs://example1/main/path/to/object`
`lakectl log lakefs://example1/main/some/prefix/`

*This was brought up by a user

lynnro314:
Closes #2251 

This feature gives the ability to get the log of all commits that contains changes to list of paths.
Log api now can get a list of prefixes and/or a list of objects, and returns all commits containing changes to one or more of the paths mentioned. 
I added the functionality under the log commits function. 
Also, I changed the api and lakectl log file to work with the new feature. 

Usage example:
`lakectl log lakefs://example1/main --objects "a/a.txt,a/b.txt" --prefixes b/`
will show all commits that relevant to either `a/a.txt`, `a/b.txt` or `b/`

### DIFF
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -1949,19 +1949,19 @@ paths:
       tags:
         - refs
       operationId: logCommits
-      summary: get commit log from ref
+      summary: get commit log from ref. If both objects and prefixes are empty, return all commits.
       parameters:
         - $ref: "#/components/parameters/PaginationAfter"
         - $ref: "#/components/parameters/PaginationAmount"
         - in: query
-          name: object_list
+          name: objects
           description: list of paths, each element is a path of a specific object
           schema:
             type: array
             items:
               type: string
         - in: query
-          name: prefix_list
+          name: prefixes
           description: list of paths, each element is a path of a prefix
           schema:
             type: array
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -895,6 +895,7 @@ components:
           minLength: 1
       required:
         - pattern
+
 paths:
   /setup_lakefs:
     post:
@@ -1952,6 +1953,20 @@ paths:
       parameters:
         - $ref: "#/components/parameters/PaginationAfter"
         - $ref: "#/components/parameters/PaginationAmount"
+        - in: query
+          name: object_list
+          description: list of paths, each element is a path of a specific object
+          schema:
+            type: array
+            items:
+              type: string
+        - in: query
+          name: prefix_list
+          description: list of paths, each element is a path of a prefix
+          schema:
+            type: array
+            items:
+              type: string
       responses:
         200:
           description: commit log
--- a/cmd/lakectl/cmd/common_helpers.go
+++ b/cmd/lakectl/cmd/common_helpers.go
@@ -24,7 +24,7 @@ var isTerminal = true
 var noColorRequested = false
 
 // ErrInvalidValueInList is an error returned when a parameter of type list contains an empty string
-var ErrInvalidValueInList = errors.New("invalid value in list. All values in list should be non-empty string")
+var ErrInvalidValueInList = errors.New("empty string in list")
 
 const (
 	LakectlInteractive        = "LAKECTL_INTERACTIVE"
--- a/cmd/lakectl/cmd/common_helpers.go
+++ b/cmd/lakectl/cmd/common_helpers.go
@@ -23,6 +23,9 @@ import (
 var isTerminal = true
 var noColorRequested = false
 
+// ErrInvalidValueInList is an error returned when a parameter of type list contains an empty string
+var ErrInvalidValueInList = errors.New("invalid value in list. All values in list should be non-empty string")
+
 const (
 	LakectlInteractive        = "LAKECTL_INTERACTIVE"
 	LakectlInteractiveDisable = "no"
--- a/docs/assets/js/swagger.yml
+++ b/docs/assets/js/swagger.yml
@@ -1949,19 +1949,19 @@ paths:
       tags:
         - refs
       operationId: logCommits
-      summary: get commit log from ref
+      summary: get commit log from ref. If both objects and prefixes are empty, return all commits.
       parameters:
         - $ref: "#/components/parameters/PaginationAfter"
         - $ref: "#/components/parameters/PaginationAmount"
         - in: query
-          name: object_list
+          name: objects
           description: list of paths, each element is a path of a specific object
           schema:
             type: array
             items:
               type: string
         - in: query
-          name: prefix_list
+          name: prefixes
           description: list of paths, each element is a path of a prefix
           schema:
             type: array
--- a/docs/assets/js/swagger.yml
+++ b/docs/assets/js/swagger.yml
@@ -225,6 +225,9 @@ components:
           description: Unix Epoch in seconds
         metadata:
           $ref: "#/components/schemas/ObjectUserMetadata"
+        content_type:
+          type: string
+          description: Object media type
 
     ObjectStatsList:
       type: object
@@ -259,6 +262,9 @@ components:
           description: Unix Epoch in seconds
         metadata:
           $ref: "#/components/schemas/ObjectUserMetadata"
+        content_type:
+          type: string
+          description: Object media type
 
     ObjectUserMetadata:
       type: object
@@ -819,6 +825,9 @@ components:
           type: object
           additionalProperties:
             type: string
+        content_type:
+          type: string
+          description: Object media type
       required:
         - staging
         - checksum
@@ -886,6 +895,7 @@ components:
           minLength: 1
       required:
         - pattern
+
 paths:
   /setup_lakefs:
     post:
@@ -1943,6 +1953,20 @@ paths:
       parameters:
         - $ref: "#/components/parameters/PaginationAfter"
         - $ref: "#/components/parameters/PaginationAmount"
+        - in: query
+          name: object_list
+          description: list of paths, each element is a path of a specific object
+          schema:
+            type: array
+            items:
+              type: string
+        - in: query
+          name: prefix_list
+          description: list of paths, each element is a path of a prefix
+          schema:
+            type: array
+            items:
+              type: string
       responses:
         200:
           description: commit log
@@ -2419,7 +2443,11 @@ paths:
           # This actually violates HTTP, which requires returning 201 if a new object was
           # created or 200 if an existing object was modified,
           # https://tools.ietf.org/html/rfc7231#section-4.3.4
-          description: successfully linked
+          description: object metadata
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ObjectStats"
         400:
           $ref: "#/components/responses/ValidationError"
         401:
--- a/docs/reference/commands.md
+++ b/docs/reference/commands.md
@@ -1925,7 +1925,8 @@ Show log of commits
 #### Synopsis
 {:.no_toc}
 
-Show log of commits for a given branch
+Show log of commits for a given branch. 
+With objects or prefixes, show log of commits containing changes to at least one of the paths.
 
 ```
 lakectl log <branch uri> [flags]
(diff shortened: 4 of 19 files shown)


api/swagger.yml
```diff
@@ -1952,6 +1953,20 @@ paths:
...
+            type: array
+            items:
+              type: string
+        - in: query
+          name: prefix_list
+          description: list of paths, each element is a path of a prefix
+          schema:
+            type: array
+            items:
+              type: string
```
arielshaqed, line 1969: Yay!  This gives a very clean API.

But some comments (APIs are hard).

1. How do I pass an array as a query param in OpenAPI?  If by repeating the keyword, then the suffix "`_list`" is janky, I need to say `...&object_list=a&object_list=b`.

2. There is a URL size limit, it might not be enough for all users.  Ideally we could also pass these in the body rather than in the query.  (This would also fix (1) above!)

3. The behaviour of passing both empty `object_list` _and_ empty `prefix_list` is inconsistent -- could we document it? 
    E.g. in the documentation of the entire GET, or on each of the 2 parameters something like
    > If both `object_list` and `prefix_list` are empty, return _all_ objects.

   The alternative might be to _avoid the inconsistency_.  E.g. if we put these params in a body then I would additionally suggest that they be inside an extra object -- call it `filter`.  Now if `filter` is *missing* then everything is returned, but if `filter` is *present* then only objects matching one of the lists are returned.

cmd/lakectl/cmd/common_helpers.go
```diff
@@ -23,6 +23,9 @@ import (
 var isTerminal = true
 var noColorRequested = false
 
+// ErrInvalidValueInList is an error returned when a parameter of type list contains an empty string
+var ErrInvalidValueInList = errors.New("invalid value in list. All values in list should be non-empty string")
```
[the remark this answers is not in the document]
arielshaqed: We probably missed this comment :-(  Can you address it?

pkg/api/controller_test.go
```diff
@@ -176,6 +186,25 @@ func TestController_GetRepoHandler(t *testing.T) {
...
+	t.Helper()
+	s := strconv.Itoa(params.seed)
+	for _, p := range params.paths {
+		err := cat.CreateEntry(ctx, params.repo, params.branch,
+			catalog.DBEntry{
+				Path:            p,
+				PhysicalAddress: onBlock(deps, "add"+s+"of:"+p),
+				CreationDate:    time.Now(),
+				Size:            int64(params.seed),
+				Checksum:        "cksum" + s,
```
arielshaqed, line 199: I don't understand `s` here: it will always be the same in the loop, so all `Checksum`s are equal.  If we don't care then just keep a constant `Checksum`, but if we do then please increment it or seed a PRNG or something.

arielshaqed (changes requested):
Thanks!

Still requesting one comment that you missed.

Also arrived late at the OpenAPI party, but the way we pass parameters is a bit inconsistent and will not allow large lists; maybe change there, and at _least_ please document what happens when both lists are empty.  (This is not a required change, to be clear)

api/swagger.yml
lynnro314: 1. I changed the name :)
2+3. Since it is a get request I think it should be a query param. I added a documentation. 
We tried at first to use a single param (instead of two lists), but we wanted the object to be flagged with 'object/prefix', and it didn't work.

pkg/api/controller_test.go
lynnro314: I'm passing seed and choosing it according to commit graph I created. I'm using different seed when I want to create a commit in which a file changed. Sounds Ok?

arielshaqed: Maybe the word "seed" in confusing me.  I understand it in this computing context, of a [PRNG seed](https://en.wikipedia.org/wiki/Random_seed).  Using the value and never growing anything from it confuses me, that's not what a seed does.

Also I don't understand why use `s` here and on line 196, but `params.seed` on l.198.  Also the type of `params.seed` is strange: it is literally cast *everywhere* it is used.

Can we just rename it to reflect what it is, give it type `int64`, and then, rather than `strconv` it and concatenate, use `fmt.Sprintf`?

api/swagger.yml
arielshaqed: Thanks for adding the doc!

In order to fix (2) I believe it is best to show that URLs have reasonable max sizes.  E.g. IE used to have a 2KiB maximum -- clearly not enough but also clearly a dead platform.  Chromium is a thousand times better and has a 2MiB maximum, which will be enough almost always.  AWS API gateway doesn't appear to have a limit.  Etc., etc.

AFAIU from reading [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-4.3.1), GETs with bodies are allowed.  [Stack Overflow](https://stackoverflow.com/questions/978061/http-get-with-request-body) has a nice explanation.

I guess we can start with just query params and then do the right thing after things break down for our users.

arielshaqed (approved):
Thanks!

pkg/api/controller.go
```diff
@@ -2996,6 +3014,36 @@ func paginationAmount(v *PaginationAmount) int {
 	return i
 }
 
+func resolvePathList(objectList *[]string, prefixList *[]string) []catalog.PathRecord {
+	var pathList []catalog.PathRecord
```
talSofer, line 3018: a suggestion, it's up to you
```suggestion
	var pathRecords []catalog.PathRecord
```

pkg/api/controller.go
```diff
@@ -2996,6 +3014,36 @@ func paginationAmount(v *PaginationAmount) int {
 	return i
 }
 
+func resolvePathList(objectList *[]string, prefixList *[]string) []catalog.PathRecord {
```
talSofer, line 3017: it's up to you but this is a suggestion 
```suggestion
func resolvePathList(objects *[]string, prefixes *[]string) []catalog.PathRecord {
```

pkg/api/controller.go
```diff
@@ -2996,6 +3014,36 @@ func paginationAmount(v *PaginationAmount) int {
...
+	var pathList []catalog.PathRecord
+	if objectList == nil && prefixList == nil {
+		return make([]catalog.PathRecord, 0)
+	}
+	if objectList != nil {
+		for _, path := range *objectList {
+			if path != "" {
+				path := path
```
talSofer, line 3025: I'm confused by this line, why do we need it?

pkg/api/controller_test.go
```diff
@@ -216,6 +245,189 @@ func TestController_CommitsGetBranchCommitLogHandler(t *testing.T) {
...
+					⁃	added data/a/foo.txt
+					⁃	added data/b/bar.txt
+				•	commit C
+					⁃	changed data/a/a.txt
+				•	commit D
+					⁃	changed data/b/b.txt
+				•	commit P
+					⁃	added data/c/banana.txt
+				•	commit R
+					⁃	merged branch-a to main
```
talSofer, line 277: Did you mean to write commit R - merged branch-b to main and commit N merged branch-a to main?

pkg/api/controller_test.go
```diff
@@ -216,6 +245,189 @@ func TestController_CommitsGetBranchCommitLogHandler(t *testing.T) {
...
+	})
+
+	cases := []struct {
+		name            string
+		objectList      *[]string
+		prefixList      *[]string
+		expectedCommits []string
+	}{
+		{
+			name:            "simpleObjectWithOneCommit",
```
talSofer, line 372: does simple means "single"?

pkg/catalog/catalog.go
```diff
@@ -891,22 +893,70 @@ func (c *Catalog) ListCommits(ctx context.Context, repository string, branch str
 		for _, parent := range v.Parents {
 			commit.Parents = append(commit.Parents, parent.String())
 		}
-		commits = append(commits, commit)
-		if len(commits) >= limit+1 {
+
+		if len(params.PathList) != 0 && len(v.Parents) == 1 {
```
talSofer, line 897: You may want to use one of the following suggestions to increase readability

```suggestion
		if len(params.PathList) != 0 && len(v.Parents) == NumberOfParentsOfNonMergeCommit {
```

```suggestion
		if len(params.PathList) != 0 && !isMergeCommit(v) {
```

talSofer: And also, can you explicitly check if the objects or prefix query params are there instead of checking the length of params list?

pkg/catalog/catalog.go
```diff
@@ -891,22 +893,70 @@ func (c *Catalog) ListCommits(ctx context.Context, repository string, branch str
...
+			// if we need to log only commits associated to the path, and also the current commit
+			// is not a merge commit, then we check if the current commit contains changes to the path
```
talSofer, line 899: Can you please edit this comment so that it is clearer? it explains the logic you implemented. 

To log only commits associated with a path, we check the diff between etc etc..

talSofer (commented):
@lynnro314 this is a really cool feature!! The comments are mainly around documentation and readability

pkg/api/controller.go
lynnro314: This is because of linter. I assign path later in line 3027, but it's address, and as I understand it shouldn't take the address of a loop variable.  This is why I reassign path variable inside the loop.

pkg/api/controller_test.go
lynnro314: Yes, thanks!

pkg/api/controller_test.go
lynnro314: Yes, I'll change it.

pkg/catalog/catalog.go
lynnro314: I'm not sure I fully understood, but I tried to make it more clear, and also the explanation about the diff is in the 'pathInCommit' function.

docs/assets/js/swagger.yml
```diff
@@ -1939,10 +1949,24 @@ paths:
       tags:
         - refs
       operationId: logCommits
-      summary: get commit log from ref
+      summary: get commit log from ref. If both objects and prefixes are empty, return all commits.
```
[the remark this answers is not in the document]
lynnro314: As Ariel requested, we want to specify how to return all commits.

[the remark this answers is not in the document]
talSofer: Ok, makes sense. thanks

pkg/catalog/catalog.go
```diff
@@ -894,9 +896,9 @@ func (c *Catalog) ListCommits(ctx context.Context, repository string, branch str
 			commit.Parents = append(commit.Parents, parent.String())
 		}
 
-		if len(params.PathList) != 0 && len(v.Parents) == 1 {
-			// if we need to log only commits associated to the path, and also the current commit
-			// is not a merge commit, then we check if the current commit contains changes to the path
+		if len(params.PathList) != 0 && len(v.Parents) == NumberOfParentsOfNonMergeCommit {
+			// if path list ins't empty, and also the current commit isn't a merge commit -
```
talSofer, line 900: nit; ins't -> isn't

docs/reference/commands.md
```diff
@@ -1988,7 +1987,7 @@ Copy or merge table
 #### Synopsis
 {:.no_toc}
 
-copy or merge table. the destination table will point to the selected branch
+Copy or merge table. the destination table will point to the selected branch
```
talSofer, line 1990: Is this change related to this pr?

talSofer (commented):
Looks good! added few more minor comments

docs/reference/commands.md
lynnro314: No, this was generated with make gen-doc. I believe sone changes from another PR was generated, and since it wasn't included in the relevant PR Barak and I thought it would be Ok to include them here.

talSofer (approved):

PR #2627, opened 2021-10-29
Fix login worked on second try

### LINKED ISSUE #2524: Pressing "Logout" button requires two logins to get back to a useful page
# Steps to reproduce

1. Log out of GUI.
   *Result:* Get logged out, redirect to page `auth/login?next=%2Fauth%2Flogin`.
2. Log in.
   *Result:* Get logged in, _but redirected to `auth/login`_ as requested.
3. Ignore URL bar, ask "what just happened?" and log in again.
   *Result:* Get logged in, redirect to main page.

# Analysis

The `next` query parameter after logging out seems wrong.  It should probably try to be the referring page.

nopcoder:
The first time the login acquired credentials but redirect to the login page.
Second time was optional, but this time there was no redirect instruction and we moved to root.

Fix include:
1. Redirect to login when route to login is skipped
2. Handle react state update on unmounted component - https://stackoverflow.com/questions/53949393/cant-perform-a-react-state-update-on-an-unmounted-component. Triggered when API call fail with authorization.

Fix #2524

### DIFF
--- a/webui/src/lib/hooks/api.jsx
+++ b/webui/src/lib/hooks/api.jsx
@@ -55,10 +55,14 @@ export const useAPI = (promise, deps = []) => {
 
     useEffect(() => {
         if (login) {
-            router.push({
-                pathname: '/auth/login',
-                query: {next: router.route}
-            });
+            const loginPathname = '/auth/login';
+            const place = {
+                pathname: loginPathname,
+            }
+            if (router.route !== loginPathname) {
+                place.next = router.route;
+            }
+            router.push(place);
             setLogin(false);
         }
     }, [login, router])
--- a/webui/src/lib/hooks/api.jsx
+++ b/webui/src/lib/hooks/api.jsx
@@ -51,29 +51,28 @@ const initialAPIState = {
 export const useAPI = (promise, deps = []) => {
     const router = useRouter();
     const [request, setRequest] = useState(initialAPIState);
-    const [ login, setLogin ] = useState(false);
+    const [login, setLogin] = useState(false);
 
     useEffect(() => {
         if (login) {
             const loginPathname = '/auth/login';
-            const place = {
+            if (router.route === loginPathname) {
+                return;
+            }
+            router.push({
                 pathname: loginPathname,
-            }
-            if (router.route !== loginPathname) {
-                place.next = router.route;
-            }
-            router.push(place);
+                query: {next: router.route},
+            });
             setLogin(false);
         }
     }, [login, router])
 
     useEffect(() => {
+        let isMounted = true;
+        setRequest(initialAPIState);
         const execute = async () => {
-            setRequest(initialAPIState);
             try {
                 const response = await promise();
-                // setResponse(response)
-                // setLoading(false)
                 setRequest({
                     loading: false,
                     error: null,
@@ -81,8 +80,10 @@ export const useAPI = (promise, deps = []) => {
                 });
             } catch (error) {
                 if (error instanceof AuthenticationError) {
-                    setLogin(true);
-                    return;
+                    if (isMounted) {
+                        setLogin(true);
+                    }
+                    return 
                 }
                 setRequest({
                     loading: false,
@@ -90,8 +91,9 @@ export const useAPI = (promise, deps = []) => {
                     response: null
                 });
             }
-        }
+        };
         execute();
+        return () => isMounted = false;
     }, deps);
     return {...request};
 }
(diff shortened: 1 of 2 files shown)


webui/src/pages/auth/login.jsx
```diff
@@ -23,18 +23,16 @@ const LoginForm = () => {
...
-                                    return router.push((!!next) ? next : '/');
-                                })
-                                .catch(err => {
-                                    setLoginError(err);
-                                })
-
+                            try {
+                                await auth.login(e.target.username.value, e.target.password.value);
+                                setLoginError(null);
+                                const nextOrRoot = next && next !== router.route ? next : '/';
```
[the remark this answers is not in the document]
arielshaqed: Not a fan of this change, then.  In general I do not trust setting a known-incorrect value of `next` and then attempting to fix it at the last moment; I would much prefer to set a correct value.

To amplify on my previous points:

## 1. Where is the special treatment of a falsy `next`?

Before the change it would redirect to `/`, now it redirects to whatever the falsy value does.

## 2. Let's set the correct value!

## 3. Behaviour on JWT timeout

I believe that this does the wrong thing whenever I am logged out, e.g. (also) if my cookie times out.

I tried this experiment:
1. Log in.
2. Navigate to repos page.
3. Erase the token cookie.
4. Click a repo, get redirected to login page (expected behaviour).
5. ... login page has a bad `next`.

With this patch, the login page will _still_ have  a bad `next`, and my reading of the fix above is that it will bring me to the main page.  The **expected** behaviour is to redirect to the old page after logging in -- AFAICT this patch does not do that.

arielshaqed (changes requested):
Sorry, I will need more convincing not to change it at the source.  I believe that this sometimes redirects to the right page but other times does not... and that it *usually* goes to the wrong place.

webui/src/pages/auth/login.jsx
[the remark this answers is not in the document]
nopcoder: Moved the check before we set the value of 'next' when route to the login page.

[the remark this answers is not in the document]
nopcoder: Moved the check before we set the value of 'next' when route to the login page.

webui/src/lib/hooks/api.jsx
```diff
@@ -51,43 +51,49 @@ const initialAPIState = {
...
         const execute = async () => {
-            setRequest(initialAPIState);
             try {
                 const response = await promise();
-                // setResponse(response)
-                // setLoading(false)
                 setRequest({
                     loading: false,
                     error: null,
                     response
```
arielshaqed, line 79:
```suggestion
                    response,
```

webui/src/pages/auth/login.jsx
```diff
@@ -23,18 +23,19 @@ const LoginForm = () => {
...
-                        }}>
+                            auth
+                                .login(e.target.username.value, e.target.password.value)
+                                .then(user => {
+                                    setLoginError(null);
+                                    return router.push((!!next) ? next : '/');
+                                })
+                                .catch(err => {
+                                    setLoginError(err);
+                                })
```
arielshaqed, line 36: I think this may be more readable as an async function (cannot put it as a suggestion because GitHub).

Something like this (untested, of course):
```js
async (e) => {
  e.preventDefault();
  try {
    const _user = await auth.login(e.target.username.value, e.target.password.value);
    setLoginError(null);
    return router.push((!!next) ? next : '/');
    } catch (e) {
      setLoginError(e);
    }
}
```

webui/src/pages/auth/login.jsx
```diff
@@ -23,18 +23,19 @@ const LoginForm = () => {
...
-                                setLoginError(err);
-                            }
-                        }}>
+                            auth
+                                .login(e.target.username.value, e.target.password.value)
+                                .then(user => {
+                                    setLoginError(null);
+                                    return router.push((!!next) ? next : '/');
+                                })
+                                .catch(err => {
```
arielshaqed: What is "`err`"?  The golang is up outside of webui and down cmd or pkg :stuck_out_tongue_closed_eyes:

arielshaqed (approved):
Thanks!  Reckon this should fix it nicely in all scenarios.

(To be fair, the redirect-to-login logic seems a bit convoluted.)

webui/src/pages/auth/login.jsx
nopcoder: :) I revered this change because I didn't wanted to add code style changes to the PR - but I also found the async syntax better to read.

PR #2634, opened 2021-11-02
Reset after param when prefix changes

itaiad200:
`after` param is a state managed by React. The state indicates the last entry in a given "TreeItem" and it's being used in future paginations for continuing to read from the last entry.
Since it's a state, we need to reset it when the prefix changes since the pagination state is completely different.

### DIFF
--- a/webui/src/lib/components/repository/changes.jsx
+++ b/webui/src/lib/components/repository/changes.jsx
@@ -60,9 +60,9 @@ const ChangeRowActions = ({ entry, onRevert }) => {
     getMore: callback to be called when more items need to be rendered
     depth: the item's depth withing the tree
  */
-export const TreeItem = ({ entry, repo, reference, internalRefresh, onRevert, onNavigate, delimiter, after, relativeTo, getMore, depth=0 }) => {
+export const TreeItem = ({ entry, repo, reference, internalRefresh, onRevert, onNavigate, delimiter, relativeTo, getMore, depth=0 }) => {
     const [expanded, setExpanded] = useState(false); // state of the item expansion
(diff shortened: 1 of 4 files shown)


webui/src/lib/components/repository/changes.jsx
```diff
@@ -60,9 +60,9 @@ const ChangeRowActions = ({ entry, onRevert }) => {
     getMore: callback to be called when more items need to be rendered
     depth: the item's depth withing the tree
  */
-export const TreeItem = ({ entry, repo, reference, internalRefresh, onRevert, onNavigate, delimiter, after, relativeTo, getMore, depth=0 }) => {
+export const TreeItem = ({ entry, repo, reference, internalRefresh, onRevert, onNavigate, delimiter, relativeTo, getMore, depth=0 }) => {
     const [expanded, setExpanded] = useState(false); // state of the item expansion
-    const [afterUpdated, setAfterUpdated] = useState(after); // state of pagination of the item's children
+    const [afterUpdated, setAfterUpdated] = useState(""); // state of pagination of the item's children
```
talSofer, line 65: Why not call this const `lastEntry`? given your review description, this makes more sense to me. WDYT?

itaiad200: `after` is associated with the pagination results that determines it, my description was kind of leaning toward the implementation in the server side which is the last item in the list. We use `after` everywhere in the API and UI, and don't actually set it ourselves (we just use the next token from the response)

talSofer: Ok, I see. thanks.

PR #2635, opened 2021-11-02
Use go-pretty/v6

arielshaqed:
This bypasses a dependabot alert on using mongo-driver (!).  It's a harmless
warning, users can only harm their own lakectl run on their own machine, but
it does shut up a bot.


arielshaqed:
# How I Wrote This PR All By Myself

1. Replaced `go-pretty/` imports with `go-pretty/v6/` (just like https://github.com/jedib0t/go-pretty#v600 says).
2. `go get && go mod tidy`.
3. Committed and pushed.

nopcoder (changes requested):
Thanks - can you check if by pulling the latest (or v6.x) of go-pretty it will solve the problem too.
The package go-pretty supports go modules so it is not required to use the path based versioning to get the newer version.

arielshaqed:
Tested CLI by running `lakectl branch` (tables work) and running `lakectl diff` (colours work).

arielshaqed:
> Thanks - can you check if by pulling the latest (or v6.x) of go-pretty it will solve the problem too. The package go-pretty supports go modules so it is not required to use the path based versioning to get the newer version.

Looked at it F2F, that doesn't work because the `go.mod` file in `go-pretty` actually wants that "`v6`".

Thanks!

nopcoder (approved):
after face time I see that the go modules they added was on the 'v6' path - so approved :)

arielshaqed:
Thanks!

PR #2636, opened 2021-11-02
Use dockertest/v3.8.0

arielshaqed:
This bypasses a dependabot alert on using runc.  The alert itself is
relevant only for running tests.  These versions should no longer be
affected by the alerts.


arielshaqed:
Should fix https://github.com/treeverse/lakeFS/security/dependabot/go.sum/github.com%2Fopencontainers%2Frunc/open

nopcoder (approved):
Looks good

Issue #2629, opened 2021-10-31
Check lakeFS Hadoop Filesystem with Hadoop CLI


itaiad200:
There seems to be some missing functionality in the filesystem to support hadoop cli. A common error between `cp` and `mv` commands:

```
hadoop fs \
> -D fs.lakefs.access.key=<key> \
> -D fs.lakefs.secret.key=<secret> \
> -D fs.lakefs.endpoint=<endpoint>/api/v1 \
> -D fs.lakefs.impl=io.lakefs.LakeFSFileSystem \
> -libjars hadoop-lakefs-assembly-0.1.0.jar \
> -cp lakefs://example-repo/main/user2/yaaqov   lakefs://example-repo/main/user2/yaaqov2
cp: `tal-test': File exists
```

itaiad200:
`rm` command fails differently:
```
hadoop fs -D fs.lakefs.access.key=<key> -D fs.lakefs.secret.key=<secret> -D fs.lakefs.endpoint=<endpoint>/api/v1 -D fs.lakefs.impl=io.lakefs.LakeFSFileSystem -libjars hadoop-lakefs-assembly-0.1.0.jar -rm -r -f lakefs://example-repo/main/user2/yaaqov
rm: statObject: Bad Request. Consider using -skipTrash option
```

arielshaqed:
> There seems to be some missing functionality in the filesystem to support hadoop cli. A common error between `cp` and `mv` commands:
> 
> ```
> hadoop fs \
> > -D fs.lakefs.access.key=<key> \
> > -D fs.lakefs.secret.key=<secret> \
> > -D fs.lakefs.endpoint=<endpoint>/api/v1 \
> > -D fs.lakefs.impl=io.lakefs.LakeFSFileSystem \
> > -libjars hadoop-lakefs-assembly-0.1.0.jar \
> > -cp lakefs://example-repo/main/user2/yaaqov   lakefs://example-repo/main/user2/yaaqov2
> ```

Perhaps the error message is missing here?

itaiad200:
Right, edited above.

PR #2643, opened 2021-11-03
Fix bad logging

talSofer:
Discovered during a live session with users. 

The globStatus operation was not implemented because it was not called by Spark in any of the flows we tested. This pr only mitigates a bad log message but the real issue will be addressed with https://github.com/treeverse/lakeFS/issues/2629

I'm not changing globStatus to an unsupported operation because it may cause unexpected issues and require more thorough testing that will be done as part of the work on the issue above.

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -759,8 +759,9 @@ public class LakeFSFileSystem extends FileSystem {
 
     @Override
     public FileStatus[] globStatus(Path pathPattern) throws IOException {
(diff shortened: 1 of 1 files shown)


nopcoder (approved):

talSofer:
> @talSofer consider check the base implementation - if it uses existing API we may not need to override this method.

@nopcoder thanks! I agree with this approach and looked into it. I still rather backing up such a change with proper testing so will go with the current version.

nopcoder:
> > @talSofer consider check the base implementation - if it uses existing API we may not need to override this method.
> 
> @nopcoder thanks! I agree with this approach and looked into it. I still rather backing up such a change with proper testing so will go with the current version.

sure, I didn't want to implementation something that we may have already working.Check if the test suite that Yoni added may have tests also for this capability.

PR #2650, opened 2021-11-05
Fix s3 adapter multipart upload working with encrypted bucket

### LINKED ISSUE #2656: Multipart upload part content verification failed on s3 encrypted bucket
Multipart upload issue was discovered by user while upload large data using Apache Spark which triggered multipart upload and failed with the following error:
```
com.amazonaws.SdkClientException: Unable to verify integrity of data upload.  Client calculated content hash (contentMD5: <HEX> in hex) didn't match hash (etag: <HEX> in hex) calculated by Amazon S3.  You may need to delete the data stored in Amazon S3.
```
The underlying storage was using server side encryption (SSE-KMS), in this case the content checksum calculation for ETag is not MD5 and the underlying storage returns additional headers on create and upload of multipart to inform the client.

nopcoder:
Fix #2656

Multipart upload ETag calculation is different for s3 bucket when it is encrypted.
AWS S3 send server side headers as part of create/upload part responses so the caller can perform content verification currently.
In this PR:
1. S3 block adapter returns the all server side headers accepted by AWS.
2. Align ETag surrounded with double-quotes. Block adapters accept/return ETag without double-quotes and S3 Gateway is responsible to wrap the value.

### DIFF
--- a/pkg/block/adapter.go
+++ b/pkg/block/adapter.go
@@ -64,7 +64,28 @@ type WalkOpts struct {
 	Prefix           string
 }
 
-// CreateMultiPartOpts contains optional arguments for
+// CreateMultiPartUploadResponse multipart upload ID and additional headers (implementation specific) currently it targets s3
+// capabilities to enable encryption properties
+type CreateMultiPartUploadResponse struct {
+	UploadID string
+	Header   http.Header
+}
+
+// CompleteMultiPartUploadResponse complete multipart etag, content length and additional headers (implementation specific) currently it targets s3
+type CompleteMultiPartUploadResponse struct {
+	ETag          string
+	ContentLength int64
+	Header        http.Header
+}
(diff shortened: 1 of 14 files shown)


guy-har (approved):
Looks good, Thanks!
Maybe we should document somewhere why we wrap and unwrap the ETag ,WDYT?

guy-har:
@nopcoder, please provide a comment for the changelog message, or change the title accordingly.

nopcoder:
> Looks good, Thanks!
> Maybe we should document somewhere why we wrap and unwrap the ETag ,WDYT?

Good idea, I'll add it to the block/adapter code

nopcoder:
> @nopcoder, please provide a comment for the changelog message, or change the title accordingly.

I'll also associate bug

PR #2654, opened 2021-11-07
Better documentation and error handling of uploadObject

### LINKED ISSUE #2652: docs: add details to REST API uploadObject endpoint
I'm a new user.  I'm writing a prototype and I wanted to share that the new user experience is a little rough, particularly in one area: getting data into lakeFS via the REST API.  I had to read source code to figure out how to use the API.  

https://docs.lakefs.io/reference/api.html#/objects/uploadObject

Details that would help future users: 

1. _Only_ multipart uploads appear to be supported.
2. One file per upload.  Named content.  If you do not name your file content you get the unfriendly error message in the trace logs "http: no such file".  I figured this out after poking around the source code.  
3. path is relative to the branch (obvious, but stating the obvious in api docs would help confusion and the sense of being lost).

Thanks for lakeFS!

itaiad200:
closes #2652 
1. Better description of the path as relative to the branch/ref.
2. Informing when `content` key is missing from multipart upload.


guy-har (approved):
LGTM thanks!

Issue #2661, opened 2021-11-08
Branch creation concurrrency


johnnyaug:
Two concurrent branch creations can both succeed.

In graveler.go, CreateBranch checks for the branch existence, and then calls SetBranch from RefManager, which updates the branch reference on conflict.

Since there is no concurrency handling, two concurrent requests may both succeed.

Issue #2405, opened 2021-08-25
Bug: race condition could lead to lost writes


ozkatz:
I believe there's a race condition in Graveler, that in some cases could lead to lost writes.
This is a design flaw with [graveler.Set()](https://github.com/treeverse/lakeFS/blob/7377b6e1fc3c5968dcce02e01154c001948b2fd1/pkg/graveler/graveler.go#L1015):

As it is currently implemented, it relies on two independent DB connections:
1. one that tries to ensure mutual exclusion between _writing_ to a branch, and _changing_ it (for example, a commit will delete all records with the branch's current staging token, replacing it with a new one).
2. one that, once acquired, relies on said mutual exclusion to read the branch's current staging token, and add records to it.

Since these are two separate mechanisms, lock safety is not guaranteed, especially in the face of a connection error or reset: the use of [`pg_advisory_xact_lock_shared`](https://www.postgresql.org/docs/9.1/functions-admin.html) as the locking mechanism, ties the lock's liveness to the [life of the PostgreSQL transaction](https://stackoverflow.com/a/32447870).

This means that the branch locker is essentially broken: it can not guarantee there are no modifications to a branch while the callback is running, which will in this case, result in writes being lost, as illustrated below:

![image](https://user-images.githubusercontent.com/205955/130766673-4c8c74d9-4985-4a08-a2c4-bd7aa6885476.png)

While this is a possible scenario, it relies on 2 rather uncommon events happening at the same time:

1. a network failure (or some other reason causing the Postgresql server to terminate the locker's transaction) but one that does not affect the successful completion of the writer's transaction underlying connection
2. simultaneously, a commit (or another event that mutates the branch's staging token) must begin after the writer function has read the current staging token for the branch.

While this is (arguably?) a rare scenario, it violates a basic guarantee provided by lakeFS.

arielshaqed:
After F2F discussions I agree that this scenario is possible.

In general, as soon as we have 2 transactions, the onus of proving consistency is on us.  I propose that we change the commit/write transaction to be safe, and use the separate locking transaction purely in order to give the desired blocking semantics.

[...]

PR #2663, opened 2021-11-08
Fix branch creation concurrency bug

johnnyaug:
Solves #2661
﻿

### DIFF
--- a/pkg/graveler/ref/manager.go
+++ b/pkg/graveler/ref/manager.go
@@ -175,6 +175,20 @@ func (m *Manager) GetBranch(ctx context.Context, repositoryID graveler.Repositor
 	return branch.(*graveler.Branch), nil
 }
 
+func (m *Manager) CreateBranch(ctx context.Context, repositoryID graveler.RepositoryID, branchID graveler.BranchID, branch graveler.Branch) error {
+	_, err := m.db.Transact(ctx, func(tx db.Tx) (interface{}, error) {
+		_, err := tx.Exec(`
+			INSERT INTO graveler_branches (repository_id, id, staging_token, commit_id)
+			VALUES ($1, $2, $3, $4)`,
+			repositoryID, branchID, branch.StagingToken, branch.CommitID)
+		return nil, err
+	})
+	if errors.Is(err, db.ErrAlreadyExists) {
+		return graveler.ErrBranchExists
+	}
+	return err
+}
+
 func (m *Manager) SetBranch(ctx context.Context, repositoryID graveler.RepositoryID, branchID graveler.BranchID, branch graveler.Branch) error {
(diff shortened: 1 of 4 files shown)


pkg/graveler/ref/manager_test.go
```diff
@@ -177,6 +177,49 @@ func TestManager_GetBranch(t *testing.T) {
 		}
 	})
 }
+func TestManager_CreateBranch(t *testing.T) {
```
nopcoder, line 180: missing empty line before this one

pkg/graveler/ref/manager_test.go
```diff
@@ -177,6 +177,49 @@ func TestManager_GetBranch(t *testing.T) {
...
+		CreationDate:     time.Now(),
+		DefaultBranchID:  "main",
+	}, ""))
+
+	err := r.CreateBranch(ctx, "repo1", "f1", graveler.Branch{CommitID: "c1", StagingToken: "s1"})
+	testutil.MustDo(t, "create branch f1", err)
+
+	br, err := r.GetBranch(ctx, "repo1", "f1")
+	testutil.MustDo(t, "get f1 branch", err)
+	if br == nil {
```
nopcoder, line 194: note this should get an error base on code - but I guess it is always test all possibilities.

nopcoder (approved):
Looks good!

pkg/graveler/ref/manager.go
```diff
@@ -175,6 +175,20 @@ func (m *Manager) GetBranch(ctx context.Context, repositoryID graveler.Repositor
...
+	_, err := m.db.Transact(ctx, func(tx db.Tx) (interface{}, error) {
+		_, err := tx.Exec(`
+			INSERT INTO graveler_branches (repository_id, id, staging_token, commit_id)
+			VALUES ($1, $2, $3, $4)`,
+			repositoryID, branchID, branch.StagingToken, branch.CommitID)
+		return nil, err
+	})
+	if errors.Is(err, db.ErrAlreadyExists) {
+		return graveler.ErrBranchExists
+	}
```
arielshaqed, line 188: IIUC this replaces `SetBranch`, which would "re-seat" the branch at the new location.  This is a cool fix, but I don't understand how this fixes any other calls on graveler racing with anything else that calls `SetBranch`, e.g. `Commit`.  Those are considerably more dangerous I reckon.

So I think the real issue remains #2405.

arielshaqed (commented):
Looks OK, but the basic problem is that _all_ `SetBranch` calls end up racing due to #2405.

pkg/graveler/ref/manager.go
nopcoder: True, this one will solve the creation of two branches creation in parallel - but not the locking issue based on the second connection pool.

PR #2665, opened 2021-11-08
Fix lakectl diff docs

johnnyaug:


nopcoder (changes requested):
1. Thanks, forgot to fix the example.
2. run 'make gen-docs'

nopcoder (approved):

PR #2672, opened 2021-11-08
golang-migrate package upgrade

nopcoder:


arielshaqed (approved):
Thanks!  Unfortunately important.

How did you test this?  I *think* it's enough to test "THE" migration (using the system test), but I would rather make sure by testing the 1-step migrate from _without_ this change to _with_ this change.  (You probably had to do it anyway, unless you wiped your local lakeFS database before trying this on your machine...).

Inception: Upgrading package `golang-migrate`... we need to go deeper

nopcoder:
Additional manual tests run by running migrate up and down using the previous version and this one, thanks @arielshaqed.

PR #2673, opened 2021-11-09
java client generation start fresh

### LINKED ISSUE #2669: Java client cleanup before generate
Java client code generation should cleanup the generated files first, in order to have old code remove and fix issues where the code generation doesn't apply changes like the test code we previously saw.

nopcoder:
Fix #2669
Clean the java client folder as part of the code generation.


Makefile
```diff
@@ -121,6 +121,7 @@ client-python: api/swagger.yml  ## Generate SDK for Python client
 		-o /mnt/clients/python
 
 client-java: api/swagger.yml  ## Generate SDK for Java (and Scala) client
+	find clients/java -mindepth 1 -delete
```
arielshaqed, line 124: This does not seem a safe thing to do.  If we care about specific names, let's delete only them.  Otherwise I'd be happiest to delete _everything_.  E.g. if everything were generated into a single directory that *only* contained generated code, we could wipe that directory and always be safe.

arielshaqed (changes requested):
Unsure (but might be convinced!) about how we delete.

arielshaqed (approved):
Neat; thanks!

PR #2676, opened 2021-11-09
remove auth migrate down statements

### LINKED ISSUE #2674: Database migrate down lower than version 33 fails
Migrate under version 33 will process database migrate down script 33 which fails with the following. 
```
% lakefs migrate goto --version 32

Failed to migrate to version 32.
2 errors occurred:
        * migration failed: null value in column "statement" violates not-null constraint, Failing row contains (5, 2021-11-09 07:57:20.116902+00, RepoManagementReadAll, null). in line 0: BEGIN;

UPDATE auth_policies
SET statement = statement - (SELECT ordinality-1 AS index
                             FROM auth_policies CROSS JOIN jsonb_array_elements(statement) WITH ORDINALITY
                             WHERE display_name = 'RepoManagementReadAll' AND value = '{"Action": ["branches:Get*"], "Effect": "allow", "Resource": "*"}')::int
WHERE display_name = 'RepoManagementReadAll' AND statement @> '[{"Action": ["branches:Get*"], "Effect": "allow", "Resource": "*"}]'::jsonb;

UPDATE auth_policies
SET statement = statement - (SELECT ordinality-1 AS index
                             FROM auth_policies CROSS JOIN jsonb_array_elements(statement) WITH ORDINALITY
                             WHERE display_name = 'RepoManagementFullAccess' AND value = '{"Action": ["branches:*"], "Effect": "allow", "Resource": "*"}')::int
WHERE display_name = 'RepoManagementFullAccess' AND statement @> '[{"Action": ["branches:*"], "Effect": "allow", "Resource": "*"}]'::jsonb;

COMMIT;
 (details: pq: null value in column "statement" violates not-null constraint)
        * pq: current transaction is aborted, commands ignored until end of transaction block in line 0: SELECT pg_advisory_unlock($1)
```

nopcoder:
keeping the updates data is a safer option than trying to remove
something that may no longer valid or the user choose to add.

Fix #2674


arielshaqed (approved):
This is great, thanks!

Do all of these fail to run?  Or are they because of our decision to keep existing data unchanged as far as possible??

nopcoder:
> This is great, thanks!
> 
> Do all of these fail to run? Or are they because of our decision to keep existing data unchanged as far as possible??

Removed all the changes in order it to be aligned.
The problem in testing it, and now I understand better that it is a problem to check migrate with policies goes like this:
We can migrate to a specific version, but auth policies data are inserted as part of 'setup', not part of migration.
Setup process performs 'migrate up' and insert initial data.
So testing migrate down (go to) to each level back in manual tests failed as it always stepped from the latest.
Testing the above just by running 'migrate goto' with clean database works as it just manipulate the table structure and the above updates works.
In order to test each step a binary without the embedded level should be generated.
Think we did right by just removing the option to manipulate data while migrate down.

PR #2681, opened 2021-11-09
Look up existing S3 client in cache only once

arielshaqed:
The entire cache is heavily optimized to use an existing S3 client -- which
happens moments after that S3 client is first used.  When no client exists,
it races (against itself) to create a new one.  The previous revision might
return a newer client during such a race than this new revision.  But if we
ever *delete* from the cache, it may return a nil.  The newer revision will
never fail in this way.  These races exist in both revisions, are harmless,
and both the older and the newer client will be safe to use.


johnnyaug (approved):
Thanks!

PR #2683, opened 2021-11-10
Add request ID fields to all DB and auth logs

### LINKED ISSUE #2682: Put request_id on (almost) every log
All logs made specifically for handling a request should include a request_id field.  This includes e.g. database queries, AWS operations, etc., but possibly **not** things like cache expiries where a particular request is not responsible for the expiry.  (When a cache expiry causes issues, a stronger correlating factor between cache expiry and Bad Things Happening to a request is concurrency with expiry -- request IDs are not necessary for this.)

arielshaqed:
Partly of #2682.  Most important gap is causing AWS clients (and, less
crucially, Azure and GCP) to log the request ID.

### DIFF
--- a/pkg/db/database.go
+++ b/pkg/db/database.go
@@ -54,11 +54,14 @@ func NewPgxDatabase(db *pgxpool.Pool) *PgxDatabase {
 	return &PgxDatabase{db: db}
 }
 
-func (d *PgxDatabase) getLogger() logging.Logger {
+func (d *PgxDatabase) getLogger(ctx context.Context, fields logging.Fields) logging.Logger {
+	var log logging.Logger
 	if d.queryOptions != nil {
-		return d.queryOptions.logger
+		log = d.queryOptions.logger
+	} else {
+		log = logging.Default()
 	}
-	return logging.Default()
+	return log.WithContext(ctx).WithFields(fields)
 }
 
 func (d *PgxDatabase) Close() {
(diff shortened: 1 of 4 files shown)


pkg/db/database.go
```diff
@@ -54,11 +54,14 @@ func NewPgxDatabase(db *pgxpool.Pool) *PgxDatabase {
...
+func (d *PgxDatabase) getLogger(ctx context.Context, fields logging.Fields) logging.Logger {
+	var log logging.Logger
 	if d.queryOptions != nil {
-		return d.queryOptions.logger
+		log = d.queryOptions.logger
+	} else {
+		log = logging.Default()
 	}
-	return logging.Default()
+	return log.WithContext(ctx).WithFields(fields)
```
nopcoder, line 64: General concern - not knowing when to call WithContext - looks like it is something we need to load on our logger as soon as possible or defer it to the last thing we do.

nopcoder (approved):
LGTM - wrote about one concern I have with WithContext

pkg/db/database.go
arielshaqed: I believe this is the right place to do it.  I changed the signature of `getLogger` to say you need to know the context in which you will be logging.  So as soon as you need it, you *must* give context and additional fields.  Now outputting a bad DB log without context and/or fields is actually more difficult than outputting a good one :-)

arielshaqed (commented):
Thanks!

PR #2685, opened 2021-11-10
Feature: Add object-level diff

### LINKED ISSUE #2437: lakefs UI diff between two objects
### User Story
User changed a file manually and uploaded it to lakeFS. Currently, lakeFS only shows that the file was changed without going into details, like number of bytes/lines that changed, showing the lines that were changed, etc.. For seeing the actual differences the user needs to download the 2 files and use an external tool to perform the diff (even 3 files if both files were changed in the compare view).
The user will prefer to use lakeFS UI in order to see the actual differences for files that are readable.

### Suggestion  
For human-readable formats, lakeFS can show a line-based diff on the UI between three/two objects on the diff and changes views.

talSofer:
Closes #2437 

### Description
Add object-level diff with the following specifications: 
- Show content-based diff for readable csv, tsv, and txt files up to 20KB size. 
- Report the change in sizes for non-readable objects. 

### Testing Matrix 
Testing is still in-progress, these are the planned test cases - 

| Test Case      | Uncommitted Changes       | Compare       | Commit
| ------------- | -------------------------- | ------------- | ------------- |
Readable csv ≤ max size added |  V | V | V |
Readable csv  ≤ max size removed | V | V | V |
Readable csv  ≤ max size changed  | V | V | V |
Readable tsv  ≤ max size added | V | V | V |
Readable tsv  ≤ max size changed | V | V | V |
Readable tsv  ≤ max size  removed | V | V | V |
Readable txt  ≤ max size added | V | V | V |
Readable txt  ≤ max size changed | V | V | V |
Readable txt ≤ max size  removed | V | V | V |
Readable csv > max size added | V |V| V |
Readable csv  > max size changed | V |V| V |
Readable csv  > max size removed | V | V | V |
Readable txt  > max size added | V | V | V |
Readable txt  > max size changed | V |V| V |
Readable txt > max size  removed | V | V | V |
Readable tsv  > max size added | V | V | V |
Readable tsv  > max size changed | V |V| V |
Readable tsv  > max size  removed | V | V | V |
Non-readable added | V | V | V |
Non-readable removed | V | V | V |
Non-readable changed | V | V | V |
Non-readable conflict | - | V | - |
Readable conflict | - | V | - |

### Sample screenshots
Screen Shot 2021-11-14 at 16 45 53

Screen Shot 2021-11-15 at 11 21 36

### DIFF
--- a/webui/package.json
+++ b/webui/package.json
@@ -11,9 +11,11 @@
     "@primer/octicons-react": "^13.0.0",
     "bootstrap": "^4.6.0",
     "http-proxy-middleware": "^1.1.1",
+    "jquery-csv": "^1.0.21",
     "moment": "^2.29.1",
     "react": "^17.0.0",
     "react-bootstrap": "^1.5.2",
+    "react-diff-viewer": "^3.1.1",
     "react-dom": "^17.0.0",
     "react-router-dom": "^5.2.0"
   },
--- a/webui/package.json
+++ b/webui/package.json
@@ -11,7 +11,6 @@
     "@primer/octicons-react": "^13.0.0",
     "bootstrap": "^4.6.0",
     "http-proxy-middleware": "^1.1.1",
-    "jquery-csv": "^1.0.21",
     "moment": "^2.29.1",
     "react": "^17.0.0",
     "react-bootstrap": "^1.5.2",
--- a/webui/src/lib/api/index.js
+++ b/webui/src/lib/api/index.js
@@ -502,28 +502,15 @@ class Objects {
         }
     }
 
-    async getWithSizeConstraint(repoId, ref, path, maxObjectSizeBytes) {
-        const response = await objects.get(repoId, ref, path);
-        const headers = response.headers;
-        const size = + headers.get("Content-Length") || 0
-        if (size > maxObjectSizeBytes) {
-            throw new Error(path + " is too big (> 20KB). To view its diff please download the objects and use an " +
-                "external diff tool.");
-        }
-        const content = await response.text();
-        return [content, headers];
-    }
-
-    async get(repoId, ref, path, additionalHeaders) {
+    async get(repoId, ref, path) {
         const query = qs({path});
         const response = await apiRequest(`/repositories/${repoId}/refs/${ref}/objects?${query}`, {
             method: 'GET',
-            headers: new Headers(additionalHeaders)
         });
         if (response.status !== 200) {
             throw new Error(await extractError(response));
         }
-        return response
+        return response.text()
     }
 
     async getStat(repoId, ref, path) {
--- a/webui/src/lib/api/index.js
+++ b/webui/src/lib/api/index.js
@@ -501,6 +501,38 @@ class Objects {
             throw new Error(await extractError(response));
         }
     }
+
+    async getWithSizeConstraint(repoId, ref, path, maxObjectSizeBytes) {
+        const response = await objects.get(repoId, ref, path);
+        const headers = response.headers;
+        if (headers.get("Content-length") > maxObjectSizeBytes) {
+            throw new Error(path + " is too big (> 20KB). To view its diff please download the objects and use an " +
+                "external diff tool.");
+        }
+        const content = await response.text();
+        return [content, headers];
+    }
+
+    async get(repoId, ref, path, additionalHeaders) {
+        const query = qs({path});
+        const response = await apiRequest(`/repositories/${repoId}/refs/${ref}/objects?${query}`, {
+            method: 'GET',
+            headers: new Headers(additionalHeaders)
+        });
+        if (response.status !== 200) {
+            throw new Error(await extractError(response));
+        }
+        return response
+    }
+
+    async getStat(repoId, ref, path) {
+        const query = qs({path});
+        const response = await apiRequest(`/repositories/${repoId}/refs/${ref}/objects/stat?${query}`);
+        if (response.status !== 200) {
+            throw new Error(await extractError(response));
+        }
+        return response.json()
+    }
 }
 
 class Commits {
--- a/webui/src/lib/api/index.js
+++ b/webui/src/lib/api/index.js
@@ -505,7 +505,8 @@ class Objects {
     async getWithSizeConstraint(repoId, ref, path, maxObjectSizeBytes) {
         const response = await objects.get(repoId, ref, path);
         const headers = response.headers;
-        if (headers.get("Content-length") > maxObjectSizeBytes) {
+        const size = + headers.get("Content-Length") || 0
+        if (size > maxObjectSizeBytes) {
             throw new Error(path + " is too big (> 20KB). To view its diff please download the objects and use an " +
                 "external diff tool.");
         }
--- a/webui/src/lib/components/repository/ObjectsDiff.jsx
+++ b/webui/src/lib/components/repository/ObjectsDiff.jsx
@@ -1,0 +1,167 @@
+import React from "react";
+import {useAPI} from "../../hooks/api";
+import {objects} from "../../api";
+import ReactDiffViewer, {DiffMethod} from "react-diff-viewer";
+import {Error, Loading} from "../controls";
+
+export const ObjectsDiff = ({diffType, repoId, leftRef, rightRef, path}) => {
+    switch (diffType) {
+        case 'changed':
+            return <ChangedDiff repoId={repoId} leftRef={leftRef} rightRef={rightRef} path={path}/>;
+        case 'added':
+            return <AddedDiff repoId={repoId} rightRef={rightRef} path={path}/>;
+        case 'removed':
+            return <RemovedDiff repoId={repoId} leftRef={leftRef} path={path}/>;
+        case 'conflict':
+            return <ConflictDiff path={path}/>;
+        default:
+            return <Error error={"Unsupported diff type " + diffType}/>;
+    }
+}
+
+// github shows 20KB diff according to https://github.blog/2016-12-06-how-we-made-diff-pages-3x-faster/
+const maxDiffSizeBytes = 20000;
+const supportedReadableFormats = ["txt", "csv", "tsv"];
+
+function readableObject(path) {
+    for (let i = 0; i < supportedReadableFormats.length; i++) {
+        if (path.endsWith(supportedReadableFormats[i])) {
+            return true;
+        }
+    }
+    return false;
+}
+
+const ConflictDiff = ({path}) => {
+    return "Conflict in " + path + " resolve the conflict manually to review objects diff.";
+}
+
+const ChangedDiff =  ({repoId, leftRef, rightRef, path}) => {
+    const readable = readableObject(path);
+    const getFn = readable ? objects.getWithSizeConstraint : objects.getStat;
+    const left = useAPI(async () => getFn(repoId, leftRef, path, maxDiffSizeBytes),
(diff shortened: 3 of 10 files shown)


webui/package.json
```diff
@@ -11,9 +11,11 @@
     "@primer/octicons-react": "^13.0.0",
     "bootstrap": "^4.6.0",
     "http-proxy-middleware": "^1.1.1",
+    "jquery-csv": "^1.0.21",
```
nopcoder, line 14: are we using this one?

talSofer: no, good catch. will remove it

webui/src/lib/components/repository/ObjectsDiff.jsx
```diff
@@ -0,0 +1,167 @@
...
+            return <Error error={"Invalid input: left is undefined"}/>;
+        }
+        statsReportBlock = <StatsReportBlock leftSize={left.size_bytes} diffType={diffType}/>;
+    }
+
+    return <>
+        <div className={"stats-diff-block"}>
+            {statsReportBlock}
+        </div>
+    </>;
```
nopcoder, line 130: something like that can make it more readable - we can pass a property when it is undefined as long as the component addresses it when needed.
```suggestion
     switch (diffType) {
        case 'changed':
            if (!left && !right) return <Error error={"Invalid input"}/>;
        case 'added':
            if (!left) return <Error error={"Invalid input: left is missing"}/>;
        case 'removed':
            if (!right) return <Error error={"Invalid input: right is missing"}/>;
        default:
            return <Error error={"Unknown diff type: " + diffType}/>;
    }
    const rightSize = right && right.size_bytes;
    const leftSize = left && left.size_bytes;
    return <>
            <div className={"stats-diff-block"}>
                <StatsReportBlock leftSize={leftSize} rightSize={rightSize} diffType={diffType}/>
            </div>
           </>;
```

webui/src/lib/components/repository/changes.jsx
```diff
@@ -52,6 +54,8 @@ const ChangeRowActions = ({ entry, onRevert }) => {
     entry: The entry the TreeItem is representing, could be either an object or a prefix.
     repo: Repository
     reference: commitID / branch
+    compareReference: commitID / branch
+    displayedAt: the page in which the TreeItem is displayed. e.g. changes, commit, or compare pages.
```
nopcoder, line 58: order it based on the function arguments order

webui/src/lib/components/repository/changes.jsx
```diff
@@ -114,15 +126,17 @@ export const TreeEntryRow = ({ entry, showActions, relativeTo="", leaf=false, ex
         <tr className={rowClass} >
             <td className="diff-indicator">{diffIndicator}</td>
             <td className="tree-path">
-                <span style={{marginLeft: depth * 20 + "px"}}>
-                    {leaf ? "" :
+                {
```
nopcoder, line 129: why do we need '{' inside the html block?

webui/src/lib/components/repository/changes.jsx
```diff
@@ -114,15 +126,17 @@ export const TreeEntryRow = ({ entry, showActions, relativeTo="", leaf=false, ex
...
+                    <span style={{marginLeft: depth * 20 + "px"}}>
```
nopcoder, line 130: style - preferred in css

webui/src/lib/components/repository/changes.jsx
```diff
@@ -156,6 +170,57 @@ export const TreeEntryPaginator = ({ path, setAfterUpdated, nextPage, depth=0, l
...
+            return [committedRef, uncommittedRef]
+        case ("commit"):
+            return [reference.parents[0], reference.id]
+        case ("compare"):
+            if (compareReference === null) {
+                return <Error error={"Missing a comparison reference"}/>
+            }
+            return [reference.id, compareReference.id]
+        default:
+            return <Error error={"Diff is unsupported for view " + displayedAt}/>
```
nopcoder, line 202: case code style
```suggestion
        case "changes":
            const committedRef = reference.id + "@"
            const uncommittedRef = reference.id
            return [committedRef, uncommittedRef]
        case "commit":
            return [reference.parents[0], reference.id]
        case "compare":
            if (compareReference === null) {
                return <Error error={"Missing a comparison reference"}/>
            }
            return [reference.id, compareReference.id]
        default:
            return <Error error={"Diff is unsupported for view " + displayedAt}/>
```

webui/src/lib/components/repository/changes.jsx
```diff
@@ -156,6 +170,57 @@ export const TreeEntryPaginator = ({ path, setAfterUpdated, nextPage, depth=0, l
...
+    }
+}
+
+const ExpandedLeafRow = ({entry, repoId, reference, compareReference, displayedAt, loading}) => {
+    let [left, right] = getDiffRefs(reference, displayedAt, compareReference)
```
nopcoder, line 207:
```suggestion
    const [left, right] = getDiffRefs(reference, displayedAt, compareReference);
```

webui/src/lib/components/repository/changes.jsx
nopcoder: display at - problematic property. it bound the view to the application business. is there a way to transform it to a visual property/properties we can pass instead of saying the place we render it.

webui/src/lib/components/repository/ObjectsDiff.jsx
```diff
@@ -0,0 +1,167 @@
...
+    if (!!left.error) return <Error error={left.error}/>;
+
+    return readable
+        ? <ContentDiff left={left.response[0]} right={right.response[0]} leftSize={(left.response[1]).get("Content-length")}
+                       rightSize={(right.response[1]).get("Content-length")} diffType={"changed"}/>
+        : <StatDiff left={left.response} right={right.response} diffType={"changed"}/>;
+}
+
+const AddedDiff =  ({repoId, rightRef, path}) => {
+    let readable = readableObject(path);
```
johnnyaug, line 58: A lot of code is duplicated between those components. 
In the end all of them (except conflict) return a ContentDiff or a StatDiff. Consider merging these into one component.

johnnyaug (changes requested):
Great work! Added some comments.

Some product notes:
* I suggest using an icon that looks more "icony" like image or image. Right now it looks like it's not clickable. 
* Also change the cursor to a pointer when hovering over the icon, and add a tooltip of "Show changes".
* In the case where the files are too large, I would strongly prefer if we could avoid the "more than 20kb" error. Instead, the button should be greyed out and the tooltip should indicate the file is too big.  I understand that this may require an API change (to return the size of both sides), so this can be handled in a different issue.
* For the "added"/"removed" in size, can use GitHub-style component image with a tooltip.
* On the same subject, the words "added"/"removed" are a bit confusing since the displayed change is only in the size. I would change the text to "+7B size" or "-17B size" to indicate this.

webui/src/lib/components/repository/ObjectsDiff.jsx
```diff
@@ -0,0 +1,138 @@
...
+            right = useAPI(async () => objects.getStat(repoId, rightRef, path),
+                [repoId, rightRef, path]);
+            break;
+        case 'removed':
+            left = useAPI(async () => objects.getStat(repoId, leftRef, path),
+                [repoId, leftRef, path]);
+            break;
+        case 'conflict':
+            return <>{path} has a conflict. Please resolve the conflict manually and then review objects diff.</>;
+            break;
```
nopcoder, line 33: minor - can be removed

webui/src/lib/components/repository/ObjectsDiff.jsx
```diff
@@ -0,0 +1,138 @@
...
+        default:
+            return <Error error={"Unknown diff type: " + diffType}/>;
+            break;
+    }
+    return <div>
+                {diffVerbSpan}
+                <span className={"diff-size"}>{size}</span>
+                <span> in size</span>
+            </div>;
```
nopcoder, line 137: suggestion if it make it clear
```suggestion
    let label = diffType;
    let size;
    switch (diffType) {
        case 'changed':
            const sizeDiff = leftSize - rightSize;
            if (sizeDiff < 0) {
                size = -sizeDiff;
                label = "added";
            } else {
                size = sizeDiff;
                label = "removed";
            }
            break;
        case 'added':
            size = rightSize;
            break;
        case 'removed':
            size = leftSize;
            break;
        default:
            return <Error error={"Unknown diff type: " + diffType}/>;
    }
    return <div>
                <span className={label}>{label} </span>
                <span className={"diff-size"}>{humanSize(size)}</span>
                <span> in size</span>
            </div>;
```

webui/src/pages/repositories/repository/changes.jsx
```diff
@@ -270,14 +270,21 @@ const ChangesBrowser = ({repo, reference, prefix, onSelectRef, }) => {
...
                                 <tbody>
-                                {results.map(entry => (
-                                    <TreeItem key={entry.path + "-tree-item"} entry={entry} repo={repo} reference={reference}
-                                              internalReferesh={internalRefresh} onNavigate={onNavigate}
-                                              onRevert={onRevert} delimiter={delimiter} relativeTo={prefix}
+                                {results.map(entry => {
+                                    const committedRef = reference.id + "@"
+                                    const uncommittedRef = reference.id
+                                    return (
+                                    <TreeItem key={entry.path + "-tree-item"} entry={entry} repo={repo}
```
nopcoder, line 277: indent

nopcoder (approved):
looks good - minor comments

webui/package.json
```diff
@@ -11,9 +11,11 @@
...
     "moment": "^2.29.1",
     "react": "^17.0.0",
     "react-bootstrap": "^1.5.2",
+    "react-diff-viewer": "^3.1.1",
```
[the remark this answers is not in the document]
nopcoder: reinstall the required package and it will update the lock file.

webui/src/lib/api/index.js
```diff
@@ -501,6 +501,38 @@ class Objects {
...
+        return [content, headers];
+    }
+
+    async get(repoId, ref, path, additionalHeaders) {
+        const query = qs({path});
+        const response = await apiRequest(`/repositories/${repoId}/refs/${ref}/objects?${query}`, {
+            method: 'GET',
+            headers: new Headers(additionalHeaders)
+        });
+        if (response.status !== 200) {
```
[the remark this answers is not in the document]
nopcoder: sure - keep it aligned.

webui/src/lib/components/repository/ObjectsDiff.jsx
nopcoder: so you need also one for the default :)

webui/package.json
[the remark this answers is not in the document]
talSofer: Done. can you please take a look?

webui/src/lib/components/repository/changes.jsx
```diff
@@ -52,6 +54,8 @@ const ChangeRowActions = ({ entry, onRevert }) => {
     entry: The entry the TreeItem is representing, could be either an object or a prefix.
     repo: Repository
     reference: commitID / branch
+    leftDiffRefID: commitID / branch
```
nopcoder, line 57: this line and the next indent one space back

nopcoder (approved):

johnnyaug (approved):
Great work!

PR #2688, opened 2021-11-10
Use int64 for StreamingReader size

arielshaqed:
It *can* be used on the API for a large streaming upload.  That will be slow
but must still give a correct result.  (Go has much 64-bit confusion, say in
the `len(...)` return type.)


guy-har (approved):
Thanks!

Issue #2684, opened 2021-11-10
Upgrade to AWS Go SDK v2


arielshaqed:
This is now supported
https://docs.aws.amazon.com/sdkref/latest/guide/version-support-matrix.html.

[Developer guide](https://aws.github.io/aws-sdk-go-v2/docs/) has a section
on [migrating](https://aws.github.io/aws-sdk-go-v2/docs/migrating/); it
appears most of the work will be in client creation.

A significant change for us right now is the addition of a [`ContextLogger`
interface](https://pkg.go.dev/github.com/aws/smithy-go/logging#ContextLogger).
We can use that to help with #2682.

nopcoder:
In favor to upgrade, more information about v2: https://aws.amazon.com/blogs/developer/aws-sdk-for-go-version-2-general-availability/

arielshaqed:
Against the upgrade is that the new version might remove some capabilities in a bid to simplify the interface.  Specifically it will be hard to write `s3Uploader`, which is a chunked uploader.

Also...
## :warning: BEWARE OF USING `manager.Uploader` :warning: 

This is dangerous.  It was dangerous, but it still is. aws/aws-sdk-go-v2#1040 is that files large enough to need multipart uploads _do not have their digests checked_.

Now it might seem the if you access S3 only via https then TLS checks everything for you.  But (of course) that reduces the issue but does not eliminate it.  If gremlins reside after the TLS endpoint and modify the data, it is entirely unchecked.  This is a significant path along multiple machines in the data centre, so is not safe.

arielshaqed:
Unassigning myself and keeping it on the backlog: no current plans for this because the v1 SDK still works for us.  Keeping open because I _still_ want my ContextLogger, it can greatly improve debugging in some cases with heavy load.

PR #2693, opened 2021-11-11
Remove "lakefs diagnose" and adapter ValidateConfiguration

arielshaqed:
This was used only to validate that S3 buckets are configured with
expiration suited for the disused expiry mechanism.  It appears never to
have been used.

This refactoring (in the sense of pushing it into a river) makes #2684
easier.


arielshaqed:
Reviewers: @ozkatz for product (to verify this is OK to kill), @nopcoder to be happy that I sometimes delete code and that he has no long-term plans for this command and/or block adapter API.

arielshaqed:
@ozkatz says "kill it with fire", so is off the hook for reviewing.  Thanks!

nopcoder (commented):
👍 Less code

nopcoder:
just re-generate the commands docs

arielshaqed (commented):
Thanks!

nopcoder (approved):

PR #2696, opened 2021-11-11
block adapter remove dependency in s3 for multipart struct

nopcoder:
Waiting for #2684 - do not merge

### DIFF
--- a/pkg/block/adapter.go
+++ b/pkg/block/adapter.go
@@ -9,7 +9,7 @@ import (
 // MultipartPart single multipart information
 type MultipartPart struct {
 	ETag       string
-	PartNumber int64
+	PartNumber int
 }
 
 // MultipartUploadCompletion parts described as part of complete multipart upload. Each part holds the part number and ETag received while calling part upload.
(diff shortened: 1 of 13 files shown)


pkg/block/adapter.go
```diff
@@ -4,15 +4,21 @@ import (
...
 	"io"
 	"net/http"
-
-	"github.com/aws/aws-sdk-go/service/s3"
 )
 
+// MultipartPart single multipart information
+type MultipartPart struct {
+	ETag       string
+	PartNumber int64
```
arielshaqed, line 12: Any point in `int64` here?  The [docs](https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html#mpu-process) say it's a number between 1 and 10_000...

pkg/block/s3/adapter.go
```diff
@@ -563,20 +563,30 @@ func (a *Adapter) AbortMultiPartUpload(ctx context.Context, obj block.ObjectPoin
 	return err
 }
 
+func convertFromBlockMultipartUploadCompletion(multipartList *block.MultipartUploadCompletion) *s3.CompletedMultipartUpload {
+	var parts []*s3.CompletedPart
```
arielshaqed, line 567:
```suggestion
	parts := make([]*s3.CompletedPart, 0, len(multipartList.Part))
```

arielshaqed (approved):
Great cleanup, thanks!

pkg/block/adapter.go
nopcoder: no point - kept the AWS types - but it will be easier to use int, we do not use code-generators yet.

nopcoder:
Transformed adapter to work with `int` instead of `int64` for multipart

arielshaqed (commented):
Cool, stillLGM...

Issue #2699, opened 2021-11-13
lakectl merge does not provide good information in case of conflict 


sthiriet:
Hello,

I have seen that in case of conflict, lakectl is not very helpful : the count is not accurate and no error message or return code != 0 is provided to the user :

```shell
$ lakectl merge lakefs://test-lakefs/modify-lb1 lakefs://test-lakefs/main 
Source: lakefs://test-lakefs/modify-lb1
Destination: lakefs://test-lakefs/main
Conflicts: 0
$ echo $?
0
```

The UI is more helpful :
![image](https://user-images.githubusercontent.com/12430903/141657690-ad889723-c020-4c7d-bd74-fc05474bcb7f.png)

It could be annoying for automations.

nopcoder:
lakectl will print "Conflict found." and exit code will be 1.

sthiriet:
Perfect! Thank you.

PR #2700, opened 2021-11-13
Propagate the merge result from graveler to the controller level

nopcoder:
Part of #2699

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -2716,7 +2716,7 @@ func (c *Controller) MergeIntoBranch(w http.ResponseWriter, r *http.Request, bod
 		writeError(w, http.StatusPreconditionFailed, err)
 		return
 	case errors.Is(err, catalog.ErrConflictFound) || errors.Is(err, graveler.ErrConflictFound):
-		writeError(w, http.StatusConflict, newMergeResultFromCatalog(res))
+		writeResponse(w, http.StatusConflict, newMergeResultFromCatalog(res))
 		return
 	}
 	if handleAPIError(w, err) {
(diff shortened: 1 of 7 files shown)


pkg/db/database.go
```diff
@@ -190,29 +190,29 @@ func (d *PgxDatabase) Transact(ctx context.Context, fn TxFunc, opts ...TxOpt) (i
 			rollbackErr := tx.Rollback(ctx)
 			if rollbackErr != nil {
 				// returning the original error and not the rollbackErr
-				return nil, err
+				return ret, err
 			}
 			// retry on serialization error
-			if IsSerializationError(err) {
+			if IsSerializationError(rollbackErr) {
```
guy-har, line 196:
```suggestion
			if IsSerializationError(err) {
```

pkg/db/database.go
```diff
@@ -190,29 +190,29 @@ func (d *PgxDatabase) Transact(ctx context.Context, fn TxFunc, opts ...TxOpt) (i
 			rollbackErr := tx.Rollback(ctx)
 			if rollbackErr != nil {
 				// returning the original error and not the rollbackErr
-				return nil, err
+				return ret, err
```
guy-har, line 193: Add a comment explaining the reason for returning ret

pkg/api/controller.go
```diff
@@ -2716,7 +2716,7 @@ func (c *Controller) MergeIntoBranch(w http.ResponseWriter, r *http.Request, bod
 		writeError(w, http.StatusPreconditionFailed, err)
 		return
 	case errors.Is(err, catalog.ErrConflictFound) || errors.Is(err, graveler.ErrConflictFound):
-		writeError(w, http.StatusConflict, newMergeResultFromCatalog(res))
```
guy-har, line 2719: Consider checking if we can change the writeError signature to work only on errors

guy-har (approved):

pkg/api/controller.go
nopcoder: Checked this option - but there are many cases that we serialize the different messages and not errors when we send a status code.

PR #2702, opened 2021-11-14
lakeFS Hadoop Filesystem to support Hadoop CLI

### LINKED ISSUE #2629: Check lakeFS Hadoop Filesystem with Hadoop CLI

itaiad200:
Fixes #2629

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -770,6 +770,8 @@ public class LakeFSFileSystem extends FileSystem {
 
         if (statuses.length > 0 &&
                 Arrays.stream(statuses).noneMatch(status -> status.getPath().equals(pathPattern))){
+            // If files were found but none equal the pattern, it means that it's an existing directory
+            // with files in it. In that case return the single status which represent that.
             return new FileStatus[]{new LakeFSFileStatus.Builder(pathPattern).isdir(true).build()};
         }
 
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -759,8 +759,7 @@ public class LakeFSFileSystem extends FileSystem {
 
     @Override
     public FileStatus[] globStatus(Path pathPattern) throws IOException {
-        // TODO: implement this method as part of https://github.com/treeverse/lakeFS/issues/2629
-        OPERATIONS_LOG.info("pathPattern({})", pathPattern);
+        OPERATIONS_LOG.trace("globStatus({})", pathPattern);
 
         FileStatus[] statuses;
         try{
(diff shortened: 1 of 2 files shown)


clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -759,10 +759,22 @@ public String getScheme() {
...
     @Override
     public FileStatus[] globStatus(Path pathPattern) throws IOException {
-        // TODO: implement this method as part of https://github.com/treeverse/lakeFS/issues/2629
-        FileStatus fStatus = new FileStatus(0, false, 1, 20, 1,
-                new Path("globStatus-test-path"));
-        return new FileStatus[]{fStatus};
+        OPERATIONS_LOG.trace("globStatus({})", pathPattern);
+
+        FileStatus[] statuses;
+        try{
```
arielshaqed, line 765: nit:
```suggestion
        try {
```

arielshaqed (changes requested):
1. How was this tested?

1. Does it would with [`count`](https://hadoop.apache.org/docs/r2.4.1/hadoop-project-dist/hadoop-common/FileSystemShell.html#count)?

1. Does the Hadoop FS spec test include something to test for this?  If so then we should ensure that it runs.

1. I don't understand how this implements [`globStatus`](https://hadoop.apache.org/docs/r2.7.1/api/org/apache/hadoop/fs/FileSystem.html#globStatus(org.apache.hadoop.fs.Path)) (and indeed IIUC the comment on the PR says that it does not).  If we provide only partial support, then I guess we would need to throw some "not implemented" exception when `path` contains any wildcards, rather than return an incorrect answer.

nopcoder (commented):
Looking at 2.7.x s3a implementation - https://github.com/apache/hadoop/blob/release-2.7.1/hadoop-tools/hadoop-aws/src/main/java/org/apache/hadoop/fs/s3a/S3AFileSystem.java it doesn't override this method.
Looking at the base implementation https://github.com/apache/hadoop/blob/release-2.7.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/FileSystem.java#L1654
It uses a Globber implementation https://github.com/apache/hadoop/blob/release-2.7.1/hadoop-common-project/hadoop-common/src/main/java/org/apache/hadoop/fs/Globber.java that calls the fs implementation to support the functionality.
Can you identify why the base implementation doesn't work for us? maybe there is a missing API or missed returning some data that will enable us this functionality?

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
itaiad200: Code deleted.

itaiad200:
Following @nopcoder & @arielshaqed advices, I checked again the base implementation.

What I assumed didn't work before, is actually in parity with the s3a behaviour. Therefore implementing the `globPattern` method is not needed and it's ok to rely on the basic implementation that will make the appropriate FS calls.

Was tested using FileSystemTracer that compares the behaviour of s3a and lakeFS FS with the Hadoop CLI which was the tool that initiated this bug (no other tool is calling the `globPattern`).

arielshaqed (approved):
This looks exactly right, only it remains untested.  I am approving as it is clearly better than what we have.  But I would be **really** happy if you could find & enable some test in the spec, and/or add a test of our own.

Thanks!

PR #2705, opened 2021-11-14
s3 block adapter option to disable bucket discovery

### LINKED ISSUE #2697: Configure S3 block adapter to disable region discovery
Current block adapter try to discover the bucket region by calling GetBucketRegion API.
Not all S3 compatible storage system supports this API - example: MinIO.

In order not to keep trying to get the region and fill the log with errors, we should enable an option to disable it and guide these underlying systems to disable this feature.

There current S3 block configuration holds property for 'region' that will be used.

nopcoder:
Fix #2697

### DIFF
--- a/docs/deploy/aws.md
+++ b/docs/deploy/aws.md
@@ -49,7 +49,7 @@ If you already have a database, take note of the connection string and skip to t
    blockstore:
      type: s3
      s3:
-       region: us-east-1
+       region: us-east-1 # optional, fallback in case discover from bucket is not supported
    ```
 
 1. [Download the binary](../index.md#downloads) to the EC2 instance.
--- a/docs/deploy/k8s.md
+++ b/docs/deploy/k8s.md
@@ -45,7 +45,7 @@ lakefsConfig: |
     blockstore:
       type: s3
       s3:
-        region: us-east-1
+        region: us-east-1 # optional, fallback in case discover from bucket is not supported
 ```
    </div>
    <div markdown="1" id="helm-tabs-2">
--- a/docs/integrations/minio.md
+++ b/docs/integrations/minio.md
@@ -21,6 +21,7 @@ blockstore:
   s3:
     force_path_style: true
     endpoint: http://<minio_endpoint>:9000
+    discover_bucket_region: false
     credentials:
       access_key_id: <minio_access_key>
       secret_access_key: <minio_secret_key>
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -161,7 +161,7 @@ auth:
 blockstore:
   type: s3
   s3:
-    region: us-east-1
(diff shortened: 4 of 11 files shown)


talSofer (commented):
Looks good! 
What about documenting this? Is the list on https://docs.lakefs.io/reference/configuration.html#reference generated or manually updated?

nopcoder:
> Looks good! What about documenting this? Is the list on https://docs.lakefs.io/reference/configuration.html#reference generated or manually updated?

good call - added the parameter reference.

nopcoder:
> Looks good! What about documenting this? Is the list on https://docs.lakefs.io/reference/configuration.html#reference generated or manually updated?

manual update, just commands.md is auto-generated.

talSofer (commented):
More thoughts about docs - 
- Should we include this param in the MinIO integration page https://docs.lakefs.io/integrations/minio.html? 
- Should we scan the rest of the docs and clear references to the region param and either remove them or add a note about this new flag and mention that the two work together? for example in https://docs.lakefs.io/deploy/aws.html#on-ec2.

nopcoder:
Updated the docs - minio and additional comments when region is used in the configuration.
We probably will need to address the gateway "region" configuration documentation too - but it will be not in this PR.

docs/deploy/aws.md
```diff
@@ -49,7 +49,7 @@ If you already have a database, take note of the connection string and skip to t
    blockstore:
      type: s3
      s3:
-       region: us-east-1
+       region: us-east-1 # optional, fallback in case discover from bucket is not supported
```
arielshaqed, line 52: This can be confusing for AWS S3 users.  Maybe
```suggestion
       # Uncomment on non-AWS S3 compatible implementations which do not support GetBucketLocation (e.g. MinIO).
       # region: us-east-1
```

arielshaqed (approved):
Neat!

I guess this will be too hard to test meaningfully.  But can you at least run it once with and without, and examine the logs?  The plumbing from configuration to execution is somewhat long.

Also not sure why specify a region (in documentation) when we *do* have discovery.  But fixing these docs can wait for another PR.

talSofer (approved):

docs/deploy/aws.md
nopcoder: In minio we will also need to add `discover_bucket_region: false`, maybe it will be better off without it.

nopcoder:
> Neat!
> 
> I guess this will be too hard to test meaningfully. But can you at least run it once with and without, and examine the logs? The plumbing from configuration to execution is somewhat long.
> 
> Also not sure why specify a region (in documentation) when we _do_ have discovery. But fixing these docs can wait for another PR.

Tested with and without discovery in local MinIO.

docs/deploy/aws.md
nopcoder: Please feedback here and I'll update the docs in a different PR.

Issue #2499, opened 2021-09-30
Add unit tests to lakectl


itaidavid:
Currently no unit tests at all
Not to be confiused with #1657

PR #2711, opened 2021-11-15
lakectl testing design

itai-david:
UPDATE: This PR was copied from a private repo, where it was mistakenly created, in order to make it publicly accessible
Comments from the original PR are copied, with the commentor mentioned.

Original description:
>An initial design for lakectl testing framework
>The document raises 2 questions:
>* Should we test lakectl as binary, or as code using golang tests
>* Should we bring up a live lakefs server or use mock(s)
>Comments and opinions are mostly welcome


design/lakectl-testing.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Architecture Outline
+
+![lakectl testing architecture](./diagrams/lakectl-testing-arch.png)
+
+### Test Case Structure
+
+The test case structure should contain at least the following details:
+* pre-condition: the system state right before the tested command starts. This can be a function name to call, a mock type to use or a file containing lakectl commands to run in order to bring the system to that state (in case real server is in use)
+* lakectl command: a string with the `lakectl` command to test.
```
itai-david, line 37: Commented by @itaiad200 
Must be just a single command? What's the downside of writing a more elaborated test that covers more than a single command?

design/lakectl-testing.md
```diff
@@ -0,0 +1,102 @@
...
+* expected result: the expected output of a correct execution of the above command. This can be a string or, better, a file containing the output string
```
itai-david, line 38: Commented by @itaiad200 
I think that the expected result should also contain the request(s) that were sent to the server. I know it depends on the architecture that is described below, but many of the lakectl commands have no output at all, and other may display the correct output without performing the right actions.

design/lakectl-testing.md
```diff
@@ -0,0 +1,102 @@
...
+
+On the upside:
+* Straightforward - the implementation is already there, just bring the server up.
+  * Moreover, `lakefs` modifications are immediately reflected into the tests
+* The best possible simulation - this is the real server `lakectl` is communicating with
+
+And then, on the downside:
+* Bringing up the server requires additional dependencies - underlying storage, DB etc. While this supports the advantage of testing a **real** server environment, it complicates the test setup, and although this is easily solved by containers, it still worth mentioning
+* Performance-wise, creating a server environment for each test (thousands of them, ideally) is very expensive. We can, of course, group several tests (all of them?) to use a single server session, but this will eliminate the isolative and stateless nature of the tests, which is undesired
+* Flexibility is limited to `lakefs` server implementation. This means that testing `lakectl` behavior, with improbable response from `lakefs` server (e.g. timeout, disconnection etc.) becomes very difficult
```
itai-david, line 56: Commented by @itaiad200 
Kind of means that this test framework also becomes sort of a system test for lakeFS, which I believe we should be ok with.

design/lakectl-testing.md
```diff
@@ -0,0 +1,102 @@
...
+#### **Live Server**
+
+On the upside:
+* Straightforward - the implementation is already there, just bring the server up.
+  * Moreover, `lakefs` modifications are immediately reflected into the tests
+* The best possible simulation - this is the real server `lakectl` is communicating with
+
+And then, on the downside:
+* Bringing up the server requires additional dependencies - underlying storage, DB etc. While this supports the advantage of testing a **real** server environment, it complicates the test setup, and although this is easily solved by containers, it still worth mentioning
+* Performance-wise, creating a server environment for each test (thousands of them, ideally) is very expensive. We can, of course, group several tests (all of them?) to use a single server session, but this will eliminate the isolative and stateless nature of the tests, which is undesired
```
itai-david, line 55: _Commented by @itaiad200_
Yeah, true isolation sucks. The way we achieve isolation in Nessie is by creating a different repo for each test-case. It's not 100% isolated like running a fresh server, but it's a decent tradeoff.

design/lakectl-testing.md
```diff
@@ -0,0 +1,102 @@
...
+* No need for initial state setup steps
+* Server bugs are not included
+
+Disadvantages:
+* Requires implementation - unlike `lakefs` server, we currently don't have a mock
+* Requires (a lot of) maintenance - every new feature and every bug fix in `lakefs` may, and probably will, need to be reflected in the mock
+
+#### Conclusion ?
+
+For the long term it look like the better approach is to use a full scale mock, which should evolve as tests are created. While this will certainly have an impact on new tests creation ease, the benefits of lack independency on `lakefs` are too important to ignore, especially when dealing with a young product such as `lakefs` which may be subdued to significant changes in the future
```
itai-david, line 74: _Commented by @itaiad200_
After reading the 2 options, I ended up with a different conclusion smile
IMO a full scale mock will end up with writing a parallel mocked lakeFS implementation, which will be hard to maintain, expand and debug. Running a real lakeFS instance means that we would be testing it too, which from the scope of lakectl testing sounds like a downside, but lakectl always interacts with a lakeFS server, so we really need to test the complete lakectl -> lakeFS flow.

design/lakectl-testing.md
```diff
@@ -0,0 +1,102 @@
...
+#### Testing `lakectl` Code with Golang Tests
+
+Pros:
+* As simple as it gets - use golang testing framework on function `main()`
+* No need to build `lakectl` binary as part of the test
+* Easy to debug - just use your favorite IDE/debugger
+
+Cons:
+* Everything that happens before `main()` is lost (process start etc.)
+* Might be challenging to create the same env as the process
```
itai-david, line 98: _Commented by @itaiad200_
You can also kind of have both with cmd execution and golang code:

    	  os.Setenv("FOO", "bar")
	  out, err := exec.Command("lakectl", "repo", "list").Output()
	  if err != nil {
		log.Fatal(err)
	  }
          // output validations ...

itai-david: _Responded by @johnnyaug_
This is not having both: this tests the binary

design/lakectl-testing.md
```diff
@@ -0,0 +1,102 @@
...
+This document does not specify the cli tests to run, but the system intended to run them
+
+## Requirements
+* Usable both locally and as part of CI/CD
+* Ease of use - adding new tests should be as easy as possible. Moreover, adding new tests, or modifying existing ones, should not draw modifications to the testing framework itself
+  * It is obvious that changing the test **format** (e.g. expanding) will require framework modifications, in order to support the modified format
+* Decoupling of Framework and Data - the test data format should contain all the required info for the test itself, regardless of the framework design. In other words, we would like the ability to rewrite/replace the framework and still use the same tests data
+*
+
+## What We Want To Test
```
itai-david, line 16: _Commented by @nopcoder_
how about output testing, so we like to verify specific format, colors and etc?

design/lakectl-testing.md
```diff
@@ -0,0 +1,102 @@
...
+For the long term it look like the better approach is to use a full scale mock, which should evolve as tests are created. While this will certainly have an impact on new tests creation ease, the benefits of lack independency on `lakefs` are too important to ignore, especially when dealing with a young product such as `lakefs` which may be subdued to significant changes in the future
+
+### How should we test `lakectl`
+
+While the decision on `lakefs` usage above, has many significant aspects, the 2 approaches on how to test `lakectl` - as a binary or as code - seems to have much lesser impact. Here are some of the advantages and disadvantages of both
+
+#### Testing `lakectl` Binary
+
+Pros:
+* Tests are performed on the real binary and includes all the surroundings, such as env variables
```
itai-david, line 83: _Commented by @nopcoder_
* platform issues

design/lakectl-testing.md
```diff
@@ -0,0 +1,102 @@
...
+* Allows usage of 3rd party testing tools, that were designed to test CLIs
+
+Cons:
+* May be hard to debug a failed test
```
itai-david, line 87: _Commented by @nopcoder_
Depends - sometimes it is very easy that you can your command line that didn't print out the expected result. You just need to execute it again to debug what failed.

design/lakectl-testing.md
```diff
@@ -0,0 +1,102 @@
...
+
+#### Conclusion ?
+
+For the long term it look like the better approach is to use a full scale mock, which should evolve as tests are created. While this will certainly have an impact on new tests creation ease, the benefits of lack independency on `lakefs` are too important to ignore, especially when dealing with a young product such as `lakefs` which may be subdued to significant changes in the future
+
+### How should we test `lakectl`
+
+While the decision on `lakefs` usage above, has many significant aspects, the 2 approaches on how to test `lakectl` - as a binary or as code - seems to have much lesser impact. Here are some of the advantages and disadvantages of both
+
+#### Testing `lakectl` Binary
```
itai-david, line 80: _Commented by @nopcoder_
Note you can still use golang testing framework and test binary - https://github.com/google/go-cmdtest

design/lakectl-testing.md
itai-david: That is an important insight!
I'm trying to think how can we perform that. May be the testing framework should act as a proxy between `lakectl` to `lakefs` server/mock? That way we can verify the request contents and delegate it forwards to the recipient?
@itaiad200 WDYT?

design/lakectl-testing.md
itai-david: In that case I think it is no longer `lakectl` testing, but system/end2end testing, which, BTW I'm totally OK about - but then we shoudl change the scope of this issue and design.
Maybe that's what we should do?!

design/lakectl-testing.md
itai-david: Using different repos is an option, given we decide to go we `lakefs` live server. Or, we can decide on mocking `lakefs` and get 100% isolation plus mocking improbable server states

design/lakectl-testing.md
itai-david: _Thumbed up by @johnnyaug_
:+1:

itai-david: If so, let's change the scope of this ticket to end-2-end tests which will cover client to server to client. We can design it to verify results at each step of the way, i.e. verify the requests sent to the server. verify the response returned from the server and verify the client output (no-output expected is also verifiable)
We can support multi-action or multi-command flows, making the dependencies a desirable side effect, with a single command test being a trivial case of the above.
We can start with `lakectl` as the client to use, but design the framework to be extensible to other clients (WebUI, WebAPI etc)
**And** for quick development of `lakectl` tests, using mocked environment, we can start pushing [lakectl usin tests](https://github.com/treeverse/lakeFS/issues/2499)

design/lakectl-testing.md
itai-david: Definitely.
I'm not sure how to achieve that, but it is definitely a requirement.
Adding to design doc

design/lakectl-testing.md
itai-david: Added

design/lakectl-testing.md
itai-david: Right. I will emphasize the "May" part

design/lakectl-testing.md
itai-david: Either that or we can use @itaiad200's suggestion

nopcoder: The suggestion library does exactly that with an option to execute several commands and match the output with expected result.

design/lakectl-testing.md
itaiad200: lakeFS system testing is covered by Nessie. I think that we should scope this to `lakectl` testing, which is easier and more reliable when you have a running lakeFS. I mentioned that lakeFS is being tested too, because now, for some system flows, will be covering it more than once. There's no downside to it IMO, as the focus of this tests is `lakectl`.

design/lakectl-testing.md
itaiad200: I don't think that we should change the scope of it from `lakectl` testing to `end-2-end` testing since we already have that. Reading thru my comments I understand why it sounds like I'm pushing for a `end-2-end` testing :)  see my previous comment ..

itaiad200 (approved):
Approving since it's seems like we're leaning towards a real lakeFS server (and I'm indifferent regarding `How should we test lakectl`). Great work on this design!

design/lakectl-testing.md
arielshaqed: :+1: for using an existing library, but :-1: for having a spec file -- can we just pipe the commands into a shell and match their outputs using a Real Programming Language?  Otherwise we will  end up with harder-to-write test because the matching primitives are never enough.

E.g. consider testing a digest.  Need to commit, then run `lakectl log`, then extract the digest from the output, then running some command on `lakefs://$repo/$digest`.  Now you need the library to be able to extract the digest from the output of the previous command.

arielshaqed (commented):
# My 2 agorot

I am heavily in favour of testing the real binary.  I shall easily be swayed to test first using the real lakeFS or a mock lakeFS, because I expect we shall end up with *both*.
* If testing a mock lakeFS, then (only!) using a generic mock that reads a swagger definition and allows configuring its behaviour.

For testing the real binary, I would like to write as much of the matching code as possible in a real programming language rather than a "framework".

design/lakectl-testing.md
```diff
@@ -0,0 +1,102 @@
...
+* Totally independent from the environment. No need for DB, underlying storage, or any other future dependency that may be added to `lakefs` server. Moreover, future removal of any dependency from `lakefs` (Postgres?!) will not require framework modification. And, once again, although this advantage is pretty much diminished by the usage of containers, it is still worth noting
+* No uptime, as the mock is a part of the test itself
+* Totally stateless - again, as the mock is part of each test
+* Maximal flexibility - each desired server response can be easily hard-coded
+* No need for initial state setup steps
+* Server bugs are not included
+
+Disadvantages:
+* Requires implementation - unlike `lakefs` server, we currently don't have a mock
+* Requires (a lot of) maintenance - every new feature and every bug fix in `lakefs` may, and probably will, need to be reflected in the mock
```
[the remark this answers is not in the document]
itai-david: Makes sense

design/lakectl-testing.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Architecture
+
+### Architecture Outline
+
+![lakectl testing architecture](./diagrams/lakectl-testing-arch.png)
+
+### Test Case Structure
+
+The test case structure should contain at least the following details:
```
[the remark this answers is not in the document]
itai-david: I see what you mean.
This really madkes me rethink the whole approach.
Thanks for that comment!

design/lakectl-testing.md
itai-david: That is absolutely correct, when using a live server. When using mocks, you can easily generate whatever "`lakefs`" output that you want, and don't really need the dependency between the two commands.
I can totally see where are the difficulties in my aproach (as you wrote - it treats the tests more like unti tests and less like system tests) and the true benefits of letting the real `lakefs` do its thing.
Moreover, I did wrote some simillar mocks in the past, where the mock itself verified the input it received (the client request) and returned hardcoded, tailored output which the client processed and generated its own, "real", output, which was finally verified by the test running tool. I must admit, though, that the testing framework was not as generic as one may hope, and it forced us to write and rewrite the mocks very often, so I would definitley don't want to go that way!

itai-david:
If I try to subsummarize the discussion above:
* Testing `lakectl` code is a no go - the benefits are minor comapring to the huge disadvantage of not testing the real thing
* We understand that we will need to use some hybrid server-mock `lakefs`, as we would like to enjoy the advantages of both
* Favoring live `lakefs` over the mock, for everything that can be done easily (or easy enough)
* Inventing a new spec/format/language (you name it), for defining tests is a big no-no, as we will be chasing our own tail, trying to ckeep up with the need to support new abilities in our tests
* We are not aiming at full-scale end-2-end tests, but we do realize that having a complex test, where several commands are dependent, is both not a problem and have the benefit of imitating natural `lakectl` <--> `lakefs` flow
* Repo level isolation is more than enough, as we alreadt successfully use it

Did I miss anything?

nopcoder (approved):

itaiad200 (approved):

PR #2715, opened 2021-11-17
initial metastore design

nopcoder:


design/metastore.md
```diff
@@ -0,0 +1,102 @@
...
+- Support storing metadata inside existing Hive Metastore implementations
+- Provide any significant change in performance or throughput of metadata operations
+- Provide compatibility with other specific versions of Hive or Hive Metastore
+
+
+### Design Diagram
+
+[lakeFS Metastore](diagrams/metastore.png)
+
+In this diagram:
```
arielshaqed, line 27: Can we direct to the backing S3 store implementation?  This will make some things much easier and faster.

design/metastore.md
```diff
@@ -0,0 +1,102 @@
...
+   - Metadata catalog that will wrap Graveler to serve metadata like we do for the object store.
+
+The diagram show how Spark/Hive/Trino connects to lakeFS's metastore proxy (which acts as Hive Metastore).
+The request to the metastore proxy will invoke the equivalent operation on lakeFS's OpenAPI.
+Inside lakeFS API implementation will use the metastore catalog, that is implemented by Graveler as the storage engine.
+
+
+### Data Model
+
+Using Gravler we can implement a Metastore catalog for all the metadata properties.
```
arielshaqed, line 41: Agreed.  The obvious alternative is to keep the metastore catalog as "normal" files with "special" names inside the repo.  That makes _some_ things much easier (e.g. viewing history, accessing from a client, etc.).  Other things might become harder to do.

Let's explain here why we're adding HMS as more metadata rather than as more data.

There are no advantages that I can see in terms of using Graveler to manage huge files -- I do not believe catalogs can be that big.

The obvious place where storing inside a Graveler is an advantage is *if* merges become trivial.  Then again, when merges are trivial then merging a single structured file should also be quite easy.

arielshaqed (commented):
Thanks, looks like one of our most important features!

design/metastore.md
```diff
@@ -0,0 +1,102 @@
...
+Must of the Hive Metastore Thrift interface kept backward compatibility v2.3.x and v3.x interfaces are almost identical, where version 3 adds Catalog and Workload management metadata.
+
+Code generation should be considered. Having generated bride that will transfer each Thrift API call to our OpenAPI. And/or generate code for each entity described in the Thrift interface can reduce the code in order serialize, map and filter the data model.
+
+Thrift API communicate over a socket, the first step will enable non-secure communication over an additional port. A secure communication can be handled by a external TCP load-balancer first, later the proxy can supply this capability out of the box.
+
+
+Hooks - The Thrift interface include a set of API ends with '_environment_context', which enables the caller passing additional information to the hooks. The hooks mechanism in Hive Metastore is a Java interface implementation that the metastore load and calls on specific operations. We will not support this mechanism at first, but we can have a reference implementation of what to expect in case of adding it to our hooks mechanism.
```
johnnyaug, line 88:
```suggestion
Hooks - The Thrift interface includes a set of API endpoints with the suffix '_environment_context', which allow the caller to pass additional information to the hooks. The hooks mechanism in Hive Metastore is a Java interface implementation that the metastore loads and calls on specific operations. We will not support this mechanism at first, but we can have a reference implementation of what to expect in case of adding it to our hooks mechanism.
```

design/metastore.md
```diff
@@ -0,0 +1,102 @@
...
+
+
+### Discovery
+
+Metastore implementation provides a reference implementation we can test and explore. It still needs to be discovered how to extend functionality to enable versioning.
```
johnnyaug, line 93: > Metastore implementation provides a reference implementation

Is that the Hive metastore?

design/metastore.md
```diff
@@ -0,0 +1,102 @@
...
+
+Finding out more about the following items will help us evaluate our design options:
+
+1. Passing lakeFS's reference/branch information from the metastore client. Explore ways to pass repository/branch/ref information from the client to our metastore.
+1. Side by side - very similar to the current work on side-by-side work with lakeFS and the current object-storage, the user like to run the same application with minimum changes, while enabling some of the data to live inside lakeFS. What needs to change or requirements in order to have the same in the metadata level? Do we need to sync information, manage part of the data and enable mapping on the client side? Can we replace the client side implementation? do we need to spin different endpoints to serve different repositories, branches in lakeFS for parts of the metadata? fallback to existing metastore?
```
johnnyaug, line 98: I think an example is in order here, not sure what it means to have a side-by-side here.

> ... enable mapping on the client side? Can we replace the client side implementation?

Client side of what?

johnnyaug (commented):
Very interesting read.
A lot of parts need more context - see comments.

design/metastore-project.md
```diff
@@ -0,0 +1,19 @@
+# Next Generation Metastore - Project
+
+
+## Milestone 1 - lakeFS Metastore Proposal
+
+Items for discover listed with hierarchy as dependency:
+
+- Passing reference/branch information
+    - Authentication with lakeFS
+    - Side by side - very similar to the current work on side-by-side work with lakeFS and the current object-storage, the user like to run the same application with minimum changes, while enabling some of the data to live inside lakeFS. What needs to change or requirements in order to have the same in the metadata level? Do we need to sync information, manage part of the data and enable mapping on the client side? Can we replace the client side implementation? do we need to spin different endpoints to serve different repositories, branches in lakeFS for parts of the metadata? fallback to existing metastore?
```
talSofer, line 10: I would call this "co-exist with other metastores"

design/metastore.md
```diff
@@ -0,0 +1,102 @@
...
+db/db2/table/tbl1 -> {tableName,dbName,owner,createTime,lastAccessTime,retention,sd,partitionKeys,...}
+...
+```
+
+The data model that Hive Metastore provides enables more than the above. We will implement the missing entities as required by the data tools.
+
+
+### Thrift API
+
+Implementing the Hive Metastore 2.3.x will be the starting point of lakeFS's Metastore.
```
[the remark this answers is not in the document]
talSofer: Implementing the Hive Metastore 2.3.x **API**

design/metastore.md
```diff
@@ -0,0 +1,102 @@
...
+1. Data model diff/merge - describe the level of entities we use for diff/merge. The operation we need to support in order to resolve conflict. How it effects our import/export path?
+1. Authentication with lakeFS - currently lakeFS secure all access by authenticating each request. What will be with metadata?
+1. Data model - best way to model the metadata information into Graveler to enable 3-way diff / merge that will work in optimal way. Enable diff and merge with context, for example, when a column is added we need to know on which table and merge can identify two new partitions to the same table without a conflict.
+
```
talSofer, line 102: Sharing an observation, there are two logical milestones I see here - 
1. lakeFS's Versioned Metastore (LVMS?)
2. Metastore clients integration with LVMS

Given this distinction, the discovery items split as follows: 
### lakeFS's Versioned Metastore
1. Data model diff/merge - describe the level of entities we use for diff/merge.
2. Data model - best way to model the metadata information into Graveler to enable 3-way diff / merge that will work in optimal way.

I think that it will be good to work on the items above in order. The level of entities used for diff/merge will affect how we model the information to optimize these operations. 

### Metastore clients integration with LVMS
1. Passing lakeFS's reference/branch information from the metastore client. Explore ways to pass repository/branch/ref information from the client to our metastore.
2. co-exist with other metastores - explore potential ways to achieve that. 
3. Authentication with lakeFS

IMO 3 is independent of 1 & 2

design/metastore-project.md
```diff
@@ -0,0 +1,19 @@
...
+- Data model - Graveler to enable 3-way diff / merge that will work in optimal way.
```
itaiad200, line 11: Please clarify how this item is different from the 3-way diff that Graveler supports today.

design/metastore.md
```diff
@@ -0,0 +1,102 @@
...
+### Non-Goals
+
+- Support storing metadata inside existing Hive Metastore implementations
+- Provide any significant change in performance or throughput of metadata operations
+- Provide compatibility with other specific versions of Hive or Hive Metastore
+
+
+### Design Diagram
+
+[lakeFS Metastore](diagrams/metastore.png)
```
itaiad200, line 25: Embed it in here instead of linking to it.

design/metastore.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Overview
+
+Deliver manageability of metastore information alongside the object-store data as part of lakeFS.
+
+### Goals
+
+- Metastore entities are versioned: using the same commit and branching abilities that lakeFS provides
+- Metastore entities are versioned together with the data: an atomic merge operation should contain both data mutations as well as metastore mutations.
+- Metastore entities are diff-able: can view the changes made to the metadata between commits
```
itaiad200, line 11: `metadata` has multiple meanings, can you be more explicit?

design/metastore.md
```diff
@@ -0,0 +1,102 @@
...
+- Provide any significant change in performance or throughput of metadata operations
+- Provide compatibility with other specific versions of Hive or Hive Metastore
+
+
+### Design Diagram
+
+[lakeFS Metastore](diagrams/metastore.png)
+
+In this diagram:
+1. Spark/Hive/Trino - data and metadata clients
```
itaiad200, line 28: Do any of these impose limitations on the `metastoreEndpoint`? Any other metastore client, open-source or not, that we need to verify?

design/metastore.md
```diff
@@ -0,0 +1,102 @@
...
+1. lakeFS Metastore Thrift Proxy - act as the client's metastore by implementing the Hive Metastore Thrift protocol
```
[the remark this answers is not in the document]
itaiad200: gateway?

design/metastore.md
```diff
@@ -0,0 +1,102 @@
...
+- Databases
+- Tables
+- Partitions
+- Indices
+- Functions
+- Columns (TODO: how do we store column statistics?)
+- Constraints
+- Serdes
+
+Mapping the above model into our key/value and compare/merge strategy by a table level can be implemented as first step. Compare and merge changes in database/table level, which means that any changes under the table will consider a table modification and the user will handle to resolve a conflcit in case two diffrent nested information would have changed.
```
itaiad200, line 53: I don't understand the `first step` you suggested. Also, what are the possible next steps?

design/metastore.md
```diff
@@ -0,0 +1,102 @@
...
+
+The data model that Hive Metastore provides enables more than the above. We will implement the missing entities as required by the data tools.
+
+
+### Thrift API
+
+Implementing the Hive Metastore 2.3.x will be the starting point of lakeFS's Metastore.
+Spark works with Hive Metastore version 2.3.9 (https://spark.apache.org/docs/latest/sql-data-sources-hive-tables.html), but can also work with version 3.x.
+Trino can work with Hive 2 and 3, while having specific capabilities enabled for version 3.
+Must of the Hive Metastore Thrift interface kept backward compatibility v2.3.x and v3.x interfaces are almost identical, where version 3 adds Catalog and Workload management metadata.
```
itaiad200, line 81: Most?

design/metastore.md
itaiad200, line 88: Can you explain the rational of not supporting that at first (or really any other item which is kept for later)? If there was a discovery that made you realize this shouldn't be implemented first, please elaborate on the findings here. If there wasn't, let's add a discovery item.

design/metastore.md
```diff
@@ -0,0 +1,102 @@
...
+
+Finding out more about the following items will help us evaluate our design options:
+
+1. Passing lakeFS's reference/branch information from the metastore client. Explore ways to pass repository/branch/ref information from the client to our metastore.
+1. Side by side - very similar to the current work on side-by-side work with lakeFS and the current object-storage, the user like to run the same application with minimum changes, while enabling some of the data to live inside lakeFS. What needs to change or requirements in order to have the same in the metadata level? Do we need to sync information, manage part of the data and enable mapping on the client side? Can we replace the client side implementation? do we need to spin different endpoints to serve different repositories, branches in lakeFS for parts of the metadata? fallback to existing metastore?
+1. Data model diff/merge - describe the level of entities we use for diff/merge. The operation we need to support in order to resolve conflict. How it effects our import/export path?
```
itaiad200, line 99: I don't understand how `import/export` are related here.

design/metastore.md
```diff
@@ -0,0 +1,102 @@
...
+1. Authentication with lakeFS - currently lakeFS secure all access by authenticating each request. What will be with metadata?
```
itaiad200, line 100: Why is this different?

design/metastore.md
itaiad200: I don't understand the emphasis we put on `optimising` diff/merge. How many items are there in the context of a table? Few 100s/1000s at best? Lets think about worst-case scenario here and see if we really need to think about optimisation of a merge/diff..

itaiad200 (changes requested):

design/metastore.md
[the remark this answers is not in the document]
johnnyaug: For me it's just "lakeFS Metastore"

design/metastore-project.md
```diff
@@ -0,0 +1,19 @@
+# Next Generation Metastore - Project
+
+
+## Milestone 1 - lakeFS Metastore Proposal
+
+Items for discover listed with hierarchy as dependency:
+
+- Passing reference/branch information
+    - Authentication with lakeFS
```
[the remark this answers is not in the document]
nopcoder: I think both items can make a decision that can effect one another or not.
The issues is that OpenAPI currently check credentials. Assume we will have a solution on how to pass information from the client, we may be using the same way to pass the user credentials.
Alternative is to defer the use of any authentication while working with metadata and rethink how we enable these changes - additional port.

design/metastore.md
nopcoder: See it as part of how we handle the data model, but can split it.
In case we choose as part of the data model handle specific entities - how we plan to import and export the data? do we loose part of it when we export back. Different Metastore solutions like Glue can have differences and she need to address them as part of our data model.
Think it will be easier to processes as a different item or part of how the data model looks like?

design/metastore.md
nopcoder: Hive Metastore doesn't provide any authentication mechanism - our API interface does. It means that if we plan to use OpenAPI to provide metastore management we need to address the authorization and authentication in some way. Or guide the user on how to enable that securely.

design/metastore-project.md
nopcoder: Graveler works on a entry level - each entry change will appear as added/removed/changed.
The data model holds different entries with relations - example: partition is an entity you can append to a table. if we model the partition is an entry in graveler a diff between the two will not explain to the end-user what what database/table was updated. Also without understanding the change it may render as conflict if both sides of the merge will append a partition, while we like to merge both of them.

nopcoder:
@itaiad200 reply about the data model, diff 3-way in optimal way - reporting a list of items that where added/removed/changed like Gravler enables for single entities will work for some but not for all. When we will report that a new partition is added it means that we will need to report the table it was associated. On our current object metadata there is no case to where we report a change and need to fetch the context. The item listed is to discover the best way to we render the model to answer diff, listing (pagination?) and merge of changes when going over a list item keys one time will probably works just for some of our operations.

design/metastore.md
nopcoder: Missed this one - comment about the diff/merge and data model at the end.  Optimizing is not something we need to focus on, just need to think of a data model that will work for us and map the pros and cons in the context of storage and operations we currently support.

@talSofer about the milestones - the first milestone is getting all the known unknowns and decide on the way we like to address the open/discovery items.
I like LMS or lakeMS - will get to naming pretty soon I guess...
Agree we can parallel the client part (although passing ref can be also done by syntax) to the data model and how we going to use lakeFS in implementing our own metastore.
As I see it the authentication we can start simple - do what HMS does - give none and do not open API. But I thought that we already discover what we can pass from the client (branch, ref and etc) so maybe we can use the same technique to pass credentials.
I'll update the metastore-project with the above and add the additional discovery items.

design/metastore.md
```diff
@@ -22,81 +23,79 @@ Deliver manageability of metastore information alongside the object-store data a
...
 In this diagram:
-1. Spark/Hive/Trino - data and metadata clients
-1. lakeFS Metastore Thrift Proxy - act as the client's metastore by implementing the Hive Metastore Thrift protocol
-1. lakeFS - Include the new following components:
-   - Additional operations on OpenAPI that will 
-   - Metadata catalog that will wrap Graveler to serve metadata like we do for the object store.
+1. Spark/Hive/Trino - data and metadata clients (connects using Thrift endpoint)
+1. lakeFS Metastore Thrift Proxy - acts as the client's metastore by implementing the Hive Metastore Thrift protocol
+1. lakeFS - Includes the metatore implementation by:
+   - Additional operations on OpenAPI that will serve the metastore information and operations like we have for objects.
```
arielshaqed, line 32: This is only if we don't implement _on top of_ lakeFS, right?  I think we can get by with implementing a metastore as an external lakeFS client; maybe it will require some pluggable logic inside merges and/or commits, but if that's the sort of thing that we need then these are possibly required OpenAPI _general_ operations rather than _HMS-specific_ ones.

design/metastore.md
```diff
@@ -22,81 +23,79 @@ Deliver manageability of metastore information alongside the object-store data a
...
+   - Metadata catalog that will use Graveler and store metastore information.
```
arielshaqed, line 33: Again, I would much have the metastore on top of lakeFS rather than inside it.  The metastore data are really quite tiny, and I am not sure why we would need to go full Graveler on them, rather than just keeping a JSON-equivalent structure inside a file.

arielshaqed (commented):
Still in favour of implementing on top of lakeFS rather than inside it.   :-)

arielshaqed:
Could you sketch out a bit more the "on top of" option?

johnnyaug (approved):

nopcoder:
> Could you sketch out a bit more the "on top of" option?

Adding to the project plan as first to items to discover and exclude the pros/cons of the two big design options - metastore implementation as part of lakeFS or building the metastore using (on top of) lakeFS.

design/metastore.md
```diff
@@ -0,0 +1,116 @@
...
+db/db1/table/tbl2 -> {tableName,dbName,owner,createTime,lastAccessTime,retention,sd,partitionKeys,...}
+db/db2/table/tbl1 -> {tableName,dbName,owner,createTime,lastAccessTime,retention,sd,partitionKeys,...}
+...
+```
+
+Location - entities location points to the object store location use a global location called warehouse location to store the data.  The location is a path to a directory.
+The location can't reference the lakeFS underlying storage (physical address) as it doesn't contain the structure and object names. We can look into integrating with lakeFS's Hadoop FS where mapping object locations under gateway address can be directly access using 'lakefs://' schema.
+
+The above list of entries are just part of the data model that Hive Metastore provides and the one we assume it is required to work. Supporting the full interface will map the additional entities we will support with one exception we like to discover which is statistics to verify if the same storage should include the statistics, do we need to provide diff/merge the same way and can we update statistics for committed metastore information.
```
itaiad200, line 89: Could you clarify the statistics part? Not really clear..

design/metastore.md
```diff
@@ -0,0 +1,116 @@
...
+
+
+### Thrift API
+
+Implementing the Hive Metastore 2.3.x API will be the starting point of lakeFS Metastore. The must of the Thrift interface of the metastore is CRUD operation on the data model (entities listed above)
+Spark works with Hive Metastore version 2.3.9 (https://spark.apache.org/docs/latest/sql-data-sources-hive-tables.html), but can also work with version 3.x.
+Trino can work with Hive 2 and 3, while having specific capabilities enabled for version 3.
+Must of the Hive Metastore Thrift interface kept backward compatibility. The v2.3.x and v3.x interfaces are almost identical, where version 3 adds Catalog and Workload management metadata.
```
itaiad200, line 97: Should we support that additional metadata?

design/metastore.md
```diff
@@ -0,0 +1,116 @@
...
+
+Implementing Thrift interface include code generation (stub/skeleton), as part of the server side we implement, we should consider using additional code generation or dynamic code that will enable us implement the proxy and pass the request to the lakeFS OpenAPI implementation. It can reduce the work needed and prevent bugs.
```
itaiad200, line 99: Not sure what's the `additional code generation or dynamic code` purpose is, given that the Thrift interface is generated to begin with..

design/metastore.md
```diff
@@ -0,0 +1,116 @@
...
+Hooks - The Thrift interface includes a set of API endpoints with the suffix '_environment_context', which allow the caller to pass additional information to the hooks. The hooks mechanism in Hive Metastore is a Java interface implementation that the metastore loads and calls on specific operations. We will not support this mechanism at first, but we can have a reference implementation of what to expect in case of adding it to our hooks mechanism
+
+
+### Discovery
+
+Hive Metastore implementation provides a reference implementation we can test and explore. It still needs to be discovered how to extend functionality to enable versioning.
+
+Finding out more about the following items will help us evaluate our design options:
+
+1. lakeFS metastore implementation over lakeFS or as part of lakeFS. Manage Agreed. Manage the metadata over lakeFS or over a new Graveler based catalog. What is the relationship between lakeFS repository and the metastore 1:1, 1:m or n:m.
```
itaiad200, line 110: Could you clarify `Manage Agreed`?

design/metastore.md
```diff
@@ -0,0 +1,116 @@
...
+Implementing Thrift interface include code generation (stub/skeleton), as part of the server side we implement, we should consider using additional code generation or dynamic code that will enable us implement the proxy and pass the request to the lakeFS OpenAPI implementation. It can reduce the work needed and prevent bugs.
+
+Hooks - The Thrift interface includes a set of API endpoints with the suffix '_environment_context', which allow the caller to pass additional information to the hooks. The hooks mechanism in Hive Metastore is a Java interface implementation that the metastore loads and calls on specific operations. We will not support this mechanism at first, but we can have a reference implementation of what to expect in case of adding it to our hooks mechanism
+
+
+### Discovery
+
+Hive Metastore implementation provides a reference implementation we can test and explore. It still needs to be discovered how to extend functionality to enable versioning.
+
+Finding out more about the following items will help us evaluate our design options:
```
itaiad200, line 108: One more items I can think of:
1. Importing/Exporting from the metastore

itaiad200 (approved):
Thanks for the initial design! Can't wait to get the discovery items going..

PR #2717, opened 2021-11-17
[DO NOT PUSH] Use aws sdk go v2

### LINKED ISSUE #2732: Garbage collection uses superfluous '/' at beginning of storage identifier
When I enable trace logging and run the garbage collection test with #2717 I see [in logs](https://github.com/treeverse/lakeFS/runs/4278534206?check_suite_focus=true#step:11:337), line 337 of step 11:
```
GET /5351-spark3-metaclient//_lakefs/retention/gc/rules/config.json?x-id=GetObject HTTP/1.1
```
Complete logs attached:  [logs_22472.zip](https://github.com/treeverse/lakeFS/files/7576430/logs_22472.zip)

Note the 2 slashes: these occur because the AWS Go SDK v2 no longer strips double slashes when they appear at the start of a bucket.

This is a repeat of a similar bug in Graveler; however because we use the block adapter directly we are again exposed.  The issue is in [these lines in `garbage_collection_manager.go`](https://github.com/treeverse/lakeFS/blob/b1f64040c258ee923c604488dd65ccbf90405919/pkg/graveler/retention/garbage_collection_manager.go#L22-L24):
```go
const (
	configFileSuffixTemplate    = "/%s/retention/gc/rules/config.json"
	addressesFilePrefixTemplate = "/%s/retention/gc/addresses/"
	commitsFileSuffixTemplate   = "/%s/retention/gc/commits/run_id=%s/commits.csv"
)
```

Those leading slashes end up (mistakenly!) being trimmed by the AWS Go S3 SDK (v1).  If we keep this bug we *will* have an issue (objects misplaced on block storage) when porting GC to Azure and/or GCS.  We _also_ cannot upgrade to AWS Go SDK v2 with this.

One place where the AWS Go S3 SDK (v1) might _keep on using **two** slashes_ is if you define garbage collection on a repo whose storage namespace is nested in a bucket rather than a complete bucket.

arielshaqed:
Advantages:
1. Long-term support.
1. (Somewhat) cleaner semantics.
1. Allows context-logging -- add lakeFS request ID to AWS logs.  (That's impossible in `aws-sdk-go` with the `v2`...).

Does _not_ migrate the CLIs, including especially `lakectl metastore`.

TODO:
- [ ] Validate large (**non-multipart**) uploads.  These should work with SigV4 with no home-grown code.  **Update:** Done manually, need to repeat on latest version.
- [x] Verify with MinIO -- a non-AWS server.  **Update:** Cool, Nessie multipart upload does this.
- [ ] **Must fix #2732 before we can safely pull this!**
- [ ] Decide we are happy with non-chunked upload behaviour of aws-sdk-go-v2, or rewrite our s3uploader and re-introduce it here.


arielshaqed:
**Releaser**: Please _do_ mention this in changelog, e.g.:
> Use version 2 of the AWS Go SDK.  This should have no user-visible consequences, but is a large change.

arielshaqed:
So... ready-ish for review in the sense that it all works.

BUT 88ebedd should really not be pulled without many tests, which is #2732.  Keeping as "**DRAFT**" so as not to pull it.

pkg/logging/aws.go
```diff
@@ -1,9 +1,53 @@
 package logging
 
+import aws_logging "github.com/aws/smithy-go/logging"
+
 type AWSAdapter struct {
 	Logger Logger
 }
 
-func (l *AWSAdapter) Log(vars ...interface{}) {
+func (l *AWSAdapter) logFunc(cls aws_logging.Classification) func(...interface{}) {
```
nopcoder, line 9: Guess we need to wire this one to the adapter's cache creation and have the s3 client use it.

pkg/testutil/db.go
```diff
@@ -212,32 +213,48 @@ func MustDo(t testing.TB, what string, err error) {
...
 	case block.BlockstoreTypeS3:
+		var (
+			configLoadOpts []func(*aws_config.LoadOptions) error
+		)
+
+		configLoadOpts = append(configLoadOpts,
+			func(o *aws_config.LoadOptions) error {
+				o.LogConfigurationWarnings = aws.Bool(true)
+				return nil
+			})
```
nopcoder, line 234:
```suggestion
		configLoadOpts = append(configLoadOpts, aws_config.WithLogConfigurationWarnings(true))
```

too bad that they had a type function type and didn't use it for the setup - LoadOptionsFunc

pkg/testutil/db.go
```diff
@@ -212,32 +213,48 @@ func MustDo(t testing.TB, what string, err error) {
...
+		}
+		// (if no keys, the default shared key providers should load by default).
+
+		cfg, err := aws_config.LoadDefaultConfig(ctx, configLoadOpts...)
+		if err != nil {
+			panic(fmt.Errorf("new S3 block adapter: %w", err))
+		}
+
+		params := block_params.AWSParams{
+			// TODO(ariels)...
```
nopcoder, line 253: ?

pkg/cloud/aws/s3inventory/utils_test.go
```diff
@@ -76,21 +77,28 @@ func uploadFile(t *testing.T, s3 s3iface.S3API, inventoryBucket string, inventor
...
-		S3ForcePathStyle: aws.Bool(true),
-	}
-	newSession, err := session.NewSession(s3Config)
+	cfg, err := aws_config.LoadDefaultConfig(ctx,
+		aws_config.WithCredentialsProvider(
+			aws_credentials.NewStaticCredentialsProvider("YOUR-ACCESSKEYID", "YOUR-SECRETACCESSKEY", "")),
+		func(o *aws_config.LoadOptions) error {
+			o.Region = "eu-central-1"
+			return nil
+		},
```
nopcoder, line 92:
```suggestion
		aws_config.WithRegion("eu-central-1"),
```

pkg/testutil/setup.go
```diff
@@ -101,18 +101,32 @@ func SetupTestingEnv(params *SetupTestingEnvParams) (logging.Logger, api.ClientW
...
+		aws_config.WithCredentialsProvider(
+			aws_credentials.NewStaticCredentialsProvider(
+				viper.GetString("access_key_id"),
+				viper.GetString("secret_access_key"),
+				"")),
+		func(o *aws_config.LoadOptions) error {
+			o.Region = "us-east-1"
+			o.LogConfigurationWarnings = aws.Bool(true)
+			return nil
+		},
```
nopcoder, line 114:
```suggestion
			aws_config.WithRegion("us-east-1"),
			aws_config.WithLogConfigurationWarnings(true),
```

nopcoder (changes requested):

arielshaqed (commented):
Thanks!

PR #2720, opened 2021-11-17
Bump lakeFS hadoopFS version in docs

johnnyaug:


nopcoder:
wonder if we should include version information as part of the site config, and use variables in the docs. so, it will be easy to update the version if these packages in the docs.

talSofer (approved):

talSofer:
> wonder if we should include version information as part of the site config, and use variables in the docs. so, it will be easy to update the version if these packages in the docs.

@nopcoder I like the suggestion. Do you want to open a discovery issue for that?

nopcoder:
Opened https://github.com/treeverse/lakeFS/issues/2729

PR #2721, opened 2021-11-18
document the reasoning for using hadoop 2.7.7 in lakeFSFileSytem

talSofer:


johnnyaug (approved):

arielshaqed:
# Why 2.7.7
## (A guide for the perplexed)

We officially support Spark 2.4.7 and 3.0.1, which we test regularly. The common Hadoop version for these is 2.7.7.  Spark appears to have poor plugin support, and Java / Scala have limited cross-building capabilities. Supporting newer Spark versions requires some effort because they require compiling different code.

# But I really need another version!

If you require support ideally open an issue and give the Spark and Hadoop versions which you are running, so we can direct effort on the right directions.

johnnyaug:
Just writing here for it to be documented: a user has expressed their concern about `hadoop-aws:2.7.7` not supporting per-bucket configuration. This potentially means that we will have to expand our support matrix to include newer versions in the near future.

Issue #2722, opened 2021-11-18
Cannot scope to tags with names that doesn't match the branchID naming format 


talSofer:
While creating tags with a name that's considered an invalid branch ID. (e.g. v.0.1.1) the following error occurs - 
![Screen Shot 2021-11-18 at 10 51 25](https://user-images.githubusercontent.com/14786381/142391084-f71dab68-755c-42ac-b999-7fa8705bcd34.png)

While resolving a ref, we first check if there is a branch with the refID, if there isn't we try to get a tag with this refID. However, while creating a tag with tag name = invalid branch id the condition https://github.com/treeverse/lakeFS/blob/e56d1aad2ae0a288291e28172b02918abc315174/webui/src/lib/hooks/repo.jsx#L13 is true because the error we are getting an "invalid id" error from the server. and the we end up not trying to relive the refID to a tag. 

### Steps to reproduce 

1. Create a tag with a name that is considered an invalid branch ID. e.g. v.0.1.1
2. Go to "Tags" -> click the tag you created.

PR #2723, opened 2021-11-18
Fix bug: Can't view tag when name is invalid as branch id

johnnyaug:
Solves #2722

### DIFF
--- a/webui/src/lib/hooks/repo.jsx
+++ b/webui/src/lib/hooks/repo.jsx
@@ -10,7 +10,7 @@ export const resolveRef = async (repoId, refId) => {
         const branch = await branches.get(repoId, refId);
         return {id: branch.id, type: 'branch'};
     } catch(error) {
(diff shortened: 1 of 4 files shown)


webui/src/lib/hooks/repo.jsx
```diff
@@ -28,7 +28,7 @@ export const resolveRef = async (repoId, refId) => {
         const commit = await commits.get(repoId, refId);
         return {id: commit.id,  type: 'commit'};
     } catch(error) {
-        if (!(error instanceof NotFoundError)) {
+        if (!error instanceof NotFoundError && !error instanceof BadRequestError) {
```
talSofer, line 31: In case we've reached here and ref is not a branch, tag or a commit and the error is aBadRequestError - shouldn't we throw the actual error instead of throwing a NotFoundError in line 36

johnnyaug: Done :)

talSofer (approved):
LGTM!

PR #2726, opened 2021-11-18
Give more chars from body in TRACE logs and make request readable

arielshaqed:
Modify API request tracing middleware to give more bytes in the response body string `response_body`.  And trim
away the response byte *numbers*, which merely bloat the logs but are never useful.

### DIFF
--- a/pkg/httputil/tracing.go
+++ b/pkg/httputil/tracing.go
@@ -9,7 +9,7 @@ import (
 )
 
 const (
-	MaxBodyBytes                      = 100
+	MaxBodyBytes                      = 750
 	RequestTracingMaxRequestBodySize  = 1024 * 1024 * 50  // 50KB
 	RequestTracingMaxResponseBodySize = 1024 * 1024 * 150 // 150KB
 )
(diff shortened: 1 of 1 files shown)


pkg/httputil/tracing.go
```diff
@@ -137,7 +137,6 @@ func TracingMiddleware(requestIDHeaderName string, fields logging.Fields, traceR
 				"took":             time.Since(startTime),
 				"status_code":      responseWriter.StatusCode,
 				"sent_bytes":       responseWriter.ResponseSize,
-				"request_body":     requestBodyTracer.bodyRecorder.Buffer,
```
guy-har, line 140: We are we removing the request body?

can we remove line 130-132
```
// record request body as well
			requestBodyTracer := newRequestBodyTracer(r.Body, RequestTracingMaxRequestBodySize)
			r.Body = requestBodyTracer
```

guy-har (approved):

pkg/httputil/tracing.go
arielshaqed: You were actually right that I should never have removed the request.  Restored that, now logs are glorious!

arielshaqed (commented):
Thanks!  PTAL...

guy-har (approved):
LGTM, Thanks!

PR #2737, opened 2021-11-22
Optimize merge by skipping ranges with same bounds

guy-har:


pkg/graveler/committed/meta_range.go
```diff
@@ -54,8 +54,9 @@ type DiffIterator interface {
 
 // RangeDiff represents a change in Range
 type RangeDiff struct {
-	Type  graveler.DiffType
-	Range *Range
+	Type         graveler.DiffType
+	Range        *Range
+	LeftIdentity ID
```
nopcoder, line 59: can you add a doc code saying when it is populated

pkg/graveler/committed/apply.go
```diff
@@ -267,6 +279,15 @@ func (a *applier) applyDiffRangeSourceKey(diffRange *Range, sourceValue *gravele
 
 func (a *applier) applyBothRanges(diffRange *Range, sourceRange *Range) error {
 	switch {
+	case !diffRange.Tombstone && bytes.Equal(sourceRange.MinKey, diffRange.MinKey) && bytes.Equal(sourceRange.MaxKey, diffRange.MaxKey):
+		// insert diff move both
+		if err := a.writer.WriteRange(*diffRange); err != nil {
```
johnnyaug, line 284: How can you skip this range?
The ranges have the same min-max keys, but the `diffRange` only contains _changes_ to be applied on the source range. You can't write it entirely, if I understand correctly.

johnnyaug (changes requested):
Didn't go over it all yet, posting a comment which I believe is a major issue.

pkg/graveler/committed/apply.go
johnnyaug: Synced f2f: there is an assumption here that if both ranges here have the same boundaries - then `diffRange` can be written to the result as is, and there is no need to look at `sourceRange`.

This is a strange assumption, and I see it as a bug in the apply. Will think of a way to change the contract of the applyer to make it seem more reasonable, and will update here.

pkg/graveler/committed/meta_range.go
guy-har: Added, Thanks

pkg/graveler/committed/iterator.go
```diff
@@ -140,6 +146,11 @@ func (rvi *iterator) SeekGE(key graveler.Key) {
 		return // Reached end.
 	}
 	rvi.started = true // "Started": rangesIt is valid.
+	if bytes.Compare(key, rvi.rng.MinKey) <= 0 {
+		// return range if range equal or greater than key
```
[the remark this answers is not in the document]
guy-har: changed

pkg/graveler/committed/iterator.go
```diff
@@ -16,6 +17,7 @@ type iterator struct {
 	it        graveler.ValueIterator // nil at start of range
 	err       error
 	namespace Namespace
+	preRange  bool
```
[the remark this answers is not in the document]
guy-har: Done

pkg/graveler/committed/apply.go
guy-har: The same bounds case should be similar to the same record case. They both state that X was changed and we need to insert only the change ( doesn't matter if we are referencing a record or a range )
I set the case of overlapping bounds as an error for now, hope it will make it more clear.

guy-har:
@johnnyaug, I handled your comments. PTAL

pkg/graveler/committed/diff.go
```diff
@@ -143,16 +154,20 @@ func (d *diffIterator) compareDiffIterators() diffIteratorCompareResult {
...
 		rightBeforeLeft = bytes.Compare(rightRange.MaxKey, leftRange.MinKey) < 0
+		sameBounds = bytes.Equal(leftRange.MinKey, rightRange.MinKey) && bytes.Equal(leftRange.MaxKey, rightRange.MaxKey)
 	}
 	comp := d.compareDiffKeys()
 	switch {
 	case leftBeforeRight:
 		return diffItCompareResultLeftRangeBeforeRight
 	case rightBeforeLeft:
 		return diffItCompareResultRightRangeBeforeLeft
+	case leftStartRange && rightStartRange && comp == 0 && sameBounds:
```
johnnyaug, line 169:
```suggestion
	case comp == 0 && sameBounds:
```
Should be enough?

eden-ohana (commented):
Code looks great :)

pkg/graveler/committed/merge_iterator.go
```diff
@@ -209,23 +211,36 @@ func (d *compareIterator) stepRange() (hasMore bool, done bool) {
...
 		return d.diffIt.Next(), false
+
+	case graveler.DiffTypeChanged:
+		if baseRange != nil && leftID == baseRange.ID {
+			// changed only in source
+			d.setRangeDiff(rngDiff)
+			d.val = nil
+			return true, true
+		}
+		if bytes.Compare(rng.MaxKey, baseRange.MinKey) < 0 || baseRange != nil && rng.ID == baseRange.ID {
```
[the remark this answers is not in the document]
johnnyaug: Also I'm seeing that merge test doesn't test conflicts, which is strange (not related to this PR, just noting)

pkg/graveler/committed/apply.go
```diff
@@ -157,7 +163,7 @@ func (a *applier) apply() error {
 		}
 	}
 
-	if !a.opts.AllowEmpty && !a.hasChanges(a.summary) {
+	if !a.opts.AllowEmpty && !a.hasChanges(a.summary) && !a.summary.MissingInfo {
```
johnnyaug, line 166: There is an assumption here that - if there is missing info, then the optimization was used, which means there _are_ changes, right?
Can you add a comment to explain that?

pkg/graveler/graveler.go
```diff
@@ -51,7 +51,8 @@ type RawRef struct {
 }
 
 type DiffSummary struct {
-	Count map[DiffType]int
+	Count       map[DiffType]int
+	MissingInfo bool // true when Diff summary has missing Information (could happen when skipping ranges with same bounds)
```
johnnyaug, line 55: Incomplete
```suggestion
	Incomplete bool // true when Diff summary has missing Information (could happen when skipping ranges with same bounds)
```

Should this also be used on the command line when showing the merge summary?

johnnyaug (changes requested):

pkg/graveler/committed/apply.go
```diff
@@ -265,33 +279,77 @@ func (a *applier) applyDiffRangeSourceKey(diffRange *Range, sourceValue *gravele
 	return nil
 }
 
-func (a *applier) applyBothRanges(diffRange *Range, sourceRange *Range) error {
+type RangeCompareSate int
+
+const (
+	RangeCompareBefore RangeCompareSate = iota
+	RangeCompareAfter
+	RangeCompareSameRange
```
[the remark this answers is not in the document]
guy-har: Changed, Thanks

pkg/graveler/committed/diff.go
guy-har: In our iterator we know for sure that we are at the beginning of a range if the record is nil. Currently in the case of `sameBounds` we don't receive both range and record ( `sameBounds` => `leftSrartRange` && `rightStartRange`) meaning, your suggestion will work. But may not always be the case, I prefer to leave it as is.

pkg/graveler/committed/apply.go
```diff
@@ -265,33 +279,77 @@ func (a *applier) applyDiffRangeSourceKey(diffRange *Range, sourceValue *gravele
...
+	default:
+		return RangeCompareOverlapping
+	}
+}
+
+func (a *applier) applyBothRanges(diffRange *Range, sourceRange *Range) error {
+	comp := compareRanges(sourceRange, diffRange)
+	switch comp {
+	case RangeCompareOverlapping:
+		return fmt.Errorf("%w - overlapping ranges", ErrInvalidState)
```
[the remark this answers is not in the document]
guy-har: Changed to support overlapping even though we shouldn't have that case

pkg/graveler/committed/apply.go
guy-har: Moved to `hasChanges` and documented. Thanks

pkg/graveler/graveler.go
guy-har: Changed.
Not as part of this PR.

pkg/graveler/committed/merge_iterator.go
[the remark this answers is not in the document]
guy-har: It is a conflict.
Added a test for this case as well
Thanks!

pkg/graveler/committed/diff_test.go
```diff
@@ -275,11 +275,13 @@ func TestDiffCancelContext(t *testing.T) {
 		t.Fatalf("Err() returned %v, should return context.Canceled", err)
 	}
 }
+
+// TODO(Guys): add test for range changed
```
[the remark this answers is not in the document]
guy-har: Added a test for range changed

johnnyaug (approved):

PR #2740, opened 2021-11-22
Run unit tests on Windows

arielshaqed:
Use windows-latest, this is *not* our primary platform.  Mark experimental,
for now failures are not blocking.


arielshaqed:
Diffs for how far I got:
[windows.diffs.gz](https://github.com/treeverse/lakeFS/files/7587390/windows.diffs.gz)

Issue #2675, opened 2021-11-09
[Discovery] Changes/Compare views: on-demand stats 


johnnyaug:
As a user, before committing changes or merging, I would like to be able to calculate stats on the changes or a prefix. These stats include: total size and number of added/deleted/changed objects.

Design a solution to show these stats. The usability can be inspired by S3's "Calculate size" UI feature.

PR #2744, opened 2021-11-23
UI: change summary under prefix

johnnyaug:
When viewing a change, this feature adds a button to calculate a summary of the change under every key prefix.
When the button is clicked, the user will see a number of added/changed/deleted/conflicting objects under the prefix.

This was added to all screens where changes can be viewed: Uncommitted changes, Compare view, and when diving into a specific commit.

Peek 2021-11-23 11-52

Solves #2675.


webui/src/lib/components/repository/ChangeSummary.jsx
```diff
@@ -0,0 +1,109 @@
...
+import {OverlayTrigger, Tooltip} from "react-bootstrap";
+import {humanSize} from "./tree";
+
+const MAX_NUM_OBJECTS = 10_000;
+const PAGE_SIZE = 1_000;
+
+class SummaryEntry {
+    constructor() {
+        this.count = 0
+        this.size = 0
```
talSofer, line 19:
```suggestion
        this.sizeBytes = 0
```

webui/src/lib/components/repository/ChangeSummary.jsx
```diff
@@ -0,0 +1,109 @@
...
+        prev[current.type].add(1, current.size_bytes)
+        return prev
+    }, new SummaryData())
+    const detailsTooltip = <Tooltip>
+        <div className={"m-1 small text-left"}>
+            {summaryData.added.count > 0 &&
+                <><span className={"color-fg-added"}>{summaryData.added.count}</span> objects added (total {humanSize(summaryData.added.size)})<br/></>}
+            {summaryData.removed.count > 0 &&
+                <><span className={"color-fg-removed"}>{summaryData.removed.count}</span> objects removed<br/></>}
+            {summaryData.changed.count > 0 &&
```
talSofer, line 88: Can you please explain your choice not to show the size for the items removed?

webui/src/lib/components/repository/changes.jsx
```diff
@@ -89,65 +96,79 @@ export const TreeItem = ({ entry, repo, reference, leftDiffRefID, rightDiffRefID
...
-    let actions = entryActions(showActions, entry, onRevert);
-
-    if (entry.path_type === "common_prefix"){
+    const [showSummary, setShowSummary] = useState(false);
+    if (entry.path_type === "common_prefix") {
         pathSection = <Link href={onNavigate(entry)}>{pathSection}</Link>
     }
+    const rowActions = []
+    if (onClickExpandDiff) {
+        rowActions.push(new RowAction(<FileDiffIcon/>, diffExpanded ? "Hide changes" : "Show changes", diffExpanded, onClickExpandDiff))
```
talSofer, line 147: The FileDiffIcon should be visible by default. I think that you have a bug here because the icon becomes visible only while hovering a leaf row.

webui/src/lib/components/repository/changes.jsx
```diff
@@ -89,65 +96,79 @@ export const TreeItem = ({ entry, repo, reference, leftDiffRefID, rightDiffRefID
...
-            <td className="diff-indicator">{diffIndicator}</td>
-            <td className="tree-path">
-                <span style={{marginLeft: depth * 20 + "px"}}>
+        <tr className={rowClass}>
+            <td className="pl-4 col-auto p-2">{diffIndicator}</td>
+            <td className="col-9 tree-path">
+                <span style={{marginLeft: (depth * 20) + "px"}}>
                     <span onClick={onClick}>
-                        {leaf
-                            ? entry.type === 'conflict'
```
talSofer, line 131: Where are you preserving this logic which means - don't show the content diff option in case of a conflict?

talSofer (changes requested):

webui/src/lib/components/repository/ChangeSummary.jsx
```diff
@@ -0,0 +1,108 @@
...
+        this.added = new SummaryEntry()
+        this.changed = new SummaryEntry()
+        this.removed = new SummaryEntry()
+        this.conflict = new SummaryEntry()
+    }
+}
+
+/**
+ * Widget to display a summary of a change: the number of added/changed/deleted/conflicting objects.
+ * Shows an error if the change is has more than {@link MAX_NUM_OBJECTS} entries.
```
itaiad200, line 38:
```suggestion
 * Shows an error if the change has more than {@link MAX_NUM_OBJECTS} entries.
```

webui/src/lib/components/repository/ChangeSummary.jsx
```diff
@@ -0,0 +1,108 @@
...
+            return
+        }
+        const {results, pagination} = await getMore(resultsState.pagination.next_offset || "", prefix, false, PAGE_SIZE)
+        if (!pagination.has_more) {
+            setLoading(false)
+        }
+        setResultsState({results: resultsState.results.concat(results), pagination: pagination, tooManyPages: false})
+    }, [resultsState.results, loading])
+
+    if (loading) return <ClockIcon/>
```
itaiad200, line 63: Why take this approach instead of showing results so far and keep update it forever..?

webui/src/lib/components/repository/changes.jsx
```diff
@@ -89,65 +96,79 @@ export const TreeItem = ({ entry, repo, reference, leftDiffRefID, rightDiffRefID
         return <Error error={error}/>
 
     if (loading && results.length === 0)
-        return <TreeEntryRow key={entry.path+"entry-row"} entry={entry} showActions={true} loading={true} relativeTo={relativeTo} depth={depth} onRevert={onRevert} onNavigate={onNavigate} repo={repo} reference={reference}/>
+        return <TreeEntryRow key={entry.path+"entry-row"} entry={entry} loading={true} relativeTo={relativeTo} depth={depth} onRevert={onRevert} onNavigate={onNavigate}repo={repo} reference={reference}
```
itaiad200, line 99:
```suggestion
        return <TreeEntryRow key={entry.path+"entry-row"} entry={entry} loading={true} relativeTo={relativeTo} depth={depth} onRevert={onRevert} onNavigate={onNavigate} repo={repo} reference={reference}
```

webui/src/lib/components/repository/changes.jsx
johnnyaug: oopsy

webui/src/lib/components/repository/ChangeSummary.jsx
```diff
@@ -0,0 +1,108 @@
...
+        }
+        if (!loading) {
+            return
+        }
+        const {results, pagination} = await getMore(resultsState.pagination.next_offset || "", prefix, false, PAGE_SIZE)
+        if (!pagination.has_more) {
+            setLoading(false)
+        }
+        setResultsState({results: resultsState.results.concat(results), pagination: pagination, tooManyPages: false})
+    }, [resultsState.results, loading])
```
itaiad200, line 61: This is inconsistent with file viewer. You may show statistics for 3 files but present 2 in the ui.
This happens because:
1. Listing happens separately.
2. Files are cached and statistics are not.

itaiad200 (changes requested):
Requesting changes mostly for checking consistency with the actual file viewers. If it's too much for this PR let's open a separate issue for consistency and cache handling of entries in the ChangeViewer.

webui/src/lib/components/repository/ChangeSummary.jsx
```diff
@@ -0,0 +1,108 @@
...
+const MAX_NUM_OBJECTS = 10_000;
+const PAGE_SIZE = 1_000;
+
+class SummaryEntry {
+    constructor() {
+        this.count = 0
+        this.size = 0
+    }
+    add(count, size) {
+        this.count += count
```
nopcoder, line 22: guess it will be always 1

webui/src/lib/components/repository/ChangeSummary.jsx
johnnyaug: I think it's not a good experience to show an updating number of objects.

webui/src/lib/components/repository/ChangeSummary.jsx
johnnyaug: Solving this basically means I have to refresh all expanded paths in the tree, together with all the stats that are already shown. 

Explanation:

Since summaries are calculated on demand - the only way for me to be consistent with the expanded tree is to refresh all expanded paths under the given path.

After doing so, shown summaries on upper levels should be updated to be consistent with their children. 
In turn, all expanded paths under these summaries should be updated to be consistent with them.

This potentially requires updating the whole tree.

I'm not sure it's worth the effort and the performance risk. Since it is clearly shown to the user that the summary is _calculated_, I think inconsistencies can be expected.

webui/src/lib/components/repository/ChangeSummary.jsx
johnnyaug: Done

webui/src/lib/components/repository/changes.jsx
johnnyaug: Done

webui/src/lib/components/repository/ChangeSummary.jsx
johnnyaug: Done

webui/src/lib/components/repository/changes.jsx
johnnyaug: Done

webui/src/lib/components/repository/ChangeSummary.jsx
johnnyaug: Fixed.

(I was under the false impression that the size wasn't returned for "deleted" type diffs)

webui/src/lib/components/repository/changes.jsx
johnnyaug: Couldn't find a reasonable way to do it, I don't want all the side-actions to be visible by default.

webui/src/lib/components/repository/changes.jsx
```diff
@@ -143,7 +143,7 @@ export const TreeEntryRow = ({entry, relativeTo = "", leaf = false, dirExpanded,
         pathSection = <Link href={onNavigate(entry)}>{pathSection}</Link>
     }
     const rowActions = []
-    if (onClickExpandDiff) {
+    if (onClickExpandDiff && entry.type !== 'conflict') {
```
talSofer, line 146: You don't have to but consider exporting this and other diff types to the new constants.jsx file

webui/src/lib/components/repository/ChangeSummary.jsx
talSofer: @johnnyaug to clarify - are you referring to a situation where you added let's say 100 objects to partition "a", and at a time 50 objects are loaded to the tree but the stats show that 100 objects were added?

johnnyaug: What is described here is a scenario where data is constantly written to the repo, and you are viewing the change at different points in time.

itaiad200: Just to make sure I understand, this limitation is due to the way we built the tree structure. "Refreshing" means we need to go and fetch everything again, while the summary already did it. I don't want to block on it because I don't think it would be that common. Can we open an issue for consolidation of the summary and the tree fetching?

itaiad200 (approved):
Approving, great work! I mostly played around with the UI, didn't look too deeply into the code.

talSofer (approved):
Awsome feature!

PR #2746, opened 2021-11-23
botwo design

eden-ohana:


design/boto-design.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Value Propositions
+When a user with an existing python code that uses boto3 client to access s3 wants to move only part of its buckets/prefixes to lakefs, botwo allows using lakefs and s3 together with minimum user's code changes.
+
+## Proposed Design
+The botwo package provides a boto3 like client with two boto3 clients.
+It handles several things for the user:
+* Holds two boto clients and routes the requests between them by the bucket/prefix configuration.
+* Maps 1:1 with the boto AWS service API. All botwo client AWS functions will be generated on the fly according to boto3 client AWS functions.
+* Holds special treatment for two buckets functions like `upload_part_copy`. If more two buckets functions will be added to s3 in the future, we will need to implement them as well.
```
johnnyaug, line 28:
```suggestion
* Holds special treatment for functions that operate on multiple buckets like `upload_part_copy`. If more such functions are added to s3 in the future, we will need to implement them as well.
```

design/boto-design.md
```diff
@@ -0,0 +1,101 @@
...
+
+### Route all the requests to lakefs server and implement the bucket configuration there
+It can result in server load because requests like put_object can call directly to s3.
+
+### Wrap boto core fs
+Possible but complex because we will need to implement boto core interface which has a lot of functionality.
+
+## Usage
+
+Current Usage (only with s3):
```
johnnyaug, line 59:
```suggestion
Before introducing the new client, the user calls interacts only with S3:
```

design/boto-design.md
```diff
@@ -0,0 +1,101 @@
...
+```
+import boto3
+
+s3 = boto3.client('s3')
+
+s3.get_object(Bucket="test-bucket", Key="test/object.txt") # in a lot of places
+```
+
+Future Usage: change in every place in the code where there is boto client initialized.
```
johnnyaug, line 68:
```suggestion
With the new client: change in every place in the code where a boto S3 client is initialized.
```

design/boto-design.md
```diff
@@ -0,0 +1,101 @@
...
+## Non Goals
+* Provide new features that weren't previously possible through lakefs
+
+## Value Propositions
+When a user with an existing python code that uses boto3 client to access s3 wants to move only part of its buckets/prefixes to lakefs, botwo allows using lakefs and s3 together with minimum user's code changes.
+
+## Proposed Design
+The botwo package provides a boto3 like client with two boto3 clients.
+It handles several things for the user:
+* Holds two boto clients and routes the requests between them by the bucket/prefix configuration.
```
johnnyaug, line 26: If I understand correctly, it also prepends a prefix (e.g. "main/") to all requests made to lakeFS. This should be noted here

design/boto-design.md
```diff
@@ -0,0 +1,101 @@
...
+
+Future Usage: change in every place in the code where there is boto client initialized.
+```
+import boto3
+import botwo
+
+client1 = boto3.client('s3')
+client2 = boto3.client('s3', endpoint_url='https://lakefs.example.com')
+
+mapping = [('bucket-a/*', {'endpoint': 'lakefs', 'bucket': 'my-repo', 'prefix': 'dev/'}),
```
johnnyaug, line 76: About allowing a bucket+prefix combo as the key to routing the request, I'm not sure the benefits justify the complexity.
Also if I understand correctly, we map here from the bucket name to a repo name in lakeFS. This does give some value (no path changes involved) but again makes things more complicated.

Consider the following mapping structure, or something in the middle between this and what you suggested:
```json
[
  {
    "bucket_name": "example-repo",  // will require the user to change a path
    "endpoint": "lakefs",
    "key_prefix": "main/"
  },
  {
    "bucket_name": "example-old-bucket",
    "mapped_bucket_name": "example-repo", // optional - no path change required
    "endpoint": "lakefs",
    "key_prefix": "main/"
  },
  {
    "bucket_name": "*",
    "endpoint": "s3",
    "key_prefix": ""
  }
]

File: design/boto-design.md [hunk 6]
```diff
@@ -0,0 +1,101 @@
...
+('bucket-b/foo', {'endpoint': 'lakefs', 'bucket': 'bucket-b', 'prefix': 'master/'}),
+# default
+('*', 's3')]
+
+s3 = botwo.client({"s3": client1, "lakefs": client2}, mapping)
+
+s3.get_object(Bucket="test-bucket", Key="test/object.txt") # stays the same
+```
+
+Wildcard matching- every time s3 function is called, it will pass on the mapping prefix list by order and stop if there is a match.
```
johnnyaug, line 86:
```suggestion
Wildcard matching- every time s3 function is called, it will go over the mapping prefix list by order and stop if there is a match.
```

johnnyaug (changes requested):
Great work! Very clear design.

Considering the name (sorry): it is more smart and witty than descriptive - I think it will be confusing (mainly because boto3 is the current name of boto, so in a way BoTwo sounds like an older version). 

Added some  other comments.

design/boto-design.md
```diff
@@ -0,0 +1,101 @@
...
+1. Clients (namely, boto3 and pandas) don't support multiple S3 endpoints with differing credentials. This makes it hard to change only a subset of the data to be in lakeFS because the client, once initialized, can talk to either lakeFS or S3 but not both.
+2. Once data is migrated from S3 to lakeFS, even if the repo has the same name as the previous bucket, it's still not a drop-in replacement, because in lakeFS you'd also need to specify the version ref (branch/commit/tag) you're using, as a prefix.
+
+Solving 1 is hard because it requires a big infrastructure change: instead of using a single initialized client (sometimes it's a default or global one), now multiple clients need to be introduced (one for S3, one for lakeFS) with custom logic for when to use which one.
+Solving 2 is hard because it involves changing many paths scattered throughout the code base: some are dynamically generated by the code, some come from env vars or configuration files - it's not always easy to ensure coverage.
+
+## Goals
+* Make it easy for new users to transfer existing data to lakefs
+* Isolation - Users should be able to separate access to bucket/prefix between lakefs and s3
```
arielshaqed, line 14: Consider avoiding the word "isolation", it often means something else than what I think is meant here.  This is maybe more like "routing"?

design/boto-design.md
```diff
@@ -0,0 +1,101 @@
...
+* Users should do the minimal set of changes possible in code to set lakefs
+
+## Non Goals
+* Provide new features that weren't previously possible through lakefs
+
+## Value Propositions
+When a user with an existing python code that uses boto3 client to access s3 wants to move only part of its buckets/prefixes to lakefs, botwo allows using lakefs and s3 together with minimum user's code changes.
+
+## Proposed Design
+The botwo package provides a boto3 like client with two boto3 clients.
```
arielshaqed, line 24: Is there any reason not to support any number of clients, say with one of them being the default?  I believe it will also simplify the configuration.

design/boto-design.md
```diff
@@ -0,0 +1,101 @@
...
+It handles several things for the user:
+* Holds two boto clients and routes the requests between them by the bucket/prefix configuration.
+* Maps 1:1 with the boto AWS service API. All botwo client AWS functions will be generated on the fly according to boto3 client AWS functions.
```
arielshaqed, line 27: Which version of the API?

design/boto-design.md
```diff
@@ -0,0 +1,101 @@
...
+* Holds special treatment for two buckets functions like `upload_part_copy`. If more two buckets functions will be added to s3 in the future, we will need to implement them as well.
+* It has more functionality like the event-system that will be implemented by using one of its clients (the default one).
```
arielshaqed, line 29: Without reading the code, I am not sure what this means for the user.

design/boto-design.md
```diff
@@ -0,0 +1,101 @@
...
+s3 = boto3.client('s3')
+if bucket == "bucket_a" and prefix == "my_path":
+  client.endpoint_url  = 'https://lakefs.example.com'
+```
+Does not work because once you initialize boto client with parameters you can’t change them in run time.
+
+### Method Overriding
+
+Monkey patching / overriding base class functions in boto3 with inheritance.
+Does not work because Boto3 functions are generated on the fly from service definitions, so there is no static put_object function to inherit from.
```
arielshaqed, line 44: Understood about inheritance but not about patching.

design/boto-design.md
```diff
@@ -0,0 +1,101 @@
...
+```
+s3 = boto3.client('s3')
+if bucket == "bucket_a" and prefix == "my_path":
+  client.endpoint_url  = 'https://lakefs.example.com'
+```
+Does not work because once you initialize boto client with parameters you can’t change them in run time.
+
+### Method Overriding
+
+Monkey patching / overriding base class functions in boto3 with inheritance.
```
arielshaqed, line 43: What about wrapping using some explicit AOP (aspect-oriented programming) method?

design/boto-design.md
arielshaqed: Can we perform the routing using just a _function_?  Having to define a complex Pythonic object means it is really hard to understand the object and its logic.  E.g. consider in all of the above: how do I say "all bucket names starting with `ariels-`?  "all JSONs inside bucket `config`?"  **You add syntax.** Now every user suffers from the added syntax, and at the same time users who do need a feature might not have it.  In many cases writing my own function is (much) simpler.

For the cases where you _do_ want to define some structure, we could define **at a second layer** a function `mapping` that takes an object like the above and returns a suitable function.

Alternatively, if we *do* do things like this, it will be easier to understand if we just matched URLs directly, rather than having to split things up like this.  I do not see the advantage of requiring the user to understand "bucket" and "key" separately.

arielshaqed, line 76: AFAIU this is _not_ an `endpoint`, this is a `client`.

design/boto-design.md
```diff
@@ -0,0 +1,101 @@
...
+If there isn't a match it will go to the default client (the user must provide a default client).
+
+## Open Questions
+![Drawing of boto architecture](./boto-architecture.png)
+
+* There are two different abstractions within the boto3 SDK for making AWS service requests: client and resource.
+  * Client - provides low-level AWS service access, maps 1:1 with the AWS service API.
+  * Resource - provides high-level, object-oriented API. It exposes sub-resources and collections of AWS resources and has actions (operations on resources).
+  
+  This proposal provides a solution only for boto client, It can be expended to resource as well per user requests.
```
arielshaqed, line 96: Can you give an example of a resource?

design/boto-design.md
```diff
@@ -0,0 +1,101 @@
...
+* Isolation - Users should be able to separate access to bucket/prefix between lakefs and s3
+* Users should do the minimal set of changes possible in code to set lakefs
+
+## Non Goals
+* Provide new features that weren't previously possible through lakefs
+
+## Value Propositions
+When a user with an existing python code that uses boto3 client to access s3 wants to move only part of its buckets/prefixes to lakefs, botwo allows using lakefs and s3 together with minimum user's code changes.
+
+## Proposed Design
```
arielshaqed, line 23: Please add a picture here, it should make things easier to understand.

arielshaqed (commented):
Thanks!  Very important and useful feature!

- Please see comments.
- Please also include the Excalidraw source for the drawing, so we can continue to edit it.

design/boto-design.md
nopcoder: Didn't understand from the design how do we manage to handle all boto requests? do hard code each method boto currently support? do we dynamically generate them? it is important to understand this part to know how a call to our client on a method that boto holds, implemented after we published the package behaves.

design/boto-design.md
```diff
@@ -0,0 +1,101 @@
...
+## Open Questions
+![Drawing of boto architecture](./boto-architecture.png)
+
+* There are two different abstractions within the boto3 SDK for making AWS service requests: client and resource.
+  * Client - provides low-level AWS service access, maps 1:1 with the AWS service API.
+  * Resource - provides high-level, object-oriented API. It exposes sub-resources and collections of AWS resources and has actions (operations on resources).
+  
+  This proposal provides a solution only for boto client, It can be expended to resource as well per user requests. 
+
+* Pandas: there are two ways to access s3 - Boto and s3fs-supported pandas API (S3Fs is a Pythonic file interface to S3. It builds on top of botocore) which is not addressed as well.
```
nopcoder, line 98: So if we target botocore we cloud address all the above?

design/boto-design.md
```diff
@@ -0,0 +1,101 @@
...
+s3.get_object(Bucket="test-bucket", Key="test/object.txt") # stays the same
+```
+
+Wildcard matching- every time s3 function is called, it will pass on the mapping prefix list by order and stop if there is a match.
+If there isn't a match it will go to the default client (the user must provide a default client).
+
+## Open Questions
+![Drawing of boto architecture](./boto-architecture.png)
+
+* There are two different abstractions within the boto3 SDK for making AWS service requests: client and resource.
```
nopcoder, line 92: Which raise the question if we should address the request route in a lower level as the client functionality - resource, client, session will eventually will go down to the S3 API to perform the actual operation. While addressing the higher level interface will require us always to make sure we cover more and more entry points.

design/boto-design.md
```diff
@@ -0,0 +1,101 @@
+# Co-Existing with Boto design
```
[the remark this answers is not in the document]
eden-ohana: Done

design/boto-design.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Problem Description
+
+Existing code bases where paths are scattered throughout the codebase (which is rather common for DE/DS code) have a hard time incorporating the lakeFS S3 gateway for several reasons:
+1. Clients (namely, boto3 and pandas) don't support multiple S3 endpoints with differing credentials. This makes it hard to change only a subset of the data to be in lakeFS because the client, once initialized, can talk to either lakeFS or S3 but not both.
```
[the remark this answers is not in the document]
eden-ohana: Done

design/boto-design.md
```diff
@@ -0,0 +1,101 @@
...
+2. Once data is migrated from S3 to lakeFS, even if the repo has the same name as the previous bucket, it's still not a drop-in replacement, because in lakeFS you'd also need to specify the version ref (branch/commit/tag) you're using, as a prefix.
+
+Solving 1 is hard because it requires a big infrastructure change: instead of using a single initialized client (sometimes it's a default or global one), now multiple clients need to be introduced (one for S3, one for lakeFS) with custom logic for when to use which one.
+Solving 2 is hard because it involves changing many paths scattered throughout the code base: some are dynamically generated by the code, some come from env vars or configuration files - it's not always easy to ensure coverage.
+
+## Goals
+* Make it easy for new users to transfer existing data to lakefs
```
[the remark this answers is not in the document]
eden-ohana: Done

design/boto-design.md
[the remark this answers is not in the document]
eden-ohana: Yes

design/boto-design.md
```diff
@@ -0,0 +1,101 @@
...
+* Isolation - Users should be able to separate access to bucket/prefix between lakefs and s3
+* Users should do the minimal set of changes possible in code to set lakefs
+
+## Non Goals
+* Provide new features that weren't previously possible through lakefs
+
+## Value Propositions
+When a user with an existing python code that uses boto3 client to access s3 wants to move only part of its buckets/prefixes to lakefs, botwo allows using lakefs and s3 together with minimum user's code changes.
```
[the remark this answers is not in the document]
eden-ohana: Done

design/boto-design.md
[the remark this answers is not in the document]
eden-ohana: Done

design/boto-design.md
eden-ohana: Done

design/boto-design.md
eden-ohana: Done

design/boto-design.md
eden-ohana: Done

design/boto-design.md
eden-ohana: Done

design/boto-design.md
eden-ohana: All bucket names starting with ariels = ariels*
Can you give an example of such function?

design/boto-design.md
eden-ohana: Done

design/boto-design.md
eden-ohana: Yes

design/boto-design.md
eden-ohana: The latest

design/boto-design.md
eden-ohana: def route_get_object(kwargs):
 #route the params

boto.client.get_object = route_get_object

s3 = boto.client("s3")
s3.get_object() # boto get_object is called

design/boto-design.md
eden-ohana: I will check it out.

design/boto-design.md
eden-ohana: Done

design/boto-design.md
eden-ohana: Done

design/boto-design.md
eden-ohana: It will be published to pypi as a package. 
The implementation will be such that we will always be sync with boto latest version because we dynamically generate the s3 api functions in run time according to boto.

design/boto-design.md
```diff
@@ -0,0 +1,101 @@
...
+2. Once data is migrated from S3 to lakeFS, even if the repo has the same name as the previous bucket, it's still not a drop-in replacement, because in lakeFS you'd also need to specify the version ref (branch/commit/tag) you're using, as a prefix.
+
+Solving 1 is hard because it requires a big infrastructure change: instead of using a single initialized client (sometimes it's a default or global one), now multiple clients need to be introduced (one for S3, one for lakeFS) with custom logic for when to use which one.
+Solving 2 is hard because it involves changing many paths scattered throughout the code base: some are dynamically generated by the code, some come from env vars or configuration files - it's not always easy to ensure coverage.
+
+## Goals
+* Make it easy for new users to transfer existing data to lakefs
+* Isolation - Users should be able to separate access to bucket/prefix between lakefs and s3
+* Users should do the minimal set of changes possible in code to set lakefs
```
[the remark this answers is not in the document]
eden-ohana: Done

design/boto-design.md
eden-ohana: Yes

design/boto-design.md
eden-ohana: That question came into my mind too, I still think it will be more complex.

design/boto-design.md
nopcoder: So we will not require any specific version of boto as dependency of the 'botwo' package?

johnnyaug (approved):

nopcoder (approved):

PR #2751, opened 2021-11-24
Enable content diff while comparing references other than branches

### LINKED ISSUE #2749: UI: content comparison only works if right hand side is a branch
Trying to expand an object in the compare tab works only if the "compared to" picker has a branch selected.
When trying to select a tag or a commit, it results in a (confusing) error message:

image

_looking at the network tab, it seems we always append a `@` to the ref, which works for branches, but is not a notation that commits and tags support._

talSofer:
Fixes #2749

Tested that content diff is available while comparing branches, commit ids and tags.


webui/src/pages/repositories/repository/compare.jsx
```diff
@@ -98,8 +98,14 @@ const CompareList = ({ repo, reference, compareReference, prefix, onSelectRef, o
                             <Table borderless size="sm">
                                 <tbody>
                                 {results.map(entry => {
-                                    const leftCommittedRef = reference.id + "@"
-                                    const rightCommittedRef = compareReference.id + "@"
+                                    let leftCommittedRef = reference.id;
+                                    let rightCommittedRef = compareReference.id;
+                                    if (reference.type === 'branch') {
```
ozkatz, line 103: Could be done now or in a future refactor, but this is a constant and should be defined as such. I think it's used widely so probably doesn't belong in this PR though.

ozkatz (approved):
Thanks! looks good.

johnnyaug (approved):

webui/src/pages/repositories/repository/compare.jsx
talSofer: Fixed, thanks

PR #2755, opened 2021-11-25
Correctly and explicitly report too-short files

### LINKED ISSUE #2508: Incorrect exception thrown while passing a non-sst file to lakeFS's SST parser
The parser fails with IlligalArgumentException on getting a non-sst file instead of failing with BadFileFormatException in this case. 

After fixing this bug, we need to enable the unit test described as "with non-sst file of size 0" which is currently ignored. This unit-test is implemented under BlockParserSpec.scala.

arielshaqed:
Fixes #2508.

### DIFF
--- a/clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
+++ b/clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
@@ -377,9 +377,9 @@ class GolangContainerSpec extends AnyFunSpec with ForAllTestContainer {
     }
 
     describe("with empty file") {
-      ignore("should fail parsing") {
-        assertThrows[BadFileFormatException]{ // TODO (tals): enable the test after closing https://github.com/treeverse/lakeFS/issues/2508
-          withGeneratedEmptyTestFile("empty.sfile", verifyBlockParserOutput)
+      it("should fail parsing") {
+        assertThrows[BadFileFormatException] {
+          withGeneratedEmptyTestFile("empty.file", verifyBlockParserOutput)
         }
       }
     }
--- a/clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
+++ b/clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
@@ -20,7 +20,7 @@ class BlockReadableSpec extends AnyFunSpec with Matchers {
   describe("instantiate BlockReadableFileChannel") {
     describe("with null file channel") {
       it("should fail with a Null Pointer Exception") {
-        assertThrows[NullPointerException]{
+        assertThrows[NullPointerException] {
           new BlockReadableFileChannel(null)
         }
       }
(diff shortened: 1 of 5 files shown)


arielshaqed:
Proposed changelog (in _Spark client_, **not** the lakeFS release):
> Report the correct error when a file is too short.

clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
```diff
@@ -392,38 +414,42 @@ class GolangContainerSpec extends AnyFunSpec with ForAllTestContainer {
...
 
-  /**
-   * Copies data from a file inside the test container to a temporary file.
-   * @param baseFileName the file name without a suffix of the files to copy from within the container
-   * @param suffix of the file to copy from the container
-   * @return
+  /** Copies data from a file inside the test container to a temporary file.
+   *  @param baseFileName the file name without a suffix of the files to copy from within the container
+   *  @param suffix of the file to copy from the container
+   *  @return
```
talSofer, line 427:
```suggestion
  /**
   *  Copies data from a file inside the test container to a temporary file.
   *  @param baseFileName the file name without a suffix of the files to copy from within the container
   *  @param suffix of the file to copy from the container
   *  @return
```

clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
```diff
@@ -392,38 +414,42 @@ class GolangContainerSpec extends AnyFunSpec with ForAllTestContainer {
...
+    )
 
-  /**
-   * Lightweight fixture for running particular tests with SSTables and JSON fules generated by a go application on
-   * test startup. The fixture uses sstables as the parser input, and the json as the source of the expected output of
-   * the parsing operation.
+  /** Lightweight fixture for running particular tests with SSTables and JSON fules generated by a go application on
+   *  test startup. The fixture uses sstables as the parser input, and the json as the source of the expected output of
+   *  the parsing operation.
    */
```
talSofer, line 448:
```suggestion
  /** 
   *  Lightweight fixture for running particular tests with SSTables and JSON fules generated by a go application on
   *  test startup. The fixture uses sstables as the parser input, and the json as the source of the expected output of
   *  the parsing operation.
   */
```

talSofer (approved):
Looks good!

clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
arielshaqed: Not done.  We actually *enforce* `scalafmtCheck`, so I ran `sbt scalafmt`.  I refuse to have anything further to do with formatting once the automated tool has done its thing.

Consider opening a PR to fix whatever `scalafmt` options are causing this, or to run just the one `scalafmtCheck` (I think that it runs it twice, once for each Scala version), or anything else.  Just out of scope for this PR.

clients/spark/core/src/test/scala/io/treeverse/jpebble/BlockParserSpec.scala
arielshaqed: As above, out of scope.

arielshaqed (commented):
Thanks!

Issue #2756, opened 2021-11-25
Documention site broken links found


nopcoder:
https://docs.lakefs.io/setup/hooks.html -> OpenAPI link to https://docs.lakefs.io/setup/reference/api.md
https://docs.lakefs.io/setup/hooks.html -> lakectl link to https://docs.lakefs.io/setup/reference/commands.md/#lakectl-actions
https://docs.lakefs.io/setup/storage/blob.html -> installing lakeFS link to https://docs.lakefs.io/setup/storage/install.md
https://docs.lakefs.io/integrations/kubeflow.html -> treeverse/lakectl link to https://docs.lakefs.io/integrations/(https://hub.docker.com/r/treeverse/lakectl)

PR #2759, opened 2021-11-25
docs: fix broken links

nopcoder:
Fix: #2756

### DIFF
--- a/docs/setup/storage/blob.md
+++ b/docs/setup/storage/blob.md
@@ -19,7 +19,8 @@ has_children: false
 {: .no_toc }
 
 In case you want to use the secret key for authentication you will need to use the account key in the configuration
-Go to the `Access Keys` tab and click on `Show Keys` save the values under `Storage account name` and `Key` we will need them in the [installing lakeFS](install.md) step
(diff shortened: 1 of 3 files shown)


nopcoder:
@johnnyaug I'm just not sure about the `docs/setup/storage/blob.md` - not the link is the issue, but actual just to this step.

docs/setup/storage/blob.md
```diff
@@ -19,7 +19,8 @@ has_children: false
 {: .no_toc }
 
 In case you want to use the secret key for authentication you will need to use the account key in the configuration
-Go to the `Access Keys` tab and click on `Show Keys` save the values under `Storage account name` and `Key` we will need them in the [installing lakeFS](install.md) step
+Go to the `Access Keys` tab and click on `Show Keys` save the values under `Storage account name` and `Key` we will need them in the [installing lakeFS](../../quickstart/installing.md) step
```
johnnyaug, line 22: If I understand correctly, what we meant here is:

```
If you want lakeFS to authenticate with your storage using the storage account key, go to the `Access Keys` tab and click `Show Keys`. Use the values under `Storage account name` and `Key` in the [lakeFS configuration](https://docs.lakefs.io/deploy/azure.html#on-azure-vm)
```

johnnyaug (approved):

PR #2770, opened 2021-12-01
Ignore .env file working with the Everything Bagel compose file

talSofer:


deployments/compose/.gitignore
```diff
@@ -0,0 +1 @@
+.env
```
arielshaqed, line 1: Maybe put this in a git ignore for the entire repo?

arielshaqed (approved):
Cool, very useful!

deployments/compose/.gitignore
talSofer: I've considered that but some parts of our repo rely on .env files. for example https://github.com/treeverse/lakeFS/blob/master/test/spark/.env

WDYT?

arielshaqed: Hmmm... so maybe _not_ rely on the default file, but pass that explicitly with a CLI flag?

nopcoder: I also think we need to switch to use .env in the repo level and fix the spark test not to use it.

nopcoder (approved):
Approved because it is serve what we wanted

deployments/compose/.gitignore
talSofer: There are many other places that use .env files. let me look at it

arielshaqed: > There are many other places that use .env files. let me look at it

I'm saying essentially that I would like us **not** to use this default.  It is suitable for interactive use, less so for committed code.  If a docker-compose uses an env file and is stored in the VCS, then that file is liable to be checked in.  At the very least it should not be a hidden file.

So I _do_ consider ignoring all `.env` files (the docker-compose default name) part of this task.  And I further believe that we should rename all other `.env` -> `env`, and use the [`--env-file` flag](https://docs.docker.com/compose/env-file/) to refer to that file _explicitly_.

talSofer: Done

nopcoder (approved):
LGTM

Issue #2235, opened 2021-07-13
Link to setup on Auth Login Page


adunuthulan:
When running lakeFS without easy access to the logs (like when lakeFS is deployed on the cloud), it’s not immediately clear how to reach the setup page from the initial auth ui page. Should there be a link to the setup page on this page? Should the user be redirected to the setup page on their first run of lakeFS?

![Screen Shot 2021-07-13 at 10 47 04 AM](https://user-images.githubusercontent.com/29340929/125500663-d62d4c20-aa27-4cff-b5bc-8a5d1437be79.png)

johnnyaug:
Similar: #2388

PR #2772, opened 2021-12-01
UI: Auto-redirect according to whether lakeFS has been setup

johnnyaug:
When going to the login page when lakeFS has not yet been setup - redirect to the setup page.

When going to the setup page when lakeFS has already been setup - redirect to "/".
Solves #2235.


api/swagger.yml
```diff
@@ -508,7 +508,11 @@ components:
           type: string
       required:
         - id
-
+    SetupState:
+      type: object
+      properties:
+        initialized:
+          type: boolean
```
arielshaqed, line 515: I don't like booleans for something like this, because they have only 2 values...  E.g. suppose we had a setup _wizard_, and you could be half set up.

Could we instead have an enum (i.e. a string with known names)?

webui/src/lib/api/index.js
```diff
@@ -710,6 +710,18 @@ class Retention {
...
 class Setup {
+    async getState() {
+        const response = await apiRequest('/setup_lakefs', {
+            method: 'GET',
+        });
+        switch (response.status) {
+            case 200:
+                return response.json();
+            default:
+                throw new Error('Unknown');
```
arielshaqed, line 721: This error is somewhat uninformative.  Can we include `response` on it somehow (and maybe get the status message into the error string)?

webui/src/pages/auth/login.jsx
```diff
@@ -1,4 +1,4 @@
-import React from "react";
+import React, {useEffect} from "react";
```
arielshaqed, line 1: Not sure where we use `useEffect`.

arielshaqed (approved):
Neat!

Would prefer an API that returns an enum to an API that returns a boolean.

As discussed F2F, it may be better to have the same path for both "login" and "setup".  While it is true that having a descriptive URL slideshow (first I go to the basic URL, then I get redirected to "setup"), as soon as we have both routes we get to worry about whether the 2 will loop around between each other if the server cannot make up its mind.
This is more an abstract worry than something that seems likely, so definitely approving and your call.

arielshaqed (approved):
StillLGTM!

Thanks for bearing with me :-)

guy-har (approved):
Looks good!

Issue #2773, opened 2021-12-03
Insufficient check for Distinguished Name login (ldap authentication)


roman-komac:
While testing LakeFS I have discovered that the LDAP authentication fails with error="create backing user for LDAP user..." 
This comes as a result of a regular expression check that seems to be too strict, as a distinguished name can contain almost any character ([with the exception of some that have to be escaped](https://ldapwiki.com/wiki/DN%20Escape%20Values)).
The validation check therefore fails when any whitespace character is present, e.g.
CN=Name Surname,OU=ResearchAndDev,OU=Users,DC=company,DC=com
as the current implementation does a rudimentary RegEx evaluation against
https://github.com/treeverse/lakeFS/blob/df12f179ad4c595c97ffdfd739f6d500e4faf2c0/pkg/auth/model/validation.go#L15

On the subject of distinguished names:
While case-sensitivity for distinguished names is [implementation dependent](https://ldapwiki.com/wiki/Distinguished%20Name%20Case%20Sensitivity), most implementations treat them as case insensitive,
meaning that the following should be regarded as equal:
CN=Name Surname,DC=company,DC=com
cn=Name Surname,Dc=company,dC=com
Furthermore the corresponding relative distinguished names only have to appear
in the [same order relative to each other](https://ldapwiki.com/wiki/DistinguishedNameMatch). If I understood the ldap wiki entry correctly this example should also pass equality:
CN=Name Surname,OU=ResearchAndDev,OU=Users,DC=company,DC=com
CN=Name Surname,DC=company,DC=com,OU=ResearchAndDev,OU=Users

Please consider changing the distinguished name validation rules to at least include all permitted characters.

arielshaqed:
Thanks, @roman-komac! Clear oversights on my part 😞.

I would rank validation as more urgent (and,  thankfully, easier...) than equality. There is no workaround for not being able to log in; meanwhile (I _expect_, but with no empirical backing) that most LDAP lookups will return DNs in some fixed ordering.

Would you agree?

roman-komac:
@arielshaqed  Thank you for addressing this bug and apologies for the late reply, I somehow overlooked the mail notification over the weekend.
I would agree that validation should have much higher priority. The only edge case that I can really think of is if the Ldap directory was rebuilt from scratch. In that case the ordering might change.

Kudos also to @nopcoder, I am pleasantly surprised by how quickly this issue was resolved 😃

arielshaqed:
[...]

PR #2774, opened 2021-12-05
Auth entity validate not empty

nopcoder:
Remove pattern validation for auth entity ID, also known as username.

Part of #2773.

### DIFF
--- a/pkg/auth/model/validation.go
+++ b/pkg/auth/model/validation.go
@@ -14,7 +14,7 @@ var (
 
 func ValidateAuthEntityID(name string) error {
 	if len(name) == 0 {
-		return ErrValidationError
+		return fmt.Errorf("empty name: %w", ErrValidationError)
 	}
 	return nil
 }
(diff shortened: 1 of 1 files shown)


pkg/auth/model/validation.go
```diff
@@ -3,20 +3,17 @@ package model
...
 var (
 	ErrValidationError = errors.New("validation error")
-
-	EntityIDRegexp = regexp.MustCompile(`^[\w+=.,@\-]{1,127}$`)
 )
 
 func ValidateAuthEntityID(name string) error {
-	if !EntityIDRegexp.MatchString(name) {
+	if len(name) == 0 {
 		return ErrValidationError
```
arielshaqed, line 17: maybe give more details?

```suggestion
	if len(name) == 0 {
		return fmt.Errorf("empty name: %w", ErrValidationError)
```

arielshaqed (approved):
Thanks!

nopcoder:
> Thanks!

Thank you!

PR #2778, opened 2021-12-05
design of the export functionality

lynnro314:


design/export-functionality.md
```diff
@@ -0,0 +1,66 @@
...
+## Possible Solutions
+
+### Spark client stand-alone
+
+Create a standalone Docker container that runs the spark client. 
+
+Pros:
+- Utilizes the existing spark client, and gives the same functionality. 
+ 
+Cons:
```
nopcoder, line 24: Think we need to explain better what is "effort" or "adjustments" in this option

design/export-functionality.md
```diff
@@ -0,0 +1,66 @@
...
+- Not a built-in command in lakeFS, therefore require some adjustments from the users.
+  - This solution probably requires the greatest effort from the users.
+
+
+### Using the java client
+Duplicate the spark client functionality and create an Exporter class that doesn't use spark. This class can be used with the java client. 
+
+Pros:
+- An intermediate solution, that require some little adjustments from the user, but also compatible with the issue.
+Cons:
```
nopcoder, line 34: Maintain another code repo

design/export-functionality.md
```diff
@@ -0,0 +1,66 @@
...
+- Not a built-in command in lakeFS, therefore require some adjustments from the users
+- Require checking how spark is being used in the current implementation (in order to rewrite its functions). Not necessarily possible in a naive way.
+
+
+### Using Rclone
+[Rclone](https://rclone.org/) is a command line program to sync files and directories between cloud providers.
+Users can use Rclone to copy files from lakeFS to a designated object store location, or use Rclone's sync command which is similar to export only the diff between a specific branch and a commit reference.
+
+Pros:
```
nopcoder, line 43: Simple

design/export-functionality.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Using the java client
+Duplicate the spark client functionality and create an Exporter class that doesn't use spark. This class can be used with the java client. 
+
+Pros:
+- An intermediate solution, that require some little adjustments from the user, but also compatible with the issue.
+Cons:
+- Not a built-in command in lakeFS, therefore require some adjustments from the users
+- Require checking how spark is being used in the current implementation (in order to rewrite its functions). Not necessarily possible in a naive way.
+
```
itaiad200, line 37: Also - developing and maintaining a second export functionality.

design/export-functionality.md
```diff
@@ -0,0 +1,66 @@
...
+### Spark client stand-alone
+
+Create a standalone Docker container that runs the spark client. 
+
+Pros:
+- Utilizes the existing spark client, and gives the same functionality. 
+ 
+Cons:
+- Not a built-in command in lakeFS, therefore require some adjustments from the users.
```
itaiad200, line 25: When you mention `adjustments`, can you be more explicit? How do you imagine the user using that option vs. what she has today.

design/export-functionality.md
```diff
@@ -0,0 +1,66 @@
...
+[Rclone](https://rclone.org/) is a command line program to sync files and directories between cloud providers.
+Users can use Rclone to copy files from lakeFS to a designated object store location, or use Rclone's sync command which is similar to export only the diff between a specific branch and a commit reference.
+
+Pros:
+- Rclone gives the functionality of exporting data from lakeFS to a designated object store location. Doesn't require any additional work. 
+
+Cons:
+- Not a built-in command in lakeFS, therefore require some adjustments from the users.
+- Doesn't support success/failure indications on the exported objects.
+- Require a designated object store location that doesn't contain any other data but the data associated to the lakeFS branch. That is because Rclone's sync command will delete all files that don't exist in the branch.
```
itaiad200, line 49: Also no support for diff import..

design/export-functionality.md
```diff
@@ -0,0 +1,66 @@
...
+- Require a designated object store location that doesn't contain any other data but the data associated to the lakeFS branch. That is because Rclone's sync command will delete all files that don't exist in the branch.   
+
+
+### Add export functionality to lakectl
+
+Pros:
+- A built-in command in lakeFS, doesn't require adjustments from the users.
+
+Cons:
+- Require a lot of work and adjustments, may not be compatible with the issue.
```
itaiad200, line 58: What do you refer to as the "adjustments" here? In the `Pros` you wrote "doesn't require adjustments from the users."

design/export-functionality.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Add export functionality to lakeFS (like lakefs run)
```
itaiad200, line 60: Is this option exactly like the lakectl one? From the pros and cons it looks like it does.

itaiad200 (changes requested):
Thanks!

I think the possible solutions section needs some more clarifications. Like:
1. Where should the user run it and how.
2. An example usage.
3. What's the expected inputs from lakeFS (if any) including some that are missing today.
4. Failures recovery.
5. Support for continuous export.
6.

design/export-functionality.md
```diff
@@ -0,0 +1,66 @@
...
+Pros:
+- Utilizes the existing spark client, and gives the same functionality. 
+ 
+Cons:
+- Not a built-in command in lakeFS, therefore require some adjustments from the users.
+  - This solution probably requires the greatest effort from the users.
+
+
+### Using the java client
+Duplicate the spark client functionality and create an Exporter class that doesn't use spark. This class can be used with the java client.
```
talSofer, line 30: Do you mean to add an export capability to the generated lakeFS Java client? that is, only to the Java client?

design/export-functionality.md
```diff
@@ -0,0 +1,66 @@
...
+- Not a built-in command in lakeFS, therefore require some adjustments from the users
+- Require checking how spark is being used in the current implementation (in order to rewrite its functions). Not necessarily possible in a naive way.
+
+
+### Using Rclone
+[Rclone](https://rclone.org/) is a command line program to sync files and directories between cloud providers.
+Users can use Rclone to copy files from lakeFS to a designated object store location, or use Rclone's sync command which is similar to export only the diff between a specific branch and a commit reference.
+
+Pros:
+- Rclone gives the functionality of exporting data from lakeFS to a designated object store location. Doesn't require any additional work.
```
talSofer, line 44: > Doesn't require any additional work.

Except for proper documentation that suggests using rclone for export in a non-distributed environment.

design/export-functionality.md
```diff
@@ -0,0 +1,66 @@
...
+Pros:
+- Rclone gives the functionality of exporting data from lakeFS to a designated object store location. Doesn't require any additional work. 
+
+Cons:
+- Not a built-in command in lakeFS, therefore require some adjustments from the users.
+- Doesn't support success/failure indications on the exported objects.
+- Require a designated object store location that doesn't contain any other data but the data associated to the lakeFS branch. That is because Rclone's sync command will delete all files that don't exist in the branch.   
+
+
+### Add export functionality to lakectl
```
talSofer, line 52: Exposing the command in lakectl is how we make this accessible to the user but it does not imply how the actual export is implemented. what workflow will the lakectl export command trigger? 

We are discussing two orthogonal aspects of the problem:
- How to implement export that satisfies the requirements above 
- How to make the new export functionality accessible to the user. Things that will affect the second decision is understanding if users are ok with running export outside of lakeFS with something that's not Spark or do they expect this to be part of what lakeFS provides? My understanding is that the first is correct.

talSofer (changes requested):
Thanks @lynnro314 !

The solutions require some more details, and you may want to draw the distinction between the possible export implementations and the ways to expose the export functionality (see my comment inline)

design/export-functionality.md
```diff
@@ -0,0 +1,66 @@
+# Export
+
+## Requirements
+The functionality currently provided by the lakeFS Export utility but without a dependency on Apache Spark.
+This includes:
+1. Export all data from a given lakeFS reference (commit or branch) to a designated object store location.
+2. Export only the diff between a specific branch and a commit reference.
```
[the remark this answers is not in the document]
lynnro314: Diff between the HEAD of a branch and a commit. Hope it's more clear.

design/export-functionality.md
lynnro314: Actually it does support for diff export, with using the sync command. I wrote it in the first paragraph, but changes it to be more clear.

design/export-functionality.md
lynnro314: Done

design/export-functionality.md
lynnro314: Yes

design/export-functionality.md
```diff
@@ -0,0 +1,85 @@
...
+The functionality currently provided by the [lakeFS Export utility](https://docs.lakefs.io/reference/export.html) but without a dependency on Apache Spark.
+This includes:
+1. Export all data from a given lakeFS reference (commit or branch) to a designated object store location.
+1. Export only the diff between the HEAD of a branch and a commit reference (helpful for continuous exports).
+1. Success/Failure indications on the exported objects. 
+
+
+## Non-Requirements
+1. Support distribution across machines.
+1. The functionality doesn't need to be supported from lakeFS.
```
talSofer, line 13:
```suggestion
1. Support from lakeFS.
```

design/export-functionality.md
```diff
@@ -0,0 +1,85 @@
...
+In order to use Docker containers, users will need to download it if necessary, learn it, and have it in their toolset.
+
+Usage example:
+
+`docker run lakefs-export --conf spark.hadoop.lakefs.api.url=https://<LAKEFS_ENDPOINT>/api/v1 \
+--conf spark.hadoop.lakefs.api.access_key=<LAKEFS_ACCESS_KEY_ID> \
+--conf spark.hadoop.lakefs.api.secret_key=<LAKEFS_SECRET_ACCESS_KEY> \
+--packages io.lakefs:lakefs-spark-client-301_2.12:0.1.0 \
+--class io.treeverse.clients.Main export-app example-repo s3://example-bucket/prefix \
+--branch=example-branch`
```
talSofer, line 31:
```suggestion
```docker run lakefs-export --conf spark.hadoop.lakefs.api.url=https://<LAKEFS_ENDPOINT>/api/v1 \
--conf spark.hadoop.lakefs.api.access_key=<LAKEFS_ACCESS_KEY_ID> \
--conf spark.hadoop.lakefs.api.secret_key=<LAKEFS_SECRET_ACCESS_KEY> \
--packages io.lakefs:lakefs-spark-client-301_2.12:0.1.0 \
--class io.treeverse.clients.Main export-app example-repo s3://example-bucket/prefix \
--branch=example-branch```
```

talSofer (commented):
Great document @lynnro314! 
Reading the solutions you suggested I think that implementing the standalone Spark docker is most cost-effective and requires insignificant effort from the users. if they are using docker. 

Do you know if the users that asked for this feature are ok with running docker? and the same question for rclone?

design/export-functionality.md
nopcoder: adding to the above - use ```shell
in order to have syntax highlight for the docker run command

design/export-functionality.md
```diff
@@ -0,0 +1,85 @@
...
+
+In order to use java code, users will need to download it if necessary, learn it, and have it in their toolset.
+
+Usage example:
+
+`val exporter = new Exporter(apiClient, repoName, rootLocation)`
+
+`exporter.exportAllFromBranch(branchName)`
+
+`exporter.exportFrom(branchName, commitRef)`
```
nopcoder, line 53: wrap java code with: ```java
```java
val exporter = new Exporter(apiClient, repoName, rootLocation)
exporter.exportAllFromBranch(branchName)
exporter.exportFrom(branchName, commitRef)
```

design/export-functionality.md
```diff
@@ -0,0 +1,85 @@
...
+- Utilizes the existing spark client, and gives the same functionality. 
+ 
+Cons:
+- Require creating and maintaining the standalone container.
+
+### Using the java client
+Reimplement the spark client export functionality by creating a new Exporter class that doesn't use spark.
+This class can be part of the java cliet, or a new "Export" client that uses the java client.
+
+The usage will be with java code (in a similar way of using custom code with the spark client).
```
nopcoder, line 43:
```suggestion
The usage will be with Java code (in a similar way of using custom code with the spark client).
```

design/export-functionality.md
```diff
@@ -0,0 +1,85 @@
...
+The copy command can be used to copy files from lakeFS to a designated object store location.
+The sync command can be used to export only the diff between a specific branch and a commit reference, since sync makes the source and the dest identical (modifying destination only).
+
+In order to use Rclone, users will need to download it if necessary, learn it, and have it in their toolset.
+
+Usage example:
+
+`rclone copy source:lakefs:example-repo/main/ dest:s3://example-bucket/prefix`
+
+`rclone sync source:lakefs:example-repo/main/ dest:s3://example-bucket/prefix`
```
nopcoder, line 76:
```suggestion
```sh
$ rclone copy source:lakefs:example-repo/main/ dest:s3://example-bucket/prefix
$ rclone sync source:lakefs:example-repo/main/ dest:s3://example-bucket/prefix
```
```

**Review by nopcoder — COMMENTED** [review]
Great, will you be adding which suggestion design we would like to use?

File: design/export-functionality.md [hunk 18]
```diff
@@ -0,0 +1,93 @@
...
+```
+
+Pros:
+- Rclone gives the functionality of exporting data from lakeFS to a designated object store location. Require only few additional work of documentation for the usage of Rclone as an export functionality.
+- Simple to use. 
+- Doesn't require maintaining a new feature.
+
+Cons:
+- Doesn't support success/failure indications on the exported objects.
+- Require a designated object store location that doesn't contain any other data but the data associated to the lakeFS branch. That is because Rclone's sync command will delete all files that don't exist in the branch.
```
arielshaqed, line 93: 1. AFAIK deleting missing files is desired behaviour of export.
1. If you truly want only to copy,  won't `rclone copy` do the job?

design/export-functionality.md
```diff
@@ -0,0 +1,93 @@
...
+rclone sync source:lakefs:example-repo/main/ dest:s3://example-bucket/prefix
+```
+
+Pros:
+- Rclone gives the functionality of exporting data from lakeFS to a designated object store location. Require only few additional work of documentation for the usage of Rclone as an export functionality.
+- Simple to use. 
+- Doesn't require maintaining a new feature.
+
+Cons:
+- Doesn't support success/failure indications on the exported objects.
```
arielshaqed, line 92: Could we `rclone check` in case of failure?

arielshaqed (commented):
Thanks!

design/export-functionality.md
lynnro314: I meant that if you have files in the destination location that aren't related to the files in lakeFS - the sync command will delete those files. 
That is the reason users should allocate a designated object store location to export the files from lakeFS.

design/export-functionality.md
lynnro314: Yes this is quite similar, thanks! I will update the doc.

talSofer (approved):
Great work @lynnro314!

design/export-functionality.md
arielshaqed: As discussed f2f, this is either a bug (I kept files in the same "subtree" used for export, but didn't mean to delete those files!)  or a feature (after I export, I want the destination s3 prefix to look exactly like lakeFS does). You're not going to please everyone here.

FTR, what our Spark-based exporter does is a third thing, that relies on separately managing the "last exported revision". I f believe that only makes sense for large-scale installations - and the sync option is close enough if we document that the entire subtree is intended only for export.

design/export-functionality.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Chosen solution
+
+Wrap Rclone with a bash script to match the behavior of the new export functionality to the one of the Spark client.
+This solution utilizes Rclone and its features, it is easy to implement and achives the required behavior.
+
+For example:
+
+```shell
+$ bash lakefs_export.sh source=lakefs:example-repo/main/ dest=s3://example-bucket/prefix export_diff=true
```
arielshaqed, line 106:
```suggestion
$ bash lakefs_export.sh source=lakefs://example-repo/main/ dest=s3://example-bucket/prefix export_diff=true
```

design/export-functionality.md
```diff
@@ -0,0 +1,114 @@
...
+- Simple to use. 
+- Doesn't require maintaining a new feature.
+
+Cons:
+- Require a designated object store location that doesn't contain any other data but the data associated to the lakeFS branch. That is because Rclone's sync command will delete all files that don't exist in the branch.   
+
+
+## Chosen solution
+
+Wrap Rclone with a bash script to match the behavior of the new export functionality to the one of the Spark client.
```
arielshaqed, line 100: Please consider Python rather than bash. Shell scripts are notoriously hard to write correctly.

arielshaqed (approved):
Nice,  thanks!

design/export-functionality.md
nopcoder, line 100: in the rclone section it says the user needs to download and use rclone.
Does our solution still require that?
Can you add a note about how our solution is going to be delivered?

nopcoder (approved):

design/export-functionality.md
lynnro314: Thanks, I'm checking this option.

lynnro314:
Forced merged since all changes requested were addressed and the PR was approved.

PR #2780, opened 2021-12-06
Use https links when possible in code and code examples

nopcoder:
Prefer to reference https links in our code to documentation and other references

### DIFF
--- a/pkg/auth/wildcard/match.go
+++ b/pkg/auth/wildcard/match.go
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *     https://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
--- a/pkg/auth/wildcard/match.go
+++ b/pkg/auth/wildcard/match.go
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     https://www.apache.org/licenses/LICENSE-2.0
+ *     http://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


pkg/auth/wildcard/match.go
```diff
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *     https://www.apache.org/licenses/LICENSE-2.0
```
arielshaqed, line 8: This is a verbatim copy of the license (supplied in the copy in MinIO, I believe).  Please revert this change.

pkg/gateway/errors/errors.go
```diff
@@ -14,7 +14,7 @@ import (
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *     https://www.apache.org/licenses/LICENSE-2.0
```
arielshaqed, line 17: Also here.

pkg/gateway/sig/sig.go
```diff
@@ -32,7 +30,7 @@ var (
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *     https://www.apache.org/licenses/LICENSE-2.0
```
arielshaqed, line 33: ... & here...

pkg/gateway/sig/v4_streaming_reader.go
```diff
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *     http://www.apache.org/licenses/LICENSE-2.0
+ *     https://www.apache.org/licenses/LICENSE-2.0
```
arielshaqed, line 8: ... & here...

arielshaqed (changes requested):
Thanks!

Requesting reverting the in-license links because legal, other than that this is a great change!

arielshaqed (approved):
👍 😺

Issue #2768, opened 2021-12-01
Delete multiple objects using lakeFS API


nopcoder:
Having to delete a large number of objects requires multiple calls to the server.
Currently alternatives will be to use the client side implementation using lakectl recursive delete or the S3 gateway delete-objects request.
Our open API is missing the use-case of deleting a large set of specific objects.

The API should enable the same functionality as the s3 delete-objects.
When all objects are delete it will return 204 No Content to indicate that all objects where deleted.
In case there is an error 200 OK will be return with a list of responses where the caller can identify which object failed to delete.
The operation is not atomic and in case of any error all objects that were deleted will not be restored as part of the request.

ozkatz:
for reference, the documentation of S3's bulk delete API: [DeleteObjects](https://docs.aws.amazon.com/AmazonS3/latest/API/API_DeleteObjects.html).

A few notes about it:
1. it always returns a successful response. Errors, if any exist in the response body, so it must be parsed.
2. it supports a "quiet" mode, in which successful deletions are not returned in the response body - only errors are returned.
3. the default mode is "verbose" in which all keys are returned, regardless of their success/error status
4. just like deleting a single object, a successful deletion does not indicate the object existed at the time of deletion.

adijesori:
It would also be nice to add support for query option called “failFast” that will cause the request to return immediately when one of the object deletions were failed.

PR #2788, opened 2021-12-12
Delete objects open-api

nopcoder:
Fix https://github.com/treeverse/lakeFS/issues/2768

Adding delete object to lakeFS open-api.
New route under objects that uses POST to send a list of paths to delete (max 1000 items)
Return 204 (no content) in case delete was successful
Return 200 and list of errors - each one with key, status code and message in case delete fails.
Note that object not found is not a failure and will not be reported per object.


api/swagger.yml
```diff
@@ -198,6 +198,17 @@ components:
...
           example: "main"
 
+    PathList:
+      type: object
+      required:
+        - paths
+      properties:
+        paths:
+          type: array
+          items:
```
arielshaqed, line 208: Not sure what "items" means.

api/swagger.yml
```diff
@@ -478,6 +489,32 @@ components:
...
+      properties:
+        status_code:
+          type: integer
+          description: http status code associated with the path/operation
+        path:
+          type: string
+          description: path to the associated object
+        message:
+          type: string
+          description: short message explaining the error
```
arielshaqed, line 506:
```suggestion
        path:
          type: string
          description: affected path
        status_code:
          type: integer
          description: HTTP status code associated for operation on path
        message:
          type: string
          description: short message explaining status_code
```

arielshaqed: This puts code and message together, and also changes wording slightly.

api/swagger.yml
```diff
@@ -2592,6 +2629,45 @@ paths:
...
+              $ref: "#/components/schemas/PathList"
+      responses:
+        200:
+          description: Delete objects response
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/ObjectErrorList"
+        204:
+          description: object deleted successfully
```
arielshaqed, line 2663:
```suggestion
          description: all requested objects successfully deleted
```
(?)

api/swagger.yml
```diff
@@ -2592,6 +2629,45 @@ paths:
...
+        401:
+          $ref: "#/components/responses/Unauthorized"
```
arielshaqed, line 2665: When do we get these errors?

pkg/api/controller.go
```diff
@@ -76,6 +78,74 @@ type Controller struct {
...
+				Message:    http.StatusText(http.StatusUnauthorized),
+			})
+			continue
+		}
+
+		lg := c.Logger.WithField("path", objectPath)
+		err := c.Catalog.DeleteEntry(ctx, repository, branch, objectPath)
+		switch {
+		case errors.Is(err, catalog.ErrNotFound):
+			lg.Debug("tried to delete a non-existent object")
```
arielshaqed, line 114:
```suggestion
			lg.Debug("tried to delete a non-existent object (OK)")
```

arielshaqed (approved):
Neat; thanks!

api/swagger.yml
nopcoder: Swagger require you to describe each item in the array - the key of the array is 'paths' so each item is a path.
I'll add a description too.

api/swagger.yml
nopcoder: Thanks

api/swagger.yml
nopcoder: auth middleware

PR #2805, opened 2021-12-16
UI testing proposal

### LINKED ISSUE #1812: Discovery: Testing UI
Our webui packages isn't being tested as part of our CI.
This issue is for exploring the different solutions exist for testing the functionality (we're less interested in the design part), and come up with a proposal.

talSofer:
Closes #1812

### DIFF
--- a/design/ui-testing.md
+++ b/design/ui-testing.md
@@ -176,9 +176,10 @@ In the proposed increment, Jest is used as the E2E tests runner. As a second ste
 other React testing libraries to to unit-test the UI. As a third increment, when the UI becomes more stable we can use more advanced features provided by the testing automation framework such as screenshot
 testing, performance testing, etc.
 
-### Conclusion
+### Decisions
 
-Use Jest and Puppeteer. 
+* Use Jest and Puppeteer.
+* Use [Testcontainers](https://www.npmjs.com/package/testcontainers) to spin up containerized lakeFS test instances.
 
 ### References
 * https://medium.com/touch4it/end-to-end-testing-with-puppeteer-and-jest-ec8198145321
--- a/design/ui-testing.md
+++ b/design/ui-testing.md
@@ -143,6 +143,7 @@ The main arguments that turn my recommendation towards **Puppeteer** as the test
 * The lakeFS UI is written in React, and using a framework that lives at the same space is more natural and convenient, 
 as opposed to Selenium which is used for multiple types of applications. 
 * The development experience working with Puppeteer was much better.
+* Online resources testify that Puppeteer executes faster than Selenium.  
 
 ### Why test against a live server?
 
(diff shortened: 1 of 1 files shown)


design/ui-testing.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+As for CI, the [node workflow](../.github/workflows/node.yaml) is already running the webui tests that currently does 
+not exist. after imlementing the tests this workflow will be responsible for running them.   
+
+_**Notes:**_
+1. The decision on the testing automation framework will not affect this section. 
+2. Its TBD to decide exactly how to work with Testcontainers to do the tests setup; what the containers 
+should include, and the frequency of creating and tearing them down (In each test file, single instance for the whole 
+test  suit etc.)
```
johnnyaug, line 170:
```suggestion
test suite etc.)
```

design/ui-testing.md
```diff
@@ -0,0 +1,184 @@
...
+
+### Future improvements
+
+In the proposed increment, Jest is used as the E2E tests runner. As a a second step we can use Jest in combination with
```
johnnyaug, line 174:
```suggestion
In the proposed increment, Jest is used as the E2E tests runner. As a second step we can use Jest in combination with
```

johnnyaug (approved):
Sounds great!
Don't see a reason not to use Puppeteer

arielshaqed (approved):
**Disclaimer:** If it's not obvious, I'm mot a UI or UI testing expert by any means.

Argument for Puppeteer is very compelling.  I do not find "other browsers" a compelling argument for us as long as we test operations rather than what it looks like.  And we do (try to) write standard HTML.  And the next browser to support after Chrome is most likely Safari, which will be hard on CI/CD but easy on Treeverse dev laptops.

About the only thing that seems to me that might block it is if Puppeteer is _significantly_ slower than Selenium -- we will have many tests.  This also applies if Puppeteer takes a lot of memory, which may prevent us from parallelizing our tests.

talSofer:
> **Disclaimer:** If it's not obvious, I'm mot a UI or UI testing expert by any means.
> 
> Argument for Puppeteer is very compelling. I do not find "other browsers" a compelling argument for us as long as we test operations rather than what it looks like. And we do (try to) write standard HTML. And the next browser to support after Chrome is most likely Safari, which will be hard on CI/CD but easy on Treeverse dev laptops.
> 
> About the only thing that seems to me that might block it is if Puppeteer is _significantly_ slower than Selenium -- we will have many tests. This also applies if Puppeteer takes a lot of memory, which may prevent us from parallelizing our tests.

Thanks @arielshaqed! good point. I haven't benchmarked it myself but I see online testimonies that Puppeteer is faster than Selenium. e.g. https://www.blazemeter.com/blog/selenium-vs-puppeteer-for-test-automation and https://blog.testproject.io/2020/02/20/selenium-vs-puppeteer-when-to-choose-what/. 
I added this point to the doc.

itaiad200 (approved):

nopcoder (approved):

ortz (approved):

PR #2808, opened 2021-12-18
Optimize SeekGE in committed iterator

johnnyaug:
When seeking multiple times in the same range, do not instantiate a new iterator to the underlying SSTable.

### DIFF
--- a/pkg/graveler/committed/iterator.go
+++ b/pkg/graveler/committed/iterator.go
@@ -136,23 +136,24 @@ func (rvi *iterator) Close() {
 
 func (rvi *iterator) SeekGE(key graveler.Key) {
 	var err error
-	// TODO(ariels): rangesIt might already be on correct range.
-	rvi.rangesIt.SeekGE(Key(key))
-	if err = rvi.rangesIt.Err(); err != nil {
-		rvi.err = err
-		return
-	}
-	if !rvi.NextRange() {
-		return // Reached end.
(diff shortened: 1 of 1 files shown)


pkg/graveler/committed/iterator.go
```diff
@@ -134,25 +134,30 @@ func (rvi *iterator) Close() {
...
 	rvi.rangesIt.SeekGE(Key(key))
-	if err = rvi.rangesIt.Err(); err != nil {
+	if err := rvi.rangesIt.Err(); err != nil {
 		rvi.err = err
-		return
+		return false
 	}
 	if !rvi.NextRange() {
-		return // Reached end.
+		return false // Reached end.3
```
guy-har, line 144:
```suggestion
		return false // Reached end
```

guy-har (approved):
Verry nice optimization!!
Thanks!

arielshaqed (approved):
Neat.

Do you have a measurement of the number of calls to SeekGE it saves, or of some other calls saved / other speedup?

johnnyaug:
@arielshaqed, we are going to post some more numbers today, but I will add something to the PR description for the record

arielshaqed:
> @arielshaqed, we are going to post some more numbers today, but I will add something to the PR description for the record

Just to write it out in the largest font possible...

# 3x faster!

PR #2811, opened 2021-12-19
lakectl: fix crash when diff is big

johnnyaug:
`lakectl diff` uses exponentially growing page sizes, starting from 50 and doubling on every page.

The server doesn't allow page size > 1000.


arielshaqed (approved):
Cool, sounds legit.

Shouldn't fail if it gets too few diffs in a page, though.

PR #2824, opened 2021-12-21
Align Hadoop version with that used in Spark 2.4.7 builds

arielshaqed:
But will it work with Spark 3.0.x?


nopcoder:
Think the question is not about Spark 3.x, but Hadoop 3.3

nopcoder (approved):

arielshaqed:
Will do this separately -- close this draft.

PR #2825, opened 2021-12-21
Added lakectl CLI command 'annotate' (blame) 

### LINKED ISSUE #2659: CLI: Provide a way to view the latest change made to an object or prefix
### What

being able to take a data object (or prefix, like a partition or collection) and get back the commits that added/modified its content.

### Why

This is valuable lineage information that is currently available in lakeFS but not exposed easily, and mimics the behavior of `git blame`

### How

Given the lakeFS API already supports listing the log of commits for an object or prefix (🎉), this could be a `lakectl` addition that  simply takes the first change from the list.

Additionally, `blame` could be used as an internal alias for those who really want to stick with Git's taxonomy - I would suggest something more positive like `origin`, e.g.:

```
# lakectl origin --prefix lakefs://repo/branch/collection/

lakefs://repo/branch/collection/date=2021/  2021-12-31  7c3cded Spark added 2021 partition (through Airflow)
lakefs://repo/branch/collection/date=2020/  2020-12-31  82ba1e8 Perfromed a rerun for 2020, manually
lakefs://repo/branch/collection/date=2019/  2019-12-31  91f6ab7 Spark added 2019 partition (using cronjob)
```

shimi9276:
Added lakectl CLI command 'annotate' (blame)  providing the ability to know who committed (as well as some basic commit info )  the latest change to an object or list of objects under a given path. This command mimics the Git command 'blame'. (Behaves similarly to the command "fs ls" but with added functionality including info regarding the latest commit to the object or objects.)
No tests have been written.
These manual tests have been performed 

- A basic sanity check. Ran the annotate function on a branch with objects and verified the relevant data was retrieved and properly displayed. This includes prefixes and the call to annotate on an entire branch.
-  Committed an object with a very long name (The longest Windows OS allows). To verify it is properly displayed.
-  Committed an object with /n and a very large message size to verify it is split by the required length.
-  Ran on a non-existing prefix to verify no data was returned and that no failure has occurred. 

**Once issue #2763 will be merged additional testing should be written.**

Close #2659.


arielshaqed (commented):
Note linter comments -- `init` should have an anti-linter comment above it, but naming with underscores is not done.

Mostly worried about missing all  objects that used to live under the prefix.  It's a requirements question what to do here, TBH.

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,52 @@
...
+	Use:   "annotate <path uri>",
+	Short: "Show who created the latest commit",
+	Long:  "Show who created the latest commit to an object or objects in a give path",
+	Args:  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, c_args []string) {
+		pathURI := MustParsePathURI("path", c_args[0])
+		client := getClient()
+		amount := 1
+		pfx := api.PaginationPrefix(*pathURI.Path)
+		res, err := client.ListObjectsWithResponse(cmd.Context(), pathURI.Repository, pathURI.Ref, &api.ListObjectsParams{Prefix: &pfx})
```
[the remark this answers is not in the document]
shimi9276: An anti-inter comment will be added, and the naming will be changed. 
Regarding deleted items, that is correct, i.e. all deleted objects will not be displayed. Just to compare to git blame, git blame requires a file as an argument, in other words, git blame cannot be used on a deleted file, the "annotate" command should be the same by default. In the future, we could add a flag that adds deleted items as well (something on the lines of, --show-deleted-files).

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,53 @@
...
+		DieOnResponseError(res, err)
+		objList := res.JSON200.Results
+		logCommitsParams := &api.LogCommitsParams{
+			Amount: api.PaginationAmountPtr(amount),
+		}
+		pagination := api.Pagination{HasMore: true}
+		for _, obj := range objList {
```
nopcoder, line 29: as far as I remember aws s3 and lakectl object listing continue to list all the object until no mo data is sent or a limit is specified.

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,53 @@
...
+			res, err := client.LogCommitsWithResponse(cmd.Context(), pathURI.Repository, pathURI.Ref, logCommitsParams)
+			DieOnResponseError(res, err)
+			pagination = res.JSON200.Pagination
+			logCommitsParams.After = api.PaginationAfterPtr(pagination.NextOffset)
```
nopcoder, line 33: why do we handle log commits pagination is we ask only for the last entry?

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,53 @@
...
+			data := struct {
+				Commits    []api.Commit
```
nopcoder, line 35: having a single entry - we can simplify the code by passing just a single commit and not let the template handle something it should not render.

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"github.com/spf13/cobra"
+	"github.com/treeverse/lakefs/pkg/api"
+)
+
+const annotateTemplate = `{{ range $val := .Commits}}
+{{ $.Object|ljust 10}} {{ $val.Committer }} {{ $val.Id | printf "%.16s"|ljust 20 }} {{ $val.Message |ljust 30 }} {{ $val.CreationDate|date }}{{ end -}}
+`
+
+var annotateCmd = &cobra.Command{
```
nopcoder, line 12: did we wanted to add an alias called "blame"?

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,53 @@
...
+	Use:   "annotate <path uri>",
+	Short: "Show who created the latest commit",
+	Long:  "Show who created the latest commit to an object or objects in a give path",
+	Args:  cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, c_args []string) {
```
nopcoder, line 17: c_args - non Go naming - suggest to call it `args`

nopcoder (changes requested):

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/treeverse/lakefs/pkg/api"
+)
+
+const annotateTemplate = `{{ range $val := .Commits}}
+{{ $.Object|ljust 10}} {{ $val.Committer }} {{ $val.Id | printf "%.16s"|ljust 20 }} {{ $val.Message |ljust 30 }} {{ $val.CreationDate|date }}{{ end -}}
+`
+
+var annotateCmd = &cobra.Command{
+	Use:   "annotate <path uri>",
+	Short: "Show who created the latest commit",
```
ozkatz, line 14:
```suggestion
	Short: "List entries under a given path, annotating each with the last commit to modify it",
```

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,53 @@
...
+	Long:  "Show who created the latest commit to an object or objects in a give path",
```
ozkatz, line 15: Can probably skip if the `Short` I suggested is OK with you

cmd/lakectl/cmd/annotate.go
ozkatz: this command should be at least consistent with `lakectl fs ls` in behavior. If we do pagination there, we should probably do it here as well.

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,53 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+	"github.com/treeverse/lakefs/pkg/api"
+)
+
+const annotateTemplate = `{{ range $val := .Commits}}
+{{ $.Object|ljust 10}} {{ $val.Committer }} {{ $val.Id | printf "%.16s"|ljust 20 }} {{ $val.Message |ljust 30 }} {{ $val.CreationDate|date }}{{ end -}}
```
ozkatz, line 9: Commit Message is much more likely than other fields to be very long and to contain line breaks, so I suggest:

1. moving it to be the last item per line - it will also help anyone looking to parse the output of the command
1. redacting it after say 100 chars (and adding a trailing "...")
1. in case of a line break (\r\n or \n), redact just before it, with a trailing "...".

ozkatz (changes requested):
Thanks! Great start :) See my comment on the template and short/long descriptions

cmd/lakectl/cmd/annotate.go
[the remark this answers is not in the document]
arielshaqed: I believe that the correct comparison is to `git log` (and to `lakectl log`).  This code gives inconsistent behaviour if I give an exact path or if I give a prefix.  Not sure about `--show-deleted-files`: when would a user *not* want to use this flag?

It is also unlike git, and git is very useful here.

E.g. on our repo I can discover when a file last existed:
```shell
[ariels@fedora lakeFS]$ git log -n 1 -u -- nessie/ops/.env
commit c43ec400f24f610005ca3174e90357c363dc6d4d
Author: talSofer <tal.sofer@treeverse.io>
Date:   Thu Dec 2 11:01:30 2021 +0200

    gitIgnore .env files

diff --git a/nessie/ops/.env b/nessie/ops/.env
deleted file mode 100644
index b62ac132..00000000
--- a/nessie/ops/.env
+++ /dev/null
@@ -1,4 +0,0 @@
-TAG=dev
-LAKEFS_BLOCKSTORE_TYPE=local
-REPO=treeverse
-LAKEFS_ROOT=.
\ No newline at end of file
```

and this works even with just a prefix (`origin/master` at e636d01122416c49acd4424d8f5ae754d1e9c023 here):

```shell
[ariels@fedora lakeFS]$ git log -n 1 -u origin/master  -- nessie/ops/
commit c43ec400f24f610005ca3174e90357c363dc6d4d
Author: talSofer <tal.sofer@treeverse.io>
Date:   Thu Dec 2 11:01:30 2021 +0200

    gitIgnore .env files

diff --git a/nessie/ops/.env b/nessie/ops/.env
deleted file mode 100644
index b62ac132..00000000
--- a/nessie/ops/.env
+++ /dev/null
@@ -1,4 +0,0 @@
-TAG=dev
-LAKEFS_BLOCKSTORE_TYPE=local
-REPO=treeverse
-LAKEFS_ROOT=.
\ No newline at end of file
diff --git a/nessie/ops/docker-compose.yaml b/nessie/ops/docker-compose.yaml
index 3a412daf..29602ab8 100644
--- a/nessie/ops/docker-compose.yaml
+++ b/nessie/ops/docker-compose.yaml
@@ -1,7 +1,7 @@
 version: '3'
 services:
   lakefs:
-    image: "${REPO}/lakefs:${TAG}"
+    image: "${REPO:-treeverse}/lakefs:${TAG:-dev}"
     ports:
[... snipped many lines...]
```

Even this works, if I cannot remember where we have `.env` files and when we deleted them:
```shell
[ariels@fedora lakeFS]$ git log -1 -u -- '**.env'
commit c43ec400f24f610005ca3174e90357c363dc6d4d
Author: talSofer <tal.sofer@treeverse.io>
Date:   Thu Dec 2 11:01:30 2021 +0200

    gitIgnore .env files

diff --git a/nessie/ops/.env b/nessie/ops/.env
deleted file mode 100644
index b62ac132..00000000
--- a/nessie/ops/.env
+++ /dev/null
@@ -1,4 +0,0 @@
-TAG=dev
-LAKEFS_BLOCKSTORE_TYPE=local
-REPO=treeverse
-LAKEFS_ROOT=.
\ No newline at end of file
diff --git a/test/spark/.env b/test/spark/.env
deleted file mode 100644
index db32cccd..00000000
--- a/test/spark/.env
+++ /dev/null
@@ -1,3 +0,0 @@
-TAG=latest
-SPARK_TAG=3
-REPO=treeverse
```

If the issue is that this needs unbounded history search, then let's fix *that* issue or do the minimum possible and fix it later: possible solutions:
- Add a flag to count how far to go back (anyway needed for a long history to find last access to an old unchanged file); this is a good minimal solution.
- Add a flag `--only-at-head` or something that will find the last change on each object that exists at head.
- Agree to take the performance hit.
- Find a way to add speedup data for long historical searches.

My point is that right now **nobody** has long history (only wide history).  So long runtimes are probably not critical right now.  By limiting ourselves to what will be fast for users in a few years' time, we give them less current value.

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,53 @@
...
+		pfx := api.PaginationPrefix(*pathURI.Path)
+		res, err := client.ListObjectsWithResponse(cmd.Context(), pathURI.Repository, pathURI.Ref, &api.ListObjectsParams{Prefix: &pfx})
+		DieOnResponseError(res, err)
+		objList := res.JSON200.Results
+		logCommitsParams := &api.LogCommitsParams{
+			Amount: api.PaginationAmountPtr(amount),
+		}
+		pagination := api.Pagination{HasMore: true}
+		for _, obj := range objList {
+			res, err := client.LogCommitsWithResponse(cmd.Context(), pathURI.Repository, pathURI.Ref, logCommitsParams)
```
arielshaqed, line 30: If we are already focusing on efficiency (I am far from sure that we should be), then there is a potentially massive gain here from logging on more than one object at a time.  Let's open an issue to speed things up here, and add a TODO comment along the lines of
```
// TODO(#9876): Query multiple objects with each Log* call.

File: cmd/lakectl/cmd/annotate.go [hunk 11]
```diff
@@ -0,0 +1,53 @@
...
+			DieOnResponseError(res, err)
+			pagination = res.JSON200.Pagination
+			logCommitsParams.After = api.PaginationAfterPtr(pagination.NextOffset)
+			data := struct {
+				Commits    []api.Commit
+				Object     string
+				Pagination *Pagination
+			}{
+				Commits: res.JSON200.Results,
+				Object:  obj.Path,
```
arielshaqed, line 40: I don't understand this code.  This is the *only* place where `obj` is ever used.  It is never passed to the log query.  So AFAIU this will output each object currently under the prefix, and claim that it was last changed at the latest commit.

Perhaps we should move up adding `lakectl` testing to this code, that will probably allow me to understand better.  In the meantime, can you add in a comment some sequences you used for testing?  (Just copy-paste a full session from your terminal...).

arielshaqed (changes requested):
Thanks!

Worried about:
1. Not understanding the logic, specifically how the server knows which "last commit" to return if we do not pass it the object.
2.  Usability with the current behaviour that special-cases deletions.

cmd/lakectl/cmd/annotate.go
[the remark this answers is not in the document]
shimi9276: I do agree that in the log command a user would generally be intrested in all the info. In blame however, I think many users are intrested only in the latest changes to active objects. If I understood correctly, the issue being described is more of general problem with the speed of the "log" command once many files and commits are in the repo. Maybe a new issue should be opened in order to address this.

cmd/lakectl/cmd/annotate.go
shimi9276: Agree. Will be fixed.

cmd/lakectl/cmd/annotate.go
shimi9276: Good point. Fixed

cmd/lakectl/cmd/annotate.go
shimi9276: Fixed

cmd/lakectl/cmd/annotate.go
shimi9276: Added

cmd/lakectl/cmd/annotate.go
shimi9276: Fixed as well.

cmd/lakectl/cmd/annotate.go
shimi9276: Accepted

cmd/lakectl/cmd/annotate.go
shimi9276: Skipped

cmd/lakectl/cmd/annotate.go
shimi9276: I have made some changes to the code. The above should be solved.

cmd/lakectl/cmd/annotate.go
shimi9276: 1,2. Good point. Done.
3. I do agree with the redact by line breaks, though I think adding a trailing "...." in this case is not needed since the text itself is formatted in the proper form while blame should only add the most basic info of the commit.

cmd/lakectl/cmd/annotate.go
arielshaqed: Please let's also do something about testing this code.  Even if for now it's just a statement on the PR of manual steps performed to test it.

shimi9276: Will do

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,81 @@
...
+	if len(str) > size {
+		str = str[:size] + "..."
+	}
+	return str
+}
+
+func splitOnNewLine(str string) string {
+	str = strings.Split(str, "\\n")[0]
+	return str
+}
```
nopcoder, line 76:
```suggestion
func splitOnNewLine(str string) string {
	return strings.SplitN(str, "\\n", 2)[0]
}
```

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,81 @@
...
+				break
+			}
+			from = pagination.NextOffset
+		}
+	},
+}
+
+const messageSize = 100
+
+func setMessageSize(size int, str string) string {
```
nopcoder, line 66: 1. func name - would call it stringTrimLen(string, len)
2. consider passing the string first
3. should consider the '...' as part of the length

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,81 @@
...
+			pagination := resp.JSON200.Pagination
+			if !pagination.HasMore {
+				break
+			}
+			from = pagination.NextOffset
+		}
+	},
+}
+
+const messageSize = 100
```
nopcoder, line 64: Move constants up above the functions

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,81 @@
...
+					Objects: &prfx,
+				}
+				res, err := client.LogCommitsWithResponse(cmd.Context(), pathURI.Repository, pathURI.Ref, logCommitsParams)
+				DieOnResponseError(res, err)
+				data := struct {
+					Commit        api.Commit
+					Object        string
+					CommitMessage string
+				}{
+					Commit:        res.JSON200.Results[0],
```
nopcoder, line 49: with check the results length before reference it - check and print error and hope it will never happen

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,81 @@
...
+				Prefix: &pfx,
+				After:  api.PaginationAfterPtr(from),
+			}
+			resp, err := client.ListObjectsWithResponse(cmd.Context(), pathURI.Repository, pathURI.Ref, params)
+			DieOnResponseError(resp, err)
+			for _, obj := range resp.JSON200.Results {
+				prfx := []string{obj.Path}
+				logCommitsParams := &api.LogCommitsParams{
+					Amount:  api.PaginationAmountPtr(amount),
+					Objects: &prfx,
```
nopcoder, line 40: api.StringPtr

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,81 @@
...
+	Short: "List entries under a given path, annotating each with the latest modifying commit ",
+	// Long:    "Show who created the latest commit to an object or objects in a give path",
+	Aliases: []string{"blame"},
+	Args:    cobra.ExactArgs(1),
+	Run: func(cmd *cobra.Command, args []string) {
+		pathURI := MustParsePathURI("path", args[0])
+		client := getClient()
+		amount := 1
+		pfx := api.PaginationPrefix(*pathURI.Path)
+		res, err := client.ListObjectsWithResponse(cmd.Context(), pathURI.Repository, pathURI.Ref, &api.ListObjectsParams{Prefix: &pfx})
```
nopcoder, line 25: suggestion: should assign context to a variable - because we use it later

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,81 @@
...
+
+import (
+	"strings"
+
+	"github.com/spf13/cobra"
+	"github.com/treeverse/lakefs/pkg/api"
+)
+
+const annotateTemplate = `{{  $val := .Commit}}
+{{ $.Object|ljust 15}} {{ $val.Committer|ljust 20 }} {{ $val.Id | printf "%.16s"|ljust 20 }} {{ $val.CreationDate|date }}  {{ $.CommitMessage |ljust 30 }}
```
nopcoder, line 11: how is align works with long object path? how the user knows the output format? consider using a table to render the output.

nopcoder (changes requested):

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,81 @@
...
+`
+
+var annotateCmd = &cobra.Command{
+	Use:   "annotate <path uri>",
+	Short: "List entries under a given path, annotating each with the latest modifying commit ",
+	// Long:    "Show who created the latest commit to an object or objects in a give path",
```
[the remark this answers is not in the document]
shimi9276: Done

cmd/lakectl/cmd/annotate.go
shimi9276: Accepted and replaced

cmd/lakectl/cmd/annotate.go
shimi9276: Accepted and fixed.

cmd/lakectl/cmd/annotate.go
shimi9276: Done

cmd/lakectl/cmd/annotate.go
shimi9276: Done

cmd/lakectl/cmd/annotate.go
shimi9276: Not sure what you mean here.

cmd/lakectl/cmd/annotate.go
shimi9276: Done

cmd/lakectl/cmd/annotate.go
shimi9276: In the case the object name is very long it will at most overlap to the next line. The next line will always start on a new line, so the user can always assume the data is starting there.

cmd/lakectl/cmd/annotate.go
nopcoder: I revert my comment - I thought this is a string and you pass a pointer to the API - we have a function for that in 'api' package. But this is not the case - so, ignore.

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,84 @@
...
+			}
+			from = pagination.NextOffset
+		}
+	},
+}
+
+func stringTrimLen(str string, size int) string {
+	if len(str) > size {
+		str = str[:size-3] + "..."
+	}
```
nopcoder, line 73: You can set '...' as constant (ex: dots) and use len(dots) instead of using '3'.
Also you need to condition that len(str) is bigger than len(dots).

cmd/lakectl/cmd/annotate.go
nopcoder: Got it - regarding the align.
About the second part - how the user knows the output format: I think we need to specify the output format if it is not self explained.

cmd/lakectl/cmd/annotate.go
shimi9276: Accepted and fixed.

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,85 @@
...
+		pathURI := MustParsePathURI("path", args[0])
+		client := getClient()
+		amount := 1
+		pfx := api.PaginationPrefix(*pathURI.Path)
+		context := cmd.Context()
+		res, err := client.ListObjectsWithResponse(context, pathURI.Repository, pathURI.Ref, &api.ListObjectsParams{Prefix: &pfx})
+		DieOnResponseError(res, err)
+		var from string
+		for {
+			pfx := api.PaginationPrefix(*pathURI.Path)
```
nopcoder, line 32: don't need another pfx. already have one in scope.

nopcoder (changes requested):

cmd/lakectl/cmd/annotate.go
shimi9276: Good point. Fixed.

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,84 @@
...
+						Object        string
+						CommitMessage string
+					}{
+						Commit:        res.JSON200.Results[0],
+						Object:        obj.Path,
+						CommitMessage: splitOnNewLine(stringTrimLen((res.JSON200.Results[0].Message), messageSize)),
+					}
+					Write(annotateTemplate, data)
+				} else {
+					fmt.Println("Error, No commit info has been retrieved from the api call")
```
nopcoder, line 58: Sorry about this late response on this one.
Write out error without explaining the error or in this case there is unexpected result without an error.
Also there is no information associated with the error - the user will not understand which object didn't hold any commit.
I would recommend just passing an empty commit to the template in this case.
Logging a warning and/or in case of an error print out the full details without "api call" part as the user doesn't know what to do with it.

shimi9276: Decided to print out just "Error" whithout specification this way the user knows that not all the data was received. Passing an empty commit is more complex while not nessareally letting the user know the some commits are missing.

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,84 @@
...
+				Prefix: &pfx,
+				After:  api.PaginationAfterPtr(from),
+			}
+			resp, err := client.ListObjectsWithResponse(context, pathURI.Repository, pathURI.Ref, params)
+			DieOnResponseError(resp, err)
+			for _, obj := range resp.JSON200.Results {
+				path := []string{obj.Path}
+				logCommitsParams := &api.LogCommitsParams{
+					Amount:  api.PaginationAmountPtr(amount),
+					Objects: &path,
```
arielshaqed, line 42: For a large prefix this can be quite inefficient.  I believe that in many cases something like the following will be (much) faster:

* For each 1000 objects, put them in a set, and starting from HEAD commit...
   * Take the _latest_ earlier commit that mentions _any_ object(s) in that set.  [Single API call]
   * Link each of those objects in the set that occur on that commit to that commit.  [Just local processing]
   * Remove the linked objects from the set
   * If set is non-empty, go back to fetching the latest commit.
* Continue with next 1000 objects.

This minimizes API calls by several 10x or 100x in most cases and _never_ increases the number of API calls.

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,84 @@
...
+
+	"github.com/spf13/cobra"
+	"github.com/treeverse/lakefs/pkg/api"
+)
+
+const annotateTemplate = `{{  $val := .Commit}}
+{{ $.Object|ljust 15}} {{ $val.Committer|ljust 20 }} {{ $val.Id | printf "%.16s"|ljust 20 }} {{ $val.CreationDate|date }}  {{ $.CommitMessage |ljust 30 }}
+`
+const messageSize = 100
+const trailingThreeDots = "..."
```
arielshaqed, line 15: Nit:

```suggestion
const ellipsis = "..."
```

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,84 @@
...
+	"strings"
+
+	"github.com/spf13/cobra"
+	"github.com/treeverse/lakefs/pkg/api"
+)
+
+const annotateTemplate = `{{  $val := .Commit}}
+{{ $.Object|ljust 15}} {{ $val.Committer|ljust 20 }} {{ $val.Id | printf "%.16s"|ljust 20 }} {{ $val.CreationDate|date }}  {{ $.CommitMessage |ljust 30 }}
+`
+const messageSize = 100
```
arielshaqed, line 14:
```suggestion
const messageSize = 200
```
Most people nowadays have terminals that are >100 chars wide.  Let's use that!

cmd/lakectl/cmd/annotate.go
arielshaqed: Please add specific test that were performed -- at the very least, we need to be able to reproduce the tests that you already did.  (Important because clearly it is easy to miss functionality when testing this code...).

arielshaqed (commented):
Thanks, neat!

Only thing missing for approval is tests performed / to perform in future.

cmd/lakectl/cmd/annotate.go
shimi9276: Done.

cmd/lakectl/cmd/annotate.go
shimi9276: That is more proper English indeed. Fixed.

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,84 @@
...
+const trailingThreeDots = "..."
+
+var annotateCmd = &cobra.Command{
+	Use:     "annotate <path uri>",
+	Short:   "List entries under a given path, annotating each with the latest modifying commit ",
```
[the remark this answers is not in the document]
shimi9276: Fixed.

cmd/lakectl/cmd/annotate.go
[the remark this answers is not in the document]
shimi9276: Good point. This inconsistency comes in part from the fact that Object receives, by definition, a path (to the object). On the other hand, I do agree that it is quite confusing in this specific code. To avoid this issue I passed the obj.Path as is.

shimi9276: This is an interesting approach. I would consider implementing it if we indeed see that the inefficiency is noticable. After all, this is happing on the client end.

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,82 @@
...
+						Object        string
+						CommitMessage string
+					}{
+						Commit:        res.JSON200.Results[0],
+						Object:        obj.Path,
+						CommitMessage: splitOnNewLine(stringTrimLen((res.JSON200.Results[0].Message), messageSize)),
+					}
+					Write(annotateTemplate, data)
+				} else {
+					fmt.Println("Error")
```
nopcoder, line 56: 1. We should not just print "Error" to the user - no commit information as part of the entry or something that they will understand we don't have commit log information for this record.
2. Skipping the record is also not an option as the entry is there and we don't want the user to think that an object is missing.

nopcoder (changes requested):

cmd/lakectl/cmd/annotate.go
shimi9276: Fixed

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,82 @@
...
+						Object        string
+						CommitMessage string
+					}{
+						Commit:        res.JSON200.Results[0],
+						Object:        obj.Path,
+						CommitMessage: splitOnNewLine(stringTrimLen((res.JSON200.Results[0].Message), messageSize)),
+					}
+					Write(annotateTemplate, data)
+				} else {
+					fmt.Println(obj.Path, "Could not retrieve commit info about this object")
```
nopcoder, line 56: 1. we retrieved the commits, but there was none.
2. I still think it should be part of the template. by passing no commit information. in case the template changes the way the path is printed, we will have to change both places.

nopcoder (changes requested):

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,86 @@
...
+				DieOnResponseError(res, err)
+				data := objectCommitData{
+					Object: obj.Path,
+				}
+				if len(res.JSON200.Results) > 0 {
+					data = objectCommitData{
+						Object:        data.Object,
+						Commit:        res.JSON200.Results[0],
+						CommitMessage: splitOnNewLine(stringTrimLen((res.JSON200.Results[0].Message), annotateMessageSize)),
+					}
```
nopcoder, line 59: You need to assign the value of each field

nopcoder (changes requested):

cmd/lakectl/cmd/annotate.go
shimi9276: This is done above, in the definition of the structure.

nopcoder (approved):
LGTM!

cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,83 @@
...
+				}
+				res, err := client.LogCommitsWithResponse(context, pathURI.Repository, pathURI.Ref, logCommitsParams)
+				DieOnResponseError(res, err)
+				data := objectCommitData{
+					Object: obj.Path,
+				}
+				if len(res.JSON200.Results) > 0 {
+					data.Commit = res.JSON200.Results[0]
+					data.CommitMessage = splitOnNewLine(stringTrimLen((res.JSON200.Results[0].Message), annotateMessageSize))
+				}
```
arielshaqed, line 57: When the LogCommits call fails, this prints the object path with no information what happened.  Can we do better?

arielshaqed (approved):
Thanks!  This will be useful for those of our users who do data archaeology.  As a code archaeologist, I approve -- you never know what you'll find when you start digging. :-)

Still concerned that we document _no tests_ that would have discovered the bug that existed in the code when blaming >1 objects that had not all been introduced in the same commit.  We *must* test this manually.  Then document (e.g. in the PR documentation) that we did so -- that way when we write automatic tests we will be able to include these tests.

Approving with this added documentation in the PR description.

Issue #2784, opened 2021-12-08
Refactor the merge operation


guy-har:
Our code for the merge operation got to a level where it's too complex and it is very hard to follow when inserting changes. 

## Current Implementation 

As part of our design
- `Apply`  is used for both commit and merge
- Merge iterator is implemented using the Diff iterator 

The iterators used by `apply` for applying a merge are:
![image](https://user-images.githubusercontent.com/60321938/145203891-959e4927-b9ac-4b19-b26f-8475a40fb40f.png)

- Apply receives two iterators 
    1. The destination iterator 
    2. The compare iterator (only the changes we need to apply on destination)
 - Compare iterator wrapping the merge iterator in order to show only changes as records and ranges
 - Merge iterator 
 - Base Iterator used by the merge iterator in order to perform a three-way diff
 - Diff Iterator shows differences between two metarange iterators 
 - Left (Destination) iterator
 - Right ( source ) iterator 
 
Also, the merge iterator uses the terms `source` and `destination` and the diff iterator uses the terms `left` and `right`

## Suggestion 

We should try to separate between:
- Apply and Merge
- Diff and Merge

Examine if we could simplify by implementing merge using three Iterators (source destination and base) and The metarange writer. 

This suggestion might raise complexity elsewhere, we should first investigate and see if it could be simplified. 

This suggestion is should contain a detailed design

PR #2829, opened 2021-12-23
merge refactor design

eden-ohana:
related issue - #2784


johnnyaug (commented):
The drawing is really great!
The text should have more details, and maybe some examples. But we can improve it while we do the actual refactor, maybe even combine them to a single PR. The way I see it, no rush to merge this.

design/merge-refactor.md
```diff
@@ -0,0 +1,26 @@
+# Design: Refactor the Merge Operation
+
+## Problem Description
+Currently, as part of our design:
+* Apply is used for both commit and merge
+* Merge iterator is implemented using the Diff iterator
+
+Our code for the merge operation got to a level where it's too complex and it is very hard to follow when inserting changes.
```
nopcoder, line 8: Think this is the point where we need to explain that apply/merge uses an iterator which communicates a stream of changes that the diff found.
Working with changes that we identify in a different level of the api leave small space to optimize in the space.

design/merge-refactor.md
```diff
@@ -0,0 +1,26 @@
+# Design: Refactor the Merge Operation
+
+## Problem Description
```
nopcoder, line 3: general note about formatting. can you keep a blank line after the header.
https://google.github.io/styleguide/docguide/style.html#add-spacing-to-headings

design/merge-refactor.md
```diff
@@ -0,0 +1,26 @@
...
+
+## Proposed Design
+![merge refactor design](diagrams/merge-refactor-design.png)
+
+
+### Separate between Diff and Merge:
+Currently, the merge operation uses the diff iterator and patches it with the merge iterator.
+Instead of using the diff iterator that holds source iterator and destination iterator, we can use a compare iterator that will get three iterators: source, destination and base. That way, we can optimize the code by comparing the 3 iterators together and skipping ranges by need.
+
+### Separate between Apply and Merge:
```
nopcoder, line 22: Do we need an Apply? from the description of this section it looks like both commit and merge needs a Writer. Make the actual implementation is written into apply as one that processes incoming Iterator. But it looks like both cases write a commit or metarange writer to serialize changes.

nopcoder (approved):

nopcoder:
Forgot to update as part of the approve that I've added comments.

design/merge-refactor.md
arielshaqed: Me too!  Some history: Originally Apply was _just_ for the commit operation.  Then we changed it to support merging because we thought that way we would have less code and it would be simpler.  But I think that the opposite occurred.

It is actually harder to write code for an iterator than it is to write code for a writer.  That's because iterator code has to remember where it was at every step, and it must also handle various distractions (e.g. entering a new range and discovering that it is empty).  In contrast, a writer is just a single loop (E.g. it can use just variables rather than instance variables, and it can use many consecutive loops in the code, and if it enters a new range and discovers that it is empty, it can just write out an empty range or skip that write...).

My ideal refactor would probably be:

* Separate "Apply" and "Merge" operations (maybe "Apply" :arrow_right: "Commit").
* Apply and Merge receive Iterator inputs and a metarange writer (in particular, there is *no* iterator over the commit or merge results).
   * Apply receives _two_ iterators: the original values and the diffs.
   * Merge receives _three_ iterators: from, to, and base.

eden-ohana:
Hi all,
Thanks for your review.
I think we all agree on the main guideline for the refactor. 
* Separate "Commit" and "Merge" operations (currently, apply is used for both commit and merge, doing compare range operations)
* Commit operation will get two iterators: source, changes
* Merge operation will get three iterators: base, source, destination (to allow optimization)
* The writing will be done or during the comparison or using a writer (depending on what will be simpler and will allow optimization).

I’m not merging it right now because the exact details will be known during implementation.

arielshaqed (approved):
This should be a vital improvement!

guy-har (approved):

PR #2832, opened 2021-12-25
Include Jupyter notebook in our everything bagel

### LINKED ISSUE #2758: Add Jupyter to the everything bagel
configured against the Spark container

nopcoder:
Close #2758

Adding another service to run Jupyter notebook on exposed port 8888.

Include additional packages for ease of use:
- findspark
- pyspark-kernel
- spylon-kernel

### DIFF
--- a/deployments/compose/docker-compose.yml
+++ b/deployments/compose/docker-compose.yml
@@ -210,3 +210,21 @@ services:
       - ./dbt/profiles.yml:/root/.dbt/profiles.yml
     entrypoint: dbt
 
+  notebook:
+    build: notebook
+    ports:
+      - 8888:8888
+    environment:
+      - SPARK_MODE=worker
+      - SPARK_MASTER_URL=spark://spark:7077
+      - SPARK_WORKER_MEMORY=1G
+      - SPARK_WORKER_CORES=1
+      - SPARK_RPC_AUTHENTICATION_ENABLED=no
+      - SPARK_RPC_ENCRYPTION_ENABLED=no
+      - SPARK_LOCAL_STORAGE_ENCRYPTION_ENABLED=no
+      - SPARK_SSL_ENABLED=no
+      - PYSPARK_DRIVER_PYTHON=jupyter
+      - PYSPARK_DRIVER_PYTHON_OPTS='notebook --no-browser --port=8888'
+    volumes:
+      - ./etc/hive-site.xml:/opt/bitnami/spark/conf/hive-site.xml
+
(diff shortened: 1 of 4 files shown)


deployments/compose/docker-compose.yml
```diff
@@ -210,3 +210,21 @@ services:
...
+      - 8888:8888
+    environment:
+      - SPARK_MODE=worker
+      - SPARK_MASTER_URL=spark://spark:7077
+      - SPARK_WORKER_MEMORY=1G
+      - SPARK_WORKER_CORES=1
+      - SPARK_RPC_AUTHENTICATION_ENABLED=no
+      - SPARK_RPC_ENCRYPTION_ENABLED=no
+      - SPARK_LOCAL_STORAGE_ENCRYPTION_ENABLED=no
+      - SPARK_SSL_ENABLED=no
```
guy-har, line 225: MIght be time to move it into a configuration file or an env file

deployments/compose/notebook/Dockerfile
```diff
@@ -0,0 +1,12 @@
...
+RUN useradd -ms /bin/bash notebook
+USER notebook
+WORKDIR /home/notebook
+ENV HOME /home/notebook
+ENV PATH ${PATH}:/home/notebook/.local/bin
+RUN pip install --user notebook findspark pyspark-kernel spylon-kernel
+COPY notebook-config.py /home/notebook/.jupyter/jupyter_notebook_config.py
+RUN mkdir /home/notebook/notebooks
+ENTRYPOINT []
+CMD ["jupyter", "notebook"]
```
guy-har, line 12: Why did you choose not to use an existing one from docker-hub Such as:
https://hub.docker.com/r/jupyter/all-spark-notebook
https://hub.docker.com/r/jupyter/pyspark-noteboo

guy-har (approved):
A great addition to the bagel!

deployments/compose/docker-compose.yml
nopcoder: True - I like it when it is very explicit and can update in one place. But I think it will be a good followup PR

deployments/compose/notebook/Dockerfile
nopcoder: At this point I wanted to use the same base image with the same spark that we currently have in out bagel.

nopcoder: If the out of the box experience will lake something - I agree we should move to another solution.

peacing:
I like this!

PR #2834, opened 2021-12-26
Pre check a branch exists before any work is done

### LINKED ISSUE #2509: Multipart upload - repo/branch existence not verified in initial request
To reproduce, use the AWS cli to copy a big file to a non-existent branch:

```
aws s3 cp --debug --endpoint-url http://my-lakefs.com bigfile.txt s3://example-repo/non-existent-branch
```

The initial create-multipart request doesn't fail, so many parts are sent to lakeFS.
Fix the initial create multi-part request to do this validation.

It should also verify that the branch is not [protected](./2369).

shimi9276:
Comes fixing a bug, when using AWS CLI to copy to a non-existing branch. Initially, Lakefs would begin uploading file parts to the server only to be rejected by the fact the branch does not exist. This creates unnecessary and unwanted work. The fix comes in the form of verifying the branch exists before any action or work is created and not only at the upload phase (which obviously is required in any case). 
It is worth noting that in any case, the AWS CLI does do some background work in every copy even if the copy fails. This is non Lakefs related as can be seen in their documentation. To additionally verify this, a cp to a non-existing AWS bucket was done and the results were the same (with no Lakefs involved).
Closes #2509


cmd/lakectl/cmd/annotate.go
```diff
@@ -0,0 +1,81 @@
+package cmd
```
itaiad200, line 1: Is this related to the change?

shimi9276:
Two merges have collided. Closing in order to open a new one.

PR #2835, opened 2021-12-26
Added branch exsitence check to createMultiPart

### LINKED ISSUE #2509: Multipart upload - repo/branch existence not verified in initial request

shimi9276:
Comes fixing a bug, when using AWS CLI to copy to a non-existing branch. Initially, Lakefs would begin uploading file parts to the server only to be rejected by the fact the branch does not exist. This creates unnecessary and unwanted work. The fix comes in the form of verifying the branch exists before any action or work is created and not only at the upload phase (which obviously is required in any case).
It is worth noting that in any case, the AWS CLI does do some background work in every copy even if the copy fails. This is non Lakefs related as can be seen in their documentation. To additionally verify this, a cp to a non-existing AWS bucket was done and the results were the same (with no Lakefs involved).
Closes #2509

### DIFF
--- a/pkg/gateway/operations/postobject.go
+++ b/pkg/gateway/operations/postobject.go
@@ -39,6 +39,17 @@ func (controller *PostObject) RequiredPermissions(_ *http.Request, repoID, _, pa
 
 func (controller *PostObject) HandleCreateMultipartUpload(w http.ResponseWriter, req *http.Request, o *PathOperation) {
 	o.Incr("create_mpu")
+	branchExists, err := o.Catalog.BranchExists(req.Context(), o.Repository.Name, o.Reference)
+	if err != nil {
+		o.Log(req).WithError(err).Error("could not check if branch exists")
+		_ = o.EncodeError(w, req, gatewayErrors.Codes.ToAPIErr(gatewayErrors.ErrInternalError))
+		return
+	}
+	if !branchExists {
+		o.Log(req).Debug("branch not found")
+		_ = o.EncodeError(w, req, gatewayErrors.Codes.ToAPIErr(gatewayErrors.ErrNoSuchBucket))
+		return
+	}
 	uuidBytes := [16]byte(uuid.New())
 	objName := hex.EncodeToString(uuidBytes[:])
 	storageClass := StorageClassFromHeader(req.Header)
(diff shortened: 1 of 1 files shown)


johnnyaug (approved):
LGTM

pkg/gateway/operations/postobject.go
```diff
@@ -149,6 +149,18 @@ func (controller *PostObject) Handle(w http.ResponseWriter, req *http.Request, o
 	// POST is only supported for CreateMultipartUpload/CompleteMultipartUpload
 	// https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateMultipartUpload.html
 	// https://docs.aws.amazon.com/AmazonS3/latest/API/API_CompleteMultipartUpload.html
+	branchExists, err := o.Catalog.BranchExists(req.Context(), o.Repository.Name, o.Reference)
```
nopcoder, line 152: 1. question- this is relevant only for create multipart?
2. do we like to push this code down one level to get the operation stat?

shimi9276: 1. That's correct. Indeed moved the branch check to the create multipart after getting the stats.

shimi9276: Good point. We indeed only need to double check before the create multipart phase and not in the rest of the handle function. Fixed this by moving the check to HandleCreateMultipartUpload function specifically.

pkg/gateway/operations/postobject.go
```diff
@@ -39,6 +39,17 @@ func (controller *PostObject) RequiredPermissions(_ *http.Request, repoID, _, pa
 
 func (controller *PostObject) HandleCreateMultipartUpload(w http.ResponseWriter, req *http.Request, o *PathOperation) {
 	o.Incr("create_mpu")
+	branchExists, err := o.Catalog.BranchExists(req.Context(), o.Repository.Name, o.Reference)
+	if err != nil {
+		o.Log(req).WithError(err).Error("could not check if branch exists")
```
arielshaqed, line 44: Not sure we should fail here. Why not just log and keep going? Either the user will succeed or they will fail - so bailing out here only increases the chance of failure when the "impossible" happens.

shimi9276: Letting the function continue running causes the server to begin creating and sending parts to be uploaded. This will fail in any case since the branch does not exist. We might as well fail at the first point to avoid unnecessary work.

arielshaqed: Sorry, I was unclear.  `BranchExists` will _fail_ and return an error when something went wrong.  It doesn't say the branch does or does not exist, only that it does not know.  If the failure was sporadic, we may still succeed.

It's not vital, just failing an informational test should IMO log-and-continue ("best effort").

shimi9276: If this is a sporadic error, I don't feel it's that bad if we exit with en error, at max the user will try again. I can fix it if you'd like, it's your call.

Issue #2231, opened 2021-07-13
Support an `exec` hook execution driver


ozkatz:
Currently, lakeFS supports only Webhooks. While easy to understand, they are not always easy in terms of operations:

* They require a server to be deployed and provide an uptime SLA
* They require setting up network communication between said server and lakeFS
* Depending on security protocols, they often require some sort of authetication
* They make very simple things more complex than needed (i.e. a one-liner in bash now needs to be wrapped in a web application)
* They make it hard to experiment locally, if lakeFS is running in Docker container due to Docker networking woes (a Docker container by default cannot access network ports of the hosting machine)

To make hooks easier to get started with, lakeFS should support a simpler type of hooks - that execute some behavior directly by the lakeFS server. 

This could be calling out to a command line directly, doing `docker run` on a provided image, or integrating with existing orchestration/scheduling tools that already exist in the organization such as Kubernetes (**discovery required**)

nopcoder:
@ozkatz can we close this one? 

https://github.com/treeverse/lakeFS/blob/0ca69c4f0f7550e8af4e513a305a6c8732ae5d3a/design/rejected/k8s-job-hook.md

ozkatz:
Yes. As we're rethinking how to execute hooks in a better way, I suggest closing this issue in the meantime as it is unlikely to be implemented in its current form.

PR #2836, opened 2021-12-26
design document for execute hook

nopcoder:
addressing https://github.com/treeverse/lakeFS/issues/2231

### DIFF
--- a/design/hook-execution.md
+++ b/design/hook-execution.md
@@ -5,7 +5,7 @@
 
 (from https://github.com/treeverse/lakeFS/issues/2231)
 
-Currently, lakeFS supports only WebHooks. While easy to understand, they are not always easy in terms of operations:
+Currently, lakeFS supports web and airflow hooks. While they are very easy to understand, they are not always easy in terms of operations:
 
  - They require a server to be deployed and provide an uptime SLA
  - They require setting up network communication between said server and lakeFS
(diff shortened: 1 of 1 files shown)


design/hook-execution.md
```diff
@@ -0,0 +1,68 @@
+# Hook Execution
```
itaiad200, line 1: This entire document assumes we have a list of different applications for hooks that users requested for,  and we prefer to let them do that themselves with webhooks or local execution. Is that the case (i.e. do we have numerous use-cases, or are we're trying to generalise from the start) ?

itaiad200 (changes requested):
I don't think this is needed :)

design/hook-execution.md
nopcoder: As I understand the use case is the entry level work needed to use the hooks for integration and by enabling a simple execution mechanism will reduce this entry barrier.

nopcoder:
> I don't think this is needed :)

the document for the discussion or the issue?

design/hook-execution.md
itaiad200: Not sure I understand. This more of a product question. I know of a single user which consider webhooks as a struggle. I also think we can manage without running custom code. Are there any other feedbacks supporting that?

nopcoder: Quoted the issue by product.
Will take the discussion with product to sharpen the requirements in order to have better understanding the use-case.

nopcoder:
The issue was updated, closing this one as it is no longer relevant.

Issue #2418, opened 2021-08-29
hadoop-lakefs integration tests to run over Databricks


nopcoder:
Currently, we run an integration test using docker on Spark cluster 2.x and 3.x.
Running the same on Databricks will help us catch more cases and issues before publishing a new package.

johnnyaug:
In our integration tests we start lakeFS in a docker-compose in GitHub actions.
If we use the same way here, we need to somehow give Databricks access to this installation.
I tried to solve this with an ngrok tunnel, but in the end found out that they have a limit of 1 online process at a time (2 in the Business plan) so it's not a suitable solution for us.

PR #2840, opened 2021-12-28
Nessie Improvements

johnnyaug:
As a prep for #2418:
* Extract bootstrapping lakeFS to a composite action
* Generate only the necessary code to save time in code generation


.github/workflows/nessie.yaml
```diff
@@ -445,47 +405,31 @@ jobs:
       TAG: ${{ needs.deploy-image.outputs.tag }}
       REPO: ${{ secrets.AWS_ACCOUNT_ID }}.dkr.ecr.us-east-1.amazonaws.com
     steps:
+      - name: Check-out code
```
arielshaqed, line 408: Why do we now need this step?  If it's for "Build lakectl for dbt container", why not put lakectl into the artifact?

arielshaqed (commented):
Neat!  Just a few short questions for my understanding.

Also, out of curiousity: is it measurabily faster?

johnnyaug:
@arielshaqed, the code generation is now faster - about 2 minutes instead of 4 minutes

.github/workflows/nessie.yaml
johnnyaug: This was added in all steps because we need access to the new composite action

.github/workflows/nessie.yaml
```diff
@@ -65,9 +65,10 @@ jobs:
 
       - name: Generate code
         run: |
-          make gen
+          make gen-ddl
+          make gen-api
+          make gen-ui
```
nopcoder, line 70:
```suggestion
          make -j3 gen-ddl gen-api gen-ui
```

nopcoder (approved):
Good reuse. The only thing that bother me in this change was the changes made for the lakectl_metastore files. Having a general format where you pass AWS_* variable is what I expect in the docker and s3 properties. Setting them to use NESSIE_* by default be cause it is run by nessie is a private case - would not want enforce NESSIE_* envvars on non nessie files.

johnnyaug:
@nopcoder changes environment variables in trino catalog to match the ones in the docker-compose in the same directory - so they are not nessie-specific. Does this make more sense?

nopcoder:
> @nopcoder, I changed the environment variables in trino catalog to match the ones in the docker-compose in the same directory - so they are not Nessie-specific. Does this make more sense?

Perfect! thanks.

nopcoder (approved):

arielshaqed (approved):
Thanks, looks great!

> @arielshaqed, the code generation is now faster - about 2 minutes instead of 4 minutes

Faster is betterer!

Issue #2703, opened 2021-11-14
lakeFS export function that doesn't depend on Apache Spark


ozkatz:
## What?

The functionality currently provided by the [lakeFS Export utility](https://docs.lakefs.io/reference/export.html) but without a dependency on Apache Spark. 

This could be a stand-alone command that doesn't support distribution across machines.

## Who?

lakeFS users who don't have Spark as part of their stack: data science and backend engineering teams - or data engineering teams in verticals where Apache Spark is less dominant.

## Why?

Adding Apache Spark for auxiliary uses has a very high cost involved: it has a steep learning curve, is relatively expensive to run for smaller workloads - and operationally complex. 

Smaller deployments and users in orgs where Spark is not already a first class citizen would like to avoid introducing it into their stacks, even at the cost of lower performance/parallelism for export.

PR #2842, opened 2021-12-29
lakeFS export functionality that doesn't depend on the spark client

lynnro314:
close: #2703 

This gives the functionality currently provided by the lakeFS Export utility but without a dependency on Apache Spark.
The solution is given with a docker container that wraps rclone with a python script.

Usage example:
```docker run -e LAKEFS_ACCESS_KEY=XXX -e LAKEFS_SECRET_KEY=YYY -e LAKEFS_ENDPOINT=https://<LAKEFS_ENDPOINT>/ -e S3_ACCESS_KEY=XXX -e S3_SECRET_KEY=YYY  -it lakefs-export repo-name s3://some-bucket/some_path/ --branch="branch-name"```
 
The full design document is [here](https://github.com/treeverse/lakeFS/blob/master/design/export-functionality.md).

### CODE REVIEW

File: deployments/tools/export/lakefs_export.py [hunk 1]
```diff
@@ -0,0 +1,113 @@
...
+    reference = ""
+    source = "lakefs:" + args.Repo + "/"
+    if args.branch != None and args.commit_id == None:
+    	source += args.branch + "/"
+    	reference = args.branch
+    elif args.branch == None and args.commit_id != None:
+    	source += args.commit_id + "/"
+    	reference = args.commit_id
+    else:
+    	raise Exception("Should assign branch or commit, but not both.")
```
talSofer, line 75:
```suggestion
    has_branch = args.branch != None
    has_commit = args.commit_id != None
    if has_branch and !has_commit:
    	source += args.branch + "/"
    	reference = args.branch
    elif !has_branch and has_commit:
    	source += args.commit_id + "/"
    	reference = args.commit_id
    else:
    	raise Exception("Should assign branch or commit, but not both.")
```

talSofer (commented):
Great and thorough work @lynnro314! this is a very valuable feature!

Added some comments and questions.

.github/workflows/nessie.yaml
```diff
@@ -395,6 +434,41 @@ jobs:
         working-directory: test/spark
         run: docker-compose logs --tail=1000 lakefs
 
+  export:
+    name: Test lakeFS export functionality
```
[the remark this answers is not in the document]
lynnro314: Done

deployments/tools/export/lakefs_export.py
```diff
@@ -0,0 +1,113 @@
...
+        f.write(res)
+    return
+
+
+def set_args():
+    parser = argparse.ArgumentParser(description='Process lakeFS export.')
+
+    parser.add_argument('Repo', metavar='repo', type=str, help='the name of lakeFS repository')
+    parser.add_argument('Dest', metavar='dest', type=str, help='path of destination')
+    parser.add_argument('--branch', metavar='branch', type=str, action='store', help='the relevant branch in the repository to export from')
```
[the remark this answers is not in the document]
lynnro314: You have to assign branch or commit_id, if you assign both or none of them - you get an error massage.

deployments/tools/export/lakefs_export.py
```diff
@@ -0,0 +1,113 @@
...
+    if args.branch != None and args.commit_id == None:
+    	source += args.branch + "/"
+    	reference = args.branch
+    elif args.branch == None and args.commit_id != None:
+    	source += args.commit_id + "/"
+    	reference = args.commit_id
+    else:
+    	raise Exception("Should assign branch or commit, but not both.")
+
+    if args.prev_commit_id == None:
```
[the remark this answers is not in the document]
lynnro314: Done

deployments/tools/export/lakefs_export.py
```diff
@@ -0,0 +1,113 @@
...
+    process = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+
+    # check export and create status file
+    if only_copy:
+        # if we check the copy command we add the flag --one-way since we only want to check that the source files were copied to the destination
+        check_process = subprocess.run(["rclone", "check", source, args.Dest, "--config=rclone.conf", "--combined=errors", "--one-way"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    else:
+        check_process = subprocess.run(["rclone", "check", source, args.Dest, "--config=rclone.conf", "--combined=errors"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+
+    # rclnoe check --combined contains files that are identical, so we'll remove those files from the errors file
```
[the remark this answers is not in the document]
lynnro314: This is the behavior of [rclone check](https://rclone.org/commands/rclone_check/). 
I updated the comment.

deployments/tools/export/lakefs_export.py
lynnro314: Done :)

docs/reference/export.md
```diff
@@ -153,3 +153,37 @@ you can alternatively export just the difference between `main` branch and the c
...
+
+
+1. Export all objects from a commit `c805e49bafb841a0875f49cd555b397340bbd9b8` on `example-repo` repository to s3 location `s3://destination-bucket/prefix/`:
+
+   ```shell
+   .... example-repo s3://destination-bucket/prefix/ --commit_id=c805e49bafb841a0875f49cd555b397340bbd9b8
+   ```
+
+
+1. Export only the diff between branch `example-branch` and commit `c805e49bafb841a0875f49cd555b397340bbd9b8`
```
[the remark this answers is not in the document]
lynnro314: I wanted to implement the feature in a similar way to the spark export.. There it is possible and it's not documented. 
So I think in this case I'll block it by code.

docs/reference/export.md
```diff
@@ -153,3 +153,37 @@ you can alternatively export just the difference between `main` branch and the c
    exporter.exportFrom(branch, commit)
    ```
 
+## Exporting Data With Docker
```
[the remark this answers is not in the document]
lynnro314: Done

docs/reference/export.md
```diff
@@ -153,3 +153,37 @@ you can alternatively export just the difference between `main` branch and the c
...
+
+In a similar way to the spark export - you can export data frome lakeFS to S3 using 3 export options:
```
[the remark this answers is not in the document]
lynnro314: Right! Done

test/rclone_export/run-test.sh
```diff
@@ -0,0 +1,55 @@
...
+
+docker-compose exec -T lakefs lakectl fs ls --recursive --no-color lakefs://${REPOSITORY}/main/ | awk '{print $8}' | sort > ${lakectl_out}
+
+aws s3 ls --recursive ${EXPORT_LOCATION} | awk '{print $4}'| cut -d/ -f ${n}-  | grep -v EXPORT_ | sort > ${s3_out}
+
+if ! diff ${lakectl_out} ${s3_out}; then
+  echo "export location and lakefs should contain same objects"
+  exit 1
+fi
+
```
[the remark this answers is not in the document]
lynnro314: I think it won't be necessary in this case, since lakeFS gets the url ref in the same way. 
(there is no test like that in the spark export)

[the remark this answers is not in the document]
talSofer: Thanks! What it will test is the export python script. how much time does it take for this test to complete?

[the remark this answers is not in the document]
lynnro314: I'll add it :)

guy-har (approved):
LGTM

talSofer (approved):

talSofer (commented):
What release process are we planning for this feature? 
Need to publish to docker and announce + consider a blogpost.

deployments/tools/export/lakefs_export.py
```diff
@@ -0,0 +1,117 @@
...
+        res = src.substitute(vars)
+    with open('rclone.conf', 'w+') as f:
+        f.write(res)
+    return
+
+
+def set_args():
+    parser = argparse.ArgumentParser(description='Process lakeFS export.')
+
+    parser.add_argument('Repo', metavar='repo', type=str, help='the name of lakeFS repository')
```
arielshaqed, line 31:
```suggestion
    parser.add_argument('Repo', metavar='repo', type=str, help='name of lakeFS repository')
```

Shorter is better in these cases, see e.g. most commands (I like to read `ls --help` for inspiration in documenting command-line flags).  This also occurs below.

deployments/tools/export/lakefs_export.py
```diff
@@ -0,0 +1,117 @@
...
+    parser.add_argument('Dest', metavar='dest', type=str, help='path of destination')
+    parser.add_argument('--branch', metavar='branch', type=str, action='store', help='the relevant branch in the repository to export from')
+    parser.add_argument('--commit_id', metavar='commit', type=str, action='store', help='the relevant commit on the repository to export from')
+    parser.add_argument('--prev_commit_id', metavar='previous-commit', type=str, action='store', help='a previous commit reference to export only the diff between it and the head of the branch')
```
arielshaqed, line 35: Arguments that change behaviour of the command are often easier to understand if phrased imperatively.  Perhaps more along these lines:
```suggestion
    parser.add_argument('--prev_commit_id', metavar='previous-commit', type=str, action='store', help='if specified, export only the difference between this commit ID and the head of the branch')
```

deployments/tools/export/lakefs_export.py
```diff
@@ -0,0 +1,117 @@
...
+
+def write_status_file(success, status_file_name):
+    if success:
+        with open(status_file_name,'w') as f:
+            f.write(SUCCESS_MSG)
+    else:
+        with open('errors','r') as errors_file, open(status_file_name,'w') as status_file:
+            for line in errors_file:
+                if line.startswith('-'):
+                    status_file.write("path was missing on the source: " + line)
```
arielshaqed, line 49:
```suggestion
                    status_file.write("path missing in source: " + line)
```

deployments/tools/export/lakefs_export.py
```diff
@@ -0,0 +1,117 @@
...
+                elif line.startswith('+'):
+                    status_file.write("path was missing on the destination: " + line)
+                elif line.startswith('*'):
+                    status_file.write("path was present in source and destination but different: " + line)
+                elif line.startswith('!'):
+                    status_file.write("there was an error reading or hashing the source or dest: " + line)
```
arielshaqed, line 55: AFAIU `write` never writes a newline.  All of these lines will get squished together.  (Of course if you tested and it works, ignore this comment...).

deployments/tools/export/lakefs_export.py
```diff
@@ -0,0 +1,117 @@
...
+    return
```
arielshaqed, line 56: If an unrecognized line appears, we tell the user nothing. All this function does is try to _add_ information to the lines.  So if it does not understand a line, it should not _delete_ it.  Just copy it unchanged or something.

deployments/tools/export/lakefs_export.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def main():
+
+    # create rclone configuration file
+    create_rclone_conf_file()
```
arielshaqed, line 62: Why do we need to create a file?  That can fail.  Why not just set [environment variables](https://rclone.org/docs/#environment-variables) for rclone?

deployments/tools/export/lakefs_export.py
```diff
@@ -0,0 +1,117 @@
...
+        source += args.branch + "/"
+        reference = args.branch
+    elif not has_branch and has_commit:
+        source += args.commit_id + "/"
+        reference = args.commit_id
+    else:
+        raise Exception("Should assign branch or commit, but not both.")
+
+    if has_commit and export_diff:
+        raise Exception("Could not export diff between two commits.")
```
arielshaqed, line 81:
```suggestion
        raise Exception("Cannot export diff between two commits.")
```
(I think)

arielshaqed, line 81: But why not, actually?  I can imagine it being useful.

deployments/tools/export/lakefs_export.py
```diff
@@ -0,0 +1,117 @@
...
+    return
+
+
+def set_args():
+    parser = argparse.ArgumentParser(description='Process lakeFS export.')
+
+    parser.add_argument('Repo', metavar='repo', type=str, help='the name of lakeFS repository')
+    parser.add_argument('Dest', metavar='dest', type=str, help='path of destination')
+    parser.add_argument('--branch', metavar='branch', type=str, action='store', help='the relevant branch in the repository to export from')
+    parser.add_argument('--commit_id', metavar='commit', type=str, action='store', help='the relevant commit on the repository to export from')
```
arielshaqed, line 34: I don't understand the difference between `commit_id` and `branch`.  They are both examples of what we call "refs".  And there are many very useful refs that are not commit IDs or branches, such at _tags_ (think "export version `1.1.4` of my data) and relative names (think "export `main^`, the previous version of `main`).

I would just unify two into a single arg, unless it makes a difference.  You can add examples to the help message to make it clear what you can pass in, e.g.
```
"ref to export from repository, e.g. `main', `v1.0.1`, `main~3`"
```

deployments/tools/export/lakefs_export.py
```diff
@@ -0,0 +1,117 @@
...
+    process = subprocess.run(cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+
+    # check export and create status file
+    if export_diff:
+        check_process = subprocess.run(["rclone", "check", source, args.Dest, "--config=rclone.conf", "--combined=errors"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    else:
+        # if we check the copy command we add the flag --one-way since we only want to check that the source files were copied to the destination
+        check_process = subprocess.run(["rclone", "check", source, args.Dest, "--config=rclone.conf", "--combined=errors", "--one-way"], stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+
+    # the command 'rclnoe check --combined' returns files that are identical, as well as the errors. So we'll remove the identical files from the errors file
```
arielshaqed, line 97: nit:

```suggestion
    # the command 'rclone check --combined' returns files that are identical, as well as errors. So we'll remove the identical files from the errors file
```

deployments/tools/export/lakefs_export.py
```diff
@@ -0,0 +1,117 @@
...
+    with open("errors", "r") as input, open("temp", "w") as output:
```
arielshaqed, line 98: This again writes a file and reads it.  That can (and will) fail in various ways.  E.g.:

* Program doesn't run if the current directory is not writable.
* If you the directory _is_ writable but already has a file `temp`, it deletes that file.
* If filesystem is close to full (common when running on very small instances, or inside a container), it may fail for no good reason.

Can RClone just write to a pipe, which we can then read and process?

docs/reference/export.md
```diff
@@ -153,3 +153,43 @@ you can alternatively export just the difference between `main` branch and the c
    exporter.exportFrom(branch, commit)
    ```
 
+## Using Docker
+
+This option is recommanded if you don't have spark at your tool-set.
```
arielshaqed, line 158: nits:
```suggestion
This option is recommended if you don't have Spark at your tool-set.
```
(and elsewhere in this dock)

docs/reference/export.md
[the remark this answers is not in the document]
arielshaqed: As above: I think this decreases usability.

test/rclone_export/run-test.sh
```diff
@@ -0,0 +1,81 @@
...
+echo "lakectl_out"
+cat ${lakectl_out}
+echo "s3_out"
+cat ${s3_out}
+
+if ! diff ${lakectl_out} ${s3_out}; then
+  echo "export location and lakefs should contain same objects"
+  exit 1
+fi
```
arielshaqed, line 81: Please open an issue for also testing error reporting.  Errors are usually the hardest thing :-/

arielshaqed (approved):
Very neat stuff!

Somewhat worried about the number of files which we use.

1. If a program writes files, it *must* offer flags to configure their names.  (These flags may have default values, of course). If the requirement specifies it is easier for users to write files, then do that -- but allow configuring the filename.  (See last item for a good alternative!)
1. Temporary files are *really* hard to get right.  They *must* be created with a unique name, they *must not* race against running the same program twice concurrently, they *must* be written to a writable directory with enough space in it, they *must not* remain after the program exits even if it failed, their location *must be* configurable, and many other limitations.

   Python uses the [tempfile](https://docs.python.org/3/library/tempfile.html) module to do this -- and you should use the highest-level facility offered there that does what you want.
1. Even better: _don't create files_.  Instead of temporary files, pipe the output back to your program and read it.  Instead of output files, write to standard output and/or standard error streams.

File work over 95% of the time.  The rest of the time, they create a mess that is easier for the programmer to deal with than for the user. :-(

deployments/tools/export/lakefs_export.py
lynnro314: Tested it again to be sure. The errors we get from rclone are with new line, so it won't be squished together. Thanks for letting me know!

deployments/tools/export/lakefs_export.py
lynnro314: As Tel mentioned in a previous comment, because it's not something given with the Spark export.

deployments/tools/export/lakefs_export.py
lynnro314: You'r very right. But this is how the Spark export works. Other refs are not supported.. I can open an issue for that and we'll add it to both.

test/rclone_export/run-test.sh
lynnro314: Good point, thanks! I'll open it.

deployments/tools/export/lakefs_export.py
lynnro314: I'll check and open an issue if necessary.

PR #2845, opened 2021-12-30
Removed '\n' so the output is purely identical to origin when calling Lakectl fs cat

### LINKED ISSUE #2821: lakectl fs cat print content to output with new-line
lakectl fs cat should stream the exact content of the file in order to get the exact content of the object without converting, formatting or adding new-line at the end.

Current implementation reads the complement content and prints it at the end as string with additional new-line.

shimi9276:
Fixing a bug that printed a new line to the output, when fs cat was called. This can cause inconsistencies when piping the output for instance into a file. 
This does **not** the issue that the file is fully copied to memory and not piece by piece to a set buffer of fixed length. A separate issue was opened for that. Issue #2846.
Closes #2821

### DIFF
--- a/cmd/lakectl/cmd/fs.go
+++ b/cmd/lakectl/cmd/fs.go
@@ -136,7 +136,7 @@ var fsCatCmd = &cobra.Command{
 			DieOnResponseError(resp, err)
 			contents = resp.Body
 		}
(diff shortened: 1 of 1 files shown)


nopcoder (approved):
Looks good. Please update the PR description to include 'lakectl' and '\n'

PR #2848, opened 2021-12-30
Clarify what is a valid source for lakectl ingest command

### LINKED ISSUE #2844: Docs: clarify what is a valid source for lakectl ingest
The description of the `lakectl ingest` command says 
```
lakectl ingest --help 
Ingest objects from an external source into a lakeFS branch (without actually copying them)
```
We need to document what an **external source** is and clarify that it means a location in the object store but not in the storage namespace.

talSofer:
Closes #2844


nopcoder (approved):
just update the bucket pattern

PR #2849, opened 2021-12-30
Document environment variable usage in lakectl

### LINKED ISSUE #2847: Docs: document the existence of $LAKECTL_BASE_URI
It's a very handy environment variable that isn't documented anywhere. For example, [this suggestion on the lakeFS Slack space](https://lakefs.slack.com/archives/C01678929QT/p1640820012014500) could be solved quite nicely using it:

```bash
$ export LAKECTL_BASE_URI="lakefs://my-repo/my-branch"
# Once set, use relative lakefs uri's:
$ lakectl ingest --from s3://bucket/prefix --to /path
$ lakectl fs ls /path
```

talSofer:
Fixes #2847

### DIFF
--- a/cmd/lakectl/cmd/docs.go
+++ b/cmd/lakectl/cmd/docs.go
@@ -48,6 +48,9 @@ This will setup a ` + "`" + `$HOME/.lakectl.yaml` + "`" + ` file with the creden
 When setting up a new installation and creating initial credentials (see [Quick start](../quickstart/index.md)), the UI
 will provide a link to download a preconfigured configuration file for you.
 
+` + "`" + `lakectl` + "`" + ` configuration items can each be controlled by an environment variable. The variable name will have a prefix of
+*LAKECTL_*, followed by the name of the configuration, replacing every '.' with a '_'. Example: ` + "`" + `LAKECTL_SERVER_ENDPOINT_URL` + "`" + ` 
+controls ` + "`" + `server.endpoint_url` + "`" + `.
 `
 
 func printOptions(buf *bytes.Buffer, cmd *cobra.Command) error {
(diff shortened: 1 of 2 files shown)


ozkatz (changes requested):
Not the right place to change :)
This file is auto generated by this: https://github.com/treeverse/lakeFS/blob/master/cmd/lakectl/cmd/docs.go

See `gen-docs` in the Makefile.

nopcoder (commented):
Need to update cmd/lakectl/cmd/docs.go - this doc is generated.

talSofer:
> Not the right place to change :) This file is auto generated by this: https://github.com/treeverse/lakeFS/blob/master/cmd/lakectl/cmd/docs.go
> 
> See `gen-docs` in the Makefile.

:) fixed this, thanks. @nopcoder @ozkatz Please let me know if there is a more elegant way to do add the note and usage example to `lakectl` root command. I didn't find a field of Cobra's command struct that looked like a good fit.

cmd/lakectl/cmd/docs.go
```diff
@@ -48,6 +48,9 @@ This will setup a ` + "`" + `$HOME/.lakectl.yaml` + "`" + ` file with the creden
 When setting up a new installation and creating initial credentials (see [Quick start](../quickstart/index.md)), the UI
 will provide a link to download a preconfigured configuration file for you.
 
+` + "`" + `lakectl` + "`" + ` configuration items can each be controlled by an environment variable. The variable name will have a prefix of
```
nopcoder, line 51: suggestion, it will have the same output
replace all: "`" + `lakectl` + "`"
with: "`lakectl`"

cmd/lakectl/cmd/docs.go
```diff
@@ -60,6 +63,13 @@ func printOptions(buf *bytes.Buffer, cmd *cobra.Command) error {
 	}
 	parentFlags := cmd.InheritedFlags()
 	parentFlags.SetOutput(buf)
+
+	name := cmd.CommandPath()
+	if name == "lakectl" {
```
nopcoder, line 68: Alternative solution, it can help more changes without looking for constant later.
```suggestion
	if cmd == rootCmd {
```

nopcoder (approved):
Two small suggestions, other than that looks good.

cmd/lakectl/cmd/docs.go
talSofer: Neat, thanks!

cmd/lakectl/cmd/docs.go
talSofer: done here and in this text block

ozkatz (approved):
Thanks!

PR #2851, opened 2022-01-01
Prevent Graveler from delete default branch

### LINKED ISSUE #2850: Delete default branch should be blocked
Currently it is possible to delete all branches from the repository, including the default branch which is used to create other branch.
In order not to keep a repository in unused condition we should block at least the delete of the repository default branch.
 
Example of how to reproduce

```sh
lakectl repo create lakefs://example s3://bucket
lakectl branch delete --yes main
```

nopcoder:
Fix #2850


talSofer (approved):
Looks good!

Issue #1657, opened 2021-03-21
lakectl command line testing


nopcoder:
Testing the functionality of lakectl using scripting will verify the client and the api functionality

arielshaqed:
Speaking as a disinterested observer, a truly great open-source tool that helps test CLI clients with a server is https://github.com/arielshaqed/specshell.  It just gives you a shell that your test can control.  At Binaris we used it to do CLI testing.  Think of it as like Expect, but easier to use (and lets you code in JavaScript not TCL).

nopcoder:
@itai-david can we close this one?

itai-david:
@nopcoder - Absolutely. We got this one done, and we got open issues for specific missing prats. Closing

PR #2852, opened 2022-01-02
Fixed Lakectl cat fs, not to copy entire file to memory, added as wel…

### LINKED ISSUE #2846: Lakectl fs cat writes the entire file to memory
When the cat command is performed, Lakectl retrieves the file and fully copies it to memory. This can potentially cause a memory bottleneck. A suggested solution would be to read from the file at a set size and then write to output, continuously doing so until the end of the file is reached.

shimi9276:
Fixed Lakectl fs cat, so when printing the output, the file is not fully copied to memory. The copy function was used which copies it in parts, thus not clogging the memory.
We should use issue #2864 to preform testing. 
Closes #2846


cmd/lakectl/cmd/fs.go
```diff
@@ -115,28 +115,31 @@ var fsCatCmd = &cobra.Command{
...
 				DieErr(err)
 			}
 			defer body.Close()
-			contents, err = io.ReadAll(body)
-			if err != nil {
-				DieErr(err)
-			}
 		} else {
-			resp, err := client.GetObjectWithResponse(cmd.Context(), pathURI.Repository, pathURI.Ref, &api.GetObjectParams{
+			client := getClientWithoutResponse()
```
arielshaqed, line 130: I suspect that the client fully downloads the file here (otherwise we'd probably have Close'd the Body). Best look at the generated code - note that this path is the common one for our users).

cmd/lakectl/cmd/root.go
```diff
@@ -1,178 +1,194 @@
...
-		viper.SetConfigName(".lakectl")
-	}
-
-	viper.SetEnvPrefix("LAKECTL")
-	viper.SetEnvKeyReplacer(strings.NewReplacer(".", "_")) // support nested config
-	viper.AutomaticEnv()                                   // read in environment variables that match
-
-	cfg = config.ReadConfig()
-}
+package cmd
```
arielshaqed, line 1: Why this change?

arielshaqed (changes requested):
Thanjavur! Very important feature, but since it's for large files it's important to verify that with the change no download to memory occurs. E.g., use ulimit to forbids process from taking up more than 500MiB and download a 1GiB file.

cmd/lakectl/cmd/fs.go
shimi9276: It does not download, it returns an HTTP response with a reader object. The file should indeed be closed. Fixed.

cmd/lakectl/cmd/root.go
```diff
@@ -85,7 +85,7 @@ lakectl is a CLI tool allowing exploration and manipulation of a lakeFS environm
 	Version: version.Version,
 }
 
-func getClient() api.ClientWithResponsesInterface {
+func getEnpointWithAuth() (string, *securityprovider.SecurityProviderBasicAuth) {
```
nopcoder, line 88: why reason to extract this one out?

cmd/lakectl/cmd/root.go
```diff
@@ -109,7 +109,11 @@ func getClient() api.ClientWithResponsesInterface {
 	if u.Path == "" || u.Path == "/" {
 		serverEndpoint = strings.TrimRight(serverEndpoint, "/") + api.BaseURL
 	}
+	return serverEndpoint, basicAuthProvider
+}
 
+func getClient() *api.ClientWithResponses {
```
nopcoder, line 115: We don't need another client func - we can extract the client with response from the client

cmd/lakectl/cmd/fs.go
```diff
@@ -115,28 +115,28 @@ var fsCatCmd = &cobra.Command{
 	Short: "Dump content of object to stdout",
 	Args:  cobra.ExactArgs(1),
 	Run: func(cmd *cobra.Command, args []string) {
-		client := getClient()
 		pathURI := MustParsePathURI("path", args[0])
 		direct, _ := cmd.Flags().GetBool("direct")
-		var contents []byte
+		var err error
+		var body io.ReadCloser
+		var resp *http.Response
```
nopcoder, line 122: can be moved into the GetObject scope

nopcoder (changes requested):

cmd/lakectl/cmd/fs.go
```diff
@@ -115,28 +115,28 @@ var fsCatCmd = &cobra.Command{
...
-			}
-			defer body.Close()
-			contents, err = io.ReadAll(body)
-			if err != nil {
-				DieErr(err)
-			}
+			_, body, err = helpers.ClientDownload(cmd.Context(), client, pathURI.Repository, pathURI.Ref, *pathURI.Path)
 		} else {
-			resp, err := client.GetObjectWithResponse(cmd.Context(), pathURI.Repository, pathURI.Ref, &api.GetObjectParams{
+			resp, err = client.GetObject(cmd.Context(), pathURI.Repository, pathURI.Ref, &api.GetObjectParams{
```
arielshaqed, line 127: My reading of the generated code for `GetObjectWithResponse` is that `GetObject` **does _not_** handle HTTP status codes in any way.  Assuming this is true, if I `GetObject` with a bad path (say one that does not exist) -- this code will not notice the 404 status code and claim that it succeeded.

cmd/lakectl/cmd/root.go
arielshaqed, line 88: nit:
```suggestion
func getEndpointWithAuth() (string, *securityprovider.SecurityProviderBasicAuth) {
```

arielshaqed (changes requested):
Thanks!

Still not sure about `GetObject`; AFAICS you will need to handle HTTP status codes yourself once you use it and not `GetObjectWithResponse`.

cmd/lakectl/cmd/fs.go
shimi9276: Good point. The workflow tests in Github did identify this as well. Fixed

cmd/lakectl/cmd/root.go
shimi9276: Was done initially in order to give the ability to create a client without response. Once realized creating a new client was unnecessary, it still seemed to be correct to be split into two separate functions. Can reset you think it's better.

shimi9276: Fixed.

cmd/lakectl/cmd/root.go
shimi9276: Correct. Fixed.

cmd/lakectl/cmd/fs.go
arielshaqed: Previously this was less important -- we just used the existing known-good method.  Now that we've seen it happen, can we add a test?  (That can happen in a future PR -- just open an issue, and do it after you merge this one.)

pkg/api/helpers/errors.go
```diff
@@ -108,25 +108,29 @@ func ResponseAsError(response interface{}) error {
...
-	if !f.IsValid() {
-		return fmt.Errorf("[no HTTPResponse]: %w", ErrRequestFailed)
-	}
-	httpResponse, ok := f.Interface().(*http.Response)
-	if !ok {
-		return fmt.Errorf("%w: no HTTPResponse", ErrRequestFailed)
+	var f reflect.Value
+	var httpResponse *http.Response
+	var ok bool
+	if reflect.TypeOf(response).String() != "*http.Response" {
```
arielshaqed, line 114: I would rather _not_ do this: now `ResponseAtError` can take `*http.Response` (from package `net/http`), or it can take some`*Response` type from code generated by the API.

The only reason for using reflection here is that there is no interface that generalizes the `*Response` types and lets us get the `*http.Response`; that's a bug in the API generator or a bug in Go not having generics.  But in order to handle passing a known `*http.Response`, you could simply add a new function `HTTPResponseAsError` -- it's this function starting at line 126 "` 	if httpResponse == nil || isOK(httpResponse.StatusCode) {`".  And the old `ResponseAsError` would simply end before that line with something like `return HTTPResponseAsError(httpResponse, err)`.

pkg/api/helpers/errors.go
```diff
@@ -108,25 +108,29 @@ func ResponseAsError(response interface{}) error {
...
+		if !f.IsValid() {
+			return fmt.Errorf("[no HTTPResponse]: %w", ErrRequestFailed)
+		}
+		httpResponse, ok = f.Interface().(*http.Response)
+		if !ok {
+			return fmt.Errorf("%w: no HTTPResponse", ErrRequestFailed)
+		}
+	} else {
+		httpResponse = response.(*http.Response)
 	}
```
arielshaqed, line 125: As an *alternative* to the above suggestion: even if you want to keep the same function for both cases, you don't need reflection here.  It's  just a checked cast, e.g.:

```suggestion
	if httpResponse, ok = response.(*http.Response); !ok {
		f = r.FieldByName("HTTPResponse")
		if !f.IsValid() {
			return fmt.Errorf("[no HTTPResponse]: %w", ErrRequestFailed)
		}
		httpResponse, ok = f.Interface().(*http.Response)
		if !ok {
			return fmt.Errorf("%w: no HTTPResponse", ErrRequestFailed)
		}
	}
```

arielshaqed (changes requested):
Thanks!

Hurrah!  Looks like our final round.  The modifications to `ResponseAsError` are awkward and make the function signature and its implementation even harder to understand than they are already.  (I'm allowed to say it's a horrible function because I wrote it.)  I made two suggestions; I prefer the first but either one will make the function a bit more clear.

cmd/lakectl/cmd/fs.go
```diff
@@ -115,28 +115,29 @@ var fsCatCmd = &cobra.Command{
...
 				Path: *pathURI.Path,
 			})
 			DieOnResponseError(resp, err)
-			contents = resp.Body
+			body = resp.Body
+		}
+		if err != nil {
+			DieErr(err)
+		}
```
nopcoder, line 135: You can move this one to the first 'if' block as this is no longer common check - the second part checks for error using 'DieOnResponseError'.

nopcoder (changes requested):

pkg/api/helpers/errors.go
```diff
@@ -132,12 +133,23 @@ func ResponseAsError(response interface{}) error {
 		statusText = http.StatusText(statusCode)
 	}
 	var message string
-	f = r.FieldByName("Body")
-	if f.IsValid() && f.Type().Kind() == reflect.Slice && f.Type().Elem().Kind() == reflect.Uint8 {
-		body := f.Bytes()
-		var apiError api.Error
-		if json.Unmarshal(body, &apiError) == nil && apiError.Message != "" {
-			message = apiError.Message
+	r := reflect.Indirect(reflect.ValueOf(httpResponse))
```
arielshaqed, line 136: Not sure I understand this code.  Now `httpResponse` has a known type `*http.Response`.  Why do we need to reflect on it?

arielshaqed (commented):
I don't understand how we're reporting errors from `http.Response`: it seems like massive overuse of reflection on a known type (and one with no `interface` that I can see.  Apologies if I'm jumping the gun and reacting in the middle of a sequence of commits!

pkg/api/helpers/errors.go
shimi9276: Yeah. I cannot run tests on my PC due to env issues, so the only way to test is by committing and pushing to github. Please disregard any commits until it's final.

pkg/api/helpers/errors.go
```diff
@@ -101,19 +102,23 @@ func (e UserVisibleAPIError) Unwrap() error {
...
 	}
-
+	var httpResponse *http.Response
+	var ok bool
 	f := r.FieldByName("HTTPResponse")
 	if !f.IsValid() {
 		return fmt.Errorf("[no HTTPResponse]: %w", ErrRequestFailed)
 	}
-	httpResponse, ok := f.Interface().(*http.Response)
+	httpResponse, ok = f.Interface().(*http.Response)
```
nopcoder, line 121: remove the above definition - we use them from here
```suggestion
	httpResponse, ok := f.Interface().(*http.Response)
```

nopcoder (approved):
Approved with a small comment about the code style

pkg/api/helpers/errors.go
shimi9276: Agree. Fixed.

cmd/lakectl/cmd/fs.go
shimi9276: Fixed

cmd/lakectl/cmd/fs.go
arielshaqed: @shimi9276 please note: PR is good.  The (only) thing blocking my approval is adding a test that _will_ identify this error.  Now that  #1657 has been merged, please add a test to CLI testing.

If you think it would take more than 1-2 hours, please instead open an issue and reply to this comment referring to that issue.

Thanks!

arielshaqed (commented):
Approving -- thanks for opening #2864!

arielshaqed (approved):
I had one job...
stamp LGTM!

PR #2854, opened 2022-01-03
List all generated files

arielshaqed:
Partially in order to see what files exist -- but also as a manual test.

### DIFF
--- a/test/spark/app/src/main/scala/Sonnets.scala
+++ b/test/spark/app/src/main/scala/Sonnets.scala
@@ -1,9 +1,30 @@
 import org.apache.spark.sql.{SaveMode, SparkSession}
+import org.apache.hadoop.fs.{Path, RemoteIterator}
 import org.apache.log4j.Logger
 
 import scala.collection.mutable.ListBuffer
+import scala.util.matching.Regex
 
 object Sonnets {
+  // Make RemoteIterator a Scala iterator.  Reformatted version of
+  // https://gist.github.com/timvw/4ec727de9b76d9afc51298d9e68c4241.
+  /**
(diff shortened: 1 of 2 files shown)


test/spark/app/src/main/scala/Sonnets.scala
```diff
@@ -41,6 +62,27 @@ object Sonnets {
         )
         target.save(outputPath)
 
+        {
```
nopcoder, line 65: question: any reason why it needs to be scoped?

test/spark/app/src/main/scala/Sonnets.scala
```diff
@@ -41,6 +62,27 @@ object Sonnets {
...
+           */
+          val pattern: Regex = s"${Regex.quote(outputPath)}/partition_key=[^/]*/.*/([^/]*\\.${Regex.quote(fmt)}|_SUCCESS)".r
+          val path = new Path(outputPath)
+          val fs = path.getFileSystem(sc.hadoopConfiguration)
+          val badFiles = convertToScalaIterator(fs.listFiles(path, true)).
+            map(file => s"${file.getPath.getParent}/${file.getPath.getName}").
+            filter(name => name match {
+              case pattern() => false
+              case _ => true
+            })
```
nopcoder, line 79: I think you can keep pattern as string and just use something like 'filter(name => name match pattern)

nopcoder (approved):
Approved - just question and a small suggestion

test/spark/app/src/main/scala/Sonnets.scala
arielshaqed: Just limiting the scope of the variables.

arielshaqed (commented):
Thanks!

Pulling -- I think the added test coverage is worth it.

PR #2856, opened 2022-01-04
bugfix - applyAll over changes iterator with tombstone record

eden-ohana:
bug:
apply operation
applyAll (over changes iterator because base is over) and we're in the middle of a range (iterValue != nil)
if the record is tombstone and we're not in Log trace
we will write the record instead of skipping it

Added a test to cover this case


guy-har (approved):
Nice catch!
Thanks

PR #2865, opened 2022-01-10
Fix diff pagination

guy-har:


eden-ohana:
Nice catch 🥇

nopcoder (approved):
LGTM

will require server side fix to remove the swagger restriction and cap on max server side support.

PR #2869, opened 2022-01-12
bagel switch to use jupyter/pyspark-notebook for notebook

nopcoder:


itaiad200 (approved):
Very elegant. Thank you!

PR #2871, opened 2022-01-13
Fix bagel for Windows users

nopcoder:
fix docker build to include entrypoint.sh as unix line format

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+*.sh text eol=lf
+*.bash text eol=lf


itaiad200 (commented):
Checking this and approving if it works :)

deployments/compose/docker-compose.yml
```diff
@@ -182,7 +182,12 @@ services:
 
   spark-thrift:
     image: docker.io/bitnami/spark:3
-    command: ["bash","-c", "/opt/bitnami/entrypoint.sh"]
+    # Windows bug (see https://github.com/treeverse/lakeFS/pull/2871)
+    # Need to copy the mounted file and replace the line endings for Windows hosts.blkio_config:
+    # For linux machines, this is a no-op
+    command: ["bash","-c", "cp /opt/bitnami/entrypoint.sh /opt/bitnami/entrypoint-copy.sh &&
```
nopcoder, line 188: prefer not to enable this hack - I assume that volume mount should work if the file format is kept

arielshaqed: Would it work if we created a new Dockerfile that started with docker.io/bitnami/spark:3 and edited that file, creating a new good Docker image?

arielshaqed (commented):
*WARNING*.  :warning:  Potentially stupid suggestion.

itaiad200 (approved):
It works! amazing work!

PR #2878, opened 2022-01-20
Add date flag to commit api

guy-har:


pkg/graveler/graveler.go
```diff
@@ -1198,6 +1200,10 @@ func (g *Graveler) Commit(ctx context.Context, repositoryID RepositoryID, branch
 
 		// fill commit information - use for pre-commit and after adding the commit information used by commit
 		commit = NewCommit()
+
+		if params.Date != nil {
+			commit.CreationDate = time.Unix(*params.Date, 0)
```
eden-ohana, line 1205: If date is not provided it will be set to 0?

clients/python/docs/CommitsApi.md
```diff
@@ -67,6 +67,7 @@ with lakefs_client.ApiClient(configuration) as api_client:
         metadata={
             "key": "key_example",
         },
+        date=1,
```
eden-ohana, line 70: Maybe give an example of a real date, 1640988000 (1/1/22)

eden-ohana (approved):
Code looks great !

pkg/graveler/graveler.go
guy-har: No, if the date is provided ( != nil) it will set the creation date. the second value in `time.Unix(*params.Date, 0)` is the nanoseconds

clients/python/docs/CommitsApi.md
guy-har: This code is generated, tried adding an example in the sagger file, didn't seem to affect the python docs.
leaving as is
Thanks

PR #2881, opened 2022-01-21
Disable hooks run configuration

### LINKED ISSUE #2879: Allow an option to disable the hooks entirely
In some environments hooks are not needed and misuse of configuring them can lead to arbitrary http requests. It's better to add a feature flag that can disable them.

itaiad200:
closes #2879 
Adding a configuration to disable the execution of hooks. The user can still see the results of old hooks that ran before, when the hooks weren't disabled. Since all API calls and UI views that involved hooks are only for fetching the data, they were not blocked. 
The prefix `_lakefs_actions/...` is still writeable but files uploaded to that path are treated similarly to any other path. A log line indicating that the hooks were skipped as written when appropriate.

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -29,12 +29,12 @@ This reference uses `.` to denote the nesting of values.
 * `logging.format` `(one of ["json", "text"] : "text")` - Format to output log message in
 * `logging.level` `(one of ["DEBUG", "INFO", "WARN", "ERROR", "NONE"] : "DEBUG")` - Logging level to output
 * `logging.output` `(string : "-")` - Path name to write logs to. `"-"` means Standard Output
+* `actions.enabled` `(bool : true)` - Setting this to false will block hooks from being executed
 * `database.connection_string` `(string : "postgres://localhost:5432/postgres?sslmode=disable")` - PostgreSQL connection string to use
 * `database.max_open_connections` `(int : 25)` - Maximum number of open connections to the database
 * `database.max_idle_connections` `(int : 25)` - Sets the maximum number of connections in the idle connection pool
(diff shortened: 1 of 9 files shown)


docs/reference/configuration.md
```diff
@@ -34,6 +34,7 @@ This reference uses `.` to denote the nesting of values.
 * `database.max_idle_connections` `(int : 25)` - Sets the maximum number of connections in the idle connection pool
 * `database.connection_max_lifetime` `(duration : 5m)` - Sets the maximum amount of time a connection may be reused
 * `listen_address` `(string : "0.0.0.0:8000")` - A `<host>:<port>` structured string representing the address to listen on
+* `actions_disabled` `(bool : false)` - Setting this to true will block hooks from being executed
```
arielshaqed, line 37: `actions.enabled` perhaps?

1. Positives are easier on my mind than double negatives
2. Hierarchy allows us to add more properties in future.
3. Matches `cache.enabled`, below.

pkg/actions/service.go
```diff
@@ -87,16 +88,17 @@ const defaultFetchSize = 1024
 
 var ErrNotFound = errors.New("not found")
 
-func NewService(ctx context.Context, db db.Database, source Source, writer OutputWriter, stats stats.Collector) *Service {
+func NewService(ctx context.Context, db db.Database, source Source, writer OutputWriter, stats stats.Collector, disabled bool) *Service {
```
arielshaqed, line 91: Consider a more descriptive name (and positive?) here, perhaps `executeActions`?

arielshaqed (approved):
Neat! Mere naming comments.

docs/reference/configuration.md
itaiad200: Done.

pkg/actions/service.go
itaiad200: Done.

PR #2885, opened 2022-01-25
Publish lakectl binaries to our public S3 bucket

johnnyaug:


.github/workflows/goreleaser.yaml
```diff
@@ -32,5 +32,7 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
-          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}  
+      - name: Make lakectl public
+        run: aws s3 cp --recursive --acl public-read "s3://treeverse-clients-us-east/lakectl/$(echo ${GITHUB_REF##*/} | cut -d. -f1-2)" "s3://treeverse-clients-us-east/lakectl/$(echo ${GITHUB_REF##*/} | cut -d. -f1-2)" --metadata-directive REPLACE
```
nopcoder, line 37: You can use the `tag_name` from the release event information - https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#release

.goreleaser.yml
```diff
@@ -54,3 +60,10 @@ changelog:
     exclude:
     - '^docs:'
     - '^test:'
+blobs:
+- provider: s3
+  region: us-east-1
+  bucket: treeverse-clients-us-east
+  folder: "lakectl/{{.Version}}"
+  ids:
+   - lakectl
```
nopcoder, line 69: This was not enough? if not (acl) do we need to keep it?

nopcoder (approved):

.github/workflows/goreleaser.yaml
nopcoder: keep it in case someone will add manual flow

PR #2901, opened 2022-02-02
docs: k8s job hook proposal

nopcoder:


design/k8s-job-hook.md
```diff
@@ -0,0 +1,176 @@
...
+## Non Goals
+
+- GitHub Actions - fully maintained, virtualized environment with pre-build images to execute pre/post hooks
+
+
+## Proposition
+
+### Overview
+
+New lakeFS hook type, triggered by pre/post commit/merge events to execute jobs on k8s cluster.
```
itaiad200, line 19: Please link to our docs page describing Actions & Hooks.

itaiad200: It should clear that understanding what we have today is a prerequisite to reading this documentation.

design/k8s-job-hook.md
```diff
@@ -0,0 +1,176 @@
...
+      command: ["python"]
+      args: ["bump-version.py"]
+      env:
+      - name: REPOSITORY
+        value: customers
+      - name: PROJECT
+        value: alpha
+```
+
+In this example we specified a post merge hook to execute a job. The job will use the image `next/lakefs-hooks:4` with the command python with the argument `bump-version.py`.
```
itaiad200, line 52: It should be super clear that:
1. `registry/lakefs-hooks:4` is defined and managed by the user. The name is misleading and may suggest we maintain it.
2. Same goes to `bump-version.py`.

design/k8s-job-hook.md
```diff
@@ -0,0 +1,176 @@
...
+Name - unique identifier to identify the specific job execution
+Image, command and args - set based on the hook information
+In case we would like to set a different job specification, we can specify in the lakeFS configuration (lakefs.yaml):
+
+```yaml
+hooks:
+  k8s-job:
+    spec_files:
+      default: /etc/lakefs/job_spec.yaml
+```
```
itaiad200, line 98: Is this a requirement? It's something we can add later on, but I think it shouldn't be part of the first version.

design/k8s-job-hook.md
```diff
@@ -0,0 +1,176 @@
...
+          resources:
+            limits:
+              cpu: "2"
+              memory: 4G
+            requests:
+              cpu: "1"
+              memory: 2G
+```
+
+We can mount the job spec using a configmap having lakeFS use it as the base to the job template used for all k8s-job hooks.
```
itaiad200, line 136: Can you please rephrase? I don't understand this line.

design/k8s-job-hook.md
```diff
@@ -0,0 +1,176 @@
...
+
+*Limit the end-user image use*
+
+Using the lakeFS configuration, we can specify a list of allowed images that the end-user and use in the k8s-job hook:
```
itaiad200, line 140: Again, not sure this is needed. But if you decide to keep this, let's start by explaining the problem: You don't want to allow lakeFS to run any image, therefore you can limit it with this configuration.

design/k8s-job-hook.md
```diff
@@ -0,0 +1,176 @@
...
+
+lakeFS requests job creation from the cluster.  In case the request fails it will log the job information and fail the hook execution.
+By default the hook waits for the job execution to complete, and based on the job status consider the execution to be successful.
+It will capture the job output into the step output.
+In case we would not like to block the operation (merge/commit) until the job execution is completed, for example on post-* events. We can change the default behavior by setting wait_for_complete to false:
+
+```yaml
+properties:
+  wait_for_complete: false
+```
```
itaiad200, line 163: `post-*` events are by default non-blocking. They can't block something that already happen. `Blocking` here simply means postponing the response to the client for something that already available to others.

design/k8s-job-hook.md
```diff
@@ -0,0 +1,176 @@
...
+
+Without waiting for the job to complete, lakeFS will consider job creation as successful execution. Will log the job information without waiting for the job to complete or capturing the output.
+
+### Considerations
```
itaiad200, line 167: Missing: The additional permissions lakeFS needs for managing jobs in k8s (consider both inside and outside the cluster, in terms of authentication as well)

itaiad200 (changes requested):

design/k8s-job-hook.md
nopcoder: 1. Update to 'myregistry/myhook:4' and removed any NEXT from the paragraph.
2. Added that his is a user-supplied image.
3. The bump version is related to next - but it is a valid use-case we can describe.

design/k8s-job-hook.md
nopcoder: updated

design/k8s-job-hook.md
nopcoder: We should enable a way to run different types of images.
If we do not enable to specify it in the action - we will need to support multiple specs.
Because  spec files are part of the lakeFS configuration, it means we will need to redeploy.

design/k8s-job-hook.md
```diff
@@ -0,0 +1,233 @@
+# Kubernetes Job Hook Proposal
+
+## Goals
+
+- Enable hooks that trigger execution of some behavior directly by the lakeFS server using Kubernetes
```
arielshaqed, line 5: Initially I read this as allowing K8s to trigger execution of some lakeFS behaviour.  I think this phrasing is indeed ambiguous, perhaps this will be better?
```suggestion
- Allow lakeFS actions directly to trigger actions running on Kubernetes
```

design/k8s-job-hook.md
arielshaqed: This is not a good way of handling user permissions.  E.g. it is very easy to break out of many images.

But I don't think we need to do this!  lakeFS can just specify what K8s role to use for the job.  The K8s admin can specify what that role is allowed to do.

design/k8s-job-hook.md
```diff
@@ -0,0 +1,233 @@
...
+### Considerations
+
+*Job lifetime*
+
+Once a job is created and executed in the cluster, the lakeFS server will not take ownership of the object. A mechanism should be in place to clean up all jobs lakeFS applied and completed (successfully or not).
+[Automatic Clean-up for Finished Jobs](https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/) capability is currently found on Kubernetes 1.23 (which we don’t have yet on AWS for example) which can help with that.
+
+*Job security*
+
+A user-supplied image can be run within the scope specified in the job spec, but lakeFS has no control over the job and the code itself, except for the image name that is limited by the lakeFS configuration.
```
arielshaqed, line 231: As above -- I would not regard whitelisting images a usable security mechanism in most environments.  Instead, we should hook into the existing K8s permissions structure, and just request a configurable role to run the job.

arielshaqed (commented):
Neat!

design/k8s-job-hook.md
nopcoder: added, thanks

design/k8s-job-hook.md
nopcoder: reduced the proposal - will not include blocking capability

design/k8s-job-hook.md
nopcoder: Rewrote this section, but I didn't removed the whitelist part yet because I didn't find any solution inside the cluster to limit the execution.

design/k8s-job-hook.md
```diff
@@ -150,27 +111,65 @@ hooks:
...
-  wait_for_complete: false
+### Authorizations
+
+Base on the above, lakeFS will require the following permissions:
+
+- `job` get, create and watch
+- `pod` get
+- `pod/log` get, list, watch
+
+The following describes possible ClusteRole that enables the above. We can limit the scope to a single namespace using Role.
```
itaiad200, line 129:
```suggestion
The following describes possible ClusterRole that enables the above. We can limit the scope to a single namespace using Role.
```

itaiad200, line 129: But you used `lakefs-hook` namespace above, so maybe use a role here to be aligned with the above example?

itaiad200 (approved):
Minor comments, overall looks great!

nopcoder:
Thanks @arielshaqed @itaiad200 @Jonathan-Rosenberg for the feedback, I've reduced the proposal functionality.

design/k8s-job-hook.md
```diff
@@ -0,0 +1,178 @@
+# Kubernetes Job Hook Proposal
+
+## Goals
+
+- Allow lakeFS actions directly to trigger actions running on Kubernetes
+- Enable non-blocking execution for post-merge and post-merge
```
Jonathan-Rosenberg, line 6:
```suggestion
- Enable non-blocking execution for post-commit and post-merge
```

design/k8s-job-hook.md
```diff
@@ -0,0 +1,178 @@
...
+          args: ["bump-version.py"]
+```
+
+In this example we specified a post merge hook to trigger a job creation in cluster using our sidecar.
+The job will use the image `myregistry/myhook:4` with the command `python` using the argument `bump-version.py`.
+The following environment variables will be populated by the sidecar based on the event information triggered by the web hook:
+
+```
+LAKEFS_HOOK_EVENTTYPE - Type of the event that triggered the action
+LAKEFS_HOOK_EVENTTIME - Time of the event that triggered
```
Jonathan-Rosenberg, line 66:
```suggestion
LAKEFS_HOOK_EVENTTIME - Time of the event that triggered the action
```

Jonathan-Rosenberg:
I see that there's no reference to the Jenkins (or possibly other integration tool) integration, meaning that the proposal is limited to a side-car that itself calling the K8s API server to run the job... 
Should we include it as an option?

design/k8s-job-hook.md
```diff
@@ -20,15 +20,21 @@
 
 Current hooks mechanism described [here](https://docs.lakefs.io/setup/hooks.html)
 
-New lakeFS hook type, triggered by pre/post commit/merge events to execute jobs on k8s cluster.
+Enabling additional side-car deployment will enable the current web-hook to trigger, using a side-car a job execution on a Kubernetes cluster.
 
-The user will provide a hook definition that will include the name of the image (docker image) and arguments that will be used for the hook.
-During commit/merge, lakeFS will schedule a job to run in the cluster. Successful response from the cluster for the job creation, will be considered successful and complete the commit/merge operation.
+The user will provide a web-hook definition that will include query parameters with the image and command to execute. The side-car use the posted information to execute a job based on the image and command line parameters supplied in the hook.
+During commit/merge, lakeFS will trigger web-hook that will post the request to the side-car, it will create a job to run in the cluster.
```
itaiad200, line 26:
```suggestion
During commit/merge, lakeFS will trigger web-hook that will post the request to the side-car, the side-car will create a job to run in the cluster.
```

design/k8s-job-hook.md
```diff
@@ -91,43 +93,41 @@ spec:
...
-      - rclone/rclone:1.57
-      - alpine
+allowed_images:
+  - rclone/rclone:1.57
+  - alpine
 ```
 
-Each item list will match the `registry/name:tag` used in the hook.
-If the tag is omitted, any tag will be allowed.
+Each item list will match the `registry/name:tag` used in the `image` name. If `tag` is omitted, any tag will be allowed.
```
itaiad200, line 109: Care to rephrase? Not clear to me..

itaiad200 (changes requested):
Thanks! Some clarifications needed

nopcoder:
@arielshaqed regarding the image allow list:
- it is not a security feature, you can try to look at it as a way to limit the end-user capability. Alternatively, as github let you select one of the virtual environment (https://github.com/actions/virtual-environments).
- As part of our current proposal, we will need to provide a predictable way to identify the job name. This will enable us to control the service account per job.
It will be a challenge to generate a unique name for a job within each action run, based on a service account that the cluster admin can predict.
I removed the job spec file to simplify - we can add a user-controlled job spec file/configuration, where we can support multiple objects yaml, where we will fill in placeholders for name, unique id, command, args, and image.

nopcoder:
Thanks for all the reviewers. Update the proposal with the decision: 
No changes to lakeFS itself required to enable the above, and we will keep it external. Stash the proposal under rejected.

itaiad200 (approved):

PR #2903, opened 2022-02-02
Update index.md

michalwosk:


johnnyaug (approved):

arielshaqed:
Late to the party, sorry.  I worry that we are removing the word "atomic" from most of what we do.  This is truly useful information for technically-oriented users.

PR #2927, opened 2022-02-08
Allow empty commit message with a specified flag

### LINKED ISSUE #1395: Disallow commit without a message
Should block at least from the CLI/UI a commit without a message

Jonathan-Rosenberg:
Added a `--allow-empty-message` flag that allows a commit to have an empty message, i.e.:
`lakectl commit lakefs://<repo>/<branch> -m "<whitespace character/s or nothing at all>" --allow-empty-message`
will be allowed, yet 
`lakectl commit lakefs://<repo>/<branch> -m "<whitespace character/s or nothing at all>"`
will not be allowed.

Closes #1395


cmd/lakectl/cmd/commit.go
```diff
@@ -77,8 +88,9 @@ func getKV(cmd *cobra.Command, name string) (map[string]string, error) {
 func init() {
 	rootCmd.AddCommand(commitCmd)
 
-	commitCmd.Flags().StringP("message", "m", "", "commit message")
-	_ = commitCmd.MarkFlagRequired("message")
+	commitCmd.Flags().BoolVar(&allowEmptyCommitMessageFlag, AllowEmptyMessageFlagName, false, "allow an empty commit message")
+	commitCmd.Flags().StringP(MessageFlagName, "m", "", "commit message")
```
johnnyaug, line 92: What is the justification for marking this flag as required? In Git, omitting this flag starts an "interactive" commit, which open your text editor for a message. This would force you to add something like `-m ''` to create an empty commit message. In our case, maybe it's not necessary: maybe using the new `allow-empty-message` flag is enough.

cmd/lakectl/cmd/commit.go
```diff
@@ -9,30 +9,41 @@ import (
 	"github.com/treeverse/lakefs/pkg/uri"
 )
 
-var commitCreateTemplate = `Commit for branch "{{.Branch.Ref}}" completed.
+var (
+	errInvalidKeyValueFormat    = fmt.Errorf("invalid key/value pair - should be separated by \"=\"")
```
johnnyaug, line 13: nit: separate errors to another var block.

cmd/lakectl/cmd/commit.go
```diff
@@ -77,8 +88,9 @@ func getKV(cmd *cobra.Command, name string) (map[string]string, error) {
 func init() {
 	rootCmd.AddCommand(commitCmd)
 
-	commitCmd.Flags().StringP("message", "m", "", "commit message")
-	_ = commitCmd.MarkFlagRequired("message")
+	commitCmd.Flags().BoolVar(&allowEmptyCommitMessageFlag, AllowEmptyMessageFlagName, false, "allow an empty commit message")
```
johnnyaug, line 91: We usually don't bind the flags to variables, but rather we just define them here in the init method, and later consume them within the command using getters.

cmd/lakectl/cmd/commit.go
```diff
@@ -9,30 +9,41 @@ import (
...
-		kvPairs, err := getKV(cmd, "meta")
+		kvPairs, err := getKV(cmd, MetaFlagName)
 		if err != nil {
 			DieErr(err)
 		}
-		message, err := cmd.Flags().GetString("message")
+		message, err := cmd.Flags().GetString(MessageFlagName)
 		if err != nil {
 			DieErr(err)
+		} else if strings.TrimSpace(message) == "" && !allowEmptyCommitMessageFlag {
```
johnnyaug, line 45: don't need this else

docs/assets/js/swagger.yml
```diff
@@ -425,6 +425,10 @@ components:
           type: object
           additionalProperties:
             type: string
+        date:
```
johnnyaug, line 428: [belong in another PR]

johnnyaug (changes requested):
Very nice! Really like the tests.
Left minor comments

cmd/lakectl/cmd/commit.go
Jonathan-Rosenberg: Is that for a reason?

johnnyaug: Not sure of the reason, but I think we should do it only one way or the other (unless there's a specific reason to bind the variable here)

itai-david:
Commenting here as I couldn't find how to comment on file rename - why was nessie/files/ro_1k renamed to nessie/files/ro_1k.txt?

nessie/golden/lakectl_fs_second_upload.golden
```diff
@@ -0,0 +1,7 @@
+Path: upload_file.txt.1
+Modified Time: <DATE> <TIME> <TZ>
+Size: 20 bytes
+Human Size: 20 B
+Physical Address: ${STORAGE}/<OBJECT_KEY>
+Checksum: bcd52aa37b0c6151513579d6f21c4414
+Content-Type: application/octet-stream
```
itai-david, line 7: Looks like you were forced to create another upload output golden file, due to lakectl tests limitation
Can you please open an issue on lakectl-testing so we can parametrized the path, size and checksum, and reuse the golden file?
Thanks

nessie/lakectl_test.go
```diff
@@ -104,9 +104,12 @@ func TestCommit(t *testing.T) {
...
 	RunCmdAndVerifySuccessWithFile(t, Lakectl()+" log lakefs://"+repoName+"/"+mainBranch, false, "lakectl_log_initial", vars)
-	RunCmdAndVerifySuccessWithFile(t, Lakectl()+" fs upload -s files/ro_1k lakefs://"+repoName+"/"+mainBranch+"/ro_1k.1", false, "lakectl_fs_upload", vars)
+	RunCmdAndVerifySuccessWithFile(t, Lakectl()+" fs upload -s files/ro_1k.txt lakefs://"+repoName+"/"+mainBranch+"/ro_1k.txt.1", false, "lakectl_fs_upload", vars)
 	RunCmdAndVerifySuccessWithFile(t, Lakectl()+" log lakefs://"+repoName+"/"+mainBranch, false, "lakectl_log_initial", vars)
 	RunCmdAndVerifyFailureWithFile(t, Lakectl()+" commit lakefs://"+repoName+"/"+mainBranch, false, "lakectl_commit_no_msg", vars)
+	RunCmdAndVerifyFailureWithFile(t, Lakectl()+" commit lakefs://"+repoName+"/"+mainBranch+" -m \" \"", false, "lakectl_commit_empty_msg_err", vars)
+	RunCmdAndVerifySuccessWithFile(t, Lakectl()+" commit lakefs://"+repoName+"/"+mainBranch+" --allow-empty-message -m \" \"", false, "lakectl_commit_with_empty_msg_flag", vars)
+	RunCmdAndVerifySuccessWithFile(t, Lakectl()+" fs upload -s files/upload_file.txt lakefs://"+repoName+"/"+mainBranch+"/upload_file.txt.1", false, "lakectl_fs_second_upload", vars)
```
itai-david, line 112: I tyhink you can use the same source file and just (ro_1k) and just upload it to a different destination. That way you don't have to create a new file for uploading (the upload_file.txt)

cmd/lakectl/cmd/commit.go
nopcoder: also, you can use back quote (`) so you don't need to escape \" (suggestion)

cmd/lakectl/cmd/commit.go
```diff
@@ -9,31 +9,43 @@ import (
 	"github.com/treeverse/lakefs/pkg/uri"
 )
 
-var commitCreateTemplate = `Commit for branch "{{.Branch.Ref}}" completed.
+var (
+	errInvalidKeyValueFormat    = fmt.Errorf("invalid key/value pair - should be separated by \"=\"")
+	errEmptyMessage             = fmt.Errorf("commit with an empty message without specifying the \"--allow-empty-message\" flag")
+	allowEmptyCommitMessageFlag bool
```
nopcoder, line 15: Don't use global variable for flag value

cmd/lakectl/cmd/commit.go
```diff
@@ -9,31 +9,43 @@ import (
...
 Parents: {{.Commit.Parents|join ", "}}
 
 `
+)
 
-var errInvalidKeyValueFormat = fmt.Errorf("invalid key/value pair - should be separated by \"=\"")
+const (
+	MessageFlagName           string = "message"
+	AllowEmptyMessageFlagName string = "allow-empty-message"
+	MetaFlagName              string = "meta"
```
nopcoder, line 29: `string` is redundant - prefer to use without so it can be used for all `string` kind as it takes the type when it is used.

cmd/lakectl/cmd/commit.go
```diff
@@ -9,31 +9,43 @@ import (
...
+		kvPairs, err := getKV(cmd, MetaFlagName)
 		if err != nil {
 			DieErr(err)
 		}
-		message, err := cmd.Flags().GetString("message")
+		message, err := cmd.Flags().GetString(MessageFlagName)
 		if err != nil {
 			DieErr(err)
 		}
+		if strings.TrimSpace(message) == "" && !allowEmptyCommitMessageFlag {
```
nopcoder, line 46: prefer to check the boolean before you do any work - just if eval order and work

cmd/lakectl/cmd/commit.go
```diff
@@ -9,31 +9,43 @@ import (
 	"github.com/treeverse/lakefs/pkg/uri"
 )
 
-var commitCreateTemplate = `Commit for branch "{{.Branch.Ref}}" completed.
+var (
+	errInvalidKeyValueFormat    = fmt.Errorf("invalid key/value pair - should be separated by \"=\"")
+	errEmptyMessage             = fmt.Errorf("commit with an empty message without specifying the \"--allow-empty-message\" flag")
```
nopcoder, line 14: Don't think we need an `error` in this case.
This is an error message to the user that we can just DieFmt.
As I see it - there is a difference between an error generated by a function - like "value is require", "empty commit message is not allowed" vs the error message to the user like you specified that explains the user how to use it with the right flag.

nopcoder (changes requested):

nessie/lakectl_test.go
Jonathan-Rosenberg: Absolutely!
Thanks

cmd/lakectl/cmd/commit.go
Jonathan-Rosenberg: Not a new flag but removed it anyway.
Handling the requirements in lines 41-46

johnnyaug (approved):
Nice!

cmd/lakectl/cmd/commit.go
```diff
@@ -1,39 +1,51 @@
...
 
 	"github.com/spf13/cobra"
 	"github.com/treeverse/lakefs/pkg/api"
 	"github.com/treeverse/lakefs/pkg/uri"
 )
 
-var commitCreateTemplate = `Commit for branch "{{.Branch.Ref}}" completed.
+var (
+	errInvalidKeyValueFormat = errors.New(`invalid key/value pair - should be separated by "="`)
+	fmtErrEmptyMessage       = `commit with no message without specifying the "--allow-empty-message" flag`
```
nopcoder, line 14: const

nopcoder (approved):
LGTM, after attending the rest of the reviewers question.

nessie/golden/lakectl_fs_second_upload.golden
Jonathan-Rosenberg: Sure

Issue #2931, opened 2022-02-08
lakectl testing - no windows support


itai-david:
As lakectl can be built and used on windows systems, need to support its testing on windows too

itaiad200:
I think a good first step is to test only lakectl on Windows. Users can spin up a container almost everywhere. Running the lakectl binary locally on linux if you have a Windows machine is much harder..

N-o-Z:
We're stuck with this task - as running a mix configuration of windows and linux containers requires enabling the docker experimental mode. 
Posted a question on Github community to see if someone can provide an answer.
Moving back to planned column and removing myself -
Created a draft [PR](https://github.com/treeverse/lakeFS/pull/6481) to test the pipeline

PR #2932, opened 2022-02-08
lakectl: convert windows paths to S3 style paths on upload

itai-david:
lakectl converts Windows style paths (i.e. backslash delimiter) Unix style paths (forward slash) to meet S3 behavior

Tested manually by running lakectl on windows emulator and verify the files are stored and displayed the same way as on Unix.

Opened #2931 to follow up with automated tests

### DIFF
--- a/cmd/lakectl/cmd/fs.go
+++ b/cmd/lakectl/cmd/fs.go
@@ -150,14 +150,14 @@ func upload(ctx context.Context, client api.ClientWithResponsesInterface, source
 	defer func() {
 		_ = fp.Close()
 	}()
-	filePath := api.StringValue(destURI.Path)
+	objectPath := api.StringValue(destURI.Path)
 	if direct {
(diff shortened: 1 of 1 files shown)


cmd/lakectl/cmd/fs.go
```diff
@@ -166,7 +165,7 @@ func uploadObject(ctx context.Context, client api.ClientWithResponsesInterface,
 		defer func() {
 			_ = pw.Close()
 		}()
-		filename := path.Base(filePath)
+		filename := filepath.Base(filePath)
```
talSofer, line 168: nit; It is confusing that both the library and variable names called filePath. maybe rename the variable to path?

talSofer (approved):
Looks good! thanks for opening a follow-up task for automated tests.

cmd/lakectl/cmd/fs.go
itai-david: @talSofer - I checked and the filepath here is actually the destination path, so decided to go with objectPath after all

PR #2934, opened 2022-02-09
feat: s3 add disable ssl conf

rongfengliang:
s3 add disable SSL conf (for better support minio without SSL conf)


johnnyaug (commented):
Can you please add this to the docs?

PR #2941, opened 2022-02-11
docs: Header button direct to join beta

JamilAhmad:
Get started button in the header menu renamed to "lakeFS Cloud" and linked to https://lakefs.io/join-beta/ as per Michal Wosk instruction
Please review and approve

### DIFF
--- a/docs/_config.yml
+++ b/docs/_config.yml
@@ -25,8 +25,8 @@ aux_links:
     - 'https://lakefs.io/blog/'
 
 buttons:
-  'Get Started':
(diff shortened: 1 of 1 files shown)


nopcoder (changes requested):
Something doesn't looks right to me:
lakefs.io is got "lakeFS Cloud" as a label and "Get Started" button to the docs, and here we don't have the label and we change the button to get to the beta which holds lakefs.io style.
I think both should be aligned. 
cc @michalwosk

nopcoder (approved):

Issue #2625, opened 2021-10-28
Spark: co-existing with existing underlying object store


ozkatz:

johnnyaug:
@ozkatz, can you add some information to this issue?

talSofer:
@ozkatz can you please add more info to this issue? thanks!

ozkatz:
@talSofer @johnnyaug description updated with more information.

talSofer:
>Since this is probably useful for non-lakeFS users as well, this should be a standalone component with no direct dependency on anything other than Spark itself

@ozkatz is this a nice-to-have requirement? I mean, it will be great if we can find such a generic solution that will be useful for non-lakeFS users, but I think it may limit us if we treat it as a must-do.

ozkatz:
@talSofer it is indeed nice-to-have! Edited the original comment to reflect that.

talSofer:
We've decided to split this delivery into multiple steps, and the following is the scope of the first step: 
- RouterFS to work with lakeFS only via S3Gateway
- RouterFS' Jar is not published to maven central. Users can build it according to provided build instructions
- The mapping configuration format is validated, and the intent is to keep it. 
- Usage it documented on [github](https://github.com/treeverse/hadoop-router-fs)

PR #2942, opened 2022-02-12
Proposal: using RouterFileSytem to co-exist with existing underlying …

talSofer:
A proposed solution for #2625


design/spark-co-exist-with-object-storages.md
```diff
@@ -0,0 +1,141 @@
...
+
+Considering the example above, the relevant object store and file system are S3 and `S3AFileSystem` compatibly. `RouterFileSystem` will use `S3AFileSystem` to perform file system operations against S3. e.g. to open `s3://example-repo/dev/foo.parquet`
+
+#### Getting the relevant File System 
+
+Given that `RouterFileSystem` is configured as the file system for URIs with a certain scheme (e.g. `scheme=s3`),
+we need to provide a way to get the relevant file system that can perform actual file system operations against the 
+relevant object store, e.g. `S3AFileSystem`. 
+To do that, we would follow a two-step process: 
+1. add the following Spark configuration:
```
nopcoder, line 69: Note we will need to use the per bucket configuration or verify that it will be ok to initialize the both filesystems with tsam s3a configurations.
In case the underlying and s3 (for lakefs) will share the same implementation, so they will access the same `fs.s3a.*` configurations.

design/spark-co-exist-with-object-storages.md
```diff
@@ -0,0 +1,141 @@
...
+#### Example: Read an object managed by lakeFS
+
+![RouterFS with lakeFS URI](diagrams/routerFS_lakefs_path.png)
+
+### How RouterFileSystem works with URIs that are not mapped to lakeFS objects   
+
+`RouterFileSystem` will check the mapping configurations, and won't change paths that are not mapped to lakeFS objects. 
+e.g. given the URI `s3://bucket-b/bar.parquet` and the following mapping configurations:
+```yaml
+MapperConfig = {
```
nopcoder, line 101: 1. consider using properties - trying to act like the ecosystem
2. can we simplify it to source prefix and target prefix? the suggestion forces a full bucket as the source.

design/spark-co-exist-with-object-storages.md
```diff
@@ -0,0 +1,141 @@
...
+
+### Handle any interaction with the underlying object store
+
+To allow Spark users to integrate with lakeFS without changing their Spark code, `RouterFileSystem` is configured to be 
+the file system for object store URIs with a certain scheme. For example, the following Spark configurations make 
+`RouterFileSystem` the file system for URIs with `scheme=s3`:
+```shell
+fs.s3.impl=RouterFileSystem
+```
+This will force any file system operation performed on an ibject URI with `scheme=s3` to go through `RouterFileSystem`.
```
Jonathan-Rosenberg, line 39:
```suggestion
This will force any file system operation performed on an object URI with `scheme=s3` to go through `RouterFileSystem`.   
```

design/spark-co-exist-with-object-storages.md
```diff
@@ -0,0 +1,141 @@
...
+This will force any file system operation performed on an ibject URI with `scheme=s3` to go through `RouterFileSystem`.   
+
+### URI translation
+
+`RouterFileSystem` has access to a configurable mapping that maps bucket names into lakeFS repositories and prefixes, e.g.
+```yaml
+MapperConfig = {
+    {
+        "source_bucket_pattern": "bucket-a",
+        "mapped_bucket_name": "example-repo",
```
idanovo, line 48: maybe we should consider more "git-like" names? (e.g. "mapped-repository" and "mapped branch")

design/spark-co-exist-with-object-storages.md
```diff
@@ -0,0 +1,141 @@
...
+```
+This will force any file system operation performed on an ibject URI with `scheme=s3` to go through `RouterFileSystem`.   
+
+### URI translation
+
+`RouterFileSystem` has access to a configurable mapping that maps bucket names into lakeFS repositories and prefixes, e.g.
+```yaml
+MapperConfig = {
+    {
+        "source_bucket_pattern": "bucket-a",
```
idanovo, line 47: we won't support prefix routing for source? ("bucket/path/*"), only bucket?

design/spark-co-exist-with-object-storages.md
```diff
@@ -0,0 +1,141 @@
...
+#### Getting the relevant File System 
+
+Given that `RouterFileSystem` is configured as the file system for URIs with a certain scheme (e.g. `scheme=s3`),
+we need to provide a way to get the relevant file system that can perform actual file system operations against the 
+relevant object store, e.g. `S3AFileSystem`. 
+To do that, we would follow a two-step process: 
+1. add the following Spark configuration:
+```shell
+# fs.underlying.impl=relevant filesystem
+fs.underlying.impl=S3AFileSystem
```
johnnyaug, line 72: I have a problem with the concept of "underlying". The way I see it, the mapping config should also include the filesystem to direct to, and can direct to multiple filesystems, not just the one. This will also allow us not to rely on per-bucket configuration which is not always supported and always is a pain.

johnnyaug:
Very interesting and clear! Thank you.
I agree that implementing and maintaining a Spark library is hard and I would even say we should avoid this until we really can't.

If we decide to go with a Spark library, this sounds like a very good direction to go.

arielshaqed (commented):
As a developer, I believe this is the only way forward with these requirements.

As a user I believe that the proposed configuration details require me to learn too many things, and will not fit into my existing Spark configurations.

1. The configuration should be more Spark-like. Both for convenience, and for maintainability as part of a larger managed configuration with overrides. 
2. I have no mental model of the replacement engine. I will need to learn and understand its capabilities (which also means that it will be both hard and important to document properly!).
3. If any feature is missing, I cannot use the router fs until someone codes them for me. This raises my barrier to adoption. 
4. Some features are already known to be missing. E.g., no way to change schema (filesystem). So I expect this not to work with older versions of s3a,  that are missing per-bucket configuration. (And it will be a difficult configuration.) 

I would *really* like to see this configuration done with regexps,  or at least with a string prefix-replacement syntax (and,  say,  a longest-match-wins policy for specificity).

design/spark-co-exist-with-object-storages.md
talSofer: 1. done
2. let me know what you think now

ozkatz (commented):
Thanks for this @talSofer ! This seems to be the right approach.
There's some prior art that we can look at, as it appears we're not the first project to think about providing a wrapper for a Hadoop Filesystem. 

Hadoop itself provides [FilterFilesystem](https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/fs/FilterFileSystem.html) which seems to do something similar (`"A FilterFileSystem contains some other file system, which it uses as its basic file system, possibly transforming the data along the way or providing additional functionality. "`).

There's even an example of how to override it: in this case, to write a checksum for every file that's written; see [ChecksumFileSystem](https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/fs/ChecksumFileSystem.html).

Not really a review (yet) but I think this could be something to research when thinking about implementing this (or a similar) design.

design/spark-co-exist-with-object-storages.md
```diff
@@ -20,122 +18,178 @@ Make the solution usable by non-lakeFS users by:
...
+routerfs.mapping.1.s3a1.replace='^s3a://bucket'
+routerfs.mapping.1.s3a1.with='s3a1://another-bucket'
+routerfs.mapping.1.minio.replace='^s3a://minio-bucket'
+routerfs.mapping.1.minio.with='minio://another-minio-bucket'
+
+# File System configurations
+fs.s3a.impl=RouterFileSystem
+fs.lakefs.impl=S3AFileSystem
+fs.s3a1.impl=S3AFileSystem
+fs.minio.impl=S3AFileSystem
```
arielshaqed, line 103: Very neat!  Perhaps show how to configure endpoint and access keys for one or more of the `S3AFileSystem`s?  Not required here, but the end-result docs will probably need something like that.

design/spark-co-exist-with-object-storages.md
```diff
@@ -20,122 +18,178 @@ Make the solution usable by non-lakeFS users by:
...
+```
+
+##### Default mapping configuration 
+
+`RouterFileSystem` requires a default mapping configuration in case that none of the mappings matches a URI.
```
arielshaqed, line 108: Is this even required?  I mean it's obviously a best practice, but I reckon it's OK to document what happens if you don't do this.  For instance, suppose it failed on error if no mapping matched.  Why is that a bad choice?  S3A with default settings does the same thing if I give it a bad URL: my access is denied or I get some other error message, and I know what to do.

arielshaqed (approved):
# 👍🏽 

Neat and looks very Spark-ly.

The running index used in configuration is somewhat unpleasant, but (as when we spoke F2F) I know of nothing better.  Perhaps @adipolak has a better idea for how to pass a (structured) _list_ in a Spark configuration?

design/spark-co-exist-with-object-storages.md
```diff
@@ -20,122 +18,178 @@ Make the solution usable by non-lakeFS users by:
...
-}
+`RouterFileSystem` has access to a configurable mapping that maps any object store URI to any type of URI. This proposal 
+uses s3 object store paths (with URI `scheme=s3a`) and lakeFS paths as examples.
+
+#### Mapping configurations
+
+The mapping configurations are spark properties of the following form:
+```properties
+routerfs.mapping.${mappingIdx}.${toFsScheme}.replace='^${fromFsScheme}://bucket/prefix'
+routerfs.mapping.${mappingIdx}.${toFsScheme}.with='${toFsScheme}://another-bucket/prefix/*'
```
Jonathan-Rosenberg, line 52: You might consider replacing the `${mappingIdx}` and `${toFsScheme}` position (i.e. `${toFsScheme}.${mappingIdx}`) for readability (the index is running "inside" of the scheme)

design/spark-co-exist-with-object-storages.md
```diff
@@ -0,0 +1,195 @@
...
+
+`RouterFileSystem` has access to a configurable mapping that maps any object store URI to any type of URI. This proposal 
+uses s3 object store paths (with URI `scheme=s3a`) and lakeFS paths as examples.
+
+#### Mapping configurations
+
+The mapping configurations are spark properties of the following form:
+```properties
+routerfs.mapping.${mappingIdx}.${toFsScheme}.replace='^${fromFsScheme}://bucket/prefix'
+routerfs.mapping.${mappingIdx}.${toFsScheme}.with='${toFsScheme}://another-bucket/prefix/*'
```
johnnyaug, line 52:
```suggestion
routerfs.mapping.${mappingIdx}.${toFsScheme}.with='${toFsScheme}://another-bucket/prefix/'
```

design/spark-co-exist-with-object-storages.md
```diff
@@ -0,0 +1,195 @@
...
+# File System configurations
+fs.s3a.impl=RouterFileSystem
+fs.lakefs.impl=S3AFileSystem
+fs.default.impl=S3AFileSystem
+```
+
+This configuration is required because otherwise `RouterFileSystem` will get stuck in an infinite loop, by calling itself 
+as the file system for `scheme=s3a` in the example. 
+
+**Note** Hadoop configurations include a `fs.default.impl` and we will need to test that this configuration is working as expected.
```
johnnyaug, line 129: From your examples I understand that the user can choose any scheme name for the default - so in the example I would use `s3a-wrapped`, `s3a1` or something less general than "default".

design/spark-co-exist-with-object-storages.md
```diff
@@ -0,0 +1,195 @@
...
+
+### Pros & Cons
+
+### Pros
+
+1. We already have experience developing Hadoop file systems, therefore, the ramp up should not be significant.
+2. `RouterFileSystem` suggests much simpler functionality than what lakeFSFS supports (it only needs to receive calls, translate paths, and route to the right file system), which reduces the estimated number of unknowns unknowns.
+3. It does not change anything related to the existing Spark<>lakeFS integrations.
+4. `RouterFileSystem` can support non-lakeFS use-cases because it does not relay on any lakeFS client.
+5. `RouterFileSystem` and can be developed in a separate repo and delivered as a standalone OSS product, we may be able to contribute it to Spark.
```
johnnyaug, line 187:
```suggestion
5. `RouterFileSystem` can be developed in a separate repo and delivered as a standalone OSS product, we may be able to contribute it to Spark. 
```

design/spark-co-exist-with-object-storages.md
```diff
@@ -0,0 +1,195 @@
...
+### Pros & Cons
+
+### Pros
+
+1. We already have experience developing Hadoop file systems, therefore, the ramp up should not be significant.
+2. `RouterFileSystem` suggests much simpler functionality than what lakeFSFS supports (it only needs to receive calls, translate paths, and route to the right file system), which reduces the estimated number of unknowns unknowns.
+3. It does not change anything related to the existing Spark<>lakeFS integrations.
+4. `RouterFileSystem` can support non-lakeFS use-cases because it does not relay on any lakeFS client.
+5. `RouterFileSystem` and can be developed in a separate repo and delivered as a standalone OSS product, we may be able to contribute it to Spark. 
+6. `RouterFileSystem` does not relay on per-bucket configurations that are only supported for hadoop-aws (includes S3AFileSystem implementation) versions >= [2.8.0](https://hadoop.apache.org/docs/r2.8.0/hadoop-aws/tools/hadoop-aws/index.html#Configurations_different_S3_buckets).
```
johnnyaug, line 188:
```suggestion
6. `RouterFileSystem` does not rely on per-bucket configurations that are only supported for hadoop-aws (includes S3AFileSystem implementation) versions >= [2.8.0](https://hadoop.apache.org/docs/r2.8.0/hadoop-aws/tools/hadoop-aws/index.html#Configurations_different_S3_buckets).  
```

johnnyaug (approved):
Very clear, thank you!

Jonathan-Rosenberg (approved):
Cool stuff

design/spark-co-exist-with-object-storages.md
talSofer: how about using `s3a-actual`?

design/spark-co-exist-with-object-storages.md
talSofer: Without a default mapping, how would routerFS find an underlying filesystem that will invoke the file system operation? 
For example, given the following configurations 
```properties
routerfs.mapping.1.lakefs.replace='^s3a://bucket/prefix'
routerfs.mapping.1.lakefs.with='lakefs://example-repo/dev/prefix'

# File System configurations
fs.s3a.impl=RouterFileSystem
fs.lakefs.impl=S3AFileSystem
```
and the path `"s3a://another-bucket/prefix'"`
What underlying fs RouterFS should use to invoke a `fs.open("s3a://another-bucket/prefix")`?

Please let me know if I'm missing something!

design/spark-co-exist-with-object-storages.md
talSofer: I renamed to s3a-default and removed the note, you are right thanks for the suggestion.

design/spark-co-exist-with-object-storages.md
talSofer: done

design/spark-co-exist-with-object-storages.md
talSofer: done

nopcoder:
lgtm, two things that will probably get at some point:
- replace regexp will probably will require group match
- document limitations for operations while working this package. How operations will execute when there are overlapping configurations. it can be mapping one file system under another.
Rx: delete recursive, /data, while one file system is the complete package, and the /data/lakefs is mapped to lakeFS.

talSofer:
> Thanks for this @talSofer ! This seems to be the right approach. There's some prior art that we can look at, as it appears we're not the first project to think about providing a wrapper for a Hadoop Filesystem.
> 
> Hadoop itself provides [FilterFilesystem](https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/fs/FilterFileSystem.html) which seems to do something similar (`"A FilterFileSystem contains some other file system, which it uses as its basic file system, possibly transforming the data along the way or providing additional functionality. "`).
> 
> There's even an example of how to override it: in this case, to write a checksum for every file that's written; see [ChecksumFileSystem](https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/fs/ChecksumFileSystem.html).
> 

Thanks @ozkatz! 
I took a look at  [FilterFilesystem](https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/fs/FilterFileSystem.html), and I don't think it's the right fit for us. FilterFS is initialized with another **(single)** file system, but we need to route file system operations to multiple file systems. Extending FilterFilesystem will be similar to extending [FileSystem](https://hadoop.apache.org/docs/stable/api/org/apache/hadoop/fs/FileSystem.html) because in both cases, we will need to do extra work to access more than a single file system at runtime. 

WDYT?

talSofer:
> lgtm, two things that will probably get at some point:
> 
> * replace regexp will probably will require group match
> * document limitations for operations while working this package. How operations will execute when there are overlapping configurations. it can be mapping one file system under another.
>   Rx: delete recursive, /data, while one file system is the complete package, and the /data/lakefs is mapped to lakeFS.

@nopcoder great points (especially 2 which is scary!), I'm adding this as a con to the proposal and we will need to do some discovery work there.

nopcoder (approved):

PR #2949, opened 2022-02-14
Live configuration reload will change logging level

itaiad200:
lakeFS reads the logging level when it is changed in the configuration file, and change the logging on the fly without restarting the app.

To enable more configuration properties to change, we should propagate them properly in the same manner. Since this isn't always trivial, I suggest we do it per property, when needed.

### DIFF
--- a/cmd/lakefs/cmd/run.go
+++ b/cmd/lakefs/cmd/run.go
@@ -96,7 +96,7 @@ var runCmd = &cobra.Command{
 		viper.WatchConfig()
 		viper.OnConfigChange(func(in fsnotify.Event) {
 			lvl := viper.GetString(config.LoggingLevelKey)
-			fmt.Println("Changing log level to", lvl)
+			logger.WithField("toLevel", lvl).Info("Changing log level")
 			logging.SetLevel(lvl)
 		})
 
--- a/cmd/lakefs/cmd/run.go
+++ b/cmd/lakefs/cmd/run.go
@@ -15,10 +15,12 @@ import (
 	"time"
 
 	"github.com/dlmiddlecote/sqlstats"
+	"github.com/fsnotify/fsnotify"
 	"github.com/go-ldap/ldap/v3"
 	"github.com/golang-migrate/migrate/v4"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 	"github.com/treeverse/lakefs/pkg/actions"
 	"github.com/treeverse/lakefs/pkg/api"
 	"github.com/treeverse/lakefs/pkg/auth"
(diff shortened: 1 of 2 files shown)


guy-har (approved):
Cool!

cmd/lakefs/cmd/run.go
```diff
@@ -91,6 +93,13 @@ var runCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		logger := logging.Default()
 		cfg := loadConfig()
+		viper.WatchConfig()
+		viper.OnConfigChange(func(in fsnotify.Event) {
+			lvl := viper.GetString("logging.level")
+			fmt.Println("Changing log level to", lvl)
+			logging.SetLevel(lvl)
```
Jonathan-Rosenberg, line 100: Do we want to handle an empty `lvl` string (in case that `GetString` returned such one)?

itaiad200: We handle it here the same way we handle it on startup -> Let viper decide.

cmd/lakefs/cmd/run.go
```diff
@@ -91,6 +93,13 @@ var runCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		logger := logging.Default()
 		cfg := loadConfig()
+		viper.WatchConfig()
+		viper.OnConfigChange(func(in fsnotify.Event) {
+			lvl := viper.GetString(config.LoggingLevelKey)
```
arielshaqed, line 98: Can we (additionally) read the rest of the config, and warn if something else changed?

cmd/lakefs/cmd/run.go
arielshaqed: What do we do if SetLevel failed?  I think we should at least issue an error here.

cmd/lakefs/cmd/run.go
```diff
@@ -91,6 +93,13 @@ var runCmd = &cobra.Command{
...
+			fmt.Println("Changing log level to", lvl)
```
arielshaqed, line 99: This should be a log, with times and everything.

arielshaqed (approved):
Cool!

cmd/lakefs/cmd/run.go
itaiad200: It can't fail and it doesn't return an error.

cmd/lakefs/cmd/run.go
itaiad200: Opening an issue, there's a lot to do here (save the previous config, compare, take defaults into account while only the logging level can really change), while the value is limited IMO (Better visibility for items that don't have any effect).

PR #2952, opened 2022-02-17
Default storage namespace

johnnyaug:
Added a new configuration `blockstore.default_namespace_prefix`.
When specified, the storage namespace field in the UI will be auto-filled.
The specified value will be the prefix, and the repository name will be added to it.

image

### Extra
* Added a (?) icon near the storage namespace field, linking to the docs.
* Small docs improvements.

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -51,6 +51,9 @@ This reference uses `.` to denote the nesting of values.
 * `auth.ldap.default_user_group` `(string : )` - Create all LDAP users in this group.  Defaults to `Viewers`.
 * `auth.ldap.user_filter` `(string : )` - Additional filter for users.
 * `blockstore.type` `(one of ["local", "s3", "gs", "azure", "mem"] : required)`.  Block adapter to use. This controls where the underlying data will be stored
+* `blockstore.default_namespace_prefix` `(string : )` - Use this to help your users choose a storage namespace for their repositories. 
+   When creating a repository from the UI, the storage namespace will be filled with this default value as a prefix.
+   The user may still change it to something else 
 * `blockstore.local.path` `(string: "~/lakefs/data")` - When using the local Block Adapter, which directory to store files in
 * `blockstore.gs.credentials_file` `(string : )` - If specified will be used as a file path of the JSON file that contains your Google service account key
 * `blockstore.gs.credentials_json` `(string : )` - If specified will be used as JSON string that contains your Google service account key (when credentials_file is not set)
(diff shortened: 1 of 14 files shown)


docs/reference/configuration.md
```diff
@@ -51,6 +51,9 @@ This reference uses `.` to denote the nesting of values.
 * `auth.ldap.default_user_group` `(string : )` - Create all LDAP users in this group.  Defaults to `Viewers`.
 * `auth.ldap.user_filter` `(string : )` - Additional filter for users.
 * `blockstore.type` `(one of ["local", "s3", "gs", "azure", "mem"] : required)`.  Block adapter to use. This controls where the underlying data will be stored
+* `blockstore.default_namespace_prefix` `(string : )` - Use this to help your users choose a storage namespace for their repositories. 
+   When creating a repository from the UI, the storage namespace will be filled with this default value as a prefix.
+   The user may still change it to something else
```
arielshaqed, line 56: "If missing, ..."?

docs/reference/object-model.md
```diff
@@ -146,9 +146,13 @@ other user-defined merge strategies for handling conflicts are on the roadmap.
 _Underlying storage_ is the area on some other object store that lakeFS uses to store object
 contents and some of its metadata.  We sometimes refer to underlying storage as _physical_.
 The path used to store the contents of an object is then termed a _physical path_.  The object
-itself on underlying storage is never modified, except to remove it entirely during some
+itself in the underlying storage is never modified, except to remove it entirely during some
```
arielshaqed, line 149: I prefer the original statement.

webui/src/lib/components/repositoryCreateForm.jsx
```diff
@@ -15,19 +18,24 @@ export const RepositoryCreateForm = ({ config, onSubmit, onCancel, error = null,
     const repoValidityRegex = /^[a-z0-9][a-z0-9-]{2,62}$/;
 
     const [formValid, setFormValid] = useState(false);
-    const [repoValid, setRepoValid] = useState(true);
+    const [repoValid, setRepoValid] = useState(null);
+    const defaultNamespacePrefix = config.default_namespace_prefix
 
-    const [storageNamespaceValid, setStorageNamespaceValid] = useState(true);
+    const [storageNamespaceValid, setStorageNamespaceValid] = useState(defaultNamespacePrefix ? true : null);
```
arielshaqed, line 24: Not sure the prefix itself is necessarily "valid"; don't we still need to check?

talSofer (commented):
Very nice!

> The specified value will be the prefix, and the repository name will be added to it.

Why are we adding the repository name to the prefix? to create a separate storage space for each repository (if so, why)?

webui/src/lib/components/repositoryCreateForm.jsx
```diff
@@ -52,7 +60,7 @@ export const RepositoryCreateForm = ({ config, onSubmit, onCancel, error = null,
             if (!formValid) {
                 return;
             }
-            onSubmit({
+             onSubmit({
```
arielshaqed, line 63: ?

webui/src/lib/components/repositoryCreateForm.jsx
```diff
@@ -63,20 +71,26 @@ export const RepositoryCreateForm = ({ config, onSubmit, onCancel, error = null,
             <Form.Group as={Row} controlId="id">
                 <Form.Label column sm={fieldNameOffset}>Repository ID</Form.Label>
                 <Col sm={sm}>
-                    <Form.Control type="text" autoFocus ref={repoNameField} onChange={checkRepoValidity}/>
-                    {!repoValid &&
+                    <Form.Control type="text" autoFocus ref={repoNameField} onChange={onRepoNameChange}/>
+                    {repoValid === false &&
```
arielshaqed, line 75:
```suggestion
                    {!repoValid &&
```
(think I prefer this.)

arielshaqed (approved):
Neat, thanks!

(Please also wait for OSS TF approval)

johnnyaug:
> Why are we adding the repository name to the prefix? to create a separate storage space for each repository (if so, why)?
@talSofer, yes.

Repositories should be in separate namespaces. We maybe not stress this enough in our docs, but some settings (like GC, branch protection) will override each other if repositories are in the same namespace.

webui/src/lib/components/repositoryCreateForm.jsx
johnnyaug: This is actually fixing a bug where the "Create repository" button becomes enabled even if you don't fill the storage namespace.

docs/reference/object-model.md
johnnyaug: Done

talSofer:
> @talSofer, yes.
Repositories should be in separate namespaces. We maybe not stress this enough in our docs, but some settings (like GC, branch protection) will override each other if repositories are in the same namespace.

I see, can you open an issue to improve docs and maybe force this by code + migrate environments that are structured differently?

PR #2954, opened 2022-02-18
Adding Class to all images for GTM Purpose

JamilAhmad:
Need to add a class attribute with value (tmtrack) to all images for google tag manager as per Noy Davidson instruction, it will be time-consuming to add class to all images one by one so I have added a script that will do this process globally, please review and approve if all good

### DIFF
--- a/docs/_includes/head_custom.html
+++ b/docs/_includes/head_custom.html
@@ -18,4 +18,8 @@
     $(function () {
         $(".tabs").tabs();
     });
+    //add class to all images for google Tag Manager
+    jQuery(document).ready(function(){
+        jQuery('img').addClass('tmtrack');
+    });
 </script>


JamilAhmad:
we can use both $ or Jquery both works, we are only about adding class to all images it's not about clickable or label, etc.
Noy Davidson mentioned that He needs the same Class for all images that's why need to run that script

nopcoder:
The $ is to be consistent with the rest of the code in the same file.
About the images, @NoyDavidson2 from what I understand we already get notified of all images getting clocked, and we like to select just the clickable?

NoyDavidson2:
@JamilAhmad After debugging the problem, it seems that the problem has been solved by filtering images on Google Tag Manager. 
No need to add the unique class to images. Thanks!

JamilAhmad:
Okay sounds good @NoyDavidson2

PR #2963, opened 2022-02-23
Exclude body validation on openapi object upload

### LINKED ISSUE #2905: Stream content in UploadObject
[`api.Controller.UploadObject`](https://github.com/treeverse/lakeFS/blob/0cb453004eae50665d8ea32e680e091ae1260f5b/pkg/api/controller.go#L1839) reads multipart uploads using `http.Request.FormFile` -- which can use up to 32MiB of memory _and_ create temporary files.

We probably don't need any of that (unless OpenAPI has already used `ParseMultipart` for us, in which case we're doomed...), because we know exactly what parts to expect.  Just use `MultipartReader()`, and stream the relevant part in with minimal use of memory.

### LINKED ISSUE #2907: UploadObject reads entire body into server memory
(This from reading the code, not yet from a live test.).

The [`UploadObject` OpenAPI definition](https://github.com/treeverse/lakeFS/blob/b213752a7192eca9129f8d12444639db9bc0c373/api/swagger.yml#L2583) specifies:
```yaml
      x-validation-exclude-body: false
```

So `ValidateRequest` (in package kin-openapi) calls `ValidateRequestBody`, and reads the entire body into memory.

nopcoder:
Fix #2907 

Skip upload object body validation that read the complete request body

### DIFF
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -2581,7 +2581,7 @@ paths:
       tags:
         - objects
       operationId: uploadObject
-      x-validation-exclude-body: false
+      x-validation-exclude-body: true
       requestBody:
         content:
           multipart/form-data:
--- a/clients/java/api/openapi.yaml
+++ b/clients/java/api/openapi.yaml
@@ -3128,7 +3128,7 @@ paths:
           description: Internal Server Error
       tags:
       - objects
-      x-validation-exclude-body: false
+      x-validation-exclude-body: true
       x-contentType: multipart/form-data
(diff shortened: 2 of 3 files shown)


nopcoder:
Note the docs yaml update have missing update from previous change

arielshaqed (approved):
Approved, but I do not think this fixes #2905.  That issue is that we use the multipart reader in a weird way, and it allocates 32 MiB (?) and reads into that.

We should also fix that one -- just this PR does not do it.

PR #2966, opened 2022-02-24
Bugfix/2935 lakectl bug on not found html

idanovo:


BinyaminSchein:
Loving thisssss

cmd/lakectl/cmd/annotate.go
```diff
@@ -31,29 +33,29 @@ var annotateCmd = &cobra.Command{
...
 				res, err := client.LogCommitsWithResponse(context, pathURI.Repository, pathURI.Ref, logCommitsParams)
-				DieOnResponseError(res, err)
+				DieOnErrorOrUnexpectedStatusCode(resp, err, http.StatusOK)
 				data := objectCommitData{
 					Object: obj.Path,
 				}
 				if len(res.JSON200.Results) > 0 {
 					data.Commit = res.JSON200.Results[0]
-					data.CommitMessage = splitOnNewLine(stringTrimLen((res.JSON200.Results[0].Message), annotateMessageSize))
+					data.CommitMessage = splitOnNewLine(stringTrimLen(res.JSON200.Results[0].Message, annotateMessageSize))
```
arielshaqed, line 58: This seems unrelated.  Also I am not sure it does what is intended here (I am not sure what is intended here).  Maybe leave it to a separate PR?

cmd/lakectl/cmd/common_helpers.go
```diff
@@ -206,6 +210,31 @@ func DieOnResponseError(response interface{}, err error) {
 		DieErr(retrievedErr)
 	}
 }
+func DieOnErrorOrUnexpectedStatusCode(response interface{}, err error, expectedStatusCode int) {
```
arielshaqed, line 213: Could we refactor parts of HTTPResponseAsError with this?  I think it copies some parts.

cmd/lakectl/cmd/doctor.go
```diff
@@ -127,7 +128,7 @@ func ListRepositoriesAndAnalyze(ctx context.Context) error {
 	case resp.JSON401 != nil:
 		return &CredentialsError{msgOnErrCredential, resp.JSON401.Message}
 	// In case we get the "not found" HTML page (the status is 200 and not 404 in this case).
-	case resp.HTTPResponse != nil && resp.HTTPResponse.StatusCode == 200:
+	case resp.HTTPResponse != nil && resp.HTTPResponse.StatusCode == 302:
```
arielshaqed, line 131: The comment doesn't change?

docs/reference/commands.md
```diff
@@ -1339,7 +1339,7 @@ lakectl branch-protect add <repo uri> <pattern> [flags]
 {:.no_toc}
 
 ```
-lakectl add lakefs://<repository> 'stable_*'
+lakectl branch-protect add lakefs://<repository> 'stable_*'
```
arielshaqed, line 1342: These seem unrelated, and I'd rather the author reviewed this.

arielshaqed (changes requested):
Thanks!

1 res/resp mixup that has to be fixed, and some things don't belong in this PR I think.

cmd/lakectl/cmd/doctor.go
idanovo: Which comment? The output comment of the `doctor`?

cmd/lakectl/cmd/common_helpers.go
idanovo: I've tried to, but it seems to be kind of difficult because of the structure of `ResponseAsError` func. All we want in our case is to get the status code from the response when `ResponseAsError` checks and returns an error on its way to get it.  WDYT?

docs/reference/commands.md
idanovo: Opened #2976

cmd/lakectl/cmd/annotate.go
```diff
@@ -31,29 +33,29 @@ var annotateCmd = &cobra.Command{
...
-			DieOnResponseError(resp, err)
+			DieOnErrorOrUnexpectedStatusCode(resp, err, http.StatusOK)
 			for _, obj := range resp.JSON200.Results {
 				logCommitsParams := &api.LogCommitsParams{
 					Amount:  api.PaginationAmountPtr(1),
 					Objects: &[]string{obj.Path},
 				}
 				res, err := client.LogCommitsWithResponse(context, pathURI.Repository, pathURI.Ref, logCommitsParams)
-				DieOnResponseError(res, err)
+				DieOnErrorOrUnexpectedStatusCode(resp, err, http.StatusOK)
```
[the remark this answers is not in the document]
arielshaqed: `tag.go` still contains a variable `res`:
```sh
[ariels@fedora lakeFS]$ git grep -w res cmd/
cmd/lakectl/cmd/tag.go:                 res, err := client.GetCommitWithResponse(ctx, tagURI.Repository, commitRef)
cmd/lakectl/cmd/tag.go:                 DieOnErrorOrUnexpectedStatusCode(res, err, http.StatusOK)
```

This inconsistency is precisely what got us into trouble on this line.  So could you fix that one too, please, even though it is not strictly related to this PR?

cmd/lakectl/cmd/tag.go
```diff
@@ -78,8 +78,8 @@ var tagCreateCmd = &cobra.Command{
 		force, _ := cmd.Flags().GetBool("force")
 		if force {
 			// checking validity of the commitRef before deleting the old one
-			resp, err := client.GetCommitWithResponse(ctx, tagURI.Repository, commitRef)
-			DieOnErrorOrUnexpectedStatusCode(resp, err, http.StatusOK)
+			res, err := client.GetCommitWithResponse(ctx, tagURI.Repository, commitRef)
```
arielshaqed, line 81: Sorry for being unclear: I really liked this change, it's part of the res -> resp fix.  It caused an issue with this PR, so I think we have no choice but to make this unrelated change here.  (Technically we could have a preview PR that just fixed this, and then your PR.  Too much effort for too little gain probably.)

arielshaqed (approved):
Thanks!

cmd/lakectl/cmd/tag.go
idanovo: not sure how we can fix this in this case. We are trying to assign `resp` in l.84 with a different type. Maybe change `res` here to be `_resp`?

cmd/lakectl/cmd/abuse.go
```diff
@@ -119,7 +120,7 @@ var abuseRandomWritesCmd = &cobra.Command{
 
 		client := getClient()
 		resp, err := client.GetRepositoryWithResponse(cmd.Context(), u.Repository)
-		DieOnResponseError(resp, err)
+		DieOnErrorOrUnexpectedStatusCode(resp, err, http.StatusOK)
```
nopcoder, line 123: suggestion - pick a shorter name

cmd/lakectl/cmd/annotate.go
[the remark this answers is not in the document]
nopcoder: My 2 cents - use different names so we will not verify or check the wrong response by mistake

cmd/lakectl/cmd/branch_protect.go
```diff
@@ -1,7 +1,10 @@
 package cmd
 
 import (
+	"net/http"
+
 	"github.com/spf13/cobra"
+
```
nopcoder, line 7: remove whitespace

cmd/lakectl/cmd/common_helpers.go
```diff
@@ -41,6 +43,8 @@ const (
 const internalPageSize = 1000          // when retreiving all records, use this page size under the hood
 const defaultAmountArgumentValue = 100 // when no amount is specified, use this value for the argument
 
+const defaultResponseOnSwaggerClient = http.StatusNoContent
```
nopcoder, line 46: Prefer to use http.StatusNoContent when needed - the code will be readable without this indirection

cmd/lakectl/cmd/root.go
```diff
@@ -108,9 +109,18 @@ func getClient() *api.ClientWithResponses {
 	if u.Path == "" || u.Path == "/" {
 		serverEndpoint = strings.TrimRight(serverEndpoint, "/") + api.BaseURL
 	}
+	httpClient := &http.Client{
+
```
nopcoder, line 113: remove whitespace

pkg/api/ui_handler.go
```diff
@@ -80,6 +80,8 @@ func NewHandlerWithDefault(root http.FileSystem, handler http.Handler, defaultPa
 		}
 		_, err := root.Open(path.Clean(urlPath))
 		if err != nil && os.IsNotExist(err) {
+			fmt.Println("redirecting...")
```
nopcoder, line 83: remove println

pkg/api/ui_handler.go
```diff
@@ -80,6 +80,8 @@ func NewHandlerWithDefault(root http.FileSystem, handler http.Handler, defaultPa
...
+			http.Redirect(w, r, defaultPath, http.StatusFound)
```
nopcoder, line 84: As far as I remember, you need to return after redirect and not continue the code flow

nopcoder (changes requested):

cmd/lakectl/cmd/abuse.go
idanovo: DieOnUnexpectedResponse?

nopcoder: Sorry, no good suggestion

nopcoder (approved):
Just address other comments

PR #2968, opened 2022-02-26
FindMergeBase - Removing redundant commit node visits

### LINKED ISSUE #2962: Diff takes more than 4 minutes to complete

itai-david:
Fixes #2962

This PR is about an efficiency issue discovered in ```FindMergeBase```
The function performs a BFS from 2 nodes, in order to find its best common ancestor
The BFS analyzes the parent commit nodes for each of the commit nodes and their parent commit nodes in turn
Since a certain commit can act as a parent for more than one commit, it is possible that it is met multiple times during the BFS and it is added multiple times. Once this is done, the entire ancestry chain is also processed multiple times and in case grid-like commits graph, this can quicky lead to a lot of redundant processing of the same nodes

This code removes the redundant processing as nodes are only processed (queued and scanned) once, but their 'reached from' status is still updated and checked each time these nodes are met. The first commit node to be reached from both source commits, is the requested ancestor - same as the old logic

Performance improvement:
Running locally, a diff that previously ran millions of iterations, ran 155 iterations instead, and finished within seconds.

### DIFF
--- a/pkg/graveler/ref/merge_base_finder.go
+++ b/pkg/graveler/ref/merge_base_finder.go
@@ -62,11 +62,7 @@ func FindMergeBase(ctx context.Context, getter CommitGetter, repositoryID gravel
 				// commit was reached from both left and right nodes
 				// Since it is not certain that the commit exist in the commits queue, a call
 				// to GetCommit is issued
-				commit, err := getter.GetCommit(ctx, repositoryID, parent)
-				if err != nil {
-					return nil, err
-				}
-				return commit, nil
+				return getter.GetCommit(ctx, repositoryID, parent)
 			}
 		}
 	}
--- a/pkg/graveler/ref/merge_base_finder.go
+++ b/pkg/graveler/ref/merge_base_finder.go
@@ -60,8 +60,8 @@ func FindMergeBase(ctx context.Context, getter CommitGetter, repositoryID gravel
 			reached[parent] |= commitFlags
 			if reached[parent]&fromLeft != 0 && reached[parent]&fromRight != 0 {
 				// commit was reached from both left and right nodes
-				// Since it is not certain that the commit exist in the commits queue, a call
-				// to GetCommit is issued
+				// At this point, we don't have the commit in-memory but only the corresponding commit record in the queue,
+				// therefore a call to GetCommit is issued.
 				return getter.GetCommit(ctx, repositoryID, parent)
 			}
 		}
--- a/pkg/graveler/ref/merge_base_finder.go
+++ b/pkg/graveler/ref/merge_base_finder.go
@@ -60,8 +60,6 @@ func FindMergeBase(ctx context.Context, getter CommitGetter, repositoryID gravel
 			reached[parent] |= commitFlags
 			if reached[parent]&fromLeft != 0 && reached[parent]&fromRight != 0 {
 				// commit was reached from both left and right nodes
-				// At this point, we don't have the commit in-memory but only the corresponding commit record in the queue,
-				// therefore a call to GetCommit is issued.
 				return getter.GetCommit(ctx, repositoryID, parent)
 			}
 		}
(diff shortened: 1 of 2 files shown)


arielshaqed (approved):
LGTM!  Consider also consulting @johnnyaug about this code :-)

Adding comments about how we might further greatly speed up this function to the issue: while the current algorithm seems very precise and fast, ACAIFT it might performs many redundant and *intentionally slow* database accesses for each `GetCommit` call.

pkg/graveler/ref/merge_base_finder.go
```diff
@@ -26,28 +26,57 @@ func FindMergeBase(ctx context.Context, getter CommitGetter, repositoryID gravel
...
-			heap.Push(&queue, &graveler.CommitRecord{CommitID: parent, Commit: parentCommit})
-			// mark the parent with the flag values from its descendents:
+			// mark the parent with the flag values from its descendents. This is done regardless
+			// of whether this parent commit is being queued in the current iteration or not. In
+			// both cases, if the 'reached' update signifies it was reached from both left and
+			// right nodes - it is the requested parent node
 			reached[parent] |= commitFlags
+			if reached[parent]&fromLeft != 0 && reached[parent]&fromRight != 0 {
+				// commit was reached from both left and right nodes
+				// Since it is not certain that the commit exist in the commits queue, a call
```
talSofer, line 63: Can you give an example when the parent commit is not in the queue at this point? 
at the first time the parent is reached it will be enqueued in line 51.

pkg/graveler/ref/merge_base_finder.go
```diff
@@ -26,28 +26,57 @@ func FindMergeBase(ctx context.Context, getter CommitGetter, repositoryID gravel
...
+				// to GetCommit is issued
+				commit, err := getter.GetCommit(ctx, repositoryID, parent)
+				if err != nil {
+					return nil, err
+				}
+				return commit, nil
```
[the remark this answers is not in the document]
talSofer: done

pkg/graveler/ref/merge_base_finder.go
johnnyaug: I think the comment is misleading - it can be removed.
We simply don't have the commit in memory at this point, so GetCommit is called.

arielshaqed: When we cache commits in memory, it will be.  Otherwise we might have read it or not, depending on the structure of the graph.  Agree wtih @johnnyaug the comment can be removed.

arielshaqed (approved):
Nice addition to the test!

pkg/graveler/ref/merge_base_finder.go
talSofer: I see, you are right we only have a CommitRecord at this point instead of graveler.Commit. thanks, I changed the comment.

johnnyaug: I don't think any comment is needed - it confuses more than it clears anything

talSofer: I removed the comment, thanks

pkg/graveler/ref/merge_base_finder.go
```diff
@@ -26,28 +26,57 @@ func FindMergeBase(ctx context.Context, getter CommitGetter, repositoryID gravel
...
+		return nil, err
+	}
+	if leftID == rightID {
+		return commit, nil
+	}
+
+	_, err = getCommitAndEnqueue(ctx, getter, &queue, repositoryID, rightID)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
itai-david: @johnnyaug - The fix is relying on the ```reached``` map to identify commit nodes that were already visited. Since the initial state marks the 2 target commit as reached, they will not be queued for further traversal when "reached" from the hypothetical node, so I had to enqueue both of the nodes, as part of the initial state

PR #2975, opened 2022-02-27
Log with multiple outputs

nopcoder:
Enable a way to log for multiple outputs. This change will apply to lakefs and lakectl.
- The `logging.output` (for lakeFS) and `--log-output` (for lakectl) can accept a list of comma separated outputs to use.
- The string "-" represents standard output
- The string "=" represents standard error
- Output files will be automatically rotated using lumberjack library


nessie/golden/lakectl_help.golden
```diff
@@ -29,13 +29,13 @@ Available Commands:
...
-      --base-uri string     base URI used for lakeFS address parse
-  -c, --config string       config file (default is $HOME/.lakectl.yaml)
-  -h, --help                help for lakectl
-      --log-format string   set logging output format
-      --log-level string    set logging level (default "none")
-      --log-output string   set logging output file
-      --no-color            don't use fancy output colors (default when not attached to an interactive terminal)
-  -v, --version             version for lakectl
+      --base-uri string      base URI used for lakeFS address parse
+  -c, --config string        config file (default is /home/barak/.lakectl.yaml)
```
ortz, line 33: I bet this is by mistake :)

ortz:
```suggestion
  -c, --config string        config file (default is $HOME/.lakectl.yaml)
```

pkg/logging/logger_test.go
```diff
@@ -0,0 +1,60 @@
...
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+func TestSetOutputs(t *testing.T) {
+	t.Run("default", func(t *testing.T) {
+		SetOutputs(nil, 0, 0)
+		if defaultLogger.Out != os.Stderr {
+			t.Error("Logger output should be stderr by default")
```
ortz, line 15: Is it? looking at the switch case under logger.go, it should be either lumberjack or none (eventually io.Discard)?

ortz (approved):
hey @nopcoder , some comments within the pull request.
Also, I must admit that using "-" and "=" for picking stderr or stdout is a bit confusing.
I think that we should either using the plain-text "stderr" / "stdout" (or the Linux's full path if it makes more sense to you - "/dev/stderr" and "/dev/stdout", or the Linux output redirection flags, "1" for stdout and "2" for stderr)

nessie/golden/lakectl_help.golden
nopcoder: > I bet this is by mistake :)

yes, the difference between `lakectl` vs `lakectl help`

pkg/logging/logger_test.go
nopcoder: I kept the default behavior where if nothing is configured the default logger (logrus) is using stderr

nopcoder:
> hey @nopcoder , some comments within the pull request. Also, I must admit that using "-" and "=" for picking stderr or stdout is a bit confusing. I think that we should either using the plain-text "stderr" / "stdout" (or the Linux's full path if it makes more sense to you - "/dev/stderr" and "/dev/stdout", or the Linux output redirection flags, "1" for stdout and "2" for stderr)

Thanks for the review. The "-" is there for backward compatability with the current lakefs configuration.
The value "-" is usually used in command line application like `aws cli` where a filename can use `-` to specify the standard output.
It is always true we can specify the linux path to stderr and stdour and it will work, but for the windows users and to have something that will work on all cases I've kept '-' and added '=' (as far fd 2) for the stderr.

PR #2978, opened 2022-02-27
Refactor AWS key generators to a separate package

arielshaqed:
Allows using them from other packages.

### DIFF
--- a/pkg/auth/service.go
+++ b/pkg/auth/service.go
@@ -689,8 +689,8 @@ func (s *DBAuthService) ListPolicies(ctx context.Context, params *model.Paginati
 }
 
 func (s *DBAuthService) CreateCredentials(ctx context.Context, username string) (*model.Credential, error) {
-	accessKeyID := compat.GenAccessKeyID()
-	secretAccessKey := compat.GenSecretAccessKey()
+	accessKeyID := keys.GenAccessKeyID()
+	secretAccessKey := keys.GenSecretAccessKey()
 	return s.AddCredentials(ctx, username, accessKeyID, secretAccessKey)
 }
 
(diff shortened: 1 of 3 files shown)


nopcoder (approved):
LGTM

pkg/auth/service.go
```diff
@@ -699,8 +689,8 @@ func (s *DBAuthService) ListPolicies(ctx context.Context, params *model.Paginati
 }
 
 func (s *DBAuthService) CreateCredentials(ctx context.Context, username string) (*model.Credential, error) {
-	accessKeyID := genAccessKeyID()
-	secretAccessKey := genSecretAccessKey()
+	accessKeyID := compat.GenAccessKeyID()
+	secretAccessKey := compat.GenSecretAccessKey()
```
Jonathan-Rosenberg, line 693:
```suggestion
	accessKeyID := keys.GenAccessKeyID()
	secretAccessKey := keys.GenSecretAccessKey()
```
?

pkg/gateway/operations/base.go
```diff
@@ -150,7 +150,7 @@ func (o *Operation) EncodeError(w http.ResponseWriter, req *http.Request, e erro
 
 func generateHostID() string {
 	const generatedHostIDLength = 8
-	return auth.HexStringGenerator(generatedHostIDLength)
+	return compat.HexStringGenerator(generatedHostIDLength)
```
Jonathan-Rosenberg, line 153:
```suggestion
	return keys.HexStringGenerator(generatedHostIDLength)
```
?

Jonathan-Rosenberg:
Other than the comments it looks good

pkg/gateway/operations/base.go
nopcoder: good catch - I think less inner-packages less issues

pkg/auth/service.go
arielshaqed: Yikes!  Bad rename, sorry (and thanks...)

pkg/gateway/operations/base.go
arielshaqed: Agreed.  But exporting all of `auth` drags in too much.  I just want to generate AWS credentials somewhere else.

arielshaqed (commented):
Thanks!

PR #2987, opened 2022-02-28
Fix  access key validation

idanovo:


cmd/lakectl/cmd/common_helpers.go
```diff
@@ -29,7 +29,7 @@ var noColorRequested = false
 // ErrInvalidValueInList is an error returned when a parameter of type list contains an empty string
 var ErrInvalidValueInList = errors.New("empty string in list")
 
-var accessKeyRegexp = regexp.MustCompile(`^AKIA[I|J][A-Z0-9]{14}Q$`)
+var accessKeyRegexp = regexp.MustCompile(`^AKIAJ[A-Z2-7]{14}Q$`)
```
arielshaqed, line 32: As explained at length in the summary, not a fan.  But if we're here, we mightfix the stray `|` in the regexp:
```suggestion
 var accessKeyRegexp = regexp.MustCompile(`^AKIA[IJ][A-Z0-9]{14}Q`)
```

arielshaqed (approved):
Sorry, I do not see the benefit of this change.  This functionality in the doctor is purely there in order to verify that something _looks like_ an access key.  That the access key is valid is verified by trying to use it with the lakeFS server.

I am not sure in what cases this change will improve the user experience.

I do however believe that it might make our lives harder in future, if we decide (for whatever reason) to use a `AKIAI...`-style access keys.  Then it will be just another place to remember to replace keys.  In fact, we _already_ use such keys!  The standard AWS example access key is AKIAIOSFODNN7EXAMPLE, we use it in some tests, and users might use it on (protected) testing environments.  With this change IIUC it will stop working.

So approving, but I am probably happier with the current situation... or with the removal of the `|` from the regexp.

cmd/lakectl/cmd/common_helpers.go
idanovo: @arielshaqed I agree. Thanks!

PR #2988, opened 2022-02-28
DBT: add branch creation capability

### LINKED ISSUE #2928: create-branch-schema -  create lakefs branch
create-branch-schema command should have an option to create the branch in lakefs

Jonathan-Rosenberg:
Currently when users want to copy their dbt schema from one branch to another they need to create this branch and then use the `lakectl dbt create-branch-schema` to copy the schema to the dedicated branch created.
This PR introduces the `--create-branch` flag which enables the user to create the branch in the context of `lakectl dbt create-branch-schema`. Since a schema is defined in the dbt project and is associated with a single branch, the source branch (from which we diverge) is the same as the schema's. 

Closes #2928


cmd/lakectl/cmd/branch_tools.go
```diff
@@ -0,0 +1,40 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"github.com/treeverse/lakefs/pkg/api"
```
idanovo, line 6: You can use `goimports` command in order to fix imports order issues

cmd/lakectl/cmd/branch_tools.go
```diff
@@ -0,0 +1,40 @@
...
+func CreateBranch(ctx context.Context, sourceLakefsBranchUri, destinationLakefsBranchUri string) {
+	branchURI := MustParseRefURI("destination branch uri", destinationLakefsBranchUri)
+	sourceURI := MustParseRefURI("source branch uri", sourceLakefsBranchUri)
+
+	client := getClient()
+	resp, err := client.CreateBranchWithResponse(ctx, branchURI.Repository, api.CreateBranchJSONRequestBody{
+		Name:   branchURI.Ref,
+		Source: sourceURI.Ref,
+	})
+	DieOnResponseError(resp, err)
```
idanovo, line 38: If you merge this after I merge [this](https://github.com/treeverse/lakeFS/pull/2966/files), please change this line to be `DieOnErrorOrUnexpectedStatusCode(resp, err, http.StatusCreated)`

cmd/lakectl/cmd/dbt.go
```diff
@@ -1,36 +1,90 @@
...
+var (
+	noSourceBranchError = errors.New(`given schema has no source lakefs branch associated with it`)
+)
+
+func (d DbtClient) ValidateGenerateSchemaMacro() {
+	err := dbtutil.ValidateGenerateSchemaMacro(d.projectDir, macrosDirName, generateSchemaName, lakectlIdentifier)
+	if err != nil {
+		switch e := err.(type) {
+		case *dbtutil.MissingSchemaIdentifierError:
+			DieFmt(e.Error())
```
idanovo, line 52: Why not use `DieErr` in this case?

cmd/lakectl/cmd/dbt.go
```diff
@@ -1,36 +1,90 @@
...
+	if err != nil {
+		fmt.Println(schemaOutput)
+		switch e := err.(type) {
+		case *dbtutil.MissingSchemaInDebugError:
+			DieFmt(e.Error())
+		default:
+			DieErr(e)
+		}
+	}
+	fmt.Printf("dbt debug succeeded with schemaOutput %s\n", schemaOutput)
```
idanovo, line 74:
```suggestion
	fmt.Println("dbt debug succeeded with schemaOutput:", schemaOutput)
```

cmd/lakectl/cmd/dbt.go
Jonathan-Rosenberg: As I understand it, `DieFmt` is used for invalid input. 
In this case the given schema identifier (in the dbt macro) is not the expected `lakectlIdentifier` identifier given here. Although it's not a "classic" input it does seem fit..

idanovo (commented):
Looks good! Only the comment about the `DieOnResponseError` is important

cmd/lakectl/cmd/branch_tools.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/treeverse/lakefs/pkg/api"
+	"github.com/treeverse/lakefs/pkg/metastore"
+)
+
+func ExtractRepoAndBranchFromDBName(ctx context.Context, dbName string, client metastore.Client) (string, string, error) {
+	metastoreDB, err := client.GetDatabase(ctx, dbName)
+
+	if err != nil {
+		return "", "", fmt.Errorf("failed to get database on extractBranchFromSchema from '%s': %w", dbName, err)
```
arielshaqed, line 15:
```suggestion
		return "", "", fmt.Errorf("\get database on extractBranchFromSchema from '%s': %w", dbName, err)
```
(Go errors describe what happened but do not say it is bad)

Also below.

cmd/lakectl/cmd/branch_tools.go
```diff
@@ -0,0 +1,41 @@
...
+	branchURI := MustParseRefURI("destination branch uri", destinationLakefsBranchUri)
+	sourceURI := MustParseRefURI("source branch uri", sourceLakefsBranchUri)
+
+	client := getClient()
+	resp, err := client.CreateBranchWithResponse(ctx, branchURI.Repository, api.CreateBranchJSONRequestBody{
+		Name:   branchURI.Ref,
+		Source: sourceURI.Ref,
+	})
+	DieOnResponseError(resp, err)
+	Fmt("created branch '%s' %s\n", branchURI.Ref, string(resp.Body))
```
arielshaqed, line 40: Please use a formatting output with a template.  It supports colouring and other neat features, and is generally simpler to maintain (for user-visible outputs).

cmd/lakectl/cmd/branch_tools_test.go
```diff
@@ -0,0 +1,105 @@
...
+	tests := []struct {
+		name      string
+		args      args
+		want      string
+		want1     string
+		wantErr   bool
+		validPath bool
+	}{
+		{
+			name: "Happy flow",
```
arielshaqed, line 37: Nit:
```suggestion
			name: "Valid",
```
(or similar).  The other test flows are less happy, but I don't want to say "Sad" about each of them.

cmd/lakectl/cmd/branch_tools_test.go
```diff
@@ -0,0 +1,105 @@
...
+)
+
+var (
+	validPath   = fmt.Sprintf("s3://%s/%s/path/to/schema", repoName, branchName)
+	invalidPath = fmt.Sprintf("s3/%s/%s/", repoName, branchName)
+)
+
+func TestExtractRepoAndBranchFromDBName(t *testing.T) {
+	type args struct {
+		ctx    context.Context
```
arielshaqed, line 25: If we never pass a context, can we just have `dbName` here?  (If we _do_ pass a context, Go generally prefers `context.Background()` or `context.TODO()` to the `nil` context...).

cmd/lakectl/cmd/branch_tools_test.go
```diff
@@ -0,0 +1,105 @@
...
+		dbName string
+	}
+	tests := []struct {
+		name      string
+		args      args
+		want      string
+		want1     string
+		wantErr   bool
```
arielshaqed, line 33: This test is very brittle.

Can `wantErr` be an `error`, that we can then check with `errors.Is`?  Or, if the is impossible, a substring of the expected error??  Otherwise any scenario with `wantErr == true` (most of them) will happily pass as long as _something_ on the execution path fails.

cmd/lakectl/cmd/branch_tools_test.go
```diff
@@ -0,0 +1,105 @@
...
+func TestExtractRepoAndBranchFromDBName(t *testing.T) {
+	type args struct {
+		ctx    context.Context
+		dbName string
+	}
+	tests := []struct {
+		name      string
+		args      args
+		want      string
+		want1     string
```
arielshaqed, line 32: Please document these fields.

cmd/lakectl/cmd/branch_tools_test.go
```diff
@@ -0,0 +1,105 @@
...
+		wantErr   bool
+		validPath bool
```
arielshaqed, line 34: Please export all fields in a `tests` struct.  It is often tempting to use`deep.Equal` or something else that relies on reflection on these at some later stage.  And that will compare no fields because Go reflection cannot see lowercase (unexported) fields.

cmd/lakectl/cmd/dbt.go
```diff
@@ -1,36 +1,91 @@
...
 	"github.com/spf13/cobra"
+	"github.com/treeverse/lakefs/pkg/dbtutil"
 	"github.com/treeverse/lakefs/pkg/metastore"
 	mserrors "github.com/treeverse/lakefs/pkg/metastore/errors"
 )
 
+type DbtCommandClient interface {
+	// Run performs 'dbt run' over the given schema and selected materializations
+	Run(schema string, selectValues []string) (string, error)
+	// Debug validates dbt connection using dbt debug, return the schema configured by the target environment (configured in the dbt profiles file)
```
arielshaqed, line 22:
```suggestion
	// Debug validates dbt connection using dbt debug and returns the schema
	// configured by the target environment (configured in the dbt profiles file)
```

cmd/lakectl/cmd/dbt.go
```diff
@@ -1,36 +1,91 @@
...
+	Debug() string
+	// Ls performs 'dbt ls' over the given schema and selected materializations
+	Ls(resourceType string, selectStatement []string) []dbtutil.DBTResource
+	// ValidateGenerateSchemaMacro validates that the "generate_schema_name.sql" file was created and used correctly (according to implementation)
+	ValidateGenerateSchemaMacro()
```
arielshaqed, line 27: Consider documenting that it dies if something is invalid.

cmd/lakectl/cmd/dbt.go
```diff
@@ -1,36 +1,91 @@
...
 	"regexp"
-	"strings"
 
 	"github.com/spf13/cobra"
+	"github.com/treeverse/lakefs/pkg/dbtutil"
 	"github.com/treeverse/lakefs/pkg/metastore"
 	mserrors "github.com/treeverse/lakefs/pkg/metastore/errors"
 )
 
+type DbtCommandClient interface {
```
arielshaqed, line 19:
```suggestion
type DBTCommandClient interface {
```
(and in other `type`s and `func`s below): the Go way is to keep capitalization of acronyms and abbreviations.  So we always have `XyzzyID` (with a capital D), etc.

cmd/lakectl/cmd/dbt.go
arielshaqed: `DieErr` is the way to die on an error.  `DieFmt` is if you have a message but not an error.  (Yes, one could be implemented in terms of the other.)

In any case, `DieFmt` is dangerous to use like this because the error message may contain formatting chars.  If you must, use `DieFmt("%s", ...)`.

cmd/lakectl/cmd/dbt.go
```diff
@@ -45,53 +100,59 @@ var dbtCreateBranchSchema = &cobra.Command{
...
 		branchName, _ := cmd.Flags().GetString("branch")
+		shouldCreateBranch, _ := cmd.Flags().GetBool(createLakefsBranchFlag)
 		toSchema, _ := cmd.Flags().GetString("to-schema")
 		projectRoot, _ := cmd.Flags().GetString("project-root")
 		skipViews, _ := cmd.Flags().GetBool("skip-views")
 		dbfsLocation, _ := cmd.Flags().GetString("dbfs-location")
 		continueOnError, _ := cmd.Flags().GetBool("continue-on-error")
 		continueOnSchemaExists, _ := cmd.Flags().GetBool(continueOnSchemaFlag)
 
+		if len(toSchema) == 0 {
```
arielshaqed, line 111:
```suggestion
		if toSchema == "" {
```

cmd/lakectl/cmd/dbt.go
```diff
@@ -45,53 +100,59 @@ var dbtCreateBranchSchema = &cobra.Command{
...
+			toSchema = branchName
+		}
+
+		dbtClient := DbtClient{projectDir: projectRoot}
+		metastoreClient, clientDeferFunc := getMetastoreClient(clientType, "")
+		defer clientDeferFunc()
```
arielshaqed, line 117:
```suggestion
		metastoreClient, clos := getMetastoreClient(clientType, "")
		defer close()
```
(name the variable for what it is, not for how it will be used on the next line).

arielshaqed (commented):
Thanks, looks really good!

Added many Go style comments (sorry).  Not enough of an expert on DBT to say much interesting on that side, except for... "how was this tested?".

Please also fix Go linter warnings.

cmd/lakectl/cmd/branch_tools_test.go
Jonathan-Rosenberg: The other flows are sad from my POV but sure 😉

cmd/lakectl/cmd/branch_tools.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// CreateBranch creates a new branch with the given source and destination branch URIs
+func CreateBranch(ctx context.Context, sourceLakeFSBranchURI, destinationLakeFSBranchURI string) {
+	branchURI := MustParseRefURI("destination branch URI", destinationLakeFSBranchURI)
+	sourceURI := MustParseRefURI("source branch URI", sourceLakeFSBranchURI)
+
+	client := getClient()
+	resp, err := client.CreateBranchWithResponse(ctx, branchURI.Repository, api.CreateBranchJSONRequestBody{
+		Name:   branchURI.Ref,
```
johnnyaug, line 85: You had the repository and branch names in hand - then you made a URI out of them, and parsed the URI again to extract the names.

Why not just use the names in the first place?

pkg/dbtutil/dbt_util.go
```diff
@@ -0,0 +1,103 @@
+package dbtutil
```
johnnyaug, line 1: IMO this file doesn't justify a top-level package.
Maybe it can go under metastore > dbt > utils.go.
@nopcoder may have an opinion.

pkg/dbtutil/dbt_util_test.go
```diff
@@ -0,0 +1,285 @@
...
+				return
+			}
+			if got != tt.runOutput {
+				t.Errorf("DbtRun() got = %v, schemaName %v", got, tt.runOutput)
+			}
+		})
+	}
+}
+
+func generateJsonStringFromResources(resources ...DBTResource) string {
```
johnnyaug, line 277:
```suggestion
func toJSON(resources ...DBTResource) string {
```

cmd/lakectl/cmd/branch_tools.go
```diff
@@ -0,0 +1,95 @@
...
+var branchCreatedTemplate = `created branch "{{.Branch.Ref}}" {{.Resp}}
+`
+
+// MissingDBError means that the requested DB is missing/not found
+type MissingDBError struct {
+	dbName string
+	err    error
+}
+
+func (mdbe MissingDBError) Error() string {
```
johnnyaug, line 23: That's a lot of work to create custom errors - and I'm not sure why they are needed.

Can something like the following work?
```
fmt.Errorf("trying to reach database: %w", originalErr)
```

johnnyaug (changes requested):
Looking good! And totally makes sense to implement.

I think some parts can be shorter and some naming can be improved - left comments.

cmd/lakectl/cmd/branch_tools.go
```diff
@@ -0,0 +1,95 @@
...
+	if err != nil {
+		return "", "", ExtractSourceBranchError{locationURI: metastoreDB.LocationURI, err: err}
+	}
+
+	return repo, branch, nil
+}
+
+// GenerateLakeFSBranchURIFromRepoAndBranchName generates a valid URI from the given repository and branch names
+func GenerateLakeFSBranchURIFromRepoAndBranchName(repoName, branchName string) (string, error) {
+	if len(strings.TrimSpace(repoName)) == 0 || len(strings.TrimSpace(branchName)) == 0 {
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: As a general purpose method (not part of a business logic flow) I don't think we can assume anything about the input...
WDYT?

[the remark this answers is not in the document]
johnnyaug: I think it's fine to assume that the arguments are valid at this point

cmd/lakectl/cmd/branch_tools.go
johnnyaug: What I mean is, why do you need a URI at all?

Jonathan-Rosenberg: The purpose was to create a function that takes URI strings and builds a new branch out of them, that way if other flows will need to create a branch from given URI strings it would be easier for them (they won't need to parse the URIs but just pass it to this function).
I could have created a function that takes the repo and branch as you said but I thought it would be more useful to receive the URI strings because the convention is that the user inputs the full lakefs path `lakefs://{repo}/{branch}` rather then the repo and branch separately, thus making the use of this function easier...

cmd/lakectl/cmd/branch_tools.go
Jonathan-Rosenberg: This was done to test the code not by the string but by the type of the error...
Go doesn't consider custom errors to be a good practice? (truly don't know 😅 )

test/lakectl_metastore/dbt-entrypoint.sh
```diff
@@ -14,6 +14,16 @@ export LAKEFS_SCHEMA=test
 
 dbt run --select star_rating
 
+unset LAKEFS_SCHEMA
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: If we take it off then we'll need to `lakectl commit` the previous step (`lakectl dbt create-branch-schema ..`) so the next will succeed. 
Why is it diverting our attention?

cmd/lakectl/cmd/common_helpers.go
```diff
@@ -201,14 +201,14 @@ func RetrieveError(response interface{}, err error) error {
 	return helpers.ResponseAsError(response)
 }
 
-func DieOnResponseError(response interface{}, err error) {
+func dieOnResponseError(response interface{}, err error) {
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: It might be a handy function but as I see it, no method is using it expect for `DieOnErrorOrUnexpectedStatusCode`, so what are the odds the a new one will?
On top of that, developers that used to use `DieOnResponseError` might use it again as a force of habit which we want to prevent.
wdyt?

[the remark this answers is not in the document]
idanovo: Agree with @Jonathan-Rosenberg about this. we want to prevent users from using `DieOnResponseError`

cmd/lakectl/cmd/branch_tools.go
johnnyaug: I think we should not generate lakeFS URIs in our code.
This is error prone. URIs should be used from the outside to identify resources in lakeFS.

Also, if you choose to stay with this function it may belong in the "uri" package.

johnnyaug:
Nice! I think it's close to being ready.
Not approving because there is one open discussion about whether to generate URIs in the code, and also one unaddressed comment about the creation of the "dbt_utils" package.

cmd/lakectl/cmd/branch_tools.go
```diff
@@ -0,0 +1,90 @@
...
+	repo, branch, err := metastore.ExtractRepoAndBranch(metastoreDB.LocationURI)
+	if err != nil {
+		return "", "", ExtractSourceBranchError{locationURI: metastoreDB.LocationURI, err: err}
+	}
+
+	return repo, branch, nil
+}
+
+// GenerateLakeFSURI generates a valid URI from the given repository and branch names
+func GenerateLakeFSURI(repoName, branchName string) string {
```
nopcoder, line 69: Looks like internal func and not public.
Name suggestion - use Format... and not Generate.

cmd/lakectl/cmd/branch_tools_test.go
```diff
@@ -0,0 +1,117 @@
...
+	wantErr   bool
+}
+
+func TestGenerateLakeFSURI(t *testing.T) {
+	got := GenerateLakeFSURI(repoName, branchName)
+	expectedURI := fmt.Sprintf("lakefs://%s/%s", repoName, branchName)
+	if got != expectedURI {
+		t.Errorf("GenerateLakeFSURI() got = %v, repositoryName %v", got, expectedURI)
+	}
+}
```
nopcoder, line 117: Testing the implementing by the code the same implementation - this is usually the point where I stop and explain to myself why and what I want to write test code in the first place.
In this case, I don't see the point in testing Sprintf and use a validation function or working code that use the data as an input to validate the value is what I expected - without re-implement it.

cmd/lakectl/cmd/branch_tools.go
```diff
@@ -0,0 +1,90 @@
...
+	return fmt.Sprintf(`lakefs://%s/%s`, repoName, branchName)
+}
+
+// CreateBranch creates a new branch with the given source and destination branch URIs
+func CreateBranch(ctx context.Context, sourceLakeFSBranchURI, destinationLakeFSBranchURI string) {
+	branchURI := MustParseRefURI("destination branch URI", destinationLakeFSBranchURI)
```
nopcoder, line 75: I think in this level, you should accept the parsed version and not `MustXXX`. Calling `MustParseRefURI` is handling user input or constants not our generated input.
In this case, it looks like we could just accept each repository and branch name.

cmd/lakectl/cmd/branch_tools_test.go
```diff
@@ -0,0 +1,117 @@
...
+
+const (
+	dbName     = "source"
+	repoName   = "repo"
+	branchName = "branch"
+)
+
+var (
+	validPath   = fmt.Sprintf("s3://%s/%s/path/to/schema", repoName, branchName)
+	invalidPath = fmt.Sprintf("s3/%s/%s/", repoName, branchName)
```
nopcoder, line 21: Can be added to the constants above just by using '+'. But I really prefer to see these values inside the single func scope it is used.  Using globals, even for test code, without a specifc name - it is something I try to avoid.

cmd/lakectl/cmd/branch_tools_test.go
```diff
@@ -0,0 +1,117 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			var schemaPath string
+			if tt.validPath {
+				schemaPath = validPath
+			} else {
+				schemaPath = invalidPath
+			}
```
nopcoder, line 78: Why not passing the correct constant in the test data? It will also open the case where we like to pass something else like empty string.

cmd/lakectl/cmd/branch_tools_test.go
```diff
@@ -0,0 +1,117 @@
...
+				t.Errorf("ExtractRepoAndBranchFromDBName() got = %v, repositoryName %v", got, tt.repositoryName)
+			}
+			if got1 != tt.branchName {
+				t.Errorf("ExtractRepoAndBranchFromDBName() got1 = %v, repositoryName %v", got1, tt.branchName)
+			}
+		})
+	}
+}
+
+func initializeMockClient(t *testing.T, schemaPath string) *mock.MSClient {
```
nopcoder, line 95: Naming: looks like NewMockClient

cmd/lakectl/cmd/dbt.go
```diff
@@ -15,22 +12,73 @@ import (
...
-	continueOnSchemaFlag = "continue-on-schema-exists"
+	lakectlIdentifier      = "LAKEFS_SCHEMA"
+	macrosDirName          = "macros"
+	generateSchemaName     = "generate_schema_name.sql"
+	continueOnSchemaFlag   = "continue-on-schema-exists"
+	createLakefsBranchFlag = "create-branch"
+)
+
+var (
+	errorNoSourceBranch = errors.New(`given schema has no source lakeFS branch associated with it`)
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: 1. Will be changed.
2. Why don't you consider that an error? If the user asks for a branch to be created and a schema be created with relation to it, the user needs its current dbt defined schema to be initialized with a branch location. If the user asks for a branch to get created and there's no such schema, it's an error just like if the user would want to create a branch with a source branch that doesn't exist, no?

cmd/lakectl/cmd/dbt_util_test.go
```diff
@@ -0,0 +1,263 @@
...
+			},
+			RunOutput: fmt.Sprintf("hello %s=%s", schemaKey, schemaValue),
+			Err:       dummyCommandExecutorError{},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.Name, func(t *testing.T) {
+			runOutput, err := DbtRun(tt.Args.ProjectRoot, tt.Args.Schema, tt.Args.SchemaEnvVarIdentifier, tt.Args.SelectValues, tt.Args.Executor)
+			if (err != nil) && !errors.Is(err, tt.Err) {
+				t.Errorf("DbtRun() error = %v, ErrType %v", err, tt.Err)
```
nopcoder, line 245:
```suggestion
				t.Errorf("DbtRun() error = '%v', expected error = '%v'", err, tt.Err)
```

cmd/lakectl/cmd/dbt_util_test.go
```diff
@@ -0,0 +1,263 @@
...
+				return
+			}
+			if runOutput != tt.RunOutput {
+				t.Errorf("DbtRun() runOutput = %v, schemaName %v", runOutput, tt.RunOutput)
```
nopcoder, line 249:
```suggestion
				t.Errorf("DbtRun() runOutput = %v, expected = %v", runOutput, tt.RunOutput)
```

cmd/lakectl/cmd/dbt_util_test.go
```diff
@@ -0,0 +1,263 @@
...
+					envReturn: false,
+				},
+			},
+			RunOutput: "hello",
+			Err:       nil,
+		},
+		{
+			Name: "command returns an error",
+			Args: args{
+				ProjectRoot:            "nevermind",
```
nopcoder, line 227: when every arg is the same for all tests I usually remove it from the code and just pass it as argument

cmd/lakectl/cmd/dbt_util_test.go
```diff
@@ -0,0 +1,263 @@
...
+	err       error
+	output    string
+	envReturn bool
+}
+
+type dummyCommandExecutorError struct{}
+
+func (dcee dummyCommandExecutorError) Error() string {
+	return "BOOM"
+}
```
nopcoder, line 24: can we use:
```go
var errDummyCommandFailed = errors.New("boom!")
```

nopcoder (approved):
looks good - added some suggestions and alignments.

cmd/lakectl/cmd/dbt_util.go
```diff
@@ -41,14 +26,16 @@ type CommandExecutor interface {
...
+
 func ValidateGenerateSchemaMacro(projectRoot, macrosDirName, generateSchemaFileName, schemaIdentifier string) error {
 	p := path.Join(projectRoot, macrosDirName, generateSchemaFileName)
 	data, err := ioutil.ReadFile(p)
 	if err != nil {
 		return err
 	}
 	if !strings.Contains(string(data), schemaIdentifier) {
-		return MissingSchemaIdentifierError{schemaIdentifier: schemaIdentifier, generateSchemaFileName: generateSchemaFileName}
+		return fmt.Errorf("generate_schema_name does not contain %s addition. Handle lakefs support to %s or use skip-views flag: %w", schemaIdentifier, generateSchemaFileName, err)
```
arielshaqed, line 38: I don't understand this error message, specifically "Handle lakefs support".

In general error messages should not contain more than a single phrase, because `Errorf` will be used on them.

cmd/lakectl/cmd/dbt.go
```diff
@@ -110,53 +58,57 @@ var dbtCreateBranchSchema = &cobra.Command{
...
-		DieErr(err)
+		DieFmt(`Given schema has no source branch associated with it %s`, err)
 	}
-	CreateBranch(ctx, sourceBranchURI, destinationBranchURI)
+	CreateBranch(ctx, sourceRepo, sourceBranch, branchName)
 }
 
 // copySchemaWithDbtTables create a copy of fromSchema and copies all dbt models materialized as table or incremental
-func copySchemaWithDbtTables(ctx context.Context, continueOnError, continueOnSchemaExists bool, fromSchema, toSchema, branchName, dbfsLocation string, dbtClient dbtCommandClient, client metastore.Client) {
+func copySchemaWithDbtTables(ctx context.Context, continueOnError, continueOnSchemaExists bool, fromSchema, toSchema, branchName, dbfsLocation, projectRoot string, client metastore.Client) {
```
arielshaqed, line 111: Another nit if we're already here: style is "DBT" not "Dbt".

```suggestion
func copySchemaWithDBTTables(ctx context.Context, continueOnError, continueOnSchemaExists bool, fromSchema, toSchema, branchName, dbfsLocation, projectRoot string, client metastore.Client) {
```

cmd/lakectl/cmd/dbt.go
```diff
@@ -166,24 +118,24 @@ func copySchemaWithDbtTables(ctx context.Context, continueOnError, continueOnSch
...
+	models := getDbtTables(projectRoot)
 	if err := copyModels(ctx, models, continueOnError, fromSchema, toSchema, branchName, dbfsLocation, client); err != nil {
 		DieErr(err)
 	}
 }
 
-func createViews(dbtClient dbtCommandClient, toSchema string) {
-	output, err := dbtClient.Run(toSchema, []string{"config.materialized:view"})
+func createViews(projectRoot, toSchema string) {
+	output, err := DbtRun(projectRoot, toSchema, lakectlIdentifier, []string{"config.materialized:view"}, commandExecutioner{})
```
arielshaqed, line 131: More nits:
```suggestion
	output, err := DBTRun(projectRoot, toSchema, lakectlIdentifier, []string{"config.materialized:view"}, commandExecutioner{})
```
(Can you go over the entire file and see we have no "Dbt", please?  (Can be in another PR).

arielshaqed (approved):
Thanks!

Sorry for the lengthy review.  Approving; please see if you can:
1. Improve strange error message.
2. Rename `commandExecutor` (many comments).
3. Get rid of "Dbt" in favour of "DBT" (can be in a separate PR).

johnnyaug (approved):
Looks very good!
Thanks for addressing all the comments!

PR #2989, opened 2022-02-28
Simplify create repo docs section

johnnyaug:


peacing (approved):
I like this better. More clear and to the point.

PR #3011, opened 2022-03-06
Fixing restore refs performance issues for old commit dups without "generation" field

idanovo:
Changing the FillGeneration func to read all commits to memory and calculate generation using BFS and update the DB with the results

### DIFF
--- a/pkg/graveler/ref/manager.go
+++ b/pkg/graveler/ref/manager.go
@@ -28,6 +28,12 @@ type Manager struct {
 	batchExecutor   batch.Batcher
 }
 
+type CommitNode struct {
+	childCommitsIds   []graveler.CommitID
+	notVisitedParents int
+	generation        int
+}
+
 func NewPGRefManager(executor batch.Batcher, db db.Database, addressProvider ident.AddressProvider) *Manager {
 	return &Manager{
 		db:              db,
@@ -387,6 +393,13 @@ func (m *Manager) addCommit(tx db.Tx, repositoryID graveler.RepositoryID, commit
 	return err
 }
 
+func (m *Manager) updateCommitGeneration(tx db.Tx, repositoryID graveler.RepositoryID, commitID graveler.CommitID, generation int) error {
+	_, err := tx.Exec(`UPDATE graveler_commits SET generation=$3 WHERE repository_id=$1 AND id=$2`,
+		repositoryID, commitID, generation)
+
+	return err
+}
+
 func (m *Manager) FindMergeBase(ctx context.Context, repositoryID graveler.RepositoryID, commitIDs ...graveler.CommitID) (*graveler.Commit, error) {
 	const allowedCommitsToCompare = 2
 	if len(commitIDs) != allowedCommitsToCompare {
@@ -412,19 +425,77 @@ func (m *Manager) ListCommits(ctx context.Context, repositoryID graveler.Reposit
 }
 
 func (m *Manager) FillGenerations(ctx context.Context, repositoryID graveler.RepositoryID) error {
-	_, err := m.db.Transact(ctx, func(tx db.Tx) (interface{}, error) {
-		return tx.Exec(`WITH RECURSIVE cte AS
-								(
-									SELECT id, 1 AS generation,creation_date
-									FROM graveler_commits
-									WHERE repository_id = $1 AND parents IS NULL OR array_length(parents,1) = 0
-									UNION ALL
-									SELECT DISTINCT(c.id), cte.generation+1 AS generation,c.creation_date
-									FROM graveler_commits c INNER JOIN cte
-									ON cte.id = ANY(c.parents) WHERE repository_id = $1
-								)
-								UPDATE graveler_commits u SET generation = cte.generation FROM cte WHERE u.id = cte.id;`,
-			repositoryID)
-	})
-	return err
+	iter, err := m.ListCommits(ctx, repositoryID)
+	if err != nil {
+		return err
+	}
+	defer iter.Close()
+
+	commitIdToNode := make(map[graveler.CommitID]CommitNode)
+	var rootCommitID graveler.CommitID
+
+	// make a commits "tree" from map data structure
+	for iter.Next() {
+		commit := iter.Value()
+		if len(commit.Parents) == 0 {
+			rootCommitID = commit.CommitID
+		}
+
+		// adding current node as a child to all parents nodes
+		for _, commitParentId := range commit.Parents {
+			commitParentNode, exist := commitIdToNode[commitParentId]
+			var parentCommitChildrenCommitsIds = []graveler.CommitID{commit.CommitID}
+			var parentCommitNotVisitedParents int
+			if exist {
+				parentCommitChildrenCommitsIds = append(commitParentNode.childCommitsIds, commit.CommitID)
+				parentCommitNotVisitedParents = commitParentNode.notVisitedParents
+			}
+			commitIdToNode[commitParentId] = CommitNode{parentCommitChildrenCommitsIds, parentCommitNotVisitedParents, 0}
+		}
+
+		currentNode, exist := commitIdToNode[commit.CommitID]
+		var currentCommitChildrenIds []graveler.CommitID
+		if exist {
+			currentCommitChildrenIds = currentNode.childCommitsIds
+		}
+		commitIdToNode[commit.CommitID] = CommitNode{currentCommitChildrenIds, len(commit.Parents), 0}
+	}
+	if err := iter.Err(); err != nil {
+		return fmt.Errorf("getting value from iterator: %w", err)
+	}
+
+	nodesQueue := make([]graveler.CommitID, 0)
+	nodesQueue = append(nodesQueue, rootCommitID)
+	currentGeneration := 1
+	// update commitNodes' generation in commitIdToNode "tree" using BFS algorithm.
+	// using a queue implementation
+	// adding a node to the queue only after all of its parents were visited in order to avoid redundant visits of nodes
+	for ; len(nodesQueue) > 0; currentGeneration++ {
+		var nodesIdsToIterateNextIteration []graveler.CommitID
+		for len(nodesQueue) > 0 {
+			currentCommitId := nodesQueue[0]
+			currentNode := commitIdToNode[currentCommitId]
+			commitIdToNode[currentCommitId] = CommitNode{currentNode.childCommitsIds, 0, currentGeneration}
+			_, err := m.db.Transact(ctx, func(tx db.Tx) (interface{}, error) {
+				return nil, m.updateCommitGeneration(tx, repositoryID, currentCommitId, currentGeneration)
+			})
+			if err != nil {
+				return err
+			}
+			nodesQueue = nodesQueue[1:]
+			for _, childNodeId := range currentNode.childCommitsIds {
+				childNode := commitIdToNode[childNodeId]
+				commitIdToNode[childNodeId] = CommitNode{childNode.childCommitsIds, childNode.notVisitedParents - 1, 0}
+				// if all parents already have been visited
+				if childNode.notVisitedParents == 1 {
+					nodesIdsToIterateNextIteration = append(nodesIdsToIterateNextIteration, childNodeId)
+				}
+			}
+		}
+		for _, nodeId := range nodesIdsToIterateNextIteration {
+			nodesQueue = append(nodesQueue, nodeId)
+		}
+	}
+
+	return nil
 }


idanovo:
Restored successfully repository with more than 10K commits in less than 40 secs (took a couple of hours before this change)

pkg/graveler/ref/manager.go
```diff
@@ -387,6 +395,31 @@ func (m *Manager) addCommit(tx db.Tx, repositoryID graveler.RepositoryID, commit
...
 
+func (m *Manager) updateCommitGeneration(tx db.Tx, repositoryID graveler.RepositoryID, commitIDToNode map[graveler.CommitID]CommitNode) error {
+	for len(commitIDToNode) != 0 {
+		command := "WITH updated(id, generation) AS (VALUES "
+		var updatingRows int
+		for commitID, commitNode := range commitIDToNode {
+			if updatingRows != 0 {
+				command += ","
+			}
+			command += fmt.Sprintf("('%s', %d)", commitID, commitNode.generation)
```
arielshaqed, line 406: This is SQL injection if `commitID` contains bad characters.  Rather than trust the file, let's quote the string!  Typically the golang DB packages have ways of quoting a literal string.

pkg/graveler/ref/manager.go
```diff
@@ -412,19 +445,72 @@ func (m *Manager) ListCommits(ctx context.Context, repositoryID graveler.Reposit
...
+	}
+	defer iter.Close()
+
+	commitIDToNode := make(map[graveler.CommitID]CommitNode)
+	var rootCommitID graveler.CommitID
+
+	// make a commits "tree" from map data structure
+	for iter.Next() {
+		commit := iter.Value()
+		if len(commit.Parents) == 0 {
```
idanovo, line 460: can we possibly have more than one commit without any parents? @arielshaqed @talSofer

pkg/graveler/ref/manager.go
```diff
@@ -412,19 +445,72 @@ func (m *Manager) ListCommits(ctx context.Context, repositoryID graveler.Reposit
...
+	currentGeneration := 1
+	// update commitNodes' generation in commitIDToNode "tree" using BFS algorithm.
+	// using a queue implementation
+	// adding a node to the queue only after all of its parents were visited in order to avoid redundant visits of nodes
+	for ; len(nodesQueue) > 0; currentGeneration++ {
+		var nodesIDsToIterateNextIteration []graveler.CommitID
+		for len(nodesQueue) > 0 {
+			currentCommitID := nodesQueue[0]
+			currentNode := commitIDToNode[currentCommitID]
+			commitIDToNode[currentCommitID] = CommitNode{currentNode.childCommitsIDs, 0, currentGeneration}
```
arielshaqed, line 499: This copies elements back into the node.  Can't we just say
```suggestion
			commitIDToNode[currentCommitID].generation = currentGeneration
			# (don't really need the next line, but if you like...)
			commitIDToNode[currentCommitID].notVisitedParents = 0
```
?

Similarly, I would rather modify just a single field on l. 513 than replace the entire `struct`.

arielshaqed (changes requested):
Thanks!

Some comments about readability on the computing function.  That one might also benefit from _shorter_ variable and field names.

Required change is to protect against malicious commit IDs.  Please try to use some Postgres library support rather than writing our own, if at all possible.  If not possible, please split off the literal quoter into its own function, and provide a reference that shows that it is correct.  (Thanks to `WITH RECURSIVE`, SQL is a programming language -- and now I can hate it like I do all other programming languages!)

pkg/graveler/ref/manager.go
```diff
@@ -412,19 +445,72 @@ func (m *Manager) ListCommits(ctx context.Context, repositoryID graveler.Reposit
 }
 
 func (m *Manager) FillGenerations(ctx context.Context, repositoryID graveler.RepositoryID) error {
```
Jonathan-Rosenberg, line 447: I would extract the code in this function into 2-3 functions for readability and possibly modularity (if some parts will be needed for different use cases).. 
For example a function that builds the tree and a function that executes the BFS is quite reasonable I think

pkg/graveler/ref/manager.go
```diff
@@ -387,6 +395,31 @@ func (m *Manager) addCommit(tx db.Tx, repositoryID graveler.RepositoryID, commit
...
+	for len(commitIDToNode) != 0 {
+		command := "WITH updated(id, generation) AS (VALUES "
+		var updatingRows int
+		for commitID, commitNode := range commitIDToNode {
+			if updatingRows != 0 {
+				command += ","
+			}
+			command += fmt.Sprintf("('%s', %d)", commitID, commitNode.generation)
+
+			delete(commitIDToNode, commitID)
```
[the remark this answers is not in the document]
idanovo: I need this since I'm running the SQL updates in batches so when I iterate again over the dictionary, I want to avoid iterating nodes I've already visited.

pkg/graveler/ref/manager.go
idanovo: Which string do you think we should quote? the `commitID` field is already quoted. What am I missing? 😅

idanovo: Which string do you mean we should quote? I'm already quoting the `commitID`. Am I'm missing something? 😅

pkg/graveler/ref/manager.go
```diff
@@ -22,12 +22,20 @@ const IteratorPrefetchSize = 1000
...
+
 type Manager struct {
 	db              db.Database
 	addressProvider ident.AddressProvider
 	batchExecutor   batch.Batcher
 }
 
+type CommitNode struct {
+	childCommitsIDs   []graveler.CommitID
+	notVisitedParents int
```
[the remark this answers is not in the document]
idanovo: @johnnyaug Hey, sorry. Changed my code, please look again 🙏

johnnyaug (changes requested):
The algorithm looks great.
1. Please move this to another struct and file.
2. Names should be simpler and shorter.
3. You can update fields on structs that you get from maps like this:
   ```
    s := myMap[key]
    s.field = 4
    ```

pkg/graveler/ref/manager.go
```diff
@@ -412,19 +445,89 @@ func (m *Manager) ListCommits(ctx context.Context, repositoryID graveler.Reposit
...
+	commitIDToNode := make(map[graveler.CommitID]CommitNode)
+
+	for iter.Next() {
+		commit := iter.Value()
+		notVisitedParents := map[graveler.CommitID]struct{}{}
+		for _, parentID := range commit.Parents {
+			notVisitedParents[parentID] = struct{}{}
+		}
+
+		commitIDToNode[commit.CommitID] = CommitNode{childCommitsIDs: []graveler.CommitID{}, notVisitedParents: notVisitedParents}
```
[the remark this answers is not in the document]
idanovo: It won't change the dictionary itself, am I wrong?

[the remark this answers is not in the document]
johnnyaug: It will change the node in the dictionary - which is what you want

johnnyaug (approved):
Nice work!

pkg/graveler/ref/manager.go
```diff
@@ -412,19 +445,78 @@ func (m *Manager) ListCommits(ctx context.Context, repositoryID graveler.Reposit
...
+func (m *Manager) mapCommitNodesToChildren(nodes map[graveler.CommitID]*CommitNode) {
+	for commitID, commitNode := range nodes {
+		// adding current node as a child to all parents in commitNode.parentsToVisit
+		for parentID := range commitNode.parentsToVisit {
+			nodes[parentID].children = append(nodes[parentID].children, commitID)
+		}
+	}
+}
+
+func (m *Manager) addGenerationToNodes(nodes map[graveler.CommitID]*CommitNode, rootCommits []graveler.CommitID) {
```
arielshaqed, line 506: I think this function should be tested.  Could open a TODO/issue to do this rather than delay this PR.

arielshaqed (changes requested):
Thanks!

Still blocked on unsafe SQL quoting, sorry.

pkg/graveler/graveler.go
```diff
@@ -1880,6 +1880,7 @@ func (g *Graveler) LoadCommits(ctx context.Context, repositoryID RepositoryID, m
 	if iter.Err() != nil {
 		return iter.Err()
 	}
+	// after removing this code, we won't be able to support dump that created using lakFS version that's lower than v0.61.0
```
arielshaqed, line 1883: Neat comment!  May I suggest:
1. Open an issue to describe what you intend to do (later).  Maybe even write a suggested date for it, and/or set a reminder on your calendar to do it in 9 months or whatever you decide.
1. Reference that issue in your comment.  The way I like to do it is to write something like
   ```go
   // TODO(#27183): Remove this code to drop support for dumps created by lakeFS versions below v0.61.0.
   ```
   (assuming the issue you opened got number 27183).

pkg/graveler/ref/manager.go
```diff
@@ -397,21 +398,21 @@ func (m *Manager) addCommit(tx db.Tx, repositoryID graveler.RepositoryID, commit
...
+			command += fmt.Sprintf(`(%s, %d)`, pq.QuoteLiteral(string(commitID)), node.generation)
 
 			delete(nodes, commitID)
 			updatingRows += 1
 			if updatingRows == BatchUpdateSQLSize {
 				break
 			}
 		}
-		command += fmt.Sprintf(") UPDATE graveler_commits SET generation = updated.generation FROM updated WHERE (graveler_commits.id=updated.id AND graveler_commits.repository_id='%s');", repositoryID)
+		command += fmt.Sprintf(`) UPDATE graveler_commits SET generation = updated.generation FROM updated WHERE (graveler_commits.id=updated.id AND graveler_commits.repository_id='%s');`, repositoryID)
```
arielshaqed, line 415: `"%s"` for `repository_id` is probably another SQL injection.  Luckily here we can just use standard SQL statement placeholders.  Look for "$1" just about anywhere for how to do that: it is the best and absolutely safest and simplest way to introduce strings into SQL statements.

arielshaqed (commented):
Thanks!

Looks very nice.  One SQL injection vector remains and blocks approval.

I would also be _happy_ if you got rid of the map in favour of a reference count.  Right now I do not think it should block us; however if people do a commit every 10s then it may become significant.  In any case we can do it on a separate PR.

Similarly, I'd really like for you to test the generations computation.  That too does not need to happen in this PR.

pkg/graveler/ref/manager.go
```diff
@@ -412,19 +445,78 @@ func (m *Manager) ListCommits(ctx context.Context, repositoryID graveler.Reposit
...
+			for _, childNodeID := range currentNode.children {
+				delete(nodes[childNodeID].parentsToVisit, nodeCommitID)
+				if len(nodes[childNodeID].parentsToVisit) == 0 {
+					nextIterationNodes = append(nextIterationNodes, childNodeID)
+				}
+			}
+		}
+		nodesCommitsIDs = nextIterationNodes
+	}
+}
```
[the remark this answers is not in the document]
idanovo: > Thanks!
> 
> Looks very nice. One SQL injection vector remains and blocks approval.
> 
> I would also be _happy_ if you got rid of the map in favour of a reference count. Right now I do not think it should block us; however if people do a commit every 10s then it may become significant. In any case we can do it on a separate PR.
> 
> Similarly, I'd really like for you to test the generations computation. That too does not need to happen in this PR.

@arielshaqed about the change of `parentsToVisit` to be counter, I'm using this map in `mapCommitNodesToChildren` so I can't really change it to be a counter

pkg/graveler/ref/manager.go
idanovo: Thanks!

Looks very nice. One SQL injection vector remains and blocks approval.

I would also be happy if you got rid of the map in favour of a reference count. Right now I do not think it should block us; however if people do a commit every 10s then it may become significant. In any case we can do it on a separate PR.

Similarly, I'd really like for you to test the generations computation. That too does not need to happen in this PR.

@arielshaqed about the change of `parentsToVisit` to be counter- I'm using this map in mapCommitNodesToChildren so I can't really change it to be a counter
About the Generation calculation- we already have a test for it called `TestManager_FillGenerations` 😃 .

Forgot to answer those comments, sorry.

arielshaqed (approved):
Thanks!  Sorry for long number of rounds; I think this PR is really good.

A note about `mapCommitNodesToChildren`: That function is part of setup.  You could avoid the parents map simply by constructing the graph directly, rather than nodes first followed by edges.  Seen this way, you build a node the first time that you need it (so usually as a parent of some node), then add its information in when your scan finally reaches it.  This does not block this PR.  I anticipate that it will cause the next performance issue here, once people have many millions of commits (think of a user who streams data in and commits every 10 seconds, for 8_640 commits every day).  So we should have at least a year :-)

idanovo:
> Thanks! Sorry for long number of rounds; I think this PR is really good.
> 
> A note about `mapCommitNodesToChildren`: That function is part of setup. You could avoid the parents map simply by constructing the graph directly, rather than nodes first followed by edges. Seen this way, you build a node the first time that you need it (so usually as a parent of some node), then add its information in when your scan finally reaches it. This does not block this PR. I anticipate that it will cause the next performance issue here, once people have many millions of commits (think of a user who streams data in and commits every 10 seconds, for 8_640 commits every day). So we should have at least a year :-)

Thanks!

About `mapCommitNodesToChildren`- you are right and this is the way I've implemented it at first, but I preferred to split the methods the way it looks now for readability. I guess there's more we can do here to improve this logic, but I really hope we will remove all of it very soon!

Issue #3002, opened 2022-03-03
lakectl doctor command improvements 


talSofer:
- Add documentation to lakectl doctor - what is it checking? add a verbose mode to the command so that it prints what it is checking.
- Instead of calling out the “endpoint not ending with api/v1”, let’s add it from our code. the user should not add it.

PR #3023, opened 2022-03-09
LakeCTL doctor command improvements. Part of #3002

idanovo:


cmd/lakectl/cmd/doctor.go
```diff
@@ -61,56 +61,82 @@ var errorTemplate = `{{ .Message |red }}
...
 	Run: func(cmd *cobra.Command, args []string) {
-		err := ListRepositoriesAndAnalyze(cmd.Context())
+		verbose, _ := cmd.Flags().GetBool("verbose")
+		err := ListRepositoriesAndAnalyze(cmd.Context(), verbose)
 		if err == nil {
 			Write(successMessageTemplate, &UserMessage{"Valid configuration"})
 			return
 		}
 
+		if verbose {
```
Jonathan-Rosenberg, line 80: A thought: instead of using the 
```
if verbose {
...
}
```
clause, you can create a:
```
func WriteVerbose(tpl string, data interface{}, verbose bool) {
    if verbose {
        Write(tpl, data, os.Stdout)
    }
}
```
function that will print only if the verbose flag was given and use it instead.
I think it would make the code a bit cleaner.

cmd/lakectl/cmd/doctor.go
```diff
@@ -61,56 +61,82 @@ var errorTemplate = `{{ .Message |red }}
...
+			Write(analyzingMessageTemplate, &UserMessage{"Got error while trying to run a simple command.\nTrying to analyze error."})
```
Jonathan-Rosenberg, line 81:
```suggestion
			Write(analyzingMessageTemplate, &UserMessage{"Got error while trying to run a sanity command.\nTrying to analyze error."})
```
?

cmd/lakectl/cmd/doctor.go
```diff
@@ -61,56 +61,82 @@ var errorTemplate = `{{ .Message |red }}
...
+	}
 	// getClient might die on url.Parse error, so check it first.
 	serverEndpoint := cfg.Values.Server.EndpointURL
 	_, err := url.Parse(serverEndpoint)
 	if err != nil {
 		return &WrongEndpointURIError{msgOnErrWrongEndpointURI, err.Error()}
 	}
 	client := getClient()
+	if verboseMode {
+		Write(analyzingMessageTemplate, &UserMessage{"Trying to run a simple command using current configuration."})
```
Jonathan-Rosenberg, line 138:
```suggestion
		Write(analyzingMessageTemplate, &UserMessage{"Trying to run a sanity command using current configuration."})
```
if also accepted above..

Jonathan-Rosenberg (approved):
Very cool!

N-o-Z:
While modifying the doctor.go file you might want to fix the typo and reformat the message in line 149
`	return &UnknownConfigError{msgOnErrUnknownConfig, "An unknown error accourd while trying to analyzing LakeCtl configuration."}
`
*_An unknown error occurred while trying to analyze LakeCtl configuration._

N-o-Z:
General suggestion - consider enabling coloring for verbose messages

cmd/lakectl/cmd/doctor.go
```diff
@@ -132,10 +146,11 @@ func ListRepositoriesAndAnalyze(ctx context.Context) error {
 	case resp.JSONDefault != nil:
 		return &UnknownConfigError{msgOnErrUnknownConfig, resp.JSONDefault.Message}
 	}
-	return &UnknownConfigError{msgOnErrUnknownConfig, "An unknown error accourd while trying to analyzing LakeCtl configuration."}
+	return &UnknownConfigError{msgOnErrUnknownConfig, "An unknown error accord while trying to analyze LakeCtl configuration."}
```
idanovo, line 149:
```suggestion
	return &UnknownConfigError{msgOnErrUnknownConfig, "An unknown error occurred while trying to analyze LakeCtl configuration."}
```

cmd/lakectl/cmd/root.go
```diff
@@ -105,7 +105,7 @@ func getClient() *api.ClientWithResponses {
 		DieErr(err)
 	}
 	// if no uri to api is set in configuration - set the default
-	if u.Path == "" || u.Path == "/" {
+	if u.Path == "" || u.Path == "/" || !strings.HasSuffix(serverEndpoint, api.BaseURL) {
```
nopcoder, line 108: I think we should keep helping the user just in case they didn't specified the path.  If there is a specific path I would like to use it.

nopcoder (changes requested):

cmd/lakectl/cmd/doctor.go
```diff
@@ -61,56 +61,70 @@ var errorTemplate = `{{ .Message |red }}
...
 
+	WriteIfVerbose(analyzingMessageTemplate, &UserMessage{"Trying to get endpoint URL and parse it as an URL format."}, verboseMode)
 	// getClient might die on url.Parse error, so check it first.
 	serverEndpoint := cfg.Values.Server.EndpointURL
 	_, err := url.Parse(serverEndpoint)
 	if err != nil {
 		return &WrongEndpointURIError{msgOnErrWrongEndpointURI, err.Error()}
 	}
 	client := getClient()
+	WriteIfVerbose(analyzingMessageTemplate, &UserMessage{"Trying to run a sanity command using current configuration."}, verboseMode)
```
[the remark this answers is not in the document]
idanovo: I don't think It's necessary, It's more implementation details that might confuse the user. WDYT?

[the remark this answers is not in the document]
N-o-Z: I believe that providing the specific tested command and output enables the user to get a better understanding of what might be the issue in situations where the error message is not clear enough.
But maybe we can get an additional view about this suggestion?

nopcoder:
Just one question - why do we need the verbose mode? If we log extra messages just as part of the 'doctor' command, which we like to identity issues, I would always want to have verbose on to understand whats work or not.

idanovo:
> Just one question - why do we need the verbose mode? If we log extra messages just as part of the 'doctor' command, which we like to identity issues, I would always want to have verbose on to understand whats work or not.

@nopcoder the verbose mode is used just for printing what are the steps we are doing during the check, if we find anything wrong or suspicious we will print it to the user also if he's not using the verbose mode.

nopcoder (approved):

Issue #2930, opened 2022-02-08
hive-ms: merge meta data


chaimt:
after branch has been merged to master, if schema was changed we need a way to merge the changes from the branch hive-ms to master hive-ms

Jonathan-Rosenberg:
Hey @chaimt,
Thank you for opening this issue!
In order to further understand the issue it would be very helpful if you'll share with us what you would expect to get merged as a result of the merge process. 
If you can also share a diff between schemas, and a desired result that would be awesome!

Thanks!

Issue #2929, opened 2022-02-08
delete-branch-schema


chaimt:
There should be a command delete-branch-schema from hive-ms if i don't want to use the branch and have deleted it from lakefs

chaimt:
we will be running dbt commands. if the command fails we will delete the branch in lakefs. in this case we also need to delete all the tables that were create in hive

PR #3024, opened 2022-03-09
DBT delete and merge branch-schema proposal

Jonathan-Rosenberg:
A proposal to add dbt subcommands for branch and schema deletion and merging.
Relates to #2929 and #2930


idanovo:
Sound good to me 🙂 . The only thing I'm missing here is my lack of understanding about how the connection between commits and branches in terms of dbt schemas. e.g. Do we have a different schema for each commit? In case we have a different schema for each commit- when we delete the branch scheme, do we delete it for each commit that was created using this branch or only the one that the branch point to?

Jonathan-Rosenberg:
Schema and commits are orthogonal, i.e. commits are unaware of the schemas and schemas are unaware of commits (a schema in our context is a namespace or directory under a database).
When we say that we remove a schema, it means deleting it and all associated tables under it from the metastore (a DB that holds the structure and properties of various database objects such as tables, columns and schemas, of a storage location [like S3 bucket] that represent those objects).
All commits in a given dbt DAG are under the same schema. Also, there is only one schema for a given branch.

idanovo:
> Schema and commits are orthogonal, i.e. commits are unaware of the schemas and schemas are unaware of commits (a schema in our context is a namespace or directory under a database). When we say that we remove a schema, it means deleting it and all associated tables under it from the metastore (a DB that holds the structure and properties of various database objects such as tables, columns and schemas, of a storage location [like S3 bucket] that represent those objects). All commits in a given dbt DAG are under the same schema. Also, there is only one schema for a given branch.

@Jonathan-Rosenberg thanks for the explanation!
If this is the case, wouldn't it be better to have different commands `delete-branch` (which we have) and `delete-schema`?

Jonathan-Rosenberg:
@idanovo, so basically when dbt users want to delete a branch using the `lakectl dbt` command it's coupled with the deletion of the schema, because as I mentioned:
> Also, there is only one schema for a given branch.

so if we delete a branch and don't delete the schema, the schema will be left untouched pointing to a deleted branch location.
We can limit the command to delete only the schema and leave the branch as is, but I imagine, as also happened with the `create-branch-schema` command, that a feature request will be created to delete the associated branch because it has no use dbt-wise...

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,30 @@
...
+- merge them into master/main/production/other branch
+- both.
+
+`lakectl dbt delete-branch-schema` will allow the deletion of a branch and the schema associated with it, or just the schema.  
+`lakectl dbt merge-branch-schema` will allow the merging of a branch to a given branch  and the schema associated with it. 
+### Why do we need it?
+The users that use the dbt command are, well, dbt users. They will interact with the `lakectl` tool in the context of the dbt command (and some basic commands such as `commit` and `merge`).
+Adding the `delete-branch-schema` and `merge-branch-schema` subcommands will make it easier for them to integrate lakeFS into their framework.
+
+## `delete-branch-schema` Usage
```
talSofer, line 16: I think that the current usage may come out a bit confusing. The main purpose of the `delete-branch-schema` command should be to delete the **schema**. 
I see two use cases here:
1. I already deleted my dbt branch and I want to delete the schema created for that branch. 
2. I want to delete a dbt schema and its corresponding branch together. 

To support 1, `lakectl dbt delete-branch-schema --schema <schema name>` you suggested is sufficient. If we want to support 2, I suggest different wording that speaks from dbt's point of view and is consistent with the `--create-branch` flag of `create-branch-schema`. 
It can be something like something like 
`lakectl dbt delete-branch-schema  --schema <schema name> --delete-branch <branch name>`

TBH, I'm not sure that we even want to support the second use case. Let me know your thoughts and we can discuss more :)

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,30 @@
...
+3. `lakectl dbt delete-branch-schema --schema <schema name>`: Delete the given schema.
+4. `lakectl dbt delete-branch-schema`: Failure. Must specify at least `--branch` or `--schema`
+
+## `merge-branch-schema` Usage
+4. `lakectl dbt merge-branch-schema <source branch name> <destination branch name>`: Merge the source branch to the destination branch and use the default schemas of the source and destination (branches names).
+2. `lakectl dbt merge-branch-schema <source branch name> <destination branch name> --source-schema <source schema name>`: Merge source branch to destination branch in the repository that the dbt project's target schema resides at. Merge the custom source schema to the default destination schema (branch name).2. `lakectl dbt merge-branch-schema <source branch name> <destination branch name> --source-schema <source schema name>`: Merge source branch to destination branch in the repository that the dbt project's target schema resides at. Merge the custom source schema to the default destination schema (branch name).
```
talSofer, line 24: you probably copy-pasted this bullet again

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,30 @@
...
+3. `lakectl dbt merge-branch-schema <source branch name> <destination branch name> --destination-schema <destination schema name>`: Merge source branch to destination branch in the repository that the dbt project's target schema resides at. Merge the default source schema (branch name) to a custom destination schema.
+1. `lakectl dbt merge-branch-schema <source branch name> <destination branch name> --source-schema <source schema name> --destination-schema <destination schema name>`: Merge source branch to destination branch in the repository that the dbt project's target schema resides at. Merge the source schema to the destination schema.
+5. `lakectl dbt merge-branch-schema <source branch name> <destination branch name> --target-schema`:  Merge the source branch to the destination branch and use the default source schema (branch name) and the dbt target schema specified in the dbt project.
+6. `lakectl dbt merge-branch-schema --branch <branch name> --schema <schema name>`: Delete branch and schema with the given names.
```
talSofer, line 28: Did you mean to suggest a delete operation here?

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,30 @@
...
+The users that use the dbt command are, well, dbt users. They will interact with the `lakectl` tool in the context of the dbt command (and some basic commands such as `commit` and `merge`).
+Adding the `delete-branch-schema` and `merge-branch-schema` subcommands will make it easier for them to integrate lakeFS into their framework.
+
+## `delete-branch-schema` Usage
+1. `lakectl dbt delete-branch-schema --branch <branch name> --schema <schema name>`: Delete branch and schema with the given names.
+2. `lakectl dbt delete-branch-schema --branch <branch name>`: Delete the branch from the repository that the dbt project's target schema resides at. The command will also try to delete a schema with the same name as the branch (when creating a branch using the `lakectl dbt create-branch-schema` command, the default schema name is the branch name).
+3. `lakectl dbt delete-branch-schema --schema <schema name>`: Delete the given schema.
+4. `lakectl dbt delete-branch-schema`: Failure. Must specify at least `--branch` or `--schema`
+
+## `merge-branch-schema` Usage
```
talSofer, line 22: I feel that the combination of branch and schema names creates complexity here. let's try to brainstorm and think about how we can simplify. 
I'm not sure that we need more than the following command at this point
`lakectl dbt merge-branch-schema <source schema name> <destination schema name>`: Merge the source schema to the destination schema.

The reason why I think that this is a good starting point is that we have limited amount of feedback and input from users, and this command supports the request to [merge metadata] (https://github.com/treeverse/lakeFS/issues/2930)

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,30 @@
...
+- both.
+
+`lakectl dbt delete-branch-schema` will allow the deletion of a branch and the schema associated with it, or just the schema.  
+`lakectl dbt merge-branch-schema` will allow the merging of a branch to a given branch  and the schema associated with it. 
+### Why do we need it?
+The users that use the dbt command are, well, dbt users. They will interact with the `lakectl` tool in the context of the dbt command (and some basic commands such as `commit` and `merge`).
+Adding the `delete-branch-schema` and `merge-branch-schema` subcommands will make it easier for them to integrate lakeFS into their framework.
+
+## `delete-branch-schema` Usage
+1. `lakectl dbt delete-branch-schema --branch <branch name> --schema <schema name>`: Delete branch and schema with the given names.
```
talSofer, line 17: a general question - 
Is there a way for users to get the schema associated with a branch via lakectl? I mean, as a user, how do I know the name of the schema I'm messing with?

talSofer (changes requested):
Thanks @Jonathan-Rosenberg for getting into the guts of dbt and for this proposal!
Your work brings up MANY lakefs-dbt usability issues that we should think of and improve, thanks for that. 

There is a lot to discuss here, I added my thoughts.

design/proposals/dbt-delete-merge-schema-branch.md
Jonathan-Rosenberg: I agree that the primary usage of this command is to delete the schema and not the branch, but when I think of the users that will use this command, these are analysts that probably want to use mainly the `lakefs dbt` command and not the other ones... 
We could definitely create a `delete-schema` command and use it as a subcommand for the `lakectl dbt` or `lakectl metastore` commands, but then we might drift apart from the users that probably require the whole set of actions under the `lakectl dbt` command.
We can start by implementing the schema deletion part and stay tuned for branch deletion capability request (as happened with the create-branch-schema command) if shall ever come.
WDYT?

design/proposals/dbt-delete-merge-schema-branch.md
Jonathan-Rosenberg: It's a bit of a mess, I'll fix that :sweat_smile:

design/proposals/dbt-delete-merge-schema-branch.md
Jonathan-Rosenberg: I agree that we can start with that (also the delete part can start only with a schema delete), but then again, I do believe that a request will arrive (or questions will get asked) regarding branch merge in the context of this command (or branch delete in the context of the delete command), as it's a convenient method for dbt users.
But I haven't completed my prophecy course on Udemy yet so stay tuned :wink:

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,45 @@
...
+Adding the `delete-branch-schema` and `merge-branch-schema` subcommands will make it easier for them to integrate lakeFS into their framework.
+
+### Goals 
+- Delete schema, tables, partitions and associated metadata from the metastore using a single command.
+- Merge schema, tables, partitions and associated metadata from the metastore using a single command.
+
+### Non-goals
+
+- Deleting a branch before/after deleting an associated schema.
+- Merging a branch before/after deleting an associated schema.
```
talSofer, line 24:
```suggestion
- couple branch deletion and schema deletion.
- couple lakectl merge and schema merge operations.
```

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Suggestions
+
+- `delete-branch-schema` will allow deleting a schema based on a branch name.
+- `merge-branch-schema` will allow merging a schema and all dbt model underneath it to a destination schema by branch names.
```
talSofer, line 29:
```suggestion
- `merge-branch-schema` will allow merging a schema and all dbt models underneath it to a destination schema by branch names.
```

What does this mean to merge the dbt models under a schema? isn't a dbt model a piece of code?

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,45 @@
...
+
+### `delete-branch-schema` Usage
+
+`lakectl dbt delete-branch-schema --branch <branch name>`: Delete the schema associated with the input branch name, e.g. `lakectl dbt delete-branch-schema --branch dbt_jon` will delete the schema associated with (mapped to) the `dbt_jon` branch.
+
+### `merge-branch-schema` Usage
+
+`lakectl dbt merge-branch-schema --source-branch <source branch name> --destination-branch <destination branch name>`: Merge the source schema (associated with the source branch name) to the destination schema (associated with the destination branch name), e.g. `lakectl dbt merge-branch-schema --source-branch dbt_jon --destination-branch master` will merge the schema and models associated with `dbt_jon` branch to `main` branch.
```
talSofer, line 37:
```suggestion
`lakectl dbt merge-branch-schema --source-branch <source branch name> --destination-branch <destination branch name>`: Merge the source schema (associated with the source branch name) to the destination schema (associated with the destination branch name), e.g. `lakectl dbt merge-branch-schema --source-branch dbt_jon --destination-branch master` will merge the schema and models associated with `dbt_jon` branch to `master` branch.
```

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Required change
+
+When creating a new schema for a branch using the `lakectl dbt create-branch-schema` command, there will be added a mapping of that branch to that schema in a file (probably under the dbt project directory).  
+This is necessary because the branch and the schema might have different names (when using the `--to-schema` flag) and not use the default behavior (which is branch and schema having the same name).
+
+### Suggestion
```
talSofer, line 44: Consider including this part in the Required changes section

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,45 @@
...
+When the users start working with lakectl and dbt together they'll also need to map the main/master/production branch they are working with to a schema. If not, the target schema will get mapped to it by default. We can do that by adding a `lakectl dbt map-branch-schema` command.
```
talSofer: Can you please explain the problem your suggestion solves? (I have the context to understand this because we talked) but I think it is required by readers. 

and a question, in case the user didn't map their main branch to a schema, how would we know what is the default branch to create mapping for?

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,45 @@
...
+### Goals 
+- Delete schema, tables, partitions and associated metadata from the metastore using a single command.
+- Merge schema, tables, partitions and associated metadata from the metastore using a single command.
+
+### Non-goals
+
+- Deleting a branch before/after deleting an associated schema.
+- Merging a branch before/after deleting an associated schema.
+
+## Suggestions
```
talSofer, line 26:
```suggestion
## Suggested solution
```

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,45 @@
...
+
+### `delete-branch-schema` Usage
+
+`lakectl dbt delete-branch-schema --branch <branch name>`: Delete the schema associated with the input branch name, e.g. `lakectl dbt delete-branch-schema --branch dbt_jon` will delete the schema associated with (mapped to) the `dbt_jon` branch.
+
+### `merge-branch-schema` Usage
+
+`lakectl dbt merge-branch-schema --source-branch <source branch name> --destination-branch <destination branch name>`: Merge the source schema (associated with the source branch name) to the destination schema (associated with the destination branch name), e.g. `lakectl dbt merge-branch-schema --source-branch dbt_jon --destination-branch master` will merge the schema and models associated with `dbt_jon` branch to `main` branch.
+
+## Required change
```
talSofer, line 39:
```suggestion
## Required changes
```

talSofer (commented):
Thanks @Jonathan-Rosenberg, the suggested change makes sense and will simplify our current integration with dbt. 
Added a couple of questions and comments, let me know your thouhgts

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,46 @@
...
+
+When creating a new schema for a branch using the `lakectl dbt create-branch-schema` command, there will be added a mapping of that branch to that schema in a file (probably under the dbt project directory).  
+This is necessary because the branch and the schema might have different names (when using the `--to-schema` flag) and not use the default behavior (which is branch and schema having the same name).
```
arielshaqed, line 42: I don't understand how this should work.

E.g. say I have branch `main` with DBT schema `main-schema`. If I branch out to branch `staging`, we write the DBT schema name `staging-schema` to the naming the schema in the DBT project directory.  Now I merge back from `staging` to `main`.  Won't that change the file, thus modifying branch `main` to use schema `staging-schema`?

arielshaqed, line 42: Here is one proposal.  The schema to use for a branch sounds like a property _of the branch_.  We are not going to be able to support it inside the existing data model (we could use pre-merge hooks, but those would have to perform careful and drastic rewriting of objects, which is scary, hard, and can fail with difficult errors).

So let's change lakeFS to add user metadata to branches, just like we already have for objects and commits!  We are talking about storing a mapping from a branch to a value.  We already have branches as first-class objects, and user metadata on other objects.  This is a small step, and we have a first use-case for it.

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,46 @@
...
+- `delete-branch-schema` will allow deleting a schema based on a branch name.
+- `merge-branch-schema` will allow merging a schema and all dbt model underneath it to a destination schema by branch names.
+
+### `delete-branch-schema` Usage
+
+`lakectl dbt delete-branch-schema --branch <branch name>`: Delete the schema associated with the input branch name, e.g. `lakectl dbt delete-branch-schema --branch dbt_jon` will delete the schema associated with (mapped to) the `dbt_jon` branch.
+
+### `merge-branch-schema` Usage
+
+`lakectl dbt merge-branch-schema --source-branch <source branch name> --destination-branch <destination branch name>`: Merge the source schema (associated with the source branch name) to the destination schema (associated with the destination branch name), e.g. `lakectl dbt merge-branch-schema --source-branch dbt_jon --destination-branch master` will merge the schema and models associated with `dbt_jon` branch to `main` branch.
```
arielshaqed, line 37: I worry about lack of atomicity, and as part of that about what to do if merging succeeds but schema update fails.

I don't think this is a hypothetical, I believe it can (and should) happen when 2 branches have incompatible changes.  For instance, suppose you add on your branch an optional field `timestamp` with a string format and I add on my branch an optional field `timestamp` with a number format (and each of us adds a few files that use this field, of course).  Then the second merge to `main` should fail, but AFAICS there are no lakeFS merge conflict.  But there will be a DBT merge conflict, no?

arielshaqed: One way to resolve might be to order these operations specifically.  However we must also decide and understand how to handle read operations occurring during a merge -- especially given that a nonatomic merge can involve humans and thus take a _long_ time.

arielshaqed (changes requested):
Great proposal -- important part of our DLM vision and much-anticipated by users!

I don't fully understand the proposed semantics of these operations.  We should either document them (if we have useful semantics) or else figure them out.

design/proposals/dbt-delete-merge-schema-branch.md
Jonathan-Rosenberg: Yes you're correct. I used it interchangeably with the materialization of the model instead (a table or a view).
I'll change that.

design/proposals/dbt-delete-merge-schema-branch.md
Jonathan-Rosenberg: I'll try to clarify:
The mapping is a `branch name` -> `schema name` mapping.
So for the given example we start of with the following mapping: `main` -> `main-schema`.
When you branch out to the `staging` branch and create a `staging-schema` schema (associated with the `staging` branch) the following mapping will be created in the map:
`staging` -> `staging-schema`.
Now when you'll merge back to `main`, the process will start of figuring which schema is associated with `main` in order to merge `staging-schema` into it. It will find that `main-schema` is associated with the `main` branch, and will conduct a merge between `staging-schema` and `main-schema`, resulting with a modified `main-schema`.
The outcome will be `main` using a merged `main-schema`.

Hopes that makes sense...

Jonathan-Rosenberg: This sounds really cool!

design/proposals/dbt-delete-merge-schema-branch.md
Jonathan-Rosenberg: What do you mean by ordering these operations?

design/proposals/dbt-delete-merge-schema-branch.md
Jonathan-Rosenberg: It's under that section (notice the heading cascade)

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,66 @@
...
+* Example: `lakectl dbt merge-branch-schema --source-branch dbt_jon --destination-branch master` will merge the schema associated with the `dbt_jon` branch to the schema associated with the `master` branch.
+
+## Required changes
+
+In order to implement the new commands, there's a need to map branches to schemas.
+This is necessary because the branch and the schema might have different names (when adding the `--to-schema <custom name>` flag when creating the schema using the `lakectl dbt create-branch-schema` command) and not use the default behavior (which is the same name for the branch and schema).
+
+### Suggested implementation
+
+The mapping of branches to schemas will be facilitated under the branch's (or more accurately, the commit that this branch is pointing to) metadata property.
```
johnnyaug, line 47: How does this work?
When I run `lakectl dbt create-branch-schema --create-branch` - you will create a branch for me, and it will point to the same commit as my main branch. 

After I commit to my new branch, who will edit the metadata to point to the new schema?

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,66 @@
...
+* Example: `lakectl dbt delete-branch-schema --branch dbt_jon` will delete the schema associated with the `dbt_jon` branch.
+
+### `merge-branch-schema` Usage
+
+`lakectl dbt merge-branch-schema --source-branch <source branch name> --destination-branch <destination branch name>`: Merge the source schema (associated with the source branch name) to the destination schema (associated with the destination branch name)
+* Example: `lakectl dbt merge-branch-schema --source-branch dbt_jon --destination-branch master` will merge the schema associated with the `dbt_jon` branch to the schema associated with the `master` branch.
+
+## Required changes
+
+In order to implement the new commands, there's a need to map branches to schemas.
```
johnnyaug, line 42: Suppose that we have this mapping, how does the merge work?
Is it something that we already have? Is it the same as `lakectl metastore copy`?

Jonathan-Rosenberg: Yes it's exactly that

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,66 @@
...
+This is necessary because the branch and the schema might have different names (when adding the `--to-schema <custom name>` flag when creating the schema using the `lakectl dbt create-branch-schema` command) and not use the default behavior (which is the same name for the branch and schema).
```
arielshaqed, line 43: Can we get rid of this?  (E.g. clarify requirements by asking users.)

The fundamental issue here is that we have 2 different objects (lakeFS branch, DBT schema) which must be coordinated but are separately managed.  To be very clear: this is not caused in any way by this proposal, or indeed by the entire `lakectl dbt` feature -- it is just the way things are.  At the very least these objects must have a 1:1 mapping!  So any feature which allows this 1:1 property to fail can have really unpleasant consequences.

One possible nasty consequence:
1. Associate branch `main` with `working-schema`.
2. Work on `main` and on the schema.
3. Branch out of `main` to `ariels-dev`, then map it separately to `working-schema` (this is a mistake).
4. Delete branch `ariels-dev` because I made a mistake.
   * `lakectl` deletes `working-schema`.

This is pretty much my fault, but it would be entirely avoidable if schema names were identical to branch names.  (It is quite hard to avoid this programmatically by checking for re-use, because that assumes we know of all accesses to schema file contents.  E.g. if I restore some DBT files from backup, any checks made at time of creation are no longer valid.)

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,66 @@
...
+
+### Suggested implementation
+
+The mapping of branches to schemas will be facilitated under the branch's (or more accurately, the commit that this branch is pointing to) metadata property.
+The `schema` key will be used to point to the name of the schema that is configured with the URI of the containing branch: `schema -> <schema name>`.
```
arielshaqed, line 48: I no longer support this.  It makes sense only in the presence of a single DBT instance.  But if you have your own DBT instance and I have my own, "obviously" each of us has to keep their own version of the mapping.  So the mapping cannot live on lakeFS, unless we have an additional key (prefix) to separate keys `jon-schema` from `ariels-schema`.

Let's go back to saving this data on the DBT side.  *That* will require an entire subdirectory that we control, holding a `branch -> schema` mapping.  (Ideally _that_ would use a separate file for each branch, to avoid races on a single file when multiple `lakectl` instances run concurrently!)

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,66 @@
...
+* The complete process of merge, i.e. both branch and schema merging, is lacking atomicity. That means that the merge can be partially successful. For example, it's possible that the branch merge was successful but the schema merge was not. This will be resolved in future work.
+
+## Future improvements
+
+### Switch used schema by branch
+
+Make it easier for users to switch the schema they are running their dbt commands over.
+Currently, users need to manually change the schema they want to run their dbt commands over in the profile target's schema property. This means they need to be aware of changes they make to the dbt profile file and not check-in those changes to their version control system later. It also means they need to book the different branch to schema mapping.
+- `lakectl dbt use-schema --branch <branch name>`: Use the provided branch's mapped schema within the terminal that the user uses to run dbt commands.
+
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: Yes of course.
Once the proposal will be approved I'll create issues for anything that was proposed.

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,66 @@
...
+* The complete process of merge, i.e. both branch and schema merging, is lacking atomicity. That means that the merge can be partially successful. For example, it's possible that the branch merge was successful but the schema merge was not. This will be resolved in future work.
+
+## Future improvements
+
+### Switch used schema by branch
+
+Make it easier for users to switch the schema they are running their dbt commands over.
+Currently, users need to manually change the schema they want to run their dbt commands over in the profile target's schema property. This means they need to be aware of changes they make to the dbt profile file and not check-in those changes to their version control system later. It also means they need to book the different branch to schema mapping.
+- `lakectl dbt use-schema --branch <branch name>`: Use the provided branch's mapped schema within the terminal that the user uses to run dbt commands.
```
arielshaqed, line 65: I don't think this syntax works in UN*X shells.  The phrasing hints that the command changes an environment variable, but subprocesses cannot change environment variables in the parent.  The best you could do would be e.g.
```sh
`lakectl dbt use-schema --branch <branch_name>`
```
to run the command and interpolate its output into the same CLI used to call it.

design/proposals/dbt-delete-merge-schema-branch.md
```diff
@@ -0,0 +1,66 @@
...
+The `schema` key will be used to point to the name of the schema that is configured with the URI of the containing branch: `schema -> <schema name>`.  
+When users start working with lakectl and dbt together they'll also need to map the main/master/production branch they are working with to a schema, otherwise they will be blocked from working with `lakectl dbt` commands.
+This is necessary because, as mentioned in the [Background section](#background), when users first initialize a schema it points to the branch's location, but they didn't take any action to **map the branch to the schema**.
+It might cause a problem if, for example, a user branches out of the main branch (the branch that the target schema's location is pointing to), commits changes with the newly created schema, merges the branches and then wants to merge the schemas. The user wouldn't be able to merge the schemas because the main branch is not aware of the schema associated with it.  
+We can allow users to map a branch to a schema by providing a `lakectl dbt map-branch-schema <branch> <schema>` command.
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: Yes, I'll write it.

arielshaqed (approved):
Thanks!

Need to be careful about how to store the `branch -> schema` mapping -- ideally we can get rid of the need to use it!

design/proposals/dbt-delete-merge-schema-branch.md
Jonathan-Rosenberg: So the use of it would be something like:
`export LAKEFS_SCHEMA=$(lakectl dbt use-schema --branch <branch_name>)` 
?

design/proposals/dbt-delete-merge-schema-branch.md
Jonathan-Rosenberg: I agree that this flag might have a very limited use and therefore can be omitted, but even if so, it leaves us with the case where the user wants to merge back to the main branch (or whatever branch that was associated with the dbt target schema initially). We have no record of which schema is associated with that branch. We can impose a requirement to name the schema as the branch being used, but it sounds a bit too restrictive to me.

PR #3025, opened 2022-03-09
Fix error capturing and formatting in DB operation failures

### LINKED ISSUE #2974: Change output on user/group creation with exists user id on lakectl
When trying to create a user (`lakectl auth users create`) with a user id that's already exists we get "query INSERT INTO auth_users (display_name, created_at, friendly_name, source) VALUES ($1, $2, $3, $4) RETURNING id: scany: rows final error: ERROR: duplicate key value violates unique constraint "auth_users_unique_display_name" (SQLSTATE 23505)"
We want to change the output in this case to more user-friendly text.

N-o-Z:
- Created a helper function in db errors to capture, log and transform sql error to a readable format.
- As part of that added 2 new labels to the prometheus error counters (query, select)
- Identified db errors will now result in http status 400 from server (instead of internal server error)
fixes #2974


N-o-Z:
@nopcoder @arielshaqed  Hi guys, please note that I plan on adding test cases to verify changes didn't break anything but I prefer to first get your initial feedback to understand if I'm on the right path.

pkg/db/errors.go
```diff
@@ -18,3 +20,24 @@ func isDialError(err error) bool {
...
+	if isUniqueViolation(err) {
+		err = ErrAlreadyExists
+	} else if pgxscan.NotFound(err) || errors.Is(err, pgx.ErrNoRows) {
+		log.Trace("SQL query returned no results")
+		err = ErrNotFound
+	} else {
+		err = fmt.Errorf("query %s: %w", query, err)
+	}
+
+	return err
```
arielshaqed, line 42: I'm a fan of early-exit, aka minimal indentation / `else`s.  It simplifies understanding of return flow.  In this case it would be
```suggestion
	if isUniqueViolation(err) {
		return ErrAlreadyExists
	}
	if pgxscan.NotFound(err) || errors.Is(err, pgx.ErrNoRows) {
		log.Trace("SQL query returned no results")
		return ErrNotFound
	}
	return fmt.Errorf("query %s: %w", query, err)
```

Also we should be careful about including queries in errors, in case they leak.

arielshaqed (changes requested):
Worried about leaking more data.

pkg/api/controller.go
```diff
@@ -1632,13 +1632,21 @@ func handleAPIError(w http.ResponseWriter, err error) bool {
 
 	case errors.Is(err, graveler.ErrLockNotAcquired):
```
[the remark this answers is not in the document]
arielshaqed: Indentation is still weird, not sure what's going on here.

pkg/api/controller.go
```diff
@@ -1632,13 +1632,21 @@ func handleAPIError(w http.ResponseWriter, err error) bool {
...
 		writeError(w, http.StatusInternalServerError, "branch is currently locked, try again later")
+
 	case errors.Is(err, adapter.ErrDataNotFound):
 		writeError(w, http.StatusGone, "No data")
+
+	case errors.Is(err, db.ErrNotFound),
```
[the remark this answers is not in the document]
arielshaqed: 1. This should probably be `auth.ErrNotFound` not `db.ErrNotFound` (currently they're the same, but they might not be...).
2. Why do we return http.StatusNotFound for other `ErrNotFound`s above?  More generally I don't think it will always be correct to return a 400 for all DB not founds.

pkg/api/controller.go
```diff
@@ -1632,13 +1632,21 @@ func handleAPIError(w http.ResponseWriter, err error) bool {
...
+		errors.Is(err, db.ErrAlreadyExists):
+		writeError(w, http.StatusBadRequest, err)
```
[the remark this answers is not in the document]
arielshaqed: Still open?

arielshaqed (commented):
Thanks!  Let's talk f2f if unclear what I'm worried about.

pkg/db/errors.go
```diff
@@ -18,3 +20,22 @@ func isDialError(err error) bool {
 	netError := &net.OpError{}
 	return errors.As(err, &netError) && netError.Op == "dial"
 }
+
+func handleSQLError(query string, err error, log logging.Logger, cmdType string) error {
+	if err == nil {
+		return nil
+	}
```
nopcoder, line 27: 1. can be method of `dbTx` - don't need to pass the logger in this case
2. no need to check the err - I think you need to call this func as part of returning an error
3. would change the method signature to accept the (type, query and err) or (err, type, query) - find it hard in the current order to find a meaning (maybe I should not)

pkg/db/tx.go
```diff
@@ -51,16 +49,24 @@ func (d *dbTx) Query(query string, args ...interface{}) (pgx.Rows, error) {
 			"args":  args,
 			"query": queryToString(query),
 		})
+	err = handleSQLError(query, err, log, "query")
```
nopcoder, line 52: we use this method to convert the error - so call it when you return the error, we already require to do so in order to return.
```go
if err != nil {
	return nil, handleSQLError("query", query, err)
}
```

nopcoder (changes requested):

pkg/api/controller.go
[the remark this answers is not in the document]
N-o-Z: As discussed I'll leave things as they are and in the new case for alreadyExists will pass a literal

pkg/db/tx.go
```diff
@@ -52,19 +50,26 @@ func (d *dbTx) Query(query string, args ...interface{}) (pgx.Rows, error) {
 			"query": queryToString(query),
 		})
 	if err != nil {
-		log.WithError(err).Error("SQL query failed with error")
-		return nil, err
+		return nil, d.handleSQLError(err, "query", query)
 	}
-	log.Trace("SQL query started successfully")
+	log.Trace("SQL query executed successfully")
```
arielshaqed, line 55:
```suggestion
	log.Trace("SQL query started successfully")
```

**Explanation:** The query has not yet been executed!  This function returns a `pgx.Rows`.  The caller will iterate over that to read each row.  So an error _can_ be returned, for instance if `query` is really broken... but the function returns before the query is over.

pkg/api/controller.go
```diff
@@ -1611,7 +1611,8 @@ func handleAPIError(w http.ResponseWriter, err error) bool {
 	switch {
 	case errors.Is(err, catalog.ErrNotFound),
 		errors.Is(err, graveler.ErrNotFound),
-		errors.Is(err, actions.ErrNotFound):
+		errors.Is(err, actions.ErrNotFound),
+		errors.Is(err, db.ErrNotFound):
```
arielshaqed, line 1615:
```suggestion
		errors.Is(err, actions.ErrNotFound),
		errors.Is(err, auth.ErrNotFound),
		errors.Is(err, db.ErrNotFound):
```
(and the same below).  The error returned by `auth` just happens to be an error from package `db`, but "obviously" it will be that error but from package `auth`.  In fact this is probably wrong; probably a different identical type should be returned by package `auth`.  So `errors.go` should say
```go
var (
	ErrNotFound                = fmt.Sprintf("%w", db.ErrNotFound)
	ErrAlreadyExists           = fmt.Sprintf("%w", db.ErrAlreadyExists)
// ...
)
```
or something, and not
```go
var (
	ErrNotFound                = db.ErrNotFound
	ErrAlreadyExists           = db.ErrAlreadyExists
// ...
)
```
.

Alternatively, _everything_ should wrap a separate "base" error `ErrNotFound`, and then a single `errors.Is` here would suffice.

None of this errorefactoring is needed here (of course).  But I would like you to catch errors from the package(s) that we called to generate them, in this case `auth`.

arielshaqed (approved):
Neat, thanks!

*Please* prefer to catch errors from package `auth` and not from package `db`, which is at a lower level.  Currently we (possibly incorrectly!) return _the same_ error as `db` does, when in fact we should probably do something better.  So for now let's just use `auth.ErrNotFound` and friends instead of (or in addition to) `db.ErrNotFound`.

pkg/api/controller.go
N-o-Z: I'll open a separate issue with a suggestion to refactor the error types and possibly handling

pkg/db/errors.go
```diff
@@ -18,3 +19,18 @@ func isDialError(err error) bool {
 	netError := &net.OpError{}
 	return errors.As(err, &netError) && netError.Op == "dial"
 }
+
+func (d *dbTx) handleSQLError(err error, cmdType string, query string) error {
```
nopcoder, line 23: can you move the implementation to `pkg/db/tx.go`

nopcoder (changes requested):

pkg/db/errors.go
N-o-Z: Can you please explain the rational behind it?

nopcoder: Sorry about that, I should have explained first. Usually, you keep the methods in the same file you define your receiver. It is easy to find the implementation this way. There are exceptions, and typically, they are baked by a file name convention - for example large routers or controllers where each method defines a request implementation - some choose to have a file per handler or group them by categories. In that case, you will find a common prefix or suffix file name.
The handleSQLError method is part of `dbTX` which is implemented in `tx.go`, I would not expect to find one of its method in `errors.go`.

N-o-Z: Thanks for the explanation

pkg/db/tx.go
```diff
@@ -128,18 +120,28 @@ func (d *dbTx) Exec(query string, args ...interface{}) (pgconn.CommandTag, error
...
+
+	// Each error that is added here should be updated also in controller.go:handleAPIError
+	if isUniqueViolation(err) {
+		return ErrAlreadyExists
+	}
+	if pgxscan.NotFound(err) {
+		d.logger.Trace("SQL query returned no results")
+		return ErrNotFound
+	}
+	return fmt.Errorf("query %s: %w", query, err)
```
nopcoder, line 142: Consider remove/change "query" to "%s statement: ..." or just "%s: ..." because the error can be from exec of insert operation and the word query can be confusing.

pkg/db/tx.go
```diff
@@ -128,18 +120,28 @@ func (d *dbTx) Exec(query string, args ...interface{}) (pgconn.CommandTag, error
...
-		return res, err
+		return res, d.handleSQLError(err, "exec", query)
 	}
 	log.Trace("SQL query executed successfully")
 	return res, err
 }
 
+func (d *dbTx) handleSQLError(err error, cmdType string, query string) error {
+	dbErrorsCounter.WithLabelValues(cmdType).Inc()
+	d.logger.WithError(err).Error("SQL query failed with error")
```
nopcoder, line 132: The main concern I have in with this one, maybe it can be addressed outside this pr - we log error level and count for all errors.
Before this change, In case we 'select' all repositories and didn't get any result - now for this case, we log an error and increase the error metric.
Also, additional logging after this log, like in like 139 will just fill the logs with additional information which is already known.
Verify that I understood the code right, but if this is the case, we may want to move these lines after the 'exists' and 'not found' issues.

pkg/db/tx_test.go
```diff
@@ -3,10 +3,9 @@ package db_test
 import (
 	"context"
 	"errors"
-	"testing"
-
 	"github.com/treeverse/lakefs/pkg/db"
 	"github.com/treeverse/lakefs/pkg/db/params"
+	"testing"
```
nopcoder, line 8: Use goimports it will take "testing" up with the rest of the packages and not here

pkg/db/tx_test.go
```diff
@@ -91,4 +90,18 @@ func TestGet(t *testing.T) {
 			t.Errorf("got %s wanted not found", err)
 		}
 	})
+
+	t.Run("failure unique", func(t *testing.T) {
+		_, err := d.Transact(ctx, func(tx db.Tx) (interface{}, error) {
+			var r R
+			_ = tx.Get(&r, `CREATE TABLE test_table (ID int NOT NULL UNIQUE)`)
```
nopcoder, line 97: would call the table "test_failure_unique" just in case somebody else testing the database here and will come up with the same name

nopcoder (changes requested):
Need your input on the main concern of log and count all the errors as part of this change.

pkg/db/tx_test.go
N-o-Z: Do you think we should delete the table at the end as well?

pkg/db/tx.go
nopcoder: The concern is that the original code condition some specific errors (isUniqueViolation for example).
The previous version didn't log as error and didn't count the error metric in this case.

N-o-Z: Got it!

pkg/db/tx.go
```diff
@@ -128,18 +120,28 @@ func (d *dbTx) Exec(query string, args ...interface{}) (pgconn.CommandTag, error
...
-		log.WithError(err).Error("SQL query failed with error")
-		return res, err
+		return res, d.handleSQLError(err, "exec", query)
 	}
 	log.Trace("SQL query executed successfully")
 	return res, err
 }
 
+func (d *dbTx) handleSQLError(err error, cmdType string, query string) error {
+	d.logger.Trace("SQL operation err:", err)
```
nopcoder, line 131: - Use `.WithError(err)` to capture `err` in a structure way - prefer to pass fields to the log in this format as later when we process the logs we can extract the information easily.
- Previously we used the scoped `log` variable which captures cmd type and query - we need to log this information here and in the error too, when we switched to use the receiver.

pkg/db/tx.go
```diff
@@ -128,18 +120,28 @@ func (d *dbTx) Exec(query string, args ...interface{}) (pgconn.CommandTag, error
...
+	// Each error that is added here should be updated also in controller.go:handleAPIError
+	if isUniqueViolation(err) {
+		return ErrAlreadyExists
+	}
+	if pgxscan.NotFound(err) {
+		return ErrNotFound
+	}
+
+	dbErrorsCounter.WithLabelValues(cmdType).Inc()
+	d.logger.WithError(err).Error("SQL operation unhandled error")
```
nopcoder, line 141: Same as in line 131 - WithError and log additional fields.
About the message - "unhandled" is something we don't know yet - prefer to report that the operation completed with error

nopcoder (changes requested):

pkg/db/tx.go
N-o-Z: Decided to pass the logger to handleSQLError not lose context, it is now no longer a member function anymore. Let me know if I should move it back to errors.go

nopcoder (approved):
LGTM

PR #3042, opened 2022-03-14
Improve Airbyte documentation

arielshaqed:
* Fix some whitespace, grammar, and phrasing.
* Use playground URL in example.

### DIFF
--- a/docs/integrations/airbyte.md
+++ b/docs/integrations/airbyte.md
@@ -1,37 +1,37 @@
 ---
 layout: default
 title: Airbyte
-description: Use Airbyte with lakeFS to easily sync data between applications and S3 with lakeFS version control. 
+description: Use Airbyte with lakeFS to easily sync data between applications and S3 with lakeFS version control.
 parent: Integrations
 nav_order: 57
 has_children: false
 ---
 
-[Airbyte](https://airbyte.io//) is an open-source platform to sync data from applications, APIs & databases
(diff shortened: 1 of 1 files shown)


docs/integrations/airbyte.md
```diff
@@ -1,53 +1,53 @@
...
+with lakeFS in the same way they would with Amazon S3.
 You can use Airbyte's [S3 Destination](https://airbyte.io/destinations/s3) for uploading the data to lakeFS.
 
 ### Configuring lakeFS using the connector
 Set the following parameters when creating a new Destination of type S3:
 
 | Name             | Value                                                        | Example                                                                                                             |
 |------------------|--------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------|
-| Endpoint         | The lakeFS S3 gateway URL                                    | `http://lakefs.example.com`                                                                                       |
+| Endpoint         | The lakeFS S3 gateway URL                                    | `https://cute-axolotol.lakefs-demo.io`
```
johnnyaug, line 42: Wondering if our docs should use lakefs Playground URLs ase examples.

arielshaqed: I (obviously) think that they should.

But I agree this is an unrelated change and should not block the other fixes.  Let me know if you would prefer me to drop this change -- perhaps we can/should discuss it separately

johnnyaug (approved):
Don't want to block this. Since many doc improvements are on the way, the URL thing will probably be addressed soon

Issue #2953, opened 2022-02-17
Use assumed roles in GitHub actions


johnnyaug:
In many of our GitHub workflows, we use this step:
```
      - name: Configure AWS Credentials
        uses: aws-actions/configure-aws-credentials@v1
        with:
          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
          aws-region: us-east-1
```
Instead of using secret keys, use assumed roles. 

I made an attempt to achieve this in 5db78fce, but couldn't get all the tests to pass and eventually gave up. You can start from this commit and debug.

PR #3043, opened 2022-03-14
lakefs release workflows without using credentials

nopcoder:
To be tested as part of a new release.

Part of #2953


johnnyaug (approved):

johnnyaug:
(Part of #2953)

PR #3044, opened 2022-03-14
Remove unused workflow

nopcoder:
Part of cleanup using aws key credentials.
I didn't saw it is being used for along time, wanted to remove stuff we don't maintain.


arielshaqed:
Replacing myself with @peacing as reviewer as I think he's closest to this workflow.

PR #3045, opened 2022-03-14
Allow non-interactive config for basic configuration

### LINKED ISSUE #3040: lakectl config to allow non-interactive single set of value
Allow all the below at least:

./lakectl config  set lakefs_access_key_id ACCESS_KEY
./lakectl config  set lakefs_secret_access_key SECRET_ACCESS_KEY
./lakectl config  set lakefs_endpoint  https://endpoint/

mishraprafful:
**Related Issues**
Closes #3040 

**Introduced Changes**
* Added `lakectl config set lakefs-access-key-id <access key id>` command
* Added `lakectl config set lakefs-secret-access-key <secret access key>` command
* Added `lakectl config set lakefs-endpoint-url <endpoint url>` command


nopcoder (changes requested):
Thanks @mishraprafful on your contribution.
The code addresses the examples in the issue, but a complete solution should enable using any of the possible keys allowed in our configuration.

You can use the metadata on `pkg/config/template.go`'s configuration structure to validate the configuration key we try to set.

The value validation is be done is couple of ways - based on the field type and metadata associated. The configuration can be complex there can be relationship between fields - one of the reasons we have a Validate method on our config (pkg/config/config.go), so we can just populate it and call validate.  But keep in mind that currently it will also validate required field.  In case we like to populate the lakefs config from zero it means that we will be required to set more than one key/value in order to get a valid configuration file after running the first command.

mishraprafful:
@nopcoder So if I understand the expected behavior correctly, we wish to let the user specify any of the possible keys from `pkg/config/configuration.go`. 

Also runs a config validation before writing the config, but the existing validation might fail as it ensures that all the required values are set. 

In case the user is setting the `access key` only from a zero config, should this be validated and added to the config, maybe validating with a warning stating something like 
```bash
Required Values Missing:
     secret-access-key
     endpoint-url
``` 
makes sense, let me know your thoughts?

nopcoder:
@mishraprafful as the current validation only checks required fields, I would keep the value validation based on the field type as trying to load the config after setting invalid type will fail to parse the value.

PR #3047, opened 2022-03-14
Use Go's embed feature instead of statik

nopcoder:
Reduce the dependencies and build complexity.
Currently we use 'statik' a command line tool to build go code to embed our UI and migration ddl files.
Using the Go's embed feature we can skip the code generation step (UI build is still required) to build.
Updated the way or UI handle identify page not found and redirect temporary.


Makefile
```diff
@@ -65,7 +65,6 @@ clean:
 		$(UI_BUILD_DIR) \
 		pkg/actions/mock \
 		pkg/api/lakefs.gen.go \
-		pkg/ddl/statik.go \
 		pkg/graveler/sstable/mock \
 		pkg/webui \
```
arielshaqed, line 69: I *think* this should also be removed... but I am not sure.  Could you either explain what goes in each `webui` directory, or remove this line and `fgrep -R webui` to be sure we got all bits?

arielshaqed (approved):
Thanks!

Please let's have someone from core/OSS (name TBD) team review this -- it's a fairly major change to the build process.

Makefile
nopcoder: Right, I missed this one - thanks!

nopcoder: Removing the extra directory - it was left for the cleanup and it is no longer used.
We will keep content.go file under the real webui to include the UI dist folder.

nopcoder:
Adding @itai-david to review this change

nopcoder:
Thanks @idanovo for helping with debugging the ui handler change with lakectl doctor.

idanovo (approved):
Looks good! one small question/fix

cmd/lakectl/cmd/dbt.go
```diff
@@ -186,7 +188,7 @@ generate_schema_name.sql
 {%- endmacro %}
 `
 		//nolint:gosec
-		err := ioutil.WriteFile(macroPath, []byte(generateSchemaData), 0644)
+		err := ioutil.WriteFile(macroPath, []byte(generateSchemaData), 0o644)
```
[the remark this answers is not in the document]
nopcoder: its an alternative way to write a number which is a file permission in Go - like: 10000 can be 10_000. In this case I think its my file formatting tool tries to make me look cool.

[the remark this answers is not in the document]
nopcoder: Sure, will be good as a separate PR

pkg/db/migration.go
```diff
@@ -52,12 +52,8 @@ func (d *DatabaseMigrator) Migrate(ctx context.Context) error {
 }
 
 func getStatikSrc() (source.Driver, error) {
```
itai-david, line 54: Maybe we can rename it, now that statik is out?

itai-david (approved):
One minor comment regarding possible func renaming. Other than that LGTM

pkg/db/migration.go
nopcoder: good idea - updated, thanks.

PR #3070, opened 2022-03-17
remove statik output from git ignore

nopcoder:
enable easy way to remove non generated files by statik which we no longer use.


itai-david (approved):

tshradheya:
@nopcoder should this change also include adding to go.mod dependencies
`	github.com/rakyll/statik v0.1.7`

Issue #2364, opened 2021-08-12
Dump configuration and details on startup with errors


arielshaqed:
Define a "startup phase" and/or "warmup period" for the lakeFS daemon.  If errors (certainly _fatal_ errors) occur during that time, dump *all* configuration.  Also dump its sources, if #2354 lets us do that.

Possible definitions:
* _Startup phase_: Right until all web services start.
* _Warmup period_: first 60 (?) seconds after startup.

edmondo1984:
Isn't there the risk to expose secret in logs with this solution? I imagine that if there is a process tailing logs, this could create a problem

itaiad200:
Good observation @edmondo1984! The dump should mask all potential secret information.

itaiad200:
I think you're right @edmondo1984, it looks odd. It's not clear to me why we're not panicking on that error too. Maybe @nopcoder sees something that we're missing?

nopcoder:
Also, agree it looks strange, and we need to address it.
But I'll try to explain the reason I think it is written this way. TL;DR - Startup database connection retries and migrate driver.

- This code verify that current db schema level match the version embedded
- We no longer migrate, so we enable a startup without db schema
- The migration connects to the database with its own driver / connection
- At some point, we added database connect retry on startup, it doesn't include to migrate driver
- In case the schema version check will exist on all errors - it probably will exit on fail to connect and there will be no retries.

arielshaqed:
Thanks for your work on this!

> Isn't there the risk to expose secret in logs with this solution? I imagine that if there is a process tailing logs, this could create a problem

I think that we are set up to handle this correctly.  Specifically, the config template [defines a type `SecureString`](https://github.com/treeverse/lakeFS/blob/234018be2d7e6f95db9cdbfad74bb3ab2bba29b5/pkg/config/template.go#L32) to use for secrets in configuration.  I opened #3074 to change the definition of its `String` method, making it _very_ hard to spill these secrets to the log.

itaiad200:
> Also, agree it looks strange, and we need to address it. But I'll try to explain the reason I think it is written this way. TL;DR - Startup database connection retries and migrate driver.
> 
> * This code verify that current db schema level match the version embedded
> * We no longer migrate, so we enable a startup without db schema
> * The migration connects to the database with its own driver / connection
> * At some point, we added database connect retry on startup, it doesn't include to migrate driver
[...]

PR #3074, opened 2022-03-20
Don't log secure strings by default

arielshaqed:
Prevent type `SecureString` from being logged or otherwise stringified by
accident.  Add a method `SecureValue` to allow access to their value that
cannot happen accidentally and is clearly flagged.


arielshaqed:
Relevant for #2364.

nopcoder (approved):
Thanks, just make the linter happy about the unused name

arielshaqed:
Yikes, thanks!

PR #3079, opened 2022-03-20
Add draining of stats metrics

itaiad200:


pkg/stats/collector.go
```diff
@@ -179,6 +179,11 @@ func (s *BufferedCollector) Run(ctx context.Context) {
 			s.cache = make(keyIndex)
 			go s.send(metrics) // no need to block on this
 		case <-ctx.Done(): // we're done
+			close(s.writes)
```
ozkatz, line 182: This means canceling the context passed to `Run()` might lead to a panic if `CollectEvent` is called after. Perhaps we should at least document that somewhere.

ozkatz (changes requested):
I feel this implementation is a bit risky. contexts are generally passed around and wrapped. canceling/timing-out/deadlining them shouldn't trigger a panic. perhaps we need a more explicit `Shutdown()` or `Drain()` API?

pkg/stats/collector.go
itaiad200: That's right. I'll remove the channel closing, it does no good other than the possible panic..

nopcoder: Yes, suggest we just enable Close method that will stop and wait for the collector to complete its job.

pkg/stats/collector.go
```diff
@@ -179,6 +179,11 @@ func (s *BufferedCollector) Run(ctx context.Context) {
 			s.cache = make(keyIndex)
 			go s.send(metrics) // no need to block on this
```
nopcoder, line 180: Think the issue is here to - sending using a background goroutine doesn't make sure the metrics are sent.
We can use the context of this goroutine to post the data and keep a large buffer for our channel.

nopcoder:
The current collector uses channel to collect events, we pass it to many services, and we don't track the actual sending of the metrics. At the point we like to go and make sure we flush the information, we don't have to stop the incoming event into the channel. Without us controlling the point of adding new events to the collector's channel and tracking the goroutine that sends the data, we not be certain that the metrics were sent.

nopcoder:
@itaiad200 added a branch with a suggestion how to handle the processing of events based on `writers` channel - https://github.com/treeverse/lakeFS/tree/139-bi-missing-init-alt

pkg/stats/collector.go
itaiad200: Thanks, looked at your PR and used some of it.

pkg/stats/collector.go
itaiad200: Handled that now.

nopcoder (changes requested):
We need to make sure that calling Close will always work and not wait for Run to end.
By using defer Close() we make sure that we always close and flush the collector resources.
So, the implementation should support this use case.

nopcoder (approved):
Had just comment about using defer while Close.
In a different, PR would suggest simplifying the metrics processing.

pkg/stats/collector.go
```diff
@@ -142,49 +158,88 @@ func (s *BufferedCollector) incr(k primaryKey) {
...
-			s.done <- true
-			return
+	atomic.StoreInt32(&s.runCalled, 1)
+	go func() {
+		for {
+			select {
+			case w := <-s.writes: // collect events
+				s.incr(w)
+			case <-s.heartbeatTicker.Tick():
+				s.CollectEvent("global", "heartbeat")
```
nopcoder, line 204: Use `s.incr` to update the cache with heartbeat information - at this scope this the recommended place to update the cache.
Collect event writes to the channel this goroutine processes. It can be tricky if the channel is full, and we try to send something while we can't process - it will block the code forever.

nopcoder (approved):

ozkatz (approved):
Thanks!

PR #3081, opened 2022-03-21
2983 lakectl ingest stages more objects than source s3 bucket

itai-david:
A bug fix, in the way lakectl's S3 store used pagination during walk + a test

### DIFF
--- a/cmd/lakectl/cmd/store/factory.go
+++ b/cmd/lakectl/cmd/store/factory.go
@@ -49,7 +49,7 @@ func Walk(ctx context.Context, storageURI string, walkFn func(e ObjectStoreEntry
 		if err != nil {
 			return err
 		}
-		walker = &S3Walker{S3: svc}
+		walker = &S3Walker{s3: svc}
 	case "gs":
 		svc, err := GetGCSClient(ctx)
 		if err != nil {
--- a/cmd/lakectl/cmd/store/factory.go
+++ b/cmd/lakectl/cmd/store/factory.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"net/url"
 	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
 )
 
 var (
@@ -45,7 +47,7 @@ func Walk(ctx context.Context, storageURI string, walkFn func(e ObjectStoreEntry
 	}
 	switch uri.Scheme {
 	case "s3":
-		svc, err := GetS3Client()
+		svc, err := GetS3Client(*aws.NewConfig())
 		if err != nil {
 			return err
 		}
--- a/cmd/lakectl/cmd/store/factory.go
+++ b/cmd/lakectl/cmd/store/factory.go
@@ -51,7 +51,7 @@ func Walk(ctx context.Context, storageURI string, walkFn func(e ObjectStoreEntry
 		if err != nil {
 			return err
 		}
-		walker = &S3Walker{s3: svc}
+		walker = &S3Walker{S3: svc}
 	case "gs":
 		svc, err := GetGCSClient(ctx)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/lakectl/cmd/store/s3.go
```diff
@@ -27,6 +28,10 @@ type S3Walker struct {
 	s3 s3iface.S3API
 }
 
+func (s *S3Walker) SetS3(s3 s3iface.S3API) {
```
nopcoder, line 31: Would prefer to expose the member, so I can initial the structure set/get this member, vs just enable a setter func.
But for this test, we don't need to expose a setter.  Calling `store.Walk` function will get us this walker and better code cover.

cmd/lakectl/cmd/store/factory.go
```diff
@@ -45,7 +47,7 @@ func Walk(ctx context.Context, storageURI string, walkFn func(e ObjectStoreEntry
 	}
 	switch uri.Scheme {
 	case "s3":
-		svc, err := GetS3Client()
+		svc, err := GetS3Client(*aws.NewConfig())
```
nopcoder, line 50: Just question - what is the difference between the default we used to have and passing the default new configuration?

nessie/store_test.go
```diff
@@ -0,0 +1,35 @@
...
+	"context"
+	"net/url"
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/stretchr/testify/require"
+	"github.com/treeverse/lakefs/cmd/lakectl/cmd/store"
+)
+
+const IngestTestBacketPath = "s3://nessie-system-testing-data/ingest-test-data/"
```
nopcoder, line 13: Bucket

nessie/store_test.go
```diff
@@ -0,0 +1,35 @@
...
+
+func TestS3Walk(t *testing.T) {
```
nopcoder, line 15: Using the Walk function, we can test all the storage we support - not saying we need to add configuration currently to Nessie now. But writing the test function using table-driven style, we add just pass different bucket Path
Can drop the S3 from the name and use table-driven code, so adding another address we can check more storage systems.

Something like:
```go
func TestWalk(t *testing.T) {
	// Test bucket was uploaded with 2100 as the test is written. If the test fails on this number,
	// make sure there were no changes made to the bucket, or update this number accordingly
	const expectedNumObjs = 2100
	tests := []struct {
		Name       string
		BucketPath string
	}{
		{Name: "s3", BucketPath: "s3://nessie-system-testing-data/ingest-test-data/"},
	}

	for _, tt := range tests {
		t.Run(tt.Name, func(t *testing.T) {
			uri, err := url.Parse(tt.BucketPath)
			require.NoError(t, err, "URL parsing failed", tt.BucketPath)
			numObjs := 0
			store.Walk(context.Background(), uri, func(store.ObjectStoreEntry) error {
				numObjs++
				return nil
			})
			require.Equal(t, expectedNumObjs, numObjs, "Wrong number of objects detected", tt.Name)
		})
	}
}

```

nopcoder (changes requested):

cmd/lakectl/cmd/store/factory.go
itai-david: No difference in this context, which is exactly what I needed                                                                       .
On the other hand, from the test, I can force certain params, such as the region, which is required in order to access the `nessie-system-testing-data` bucket

nessie/store_test.go
itai-david: Done

nessie/store_test.go
itai-david: I can open an issue to address this in the future - WDYT?

nopcoder (approved):

nessie/store_test.go
itai-david: Opened #3101 and #3102

nopcoder: Think you can close/delete 3102, we would like to run all the tests and locally with the suggested code we can run specific storage.

nopcoder (approved):
LGTM - small comment about the issue and the test code

PR #3083, opened 2022-03-21
Add draining of stats metrics

itaiad200:


pkg/stats/collector.go
```diff
@@ -156,35 +162,78 @@ func (s *BufferedCollector) send(metrics []Metric) {
...
+	}()
+}
+
+func (s *BufferedCollector) sendCache() {
+	if len(s.cache) == 0 {
+		return
+	}
+	metrics := makeMetrics(s.cache)
+	for k := range s.cache {
+		delete(s.cache, k)
```
itaiad200, line 234: This is dangerous if someone added to the cache between the `metrics` were created and the cache was deleted

itaiad200:
Irrelevant.

PR #3084, opened 2022-03-21
Allow ingesting from a non-default S3 endpoint

### LINKED ISSUE #2886: lakectl ingest from Minio
Is it possible to use `lakectl ingest ` to get data from minio instead of amazon s3 ?

tshradheya:
Fixes #2886 

Command used to ingest:

```
 LAKECTL_S3_ENDPOINT_URL="http://127.0.0.1:9000"
  ./lakectl ingest --from s3://src-buck --to lakefs://dest-buck/main/ --intg minio 

```


tshradheya:
@johnnyaug i have coded a working feature but need some advice on why this happens when i ingest the data 

image

The data is always referenced inside an empty directory 

Also whenever i click to download the file, it download a `Failed-Server prolem objects.json`

johnnyaug:
@tshradheya, thank you for tackling this issue!
1. Regarding the empty directory, I think it's a bug in the existing ingest code. I will look into it and open an issue.
1. About the error you're getting, maybe your installation cannot access your minio storage? Is lakeFS configured on top of minio? Are you able to upload a file into your repository from the UI?
1. In general, I think the implementation should not use a separate MinIO client. Instead, I would use the MinIO S3-compatible API. You can introduce a new LAKECTL_S3_ENDPOINT_URL variable, and use the same S3 client - only adding the endpoint parameter to use MinIO instead of S3.

tshradheya:
1. Thanks, appreciate the help
2. Currently my local lakefs has a repository with namespace of a local minio server. And i am trying ingesting from http://play.min.io . From my understanding after ingestion I should be able to read that data from my local minio server? Please correct my understanding if i am wrong
3. Thanks for this suggestion! New to the world of S3 and block storage so let me read up on this and change implementation accordingly 😄

johnnyaug:
I think this feature is a bit more tricky than we thought.
As mentioned in our [docs for ingest](https://docs.lakefs.io/setup/import.html#zero-copy-import), the lakeFS installation needs to be able to read the source objects. Note that ingesting does not copy any objects - it only creates pointers to the objects in the source bucket.

In the case of MinIO - it also means that the installation **needs to use the same MinIO installation as the source.** We will have to document this once this is merged.
So in short, you can't ingest data from one MinIO installation to another- you can only ingest from a MinIO bucket to a lakeFS repository which is on top of the same MinIO installation.

tshradheya:
> In the case of MinIO - it also means that the installation needs to use the same MinIO installation as the source. We will have to document this once this is merged.

Yes, this makes more sense in my head now haha
I fixed my issue of download and it was because of wrong address. 


Will work on simplifying the logic as you said and then update the docs as well :)

tshradheya:
> In general, I think the implementation should not use a separate MinIO client. Instead, I would use the MinIO S3-compatible API. You can introduce a new LAKECTL_S3_ENDPOINT_URL variable, and use the same S3 client - only adding the endpoint parameter to use MinIO instead of S3.

I have added three different variable for minio when ingesting. This will allow ingestion from Minio to any S3 blockstore like AWS S3 as well if my understanding is correct. 
I have updated the implementation, please do take a look and let me know if anything needs to be revised and then i can accordingly document

Ingest command:
```
MINIO_STORAGE_ACCOUNT="127.0.0.1:9000" MINIO_STORAGE_ACCESS_KEY="minioadmin" MINIO_STORAGE_SECRET_ACCESS_KEY="minioadmin"
  ./lakectl ingest --from s3://src-buck --to lakefs://dest-buck/main/ --intg minio

```

cmd/lakectl/cmd/ingest.go
```diff
@@ -132,6 +133,7 @@ func init() {
 	ingestCmd.Flags().String("to", "", "lakeFS path to load objects into (e.g. \"lakefs://repo/branch/sub/path/\")")
 	_ = ingestCmd.MarkFlagRequired("to")
 	ingestCmd.Flags().Bool("dry-run", false, "only print the paths to be ingested")
+	ingestCmd.Flags().String("intg", "", "integration source which must be used for ingestion")
```
arielshaqed, line 136:
```suggestion
	ingestCmd.Flags().String("integration-source", "", "integration source to use for ingestion")
```

"intg" is a big hard for me to understand.  I'd be happy for an alias or even a one-letter short flag if `integration-source` is too much to type.  But I _do_ like the descriptive flag name for a flag which is not used much, it is easier for me to read.

cmd/lakectl/cmd/store/s3.go
```diff
@@ -12,15 +14,38 @@ import (
...
-func GetS3Client() (*s3.S3, error) {
-	sess, err := session.NewSessionWithOptions(session.Options{
-		SharedConfigState: session.SharedConfigEnable,
-	})
-	if err != nil {
-		return nil, err
+func GetS3Client(intg string) (*s3.S3, error) {
+	if intg == "minio" {
+		// Configure to use MinIO Server
+		accountName, accessKey, secretKey := os.Getenv("MINIO_STORAGE_ACCOUNT"), os.Getenv("MINIO_STORAGE_ACCESS_KEY"), os.Getenv("MINIO_STORAGE_SECRET_ACCESS_KEY")
```
arielshaqed, line 20: I cannot find documentation for these environment variables.  There is [no reference](https://www.google.com/search?q=site%3Amin.io+"MINIO_STORAGE_ACCOUNT") to these environment variables in min.io or on [the Minio organization on GitHub](https://github.com/search?q=org%3Aminio+%22MINIO_STORAGE_ACCOUNT%22&type=code).

arielshaqed (commented):
Hi @tshradheya,

Thanks for this extension of `lakectl ingest` to work with MinIO!  As you can probably imagine, we _really_ want it to be easy to ingest data into lakeFS.

I am having difficulty understanding the solution in this PR, possibly because I come from a more AWS background.  It should of course be possible to hook up `lakectl` to any S3-like source.  But as an S3 user I would expect that to happen by configuring the regular `AWS_*` environment variables (either `AWS_PROFILE` or `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` and possibly a few more).  So what I would primarily be missing might be a way to specify a different S3 endpoint URL for lakectl to use.

I am not sure how many of our users will have these `MINIO_*` environment variables set.  Ideally these variables would be appear in minIO documentation (maybe for the `mc` command?).  Then the documentation for the `integration-source` flag that you are adding could link to that documentation.  Of course these variables might commonly be used in the MinIO community -- we could link to any page that talks about them, really.

Thanks!  Really looking forward to improving onboarding from MinIO to lakeFS.

tshradheya:
Hi, thanks for the feedback.

You are right, these env variables aren't in anyway related to MINIO_*

I wanted to check if the following scenario is possible when lakefs has namespace from AWS S3 but wants to ingest data from MinIO?

To support the above, is the main reason why I introduced my own MINIO env variables so they don't clash with the S3 variables set by lakefs

If above doesn't need to supported then i can simplify the logic to just accept S3_ENDPOINT and use the already defined AWS tokens for the credentials.


Once the implementation has a firstpass i will write up documentation for this as well 😀

arielshaqed:
> Hi, thanks for the feedback.

Sure, @tshradheya !  As you can imagine, it's really important for us to make it easy to ingest data into lakeFS.  So thanks for helping to do that!

> 
> You are right, these env variables aren't in anyway related to MINIO_*
> 
> I wanted to check if the following scenario is possible when lakefs has namespace from AWS S3 but wants to ingest data from MinIO?

This will be difficult, but for other reasons.  "Ingest" is not a copy operation, rather it creates metadata on the lakeFS side that points at existing objects.  So those objects must be accessible from the same S3 endpoint that lakeFS uses.

There is no support currently for _multiple block adapters_.  The challenge here is not so much to code it as to ensure consistent behaviour and semantics.  We'll need (several) very concrete use-cases for such support before we can implement it.

So you will need to copy your files from MinIO to lakeFS over S3.  There are numerous 3rd-party utilities to perform bulk S3 copies, depending on which tooling you are most familiar with and your performance requirements.  I would probably start with our integration guide [Copying data with Rclone](https://docs.lakefs.io/integrations/rclone.html), and just copy everything from MinIO into lakeFS and then commit.
If performance turns out to be a limiting factor then you have a _lot_ of data, so you will want to have a Spark cluster.  Assuming you do, you could [use DistCp to copy data](https://docs.lakefs.io/integrations/distcp.html).

> 
> To support the above, is the main reason why I introduced my own MINIO env variables so they don't clash with the S3 variables set by lakefs
> 
> If above doesn't need to supported then i can simplify the logic to just accept S3_ENDPOINT and use the already defined AWS tokens for the credentials.

Your code only touches `lakectl` (the command-line tool), and `lakefs` (the daemon/server) will be entirely unaffected by whatever settings you change here.  So there is no need to read credentials from another source, just use whatever credentials mechanism in boto is easiest for you.

You *will* most likely need to change the endpoint URL, so I would consider adding a CLI flag that changes that.
 
> Once the implementation has a firstpass i will write up documentation for this as well grinning

Sure!  Thanks for taking the time to explain.  Just please note you will *not* be able to ingest data from MinIO into a lakeFS that runs on S3, you will need to copy as explained above.

If you do need a lakeFS that can handle multiple backends at once, please open an issue.  It will be a **big** discussion, and we will have some questions.  🐘 ❓ 🐘 ❓ 🐘

tshradheya:
@arielshaqed Thanks for being patient and explaining 😄 

I now have a much better understanding of how things work.

I have updated the implementation to use the shared config from AWS CLI. I am thinking we should support `config.Profile` as well since people may have multiple S3 creds stored in AWS CLI 

Updated the docs as well! Would be great if you could help take a look :)

cmd/lakectl/cmd/store/s3.go
```diff
@@ -12,15 +13,34 @@ import (
...
+				Region:           aws.String("us-east-1"), // Needs region for validation as it is AWS client
+				S3ForcePathStyle: aws.Bool(true),
+			},
+			SharedConfigState: session.SharedConfigEnable,
+		})
+	} else {
+		sess, err = session.NewSessionWithOptions(session.Options{
+			SharedConfigState: session.SharedConfigEnable,
+		})
+	}
```
arielshaqed, line 37: 1. Can we get the URL from a flag, instead, say `s3-endpoint-url`?  Every other lakectl option uses flags, this should too.
1. I would rather not have to mention "minio" at all.  If I specify `--s3-endpoint-url`, then presumably I want to use that endpoint.  It might be MinIO, it might be any other S3 compatible storage, it might even be AWS S3 just using a different endpoint.  So I would drop the flag `integration-source` entirely.

cmd/lakectl/cmd/store/s3.go
```diff
@@ -12,15 +13,34 @@ import (
...
 	if err != nil {
 		return nil, err
 	}
 	svc := s3.New(sess)
 	return svc, nil
+
```
arielshaqed, line 43: nit:
```suggestion
```
We prefer not to accumulate blank lines at the ends of functions.

arielshaqed (changes requested):
Thanks!

I'd be happy to approve if we have just an `endpoint-url` flag, and no `integration-source` flag.

Please note that ingestion does not copy objects.  So this code _will_ ingest objects from your MinIO installation to a lakeFS set up to use that same MinIO installation.  I believe you indicated that your use-case might be to copy files on a MinIO installation into a lakeFS set up to write to S3.  This code will **not** be able to make that work.

cmd/lakectl/cmd/store/s3.go
tshradheya: Nice, that makes it much more clearer 👍

tshradheya:
> I believe you indicated that your use-case might be to copy files on a MinIO installation into a lakeFS set up to write to S3. This code will not be able to make that work.

Yes it makes sense to me now. I currently dont need that functionality but was assuming that this might have to support that as well since both are S3 compatible but as you explained i was wrong to assume so

tshradheya:
@arielshaqed could you help review the updated PR? Thanks :)

cmd/lakectl/cmd/ingest.go
```diff
@@ -132,6 +133,7 @@ func init() {
 	ingestCmd.Flags().String("to", "", "lakeFS path to load objects into (e.g. \"lakefs://repo/branch/sub/path/\")")
 	_ = ingestCmd.MarkFlagRequired("to")
 	ingestCmd.Flags().Bool("dry-run", false, "only print the paths to be ingested")
+	ingestCmd.Flags().String("s3-endpoint-url", "", "for ingesting from non-AWS S3 compatible sources which need endpoint (e.g. MinIO)")
```
arielshaqed, line 136:
```suggestion
	ingestCmd.Flags().String("s3-endpoint-url", "", "URL to access S3 storage API (by default, use regular AWS S3 endpoint)")
```

cmd/lakectl/cmd/store/s3.go
```diff
@@ -12,10 +12,23 @@ import (
...
 )
 
-func GetS3Client() (*s3.S3, error) {
-	sess, err := session.NewSessionWithOptions(session.Options{
-		SharedConfigState: session.SharedConfigEnable,
-	})
+func GetS3Client(s3EndpointUrl string) (*s3.S3, error) {
+	var sess *session.Session
+	var err error
+	if len(s3EndpointUrl) != 0 {
```
arielshaqed, line 18:
```suggestion
	if s3EndpointUrl != "" {
```

cmd/lakectl/cmd/store/s3.go
```diff
@@ -12,10 +12,23 @@ import (
...
+				Region:           aws.String("us-east-1"), // Needs region for validation as it is AWS client
+				S3ForcePathStyle: aws.Bool(true),
+			},
+			SharedConfigState: session.SharedConfigEnable,
+		})
+	} else {
+		sess, err = session.NewSessionWithOptions(session.Options{
+			SharedConfigState: session.SharedConfigEnable,
+		})
+	}
```
arielshaqed, line 31: (Not required!)

Usually I prefer to extract shared code outside of conditionals.  Here this would be dropping the :
```suggestion
	var config aws.Config{}
	if s3EndpointUrl != "" {
		config = aws.Config{
			Endpoint:         aws.String(s3EndpointUrl),
			Region:           aws.String("us-east-1"), // Needs region for validation as it is AWS client
			S3ForcePathStyle: aws.Bool(true),
			}
	}
	sess, err := session.NewSessionWithOptions(session.Options{
		SharedConfigState: session.SharedConfigEnable,
		Config: config,
	})
	if err != nil {
		return nil, err
	}
```

docs/setup/import.md
```diff
@@ -48,6 +48,7 @@ lakectl ingest \
 
 The `lakectl ingest` command will attempt to use the current user's existing credentials and will respect instance profiles,
 environment variables and credential files [in the same way that the AWS cli does](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-quickstart.html){: target="_blank" }
+If you wish to ingest from other S3 compatible storage solution you could use the flag to specify the endpoint `--s3-endpoint-url https://play.min.io`
```
arielshaqed, line 51: I would shorten phrasing here:
```suggestion
Specify an endpoint to ingest from other S3 compatible storage solutions, e.g. add `--s3-endpoint-url https://play.min.io`.
```

arielshaqed (commented):
Thanks for your contribution and you patience with me!

I'm running the tests and everything should be cool now.  I would obviously prefer for you to handle the minor comments I added in this round, but that is **not** required.

I am renaming this PR to better reflect what it does.

arielshaqed (approved):
_I had **one job**_... and I forgot to approve.  Sorry :-/

arielshaqed:
Please note test and lint failures.  No worries, this is just CI/CD doing its job.  Let me know if you need help fixing them.

arielshaqed:
Thanks!  We should be good to go as soon as tests pass :-)

arielshaqed:
Tests passed on #3140!

arielshaqed:
enforce-label check is incorrect, everything else is green ✅ , so pulling with extreme administrative force.

Thanks again, @tshradheya !

PR #3104, opened 2022-03-23
Validate new repo isn't using existing storage namespace

### LINKED ISSUE #3077: Fail repository creation when its storage namespace is already used by another repository
lakeFS repositories must be created with a unique storage namespace. that is, two repositories cannot share the same storage namespaces. 

#### Why 
It is a prerequisite for some advanced lakeFS features such as Garbage collection.

itaiad200:
close #3077

### DIFF
--- a/.github/workflows/nessie.yaml
+++ b/.github/workflows/nessie.yaml
@@ -455,9 +455,6 @@ jobs:
           LAKEFS_BLOCKSTORE_S3_CREDENTIALS_ACCESS_KEY_ID: ${{ secrets.NESSIE_AWS_ACCESS_KEY_ID }}
           LAKEFS_BLOCKSTORE_S3_CREDENTIALS_SECRET_ACCESS_KEY: ${{ secrets.NESSIE_AWS_SECRET_ACCESS_KEY }}
 
-      - name: Copy repository ref
-        run: aws s3 cp --recursive s3://nessie-system-testing-data/golden-files/gc-test-data s3://nessie-system-testing/${{ github.run_number }}-storage-rclone-export
-
       - name: Setup lakeFS for tests
         env:
           STORAGE_NAMESPACE: s3://nessie-system-testing/${{ github.run_number }}-storage-rclone-export
--- a/.github/workflows/nessie.yaml
+++ b/.github/workflows/nessie.yaml
@@ -325,8 +325,8 @@ jobs:
 
       - name: Test lakeFS S3 with Spark 2.x
         env:
-          STORAGE_NAMESPACE: s3://nessie-system-testing/${{ github.run_number }}-spark
-          REPOSITORY: gateway-test
+          STORAGE_NAMESPACE: s3://nessie-system-testing/${{ github.run_number }}-spark2
+          REPOSITORY: gateway-test-spark2
           SONNET_JAR: sonnets-246/target/sonnets-246/scala-2.11/sonnets-246_2.11-0.1.0.jar
         working-directory: test/spark
         run: ./run-test.sh
@@ -396,8 +396,8 @@ jobs:
       - name: Test lakeFS S3 with Spark 3.x
         continue-on-error: true
         env:
-          STORAGE_NAMESPACE: s3://nessie-system-testing/${{ github.run_number }}-spark
-          REPOSITORY: gateway-test
+          STORAGE_NAMESPACE: s3://nessie-system-testing/${{ github.run_number }}-spark3
+          REPOSITORY: gateway-test-spark3
           SONNET_JAR: sonnets-311/target/sonnets-311/scala-2.12/sonnets-311_2.12-0.1.0.jar
         working-directory: test/spark
         run: ./run-test.sh
--- a/nessie/repository_test.go
+++ b/nessie/repository_test.go
@@ -6,7 +6,6 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/require"
-	"github.com/treeverse/lakefs/pkg/api"
 )
 
 func TestRepositoryBasicOps(t *testing.T) {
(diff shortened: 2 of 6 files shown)


eden-ohana (approved):
LGTM

.github/workflows/nessie.yaml
```diff
@@ -455,9 +455,6 @@ jobs:
           LAKEFS_BLOCKSTORE_S3_CREDENTIALS_ACCESS_KEY_ID: ${{ secrets.NESSIE_AWS_ACCESS_KEY_ID }}
           LAKEFS_BLOCKSTORE_S3_CREDENTIALS_SECRET_ACCESS_KEY: ${{ secrets.NESSIE_AWS_SECRET_ACCESS_KEY }}
 
-      - name: Copy repository ref
```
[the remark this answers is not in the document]
itaiad200: It was copied from some other test and it's meaningless in this context.
But it does put files in the storage namespace, which is not something you can do with this PR.

pkg/api/controller.go
```diff
@@ -1238,20 +1240,33 @@ func (c *Controller) CreateRepository(w http.ResponseWriter, r *http.Request, bo
...
 	if err != nil {
 		return err
 	}
-	_, err = adapter.Get(ctx, obj, objLen)
-	return err
+	if _, err = c.BlockAdapter.Get(ctx, obj, objLen); err != nil {
+		return err
+	}
+
+	// if a single file exists under the lakeFS prefix in the storage namespace
```
arielshaqed, line 1261: AFAICS this allows multiple repositories to use the same storage namespace, as long as I don't use them first.  (So it is also a race if I create both repos and then upload a file to each.)

Why not just check for existence of the "dummy" file?  (It would _still_ be a race, but at least it would prevent this case.)

Thinking about it, one safe way to do this without a full database would be to keep a Graveler object with all storage namespaces.  Then you could verify-nonoverlap-and-add, and use a commit-like operation to prevent races.  Maybe you could have a hidden "global lakeFS" repo and keep a file named for each storage namespace.

Anyway, the current code is racy -- and since the time between creating a repo and using it is **unbounded**, there is no protection against this race.

arielshaqed (changes requested):
Not sure how this should work.  What if I type this sequence of commands?

```sh
lakectl repo create lakefs://foo s3://bucket/quux & lakectl repo create lakefs://bar s3://bucket/quux
# Both might succeed, the empty commit only occurs after `ensureStorageNamespace`.
lakectl fs upload -s file lakefs://foo/main/file
lakectl fs upload -s file lakefs://boo/main/file
# Now I have two nonempty repos!
```

pkg/api/controller.go
johnnyaug: +1 for using the dummy file.
I think we can live with the race- but if we really want to solve it we can implement a conditional put in the block adapter (like If-None-Match in S3).

Note that using a "global lakeFS" repository will not solve the race for multiple lakeFS installations (to me it's not really a problem but just noting).

itaiad200: Yeah, I missed Oz remarks of the requirements in #3077 - we need to tighten it some more.
The requirements states cross-installations enforcement, so a graveler object won't do the trick here.

I avoided checking the `dummy` object because I feel less comfortable treating it as a contract between repositories. It's there because of a one-time check, no one ever checks it the second time, and it may very well be deleted without us knowing (GC, lifecycle policies, etc..). But I seem to be the only one who believes it's fragile, so I'll do this check..

itaiad200: For example, now that we delete a repo we also need to delete that `dummy` file.

nessie/repository_test.go
```diff
@@ -24,10 +25,39 @@ func TestRepositoryBasicOps(t *testing.T) {
 		require.Contains(t, listedRepos, repo, "repository missing in listing")
 	}
 
+	// get repo and try creating another repo with the same storage namespace
+	testDupStorageNamespace(t, repos[0])
```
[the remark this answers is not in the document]
itaiad200: Deleted and added a UT.

pkg/api/controller.go
```diff
@@ -1238,19 +1242,28 @@ func (c *Controller) CreateRepository(w http.ResponseWriter, r *http.Request, bo
 	writeResponse(w, http.StatusCreated, response)
 }
 
-func ensureStorageNamespace(ctx context.Context, adapter block.Adapter, storageNamespace string) error {
+var errorStorageNamespaceInUse = errors.New("lakeFS repositories can't share storage namespace")
```
arielshaqed, line 1245:
```suggestion
var errStorageNamespaceInUse = errors.New("lakeFS repositories can't share storage namespace")
```

pkg/api/controller.go
```diff
@@ -1238,19 +1242,28 @@ func (c *Controller) CreateRepository(w http.ResponseWriter, r *http.Request, bo
...
 		dummyKey  = "dummy"
 		dummyData = "this is dummy data - created by lakeFS in order to check accessibility"
 	)
 
 	obj := block.ObjectPointer{StorageNamespace: storageNamespace, Identifier: dummyKey}
 	objLen := int64(len(dummyData))
-	err := adapter.Put(ctx, obj, objLen, strings.NewReader(dummyData), block.PutOpts{})
-	if err != nil {
+	if _, err := c.BlockAdapter.Get(ctx, obj, objLen); err == nil {
+		return fmt.Errorf("found lakeFS objects in the storage namespace(%s): %w",
```
arielshaqed, line 1256:
```suggestion
		return fmt.Errorf("%s: %w",
```
(and maybe rephrase the error itself if that doesn't look nice enough).

arielshaqed (approved):
Thanks!

This is cool and certainly (much _much_) better than the existing, but still keeps the race condition.  And it **can** happen, e.g. if automation sets up the two repos.

Obviously _loads_ better than the old way of doing nothing, but still dangerous.  Perhaps we will be able to do better with the refstore?

(Note that **no matter what you do** there is danger!  E.g. I can set up two distinct lakeFS installations and share storage namespaces between them.  There will always be danger, unless we keep our own directory listings for each lakeFS repo _and_ avoid re-using filenames...)

pkg/api/controller.go
itaiad200: Would rather keep it as is. The first sentence makes it clearer IMO.

pkg/api/controller.go
```diff
@@ -1203,22 +1203,26 @@ func (c *Controller) CreateRepository(w http.ResponseWriter, r *http.Request, bo
 		return
 	}
 
-	err := ensureStorageNamespace(ctx, c.BlockAdapter, body.StorageNamespace)
+	err := c.ensureStorageNamespace(ctx, body.StorageNamespace)
 	if err != nil {
+		msg := "Could not access storage namespace"
```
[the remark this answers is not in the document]
itaiad200: Done.

PR #3108, opened 2022-03-23
3085 lakectl ingest adds multiple excess slash to object name

itai-david:
Fixed `ingest` function not to add a `"/"` if the to path ends wirh rhe branch name (and thus ends with `"/"`)
Fixed `S3` `Walk` to correctly trim prefixes: with `"/"` if exists, and partial prefixes too

Tested manually with `--from` `s3:://mybucket/data/`, `s3:://mybucket/data` and `s3:://mybucket/da`
Opened #3109 for adding tests

### DIFF
--- a/cmd/lakectl/cmd/ingest.go
+++ b/cmd/lakectl/cmd/ingest.go
@@ -64,7 +64,7 @@ var ingestCmd = &cobra.Command{
 		if lakefsURI.Path != nil {
 			path = *lakefsURI.Path
 		}
-		if !strings.HasSuffix(path, PathDelimiter) {
+		if len(path) > 0 && !strings.HasSuffix(path, PathDelimiter) {
 			path = path + PathDelimiter // append a path delimiter (slash) if not passed by the user
 		}
 		go func() {
--- a/cmd/lakectl/cmd/ingest.go
+++ b/cmd/lakectl/cmd/ingest.go
@@ -65,7 +65,7 @@ var ingestCmd = &cobra.Command{
 			path = *lakefsURI.Path
 		}
 		if len(path) > 0 && !strings.HasSuffix(path, PathDelimiter) {
-			path = path + PathDelimiter // append a path delimiter (slash) if not passed by the user
+			path = path + PathDelimiter // append a path delimiter (slash) if not passed by the user, and it's not an empty path in lakeFS
 		}
 		go func() {
 			err := store.Walk(ctx, from, func(e store.ObjectStoreEntry) error {
--- a/cmd/lakectl/cmd/store/s3.go
+++ b/cmd/lakectl/cmd/store/s3.go
@@ -35,7 +35,7 @@ func (s *S3Walker) Walk(ctx context.Context, storageURI *url.URL, walkFn func(e 
 	// trimPrefix will be used to hold the actual value needed to be trimmed from the keys returned by the
 	// ListObjects function below. As the prefix itself might contain partial key, it cannot be used for the
 	// trim purpose, as this will create partial object names. The trimPrefix is the 'complete' prefix of the
-	// prefix, i.e. up to the first seperator ("/"), if exists, or an empty string. When the trimPrefix is
+	// prefix, i.e. up to the first separator ("/"), if exists, or an empty string. When the trimPrefix is
 	// trimmed from the key, the remains will be the object name.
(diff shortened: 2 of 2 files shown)


cmd/lakectl/cmd/ingest.go
```diff
@@ -64,7 +64,7 @@ var ingestCmd = &cobra.Command{
 		if lakefsURI.Path != nil {
 			path = *lakefsURI.Path
 		}
-		if !strings.HasSuffix(path, PathDelimiter) {
+		if len(path) > 0 && !strings.HasSuffix(path, PathDelimiter) {
 			path = path + PathDelimiter // append a path delimiter (slash) if not passed by the user
```
itaiad200, line 68:
```suggestion
			path = path + PathDelimiter // append a path delimiter (slash) if not passed by the user, and it's not an empty path in lakeFS
```

cmd/lakectl/cmd/store/s3.go
```diff
@@ -31,6 +31,10 @@ func (s *S3Walker) Walk(ctx context.Context, storageURI *url.URL, walkFn func(e
 	var continuation *string
 	const maxKeys = 1000
 	prefix := strings.TrimLeft(storageURI.Path, "/")
+	var trimPrefix string
+	if idx := strings.LastIndex(prefix, "/"); idx != -1 {
+		trimPrefix = prefix[:idx+1]
```
itaiad200, line 36: It took me some time to understand why this is the right value, it's based on the `ObjectStoreEntry` definition:
```
	// RelativeKey represents a path relative to prefix (or directory). If none specified, will be identical to FullKey
```

Would you mind adding a comment explaining the `trimPrefix` part?

itaiad200 (approved):
Thanks for this fix! Looking to see the tests, but probably not part of this PR.
See my 2 comments

cmd/lakectl/cmd/store/s3.go
itai-david: Done

arielshaqed:
Please rebase on `origin/master` (or merge that branch in): the name of a required test has changed, and GitHub is waiting for the new name.

https://lakefs.slack.com/archives/C01APUNTSGH/p1648120054377969?thread_ts=1648120054.377969&cid=C01APUNTSGH

cmd/lakectl/cmd/store/s3.go
```diff
@@ -31,6 +31,24 @@ func (s *S3Walker) Walk(ctx context.Context, storageURI *url.URL, walkFn func(e
...
+	// trimmed from the key, the remains will be the object name.
+	// Example:
+	// Say we have the following keys:
+	// pref/object
+	// pref/obj/another
+	// If we specify prefix="pref/obj" (both keys will be listed) then trimPrefix="pref/" and the trim result
+	// for the keys will be:
+	// object
+	// obj/another
+	var trimPrefix string
```
johnnyaug, line 48:
```suggestion
	var basePath string
```

cmd/lakectl/cmd/store/s3.go
```diff
@@ -31,6 +31,24 @@ func (s *S3Walker) Walk(ctx context.Context, storageURI *url.URL, walkFn func(e
...
+	// prefix, i.e. up to the first separator ("/"), if exists, or an empty string. When the trimPrefix is
+	// trimmed from the key, the remains will be the object name.
+	// Example:
+	// Say we have the following keys:
+	// pref/object
+	// pref/obj/another
+	// If we specify prefix="pref/obj" (both keys will be listed) then trimPrefix="pref/" and the trim result
+	// for the keys will be:
+	// object
+	// obj/another
```
johnnyaug, line 47:
```suggestion
	// basePath is the path relative to which the walk is done. The key of the resulting entries will be relative to this path.
	// As the original prefix might not end with a separator, it cannot be used for the
	// trim purpose, as this will create partial "folder" names. When the basePath is
	// trimmed from the key, the remains will be the object name.
	// Example:
	// Say we have the following keys:
	// pref/object
	// pref/obj/another
	// If we specify prefix="pref/obj" (both keys will be listed) then basePath="pref/" and the trim result
	// for the keys will be:
	// object
	// obj/another
```

johnnyaug (approved):

PR #3117, opened 2022-03-25
3109 lakectl ingest tests

itai-david:
Verification tests for `lakectl ingest` command output format


nessie/lakectl_test.go
```diff
@@ -295,3 +301,52 @@ func TestLakectlAuthUsers(t *testing.T) {
...
+	// location. This is done as the `lakectl ingest` commad runs with local aws settings and can only access the
+	// corresponding region
+
+	s3Client, err := store.GetS3Client()
+	require.NoError(t, err, "Failed to initialize S3 session")
+
+	_, err = s3Client.CreateBucket(&s3.CreateBucketInput{Bucket: aws.String(tempBucketName)})
+	require.NoError(t, err, "Failed to create bucket", tempBucketName)
+
+	defer cleanS3Bucket(s3Client, tempBucketName)
```
itaiad200, line 325: Creating a bucket per run can blow up if deleting the bucket doesn't work as expected, or the test is stopped half way through.

I suggest instead to use a predefined bucket and use prefixes for the ingestion and the repo part.
For example `s3://test-bucket-for-lakectl-ingestion/` where the prefix ingestion `ingestion-raw-run-xxxxx` and storage namespace for the matching repo `ingestion-repo-run-xxxxx`. Then this bucket can have lifecycle policies for cleanups (let's say 2 weeks) and you don't have to worry about dangling buckets.

You can also use the bucket `s3://nessie-system-testing-data` for the raw to-be-ingested data. That bucket data isn't deleted and you won't have to upload these files yourself.

nessie/lakectl_test.go
```diff
@@ -295,3 +301,52 @@ func TestLakectlAuthUsers(t *testing.T) {
 	// Cleanup
 	RunCmdAndVerifySuccess(t, Lakectl()+" auth users delete --id "+userName, false, "User deleted successfully\n", vars)
 }
+
+func TestLakectlIngest(t *testing.T) {
```
itaiad200, line 305: This test can't work with Azure and GCP.

itaiad200 (changes requested):
Thanks for this much needed tests!

I do think that you should use predefined bucket and data, instead of creating buckets for each tests. Also - you need to disable the test until we support ingest for gcp and azure.

N-o-Z (approved):
For me it looks good, I don't really have anything to add to what @itaiad200 already mentioned

nessie/lakectl_test.go
itai-david: Correct
Added the adapter type to nessie via env var. Test is skipped if adapter is not S3

.github/workflows/nessie.yaml
```diff
@@ -724,6 +727,7 @@ jobs:
           LAKEFS_BLOCKSTORE_TYPE: azure
           LAKEFS_BLOCKSTORE_AZURE_STORAGE_ACCOUNT: ${{ secrets.LAKEFS_BLOCKSTORE_AZURE_STORAGE_ACCOUNT }}
           LAKEFS_BLOCKSTORE_AZURE_STORAGE_ACCESS_KEY: ${{ secrets.LAKEFS_BLOCKSTORE_AZURE_STORAGE_ACCESS_KEY }}
+          NESSIE_BLOCKSTORE_TYPE: azure
```
itaiad200, line 730: Can this be different than `LAKEFS_BLOCKSTORE_TYPE`? Why not use the same env var?

itaiad200 (approved):
Approved with a single minor comment, thx!

.github/workflows/nessie.yaml
itai-david: It is used inside `nessiecontext, and `viper` only reads variables with `NESSIE_` prefix

PR #3120, opened 2022-03-27
Fix: Modify hooks to run on committed data only

N-o-Z:
Hooks data (action files) should be taken from committed data only and not from unstaged data.
Modify hook records source ref to the relevant commit ID:

- in Merge action
- - provide the commit ID of the destination branch
- in Commit action
- - Provide the commit ID of the base commit


N-o-Z:
Decided to keep implementation as is

PR #3122, opened 2022-03-27
Adding email functionality to Lakefs

### LINKED ISSUE #3103: Support email and smtp in Lakefs
Add email functionality to lakefs in order to support actions that require email, such as reset password or invite user.

shimi9276:
Closes #3103

### DIFF
--- a/cmd/lakefs/cmd/run.go
+++ b/cmd/lakefs/cmd/run.go
@@ -30,6 +30,7 @@ import (
 	"github.com/treeverse/lakefs/pkg/catalog"
 	"github.com/treeverse/lakefs/pkg/config"
 	"github.com/treeverse/lakefs/pkg/db"
+	"github.com/treeverse/lakefs/pkg/email"
 	"github.com/treeverse/lakefs/pkg/gateway"
 	"github.com/treeverse/lakefs/pkg/gateway/multiparts"
 	"github.com/treeverse/lakefs/pkg/gateway/sig"
@@ -200,7 +201,8 @@ var runCmd = &cobra.Command{
 		done := make(chan bool, 1)
 		quit := make(chan os.Signal, 1)
 		signal.Notify(quit, syscall.SIGINT, syscall.SIGTERM)
-		email, _ := cfg.GetEmailParams()
+		emailParams, _ := cfg.GetEmailParams()
+		emailer := email.NewEmailer(emailParams)
 
 		apiHandler := api.Serve(
 			cfg,
@@ -215,7 +217,7 @@ var runCmd = &cobra.Command{
 			actionsService,
 			auditChecker,
 			logger.WithField("service", "api_gateway"),
-			email,
+			*emailer,
 			cfg.GetS3GatewayDomainNames(),
 		)
 
--- a/cmd/lakefs/cmd/run.go
+++ b/cmd/lakefs/cmd/run.go
@@ -203,7 +203,6 @@ var runCmd = &cobra.Command{
 		signal.Notify(quit, syscall.SIGINT, syscall.SIGTERM)
 		emailParams, _ := cfg.GetEmailParams()
 		emailer := email.NewEmailer(emailParams)
-
 		apiHandler := api.Serve(
 			cfg,
 			c,
@@ -217,7 +216,7 @@ var runCmd = &cobra.Command{
 			actionsService,
 			auditChecker,
 			logger.WithField("service", "api_gateway"),
-			*emailer,
+			emailer,
 			cfg.GetS3GatewayDomainNames(),
 		)
 
--- a/cmd/lakefs/cmd/run.go
+++ b/cmd/lakefs/cmd/run.go
@@ -200,6 +200,7 @@ var runCmd = &cobra.Command{
 		done := make(chan bool, 1)
 		quit := make(chan os.Signal, 1)
 		signal.Notify(quit, syscall.SIGINT, syscall.SIGTERM)
+		email, _ := cfg.GetEmailParams()
 
 		apiHandler := api.Serve(
 			cfg,
@@ -214,6 +215,7 @@ var runCmd = &cobra.Command{
 			actionsService,
 			auditChecker,
 			logger.WithField("service", "api_gateway"),
+			email,
 			cfg.GetS3GatewayDomainNames(),
 		)
 
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -102,6 +102,11 @@ This reference uses `.` to denote the nesting of values.
   `max_range_size_bytes`).
 + `committed.sstable.memory.cache_size_bytes` (`int` : `200_000_000`) - maximal size of
   in-memory cache used for each SSTable reader.
++ `email.smtp_host` (`string` :   ) - A string representing the URL of the SMTP host.
++ `email.port` (`int` :   ) - An integer representing the port of the SMTP service (465, 587, 993, 25 are some standard ports)
++ `email.username` (`string` :   ) - A string representing the username of the specific account at the SMTP. It's recommended to provide this value at runtime from a secret vault of some sort.
++ `email.password` (`string` :   ) - A string representing the password of the account. It's recommended to provide this value at runtime from a secret vault of some sort.
++ `email.sender` (`string` :   ) - A string representing the email account which is set as the sender.
 * `gateways.s3.domain_name` `(string : "s3.local.lakefs.io")` - a FQDN
   representing the S3 endpoint used by S3 clients to call this server
   (`*.s3.local.lakefs.io` always resolves to 127.0.0.1, useful for
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -102,11 +102,11 @@ This reference uses `.` to denote the nesting of values.
   `max_range_size_bytes`).
 + `committed.sstable.memory.cache_size_bytes` (`int` : `200_000_000`) - maximal size of
   in-memory cache used for each SSTable reader.
-+ `email.smtp_host` (`string` :   ) - A string representing the URL of the SMTP host.
++ `email.smtp_host` `(string)` - A string representing the URL of the SMTP host.
 + `email.port` (`int` :   ) - An integer representing the port of the SMTP service (465, 587, 993, 25 are some standard ports)
-+ `email.username` (`string` :   ) - A string representing the username of the specific account at the SMTP. It's recommended to provide this value at runtime from a secret vault of some sort.
-+ `email.password` (`string` :   ) - A string representing the password of the account. It's recommended to provide this value at runtime from a secret vault of some sort.
-+ `email.sender` (`string` :   ) - A string representing the email account which is set as the sender.
++ `email.username` `(string)` - A string representing the username of the specific account at the SMTP. It's recommended to provide this value at runtime from a secret vault of some sort.
++ `email.password` `(string)` - A string representing the password of the account. It's recommended to provide this value at runtime from a secret vault of some sort.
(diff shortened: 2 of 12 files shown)


pkg/config/template.go
```diff
@@ -177,4 +177,11 @@ type configuration struct {
 		AuditCheckInterval time.Duration `mapstructure:"audit_check_interval"`
 		AuditCheckURL      string        `mapstructure:"audit_check_url"`
 	} `mapstructure:"security"`
+	Email struct {
+		SMTPHost string `mapstructure:"smtp_host"`
+		Port     int    `mapstructure:"port"`
+		Username string `mapstructure:"username"`
+		Password string `mapstructure:"password"`
+		Sender   string `mapstructure:"sender"`
```
guy-har, line 185: renaming using `mapstructure` ins't needed here

pkg/email/send_email.go
```diff
@@ -0,0 +1,44 @@
...
+	msg.SetHeader("Subject", e.Subject)
+	msg.SetBody("text/html", e.Body)
+	if len(e.AttachmentFilePath) > 0 {
+		for _, f := range e.AttachmentFilePath {
+			msg.Attach(f)
+		}
+	}
+	d := gomail.NewDialer(e.SMTPHost, e.Port, e.Username, e.Password)
+	return d.DialAndSend(msg)
+}
```
guy-har, line 36: From what I understand, to send an email I would need to initiate a `params.Email` struct with all the fields (such as `SMTPHost`, `Port`, etc...) and then call SendEmail with the struct.

I would expect something like 
```
NewMailer(SmtpHost, Port, Username, Password, Sender) &Mailer
```
And then send mail using 
```
mailer.mail(Receivers, Subject, Body)
```
That way we could initiate lakeFS with some Sender and then sending a mail all I need is Receivers, Subject and body

So for example, to send an email from the controller I would need to run something like
```
c.mailer.mail(receivers, subject, body)
```

cmd/lakefs/cmd/run.go
```diff
@@ -183,6 +183,7 @@ var runCmd = &cobra.Command{
 			interval := cfg.GetSecurityAuditCheckInterval() + jitter
 			auditChecker.StartPeriodicCheck(ctx, interval, logger)
 		}
+		// Wiring required for sending emails once it's decided where exactly it will be call from
```
guy-har, line 186: what about this part? not implementing yet?
If so, leave it as a TODO (`// TODO(shimi): `)

guy-har (changes requested):
Great job

I added some comments
The main reason for requesting changes is for adding the `static` to `go.mod`

docs/reference/configuration.md
```diff
@@ -102,6 +102,11 @@ This reference uses `.` to denote the nesting of values.
   `max_range_size_bytes`).
 + `committed.sstable.memory.cache_size_bytes` (`int` : `200_000_000`) - maximal size of
   in-memory cache used for each SSTable reader.
++ `email.smtp_host` (`string` :   ) - A string represnting the URL of the smtp host.
```
ortz, line 105:
```suggestion
+ `email.smtp_host` (`string` :   ) - A string representing the URL of the SMTP host.
```

docs/reference/configuration.md
```diff
@@ -102,6 +102,11 @@ This reference uses `.` to denote the nesting of values.
...
++ `email.port` (`int` :   ) - A intiger represnting the port of the SMTP service (465, 587, 993, 25 are some standerd ports)
```
ortz, line 106:
```suggestion
+ `email.port` (`int` :   ) - An integer representing the port of the SMTP service (465, 587, 993, 25 are some standard ports)
```

docs/reference/configuration.md
```diff
@@ -102,6 +102,11 @@ This reference uses `.` to denote the nesting of values.
...
++ `email.username` (`string` :   ) - A string represnting the username of the specific account at the SMTP. It's recommended to provide this value at runtime from a secret vault of some sort.
```
ortz, line 107:
```suggestion
+ `email.username` (`string` :   ) - A string representing the username of the specific account at the SMTP. It's recommended to provide this value at runtime from a secret vault of some sort.
```

docs/reference/configuration.md
```diff
@@ -102,6 +102,11 @@ This reference uses `.` to denote the nesting of values.
...
++ `email.password` (`string` :   ) - A string represnting the password of the account. It's recommended to provide this value at runtime from a secret vault of some sort.
```
ortz, line 108:
```suggestion
+ `email.password` (`string` :   ) - A string representing the password of the account. It's recommended to provide this value at runtime from a secret vault of some sort.
```

docs/reference/configuration.md
```diff
@@ -102,6 +102,11 @@ This reference uses `.` to denote the nesting of values.
...
++ `email.sender` (`string` :   ) - A string represnting the email account of which is set as the sender.
```
ortz, line 109:
```suggestion
+ `email.sender` (`string` :   ) - A string representing the email account which is set as the sender.
```

pkg/email/send_email.go
```diff
@@ -0,0 +1,44 @@
+package email
+
+import (
+	"github.com/treeverse/lakefs/pkg/email/params"
+	gomail "gopkg.in/gomail.v2"
+)
+
+func NewEmail(e params.Email, err error) (*params.Email, error) {
```
ortz, line 8: What's the purpose of this function?

ortz (changes requested):
hey @shimi9276 , I had some comments, most of them are semantics and I noticed that @guy-har already asked some functional questions...
I did have one regarding a function I couldn't find a reference to.

thanks!

pkg/config/template.go
shimi9276: It's needed for SMTPHost as it contains an underscore. From my understanding of the docs, it is preferable to add it even in the case it might not be needed.

pkg/email/send_email.go
```diff
@@ -0,0 +1,44 @@
+package email
+
+import (
+	"github.com/treeverse/lakefs/pkg/email/params"
+	gomail "gopkg.in/gomail.v2"
```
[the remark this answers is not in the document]
shimi9276: Oh. Will remove it if that is the case.

pkg/email/send_email.go
shimi9276: Such a struct is automatically created during the init of the Lakefs config file, so the assumption is that such a struct exists and is initialized already and all that is needed is to add params such as body or subject, and for that, there is the function of `InsertAdditionalParamsToEmail` to give that capability.

pkg/email/send_email.go
```diff
@@ -0,0 +1,44 @@
...
+	return d.DialAndSend(msg)
+}
+
+func InsertAdditionalParamsToEmail(e params.Email, receivers []string, subject string, body string, attachmentFilePath ...string) *params.Email {
+	e.Receivers = receivers
+	e.Subject = subject
+	e.Body = body
+	e.AttachmentFilePath = attachmentFilePath
+	return &e
+}
```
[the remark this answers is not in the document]
shimi9276: See the above. I do agree though that it could be a method of the `params.Email` function.

pkg/email/send_email.go
shimi9276: The idea is the basic email params (i.e. SMTP, port, password, etc) should be initialized on `run`, this is done using this function as it takes in the config email struct and creates a new one that can be passed on. Please see the above comment regarding the initilization, and that some of the wiring will be done in a different PR as it is to be determined where it should reside.

ortz: I'm a bit confused on which functions will trigger this one if we have the `SendEmail` function.
Also, it looks like this function is only storing the email struct information, within a new but same struct and returns it.

We sometimes use "NewStruct" functions in order to simulate object creation (from OOP) using the receiver function, but this is not the case here...
Maybe I miss some context or the rest of the functions to understand it better, let's discuss this tomorrow during our PR review.

docs/reference/configuration.md
```diff
@@ -102,6 +102,11 @@ This reference uses `.` to denote the nesting of values.
   `max_range_size_bytes`).
 + `committed.sstable.memory.cache_size_bytes` (`int` : `200_000_000`) - maximal size of
   in-memory cache used for each SSTable reader.
++ `email.smtp_host` (`string` :   ) - A string representing the URL of the SMTP host.
++ `email.port` (`int` :   ) - An integer representing the port of the SMTP service (465, 587, 993, 25 are some standard ports)
++ `email.username` (`string` :   ) - A string representing the username of the specific account at the SMTP. It's recommended to provide this value at runtime from a secret vault of some sort.
++ `email.password` (`string` :   ) - A string representing the password of the account. It's recommended to provide this value at runtime from a secret vault of some sort.
++ `email.sender` (`string` :   ) - A string representing the email account which is set as the sender.
```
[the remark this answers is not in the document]
shimi9276: Fixed

pkg/api/controller.go
```diff
@@ -77,6 +78,7 @@ type Controller struct {
 	Actions               actionsHandler
 	AuditChecker          AuditChecker
 	Logger                logging.Logger
+	Email                 email.Email
```
guy-har, line 81: Consider using Emailer or Mailer (just a suggestion)

guy-har (approved):

pkg/email/send_email.go
```diff
@@ -0,0 +1,38 @@
...
+	Sender   string
+}
+
+type EmailParams struct {
+	SMTPHost string
+	Port     int
+	Username string
+	Password string
+	Sender   string
+}
```
ortz, line 21: Why do we need two structs representing the exact same thing and the NewEmail is only passing parameters between the two structs?

pkg/email/send_email.go
```diff
@@ -0,0 +1,38 @@
...
+
+func NewEmailer(e EmailParams) Emailer {
+	return Emailer(e)
+}
+
+func (e Emailer) SendEmail(receivers []string, subject string, body string, attachmentFilePath ...string) error {
```
ortz, line 27:
```suggestion
func (e Emailer) SendEmail(receivers []string, subject string, body string, attachmentFilePath []string) error {
```
I think it'd be better to use array of string for the attachmentFilePath compared the never-endning fiesca of parameters, and it basically limits our signature modification for future use, due to the fact the parameters are not explicit anymore.

ortz (approved):

pkg/email/send_email.go
shimi9276: One holds config email params and one holds the email functionality. Its kind of weird a config struct should hold functionality on it.

pkg/email/send_email.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"gopkg.in/gomail.v2"
+)
+
+type Emailer struct {
+	SMTPHost string
+	Port     int
+	Username string
+	Password string
+	Sender   string
```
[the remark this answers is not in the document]
shimi9276: Done.

pkg/email/send_email.go
```diff
@@ -0,0 +1,47 @@
...
+func (e Emailer) SendEmail(receivers []string, subject string, body string, attachmentFilePath []string) error {
+	msg := gomail.NewMessage()
+	msg.SetHeader("From", e.Sender)
+	msg.SetHeader("To", receivers...)
+	msg.SetHeader("Subject", subject)
+	msg.SetBody("text/html", body)
+	for _, f := range attachmentFilePath {
+		msg.Attach(f)
+	}
+	// d := gomail.NewDialer(e.SMTPHost, e.Port, e.Username, e.Password)
```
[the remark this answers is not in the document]

PR #3130, opened 2022-03-28
Feature/add additional hook locations

N-o-Z:
Add hook locations:
Pre/Post - Create/Delete - Branch/Tag
#3066


pkg/graveler/graveler.go
```diff
@@ -830,10 +830,53 @@ func (g *Graveler) CreateBranch(ctx context.Context, repositoryID RepositoryID,
...
+		}
+	}
+
 	err = g.RefManager.CreateBranch(ctx, repositoryID, branchID, newBranch)
 	if err != nil {
 		return nil, fmt.Errorf("set branch '%s' to '%s': %w", branchID, newBranch, err)
 	}
+
+	postRunID := NewRunID()
+	err = g.hooks.PostCreateBranchHook(ctx, HookRecord{
```
nopcoder, line 864: For all PostXXX Pass the PreRunID

pkg/graveler/graveler.go
```diff
@@ -891,11 +934,117 @@ func (g *Graveler) GetTag(ctx context.Context, repositoryID RepositoryID, tagID
...
+	// Create tag
+	err = g.RefManager.CreateTag(ctx, repositoryID, tagID, commitID)
+	if err != nil {
+		return err
+	}
+
+	// Run post-hook
+	postRunID := NewRunID()
+	err = g.hooks.PostCreateTagHook(ctx, HookRecord{
+
```
nopcoder, line 976: remove whitespace

pkg/graveler/graveler.go
```diff
@@ -891,11 +934,117 @@ func (g *Graveler) GetTag(ctx context.Context, repositoryID RepositoryID, tagID
 }
 
 func (g *Graveler) CreateTag(ctx context.Context, repositoryID RepositoryID, tagID TagID, commitID CommitID) error {
-	return g.RefManager.CreateTag(ctx, repositoryID, tagID, commitID)
+	// Check that Tag doesn't exist before running hook
```
nopcoder, line 937: Just to make the API more expected, fetch the repo first - so you will get repository not found and not tag not found.

pkg/graveler/graveler.go
```diff
@@ -891,11 +934,117 @@ func (g *Graveler) GetTag(ctx context.Context, repositoryID RepositoryID, tagID
...
+	})
+	if err != nil {
+		return &HookAbortError{
+			EventType: EventTypePreDeleteTag,
+			RunID:     preRunID,
+			Err:       err,
+		}
+	}
+
+	// Create tag
```
nopcoder, line 1025: one of the reasons I don't like code comments when the code is clear

pkg/graveler/graveler.go
```diff
@@ -931,6 +1080,9 @@ func (g *Graveler) ListBranches(ctx context.Context, repositoryID RepositoryID)
 }
 
 func (g *Graveler) DeleteBranch(ctx context.Context, repositoryID RepositoryID, branchID BranchID) error {
+	var preRunID string
+	var storageNamespace StorageNamespace
+	var commitID CommitID
```
nopcoder, line 1085:
```suggestion
	var (
	    preRunID string
	    storageNamespace StorageNamespace
	    commitID CommitID
	)
```

nopcoder (changes requested):

pkg/graveler/graveler.go
N-o-Z: @nopcoder I think this might be problematic since RunID is a primary key in the actions_runs table

pkg/actions/action.go
```diff
@@ -107,6 +121,35 @@ func (a *Action) Validate() error {
...
+					if strings.Contains(string(k), "tag") {
+						return fmt.Errorf("'branches' is not supported in tag event types. %w", ErrInvalidEventParameter)
+					}
+				}
+			default:
+				{
+					return fmt.Errorf("should not reach! %s, %w", field, ErrInvalidEventParameter)
+				}
+			}
+		}
```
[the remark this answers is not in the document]
N-o-Z: In the future we will probably want to add some filter parameters for Tags events as well which will be illegal for Branch based events. I wanted to keep it in a switch format as to reflect this logic and that in the future the validation will be done also for non-tag events. If you think it's a big issue I can just document it and change it to an if statement

pkg/actions/action.go
```diff
@@ -65,15 +67,23 @@ var (
 	reName   = regexp.MustCompile(`^\w[\w\-. ]+$`)
 	reHookID = regexp.MustCompile(`^[_a-zA-Z][\-_a-zA-Z0-9]{1,255}$`)
 
-	ErrInvalidAction    = errors.New("invalid action")
-	ErrInvalidEventType = errors.New("invalid event type")
+	ErrInvalidAction         = errors.New("invalid action")
+	ErrInvalidEventParameter = errors.New("invalid event parameter")
 )
 
 var supportedEvents = map[graveler.EventType]bool{
```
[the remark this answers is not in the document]
N-o-Z: @nopcoder Can you explain the advantage of using a function with a switch case over this map?

nessie/hooks_test.go
```diff
@@ -92,6 +194,14 @@ func TestHooksSuccess(t *testing.T) {
 	parseAndUpload(t, ctx, repo, branch, actionPreCommitYaml, "action-pre-commit", "testing_pre_commit")
 	parseAndUpload(t, ctx, repo, branch, actionPostCommitYaml, "action-post-commit", "testing_post_commit")
 	parseAndUpload(t, ctx, repo, branch, actionPostMergeYaml, "action-post-merge", "testing_post_merge")
+	parseAndUpload(t, ctx, repo, branch, actionPreCreateTagYaml, "action-pre-create-tag", "testing_pre_create_tag")
```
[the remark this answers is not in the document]
N-o-Z: I think there's an advantage to uploading all the files for each test. We get the extra value of verifying that the wrong actions are not triggered. WDYT?

pkg/graveler/hooks_handler.go
```diff
@@ -26,13 +34,22 @@ type HookRecord struct {
 	Commit           Commit
 	CommitID         CommitID
 	PreRunID         string
+	TagID            TagID
```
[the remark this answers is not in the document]
N-o-Z: Suggest we document the HookRecord struct and open an issue to examine the possibility to create specific record types. What do you think?

pkg/actions/action.go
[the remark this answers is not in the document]
nopcoder: First, both will work and both have pros and cons. The difference is runtime vs compile time - there are cases where the compiled option have more options to optimize the lookup.

pkg/actions/action.go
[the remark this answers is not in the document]
nopcoder: You can use type assertion to check Branch. Tools and compiler doesn't see "Branch" as the field Branch - this is why the static type assertion is preferred here.
Would go with reflection when the type you check is dynamic, or your type you validate is known only at run time.

nessie/hooks_test.go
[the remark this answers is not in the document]
nopcoder: Guess it also depends on the current written tests - set up a more data in the env sounds good as long as it doesn't add noise for some tests.
My comment was just to have the above as part of looping over the embedded files - as you did :)

nessie/hooks_test.go
```diff
@@ -190,26 +170,193 @@ func TestHooksSuccess(t *testing.T) {
...
+	t.Helper()
 	// render actions based on templates
 	docData := struct {
 		URL string
 	}{
 		URL: server.BaseURL(),
 	}
 
-	var doc bytes.Buffer
+	fs.WalkDir(actions, ".", func(path string, d fs.DirEntry, err error) error {
```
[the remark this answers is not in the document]
N-o-Z: I'm not sure it will help since we will still have '.' and '..' in the dir entries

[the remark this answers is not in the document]
N-o-Z: I'm not sure it will help since we will still have '.' and '..' in the dir entries

pkg/actions/action.go
```diff
@@ -107,6 +121,35 @@ func (a *Action) Validate() error {
 	return nil
 }
 
+func validateOnParameters(on map[graveler.EventType]*ActionOn) error {
```
itaiad200, line 129: must run `gofmt`

pkg/actions/action.go
```diff
@@ -87,9 +102,13 @@ func (a *Action) Validate() error {
 		return fmt.Errorf("'on' is required: %w", ErrInvalidAction)
 	}
 	for o := range a.On {
-		if !supportedEvents[o] {
+		if !isEventSupported(o) {
 			return fmt.Errorf("event '%s' is not supported: %w", o, ErrInvalidAction)
 		}
```
itaiad200, line 107: I think this belongs in `validateOnParameters`

pkg/actions/action.go
```diff
@@ -87,9 +102,13 @@ func (a *Action) Validate() error {
...
+		err := validateOnParameters(a.On)
```
itaiad200, line 108: You iterate on `a.On` and then pass the entire struct `a.On` with every iteration?

pkg/actions/service.go
```diff
@@ -453,6 +453,42 @@ func (s *Service) PostMergeHook(ctx context.Context, record graveler.HookRecord)
...
+
+func (s *Service) PreDeleteBranchHook(ctx context.Context, record graveler.HookRecord) error {
+	return s.Run(ctx, record)
+}
+
+func (s *Service) PostDeleteBranchHook(_ context.Context, record graveler.HookRecord) error {
+	s.asyncRun(record)
+	return nil
+}
+
```
itaiad200, line 491: I think this demonstrates quite well why we need a refactor here..

pkg/graveler/graveler.go
```diff
@@ -891,11 +935,115 @@ func (g *Graveler) GetTag(ctx context.Context, repositoryID RepositoryID, tagID
...
 
 func (g *Graveler) CreateTag(ctx context.Context, repositoryID RepositoryID, tagID TagID, commitID CommitID) error {
-	return g.RefManager.CreateTag(ctx, repositoryID, tagID, commitID)
+	repo, err := g.RefManager.GetRepository(ctx, repositoryID)
+	if err != nil {
+		return fmt.Errorf("get repository: %w", err)
+	}
+	storageNamespace := repo.StorageNamespace
+
+	// Check that Tag doesn't exist before running hook
```
itaiad200, line 944: I would add 
```suggestion
	// Sanity check that Tag doesn't exist before running hook.
	// A race may still occur .
```

pkg/graveler/graveler.go
```diff
@@ -891,11 +935,115 @@ func (g *Graveler) GetTag(ctx context.Context, repositoryID RepositoryID, tagID
...
+	_, err = g.RefManager.GetTag(ctx, repositoryID, tagID)
+	if !errors.Is(err, ErrTagNotFound) {
```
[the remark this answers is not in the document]
itaiad200: Let's add a test for it too.

pkg/graveler/hooks_handler.go
```diff
@@ -10,29 +10,59 @@ import (
...
 	PreMergeHook(ctx context.Context, record HookRecord) error
 	PostMergeHook(ctx context.Context, record HookRecord) error
+	PreCreateTagHook(ctx context.Context, record HookRecord) error
+	PostCreateTagHook(ctx context.Context, record HookRecord) error
+	PreDeleteTagHook(ctx context.Context, record HookRecord) error
+	PostDeleteTagHook(ctx context.Context, record HookRecord) error
+	PreCreateBranchHook(ctx context.Context, record HookRecord) error
+	PostCreateBranchHook(ctx context.Context, record HookRecord) error
+	PreDeleteBranchHook(ctx context.Context, record HookRecord) error
+	PostDeleteBranchHook(ctx context.Context, record HookRecord) error
```
itaiad200, line 65: Why are the `Post` returning errors when they can't fail?

pkg/graveler/hooks_handler.go
```diff
@@ -10,29 +10,59 @@ import (
...
+// EventType
+// RepositoryID
+// StorageNamespace
+// SourceRef - The reference from which the actions files are read from
+// Event specific fields:
+// BranchID - Relevant for all event types except tags. For merge events this will be the ID of the destination branch
+// Commit - Relevant only for commit and merge events. In both it will contain the new commit data created from the operation
+// CommitID - Not relevant in delete branch. In commit and merge will not exist in pre-action. In post actions will contain the new commit ID
+// PreRunID - Exists only in post actions. Contains the ID of the pre-action associated with this post-action
+// TagID - Exists only in tag actions.
```
itaiad200, line 39: Thanks for the documentation!
Usually field comments go above the actual field.

itaiad200 (changes requested):
Thanks for the great hooks additions alongside many good and much needed refactoring. Unittests are also pretty impressive. I wonder if you spin up a hooks server and tested it yourself? If so, might be worth to share the code too, we have a hooks repo for it.

Other than that, few comments that needs fixing.

nopcoder (approved):
looks good - minor comments about the code style

pkg/graveler/hooks_handler.go
N-o-Z: In this case - do you think it's better to leave it like this or to break the documentation per field?

itaiad200: I think it's better to have per field.

pkg/actions/action.go
```diff
@@ -107,6 +126,29 @@ func (a *Action) Validate() error {
...
+func validateOnParameters(on map[graveler.EventType]*ActionOn) error {
+	for k, v := range on {
+		if v == nil {
+			continue
+		}
+		switch {
+		// Add a case for any additional field added to ActionOn struct
+		case len(v.Branches) > 0:
+			{
+				if strings.HasSuffix(string(k), "-tag") {
```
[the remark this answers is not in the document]
N-o-Z: How do you feel about deferring this until we get more requirements for parameters and then perhaps we will have a better understanding of how to approach this?

N-o-Z:
> Thanks for the great hooks additions alongside many good and much needed refactoring. Unittests are also pretty impressive. I wonder if you spin up a hooks server and tested it yourself? If so, might be worth to share the code too, we have a hooks repo for it.
> 
> Other than that, few comments that needs fixing.

Thanks! I actually used the hooks server to test some of the changes but mostly I relied on unit/system test

itaiad200 (approved):
Great work on this!

N-o-Z:
Thank you @nopcoder @itaiad200 for the professional and thorough review!
Waiting on #3142 to merge together

PR #3155, opened 2022-04-02
Fix create repo min character limit issue

### LINKED ISSUE #2967: webui - Create A New Repository declares repo ID should be atleast 2 chars but does not allow less than 3
image

The regex at webui/src/lib/components/repositoryCreateForm.jsx, line 18 actually requires 3-63 chars
```const repoValidityRegex = /^[a-z0-9][a-z0-9-]{2,62}$/;```
while the warning at line 77 states `Min 2 characters`

Should fix either one

### PR DESCRIPTION
**Pull request by DataDavD**
Currently, there exists a bug upon which the Create A New Repository
form states that repo ID's must be at least 2 characters, but in
actuality it does not allow less than 3. We update the regex string so
that the minimum is now 2 characters.

Closes #2967

### DIFF
--- a/webui/src/lib/components/repositoryCreateForm.jsx
+++ b/webui/src/lib/components/repositoryCreateForm.jsx
@@ -74,7 +74,7 @@ export const RepositoryCreateForm = ({ config, onSubmit, onCancel, error = null,
                     <Form.Control type="text" autoFocus ref={repoNameField} onChange={onRepoNameChange}/>
                     {repoValid === false &&
                     <Form.Text className="text-danger">
-                        Min 2 characters. Only lowercase alphanumeric characters and {'\'-\''} allowed.
+                        Min 3 characters. Only lowercase alphanumeric characters and {'\'-\''} allowed.
                     </Form.Text>
                     }
                 </Col>

### CODE REVIEW

File: webui/src/lib/components/repositoryCreateForm.jsx [hunk 1]
```diff
@@ -15,7 +15,7 @@ const DEFAULT_BLOCKSTORE_VALIDITY_REGEX = new RegExp(`^s3://`);
 
 export const RepositoryCreateForm = ({ config, onSubmit, onCancel, error = null, inProgress = false, sm = 6 }) => {
     const fieldNameOffset = 3;
-    const repoValidityRegex = /^[a-z0-9][a-z0-9-]{2,62}$/;
+    const repoValidityRegex = /^[a-z0-9][a-z0-9-]{1,62}$/;
```
nopcoder, line 18: I think the pattern was ok - based on "Bucket names must be between 3 (min) and 63 (max) characters long" like s3.

nopcoder (changes requested):

webui/src/lib/components/repositoryCreateForm.jsx
DataDavD: Thanks @nopcoder I didn't realize original naming req's were based on s3 naming req's. Totally makes sense now. I'll update, rebase, and push corrections. Thank you!

DataDavD: @nopcoder I was looking over s3 bucket naming rules and noticed s3 buckets must also adhere to these 2 specific requirements
- `Bucket names must not start with the prefix xn--.`
- `Bucket names must not end with the suffix -s3alias`

I noticed with the current regex iteration, users are able to make repo names that match these restrictions. Should I improve the regex to avoid these 2 specific prefix and suffix restrictions?

DataDavD: FYI code has been updated so that wording now states that 3 characters is the minimum character limit for Repository ID.

nopcoder: Thanks, @DataDavD, I would open a different issue for the extra validation as the enforcement should be first implemented on the backend. 
Compatibility is part of S3 gateway, if all S3 clients or SDKs do not validate 'xn--' and '-s3alias' - lakefs can support it, as the limitation is from the underlying storage.

nopcoder (approved):
lgtm!

DataDavD:
thanks @nopcoder!!! Glad to get another lakeFS PR completed, regardless of how small it may be 😆

PR #3167, opened 2022-04-04
Adding password reseting functionality

### LINKED ISSUE #3175: Avoiding an email flooding attack
There is a trivial email-flooding attack on this implementation.  If I know ariels@bigcorp.com has an account, I can flood the lakeFS server with "forgot password" requests.  This has a >>10x amplification factor (I send >> 10x fewer bytes than lakeFS will, possible even 100x fewer).

What can I attack with this?

* ariels@bigcorp.com's mailbox.
* The operator (who will get notified of abuse)
* lakeFS (which may get cut off from emailing by the email provider)
* lakeFS (which may end up being marked as spam by its users)

Typically features such as this one have a rate-limiter built into them.  We can either use an existing rate-limiter or write our own -- maybe initially even just in memory.  But something has to prevent unlimited password reset requests.

_Originally posted by @arielshaqed in https://github.com/treeverse/lakeFS/issues/3167#issuecomment-1090872018_

shimi9276:
Closes #3175


api/swagger.yml
```diff
@@ -539,6 +539,8 @@ components:
           type: string
         email:
           type: string
+        password:
+          type: string
```
guy-har, line 543:
```suggestion

```
We don't want to return this information about the user

api/swagger.yml
```diff
@@ -591,6 +601,24 @@ components:
...
+      
+    SetPassword:
+      type: object
+      properties:
+        token:
+          description: a new password for the given email
+          type: string
+        newPassword:
+          description: auth token for password reset
+          type: string
```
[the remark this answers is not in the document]
shimi9276: True. Fixed

api/swagger.yml
```diff
@@ -1063,6 +1091,50 @@ paths:
         default:
           $ref: "#/components/responses/ServerError"
 
+  /auth/set_password:
```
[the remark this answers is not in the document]
shimi9276: Done

api/swagger.yml
```diff
@@ -1063,6 +1091,50 @@ paths:
...
+        responses:
+          200:
+            description: successful reset
+          401:
+            $ref: "#/components/responses/Unauthorized"
+          default:
+            $ref: "#/components/responses/ServerError"
+ 
+  /auth/reset_password:
+      post:
```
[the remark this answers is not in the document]
shimi9276: Good point. We should indeed seperate. Fixed

api/swagger.yml
shimi9276: True. Added it during debugging and forgot to remove it. Fixed

pkg/api/controller.go
```diff
@@ -3006,6 +3010,48 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+	// TODO (@shimi9276) create template for sending the email with link for reset
+	err = c.Emailer.SendEmail([]string{body.Email}, token, token, []string{})
+	if err != nil {
+		writeError(w, http.StatusBadRequest, err)
+	}
+	writeResponse(w, http.StatusOK, err)
+}
+
+func (c *Controller) SetPassword(w http.ResponseWriter, r *http.Request, body SetPasswordJSONRequestBody) {
+	err := VerifyResetPasswordToken(c.Auth, body.Token)
```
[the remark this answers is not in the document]
shimi9276: Good point. The problem is though that we don't have a user ID when we create the token, only the email. This basically means we need to add the email to the token. Done

pkg/api/controller.go
```diff
@@ -3006,6 +3010,48 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		writeError(w, http.StatusForbidden, err)
+	}
+	u := model.User{}
+	err = u.UpdatePassword(body.NewPassword)
```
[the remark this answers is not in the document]
shimi9276: Exposed hashPassword (Tryed not to do so initially, but as it seems it makes construct an unnecessary struct).

pkg/api/login_handler.go
```diff
@@ -21,8 +21,12 @@ const (
 	JWTCookieName          = "access_token"
 )
 
-func GenerateJWT(secret []byte, userID int, issuedAt, expiresAt time.Time) (string, error) {
+// Be aware! The aud field should *only* be used for new applications! Otherwise it should be passed as an empty string, so it supports
```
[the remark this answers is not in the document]
shimi9276: Done, We are not limiting the aud field as of now, as it may come in handy in the future.

pkg/api/login_handler.go
```diff
@@ -21,8 +21,12 @@ const (
...
+// backward compatibility.
+
```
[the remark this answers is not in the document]
shimi9276: Done

pkg/api/login_handler.go
```diff
@@ -21,8 +21,12 @@ const (
...
+func GenerateJWT(aud string, secret []byte, userID int, issuedAt, expiresAt time.Time) (string, error) {
```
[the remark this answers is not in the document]
shimi9276: Done

pkg/auth/service.go
```diff
@@ -959,3 +960,11 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
 	// we're allowed!
 	return &AuthorizationResponse{Allowed: true}, nil
 }
+
+func (s *DBAuthService) UpdatePassword(ctx context.Context, email string, newPassword string) error {
```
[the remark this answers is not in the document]
shimi9276: Done

pkg/auth/service.go
```diff
@@ -959,3 +960,11 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
...
+	_, err := s.db.Transact(ctx, func(tx db.Tx) (interface{}, error) {
+		exec, err := tx.Exec(`UPDATE auth_users SET encrypted_password = $1 WHERE email = $2`, newPassword, email)
```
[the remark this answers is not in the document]
shimi9276: Done

arielshaqed (commented):
This PR has (major) security impact and I would like to review it carefully.

It is somewhat large for rapid review; apologies for the time it will take.  In order to save time, if you have already done some security review or analysis of the intended flow could you please link to it here?

THANKS!

pkg/api/controller.go
```diff
@@ -56,6 +56,10 @@ const (
 	setupStateNotInitialized = "not_initialized"
 
 	DefaultMaxDeleteObjects = 1000
+
+	ResetPasswordAudience = "password_reset"
+	tokenExpiryDuration   = 5 * time.Minute
+	loginAudiance         = ""
```
arielshaqed, line 62:
```suggestion
	loginAudience         = ""
```

Consistent spelling here is **critical**.  If I need to verify all audiences for a security audit, I will miss this audience.  Also, ideally spelling of this word would be correct in the comments below, too.

arielshaqed, line 62: Why do we also have `loginTokenAudience` defined separately and above?  A const that is defined with different names in two locations but must have equal values is a bug waiting to happen, and is probably worse than just using a literal.

Can you export the audience values from a single location?  Possibly even create a sub-package just for that (if `import` cycles are an issue).

pkg/api/controller.go
```diff
@@ -175,7 +179,9 @@ func (c *Controller) Login(w http.ResponseWriter, r *http.Request, body LoginJSO
 	secret := c.Auth.SecretStore().SharedSecret()
 	// user.Username will be different from username/access_key_id on
 	// LDAP login.  Use the stored value.
-	tokenString, err := GenerateJWT(secret, user.ID, loginTime, expires)
+	// audiance and email set here to be an empty string for backward compatibility
```
arielshaqed, line 182: I don't understand why.  The user has logged in, we have their email.  Why don't we allow them to use it?

Also can cause issues with users having 2 different identities depending on how they choose to login.  This is also relevant to audits conducted by the lakeFS admin, of course.

pkg/api/controller.go
```diff
@@ -3006,6 +3012,58 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
 	writeResponse(w, http.StatusOK, response)
 }
 
+func (c *Controller) sendResetPasswordEmail(email []string, token string) error {
+	return c.Emailer.SendEmail(email, token, token, []string{})
```
arielshaqed, line 3016: This *MUST* be logged, and *SHOULD PROBABLY* be rate-limited.  Otherwise you have an easy unauthenticated email flooding vector (a kind of [CWE 770](https://cwe.mitre.org/data/definitions/770.html).  By exploiting this an attacker can block the sending email at the provider, preventing (at least) legitimate users from resetting passwords, allowing an attacker to email-flood legitimate users, and creating a great deal of hassle for the lakeFS administrator.

The feature proposal must cover what is acceptable behaviour on this flow.

pkg/api/controller.go
```diff
@@ -3006,6 +3012,58 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+		writeError(w, http.StatusInternalServerError, err)
+		return
+	}
+	if strings.Compare(*user.Email, body.Email) != 0 {
+		writeError(w, http.StatusNotFound, "No such email")
+		return
+	}
+	secret := c.Auth.SecretStore().SharedSecret()
+	currentTime := time.Now()
+	token, err := GenerateJWT(secret, ResetPasswordAudience, auth.InvalidUserID, body.Email, currentTime, currentTime.Add(tokenExpiryDuration))
```
arielshaqed, line 3031: Why use the wrong user ID?  At the very least it should be incorporated into the to-be-signed part of the JWT, if not the returned token itself.  Otherwise if I time things right and we swap emails, I can take control of your account.  "Obviously" people would never do that, but for service accounts this is a valid concern.

pkg/api/controller.go
```diff
@@ -3006,6 +3012,58 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+	writeResponse(w, http.StatusOK, err)
+}
+
+func (c *Controller) SetPassword(w http.ResponseWriter, r *http.Request, body SetPasswordJSONRequestBody) {
+	claims, err := VerifyResetPasswordToken(c.Auth, body.Token)
+	if err != nil {
+		writeError(w, http.StatusForbidden, err)
+		return
+	}
+	if strings.Compare(claims.Id, body.Email) != 0 {
```
arielshaqed, line 3051: Can we use == here to compare strings?

pkg/api/controller.go
```diff
@@ -3006,6 +3010,48 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
 }
 
+const ResetPasswordAudience = "password reset request"
+const tokenExpiryTime = 5 // minutes
+
+func (c *Controller) RequestPasswordReset(w http.ResponseWriter, r *http.Request, body RequestPasswordResetJSONRequestBody) {
+	user, err := c.Auth.GetUserByEmail(r.Context(), body.Email)
+	if err != nil {
+		return
+	}
```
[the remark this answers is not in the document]
shimi9276: Fixed

pkg/api/controller.go
```diff
@@ -3006,6 +3010,48 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+	currentTime := time.Now()
+	token, err := GenerateJWT(ResetPasswordAudience, secret, -1, currentTime, currentTime.Add(time.Minute*tokenExpiryTime))
+	if err != nil {
+		return
+	}
+	// TODO (@shimi9276) create template for sending the email with link for reset
+	err = c.Emailer.SendEmail([]string{body.Email}, token, token, []string{})
+	if err != nil {
+		writeError(w, http.StatusBadRequest, err)
+	}
```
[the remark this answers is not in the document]
shimi9276: Fixed

pkg/api/controller.go
nopcoder: + there should be a follow-up PR that should also address human request, like we have in the demo.

api/swagger.yml
```diff
@@ -1063,6 +1089,60 @@ paths:
         default:
           $ref: "#/components/responses/ServerError"
 
+  /auth/password:
+      post:
+        tags:
+          - auth
+        operationId: setPassword
+        summary: update password for the given password
```
nopcoder, line 1097: Update user password by reset_password token

pkg/api/login_handler.go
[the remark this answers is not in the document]
shimi9276: Set to one constant and mentioned it in the comment.

pkg/auth/service.go
```diff
@@ -959,3 +960,14 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
 	// we're allowed!
 	return &AuthorizationResponse{Allowed: true}, nil
 }
+
+func (s *DBAuthService) UpdatePassword(ctx context.Context, email string, encryptedPassword string) error {
+	_, err := s.db.Transact(ctx, func(tx db.Tx) (interface{}, error) {
+		exec, err := tx.Exec(`UPDATE auth_users SET encrypted_password = $1 WHERE email = $2`, encryptedPassword, email)
```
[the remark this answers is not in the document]
shimi9276: Fixed as above

api/swagger.yml
```diff
@@ -1063,6 +1089,60 @@ paths:
...
+        responses:
+          200:
+            description: successful reset
+          401:
+            $ref: "#/components/responses/Unauthorized"
+          default:
+            $ref: "#/components/responses/ServerError"
+ 
+  /auth/password/forgot:
+      post:
```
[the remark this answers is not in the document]
arielshaqed: You are correct and I am wrong.  POST is used for non-idempotent operations (and therefore warns in the browser).

Thanks!

pkg/api/controller.go
```diff
@@ -175,7 +179,9 @@ func (c *Controller) Login(w http.ResponseWriter, r *http.Request, body LoginJSO
 	secret := c.Auth.SecretStore().SharedSecret()
 	// user.Username will be different from username/access_key_id on
 	// LDAP login.  Use the stored value.
-	tokenString, err := GenerateJWT(secret, user.ID, loginTime, expires)
+	// Audience and email set here to be an empty string for backward compatibility
+	tokenString, err :=
+		GenerateJWT(secret, LoginAudience, user.ID, *user.Email, loginTime, expires)
```
arielshaqed, line 184: I don't understand how this works.  Say I have a user with no email.  I would expect to receive `user.Email == nil`.  And when that user logs in with their access key (supported!), does this panic?

pkg/api/controller.go
```diff
@@ -3006,6 +3012,65 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
 	writeResponse(w, http.StatusOK, response)
 }
 
+func (c *Controller) sendResetPasswordEmail(email []string, token string) error {
+	err := c.Emailer.SendEmail(email, token, token, []string{})
+	if err != nil {
+		c.Logger.Error("Error sending email")
```
arielshaqed, line 3018: This log is not useful.  How do you debug an error with just a log line that says "Error sending email"?  And if there are 10 organizations with 100 users each?

Please add (all) emails, the error, and more details as fields on the log.

pkg/api/controller.go
```diff
@@ -3006,6 +3012,65 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+		return err
+	}
+	c.Logger.Info("Succefully sent email")
```
arielshaqed, line 3021: We need log fields here, too!

pkg/api/controller.go
arielshaqed: I am not sure why this is an issue.  AFAIK this was already exposed to users in previous JWTs.

If we *are* worried about this (e.g. because it hints at number of users), then we should simply avoid using consecutive user IDs.

pkg/api/controller.go
arielshaqed: I don't see the new code, did you forget to push?

pkg/api/login_handler.go
```diff
@@ -21,11 +21,16 @@ const (
 	JWTCookieName          = "access_token"
 )
 
-func GenerateJWT(secret []byte, userID int, issuedAt, expiresAt time.Time) (string, error) {
+// Be aware! The aud field should *only* be used for new applications! Otherwise it should be passed as an empty string i.e. pass LoginAudience or the proper audience,
```
arielshaqed, line 24: godoc comments should begin "GenerateJWT" and then a present-tense description of what it does.  I believe this has already been mentioned; did you forget to push?

arielshaqed (commented):
Most changes mentioned in responses seem to be missing; did you forget to push?

pkg/api/controller.go
nopcoder, line 184: 1. We didn't had to change how we generate tokens today
2. if the email is optional - you need to use the email value if it is not nil

pkg/api/login_handler.go
nopcoder, line 24: Previous review commented about the code doc - first there is the func name.

pkg/api/login_handler.go
```diff
@@ -21,11 +21,16 @@ const (
...
+// so it supports backward compatibility. As well Id field contains the email and not the Id for backward compatibility. The secret field is the secret key
+// used for encryption and set in the config file
+func GenerateJWT(secret []byte, aud string, userID int, email string, issuedAt, expiresAt time.Time) (string, error) {
```
nopcoder, line 27: looks like we need one function that accepts claims that we can reuse it to generate reset token and login token.

pkg/auth/model/model.go
```diff
@@ -41,23 +41,24 @@ type User struct {
...
 	u.EncryptedPassword = pw
 	return nil
 }
 
+func HashPassword(password string) ([]byte, error) {
+	pw, err := bcrypt.GenerateFromPassword([]byte(password), bcrypt.DefaultCost)
+	if err != nil {
+		return nil, err
+	}
+	return pw, nil
```
nopcoder, line 59:
```suggestion
	return bcrypt.GenerateFromPassword([]byte(password), bcrypt.DefaultCost)
```

pkg/api/controller.go
```diff
@@ -3006,6 +3012,65 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+	p, err := model.HashPassword(body.NewPassword)
+	if err != nil {
+		writeError(w, http.StatusServiceUnavailable, err)
+		return
+	}
+
+	err = c.Auth.UpdatePassword(r.Context(), body.Email, p)
+	if err != nil {
+		writeError(w, http.StatusConflict, err)
+	}
```
nopcoder, line 3070: please pay attention - the previous review comments - you need to return if you response to the call and not continue the code on the happy path.

pkg/api/controller.go
```diff
@@ -3006,6 +3012,65 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+	writeResponse(w, http.StatusOK, err)
+}
+
+func (c *Controller) UpdatePassword(w http.ResponseWriter, r *http.Request, body UpdatePasswordJSONRequestBody) {
+	claims, err := VerifyResetPasswordToken(c.Auth, body.Token)
+	if err != nil {
+		writeError(w, http.StatusForbidden, err)
+		return
+	}
+	if strings.Compare(claims.Id, body.Email) != 0 {
```
nopcoder, line 3057: use !=

pkg/api/controller.go
```diff
@@ -3006,6 +3012,65 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+		writeError(w, http.StatusForbidden, "Unauthorized to switch password for given email")
```
nopcoder, line 3058: "Unauthorized update password"

pkg/api/controller.go
```diff
@@ -3006,6 +3012,65 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+	}
+	p, err := model.HashPassword(body.NewPassword)
+	if err != nil {
+		writeError(w, http.StatusServiceUnavailable, err)
+		return
+	}
+
+	err = c.Auth.UpdatePassword(r.Context(), body.Email, p)
+	if err != nil {
+		writeError(w, http.StatusConflict, err)
```
nopcoder, line 3069: StatusConflict?

nopcoder (changes requested):

pkg/api/auth_middleware.go
```diff
@@ -116,12 +118,20 @@ func userByToken(ctx context.Context, logger logging.Logger, authService auth.Se
 		return authService.SecretStore().SharedSecret(), nil
 	})
 	if err != nil {
-		return nil, ErrAuthenticatingRequest
+		return nil, err
```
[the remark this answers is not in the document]
shimi9276: I think this was fixed in an earlier push. In any case reverified it is up to date.

pkg/api/controller.go
shimi9276: Good point. In any case, as @nopcoder suggested, the JWT creation function was split into two separate functions, so the email will not be passed in any case.

shimi9276: Fixed indeed, as you suggested to split into separate functions solved these problems as well.

pkg/api/controller.go
shimi9276: Add logging with fields. It should be noted, that besides for the email address there are no other fields with relevant info (i.e. org id is not available in OSS)

pkg/api/controller.go
shimi9276: Added as welll

pkg/api/controller.go
```diff
@@ -3006,6 +3012,58 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+func (c *Controller) sendResetPasswordEmail(email []string, token string) error {
+	return c.Emailer.SendEmail(email, token, token, []string{})
+}
+
+func (c *Controller) PasswordForgot(w http.ResponseWriter, r *http.Request, body PasswordForgotJSONRequestBody) {
+	user, err := c.Auth.GetUserByEmail(r.Context(), body.Email)
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, err)
+		return
+	}
```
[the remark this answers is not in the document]
shimi9276: Fixed.

pkg/api/controller.go
```diff
@@ -3006,6 +3012,58 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+	if strings.Compare(*user.Email, body.Email) != 0 {
+		writeError(w, http.StatusNotFound, "No such email")
```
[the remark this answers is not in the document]
shimi9276: Set them both (email not found or email mismatch) to a general forbidden error (403)

pkg/api/controller.go
shimi9276: Accepted. Set it to the current user.

pkg/api/controller.go
shimi9276: Sorry, Used it in many places in the code and missed fixing this one.

pkg/api/controller.go
shimi9276: Fixed. As well

pkg/api/controller.go
shimi9276: Fixed

pkg/api/controller.go
```diff
@@ -3006,6 +3012,58 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+	claims, err := VerifyResetPasswordToken(c.Auth, body.Token)
+	if err != nil {
+		writeError(w, http.StatusForbidden, err)
+		return
+	}
+	if strings.Compare(claims.Id, body.Email) != 0 {
+		writeError(w, http.StatusForbidden, "Unautherized to switch password for given email")
+		return
+	}
+	p, err := model.HashPassword(body.NewPassword)
```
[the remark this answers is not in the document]
shimi9276: Done!

pkg/api/controller.go
shimi9276: Autocomplete... fixed

pkg/api/controller.go
shimi9276: Agreed, missed fixing this one. fixed

pkg/api/login_handler.go
shimi9276: Fixed the docs according to the new code. Tried my best to write it by the godoc rules.

pkg/api/login_handler.go
shimi9276: Done!

pkg/auth/model/model.go
shimi9276: Not relevant to the new function created.

shimi9276 (commented):
As there are multiple comments and requested changes, I may have missed some. If that is the case please call them out. Thanks!

pkg/api/controller.go
```diff
@@ -3006,6 +3011,66 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+		c.Logger.
+			WithError(err).
+			WithFields(logging.Fields{
+				"email": email,
+			}).Debug("Error sending email")
+		return err
+	}
+	c.Logger.WithFields(logging.Fields{
+		"email": email,
+	}).Info("Succefully sent email")
```
arielshaqed, line 3026:
```suggestion
	}).Info("Successfully sent email")
```
Log messages are most useful if words in them are spelled consistently.  Perhaps the IDE check spelling on log messages?

pkg/api/controller.go
```diff
@@ -3006,6 +3011,66 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+	return err
+}
+
+func (c *Controller) PasswordForgot(w http.ResponseWriter, r *http.Request, body PasswordForgotJSONRequestBody) {
+	user, err := c.Auth.GetUserByEmail(r.Context(), body.Email)
+	if err != nil {
+		writeError(w, http.StatusForbidden, http.StatusText(http.StatusForbidden))
```
arielshaqed, line 3033: I think this is wrong, here and on l. 3037 you need to send 401 (Unauthorized).  See [403 Forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403), [401 Unauthorized](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401).

Confusingly, "forbidden" in HTTP generally corresponds to authorization failure ("I know who you are and I'm not going to let you do it"), while "unauthorized" generally corresponds to authentication failure ("you didn't prove to me who you are").

pkg/api/login_handler.go
```diff
@@ -21,12 +21,30 @@ const (
 	JWTCookieName          = "access_token"
 )
 
-func GenerateJWT(secret []byte, userID int, issuedAt, expiresAt time.Time) (string, error) {
+func generateJWT(claims *jwt.StandardClaims, secret []byte) (string, error) {
+	token := jwt.NewWithClaims(jwt.SigningMethodHS256, claims)
+	return token.SignedString(secret)
+}
+
+// GenerateJWTLogin supports backward compatibilty for creating a login jwt. The audience is set to LoginAudience which is
```
arielshaqed, line 29: Please note linter comments (here, about spelling).

arielshaqed, line 29: This is still not the Go way to write func docs.  The comment should start by saying what it does.  After that finer details may appear.

pkg/auth/model/model.go
arielshaqed: I'm not following this comment thread.  Isn't the new function _exactly_ the single line in the comment above?

api/swagger.yml
```diff
@@ -591,6 +599,24 @@ components:
       required:
         - username
 
+      
+    UpdatePasswordByToken:
+      type: object
+      properties:
+        token:
+          description: token used for authentification
```
arielshaqed, line 607:
```suggestion
          description: token used for authentication
```

api/swagger.yml
```diff
@@ -591,6 +599,24 @@ components:
...
+          type: string
+        newPassword:
+          description: the new password to be updated
```
arielshaqed, line 610: No need for articles when documenting parameters.
```suggestion
          description: new password to update
```

arielshaqed (commented):
Thanks!

Please go over spelling, it is quite confusing and will be hard to search docs, logs, etc.

Also you may have missed the non-line comments in my original review.  There remain feature issues with email flooding.

pkg/api/login_handler.go
```diff
@@ -21,11 +21,16 @@ const (
 	JWTCookieName          = "access_token"
 )
 
-func GenerateJWT(secret []byte, userID int, issuedAt, expiresAt time.Time) (string, error) {
+// Be aware! The aud field should *only* be used for new applications! Otherwise it should be passed as an empty string, so it supports
+// backward compatibility. As well Id field contains the email and not the Id for backward compatibility. The secret field is the secret key
```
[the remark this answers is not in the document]
arielshaqed: This will be a security and functionality headache: whenever a user is used, there is no way to know whether it's a new user-with-email or an old user-without-email.  So dereferencing the `email` thread will be a bit of a gamble.

arielshaqed:
> Also you may have missed the non-line comments in my original review. There remain feature issues with email flooding.

Sorry, my mistake: _**I never posted those comments**_.  So here goes.

There is a trivial email-flooding attack on this implementation.  If I know ariels@bigcorp.com has an account, I can flood the lakeFS server with "forgot password" requests.  This has a >>10x amplification factor (I send >> 10x fewer bytes than lakeFS will, possible even 100x fewer).

What can I attack with this?

* ariels@bigcorp.com's mailbox.
* The operator (who will get notified of abuse)
* lakeFS (which may get cut off from emailing by the email provider)
* lakeFS (which may end up being marked as spam by its users)

Typically features such as this one have a rate-limiter built into them.  We can either use an existing rate-limiter or write our own -- maybe initially even just in memory.  But something has to prevent unlimited password reset requests.

api/swagger.yml
shimi9276: Fixed

api/swagger.yml
shimi9276: Fixed

pkg/api/controller.go
shimi9276: Fixed. Bad spelling on my side. (That's what happens when coding at 12 AM)

pkg/api/controller.go
shimi9276: Your right. Fixed

pkg/api/login_handler.go
shimi9276: Fixed

shimi9276: Fixed (I think)

pkg/auth/model/model.go
shimi9276: No, here we must check for errors before we transact with the DB.

shimi9276 (commented):
Thanks for allocating time for multiple requests!

pkg/api/controller.go
```diff
@@ -3006,6 +3011,72 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+			}).Debug("Error sending email")
+		return err
+	}
+	c.Logger.WithFields(logging.Fields{
+		"email": email,
+	}).Info("Successfully sent email")
+	return err
+}
+
+func (c *Controller) PasswordForgot(w http.ResponseWriter, r *http.Request, body PasswordForgotJSONRequestBody) {
```
nopcoder, line 3030: We should log/metric the error in this flow - but not notify the user about anything - as this request is not authenticated. If it is the real user - the user will accept email - this is the way the request is verified.

pkg/api/auth_middleware.go
```diff
@@ -120,6 +120,14 @@ func userByToken(ctx context.Context, logger logging.Logger, authService auth.Se
...
 	claims, ok := token.Claims.(*jwt.StandardClaims)
 	if !ok || !token.Valid {
+		return nil, err
+	}
+	return claims, nil
+}
+
+func userByToken(ctx context.Context, logger logging.Logger, authService auth.Service, tokenString string) (*model.User, error) {
+	claims, err := verifyToken(authService, tokenString)
+	if err != nil || claims.Audience != LoginAudience {
```
nopcoder, line 130: claims have VerifyAudience

pkg/api/auth_middleware.go
```diff
@@ -154,3 +162,11 @@ func userByAuth(ctx context.Context, logger logging.Logger, authenticator auth.A
 	}
 	return user, nil
 }
+
+func VerifyResetPasswordToken(authService auth.Service, tokenString string) (*jwt.StandardClaims, error) {
+	claims, err := verifyToken(authService, tokenString)
+	if err != nil || claims.Audience != ResetPasswordAudience {
```
nopcoder, line 168: use VerifyAudience method

pkg/api/controller.go
```diff
@@ -3006,6 +3011,72 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+	writeResponse(w, http.StatusOK, err)
+}
+
+func (c *Controller) UpdatePassword(w http.ResponseWriter, r *http.Request, body UpdatePasswordJSONRequestBody) {
+	claims, err := VerifyResetPasswordToken(c.Auth, body.Token)
+	if err != nil {
+		writeError(w, http.StatusUnauthorized, err)
+		return
+	}
+	user, err := c.Auth.GetUserByEmail(r.Context(), claims.Id)
```
nopcoder, line 3063: here we will need to pull the email from a different field (not Id) - related to another comment in this review

nopcoder (changes requested):

pkg/api/controller.go
```diff
@@ -3006,6 +3011,72 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		writeError(w, http.StatusUnauthorized, http.StatusText(http.StatusUnauthorized))
```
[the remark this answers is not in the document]
shimi9276: Fixed

pkg/api/controller.go
```diff
@@ -3006,6 +3011,72 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	if claims.Id != *user.Email {
```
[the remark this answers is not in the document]
shimi9276: Fixed

pkg/auth/service.go
```diff
@@ -959,3 +961,12 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
 	// we're allowed!
 	return &AuthorizationResponse{Allowed: true}, nil
 }
+
+func (s *DBAuthService) HashAndUpdatePassword(ctx context.Context, email string, password string) error {
```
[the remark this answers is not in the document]
shimi9276: Fixed

pkg/api/controller.go
shimi9276: Fixed

api/swagger.yml
```diff
@@ -1063,6 +1089,60 @@ paths:
...
+            description: password reset request submitted
+          400:
+            description: bad request
+            content:
+              application/json:
+                schema:
+                  $ref: "#/components/schemas/Error"
+          500:
+            $ref: "#/components/responses/ServerError"
+          default:
```
nopcoder, line 1139: keep the server error as default - use 204 instead of 200

pkg/api/controller.go
```diff
@@ -3006,6 +3011,62 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+	secret := c.Auth.SecretStore().SharedSecret()
+	currentTime := time.Now()
+	token, err := GenerateJWTResetPassword(secret, ResetPasswordAudience, user.ID, body.Email, currentTime, currentTime.Add(tokenExpiryDuration))
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, err)
+		c.Logger.WithError(err).WithField("email", *user.Email).Debug("failed to create a token")
+		return
+	}
+	// TODO (@shimi9276) create template for sending the email with link for reset
+	err = c.sendResetPasswordEmail([]string{body.Email}, token)
```
nopcoder, line 3040: user user.email and not the one the use posted

pkg/api/controller.go
```diff
@@ -3006,6 +3011,62 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
 	writeResponse(w, http.StatusOK, response)
 }
 
+func (c *Controller) sendResetPasswordEmail(email []string, token string) error {
+	err := c.Emailer.SendEmail(email, token, token, []string{})
```
nopcoder, line 3015: can pass nil instead of empty slice

pkg/api/controller.go
```diff
@@ -3006,6 +3011,62 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
 	writeResponse(w, http.StatusOK, response)
 }
 
+func (c *Controller) sendResetPasswordEmail(email []string, token string) error {
```
nopcoder, line 3014: can accept just an email and not mailing list

pkg/api/controller.go
```diff
@@ -3006,6 +3011,62 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+	err := c.Emailer.SendEmail(email, token, token, []string{})
+	if err != nil {
+		c.Logger.WithError(err).WithField("email", email).Debug("failed sending reset password email")
+		return err
+	}
+	c.Logger.WithField("email", email).Info("Successfully sent reset password email")
+	return err
```
nopcoder, line 3021: nil

pkg/api/controller.go
```diff
@@ -3006,6 +3011,62 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+}
+
+func (c *Controller) PasswordForgot(w http.ResponseWriter, r *http.Request, body PasswordForgotJSONRequestBody) {
+	user, err := c.Auth.GetUserByEmail(r.Context(), body.Email)
+	if err != nil {
+		writeError(w, http.StatusNoContent, http.StatusText(http.StatusNoContent))
+		c.Logger.WithError(err).WithField("request_body", body).Debug("failed to retrieve user by email")
+		return
```
nopcoder, line 3029: prefer keep log before the action in this case

pkg/api/controller.go
```diff
@@ -3006,6 +3011,62 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+		return err
+	}
+	c.Logger.WithField("email", email).Info("Successfully sent reset password email")
+	return err
+}
+
+func (c *Controller) PasswordForgot(w http.ResponseWriter, r *http.Request, body PasswordForgotJSONRequestBody) {
+	user, err := c.Auth.GetUserByEmail(r.Context(), body.Email)
+	if err != nil {
+		writeError(w, http.StatusNoContent, http.StatusText(http.StatusNoContent))
```
nopcoder, line 3027: w.WriteHeader(http.StatusNoContent)

pkg/api/controller.go
```diff
@@ -3006,6 +3011,62 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+	token, err := GenerateJWTResetPassword(secret, ResetPasswordAudience, user.ID, body.Email, currentTime, currentTime.Add(tokenExpiryDuration))
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, err)
+		c.Logger.WithError(err).WithField("email", *user.Email).Debug("failed to create a token")
+		return
+	}
+	// TODO (@shimi9276) create template for sending the email with link for reset
+	err = c.sendResetPasswordEmail([]string{body.Email}, token)
+	if err != nil {
+		writeError(w, http.StatusNoContent, err)
```
nopcoder, line 3042: replace with log

pkg/api/controller.go
```diff
@@ -3006,6 +3011,62 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+		c.Logger.WithError(err).WithField("email", user.Email).Debug("failed to retrieve user by email")
+		return
+	}
+	err = c.Auth.HashAndUpdatePassword(r.Context(), claims.Subject, body.NewPassword)
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, err)
+		c.Logger.WithError(err).WithField("email", claims.Subject).Debug("failed to update password")
+		return
+	}
+	writeResponse(w, http.StatusCreated, err)
```
nopcoder, line 3067: w.WriteHeader

pkg/api/controller.go
```diff
@@ -3006,6 +3011,62 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+	}
+	user, err := c.Auth.GetUserByEmail(r.Context(), claims.Subject)
+	if err != nil {
+		writeError(w, http.StatusNotFound, http.StatusText(http.StatusNotFound))
+		c.Logger.WithError(err).WithField("email", user.Email).Debug("failed to retrieve user by email")
+		return
+	}
+	err = c.Auth.HashAndUpdatePassword(r.Context(), claims.Subject, body.NewPassword)
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, err)
```
nopcoder, line 3063: check that 'err' doesn't container sensitive data or replace with error message

pkg/auth/service.go
```diff
@@ -68,6 +69,7 @@ type Service interface {
 	GetCredentialsForUser(ctx context.Context, username, accessKeyID string) (*model.Credential, error)
 	GetCredentials(ctx context.Context, accessKeyID string) (*model.Credential, error)
 	ListUserCredentials(ctx context.Context, username string, params *model.PaginationParams) ([]*model.Credential, *model.Paginator, error)
+	HashAndUpdatePassword(ctx context.Context, email string, encryptedPassword string) error
```
nopcoder, line 72:
```suggestion
	HashAndUpdatePassword(ctx context.Context, email string, password string) error
```

nopcoder (changes requested):

api/swagger.yml
shimi9276: Ok. fixed

pkg/api/controller.go
shimi9276: Fixed

pkg/api/controller.go
shimi9276: Fixed

pkg/api/controller.go
shimi9276: Fixed

pkg/api/controller.go
shimi9276: Fixed

pkg/api/login_handler.go
```diff
@@ -21,12 +21,32 @@ const (
...
 	}
-	token := jwt.NewWithClaims(jwt.SigningMethodHS256, claims)
-	return token.SignedString(secret)
+	return generateJWT(claims, secret)
+}
+
+// GenerateJWTResetPassword creates a jwt token with the field subject set the email passed.
+func GenerateJWTResetPassword(secret []byte, aud string, userID int, email string, issuedAt, expiresAt time.Time) (string, error) {
+	claims := &jwt.StandardClaims{
+		Audience:  aud,
```
nopcoder, line 45: The audience should be set by the implementation as the func name, we need to set the audience because this is not the generic way.

pkg/auth/model/model.go
```diff
@@ -41,23 +41,20 @@ type User struct {
 }
 
 // hashPassword generates a hashed password from a plaintext string
```
nopcoder, line 43: 1. Move the comment to the function
2. The function is now exposed so start with upper case

nopcoder (changes requested):

pkg/api/login_handler.go
shimi9276: ok, fixed

pkg/auth/model/model.go
shimi9276: moved and was already uppercase.

pkg/api/controller.go
```diff
@@ -3006,6 +3011,63 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+		return
+	}
+	// TODO (@shimi9276) create template for sending the email with link for reset
+	err = c.sendResetPasswordEmail(*user.Email, token)
+	if err != nil {
+		c.Logger.WithError(err).WithField("email", user.Email).Debug("failed sending reset password email")
+		return
+	}
+	c.Logger.WithField("email", user.Email).Info("Successfully sent reset password email")
+	writeResponse(w, http.StatusNoContent, err)
```
nopcoder, line 3046: don't write err in no content response. just update the status code

pkg/api/controller.go
```diff
@@ -3006,6 +3011,63 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+	token, err := GenerateJWTResetPassword(secret, user.ID, body.Email, currentTime, currentTime.Add(tokenExpiryDuration))
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, err)
+		c.Logger.WithError(err).WithField("email", *user.Email).Debug("failed to create a token")
+		return
+	}
+	// TODO (@shimi9276) create template for sending the email with link for reset
+	err = c.sendResetPasswordEmail(*user.Email, token)
+	if err != nil {
+		c.Logger.WithError(err).WithField("email", user.Email).Debug("failed sending reset password email")
```
nopcoder, line 3042: remove this one. or the internal one inside the method

pkg/api/controller.go
```diff
@@ -3006,6 +3011,63 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+	}
+	c.Logger.WithField("email", user.Email).Info("Successfully sent reset password email")
+	writeResponse(w, http.StatusNoContent, err)
+}
+
+func (c *Controller) UpdatePassword(w http.ResponseWriter, r *http.Request, body UpdatePasswordJSONRequestBody) {
+	claims, err := VerifyResetPasswordToken(c.Auth, body.Token)
+	if err != nil {
+		writeError(w, http.StatusUnauthorized, err)
+		c.Logger.WithError(err).WithField("token", body.Token).Debug("failed to verify token")
```
nopcoder, line 3053: previous feedback.  log and write response.

pkg/auth/model/model.go
```diff
@@ -41,23 +41,19 @@ type User struct {
...
-	pw, err := hashPassword(password)
+	pw, err := HashPassword(password)
 	if err != nil {
 		return err
 	}
 	u.EncryptedPassword = pw
 	return nil
 }
 
+func HashPassword(password string) ([]byte, error) {
```
nopcoder, line 53: the code doc left above and does not match the func case.

nopcoder (changes requested):

pkg/api/controller.go
shimi9276: Fixed. Forgot to commit and push

pkg/api/controller.go
shimi9276: Fixed in all places (as above)

pkg/auth/model/model.go
shimi9276: Not code I wrote but moved the description to the proper function.

shimi9276 (commented):
Missed a merge an push, sorry about that

api/swagger.yml
```diff
@@ -1063,6 +1089,56 @@ paths:
...
+          204:
+            description: No content
+          400:
+            description: bad request
+            content:
+              application/json:
+                schema:
+                  $ref: "#/components/schemas/Error"
+          500:
+            $ref: "#/components/responses/ServerError"
```
nopcoder, line 1138: delete this one too

pkg/api/controller.go
```diff
@@ -3006,6 +3011,57 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
 }
 
+func (c *Controller) sendResetPasswordEmail(email string, token string) error {
+	return c.Emailer.SendEmail([]string{email}, token, token, nil)
+}
+
+func (c *Controller) PasswordForgot(w http.ResponseWriter, r *http.Request, body PasswordForgotJSONRequestBody) {
+	user, err := c.Auth.GetUserByEmail(r.Context(), body.Email)
+	if err != nil {
+		c.Logger.WithError(err).WithField("request_body", body).Debug("failed to retrieve user by email")
```
nopcoder, line 3021:
```suggestion
		c.Logger.WithError(err).WithField("email", body.Email).Debug("forgot password, user not found by email")
```

pkg/api/controller.go
```diff
@@ -3006,6 +3011,57 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
 	writeResponse(w, http.StatusOK, response)
 }
 
+func (c *Controller) sendResetPasswordEmail(email string, token string) error {
+	return c.Emailer.SendEmail([]string{email}, token, token, nil)
+}
+
+func (c *Controller) PasswordForgot(w http.ResponseWriter, r *http.Request, body PasswordForgotJSONRequestBody) {
+	user, err := c.Auth.GetUserByEmail(r.Context(), body.Email)
```
nopcoder, line 3019: We need to verify the request input (email), before you call get user by email.
You can use the following before calling the lookup:
```go
addr, err := mail.ParseAddress(body.Email)
if err != nil {
    c.Logger.WithError(err).WithField("email", body.Email).Debug("forgot password with invalid email")
    w.WriteHeader(http.StatusNoContent)
    return
}
```

pkg/api/controller.go
```diff
@@ -3006,6 +3011,57 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
 	writeResponse(w, http.StatusOK, response)
 }
 
+func (c *Controller) sendResetPasswordEmail(email string, token string) error {
+	return c.Emailer.SendEmail([]string{email}, token, token, nil)
+}
+
+func (c *Controller) PasswordForgot(w http.ResponseWriter, r *http.Request, body PasswordForgotJSONRequestBody) {
```
nopcoder, line 3018: I know it is a bit late but 'passwordForgot' (operation id) should be 'forgotPassword' - it is API and I think we better have 'ForgotPassword' as a method and not the other way around.
It is hard to change API later and I think we will be sorry to lookup PasswordForgot later.

nopcoder (changes requested):

nopcoder:
LGTM, please create an issue and address the missing test code.

nopcoder (approved):

arielshaqed (changes requested):
Thanks!

## Limit rates
Please add a rate-limiter on sending emails.  Package `time/rate` is a good token bucket.  You might initialize a limiter with `limit.NewLimiter(Limit(1.0/60.0), 10)` and then check `if Allow()` before you send an email.

### Explanation of the parameters

1/60 gives you on average at most 1 email every minute (so 1440 a day).  The way it works is that every 60 seconds it gets a "voucher" allowing it to send an email, and every time it wants to send an email it has to pay a voucher.  So if nobody forgets their password for 5 minutes, there are 5 vouchers and 5 emails can be sent _immediately_.  **BUT**: lakeFS only gets to keep up to 10 vouchers.  So if nobody forgets their password for 24 hours, there are still only 10 vouchers and no more than 10 emails can be sent immediately.

pkg/api/controller.go
```diff
@@ -3006,6 +3025,69 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
 }
 
+func (c *Controller) sendResetPasswordEmail(email string, token string) error {
+	if c.Emailer.Limiter.Allow() {
+		return c.Emailer.SendEmail([]string{email}, token, token, nil)
+	}
+	err := ErrRateLimitExceeded
+	c.Logger.WithError(err).WithField("email", email).Debug("Email sending rate limit was exceeded")
+	return err
+}
```
arielshaqed, line 3035: The usual style for early-return is to  return quickly on exceptional conditions.  In this case it would be:
```suggestion
	if !c.Emailer.Limiter.Allow() {
		err := ErrRateLimitExceeded
		c.Logger.WithError(err).WithField("email", email).Warn("Email sending rate limit exceeded")
	}
	return c.Emailer.SendEmail([]string{email}, token, token, nil)
}
```

Also we need at least to **warn** here: this is a condition that interests operators.

pkg/api/controller.go
```diff
@@ -3026,7 +3026,12 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
 }
 
 func (c *Controller) sendResetPasswordEmail(email string, token string) error {
-	return c.Emailer.SendEmail([]string{email}, token, token, nil)
+	if c.Emailer.Limiter.Allow() {
+		return c.Emailer.SendEmail([]string{email}, token, token, nil)
+	}
+	err := ErrRateLimitExceeded
+	c.Logger.WithError(err).WithField("email", email).Debug("Email sending rate limit was exceeded")
+	return err
```
arielshaqed, line 3034: This code should go inside `c.Emailer.SendEmail()` (or `SendEmailWithLimit`), not inside the Controller.  The general rule is that you should prefer not to dig inside another object's internals.

docs/reference/configuration.md
```diff
@@ -107,6 +107,8 @@ This reference uses `.` to denote the nesting of values.
 + `email.username` `(string)` - A string representing the username of the specific account at the SMTP. It's recommended to provide this value at runtime from a secret vault of some sort.
 + `email.password` `(string)` - A string representing the password of the account. It's recommended to provide this value at runtime from a secret vault of some sort.
 + `email.sender` `(string)` - A string representing the email account which is set as the sender.
++ `email.rate_limit` `(float)` - A float representing the amout emails that can be sent per second.
```
arielshaqed, line 110: This is a really hard number to select correctly.  Indeed, to set the default you explicitly inverted a useful number.  I would suggest:
1. `email.limit_every` (Duration): Average time to wait between sending emails.
1. `email.burst` (int): Maximal burst of emails before applying `limit_every`.

Then you can compute the rate limit param as something like `float(time.Second) / float(config.Email.LimitEvery)`.

WDYT?

arielshaqed (commented):
Thanks!  We're there.  I am not blocking but I **do** want you to invert the `rate_limit` config param: it is very hard to use correctly in its current form.  Please do try to do it :-)

arielshaqed (approved):

nopcoder (changes requested):

pkg/api/controller.go
```diff
@@ -3030,7 +3026,12 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
 }
 
 func (c *Controller) sendResetPasswordEmail(email string, token string) error {
-	return c.Emailer.SendEmail([]string{email}, token, token, nil)
+	if c.Emailer.Limiter.Allow() {
```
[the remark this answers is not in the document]
shimi9276: Fixed

pkg/api/controller.go
```diff
@@ -3030,7 +3026,12 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+		return c.Emailer.SendEmail([]string{email}, token, token, nil)
+	}
+	err := ErrRateLimitExceeded
+	c.Logger.WithError(err).WithField("email", email).Debug("Email sending rate limit was exceeded")
```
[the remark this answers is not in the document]
shimi9276: fixed

pkg/api/controller.go
shimi9276: Went with SendEmailWithLimit, to give the ability to send email without limits if the caller is intrested

pkg/api/controller.go
shimi9276: fixed

pkg/api/errors.go
```diff
@@ -14,4 +14,6 @@ var (
 	ErrInvalidAPIEndpoint = errors.New("invalid API endpoint")
 
 	ErrRequestSizeExceeded = errors.New("request size exceeded")
+
+	ErrRateLimitExceeded = errors.New("rate limit exceeded")
```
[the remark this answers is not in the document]
shimi9276: fixed

docs/reference/configuration.md
shimi9276: accepted and done

pkg/config/template.go
```diff
@@ -178,12 +178,12 @@ type configuration struct {
...
-		Password  string  `mapstructure:"password"`
-		Sender    string  `mapstructure:"sender"`
-		RateLimit float64 `mapstructure:"rate_limit"`
-		Burst     int     `mapstructure:"burst"`
+		SMTPHost   string  `mapstructure:"smtp_host"`
+		Port       int     `mapstructure:"port"`
+		Username   string  `mapstructure:"username"`
+		Password   string  `mapstructure:"password"`
+		Sender     string  `mapstructure:"sender"`
+		LimitEvery float64 `mapstructure:"limit_every"`
```
arielshaqed, line 186: This parameter *must* include units.  Otherwise I have no idea.  (Or it could be `time.Duration`, which has units).

docs/reference/configuration.md
```diff
@@ -107,8 +107,8 @@ This reference uses `.` to denote the nesting of values.
 + `email.username` `(string)` - A string representing the username of the specific account at the SMTP. It's recommended to provide this value at runtime from a secret vault of some sort.
 + `email.password` `(string)` - A string representing the password of the account. It's recommended to provide this value at runtime from a secret vault of some sort.
 + `email.sender` `(string)` - A string representing the email account which is set as the sender.
-+ `email.rate_limit` `(float)` - A float representing the amout emails that can be sent per second.
-+ `email.burst` `(int)` - An integer representing the maximum amount of credits that can be obtained.
++ `email.limit_every` `(Duration)` - Average time in seconds to wait between sending emails.
```
arielshaqed, line 110: But the actual type is `float64`!  As suggested, please use `time.Duration` and code the conversion.  Also our config already has a way to implement default values; please use it.

arielshaqed (changes requested):
Thanks!

pkg/email/send_email.go
```diff
@@ -11,30 +14,35 @@ type Emailer struct {
 	Limiter *rate.Limiter
 }
 
+var (
+	ErrRateLimitExceeded = errors.New("rate limit exceeded")
+)
```
nopcoder, line 19: just style for single var
```suggestion
var ErrRateLimitExceeded = errors.New("rate limit exceeded")
```

pkg/email/send_email.go
```diff
@@ -11,30 +14,35 @@ type Emailer struct {
...
-		e.RateLimit = defaultRateLimit
+	if e.LimitEvery == 0 {
+		e.LimitEvery = defaultRateLimit
 	}
 	if e.Burst == 0 {
 		e.Burst = defaultBurstRate
 	}
-	l := rate.NewLimiter(rate.Limit(e.RateLimit), e.Burst)
+	limit := float64(time.Second) / float64(e.LimitEvery)
+	l := rate.NewLimiter(rate.Limit(limit), e.Burst)
```
nopcoder, line 45: If there was already discussion about the rate limit, but my 2cent:
1. https://pkg.go.dev/golang.org/x/time/rate#Every - based on the configuration, this is the method you are looking for.
2. based on the current implementation, if both of us asks for password reset - at the same time, my request will fail. We don't tell the user about our failures. so, I will not get any email.  So as I see it, we will require to tell the user we failed, so it can retry (not something I like).  Have different rate limit, one that holds different configuration for the global (all emails) and per email.

pkg/email/send_email.go
```diff
@@ -11,30 +14,35 @@ type Emailer struct {
...
-	if e.RateLimit == 0 {
-		e.RateLimit = defaultRateLimit
+	if e.LimitEvery == 0 {
+		e.LimitEvery = defaultRateLimit
 	}
 	if e.Burst == 0 {
 		e.Burst = defaultBurstRate
 	}
-	l := rate.NewLimiter(rate.Limit(e.RateLimit), e.Burst)
+	limit := float64(time.Second) / float64(e.LimitEvery)
```
nopcoder, line 44: check the linter error

pkg/email/send_email.go
```diff
@@ -53,3 +61,11 @@ func (e Emailer) SendEmail(receivers []string, subject string, body string, atta
 	}
 	return e.Dialer.DialAndSend(msg)
 }
+
+func (e Emailer) SendEmailWithLimit(receivers []string, subject string, body string, attachmentFilePath []string) error {
+	if !e.Limiter.Allow() {
+		err := ErrRateLimitExceeded
+		return err
```
nopcoder, line 68: code style
```suggestion
		return ErrRateLimitExceeded
```

nopcoder (changes requested):

pkg/email/send_email.go
arielshaqed: (1) is really true, thanks!

(2) -- this is _exactly_ what the burst is for.  We can send _burst_ emails at once when we do not suspect an attack.  So not a problem when not under attack.  We might additionally report an informative error to users (of course).

pkg/email/send_email.go
arielshaqed: The warning is because `LimitEvery` is still not a `Duration`.  But: this place is where @nopcoder suggested simply using the `Every` method!

arielshaqed (commented):
Thanks, @nopcoder  -- Every is exactly the workaround for writing the nasty computation!

pkg/email/send_email.go
nopcoder: Thanks @arielshaqed this is what I was looking for - item 2.

pkg/email/send_email.go

pkg/email/send_email.go
shimi9276: Used indeed! Thanks

pkg/email/send_email.go

pkg/config/config.go
```diff
@@ -161,6 +164,9 @@ const (
 
 	SecurityAuditCheckURLKey     = "security.audit_check_url"
 	DefaultSecurityAuditCheckURL = "https://audit.lakefs.io/audit"
+
+	EmailLimitEveryKey = "email.limit_every_duration"
```
nopcoder, line 168:
```suggestion
	EmailLimitEveryDurationKey = "email.limit_every_duration"
```

pkg/config/config.go
```diff
@@ -212,6 +218,8 @@ func setDefaults() {
 
 	viper.SetDefault(SecurityAuditCheckIntervalKey, DefaultSecurityAuditCheckInterval)
 	viper.SetDefault(SecurityAuditCheckURLKey, DefaultSecurityAuditCheckURL)
+	viper.SetDefault(EmailLimitEveryKey, DefaultEmailLimitEvery)
```
nopcoder, line 221:
```suggestion
	viper.SetDefault(EmailLimitEveryDurationKey, DefaultEmailLimitEveryDuration)
```

pkg/email/send_email.go
```diff
@@ -14,43 +14,29 @@ type Emailer struct {
...
+	SMTPHost           string
+	Port               int
+	Username           string
+	Password           string
+	Sender             string
+	LimitEveryDuration time.Duration
+	Burst              int
 }
 
 func NewEmailer(e EmailParams) Emailer {
```
nopcoder, line 29: Prefer to return a pointer in this case. Interface that holds implementation or case where copy the Emailer value will not support value copy.

nopcoder (changes requested):

pkg/api/serve.go
```diff
@@ -85,7 +85,7 @@ func Serve(
 		actions,
 		auditChecker,
 		logger,
-		emailer,
+		*emailer,
```
nopcoder, line 88: You want to keep a pointer to emailer - later when you add the tests you will need to replace it with interface

pkg/config/config.go
```diff
@@ -218,7 +218,7 @@ func setDefaults() {
 
 	viper.SetDefault(SecurityAuditCheckIntervalKey, DefaultSecurityAuditCheckInterval)
 	viper.SetDefault(SecurityAuditCheckURLKey, DefaultSecurityAuditCheckURL)
-	viper.SetDefault(EmailLimitEveryKey, DefaultEmailLimitEvery)
+	viper.SetDefault(EmailLimitEveryDurationKey, DefaultEmailLimitEvery)
```
nopcoder, line 221: DefaultEmailLimitEvery -> DefaultEmailLimitEveryDuration

nopcoder (changes requested):

pkg/config/config.go
shimi9276: fixed

pkg/api/serve.go
shimi9276: Done

nopcoder: The previous feedback I wrote - passing emailer implementation as a pointer and not copy the value from the creation, is true here too.
Don't want to copy the emailer struct by value - we like to pass the pointer we created also the controller.

nopcoder (changes requested):

pkg/api/serve.go
shimi9276: Got it, made two pushes one implementing it via pointers (i.e. the emailer is passed by reference) and the second via interface. I think the interface one is more proper.

arielshaqed (approved):
Nice!  I like how clean the code is with the rate-limiter.  It seems to have a really good interface.

Thanks...

pkg/email/send_email.go
```diff
@@ -8,6 +8,11 @@ import (
 	"gopkg.in/gomail.v2"
 )
 
+type Email interface {
```
nopcoder, line 11: This is not what I meant in my previous comment.
- Interface should describe the operation, like: EmailSender
- It would probably include just a method and the one that passes the implementation should pass one that use or not use limit.
- The mailer implementation can hold two ways to send email - with and without limit, but in our configuration we configure one mail - so from the user perspective we configure it.  A preferred way so having two methods in mailer is good for testing and internal implementation - but not for reflecting the mailer behavior.

nopcoder (changes requested):

nopcoder (approved):

Issue #3173, opened 2022-04-06
Create UI and modify UI pages for login and reset password


shimi9276:
The ability to use email/password credentials creates the need for supporting UI pages to log in and reset the password. 
See [here](https://www.figma.com/file/76Vk9s608aRH7HQEe6CbNl/lakeFS-Cloud?node-id=6%3A230) for exact mockup. 
Depends on #3167

PR #3176, opened 2022-04-07
Add client side reset password workflow

talSofer:
closes https://github.com/treeverse/lakeFS/issues/3173

Important Note:
This pr does not make any backend call. I will add it after https://github.com/treeverse/lakeFS/pull/3167 will be closed. 

Screen Shot 2022-04-07 at 13 42 04
Screen Shot 2022-04-07 at 13 40 15
Screen Shot 2022-04-07 at 13 40 09

Screen Shot 2022-04-07 at 14 08 22

Screen Shot 2022-04-07 at 14 56 48


webui/package-lock.json
```diff
@@ -1,8 +1,7715 @@
 {
   "name": "lakefs-ui",
   "version": "0.1.0",
-  "lockfileVersion": 1,
+  "lockfileVersion": 2,
```
nopcoder, line 4: This happens when we use npm v7 or above.  It is probably before you are using newer version of nodejs.
We can upgrade the current node version requirements, update the contribution doc and workflows - but maybe in a different pr.

webui/src/lib/api/index.js
```diff
@@ -764,9 +764,14 @@ class Config {
...
     async getLakeFSVersion() {
-        const response = await apiRequest('/config/version', {
-            method: 'GET',
-        });
+        try {
+            const response = await apiRequest('/config/version', {
+                method: 'GET',
+            });
+        } catch (e) {
+            // return an empty response in case we are not logged in
```
johnnyaug, line 772:
```suggestion
            // avoid crash in case we are not logged in
```

webui/src/pages/auth/reset-password.jsx
```diff
@@ -0,0 +1,188 @@
...
+        if (validator.isEmail(email.current.value)) {
+            setFormValid(true);
+            setEmailValid(true);
+        } else {
+            setFormValid(false);
+            setEmailValid(false);
+        }
+    };
+
+    const [reqResetPwdError, setReqResetPwdError] = React.useState(null);
```
johnnyaug, line 30: What is the difference between React.useState and useState?

webui/src/pages/auth/reset-password.jsx
```diff
@@ -0,0 +1,188 @@
...
+    } else {
+        return (
+            <Row>
+                <Col md={{offset: 4, span: 4}}>
+                    <div className="reset-pwd-risks">
+                        <div className="risk-icon">
+                            <AlertIcon size={20}/>
+                        </div>
+                        <div className="warn-message">
+                            We don&#39;t store plain text passwords, therefore, once reset, they&#39;ll be lost.<br/>
```
johnnyaug, line 54: Nit, opinion: no need for all this info: 
1. It goes without saying that we don't store them.
2. If they don't have access to the mailbox, they can't reset the password.
3. Reset link is available for six hours - this info should be in the email itself

johnnyaug (changes requested):
Excited for this feature!

Left comments, nothing major.

webui/src/lib/api/index.js
talSofer: done

webui/src/pages/auth/login.jsx
```diff
@@ -33,17 +33,18 @@ const LoginForm = () => {
...
                             <Form.Group controlId="password">
-                                <Form.Control type="password" placeholder="Secret Access Key"/>
+                                <Form.Control type="password" placeholder="Password / Secret Access Key"/>
                             </Form.Group>
 
                             {(!!loginError) && <Error error={loginError}/>}
 
                             <Button variant="primary" type="submit">Login</Button>
                         </Form>
+                        <Button variant="link" className="reset-pwd" onClick={()=> {router.push("/auth/resetpassword")}}>Reset Password</Button>
```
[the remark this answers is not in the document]
talSofer: done

webui/src/pages/auth/reset-password.jsx
```diff
@@ -0,0 +1,188 @@
...
+import Button from "react-bootstrap/Button";
+import {
+    AlertIcon,
+    MailIcon
+} from "@primer/octicons-react";
+import {Error} from "../../lib/components/controls";
+import {auth} from "../../lib/api";
+import validator from "validator/es";
+
+const TOKEN = "token";
```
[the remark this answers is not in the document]
talSofer: done

webui/src/pages/auth/reset-password.jsx
```diff
@@ -0,0 +1,188 @@
...
+const RequestResetPasswordForm = () => {
+
+    const onEmailChange = () => {
+        if (validator.isEmail(email.current.value)) {
+            setFormValid(true);
+            setEmailValid(true);
+        } else {
+            setFormValid(false);
+            setEmailValid(false);
+        }
```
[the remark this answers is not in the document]
talSofer: done, thanks :)

webui/src/pages/auth/reset-password.jsx
talSofer: no idea, removed the React prefix thanks

webui/src/pages/auth/reset-password.jsx
```diff
@@ -0,0 +1,188 @@
...
+                                    }
+                                </Form.Group>
+
+                                {(!!reqResetPwdError) && <Error error={reqResetPwdError}/>}
+
+                                <div>
+                                    <input type="checkbox" checked={riskAccepted} onChange={() => {
+                                        setRiskAccepted(!riskAccepted);
+                                    }}/>
+                                    <label className="accept-risk-label">I understand the risks listed above</label>
```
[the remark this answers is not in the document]
talSofer: yeah, I actually agree.. @ortz how do you want this workflow to look like? 
Options:
1. remove the message, warning sign, and checkbox altogether
2.  change "I understand the risks listed above" to "I have read the message above"

webui/src/pages/auth/reset-password.jsx
```diff
@@ -0,0 +1,188 @@
...
+        </div>
+    )
+}
+
+const ResetPasswordForm = () => {
+
+    const onConfirmPasswordChange = () => {
+        setPwdConfirmValid(true)
+        if (newPwdField.current != null) {
+            console.log("here")
```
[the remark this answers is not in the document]
talSofer: removed

webui/src/pages/auth/reset-password.jsx
ortz: It's not the first time I'm hearing this, so I think that we can drop this text and if we'll have any complaints or misalignments, we'll figure out what text to our here.
At first, this warning was for the application credentials, this is why the warning was added in the first place.

webui/src/pages/auth/reset-password.jsx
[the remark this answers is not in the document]
ortz: I agree, I think we can drop it. (detailed explanation in the comment above)

ortz (approved):

webui/src/pages/auth/reset-password.jsx
talSofer: Cool, I removed it

webui/package-lock.json
talSofer: I'm using npm 8.5.2, and node v14.18.1 (which is what our contribution guide has). 
Does this make sense that I'm getting this change?

nopcoder: https://nodejs.org/en/download/releases/
There is the node+npm version out of the box and the latest.
my concern was that we ask for node 14 that comes with an older npm, and we can use a newer node.

webui/src/lib/api/index.js
```diff
@@ -118,6 +118,36 @@ export class MergeError extends Error {
 // actual actions:
 class Auth {
 
+    async password(token, newPassword, newPasswordConfirm) {
```
nopcoder, line 121: Why do we need to post both new password and password confirm? It is something the at verify on the client side before you post a request?
There is no need to get to the server side to understand that they are not matched.

webui/src/lib/api/index.js
```diff
@@ -118,6 +118,36 @@ export class MergeError extends Error {
...
+            headers: new Headers(defaultAPIHeaders),
+            method: 'POST',
+            body: json({email: email})
+        });
+
+        if (response.status === 400) {
+            throw new BadRequestError("invalid email");
+        }
+        if (response.status !== 204) {
+            throw new Error('unknown authentication error');
```
nopcoder, line 147: failed to request password reset?

nopcoder (changes requested):

nopcoder (changes requested):
Can you explain the update password with confirmation rational?

webui/src/lib/api/index.js
talSofer: > Why do we need to post both new password and password confirm? It is something the at verify on the client side before you post a request? There is no need to get to the server-side to understand that they are not matched.

@ortz I think you wanted it to avoid someone abusing the API? but I actually don't think it is necessary.. 

> Method name should describe the action - update password by token - something like that or shorter

I agree this is bad naming, but I'm mirroring the name we have in swagger. should we change both?

webui/src/lib/api/index.js
talSofer: done

webui/src/pages/auth/reset-password.jsx
```diff
@@ -0,0 +1,162 @@
...
+            <Col md={{offset: 4, span: 4}}>
+                <Card className="reset-pwd-widget">
+                    <Card.Header>Reset Password</Card.Header>
+                    <Card.Body>
+                        <Form onSubmit={async (e) => {
+                            e.preventDefault()
+                            try {
+                                await auth.password(token, e.target.newPassword.value,
+                                    e.target.confirmPassword.value)
+                                setResetPwdError(null);
```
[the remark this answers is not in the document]
talSofer: done

webui/src/lib/api/index.js
shimi9276: I think the name in swagger is consistent with the other paths in swagger. For instance /auth/groups, when posting it creates a group, although "groups" doesn't suggest it creates a group.

talSofer: renamed it to updatePasswordByToken

ortz: If you both think it's preferred not to send it, let's drop it.

webui/src/lib/api/index.js
```diff
@@ -118,6 +118,36 @@ export class MergeError extends Error {
...
+            headers: new Headers(defaultAPIHeaders),
+            method: 'POST',
+            body: json({token: token, newPassword: newPassword})
+        });
+
+        if (response.status === 401) {
+            throw new AuthorizationError('user unauthorized');
+        }
+        if (response.status !== 201) {
+            throw new Error('unknown authentication error');
```
nopcoder, line 132: Failed to update password? or something that explain that we didn't managed to update the password.

webui/src/lib/api/index.js
```diff
@@ -764,9 +794,14 @@ class Config {
         }
     }
     async getLakeFSVersion() {
-        const response = await apiRequest('/config/version', {
-            method: 'GET',
-        });
+        try {
+            const response = await apiRequest('/config/version', {
```
nopcoder, line 798: Think it is working because the final code uses 'var' - but we declare 'response' inside the scope and reference it outside the scope.

webui/src/lib/api/index.js
```diff
@@ -764,9 +794,14 @@ class Config {
...
+                method: 'GET',
+            });
+        } catch (e) {
+            // avoid crush in case we are not logged in
+            return Promise.resolve({})
```
nopcoder, line 803: Think we need to open an issue to fix this one - screens which doesn't have user authenticated / user logged in, should not render data using authenticated API. Also, it hides the fact that we failed and the empty object response is not the response from the server. Will be hard to debug later.

webui/src/lib/api/index.js
```diff
@@ -118,6 +118,36 @@ export class MergeError extends Error {
...
+        if (response.status === 401) {
+            throw new AuthorizationError('user unauthorized');
+        }
+        if (response.status !== 201) {
+            throw new Error('unknown authentication error');
+        }
+    }
+
+    async passwordForgot(email) {
+        const response = await fetch(`${API_ENDPOINT}/auth/password/forgot`, {
```
nopcoder, line 137: wrote it also in the serve side pr - I think this flow requires captcha - it something that will require a human token as part of an unauthenticated request.

nopcoder (changes requested):

webui/src/lib/api/index.js
talSofer: opened https://github.com/treeverse/lakeFS/issues/3201, will add more details

PR #3178, opened 2022-04-07
Add API implementation for auth service

guy-har:


pkg/auth/service.go
```diff
@@ -959,3 +964,626 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
...
+		}
+		if u.EncryptedPassword != nil {
+			user.EncryptedPassword = *u.EncryptedPassword
+		}
+		return user, err
+	})
+}
+
+// getNonRequiredString returns empty string in case of nil
+func getNonRequiredString(s *string) string {
```
nopcoder, line 1076: This one can be found in swag and api packages - we can move in a different pr all the api.StringValue functions into apiutil package

pkg/auth/service.go
```diff
@@ -959,3 +964,626 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
...
+	if s == nil {
+		return ""
+	}
+	return *s
+}
+
+func toPagination(paginator Pagination) *model.Paginator {
```
nopcoder, line 1083: same as the above function - looks like a set of functions we require to build api client or api user

pkg/auth/service.go
```diff
@@ -959,3 +964,626 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
...
+	resp, err := a.apiClient.CreateGroupWithResponse(ctx, CreateGroupJSONRequestBody{
+		Id: group.DisplayName,
+	})
+	if err != nil {
+		return err
+	}
+	return expectedStatusCode(resp, 201)
+}
+
+var ErrUnexpectedStatusCode = errors.New("unexpected status code")
```
nopcoder, line 1132: errors.go?

pkg/auth/service.go
```diff
@@ -959,3 +964,626 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
...
+}
+
+func (a *APIAuthService) CreateUser(ctx context.Context, user *model.User) (int, error) {
+	resp, err := a.apiClient.CreateUserWithResponse(ctx, CreateUserJSONRequestBody{
+		Username: user.Username,
+	})
+	if err != nil {
+		return InvalidUserID, err
+	}
+	if err := expectedStatusCode(resp, 201); err != nil {
```
nopcoder, line 985: As I see it we can have one function that handles all the remote api return values (errors) and map them to the local error. Alternative will be to have a switch case per status code to handle the local error mapping.

pkg/auth/service.go
```diff
@@ -1,15 +1,20 @@
 package auth
 
+//go:generate oapi-codegen -package auth -generate "types,client,spec"  -o client.gen.go api/swagger.yml
```
nopcoder, line 3: I think it should in placed on the <project>/api directory.

pkg/auth/service.go
```diff
@@ -959,3 +964,626 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
...
 	return &AuthorizationResponse{Allowed: true}, nil
 }
+
+type APIAuthService struct {
+	apiClient   *ClientWithResponses
+	secretStore crypt.SecretStore
+	cache       Cache
+}
+
+func (a *APIAuthService) SecretStore() crypt.SecretStore {
```
nopcoder, line 974: I think the only thing the service needs to expose is the shared secret. Can be handled in a different PR.

pkg/auth/service.go
```diff
@@ -959,3 +964,626 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
...
+	}
+	if err := expectedStatusCode(resp, 201); err != nil {
+		return InvalidUserID, err
+	}
+
+	return resp.JSON201.Id, nil
+}
+
+func (a *APIAuthService) DeleteUser(ctx context.Context, username string) error {
+	u, err := a.GetUser(ctx, username)
```
nopcoder, line 993: why do we need to get the user first - this is not the remote implementation's work?

pkg/auth/service.go
```diff
@@ -959,3 +964,626 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
...
+	}
+	client, err := NewClientWithResponses(
+		apiEndpoint,
+		WithRequestEditorFn(basicAuthProvider.Intercept),
+		WithHTTPClient(httpClient),
+	)
+	if err != nil {
+		return nil, err
+	}
+	logging.Default().Info("initialized Auth service")
```
nopcoder, line 1577: consider accepting logger - or initialize one here with field "service" set to remote authorization service

pkg/auth/service.go
```diff
@@ -959,3 +964,626 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
...
+	// TODO(Guys): currently token is used, maybe we should consider using APIKey
+	basicAuthProvider, err := securityprovider.NewSecurityProviderBearerToken(token)
+	if err != nil {
+		return nil, err
+	}
+	httpClient := &http.Client{
+		// avoid redirect automatically
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
+			return http.ErrUseLastResponse
+		},
```
nopcoder, line 1567: 1. not sure this one is required for this client
2. needs a way to handle timeout or at least have a default set

pkg/auth/service.go
```diff
@@ -959,3 +964,626 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
...
+func (a *APIAuthService) ListUserPolicies(ctx context.Context, username string, params *model.PaginationParams) ([]*model.Policy, *model.Paginator, error) {
+	return a.listUserPolicies(ctx, username, params, false)
+}
+
+func (a *APIAuthService) ListEffectivePolicies(ctx context.Context, username string, params *model.PaginationParams) ([]*model.Policy, *model.Paginator, error) {
+	if params.Amount == -1 {
+		// read through the cache when requesting the full list
+		policies, err := a.cache.GetUserPolicies(username, func() ([]*model.Policy, error) {
+			policies, _, err := a.listUserPolicies(ctx, username, params, true)
+			return policies, err
```
nopcoder, line 1489: use a different names inside the scope so the reader will not get mixed. or just return the 3 of them and ignore the middle one outside

nopcoder (commented):
some comments I had while going over the code - will continue sunday morning

pkg/auth/service.go
```diff
@@ -959,3 +964,632 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
...
+	}
+	policies := make([]*model.Policy, len(resp.JSON200.Results))
+
+	for i, r := range resp.JSON200.Results {
+		policies[i] = serializePolicyToModalPolicy(r)
+	}
+	return policies, toPagination(resp.JSON200.Pagination), nil
+}
+
+func (a *APIAuthService) Authorize(ctx context.Context, req *AuthorizationRequest) (*AuthorizationResponse, error) {
```
nopcoder, line 1541: why do we need this one - looks like we need to extract it from the interface

nopcoder (changes requested):
🥇 added some comments, the main one is error handling and the second one is the authorized which looks like a client-side implementation and not server side implemented. which make you think it should be extracted out of the service that provides the authorization information.

pkg/auth/service.go
```diff
@@ -959,3 +964,632 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
...
+		}, nil
+	}
+
+	// we're allowed!
+	return &AuthorizationResponse{Allowed: true}, nil
+}
+
+func NewAPIAuthService(apiEndpoint, token string, secretStore crypt.SecretStore, cacheConf params.ServiceCache) (*APIAuthService, error) {
+	// TODO(Guys): currently token is used, maybe we should consider using APIKey
+	basicAuthProvider, err := securityprovider.NewSecurityProviderBearerToken(token)
```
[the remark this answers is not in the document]
guy-har: 1. Done
2. It will be done soon in a different PR

pkg/auth/service.go
```diff
@@ -959,3 +964,632 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
...
+	if err != nil {
+		return nil, err
+	}
+	httpClient := &http.Client{
+		// avoid redirect automatically
+		CheckRedirect: func(req *http.Request, via []*http.Request) error {
```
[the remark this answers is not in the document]
guy-har: Done

pkg/auth/service.go
```diff
@@ -959,3 +964,632 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
...
+		Username: user.Username,
+	})
+	if err != nil {
+		return InvalidUserID, err
+	}
+	if err := validateResponse(resp, 201); err != nil {
+		return InvalidUserID, err
+	}
+
+	return resp.JSON201.Id, nil
```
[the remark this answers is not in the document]
guy-har: `resp.json` could not be nil In case the return status code is 201 which is validated in `validateResponse`
@nopcoder Do you think I should add this anyway?

pkg/api/controller.go
```diff
@@ -2932,7 +2932,7 @@ func (c *Controller) GetSetupState(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	state := setupStateNotInitialized
-	if initialized {
+	if initialized || c.Config.IsAuthTypeAPI() {
```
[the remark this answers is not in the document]
guy-har: Because initialized is set on setup, which is not used or needed when using  external authentication
The UI redirects to setup page in case initialized is false.
@nopcoder would you suggest a different way to skip this?

pkg/auth/service.go
```diff
@@ -959,3 +964,632 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
...
+			return nil, err
+		}
+		u := resp.JSON200
+		return &model.User{
+			ID:                u.Id,
+			CreatedAt:         time.Unix(u.CreationDate, 0),
+			Username:          u.Name,
+			FriendlyName:      u.FriendlyName,
+			Email:             u.Email,
+			EncryptedPassword: nil,
```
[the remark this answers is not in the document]
guy-har: we authenticate by password only when they login with email.
@nopcoder Do you think we should pass it also here?

pkg/auth/service.go
```diff
@@ -959,3 +964,632 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
...
+		user := &model.User{
+			ID:           u.Id,
+			CreatedAt:    time.Unix(u.CreationDate, 0),
+			Username:     u.Name,
+			FriendlyName: u.FriendlyName,
+			Email:        u.Email,
+			Source:       getNonRequiredString(u.Source),
+		}
+		if u.EncryptedPassword != nil {
+			user.EncryptedPassword = *u.EncryptedPassword
```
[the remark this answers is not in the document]
guy-har: It's a byte array

pkg/auth/service.go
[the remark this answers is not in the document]
nopcoder: not for this pr 👍

pkg/auth/service.go
nopcoder: delete user is the only one that works with ID - the service can delete user by name

nopcoder (approved):
lgtm! optional, service can delete user by name

nopcoder (approved):

PR #3179, opened 2022-04-07
auth: generate jwt token ID

### LINKED ISSUE #3177: JWT tokens should include unique identifyer
JSON web tokens (JWTs) claims are pieces of information asserted about a subject.
jti (JWT ID) Claim: The “jti” claim gives a unique identifier for the JWT.
The “jti” value is a case-sensitive string and it should be assigned in such a manner that ensures that there is a negligible probability that the same value will be repeated.
Currently we do not populate this information into our tokens.
Using this information we can identity different tokens we created and have better audit information.

nopcoder:
Close #3177 

- New JWT token will include new UUID as ID
- Log token ID on verify user error
- Small code refactor to make it clear on claims

### DIFF
--- a/pkg/api/auth_middleware.go
+++ b/pkg/api/auth_middleware.go
@@ -132,14 +132,18 @@ func userByToken(ctx context.Context, logger logging.Logger, authService auth.Se
 	const bitSize = 32
 	id, err := strconv.ParseInt(claims.Subject, base, bitSize)
 	if err != nil {
-		logger.WithField("subject", claims.Subject).Info("could not parse user ID on token")
+		logger.WithFields(logging.Fields{
+			"subject":  claims.Subject,
+			"token_id": claims.Id,
+		}).Info("could not parse user ID on token")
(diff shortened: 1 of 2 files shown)


pkg/api/auth_middleware.go
```diff
@@ -115,13 +115,10 @@ func userByToken(ctx context.Context, logger logging.Logger, authService auth.Se
 		}
 		return authService.SecretStore().SharedSecret(), nil
 	})
-	if err != nil {
-		return nil, ErrAuthenticatingRequest
-	}
-	claims, ok := token.Claims.(*jwt.StandardClaims)
```
guy-har, line 121: Why did you remove this?

guy-har (changes requested):

pkg/api/auth_middleware.go
guy-har: OK! Thanks

guy-har (approved):

PR #3180, opened 2022-04-07
Spark 3.1.2 support

arielshaqed:
This is currently suitable *only* for internal use; do *NOT* merge to
master!

PR #3183, opened 2022-04-08
Use aws sdk v1 in spark metadata client

arielshaqed:
Really need to use 1.7.4, which is what Hadoop 2.7.7 used.  But that version is **loooooooonnnnggg** gone.  So go to 1.12.140, which a version from 2022-01 (which should alleviate some AWS log4j worries).

In any case depend on the *provided* version, so we get whatever comes with hadoop-aws (which we do need to bring in some cases).

### DIFF
--- a/clients/spark/build.sbt
+++ b/clients/spark/build.sbt
@@ -95,13 +95,17 @@ lazy val spark2Type =
   new BuildType("247", scala211Version, "2.4.7", "0.9.8", "2.7.7", "hadoop2-2.0.1")
 lazy val spark3Type =
   new BuildType("301", scala212Version, "3.0.1", "0.10.11", "2.7.7", "hadoop2-2.0.1")
+lazy val spark312Type =
+  new BuildType("312", scala212Version, "3.1.2", "0.10.11", "3.2.1", "hadoop2-2.0.1")
 
 lazy val core2 = generateCoreProject(spark2Type)
 lazy val core3 = generateCoreProject(spark3Type)
+lazy val core312 = generateCoreProject(spark312Type)
 lazy val examples2 = generateExamplesProject(spark2Type).dependsOn(core2)
 lazy val examples3 = generateExamplesProject(spark3Type).dependsOn(core3)
+lazy val examples312 = generateExamplesProject(spark312Type).dependsOn(core312)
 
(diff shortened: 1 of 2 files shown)


talSofer (approved):
How was this tested (a sanity test)?

arielshaqed:
> How was this tested (a sanity test)?

Not yet tested.  Will try to rebase #3180 on top of this, to get a better e2e test.  Also ideally some integration testing :-)

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -381,18 +388,18 @@ object S3BulkDeleter {
...
-      System.exit(1)
-    }
-    val repo = args(0)
-    val runID = args(1)
-    val region = args(2)
-    val addressesDFLocation = args(3)
-    val spark = SparkSession.builder().getOrCreate()
-    remove(repo, addressesDFLocation, runID, region, spark)
-  }
+  // def main(args: Array[String]): Unit = {
```
talSofer, line 391: How does this Spark job run without the main function?

arielshaqed: There's another "main" function that gets used.  This used to be another execution method, it was never finished -- so I unified the two objects and got rid of this one.

Sorry, will delete the comment.

clients/spark/build.sbt
```diff
@@ -95,13 +96,17 @@ lazy val spark2Type =
   new BuildType("247", scala211Version, "2.4.7", "0.9.8", "2.7.7", "hadoop2-2.0.1")
 lazy val spark3Type =
   new BuildType("301", scala212Version, "3.0.1", "0.10.11", "2.7.7", "hadoop2-2.0.1")
+lazy val spark312Type =
+  new BuildType("312", scala212Version, "3.1.2", "0.10.11", "3.2.1", "hadoop2-2.0.1")
```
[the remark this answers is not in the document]
arielshaqed: Right. Added a comment, thanks!

arielshaqed (commented):
Thanks!

PR #3186, opened 2022-04-10
fix redirect url

johnnyaug:


idanovo (approved):
👍 kind of PRs that makes me happy

PR #3192, opened 2022-04-10
docs: lakefs on kv execution plan

nopcoder:


design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Transform from using db to kv package
+
+High level interface described in the kv proposal, we should consider taking the DynamoDB interface to address the way we communicate `set-if` functionality while marshal the same value without fetching data first.
+
+Key/value format can be done using ProtoBuf / JSON.
+
+- key format based on path - a format with escaping should be supported to enable easy way to encode values as part of the key
+- each model should include version as part of the model information
+- migrate from old model to new one without downtime
```
N-o-Z, line 14: I think we need to consider the migration as a 2 step scenario and understand how / if support both steps without downtime.
1. Migration from db to kv on postgres
2. Migration from kv on postgres to another db
relevant for both OSS and cloud I think?

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
+# lakeFS on KV - Execution Plan
+
+Key/value data storage package (kv) that will supply key/value access to data to replace the current storage and locking done by Postgres.
+High level overview of the interface described in https://github.com/treeverse/lakeFS/blob/ed55edc2c24b24ed4e6fcccad1ae95844d38b9ad/design/open/metadata_kv/index.md. The package will support multiple back-end implementations.
+
+## Transform from using db to kv package
+
+High level interface described in the kv proposal, we should consider taking the DynamoDB interface to address the way we communicate `set-if` functionality while marshal the same value without fetching data first.
+
+Key/value format can be done using ProtoBuf / JSON.
```
N-o-Z, line 10: It will be very helpful if we can have a discussion on this to understand the tradeoffs between using ProtoBuf / JSON

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+
+- key format based on path - a format with escaping should be supported to enable easy way to encode values as part of the key
+- each model should include version as part of the model information
+- migrate from old model to new one without downtime
+
+## Per package changes
+
+For the following packages we should handle to move from 'db' package to 'kv'.
```
N-o-Z, line 18:
```suggestion
For the following packages we should handle the move from 'db' package to 'kv'.
```

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+During development it should enable a configuration flag to switch between using kv and db implementation.
```
N-o-Z, line 19:
```suggestion
During development, allow an internal configuration flag to switch between using kv and db implementation.
```

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+    pkg/api
+        - calls migrate as part of setup lakefs
+    pkg/catalog
+        - initialize storage with db and lock db
+    pkg/diagnostics
+        - runs list of queries to collect information on the user's environment
+
+## Global changes
+
+- Transition from db to kb should include a configuration flag that enables the use of 'kv' vs the current 'db' implementation.
```
N-o-Z, line 54:
```suggestion
- Transition from db to kv should include a configuration flag that enables the use of 'kv' vs the current 'db' implementation.
```

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+- Migration from old format performance - can't take 24h to switch to new version (downtime?)
+- New locking mechanism functionality
+
+## Milestone
+
+- implement adapter k/v for Postgres
+- feature flag (internal) used for kv development
+- each milestone will include migrate db that will use the new kv format. will support only the part we decided as feature complete and will no longer control by the feature flag.
+- for each step: implement the new models to k/v and migrate code
+- implement adapter k/v for other implementations
```
N-o-Z, line 72:
```suggestion
- Implement adapter k/v for Postgres
- Feature flag (internal) used for kv development
- Each milestone will include migrate db that will use the new kv format. will support only the part we decided as feature complete and will no longer be controlled by the feature flag.
- For each step: implement the new models to k/v and migrate code
- Implement adapter k/v for other implementations
```

N-o-Z: Do we plan on releasing versions while some modules are migrated to KV and some not? If so maybe until we complete the entire system transition to KV we should write the data using both formats as a backup until the transition is complete?

N-o-Z (commented):
Thank you for the detailed plan. I added some thoughts about the migration process I think we need to address.

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
+# lakeFS on KV - Execution Plan
+
+Key/value data storage package (kv) that will supply key/value access to data to replace the current storage and locking done by Postgres.
+High level overview of the interface described in https://github.com/treeverse/lakeFS/blob/ed55edc2c24b24ed4e6fcccad1ae95844d38b9ad/design/open/metadata_kv/index.md. The package will support multiple back-end implementations.
+
+## Transform from using db to kv package
+
+High level interface described in the kv proposal, we should consider taking the DynamoDB interface to address the way we communicate `set-if` functionality while marshal the same value without fetching data first.
```
itaiad200, line 8: Why just DynamoDB? I think Postgres is the first one, no?

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+
+Key/value format can be done using ProtoBuf / JSON.
+
+- key format based on path - a format with escaping should be supported to enable easy way to encode values as part of the key
```
itaiad200, line 12: I don't understand this line. Can you share an example? What would work, what wouldn't and why is it important.

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+- each model should include version as part of the model information
```
itaiad200, line 13: What's a `model`? This is the first reference to it in this plan..

design/open/lakefs-kv-execution-plan.md
itaiad200, line 14: Is it the same model as the previous line? Or are we talking about current implementation vs. new kv store implementation?

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+- migrate from old model to new one without downtime
+
+## Per package changes
+
+For the following packages we should handle to move from 'db' package to 'kv'.
+During development it should enable a configuration flag to switch between using kv and db implementation.
+Any move transition should support migration from old Postgres data to new Postgres data format.
```
itaiad200, line 20: What's a `move transition`? Are you referring to `new Postgres data format` per package?

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+When a complete data model / functionality completes the transition we can exclude the 'option' to swtich from/to and just use the new format.
```
itaiad200, line 21: Do you mean when all packages were migrated?

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+
+- Map the use of transactions in order to provide a valid solution using the kv interface
+- Map any database locking to specific solution using the kv implementation
+- Map models from db to key/value key and value format (ProtoBuf?)
+- Map secondary index used - ex: lookup user by id and email
```
itaiad200, line 26: What is this list? I assume it's a checklist for each package? Can we state that?

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+## Per package changes
+
+For the following packages we should handle to move from 'db' package to 'kv'.
+During development it should enable a configuration flag to switch between using kv and db implementation.
+Any move transition should support migration from old Postgres data to new Postgres data format.
+When a complete data model / functionality completes the transition we can exclude the 'option' to swtich from/to and just use the new format.
+
+- Map the use of transactions in order to provide a valid solution using the kv interface
+- Map any database locking to specific solution using the kv implementation
+- Map models from db to key/value key and value format (ProtoBuf?)
```
itaiad200, line 25: Please rephrase.

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Per package changes
+
+For the following packages we should handle to move from 'db' package to 'kv'.
+During development it should enable a configuration flag to switch between using kv and db implementation.
+Any move transition should support migration from old Postgres data to new Postgres data format.
+When a complete data model / functionality completes the transition we can exclude the 'option' to swtich from/to and just use the new format.
+
+- Map the use of transactions in order to provide a valid solution using the kv interface
+- Map any database locking to specific solution using the kv implementation
```
itaiad200, line 24: I think we need to expand on these two and offer alternatives. It basically offers no help in the kv-world, and I'm sure we can come up with more primitives on top of the existing interface that can set a common ground to solve most of the common issues around locking, transactions, secondary indexes, etc.. If each package implementor is left to build her own on top of the kv interface, we'll end up with a much more complex solution.

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+- Map models from db to key/value key and value format (ProtoBuf?)
+- Map secondary index used - ex: lookup user by id and email
+
+    pkg/auth
+        - metadata information: installation id, version and etc
+        - authorization information: users, group, policy, credentials and etc
+    pkg/actions
+        - actions information: runs, hooks, status
```
itaiad200, line 32: Add a note about the duplication with the blockstore. Might be a good time to get rid of it.

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+        - tracking gateway multipart requests
+
+    pkg/graveler/ref
+        - crud: repository, branch, commits, tags (+log)
+        - branch level locking
+        - lock free commit
+    pkg/graveler/retention
+        - uses graveler/ref to iterate over branches and commit log
+    pkg/graveler/staging
+        - staged k/v storage: key/identity/value used to get/set/drop/drop by prefix/list
```
itaiad200, line 43: Add a note referring to the suggested design in the proposal.

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+        - initialize storage with db and lock db
+    pkg/diagnostics
+        - runs list of queries to collect information on the user's environment
+
+## Global changes
+
+- Transition from db to kb should include a configuration flag that enables the use of 'kv' vs the current 'db' implementation.
+  As complete features move to use kv, and migration plan is implemented, we can remove the flag check for the feature and just enable the new functionality.
+- Migrate from current DDL to kv will be supported only for Postgres and dropped when all features works using kv.
+- Basic dump/restore functionality should be supported for the k/v implementation, how do we migrate the information - discovery
```
itaiad200, line 57: It's easy to miss this big discovery item :)
Why is it a discovery? Why isn't it straightforward? Details are missing...

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+    pkg/catalog
+        - initialize storage with db and lock db
+    pkg/diagnostics
+        - runs list of queries to collect information on the user's environment
+
+## Global changes
+
+- Transition from db to kb should include a configuration flag that enables the use of 'kv' vs the current 'db' implementation.
+  As complete features move to use kv, and migration plan is implemented, we can remove the flag check for the feature and just enable the new functionality.
+- Migrate from current DDL to kv will be supported only for Postgres and dropped when all features works using kv.
```
itaiad200, line 56: How do we migrate to a semi-implemented lakeFS version (i.e. not all packages are migrated yet)

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+- Basic dump/restore functionality should be supported for the k/v implementation, how do we migrate the information - discovery
+
+## Testing
+
+- Migration - running the same tests on data that was generated with previous db/model
```
itaiad200, line 61: What's `data` here? Are we generating it once or with every gradual migration?

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+- Functionality performance - set of tests that will check that we didn't degrade
+- Migration from old format performance - can't take 24h to switch to new version (downtime?)
+- New locking mechanism functionality
```
itaiad200, line 64: We don't have locking. What do you suggest we test here?

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Milestone
+
+- implement adapter k/v for Postgres
+- feature flag (internal) used for kv development
```
itaiad200, line 69: What's the milestone here? simply a flag?

design/open/lakefs-kv-execution-plan.md
itaiad200, line 72: I don't understand. What are the actual milestones? An example plan would be:
- Implement the KV interface for Postgres and introduce a feature flag that turns it on for implementors of kv.
- Auth package to use the KV interface (use the adapter, tests, migration code, supporting feature flag, etc.)
- api&catalog&actions&gateway&diagnostics to use KV.
- graveler to use KV.
- removing the previous implementation, feature flag deletion.

Not a milestone but should happen not after the second bullet: All tests (performance, migration, locking, alongside existing functionality tests) are in-place to check the migration.

itaiad200 (approved):
Thanks for the plan!
I think some details are missing and other needs more clarifications, other than that I can see us working according to this.

itaiad200 (changes requested):
Clicked `Approve` by mistake 😮‍💨

design/open/lakefs-kv-execution-plan.md
nopcoder: Postgres is the first one - dynamo db interface as an API reference implementation. The way dynamo handles conditional set.

design/open/lakefs-kv-execution-plan.md
nopcoder: Can give an example, or move it to a design document and drop the details from the exec plan. WDYT?

design/open/lakefs-kv-execution-plan.md
nopcoder: Rewrote this part

design/open/lakefs-kv-execution-plan.md
nopcoder: Removed the use of the word data. But just to be clear, yes - each data we keep in a table and would be modeled into the key/value the above text consider as 'model'.

design/open/lakefs-kv-execution-plan.md
itai-david: Regarding this feature flag - is it only to control the way we swotch from the old DB usage to the new one, or is it intended to allow us to go back too?
If the later, does that mean that for the entire process we should maintain the entire  data on both?

design/open/lakefs-kv-execution-plan.md
```diff
@@ -0,0 +1,72 @@
...
+        - runs list of queries to collect information on the user's environment
+
+## Global changes
+
+- Transition from db to kv should include a configuration flag that enables the use of 'kv' vs the current 'db' implementation.
+  As complete features move to use kv, and migration plan is implemented, we can remove the flag check for the feature and just enable the new functionality.
+- Migrate from current DDL to kv will be supported only for Postgres and dropped when all features works using kv.
+- Basic dump/restore functionality should be supported for the k/v implementation, how do we migrate the information - discovery
+
+## Testing
```
itai-david, line 59: I would start by mapping functionalities/APIs/etc. to tests and consider completing tests we don't have. It is a great opportunity to enrich our testing coverage

design/open/lakefs-kv-execution-plan.md
N-o-Z: @nopcoder You might have missed the comment above? I think it's important to address the issue of data integrity while in transition

nopcoder: @N-o-Z didn't go over all the comments yet, :/ still on it, thanks!

design/open/lakefs-kv-execution-plan.md
nopcoder: Migration is from table format to kv format. The first step is to migrate all the sql based implementation to kv.
The second part is to support to instruct how we migrate changes when all the data is found in kv.
Moving from one underlying implementation to another (ex: postgres to dynamodb) will be done by dump and restore.

design/open/lakefs-kv-execution-plan.md
nopcoder: - Remove the extra description about the transition - it is the migration from db postgres to kv based postgres.
- One we - just a way for us to develop the kv capability while db is still the version we used in the project until feature is complete. It should enable us to develop and test kv while we deliver a new release to the project, until feature complete.
-

design/open/lakefs-kv-execution-plan.md
nopcoder: When a feature is complete (using kv, dump/restore, migrate) as part of a milestone, can be only part of the model, we can exclude the feature flag and keep the new kv functionality without a way to go back.

design/open/lakefs-kv-execution-plan.md
nopcoder: The database locking alternative is based on set-if implementation in the current proposal - updated the text.
About the transaction - we listed couple of options: using set-if, check secondary index after reading etc.
@itaiad200 do you think we need to list the ones we discussed on the first item as possible options, or it should be part of the task / design?

design/open/lakefs-kv-execution-plan.md
nopcoder: Can you explain more about the specific issue? I'll create one or reference it.

design/open/lakefs-kv-execution-plan.md
nopcoder: I've removed the discovery part for the graveler information as part of dump.
We need to discuss the 'Pluggable ref-store' in the context of switching between implementation.
Based on the current feature - without dump/restore capability for all data (non ref related, like authorization) we will be required to implement additional information to enable such transition. @itaiad200 do you see this feature implementation include this capability or we need to state that it will not be possible at this point.

design/open/lakefs-kv-execution-plan.md
nopcoder: Auth for example, implementing the auth service to work with kv package and add migration functionality that will read all the current information from the current data and write it into kv can be deployed once as part of a milestone.
Today migrate run just sql statements - when kv package will be added additional migrate step will be added that will copy the data from the old format to the new one.

design/open/lakefs-kv-execution-plan.md
nopcoder: Updated the doc - data refer to the Postgres information. Specified the packages related.

design/open/lakefs-kv-execution-plan.md
nopcoder: Updated the doc - multiple commits, read and write while commit, commit after a failed commit.

design/open/lakefs-kv-execution-plan.md
nopcoder: Update the doc - so it doesn't reflect the above.
When I wrote the above, it meant that you will require to implement an adapter that you can control the effective implementation using a flag.

design/open/lakefs-kv-execution-plan.md
nopcoder: Added item in the 'testing' section for testing ref dump/restore  - any specific functionality you think testing is lacking?

design/open/lakefs-kv-execution-plan.md
```diff
@@ -51,22 +52,23 @@ When a complete data model / functionality completes the transition we can exclu
...
 
-- implement adapter k/v for Postgres
-- feature flag (internal) used for kv development
-- each milestone will include migrate db that will use the new kv format. will support only the part we decided as feature complete and will no longer control by the feature flag.
-- for each step: implement the new models to k/v and migrate code
-- implement adapter k/v for other implementations
+- Implement adapter k/v for Postgres. Unit test. Performance test.
+- Implement authorization, actions and multi part using kv. Include feature flag to control which implementation is active and migrate information to move from db to kv version. Unit test.
+- Implement graveler staging over kv, using kv to manage branch locking for commit and merge. Feature flag. Migrate information from db to kv for graveler (commit log, staging, ref-dump and restore). Unit test. Performance test.
+- Integration testing. Remove old implementation.
```
itaiad200, line 74: I think we're missing some packages like catalog and diagnostics.. Where do they go?

design/open/lakefs-kv-execution-plan.md
```diff
@@ -51,22 +52,23 @@ When a complete data model / functionality completes the transition we can exclu
...
-- Migration - running the same tests on data that was generated with previous db/model
-- Functionality performance - set of tests that will check that we didn't degrade
-- Migration from old format performance - can't take 24h to switch to new version (downtime?)
-- New locking mechanism functionality
+- Migration - upgrade from db to kv version and verify commit log, branch staging environment, authorization and actions data integrity.
+- Functionality - verify that all data kept on kv works with the current functionality (CRUD).
+- Ref-store dump/restore state test that old data is available and not corrupted.
+- Functionality performance - set of tests that will check that we didn't degrade.
+- Migration from old format performance - can't take 24h to switch to new version (downtime?).
+- New locking mechanism functionality. Multiple commits. Read and write while commit. Commit after a failed commit.
```
itaiad200, line 68: I would add a comment that testing is a discovery item that should consider all the above aspects, and should be ready for the first milestone (i.e. we can't really start to migrate if we don't have proper tests).

itaiad200 (approved):

design/open/lakefs-kv-execution-plan.md
```diff
@@ -5,31 +5,32 @@ High level overview of the interface described in https://github.com/treeverse/l
...
-High level interface described in the kv proposal, we should consider taking the DynamoDB interface to address the way we communicate `set-if` functionality while marshal the same value without fetching data first.
-
-Key/value format can be done using ProtoBuf / JSON.
-
-- key format based on path - a format with escaping should be supported to enable easy way to encode values as part of the key
-- each model should include version as part of the model information
-- migrate from old model to new one without downtime
+High level API described in the kv proposal, we should consider taking the DynamoDB API to address the way we communicate `set-if` functionality while marshal the same value without fetching data first.
+Key/value format can be done using ProtoBuf / JSON - discussion and information will be part of a design document.
+Each data currently stored using 'db' in a table will be migrated to a key value format. This is the first migration that will be supported on postgres. When all data is migrated from the current tables to kv, the migration will support data format changes inside the key/value.
```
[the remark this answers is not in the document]
nopcoder: Bug like any other issue we have today will have to be address before we can continue.
We can change the execution plan to include testing the new 'kv' functionality on all aspects first before migrating each package. This will include to verify that the kv functionality can handle each package requirements.
@itaiad200 WDYT?

design/open/lakefs-kv-execution-plan.md
nopcoder: - Catalog just passing the two DB connections to Graveler - so the code will be removed as we remove the 'db' package.
- Diagnostics I'll add a step - we will need to reimplement or drop  the functionality if we like to get the same information from all databases.

design/open/lakefs-kv-execution-plan.md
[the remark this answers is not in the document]
itaiad200: What do you mean by `all aspects`? 
I'm sure we'll figure out things as we go and we'll have to adapt like fixing already migrated packages implementations, changing the kv interface to support new ones, etc. We need to test with every packages we add but there's no guarantee that nothing won't change beyond that. Currently it states `When all data is migrated from the current tables to kv, the migration will support data format changes inside the key/value.` but I believe we need to have a way to change things for migrated packages even if the migration of all packages isn't complete.

N-o-Z (approved):

design/open/lakefs-kv-execution-plan.md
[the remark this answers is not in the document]
itaiad200: Just that :) the ability to fix migrated packages at all times.

design/open/lakefs-kv-execution-plan.md
nopcoder: The discovery part is just to map how we're going to test the new functionality? Because test first and implementing everything without any new functionality can be hard. Some functionality should be backed with unit test.

design/open/lakefs-kv-execution-plan.md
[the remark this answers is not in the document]
nopcoder: Line 10 state that "migration will support data format changes...". The point was to make sure that migration of kv data as part of each lakefs version should be supported at the first version of lakeFS with kv. But not during the development while we support migration from db to kv. The kv changes at this step is work in progress / beta / experimentation.

PR #3196, opened 2022-04-10
bugfix: webui server serve main page instead of page not found

### LINKED ISSUE #3194: Reload of users page returns to home page
When logging to [this](http://127.0.0.1:8000/auth/users) page and then reloading it returns to the home page (i.e. [here](http://127.0.0.1:8000/repositories) instead of reloading and returning the same page

### LINKED ISSUE #3197: UI: direct link doesn't route to the right location
Entering a direct link under lakeFS UI doesn't route the request to the desired page.
Instead, it redirects the request to the main page (render the '/repositories' page)

Expected behavior - entering the following URL on your browser:
`<lakefs>/repositories/test/objects` - should render the `test` repository objects page.
`<lakefs>/auth/credencials` - should render authorization credentials page.

nopcoder:
Revert behavior, where web UI server - instead of redirect to main page, we return the main page content.

Redirect to main page was added to handle the case of user misconfiguration - the lakectl or API client cloud identify non 2xx and report to the user that the endpoint is not configured.
A follow-up fix should address this in `lakectl doctor` as this fix will prevent the command to identify if the route is valid - getting 200 ok with a page instead of redirect.

Why did we revert to the old behavior:
The client React application uses browser router which handle the navigation on the client side.
Without having a server side that handles the routing or using HashRouter on the client, we need to serve the main page, which includes the main application to handle the URL requested by the browser.

In this change, we return our previous behavior and address defaults for some index pages.
- Redirect of repository name to browse the repository
- Redirect of non auth path to credentials
- Root level not found route will redirect to repositories

Fix #3197 
Closes #3194

### DIFF
--- a/cmd/lakectl/cmd/common_helpers.go
+++ b/cmd/lakectl/cmd/common_helpers.go
@@ -11,7 +11,6 @@ import (
 	"os"
 	"reflect"
 	"regexp"
-	"strconv"
 	"strings"
 	"text/template"
 	"time"
@@ -24,9 +23,11 @@ import (
 	"golang.org/x/term"
 )
 
-var isTerminal = true
-var noColorRequested = false
-var verboseMode = false
+var (
+	isTerminal       = true
+	noColorRequested = false
+	verboseMode      = false
+)
 
 // ErrInvalidValueInList is an error returned when a parameter of type list contains an empty string
 var ErrInvalidValueInList = errors.New("empty string in list")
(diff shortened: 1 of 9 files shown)


talSofer (approved):
Thanks @nopcoder!

nopcoder:
> Thanks @nopcoder!

Thank you - fixed the unit test, looking into the system tests.

idanovo (approved):

nopcoder:
Force pushed - the multipart test failure doesn't relate to any code change

PR #3198, opened 2022-04-11
Improve run time of GC integration tests

### LINKED ISSUE #3189: Test: System testing GC fail with timeout
The GC tests on spark takes a long time to run and fail with timeout:

```
Error: The action has timed out.
```

image

Jonathan-Rosenberg:
Closes #3189


arielshaqed (approved):
# 🐎 

That's a pretty cool speedup!  And 7 partitions seems like it should exercise the small sizes we're handling even better.

THANKS!

PR #3199, opened 2022-04-11
Add invite user

### LINKED ISSUE #3191: Create flow for user creation via email invite
This is for the basic functionality, A proper design issue will follow

shimi9276:
Adding the Backend functionality for lakefs admins to invite a user.
Closes #3191


api/swagger.yml
```diff
@@ -556,6 +556,17 @@ components:
         email:
           type: string
 
+    InviteUserRequest:
+      type: object
+      required:
+        - email
+        - id
```
talSofer, line 563: Why does an invite user request require an id? how would the user have an id at this point?

shimi9276: The id here is just the name of the user (Just like when creating a user, see auth/user/). The idea here is to preserve backend compatibility for createUser as much as possible.

nopcoder (changes requested):
discussed the flow of invite that will match password reset - waiting for the next pr

shimi9276:
Closed this one and opened a new and updated one 
See [here](https://github.com/treeverse/lakeFS/pull/3220#)

PR #3202, opened 2022-04-11
API endpoint to get email configuration

shimi9276:
Used for the frontend to check if an emailer exists and render email functionality based on that.

### DIFF
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -3295,6 +3295,7 @@ paths:
                 $ref: "#/components/schemas/VersionConfig"
         401:
           $ref: "#/components/responses/Unauthorized"
+          
   /config/storage:
     get:
       tags:
@@ -3308,5 +3309,21 @@ paths:
             application/json:
               schema:
                 $ref: "#/components/schemas/StorageConfig"
+        401:
+          $ref: "#/components/responses/Unauthorized"
+
+  /config/emailer:
+    get:
+      tags:
+        - config
+      operationId: getEmailConfig
+      description: retrieve emailer config
+      responses:
+        200:
+          description: emailer configuration
+          content:
+            application/json:
+              schema:
+                type: object
         401:
           $ref: "#/components/responses/Unauthorized"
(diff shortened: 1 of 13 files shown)


nopcoder (changes requested):
- API should describe the configuration structure
- There is no need to check if something is configured by compare of one field. Unless we add enable/disable configuration, just return the configuration the behavior is they expect while looking at the configuration.
- read email configuration should be associated to policy, so you will need to update the information or add documentation on how to configure it.

api/swagger.yml
```diff
@@ -3310,3 +3311,19 @@ paths:
...
+        - config
+      operationId: getEmailConfig
+      description: retrieve emailer config
+      responses:
+        200:
+          description: emailer configuration
+          content:
+            application/json:
+              schema:
+                type: object
```
arielshaqed, line 3327: Please specify an object type. Otherwise it will not be possible to use the API defined in this file to develop code.

pkg/api/controller.go
```diff
@@ -1124,6 +1124,22 @@ func (c *Controller) GetStorageConfig(w http.ResponseWriter, r *http.Request) {
...
+			Resource: permissions.All,
+		},
+	}) {
+		return
+	}
+	if len(c.Emailer.Params.SMTPHost) == 0 {
+		writeError(w, http.StatusNotFound, "No Emailer set")
+		return
+	}
+	writeResponse(w, http.StatusOK, c.Emailer.Params)
```
arielshaqed, line 1140: This exposes many params that must not be user visible. E.g. it may contain email server secrets.

arielshaqed (changes requested):
There are potential security issues here,  too. Let's split this one up into 2 PRs: first a well-defined Swagger-only PR (with dummy code that sends an empty response in the implementation), then a follow-on to add the functionality.

shimi9276:
Done is[ this pr](https://github.com/treeverse/lakeFS/pull/3220)

PR #3214, opened 2022-04-13
Add Id to get credentials

guy-har:


nopcoder (approved):
lgtm

arielshaqed (commented):
This change seems dangerous: it is in authorization.yml but not in swagger.yml. So it will never be clear whether we do or do not receive it. For instance if I call lakeFS from the GUI then I might receive it if lakeFS forwards the call, but not otherwise. And JavaScript is (very) dynamic.

Not sure how (or where) best to document this. You might change all struct names in authorization.yml, but that would negate the assumed benefits of rewriting an auth service rather than adding a specific mode to lakeFS. Perhaps this was addressed in the informal design of the separated auth services?

PR #3220, opened 2022-04-14
Invite user backend functionality

shimi9276:
Gives the ability to invite users to Lakefs


pkg/api/controller.go
```diff
@@ -3025,39 +3063,40 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
-
-func (c *Controller) ForgotPassword(w http.ResponseWriter, r *http.Request, body ForgotPasswordJSONRequestBody) {
-	addr, err := mail.ParseAddress(body.Email)
+func (c *Controller) resetPasswordRequest(ctx context.Context, email string) error {
+	addr, err := mail.ParseAddress(email)
 	if err != nil {
-		c.Logger.WithError(err).WithField("email", body.Email).Debug("forgot password with invalid email")
-		w.WriteHeader(http.StatusNoContent)
-		return
+		c.Logger.WithError(err).WithField("email", email).Debug("forgot password with invalid email")
```
guy-har, line 3069:
```suggestion
		c.Logger.WithError(err).WithField("email", email).Debug("reset password with invalid email")
```

pkg/api/controller.go
```diff
@@ -3025,39 +3063,40 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
 	if err != nil {
 		c.Logger.WithError(err).WithField("email", email).Warn("failed sending reset password email")
 	} else {
 		c.Logger.WithField("email", email).Info("reset password email sent")
 	}
+	return nil
+}
+
+func (c *Controller) ForgotPassword(w http.ResponseWriter, r *http.Request, body ForgotPasswordJSONRequestBody) {
+	err := c.resetPasswordRequest(r.Context(), body.Email)
```
guy-har, line 3096: log action

guy-har (changes requested):
Didn't finish going over the PR
Left a few comments

pkg/api/controller.go
```diff
@@ -817,12 +828,39 @@ func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body Cre
...
+		}
+		hp, err := model.HashPassword(p)
+		if err != nil {
+			return
+		}
+
+		e = normalizedEmail.String()
+		u.Username = e
+		u.Email = &e
+		u.EncryptedPassword = hp
```
guy-har, line 850: Could Use u.UpdatePassword

pkg/api/controller.go
```diff
@@ -817,12 +828,39 @@ func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body Cre
 		FriendlyName: nil,
 		Source:       "internal",
 	}
+	var e string
+	invite := swag.BoolValue(body.InviteUser)
+	if invite {
+		normalizedEmail, err := regexp.Compile(body.Id)
+		if err != nil {
+			return
+		}
```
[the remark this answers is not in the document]
shimi9276: Added the comment.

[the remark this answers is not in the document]
shimi9276: normalize the string. added as well as setting the email to lowercase.

pkg/api/controller.go
shimi9276: True. Fixed

pkg/api/controller.go
shimi9276: fixed

pkg/api/controller.go
shimi9276: resetPasswordRequest logs all actions or failures, so it would be double logging

pkg/api/controller.go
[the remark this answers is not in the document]
guy-har: @shimi9276 The comment was for you

You should write a response before returning from a Controllers method.
handleAPIError does that for you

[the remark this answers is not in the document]
guy-har: @shimi9276 What exactly do you mean by normalize? 
Don't think `regexp.Compile` is what your looking for

pkg/api/controller.go
```diff
@@ -3025,39 +3062,41 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
-		return
+		return err
 	}
-	// TODO (@shimi9276) create template for sending the email with link for reset
-	err = c.sendResetPasswordEmail(email, token)
+	// TODO (@shimi9276) create template for sending the email with link for reset using the passed template
+	err = c.Emailer.SendEmailWithLimit([]string{email}, token, token, nil)
 	if err != nil {
 		c.Logger.WithError(err).WithField("email", email).Warn("failed sending reset password email")
+		fmt.Println(err)
```
nopcoder, line 3088: remove print out

pkg/auth/service.go
```diff
@@ -260,8 +260,8 @@ func (s *DBAuthService) CreateUser(ctx context.Context, user *model.User) (int64
 		}
 		var id int64
 		err := tx.Get(&id,
-			`INSERT INTO auth_users (display_name, created_at, friendly_name, source) VALUES ($1, $2, $3, $4) RETURNING id`,
-			user.Username, user.CreatedAt, user.FriendlyName, user.Source)
+			`INSERT INTO auth_users (display_name, created_at, friendly_name, source, email, encrypted_password) VALUES ($1, $2, $3, $4, $5, $6) RETURNING id`,
```
nopcoder, line 263: Verify what is the current database default - if it is empty string or null in case there is no email and/or password.
We like to keep the same values in create we invite or not a user.

pkg/api/controller.go
```diff
@@ -819,10 +829,37 @@ func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body Cre
...
+		normalizedEmail, err := regexp.Compile(body.Id)
+		e := normalizedEmail.String()
+		userEmail = strings.ToLower(e)
+		if err != nil {
+			return
+		}
+		u.Username = userEmail
+		u.Email = &userEmail
+
+		p, err := nanoid.New()
```
nopcoder, line 844: Verify the following - I don't think we require generating a unique password and set it to the user.
Keep the password black, block the user ever to log in to the system until the password is reset.
We keep the encrypted password in the database, and empty is not an option that our hash function can return.
So, there is no need to set up a random password, as the user will not be able to log in to the system until the reset flow completes.

pkg/api/controller.go
```diff
@@ -3025,39 +3062,41 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
-func (c *Controller) ForgotPassword(w http.ResponseWriter, r *http.Request, body ForgotPasswordJSONRequestBody) {
-	addr, err := mail.ParseAddress(body.Email)
+func (c *Controller) resetPasswordRequest(ctx context.Context, email string) error {
+	addr, err := mail.ParseAddress(email)
 	if err != nil {
-		c.Logger.WithError(err).WithField("email", body.Email).Debug("forgot password with invalid email")
-		w.WriteHeader(http.StatusNoContent)
-		return
+		c.Logger.WithError(err).WithField("email", email).Debug("reset password with invalid email")
+		return err
```
nopcoder, line 3069: Need to handle write error response

pkg/api/controller.go
```diff
@@ -3025,39 +3062,41 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
 	token, err := GenerateJWTResetPassword(secret, email, currentTime, currentTime.Add(DefaultResetPasswordExpiration))
 	if err != nil {
 		c.Logger.WithError(err).WithField("email", email).Debug("failed to create a token")
-		writeError(w, http.StatusInternalServerError, err)
-		return
+		return err
 	}
-	// TODO (@shimi9276) create template for sending the email with link for reset
-	err = c.sendResetPasswordEmail(email, token)
+	// TODO (@shimi9276) create template for sending the email with link for reset using the passed template
```
nopcoder, line 3084: invite user and should be part of this pr

pkg/api/controller.go
```diff
@@ -3025,39 +3062,41 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
 	} else {
 		c.Logger.WithField("email", email).Info("reset password email sent")
 	}
+	return nil
+}
+
+func (c *Controller) ForgotPassword(w http.ResponseWriter, r *http.Request, body ForgotPasswordJSONRequestBody) {
+	err := c.resetPasswordRequest(r.Context(), body.Email)
+	if err != nil {
+		return
```
nopcoder, line 3098: Need to handle write error response

nopcoder (changes requested):
Pay attention, some comments are from previous PR
I know we discussed the API, but I suggest the following update:
Adding the API to invite a user after the user creation and not part of the user creation - I know it simplify the flow and I accept it.
But, when we fail to send an invitation, we are in the strange state where the caller / api need to choose how to handle it.
@talSofer will it make more sense to create a user first and request to invite by email - in case the user creation fails or just the invite the client will know how to handle the state that is left, and or least notify the user with the right state.

arielshaqed (commented):
Thanks!

We need to have the same behaviour for _all flows_ for logging the error and for writing an HTTP response.  I see too many variants here for me to be sure, but AFAICT we don't always log and we sometimes try to write *two* HTTP headers.

Also: @ortz we are beginning to add too many features to our auth. It needlessly complicates our code, and users will only have more and more requests (e.g. next will be "bulk invite users").  I believe that we should remove all this authorization code from lakeFS and instead interface with third-party auth providers.

pkg/api/controller.go
```diff
@@ -817,12 +828,39 @@ func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body Cre
 		FriendlyName: nil,
 		Source:       "internal",
 	}
+	var e string
+	invite := swag.BoolValue(body.InviteUser)
+	if invite {
+		normalizedEmail, err := regexp.Compile(body.Id)
+		if err != nil {
+			return
```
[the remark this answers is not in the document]
shimi9276: Good point. Added the proper response (403) in the creator is unauthorized.

pkg/api/controller.go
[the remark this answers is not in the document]
shimi9276: Got it. Fixed.

pkg/api/controller.go
```diff
@@ -819,10 +829,37 @@ func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body Cre
...
+	invite := swag.BoolValue(body.InviteUser)
+	if invite {
+		normalizedEmail, err := regexp.Compile(body.Id)
+		e := normalizedEmail.String()
+		userEmail = strings.ToLower(e)
+		if err != nil {
+			return
+		}
+		u.Username = userEmail
+		u.Email = &userEmail
```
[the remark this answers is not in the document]
shimi9276: I think we want to normalize and set all emails to lowercase, this way we verify the emails are in one format (this prevents the user from accidentally putting in an uppercase letter for instance, during init setup and won't be able to login only that way). This normalization is done throughout the creation and the various verifications done using the email.

pkg/api/controller.go
shimi9276: Agree, removed the password creation and just left it empty.

pkg/api/controller.go
```diff
@@ -819,10 +829,37 @@ func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body Cre
...
+
+		p, err := nanoid.New()
+		if err != nil {
```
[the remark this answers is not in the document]
shimi9276: Done

pkg/api/controller.go
```diff
@@ -819,10 +829,37 @@ func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body Cre
...
+			return
+		}
+		err = u.UpdatePassword(p)
+		if err != nil {
```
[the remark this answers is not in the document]
shimi9276: Done

pkg/api/controller.go
```diff
@@ -819,10 +829,37 @@ func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body Cre
...
+			return
+		}
+	}
 	_, err := c.Auth.CreateUser(ctx, u)
-	if handleAPIError(w, err) {
+	if handleAPIError(w, err) { //Notice that `handleAPIError` writes a response in case of error. Same for following cases
```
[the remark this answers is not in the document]
shimi9276: Done

pkg/api/controller.go
```diff
@@ -819,10 +829,37 @@ func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body Cre
...
 		return
 	}
+	if invite {
+		err = c.resetPasswordRequest(ctx, userEmail)
```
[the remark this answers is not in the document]
shimi9276: Done

pkg/api/controller.go
```diff
@@ -819,10 +829,37 @@ func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body Cre
...
+		if err != nil {
```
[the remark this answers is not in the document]
shimi9276: Done

pkg/api/controller.go
shimi9276: Not needed in the updated code (each call to this function will write the error on its own as @arielshaqed suggested)

pkg/api/controller.go
shimi9276: fixed the comment, Don't think we should do this in one PR, this will just add additional complexity,

pkg/api/controller.go
shimi9276: my bad, fixed

pkg/api/controller.go
shimi9276: Done I think

pkg/api/controller.go
shimi9276: Fixed as the calling function will do that

pkg/auth/service.go
shimi9276: That indeed is the case, if an empty string is passed or nil, it is simply left blank as until now

shimi9276 (commented):
Tried my best to respond to all comments (if missed any or misunderstood please let me know). Did basic manual sanity checks, but haven't written any tests as this will be done in a later stage. @nopcoder I understand the problem with the state in the case the email failed, we can definitely discuss once more the invite flow, but as of now I stuck to the original agreed flow.

api/swagger.yml
```diff
@@ -1134,6 +1144,24 @@ paths:
           default:
             $ref: "#/components/responses/ServerError"
```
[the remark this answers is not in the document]
nopcoder: we said that this is called from a place where the user forgot this password. so this request can't be authenticated. you need to specify in the swagger not to check this route.

arielshaqed (commented):
Haven't had time yet to do a full review. However I can see that I will need help to understand username normalization: we do it when handling emails but not when logging in. I believe documentation will be needed here to explain what the code is doing. Also I have difficulty finding documentation on how Unicode NFC is needed for string lookups in Postgres, or how our Postgres dB drivers interact with this and need it.

pkg/api/controller.go
```diff
@@ -802,13 +811,28 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
...
+}
+
 func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body CreateUserJSONRequestBody) {
 	if !c.authorize(w, r, permissions.Node{
 		Permission: permissions.Permission{
 			Action:   permissions.CreateUserAction,
 			Resource: permissions.UserArn(body.Id),
 		},
 	}) {
+		writeResponse(w, http.StatusUnauthorized, nil)
```
nopcoder, line 835: why? this is what authorize method does

pkg/api/controller.go
```diff
@@ -817,12 +841,27 @@ func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body Cre
 		FriendlyName: nil,
 		Source:       "internal",
 	}
+	normalizedEmail := normalizeAndSetToLowercase(body.Id)
```
nopcoder, line 844: 1. At this point this is also a user-provided data - don't normalize it here
2. You need to extract the email - verify it is an email and lower case (no need for the utf8 norm) before you check permission or initialize the user.
3. The logic in this pr changes how we treat user email at the email field - until now we respected the email case and kept it inside the user's email.

pkg/api/controller.go
```diff
@@ -817,12 +841,27 @@ func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body Cre
...
 	c.LogAction(ctx, "create_user")
 	_, err := c.Auth.CreateUser(ctx, u)
+
 	if handleAPIError(w, err) {
 		return
 	}
+
+	if invite {
+		err = c.inviteUserRequest(ctx, normalizedEmail)
+		if handleAPIError(w, err) {
```
nopcoder, line 860: Saw your answer about changing the api - which I don't think it is a big change.
Keep the current implementation the client will have to delete a user which it doesn't know the user ID after in the case we fail to send an email.

pkg/api/controller.go
```diff
@@ -3025,39 +3064,41 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
 }
 
-func (c *Controller) sendResetPasswordEmail(email string, token string) error {
-	return c.Emailer.SendEmailWithLimit([]string{email}, token, token, nil)
-}
-
-func (c *Controller) ForgotPassword(w http.ResponseWriter, r *http.Request, body ForgotPasswordJSONRequestBody) {
-	addr, err := mail.ParseAddress(body.Email)
+func (c *Controller) resetPasswordRequest(ctx context.Context, email string) error {
+	email = normalizeAndSetToLowercase(email)
```
nopcoder, line 3068: we should not do that at this point - unless we changed our decision to keep the user's email case.

pkg/api/controller.go
```diff
@@ -817,12 +841,27 @@ func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body Cre
...
 	ctx := r.Context()
 	c.LogAction(ctx, "create_user")
 	_, err := c.Auth.CreateUser(ctx, u)
+
 	if handleAPIError(w, err) {
 		return
 	}
+
+	if invite {
+		err = c.inviteUserRequest(ctx, normalizedEmail)
```
nopcoder, line 859: This method should just send invite

pkg/api/controller.go
nopcoder: Please check the code - the function returns an err - so it write an error as a response or return one - for each case something is missing here. No content is also a response.

nopcoder (changes requested):

pkg/api/controller.go
```diff
@@ -37,6 +37,7 @@ import (
 	"github.com/treeverse/lakefs/pkg/stats"
 	"github.com/treeverse/lakefs/pkg/upload"
 	"github.com/treeverse/lakefs/pkg/version"
+	"golang.org/x/text/unicode/norm"
```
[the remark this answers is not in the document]
shimi9276: Done.

api/swagger.yml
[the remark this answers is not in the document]
shimi9276: I'm afraid I still did not understand, are we are referring to server error in the case of forgot password? If that is the case, why shouldn't we check server error (this can happen if the token generation failed for instance)?

pkg/api/controller.go
```diff
@@ -802,13 +811,28 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
 	writeResponse(w, http.StatusOK, response)
 }
 
+func normalizeAndSetToLowercase(s string) string {
+	normalize := norm.NFC.String(s)
+	return strings.ToLower(normalize)
+}
+
+func setUsernameAndEmail(user *model.User, email string) {
```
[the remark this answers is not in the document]
shimi9276: Fixed

pkg/api/controller.go
```diff
@@ -802,13 +811,28 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
...
+	user.Username = email
+	user.Email = &email
+}
+
+func (c *Controller) inviteUserRequest(ctx context.Context, email string) error {
+	return c.resetPasswordRequest(ctx, email)
+}
```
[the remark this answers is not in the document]
shimi9276: Fixed

pkg/api/controller.go
shimi9276: Fixed

pkg/api/controller.go
shimi9276: Got it. fixed

pkg/api/controller.go
shimi9276: Fixed

pkg/api/controller.go
shimi9276: Got it, will be in a separate PR, by how will be decided.

pkg/api/controller.go
```diff
@@ -817,12 +841,27 @@ func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body Cre
...
 	_, err := c.Auth.CreateUser(ctx, u)
+
 	if handleAPIError(w, err) {
 		return
 	}
+
+	if invite {
+		err = c.inviteUserRequest(ctx, normalizedEmail)
+		if handleAPIError(w, err) {
+			c.Logger.WithError(err).WithField("email", normalizedEmail).Warn("failed sending reset password email during invite user")
```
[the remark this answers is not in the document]
shimi9276: Fixed

pkg/api/controller.go
shimi9276: Indeed. fixed

pkg/api/controller.go
shimi9276: your right, fixed

pkg/auth/service.go
shimi9276: fixed

shimi9276 (commented):
@arielshaqed string normalization to utf8 will not be used, so no concern on that I hope.

pkg/api/controller.go
```diff
@@ -812,27 +811,35 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
...
 
 func normalizeAndSetToLowercase(s string) string {
-	normalize := norm.NFC.String(s)
-	return strings.ToLower(normalize)
-}
-
-func setUsernameAndEmail(user *model.User, email string) {
-	user.Username = email
-	user.Email = &email
+	return strings.ToLower(s)
```
arielshaqed, line 814: Sorry, I am still confused.  Is there any documentation for this feature that would let me un-confuse myself without repeatedly asking roughly the same questions?

This time I do not understand why we lowercase an email address.  I am reading [RFC 5321 section 2.4](https://www.rfc-editor.org/rfc/rfc5321.html#section-2.4):

> The
   local-part of a mailbox MUST BE treated as case sensitive.
   Therefore, SMTP implementations MUST take care to preserve the case
   of mailbox local-parts.  In particular, for some hosts, the user
   "smith" is different from the user "Smith".  However, exploiting the
   case sensitivity of mailbox local-parts impedes interoperability and
   is discouraged.  Mailbox domains follow normal DNS rules and are
   hence not case sensitive.

To be very clear, [2.3.11](https://www.rfc-editor.org/rfc/rfc5321.html#section-2.3.11) there says:

> The
   standard mailbox naming convention is defined to be
   "local-part@domain"; contemporary usage permits a much broader set of
   applications than simple "user names".  Consequently, and due to a
   long history of problems when intermediate hosts have attempted to
   optimize transport by modifying them, the local-part MUST be
   interpreted and assigned semantics only by the host specified in the
   domain part of the address.

What is this feature attempting to accomplish?

pkg/api/controller.go
```diff
@@ -812,27 +811,35 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
 }
 
 func normalizeAndSetToLowercase(s string) string {
```
arielshaqed, line 813: (Assuming this function is needed, see below:) This function is now misnamed.

pkg/api/controller.go
```diff
@@ -82,6 +82,14 @@ type Controller struct {
 	Emailer               *email.Emailer
 }
 
+func (c *Controller) GetAuthCapabilities(w http.ResponseWriter, r *http.Request) {
+	emailSupported := c.Emailer.Params.SMTPHost != ""
```
arielshaqed, line 86: Nit: Is there a better way to switch off email than to remove some field from the configuration?  Many features allow explicitly disabling or enabling them.  That would allow me "just" to disable email without having to delete or comment-out the SMTP host configuration (which could be tricky to remember)

pkg/api/controller.go
```diff
@@ -817,12 +848,26 @@ func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body Cre
...
+	if invite {
+		u.Username = normalizedEmail
+		u.Email = &body.Id
+	}
 	ctx := r.Context()
 	c.LogAction(ctx, "create_user")
 	_, err := c.Auth.CreateUser(ctx, u)
+
 	if handleAPIError(w, err) {
 		return
```
arielshaqed, line 861: Other calls to `handleAPIError` log an error, this one does not.  Would prefer consistency here.

arielshaqed (changes requested):
Thanks!  I believe this now reports and logs all errors, except for one case I found that does not get logged.  I think the next round should have all cases covered.

I remain disoriented by the email normalization code.  I do not understand the requirement.  It seems to me that each of the 3 different versions of email normalization so far has had issues.

Right now we are arguing concurrently about the _requirement_ and about its _implementation_.  I would like to be able to settle first the one and then the other.  I believe that the fastest way to do this would be first to write up a separate requirement and only then to implement it.  (This requirement could be an issue, or a note on a design doc for this feature, or a thread on public Slack, or anywhere really.  My request here is to have a single explicit requirement for what this is trying to do.)

pkg/api/controller.go
```diff
@@ -3068,14 +3113,17 @@ func (c *Controller) UpdatePassword(w http.ResponseWriter, r *http.Request, body
...
-	tokenEmail := claims.Subject
-	user, err := c.Auth.GetUserByEmail(r.Context(), tokenEmail)
+	user, err := c.Auth.GetUserByEmail(r.Context(), claims.Subject)
 	if err != nil {
-		c.Logger.WithError(err).WithField("email", tokenEmail).Debug("failed to retrieve user by email")
+		c.Logger.WithError(err).WithField("email", claims.Subject).Debug("failed to retrieve user by email")
+		writeError(w, http.StatusNotFound, http.StatusText(http.StatusNotFound))
+		return
+	}
+	if user.Username != normalizeAndSetToLowercase(claims.Subject) {
```
nopcoder, line 3122: Why do we need this one? If the user got a token that allows a change of password - any user that we will get back from `GetUserByEmail` should be matched this description.

pkg/auth/service.go
```diff
@@ -260,8 +260,8 @@ func (s *DBAuthService) CreateUser(ctx context.Context, user *model.User) (int64
 		}
 		var id int64
 		err := tx.Get(&id,
-			`INSERT INTO auth_users (display_name, created_at, friendly_name, source) VALUES ($1, $2, $3, $4) RETURNING id`,
-			user.Username, user.CreatedAt, user.FriendlyName, user.Source)
+			`INSERT INTO auth_users (display_name, created_at, friendly_name, source, email) VALUES ($1, $2, $3, $4, $5) RETURNING id`,
+			user.Username, user.CreatedAt, user.FriendlyName, user.Source, user.Email)
```
nopcoder, line 264: Please verify that is the table default value for email and if we are now setting null as email can be a problem, or if it is aligned to the value we insert for users without email.

pkg/api/controller.go
```diff
@@ -3025,39 +3070,39 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
-	// TODO (@shimi9276) create template for sending the email with link for reset
-	err = c.sendResetPasswordEmail(email, token)
+	// TODO (@shimi9276) create template for sending the email with link for reset using the passed template
+	err = c.Emailer.SendEmailWithLimit([]string{email}, token, token, nil)
 	if err != nil {
-		c.Logger.WithError(err).WithField("email", email).Warn("failed sending reset password email")
+		return err
 	} else {
 		c.Logger.WithField("email", email).Info("reset password email sent")
 	}
```
nopcoder, line 3096: Don't need else after return

nopcoder (changes requested):

pkg/api/controller.go
shimi9276: The purpose here is the opposite, we want to verify the user definitely set up an email server and the most simple way is to see if one of such was configured. I do agree we should have such a switch to enable and disable but I don't think this is in the scope of this PR.

pkg/api/controller.go
shimi9276: Correct, was removed all together

pkg/api/controller.go
shimi9276: You are right, as far as email sending goes the SMTP server must be case insensitive. Let me try to explain. As of today, the unique identifier for a user was the username or display name (sometimes referred to as user id as well). In the invite user process all we know about the new user is his email. As such we don't have a better uniquely identify the user only by the given email. In order to retain it's uniqueness, we must set it all to lowercase to avoid having the same email twice in the system with different caps...

pkg/api/controller.go
```diff
@@ -802,11 +810,34 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
 	writeResponse(w, http.StatusOK, response)
 }
 
+func normalizeAndSetToLowercase(s string) string {
+	return strings.ToLower(s)
+}
+
+func (c *Controller) inviteUserRequest(ctx context.Context, email string) error {
```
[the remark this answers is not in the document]
shimi9276: Fixed

pkg/api/controller.go
```diff
@@ -802,11 +810,34 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
...
+	secret := c.Auth.SecretStore().SharedSecret()
+	currentTime := time.Now()
+	token, err := GenerateJWTResetPassword(secret, email, currentTime, currentTime.Add(DefaultResetPasswordExpiration))
+	if err != nil {
+		c.Logger.WithError(err).WithField("email", email).Debug("failed to create a token")
```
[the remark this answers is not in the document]
shimi9276: Fixed

pkg/api/controller.go
```diff
@@ -802,11 +810,34 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
...
+		return err
+	}
```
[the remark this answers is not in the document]
shimi9276: Done

pkg/api/controller.go
```diff
@@ -802,11 +810,34 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
...
+	// TODO (@shimi9276) create template for sending the email with link for invite using the passed template
+	err = c.Emailer.SendEmailWithLimit([]string{email}, token, token, nil)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
shimi9276: True, removed the first log. It will be done by the calling function

pkg/api/controller.go
```diff
@@ -802,11 +810,34 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
...
+	} else {
+		c.Logger.WithField("email", email).Info("invite email sent")
+	}
+	return nil
+}
+
 func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body CreateUserJSONRequestBody) {
+	normalizedEmail := normalizeAndSetToLowercase(body.Id)
```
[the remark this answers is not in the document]
shimi9276: Fixed

pkg/api/controller.go
```diff
@@ -817,12 +848,26 @@ func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body Cre
 		FriendlyName: nil,
 		Source:       "internal",
 	}
+	invite := swag.BoolValue(body.InviteUser)
+	if invite {
+		u.Username = normalizedEmail
+		u.Email = &body.Id
```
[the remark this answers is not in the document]
shimi9276: Fixed

pkg/api/controller.go
shimi9276: I haven't written this code or modified it, but I think the reason here is that two lines before we` c.LogAction(ctx, "create_user")`. If you feel we should log more explicitly, I can definitely add that

pkg/api/controller.go
shimi9276: Fixed

pkg/api/controller.go
shimi9276: True, Removed

pkg/auth/service.go
shimi9276: So both are true :). It is set as NULL and this is also the value for users without email

shimi9276 (commented):
@arielshaqed regarding your first question, normalization is for uniqueness (elaborated more on by the comment).
Regarding the requirement's we had a sync session between all parties involved and this was the outcome (putting it here and we can continue the debate offline with the involved parties if necessary. )
In the UI the admin or one with proper credentials can click on invite user, he will then be prompted to enter the new users email. Once clicking invite, it will trigger the API endpoint of create user with the optional field of "invite" set to "true" (this does not affect create user, so in the case of "create user" simply "invite" value is not passed and therefore create user continues as usual). Once the API is called with invite set to true, a user will be created with the email as the user/display name. It will as well update the email and send an invite to the user. The link in the update email will request he reset his password (which is unset up to that point and unusable since it cannot be hashed) and hence make an API call to update password (with a different UI page of course)which is already implemented. I hope this is clarifies.

pkg/api/controller.go
```diff
@@ -802,27 +810,76 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
...
 }
 
+func (c *Controller) generateResetPasswordToken(email string) (*string, error) {
+	secret := c.Auth.SecretStore().SharedSecret()
+	currentTime := time.Now()
+	token, err := GenerateJWTResetPassword(secret, email, currentTime, currentTime.Add(DefaultResetPasswordExpiration))
+	if err != nil {
+		return nil, err
+	}
+	return &token, nil
```
nopcoder, line 820: Prefer the shorter version - I know it will modify the return type, but in the case of string it is common
```suggestion
	return GenerateJWTResetPassword(secret, email, currentTime, currentTime.Add(DefaultResetPasswordExpiration))
```

pkg/api/controller.go
```diff
@@ -802,27 +810,76 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
...
+	token, err := c.generateResetPasswordToken(email)
+	if err != nil {
+		return err
+	}
+	err = c.Emailer.SendEmailWithLimit([]string{email}, *token, *token, nil)
+	if err != nil {
+		return err
+	} else {
+		c.Logger.WithField("email", email).Info("invite email sent")
+	}
```
nopcoder, line 834: no need to else after return - already commended about this one in the previous pr

pkg/api/controller.go
```diff
@@ -802,27 +810,76 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
...
+}
+
 func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body CreateUserJSONRequestBody) {
+	invite := swag.BoolValue(body.InviteUser)
+	id := body.Id
+	var addr *mail.Address
+	var err error
+	if invite {
+		addr, err = mail.ParseAddress(body.Id)
+		if handleAPIError(w, err) {
```
nopcoder, line 845: handleAPIError will not know how to handle invalid address error - in case of any error, you need to write back validation error.
The error we write here should describe to the caller that the invite user id is in invalid email format.

pkg/api/controller.go
```diff
@@ -802,27 +810,76 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
...
+			c.Logger.WithError(err).WithField("email", body.Id).Warn("failed parsing email")
```
nopcoder, line 846: I would update the message and the field name - the caller passed 'id' field. When you will read the log later, will need to get to the code and re-learn why it is talking about 'email' field when you passed just an ID.
The validation is done to the email posted as part of the 'id' with invite user.

pkg/api/controller.go
arielshaqed: Yes, now that there are multiple ways to fail it seems strange to log all the new ways but not the old one.

pkg/api/controller.go
```diff
@@ -802,27 +810,76 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
...
+	} else {
+		c.Logger.WithField("email", email).Info("invite email sent")
+	}
+	return nil
+}
+
 func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body CreateUserJSONRequestBody) {
+	invite := swag.BoolValue(body.InviteUser)
+	id := body.Id
+	var addr *mail.Address
```
nopcoder, line 841: Suggestion - if you keep the variable called 'email' as `*string` - you can use it when you construct the user later, and it will get the addr value. Without checking 'invite' again.
This way, you will not require the current `err` and `addr` will be limited to the invite scope, and 'email' will be in the function one.

arielshaqed (commented):
About caps: AFAIK the reference that I sent is both definitive and normative. It is *literally* the definition of "an email address". It also gives guidance on what you may assume about it.
It is not possible to identify uniquely a user by an email address. One should not try to,  either. E.g. on Gmail you can freely add periods,  or change case, or append a plus sign followed bt almost arbitrary text - but not on other servers.

pkg/api/controller.go
```diff
@@ -802,27 +810,76 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
...
 	c.LogAction(ctx, "create_user")
-	_, err := c.Auth.CreateUser(ctx, u)
+	_, err = c.Auth.CreateUser(ctx, u)
+
 	if handleAPIError(w, err) {
 		return
 	}
+
+	if invite {
+		err = c.inviteUserRequest(id)
```
nopcoder, line 877: Use the email/addr here and not the id we created for the user. You can also get the created user and pass the email created.

pkg/api/controller.go
```diff
@@ -802,27 +810,76 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
...
+		if handleAPIError(w, err) {
+			c.Logger.WithError(err).WithField("email", id).Warn("failed sending invite email")
```
nopcoder, line 879: The 'email' field should log the email address not the id we created by the email address.

nopcoder (changes requested):

pkg/api/controller.go
shimi9276: Fixed

pkg/api/controller.go
shimi9276: Fixed, missed the change in both places

pkg/api/controller.go
shimi9276: Done, cool idea BTW.

pkg/api/controller.go
shimi9276: I did check this one out and tried to invite a user with an empty email, it returned 500 (internal server error) with the proper error (i.e. no address), so I think it should be pretty clear what the error is, In any case, I replaced it with a more descriptive response as you suggested.

pkg/api/controller.go
shimi9276: Agree, fixed

pkg/api/controller.go
shimi9276: Done

pkg/api/controller.go
shimi9276: Done, used the email from user struct

pkg/api/controller.go
shimi9276: Done.

shimi9276 (commented):
@arielshaqed, I agree with you that we should not use an email address for **both** email and userID **at once**. What we did here instead is use the email twice, meaning, we write once the email in the user email field and column in the DB. In this case we do not modify it, only insert it as is (as you rightfully mentioned, the SMTP server should deal with caps).
We use the email once more as the userID, (which as far as we care this can be a random string), but since this is for internal use (i.e. to id the user, and not for sending email purposes), we must verify its uniqueness, and therefor we must lowercase it.

pkg/api/controller.go
```diff
@@ -802,27 +810,71 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
...
 		FriendlyName: nil,
 		Source:       "internal",
+		Email:        email,
 	}
 	ctx := r.Context()
 	c.LogAction(ctx, "create_user")
 	_, err := c.Auth.CreateUser(ctx, u)
+
 	if handleAPIError(w, err) {
+		c.Logger.WithError(err).WithField("email", id).Warn("failed creating user")
```
nopcoder, line 867: This is the general flow - "user_id"

nopcoder (changes requested):

pkg/api/controller.go
shimi9276: At this point id == u.ID, but I do agree that for readability sake we should pass u.ID. Fixed

pkg/api/controller.go
```diff
@@ -802,27 +810,71 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
...
+	var email *string
+	if invite {
+		addr, err := mail.ParseAddress(body.Id)
+		if err != nil {
+			c.Logger.WithError(err).WithField("userID", id).Warn("failed parsing email")
+			writeError(w, http.StatusBadRequest, "Invalid email format")
+			return
+		}
+		id = strings.ToLower(addr.Address)
+		email = &body.Id
```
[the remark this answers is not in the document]
shimi9276: Not sure we need to do that, because if it is an invalid email address it would have failed in the previous step (by the parsing). In any case, passed &addr.Address which is definitely a valid email.

nopcoder (changes requested):

pkg/api/controller.go
```diff
@@ -3025,38 +3077,35 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+		return err
+	}
+	c.Logger.WithField("email", email).Info("reset password email sent")
+
+	return nil
+}
+
+func (c *Controller) ForgotPassword(w http.ResponseWriter, r *http.Request, body ForgotPasswordJSONRequestBody) {
+	err := c.resetPasswordRequest(r.Context(), body.Email)
+	if handleAPIError(w, err) {
```
nopcoder, line 3106: Handle API error at this point is not good as it will return an answer to the user based on err value - in this case if we fail to lookup the user it will render an error saying that - which is a bad practice for non secure route.

pkg/api/controller.go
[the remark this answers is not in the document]
arielshaqed: I believe @nopcoder refers to the difference between "addr-spec" (what you are looking for) and "address" or "mailbox" (what the code parses).  See links on my looooong comment appearing soon below.

arielshaqed (approved):
I remain totally confused by the uniqueness requirements and what exactly goes into the "ID" field.  I hope that documentation of that is forthcoming.  So far I have requested it several times, but after >100 comments I will just give up and open a separate issue instead.

I am approving as I see no security bugs beyond those indicated by @nopcoder below in reporting errors.

Thanks!

pkg/api/controller.go
```diff
@@ -802,27 +810,71 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
...
+	c.Logger.WithField("email", email).Info("invite email sent")
+	return nil
+}
+
 func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body CreateUserJSONRequestBody) {
+	invite := swag.BoolValue(body.InviteUser)
+	id := body.Id
+	var email *string
+	if invite {
+		addr, err := mail.ParseAddress(body.Id)
```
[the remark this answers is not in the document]
shimi9276: I do agree Passing addr.Address, (as it is today) should avoid the problem.

pkg/api/controller.go
```diff
@@ -802,27 +810,71 @@ func (c *Controller) ListUsers(w http.ResponseWriter, r *http.Request, params Li
...
+		if err != nil {
+			c.Logger.WithError(err).WithField("userID", id).Warn("failed parsing email")
```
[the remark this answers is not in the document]
shimi9276: Fixed

pkg/api/controller.go
```diff
@@ -864,7 +864,7 @@ func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body Cre
 	_, err := c.Auth.CreateUser(ctx, u)
 
 	if handleAPIError(w, err) {
-		c.Logger.WithError(err).WithField("email", id).Warn("failed creating user")
+		c.Logger.WithError(err).WithField("user ID", u.ID).Warn("failed creating user")
```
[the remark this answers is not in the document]
shimi9276: Fixed

pkg/api/controller.go
shimi9276: Good point, fixed, now in any case (error or not) it will simply write no content

pkg/api/controller.go
```diff
@@ -3025,38 +3077,34 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
+		return err
+	}
+	c.Logger.WithField("email", email).Info("reset password email sent")
+
+	return nil
+}
+
+func (c *Controller) ForgotPassword(w http.ResponseWriter, r *http.Request, body ForgotPasswordJSONRequestBody) {
+	err := c.resetPasswordRequest(r.Context(), body.Email)
+	if err != nil {
```
nopcoder, line 3106: This will ignore all errors - when the UI will send invalid email - failed to parse, this is something we mapped into the API and we can return 400

nopcoder (changes requested):

pkg/api/controller.go
shimi9276: Done

pkg/api/controller.go
```diff
@@ -3105,6 +3105,7 @@ func (c *Controller) ForgotPassword(w http.ResponseWriter, r *http.Request, body
 	err := c.resetPasswordRequest(r.Context(), body.Email)
 	if err != nil {
 		c.Logger.WithError(err).WithField("email", body.Email).Debug("failed sending reset password email")
+		writeError(w, http.StatusBadRequest, "bad request")
```
nopcoder, line 3108: 1. you write two responses
2. you don't want to return 400 for all these errors - just for invalid email format - not helping me understand which email exists or not

nopcoder (changes requested):

pkg/api/controller.go
shimi9276: Fixed

pkg/api/controller.go
```diff
@@ -3102,10 +3098,14 @@ func (c *Controller) resetPasswordRequest(ctx context.Context, email string) err
 }
 
 func (c *Controller) ForgotPassword(w http.ResponseWriter, r *http.Request, body ForgotPasswordJSONRequestBody) {
-	err := c.resetPasswordRequest(r.Context(), body.Email)
+	addr, err := mail.ParseAddress(body.Email)
 	if err != nil {
-		c.Logger.WithError(err).WithField("email", body.Email).Debug("failed sending reset password email")
 		writeError(w, http.StatusBadRequest, "bad request")
```
nopcoder, line 3103: should be "invalid email"

nopcoder (approved):
There were a lot of iterations and I hope I didn't missed anything - really need to write tests to back this flow and document the expected behavior.

guy-har (approved):
Approving because this was approved and I don't want to block 
But please go over the comments, open Issues for the cases that were out of the scope of this PR (If it wasn't done yet) 
And check the fact that auth capabilities needs authentication ( could it work for reset password?)

api/swagger.yml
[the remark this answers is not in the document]
shimi9276: Thank for pointing this out, missed this one, fixed now

pkg/api/controller.go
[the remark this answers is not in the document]
shimi9276: No, it will create a user with just the email `bg@example.com`, i.e. `Barry Gibbs` is not used at all. We use just the email address. Proper documentation is forthcoming.

pkg/api/controller.go
```diff
@@ -3025,38 +3077,35 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
-		c.Logger.WithError(err).WithField("email", addr.Address).Debug("failed to retrieve user by email")
-		w.WriteHeader(http.StatusNoContent)
-		return
+		return err
 	}
-	email := StringValue(user.Email)
-	secret := c.Auth.SecretStore().SharedSecret()
-	currentTime := time.Now()
-	token, err := GenerateJWTResetPassword(secret, email, currentTime, currentTime.Add(DefaultResetPasswordExpiration))
+	err = c.Emailer.SendEmailWithLimit([]string{email}, token, token, nil)
```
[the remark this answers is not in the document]
shimi9276: Agreed, the template PR should be out soon.

shimi9276 (commented):
Thanks for approving @guy-har.

nopcoder (approved):

pkg/api/controller.go
```diff
@@ -3025,38 +3077,35 @@ func (c *Controller) GetCurrentUser(w http.ResponseWriter, r *http.Request) {
...
 		return
 	}
-	// TODO (@shimi9276) create template for sending the email with link for reset
-	err = c.sendResetPasswordEmail(email, token)
+	err = c.resetPasswordRequest(r.Context(), addr.Address)
 	if err != nil {
-		c.Logger.WithError(err).WithField("email", email).Warn("failed sending reset password email")
-	} else {
-		c.Logger.WithField("email", email).Info("reset password email sent")
+		c.Logger.WithError(err).WithField("email", body.Email).Debug("failed sending reset password email")
```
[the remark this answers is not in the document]
shimi9276: This is done for security purposes, as this an exposed endpoint we don't want it to be used as an attack vector to see which emails are in the system, so doesn't matter if successful or not, we always return 204.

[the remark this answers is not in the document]
guy-har: @shimi9276, I understand. But would expect it to be documented in the code, so nobody will accidentally "fix" this

PR #3221, opened 2022-04-16
GC: Flaky tests fix

Jonathan-Rosenberg:
Currently the integration tests for the garbage collector are a bit flaky... They sometimes fail when trying to validate that files that aren't supposed to be deleted are in fact not deleted. 
This PR tries to fix that by testing the "main" branch for the files and not the branch that the test was using (the files should always be on the main branch).

### DIFF
--- a/test/spark/run-gc-test.sh
+++ b/test/spark/run-gc-test.sh
@@ -79,9 +79,9 @@ validate_gc_job() {
     if [[ ${days_ago} -gt -1 ]]; then
       local branch_name=$(echo ${branch_props} | jq -r '.branch_name')
       for location in \
-        lakefs://${repo}/${branch_name}/not_deleted_file1 \
-        lakefs://${repo}/${branch_name}/not_deleted_file2 \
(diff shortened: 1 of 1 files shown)


johnnyaug:
@Jonathan-Rosenberg, I don't understand why this should help - can you clarify?

Jonathan-Rosenberg:
@johnnyaug this is actually a test to see if it has something to do with querying the tested branches. 
I should've written it in the description...
If it won't help, it won't hurt, and we'll see the results later on when enough GitHub checks will run through it.

johnnyaug (approved):

test/spark/run-gc-test.sh
```diff
@@ -79,9 +79,9 @@ validate_gc_job() {
     if [[ ${days_ago} -gt -1 ]]; then
       local branch_name=$(echo ${branch_props} | jq -r '.branch_name')
```
arielshaqed, line 80: No longer sure: what does this do now?

arielshaqed (commented):
Late to the party :-(

PR #3222, opened 2022-04-17
docs - nginx client body size limitation

itaiad200:


docs/deploy/k8s.md
```diff
@@ -113,6 +113,9 @@ You should have a load balancer direct requests to the lakeFS server.
 Options to do so include a Kubernetes Service of type `LoadBalancer`, or a Kubernetes Ingress.
 By default, lakeFS operates on port 8000, and exposes a `/_health` endpoint which you can use for health checks.
 
+Nginx Ingress Controller has a default client's body size of 1 MB.
```
arielshaqed, line 116:
```suggestion
The NGINX Ingress Controller by default limits the client body size to 1 MiB.
```

But I would consider calling it out more explicitly, e.g.
```suggestion
The NGINX Ingress Controller by default limits the client body size to 1 MiB.  This is not useful for working with large objects, and should be increased at least to 5 MiB to allow multipart uploads, and to much more for safe usage.
```

arielshaqed (approved):
Important stuff!

ozkatz (approved):
Approving because it's much better than before! Thanks :)

Some things I believe could improve it even further:

1. Explain which API calls benefit from this (i.e. put/get object, s3 gtw usage, etc) and why
2. give a concrete recommendation for a value for this setting (i.e. if Spark/Hadoop/S3A/?? has a setting to control the max size of data files, perhaps use that * 1.2, or any other formula we feel safe with)

itaiad200:
Thanks for the review, pulling once checks pass

PR #3208, opened 2022-04-13
API changes for invite user

guy-har:
Added The Changes in Swagger

@shimi9276 please note that I left `todos` where implementation is needed

PR #3223, opened 2022-04-17
Temporarily disable reset password capability from login form

talSofer:
Disable the reset password link until https://github.com/treeverse/lakeFS/pull/3208 is ready and we enable reset password only for users with email integration enabled.


guy-har (approved):
LGTM

PR #3228, opened 2022-04-17
Remove S3AUser system test

arielshaqed:
This was relevant for lakeFSFS, as part of testing whether S3A could
actually support it.  Property-based testing of lakeFSFS now does a much
better job of supporting it than it can ever do, so remove it.


johnnyaug (approved):

arielshaqed:
To clarify: This "system" "test" was a manually-run program / script that tried to verify the S3AFileSystem would vaguely work for writing lakeFSFS.  We have better tech now :-)

PR #3229, opened 2022-04-17
Support assuming a role (only on Hadoop 3)

arielshaqed:
# Reviewers please read!

I have tried to split this PR up into manageable commits.  You might review each commit separately, and it should make more sense:
* 027fb8ea6b1d7f3bacd63aa9a49ecb0b4cdb6648 is (just!) a facility for conditional compilation of code, allowing us to compile different files for Hadoop 2 and Hadoop 3 "flavours".
* 0a87e65f6f2503ce62e49789f326dda494b5e589 refactors `getS3Client` out of GarbageCollector.scala, once for each flavour.

Good luck!

### DIFF
--- a/.github/workflows/esti.yaml
+++ b/.github/workflows/esti.yaml
@@ -550,7 +550,9 @@ jobs:
 
       - name: Package Metaclient
         working-directory: clients/spark
-        run: sbt lakefs-spark-client-301/assembly
+        run: |
+          sbt lakefs-spark-client-301/assembly
+          sbt lakefs-spark-client-312-hadoop3/assembly
 
       - name: Prepare Metaclient location for export
         # upload-artifact cannot take a working-directory option (that only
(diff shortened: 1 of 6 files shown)


talSofer (approved):
WOW! for the clean solution and for mastering sbt!

WDYT of chaining the hadoop version into the jar name as we do with Spark and Scala versions?

arielshaqed:
> WOW! for the clean solution and for mastering sbt!

Not sure about "clean"; macro based solutions similar to `#ifdef` would be easier for me to read, but macros are still not entirely in Scala, and I was worried about having to work with differences between Scala 2.11 and 2.12.

Also wasn't sure Scala macros would actually do the job.

> WDYT of chaining the hadoop version into the jar name as we do with Spark and Scala versions?

Good idea!  Will do...

arielshaqed:
Thanks!  Pulling...

arielshaqed (commented):
Sorry, forgot to post this comment before pulling :-(

Issue #3137, opened 2022-03-29
Import: ingest bucket/path from UI


itaiad200:
For managed lakeFS, you can't use import from the lakeFS binary, so there should be an option from the UI.

itaiad200:
@ozkatz:

cc @johnnyaug for onboarding

Worked with @eden-ohana  on requirements and API for this:

# Client side multipart ingestion

## Required API operations

We're adding a few plumbing API operations to lakeFS:

### range creation API

range API can create (and later possibly get?) range files

`POST /api/v1/repositories/{repo}/ranges?fromSourceURI={storageURI}&after={keyToStartAfter}&prepend={prependToRangeKey}`

Returns 201 on success with the following body structure:

```yaml
# based on "graveler/committed/committed.proto"
id: string           # Hex of the range ID that got created
min_key: string      # First key in range
max_key: string      # last key in range
count: int           # number of records in range
estimated_size: int  # estimated size in bytes
```

Example:

```json
// POST /api/v1/repositories/my-repo/ranges?fromSourceURI=s3://my-bucket/production/collections/&after=production/collections/some/file.parquet&prepend=collections/
{
    "id": "a99241ed0732ab450f036110d89ab2d85ee1727a0fcf283f27341c5c3fe956d0",
    "min_key": "production/collections/some/file2.parquet",
    "max_key": "production/collections/some/file29348209.parquet",
    "count": 12983232,
    "estimated_size": 32940
}
```

### meta-range creation API

API to create meta-ranges from a set of non-overlapping ranges

`POST /api/v1/repositories/{repo}/metaranges`

This operation expects a POST-ed body in the schema:

```yaml
ranges: array[#range]  # a list of range entries as listed above
```

and returns a metarange with a range structure as listed above
Example:

```json
// POST /api/v1/repositories/my-repo/metaranges
{
    "ranges": [
        {
            "id": "a99241ed0732ab450f036110d89ab2d85ee1727a0fcf283f27341c5c3fe956d0",
            "min_key": "production/collections/some/file2.parquet",
            "max_key": "production/collections/some/file29348209.parquet",
            "count": 12983232,
            "estimated_size": 32940
        },
        {
            "id": "f0829e4626f7fa25d574189e5b75b208729707c56ebac493c8881bb29a1fd9d2",
            "min_key": "production/collections/some/file29348209.parquet",
            "max_key": "production/collections/some/file73489343.parquet",
            "count": 347834822,
            "estimated_size": 129023
        },
[...]

PR #3233, opened 2022-04-18
Expanding API to support import from UI

### LINKED ISSUE #3150: API to support new functionality of import from the UI
Part of https://github.com/treeverse/lakeFS/issues/3137

API needs to open new endpoints specified in #3137. 
It also requires to list objects from the object store and pass them to graveler.

itaiad200:
closes #3150 

Changes in this PR:
1. Adding 3 new functionalities to the API as described in #3137 
2. New permissions to support the new APIs (risks mapped nicely by @arielshaqed in #3150)
3. `store` package that was used for `ingest` was slightly refactored and moved to a shared place for reusing it for UI import.
4. Catalog had to be expanded. New WalkEntryIterator for managing iterators that list the object-store entries.
5. Graveler was slightly modified for the commitFromSourceMetaRange.

What not done yet:
~~1. Tests, tests, tests...
	- Unit-testing the new iterator
	- Esti test for the complete functionality
	- Unit-testing modified graveler/catalog
	- Unit-testing the new API from the controller~~

2. I'm unsure of how defensive we should be with the new API and would like to hear the reviewers opinion here. Assuming most functionality is plumbing by nature, what should we do with given ranges/metaranges? We can either do nothing, simply check for existence (roundtrip to the object-store for every range), or validate it completly (parse the range file and check for matching details).


clients/java/api/openapi.yaml
```diff
@@ -2202,6 +2202,15 @@ paths:
         schema:
           type: string
         style: simple
+      - description: Ths source metarange to commit. Branch must not have unstaged
```
N-o-Z, line 2205: What is the purpose of having this parameter in a GET operation? What do we expect to get here?

itaiad200: Closing this comment and recreating it in `swagger.yaml` file. This file is generated by Java openapi generator..

pkg/graveler/graveler.go
```diff
@@ -1489,15 +1499,19 @@ func (g *Graveler) Commit(ctx context.Context, repositoryID RepositoryID, branch
 			parentGeneration = commit.Generation
 		}
 		commit.Generation = parentGeneration + 1
-		changes, err := g.StagingManager.List(ctx, branch.StagingToken, ListingMaxBatchSize)
-		if err != nil {
-			return "", fmt.Errorf("staging list: %w", err)
-		}
-		defer changes.Close()
+		if params.SourceMetaRange != nil {
```
N-o-Z, line 1502: Not part of the changes but suggest to change the name of the variable 'commit' (line 1494) as it shadows an already defined variable in outer scope.

pkg/catalog/walk_entry_iterator.go
```diff
@@ -0,0 +1,131 @@
...
+
+	done   chan bool
+	closed *atomic.Bool
+	err    *atomic.Error
+
+	curr EntryWithMarker
+}
+
+// Mark stands for pagination information when listing objects from the blockstore.
+// It is used for server-client communication on the status of range ingestions.
```
N-o-Z, line 25:
```suggestion
// It is used for server-client communication on the status of range ingestion.
```

pkg/catalog/walk_entry_iterator.go
```diff
@@ -0,0 +1,131 @@
...
+type Mark store.Mark
+
+type EntryWithMarker struct {
+	EntryRecord
+	Mark
+}
+
+// used to determined the reason for the end of the walk
```
N-o-Z, line 33:
```suggestion
// used to determine the reason for the end of the walk
```

pkg/catalog/walk_entry_iterator.go
```diff
@@ -0,0 +1,131 @@
...
+	if it.err.Load() != nil || it.closed.Load() {
+		return false
+	}
+
+	var ok bool
+	it.curr, ok = <-it.entries
+	return ok
+}
+
+func (it walkEntryIterator) SeekGE(Path) {
```
N-o-Z, line 103: Wouldn't it be better to raise an unsupported error here?

N-o-Z (changes requested):
@itaiad200 I reviewed the parts that I could, there were a few comments / suggestion but I think unit tests for the added methods in graveler and cataloger are imperative as well as some sanity end to end test for ingest range and commit with metarange scenarios.
Also please note that some tests were broken due to the changes in controller.
Thank you!

cmd/lakectl/cmd/ingest.go
```diff
@@ -69,7 +69,14 @@ var ingestCmd = &cobra.Command{
...
 		go func() {
-			err := store.Walk(ctx, s3EndpointURL, from, func(e store.ObjectStoreEntry) error {
+			walker, err := store.GetWalker(ctx, store.WalkerOptions{
+				S3EndpointURL: s3EndpointURL,
+				StorageURI:    from,
+			})
+			if err != nil {
+				DieFmt("error creating object-store walker: %v", err)
+			}
+			err = walker.Walk(ctx, store.WalkOptions{}, func(e store.ObjectStoreEntry) error {
```
itai-david, line 79: Question - Now that we have a similar server side implementation - is this logic about to move to the server side as well, or is it here to stay?

pkg/graveler/graveler.go
itaiad200: Done.

pkg/catalog/walk_entry_iterator.go
itaiad200: Done.

pkg/catalog/walk_entry_iterator.go
itaiad200: Done.

pkg/catalog/walk_entry_iterator.go
itaiad200: I can't change the signature of the interface. Would you rather I raise a `Panic` here? Set the `it.err` to some custom error?

N-o-Z: That's a good question. using it.err can lead to a situation where if the user doesn't check Err explicitly, might accidentally assume the iterator reached the end. In this case I think I would go with panic

cmd/lakectl/cmd/ingest.go
itaiad200: Good question. They provide different capabilities. 
- `lakectl ingest` is more intuitive and stage the entries to the workspace, but doesn't scale to millions of objects.
- New server side import scales better, but has all or nothing approach to create a commit.

I think a possible followup is to consolidate the ingest/import/UI import features into 1/2, but not part of this PR.

nopcoder (changes requested):
Some feedback, still need to check the system tests

itaiad200 (commented):
Fixed some comments, still WIP

pkg/api/controller_test.go
```diff
@@ -1150,6 +1197,133 @@ func TestController_DeleteBranchHandler(t *testing.T) {
...
+		fromSourceURIWithPrefix = fromSourceURI + "/" + uriPrefix
+		after                   = "some/key/to/start/after"
+		continuationToken       = "opaque"
+		prepend                 = "some/logical/prefix"
+	)
+
+	ctx := context.Background()
+	token := continuationToken
+
+	setup := func(count int, err error) (api.ClientWithResponsesInterface, *testutils.FakeWalker) {
```
nopcoder, line 1213: Consider using as a helper function - passing t *testing.T as argument and cal Helper, it will report the line in the test as failed and not this one for all the tests.

nopcoder, line 1213: Can we use some other name than 'err' for the second argument?

pkg/api/controller_test.go
```diff
@@ -1150,6 +1197,133 @@ func TestController_DeleteBranchHandler(t *testing.T) {
...
+		w := testutils.NewFakeWalker(count, count, uriPrefix, after, continuationToken, fromSourceURIWithPrefix, err)
+		clt, deps := setupClientWithAdminAndWalkerFactory(t, testutils.FakeFactory{Walker: w})
+
+		// setup test data
+		_, err = deps.catalog.CreateRepository(ctx, "repo1", onBlock(deps, "foo1"), "main")
```
nopcoder, line 1218: Can keep `err` if we updated the name of the `err` argument. But it is a good practice to re-use it when it is input parameter.

pkg/api/controller_test.go
```diff
@@ -1150,6 +1197,133 @@ func TestController_DeleteBranchHandler(t *testing.T) {
...
+		require.Contains(t, string(resp.Body), expectedErr.Error())
+	})
+}
+
+func TestController_WriteMetaRangeHandler(t *testing.T) {
+	ctx := context.Background()
+	clt, deps := setupClientWithAdmin(t)
+
+	// setup test data
+	_, err := deps.catalog.CreateRepository(ctx, "repo1", onBlock(deps, "foo1"), "main")
```
nopcoder, line 1293: The repository name 'repo1' doesn't conflict with other tests here?

pkg/api/controller_test.go
```diff
@@ -1150,6 +1197,133 @@ func TestController_DeleteBranchHandler(t *testing.T) {
...
+	testutil.Must(t, err)
+
+	t.Run("successful metarange creation", func(t *testing.T) {
+		resp, err := clt.CreateMetaRangeWithResponse(ctx, "repo1", api.CreateMetaRangeJSONRequestBody{
+			Ranges: []api.RangeMetadata{
+				{Count: 11355, EstimatedSize: 123465897, Id: "FirstRangeID", MaxKey: "1", MinKey: "2"},
+				{Count: 13123, EstimatedSize: 123465897, Id: "SecondRangeID", MaxKey: "3", MinKey: "4"},
+				{Count: 10123, EstimatedSize: 123465897, Id: "SecondRangeID", MaxKey: "5", MinKey: "6"},
```
nopcoder, line 1301: Question - in real life, the fact that we will get two ranges with the same ID for metarange is almost not possible, right?

pkg/catalog/walk_entry_iterator.go
```diff
@@ -0,0 +1,150 @@
...
+// buffer size of the buffer between reading entries from the blockstore Walk and passing it on
+const bufferSize = 100
+
+// WalkerFactory provides an abstraction for creating Walker
+type WalkerFactory interface {
+	GetWalker(ctx context.Context, opts store.WalkerOptions) (*store.WalkerWrapper, error)
+}
+
+func NewWalkEntryIterator(ctx context.Context, factory WalkerFactory, fromSourceURI, prepend, after, continuationToken string) (*walkEntryIterator, error) {
+	if prepend != "" && prepend[len(prepend)-1:] != "/" {
```
nopcoder, line 45: For readability, I'll go with - `strings.HasPrefix(prepend, "/")`

pkg/catalog/walk_entry_iterator.go
```diff
@@ -0,0 +1,150 @@
...
+			}
+		}
+	}
+
+	return ok
+}
+
+func (it *walkEntryIterator) SeekGE(Path) {
+	// unsupported
+}
```
nopcoder, line 123: Think we will need a better way to handle this one - in other iterators, we set an error that this operation is not supported / panic.

pkg/catalog/walk_entry_iterator.go
```diff
@@ -0,0 +1,150 @@
...
+				},
+				Mark: Mark(it.walker.Marker()),
+			}
+			return nil
+		})
+		if !errors.Is(err, ErrItClosed) {
+			it.err.Store(err)
+		}
+		close(it.entries)
+		close(it.done)
```
nopcoder, line 91: prefer to call 'defer close' at the first thing in the `go func...`

pkg/catalog/walk_entry_iterator.go
```diff
@@ -0,0 +1,150 @@
...
+	go func() {
+		err := it.walker.Walk(ctx, store.WalkOptions{
+			After:             after,
+			ContinuationToken: continuationToken,
+		}, func(e store.ObjectStoreEntry) error {
+			if it.closed.Load() {
+				return ErrItClosed
+			}
+
+			it.entries <- EntryWithMarker{
```
nopcoder, line 71: We are blocking here until we read - if we use `select` we can break in case something closes, for example

pkg/catalog/walk_entry_iterator.go
```diff
@@ -0,0 +1,150 @@
...
+func (it *walkEntryIterator) Value() *EntryRecord {
+	rec := it.curr.EntryRecord
+	return &rec
+}
+
+func (it *walkEntryIterator) Err() error {
+	return it.err.Load()
+}
+
+func (it *walkEntryIterator) Close() {
```
nopcoder, line 134: I think with closing a channel and a wait group, we can close the iterator without a problem.

nopcoder (changes requested):

esti/import_test.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+const (
+	s3ImportPath    = "s3://esti-system-testing-data/import-test-data/"
+	gsImportPath    = "gs://esti-system-testing-data/import-test-data/"
+	azureImportPath = "https://esti-system-testing-data.blob.core.windows.net/import-test-data/"
+	prefixImport    = "imported/new-prefix/"
+)
+
+var filesToCheck = map[string]bool{
```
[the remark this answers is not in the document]
itaiad200: Because I was tired, I think..

pkg/api/controller_test.go
```diff
@@ -1150,6 +1197,133 @@ func TestController_DeleteBranchHandler(t *testing.T) {
...
+		fromSourceURI           = "https://valid.uri"
+		uriPrefix               = "take/from/here"
+		fromSourceURIWithPrefix = fromSourceURI + "/" + uriPrefix
+		after                   = "some/key/to/start/after"
+		continuationToken       = "opaque"
+		prepend                 = "some/logical/prefix"
+	)
+
+	ctx := context.Background()
+	token := continuationToken
```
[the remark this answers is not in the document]
itaiad200: Done.

pkg/api/controller_test.go
itaiad200: Done.

itaiad200: Done.

pkg/api/controller_test.go
itaiad200: Right, Done.

pkg/api/controller_test.go
itaiad200: Somehow it doesn't. Renaming just in case.

pkg/api/controller_test.go
itaiad200: Right, please see the second bullet in the PR description. We currently don't validate it.

pkg/api/controller.go
```diff
@@ -1704,7 +1705,81 @@ func (c *Controller) ResetBranch(w http.ResponseWriter, r *http.Request, body Re
 	writeResponse(w, http.StatusNoContent, nil)
 }
 
-func (c *Controller) Commit(w http.ResponseWriter, r *http.Request, body CommitJSONRequestBody, repository string, branch string) {
+func (c *Controller) IngestRange(w http.ResponseWriter, r *http.Request, body IngestRangeJSONRequestBody, repository string) {
+	if !c.authorize(w, r, permissions.Node{
```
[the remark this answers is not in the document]
itaiad200: I don't think so. I think that the write permission to the repo path is sufficient.

pkg/api/controller.go
```diff
@@ -1704,7 +1705,81 @@ func (c *Controller) ResetBranch(w http.ResponseWriter, r *http.Request, body Re
...
+					Action:   permissions.WriteObjectAction,
+					Resource: permissions.ObjectArn(repository, body.Prepend),
+				},
+			},
+		},
+	}) {
+		return
+	}
+
+	contToken := ""
```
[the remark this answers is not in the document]
itaiad200: Done.

esti/import_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+	metarangeResp, err := client.CreateMetaRangeWithResponse(ctx, repoName, api.CreateMetaRangeJSONRequestBody{
+		Ranges: ranges,
+	})
+
+	require.NoError(t, err, "failed to create metarange")
+	require.Equal(t, http.StatusCreated, metarangeResp.StatusCode())
+	require.NotNil(t, metarangeResp.JSON201.Id)
+
+	ingestionBranch := "ingestion"
```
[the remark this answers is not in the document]
itaiad200: Done.

esti/import_test.go
```diff
@@ -0,0 +1,112 @@
...
+	for hasMore {
+		resp, err := client.IngestRangeWithResponse(ctx, repoName, api.IngestRangeJSONRequestBody{
+			After:             after,
+			ContinuationToken: token,
+			FromSourceURI:     importPath,
+			Prepend:           prefixImport,
+		})
+		require.NoError(t, err, "failed to ingest range")
+		log.Info("Ingest range response body: %s", resp.Body)
+		require.Equal(t, http.StatusCreated, resp.StatusCode())
```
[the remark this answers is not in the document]
itaiad200: Done.

esti/import_test.go
```diff
@@ -0,0 +1,112 @@
...
+	var ranges []api.RangeMetadata
+	for hasMore {
+		resp, err := client.IngestRangeWithResponse(ctx, repoName, api.IngestRangeJSONRequestBody{
+			After:             after,
+			ContinuationToken: token,
+			FromSourceURI:     importPath,
+			Prepend:           prefixImport,
+		})
+		require.NoError(t, err, "failed to ingest range")
+		log.Info("Ingest range response body: %s", resp.Body)
```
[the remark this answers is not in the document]
itaiad200: Just debugging, will be removed once test passed.

pkg/ingest/store/factory.go
```diff
@@ -0,0 +1,112 @@
...
+			return nil, err
+		}
+		walker = &S3Walker{s3: svc, mark: Mark{HasMore: true}}
+	case "gs":
+		svc, err := GetGCSClient(ctx)
+		if err != nil {
+			return nil, err
+		}
+		walker = &GCSWalker{client: svc}
+	case "http", "https":
```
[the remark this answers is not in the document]
itaiad200: Right, but we don't have any other blockstore option here. When will add one that matched these schemas, will not to change it accordingly here and elsewhere.

pkg/ingest/store/factory.go
```diff
@@ -0,0 +1,112 @@
...
+
+var (
+	ErrNoStorageAdapter = errors.New("no storage adapter found")
+)
+
+type ObjectStoreEntry struct {
+	// FullKey represents the fully qualified path in the object store namespace for the given entry
+	FullKey string
+	// RelativeKey represents a path relative to prefix (or directory). If none specified, will be identical to FullKey
+	RelativeKey string
```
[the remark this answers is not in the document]
itaiad200: It is used by the marker as the last key and passed as-is to the client. The client should pass it to the server in the next call. It is important to keep the full key in the object store and not leave it to the client it to parse it from the combination of `RelativeKey` & `StorageURI` because of `/` complexities and the way different blockstores work.

pkg/ingest/store/azure.go
```diff
@@ -58,20 +71,31 @@ func getAzureBlobURL(containerURL *url.URL, blobName string) *url.URL {
...
+func (a *AzureBlobWalker) Walk(ctx context.Context, storageURI *url.URL, op WalkOptions, walkFn func(e ObjectStoreEntry) error) error {
 	// we use bucket as container and prefix as path
 	containerURL, prefix, err := extractAzurePrefix(storageURI)
 	if err != nil {
 		return err
 	}
 	container := azblob.NewContainerURL(*containerURL, a.client)
-	for marker := (azblob.Marker{}); marker.NotDone(); {
-		listBlob, err := container.ListBlobsFlatSegment(ctx, marker, azblob.ListBlobsSegmentOptions{Prefix: prefix})
+	for marker := (azblob.Marker{Val: &op.ContinuationToken}); marker.NotDone(); {
```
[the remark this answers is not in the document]
itaiad200: I have no idea why your suggestion doesn't work when it should 🤷

package-lock.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "lakeFS",
```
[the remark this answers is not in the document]
itaiad200: Done.

pkg/graveler/graveler.go
```diff
@@ -493,6 +488,11 @@ type Plumbing interface {
 	GetMetaRange(ctx context.Context, repositoryID RepositoryID, metaRangeID MetaRangeID) (MetaRangeAddress, error)
 	// GetRange returns information where rangeID is stored.
 	GetRange(ctx context.Context, repositoryID RepositoryID, rangeID RangeID) (RangeAddress, error)
+	// WriteRange creates a new Range from the iterator values.
+	// Keeps Range closing logic, so might not flush all values to the range.
```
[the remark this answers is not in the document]
itaiad200: Yes, rephrased.

itaiad200:
Decision: Don't be defensive with the given range and metarange inputs. These are plumbing commands and its up to the user to use them carefully. We can add validation later on if necessary.

pkg/api/controller_test.go
```diff
@@ -507,21 +510,65 @@ func TestController_CommitHandler(t *testing.T) {
...
+		})
+		verifyResponseOK(t, resp, err)
+
+		testutil.MustDo(t, "commit bar on foo4", deps.catalog.CreateEntry(ctx, repo, "foo-branch", catalog.DBEntry{Path: "foo/bar/2", PhysicalAddress: "pa", CreationDate: time.Now(), Size: 666, Checksum: "cs", Metadata: nil}))
+		resp, err = clt.CommitWithResponse(ctx, repo, "foo-branch", &api.CommitParams{SourceMetarange: &resp.JSON201.MetaRangeId}, api.CommitJSONRequestBody{
+			Message: "some message",
+		})
+
+		require.NoError(t, err)
+		require.Equal(t, http.StatusBadRequest, resp.StatusCode())
```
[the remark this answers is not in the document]
nopcoder: JSON400 != nil will verify the status code + content type application/json

pkg/catalog/walk_entry_iterator.go
```diff
@@ -0,0 +1,156 @@
...
+
+var (
+	// ErrItClosed is used to determine the reason for the end of the walk
+	ErrItClosed = errors.New("iterator closed")
+
+	errSeekGENotSupported = errors.New("SeekGE not supported for walk entry iterator")
+)
+
+// buffer size of the buffer between reading entries from the blockstore Walk and passing it on
+const bufferSize = 100
```
nopcoder, line 42: code doc starts with the name
```suggestion
// bufferSize - buffer size of the buffer between reading entries from the blockstore Walk and passing it on
const bufferSize = 100
```

pkg/catalog/walk_entry_iterator.go
```diff
@@ -0,0 +1,156 @@
...
+
+// WalkerFactory provides an abstraction for creating Walker
+type WalkerFactory interface {
+	GetWalker(ctx context.Context, opts store.WalkerOptions) (*store.WalkerWrapper, error)
+}
+
+func NewWalkEntryIterator(ctx context.Context, factory WalkerFactory, fromSourceURI, prepend, after, continuationToken string) (*walkEntryIterator, error) {
```
nopcoder, line 49: Question - if the iterator creates just one walker, why does it need a factory? We can just pass it a walker

pkg/catalog/walk_entry_iterator.go
```diff
@@ -0,0 +1,156 @@
...
+					Entry: &Entry{
+						Address:      e.Address,
+						LastModified: timestamppb.New(e.Mtime),
+						Size:         e.Size,
+						ETag:         e.ETag,
+						AddressType:  Entry_FULL,
+						ContentType:  e.Address,
+					},
+				},
+				Mark: Mark(it.walker.Marker()),
```
nopcoder, line 90: Marker() returns Mark, redundant?

pkg/catalog/walk_entry_iterator.go
```diff
@@ -0,0 +1,156 @@
...
+			}
+			return nil
+		})
+		if !errors.Is(err, ErrItClosed) {
+			it.err.Store(err)
+		}
+		close(it.entries)
```
nopcoder, line 97: also defer close?

pkg/graveler/graveler.go
```diff
@@ -329,6 +329,8 @@ type CommitParams struct {
 	// Date (Unix Epoch in seconds) is used to override commits creation date
 	Date     *int64
 	Metadata Metadata
+	// If SourceMetaRange exists, use it directly. Fail if branch has uncommitted changes
+	SourceMetaRange *MetaRangeID
```
nopcoder, line 333:
```suggestion
	// SourceMetaRange - If SourceMetaRange exists, use it directly. Fail if branch has uncommitted changes
	SourceMetaRange *MetaRangeID
```

nopcoder (approved):
Looks good - minor comments added.
Except that my main concern how the iterator implementation over the current walk API. My concern is the walk itself, which affect how the iterator is written - so I don't think it should be addressed in this PR.

pkg/api/controller_test.go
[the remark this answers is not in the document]
itaiad200: Done.

pkg/catalog/walk_entry_iterator.go
itaiad200: Done.

pkg/catalog/walk_entry_iterator.go
itaiad200: Right, done.

pkg/catalog/walk_entry_iterator.go
itaiad200: Not the same Mark type :)

pkg/catalog/walk_entry_iterator.go
itaiad200: Done.

pkg/graveler/graveler.go
itaiad200: Done.

PR #3251, opened 2022-04-23
kv implementation using postgres and memory

### LINKED ISSUE #3239: KV interface and initial Postgres implementation

nopcoder:
Initial implementation of kv for lakefs proposal

Close #3239

### DIFF
--- a/pkg/kv/kvtest/store.go
+++ b/pkg/kv/kvtest/store.go
@@ -1,0 +1,253 @@
+package kvtest
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"testing"
+
+	"github.com/go-test/deep"
+	"github.com/treeverse/lakefs/pkg/kv"
+)
+
+type makeStore func(t *testing.T, ctx context.Context) kv.Store
+
+func TestDriver(t *testing.T, name string, dsn string) {
+	ctx := context.Background()
+	ms := makeStoreByName(name, dsn)
+	t.Run("Driver_Open", func(t *testing.T) { testDriverOpen(t, ctx, ms) })
+	t.Run("Store_SetGet", func(t *testing.T) { testStoreSetGet(t, ctx, ms) })
+	t.Run("Store_SetIf", func(t *testing.T) { testStoreSetIf(t, ctx, ms) })
+	t.Run("Store_Delete", func(t *testing.T) { testStoreDelete(t, ctx, ms) })
+	t.Run("Store_Scan", func(t *testing.T) { testStoreScan(t, ctx, ms) })
+}
+
+func testDriverOpen(t *testing.T, ctx context.Context, ms makeStore) {
+	store1 := ms(t, ctx)
+	store1.Close()
+
+	store2 := ms(t, ctx)
+	store2.Close()
+}
+
+func testStoreSetGet(t *testing.T, ctx context.Context, ms makeStore) {
+	store := ms(t, ctx)
+	defer store.Close()
+
+	testKey := []byte("key")
+	testValue1 := []byte("value1")
+	testValue2 := []byte("value2")
+
+	// set test key with value1
+	err := store.Set(ctx, testKey, testValue1)
+	if err != nil {
+		t.Fatalf("failed to set key '%s', to value '%s': %s", testKey, testValue1, err)
+	}
+
+	// get test key with value1
+	val, err := store.Get(ctx, testKey)
+	if err != nil {
+		t.Fatalf("failed to get key '%s': %s", testKey, err)
+	}
+	if !bytes.Equal(testValue1, val) {
+		t.Fatalf("key='%s' value='%s' doesn't match, expected='%s'", testKey, val, testValue1)
+	}
+
+	// override key with value2
+	err = store.Set(ctx, testKey, testValue2)
+	if err != nil {
+		t.Fatalf("failed to set key '%s', to value '%s': %s", testKey, testValue2, err)
+	}
+
+	// get test key with value2
+	val2, err := store.Get(ctx, testKey)
+	if err != nil {
+		t.Fatalf("failed to get key '%s': %s", testKey, err)
+	}
+	if !bytes.Equal(testValue2, val2) {
+		t.Fatalf("key='%s' value='%s' doesn't match, expected='%s'", testKey, val2, testValue2)
+	}
+
+	// get a missing key
+	const keyNotExists = "key-not-exists"
+	val3, err := store.Get(ctx, []byte(keyNotExists))
+	if !errors.Is(err, kv.ErrNotFound) {
+		t.Fatalf("get key='%s' err=%s, expected not found", keyNotExists, err)
+	}
+	if val3 != nil {
+		t.Fatalf("get key='%s' value='%s', expected nil", keyNotExists, val3)
+	}
+}
+
+func testStoreDelete(t *testing.T, ctx context.Context, ms makeStore) {
+	store := ms(t, ctx)
+	defer store.Close()
+
+	t.Run("exists", func(t *testing.T) {
+		keyToDel := []byte("key-to-delete")
+		err := store.Set(ctx, keyToDel, []byte("value to delete"))
+		if err != nil {
+			t.Fatalf("failed to set key='%s': %s", keyToDel, err)
+		}
+		err = store.Delete(ctx, keyToDel)
+		if err != nil {
+			t.Fatalf("failed to delete key='%s': %s", keyToDel, err)
+		}
+	})
+
+	t.Run("non_exists", func(t *testing.T) {
+		keyToDel := []byte("missing-key-to-delete")
+		err := store.Delete(ctx, keyToDel)
+		if !errors.Is(err, kv.ErrNotFound) {
+			t.Fatalf("delete missing key '%s', err=%v expected=%s", keyToDel, err, kv.ErrNotFound)
+		}
+	})
+}
+
+func testStoreSetIf(t *testing.T, ctx context.Context, ms makeStore) {
+	store := ms(t, ctx)
+	defer store.Close()
+
+	t.Run("previous_value", func(t *testing.T) {
+		// setup initial value
+		key := []byte("setIfKey1")
+		value := []byte("val1")
+		err := store.Set(ctx, key, value)
+		if err != nil {
+			t.Fatalf("failed to set key=%s, value=%s: %s", key, value, err)
+		}
+
+		// should set new value if old value is kept
+		newValue := []byte("val2")
+		err = store.SetIf(ctx, key, newValue, value)
+		if err != nil {
+			t.Fatalf("failed to setIf key=%s value=%s pred=%s: %s", key, newValue, value, err)
+		}
+
+		// should NOT set new value if old value is not kept
+		newValue = []byte("val3")
+		err = store.SetIf(ctx, key, newValue, value)
+		if !errors.Is(err, kv.ErrNotFound) {
+			t.Fatalf("setIf err=%s, key=%s, value=%s, pred=%s, expected err=%s", err, key, newValue, value, kv.ErrNotFound)
+		}
+
+		// should NOT set new value if try conditional nil
+		newValue = []byte("val4")
+		err = store.SetIf(ctx, key, newValue, nil)
+		if !errors.Is(err, kv.ErrNotFound) {
+			t.Fatalf("setIf err=%s, key=%s, value=%s, pred=%s, expected err=%s", err, key, newValue, value, kv.ErrNotFound)
+		}
+	})
+
+	t.Run("no_previous_key", func(t *testing.T) {
+		// setIf without previous value
+		key := []byte("setIfKey2")
+		value := []byte("val1")
+		err := store.SetIf(ctx, key, value, nil)
+		if err != nil {
+			t.Fatalf("setIf without previous key - key=%s, value=%s: %s", key, value, err)
+		}
+
+		// should fail if we try to do the same again - as the value updated to 'val1'
+		err = store.SetIf(ctx, key, value, nil)
+		if !errors.Is(err, kv.ErrNotFound) {
+			t.Fatalf("setIf with nil while key/value exists - err=%v, key=%s value=%s, expected err=%s",
+				err, key, value, kv.ErrNotFound)
+		}
+	})
+}
+
+func testStoreScan(t *testing.T, ctx context.Context, ms makeStore) {
+	store := ms(t, ctx)
+	defer store.Close()
+
+	// setup sample data
+	const samplePrefix = "scan"
+	const sampleItems = 10
+	sampleData := setupSampleData(t, ctx, store, samplePrefix, sampleItems)
+
+	t.Run("full", func(t *testing.T) {
+		scanKeyPrefix := []byte("scan-key")
+		scan, err := store.Scan(ctx, scanKeyPrefix)
+		if err != nil {
+			t.Fatal("failed to scan", err)
+		}
+		defer scan.Close()
+
+		var entries []kv.Entry
+		for scan.Next() {
+			ent := scan.Entry()
+			if ent == nil {
+				t.Fatal("scan got nil entry")
+			}
+			if !bytes.HasPrefix(ent.Key, scanKeyPrefix) {
+				break
+			}
+			entries = append(entries, *ent)
+		}
+		if err := scan.Err(); err != nil {
+			t.Fatal("scan ended with an error", err)
+		}
+		if diff := deep.Equal(entries, sampleData); diff != nil {
+			t.Fatal("scan data didn't match:", diff)
+		}
+	})
+
+	t.Run("part", func(t *testing.T) {
+		scan, err := store.Scan(ctx, []byte("scan-key-5"))
+		if err != nil {
+			t.Fatal("failed to scan", err)
+		}
+		defer scan.Close()
+
+		var entries []kv.Entry
+		for scan.Next() {
+			ent := scan.Entry()
+			if ent == nil {
+				t.Fatal("scan got nil entry")
+			}
+			if !bytes.HasPrefix(ent.Key, []byte("scan-key-")) {
+				break
+			}
+			entries = append(entries, *ent)
+		}
+		if err := scan.Err(); err != nil {
+			t.Fatal("scan ended with an error", err)
+		}
+		if diff := deep.Equal(entries, sampleData[5:]); diff != nil {
+			t.Fatal("scan data didn't match:", diff)
+		}
+	})
+}
+
+func makeStoreByName(name string, dsn string) makeStore {
+	return func(t *testing.T, ctx context.Context) kv.Store {
+		t.Helper()
+		store, err := kv.Open(ctx, name, dsn)
+		if err != nil {
+			t.Fatalf("failed to open kv '%s' (%s) store: %s", name, dsn, err)
+		}
+		return store
+	}
+}
+
+func setupSampleData(t *testing.T, ctx context.Context, store kv.Store, prefix string, items int) []kv.Entry {
+	t.Helper()
+	entries := make([]kv.Entry, 0, items)
+	for i := 0; i < items; i++ {
+		entry := sampleEntry(prefix, i)
+		err := store.Set(ctx, entry.Key, entry.Value)
+		if err != nil {
+			t.Fatalf("failed to setup data with '%s': %s", entry, err)
+		}
+		entries = append(entries, entry)
+	}
+	return entries
+}
+
+func sampleEntry(prefix string, n int) kv.Entry {
+	k := fmt.Sprintf("%s-key-%d", prefix, n)
+	v := fmt.Sprintf("%s-value-%d", prefix, n)
+	return kv.Entry{Key: []byte(k), Value: []byte(v)}
+}
--- a/pkg/kv/mem/store.go
+++ b/pkg/kv/mem/store.go
@@ -1,0 +1,159 @@
+package mem
+
+import (
+	"bytes"
+	"context"
+	"sort"
+	"sync"
+
+	"github.com/treeverse/lakefs/pkg/kv"
+)
+
+type Driver struct{}
+
+type Store struct {
+	m    map[string][]byte
+	keys []string
+	mu   sync.RWMutex
+}
+
+type Entries struct {
+	entry *kv.Entry
+	err   error
+	start []byte
+	first bool
+	store *Store
+}
+
+const DriverName = "mem"
+
+//nolint:gochecknoinits
+func init() {
+	kv.Register(DriverName, &Driver{})
+}
+
+func (d *Driver) Open(_ context.Context, _ string) (kv.Store, error) {
+	return &Store{
+		m:    make(map[string][]byte),
+		keys: []string{},
+	}, nil
+}
+
+func (s *Store) Get(_ context.Context, key []byte) ([]byte, error) {
+	if key == nil {
+		return nil, kv.ErrMissingKey
+	}
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	value, ok := s.m[string(key)]
+	if !ok {
+		return nil, kv.ErrNotFound
+	}
+	return value, nil
+}
+
+func (s *Store) Set(_ context.Context, key, value []byte) error {
+	if key == nil {
+		return kv.ErrMissingKey
+	}
+	if value == nil {
+		return kv.ErrMissingValue
+	}
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if _, found := s.m[string(key)]; !found {
+		s.keys = append(s.keys, string(key))
+		sort.Strings(s.keys)
+	}
+	s.m[string(key)] = value
+	return nil
+}
+
+func (s *Store) SetIf(_ context.Context, key, value, valuePredicate []byte) error {
+	if key == nil {
+		return kv.ErrMissingKey
+	}
+	if value == nil {
+		return kv.ErrMissingValue
+	}
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	curr := s.m[string(key)]
+	if valuePredicate == nil {
+		if curr != nil {
+			return kv.ErrNotFound
+		}
+		s.keys = append(s.keys, string(key))
+		sort.Strings(s.keys)
+	} else if !bytes.Equal(valuePredicate, curr) {
+		return kv.ErrNotFound
+	}
+	s.m[string(key)] = value
+	return nil
+}
+
+func (s *Store) Delete(_ context.Context, key []byte) error {
+	if key == nil {
+		return kv.ErrMissingKey
+	}
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	if _, found := s.m[string(key)]; !found {
+		return kv.ErrNotFound
+	}
+	idx := sort.SearchStrings(s.keys, string(key))
+	if idx < len(s.keys) && s.keys[idx] == string(key) {
+		s.keys = append(s.keys[:idx], s.keys[idx+1:]...)
+	}
+	delete(s.m, string(key))
+	return nil
+}
+
+func (s *Store) Scan(ctx context.Context, start []byte) (kv.Entries, error) {
+	return &Entries{
+		store: s,
+		first: true,
+		start: start,
+	}, nil
+}
+
+func (s *Store) Close() {}
+
+func (e *Entries) Next() bool {
+	if e.start == nil {
+		return false
+	}
+	e.store.mu.RLock()
+	defer e.store.mu.RUnlock()
+	idx := sort.SearchStrings(e.store.keys, string(e.start))
+	if idx == len(e.store.keys) {
+		e.start = nil
+		return false
+	}
+	// point to the entry we found
+	key := e.store.keys[idx]
+	value := e.store.m[key]
+	e.entry = &kv.Entry{
+		Key:   []byte(key),
+		Value: value,
+	}
+	// set start to the next item - nil as end indicator
+	if idx+1 < len(e.store.keys) {
+		e.start = []byte(e.store.keys[idx+1])
+	} else {
+		e.start = nil
+	}
+	return true
+}
+
+func (e *Entries) Entry() *kv.Entry {
+	return e.entry
+}
+
+func (e *Entries) Err() error {
+	return e.err
+}
+
+func (e *Entries) Close() {
+	e.err = kv.ErrClosedEntries
+}
--- a/pkg/kv/postgres/store.go
+++ b/pkg/kv/postgres/store.go
@@ -1,0 +1,234 @@
+package postgres
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/jackc/pgconn"
+	"github.com/jackc/pgx/v4"
+	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/treeverse/lakefs/pkg/kv"
+)
+
+type Driver struct{}
+
+type Store struct {
+	Pool   *pgxpool.Pool
+	Params *Params
+}
+
+type Entries struct {
+	rows  pgx.Rows
+	entry *kv.Entry
+	err   error
+}
+
+const (
+	DriverName = "postgres"
+
+	defaultTableName = "kv"
+	paramTableName   = "lakefskv_table"
+)
+
+//nolint:gochecknoinits
+func init() {
+	kv.Register(DriverName, &Driver{})
+}
+
+func (d *Driver) Open(ctx context.Context, name string) (kv.Store, error) {
+	// TODO(barak): name format - we need to accept a connection string to the db + params related to db + params related to kv
+	// TODO(barak): handle reuse - open the same kv - reuse the pool
+	config, err := pgxpool.ParseConfig(name)
+	if err != nil {
+		return nil, err
+	}
+	pool, err := pgxpool.ConnectConfig(ctx, config)
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		// if we return before store uses the pool, free it
+		if pool != nil {
+			pool.Close()
+		}
+	}()
+
+	// acquire connection and make sure we reach the database
+	conn, err := pool.Acquire(ctx)
+	if err != nil {
+		return nil, err
+	}
+	defer conn.Release()
+	err = conn.Conn().Ping(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	params := parseStoreConfig(config.ConnConfig.RuntimeParams)
+
+	// migrate database
+	err = migrateDatabase(ctx, conn, params)
+	if err != nil {
+		return nil, err
+	}
+	store := &Store{
+		Pool:   pool,
+		Params: params,
+	}
+	pool = nil
+	return store, nil
+}
+
+type Params struct {
+	TableName string
+}
+
+func parseStoreConfig(runtimeParams map[string]string) *Params {
+	p := &Params{
+		TableName: defaultTableName,
+	}
+	if tableName, ok := runtimeParams[paramTableName]; ok {
+		p.TableName = tableName
+	}
+	return p
+}
+
+func migrateDatabase(ctx context.Context, conn *pgxpool.Conn, params *Params) error {
+	_, err := conn.Exec(ctx, `CREATE TABLE IF NOT EXISTS `+params.TableName+` (
+    key BYTEA NOT NULL PRIMARY KEY,
+    value BYTEA NOT NULL);`)
+	return err
+}
+
+func (s *Store) Get(ctx context.Context, key []byte) ([]byte, error) {
+	if key == nil {
+		return nil, kv.ErrMissingKey
+	}
+	row := s.Pool.QueryRow(ctx, `SELECT value FROM `+s.Params.TableName+` WHERE key = $1`, key)
+	var val []byte
+	err := row.Scan(&val)
+	if errors.Is(err, pgx.ErrNoRows) {
+		return nil, kv.ErrNotFound
+	}
+	if err != nil {
+		return nil, fmt.Errorf("%w: %s", kv.ErrOperationFailed, err)
+	}
+	return val, nil
+}
+
+func (s *Store) Set(ctx context.Context, key, value []byte) error {
+	if key == nil {
+		return kv.ErrMissingKey
+	}
+	if value == nil {
+		return kv.ErrMissingValue
+	}
+	_, err := s.Pool.Exec(ctx, `INSERT INTO `+s.Params.TableName+`(key,value) VALUES($1,$2)
+			ON CONFLICT (key) DO UPDATE SET value = $2`, key, value)
+	if err != nil {
+		return fmt.Errorf("%w: %s", kv.ErrOperationFailed, err)
+	}
+	return nil
+}
+
+func (s *Store) SetIf(ctx context.Context, key, value, valuePredicate []byte) error {
+	if key == nil {
+		return kv.ErrMissingKey
+	}
+	if value == nil {
+		return kv.ErrMissingValue
+	}
+	var (
+		res pgconn.CommandTag
+		err error
+	)
+	if valuePredicate == nil {
+		// use insert to make sure there was no previous value before
+		res, err = s.Pool.Exec(ctx, `INSERT INTO `+s.Params.TableName+`(key,value) VALUES($1,$2) ON CONFLICT DO NOTHING`, key, value)
+	} else {
+		// update just in case the previous value was same as predicate value
+		res, err = s.Pool.Exec(ctx, `UPDATE `+s.Params.TableName+` SET value=$2 WHERE key=$1 AND value=$3`, key, value, valuePredicate)
+	}
+	if err != nil {
+		return fmt.Errorf("%w: %s", kv.ErrOperationFailed, err)
+	}
+	if res.RowsAffected() != 1 {
+		return kv.ErrNotFound
+	}
+	return nil
+}
+
+func (s *Store) Delete(ctx context.Context, key []byte) error {
+	if key == nil {
+		return kv.ErrMissingKey
+	}
+	res, err := s.Pool.Exec(ctx, `DELETE FROM `+s.Params.TableName+` WHERE key=$1`, key)
+	if err != nil {
+		return fmt.Errorf("%w: %s", kv.ErrOperationFailed, err)
+	}
+	if res.RowsAffected() != 1 {
+		return kv.ErrNotFound
+	}
+	return nil
+}
+
+func (s *Store) Scan(ctx context.Context, start []byte) (kv.Entries, error) {
+	var (
+		rows pgx.Rows
+		err  error
+	)
+	if start == nil {
+		rows, err = s.Pool.Query(ctx, `SELECT key,value FROM `+s.Params.TableName+` ORDER BY key`)
+	} else {
+		rows, err = s.Pool.Query(ctx, `SELECT key,value FROM `+s.Params.TableName+` WHERE key >= $1 ORDER BY key`, start)
+	}
+	if err != nil {
+		return nil, fmt.Errorf("%w: %s", kv.ErrOperationFailed, err)
+	}
+	return &Entries{
+		rows: rows,
+	}, nil
+}
+
+func (s *Store) Close() {
+	s.Pool.Close()
+}
+
+// Next read the next key/value on any error entry will be nil, err will be set by trying to scan the results
+func (e *Entries) Next() bool {
+	e.entry = nil
+	e.err = nil
+	if !e.rows.Next() {
+		return false
+	}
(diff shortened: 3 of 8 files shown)


pkg/kv/store.go
```diff
@@ -0,0 +1,97 @@
...
+	ErrNotFound        = errors.New("not found")
+	ErrClosedEntries   = errors.New("closed entries")
+)
+
+type Entry struct {
+	Key   []byte
+	Value []byte
+}
+
+type Driver interface {
```
itaiad200, line 27: Can you document what a `Driver` is?

pkg/kv/store.go
```diff
@@ -0,0 +1,97 @@
...
+	// Open access to the database store. migrate kv if needed
+	Open(ctx context.Context, dsn string) (Store, error)
```
itaiad200, line 29: What is `dsn`?

pkg/kv/store.go
```diff
@@ -0,0 +1,97 @@
...
+
+	// Destroy will delete kv from storage
+	// Destroy(ctx context.Context, name string) error
```
itaiad200, line 32: Let's remove if unneeded.

pkg/kv/store.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+type Store interface {
```
itaiad200, line 35: This comment should probably arose sooner, but better late than never :)
How are planning to implement `Drop` or `DropByPrefix` we have for staging these days?

pkg/kv/store.go
```diff
@@ -0,0 +1,97 @@
...
+	ErrUnknownDriver   = errors.New("unknown driver")
+	ErrOperationFailed = errors.New("operation failed")
+	ErrMissingKey      = errors.New("missing key")
+	ErrMissingValue    = errors.New("missing value")
+	ErrNotFound        = errors.New("not found")
+	ErrClosedEntries   = errors.New("closed entries")
+)
+
+type Entry struct {
+	Key   []byte
```
itaiad200, line 23: Let's have a concrete `Key` & `Value` types to avoid mixup by the caller or the implementor.

pkg/kv/store.go
```diff
@@ -0,0 +1,97 @@
...
+type Store interface {
+	// Get returns a value for the given key, or ErrNotFound if key doesn't exist
+	Get(ctx context.Context, key []byte) ([]byte, error)
+
+	// Set stores the given value, overwriting an existing value if one exists
+	Set(ctx context.Context, key, value []byte) error
+
+	// SetIf returns an ErrPredicateFailed error if the valuePredicate passed
+	//  doesn't match the currently stored value. SetIf is a simple compare-and-swap operator:
+	//  valuePredicate is either the existing value, or an opaque value representing it (hash, index, etc).
```
itaiad200, line 44: So how would the caller know if `valuePredicate` should be compared to the stored `value` or `hash`, `index`?

pkg/kv/store.go
itaiad200: And `ValuePredicate`?

pkg/kv/store.go
```diff
@@ -0,0 +1,97 @@
...
+	//  this is intentionally simplistic: we can model a better abstraction on top, keeping this interface simple for implementors
+	SetIf(ctx context.Context, key, value, valuePredicate []byte) error
+
+	// Delete will delete the key, if any
```
itaiad200, line 48: Any error returned if the key is missing? Let's document the expected behaviour.

pkg/kv/store.go
```diff
@@ -0,0 +1,97 @@
...
+	Delete(ctx context.Context, key []byte) error
+
+	// Scan returns entries that can be read by key order, starting at or after the `start` position
+	Scan(ctx context.Context, start []byte) (Entries, error)
+
+	// Close access to the database store
+	Close()
+}
+
+type Entries interface {
```
itaiad200, line 58: 😢  ? 
```suggestion
type EntriesIterator interface {
```

pkg/kv/store.go
```diff
@@ -0,0 +1,97 @@
...
+	Err() error
+	Close()
+}
+
+var (
+	drivers   = make(map[string]Driver)
+	driversMu sync.RWMutex
+)
+
+func Register(name string, driver Driver) {
```
itaiad200, line 70: Register is called during runtime? Isn't the list predefined? Why would a lakeFS start with more than one instance registered?

pkg/kv/postgres/store.go
```diff
@@ -0,0 +1,239 @@
...
+
+//nolint:gochecknoinits
+func init() {
+	kv.Register(DriverName, &Driver{})
+}
+
+func (d *Driver) Open(ctx context.Context, name string) (kv.Store, error) {
+	// TODO(barak): name format - we need to accept a connection string to the db + params related to db + params related to kv
+	// TODO(barak): handle reuse - open the same kv - reuse the pool
+	config, err := pgxpool.ParseConfig(name)
```
itaiad200, line 42: Can it get a config struct like we do elsewhere and work it way from there?

pkg/kv/postgres/store.go
```diff
@@ -0,0 +1,239 @@
...
+	p := &Params{
+		TableName: defaultTableName,
+	}
+	if tableName, ok := runtimeParams[paramTableName]; ok {
+		p.TableName = tableName
+	}
+	return p
+}
+
+func migrateDatabase(ctx context.Context, conn *pgxpool.Conn, params *Params) error {
```
itaiad200, line 97: Why here and not a DDL package? How would the next migration code look like? How would this migration coexist with pkgs (auth/graveler/etc..) migrations?

pkg/kv/postgres/store.go
```diff
@@ -0,0 +1,239 @@
...
+	_, err := conn.Exec(ctx, `CREATE TABLE IF NOT EXISTS `+params.TableName+` (
+    key BYTEA NOT NULL PRIMARY KEY,
+    value BYTEA NOT NULL);`)
```
itaiad200, line 100: Shouldn't be ordered by some way to allow quick scans?

pkg/kv/postgres/store.go
```diff
@@ -0,0 +1,239 @@
...
+	if key == nil {
+		return kv.ErrMissingKey
+	}
+	if value == nil {
+		return kv.ErrMissingValue
+	}
+	_, err := s.Pool.Exec(ctx, `INSERT INTO `+s.Params.TableName+`(key,value) VALUES($1,$2)
+			ON CONFLICT (key) DO UPDATE SET value = $2`, key, value)
+	if err != nil {
+		return fmt.Errorf("%w: %s", kv.ErrOperationFailed, err)
```
itaiad200, line 130:
```suggestion
		return fmt.Errorf("%s: %w", err, kv.ErrOperationFailed)
```

pkg/kv/postgres/store.go
```diff
@@ -0,0 +1,239 @@
...
+		res, err = s.Pool.Exec(ctx, `INSERT INTO `+s.Params.TableName+`(key,value) VALUES($1,$2) ON CONFLICT DO NOTHING`, key, value)
+	} else {
+		// update just in case the previous value was same as predicate value
+		res, err = s.Pool.Exec(ctx, `UPDATE `+s.Params.TableName+` SET value=$2 WHERE key=$1 AND value=$3`, key, value, valuePredicate)
+	}
+	if err != nil {
+		return fmt.Errorf("%w: %s", kv.ErrOperationFailed, err)
+	}
+	if res.RowsAffected() != 1 {
+		return kv.ErrNotFound
```
itaiad200, line 157: I don't think that's the proper error. `kv.ErrPredicateFailed` is more appropriate in some cases.

pkg/kv/postgres/store.go
```diff
@@ -0,0 +1,239 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("%w: %s", kv.ErrOperationFailed, err)
+	}
+	return &Entries{
+		rows: rows,
+	}, nil
+}
+
+func (s *Store) Close() {
+	if s.closed {
```
itaiad200, line 195: This value anywhere except here, wouldn't it result with errors if you call other methods when the `Store` is closed?

pkg/kv/postgres/store.go
```diff
@@ -0,0 +1,239 @@
...
+		return
+	}
+	s.Pool.Close()
+	s.closed = true
+}
+
+// Next read the next key/value on any error entry will be nil, err will be set by trying to scan the results
+func (e *Entries) Next() bool {
+	e.entry = nil
+	e.err = nil
```
itaiad200, line 205: if `err` is set I don't think we should override it.

pkg/kv/postgres/store.go
```diff
@@ -0,0 +1,239 @@
...
+func (e *Entries) Entry() *kv.Entry {
+	return e.entry
+}
+
+// Err return the last scan error or the cursor error
+func (e *Entries) Err() error {
+	if e.err != nil {
+		return e.err
+	}
+	if err := e.rows.Err(); err != nil {
```
itaiad200, line 227: Why show initiative and check for this error? It will come up in the next read anyway..

pkg/kv/mem/store.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	curr := s.m[string(key)]
+	if valuePredicate == nil {
+		if curr != nil {
+			return kv.ErrNotFound
+		}
+		s.keys = append(s.keys, string(key))
+		sort.Strings(s.keys)
```
itaiad200, line 87: Efficient 😉

pkg/kv/kvtest/store.go
```diff
@@ -0,0 +1,253 @@
...
+
+	testKey := []byte("key")
+	testValue1 := []byte("value1")
+	testValue2 := []byte("value2")
+
+	// set test key with value1
+	err := store.Set(ctx, testKey, testValue1)
+	if err != nil {
+		t.Fatalf("failed to set key '%s', to value '%s': %s", testKey, testValue1, err)
+	}
```
itaiad200, line 47: I like `require.NoError(...)` better as a one liner 0 indentation check..

pkg/kv/kvtest/store.go
```diff
@@ -0,0 +1,253 @@
+package kvtest
```
itaiad200, line 1: Thanks for this awesome infra test package!

itaiad200 (changes requested):
Thanks! I find some things a little unclear and need more documentation, other than that mainly asking questions..

pkg/kv/store.go
nopcoder: Data Source Name. The short name for connection string.

pkg/kv/store.go
nopcoder: 1. Drop -> Delete
2. DropByPrefix -> DeleteByPrefix. Note that, not like staging - we don't have transaction, and it can be implemented by Scan + Delete with the current interface.

pkg/kv/store.go
nopcoder: Went over several libraries (key/value databases related) - all uses native types as key/value - it is string/[]byte for key and []byte for value:
https://github.com/dgraph-io/badger/blob/master/db.go
https://github.com/etcd-io/bbolt/blob/master/bucket.go#L261
https://github.com/recoilme/pudge/blob/master/api.go
https://git.mills.io/prologic/bitcask/src/branch/master/bitcask.go#L121
https://github.com/peterbourgon/diskv/blob/master/index.go

pkg/kv/store.go
nopcoder: Update the description - it should be the current value or nil for no key.

pkg/kv/kvtest/store.go
```diff
@@ -0,0 +1,269 @@
...
+			t.Fatalf("failed to delete key='%s': %s", keyToDel, err)
+		}
+	})
+
+	t.Run("non_exists", func(t *testing.T) {
+		keyToDel := []byte("missing-key-to-delete")
+		err := store.Delete(ctx, keyToDel)
+		if !errors.Is(err, kv.ErrNotFound) {
+			t.Fatalf("delete missing key '%s', err=%v expected=%s", keyToDel, err, kv.ErrNotFound)
+		}
```
arielshaqed, line 108: Do we need to require this behaviour of our KV stores?  It might not be natively supported by some KV stores.  E.g. DynamoDB does not easily support this AFAIK: the [manual says](https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_DeleteItem.html):

> Unless you specify conditions, the DeleteItem is an idempotent operation; running it multiple times on the same item or attribute does not result in an error response.

pkg/kv/kvtest/store.go
```diff
@@ -0,0 +1,269 @@
...
+		newValue := []byte("val2")
+		err = store.SetIf(ctx, key, newValue, value)
+		if err != nil {
+			t.Fatalf("failed to setIf key=%s value=%s pred=%s: %s", key, newValue, value, err)
+		}
+
+		// should NOT set new value if old value is not kept
+		newValue = []byte("val3")
+		err = store.SetIf(ctx, key, newValue, value)
+		if !errors.Is(err, kv.ErrNotFound) {
```
arielshaqed, line 136: This should use a different error code: it is "condition not satisfied", not "item not found" (which does not make sense for a PUT).

pkg/kv/kvtest/store.go
```diff
@@ -0,0 +1,269 @@
...
+			t.Fatalf("setIf err=%s, key=%s, value=%s, pred=%s, expected err=%s", err, key, newValue, value, kv.ErrNotFound)
+		}
+
+		// should NOT set new value if try conditional nil
```
arielshaqed, line 140: What _is_ the desired behaviour for a `nil` conditional?  If it is always to fail then I would rather make it an application error.

pkg/kv/kvtest/store.go
```diff
@@ -0,0 +1,269 @@
...
+		if err := scan.Err(); err != nil {
+			t.Fatal("scan ended with an error", err)
+		}
+		if diff := deep.Equal(entries, sampleData); diff != nil {
+			t.Fatal("scan data didn't match:", diff)
+		}
+	})
+
+	t.Run("part", func(t *testing.T) {
+		scan, err := store.Scan(ctx, []byte("scan-key-5"))
```
arielshaqed, line 209: This test only works because
```go
 	const sampleItems = 10
```
above.  It would fail with
```go
 	const sampleItems = 100
```

OTOH it would be _much_ better, because it would scan multiple items.

arielshaqed (changes requested):
I think there are some differences in behaviour between what the tests say a KV should satisfy and what I believe some KVs can offer.  So I am choosing to block on the test before going further into the PostgreSQL code.

pkg/kv/store.go
nopcoder: Went with the same pattern as Go's `database/sql` - Entries == Rows.  Can call it iterator - but we currently do not support 'Seek' or 'Prev' at this point. Will update.

pkg/kv/store.go
nopcoder: The idea is, you import the implementation you like to support in runtime.
The main application need to anonymous import the implementation packages to use the drivers in the code.
This is the same as Go's hash algorithm and database packages work.

pkg/kv/postgres/store.go
nopcoder: It should be dynamic to support all the future possible database implementations

pkg/kv/postgres/store.go
nopcoder: You are right. I called it migrate because this is the underlying implementation of kv and wanted to support changes in case of internal structure changes - not business level changes.
I'll rename the function, so it will not be confused with the application level migration, which is out of the 'kv' scope.

pkg/kv/postgres/store.go
nopcoder: More than primary key?

pkg/kv/postgres/store.go
nopcoder: I've updated the description to use not found error. I think not found will make the API simple to use - instead of handling two cases: key not found and predicate failed.
WDYT?

pkg/kv/postgres/store.go
nopcoder: Each store holds a connection pool - calling close will close the pool. This code is just protecting our code from crash in case we call to Close twice.
It is helpful in cases that we will want to wrap the Store and call to 'Close' also from the wrapper.

pkg/kv/postgres/store.go
nopcoder: Good catch - I need to check it first

pkg/kv/mem/store.go
nopcoder: optimized

pkg/kv/kvtest/store.go
nopcoder: We need to have something like that for block adapter - it will help contributors to check new implementations and for us to have one test suite.

pkg/kv/kvtest/store.go
nopcoder: May take this into consideration - all Go implemented without 'require' so I'm still fine with that - unless more people can't stand this code, and I'll replace.

pkg/kv/kvtest/store.go
```diff
@@ -0,0 +1,269 @@
...
+	"testing"
+
+	"github.com/go-test/deep"
+	"github.com/treeverse/lakefs/pkg/kv"
+)
+
+type makeStore func(t *testing.T, ctx context.Context) kv.Store
+
+func TestDriver(t *testing.T, name string, dsn string) {
+	ms := makeStoreByName(name, dsn)
```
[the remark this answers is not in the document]
nopcoder: The function `makeStoreByName` generates a function that allocates a new store based on the name and connection string.
The retuned function is a helper function that accepts `t` and identifies as helper for the test function that uses it.

pkg/kv/kvtest/store.go
```diff
@@ -0,0 +1,269 @@
...
+	defer store.Close()
+
+	testKey := []byte("key")
+	testValue1 := []byte("value1")
+	testValue2 := []byte("value2")
+
+	// set test key with value1
+	err := store.Set(ctx, testKey, testValue1)
+	if err != nil {
+		t.Fatalf("failed to set key '%s', to value '%s': %s", testKey, testValue1, err)
```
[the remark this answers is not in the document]
nopcoder: In this specific case, if calling 'Set' fails, means that calling 'Get' will probably fail too - not getting a list of errors, the first one is enough in this case.
For single operation which multiple tests, I usually user error to check - ex: Add repository, get repository details and check each expected field value and print error - or use one diff with fatal ;)

pkg/kv/kvtest/store.go
```diff
@@ -0,0 +1,269 @@
...
+}
+
+func testStoreSetGet(t *testing.T, ms makeStore) {
+	ctx := context.Background()
+	store := ms(t, ctx)
+	defer store.Close()
+
+	testKey := []byte("key")
+	testValue1 := []byte("value1")
+	testValue2 := []byte("value2")
```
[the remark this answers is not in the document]
nopcoder: Good idea - thanks

pkg/kv/kvtest/store.go
```diff
@@ -0,0 +1,269 @@
...
+	}
+	if val3 != nil {
+		t.Fatalf("get key='%s' value='%s', expected nil", keyNotExists, val3)
+	}
+}
+
+func testStoreDelete(t *testing.T, ms makeStore) {
+	ctx := context.Background()
+	store := ms(t, ctx)
+	defer store.Close()
```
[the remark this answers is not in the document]
nopcoder: I can run the tests in parallel. There is no dependency between the tests - use different keys.
The code assumes that the store implementation can have one global key namespace - so as long as we connect to the same database, we have one key namespace.
Already have control over the underlying postgers table, but it is driver specific I'll create per run ID to use with the keys to enable running the tests over and over or using the same container and still work.

pkg/kv/kvtest/store.go
nopcoder: True - we can simulate it because I'm sure not all our condition will be 1:1 with the underlying features.
Strongly agree in this case - I'll align the implementation.

pkg/kv/kvtest/store.go
nopcoder: I was looking at it as 'not found' as the item with the desired predicate not found.
But @itaiad200 and already pointed this - so I guess most of you prefer different codes to handle - one is key not found, and the second one is for the condition.
Please let me know if I misunderstood - I'll align the implementation.

pkg/kv/kvtest/store.go
nopcoder: The case if setting a key first. Passing nil as precondition is to make sure there is currently no value set.

pkg/kv/kvtest/store.go
```diff
@@ -0,0 +1,319 @@
...
+		if err := scan.Err(); err != nil {
+			t.Fatal("scan ended with an error", err)
+		}
+		if diff := deep.Equal(entries, sampleData[fromIndex:]); diff != nil {
+			t.Fatal("scan data didn't match:", diff)
+		}
+	})
+}
+
+func makeStoreByName(name string, dsn string) makeStore {
```
N-o-Z, line 249: I'd move this to the top with all the non-test functions

pkg/kv/kvtest/store.go
```diff
@@ -0,0 +1,319 @@
...
+
+	t.Run("Delete", func(t *testing.T) {
+		err := store.Delete(ctx, nil)
+		if !errors.Is(err, kv.ErrMissingKey) {
+			t.Errorf("Delete using nil key - err=%v, expected %s", err, kv.ErrMissingKey)
+		}
+	})
+}
+
+func setupSampleData(t *testing.T, ctx context.Context, store kv.Store, prefix string, items int) []kv.Entry {
```
N-o-Z, line 301: Same as above to both these functions

pkg/kv/store.go
```diff
@@ -0,0 +1,111 @@
...
+	Get(ctx context.Context, key []byte) ([]byte, error)
+
+	// Set stores the given value, overwriting an existing value if one exists
+	Set(ctx context.Context, key, value []byte) error
+
+	// SetIf returns an ErrPredicateFailed error if the key with valuePredicate passed
+	//  doesn't match the currently stored value. SetIf is a simple compare-and-swap operator:
+	//  valuePredicate is either the existing value, or nil for no key.
+	//  this is intentionally simplistic: we can model a better abstraction on top, keeping this interface simple for implementors
+	SetIf(ctx context.Context, key, value, valuePredicate []byte) error
```
N-o-Z, line 41: So if valuePredicate is nil this is the equivalent of 'set if not exist'?

pkg/kv/store.go
```diff
@@ -0,0 +1,111 @@
...
+	driversMu.RLock()
+	d, ok := drivers[name]
+	driversMu.RUnlock()
+	if !ok {
+		return nil, fmt.Errorf("%w: %s", ErrUnknownDriver, name)
+	}
+	return d.Open(ctx, dsn)
+}
+
+func (e *Entry) String() string {
```
N-o-Z, line 106: Can you move it under the Entry type definition?

pkg/kv/store.go
```diff
@@ -0,0 +1,111 @@
...
+
+// Entry holds a pair of key/value
+type Entry struct {
+	Key   []byte
+	Value []byte
+}
+
+// map drivers implementation
+var (
+	drivers   = make(map[string]Driver)
```
N-o-Z, line 77: I didn't quite understand - can you explain why we need a map of drivers? Will we not have a single implementation used in any given time?

N-o-Z: If the purpose was to just store all the available implementations, I think we still need to add protection to the open flow and allow only a single driver to be opened

pkg/kv/store.go
nopcoder: yes

pkg/kv/store.go
nopcoder: Driver implementation register to map of drivers.
The main app should anonymous import any driver it wishes to support.
The app at runtime can use one of the registered drivers as the underlying implementation to use.
Calling 'Open' enables you to use driver to open access to the Store.
We do support multiple implementations and parallel calls to open.
Why do you want to restrict a single driver or a single store?

pkg/kv/mem/store.go
```diff
@@ -0,0 +1,169 @@
...
+	m    map[string][]byte
+	keys []string
+	mu   sync.RWMutex
+}
+
+type Entries struct {
+	entry *kv.Entry
+	err   error
+	start []byte
+	first bool
```
N-o-Z, line 24: I might be missing something, but I didn't find a real usage for 'first'

pkg/kv/store.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+func Open(ctx context.Context, name string, dsn string) (Store, error) {
+	driversMu.RLock()
+	d, ok := drivers[name]
+	driversMu.RUnlock()
+	if !ok {
+		return nil, fmt.Errorf("%w: %s", ErrUnknownDriver, name)
+	}
+	return d.Open(ctx, dsn)
```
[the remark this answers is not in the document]
nopcoder: Thanks - Update the code documentation.
The implementor can return the same storage or a new one as long as it is connected to the same storage specified in the dsn.

pkg/kv/mem/store.go
nopcoder: good catch - at first I was planning to use it to select the first item by search for >= at first and later > based on previous key. modified the implementation to get the next key during 'Next' so flag is redundant.

pkg/kv/mem/store.go
```diff
@@ -0,0 +1,169 @@
...
+		start: start,
+	}, nil
+}
+
+func (s *Store) Close() {}
+
+func (e *Entries) Next() bool {
+	if e.start == nil {
+		return false
+	}
```
N-o-Z, line 136: Should we not check for e.err as well? For example I can perform Close() and then Next() and it will work
Same for Entry()?
Same question for postgres implementation

N-o-Z (changes requested):
Thanks @nopcoder, I learned a lot

pkg/kv/mem/store.go
nopcoder: 1. true - fixed in the 4min ago commit
2. entry should be covered as Next implementation will set the right value. There is no garenty on using any object after calling Close. I'll add a code doc to exlpain that.
3. same handling can be found in postgres too

nopcoder: @N-o-Z found a case where entry is not set to nil - fixing it and add test code to verify

nopcoder: ok, false alarm - it was a case where we break the iteration because we got to an entry which doesn't match the same prefix.
Think if a helper method on the iterator should help write code that using it.

pkg/kv/kvtest/store.go
```diff
@@ -0,0 +1,319 @@
...
+	nanoid "github.com/matoous/go-nanoid/v2"
+	"github.com/treeverse/lakefs/pkg/kv"
+)
+
+type makeStore func(t *testing.T, ctx context.Context) kv.Store
+
+var runTestID = nanoid.MustGenerate("abcdef1234567890", 8)
+
+func uniqueKey(k string) []byte {
+	return []byte(k + "-" + runTestID)
```
arielshaqed, line 20: Consider _prepending_ `runTestID` -- keys will sorter nicer, and it can sometimes make it easier to clean up after a test that way.

pkg/kv/kvtest/store.go
```diff
@@ -0,0 +1,319 @@
...
+	"testing"
+
+	"github.com/go-test/deep"
+	nanoid "github.com/matoous/go-nanoid/v2"
+	"github.com/treeverse/lakefs/pkg/kv"
+)
+
+type makeStore func(t *testing.T, ctx context.Context) kv.Store
+
+var runTestID = nanoid.MustGenerate("abcdef1234567890", 8)
```
arielshaqed, line 17: Nit: Why bother to limit it to digits?

pkg/kv/kvtest/store.go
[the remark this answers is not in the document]
arielshaqed: I think we've had this conversation a million times: I like to fail only when the test will crash if it continues; otherwise I like to error out and get ONE MILLION error lines in the log.

[Dr Evil: one MILLION errors!](https://imgflip.com/i/6drbbq).
I think I should stop repeating this comment, I doubt we will ever agree.

pkg/kv/kvtest/store.go
arielshaqed: Cool.  Then let's document it!

I think it would be easier for me to read if this appeared in the next test (which tests behaviour with old value `nil`); perhaps rename it to `OldValueNil`?  That way each case in the function documentation is tested by one function.  Your call, though.

pkg/kv/mem/store.go
```diff
@@ -0,0 +1,172 @@
...
+
+	"github.com/treeverse/lakefs/pkg/kv"
+)
+
+type Driver struct{}
+
+type Store struct {
+	m    map[string][]byte
+	keys []string
+	mu   sync.RWMutex
```
arielshaqed, line 17: nit: in-memory stores are almost always faster with a `Mutex` not a `RWMutex`.

pkg/kv/postgres/store.go
```diff
@@ -0,0 +1,237 @@
...
+		return nil, fmt.Errorf("%w: %s", kv.ErrConnectFailed, err)
+	}
+	defer func() {
+		// if we return before store uses the pool, free it
+		if pool != nil {
+			pool.Close()
+		}
+	}()
+
+	// acquire connection and make sure we reach the database
```
arielshaqed, line 57: At least in a test maybe this should repeat, in case the app is ready before the DB.

pkg/kv/postgres/store.go
```diff
@@ -0,0 +1,237 @@
...
+    key BYTEA NOT NULL PRIMARY KEY,
+    value BYTEA NOT NULL);`)
+	return err
+}
+
+func (s *Store) Get(ctx context.Context, key []byte) ([]byte, error) {
+	if key == nil {
+		return nil, kv.ErrMissingKey
+	}
+	row := s.Pool.QueryRow(ctx, `SELECT value FROM `+s.Params.TableName+` WHERE key = $1`, key)
```
arielshaqed, line 108: This is vaguely unsafe -- the pq  drivers allow quoting identifiers, and this might be a way to inject SQL in some situations (if I can control just the table name, I can pwn the database after this line).

pkg/kv/postgres/store.go
arielshaqed: Note however that there is still a race.  If this is really important we probably want to protect with an atomic op: the tricky cases of closing twice will involve two goroutines.

pkg/kv/postgres/store.go
```diff
@@ -0,0 +1,237 @@
...
+
+func (s *Store) Close() {
+	if s.closed {
+		return
+	}
+	s.Pool.Close()
+	s.closed = true
+}
+
+// Next read the next key/value on any error entry will be nil, err will be set by trying to scan the results
```
arielshaqed, line 199: Why not just be an iterator?
```suggestion
// Next reads the next key/value.
```

pkg/kv/store.go
```diff
@@ -0,0 +1,114 @@
...
+	ErrMissingValue        = errors.New("missing value")
+	ErrNotFound            = errors.New("not found")
+	ErrOperationFailed     = errors.New("operation failed")
+	ErrPredicateFailed     = errors.New("predicate failed")
+	ErrSetupFailed         = errors.New("setup failed")
+	ErrUnknownDriver       = errors.New("unknown driver")
+)
+
+// Driver define the root interface to access the kv database.
+// Each kv provider implements a driver which provides Store instance.
```
arielshaqed, line 24:
```suggestion
// Driver is the interface to access a kv database as a Store.
// Each kv provider implements a Driver.
```

pkg/kv/store.go
```diff
@@ -0,0 +1,114 @@
...
+type Driver interface {
+	// Open access to the database store. Implementation should give access to the same storage based on the dsn.
```
arielshaqed, line 26:
```suggestion
	// Open opens access to the database store. Implementations give access to the same storage based on the dsn.
```
(etc.) -- this should mostly be API documentation for kv users, *not* documentation for people who write Drivers.

pkg/kv/store.go
```diff
@@ -0,0 +1,114 @@
...
+	// SetIf returns an ErrPredicateFailed error if the key with valuePredicate passed
+	//  doesn't match the currently stored value. SetIf is a simple compare-and-swap operator:
+	//  valuePredicate is either the existing value, or nil for no previous key exists.
+	//  this is intentionally simplistic: we can model a better abstraction on top, keeping this interface simple for implementors
+	SetIf(ctx context.Context, key, value, valuePredicate []byte) error
+
+	// Delete will delete the key, no error in if key doesn't exist
+	Delete(ctx context.Context, key []byte) error
+
+	// Scan returns entries that can be read by key order, starting at or after the `start` position
```
arielshaqed, line 48: I am not sure whether this is the most convenient abstraction.  _Most_ users will want to iterate over all keys that start with a prefix.  That's a bit nasty, because there are 2 different stop conditions.  Perhaps add a utility function (*not* a Store method!) that does this?  ("Add another iterator type?  THIS IS TREEVERSE!")

pkg/kv/store.go
```diff
@@ -0,0 +1,114 @@
...
+
+// map drivers implementation
+var (
+	drivers   = make(map[string]Driver)
+	driversMu sync.RWMutex
+)
+
+func Register(name string, driver Driver) {
+	if name == "" {
+		panic("kv store register name is missing")
```
arielshaqed, line 93: Being able to panic probably means this func needs godoc that says as much.  Callers will not read the function, but just might read the godoc.

pkg/kv/store.go
```diff
@@ -0,0 +1,114 @@
...
+	if _, found := drivers[name]; found {
+		panic("kv store Register driver already registered " + name)
+	}
+	drivers[name] = driver
+}
+
+func Open(ctx context.Context, name, dsn string) (Store, error) {
+	driversMu.RLock()
+	d, ok := drivers[name]
+	driversMu.RUnlock()
```
arielshaqed, line 109:
```suggestion
	driversMu.Lock()
	d, ok := drivers[name]
	driversMu.Unlock()
```
and switch to a RW-mutex later, if it is actually any faster.  I worry that we do not spend enough time in the lock to want to bother will full RW mutex fairness.

arielshaqed (approved):
REally neat, and well-documented!  Thanks for explaining it all.

So I asked for even more documentation, and made a few minor suggestions.  None are required.

pkg/kv/kvtest/store.go
nopcoder: just to have something easy to read and add couple of more letters ;)

pkg/kv/kvtest/store.go
nopcoder: Cool. Update.

pkg/kv/store.go
N-o-Z: Single Open driver, at any given time having more than one 'active' driver seems like a bug to me no?

pkg/kv/kvtest/store.go
```diff
@@ -0,0 +1,319 @@
...
+			t.Fatal("scan ended with an error", err)
+		}
+		if diff := deep.Equal(entries, sampleData); diff != nil {
+			t.Fatal("scan data didn't match:", diff)
+		}
+	})
+
+	t.Run("part", func(t *testing.T) {
+		const fromIndex = 5
+		fromKey := append(samplePrefix, []byte(fmt.Sprintf("-key-%04d", fromIndex))...)
```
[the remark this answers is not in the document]
nopcoder: Switching from 10 to 100 samples - wanted the key ordered by number

pkg/kv/mem/store.go
```diff
@@ -0,0 +1,172 @@
...
+	if e.err != nil {
+		return false
+	}
+	if e.start == nil {
+		e.entry = nil
+		return false
+	}
+	e.store.mu.RLock()
+	defer e.store.mu.RUnlock()
+	idx := sort.SearchStrings(e.store.keys, string(e.start))
```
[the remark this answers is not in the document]
nopcoder: The use-case I was worried about was changes during iteration.
- I didn't wanted to capture the index and try to iterate over the original list of items (if still exists)
- Didn't want during the iteration skip new items that added - in the sql version each iteration we query for the next item. This implementation does the same.
- 
The main issue was worried about using index:
1. Scan start
2. Delete the first key based on start
3. Next success
4. Getting a key/value will return item before 'start'

pkg/kv/postgres/store.go
nopcoder: At this point I'll let the app handle backoff and retry.

pkg/kv/postgres/store.go
nopcoder: Will look for a way to do it for a table name - not like query variables table names can't be set with placeholders. The only solution I found to help was to quote it.
Currently the table name can be altered just by configuration and nothing from user data/input.

pkg/kv/store.go
nopcoder: https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/database/sql/sql.go;l=64
Good for Go's database/sql - think it will be good for us. Also I don't predict performance issue as we keep reference to a Store and not calling Open all the time during the current flow of the code.

pkg/kv/mem/store.go
```diff
@@ -0,0 +1,170 @@
...
+func (s *Store) SetIf(_ context.Context, key, value, valuePredicate []byte) error {
+	if key == nil {
+		return kv.ErrMissingKey
+	}
+	if value == nil {
+		return kv.ErrMissingValue
+	}
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	curr := s.m[string(key)]
```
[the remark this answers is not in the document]
nopcoder: I'll update the code to be more explicit - as the current implement doesn't support setting 'nil' as value - I used it as indication.

pkg/kv/store.go
nopcoder: When you say "active" driver you mean "open"? Look at it as you can "open" connection to more than one database type with different connection strings.

pkg/kv/postgres/store.go
nopcoder: Updated the code to use sanitize table name.

pkg/kv/mem/store.go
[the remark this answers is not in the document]
arielshaqed: The way to move _backwards_ during an iteration that just has an `index` is to keep adding elements before it.  Now personally I believe this entirely makes sense (this is not a useful iteration on a kv-store), but I can see why we might want a bit more.

# Suggestions

Do **both** of these!

## Define semantics

Define (on the kv Store interface) the desired semantics of iteration when elements are being added and deleted.  For instance these might be "if the Store is modified before the iteration, then iteration may skip elements or return them multiple times, but will always terminate", or "if the Store is modified, iteration will return each unmodified element exactly once", or anything else.  Note that, at least for RDBMS-backed kvs, returning a "snapshot" of precisely those keys present at start of iteration is the easiest and best result.

The C++ standard library is usually very good about this and can give some useful examples, see for instance [iterator invalidation in `std::unordered_map`](https://en.cppreference.com/w/cpp/container/unordered_map#Iterator_invalidation) or for a sorted example [iterator invalidation in `std::map::erase`](https://en.cppreference.com/w/cpp/container/map/erase) and [iterator invalidation in `std::map::insert`](https://en.cppreference.com/w/cpp/container/map/insert).

## Have the _least useful_ conformant semantics in the example Store

Make the in-memory store conform to the defined semantics, but in the least-useful way.  For instance, if the semantics are "may skip or return multiple times", then one implementation might be to keep an index _and_ the last key, and on `Next` advance it until its new value is after the last key seen.  This implementation will skip elements erased before we start -- which will help us understand whether we really want to do something like that.

pkg/kv/postgres/store.go
arielshaqed: Actually I don't think so.  The way to iterate in Go is to call `Next` until it returns `false`, then examine `Err` to determine whether `Next` stopped because of end-of-iteration or because of an error.  In this case, `e.rows.Next()` on l.203 will return false, so `Next()` will return false.  But then when the caller checks `Err()` it needs to know (immediately, on this call) that something went wrong.  So the check here is required by Go semantics.

pkg/kv/scan.go
```diff
@@ -0,0 +1,55 @@
...
+
+func (b *PrefixIterator) Err() error {
+	return b.Iterator.Err()
+}
+
+func (b *PrefixIterator) Close() {
+	b.Iterator.Close()
+}
+
+// ScanPrefix helper iterator to scan set of keys by prefix
```
arielshaqed, line 45:
```suggestion
// ScanPrefix returns an iterator on store that scan the set of keys that start with prefix
```

pkg/kv/store.go
arielshaqed: We should probably add tests for concurrent iteration on one prefix and Delete on another prefix -- in all possible orders:

* ```
  iteration
               deletion
  ```
* ```
  iteration
        deletion
  ```
* ```
  iiiiiiiiiiteration
      deletion
  ```
* ```
    iteration
  deleeeeeeeetion
  ```
* ```
      iteration
  deletion
  ```
* ```
             iteration
  deletion
  ```

pkg/kv/store.go
arielshaqed: Sure, just seriously not a fan of RWMutex unless absolutely indicated by tests.  I just ran this benchmark with varying numbers of concurrent readers and writers:
[mutex_test.go.gz](https://github.com/treeverse/lakeFS/files/8560366/mutex_test.go.gz).  You need to tune to ~10x more readers than writers to start winning on a plain map.  (And if it's just called at startup, performance doesn't matter and the simpler Mutex wins.)

arielshaqed (approved):
StilLGTM...

nopcoder:
Thanks, @arielshaqed I've created two issues related to the latest feedback to address the semantics and testing the expected behaviour of "delete while iterating" over data behaviour.

pkg/kv/postgres/store.go
itaiad200: Missed that one.

pkg/kv/kvtest/store.go
```diff
@@ -0,0 +1,404 @@
...
+				t.Fatal("scan got nil entry")
+			case ent.Key == nil:
+				t.Fatal("Key is nil while scan item", len(entries))
+			case ent.Value == nil:
+				t.Fatal("Value is nil while scan item", len(entries))
+			}
+			if !bytes.HasPrefix(ent.Key, samplePrefix) {
+				break
+			}
+			entries = append(entries, *ent)
```
itaiad200, line 232: Can you check that the entries are in-order?

itaiad200 (approved):
Thanks for fixing the many many comments. Approving with 2 minor comments.

pkg/kv/kvtest/store.go
```diff
@@ -232,22 +317,88 @@ func makeStoreByName(name string, dsn string) makeStore {
...
+			t.Fatal("ScanPrefix Key is nil on entry", len(entries))
+		case ent.Value == nil:
+			t.Fatal("ScanPrefix Value is nil on entry", len(entries))
+		}
+		entries = append(entries, *ent)
+	}
+	if err := scan.Err(); err != nil {
+		t.Fatal("ScanPrefix ended with an error", err)
+	}
+	if diff := deep.Equal(entries, sampleData); diff != nil {
```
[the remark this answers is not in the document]
nopcoder: Using `deep.Equal` to match the complete expected entries - order is included.

pkg/kv/kvtest/store.go
```diff
@@ -0,0 +1,404 @@
...
+		}
+		if err := scan.Err(); err != nil {
+			t.Fatal("scan ended with an error", err)
+		}
+		if diff := deep.Equal(entries, sampleData[fromIndex:]); diff != nil {
+			t.Fatal("scan data didn't match:", diff)
+		}
+	})
+
+	t.Run("part", func(t *testing.T) {
```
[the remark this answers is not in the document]
nopcoder: bad copy/paste - deleted. thanks!

nopcoder:
@N-o-Z added functionality to list drivers

N-o-Z (approved):

pkg/kv/kvtest/store.go
nopcoder: Using diff to compare results

PR #3252, opened 2022-04-24
Force catalog ID in metastore client config to be a string

### LINKED ISSUE #2996: Glue catalog_id parsed as int
When specifying a Glue catalog id in the config file, the value is read as an integer. In cases where the account id has a leading zero all arns generated for resources in the account will be incorrect. Either the docs need to be updated to specify that the account id / catalog id should be quoted or the parser needs a look.

```
metastore:
  type: glue
  glue:
    catalog-id: 023456789012
```

should be 

```
metastore:
  type: glue
  glue:
    catalog-id: "023456789012"
```

arielshaqed:
Add a type `OnlyString`, used in configuration, that _requires_ a type to be specified in YAML as a `string`.

Fixes #2996.

### Explanation

YAML determines that a field is a number if it looks like one.  And Viper
parses the config file as YAML, then applies the desired structure to it.
So if you specify a catalog ID with a leading zero:

1. YAML sees a number and parses it.
   * The number has a leading zero, which gets dropped.
1. Viper translates the number to a string to place in CatalogID.
1. We just lost the leading zero!

Instead force it to be a string and fail parsing if we got anything else.


talSofer (approved):
Nice catch, thanks!

arielshaqed:
Thanks for the review!

> Nice catch, thanks!

This one is all #2996, reported by @phzietsman :-)

@nopcoder: we (probably) want to do this, _automatically_, across the board for all `string` values in config.  We know that some strings are parsed incorrectly in the current setup.  (The issue is that viper "weak typing" support is too broad, and disabling it will probably kill many conversions we _do_ want.  But need to figure this out :-)

PR #3255, opened 2022-04-24
KV Store: Add KV feature flag

N-o-Z:
Add KV feature flag to configuration

Disabled by default
Added getter for KV value in config
Added GetConfigAsJson to config
Added trace log of config on LoadConfig

### DIFF
--- a/cmd/lakefs/cmd/root.go
+++ b/cmd/lakefs/cmd/root.go
@@ -51,6 +51,13 @@ func loadConfig() *config.Config {
 		fmt.Println("Failed to load config file", err)
 		os.Exit(1)
 	}
+	logger := logging.Default().WithField("phase", "startup")
+	prettyCfg, err := cfg.GetConfigAsJson()
+	if err != nil {
+		fmt.Println("Failed getting configuration JSON", err)
+		os.Exit(1)
+	}
+	logger.Trace("Configuration loaded:", string(prettyCfg))
 	return cfg
 }
 
--- a/cmd/lakefs/cmd/root.go
+++ b/cmd/lakefs/cmd/root.go
@@ -52,7 +52,7 @@ func loadConfig() *config.Config {
 		os.Exit(1)
 	}
 	logger := logging.Default().WithField("phase", "startup")
-	prettyCfg, err := cfg.GetConfigAsJson()
+	prettyCfg, err := cfg.GetConfigAsJSON()
 	if err != nil {
 		fmt.Println("Failed getting configuration JSON", err)
 		os.Exit(1)
--- a/cmd/lakefs/cmd/root.go
+++ b/cmd/lakefs/cmd/root.go
@@ -51,13 +51,6 @@ func loadConfig() *config.Config {
 		fmt.Println("Failed to load config file", err)
 		os.Exit(1)
 	}
-	logger := logging.Default().WithField("phase", "startup")
-	prettyCfg, err := cfg.GetConfigAsJSON()
-	if err != nil {
-		fmt.Println("Failed getting configuration JSON", err)
-		os.Exit(1)
-	}
-	logger.Trace("Configuration loaded:", string(prettyCfg))
 	return cfg
 }
 
@@ -116,6 +109,9 @@ func initConfig() {
 	}
 
 	logger.WithFields(cfg.ToLoggerFields()).Info("Config")
+	if viper.GetBool("database.alpha_kv_enabled") {
+		logger.Error("USING KV EXPERIMENTAL FLAG!!! USE AT YOUR OWN RISK!!!")
+	}
 }
 
 // getHomeDir find and return the home directory
--- a/cmd/lakefs/cmd/root.go
+++ b/cmd/lakefs/cmd/root.go
@@ -109,8 +109,8 @@ func initConfig() {
 	}
 
 	logger.WithFields(cfg.ToLoggerFields()).Info("Config")
-	if viper.GetBool("database.alpha_kv_enabled") {
-		logger.Error("USING KV EXPERIMENTAL FLAG!!! USE AT YOUR OWN RISK!!!")
+	if cfg.GetDatabaseParams().KVEnabled {
+		logger.Warn("USING KV EXPERIMENTAL FLAG!!! USE AT YOUR OWN RISK!!!")
 	}
 }
 
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,7 +1,6 @@
 package config
 
 import (
-	"encoding/json"
 	"errors"
 	"fmt"
 	"math"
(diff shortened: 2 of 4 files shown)


cmd/lakefs/cmd/root.go
```diff
@@ -51,6 +51,13 @@ func loadConfig() *config.Config {
 		fmt.Println("Failed to load config file", err)
 		os.Exit(1)
 	}
+	logger := logging.Default().WithField("phase", "startup")
+	prettyCfg, err := cfg.GetConfigAsJson()
+	if err != nil {
+		fmt.Println("Failed getting configuration JSON", err)
+		os.Exit(1)
+	}
+	logger.Trace("Configuration loaded:", string(prettyCfg))
```
nopcoder, line 60: Why do we need this one?

nopcoder (changes requested):
Was thinking about supporting the following in our configuration:
```yaml
kv:
  enabled: true
  type: postgres
  connection_string: postgres://lakefs:lakefs@localhost:5432/postgres?sslmode=disable
```
so for the development phase we will have a section we configure without changing the configuration all the time.

Alternative will be a field called 'type' under 'database' section:
```yaml
database:
  type: "kv"
  connection_string: postgres://lakefs:lakefs@localhost:5432/postgres?sslmode=disable
```
The default (empty) or non "kv" type will use the current postgres enabled - setting to 'kv' will enable all the kv related features.
The +/- in this solution that we are using the the 'database_connection' string for both.

cmd/lakefs/cmd/root.go
N-o-Z: Since configuration is changed dynamically - I think it will be beneficial to get the current configuration to log whenever we load it

N-o-Z:
> Was thinking about supporting the following in our configuration:
> 
> ```yaml
> kv:
>   enabled: true
>   type: postgres
>   connection_string: postgres://lakefs:lakefs@localhost:5432/postgres?sslmode=disable
> ```
> 
> so for the development phase we will have a section we configure without changing the configuration all the time.
> 
> Alternative will be a field called 'type' under 'database' section:
> 
> ```yaml
> database:
>   type: "kv"
>   connection_string: postgres://lakefs:lakefs@localhost:5432/postgres?sslmode=disable
> ```
> 
> The default (empty) or non "kv" type will use the current postgres enabled - setting to 'kv' will enable all the kv related features. The +/- in this solution that we are using the the 'database_connection' string for both.

@nopcoder  Since this is a temporary parameter I prefer it to have as little footprint as possible. I also don't really see a reason to maintain different connection strings between different implementations, and since the KV store is DB agnostic I don't really understand why we need the 'type: postgres'.
I can move the field under 'database' and rename it to 'kv_enabled' if it makes more sense to you. WDYT?

nopcoder:
kv configuration is part of the flag configuration. we will have multiple underlying implementations so the type property is required to select which one you like to use.
in addition you will need a connection string - I'm fine it will be the same one. the reason to have another one is additional parameters as we experiment. table name used for kv, number of connections and etc that you will want different values than the current connection.

pkg/config/template.go
```diff
@@ -95,6 +95,8 @@ type configuration struct {
 		MaxOpenConnections    int32         `mapstructure:"max_open_connections"`
 		MaxIdleConnections    int32         `mapstructure:"max_idle_connections"`
 		ConnectionMaxLifetime time.Duration `mapstructure:"connection_max_lifetime"`
+		DBType                string        `mapstructure:"type"`
```
nopcoder, line 98:
```suggestion
		Type                string        `mapstructure:"type"`
```

cmd/lakefs/cmd/root.go
```diff
@@ -109,6 +109,9 @@ func initConfig() {
 	}
 
 	logger.WithFields(cfg.ToLoggerFields()).Info("Config")
+	if viper.GetBool("database.alpha_kv_enabled") {
```
nopcoder, line 112: Prefer to use the configuration structure - it will hold refactor tools and 'find usage'

pkg/config/config.go
```diff
@@ -167,6 +169,8 @@ const (
 
 	EmailLimitEveryDurationKey = "email.limit_every_duration"
 	EmailBurstKey              = "email.burst"
+
+	KVStoreKey = "database.alpha_kv_enabled"
```
nopcoder, line 173:
```suggestion
	KVStoreEnabledKey = "database.kv_enabled"
```

pkg/config/template.go
```diff
@@ -190,4 +192,5 @@ type configuration struct {
 		LimitEveryDuration time.Duration `mapstructure:"limit_every_duration"`
 		Burst              int           `mapstructure:"burst"`
 	}
+	KVStore bool `mapstructure:"kv_store"`
```
nopcoder, line 195: Not required

pkg/db/params/database.go
```diff
@@ -8,4 +8,6 @@ type Database struct {
 	MaxOpenConnections    int32
 	MaxIdleConnections    int32
 	ConnectionMaxLifetime time.Duration
+	DBType                string
```
nopcoder, line 11:
```suggestion
	Type                string
```

nopcoder (changes requested):
- Trying to keep the names of the properties align to the keys
- The flag is a feature flag. If possible, not to use 'alpha' as part of the name. It is marked as beta feature - I think it is enough.
- Optional - you can remove the default set and just keep it unset, as the default will be false. Less code to remove later.

pkg/config/template.go
```diff
@@ -95,6 +95,8 @@ type configuration struct {
 		MaxOpenConnections    int32         `mapstructure:"max_open_connections"`
 		MaxIdleConnections    int32         `mapstructure:"max_idle_connections"`
 		ConnectionMaxLifetime time.Duration `mapstructure:"connection_max_lifetime"`
+		Type                  string        `mapstructure:"type"`
```
itaiad200, line 98: What is `Type`?

cmd/lakefs/cmd/root.go
```diff
@@ -109,6 +109,9 @@ func initConfig() {
 	}
 
 	logger.WithFields(cfg.ToLoggerFields()).Info("Config")
+	if cfg.GetDatabaseParams().KVEnabled {
+		logger.Error("USING KV EXPERIMENTAL FLAG!!! USE AT YOUR OWN RISK!!!")
```
nopcoder, line 113: Warn?

nopcoder (approved):
LGTM

nopcoder:
Just added a comment about the log level of the warning

pkg/config/template.go
N-o-Z: Same convention as in Blockstore Type
Used to define which DB implementation to use, it is in fact the name of the driver to load

itaiad200: Let's document that with the possible values. Isn't it an enum?

PR #3256, opened 2022-04-24
Enable configuring the cookie domain

guy-har:


docs/reference/configuration.md
```diff
@@ -42,6 +42,7 @@ This reference uses `.` to denote the nesting of values.
 * `auth.cache.ttl` `(time duration : "20s")` - How long to store an item in the auth cache. Using a higher value reduces load on the database, but will cause changes longer to take effect for cached users.
 * `auth.cache.jitter` `(time duration : "3s")` - A random amount of time between 0 and this value is added to each item's TTL. This is done to avoid a large bulk of keys expiring at once and overwhelming the database.
 * `auth.encrypt.secret_key` `(string : required)` - A random (cryptographically safe) generated string that is used for encryption and HMAC signing
+* `auth.cookie_domain` `(string : mydomain.com)` - Domain attribute to set the access_token cookie on (defaults to the same host that sets the cookie)
```
nopcoder, line 45: The 'mydomain.com' is placed at the default value location - we need to specify it is optional. That the default is empty - which we will not set on the cookie, which means the specific host will be matched. Any domain that will be set will also include sub-domains.

nopcoder (approved):
LGTM just comment about the documentation - trying to explain better when and how to set the proper value.

docs/reference/configuration.md
guy-har: Changed the default, edited it a bit to explain it better, and added a link to MDN

PR #3258, opened 2022-04-25
Add TLD for Hadoop LakeFSOutputCommitter

### LINKED ISSUE #2042: Add a lakeFS-native atomic OutputCommitter
# What

Add an OutputCommitter that uses _existing_ lakeFS operations for **atomic** commits that are efficient and safely concurrent.

# Why <a href="#user-content-why" id="user-content-why">#</a>

## Clearly correct by construction

Rather than rely on per-file atomic operations and carefully tailoring operations to Spark retry mechanisms, directly use lakeFS.

## FAST

Don't copy any data, ideally pay only the price of creating a new commit.

## Data lineage and beautiful semantics

Spark commits should be, well, _commits_.  The history of a Spark job should appear right there in lakeFS history.  Data lineage should appear in the commit metadata.

## Bypass tricky semantics

HDFS and S3A have *different* semantics.  Existing committers try to bridge these semantics, and in so doing overlay very specific but undocumented semantics on FileSystems, in particular on the `mkdirs` and `rename` operations.  A new committer would have _our_ semantics; it will work *only* with a `LakeFSFileSystem` and so can use entirely new operations `createBranch` and `mergeBranch`.

# How

To start a commit, branch out from the desired output lakeFS branch with a unique name.  (Ideally we will have some nice name for the output committer to use; at the very least it might include the Spark task ID.)  Because the name is unique, this branching out might occur only when the first file is written, allowing the committer to be stateless.

The committer writes all files to its branch.  When the time comes to commit, merge the working branch back to the output branch.  Include all available metadata.

Never retry on lakeFS conflicts: if it failed once, it will always fail.

## Why use lakeFS merges not (just) commits?

Multiple OutputCommitters may run concurrently, but share the same staging area.  Branch-per-OutputCommitter separates these staging areas.  Using a shared staging area and commits directly to the branch would require a new commit-certain-paths-only operation, which introduces new issues (and needs basically identical semantics to merges).  The branch per output committer model here is easier to get right, no server-side changes needed.

We may have a branchless OutputCommitter... later.

# Notes

Need to figure (at least) these out:

1. Parquet requires its committers to be [`ParquetOutputCommitter`](https://github.com/apache/parquet-mr/blob/5608695f5777de1eb0899d9075ec9411cfdf31d3/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputCommitter.java#L37)s (of course it does), see e.g. [Cloudera's explanation](https://docs.cloudera.com/HDPDocuments/HDP3/HDP-3.1.5/bk_cloud-data-access/content/enabling-directory-committer-spark.html).  It's not intended to be derived from, and cannot easily use another OutputCommitter.  However there's a [`BindingParquetOutputCommitter`](https://github.com/apache/spark/blob/08e6f633b5bc3a7d8d008db2a264b1607d269f25/hadoop-cloud/src/hadoop-3/main/scala/org/apache/spark/internal/io/cloud/BindingParquetOutputCommitter.scala#L37) in Spark (not in Hadoop, not in Parquet-MR, in *Spark*) that claims to transform the selected output committer into a ParquetOutputCommitter.
2. Committer definition is per output format.  Can the FileSystem intervene?  If not, our OutputCommitter will ideally gracefully degrade to the "otherwise configured OutputCommitter" when not running on lakeFS.
3. Is there a single initialization for a committer across the entire cluster?  Otherwise the unique name needs to include a hash of enough values to be deterministic _and_ unique.
4. Do we need to handle conflicts when the same file is created twice concurrently by different committers?

arielshaqed:
Fixes #2042 in the sense that it updates it an adds it as a design proposal.  We can transform it into the relevant epic and issues from here.

### DIFF
--- a/design/open/spark-outputcommitter/committer.md
+++ b/design/open/spark-outputcommitter/committer.md
@@ -53,11 +53,12 @@ These issues remain with the magic and staging committer:
   > consider it the least mature of the committers.
 * Documentation is still somewhat lacking.
 
-Other OutputCommitters are not as good, and _rename by copying_.  This is
-both slow and highly non-atomic, as well as making it difficult to recover
-from failed attempts.  The current recommendations are to use the
-partitioned staging committer for overwriting or updating partitioned data
-trees, and the magic committer is most other cases.
+The both versions 1 and 2 of the FileOutputCommitter are not as good.  They
+_rename_ files, which on S3A works by _copying and deleting_.  This is both
+slow and highly non-atomic, as well as making it difficult to recover from
+failed attempts.  The current recommendations are to use the partitioned
+staging committer for overwriting or updating partitioned data trees, and
+the magic committer is most other cases.
 
 ## LakeFSOutputCommitter
 
--- a/design/open/spark-outputcommitter/committer.md
+++ b/design/open/spark-outputcommitter/committer.md
@@ -1,0 +1,143 @@
+# Proposal
+
+Add a Hadoop OutputCommitter that uses _existing_ lakeFS operations for
+**atomic** commits that are efficient and safely concurrent.
+
+# Why <a href="#user-content-why" id="user-content-why">#</a>
+
+## Issues with existing OutputCommitters
+
+OutputCommitters are charged with writing a directory tree (a multi-part
+Hadoop file, for example a partitioned Parquet file) atomically.
+
+Hadoop comes with several OutputCommitter algorithms.  For S3 these try to
+bridge the gap between the Hadoop assumed semantics of a POSIX-ish
+filesystem (HDFS) and the actual semantics of an object store (S3).
+
+The best correct committer is probably the [magic committer][magic].  It
+works using only S3 (with its current consistency guarantees!), without
+auxiliary systems such as a small HDFS or S3Guard.  The magic committer
+uploads a directory tree by starting a multipart upload to its intended location,
+creating a completion record for that multipart upload on S3, and uploading
+the object without completing the upload.  To complete the upload, all
+completion records are processed and the objects magically appear.
+
+The [staging committer][staging] (aka "Netflix committer") is an older
+committer for S3A.  It requires a "staging" filesystem on HDFS and offers
+manual modes for overwriting outputs ("conflict resolution").
+
+These issues remain with the magic and staging committer:
+
+* Not atomic.  Multipart objects appear sequentially.  Finding a "_success"
+  object indicator is still required before a directory tree can be
+  processed.
+* The magic committer requires "magic committer" support in the FileSystem,
+  in order to write `__magic` objects to a _different_ path.  (Current
+  lakeFSFS does not contain this support.)
+* The magic committer is somewhat new:
+  > It’s also not been field tested to the extent of Netflix’s committer;
+  > consider it the least mature of the committers.
+* Documentation is still somewhat lacking.
+
+Other OutputCommitters are not as good, and _rename by copying_.  This is
+both slow and highly non-atomic, as well as making it difficult to recover
+from failed attempts.  The current recommendations are to use the
+partitioned staging committer for overwriting or updating partitioned data
+trees, and the magic committer is most other cases.
+
+## LakeFSOutputCommitter
+
+We propose to leverage the atomic capabilities of lakeFS to write a specific
+OutputCommitter for lakeFSFS.  In the initial version, it will branch out,
+prepare the desired output, and merge back in to commit.  Aborting will be
+done by dropping the branch (or repurposing it if the same job ID is
+requested again).  Regular lakeFS retention can handle dropping data
+objects; we might want to add file patterns to retention to allow temporary
+objects to be dropped rapidly.
+
+# How
+
+## Sample flow
+
+* User configures lakeFSFS and configures LakeFSOutputCommitter as the
+  OutputCommitter for `lakefs` protocol Paths.
+* **Setup (job/task TBD)**: Create a new branch for this job/task.  Its name
+  is predictable from the job ID and/or task ID, so can be easily found
+  again.  Immediately delete the entire subree of the intended output
+  path, and possibly delete branches of previous tasks.
+* **Write objects**: Everything is written to its correct path on the branch.
+* **Commit**: Merge back to the original branch.
+
+## Properties
+
+* The merge is performed by lakeFS so it is **atomic**.
+* In the **single-writer case** the merge succeeds: no other operations occur on the subtree.
+* **In-place updates** work: the old objects are deleted and replaced by new
+  objects.  This is true regardess of partitioning etc.
+* **Multiple writers** are detected: the first to commit succeeds, the
+  others fail.
+* (Conflicting) **non-OutputCommitter writes are detected** and clearly
+  handled.  LakeFSOutputCommitter cab achieve its correct semantics
+  regardless of other writers used.
+* **Clearly correct by construction**: Rather than rely on single atomic
+  operations and carefully tailoring operations to Spark retry mechanisms,
+  we use lakeFS capabilities and guarantees.  Analyzing correctness becomes
+  simpler.
+* **Fast**: No data copies, just only (required) metadata operations.  Cost
+  of commit is linear in the number of objects it touches.
+* **Good semantics**: Spark commits will be lakeFS commits.  The history of
+  a Spark job appears right in lakeFS history.  Metadata even includes some
+  data lineage -- and in future we can easily add more, for instance as
+  merge commit user metadata.
+
+### Implementation details
+
+#### Hadoop >=3.1
+
+Hadoop 3.1 offered a fairly complete overhaul of committer architecture,
+configuration, and S3A support.  Supporting new committers on older Hadoops
+will be challenging.  It also seems to be the version where the magic output
+committer is recommended for use, so potentially our users will be there or
+will agree to upgrade.
+
+#### ParquetOutputCommitters
+
+Parquet requires its committers to be
+[`ParquetOutputCommitter`](https://github.com/apache/parquet-mr/blob/5608695f5777de1eb0899d9075ec9411cfdf31d3/parquet-hadoop/src/main/java/org/apache/parquet/hadoop/ParquetOutputCommitter.java#L37)s
+(of course it does), see e.g. [Cloudera's
+explanation](https://docs.cloudera.com/HDPDocuments/HDP3/HDP-3.1.5/bk_cloud-data-access/content/enabling-directory-committer-spark.html).
+It's not intended to be derived from, and cannot easily use another
+OutputCommitter.  However there's a
+[`BindingParquetOutputCommitter`](https://github.com/apache/spark/blob/08e6f633b5bc3a7d8d008db2a264b1607d269f25/hadoop-cloud/src/hadoop-3/main/scala/org/apache/spark/internal/io/cloud/BindingParquetOutputCommitter.scala#L37)
+in Spark (not in Hadoop, not in Parquet-MR, in *Spark*) that claims to
+transform the selected output committer into a ParquetOutputCommitter.  This
+is resolvable per documentation but will require some work.
+
+## Future work
+
+### Multiple writers
+
+We can support multiple concurrent writers and allow all writers to succeed
+(keeping the results of the last writer).
+
+We add a "merge-if" operation to lakeFS: atomically merge branch B into
+branch A if branch A is at a given commit.
+
+Now change LakeFSOutputCommitter to loop during commit:
+
+* Attempt to merge the task branch into the source branch _if the source
+  branch has not moved_.
+* On failure:
+  - Merge the source branch into the task branch using the "destination
+	wins" strategy and delete all added files under the prefix (or add and
+	use a new "merge but *never* copy from source branch" strategy).
+  - Go back and attempt another merge.
+
+This is essentially (noncooperative!) locking of output paths on top of
+lakeFS, with no additional DB.  We can even add cooperation by means of
+various locking hints, _informing_ multiple jobs about attempting to update
+the same paths but keeping things safe regardless.
+
+
+[magic]:  https://hadoop.apache.org/docs/stable/hadoop-aws/tools/hadoop-aws/committers.html#The_Magic_Committer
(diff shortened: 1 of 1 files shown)


design/open/spark-outputcommitter/committer.md
```diff
@@ -0,0 +1,148 @@
...
+objects to be dropped rapidly.
+
+# How
+
+## Sample flow
+
+* User configures lakeFSFS and configures LakeFSOutputCommitter as the
+  OutputCommitter for `lakefs` protocol Paths.  **Possibly** lakeFSFS will
+  set up this OutputCommitter by default.
+* **Setup (job/task TBD)**: Create a new branch for this job/task.  Its name
```
johnnyaug, line 65: Maybe stress that these are operations/functions that are part of the OutputCommitter API.

design/open/spark-outputcommitter/committer.md
```diff
@@ -0,0 +1,148 @@
...
+  is predictable from the job ID and/or task ID, so can be easily found
+  again.  Immediately delete the entire subree of the intended output
```
johnnyaug, line 67: Does this mean that we always overwrite previous data in this path?
Do other committers also do this?

design/open/spark-outputcommitter/committer.md
```diff
@@ -0,0 +1,148 @@
...
+## Properties
+
+* The merge is performed by lakeFS so it is **atomic**.
+* In the **single-writer case** the merge succeeds: no other operations occur on the subtree.
+* **In-place updates** work: the old objects are deleted and replaced by new
+  objects.  This is true regardess of partitioning etc.
+* **Multiple writers** are detected: the first to commit succeeds, the
+  others fail.
+* (Conflicting) **non-OutputCommitter writes are detected** and clearly
+  handled.  LakeFSOutputCommitter cab achieve its correct semantics
```
johnnyaug, line 81:
```suggestion
  handled.  LakeFSOutputCommitter can achieve its correct semantics
```

design/open/spark-outputcommitter/committer.md
```diff
@@ -0,0 +1,148 @@
...
+  regardless of other writers used.
```
johnnyaug, line 82: This one needs more clarification. Why is this possible?

Does this depend on the fact that lakeFS fails to merge when the destination branch is "dirty"?

design/open/spark-outputcommitter/committer.md
```diff
@@ -0,0 +1,148 @@
...
+[`BindingParquetOutputCommitter`](https://github.com/apache/spark/blob/08e6f633b5bc3a7d8d008db2a264b1607d269f25/hadoop-cloud/src/hadoop-3/main/scala/org/apache/spark/internal/io/cloud/BindingParquetOutputCommitter.scala#L37)
+in Spark (not in Hadoop, not in Parquet-MR, in *Spark*) that claims to
+transform the selected output committer into a ParquetOutputCommitter.  This
+is resolvable per documentation but will require some work.
+
+## Future work
+
+### Multiple writers
+
+We can support multiple concurrent writers and allow all writers to succeed
```
johnnyaug, line 125: What is the use-case here and what are the expected results?
Are we talking about multiple concurrent writers to the same path?
Is it expected that the path contain only the results of the latest writer?

johnnyaug (changes requested):
Love this idea, sounds promising.
Added comments asking for more details.

design/open/spark-outputcommitter/committer.md
```diff
@@ -0,0 +1,148 @@
...
+both slow and highly non-atomic, as well as making it difficult to recover
+from failed attempts.  The current recommendations are to use the
+partitioned staging committer for overwriting or updating partitioned data
+trees, and the magic committer is most other cases.
+
+## LakeFSOutputCommitter
+
+We propose to leverage the atomic capabilities of lakeFS to write a specific
+OutputCommitter for lakeFSFS.  In the initial version, it will branch out,
+prepare the desired output, and merge back in to commit.  Aborting will be
```
talSofer, line 52: I think it will be helpful to add a note on terminology to avoid confusing foc commits with lakeFS commits

design/open/spark-outputcommitter/committer.md
talSofer, line 67: this doesn't mean that output names are still going to be dynamic right? 
Are we deleting the latest output to allow repartitioning etc.? I think it worth clarifying this point

design/open/spark-outputcommitter/committer.md
```diff
@@ -0,0 +1,148 @@
...
+  path, and possibly delete branches of previous tasks.
+* **Write objects**: Everything is written to its correct path on the branch.
+* **Commit**: Merge back to the original branch.
+
+## Properties
+
+* The merge is performed by lakeFS so it is **atomic**.
+* In the **single-writer case** the merge succeeds: no other operations occur on the subtree.
+* **In-place updates** work: the old objects are deleted and replaced by new
+  objects.  This is true regardess of partitioning etc.
```
talSofer, line 77: Ok you clarified this point here 😄

design/open/spark-outputcommitter/committer.md
```diff
@@ -0,0 +1,148 @@
...
+* **Multiple writers** are detected: the first to commit succeeds, the
+  others fail.
```
talSofer, line 79: How do we guarantee that others fail?

design/open/spark-outputcommitter/committer.md
```diff
@@ -0,0 +1,148 @@
...
+We add a "merge-if" operation to lakeFS: atomically merge branch B into
+branch A if branch A is at a given commit.
+
+Now change LakeFSOutputCommitter to loop during commit:
+
+* Attempt to merge the task branch into the source branch _if the source
+  branch has not moved_.
+* On failure:
+  - Merge the source branch into the task branch using the "destination
+	wins" strategy and delete all added files under the prefix (or add and
```
talSofer, line 137: Shouldn't we use the "source wins" strategy here? The merge destination is the task branch in this case

talSofer (approved):
A highly impactful feature, and a great write up. Thanks! Added some questions a minor suggestions

design/open/spark-outputcommitter/committer.md
arielshaqed: There is _some_ dynamism from the output format -- it decides filenames.  However the LakeFSOutputCommitter adds no further dynamism.

Added text to explain repartitioning.

design/open/spark-outputcommitter/committer.md
arielshaqed: Added a length explanation.

design/open/spark-outputcommitter/committer.md
arielshaqed: You're correct (of course; as always), so I added some text to say this.

design/open/spark-outputcommitter/committer.md
arielshaqed: Yes and yes.  Users may repeatedly rewrite the same files.  Ideally the _last_ writer should win -- even in the presence of concurrent writes.  (Other strategies used in the field include "locking" the file.  But distributed locking is hard, and in general must be either incorrect or generate spurious errors in any scenario!)

design/open/spark-outputcommitter/committer.md
arielshaqed: Not really.  This is a merge from the source to the task.  Whenever there is a conflict, we want the task branch (which will become the "latest writer" after a successful HOC commit) to win.

Added text. :-)

arielshaqed (commented):
Thanks for *great* reviews!  I think they improved things.

PTAL...

design/open/spark-outputcommitter/committer.md
```diff
@@ -0,0 +1,194 @@
...
+  processed.
+* The magic committer requires "magic committer" support in the FileSystem,
+  in order to write `__magic` objects to a _different_ path.  (Current
+  lakeFSFS does not contain this support.)
+* The magic committer is somewhat new:
+  > It’s also not been field tested to the extent of Netflix’s committer;
+  > consider it the least mature of the committers.
+* Documentation is still somewhat lacking.
+
+Other OutputCommitters are not as good, and _rename by copying_.  This is
```
talSofer, line 56: did you mean this? 
```suggestion
Other OutputCommitters are not as good, and use _rename by copying_.  This is
```
Also, isn't _rename by copying_ a file system property?

design/open/spark-outputcommitter/committer.md
```diff
@@ -0,0 +1,194 @@
...
+  to the Hadoop OutputCommitter protocol.
+* **Setup (job/task TBD)**: Create a new branch for this job/task.  Its name
+  is predictable from the job ID and/or task ID, so can be easily found
+  again.  In "overwrite" mode, immediately delete the entire subtree of the
+  intended output path, and possibly delete branches of previous tasks.
+  This handles cases where the names of the objects written by the output
+  format change, for instance because of repartitioning to a smaller (or
+  different) number of partitions or because of nondeterminism in the names
+  of the objects.
+* **Write objects**: Everything is written to its correct path on the branch.
```
talSofer, line 99: Does this mean that output names are still dynamic but due to the delete step the HOC does, we get to overwrite the previous output written to write the destination path?

talSofer, line 99: Does this mean that output names are still dynamic but due to the delete step the HOC does we get to overwrite the previous output written to write destination path?

design/open/spark-outputcommitter/committer.md
```diff
@@ -0,0 +1,194 @@
...
+## Sample flow
+
+* User configures lakeFSFS and configures LakeFSOutputCommitter as the
+  OutputCommitter for `lakefs` protocol Paths.  **Possibly** lakeFSFS will
+  set up this OutputCommitter by default.
+
+  Successive steps are controlled by Spark/Hadoop to output, and correspond
+  to the Hadoop OutputCommitter protocol.
+* **Setup (job/task TBD)**: Create a new branch for this job/task.  Its name
+  is predictable from the job ID and/or task ID, so can be easily found
```
talSofer, line 92: Do we need to easily find it for troubleshooting purposes?

talSofer, line 92: Do we need to easily find it for troubleshooting purposes?

design/open/spark-outputcommitter/committer.md
```diff
@@ -0,0 +1,194 @@
...
+  (successful) merge.
+* (Conflicting) **non-OutputCommitter writes are detected** and clearly
+  handled.  As long as other writes create _one_ object with an overlapping
+  name the merge will fail.  So LakeFSOutputCommitter can achieve its
+  correct semantics regardless of other writers used.
+* **Clearly correct by construction**: Rather than rely on single atomic
+  operations and carefully tailoring operations to Spark retry mechanisms,
+  we use lakeFS capabilities and guarantees.  Analyzing correctness becomes
+  simpler.
+* **Fast**: No data copies, just only (required) metadata operations.  Cost
```
talSofer, line 122: When you say "no data copies" here you mean that the lakeFS operations create branch and merge don't copy data or it implies that the new HOC relies on non-copying rename?

talSofer, line 122: When you say "no data copies" here you mean that the lakeFS operations create branch and merge don't copy data or it implies that the new HOC relies on non-copying rename?

talSofer (commented):
Thanks for adding more details! I asked more questions (of course :))

design/open/spark-outputcommitter/committer.md
arielshaqed: True.  Reworded to split the explanation between (File) OutputCommitter and FileSystem.

design/open/spark-outputcommitter/committer.md
arielshaqed: Yes, or for debugging, or for cleanup if things go wrong.

design/open/spark-outputcommitter/committer.md
arielshaqed: Sort-of.  We don't control path names, so they could change during an overwrite.  OutputCommitters don't really change final pathnames.  The change here is that objects are only ever written to their correct pathnames.  Again: it is much easier to see and understand what did what and see the source of each object.

design/open/spark-outputcommitter/committer.md
arielshaqed: This proposal only performs metadata operations.

We might (additionally or instead) support "rename" operations correctly.  Then we would perform only metadata operations -- but probably perform more of them, and be non-atomic.

arielshaqed (commented):
Thanks!

Tried to clarify, changed one portion of text.

We have to decide where to perform the discussion of which OC option(s) we want to perform, and when.  I'd rather not do it on this proposal.

arielshaqed:
Forgot to request that people look again; sorry...

# PTAL :-)

talSofer (approved):
Awesome. can't wait we start working on this feature!

arielshaqed:
@johnnyaug you are the only one... :-)

design/open/spark-outputcommitter/committer.md
```diff
@@ -0,0 +1,195 @@
...
+* **[HOC] Commit**: Merge back to the original branch.
+
+## Properties
+
+* The merge is performed by lakeFS so it is **atomic**.
+* In the **single-writer case** the merge succeeds: no other operations
+  occur on the subtree.
+* **In-place updates** work: the old objects are deleted and replaced by new
+  objects.  This is true regardess of partitioning etc.
+* **Multiple writers** are detected and the first fails.  The first to
```
johnnyaug, line 110: The first fails?

johnnyaug (approved):
Love it!
Sorry for the delay

design/open/spark-outputcommitter/committer.md
```diff
@@ -0,0 +1,195 @@
...
+  HOC-commit succeeds.  But all the others fail: they deleted the same
+  previous files, or created a conflicting file (at least their `_SUCCESS`
```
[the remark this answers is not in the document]
arielshaqed: Not necessarily.  Currently we do not usually dedupe these, so it will have a different physical address and conflict.

arielshaqed (commented):
Thanks!

arielshaqed:
Thanks!  Esti timeout failure not relevant to docs, so pulling.

PR #3260, opened 2022-04-26
Add uniquifying values to Esti tests with external storage access

arielshaqed:
* Add a random suffix to _all_ external `s3://...` URLs used in Esti.
 * Split metadata client test into separate GC, export modes.  This allows them to run in parallel and makes Esti succeed more often.

   Not really _required_ for this PR... but it **does** make Esti pass rather than time out when GitHub actions are loaded, so _helpful_ to produce this PR.

### DIFF
--- a/.github/workflows/esti.yaml
+++ b/.github/workflows/esti.yaml
@@ -483,7 +483,7 @@ jobs:
 
       - name: Test rclone export
         env:
-          EXPORT_LOCATION: s3://esti-system-testing/${{ github.run_number }}-rclone-export-dest
+          EXPORT_LOCATION: s3://esti-system-testing/${{ github.run_number }}-rclone-export-dest/${{ steps.unique.outputs.value }}
           S3_ACCESS_KEY: ${{ secrets.ESTI_AWS_ACCESS_KEY_ID }}
           S3_SECRET_KEY: ${{ secrets.ESTI_AWS_SECRET_ACCESS_KEY }}
         working-directory: test/rclone_export
@@ -531,7 +531,7 @@ jobs:
       - name: Setup lakeFS for tests
         working-directory: test/lakectl_metastore
         env:
-          STORAGE_NAMESPACE: s3://esti-system-testing/${{ github.run_number }}-metaclient/${{ secrets.ESTI_AWS_SECRET_ACCESS_KEY }}
+          STORAGE_NAMESPACE: s3://esti-system-testing/${{ github.run_number }}-metaclient/${{ steps.unique.outputs.value }}
         run: ./setup-test.sh
 
       - name: Test using DBT and lakectl metastore commands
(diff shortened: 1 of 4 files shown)


nopcoder (approved):
lgtm

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -366,7 +366,7 @@ object GarbageCollector {
 
   private def repartitionBySize(df: DataFrame, maxSize: Int, column: String): DataFrame = {
     val nRows = df.count()
-    val nPartitions = math.max(1, math.floor(nRows / maxSize)).toInt
+    val nPartitions = math.max(1, math.ceil(nRows / maxSize)).toInt
```
johnnyaug, line 369: Why part of this PR?

.github/workflows/esti.yaml
```diff
@@ -613,37 +637,88 @@ jobs:
...
         working-directory: test/spark
         run: ./run-gc-test.sh
 
+      - name: lakeFS Logs on Spark with gateway failure
+        if: ${{ failure() }}
+        continue-on-error: true
+        working-directory: test/spark
+        run: docker-compose logs --tail=1000 lakefs
+
+  metadata-client-export-spark3:
```
johnnyaug, line 661: New test? Is this related to this PR?

johnnyaug (changes requested):
Some parts here look like they belong to a different PR.
Can you go over the changed files?
If they are relevant, can you change the PR description to reflect it?

.github/workflows/esti.yaml
arielshaqed: Sort of, but mostly in order to pass tests.  In eb8df2b19048ef8173b6e1874b79db30b1e0331b I split the metadata client test into gc and export so that they may run in parallel -- that lets the export test (just a few minutes) run without timing out, and also helps the long-running gc test finish running on time.
Without this, Esti barely managed to sneak in under the deadline -- which would have meant we cannot trust this PR.
I shall update the PR description to reflect this addition. :-)  So now we won't have a dirty hack, we will have a **documented** dirty hack.

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
arielshaqed: Moved it out.  (It's a bug, but not one we need to fix here.)

arielshaqed (commented):
Thanks!

PTAL, especially at the updated description.

lynnro314 (approved):

johnnyaug (approved):

arielshaqed:
THANKS!

Let's hope now everything works :-) ❗   Except that Esti is still too slow, of course.

PR #3265, opened 2022-04-27
Fix TestHooksSuccess

N-o-Z:
Found the issue - it was a timing in the test. Added a backoff mechanism to wait until ListRepositoryRuns returns the expected number of results


itaiad200:
@N-o-Z can you please link the related issue?

esti/hooks_test.go
```diff
@@ -42,6 +43,29 @@ func TestHooksSuccess(t *testing.T) {
 	})
 }
 
+func waitForListRepositoryRunsLen(ctx context.Context, t *testing.T, repo, ref string, l int) *api.ActionRunList {
```
itaiad200, line 46: What action is behaving inconsistently that we need to retry it?
These retries might be hiding a real issue.

esti/hooks_test.go
```diff
@@ -42,6 +43,29 @@ func TestHooksSuccess(t *testing.T) {
...
+	bo := backoff.NewExponentialBackOff()
+	bo.MaxInterval = time.Second * 5
+	bo.MaxElapsedTime = time.Second * 30
+	listFunc := func() error {
+		runsResp, err := client.ListRepositoryRunsWithResponse(ctx, repo, &api.ListRepositoryRunsParams{
+			Commit: api.StringPtr(ref),
+		})
+		require.NoError(t, err)
+		require.Equal(t, http.StatusOK, runsResp.StatusCode())
+		runs = runsResp.JSON200
```
nopcoder, line 58: Need to check if runs / JSON200 is not nil.  Getting 200 doesn't verify that the content type was json and we managed to parse the result.

esti/hooks_test.go
nopcoder, line 46: If we extract the exponential back off configured in this function to another function - the test code just needs to use it and pass, and we will get assert failures to point to the right error during the test execution.

nopcoder (changes requested):

esti/hooks_test.go
N-o-Z: This is a test timing issue, we query for list repository runs before the post merge record is written to the storage so that we get only the pre merge event. In a local setting with a local storage and server this race happens rarely and that's why the test in it's current state fails locally whereas in github we consistently get a single entry (for pre merge) even though we should expect 2

nopcoder (approved):

PR #3267, opened 2022-04-27
[gc] Use right number of partitions to get each of limited size

arielshaqed:
We have `nRows` files to split into `k` partitions, and we want each
partition to have size at most `maxSize` (think 1000, because we can delete
up to 1000 objects in each AWS S3 bulk deletion).  So we must have

    k * maxSize >= nRows > (k-1) * maxSize

or (because `k`, `k-1` are both ints)

    k >= ceil(n / maxSize) >= k >= floor(n / maxSize) >= k-1

Since we want `k` as small as possible,

    k = ceil(n / maxSize)

and `floor` does not work.


lynnro314 (approved):

arielshaqed:

PR #3270, opened 2022-04-27
Fix bug in merge

johnnyaug:
When merging from a branch that made no changes, Graveler would create a new commit with an empty metarange ID, causing all the objects to be deleted on the merge destination.
Instead, we should return a "no changes" error.

## Test changes

Changed the behavior of the unit tests to take into account the metarange IDs, which until now were ignored.

This required changes to the test cases themselves: before this change, metaranges always got fixed IDs. After this change, IDs are generated according to the metarange's content. In some cases this causes the ID of the base metarange to be equal to the ID of the source or dest metaranges.

In order for them to get different IDs, I added some content to the end of some of the bases. Since all addition are only to the merge bases, this doesn't affect the merge results.

### DIFF
--- a/pkg/graveler/committed/merge_test.go
+++ b/pkg/graveler/committed/merge_test.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/golang/mock/gomock"
@@ -66,18 +67,35 @@ type testRunResult struct {
 	expectedErr     error
 }
 
+type testMetaRange struct {
+	ranges []testRange
+}
+
+func (t *testMetaRange) GetMetaRangeID() graveler.MetaRangeID {
+	var sb strings.Builder
+	for _, rng := range t.ranges {
+		sb.WriteString(string(rng.rng.ID))
+		sb.WriteString("#")
+	}
+	return graveler.MetaRangeID(sb.String())
+}
+
+func newTestMetaRange(ranges []testRange) *testMetaRange {
+	return &testMetaRange{ranges: ranges}
+}
+
 type testCase struct {
-	baseRange      []testRange
-	sourceRange    []testRange
-	destRange      []testRange
+	baseRange      *testMetaRange
+	sourceRange    *testMetaRange
+	destRange      *testMetaRange
 	expectedResult []testRunResult
 }
 
 type testCases map[string]testCase
 
-func createIter(tr []testRange) committed.Iterator {
+func createIter(tr *testMetaRange) committed.Iterator {
 	iter := testutil.NewFakeIterator()
-	for _, rng := range tr {
+	for _, rng := range tr.ranges {
 		addRange := rng.rng
 		iter.AddRange(&addRange)
 		for _, record := range rng.records {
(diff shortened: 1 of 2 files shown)


pkg/graveler/committed/merge_test.go
```diff
@@ -66,18 +67,35 @@ type testRunResult struct {
 	expectedErr     error
 }
 
+type testMetaRange struct {
```
nopcoder, line 70: can be []testRange this way we don't need special construction for all the test code.
```go
type testMetaRange []testRange
```

nopcoder (changes requested):
When there are no changes to the source, we want the destination to be the result.
Think that the no change should be returned when both point to base.

arielshaqed:
> When there are no changes to the source, we want the destination to be the result. Think that the no change should be returned when both point to base.

I think that this is correct.  So: we should start by agreeing on the desired test results, and only then proceed to the actual fix.

Note that we can determine the correct result by ignoring the "base === {source, dest}" optimization.  I mean, suppose I took a case where base and source (or dest) have precisely the same diffs, but one range boundary in base is different.  Then the metarange IDs are different and the full algorithm runs.  But the result should be the same.

For instance, we could consider adding a test-only flag that turns off these optimizations.  Now we can run the tests with and without the flag -- the tests should pass regardless!

Or you could add a field to `testMetaRange` that gets appended to its ID.  Now you can copy the tests in disagreement and change that field on `base`.  We should still get the same result!

pkg/graveler/committed/merge_test.go
```diff
@@ -626,29 +687,20 @@ func Test_merge(t *testing.T) {
...
-						action: actionTypeWriteRange,
-						rng:    committed.Range{ID: "dest:k1-k2", MinKey: committed.Key("k1"), MaxKey: committed.Key("k2"), Count: 2, EstimatedSize: 1234},
-					},
-					{
-						action: actionTypeWriteRange,
-						rng:    committed.Range{ID: "base:k3-k6", MinKey: committed.Key("k3"), MaxKey: committed.Key("k6"), Count: 2, EstimatedSize: 1234},
-					},
-				},
-				expectedErr: nil,
+				expectedActions: []writeAction{},
```
arielshaqed, line 698: Concretely, I believe @nopcoder is correct and I do not like this expected result -- I believe the old one was correct.  So I would appreciate an explanation of this change.

arielshaqed: Explained by @guy-har .  THANKS!

arielshaqed (commented):
Thanks!  Still want the test with and without changing the base metarange ID, to ensure we have consistent behaviour.  But this is not blocking for a bug like this.

guy-har:
Adding Here This comment just to explain what exactly happened, because it may be a bit confusing

When trying to introduce a merge with no changes an error should be returned
There was a bug that returned an empty string and no error, which then set the metarangeID of the merge commit to be an empty string

image

nopcoder (approved):
Approve as keeping the previous behaviour. Opening an issue to address the behaviour that will allow us to merge when no changes from base.

PR #3272, opened 2022-04-28
Emailer config smpt_port with additional ssl and local_name

nopcoder:
Address the following for our emailer configuration:
1. Update configuration from port to smtp_port to align with smtp_host.
2. Add use_ssl to configure mailer to use ssl not just on default ssl port.
3. Configure local_name - how we identify the domain with smtp server.


guy-har (approved):
LGTM 
Thanks!

shimi9276 (approved):
Thanks!

PR #3292, opened 2022-04-29
Changelog for version 64

guy-har:


CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## v0.64.0 - 2022-04-29
+
+- Fix bug in merge (#3270)
```
nopcoder, line 5: and you add that this is related to the case of no changes on source

nopcoder (approved):
Approved, just update the first bug description

CHANGELOG.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## v0.64.0 - 2022-04-29
+
+- Fix bug in merge - merge with no changes resolves by creating empty commit (with no data) on destinations head  (#3270)
+- Fix broken content-based diff for changed objects in compare view (#3275)
+- Bump metadata client version to 0.1.7-RC.0 (#3277)
```
itaiad200, line 7: GC isn't released together with a lakeFS release, right? In that case I think we should avoid adding it to the lakeFS changelog (But there's really no need to fix it now).

arielshaqed: Yup, my mistake there.  OTOH it _does_ sort-of state that we are now testing after cutting this release with lakeFS (we always test HEAD, but obviously a release is a nice step to indicate along the way...).

Issue #2031, opened 2021-05-31
Remove gateway playback test


arielshaqed:
The gateway test does not verify a lot, and is very hard to modify.  Relevant: #1769, #1784, #1785.  It runs pre-recorded S3 commands from some Spark run.  Now we have a live Spark run as part of Nessie.  Add whatever is missing and then get rid of the recordings.

# Known missing:

1. Multipart uploads.  See [`fs.s3a.multipart.size` and similar in hadoop-aws configuration](https://hadoop.apache.org/docs/current/hadoop-aws/tools/hadoop-aws/index.html#general_configuration) for how we might induce Spark to do this.
2. Previously Spark on local gave different results than Spark on EMR.  This is _probably_ because of local vs. cluster differences.  Can we ensure coverage?

arielshaqed:
@guy-har for any other things that we need (you _did_ tell me the above 2, do you remember any others?)

guy-har:
We currently also have recording for Presto and AWS CLI.
* The important thing to check in AWS CLI is multipart upload. We already have a test for that using the AWS SDK, I believe it covers the AWS CLI recording (we might want to add AWS CLI tests anyway, but it shouldn't be related to this issue)
* We should check if the Presto recording is significantly different from Spark recording

arielshaqed:
F2F with @guy-har :

* Dropped Spark/EMR playback as required: this is _not_ causing problems, #2418 is much more urgent (has more users).
* Opened #2657 to do Presto.

So, once #2657 is done... we can remove the gateway playback test!

guy-har:
@arielshaqed, Test for Trino was added. we could now remove the gateway playback test!

Issue #3240, opened 2022-04-19
KV model support functionality


nopcoder:
Support code that enable working with serialized data model from protobuf to the KV library

N-o-Z:
I want to suggest alternatives for using protobuf.
I did some research and came with the following viable alternatives:

1. BSON https://bsonspec.org/ (Binary JSON)
    - Advantages:
      - Ordered
      - Traverse-able
      - Smaller than JSON
      - Quick to encode decode
      - Supported by many languages (uses C data types) https://bsonspec.org/implementations.html
      - Has a built go mongo driver
    - Disadvantages:
      - Slower than protobuf
      - Requires importing the specific implementation package
      - Last update 2015?
14. https://msgpack.org/ (another binary json)
      - Same properties as BSON - seems more maintained
      - Limited in data type sizes (for example max str size 32B)
17. Flatbuffers https://google.github.io/flatbuffers/
      - Advantages:
        - supports a variety of languages
        - Developed by google
        - No additional memory to access data
        - Access data without parsing / unpacking
        - Supports optional fields
        - Last release May2020
        - Supports schema-less version
     - Disadvantages:
       - Requires schema compilation from C++

My tendency is to go with Flatbuffers as it is very memory efficient both in storing and accessing data and very fast (doesn't require unpacking of data)

itaiad200:
Thanks @N-o-Z , it made me smarter :)
I wonder how would the above protocol characteristics map into solving the challenges we arose before (copying our summary for the discussion on this topic):
```
KV Value Format aspects for comparison:
- Data readability
- Should we use ID together with the value?
	- Application level: Can customize the fields for computing the id (e.g. timestamp)
	- DB level: Just like graveler. DB can leverage it to use a shorter field for SetIf
- Backward compatibility
- protobuf has metadata, json doesn't
- Compression; size of deserialised objects
- Popularity of format
```

N-o-Z:
@itaiad200 answers for flatbuffers regarding the issues you've raised:
- Data readability - same as protobuf, the binary format doesn't allow as to read the data as is, but different from protobuf the data can be read without de-serialization if you are familiar with the data structure (or via the object file)
- Supports optionals and defaults so we should be fine with backwards compatibility
- Not needed as it is structured (with a caveat here)
- Compression - we will see that in the benchmarking I made the binary size created by flatbuffers is bigger than protobuff, on the other hand there's no need to de-serialize the data when using flatbuffers
[...]

PR #3297, opened 2022-05-01
KV Store: StoreMessage and multipart upload implementation

### LINKED ISSUE #3242: KV multipart upload implementation

N-o-Z:
Closes #3242 and #3240


cmd/lakefs/cmd/run.go
```diff
@@ -132,7 +134,18 @@ var runCmd = &cobra.Command{
...
-		multipartsTracker := multiparts.NewTracker(dbPool)
+		var multipartsTracker multiparts.Tracker
+
+		if dbParams.KVEnabled {
+			kvStore, err := kv.Open(ctx, dbParams.Type, dbParams.ConnectionString)
+			if err != nil {
+				logger.WithError(err).Fatal("failed to open KV store")
+			}
+			defer kvStore.Close()
+			multipartsTracker = multiparts.NewTracker(kv.StoreMessage{Store: kvStore})
```
itaiad200, line 145: I know `migration` isn't part of this PR, but I would add 
```suggestion
			// TODO: migration required
			kvStore, err := kv.Open(ctx, dbParams.Type, dbParams.ConnectionString)
			if err != nil {
				logger.WithError(err).Fatal("failed to open KV store")
			}
			defer kvStore.Close()
			multipartsTracker = multiparts.NewTracker(kv.StoreMessage{Store: kvStore})
```

pkg/gateway/multiparts/multipart.proto
```diff
@@ -0,0 +1,15 @@
+syntax = "proto3";
+option go_package = "github.com/treeverse/lakefs/gateway/multiparts";
+
+import "google/protobuf/timestamp.proto";
+
+package io.treeverse.lakefs.multiparts;
+
+message MultipartUploadData {
```
itaiad200, line 8: Please add documentation for the type and fields.

pkg/gateway/multiparts/tracker.go
```diff
@@ -22,14 +26,36 @@ type MultipartUpload struct {
...
 
+func (m Metadata) Set(k, v string) {
+	m[strings.ToLower(k)] = v
+}
+
+func (m Metadata) Get(k string) string {
+	return m[strings.ToLower(k)]
+}
+func (m Metadata) Value() (driver.Value, error) {
+	return json.Marshal(m)
```
itaiad200, line 37: json?

pkg/gateway/multiparts/tracker.go
```diff
@@ -22,14 +26,36 @@ type MultipartUpload struct {
...
+}
+
+func (m *Metadata) Scan(src interface{}) error {
```
itaiad200, line 40: Why not accept the concrete type `[]byte`?

pkg/gateway/multiparts/tracker.go
```diff
@@ -38,21 +64,28 @@ var (
...
-		_, err := tx.Exec(`INSERT INTO gateway_multiparts (upload_id,path,creation_date,physical_address, metadata, content_type)
-			VALUES ($1, $2, $3, $4, $5, $6)`,
-			multipart.UploadID, multipart.Path, multipart.CreationDate, multipart.PhysicalAddress, multipart.Metadata, multipart.ContentType)
-		return nil, err
+	_, err := m.Get(ctx, multipart.UploadID)
+	if err == nil {
+		return ErrAlreadyExists
+	}
+
+	err = m.store.SetMsg(ctx, []byte(multipart.UploadID), &MultipartUploadData{
```
itaiad200, line 82: We need hierarchy for the keys in the store. Imagine creating a `Branch` and an `Upload` with the same ID.

pkg/gateway/multiparts/tracker.go
```diff
@@ -61,59 +94,31 @@ func (m *tracker) Get(ctx context.Context, uploadID string) (*MultipartUpload, e
...
-	return m[strings.ToLower(k)]
-}
-func (m Metadata) Value() (driver.Value, error) {
-	return json.Marshal(m)
-}
-
-func (m *Metadata) Scan(src interface{}) error {
-	if src == nil {
-		return nil
+		return fmt.Errorf("failed on Get. key %s, %w", uploadID, err)
```
itaiad200, line 120: golang convention for wrapping errors:
```suggestion
		return fmt.Errorf("failed on Get. key (%s): %w", uploadID, err)
```

pkg/gateway/multiparts/tracker.go
```diff
@@ -61,59 +94,31 @@ func (m *tracker) Get(ctx context.Context, uploadID string) (*MultipartUpload, e
...
 	}
-	data, ok := src.([]byte)
-	if !ok {
-		return ErrInvalidMetadataSrcFormat
-	}
-	return json.Unmarshal(data, m)
+	err := m.store.Delete(ctx, []byte(uploadID))
+	return err
```
itaiad200, line 123:
```suggestion
	return m.store.Delete(ctx, []byte(uploadID))
```

itaiad200: Also - Does the store's `Delete` error on missing key? If so, no check is needed.

pkg/gateway/multiparts/multipart.pb.go
```diff
@@ -0,0 +1,218 @@
+// Code generated by protoc-gen-go. DO NOT EDIT.
```
itaiad200, line 1: Should we commit this generated file?

pkg/kv/store_message.go
```diff
@@ -0,0 +1,63 @@
+package kv
+
+import (
+	"context"
+
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+type StoreMessage struct {
```
itaiad200, line 10: Please add documentation for this Store. Not sure I understand what it does and why we need it.

pkg/kv/store_message.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+	return s.Store.SetIf(ctx, key, val, curr)
+}
+
+func (s *StoreMessage) Delete(ctx context.Context, key []byte) error {
+	return s.Store.Delete(ctx, key)
+}
+
+// TODO: niro - implement when required
+// func (s *StoreMessage) Scan(ctx context.Context, start string) (Msgs, error) {
```
itaiad200, line 56: It's better not to keep commented-out code.

pkg/kv/store_message_test.go
```diff
@@ -0,0 +1,211 @@
...
+			"blue":  3,
+		},
+		TestList: []bool{true},
+	}
+
+	// SetIf fails nil
+	err = sm.SetIf(ctx, []byte(setModel.Name), m1, nil)
+	require.Error(t, kv.ErrPredicateFailed, err)
+
+	// SetIf fails nil
```
itaiad200, line 107: wrong comment

pkg/gateway/multiparts/tracker_test.go
```diff
@@ -73,6 +86,9 @@ func TestTracker_Get(t *testing.T) {
 				t.Errorf("Get() error = %v, wantErr %v", err, tt.wantErr)
 				return
 			}
+			if !tt.wantErr {
+				got.CreationDate = got.CreationDate.UTC() // deepEqual workaround for time
```
itaiad200, line 90: So the time is the only thing that could fail?
What if you expect a different error and everything unexpectedly works as expected, except for the `CreationDate` which is flaky?

pkg/gateway/multiparts/tracker.go
```diff
@@ -38,21 +64,28 @@ var (
 	ErrInvalidMetadataSrcFormat = errors.New("invalid metadata source format")
 )
 
-func NewTracker(adb db.Database) Tracker {
+func NewTracker(ms kv.StoreMessage) Tracker {
```
itaiad200, line 67: The name is generic, but it returns an implementation of the `Tracker`, no?

pkg/gateway/playback_test.go
```diff
@@ -128,7 +128,7 @@ func getBasicHandler(t *testing.T, authService *simulator.PlayBackMockConf) (htt
 		DB:     conn,
 	})
 	testutil.MustDo(t, "build c", err)
-	multipartsTracker := multiparts.NewTracker(conn)
+	multipartsTracker := multiparts.NewDBTracker(conn)
```
itaiad200, line 131: When will we test the new tracker here?

pkg/gateway/multiparts/tracker.go
itaiad200: OK it was just moved, never mind..

itaiad200 (changes requested):
Thanks for the first pkg implementation of the KV!
I have some concerns but overall looks great.

pkg/gateway/multiparts/multipart.pb.go
N-o-Z: As in graveler - the pb files are committed to prevent repository from breaking

cmd/lakefs/cmd/run.go
N-o-Z: I'm not sure this will be the place where migration will happen, and since we don't know who might take this task, this TODO can easily be forgotten and left in the code. I'm confident we won't forget to implement the migration even without this TODO :)

pkg/gateway/multiparts/tracker.go
N-o-Z: This code was not touched - just moved in the file.
If we change this interface we will need also to touch the tracker usage which I was very careful not to disturb

pkg/gateway/multiparts/tracker.go
N-o-Z: I think this should be solved by making sure we create a unique hash (UUID / GUID) every time. Creating a conflicting BranchID / UploadID / RefID is a bug in my opinion

pkg/gateway/multiparts/tracker.go
N-o-Z: Delete doesn't error on missing key

pkg/gateway/multiparts/tracker_test.go
N-o-Z: The CreationDate is not flaky it's just a different behavior between how it was saved originally in the DB and how protobuf's time object saves it.
protobuf's time object (timestamppb) doesn't save the timezone information so when read from the protobuf object we get the UTC time whereas the DB implementation when we read the timezone information, we get the time formatted using local timezone. The DeepEqual function has a "bug" where it compares two time objects without taking into account the timezone and this is simply a workaround to allow using DeepEqual and testing both implementations

pkg/gateway/multiparts/tracker.go
N-o-Z: This is the original name of the tracker and when we will be at feature complete this will also be the only implementation

pkg/gateway/playback_test.go
N-o-Z: Correct - fixed

pkg/gateway/multiparts/multipart.proto
N-o-Z: We can add documentation for the message but I think the fields should be documented in the original struct

pkg/gateway/multiparts/tracker.go
```diff
@@ -61,59 +101,30 @@ func (m *tracker) Get(ctx context.Context, uploadID string) (*MultipartUpload, e
...
 
 func (m *tracker) Delete(ctx context.Context, uploadID string) error {
 	if uploadID == "" {
 		return ErrInvalidUploadID
 	}
-	_, err := m.db.Transact(ctx, func(tx db.Tx) (interface{}, error) {
-		res, err := tx.Exec(`DELETE FROM gateway_multiparts WHERE upload_id = $1`, uploadID)
-		if err != nil {
-			return nil, err
+	if _, err := m.Get(ctx, uploadID); err != nil {
```
nopcoder, line 123: This is good for debug, but not for performance. We can just call Delete on the store.

pkg/gateway/multiparts/tracker_test.go
```diff
@@ -7,20 +7,33 @@ import (
...
+}
 
-	creationTime := time.Now().Round(time.Second) // round in order to remove the monotonic clock
+func TestDBTracker_Get(t *testing.T) {
+	tracker := testDBTracker(t)
+	testTrackerGet(t, tracker)
+}
+
+func testTrackerGet(t *testing.T, tracker multiparts.Tracker) {
+	t.Helper()
```
nopcoder, line 34: would remove this one to get a better stack on failed test

pkg/gateway/multiparts/tracker_test.go
```diff
@@ -7,20 +7,33 @@ import (
...
 	t.Helper()
 	conn, _ := testutil.GetDB(t, databaseURI)
-	return multiparts.NewTracker(conn)
+	return multiparts.NewDBTracker(conn)
 }
 
 func TestTracker_Get(t *testing.T) {
-	ctx := context.Background()
-	tracker := testTracker(t)
+	store := kvtest.MakeStoreByName("mem", "")(t, context.Background())
```
nopcoder, line 22: 1. still prefer to test over postgres at this point
2. would split into two function to have code easy to read
3. why don't we have testTracker(t) to return a new tracker based on store - like we have for DB?

pkg/gateway/playback_test.go
```diff
@@ -45,7 +48,17 @@ var (
...
 
+func TestGatewayRecordingDB(t *testing.T) {
+	viper.Set("database.kv_enabled", false)
+	testGatewayRecording(t)
+}
+
 func TestGatewayRecording(t *testing.T) {
+	viper.Set("database.kv_enabled", true)
+	testGatewayRecording(t)
+}
```
nopcoder, line 59: 1. would not play with global viper configuration during test if possible - just passing an argument to testGatewayRecording if possible
2. check with @guy-har if this is still relevant or we can remove the recording parts

pkg/gateway/playback_test.go
```diff
@@ -128,7 +140,16 @@ func getBasicHandler(t *testing.T, authService *simulator.PlayBackMockConf) (htt
 		DB:     conn,
 	})
 	testutil.MustDo(t, "build c", err)
-	multipartsTracker := multiparts.NewTracker(conn)
+
+	var multipartsTracker multiparts.Tracker
+	dbParams := conf.GetDatabaseParams()
```
nopcoder, line 145: I don't think the test code need to use the configuration - it should setup the right services composition it would like to test

pkg/kv/store_message.go
```diff
@@ -0,0 +1,60 @@
...
+		return err
+	}
+	return s.Store.SetIf(ctx, key, val, curr)
+}
+
+func (s *StoreMessage) Delete(ctx context.Context, key []byte) error {
+	return s.Store.Delete(ctx, key)
+}
+
+// TODO: niro - implement when required
```
nopcoder, line 56: still relevant?

pkg/kv/store_message.go
```diff
@@ -0,0 +1,60 @@
...
+		return err
+	}
+	return proto.Unmarshal(val, msg)
+}
+
+func (s *StoreMessage) SetMsg(ctx context.Context, key []byte, msg protoreflect.ProtoMessage) error {
+	val, err := proto.Marshal(msg)
+	if err != nil {
+		return err
+	}
```
nopcoder, line 27: relevant for some places in this code - and probably in our storage too - the error should include more information like the key.

pkg/kv/store_message_test.go
```diff
@@ -0,0 +1,211 @@
+package kv_test
```
nopcoder, line 1: Like the non test package - we should keep the keys under specific prefix

nopcoder (changes requested):

pkg/gateway/multiparts/multipart.pb.go
itaiad200: Not all generated code is committed, but I'm fine with keeping Graveler conventions.

pkg/gateway/multiparts/tracker_test.go
itaiad200: So why do it on the condition `!tt.wantErr`?

pkg/gateway/multiparts/tracker.go
itaiad200: Still think that `NewKVTracker` is a better name, but not blocking..

pkg/gateway/multiparts/tracker.go
```diff
@@ -5,14 +5,25 @@ import (
...
+// MultipartUpload data structure
+// UploadID A unique identifier for the uploaded part
+// Path Multipart path in repository
+// CreationDate Creation date of the part
+// PhysicalAddress Physical address of the part in the storage
+// Metadata Additional metadata as required (by storage vendor etc.)
+// ContentType Original file's content-type
 type MultipartUpload struct {
 	UploadID        string    `db:"upload_id"`
 	Path            string    `db:"path"`
```
itaiad200, line 29: We usually document fields inline

```suggestion
// MultipartUpload represents an ongoing multipart upload to the blockstore
type MultipartUpload struct {
	// UploadID A unique identifier for the uploaded part
	UploadID        string    `db:"upload_id"`
	
	// Path Multipart path in repository
	Path            string    `db:"path"`
	
	...
```

pkg/gateway/multiparts/tracker_test.go
N-o-Z: Because if we got an error then got == nil and this line will fail

pkg/gateway/multiparts/main_test.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"github.com/ory/dockertest/v3"
 
+	_ "github.com/treeverse/lakefs/pkg/kv/mem"
```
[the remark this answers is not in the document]
N-o-Z: I like the approach where unit tests use 'mem' and system tests use postgres. I think the only reason we used postgres in our unit tests is because we didn't have any other option.
We can take advantage of the new mem implementation and also save a lot of runtime. WDYT?

pkg/gateway/multiparts/tracker.go
```diff
@@ -22,14 +26,36 @@ type MultipartUpload struct {
 	ContentType     string    `db:"content_type"`
 }
 
+func (m Metadata) Set(k, v string) {
```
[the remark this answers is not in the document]
N-o-Z: This is not new code - do you want me to modify it on this PR?

pkg/gateway/multiparts/tracker.go
N-o-Z: This is required since the current multiparts behavior relies on getting an error from delete in case the entry doesn't exist. I can change that but that will lead to behavior change

pkg/kv/store_message.go
N-o-Z: Yes sorry - this is a remark for implementing the Scan method - updated

pkg/gateway/playback_test.go
N-o-Z: Checked with @guy-har - the tests are not relevant anymore - removing it as part of this PR. (closes #2031)

pkg/gateway/playback_test.go
N-o-Z: Removed playback test

pkg/gateway/multiparts/tracker.go
```diff
@@ -38,82 +72,70 @@ var (
...
 
-func NewTracker(adb db.Database) Tracker {
+func NewTracker(ms kv.StoreMessage) Tracker {
 	return &tracker{
-		db: adb,
+		store: ms,
+	}
+}
+
+func multipartFromProto(pb *MultipartUploadData) *MultipartUpload {
```
nopcoder, line 81: This is great - nothing actionable - just a thought that maybe in the next iteration we can also support as part of the model's method. Similar to how the current DB can query for 'Scan' method to convert the structure to JSON. We can do the same to get from/to proto format.

nopcoder (approved):
Looks good thanks!, minor comments to consider.

pkg/gateway/main_test.go
```diff
@@ -0,0 +1,133 @@
...
+	"github.com/spf13/viper"
+	"github.com/treeverse/lakefs/pkg/block"
+	"github.com/treeverse/lakefs/pkg/catalog"
+	"github.com/treeverse/lakefs/pkg/config"
+	"github.com/treeverse/lakefs/pkg/gateway"
+	"github.com/treeverse/lakefs/pkg/gateway/multiparts"
+	"github.com/treeverse/lakefs/pkg/gateway/simulator"
+	"github.com/treeverse/lakefs/pkg/kv"
+	"github.com/treeverse/lakefs/pkg/kv/kvtest"
+	_ "github.com/treeverse/lakefs/pkg/kv/mem"
```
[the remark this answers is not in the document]
N-o-Z: See my previous response:

> I like the approach where unit tests use 'mem' and system tests use postgres. I think the only reason we used postgres in our unit tests is because we didn't have any other option.
We can take advantage of the new mem implementation and also save a lot of runtime. WDYT?

pkg/gateway/main_test.go
```diff
@@ -0,0 +1,133 @@
...
+	auth    simulator.GatewayAuthService
+	catalog catalog.Interface
+}
+
+const ReplayRepositoryName = "example"
+
+var (
+	pool         *dockertest.Pool
+	databaseURI  string
+	IdTranslator *testutil.UploadIDTranslator
```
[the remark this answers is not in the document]
N-o-Z: Used by getBasicHandler which is used by testPathWithTrailingSlash

pkg/gateway/main_test.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		log.Fatalf("Could not connect to Docker: %s", err)
+	}
+	databaseURI, closer = testutil.GetDBInstance(pool)
+	code := m.Run()
+	closer() // cleanup
+	os.Exit(code)
+}
+
+func setupTest(t *testing.T, method, target string, body io.Reader, kvEnabled bool) *http.Response {
```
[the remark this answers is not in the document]
N-o-Z: I'd rather not put a lot of thought into it as this is temporary code just until we complete the feature

pkg/gateway/multiparts/tracker.go
```diff
@@ -38,82 +72,70 @@ var (
...
-		return nil
+		return fmt.Errorf("failed on Get. Key (%s): %w", uploadID, err)
 	}
-	data, ok := src.([]byte)
-	if !ok {
-		return ErrInvalidMetadataSrcFormat
-	}
-	return json.Unmarshal(data, m)
+	path := multipartPrefix + kv.PathDelimiter + uploadID
+	return m.store.Delete(ctx, path)
```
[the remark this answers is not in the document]
N-o-Z: I think it's better to do it in the store_message level

pkg/gateway/multiparts/db_tracker.go
```diff
@@ -0,0 +1,69 @@
...
+		return ErrInvalidUploadID
+	}
+	_, err := m.db.Transact(ctx, func(tx db.Tx) (interface{}, error) {
+		res, err := tx.Exec(`DELETE FROM gateway_multiparts WHERE upload_id = $1`, uploadID)
+		if err != nil {
+			return nil, err
+		}
+		affected := res.RowsAffected()
+		if affected != 1 {
+			return nil, ErrMultipartUploadNotFound
```
itaiad200, line 64: Isn't the KV behave differently in this case?

pkg/gateway/multiparts/tracker.go
itaiad200: Nice catch, I found it confusing in the original implementation but I guess this is how S3 behaves 🤷

pkg/gateway/multiparts/tracker.go
```diff
@@ -38,82 +72,70 @@ var (
...
-	if src == nil {
-		return nil
+		return fmt.Errorf("failed on Get. Key (%s): %w", uploadID, err)
 	}
-	data, ok := src.([]byte)
-	if !ok {
-		return ErrInvalidMetadataSrcFormat
-	}
-	return json.Unmarshal(data, m)
+	path := multipartPrefix + kv.PathDelimiter + uploadID
```
itaiad200: Not part of this PR - I think that one of the utilities of the KV package should be concatenating paths. We messed up with it so many times with the block adapters..

pkg/gateway/multiparts/tracker.go
```diff
@@ -5,21 +5,55 @@ import (
...
+	"fmt"
 	"strings"
 	"time"
 
-	"github.com/treeverse/lakefs/pkg/db"
+	"github.com/treeverse/lakefs/pkg/kv"
+	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
+const multipartPrefix = "multipart"
```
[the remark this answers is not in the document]
N-o-Z: Implementing NewPath on kv

pkg/gateway/multiparts/tracker_test.go
itaiad200: Got it. The condition should actually be `(err == nil)` but the previous check makes it the same as `!tt.wantErr`. Consider changing to `(err == nil)` for clarity.

pkg/kv/store_message.go
```diff
@@ -0,0 +1,62 @@
...
+
+// StoreMessage protobuf generic implementation for kv.Store interface applicable for all data models
+type StoreMessage struct {
+	Store Store
+}
+
+func (s *StoreMessage) GetMsg(ctx context.Context, path string, msg protoreflect.ProtoMessage) error {
+	val, err := s.Store.Get(ctx, []byte(path))
+	if err != nil {
+		return fmt.Errorf("failed on Get. Path (%s): %w", path, err)
```
itaiad200: Here and elsewhere

I think that the way errors are wrapped here may lead to long error lines with repetitive information, imagine the following possible error:
```
failed on Get. Path (this/is/some/very/long/repetitive/path): failed on Get. Path (this/is/some/very/long/repetitive/path): failed on Get. Path (this/is/some/very/long/repetitive/path): postgres failed to connect. Path (this/is/some/very/long/repetitive/path): tcp: no such host
```
Now it's repetitive because there are several calls to Get with path (gateway, storeMessage, KVStore and probably others I missed.)

I suggest the following (I think it's the golang approach)
```suggestion
		return fmt.Errorf("getting from store: %w", err)
```

If the path is relevant for the error, it should be added at the very start of the chain where the path has the most meaning.

itaiad200 (approved):
Few minor comments 
LGTM

pkg/gateway/main_test.go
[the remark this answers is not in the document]
N-o-Z: This code won't be removed (checking with @guy-har if the test using it can be removed altogether) but the kvEnabled flag will no longer be relevant once we are in feature complete and we will only be using and testing one tracker implementation

pkg/gateway/main_test.go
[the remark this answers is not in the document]
N-o-Z: Since this is a unit test and we are testing the tracker functionality and not the KV functionality I believe the underlying KV implementation is irrelevant. In any case - the postgres implementation should be tested by it's own in a dedicated unit test and will be tested extensively as part of the system tests

pkg/gateway/multiparts/db_tracker.go
N-o-Z: It doesn't since message_store Delete I explicitly check if  the doesn't exist and return ErrMultipartUploadNotFound immediately in this case to simulate this behavior. This is related to @nopcoder's comment as well:

> This is required since the current multiparts implementation relies on getting an error from delete in case the entry doesn't exist. I can change that but that will lead to behavior change

pkg/gateway/multiparts/tracker.go
N-o-Z: Done

pkg/gateway/multiparts/tracker_test.go
N-o-Z: Done

arielshaqed:
Note that #3303 will soon remove almost all of the gateway test.  IIWY, I would minimize the changes there.  And I'll be happy to help resolve the coming conflicts in any way that I can.

N-o-Z:
> Note that #3303 will soon remove almost all of the gateway test. IIWY, I would minimize the changes there. And I'll be happy to help resolve the coming conflicts in any way that I can.

Perfect @arielshaqed thanks for the heads up. So I think I'll just wait for this ticket to merge and rebase my changes - since we still need to modify GetBasicHandler multipartsTracker initialization

N-o-Z:
Rebased over master after #3303 changes. Added NewPath function.

pkg/gateway/multiparts/tracker.go
nopcoder: I assume we will have to verify if the S3 protocol does it too - if it is not the case we will change the behviour as there is no way to ensure that by calling Get a second before Delete.

pkg/gateway/multiparts/tracker.go
```diff
@@ -38,82 +46,70 @@ var (
...
 
 func (m *tracker) Delete(ctx context.Context, uploadID string) error {
 	if uploadID == "" {
 		return ErrInvalidUploadID
 	}
-	_, err := m.db.Transact(ctx, func(tx db.Tx) (interface{}, error) {
-		res, err := tx.Exec(`DELETE FROM gateway_multiparts WHERE upload_id = $1`, uploadID)
-		if err != nil {
-			return nil, err
+	if _, err := m.Get(ctx, uploadID); err != nil {
```
nopcoder, line 107: I would not implement Delete in the tracker level, using Get in the tracker level - until there is another lower implementation of 'get'.
This is because usually the logging, BI, etc will be implemented at that level.
In this above case - I'll use the storage Get and not the upper level one.

pkg/kv/store.go
```diff
@@ -20,6 +23,14 @@ var (
 	ErrUnknownDriver       = errors.New("unknown driver")
 )
 
+func NewPath(p ...string) string {
+	const minPathLength = 2
+	if len(p) < minPathLength {
+		panic("invalid path length")
+	}
```
nopcoder, line 30: Why? keep it simple and just enable us to build a single word path - which will not include any sep.

nopcoder (changes requested):

pkg/gateway/multiparts/tracker.go
```diff
@@ -38,82 +46,70 @@ var (
...
-			VALUES ($1, $2, $3, $4, $5, $6)`,
-			multipart.UploadID, multipart.Path, multipart.CreationDate, multipart.PhysicalAddress, multipart.Metadata, multipart.ContentType)
-		return nil, err
-	})
+	_, err := m.Get(ctx, multipart.UploadID)
+	if err == nil {
+		return ErrAlreadyExists
+	}
+	path := kv.NewPath(multipartPrefix, multipart.UploadID)
+	err = m.store.SetMsg(ctx, path, protoFromMultipart(&multipart))
```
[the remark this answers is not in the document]
N-o-Z: @nopcoder - If we use SetIf in this instance - we will get an ambiguous error on ErrPredicateFailed which we will need to interpret as ErrAlreadyExists. We will also not save on any API calls since SetIf uses Get as well

pkg/kv/store_message.go
```diff
@@ -37,14 +37,14 @@ func (s *StoreMessage) SetIf(ctx context.Context, path string, msg protoreflect.
...
-		return fmt.Errorf("failed on Unmarshal (path: %s) (curr msg: %s): %w", path, currMsg, err)
+		return fmt.Errorf("failed on Unmarshal (path: %s) %w", path, err)
 	}
 	if !proto.Equal(pred, currMsg) {
 		return fmt.Errorf("failed on predicate. Path (%s): %w", path, ErrPredicateFailed)
 	}
 	val, err := proto.Marshal(msg)
 	if err != nil {
-		return fmt.Errorf("failed on Marshal (path: %s) (msg: %s): %w", path, msg, err)
+		return fmt.Errorf("failed on Marshal (path: %s): %w", msg, err)
```
nopcoder, line 47: path

pkg/gateway/multiparts/tracker.go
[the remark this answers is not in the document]
nopcoder: Nice - maybe a different name for the function FormatPath or something like that. The name `NewPath` suggests it will return a type Path.

pkg/gateway/multiparts/tracker.go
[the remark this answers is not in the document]
nopcoder: Understood about the error type migration - but it is in place to handle the 'transaction' part of setting a new value, just in case it doesn't exist.

nopcoder (changes requested):

pkg/gateway/multiparts/tracker.go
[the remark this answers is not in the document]
N-o-Z: I've modified the SetIf function in store_message. Please have another look

nopcoder (approved):
LGTM!

PR #3300, opened 2022-05-01
Fix: Failing HTTP request hook should be logged

### LINKED ISSUE #3099: Failing HTTP request hook should be logged
When a webhook fails before the http request was sent (e.g "no such host"), the failed action log doesn't display the error.
It does appear in the returned error to the client.

N-o-Z:
Closes #3099

### DIFF
--- a/pkg/actions/service.go
+++ b/pkg/actions/service.go
@@ -211,7 +211,8 @@ func (s *Service) runTasks(ctx context.Context, record graveler.HookRecord, task
 
 				s.stats.CollectEvent("actions_service", string(record.EventType))
 
-				if task.Err != nil { // log error in buffer
+				if task.Err != nil { 
+				// log error in buffer
 					// wrap error with more information and return
 					task.Err = fmt.Errorf("hook run id '%s' failed on action '%s' hook '%s': %w",
 						task.HookRunID, task.Action.Name, task.HookID, task.Err)
--- a/pkg/actions/service.go
+++ b/pkg/actions/service.go
@@ -211,19 +211,20 @@ func (s *Service) runTasks(ctx context.Context, record graveler.HookRecord, task
 
 				s.stats.CollectEvent("actions_service", string(record.EventType))
 
-				if task.Err != nil { 
-				// log error in buffer
-					// wrap error with more information and return
+				if task.Err != nil {
+					_, _ = fmt.Fprintf(&buf, "Error: %s\n", task.Err)
+					// wrap error with more information
 					task.Err = fmt.Errorf("hook run id '%s' failed on action '%s' hook '%s': %w",
 						task.HookRunID, task.Action.Name, task.HookID, task.Err)
-					_, _ = fmt.Fprintf(&buf, "%s\n", task.Err)
 				}
+
 				err := hookOutputWriter.OutputWrite(ctx, &buf, int64(buf.Len()))
 				if err != nil {
-					return fmt.Errorf("failed to write action log. %w", err)
+					return fmt.Errorf("failed to write action log. Run id '%s' action '%s' hook '%s': %w",
+						task.HookRunID, task.Action.Name, task.HookID, err)
 				}
-				if task.Err != nil { 
-				// stop execution of tasks and return error
+				if task.Err != nil {
+					// stop execution of tasks and return error
 					return task.Err
 				}
 			}
--- a/pkg/actions/service.go
+++ b/pkg/actions/service.go
@@ -203,16 +203,25 @@ func (s *Service) runTasks(ctx context.Context, record graveler.HookRecord, task
 					ActionName:       task.Action.Name,
 					HookID:           task.HookID,
 				}
+				buf := bytes.Buffer{}
 				task.StartTime = time.Now().UTC()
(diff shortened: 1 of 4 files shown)


N-o-Z:
Before:
image

After:
image

pkg/actions/webhook.go
```diff
@@ -147,6 +147,7 @@ func executeAndLogResponse(ctx context.Context, req *http.Request, buf *bytes.Bu
 	elapsed := time.Since(start)
 	buf.WriteString(fmt.Sprintf("\nRequest duration: %s\n", elapsed))
 	if err != nil {
+		buf.WriteString(fmt.Sprintf("\nError: %s\n", err))
```
itaiad200, line 150: What happens if it fail on line 85?

N-o-Z: In that case we didn't even try to execute the URL.
I suggest creating a new method "logRunError" which will just write error to buffer and close it.
In that case we might want to map all possible error types and format them properly

nopcoder: - Think we can long the `err` of the failed task at the `service.go` (runTasks) level.
- We can use `fmt.Fprintf(buf, "\nError: %s\n", err)` instead of the combination of `Sprintf` and `WriteString`

nopcoder (changes requested):

pkg/actions/service.go
```diff
@@ -203,16 +203,25 @@ func (s *Service) runTasks(ctx context.Context, record graveler.HookRecord, task
...
 				task.StartTime = time.Now().UTC()
-				task.Err = task.Hook.Run(ctx, record, hookOutputWriter)
+
+				task.Err = task.Hook.Run(ctx, record, &buf)
 				task.EndTime = time.Now().UTC()
 
 				s.stats.CollectEvent("actions_service", string(record.EventType))
 
-				if task.Err != nil {
+				if task.Err != nil { // log error in buffer
```
itaiad200, line 214:
```suggestion
				if task.Err != nil { 
				// log error in buffer
```

pkg/actions/service.go
```diff
@@ -203,16 +203,25 @@ func (s *Service) runTasks(ctx context.Context, record graveler.HookRecord, task
...
 					// wrap error with more information and return
 					task.Err = fmt.Errorf("hook run id '%s' failed on action '%s' hook '%s': %w",
 						task.HookRunID, task.Action.Name, task.HookID, task.Err)
+					_, _ = fmt.Fprint(&buf, task.Err)
+				}
+				err := hookOutputWriter.OutputWrite(ctx, &buf, int64(buf.Len()))
+				if err != nil {
+					return fmt.Errorf("failed to write action log. %w", err)
+				}
+				if task.Err != nil { // stop execution of tasks and return error
```
itaiad200, line 224:
```suggestion
				if task.Err != nil { 
				// stop execution of tasks and return error
```

pkg/actions/service.go
```diff
@@ -203,16 +203,25 @@ func (s *Service) runTasks(ctx context.Context, record graveler.HookRecord, task
...
-				if task.Err != nil {
+				if task.Err != nil { // log error in buffer
 					// wrap error with more information and return
 					task.Err = fmt.Errorf("hook run id '%s' failed on action '%s' hook '%s': %w",
 						task.HookRunID, task.Action.Name, task.HookID, task.Err)
+					_, _ = fmt.Fprint(&buf, task.Err)
+				}
+				err := hookOutputWriter.OutputWrite(ctx, &buf, int64(buf.Len()))
+				if err != nil {
+					return fmt.Errorf("failed to write action log. %w", err)
```
itaiad200, line 222:
```suggestion
					return fmt.Errorf("failed to write action log: %w", err)
```

itaiad200 (approved):
LGTM

pkg/actions/service.go
```diff
@@ -203,16 +203,25 @@ func (s *Service) runTasks(ctx context.Context, record graveler.HookRecord, task
...
 				task.EndTime = time.Now().UTC()
 
 				s.stats.CollectEvent("actions_service", string(record.EventType))
 
-				if task.Err != nil {
+				if task.Err != nil { // log error in buffer
 					// wrap error with more information and return
 					task.Err = fmt.Errorf("hook run id '%s' failed on action '%s' hook '%s': %w",
 						task.HookRunID, task.Action.Name, task.HookID, task.Err)
+					_, _ = fmt.Fprint(&buf, task.Err)
```
nopcoder, line 218: - Optional: long of the original error is extra information we format is for our logs as far as I remember.
- Just in case we like to print something after, and we will need a new-line at the end

```suggestion
					_, _ = fmt.Fprintf(&buf, "%s\n", task.Err)
```

pkg/actions/service.go
nopcoder: I think we will also need to format - task.HookRunID, task.Action.Name, task.HookID
For additional information

pkg/actions/service.go
nopcoder: Don't need extra comments ;)

nopcoder (approved):
LGTM - minor comments

PR #3301, opened 2022-05-02
lakeFS on KV store - testing plan - first draft

itai-david:
First draft with analysis of how we need to test the move to KV Store with a 1st MS plan
Contains analysis of current test coverage for relevant packages
Next phase - design of the testing infrastructure itself

### DIFF
--- a/design/open/lakefs-on-kv-testing-plan.md
+++ b/design/open/lakefs-on-kv-testing-plan.md
@@ -1,0 +1,128 @@
+# lakeFS on KV - Testing Plan
+
+## Global Testing
+
+### Migration
+* Global tests to verify and benchmark the migration process
+* Need to verify that the system behavior is not affected by the move to the KV DB
+* [Question] - Do we need to dispose of these once migration to KV is globally completed?
+
+#### Operational Testing
+* Set of tests that will run DB migration as part of the test
+* Currently, system tests assume lakeFS is up and running. Need to add ability to perform migration in the middle of the test (stop lakeFS -> run&verify migration -> start lakeFS ?)
+* Maybe something that can be 
+
+#### Data Level Testing
+* Dump (all) data before and after migration and compare it
+* Verifies data is preserved through migration, and is readable by the KV Store
+* Need to write appropriate dumpers. Can be done per package, as we go
+* Can be done both on DB package level, to verify the package itself, and on the using packages level, to verify data usage is not affected
+* [Question] - How do we define a covering data set, to test? How do we create
+
+### Performance
+* **A benchmark for the migration process itself:**
+  * How do we load the DB with data to migrate for benchmarking?
+  * Ideally, run on several data sizes (tens / thousands / millions branches & commits, various numbers of users, etc.) to get sense of how scaling will affect
+  * What is the expected/acceptable outcome?
+  * Should be built gradually, as we support and migrate additional package
+
+* **Benchmark comparison of both DBs**
+  * Compare performance before and after migration - does not include the migration itself
+  * Define package specific set of operation to benchmark
+  * Run with both Table DB and KV Store and compare the results
+  * What is the accepted degradation, if any?
+  * Need to consider various scales
+
+### Feature Flag
+* Dummy tests to verify the feature flag works as expected
+* Package level panic in case the wrong DB logic is used? Is that sufficient?
+
+## Per Package DB Testing
+
+### ```pkg/gateway/multiparts```
+* DB is used to track the start and end of a multipart upload. All DB accesses are done via `mutltiparts.Tracker`. Entries are created once, read-accesses multiple times and deleted upon completion
+* Currently unit tests cover correctness of DB accesses in both good and error paths.
+* 83.8% coverage
+* Currently system tests cover a simle multipart upload - a single object with 7 parts, good path only
+* What is missing:
+  * Performance - Can that sustain heavy loads: multi-multipart uploads with a lot of parts
+  * Concurrency, that is derived from the previous bullet (can we increase concurrency by using smaller parts?)
+  * Migration - Verify that data migrated from Table to KV, in the middle of multipart upload, is still usable
+
+
+### ```pkg/actions```
+* DB is used to store actions runs and hook runs with results. Db is read by the actions service to handle requests from the various clients. For each run there is a new entry in `actions_run` table and an entry for each hook in `actions_run_hooks` table.
+Updates are done for commitID, as part of `post` hooks
+* Currently unit tests cover hooks run and verify the results in DB using the service reading functions. There are also error testings (hook failures)
+* 67.6% coverage
+* What is missing:
+  * DB Failure during hook action (TBD - what is the expected behavior? Is that interesting to test?)
+  * Performance under load - not sure a load on actions/hooks is relevant
+  * Concurrent actions/hooks execution, which derives concurrent DB accesses. This gets more interesting if load is involved (in case load is relevant)
+
+### ```pkg/auth```
+* DB is used to store installation metadata, users and policies
+* Only 6.8% coverage
+* Some benchmarks exists that covers effectivePolicies (multi-table JOIN)
+* What is missing:
+  * Need to increase code coverage
+  * Comparison benchmark to toggle between Tables and KV and verify there is no degradation
+  * Migration tests:
+    * verify that data is migrated correctly
+    * Authentication transactions in during migration (data that was written to tables is read from KV)
+
+### ```pkg/diagnostics```
+* Only read access
+* No test coverage. Need to verify all is working the same after migration
+
+### ```pkg/graveler/ref```
+* Branch locking (read)
+* Repos, branches, commits and tags - read/write
+* 95.5% testing coverage
+* No performance/benchmarking
+* Missing:
+  * Migration tests
+  * Comparison benchmark
+
+### ```pkg/graveler/retention```
+* No direct DB access. Accesses are done using ```pkg/graveler/ref```, which should make the DB transition seamless
+* 42.9% coverage in unit testing
+* Missing performance tests that might help detect degradation
+
+### ```pkg/graveler/staging```
+* 89.1% coverage
+* Covers all staged KV functionality
+* No performance/benchmarking
+* Missing:
+  * Migration tests
+  * Comparison benchmark
+
+### Res-Dump/Restore
+Currently this functionality is not covered directly, but since it relies on '''pkg/graveler/ref''' for DB access, the DB migration should be seamless
+It can be leveraged, however, to extend the cover of '''pkg/graveler/ref''' and for performance, as it is quite exhaustive (traverses all branches and commits and tags, per repo)
+
+## Execution Plan
+
+### KVM1
+* Data level migration tests infrastructure
+  * Migrate data from Table to KV, dumps both and compares
+  * Implement dumpers for `gateway_multiprts`
+  * Implement tests for '''pkg/multiparts'''
+* Infrastructure for running migration during a system test execution
+  * System test to run migration during multiparts upload
+  * Currently there is a single simple multiparts system test (single file, 7 parts) - this is also an opportunity to expand that
+* KV Store unit tests 
+* `multiparts.Tracker` benchmark to run on both Table DB and KV Store (use feature flag to toggle) and verify there is no degradation
+  * Define sequence(s) of actions to perform (Create/Get/Delete etc.)
+  * Run each sequence with feature flag off and on
+  * Compare results and fail if KV performance is more than [TBD]% slower
+  * Need an infrastructure 
+
+### Next MS
+TBD
+
+
+## Open questions
+* What performance degradation is acceptable? (Can be 0)
+* How do we simulate KV Store failure? Do we need it at all?
+* What is the expected downtime for migrate?
--- a/design/open/lakefs-on-kv-testing-plan.md
+++ b/design/open/lakefs-on-kv-testing-plan.md
@@ -39,7 +39,7 @@
 * DB is used to track the start and end of a multipart upload. All DB accesses are done via `mutltiparts.Tracker`. Entries are created once, read-accesses multiple times and deleted upon completion
 * Currently unit tests cover correctness of DB accesses in both good and error paths.
 * 83.8% coverage
-* Currently system tests cover a simle multipart upload - a single object with 7 parts, good path only
+* Currently system tests cover a simple multipart upload - a single object with 7 parts, good path only
 * What is missing:
   * Performance - Can that sustain heavy loads: multi-multipart uploads with a lot of parts
   * Concurrency, that is derived from the previous bullet (can we increase concurrency by using smaller parts?)
--- a/design/open/lakefs-on-kv-testing-plan.md
+++ b/design/open/lakefs-on-kv-testing-plan.md
@@ -6,14 +6,15 @@
 * Global tests to verify and benchmark the migration process
 * Need to verify that the system behavior is not affected by the move to the KV DB
 * [Question] - Do we need to dispose of these once migration to KV is globally completed?
+  * Yes. Migration code (and related tests) should be removed from the code, once the entire deployment of KV Store is done
 
 #### Operational Testing
 * Set of tests that will run DB migration as part of the test
 * Currently, system tests assume lakeFS is up and running. Need to add ability to perform migration in the middle of the test (stop lakeFS -> run&verify migration -> start lakeFS ?)
-* Maybe something that can be 
 
 #### Data Level Testing
-* Dump (all) data before and after migration and compare it
+**Note:** This section is on a "Nice to Have" basis, as it is pretty much covered by the previous section
+* ~~Dump~~ Extract (all) data before and after migration and compare it
 * Verifies data is preserved through migration, and is readable by the KV Store
 * Need to write appropriate dumpers. Can be done per package, as we go
 * Can be done both on DB package level, to verify the package itself, and on the using packages level, to verify data usage is not affected
@@ -32,11 +33,6 @@
   * Run with both Table DB and KV Store and compare the results
   * What is the accepted degradation, if any?
   * Need to consider various scales
-
-### Feature Flag
-* Dummy tests to verify the feature flag works as expected
-* Package level panic in case the wrong DB logic is used? Is that sufficient?
-
 ## Per Package DB Testing
 
 ### ```pkg/gateway/multiparts```
@@ -97,7 +93,7 @@ Updates are done for commitID, as part of `post` hooks
   * Migration tests
   * Comparison benchmark
 
-### Res-Dump/Restore
+### Ref-Dump/Restore
 Currently this functionality is not covered directly, but since it relies on '''pkg/graveler/ref''' for DB access, the DB migration should be seamless
 It can be leveraged, however, to extend the cover of '''pkg/graveler/ref''' and for performance, as it is quite exhaustive (traverses all branches and commits and tags, per repo)
 
@@ -105,9 +101,18 @@ It can be leveraged, however, to extend the cover of '''pkg/graveler/ref''' and 
 
 ### KVM1
 * Data level migration tests infrastructure
-  * Migrate data from Table to KV, dumps both and compares
-  * Implement dumpers for `gateway_multiprts`
-  * Implement tests for '''pkg/multiparts'''
+  * Migrate data from Table to KV, extract both and compare
+    * Data extraction should be done by listing all objects in the DB, using a designated 'get' function, and compare
+  * Implement ~~dumpers for `gateway_multiprts`~~ `GetAll` for `multiparts.Tracker`, to return a list of `MultipartUpload`
+  * Implement comparison of `MultipartUploads` list. Lists are considered identical if objects are identical, but **not necessarily** at the same order
+  * Implement unit tests for `pkg/gateway/multiparts`
+    * Add multipart uploads using `multiparts.Tracker.Create` with Table DB (KV Feature Flag off)
+      * Create an entry with key representing each off the supported storages:
+        * azure, google, s3, local, mem & transient
+      * Read all entries using `GetAll` above (Table DB)
+      * Run migration for `gateway_multiparts` table
+      * Read all entries using `GetAll` (KV Store)
+      * Compare the lists and expect equality (up to order)
 * Infrastructure for running migration during a system test execution
   * System test to run migration during multiparts upload
   * Currently there is a single simple multiparts system test (single file, 7 parts) - this is also an opportunity to expand that
--- a/design/open/lakefs-on-kv-testing-plan.md
+++ b/design/open/lakefs-on-kv-testing-plan.md
@@ -1,5 +1,9 @@
 # lakeFS on KV - Testing Plan
-
+This document discusses the tests and testing methods that will be applied to support our development and transition to [lakeFS on KV](https://github.com/treeverse/lakeFS/blob/ed55edc2c24b24ed4e6fcccad1ae95844d38b9ad/design/open/metadata_kv/index.md). It is derived and aligned with [lakeFS on KV - Execution Plan](https://github.com/treeverse/lakeFS/blob/master/design/open/lakefs-kv-execution-plan.md) doc. </br>
+This document aims to describe the requirements for the testing infrastructure and the tests to be conducted. It refers both to system tests and package specific unit tests. </br>
+## Important Notes
+* Part of the tests discussed, refer to the migration process from Table DB to KV Store. These tests are valid for the development and transition phases and will not be required once `lakeFS` is completely migrated to KV Store
+* Implied from the previous comment, the migration process itself is not here to stay. The current plan is to have a transition version (V.mig) which will be the last version to support migration from Table DB to KV Store, and no migration such will be supported in later version. An update from Table DB version (V.table) prior to Vt, to KV version (V.kv) later than Vt, will have to go through Vt (i.e. 2 step update): V.table -> V.mig -> V.kv
 ## Global Testing
 
 ### Migration
@@ -22,6 +26,7 @@
 
 ### Performance
 * **A benchmark for the migration process itself:**
+  * An `Esti` system test for system-wide migration from Table DB to KV. This tests should be a WIP and will support a growing set of data to be migrated, as KV support is growing
   * How do we load the DB with data to migrate for benchmarking?
   * Ideally, run on several data sizes (tens / thousands / millions branches & commits, various numbers of users, etc.) to get sense of how scaling will affect
   * What is the expected/acceptable outcome?
@@ -130,4 +135,5 @@ TBD
 ## Open questions
 * What performance degradation is acceptable? (Can be 0)
 * How do we simulate KV Store failure? Do we need it at all?
+  * [per @N-o-Z] We can create a mock for either Store or StoreMessage to simulate failures. I would describe what kind of fault injection do we need to support to better understand the faults requirements
 * What is the expected downtime for migrate?


design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,128 @@
...
+## Execution Plan
+
+### KVM1
+* Data level migration tests infrastructure
+  * Migrate data from Table to KV, dumps both and compares
+  * Implement dumpers for `gateway_multiprts`
+  * Implement tests for '''pkg/multiparts'''
+* Infrastructure for running migration during a system test execution
+  * System test to run migration during multiparts upload
+  * Currently there is a single simple multiparts system test (single file, 7 parts) - this is also an opportunity to expand that
```
itaiad200, line 113: What would you expand here and why?

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Res-Dump/Restore
+Currently this functionality is not covered directly, but since it relies on '''pkg/graveler/ref''' for DB access, the DB migration should be seamless
+It can be leveraged, however, to extend the cover of '''pkg/graveler/ref''' and for performance, as it is quite exhaustive (traverses all branches and commits and tags, per repo)
+
+## Execution Plan
+
+### KVM1
+* Data level migration tests infrastructure
+  * Migrate data from Table to KV, dumps both and compares
```
itaiad200, line 108: Why compare the dumps and not use the implementations themselves? Like listing the two and comparing.

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,128 @@
...
+  * Implement dumpers for `gateway_multiprts`
+  * Implement tests for '''pkg/multiparts'''
+* Infrastructure for running migration during a system test execution
+  * System test to run migration during multiparts upload
+  * Currently there is a single simple multiparts system test (single file, 7 parts) - this is also an opportunity to expand that
+* KV Store unit tests 
+* `multiparts.Tracker` benchmark to run on both Table DB and KV Store (use feature flag to toggle) and verify there is no degradation
```
itaiad200, line 115: Should we run it once locally or as part of nessie/something else?

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,128 @@
...
+### Res-Dump/Restore
+Currently this functionality is not covered directly, but since it relies on '''pkg/graveler/ref''' for DB access, the DB migration should be seamless
+It can be leveraged, however, to extend the cover of '''pkg/graveler/ref''' and for performance, as it is quite exhaustive (traverses all branches and commits and tags, per repo)
+
+## Execution Plan
+
+### KVM1
+* Data level migration tests infrastructure
+  * Migrate data from Table to KV, dumps both and compares
+  * Implement dumpers for `gateway_multiprts`
```
itaiad200, line 109:
```suggestion
  * Implement dumpers for `gateway_multiparts`
```

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,128 @@
...
+  * Implement tests for '''pkg/multiparts'''
```
itaiad200, line 110: Please be more specific.

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,128 @@
+# lakeFS on KV - Testing Plan
+
+## Global Testing
+
+### Migration
+* Global tests to verify and benchmark the migration process
+* Need to verify that the system behavior is not affected by the move to the KV DB
+* [Question] - Do we need to dispose of these once migration to KV is globally completed?
```
itaiad200, line 8: I think that migration code would be eventually deleted and so would the tests

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,128 @@
...
+
+#### Operational Testing
+* Set of tests that will run DB migration as part of the test
```
itaiad200, line 11: How is this different from the previous section?

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,128 @@
...
+* Currently, system tests assume lakeFS is up and running. Need to add ability to perform migration in the middle of the test (stop lakeFS -> run&verify migration -> start lakeFS ?)
+* Maybe something that can be 
+
+#### Data Level Testing
+* Dump (all) data before and after migration and compare it
+* Verifies data is preserved through migration, and is readable by the KV Store
+* Need to write appropriate dumpers. Can be done per package, as we go
+* Can be done both on DB package level, to verify the package itself, and on the using packages level, to verify data usage is not affected
+* [Question] - How do we define a covering data set, to test? How do we create
```
itaiad200, line 20: This covers migration again I think.

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Performance
+* **A benchmark for the migration process itself:**
+  * How do we load the DB with data to migrate for benchmarking?
+  * Ideally, run on several data sizes (tens / thousands / millions branches & commits, various numbers of users, etc.) to get sense of how scaling will affect
+  * What is the expected/acceptable outcome?
```
itaiad200, line 26: When and how do we run it?

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,128 @@
...
+  * Compare performance before and after migration - does not include the migration itself
+  * Define package specific set of operation to benchmark
+  * Run with both Table DB and KV Store and compare the results
+  * What is the accepted degradation, if any?
+  * Need to consider various scales
+
+### Feature Flag
+* Dummy tests to verify the feature flag works as expected
+* Package level panic in case the wrong DB logic is used? Is that sufficient?
+
```
itaiad200, line 39: I don't think there's much to test here. All previous tests rely on this flag to work.. I can't really see how they pass when the flag doesn't do its job.

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,128 @@
...
+## Per Package DB Testing
+
+### ```pkg/gateway/multiparts```
+* DB is used to track the start and end of a multipart upload. All DB accesses are done via `mutltiparts.Tracker`. Entries are created once, read-accesses multiple times and deleted upon completion
+* Currently unit tests cover correctness of DB accesses in both good and error paths.
+* 83.8% coverage
+* Currently system tests cover a simle multipart upload - a single object with 7 parts, good path only
```
itaiad200, line 46:
```suggestion
* Currently system tests cover a simple multipart upload - a single object with 7 parts, good path only
```

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,128 @@
...
+  * Define package specific set of operation to benchmark
+  * Run with both Table DB and KV Store and compare the results
+  * What is the accepted degradation, if any?
+  * Need to consider various scales
+
+### Feature Flag
+* Dummy tests to verify the feature flag works as expected
+* Package level panic in case the wrong DB logic is used? Is that sufficient?
+
+## Per Package DB Testing
```
itaiad200, line 40: Thanks for this very detailed analysis!

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,128 @@
...
+  * DB Failure during hook action (TBD - what is the expected behavior? Is that interesting to test?)
+  * Performance under load - not sure a load on actions/hooks is relevant
+  * Concurrent actions/hooks execution, which derives concurrent DB accesses. This gets more interesting if load is involved (in case load is relevant)
+
+### ```pkg/auth```
+* DB is used to store installation metadata, users and policies
+* Only 6.8% coverage
+* Some benchmarks exists that covers effectivePolicies (multi-table JOIN)
+* What is missing:
+  * Need to increase code coverage
```
itaiad200, line 68: I don't think this should be a part of the KV effort.

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,128 @@
...
+  * Comparison benchmark to toggle between Tables and KV and verify there is no degradation
+  * Migration tests:
+    * verify that data is migrated correctly
+    * Authentication transactions in during migration (data that was written to tables is read from KV)
+
+### ```pkg/diagnostics```
+* Only read access
+* No test coverage. Need to verify all is working the same after migration
```
itaiad200, line 76: So manual testing?

itaiad200 (changes requested):
Thanks for the detailed plan. I think the execution plan for the KVM1 milestone should be more "crisp" (i.e no followup discovery tasks after this document is approved.)

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,128 @@
...
+
+### ```pkg/graveler/ref```
+* Branch locking (read)
```
eden-ohana, line 79: We can also test consistency with commit racing

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,128 @@
...
+  * Migrate data from Table to KV, dumps both and compares
+  * Implement dumpers for `gateway_multiprts`
+  * Implement tests for '''pkg/multiparts'''
+* Infrastructure for running migration during a system test execution
+  * System test to run migration during multiparts upload
+  * Currently there is a single simple multiparts system test (single file, 7 parts) - this is also an opportunity to expand that
+* KV Store unit tests 
+* `multiparts.Tracker` benchmark to run on both Table DB and KV Store (use feature flag to toggle) and verify there is no degradation
+  * Define sequence(s) of actions to perform (Create/Get/Delete etc.)
+  * Run each sequence with feature flag off and on
```
eden-ohana, line 117: Does that mean that we will need to adjust the current tests to the new implantation? so we can run the current test on it.

eden-ohana (commented):
What about the IAM tests?

design/open/lakefs-on-kv-testing-plan.md
itai-david: The current test verifies a very specific simple good path.
I thought of adding, at least, a test that uploads multiple files with multiple parts each (e.g. 100 files with 1000 parts each). This will insert a concurrency aspect to the tests, which is currently lacks
I also thought of shuffeling the parts order, rather than upload them ordered sequentially, just for the fun of it
Another aspect, I miss in the current test, is errors and retries of uploads, so this, too can be added.
This is orthogonal to our KV needs and can (and IMHO should) be added' regardless. It will, however, help with achieving higher confidence in our KV support for the multiplrt pkg
WDYT?

design/open/lakefs-on-kv-testing-plan.md
itai-david: I thought of a dump as this could later be a more uniform way to compare various data types - compatring 'text'.
Truth is, that there is no real benefit for that, so we might as well go with the implementation. In fact, it will save the need to enforce implementation of a `dump` function in the future

design/open/lakefs-on-kv-testing-plan.md
itai-david: This is intended to be a unit-test, so outside of `Esti` context. It could be used locally for developement/debugging purposes, but obviously, I would like it to run as part of our '''Go Tests''' and fail if an unacceptable degradation is detected

design/open/lakefs-on-kv-testing-plan.md
itai-david: The previous section is about looking at the DB before and after the migration, and comparing its context.
This section is about having certain operations running (e.g. several multipart uploads), then, as part of the test, a migration to KV takes place (probably need to stop lakeFS, migrate, verify it was successful (?) and start lakeFS), then continue the operations where we left (in that case, resume some and some complete multipart uploads, as well as starting new ones) and verify it works as expected.
I agree that it seems quite the same, as the mutlipart uploads DB usage is very basic, but I believe that once we get into modules that use the DB in a more complexed way (e.g. `pkg/auth`) it will be beneficial.
This step is about building the infrastructure for such test and not necessarily about the `multipart uploads` aspect

design/open/lakefs-on-kv-testing-plan.md
itai-david: Agreed. Removed

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,128 @@
...
+
+### Migration
+* Global tests to verify and benchmark the migration process
+* Need to verify that the system behavior is not affected by the move to the KV DB
+* [Question] - Do we need to dispose of these once migration to KV is globally completed?
+
+#### Operational Testing
+* Set of tests that will run DB migration as part of the test
+* Currently, system tests assume lakeFS is up and running. Need to add ability to perform migration in the middle of the test (stop lakeFS -> run&verify migration -> start lakeFS ?)
+* Maybe something that can be
```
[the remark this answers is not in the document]
itai-david: Me too :rofl: 
Incomplete thought leftover. Removed

design/open/lakefs-on-kv-testing-plan.md
itai-david: Updated

design/open/lakefs-on-kv-testing-plan.md
itai-david: Entire line delete as `dump` idea is dismissed

design/open/lakefs-on-kv-testing-plan.md
itai-david: Done

design/open/lakefs-on-kv-testing-plan.md
itai-david: Sounds fair. Added a note

design/open/lakefs-on-kv-testing-plan.md
itai-david: Correct. This is a section under the **Migration** headline. There are 2 sections: (1) Operational Tests - running the migration as part of a work flow, to verify the migration completes successfully and (probably) that  the system is stil operable, and (2) Data Level Tests - comparison of the data before and after the migration, for equality.
As I'm writing these, it seems that the first section covers the second too. Yet, I believe it will make it easier to test the migration code (at least for debugging purposes) an I think it should stay, at lease as a "Nice to Have".
Will add a note

design/open/lakefs-on-kv-testing-plan.md
itai-david: A dedicated system-wide migration tests, as part of `Esti` seems like the proper place
WDYT?

design/open/lakefs-on-kv-testing-plan.md
itai-david: You are right on this one. However, I would argue that a higher coverage will increase our confidence in the package all together, and with that in the KV support (for the `auth` package.
How about extracting it to a separate, non KV relate, issue? WDYT?

design/open/lakefs-on-kv-testing-plan.md
itai-david: I guess. Couldn't find anything in the code
Let's cover that at least from KV aspect. WDYT?

design/open/lakefs-on-kv-testing-plan.md
itai-david: Can you please specify regarding that?
How do we create this race?
Thanks

design/open/lakefs-on-kv-testing-plan.md
itai-david: I thought in the direction of defining a set of operations, for benchmarking, then run it twice: once with KV feature disabled and once with KV enabled, in order to compare the 2 results.
At least at the moment, we do not have such benchmark test(s) so it is more of writing them to support both modes, rather than modify
Does this make sense?

itai-david:
> What about the IAM tests?

Excellent catch. Thanks for that @eden-ohana
I toatlly missed it. Will need to add this

@itaiad200 @nopcoder - I think that can be done outside of KV-M1 context, WDYT?

design/open/lakefs-on-kv-testing-plan.md
itai-david: Thanks!

design/open/lakefs-on-kv-testing-plan.md
itaiad200: 👍 
Let's add it then

design/open/lakefs-on-kv-testing-plan.md
itaiad200: Fine with that - be aware that `auth` is Cloud-Native's domain.

design/open/lakefs-on-kv-testing-plan.md
itaiad200: Manual it is.

design/open/lakefs-on-kv-testing-plan.md
itaiad200: > if an unacceptable degradation is detected

How? It depends on your env where you run the test, no? I think benchmarking should be executed manually by us before releasing and checking it as we continue to develop the feature.

design/open/lakefs-on-kv-testing-plan.md
itaiad200: I don't think it's a must for the current milestone. Maybe leave it as a stretch goal?

itaiad200 (approved):

design/open/lakefs-on-kv-testing-plan.md
N-o-Z: I think the current direction is to do the "migration" by reading from a dump so I don't think we will require to start and stop lakeFS for this flow - just prepare some DB dump with data ahead of running the test.

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,133 @@
...
+* [Question] - How do we define a covering data set, to test? How do we create
+
+### Performance
+* **A benchmark for the migration process itself:**
+  * How do we load the DB with data to migrate for benchmarking?
+  * Ideally, run on several data sizes (tens / thousands / millions branches & commits, various numbers of users, etc.) to get sense of how scaling will affect
+  * What is the expected/acceptable outcome?
+  * Should be built gradually, as we support and migrate additional package
+
+* **Benchmark comparison of both DBs**
```
N-o-Z, line 30: We might want to benchmark the migration process itself to see that it does not suffer from serious performance issues once deployed in the field with costumers who have a lot of data to migrate (resulting in an unacceptably long downtime)

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,133 @@
...
+### ```pkg/auth```
+* DB is used to store installation metadata, users and policies
+* Only 6.8% coverage
+* Some benchmarks exists that covers effectivePolicies (multi-table JOIN)
+* What is missing:
+  * Need to increase code coverage
+  * Comparison benchmark to toggle between Tables and KV and verify there is no degradation
+  * Migration tests:
+    * verify that data is migrated correctly
+    * Authentication transactions in during migration (data that was written to tables is read from KV)
```
N-o-Z, line 68: Can you explain how transactions should occur during migration? Migration should be done when lakeFS is still down

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,133 @@
...
+## Execution Plan
+
+### KVM1
+* Data level migration tests infrastructure
+  * Migrate data from Table to KV, extract both and compare
+    * Data extraction should be done by listing all objects in the DB, using a designated 'get' function, and compare
+  * Implement ~~dumpers for `gateway_multiprts`~~ `GetAll` for `multiparts.Tracker`, to return a list of `MultipartUpload`
+  * Implement comparison of `MultipartUploads` list. Lists are considered identical if objects are identical, but **not necessarily** at the same order
+  * Implement unit tests for `pkg/gateway/multiparts`
+    * Add multipart uploads using `multiparts.Tracker.Create` with Table DB (KV Feature Flag off)
```
N-o-Z, line 109: As I commented before - I think the migration tests should rely on reading data from a file and not to create the data dynamically - the migration process should happen in the lakeFS entry point before server is up

design/open/lakefs-on-kv-testing-plan.md
N-o-Z: There's no need to adjust the tests as the interface is identical. We just need to decide which scenarios we want to benchmark

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,133 @@
...
+  * Compare results and fail if KV performance is more than [TBD]% slower
+  * Need an infrastructure 
+
+### Next MS
+TBD
+
+
+## Open questions
+* What performance degradation is acceptable? (Can be 0)
+* How do we simulate KV Store failure? Do we need it at all?
```
N-o-Z, line 132: We can create a mock for either Store or StoreMessage to simulate failures.
I would describe what kind of fault injection do we need to support to better understand the faults requirements

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,133 @@
+# lakeFS on KV - Testing Plan
```
N-o-Z, line 1: While in development we will also need to modify our github pipeline to run all of our tests in 2 flavors (kv enabled/disabled)
We need to decide on the following things:

1. How do we support these flavors (In Esti / in .github actions?)
2. Do we need to run all of our jobs in both flavors (maybe we can do some mitigations)
3. Will the kv flavor be blocking for merge oprerations (even on issues that are not part of KV) this can potentially block other teams development process due to a bug in our development

We will of course need this for the KV1 milestone as well

N-o-Z (commented):
Thanks for the very thorough test plan. See my comments below

design/open/lakefs-on-kv-testing-plan.md
itai-david: @itaiad200 - If eventually we are getting rid of the migration code, how do users, that currently have lakeFS with Table DB installed, upgrade to the recent version? Are they forced to go through a Table->KV migration supporting version (i.e. at least 2 updates)?

design/open/lakefs-on-kv-testing-plan.md
itai-david: @N-o-Z How do we make sure we are not missing any DB updates arriving after the dump and before KV done importing?

design/open/lakefs-on-kv-testing-plan.md
itai-david: Absolutely correct. It is coevered in the previous section

design/open/lakefs-on-kv-testing-plan.md
itai-david: `During migration` stands for `before and after migration`, rather than during the actual process of the migration. What I meant is that data can sustain the migration process, e.g. a `multipart upload` that is started and written to the `gateway_multiparts` table, can resume and complete after the migration to KV Store

design/open/lakefs-on-kv-testing-plan.md
itai-david: Sounds good to me. Thanks

design/open/lakefs-on-kv-testing-plan.md
itai-david: 1. Good question. I think we can run it as a separate workflow (similar to [Unit test Spark metadata client](https://github.com/treeverse/lakeFS/actions/workflows/spark.yaml), for both flavors, or better, for the flavor that is not currently the default, and have the default flavor run with the general testing package.
That way, during the early development stages, we will have the test running on Table DB as part of the general testing flow and with KV enabled on the dedicated workflow, and at some point, we can switch both flags, and have the KV kicking in to the general workflow, applying to all tests that are affected by the DB section that is being enabled, while we keep the Table DB at the side workflow, at least until we ditch it for good.
That way it will be easy to delete the irrelevant tests as it will be a dedicated workflow
2. Implied from (1) - only relevant jobs will be executed in the dedicated workflow. These can be the specific tests for the feature we are currently working on, or any related tests we want to add
3. I believe it should be blocking. Idealy, a bug in our development of the KV support, will not get through, as it will be blocked too

WDYT?

design/open/lakefs-on-kv-testing-plan.md
itai-david: Added to the doc

design/open/lakefs-on-kv-testing-plan.md
itai-david: I think the migration tests should rely on the migration method that will be implemented - if it is dump&read or any other method.
Either ways, I agree with @itaiad200 that we should use the store library functions to read the data to compare. As data changes and grows, the library will obviously adapt, as to support the modified data. I'm not sure the same goes for dump support, as even if the initial implementation of the migration relies on dump, and so, will be maintained with any schema change, we might decide to replace this method with a different one, leaving us with the need to keep and support the dump functionality for the sole purpose of testing

design/open/lakefs-on-kv-testing-plan.md
itaiad200: Yes. They are forced to do that. Eventually there would be a version where all lakeFS installations must go through.

design/open/lakefs-on-kv-testing-plan.md
itaiad200: To the expansion.

design/open/lakefs-on-kv-testing-plan.md
itai-david: Added

design/open/lakefs-on-kv-testing-plan.md
itai-david: How about keeping it as a stretch goal?

design/open/lakefs-on-kv-testing-plan.md
itaiad200: I think that the maintenance burden of performance testing as a gatekeeper is huge. It will sporadically fail, require a lot of tuning and generally really hard to get right. And since we agree that it's temporary testing (until we complete the feature), I think we'll get much more value from a local execution that we control and can run before each feature/milestone release.

design/open/lakefs-on-kv-testing-plan.md
```diff
@@ -0,0 +1,139 @@
...
+  * Comparison benchmark
+
+### Ref-Dump/Restore
+Currently this functionality is not covered directly, but since it relies on '''pkg/graveler/ref''' for DB access, the DB migration should be seamless
+It can be leveraged, however, to extend the cover of '''pkg/graveler/ref''' and for performance, as it is quite exhaustive (traverses all branches and commits and tags, per repo)
+
+## Execution Plan
+
+### KVM1
+* Data level migration tests infrastructure
```
itaiad200, line 108: Every item in this list (except for maybe part of the first bullet) is `multipart` specific and is not "infrastructure". Correct?

design/open/lakefs-on-kv-testing-plan.md
N-o-Z: @itai-david I think the data should really be scenario dependant. For example if you want to test migration during multipart upload - you can build the dump file in a way that continuing the multipart upload after migration you will expect a certain outcome (upload successful + file content)

design/open/lakefs-on-kv-testing-plan.md
itai-david: FWIW - all items except for the last one are (at least partially) infrastructure
The last bullet is indeed `multipart` specific, but the first and second are intended to be system wide support, with validity only for `multipart`, as part of KV1

itai-david:
Summary:
* #3328 & #3312 (p1 for kvm1)
* #3310 (p2 for kvm1)
* #3315 & #3329 (nice to have, not for kvm1)
* In addition, #3311 is part of the migration implementation iteslf

PR #3302, opened 2022-05-02
Verify sender is properly configured during load 

shimi9276:


cmd/lakefs/cmd/run.go
```diff
@@ -244,6 +244,10 @@ var runCmd = &cobra.Command{
 			}
 		}
 
+		if !emailer.IsEmailerConfigured() {
+			logger.Warn(fmt.Sprintf("Emailer has not been properly configured, check the values in the email config"))
+		}
```
guy-har, line 249: - This will now log a warning to all users that did not configure email (all our current users)
- This function doesn't only return if the email is configured but also validates it was configured correctly, In case it wasn't configured correctly we should provide information why (this is returned by `mail.ParseAddress` )
- If a user does configure invalid email parameters he will probably miss the warning

guy-har (changes requested):

pkg/email/send_email.go
```diff
@@ -75,3 +77,13 @@ func (e *Emailer) SendInviteUserEmail(receivers []string, params map[string]stri
 	}
 	return e.SendEmailWithLimit(receivers, inviteUserWEmailSubject, body, nil)
 }
+
+func (e *Emailer) IsEmailerConfigured() bool {
```
nopcoder, line 81: In this case, cache the state of the email as part of the parameters is not a good practice - you don't know when it is updated as it can be updated externally.
- Name: IsEmailerConfigured -> IsConfigured() or IsEnabled(), will remove the name of the entity from query the state
- Just return the result as recalling it doesn't cache the state and we don't need this cache at this point.

nopcoder (changes requested):

cmd/lakefs/cmd/run.go
shimi9276: Fixed. It will be fatal in the case the sender is misconfigured with the proper error.

cmd/lakefs/cmd/run.go
```diff
@@ -244,6 +244,10 @@ var runCmd = &cobra.Command{
 			}
 		}
 
+		if !emailer.IsEmailerConfigured() {
+			logger.Warn(fmt.Sprintf("Emailer has not been properly configured, check the values in the email config"))
```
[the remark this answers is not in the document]
shimi9276: Fixed

pkg/email/send_email.go

nopcoder (changes requested):
How was it tested? It looks like the suggested solution will stop exit with error when no sender is configured.
Can you move the verification to NewEmailer? it will verify the state and return an error that will fail the lakeFS startup.

shimi9276 (commented):
@nopcoder, Moved the check to the creation of the emailer (newEmailer). (I hope this is what you meant)
As for testing, there were basic sanity checks done, but no written test. The written tests are forthcoming.

pkg/api/serve_test.go
```diff
@@ -114,7 +114,7 @@ func setupHandlerWithWalkerFactory(t testing.TB, factory catalog.WalkerFactory,
 
 	auditChecker := version.NewDefaultAuditChecker(cfg.GetSecurityAuditCheckURL())
 	emailParams, _ := cfg.GetEmailParams()
-	emailer := email.NewEmailer(emailParams)
+	emailer, _ := email.NewEmailer(emailParams)
```
nopcoder, line 117: You want to check the error in the test

pkg/loadtest/local_load_test.go
```diff
@@ -98,7 +98,7 @@ func TestLocalLoad(t *testing.T) {
 	})
 	auditChecker := version.NewDefaultAuditChecker(conf.GetSecurityAuditCheckURL())
 	emailParams, _ := conf.GetEmailParams()
-	emailer := email.NewEmailer(emailParams)
+	emailer, _ := email.NewEmailer(emailParams)
```
nopcoder, line 101: verify the return error - this is a valid flow that can fail

pkg/email/send_email.go
```diff
@@ -29,7 +30,14 @@ type Params struct {
 	LakefsBaseURL      string
 }
 
-func NewEmailer(p Params) *Emailer {
+func NewEmailer(p Params) (*Emailer, error) {
+	if p.SMTPHost == "" {
+		return &Emailer{}, nil
```
nopcoder, line 35: what will happen if we send an email using this Emailer when there is no Dialer? calling a method on Nil pointer.

nopcoder (changes requested):

nopcoder:
@shimi9276 please verify the code works with and without mailer host configured and when sender is set/unset.

pkg/api/serve_test.go
shimi9276: Fixed

pkg/loadtest/local_load_test.go
shimi9276: fixed

pkg/email/send_email.go
shimi9276: Good point, it will be a nil pointer. FIxed and testing the functionalities once again to verify it works

nopcoder (changes requested):
Please verify that this code with and without mail host configured

pkg/email/send_email.go
```diff
@@ -15,6 +16,7 @@ type Emailer struct {
 }
 
 var ErrRateLimitExceeded = errors.New("rate limit exceeded")
+var ErrNoSMTPHostConfigured = errors.New("no smtp host configured")
```
nopcoder, line 19: Prefer using scope while defining multiple variables or constants
```suggestion
var (
    ErrRateLimitExceeded = errors.New("rate limit exceeded")
    ErrNoSMTPHostConfigured = errors.New("no smtp host configured")
)
```

pkg/email/send_email.go
```diff
@@ -38,10 +46,13 @@ func NewEmailer(p Params) *Emailer {
...
 		Dialer:  dialer,
 		Limiter: limiter,
-	}
+	}, nil
 }
 
 func (e *Emailer) SendEmail(receivers []string, subject string, body string, attachmentFilePath []string) error {
+	if e.Params.SMTPHost == "" {
+		return ErrNoSMTPHostConfigured
+	}
```
nopcoder, line 55: Same verification should be done for 'Sender' and 'receivers'

pkg/email/send_email.go
```diff
@@ -29,7 +31,13 @@ type Params struct {
 	LakefsBaseURL      string
 }
 
-func NewEmailer(p Params) *Emailer {
+func NewEmailer(p Params) (*Emailer, error) {
+	if p.SMTPHost != "" {
+		_, err := mail.ParseAddress(p.Sender)
+		if err != nil {
+			return nil, err
```
nopcoder, line 38: Use `fmt.Errorf` and format an error with more information - the missing data is that we validate 'sender'.
Imagine what will be the error from parse address and the fact that you need to understand what went wrong.

nopcoder (changes requested):

pkg/email/send_email.go
shimi9276: Done

pkg/email/send_email.go
shimi9276: Done

pkg/email/send_email.go
shimi9276: Done

nopcoder (approved):
lgtm

guy-har (approved):

PR #3303, opened 2022-05-02
Delete gateway playback tests

### LINKED ISSUE #2031: Remove gateway playback test

arielshaqed:
These tests are no longer required, we test with real live clients and cover
all of these.

Fixes #2031.


cmd/lakefs/cmd/run.go
```diff
@@ -34,7 +34,6 @@ import (
 	"github.com/treeverse/lakefs/pkg/gateway"
 	"github.com/treeverse/lakefs/pkg/gateway/multiparts"
 	"github.com/treeverse/lakefs/pkg/gateway/sig"
-	"github.com/treeverse/lakefs/pkg/gateway/simulator"
```
guy-har, line 37: 😄

guy-har (approved):
LGTM
THANKS!

PR #3306, opened 2022-05-03
Improve Esti Performance

### LINKED ISSUE #2883: Nessie is often timing-out in Spark tests
We should either check what causing the flakiness or make the timeout longer. Examples:
https://github.com/treeverse/lakeFS/runs/4803247848
https://github.com/treeverse/lakeFS/runs/4801417391
https://github.com/treeverse/lakeFS/runs/4800661911

johnnyaug:
Closes #2883

### Esti performance improvements

* More resources to Spark workers. Hopefully this helps us never reach the timeout.
* Parallelize GC tests to make them faster.


nopcoder (changes requested):
Suggestion - keep the original loop of setup/test/clean, but this time run then as a single function in the background (using &).
In order to keep just 8 running in the background, keep counter and every mod 8 call `wait`. (also at the end of the loop) - it will wait until the 8 or the rest will end.
This way we will setup just the tests we run and if we fail, we didn't setup all tests.

johnnyaug:
@nopcoder - if I understand your suggestion, it will not behave like a thread (process actually) pool - because I'll have to wait for all 8 to finish before I start the next batch. Am I right?

test/spark/run-gc-test.sh
```diff
@@ -92,46 +82,71 @@ validate_gc_job() {
...
-  if ! validate_gc_job "${test_case}" ${REPOSITORY} ${file_existing_ref} ${test_id}; then
-    failed_tests+=("${test_description}")
-  else
-    echo "Test: ${test_description} - SUCCEEDED"
-  fi
-  rm -f policy.json
-  clean_repo ${REPOSITORY}
+  # run prepare_for_gc in parallel
+  prepare_for_gc "${test_case}" "${test_id}" &
+  pids[${test_id}]=$!
```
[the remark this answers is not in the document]
johnnyaug: done

nopcoder:
> @nopcoder - if I understand your suggestion, it will not behave like a thread (process actually) pool - because I'll have to wait for all 8 to finish before I start the next batch. Am I right?

True, there is another bash trick for keeping 8 background jobs working.
Alternative, you can use GNU Parallel. It is re-installed, like all good tools on GitHub virtual machine.

So something like the following will simplify the execution:
```bash
#!/bin/bash

function do_something {
    echo "doing something $1"
    sleep 1
}

export -f do_something
parallel -j 8 ::: do_something ::: {1..50}
```

Let me know what you think

johnnyaug:
@nopcoder, Why is this better than the xargs thing?

nopcoder:
@johnnyaug same effect - just I wish each job was running a complete flow setup/test/cleanup. Any reason you prefer to split the setup/cleanup?

test/spark/docker-compose.yaml
```diff
@@ -65,8 +65,8 @@ services:
     environment:
       - SPARK_MODE=worker
       - SPARK_MASTER_URL=spark://spark:7077
-      - SPARK_WORKER_MEMORY=1G
-      - SPARK_WORKER_CORES=1
+      - SPARK_WORKER_MEMORY=8G
```
arielshaqed, line 68: Is this realistic on a GH actions runner?  I see [7 GiB](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) for the entire runner VM.  All of the docker containers have to fit on that VM, along with a few other containers.

arielshaqed (commented):
I have little to add to @nopcoder :-)

How much faster does it run in practice??

nopcoder (changes requested):
Question about the structure we run in parallel

test/spark/docker-compose.yaml
johnnyaug: You're right, it's not really realistic. I will change to 4G

johnnyaug:
@arielshaqed, it cuts off about 1m every time we run the Sonnet test. It improves it by roughly 25%.

nopcoder:
@johnnyaug looks good! don't know why in my head passing args vs standard input looks more scalable. but I guess it is a non issue in this case.

nopcoder (approved):
Thanks, lgtm!

PR #3307, opened 2022-05-04
Suggested changes to docs homepage tone

arielshaqed:
These are organized by commit, so consider reviewing each commit separately (and possibly asking me to drop an entire commit, if it is completely off the mark).  Also the commits probably get more controversial in order of appearance. :-)


arielshaqed:
THANKS!

PR #3309, opened 2022-05-04
Split authentication and authorization into two separate docs

shimi9276:
As first step to adding documentation on user management.

### DIFF
--- a/docs/reference/authentication.md
+++ b/docs/reference/authentication.md
@@ -3,7 +3,7 @@ layout: default
 title: Authentication 
 description: This section covers Authentication of your lakeFS server.
 parent: Reference
-nav_order: 59
+nav_order: 60
 has_children: false
 ---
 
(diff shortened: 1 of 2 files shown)


docs/reference/authentication.md
```diff
@@ -0,0 +1,101 @@
...
+### User Authentication
+
+lakeFS authenticates users from a built-in authentication database, or
+optionally from a configured LDAP server.
+
+#### Built-in database
+
+The built-in authentication database is always present and active.  Use the
+Web UI at Administration / Users to create users.  Users have an access key
+`AKIA...` and an associated secret access key.  These credentials are valid
```
johnnyaug, line 27: Don't we have email authentication as well now?

docs/reference/authentication.md
```diff
@@ -0,0 +1,101 @@
...
+to log into the Web UI, or to authenticate programmatic requests to the API
+Server, or for S3 Gateway Authentication.
```
johnnyaug, line 29:
```suggestion
to log into the Web UI, or to authenticate programmatic requests to the API
Server or the S3 Gateway.
```

johnnyaug (commented):
Can we add info about the user/password authentication? Assuming it's available today - I'm not really in the loop there.

docs/reference/authentication.md
shimi9276: We do, The idea here is to add exactly that (in a coming PR hopefully by today/tomorrow), and as a first step we felt is made more sense to split the authentication and authorization into two separate docs.

PR #3314, opened 2022-05-05
docs: fix conn type in airflow integration

nopcoder:
Fix adding connection to lakefs using airflow command-line:

```
UserWarning: The type provided to --conn-type is invalid: HTTP
```


nopcoder:
Already have a pending issue to have custom connection type, which can be a better option than customize http connection

PR #3317, opened 2022-05-05
Skip metaclient tests in esti

johnnyaug:
These tests already run as part of the _Unit test Spark metadata client_ workflow.

### DIFF
--- a/.github/workflows/esti.yaml
+++ b/.github/workflows/esti.yaml
@@ -571,8 +571,8 @@ jobs:
       - name: Package Metaclient
         working-directory: clients/spark
         run: |
-          sbt lakefs-spark-client-301/assembly
-          sbt lakefs-spark-client-312-hadoop3/assembly
+          sbt 'set core3 / assembly / test := {}' lakefs-spark-client-301/assembly
+          sbt 'set core312 / assembly / test := {}' lakefs-spark-client-312-hadoop3/assembly
 
       - name: Prepare Metaclient location for export
         # upload-artifact cannot take a working-directory option (that only


arielshaqed (changes requested):
I don't like this change: the tested assembly should be the one that we would publish,  if we did publish it.
If anything, I would skip the separate unit test flow for the metadata client. (Or understand what flow sbt expects us to use; I agree the current flow makes little sense)

johnnyaug:
@arielshaqed, Just to clarify - these tests **do** run when we publish the metadata client - here we just package the jar for the GC testing.

arielshaqed:
> @arielshaqed, Just to clarify - these tests **do** run when we publish the metadata client, and also on another workflow - here we just package the jar for the GC testing. I skip them here because it can save some time on Esti testing - which is a bottleneck

Understood, of course.  And I do realize that it takes time.  However removing them makes useful flows fragile: for instance, I can think everything is good (because these tests passed), and then fail to publish at the last moment.

What do you say we instead kill the entire [unit test Spark metadata client](https://github.com/treeverse/lakeFS/blob/10bced0f40f110836e8e2cd2dcbc95b0308c4397/.github/workflows/spark.yaml#L1-L20) flow?

arielshaqed (approved):
After chatting it over, I'm convinced!  In the worst case we don't notice that the unit test workflow fails, and try to publish, and then need to `git bisect` to determine where the unit test started failing.

Which is **NOT THAT BAD** in exchange for ~ 1m30s less time to test!

PR #3319, opened 2022-05-05
Fix: login redirects to credentials page

johnnyaug:
After doing a setup through "/setup", the next variable would be populated with the string "undefined", causing a wrong redirect.


itai-david (approved):
I know nothing about js, but as it is a really small change, I guess if it fix the problem, it can'r be wrong.

PR #3321, opened 2022-05-08
Feature airflow dag

### LINKED ISSUE #3257: Airflow hook wait for dag to finish
Airflow hook currently triggers a DAG. On successfully triggering the flow, it completes successfully.
To use the DAG completion status as part of the hook completion status, we will need to wait for the DAG to complete.
It depends on https://github.com/treeverse/lakeFS/issues/3020 to enable running a DAG without getting into timeout.
With enabling long-running hooks, we can have a property on the airflow hook to specify that it should wait for the DAG to complete.

nopcoder:
Close #3257 

- Airflow hook: add wait_for_dag property, boolean (Default: false) to indicate if we like to wait for DAG to complete.
- Airflow hook: timeout (Default: 1m). Duration to wait for DAG to complete.
- DAG Run ID updated to use hook run ID and not timestamp

Testing was done manually - using a successful and unsuccessful DAG.

### DIFF
--- a/docs/setup/hooks.md
+++ b/docs/setup/hooks.md
@@ -231,18 +231,21 @@ Example:
 ```
 
 ### Airflow Hooks
+
 Airflow Hook triggers a DAG run in an Airflow installation using [Airflow's REST API](https://airflow.apache.org/docs/apache-airflow/stable/stable-rest-api-ref.html#operation/post_dag_run).
 The hook run succeeds if the DAG was triggered, and fails otherwise.
 
 #### Action file Airflow hook properties
 
-| Property | Description                                             | Data Type | Example                 | Required | Env Vars Support |
-|----------|---------------------------------------------------------|-----------|-------------------------|----------|------------------|
-| url      | The URL of the Airflow instance                         | String    | "http://localhost:8080" | true     | no               |
-| dag_id   | The DAG to trigger                                      | String    | "example_dag"           | true     | no               |
-| username | The name of the Airflow user performing the request     | String    | "admin"                 | true     | no               |
-| password | The password of the Airflow user performing the request | String    | "admin"                 | true     | yes              |
-| dag_conf | DAG run configuration that will be passed as is         | JSON      |                         | false    | no               |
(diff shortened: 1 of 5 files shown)


docs/setup/hooks.md
```diff
@@ -231,18 +231,21 @@ Example:
...
+| username      | The name of the Airflow user performing the request             | String                                                                                    | "admin"                 | true     | no               |
+| password      | The password of the Airflow user performing the request         | String                                                                                    | "admin"                 | true     | yes              |
+| dag_conf      | DAG run configuration that will be passed as is                 | JSON                                                                                      |                         | false    | no               |
+| wait_for_dag  | Wait for DAG run to complete and reflect state (default: false) | Boolean                                                                                   |                         | false    | no               |
+| timeout       | Duration while waiting for DAG to complete (default: 1m)        | String (golang's [Duration](https://golang.org/pkg/time/#Duration.String) representation) |                         | false    | no               |
```
itaiad200, line 248:
```suggestion
| timeout       | Time to wait for the DAG run to complete (default: 1m)        | String (golang's [Duration](https://golang.org/pkg/time/#Duration.String) representation) |                         | false    | no               |
```

pkg/actions/webhook.go
```diff
@@ -128,7 +128,7 @@ func (w *Webhook) Run(ctx context.Context, record graveler.HookRecord, buf *byte
 	return nil
 }
 
-func executeAndLogResponse(ctx context.Context, req *http.Request, buf *bytes.Buffer, timeout time.Duration) (n int, err error) {
+func executeAndLogResponse(ctx context.Context, req *http.Request, buf *bytes.Buffer, timeout time.Duration, respMsg interface{}) (n int, err error) {
```
itaiad200, line 131: I would create a separate function with shared logic with this one, instead of forcing most callers to pass a nil for a value they don't use (at the very least add documentation for this arg).

itaiad200 (approved):
LGTM! minor comments

itaiad200 (approved):
Thanks for applying the changes!

PR #3325, opened 2022-05-08
KV Migration (export import phase 1) Proposal

N-o-Z:


design/accepted/kv_migration.md
```diff
@@ -0,0 +1,35 @@
...
+KV Migration Version will be a consecutive version succeeding the last DB Migration version.
+
+## Migration Mechanism
+1. Migrate up command will perform schema migration up to the latest version
+2. Check if current version equals the KV Migration Version and perform KV Migration
+3. KV Migration will consist of 2 main steps: Export, Import
+4. Export:
+   1. Each migrating package will read the data from postgres into its corresponding KV models 
+   2. Data will be saved on file per pkg in it’s KV format 
+   3. Each pkg file will consist of a header containing the migration version and etc…
```
N-o-Z, line 24: Per path prefix?

design/accepted/kv_migration.md
```diff
@@ -0,0 +1,35 @@
...
+1. KV Migration - The transition from schema based postgres implementation to key value store based postgres implementation
+2. DB Migration - The process of migrating postgres data using the postgres versioning schema
+3. KV Migration Version - “Schema” version for which the KV Migration will occur
+
+## Abstract
+Current migration mechanism relies on postgres versioning schema.
+KV Migration flow is a transitory mechanism, which will be needed only when upgrading lakeFS to the KV Migration Version. For efficiency and re-use we would like to create a preliminary version of the export-import feature to implement the KV Migration flow.
+
+## Considerations
+Although we cannot use the DB Migration flow to handle the KV migration, it must be incorporated into it and executed in the context of the DB migration
```
itaiad200, line 13: Why can't we use it?

itaiad200: And why do we must incorporated it?

design/accepted/kv_migration.md
```diff
@@ -0,0 +1,35 @@
...
+DB migration must be performed up to the latest version before the KV Migration Version and only then KV Migration can happen
+KV Migration Version will be a consecutive version succeeding the last DB Migration version.
```
itaiad200, line 15: But KV is an ongoing task, which is hidden behind a feature-flag. How would DB migration continue to work? How would we make KV migrations non-mandatory?

design/accepted/kv_migration.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Migration Mechanism
+1. Migrate up command will perform schema migration up to the latest version
```
itaiad200, line 18: schema migration == DB migration?

design/accepted/kv_migration.md
```diff
@@ -0,0 +1,35 @@
...
+2. Check if current version equals the KV Migration Version and perform KV Migration
+3. KV Migration will consist of 2 main steps: Export, Import
+4. Export:
+   1. Each migrating package will read the data from postgres into its corresponding KV models 
+   2. Data will be saved on file per pkg in it’s KV format 
+   3. Each pkg file will consist of a header containing the migration version and etc… 
+   4. Files will be saved in a designated export folder 
+5. Import:
+   1. Read files content from the ‘export’ folder directly to DB
+
```
itaiad200, line 28: How would you handle schema evolution of KV packages?

itaiad200 (changes requested):
Awesome design, thanks! I don't think I fully understand the versioning of the migrations during the transition..

design/accepted/kv_migration.md
```diff
@@ -0,0 +1,35 @@
+# KV Migration (export import phase 1) Design
+
+## Definitions
+1. KV Migration - The transition from schema based postgres implementation to key value store based postgres implementation
+2. DB Migration - The process of migrating postgres data using the postgres versioning schema
+3. KV Migration Version - “Schema” version for which the KV Migration will occur
+
+## Abstract
+Current migration mechanism relies on postgres versioning schema.
```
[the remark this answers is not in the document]
N-o-Z: specified explicitly

design/accepted/kv_migration.md
N-o-Z: Adding the information to the document

design/accepted/kv_migration.md
N-o-Z: Fixed

design/accepted/kv_migration.md
N-o-Z: Currently out of scope for this document - but once we are post KV migration, the migrate up command will be used to handle any schema changes as before, according to the current versioning scheme defined in the database. We do need to consider how do we want to save the version in the database and define the mechanism for the schema modifications.

design/accepted/kv_migration.md
N-o-Z: Since KV migration must occur on the latest schema version we must ensure we migrate to the latest and greatest while still in DB schema, only then can we perform the KV migration.

itaiad200 (approved):
This makes perfect sense :) LGTM

design/accepted/kv_migration.md
```diff
@@ -0,0 +1,41 @@
...
+4. Export:
+   1. Each migrating package will read the data from postgres into its corresponding KV models 
+   2. Data will be saved on file per pkg in its KV format 
+   3. Each pkg file will consist of a header containing the migration version and etc… 
+   4. Files will be saved in a designated export folder 
+5. Import:
+   1. Read files content from the ‘export’ folder directly to DB
+      
+6. Upon successful import - update DB migration version in DB
+7. Drop old DB tables
```
nopcoder, line 36: Let's have this step optional, one use-case will be testing.

design/accepted/kv_migration.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Failures:
+1. In event of failure on any of the KV migration steps, before the DB migration version is updated, the migration will be considered as failed
+2. Upon re-run of migrate up, the KV Migration will start from scratch, disregarding any previous changes
+3. In the event of failure after updating the DB migration version and before/during cleanup (tables drop), an event will be issued to the user, requiring manual intervention
```
nopcoder, line 41: The version update will be done as the last step of the import? if not, which step we fail in this case - or we are talking about general bug after migrate completes.

nopcoder (approved):
LGTM, just one suggestion and a question

design/accepted/kv_migration.md
N-o-Z: The version update will happen as part of the migrate up flow (not part of the import flow - but after). In case the DB migration succeeded but we failed during the cleanup of old tables - we consider it as migration successful and require the user to manually remove the redundant tables

design/accepted/kv_migration.md
```diff
@@ -0,0 +1,45 @@
+# KV Migration (export import phase 1) Design
+
+## Definitions
+1. KV Migration - The transition from schema based postgres implementation to key value store based postgres implementation
```
itai-david, line 4: Does KV has to be over postgress implementation for that? Is there a difference in the definitions for other (future) "engines"?

design/accepted/kv_migration.md
```diff
@@ -0,0 +1,45 @@
...
+2. DB Migration - The process of migrating postgres data using the postgres versioning schema
```
itai-david, line 5: Is that a migration to the latest version prior to KV?

design/accepted/kv_migration.md
```diff
@@ -0,0 +1,45 @@
...
+Migration check for KV should be enabled only when the KV Enabled flag was passed. While the flag is disabled migration will
+behave as usual, not taking into account KV migration.
+
+
+## Migration Mechanism
+1. Migrate up command will perform DB migration up to the latest version
+2. Check if current version equals the KV Migration Version and perform KV Migration
+3. KV Migration will consist of 2 main steps: Export, Import
+4. Export:
+   1. Each migrating package will read the data from postgres into its corresponding KV models
```
itai-david, line 32: Can you please explain what do you mean by "read from postgres **into KV models**"?

design/accepted/kv_migration.md
```diff
@@ -0,0 +1,45 @@
...
+   3. Each pkg file will consist of a header containing the migration version and creation date
+   4. Files will be saved in a designated export folder
+5. Import:
+   1. Read files content from the ‘export’ folder directly to DB
+
+6. Upon successful import - update DB migration version in DB
+7. Drop old DB tables - support an optional flag for testing purposes
+
+## Failures:
+1. In event of failure on any of the KV migration steps, before the DB migration version is updated, the migration will be considered as failed
```
itai-david, line 43: What happen in that case? Are we deleting the exported files?

itai-david (commented):
Thank you @N-o-Z.
A couple comments with questions, mainley to clarify things I'm not sure I fully understand
Other than that, LGTM

design/accepted/kv_migration.md
N-o-Z: KV Migrate happens only on prior lakeFS installations using postgres and will use the same DB instance for migration.
After that - migrating from postgres KV store to another DB should be done via the export import feature

design/accepted/kv_migration.md
N-o-Z: Each pkg should read it's tables and create the appropriate KV models to support it's functionality. The data will be saved in a format that can be read by the pkg's KV implementation. I will clarify it in the document

design/accepted/kv_migration.md
N-o-Z: 1. Same as today - lakeFS will not be able to run unless migrated to latest version
2. We are not creating files during the migration. For now, data is saved in-memory. If in the future we will need to write files - they should be discarded if migration fails

N-o-Z: Updated accordingly

PR #3333, opened 2022-05-10
Fix pebble panic in api unit-tests

itaiad200:


nopcoder (approved):
good catch

arielshaqed (approved):
Yeah, that ought to do the trick.  THANKS!

(I _love_ languages with no destructors, resource management is so much easier with garbage-collected languages because they only worry about a single resource...).

PR #3338, opened 2022-05-10
webUI - Get user from api when not in cache

guy-har:
Currently:
-  The user information is saved in the local cache on login.
-  The login token is saved in cookie
- user navbar (which includes logout) is shown only if the is user information in the cache

In case the local cache was erased for some reason, users will not have an option to logout 

This PR fixes this partly 
- we will see the user in the navbar and have an option to logout but the value under `user:` will not be available

### DIFF
--- a/webui/src/lib/components/navbar.jsx
+++ b/webui/src/lib/components/navbar.jsx
@@ -20,7 +20,7 @@ const NavUserInfo = () => {
         <Navbar.Collapse className="justify-content-end">
             <NavDropdown title={user.friendly_name || user.id} className="navbar-username" alignRight>
                 <NavDropdown.Header>
-                    User: <code>{user.accessKeyId}</code>
+                    User: <code>{user.accessKeyId || user.id}</code>
                 </NavDropdown.Header>
 
                 <NavDropdown.Divider/>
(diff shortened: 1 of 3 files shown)


nopcoder (changes requested):

johnnyaug (changes requested):
I don't understand this change:
1. When logging in - why would I every want to use the cache? This may be dangerous.
2. The other use of the getCurrentUser function is in lib/hooks/users.jsx and it doesn't pass the accessKeyId parameter.

webui/src/lib/components/navbar.jsx
```diff
@@ -20,7 +20,7 @@ const NavUserInfo = () => {
         <Navbar.Collapse className="justify-content-end">
             <NavDropdown title={user.friendly_name || user.id} className="navbar-username" alignRight>
                 <NavDropdown.Header>
-                    User: <code>{user.accessKeyId}</code>
+                    User: <code>{user.accessKeyId || user.id}</code>
```
johnnyaug, line 23: So this will sometimes show the user ID and sometime the access key? Better just not show anything.

johnnyaug (approved):
Suggestion to hide access key if it's not there.

webui/src/lib/components/navbar.jsx
guy-har: @johnnyaug By not showing anything you mean Not showing anything when there is no accessKey or Not showing anything at all?

johnnyaug: I say remove it altogether, otherwise strange behaviors can be expected.

arielshaqed:
Found a useful reference for the underlying issue with all technical details included: https://shouldiblamecaching.com/

PR #3339, opened 2022-05-10
test lakefs kv delete while iteration

itai-david:


pkg/kv/kvtest/store.go
```diff
@@ -369,3 +372,132 @@ func testScanPrefix(t *testing.T, ms MakeStore) {
...
+	// iterating over the input sequence and triggering the read/delete routines accordingly
+	for i := 0; i < len(sequence); i++ {
+		chMap[sequence[i]] <- true
+		err := <-chErr
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	close(chMap['R'])
```
N-o-Z, line 479: If I'm not mistaken - you're not supposed to close the channels while other goroutines are using them

N-o-Z: In general, you don't have to close channels

pkg/kv/kvtest/store.go
```diff
@@ -369,3 +372,132 @@ func testScanPrefix(t *testing.T, ms MakeStore) {
...
+			t.Fatal("Unexpected failure reading value for verification", err)
+		}
+		if !bytes.Equal(val, kve.Value) {
+			t.Fatal("Expected value not found for key", kve.Key)
+		}
+	}
+
+	for _, kve := range toDelData {
+		_, err := store.Get(ctx, kve.Key)
+		if !errors.Is(err, kv.ErrNotFound) {
```
N-o-Z, line 499: I think you should separate between not getting an error - and getting and error that is not kv.ErrNotFound. It will be more clear when it fails as to exactly what happened

pkg/kv/kvtest/store.go
```diff
@@ -369,3 +372,132 @@ func testScanPrefix(t *testing.T, ms MakeStore) {
...
+	//     iteration
+	// deletion
+	testDeleteWhileIterPrefixSingleSequence(t, ms, "DDDDRDRDDRRDDDDD")
+	//           iteration
+	// deletion
+	testDeleteWhileIterPrefixSingleSequence(t, ms, "DDDDDRRRRR")
+}
+
+// Executing a sequence of (R)ead and (D)elete operations, according to the specified sequence
+func testDeleteWhileIterPrefixSingleSequence(t *testing.T, ms MakeStore, sequence string) {
```
N-o-Z, line 398: How do you make sure that all the R entries were actually read (non was skipped due to the cuncurrent deletion)? Or is it not part of the test?

N-o-Z (changes requested):
I really like the approach for testing this scenario, it was really interesting to read - thanks!
One thing I think we want to make sure to cover is in implementations based on ordered key value stores, that the delete while iterating works also for items that are inside the scan range. In the current implementation of the test, delete and read entries have different prefixes, and so they will have disjoint ranges in the DB in that case

pkg/kv/kvtest/store.go
itai-david: I believe that [close](https://pkg.go.dev/builtin@go1.18.2#close) is the way (at least one of them) to stop the `for ... range` loop in the routine.
The channel is shut down only after the last value is received (and since it happens outside the send loop, it is guaranteed that all values are already sent), and will cause the next `receive` to return with `ok` equals to `false`, effectively terminating the for loop

pkg/kv/kvtest/store.go
itai-david: Done

pkg/kv/kvtest/store.go
itai-david: Originally, I had no intention to do so, but since you brought it up, I guess it is worth verifying
Done - thanks

itai-david:
Added a test for concurrent scan/delete on same prefix

go.mod
```diff
@@ -75,7 +75,10 @@ require (
 	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
 )
 
-require golang.org/x/time v0.0.0-20220224211638-0e9765cccd65
+require (
+	golang.org/x/sync v0.0.0-20210220032951-036812b2e83c
+	golang.org/x/time v0.0.0-20220224211638-0e9765cccd65
+)
```
itaiad200, line 81: Are you using these `sync` & `time` packages on purpose?

pkg/kv/kvtest/store.go
```diff
@@ -527,3 +529,155 @@ func testDeleteWhileIterPrefixSingleSequence(t *testing.T, ms MakeStore, sequenc
...
+
+	for cleanIter.Next() {
+		e := cleanIter.Entry()
+		if cleanIter != nil {
+			err = store.Delete(ctx, e.Key)
+			if err != nil {
+				t.Fatal(err)
+			}
+		}
+	}
```
itaiad200, line 594: Interesting, you delete while you scan as a setup to a test where you check that. Maybe be on the safe side and aggregate the scan results in an array and delete afterwards?

pkg/kv/kvtest/store.go
```diff
@@ -527,3 +529,155 @@ func testDeleteWhileIterPrefixSingleSequence(t *testing.T, ms MakeStore, sequenc
...
+
+	// verify all delete entries fits delPref
+	for k := range deleteDone {
+		if strings.Index(k, string(delPref)) != 0 {
+			t.Fatal("unexpected key deleted for prefix", k, string(delPref))
+		}
+	}
+
+	// verify all entires that fits delPref are indeed deleted, i.e. no such entry is left
+	// in the store
```
itaiad200, line 667: Is there an easy way to validate that nothing else was mistakenly deleted?

itaiad200 (approved):
Few suggestions, overall LGTM

pkg/kv/kvtest/store.go
```diff
@@ -369,3 +374,310 @@ func testScanPrefix(t *testing.T, ms MakeStore) {
...
+	}
+
+	// verify all delete entries fits delPref
+	for k := range deleteDone {
+		if strings.Index(k, string(delPref)) != 0 {
+			t.Fatal("unexpected key deleted for prefix", k, string(delPref))
+		}
+	}
+
+	// verify all entires that fits delPref are indeed deleted, i.e. no such entry is left
```
N-o-Z, line 666:
```suggestion
	// verify all entries that fits delPref are indeed deleted, i.e. no such entry is left
```

pkg/kv/kvtest/store.go
```diff
@@ -369,3 +374,310 @@ func testScanPrefix(t *testing.T, ms MakeStore) {
...
+		return nil
+	})
+
+	err = errGrp.Wait()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// verify all readDone entries fits readPref
+	for k := range readDone {
```
N-o-Z, line 653: We should add another check (verify all readPref has readDone) that scans the prefix again - and verifies that each entry has a true value in readDone

go.mod
itai-david: `sync` - yes
`time` - Not explicitly. Is it possible that `sync` requires it?

pkg/kv/kvtest/store.go
itai-david: It is not required that all entries under `readPref` will be read. Since we are reading and deleting overlapping prefixes (purpose of this test), an entry that satisfy both `delPref` and `readPref`can be deleted before it is read. We do need to guarantee that if an entry was read, it satisfies `readPref`, but not the other way around

go.mod
itai-david: Moreover, if I'm not wrong, `time` was there before

pkg/kv/kvtest/store.go
itai-david: Agreed. Done

pkg/kv/kvtest/store.go
itai-david: Yes there is :-)
Added code to collect all entries to an entries array, and iterating it to go through all **created** keys. Keys that are not expected to be deleted (not falling under `delPref`) are expected to be reachable via `store.Get`

pkg/kv/kvtest/store.go
N-o-Z: What I meant was that we need to make sure that all the entries in the range were either deleted or read. What you checked here is that everything you read, wasn't deleted which is another thing.

pkg/kv/kvtest/store.go
```diff
@@ -369,3 +374,333 @@ func testScanPrefix(t *testing.T, ms MakeStore) {
...
+	}
+
+	// verify all delete entries fits delPref
+	for k := range deleteDone {
+		if strings.Index(k, string(delPref)) != 0 {
+			t.Fatal("unexpected key deleted for prefix", k, string(delPref))
+		}
+	}
+
+	// verify entries that does not fit delPref, were not accidentally deleted
```
N-o-Z, line 675: Seems like many of the validations on both tests can be exported to a helper function to reduce code duplication

pkg/kv/kvtest/store.go
itai-david: Done

pkg/kv/kvtest/store.go
itai-david: Done

pkg/kv/kvtest/store.go
```diff
@@ -369,3 +374,327 @@ func testScanPrefix(t *testing.T, ms MakeStore) {
 		t.Fatal("ScanPrefix entries didn't match:", diff)
 	}
 }
+
+func testDeleteWhileIterPrefix(t *testing.T, ms MakeStore) {
+	// iteration
```
N-o-Z, line 379: Strange comments and formatting ??

pkg/kv/kvtest/store.go
```diff
@@ -369,3 +374,327 @@ func testScanPrefix(t *testing.T, ms MakeStore) {
...
+	defer store.Close()
+
+	readPref := uniqueKey("readPrefix")
+	toDelPref := uniqueKey("toDelPrefix")
+	naPref := uniqueKey("naPrefix")
+
+	numRead := strings.Count(sequence, "R")
+	numDel := strings.Count(sequence, "D")
+
+	initialStore := []kv.Entry{}
```
N-o-Z, line 412:
```suggestion
	var initialStore []kv.Entry
```

pkg/kv/kvtest/store.go
```diff
@@ -369,3 +374,327 @@ func testScanPrefix(t *testing.T, ms MakeStore) {
...
+
+	// Emptying the KV store from previous entries. This is essential in these tests as
+	// the verification relies on exact keys to exist
+	cleanIter, err := store.Scan(ctx, delPref)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer cleanIter.Close()
+
+	keysToClean := [][]byte{}
```
N-o-Z, line 552:
```suggestion
	var keysToClean [][]byte
```

pkg/kv/kvtest/store.go
```diff
@@ -369,3 +374,327 @@ func testScanPrefix(t *testing.T, ms MakeStore) {
...
+
+	for _, k := range keysToClean {
+		err = store.Delete(ctx, k)
+		if err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	// initialKv holds all the created items, to later verify that no items, that do not fit delPref, were deleted
+	initialKv := []kv.Entry{}
```
N-o-Z, line 570:
```suggestion
	var initialKv []kv.Entry
```

pkg/kv/kvtest/store.go
itai-david: That's just the way the different `iteration` vs `deletion` orders are depicted in the issue

N-o-Z: Got it now 🤟

PR #3340, opened 2022-05-10
KV Store: Migration and multipart

### LINKED ISSUE #3243: KV multipart upload migration
Data migration between db based multipart upload to kv

N-o-Z:
Closes #3243


itai-david (changes requested):
Thank you for this code. It really helped my in better understanding the migration mechanizm.
Just a few small comments/suggestions/requests below. Nothing too crucial

cmd/lakefs/cmd/migrate.go
```diff
@@ -42,12 +42,12 @@ var upCmd = &cobra.Command{
 	Short: "Apply all up migrations",
 	Run: func(cmd *cobra.Command, args []string) {
 		cfg := loadConfig()
-		err := db.MigrateUp(cfg.GetDatabaseParams())
+		drop, _ := cmd.Flags().GetBool("drop-tables")
```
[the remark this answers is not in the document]
N-o-Z: You are correct - but currently I don't want to expose it in the CLI - only after we release the feature WDYT?
Maybe just add a TODO there?

pkg/db/migration.go
```diff
@@ -41,7 +49,7 @@ func (d *DatabaseMigrator) Migrate(ctx context.Context) error {
 	lg := log.WithFields(logging.Fields{
 		"direction": "up",
 	})
-	err := MigrateUp(d.params)
+	err := MigrateUp(d.params, false)
```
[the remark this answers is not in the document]
N-o-Z: Migrate function is called only on setup (new installation of lakeFS), so basically you should already load in KV format. It will mean that the old migration flow (which initially creates the tables) should have been removed and we will go straight to creating the kv DB.
To emulate it, I can check for KV enabled on setup and if so, skip the normal migrate. WDYT?

pkg/db/migration.go
```diff
@@ -228,3 +237,109 @@ func MigrateVersion(ctx context.Context, dbPool Database, params params.Database
...
+
+	// Update migrate version
+	err = m.Force(kv.MigrateVersion + 1)
+	if err != nil {
+		logging.Default().WithError(err).Error("Failed setting migrate version")
+		return err
+	}
+
+	if drop {
+		tables := []string{"gateway_multiparts"}
```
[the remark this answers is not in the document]
N-o-Z: The problem with that is that we will have to import db from the multiparts package which will cause a cyclic import

pkg/gateway/multiparts/db_tracker.go
```diff
@@ -4,11 +4,22 @@ import (
...
 	"github.com/treeverse/lakefs/pkg/db"
+	"github.com/treeverse/lakefs/pkg/kv"
+	"github.com/treeverse/lakefs/pkg/kv/export"
+	"google.golang.org/protobuf/proto"
 )
 
+// TODO(niro): Remove after Migration version
+//nolint:gochecknoinits
+func init() {
+	db.KVRegister("multiparts", Migrate)
```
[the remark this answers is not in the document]
N-o-Z: Can you explain what you meant here?

pkg/kv/export/utils.go
```diff
@@ -0,0 +1,71 @@
+package export
```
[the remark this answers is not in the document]
N-o-Z: The export mechanism here is implementation specific to KV stores.
I think it belongs under the KV package, but I'm open to hear more opinions about it

cmd/lakefs/cmd/migrate.go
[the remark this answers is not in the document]
itai-david: Either that, or you can user [MarHidden](https://pkg.go.dev/github.com/spf13/pflag#FlagSet.MarkHidden), which will let allow us to start using it (e.g. for tests) and it is still "not exposed"
Your call

pkg/db/migration.go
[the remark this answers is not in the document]
itai-david: I think we are OK as it is. Thanks for the explanation

pkg/db/migration.go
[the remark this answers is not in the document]
itai-david: So basically this code is going to change with each package that will support KV?

pkg/gateway/multiparts/db_tracker.go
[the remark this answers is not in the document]
itai-david: Just a reference to https://github.com/treeverse/lakeFS/pull/3340#discussion_r869855195
No action needed here

pkg/kv/export/utils.go
```diff
@@ -0,0 +1,71 @@
...
+					return err
+				}
+			}
+			err = fd.Close()
+			if err != nil {
+				return err
+			}
+		}
+	}
+}
```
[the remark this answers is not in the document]
itai-david: :)

pkg/kv/export/utils.go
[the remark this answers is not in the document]
itai-david: I was refering to the `Untar` function (and the entire `utils.go` file), and that it looks generic to me. Am I missing something that is KV/Export specific?

pkg/kv/export/export.go
```diff
@@ -0,0 +1,141 @@
...
+	reader, err := os.Open(path)
+	if err != nil {
+		return fmt.Errorf("error opening file (%s): %w", path, err)
+	}
+	defer reader.Close()
+
+	pattern := exportPrefix + strings.Split(filepath.Base(path), ".tar.gz")[0]
+	dir, err := os.MkdirTemp("", pattern)
+	if err != nil {
+		return fmt.Errorf("error creating temp dir (%s): %w", pattern, err)
```
itaiad200, line 88:
```suggestion
		return fmt.Errorf("creating temp dir (%s): %w", pattern, err)
```

pkg/kv/export/export.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	defer os.RemoveAll(dir) // clean up
+
+	err = Untar(dir, reader)
+	if err != nil {
+		return fmt.Errorf("error extracting files from tar (%s): %w", dir, err)
+	}
+
+	var files []string
+	_ = filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
```
itaiad200, line 98: `path` is being shadowed, please change the name

pkg/kv/export/export.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func Import(ctx context.Context, store kv.Store, logger logging.Logger, path string) error {
+	reader, err := os.Open(path)
+	if err != nil {
+		return fmt.Errorf("error opening file (%s): %w", path, err)
+	}
+	defer reader.Close()
+
+	pattern := exportPrefix + strings.Split(filepath.Base(path), ".tar.gz")[0]
```
itaiad200, line 85: Can the file not end with `.tar.gz`?

pkg/kv/export/export.go
```diff
@@ -0,0 +1,141 @@
...
+		}
+		err = store.SetIf(ctx, entry.Key, entry.Value, nil)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func Import(ctx context.Context, store kv.Store, logger logging.Logger, path string) error {
```
itaiad200, line 78: Please add doc for the path..

pkg/kv/export/export.go
itaiad200: And why expect a path if you're just using the last element of it?
e.g. passing `foo/bar/bla.tar.gz` is like passing `foo2/bar2/bla.tar.gz`?

pkg/kv/export/export.go
```diff
@@ -0,0 +1,141 @@
...
+	if err != nil {
+		return fmt.Errorf("error extracting files from tar (%s): %w", dir, err)
+	}
+
+	var files []string
+	_ = filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+		if info.Mode().IsRegular() {
+			files = append(files, path)
+		} else {
+			logger.Warn("Ignoring irregular file. file:", path, "mode:", info.Mode())
```
itaiad200, line 102: Why not fail here?

pkg/kv/export/utils.go
```diff
@@ -0,0 +1,71 @@
...
+		case header == nil:
+			continue
+		}
+		path := filepath.Join(dir, header.Name) // nolint:gosec
+		if !strings.HasPrefix(path, filepath.Clean(dir)) {
+			logging.Default().WithError(ErrIllegalFilePath).Error("Illegal file path:", path)
+			return ErrIllegalFilePath
+		}
+		switch header.Typeflag {
+		case tar.TypeDir:
```
itaiad200, line 45: Are we expecting dirs here? where can I read the format?

pkg/kv/store.go
```diff
@@ -8,7 +8,10 @@ import (
 	"sync"
 )
 
-const PathDelimiter = "/"
+const (
+	MigrateVersion = 37
```
itaiad200, line 12: Please explain why this is a const :)

itaiad200 (changes requested):
Reviewing everything except the tests - will get to it next week.
I think we're missing 2 key things:
- Small writeup (design doc, comment on the issue , etc..) explaining the migration protocol and the file format. Apologies if I missed it somewhere.
- Wiring it to `migrate` when `kv` is enabled..

pkg/db/migration.go
```diff
@@ -228,3 +236,109 @@ func MigrateVersion(ctx context.Context, dbPool Database, params params.Database
...
+			return nil
+		})
+	}
+	err = g.Wait().ErrorOrNil()
+	if err != nil {
+		return err
+	}
+
+	// Update migrate version
+	err = m.Force(kv.MigrateVersion + 1)
```
nopcoder, line 300: I prefer not to bring the migration to invalid state that that we will need to maintain a constant in order to verify.
If we assume that the latest migration should always be applied the 'kv' store should hold the information if the data-migration was completed or not.

pkg/gateway/multiparts/db_tracker.go
```diff
@@ -92,3 +103,39 @@ func (m *dbTracker) Delete(ctx context.Context, uploadID string) error {
 	})
 	return err
 }
+
+func Migrate(d db.Database, writer io.Writer) error {
```
nopcoder, line 107: accept ctx

pkg/gateway/multiparts/db_tracker.go
```diff
@@ -92,3 +103,39 @@ func (m *dbTracker) Delete(ctx context.Context, uploadID string) error {
...
+	err := je.Encode(export.Header{
+		Version:   kv.MigrateVersion + 1,
+		Timestamp: time.Now(),
+	})
+	if err != nil {
+		return err
+	}
+
+	entries := make([]MultipartUpload, 0)
+	err = d.Select(ctx, &entries, "SELECT * FROM gateway_multiparts")
```
nopcoder, line 121: we don't want to select all the database content at this point - we like to query the data and migrate each part - can even read the information directly into the proto message

pkg/gateway/multiparts/db_tracker.go
```diff
@@ -92,3 +103,39 @@ func (m *dbTracker) Delete(ctx context.Context, uploadID string) error {
...
+		return err
+	}
+	for i := range entries {
+		pr := protoFromMultipart(&entries[i]) // gosec error WA
+		value, err := proto.Marshal(pr)
+		key := []byte(kv.FormatPath(multipartPrefix, entries[i].UploadID))
+		if err != nil {
+			return err
+		}
+		err = je.Encode(kv.Entry{
```
nopcoder, line 132: The Go's json encoder produce a line based file? or just serialize all the message one by one?

pkg/kv/export/export.go
nopcoder, line 78: can use use different parameter name for 'path' - its a known package name usually.

pkg/kv/export/export_test.go
```diff
@@ -0,0 +1,48 @@
...
+	"github.com/treeverse/lakefs/pkg/kv/export"
+	"github.com/treeverse/lakefs/pkg/kv/kvtest"
+	_ "github.com/treeverse/lakefs/pkg/kv/mem"
+	"github.com/treeverse/lakefs/pkg/logging"
+)
+
+func TestImport(t *testing.T) {
+	ctx := context.Background()
+	logger := logging.Default()
+	store := kvtest.MakeStoreByName("mem", "")(t, context.Background())
```
nopcoder, line 20: Calling this function returns a factory - you can keep a reference
```
kvtest.MakeStoreByName("mem", "")
```
There is no need to create a factory every time you need a storage - you can even add a different package to create the test storage.
So in case we like to test using a different implementation we can switch in a single place

pkg/kv/export/utils.go
```diff
@@ -0,0 +1,71 @@
...
+			return nil
+		case err != nil:
+			return err
+		// Skip if the header is nil
+		case header == nil:
+			continue
+		}
+		path := filepath.Join(dir, header.Name) // nolint:gosec
+		if !strings.HasPrefix(path, filepath.Clean(dir)) {
+			logging.Default().WithError(ErrIllegalFilePath).Error("Illegal file path:", path)
```
nopcoder, line 41: As this function is very basic I would keep the logging to the caller and just return more information on the error

pkg/kv/export/utils.go
```diff
@@ -0,0 +1,71 @@
...
+					return err
+				}
+			}
+		case tar.TypeReg:
+			fd, err := os.OpenFile(path, os.O_CREATE|os.O_RDWR, os.FileMode(header.Mode))
+			if err != nil {
+				return err
+			}
+			for {
+				_, err := io.CopyN(fd, tr, 1024)
```
nopcoder, line 57: very small buffer

pkg/kv/export/utils.go
```diff
@@ -0,0 +1,71 @@
...
+				if err != nil {
+					if err == io.EOF {
+						break
+					}
+					return err
```
nopcoder, line 62: return an error while the file is still open

pkg/kv/export/utils.go
```diff
@@ -0,0 +1,71 @@
...
+	"strings"
+
+	"github.com/treeverse/lakefs/pkg/logging"
+)
+
+var ErrIllegalFilePath = errors.New("illegal file path")
+
+// Untar takes a destination dir and a reader to a tar file.
+// Looping over the tar file, extracting files into dir
+func Untar(dir string, r io.Reader) error {
```
nopcoder, line 19: Ignore if I wrote it before - but still think that this is too much at this point

nopcoder (changes requested):

pkg/kv/export/utils.go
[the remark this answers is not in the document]
N-o-Z: I'm going to remove it altogether since it's not part of the migration task

pkg/db/migration.go
```diff
@@ -228,3 +236,109 @@ func MigrateVersion(ctx context.Context, dbPool Database, params params.Database
 	}
 	return version, dirty, err
 }
+
+// TODO: (niro) Remove on feature complete
```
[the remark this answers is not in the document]
N-o-Z: The comment refers to the entire code block and not to a single function

pkg/kv/export/export.go
```diff
@@ -0,0 +1,141 @@
...
+func ImportFile(ctx context.Context, reader io.Reader, store kv.Store) error {
+	jd := json.NewDecoder(reader)
+	// Read header
+	if !jd.More() {
+		return ErrEmptyFile
+	}
+	var header Header
+	err := jd.Decode(&header)
+	// Decode does not return error on failure to Unmarshal
+	if err != nil || header == (Header{}) {
```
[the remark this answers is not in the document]
N-o-Z: 2. Already checked in ` !jd.More()`

pkg/db/migration.go
```diff
@@ -228,3 +236,109 @@ func MigrateVersion(ctx context.Context, dbPool Database, params params.Database
...
+	}
+	return err
+}
+
+type MigrateFunc func(db Database, writer io.Writer) error
+
+var (
+	kvPkgs     = make(map[string]MigrateFunc)
+	registerMu sync.RWMutex
+)
```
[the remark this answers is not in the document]
N-o-Z: 1. I've put all the migration code at the bottom of the module so we can remove it quickly afterwards and with as little conflicts as possible
2. ok

pkg/db/migration.go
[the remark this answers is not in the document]
N-o-Z: Done

pkg/db/migration.go
```diff
@@ -228,3 +237,109 @@ func MigrateVersion(ctx context.Context, dbPool Database, params params.Database
...
+	dbPool := BuildDatabaseConnection(ctx, dbParams)
+	defer dbPool.Close()
+	version, dirty, err := MigrateVersion(ctx, dbPool, dbParams)
+	if err != nil {
+		return err
+	}
+	if dirty {
+		return migrate.ErrDirty{Version: int(version)}
+	}
+	if kv.MigrateVersion != version {
```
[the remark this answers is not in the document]
itaiad200: IIUC `GetLastMigrationAvailable` is now wrong since it doesn't take kv special migration into consideration.

pkg/db/migration_test.go
```diff
@@ -11,7 +11,7 @@ import (
 func TestMigrations(t *testing.T) {
 	databaseURI, closer := testutil.GetDBInstance(pool)
 	defer closer()
-	err := db.MigrateUp(params.Database{ConnectionString: databaseURI})
+	err := db.MigrateUp(params.Database{ConnectionString: databaseURI}, true)
```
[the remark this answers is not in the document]
N-o-Z: Not sure - we should probably not touch this test as it checks for something that is not supported as far as I can tell (migrate down)

pkg/kv/export/export.go
```diff
@@ -0,0 +1,141 @@
...
+	err = validateHeader(header)
+	if err != nil {
+		return err
+	}
+
+	var entry kv.Entry
+	for jd.More() {
+		err = jd.Decode(&entry)
+		// Decode does not return error on failure to Unmarshal
+		if err != nil || entry.Key == nil || entry.Value == nil {
```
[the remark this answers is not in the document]
N-o-Z: We don't support nil values in the KV store as far as I understand

pkg/db/migration.go
```diff
@@ -228,3 +236,109 @@ func MigrateVersion(ctx context.Context, dbPool Database, params params.Database
...
+
+	// Import to KV Store
+	var g multierror.Group
+	for n, f := range kvPkgs {
+		name := n
+		migrateFunc := f
+		g.Go(func() error {
+			logging.Default().Info("Starting KV migration for package:", name)
+			buf := bytes.Buffer{}
+			err := migrateFunc(dbPool, &buf)
```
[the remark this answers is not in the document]
N-o-Z: Used it temporarily to see that everything works - changed to file

pkg/gateway/multiparts/db_tracker.go
N-o-Z: https://pkg.go.dev/encoding/json#Encoder.Encode

> ...Encode writes the JSON encoding of v to the stream, followed by a newline character....

pkg/db/migration.go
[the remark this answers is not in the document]
N-o-Z: Used it temporarily to see that everything works - changed to file

cmd/lakefs/cmd/migrate.go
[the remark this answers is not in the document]
N-o-Z: Changed it to config param

itaiad200:
@N-o-Z Just tested the e2e flow with the latest lakeFS build from this PR and it [passed](https://github.com/treeverse/lakeFS/runs/6442713233?check_suite_focus=true) :)

N-o-Z:
> @N-o-Z Just tested the e2e flow with the latest lakeFS build from this PR and it [passed](https://github.com/treeverse/lakeFS/runs/6442713233?check_suite_focus=true) :)

🥳🥳🥳

eden-ohana:
what will happen if the migration fails and some of the records are already written?

N-o-Z:
> what will happen if the migration fails and some of the records are already written?

@eden-ohana, basically as long as the migration was not completed successfully - the KV DB version will not be updated. If we try to start the migration process again, we check for the version and if it's not updated we drop the KV table and start from scratch

pkg/kv/import.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+// TODO: (niro) implement export
+
+// TODO: (niro) Make private after migration
+func Import(ctx context.Context, reader io.Reader, store Store) error {
+	jd := json.NewDecoder(reader)
+	// Read header
+	if !jd.More() {
+		return fmt.Errorf("%w: no content", ErrInvalidFormat)
```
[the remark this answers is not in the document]
N-o-Z: The problem here is that the file is empty, not a decoding error

pkg/kv/postgres/migrate.go
```diff
@@ -0,0 +1,172 @@
...
+
+	store, err := kv.Open(ctx, DriverName, dbParams.ConnectionString)
+	if err != nil {
+		return fmt.Errorf("opening kv store: %w", err)
+	}
+	defer store.Close()
+
+	// Check KV version before migration
+	version, err := kv.GetDBVersion(ctx, store)
+	if err != nil && errors.Is(err, kv.ErrMissingKey) {
```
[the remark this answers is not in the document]
N-o-Z: It's actually correct (the modified version). If there's no version entry we want to continue with the migration

pkg/db/kv_migration_test.go
```diff
@@ -0,0 +1,189 @@
...
+		key:   []byte("test"),
+		value: -1,
+	})
+	if err != nil {
+		log.Fatal("Failed to encode struct")
+	}
+	return nil
+}
+
+func migrateParallel(_ context.Context, _ *pgxpool.Pool, writer io.Writer) error {
```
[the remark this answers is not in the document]
N-o-Z: Fixed

pkg/kv/import.go
```diff
@@ -0,0 +1,83 @@
...
+	}
+	var header Header
+	err := jd.Decode(&header)
+	// Decode does not return error on failure to Unmarshal
+	if err != nil {
+		return fmt.Errorf("%w: error decoding header", err)
```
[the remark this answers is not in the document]
N-o-Z: Fixed

itai-david (approved):
LGTM
Thank you for addressing all the requests

pkg/kv/postgres/migrate.go
```diff
@@ -0,0 +1,181 @@
...
+	if err = os.RemoveAll(tmpDir); err != nil {
+		logger.Error("Failed to remove migration directory") // This should not fail the migration process
+	}
+	return nil
+}
+
+// validateVersion Check KV version before migration. Version exists and smaller than InitialMigrateVersion indicates
+// failed previous migration. In this case we drop the kv table and recreate it.
+// In case version is equal or bigger - it means we already performed a successful migration and there's nothing to do
+func validateVersion(ctx context.Context, store kv.Store) (bool, error) {
```
nopcoder, line 138: we are asking if migration completed. suggest a different name. verify func that tells me to drop requires reading all the comments.

pkg/kv/postgres/migrate.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	if dbParams.DropTables {
+		err = dropTables(ctx, dbPool, tables)
+		if err != nil {
+			return err
+		}
+	}
+	if err = os.RemoveAll(tmpDir); err != nil {
+		logger.Error("Failed to remove migration directory") // This should not fail the migration process
```
nopcoder, line 130: warn is enough as this is temp. add the dir name and the error to help

pkg/kv/postgres/migrate.go
```diff
@@ -0,0 +1,181 @@
...
+			}
+			_, err = fd.Seek(0, 0)
+			if err != nil {
+				return fmt.Errorf("failed seek file on package %s: %w", name, err)
+			}
+			err = kv.Import(ctx, fd, store)
+			if err != nil {
+				return fmt.Errorf("failed import on package %s: %w", name, err)
+			}
+			fileLog.Info("Successfully migrated package to KV")
```
nopcoder, line 108: consider adding the time it took

pkg/kv/postgres/migrate.go
```diff
@@ -0,0 +1,181 @@
...
+			fileLog := logging.Default().WithField("pkg_id", name)
+			fileLog.Info("Starting KV migration for package")
+			fd, err := os.CreateTemp(tmpDir, fmt.Sprintf("migrate_%s_", name))
+			if err != nil {
+				return fmt.Errorf("create temp file: %w", err)
+			}
+			defer fd.Close()
+			err = migrateFunc(ctx, dbPool, fd)
+			if err != nil {
+				fileLog.WithError(err).Error()
```
nopcoder, line 97: remove?

pkg/kv/postgres/migrate.go
```diff
@@ -0,0 +1,181 @@
...
+		panic(fmt.Sprintf("Package already registered: %s", name))
+	}
+	kvPkgs[name] = pkgMigrate{
+		Func:   f,
+		Tables: tables,
+	}
+}
+
+// UnRegisterAll remove all loaded migrate callbacks, used for test code.
+func UnRegisterAll() {
```
nopcoder, line 161: UnregisterAll

nopcoder (changes requested):

pkg/kv/postgres/migrate.go
N-o-Z: Done - the tmpDir is already a field in this log

pkg/testutil/db.go
```diff
@@ -243,3 +254,92 @@ func NewBlockAdapterByType(t testing.TB, translator block.UploadIDTranslator, bl
 		return mem.New(mem.WithTranslator(translator))
 	}
 }
+
+// migrate functions for test scenarios
```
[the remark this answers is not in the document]
N-o-Z: All the functions below are intended for reuse. All but one actually are being reused at the moment

pkg/kv/import.go
```diff
@@ -0,0 +1,76 @@
...
+	// Read header
+	var header Header
+	if err := jd.Decode(&header); err != nil {
+		if errors.Is(err, io.EOF) {
+			return fmt.Errorf("empty file: %w", ErrInvalidFormat)
+		} else {
+			return fmt.Errorf("decoding header: %w", err)
+		}
+	}
+	// Decode does not return error on failure to Unmarshal
```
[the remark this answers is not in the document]
N-o-Z: That's what I meant - rephrasing

pkg/db/kv_migration_test.go
```diff
@@ -0,0 +1,75 @@
...
+		t.Run(tt.name, func(t *testing.T) {
+			kvpg.UnregisterAll()
+			for n, m := range tt.migrations {
+				kvpg.RegisterMigrate(n, m, nil)
+			}
+			err = db.MigrateUp(dbParams)
+			require.ErrorIs(t, err, tt.err)
+
+			if tt.err == nil {
+				data, err := kvStore.Get(ctx, []byte(kv.DBSchemaVersionKey))
```
nopcoder, line 65: We can call the public func `GetDBSchemaVersion` and verify the version and that err is nil?

pkg/gateway/multiparts/kv_migrate_test.go
```diff
@@ -0,0 +1,69 @@
...
+	testutil.MustDo(t, "Import file", kv.Import(ctx, &buf, kvStore))
+	kvTracker := multiparts.NewTracker(kv.StoreMessage{Store: kvStore})
+	for _, entry := range data {
+		m, err := kvTracker.Get(ctx, entry.UploadID)
+		require.NoError(t, err)
+		require.Equal(t, *m, entry)
+	}
+}
+
+func createData(t *testing.T, ctx context.Context, tracker multiparts.Tracker, size int) []multiparts.MultipartUpload {
```
nopcoder, line 41: createMigrateTestData? needs a better name as a global func in the scope of multiparts test package

pkg/kv/import.go
```diff
@@ -0,0 +1,70 @@
...
+	if header == (Header{}) {
+		return fmt.Errorf("bad header format: %w", ErrInvalidFormat)
+	}
+
+	// TODO(niro): Add validation the header in the future
+	logging.Default().WithFields(logging.Fields{
+		"package name":   header.PackageName,
+		"lakeFS version": header.LakeFSVersion,
+		"DB Version":     header.DBSchemaVersion,
+		"CreatedAt":      header.CreatedAt,
```
nopcoder, line 45: prefer to process snake case later when parse structure log

```suggestion
		"package_name":   header.PackageName,
		"lakefs_version": header.LakeFSVersion,
		"db_schema_version":     header.DBSchemaVersion,
		"created_at":      header.CreatedAt,
```

pkg/kv/postgres/migrate.go
```diff
@@ -0,0 +1,186 @@
...
+	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/treeverse/lakefs/pkg/db/params"
+	"github.com/treeverse/lakefs/pkg/logging"
+)
+
+type MigrateFunc func(ctx context.Context, db *pgxpool.Pool, writer io.Writer) error
+
+var (
+	kvPkgs     = make(map[string]pkgMigrate)
+	registerMu sync.RWMutex
```
nopcoder, line 28:
```suggestion
	kvPkgs     = make(map[string]pkgMigrate)
	kvPkgsMu sync.RWMutex
```

pkg/kv/postgres/migrate.go
```diff
@@ -0,0 +1,186 @@
...
+
+type pkgMigrate struct {
+	Func   MigrateFunc
+	Tables []string
+}
+
+func timeTrack(start time.Time, logger logging.Logger, name string) {
+	elapsed := time.Since(start)
+	logger.Info(fmt.Sprintf("%s took %s", name, elapsed))
+}
```
nopcoder, line 39: 1. our logging package reports the calling func - in this case, I would not encapsulate the call to this func as defer helps you, we will log the time took as part of a complete operation and not a failure (if you like that - sometimes you don't)
2. logger.Infof preferred in this case

pkg/kv/postgres/migrate.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+	err = g.Wait().ErrorOrNil()
+	if err != nil {
+		return err
+	}
+
+	// Update migrate version
+	err = store.Set(ctx, []byte(kv.DBSchemaVersionKey), []byte(strconv.Itoa(kv.InitialMigrateVersion)))
+	if err != nil {
+		return fmt.Errorf("failed setting migrate version: %w", err)
```
nopcoder, line 125:
```suggestion
		return fmt.Errorf("failed setting migrate schema version: %w", err)
```

nopcoder (approved):
Thanks, it lgtm - there is an extra file `pkg/kv/export/test_data/test.tar.gz` that should be removed and some minor comments to consider.

pkg/db/kv_migration_test.go
N-o-Z: Done

PR #3342, opened 2022-05-10
UI: fix create repository focus on name

nopcoder:
### Current behaviour

While clicking Create Reposioty button.
A modal dialog with repository information will popup.
1. The focus is not set on the repository name, so trying to write something will do nothing.
2. Using tab switching focus will go over the dialog inputs - one of the input is the information link (i).

### New behaviour

In order to have better input experience:
1. Focus is set to the repository name when the dialog apper.
2. Using tab to switch focus will not go over the hyperlink of the information icon.


N-o-Z (changes requested):
@nopcoder Since I'm not so familiar with the UI, I didn't quite understand what exactly the issue is. I tried to compare the form from the playground with running a local server but I didn't see any difference.
Can you please add some explanation to the description of the PR? Thanks!

nopcoder:
@N-o-Z sure, update description soon

N-o-Z (approved):
👍🏽

Issue #3244, opened 2022-04-20
Document user IDs for users with emails


arielshaqed:
I approved #3220 to allow us to move forward on this feature.  However I continue not to understand what goes into a "user ID" field of a user defined with an email.  As such, it is only a matter of time before I or someone else get confused and introduce a usability or a security bug.

Specifically, the best comment about what goes in this field is this:

> the unique identifier for a user was the username or display name (sometimes referred to as user id as well). In the invite user process all we know about the new user is his email. As such we don't have a better uniquely identify the user only by the given email. In order to retain it's uniqueness, we must set it all to lowercase to avoid having the same email twice in the system with different caps...

This _does not_ say what goes into the field.  It also does not explain why it makes _unique_ emails collide by lowercasing them.  Please document this field, as a start in documenting the entire feature.

Thanks!

johnnyaug:
Removed this ambiguity in: #3703

PR #3351, opened 2022-05-11
Evict user from the cache once a user reset his password

### LINKED ISSUE #3331: Evict user from cache on reset password

shimi9276:
This avoids conflict of the old password still being in the cache, resulting in failure to login with the new credentials.
Closes #3331

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -3223,6 +3223,8 @@ func (c *Controller) UpdatePassword(w http.ResponseWriter, r *http.Request, body
 		writeError(w, http.StatusInternalServerError, http.StatusText(http.StatusInternalServerError))
 		return
 	}
+	// RemoveUserFromCache to clean the cache from the old password
+	c.Auth.RemoveUserFromCache(user.Username)
 	w.WriteHeader(http.StatusCreated)
 }
 
--- a/pkg/auth/cache.go
+++ b/pkg/auth/cache.go
@@ -16,6 +16,8 @@ type Cache interface {
 	GetUser(username string, setFn UserSetFn) (*model.User, error)
 	GetUserByID(userID int64, setFn UserSetFn) (*model.User, error)
 	GetUserPolicies(userID string, setFn UserPoliciesSetFn) ([]*model.Policy, error)
+	RemoveUser(username string)
+	RemoveUserByID(userID int64)
 }
 
 type LRUCache struct {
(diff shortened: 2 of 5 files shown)


pkg/api/controller.go
```diff
@@ -3223,6 +3223,8 @@ func (c *Controller) UpdatePassword(w http.ResponseWriter, r *http.Request, body
 		writeError(w, http.StatusInternalServerError, http.StatusText(http.StatusInternalServerError))
 		return
 	}
+	// RemoveUserFromCache to clean the cache from the old password
+	c.Auth.RemoveUserFromCache(user.Username)
```
nopcoder, line 3227: I think that if we call it part of update password
- it will prevent future issues as it came to solve this specific problem
- the user that call the auth API doesn't aware of any cache, and now we expose implementation detail as part of the interface

nopcoder (changes requested):
We need to make sure that the one that manages the cache will need to update/invalidate it when needed as the user of the API doesn't know about any internal cache implementation.

arielshaqed (changes requested):
As coded I believe that this feature is broken -- see comment inline.

Before coding the solution, I am afraid we will need to\ write up a proper design.  It should address _both_ of these issues:

* **Connectivity**.  After a user changes their password, they cannot use their new password to log into some installations.  This might also apply to first-time login attempts (I don't remember if you ended up sticking a redundant random unusable password for users until their first login; if you did then that would sometimes apply to first-time logins).
* **Security**.  After a user changes their password, their old password might still be usable to log into some installations.

Note that when using central auth, all users exist on multiple installations.  And each installation can have multiple instances.

arielshaqed:
After looking through the PR again, I am afraid that I am not able to understand the functioning of this PR.  As such, I will not be able to approve this PR in its current form.

I am lacking a design that explains the following:

1. Difference between `Username` and `Email` in the [`User` struct](https://github.com/treeverse/lakeFS/blob/3b541f3327eea68e0f33671d3f348aef87ba0018/pkg/auth/model/model.go#L29-L41).  See #3244, which I previously opened to improve understanding of this.

   In this PR, it _appears_ that the user being fetched and then forgotten from the cache is fetched by email in order to determine their username.  However in previous PRs it appeared that these are actually equal.  If this were so, then fetching the user-to-be-forgotten could simply be removed from the code.  The fact that this cannot work means that I do not understand the code.
1. Caching during the "reset password" flow when a user forgets their password.  As explained in my previous comment, I continue to believe that this feature is broken in all cases where there is more than a single lakeFS daemon.

   This of course has implications both for security and for usability.
1. Deep access into the authorization subsystem from the controller.  This violates what I best know as the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter).  While I do not ascribe extreme value to this law, _in this case_ it it relevant, as can be seen from the need to duplicate it in both AuthService implementations.

I understand the blocker to doing all of these is a desire to "move quickly".  As should be clear by now, I will not be able to understand this system when it is generated at this velocity.  I suspect that the same will be true of other on the team, but obviously I cannot prove it.  As such, I can see two ways out of being stuck here (there may be others):

* **Write up a design.**  I believe that this will _save_ time.  I cannot predict the future, but I believe I can confidently state of the past that up until now the lack of a design has slowed down every single review -- and in many cases _caused_ the need for additional PRs.  This particular PR is something that I believe would come up in a design review.
* **Remove from from reviewing this code.**  I am unable to approve this code.  If we decide to go this road, please dismiss my review (with a full explanation of why my comments should be skipped) and find some other additional reviewer for this code.

Thanks!

shimi9276:
Solved with [this](https://github.com/treeverse/lakeFS/pull/3407) PR

PR #3355, opened 2022-05-12
Succeed on required Go tests when we do not need to run them

### LINKED ISSUE #3350: Required jobs are not running on documentation changes and block merge
When performing documentation changes, the required jobs: "Run Linters and Checkers" "Run Go tests" are skipped due to workflow configuration:
```
name: Test
on:
  push:
    branches:
      - master
  pull_request:
    paths-ignore:
      - "*.md"
      - "docs/**"
      - "design/**"
```
Since these jobs are required for merge, merge is blocked.
The problem is, it's currently not possible to mark jobs as not required only for a specific path. Need to find a way to work around it

arielshaqed:
See
https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks.

Fixes #3350.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -90,3 +90,27 @@ jobs:
       #   with:
       #     files: ./cover.out,api/cover.out
       #     fail_ci_if_error: false
+
+---
+# Succeed on these required tests when we do not need to run them.  See
+# https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/troubleshooting-required-status-checks#handling-skipped-but-required-checks
+name: Test
+on:
+  push:
+    paths:
+      - "*.md"
+      - "docs/**"
+      - "design/**"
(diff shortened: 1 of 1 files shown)


arielshaqed:
## How was this tested?

Sad to say I  easily cannot tested this before pulling to trunk.  But once we do that, I will send a docs-only PR to verify that merging will no longer be blocked.

arielshaqed:
> @arielshaqed how make sure that if we change code and docs this change will not turn green when the actual test will fail? it will verify that all required tests are green?

Oh wow, you're right of course -- docs are at least confusing or maybe confused.

I would *hope* that GitHub will somehow fail if 2 actions succeed, but I doubt they did that (actually, I doubt whether it is even possible to implement without a rather fancy KV...).

I shall not pull for now, and instead ask them.

arielshaqed:
Not planned.

PR #3356, opened 2022-05-12
Migration system test

### LINKED ISSUE #3310: Esti system test for DB migration to KV Store
* Esti system test to verify a system wide migration to KV, for correctness
  * Can also be used to cover the migration from the performance aspect by (currently manually) checking the time it takes
  * As the migration process will increase over time, as we transfer more parts to KV, the time to migrate cannot be well defined
  * We will use this to identify "unreasonable" migration time

* The main idea is to have a system running on SQL DB, performing various lakefs actions [TBD] and bringing lakefs to a certain state. Then lakefs is stopped, a migration to KV takes place and lakefs is restarted with KV enabled. The testing flow should continue at this point, seamlessly
* **Note**: Currently, Esti tests rely on lakefs being up and running prior to the test itself. This task needs to add support for stopping/starting lakefs as part of the test
* The time of the migration process should be "benchmarked" as part of this test. Initially, this can be done by manually verifying the migration time
* What is the sufficient dataset prior to migration?

itaiad200:
closes #3310

### DIFF
--- a/esti/migrate_test.go
+++ b/esti/migrate_test.go
@@ -2,6 +2,7 @@ package esti
 
 import (
 	"context"
+	"encoding/base64"
 	"encoding/json"
 	"net/http"
 	"testing"
@@ -64,6 +65,7 @@ type multipartState struct {
 	Repo           string                         `json:"repo"`
 	Info           s3.CreateMultipartUploadOutput `json:"info"`
 	CompletedParts []*s3.CompletedPart            `json:"completed_parts"`
+	Content        string                         `json:"state"`
 }
 
 const (
@@ -97,6 +99,7 @@ func testPreMigrateMultipart(t *testing.T) {
 	state.Multipart.Repo = repo
 	state.Multipart.Info = *resp
 	state.Multipart.CompletedParts = completedParts
+	state.Multipart.Content = base64.StdEncoding.EncodeToString(partsConcat)
 }
 
 func testPostMigrateMultipart(t *testing.T) {
@@ -110,5 +113,8 @@ func testPostMigrateMultipart(t *testing.T) {
 	getResp, err := client.GetObjectWithResponse(ctx, state.Multipart.Repo, mainBranch, &api.GetObjectParams{Path: migrateMultipartFile})
 	require.NoError(t, err, "failed to get object")
 	require.Equal(t, http.StatusOK, getResp.StatusCode())
-	require.Equal(t, multipartNumberOfParts*multipartPartSize, len(getResp.Body))
+
+	contentBytes, err := base64.StdEncoding.DecodeString(state.Multipart.Content)
+	require.NoError(t, err, "failed to decode error")
+	require.Equal(t, contentBytes, getResp.Body)
 }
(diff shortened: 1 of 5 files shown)


esti/migrate_test.go
```diff
@@ -0,0 +1,128 @@
...
+	if postMigrate {
+		postMigrateTests(t)
+	} else {
+		preMigrateTests(t)
+	}
+}
+
+func preMigrateTests(t *testing.T) {
+	// all pre tests execution
+	t.Run("TestPreMigrateMultipart", testPreMigrateMultipart)
```
N-o-Z, line 30: Does it make sense to continue the pre migrate tests if one fails?

esti/migrate_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	saveStateInLakeFS(t)
+}
+
+func postMigrateTests(t *testing.T) {
```
N-o-Z, line 35: Same as the comment before but for the general usecase.
If preMigrateTests failed - there's no point in running postMigrateTests.
We need to somehow indicate the pre tests failed (maybe via the state file?)

esti/migrate_test.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+const (
+	migrateMultipartFile     = "multipart_file"
+	migrateMultipartFilepath = mainBranch + "/" + migrateMultipartFile
+	migrateStateRepoName     = "migrate"
+	migrateStateObjectPath   = "state.json"
+)
+
+func testPreMigrateMultipart(t *testing.T) {
```
N-o-Z, line 86: Just an idea - not sure how much of a usecase it is:
How about uploading only some of the parts in the old DB format, migrating and then uploading the rest?

esti/migrate_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+	completedParts := uploadMultipartParts(t, logger, resp, parts)
+
+	state.Multipart.Repo = repo
+	state.Multipart.Info = *resp
+	state.Multipart.CompletedParts = completedParts
+	state.Multipart.Content = base64.StdEncoding.EncodeToString(partsConcat)
+}
+
+func testPostMigrateMultipart(t *testing.T) {
```
N-o-Z, line 113: I would arrange all the test functions together and order them hierarchically and put the variables declaration at the top of the module

N-o-Z (changes requested):
Very cool! 
Some comments

esti/migrate_test.go
itaiad200: We'll have to answer that when we'll have more than a single migration test.

esti/migrate_test.go
itaiad200: Right - If Pre test fails -> Esti fails -> flow stops and we'll never reach the post test.

esti/migrate_test.go
itaiad200: Done.

N-o-Z (approved):
Cool!

PR #3358, opened 2022-05-14
LAKECTL_INTERACTIVE env can control lakeCTL on/off terminal output

### LINKED ISSUE #3026: Can't run runShellCommand func with terminal mode
For some of our system-tests, we would like to run a lakectl command on terminal mode, usually on cases where there's a difference in output when running the command in terminal mode (cat-sst command for example doesn't print headers when not running in terminal mode).  Currently, there's a problem with running the `runShellCommand` function with the `isTerminal` flag as true

nopcoder:
Using `LAKECTL_INTERACTIVE` environment variable boolean (already supported), we can force lakectl terminal (interactive on) or no terminal (interactive off).

By setting a value, it will override the default behaviour where we try to identify if the terminal is currently attached to our standard output.

Fix #3026

### DIFF
--- a/cmd/lakectl/cmd/common_helpers.go
+++ b/cmd/lakectl/cmd/common_helpers.go
@@ -11,6 +11,7 @@ import (
 	"os"
 	"reflect"
 	"regexp"
+	"strconv"
 	"strings"
 	"text/template"
 	"time"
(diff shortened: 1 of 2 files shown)


idanovo:
@nopcoder thanks for doing this! 
@talSofer it means we can now test refs-dump command. (and others) properly. If you remember, we had a problem testing the output of it and check if we are missing a field on the dump output

idanovo:
@nopcoder do you think a test that checks the output of a simple command with and without the flag would help us here?

nopcoder:
> @nopcoder do you think a test that checks the output of a simple command with and without the flag would help us here?

The plan is to use this feature in all the tests that verify the lakectl output in the terminal format.

idanovo (approved):

PR #3359, opened 2022-05-14
Show skipped hooks for failed action run

### LINKED ISSUE #3098: Skipped hooks should be visible in the Action Run view
When some hook fails during an Action Run, the next hooks are not executed.
The view should explicitly mention the those hooks were skipped. Currently, the view simply doesn't display them.

tshradheya:
Fixes #3098 

image

### DIFF
--- a/pkg/actions/service.go
+++ b/pkg/actions/service.go
@@ -306,18 +306,15 @@ func buildRunManifestFromTasks(record graveler.HookRecord, tasks [][]*Task) RunM
 				ActionName: task.Action.Name,
 				StartTime:  task.StartTime,
 				EndTime:    task.EndTime,
-				Passed:     task.Err == nil && !task.StartTime.IsZero(),
+				Passed:     task.Err == nil && !task.StartTime.IsZero(), // mark skipper tasks as failed
 			})
-			// only check for non-skipped tasks
-			if !task.StartTime.IsZero() {
-				// keep min run start time
-				if manifest.Run.StartTime.IsZero() || task.StartTime.Before(manifest.Run.StartTime) {
-					manifest.Run.StartTime = task.StartTime
-				}
-				// keep max run end time
-				if manifest.Run.EndTime.IsZero() || task.EndTime.After(manifest.Run.EndTime) {
-					manifest.Run.EndTime = task.EndTime
-				}
+			// keep min run start time using non-skipper tasks
+			if manifest.Run.StartTime.IsZero() || (!task.StartTime.IsZero() && task.StartTime.Before(manifest.Run.StartTime)) {
+				manifest.Run.StartTime = task.StartTime
+			}
+			// keep max run end time
+			if manifest.Run.EndTime.IsZero() || task.EndTime.After(manifest.Run.EndTime) {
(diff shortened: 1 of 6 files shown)


pkg/actions/service.go
```diff
@@ -310,10 +306,10 @@ func buildRunManifestFromTasks(record graveler.HookRecord, tasks [][]*Task) RunM
 				ActionName: task.Action.Name,
 				StartTime:  task.StartTime,
 				EndTime:    task.EndTime,
-				Passed:     task.Err == nil,
+				Passed:     task.Err == nil && !task.StartTime.IsZero(), // mark skipped tasks as failed
 			})
-			// keep min run start time
-			if manifest.Run.StartTime.IsZero() || task.StartTime.Before(manifest.Run.StartTime) {
+			// keep min run start time using non-skipped tasks
+			if manifest.Run.StartTime.IsZero() || (!task.StartTime.IsZero() && task.StartTime.Before(manifest.Run.StartTime)) {
```
nopcoder, line 312: 1. Using a variable called 'taskStarted' set by '!manifest.Run.StartTime.IsZero()' it can simplify the code and reduce the number of checks
2. Previous commit also skipped the run end time

pkg/api/controller.go
```diff
@@ -1532,9 +1533,12 @@ func (c *Controller) ListRunHooks(w http.ResponseWriter, r *http.Request, reposi
 			StartTime: val.StartTime,
 			EndTime:   &val.EndTime,
```
nopcoder, line 1534: This is also true before your PR - but because the task didn't start and end time is optional, we should keep it nil.

nopcoder (changes requested):
Thanks for your contribution! Added my review comments - also the code needs to handle how 'lakectl actions' handles the new status.

pkg/actions/service.go
tshradheya: 1. Do you mean `taskStarted` can be set by `!task.StartTime.IsZero()`
2. I am not sure what you meant by previous commit?

tshradheya:
@nopcoder addressed the changes for `lakectl action` command. Please help take a look again

pkg/actions/service.go
nopcoder: 1. Yes
2. Previous commit in this pr, you checked if the task didn't start and skipped the set of start and end time

tshradheya: For 2. I realized that I don't need to do for endtime because when the condition is to check .After so it won't affect it

nopcoder (approved):
LGTM Thanks again!
I see we still have an issue with rendering the start time in case of zero value. I think it is related to the server-side loading it from the database.
Will address it in a different issue.

tshradheya:
Hi, can I know what's the issue with rendering the start time currently? I may have missed that

itaiad200:
Thanks for the awesome contribution @tshradheya ! @nopcoder can we merge this? I see that tests were skipped, should we run them in another branch?

nopcoder:
Thanks @itaiad200 looks all green! merge it

PR #3361, opened 2022-05-15
Fix get user by email should have different cache instance

nopcoder:
Get user by email used the user's cache while loading user information by email.
This PR fix this behaviour, as the email and the username are two different identifiers and should use different cache instances.


guy-har (approved):
Thanks!

PR #3368, opened 2022-05-16
Allow lakeFS to serve GUI from other directory

arielshaqed:
lakeFS has always served a GUI built into it.  Now when configuration option
`ui.dir` has a value, serve the UI stored there.

This allows GUI plugin-in: copy the `webui/` directory, modify it as you see
fit, then `npm run build` and point `ui.dir` at `webui-copy/dist`.  I tested
it by doing just that and modifying the CSS there.

### DIFF
--- a/pkg/api/ui_handler.go
+++ b/pkg/api/ui_handler.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io/fs"
 	"net/http"
+	"os"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


arielshaqed:
(Draft because PoC.  You can review but not sure we want to pull!)

pkg/api/ui_handler.go
```diff
@@ -16,11 +17,19 @@ import (
...
+		err     error
+	)
+	if fsDir == nil {
+		content, err = fs.Sub(webui.Content, "dist")
+		if err != nil {
+			// embedded UI content is missing
+			panic(err)
+		}
+	} else {
+		content = os.DirFS(*fsDir)
```
nopcoder, line 32: I don't know if there is a difference (from the code it looks like both works) - usually the classic implementation uses `http.Dir`

pkg/config/template.go
```diff
@@ -41,6 +41,10 @@ type configuration struct {
 		Enabled bool `mapstructure:"enabled"`
 	}
 
+	UI struct {
+		Dir *string
```
nopcoder, line 45: Can we use a name like `DocumentRoot` or `Directory` or `Root`?

nopcoder (approved):
Looks good! a quick way to enable alertnative UI or no UI :)

arielshaqed:
@eladlachmi: after your changes to #3907, can I close this?

eladlachmi:
@arielshaqed Yes, thank you!

arielshaqed:
> @eladlachmi: after your changes in #3907, can I close this?

> @arielshaqed Yes, thank you!

:-) 🔐

PR #3370, opened 2022-05-16
Improve example in java client README.md

happytomatoe:
As it's not clear from documentation how to use java api I improved example


johnnyaug:
Hey @happytomatoe! Thank you for taking the time to open this PR.
Unfortunately, this file is generated by [OpenAPI Generator](https://openapi-generator.tech/), and therefore we cannot edit it at this point. 
I agree that the documentation for this client can be better, and in the future we hope to add some better documentation on top of it.

PR #3373, opened 2022-05-17
Initial suggestion of a template server

arielshaqed:
Useful basis for decreased TTV, more maintainable Spark clients, and more
configuration of detached GUIs.

### DIFF
--- a/design/open/template-server/template-server.md
+++ b/design/open/template-server/template-server.md
@@ -107,6 +107,7 @@ spark.hadoop.fs.lakefs.impl=io.lakefs.LakeFSFileSystem
 {{with $creds := new_credentials}}
 spark.hadoop.fs.lakefs.access_key={{$creds.ID}}
 spark.hadoop.fs.lakefs.secret_key={{$creds.Secret}}
+{{end}}
 spark.hadoop.fs.lakefs.endpoint={{"local.templates.lakefs_url" | config}}
 # Or pass something from "local" to give users a different key.  In any case 
 # requires IAM authorization to fetch.
--- a/design/open/template-server/template-server.md
+++ b/design/open/template-server/template-server.md
@@ -24,26 +24,66 @@ Among the planned applications:
 ## Template format
 
 We use Golang [text/template][text/template] to expand most templates.
-For browsers (based on filename, e.g. `*.html.tt`) we instead must use
+For browsers (based on filename, e.g. `*.html`) we instead must use
 [html/template][html/template] for safety.
 
-The template is read by lakeFS from a specific (configured) prefix, by
-default `lakefs://lakefs-templates/main`.  IAM authorizes users to see
-this template, at which point expansion can start.  We will supply the
-following replacements (at least):
+Accesses to http://<lakefs>/api/v1/templates/some/path read the template
+from `some/path`.
 
-| template function | value source              | IAM permissions                              |
-|:------------------|:--------------------------|:---------------------------------------------|
-| config            | lakeFS configuration[^1]  | `arn:lakefs:template:::config/path/to/field` |
-| object            | (small) lakeFS object     | IAM read permission for that object          |
-| querystring       | URL query string[^2]      | (none)                                       |
-| contenttype       | none; set `Content-Type:` | (none)                                       |
+The template is read by lakeFS from a specific (configured) prefix, by
+default `lakefs://lakefs-templates/`.  We might decide to support
+various object storage types; `lakefs` gives the best management and lineage
+capabilities, but possible supporting the production blockstore types `s3`,
+`azure` and `gcs` will simplify deployment of some lakeFS deployments.
+
+If storing on lakeFS, IAM authorizes users to see this template by
+`fs:ReadObject`, at which point expansion can start.  If we want to store
+templates on the blockstore types, we will probably need to add a new IAM
+action type for template expansion, or alternatively overload
+`fs:ReadObject` to support off-lakeFS URIs.  We will supply the following
+replacements (at least):
+
+| template function | value source                        | IAM permissions                              |
+|:------------------|:------------------------------------|:---------------------------------------------|
+| config            | exportable lakeFS configuration[^1] | `arn:lakefs:template:::config/path/to/field` |
+| object            | (small) object on lakeFS            | IAM read permission for that object          |
+| contenttype       | none; set `Content-Type:`           | (none)                                       |
+
+A configuration variable will be "exportable" if its struct field is tagged
+`` `export:yes` ``, or (if implementing that is too hard) if its type
+implements
+
+```go
+interface Exportable {
+	Export() string
+}
+```
+
+IAM is still checked for exportable configuration variables (which will be
+ON by default for users, but can be removed).
 
 Adding a freeform dictionary to our config will allow admins to set up
 any needed configuration.
 
-We shall also supply the user object in `.user`.  We will probably add
-new functions or objects.
+We shall supply the user object in `.user` and the parsed query args in
+`.query`[^2], to allow conditional operations.  We will probably add new
+functions or objects.
+
+## Template expansion flow
+
+#### _:warning: This flow assumes templates stored on lakeFS. :warning:_
+
+1. User accesses the template
+   http://<lakefs>/api/v1/templates/main/expand/me.
+1. lakeFS expands to a lakeFS path (by default this is
+   lakefs://lakefs-templates/main/expand/me).
+1. lakeFS uses IAM to verify user has `fs:ReadObject` permission on this
+   object.
+1. Object is parsed into a text/template.
+1. lakeFS creates template
+   [`Funcs`](https://pkg.go.dev/text/template#FuncMap) that will check IAM
+   permissions as required for expansion.
+1. Template gets expanded and returned to user.
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


Jonathan-Rosenberg (commented):
Do you mind adding some kind of a flow chart to show the whole process of what you're describing here?
And just to clarify, the server will hold the templates that were configured by the administrators (and might have some predefined-ready-to-roll templates that we'll offer?), and the clients of the server will ask for the templates they need and populate them when they get them back?

design/open/template-server/template-server.md
```diff
@@ -0,0 +1,77 @@
...
+
+(All examples subject to change as we define object format, and decide
+what functions are allowed.)
+
+Set a template that looks like this:
+
+```conf
+spark.hadoop.fs.lakefs.impl=io.lakefs.LakeFSFileSystem
+spark.hadoop.fs.lakefs.access_key={{.user.credentials.access_key.id}}
+spark.hadoop.fs.lakefs.secret_key={{.user.credentials.access_key.secret}}
```
johnnyaug, line 58: Not sure it's a good idea to expose the secrets through templating

design/open/template-server/template-server.md
```diff
@@ -0,0 +1,77 @@
...
+  which to receive by path.
+  
+## Template format
+
+We use Golang [text/template][text/template] to expand most templates.
+For browsers (based on filename, e.g. `*.html.tt`) we instead must use
+[html/template][html/template] for safety.
+
+The template is read by lakeFS from a specific (configured) prefix, by
+default `lakefs://lakefs-templates/main`.  IAM authorizes users to see
```
johnnyaug, line 31: I'm not sure what we gain by storing the templates in lakeFS.
This makes it harder for us to create built-in templates.
For example: there is going to be an option to create a new repository pre-configured to be connected to Spark.
So with this suggestion, we will need to pre-upload the template to an existing repository or to the new one, just for the user to get the templated Spark snippet.

@arielshaqed, WDYT?

arielshaqed:
Thanks; PTAL!

design/open/template-server/template-server.md
```diff
@@ -0,0 +1,105 @@
...
+If storing on lakeFS, IAM authorizes users to see this template by
+`fs:ReadObject`, at which point expansion can start.  If we want to store
+templates on the blockstore types, we will probably need to add a new IAM
+action type for template expansion, or alternatively overload
+`fs:ReadObject` to support off-lakeFS URIs.  We will supply the following
+replacements (at least):
+
+| template function | value source              | IAM permissions                              |
+|:------------------|:--------------------------|:---------------------------------------------|
+| config            | lakeFS configuration[^1]  | `arn:lakefs:template:::config/path/to/field` |
```
johnnyaug, line 48: I think there should be a global, fixed whitelist of configurations we allow to expose in this way.
Some values should never be exposed, for example the installation's encryption key.

johnnyaug (changes requested):

arielshaqed (commented):
Sorry, managed not to post my 2 pending comments!

design/open/template-server/template-server.md
johnnyaug: Thanks for clarifying about these being taken from the request.

So yes, let's discuss returning the underlying storage credentials from the configuration: I'm not referring specifically to templating - I think in general we shouldn't return them from any API endpoint.

design/open/template-server/template-server.md
arielshaqed: That makes sense!

I can think of 2 ways to do this:

1. Have a list of configuration properties that may be exported in a configuration setting.
1. Specify an additional (Go...) tag on the `Config` struct that allows a configuration setting to be exposed.  I think I prefer this option: it is the most static way possible, but also easy enough for developers to add more properties if needed.

In any case, I think we should **still** enforce IAM for that setting.  Controlling everything by permissions on the path to the template is too hard to get right when you do care, and we can just add a "everything is allowed" default policy and put that on users.

design/open/template-server/template-server.md
johnnyaug: My only concern was that taking templates from lakeFS complicates our first time-to-value tasks - namely showing the user Spark snippets when creating a repository. If @Jonathan-Rosenberg is happy with the current design, I am too.

design/open/template-server/template-server.md
johnnyaug: Agreed

pkg/config/template.go
```diff
@@ -31,8 +31,7 @@ type S3AuthInfo struct {
 }
 
 // Output struct of configuration, used to validate.  If you read a key using a viper accessor
-// rather than accessing a field of this struct, that key will *not* be validated.  So don't
-// do that.
+// rather than accessing a field of this struct, that key will *not* be validated.  So don'// do that.
```
Jonathan-Rosenberg, line 34: ?

arielshaqed: Whoops, sorry!
Reverted.

arielshaqed:
Added a `new_credentials` template function, as discussed.

design/open/template-server/template-server.md
```diff
@@ -0,0 +1,130 @@
...
+(All examples subject to change as we define object format, and decide
+what functions are allowed.)
+
+Set a template that looks like this:
+
+```conf
+spark.hadoop.fs.lakefs.impl=io.lakefs.LakeFSFileSystem
+{{with $creds := new_credentials}}
+spark.hadoop.fs.lakefs.access_key={{$creds.ID}}
+spark.hadoop.fs.lakefs.secret_key={{$creds.Secret}}
```
Jonathan-Rosenberg, line 109:
```suggestion
spark.hadoop.fs.lakefs.secret_key={{$creds.Secret}}
{{end}}
```

talSofer (commented):
Very interesting, thanks! 
Added a few questions

design/open/template-server/template-server.md
```diff
@@ -0,0 +1,130 @@
...
+Adding a freeform dictionary to our config will allow admins to set up
+any needed configuration.
+
+We shall supply the user object in `.user` and the parsed query args in
+`.query`[^2], to allow conditional operations.  We will probably add new
+functions or objects.
+
+## Template expansion flow
+
+#### _:warning: This flow assumes templates stored on lakeFS. :warning:_
```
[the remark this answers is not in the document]
arielshaqed: Templates are in a repo.  I don't know why we would want to hide it.

In general I believe that **everything should be versioned**, so everything should live in a repo.  Among the many advantages of storing in a repo, for templates we have:

1. **Ease of configuration.** No need to overload the block adapter to read templates.

   In particular, note that so far every time we decided to store files on an object store outside of lakeFS, the implementation contained serious mistakes.
1. **Reproducibility and lineage.** If templates change behind the user's back, they have no way of understanding why they got a particular expansion at a particular date.
1. **Adaptability.** I want admins to be free to change templates or add new ones.
1. **Ease of development.** When developing and testing improved or new templates, having a branch makes everything easier.

I would further claim that, as people who work on a versioning product, the burden of proof is with those who want _not_ to store things inside the product, not with those who _do_.

design/open/template-server/template-server.md
```diff
@@ -0,0 +1,130 @@
...
+
+(All examples subject to change as we define object format, and decide
+what functions are allowed.)
+
+Set a template that looks like this:
+
+```conf
+spark.hadoop.fs.lakefs.impl=io.lakefs.LakeFSFileSystem
+{{with $creds := new_credentials}}
+spark.hadoop.fs.lakefs.access_key={{$creds.ID}}
```
Jonathan-Rosenberg, line 108:
```suggestion
spark.hadoop.fs.lakefs.access_key={{$creds.KEY}}
```
?

johnnyaug (approved):
Looking good!

design/open/template-server/template-server.md
```diff
@@ -0,0 +1,130 @@
...
+action type for template expansion, or alternatively overload
+`fs:ReadObject` to support off-lakeFS URIs.  We will supply the following
+replacements (at least):
+
+| template function | value source                        | IAM permissions                              |
+|:------------------|:------------------------------------|:---------------------------------------------|
+| config            | exportable lakeFS configuration[^1] | `arn:lakefs:template:::config/path/to/field` |
+| object            | (small) object on lakeFS            | IAM read permission for that object          |
+| contenttype       | none; set `Content-Type:`           | (none)                                       |
+| new_credentials   | new credentials added to user       | `auth:CreateCredentials`                     |
```
Jonathan-Rosenberg, line 51: This would potentially generate multiple credentials for each user. These credentials should be managed in some way.

design/open/template-server/template-server.md
arielshaqed: Not done: it's called the "ID" everywhere, in AWS and in our docs.

design/open/template-server/template-server.md
arielshaqed: Thanks!

arielshaqed (commented):
Thanks!

arielshaqed:
Go tests and linters stuck (never showed up as running), and this is just a design doc PR with no code.  So pulling by the powers vested in me as an admin.

PR #3374, opened 2022-05-17
create branch name validation message

### LINKED ISSUE #2894: Documentation and returned error should declare naming conventions
I got from the server on create branch: `argument branchID: invalid value: validation error`
It took me a lot of time to figure out that the branch name must match `^\w[-\w]*$` , and in one of our configurations the branch name contained `.` .
Btw I didn’t find such restriction in documentation, but anyway, if the returned error message will provide any info/hint about the error it would safe a lot of investigation time

eden-ohana:
fixes #2894 

`argument branch: branch id: invalid value: validation error can contain only letters, numbers, underscores, and dashes (not at the beginning)`

### DIFF
--- a/pkg/graveler/errors.go
+++ b/pkg/graveler/errors.go
@@ -23,7 +23,7 @@ var (
 	ErrNoMergeBase                  = errors.New("no merge base")
 	ErrInvalidRef                   = fmt.Errorf("ref: %w", ErrInvalidValue)
 	ErrInvalidCommitID              = fmt.Errorf("commit id: %w", ErrInvalidValue)
-	ErrInvalidBranchID              = fmt.Errorf("branch id: %w can contain only letters, numbers, underscores, and dashes (not at the beginning)", ErrInvalidValue)
+	ErrInvalidBranchID              = fmt.Errorf("branch id: %w", ErrInvalidValue)
 	ErrInvalidTagID                 = fmt.Errorf("tag id: %w", ErrInvalidValue)
 	ErrInvalid                      = errors.New("validation error")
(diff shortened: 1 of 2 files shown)


pkg/graveler/errors.go
```diff
@@ -23,7 +23,7 @@ var (
 	ErrNoMergeBase                  = errors.New("no merge base")
 	ErrInvalidRef                   = fmt.Errorf("ref: %w", ErrInvalidValue)
 	ErrInvalidCommitID              = fmt.Errorf("commit id: %w", ErrInvalidValue)
-	ErrInvalidBranchID              = fmt.Errorf("branch id: %w", ErrInvalidValue)
+	ErrInvalidBranchID              = fmt.Errorf("branch id: %w can contain only letters, numbers, underscores, and dashes (not at the beginning)", ErrInvalidValue)
```
arielshaqed, line 26: This will have the form

> branch id: invalid value: can contain only letters, numbers, underscores, and dashes (not at the beginning)

which looks somewhat different from `ErrInvalidRef` and friends.  Consider

```suggestion
	ErrInvalidBranchID              = fmt.Errorf("branch id must consist of letters, digits, underscores and dashes, and cannot start with a dash: %w", ErrInvalidValue)
```

However I would also like to question whether we need this validation almost anywhere except during creation.  Specifically, if I try to **create** a branch named '-not=valid--` then it should fail.  But anywhere else I only _fetch_ a branch, and I think I should be just as happy to be told "branch not found".

eden-ohana:
```message: "argument branch: branch id: invalid value: validation error: branch id must consist of letters, digits, underscores and dashes, and cannot start with a dash"```

returns only on create branch validation

itaiad200 (approved):

PR #3376, opened 2022-05-18
OIDC support initial research

johnnyaug:


design/open/oidc-support.md
```diff
@@ -0,0 +1,28 @@
...
+
+### Open questions
+
+#### What do we save in our database?
+What information about users should we save in the lakeFS database?
+
+The minimum here is to save the unique user ID generated by the OIDC provider (the "subject" claim).
+
+However, this may not be enough. 
+Since we only want authentication (and not authorization) to be handled by the external provider, we should be able to link between a user and roles/policies. It seems that to do so we need to save some more info (at least username/email). Once we also move authorizaiton outside, it may be easier to avoid saving any PII on the lakeFS side.
```
ortz, line 17: regarding the assumption that we'll have to save more information locally (username/email), I'm not sure that's true.
we can use a hash on top of the mail or the oidc user id, we will need to make sure every function returning the username/email, will now have to query the oidc in order to return good results.
thoughts?

design/open/oidc-support.md
```diff
@@ -0,0 +1,28 @@
...
+
+#### Can an installation deactivate OIDC after it's been turned on?
+
+If so, how does one migrate the existing users?
+
+#### How does one create the initial admin user?
+
+#### How does the admin set the initial permissions for new users?
+
+The users will now be managed in an external database. Once they log in to lakeFS for the first time, a user will be created in the lakeFS database. How does the admin specify the permissions for such a user?
```
ortz, line 27: how do new users today get assigned to default groups?
AFAIK it's not being handled right now, I think it's a good approach to allow a default-group (either via configuration or UI), but it doesn't change the current behavior.

design/open/oidc-support.md
```diff
@@ -0,0 +1,28 @@
...
+The minimum here is to save the unique user ID generated by the OIDC provider (the "subject" claim).
+
+However, this may not be enough. 
+Since we only want authentication (and not authorization) to be handled by the external provider, we should be able to link between a user and roles/policies. It seems that to do so we need to save some more info (at least username/email). Once we also move authorizaiton outside, it may be easier to avoid saving any PII on the lakeFS side.
+
+#### Can an installation deactivate OIDC after it's been turned on?
+
+If so, how does one migrate the existing users?
+
+#### How does one create the initial admin user?
```
ortz, line 23: can't we use our current process that the first user is still required to provide the system?
if we support both internal authentication and oidc side by side, why does it matter?

design/open/oidc-support.md
johnnyaug: I'm considering the case where I show a list of users, and need to display the email for all of them.
So you need to ask for information about users other than the one which is currently logged in. This is not part of the OIDC protocol.
Once users use a concrete provider, we should allow them to implement an `auth.Service` that uses the provider's API to fetch information about users. WDYT?

design/open/oidc-support.md
arielshaqed: LDAP provides a default group.  Otherwise it would be pretty much unusable.

design/open/oidc-support.md
johnnyaug: I was leaning towards not supporting them side-by-side, as this introduces some complications. 
Since I see there's other opinions, I added this as a question.

arielshaqed (commented):
We can avoid storing any user display name by requesting groups from the OIDC provider and using those.

That leaves user management on the OIDC provider -- which is where I would argue that it belongs.  I mean, why do we need to be better than, say, AWS?

johnnyaug:
## What MinIO does
I was able to integrate MinIO with Okta. Here are some findings:
1. Once you switch MinIO to work with OIDC, the existing users become useless. You can still manage them on MinIO, but you can't use them to log in.
2. Policies are the only thing that is still configured on MinIO side. Groups and users move to the identity provider.
3. The policies for a user are saved on the session token. The implication is that you can't change/revoke a user's policies until they log out or their session expires (24 hours). The only way to revoke permissions is to delete the policy or to rotate the secret key.
4. As far as I can tell, this way they are able to avoid storing any PII.

design/open/oidc-support.md
ortz: I think I might be missing you.
if anyone would like to use OIDC, they'll have to implement `auth.Service` in order to have a complete functioning IAM system?
I'm not sure that's a reasonable ask, we should preserve the current features.
Can't we use a separate user for listing users?
A user that the admin will need to provide (along with configuring OIDC)

design/open/oidc-support.md
ortz: > The difference is that today the admin can create a user and immediately set up permissions for them.
> 
> With my suggestion, the admin will invite the user, but the user in lakeFS will be created only once the user logs in, which can happen days later. Then the admin needs to remember to set the correct permissions.

I would stick with using a default group.
This way, on user-activation, it'll be assigned automatically, and if the admin would like to change it, he can, once the user is created.

design/open/oidc-support.md
```diff
@@ -0,0 +1,22 @@
...
+
+In such a solution, the person who configures lakeFS will supply the parameters required for implementing OIDC, including the provider's domain, the client ID and secret, and a callback URL. API access will still use credentials managed internally like today.
+
+As a first step, authorization will still be managed internally by lakeFS.
+
+## Requirements
+
+
+### lakeFS admin
+- As part of the lakeFS configuration file, you can enable an integration with an OIDC provider
```
arielshaqed, line 13:
```suggestion

- As part of the lakeFS configuration file, you can enable an integration with an OIDC provider
```

(Markdown likes a blank line after a header; it does happen to work in GFM but not in more nitpicking tools...).

design/open/oidc-support.md
```diff
@@ -0,0 +1,22 @@
...
+- The admin can configure a global default lakeFS IAM groups to create all users with.
+- Using the "policy" claim on the ID token, the admin can specify an initial group for specific user, to replace the global default group.
```
arielshaqed, line 15: Could you explain a bit more about how the admin configures this, maybe even for a specific provider?  Will this require admins to control groups on the OIDC provider, or on lakeFS, or on both???

Also, users should be able to belong to more than one group; not sure I understand how that will work.

arielshaqed (approved):
NEAT, thanks!

PR #3378, opened 2022-05-18
Document why we use Get before SetIf on the proto store

itaiad200:


arielshaqed (commented):
Not sure whether you want this change.  One "interesting" feature of protobufs is that there is no canonical encoding: you may encode the same protobuf twice and get different byte sequences.

This can happen in various ways.  For instance, if the underlying proto library code uses a hash-map to hold fields by tag but does not take care, the hash function used can change between runs -- and you may get different encodings of the same message.  Or some libraries do a JavaScript and keep tags in the same order in which they were added -- again, so much can go wrong with a comparison like this.

So I wouldn't do this unless the library documentation guaranteed something about field orderings in proto messages.

arielshaqed:
Read https://developers.google.com/protocol-buffers/docs/encoding#order and despair.

Or write a deterministic binary encoder.

N-o-Z:
> Not sure whether you want this change. One "interesting" feature of protobufs is that there is no canonical encoding: you may encode the same protobuf twice and get different byte sequences.
> 
> This can happen in various ways. For instance, if the underlying proto library code uses a hash-map to hold fields by tag but does not take care, the hash function used can change between runs -- and you may get different encodings of the same message. Or some libraries do a JavaScript and keep tags in the same order in which they were added -- again, so much can go wrong with a comparison like this.
> 
> So I wouldn't do this unless the library documentation guaranteed something about field orderings in proto messages.

And that's exactly why I left it. Thanks @arielshaqed

arielshaqed:
One alternative that @nopcoder and I discussed is to keep both the original bytes _and_ a decoded (readonly) message in a single object.  In all cases that I saw (so let's say in most cases in real life?), the SetIf pred arg is something that was previously read.  So if you keep both its bytes and its decoded form, you can read from the decoded form to create a new message, but use the still-encoded original bytes to test whether the message has changed.

(If you're still looking to save on `Get` calls during a structured `SetIf`, the easy way is probably have `SetIf` return a more-current message when it fails.)

N-o-Z (approved):

PR #3398, opened 2022-05-19
UI: Dynamic load js snippets

nopcoder:
Server-side UI configuration to inject JS into our main app.
Unit test to check inject works.
Manual test for testing with and without code snippet.

### DIFF
--- a/pkg/api/injectsnippetsfs.go
+++ b/pkg/api/injectsnippetsfs.go
@@ -7,6 +7,8 @@ import (
 	"io"
 	"io/fs"
 	"time"
+
+	"github.com/treeverse/lakefs/pkg/api/params"
 )
 
 type InjectSnippetsFS struct {
@@ -15,7 +17,7 @@ type InjectSnippetsFS struct {
 	content string
 }
 
-func NewInjectIndexFS(fsys fs.FS, name string, marker string, snippets map[string]string) (fs.FS, error) {
+func NewInjectIndexFS(fsys fs.FS, name string, marker string, snippets []params.CodeSnippet) (fs.FS, error) {
 	if len(snippets) == 0 {
 		// no snippets, return the original marker
 		return fsys, nil
@@ -34,7 +36,7 @@ func NewInjectIndexFS(fsys fs.FS, name string, marker string, snippets map[strin
 
 	// inject snippets code into content
 	codeSnippets := renderCodeSnippets(snippets)
-	contentBytes := bytes.Replace(all, []byte(marker), codeSnippets, 1)
+	contentBytes := bytes.ReplaceAll(all, []byte(marker), codeSnippets)
 	return &InjectSnippetsFS{
 		FS:      fsys,
 		name:    name,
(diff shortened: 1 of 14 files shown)


pkg/api/injectsnippetsfs.go
```diff
@@ -0,0 +1,108 @@
...
+	defer func() { _ = f.Close() }()
+	all, err := io.ReadAll(f)
+	if err != nil {
+		return nil, fmt.Errorf("%s: %w", name, err)
+	}
+
+	// inject snippets code into content
+	codeSnippets := renderCodeSnippets(snippets)
+	endElement := []byte("</head>")
+	idx := bytes.Index(all, endElement)
```
arielshaqed, line 38: I would match rather have an explicit marker for where to inject snippets.

arielshaqed (commented):
Nice!

This is pretty close to an implementation of my #3373, which will allow more general templating (and I have plans for using it outside of HTML  and GUI).  Let's go over both and decide whether to unify them!

pkg/api/injectsnippetsfs.go
nopcoder: Sure, I had this version - but didn't want to show the marker or process the file in case there are no snippets

nopcoder:
> Nice!
> 
> This is pretty close to an implementation of my #3373, which will allow more general templating (and I have plans for using it outside of HTML and GUI). Let's go over both and decide whether to unify them!

Thanks! I was looking for a way to keep the current static web and the dev workflow using 'npm run dev' and still develop server side inject without full templating mechanism.

pkg/api/injectsnippetsfs.go
```diff
@@ -0,0 +1,103 @@
...
+		FS:      fsys,
+		name:    name,
+		content: string(contentBytes),
+	}, nil
+}
+
+func renderCodeSnippets(snippets map[string]string) []byte {
+	var b bytes.Buffer
+	for k, v := range snippets {
+		_, _ = b.WriteString("<!-- snippet begin: " + html.EscapeString(k) + " -->\n")
```
arielshaqed, line 48: This works, but only because of a technicality.  HTML comments end with an explicit "`-->`"; the [exact details of what is allowed inside are complex](https://html.spec.whatwg.org/#comments) but quoting every "`<`" and "`>`" appears -- as a side effect -- to resolve them.  Maybe just strip all non-ASCII or punctuation?  Or maybe package html/template can do it for us?

pkg/api/injectsnippetsfs.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	defer func() { _ = f.Close() }()
+	all, err := io.ReadAll(f)
+	if err != nil {
+		return nil, fmt.Errorf("%s: %w", name, err)
+	}
+
+	// inject snippets code into content
+	codeSnippets := renderCodeSnippets(snippets)
+	contentBytes := bytes.Replace(all, []byte(marker), codeSnippets, 1)
```
arielshaqed, line 37:
```suggestion
	contentBytes := bytes.Replace(all, []byte(marker), codeSnippets, -1)
```
Why do we want to replace at most once?  Sure usually it won't be needed, but I can imagine that it might.  E.g. for debugging I might try to get the string into a `<pre>` block and be able to read what the server generated on the page.

pkg/api/injectsnippetsfs.go
```diff
@@ -0,0 +1,103 @@
...
+	return &InjectSnippetsFS{
+		FS:      fsys,
+		name:    name,
+		content: string(contentBytes),
+	}, nil
+}
+
+func renderCodeSnippets(snippets map[string]string) []byte {
+	var b bytes.Buffer
+	for k, v := range snippets {
```
arielshaqed, line 47: This gives no way to order snippets.  That will give unexpected behaviour, and only some of the time.  I would prefer to have a JSON array of `{key: ..., value: ...}` objects; is that possible?

arielshaqed (commented):
Neater and neater!

pkg/api/injectsnippetsfs.go
nopcoder: I prefer to encode the text as is without inject and get the user description here :)

Issue #3395, opened 2022-05-19
kv package implementation feedback


nopcoder:
- [x] entry as string should not print dump as string
- [x] verify that value can be empty without null
- [x] postgres view on kv table to list keys as strings

nopcoder:
Verified that inserting a null as inserting an empty value to bytea in postres are not the same.
We don't want to enable null for key or value.

itaiad200:
Can we add tests for all of these cases?

PR #3402, opened 2022-05-21
kv view for easy access

nopcoder:
- Add: create view for easy select on kv table
- Fix: dump slice of bytes for Entry's String 

Fix: #3395

### DIFF
--- a/pkg/kv/postgres/store.go
+++ b/pkg/kv/postgres/store.go
@@ -97,9 +97,17 @@ func parseStoreConfig(runtimeParams map[string]string) *Params {
 
 // setupKeyValueDatabase setup everything required to enable kv over postgres
 func setupKeyValueDatabase(ctx context.Context, conn *pgxpool.Conn, params *Params) error {
+	// main kv table
 	_, err := conn.Exec(ctx, `CREATE TABLE IF NOT EXISTS `+params.SanitizedTableName+` (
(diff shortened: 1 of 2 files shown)


pkg/kv/postgres/store.go
```diff
@@ -97,9 +97,17 @@ func parseStoreConfig(runtimeParams map[string]string) *Params {
...
-    value BYTEA NOT NULL);`)
+		key BYTEA NOT NULL PRIMARY KEY,
+		value BYTEA NOT NULL);`)
+	if err != nil {
+		return err
+	}
+
+	// view of kv table to help humans select from table (same as table with _v as suffix)
+	_, err = conn.Exec(ctx, `CREATE OR REPLACE VIEW `+pgx.Identifier{params.TableName + "_v"}.Sanitize()+
+		` AS SELECT ENCODE(key, 'escape') AS path, VALUE FROM `+params.SanitizedTableName)
```
itaiad200, line 110: I would prefer if we don't introduce new concepts if we don't have too.
```suggestion
		` AS SELECT ENCODE(key, 'escape') AS key, value FROM `+params.SanitizedTableName)
```

itaiad200 (approved):

pkg/kv/postgres/store.go
nopcoder: so we need to align the msg interface

Issue #3349, opened 2022-05-11
Create KV Auth models


N-o-Z:
Create the data structures to support Auth in KV mode

- Depends on #3348

PR #3405, opened 2022-05-22
Auth on KV discovery & design doc

### LINKED ISSUE #3348: KV Auth Discovery
- Gather requirements for KV implementation for Auth package
- Map entity relations and define KV models for Auth entities
- Define interfaces needed to support Auth flows with KV based store

itaiad200:
closes #3348

### DIFF
--- a/design/accepted/kv_auth.md
+++ b/design/accepted/kv_auth.md
@@ -1,0 +1,115 @@
+# Implementing Auth Package with KV Database
+
+This document describes the entities of lakeFS auth package, the relationship between them and offers
+a possible representation of them in the KV database.
+
+## Entities
+
+### User
+
+- Can be looked up by `ID`, `Username` & `Email`.
+- Deleted by `Username`.
+- Listed by `Username`.
+
+### Group
+
+- Looked up, deleted and listed by `DisplayName`.
+
+### Policies
+
+- Looked up, deleted and listed by `DisplayName`.
+
+### Credentials
+
+- Has a foreign id reference to `User.ID`.
+    - The `User.ID` is used for looking up the `User.Username`.
+    - Almost all actions are in the context of `User.Username`, i.e. it is always passed.
+- The only action without a `User.Username` passed is credentials lookup by `AccessKeyID`.
+
+
+## Entities Relationship
+
+### Group <-> User
+
+- Add/Remove `User` to `Group` by `User.Username` & `Group.DisplayName`. 
+- List all the `Group`s for a `User` by `User.Username`.
+- List all the `Users`s in a `Group` by `Group.DisplayName`.
+
+### Policy <-> User
+
+- Attach/Detach `Policy` to `User` by `User.Username` & `Policy.DisplayName`.
+- List all `Policies` for a `User` by `User.Username`.
+- List all effective `Policies` for a `User` by `User.Username`, this includes all `Policies` for `Group`s is a member of.
+
+### Policy <-> Group
+
+- Attach/Detach `Policy` to `Group` by `Group.DisplayName` & `Policy.DisplayName`.
+- List all `Policies` for a `Group` by `Group.DisplayName`.
+    - `Group` cannot be a member of another `Group`, therefore all effective policies are attached to it directly.
+
+## Representation in the KV world
+
+- All keys are prefixed by the reserved package prefix `auth/`.
+  
+- `User`s key will be in the form of `users/<UserName>`.
+- `Group`s key will be in the form of `groups/<DisplayName>`.
+- `Policies`s key will be in the form of `policies/<DisplayName>`.
+- `Credentials`s key will be in the form of `users/<UserName>/credentials/<AccessKeyID>`.
+
+### Handling Secondary Indexes
+
+Below are 2 options that vary in efficiency and complexity of the implementation.
+ 
+#### Store just the minimum indexes
+
+Keep only the minimum that is required to represent the relationship of the entities:
+
+- A `User` membership of a `Group` under `groups/<DisplayName>/users/<Username>`.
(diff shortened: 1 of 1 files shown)


design/accepted/kv_auth.md
```diff
@@ -0,0 +1,115 @@
...
+Listing by anything other than the key will require to list all the entities which are relevant.
+Some examples:
+
+1. Finding a `User` by `AccessKeyID` will require to list all the users.
+1. Listing the `User`'s effective policies will require to list all entities under `groups/`.
+
+The amount of entities in the Auth world isn't big (<10k) and cached in the server, it's unlikely it will incur a 
+notable performance degradation.
+
+#### Store all secondary indexes
```
N-o-Z, line 84: I am more worried about duplication and the risk of inconsistency than performance itself. And since your assumption is that performance in this data scale is not an issue - I think we should make sure not to duplicate data unless it is absolutely unavoidable

design/accepted/kv_auth.md
```diff
@@ -0,0 +1,115 @@
...
+For example, a deleted `Policy` would cascade to the `auth_user_policies` & `auth_group_policies` tables.
+
+
+In the KV world, we must cleanup the secondary indexes first, before deleting the entity itself.
+Since the KV is lock-free, we might still get into troubles.
+For example, a `Policy` is requested to be deleted. The operation starts by listing all `User`s & `Group`s with that `Policy` attached to it.
+While iterating, it's possible that another unlisted `User` would be attached to the `Policy`. Once the `Policy` is deleted,
+we would remain with an attachment to it. There are at least 2 mitigations for it:
+
+1. Do (a) list and delete of secondary indexes (b) delete the entity (c) another round of list and delete.
```
N-o-Z, line 108: In the case why not perform the deletion order in reverse? delete entity -> delete secondary indexes without making another round?

N-o-Z (commented):
@itaiad200  Thanks,

I didn't quite understand why we have to delete the secondary indices first and only then delete the entry. It seems to me that doing it the other way around might resolve some of the concerns you raised

design/accepted/kv_auth.md
```diff
@@ -0,0 +1,115 @@
...
+- List all `Policies` for a `User` by `User.Username`.
+- List all effective `Policies` for a `User` by `User.Username`, this includes all `Policies` for `Group`s is a member of.
+
+### Policy <-> Group
+
+- Attach/Detach `Policy` to `Group` by `Group.DisplayName` & `Policy.DisplayName`.
+- List all `Policies` for a `Group` by `Group.DisplayName`.
+    - `Group` cannot be a member of another `Group`, therefore all effective policies are attached to it directly.
+
+## Representation in the KV world
```
N-o-Z, line 50: In addition to the keys, I think it will be very very helpful to explicitly describe the KV model structs

N-o-Z (changes requested):

design/accepted/kv_auth.md
```diff
@@ -0,0 +1,115 @@
...
+
+- Add/Remove `User` to `Group` by `User.Username` & `Group.DisplayName`. 
+- List all the `Group`s for a `User` by `User.Username`.
+- List all the `Users`s in a `Group` by `Group.DisplayName`.
+
+### Policy <-> User
+
+- Attach/Detach `Policy` to `User` by `User.Username` & `Policy.DisplayName`.
+- List all `Policies` for a `User` by `User.Username`.
+- List all effective `Policies` for a `User` by `User.Username`, this includes all `Policies` for `Group`s is a member of.
```
N-o-Z, line 42:
```suggestion
- List all effective `Policies` for a `User` by `User.Username`, this includes all `Group Policies` for Groups the user is a member of.
```

design/accepted/kv_auth.md
ortz: @N-o-Z this might cause an issue as now there are users attached to a policy that doesn't exist, I think we should check how the user's effective policy generation would behave if that's the case.

design/accepted/kv_auth.md
ortz: +1 to @N-o-Z 
due to the quantitive nature of this feature (small amounts in IAM), I would prefer consistency over performance, especially with the cache.

design/accepted/kv_auth.md
N-o-Z: Assuming this is an issue, we will still encounter it in the second round of list and delete

design/accepted/kv_auth.md
itaiad200: We have a separate issue for that #3349

design/accepted/kv_auth.md
itaiad200: I agree, that's why I preferred option 1.

design/accepted/kv_auth.md
itaiad200: By design, we don't have a DB lock. Hence with with every implementation we choose, extra care must be taken.
When reading secondary index, one must make sure that the principal entities exist. If not, it's ok to delete the secondary index.

itaiad200:
> @itaiad200 how do we address the user id and group id which are currently a `serial` type in our database - auto-generated key?

To my understanding, there's no real need for those, right? It's a good opportunity to deprecate them, if there's any reason for keeping them, then yes, auto generated keys should work.

nopcoder:
> > @itaiad200 how do we address the user id and group id which are currently a `serial` type in our database - auto-generated key?
> 
> To my understanding, there's no real need for those, right? It's a good opportunity to deprecate them, if there's any reason for keeping them, then yes, auto generated keys should work.

True. So the lakefs implementation will have to change and we will need to migrate the existing IDs to generated one.

nopcoder (approved):

N-o-Z (approved):

PR #3407, opened 2022-05-22
Evict user from auth-cache on reset password

idanovo:


design/open/evict-user-from-auth-cache-on-reset-password.md
```diff
@@ -0,0 +1,13 @@
+## The problem: 
+When a user logs in to lakeFS we save his credentials in the auth-cache.
```
arielshaqed, line 2:
```suggestion
When a user logs in to lakeFS we save their credentials in the auth-cache. 
```
and also below.

See [How to write great documentation for lakeFS, "style basics"](https://docs.google.com/document/d/1vif05nDiKZYPwYdINLoq4BJ6j7bBIQ6ZA2dt7McuzyY/edit#heading=h.dqp77semtfge).

arielshaqed (commented):
This seems like an easy way to resolve many of our issues when changing user passwords.  I _think_ it will be OK: after logging in the user gets a JWT and authenticates with it.  So the password is never really used again for a significant period of time!

(However other auth information for the user, such as group memberships and groups and their policies, **are** re-used on each call and must remain cached!)

design/open/evict-user-from-auth-cache-on-reset-password.md
```diff
@@ -0,0 +1,13 @@
+## The problem: 
+When a user logs in to lakeFS we save their credentials in the auth-cache. 
+This way, we avoid checking the usr credentials through the DB on each call the user makes. 
+Currently, when a user tries to reset his password, his old password is still stored in the auth-cache, so he has to wait till it is evicted to log in using the new password.
```
johnnyaug, line 4:
```suggestion
Currently, when a user tries to reset their password, their old password is still stored in the auth-cache, so they have to wait till it is evicted to log in using the new password.
```

johnnyaug: Also further down the text

design/open/evict-user-from-auth-cache-on-reset-password.md
```diff
@@ -0,0 +1,13 @@
...
+
+
+## The suggested solution: 
+The intuitive solution is just to evict the user from the auth cache when resetting his password. 
+The problem with this solution is that we might not necessarily have a centralized cache, so we might need to evict the user for every auth cache he exists in. 
+This is why the simpler solution here might be just to not use auth cache for email-password authentication. 
+This way, we still won't need to authenticate when a user uses API calls that use access-key and secret for authentication and won't need to do the effort of evicting the user auth from all the auth cache instances.
+In case we decide to go with this solution, we have to enable email/password authentication only through the UI.
+Otherwise, user might try to use email/password authentication with our Python/Java client and won't have their credentials cached.
```
johnnyaug, line 13: And why is that so bad? 
We already state in the docs that email+password _should_ be used only for the UI.
If the user decides to use them for the API, an extra DB call will happen for every request - I think it will not cause a bottleneck. If it does in the future - we can solve it.

nopcoder (changes requested):
Remove the unused cache `userEmailCache`

design/open/evict-user-from-auth-cache-on-reset-password.md
arielshaqed: Think of a Spark app reading from multiple workers. One user in one org makes a mistake, and all users get rate limited.

A shared auth server will have this problem in trumps, even more so if it fails to protect its DB access.

PR #3408, opened 2022-05-22
kv API update to get predicate when needed for SetIf operation

### LINKED ISSUE #3400: kv API predicate handling
The current API to `SetIf` enable the developer to pass a value as the predicate ([]byte).
To enable any underlaying implementation to use `SetIf` with any value of a predicate - returned from a `Get` operation, we should design the API to same.

The Get should return an opaque value (interface{}) that should be passed to the `SetIf` operation. Using this the implementor and the developer using the API will be a clear way to implement and pass the predicate value.

nopcoder:
Close #3400

### DIFF
--- a/pkg/kv/mem/store.go
+++ b/pkg/kv/mem/store.go
@@ -51,6 +51,18 @@ func (s *Store) Get(_ context.Context, key []byte) ([]byte, error) {
 	return value, nil
 }
 
+func (s *Store) GetEntry(ctx context.Context, key []byte) (*kv.Entry, error) {
+	v, err := s.Get(ctx, key)
+	if err != nil {
+		return nil, err
+	}
+	return &kv.Entry{
+		Key:       key,
+		Value:     v,
+		Predicate: kv.Predicate(v),
+	}, nil
+}
+
 func (s *Store) Set(_ context.Context, key, value []byte) error {
 	if key == nil {
 		return kv.ErrMissingKey
@@ -79,26 +91,26 @@ func (s *Store) insertNewKey(key []byte) {
 	}
 }
 
-func (s *Store) SetIf(_ context.Context, key, value, valuePredicate []byte) error {
+func (s *Store) SetIf(_ context.Context, key, value []byte, valuePredicate kv.Predicate) (kv.Predicate, error) {
 	if key == nil {
-		return kv.ErrMissingKey
+		return nil, kv.ErrMissingKey
 	}
 	if value == nil {
-		return kv.ErrMissingValue
+		return nil, kv.ErrMissingValue
 	}
 	s.mu.Lock()
 	defer s.mu.Unlock()
 	curr, currOK := s.m[string(key)]
 	if valuePredicate == nil {
 		if currOK {
-			return kv.ErrPredicateFailed
+			return nil, kv.ErrPredicateFailed
 		}
 		s.insertNewKey(key)
-	} else if !bytes.Equal(valuePredicate, curr) {
-		return kv.ErrPredicateFailed
+	} else if !bytes.Equal(valuePredicate.([]byte), curr) {
+		return nil, kv.ErrPredicateFailed
 	}
 	s.m[string(key)] = value
-	return nil
+	return kv.Predicate(value), nil
 }
 
 func (s *Store) Delete(_ context.Context, key []byte) error {
@@ -147,8 +159,9 @@ func (e *EntriesIterator) Next() bool {
 	key := e.store.keys[idx]
 	value := e.store.m[key]
 	e.entry = &kv.Entry{
-		Key:   []byte(key),
-		Value: value,
+		Key:       []byte(key),
+		Value:     value,
+		Predicate: kv.Predicate(value),
 	}
 	// set start to the next item - nil as end indicator
 	if idx+1 < len(e.store.keys) {
--- a/pkg/kv/mem/store.go
+++ b/pkg/kv/mem/store.go
@@ -3,6 +3,7 @@ package mem
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"sort"
 	"sync"
 
@@ -39,14 +40,14 @@ func (d *Driver) Open(_ context.Context, _ string) (kv.Store, error) {
 }
 
 func (s *Store) Get(_ context.Context, key []byte) ([]byte, error) {
-	if key == nil {
+	if len(key) == 0 {
 		return nil, kv.ErrMissingKey
 	}
 	s.mu.RLock()
 	defer s.mu.RUnlock()
 	value, ok := s.m[string(key)]
 	if !ok {
-		return nil, kv.ErrNotFound
+		return nil, fmt.Errorf("key=%v: %w", key, kv.ErrNotFound)
 	}
 	return value, nil
 }
@@ -63,7 +64,7 @@ func (s *Store) GetValuePredicate(ctx context.Context, key []byte) (*kv.ValuePre
 }
 
 func (s *Store) Set(_ context.Context, key, value []byte) error {
-	if key == nil {
+	if len(key) == 0 {
 		return kv.ErrMissingKey
 	}
 	if value == nil {
@@ -91,7 +92,7 @@ func (s *Store) insertNewKey(key []byte) {
 }
 
 func (s *Store) SetIf(_ context.Context, key, value []byte, valuePredicate kv.Predicate) error {
-	if key == nil {
+	if len(key) == 0 {
 		return kv.ErrMissingKey
 	}
 	if value == nil {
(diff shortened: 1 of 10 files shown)


itaiad200 (changes requested):
I think we should replace the `Get` API and not add another one

pkg/kv/store.go
```diff
@@ -41,18 +41,22 @@ type Driver interface {
...
 }
 
+type Predicate interface{}
+
 type Store interface {
 	// Get returns a value for the given key, or ErrNotFound if key doesn't exist
 	Get(ctx context.Context, key []byte) ([]byte, error)
 
+	GetEntry(ctx context.Context, key []byte) (*Entry, error)
+
```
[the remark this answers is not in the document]
nopcoder: Renamed to GetValuaPredicate - the case where Get without predicate is a lighter version of Get with predicate.
I've tested the "options" variant, where we have one Get and one Set method and passing options to the function control if you like to pass a predicate or not.
The problem I see in this style is the contract that the implementor need to address - the combination of all the options and the expected result.
When you have a method signature - it is easy to pass the requirement for the implementor.

pkg/kv/store_message.go
```diff
@@ -13,55 +12,39 @@ type StoreMessage struct {
...
-func (s *StoreMessage) GetMsg(ctx context.Context, path string, msg protoreflect.ProtoMessage) error {
-	val, err := s.Store.Get(ctx, []byte(path))
+func (s *StoreMessage) GetMsg(ctx context.Context, path string, msg protoreflect.ProtoMessage) (Predicate, error) {
+	ent, err := s.Store.GetEntry(ctx, []byte(path))
 	if err != nil {
-		return fmt.Errorf("failed on Get (path: %s): %w", path, err)
+		return nil, err
 	}
-	return proto.Unmarshal(val, msg)
+	if msg == nil {
```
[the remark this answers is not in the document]
nopcoder: It was a quick way to get a predicate for a key without serialize the value.

pkg/kv/store.go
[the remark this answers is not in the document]
itaiad200: I get why we would like to have 2 `Set`s functions, since the predicate existence changes the behaviour.
I don't get why there's a need for 2 `Get`s. We pass the predicate and if the caller wish, she can ignore it. It's not saving any time/memory, just makes the api more confusing..

pkg/kv/store_message.go
```diff
@@ -14,54 +13,46 @@ type StoreMessage struct {
 }
 
 func (s *StoreMessage) GetMsg(ctx context.Context, path string, msg protoreflect.ProtoMessage) error {
-	val, err := s.Store.Get(ctx, []byte(path))
+	v, err := s.Store.Get(ctx, []byte(path))
 	if err != nil {
-		return fmt.Errorf("failed on Get (path: %s): %w", path, err)
+		return err
```
N-o-Z, line 18: Why did you remove the formatting? It was the only place where we log the key that failed

pkg/kv/store_message.go
```diff
@@ -14,54 +13,46 @@ type StoreMessage struct {
...
+	return proto.Unmarshal(v, msg)
 }
 
-func (s *StoreMessage) SetMsg(ctx context.Context, path string, msg protoreflect.ProtoMessage) error {
-	val, err := proto.Marshal(msg)
+func (s *StoreMessage) GetMsgPredicate(ctx context.Context, path string, msg protoreflect.ProtoMessage) (Predicate, error) {
+	vp, err := s.Store.GetValuePredicate(ctx, []byte(path))
 	if err != nil {
-		return fmt.Errorf("failed on Set (path: %s): %w", path, err)
+		return nil, err
```
N-o-Z, line 26: Same as before

pkg/kv/store_message.go
```diff
@@ -14,54 +13,46 @@ type StoreMessage struct {
...
 
-func (s *StoreMessage) SetIf(ctx context.Context, path string, msg protoreflect.ProtoMessage, pred protoreflect.ProtoMessage) error {
+func (s *StoreMessage) SetMsg(ctx context.Context, path string, msg protoreflect.ProtoMessage) error {
 	val, err := proto.Marshal(msg)
 	if err != nil {
-		return fmt.Errorf("failed on Marshal (path: %s): %w", path, err)
-	}
-	if pred == nil {
-		return s.Store.SetIf(ctx, []byte(path), val, nil)
+		return err
```
N-o-Z, line 41: Same as previous comment

pkg/kv/store_message.go
```diff
@@ -14,54 +13,46 @@ type StoreMessage struct {
...
+	return s.Store.SetIf(ctx, []byte(path), val, predicate)
 }
 
 func (s *StoreMessage) Delete(ctx context.Context, path string) error {
-	err := s.Store.Delete(ctx, []byte(path))
-	if err != nil {
-		err = fmt.Errorf("failed on Delete (path: %s): %w", path, err)
-	}
-	return err
+	return s.Store.Delete(ctx, []byte(path))
```
N-o-Z, line 55: The error was here for a reason - this changes the multiparts tracker current behavior

pkg/kv/store.go
```diff
@@ -41,18 +41,22 @@ type Driver interface {
 	Open(ctx context.Context, dsn string) (Store, error)
 }
 
+type Predicate interface{}
+
 type Store interface {
 	// Get returns a value for the given key, or ErrNotFound if key doesn't exist
 	Get(ctx context.Context, key []byte) ([]byte, error)
 
+	GetValuePredicate(ctx context.Context, key []byte) (*ValuePredicate, error)
```
N-o-Z, line 50: Missing documentation

pkg/kv/store.go
```diff
@@ -81,12 +85,18 @@ type EntriesIterator interface {
 	Close()
 }
 
-// Entry holds a pair of key/value
+// Entry holds a pair of key/value and predicate used for conditional set
```
N-o-Z, line 88: No predicate in struct

itaiad200 (changes requested):
I still think we should have a single `Get` API for the store

pkg/kv/store_message.go
nopcoder: Will address it as part of Store implementation

N-o-Z: I wonder why you have decided to move it to the store implementation, as this will require all stores to make sure to log the error with the key. When done in the store_message level you only have to take care of it once WDYT?

pkg/kv/store_message.go
N-o-Z: The comment was regarding the removal of the check in delete to see if the key exists and if not return an error

pkg/kv/store.go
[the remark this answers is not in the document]
nopcoder: Didn't want to lock the option that there can be a version of 'get' with predicate that will slower to use than a 'get' for the latest value (without predicate).
I'll discover and see how we plan to implement the 'get' for different underlying storage types - if it is the same cost as calling a simple 'get' - I'll make predicate optional and leave one API.

pkg/kv/store_message.go
nopcoder: Wanted to get the infomation also when you directly access the storage.
- We can guide the storage implementor
- Create a wrapper that will wrap the underlying storage to enritch the return errors
WDYT?

pkg/kv/store_message.go
nopcoder: The previous code - for any error in delete enrich the error with path - no extra verification at this level.
```go
        err := s.Store.Delete(ctx, []byte(path))
	if err != nil {
		err = fmt.Errorf("failed on Delete (path: %s): %w", path, err)
	}
	return err
```

N-o-Z: Sorry, my bad - it's handled somewhere else

pkg/kv/store_message.go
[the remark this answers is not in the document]
N-o-Z: Maybe we should document this functionality in the function so the user understands that passing nil msg to GetMsg is a feature :)

N-o-Z (approved):

pkg/kv/store.go
```diff
@@ -41,9 +41,23 @@ type Driver interface {
 	Open(ctx context.Context, dsn string) (Store, error)
 }
 
+// Predicate value used to update a key base on a previous fetched value.
+//   Store's GetValuePredicate used to pull the key's value with the associated predicate.
```
itaiad200, line 45:
```suggestion
//   Store's Get used to pull the key's value with the associated predicate.
```

pkg/kv/store.go
```diff
@@ -41,9 +41,23 @@ type Driver interface {
...
+//   Store's SetIf used to set the key's value based on the predicate.
+type Predicate interface{}
+
+// GetResult value with predicate - Value holds the data and Predicate a value used for conditional set.
+//   Get operation will return this struct, holding the key's information
+//   SetIf operation will use the Predicate for conditional set
+type GetResult struct {
```
itaiad200, line 52: Perhaps 
```suggestion
type ValueWithPredicate struct {
```

pkg/kv/mem/store.go
```diff
@@ -91,18 +95,18 @@ func (s *Store) SetIf(_ context.Context, key, value, valuePredicate []byte) erro
 	curr, currOK := s.m[string(key)]
 	if valuePredicate == nil {
 		if currOK {
-			return kv.ErrPredicateFailed
+			return fmt.Errorf("%w: key=%v", kv.ErrPredicateFailed, key)
```
itaiad200, line 98:
```suggestion
			return fmt.Errorf("key=%v: %w", key, kv.ErrPredicateFailed)
```

itaiad200: Here and elsewhere.

itaiad200 (approved):
LGTM
just mind the error formating

PR #3412, opened 2022-05-22
KV Store: Message Store Scan

### LINKED ISSUE #3409: Implement message_store Scan and Iterator

N-o-Z:
Implemented Scan and prefix iterator for proto messages

- Closes #3409

### DIFF
--- a/pkg/gateway/multiparts/tracker.go
+++ b/pkg/gateway/multiparts/tracker.go
@@ -9,7 +9,7 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
-const multipartPrefix = "multipart"
+const multipartsPrefix = "multiparts"
 
 type Metadata map[string]string
 
@@ -76,8 +76,8 @@ func (m *tracker) Create(ctx context.Context, multipart MultipartUpload) error {
 	if multipart.UploadID == "" {
 		return ErrInvalidUploadID
 	}
-	path := kv.FormatPath(multipartPrefix, multipart.UploadID)
-	err := m.store.SetIf(ctx, path, protoFromMultipart(&multipart), nil)
+	key := kv.FormatPath(multipartsPrefix, multipart.UploadID)
+	err := m.store.SetIf(ctx, key, protoFromMultipart(&multipart), nil)
 	return err
 }
 
@@ -86,8 +86,8 @@ func (m *tracker) Get(ctx context.Context, uploadID string) (*MultipartUpload, e
 		return nil, ErrInvalidUploadID
 	}
 	data := &MultipartUploadData{}
-	path := kv.FormatPath(multipartPrefix, uploadID)
-	err := m.store.GetMsg(ctx, path, data)
+	key := kv.FormatPath(multipartsPrefix, uploadID)
+	err := m.store.GetMsg(ctx, key, data)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


pkg/gateway/multiparts/tracker.go
```diff
@@ -9,7 +9,7 @@ import (
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
-const multipartPrefix = "multipart"
+const multipartsPrefix = "multiparts"
```
itaiad200, line 12: Just a MASSIVE breaking change that we're allowed to do for now ;)

pkg/kv/store_message.go
```diff
@@ -64,8 +64,66 @@ func (s *StoreMessage) Delete(ctx context.Context, path string) error {
 	return err
 }
 
-// TODO(niro): implement Scan when required
+// Scan returns a prefix iterator which returns entries in deserialized format. Scan in store message implementation is msg specific
+// and therefore must be PREFIX bound
```
itaiad200, line 68: So why isn't `prefix` being validated?

pkg/kv/store_message.go
```diff
@@ -64,8 +64,66 @@ func (s *StoreMessage) Delete(ctx context.Context, path string) error {
...
 	s.Store.Close()
 }
+
+type MessageIterator struct {
+	itr EntriesIterator
+	msg protoreflect.Message
+	err error
+}
+
+type MessageEntry struct {
```
itaiad200, line 83: You need a `predicate` as well.

pkg/kv/store_message.go
```diff
@@ -64,8 +64,66 @@ func (s *StoreMessage) Delete(ctx context.Context, path string) error {
...
+func (m *MessageIterator) Next() bool {
+	if m.err != nil {
+		return false
+	}
+	return m.itr.Next()
+}
+
+func (m *MessageIterator) Entry() *MessageEntry {
+	entry := m.itr.Entry()
+	if entry == nil {
```
itaiad200, line 105: What if there's an error in the underlying iterator? I think the `Err()` functionality is fine, but not sure about the `Next()` in such a case.

itaiad200 (changes requested):
I believe you're missing a predicate in the returned value?

pkg/kv/store_message.go
N-o-Z: Can you explain what kind of validations you had in mind?
Iterator-wise we are using the PrefixIterator which checks for it

pkg/kv/store_message.go
N-o-Z: Can you explain?

itaiad200: Look at #3408

N-o-Z: I would not like to rely on currently WIP PRs. I prefer addressing it only when the changes are committed

pkg/kv/store_message.go
N-o-Z: prefix is a byte slice and the comparison in PrefixIterator is `!bytes.HasPrefix(entry.Key, b.Prefix)` so it will return false for all keys and iterate over all kv store

itaiad200: So it can be unbounded and therefore the comment needs to change, right?

itaiad200 (approved):

PR #3415, opened 2022-05-23
Fix: Use repository root commit for import branch in import from UI

### LINKED ISSUE #3375: Import branches should be created from the repo root commit
**Actual**

The import branch is created from the HEAD of the given branch. When merging the import branch to the origin branch, it will delete all entries that existed when the import branch was created, due to the nature of 3-way merge.

**Desired**

Entries that existed before the import are not deleted. This can be achieved if the import branch is created from the repository root commit.

N-o-Z:
- Closes #3375

### DIFF
--- a/webui/src/pages/repositories/repository/objects.jsx
+++ b/webui/src/pages/repositories/repository/objects.jsx
@@ -133,12 +133,16 @@ const ImportButton = ({ config, repo, reference, path, onDone, onClick, variant 
             } catch (error) {
                 if (error instanceof NotFoundError) {
                     // Find root commit for repository
-                    let response = await commits.log(repo.id, repo.default_branch, "", 1000)
-                    while (response.pagination.has_more) {
-                        response = await commits.log(repo.id, repo.default_branch, response.pagination.next_offset, 1000)
-                    }
-                    let baseBranch = response.results.at(-1)
-                    await branches.create(repo.id, importBranch, baseBranch.id)
+                    let hasMore = true;
+                    let nextOffset = "";
+                    let baseCommit = reference.id;
+                    do {
+                    	let response = await commits.log(repo.id, reference.id, nextOffset, 1000);
+                    	hasMore = response.pagination.has_more;
+                    	nextOffset = response.pagination.next_offset;
+                    	baseCommit = response.results.at(-1);
+                    } while (hasMore)
+                    await branches.create(repo.id, importBranch, baseCommit.id)
                     importBranchResp = await branches.get(repo.id, importBranch)
                 } else {
                     setImportState({...initialState, error})
--- a/webui/src/pages/repositories/repository/objects.jsx
+++ b/webui/src/pages/repositories/repository/objects.jsx
@@ -59,10 +59,6 @@ const ImportDone = ({ numObjects, importBranch, currBranch}) => {
                 <div className='import-text'>
                     <p> Use the&nbsp;<Link to={`compare?ref=${ currBranch }&compare=${ importBranch }`} variant = "success" >Compare tab</Link>&nbsp;to view the changes and merge them to {currBranch}.</p>
                 </div>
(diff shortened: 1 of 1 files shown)


webui/src/pages/repositories/repository/objects.jsx
```diff
@@ -137,7 +132,13 @@ const ImportButton = ({ config, repo, reference, path, onDone, onClick, variant
                 importBranchResp = await branches.get(repo.id, importBranch)
             } catch (error) {
                 if (error instanceof NotFoundError) {
-                    await branches.create(repo.id, importBranch, currBranch)
+                    // Find root commit for repository
+                    let response = await commits.log(repo.id, repo.default_branch, "", 1000)
+                    while (response.pagination.has_more) {
+                        response = await commits.log(repo.id, repo.default_branch, response.pagination.next_offset, 1000)
+                    }
```
itaiad200, line 139: Does JS has a `do {...} while(...)` option?

webui/src/pages/repositories/repository/objects.jsx
```diff
@@ -81,7 +77,6 @@ const ImportButton = ({ config, repo, reference, path, onDone, onClick, variant
     const sourceRef = useRef(null);
     const destRef = useRef(null);
     const commitMsgRef = useRef(null);
-    const storageType = config.blockstore_type
     const storageNamespaceValidityRegexStr = config.blockstore_namespace_ValidityRegex;
     const storageNamespaceValidityRegex = RegExp(storageNamespaceValidityRegexStr);
     const sourceURIExample = config ? config.blockstore_namespace_example : "s3://my-bucket/path/";
```
itaiad200, line 82: Lines 86-90 are wrong, think about importing to a branch named `main_2`

itaiad200 (approved):
LGTM but I would wait for a native JS reader to review as well.

webui/src/pages/repositories/repository/objects.jsx
N-o-Z: There is a do-while but since I need the response I need to define it outside the scope of the while, I can separate the definition from the assignment if you think it looks better.
Can you recommend on a native JS reader to review? :D

webui/src/pages/repositories/repository/objects.jsx
N-o-Z: Opened #3416 for it, we should address it there

webui/src/pages/repositories/repository/objects.jsx
```diff
@@ -137,7 +132,13 @@ const ImportButton = ({ config, repo, reference, path, onDone, onClick, variant
                 importBranchResp = await branches.get(repo.id, importBranch)
             } catch (error) {
                 if (error instanceof NotFoundError) {
-                    await branches.create(repo.id, importBranch, currBranch)
+                    // Find root commit for repository
+                    let response = await commits.log(repo.id, repo.default_branch, "", 1000)
```
eden-ohana, line 136:
```suggestion
                    let response = await commits.log(repo.id, reference.id, "", 1000)
```
for consistency

eden-ohana (approved):
Thanks! it works smoothly

webui/src/pages/repositories/repository/objects.jsx
itaiad200: So maybe:

```suggestion
                    // Find root commit for repository
                    let hasMore = true;
                    let nextOffset = "";
                    let baseCommit = reference.id;
                    do {
                    	let response = await commits.log(repo.id, reference.id, nextOffset, 1000);
                    	hasMore = response.pagination.has_more;
                    	nextOffset = response.pagination.next_offset;
                    	baseCommit = response.results.at(-1);
                    } while (hasMore)
```

itaiad200 (approved):
LGTM
Thanks for making the fix! Please test the new version as well.

Issue #1511, opened 2021-02-24
Actions runs information and output can be access through our storage (no postgres)


nopcoder:
In order to reduce the use of Postgres, we can use the underlying storage with the run manifest information to supply the list and get of action's hook information.

### Solution
1. How to support listing of action runs without relaying on Postgres. 
    - using the fact the run_id can be sorted by date due to its naming format, we can use the bucket listing API to sort by latest date.
    - Parse the run.manifest stored in the object storage to get the action runs details
    - Using pagination, we will reduce the number of files we need to parse in each listing operation. (The action runs listing capability is paginated (p.size = 25) and therefore we don't expect performance hit with the parsing action).
2. How to support filtering by commit id and by branch 
    - We will index each action run two times as follows:
   
```
    _lakefs/actions/log/<runid>/run.manifest #includes the action run's details
    metadata/lakefs_actions/	
    	/commits/<commit_id>/<run_id>  #/commits/<commit_id>/ represents the index 
    	/branches/<branch_name>/<run_id>  #/branches/<branch_name>/ represents the index 
```
- To filter by commit id/branch:
        -   go to `metadata/lakefs_actions/commits/<commit_id>/` or `metadata/lakefs_actions/branches/<branch_name>/`
        -         get all the run ids associated with the commit/branch
        -         construct the path of the run. manifest for each run_id
        -         read the run.manifest files 

3. How to keep the action runs data consistent
    - We only keep a single source of truth with the action details - the run.manifest file.
    - The order in which we write the data will guarantees that no action run is persisted without making it filterable. The order of writes (divided into pre and post commit stages) -
        1. Pre-commit - 
        -- (1) branch_idx/run_id
        -- (2) run_id/run.manifest 
        3. Post-commit - 
        -- (1) commit_idx/run_id 
        -- (2) rewrite run.manifest
        If step a.2 and b.2 ended successfully, it is filterable by branch name and commit id compatibly. 
    - We will need to add a verification step on the filter consumer side. Before displaying an action run by branch, verify that its run.manifest file exists, and fails otherwise. That is, If step a.1 succeeded but step a.2 failed, the consumer of `branch_idx/run_id` will exclude this run.
4. How to update the commit id associated with an action run
    - Add the commit index file (see section 3, step b.1)
    - The post-commit stage will rewrite the run.manifest file adding a commit id to it.

PR #3422, opened 2022-05-24
KV Store: Actions implementation

N-o-Z:


pkg/actions/kv_run_results_iterator.go
```diff
@@ -0,0 +1,78 @@
...
+		it:       *it,
+		ctx:      ctx,
+		branchID: branchID,
+		commitID: commitID,
+	}
+}
+
+func (i *KVRunResultIterator) Next() bool {
+	entry := kv.NewMessageEntry((&RunResultData{}).ProtoReflect().Type())
+	for i.it.Next() {
```
itaiad200, line 36: The performance here would be poor. You would always have to scan and parse all runs just to get a few. I realize that we rely on Postgres for easy filtering until now, but we can't read everything all runs every time a user clicks on a `Actions` tab in the UI.

N-o-Z: The only way to avoid it is by adding the branch id and/or commit id the the key's path. WDYT? It will mean that we will need to read only the key.

itaiad200: Do we even use `SetIf` in Actions kv implementation? If not, we'll face the same issues as in #1511

N-o-Z: I don't really understand how SetIf is connected to the result iterator. Can you explain?

pkg/actions/service.go
```diff
@@ -260,29 +331,50 @@ func (s *Service) saveRunManifestObjectStore(ctx context.Context, manifest RunMa
...
+func (s *KVService) storeRun(ctx context.Context, run *RunResultData, repoID string) error {
+	runKey := getRunPath(repoID, run.RunId)
+	err := s.Store.SetMsg(ctx, runKey, run)
+	if err != nil {
+		return fmt.Errorf("save run result (runKey %s): %w", runKey, err)
+	}
+
+	// Save secondary index by BranchID
+	bk := kv.FormatPath(getRunByBranchPath(repoID, run.BranchId), run.RunId)
+	err = s.Store.SetIf(ctx, []byte(bk), []byte(runKey), nil)
```
itaiad200, line 365: You implicitly created a format for secondary indexes. I'm fine with the simplicity of it, I just wonder if it doesn't need a type of its own. When this simple schema evolves, you won't have proto for backward compatibility.

pkg/actions/service.go
```diff
@@ -326,117 +418,117 @@ func buildRunManifestFromTasks(record graveler.HookRecord, tasks [][]*Task) RunM
...
-		}
-		// return if nothing was updated
-		if res.RowsAffected() == 0 {
-			return nil, nil
+	runKey := getRunPath(repositoryID, runID)
+	run := RunResultData{}
+	_, err := s.Store.GetMsg(ctx, runKey, &run)
+	if err != nil {
+		if errors.Is(err, kv.ErrNotFound) {
+			return nil // nothing to update
```
itaiad200, line 431: How is this possible?

pkg/actions/service.go
```diff
@@ -326,117 +418,117 @@ func buildRunManifestFromTasks(record graveler.HookRecord, tasks [][]*Task) RunM
...
 		}
+		return fmt.Errorf("run id: %w", err)
+	}
+	if run.CommitId == commitID { // return if no update is required
+		return nil
```
itaiad200, line 436: And this? Which flow brings us here?

pkg/actions/service.go
```diff
@@ -326,117 +418,117 @@ func buildRunManifestFromTasks(record graveler.HookRecord, tasks [][]*Task) RunM
...
+	}
 
-		// read run information
-		runResult, err := s.getRunResultTx(tx, repositoryID, runID)
-		if err != nil {
-			return nil, err
-		}
-		manifest = &RunManifest{Run: *runResult}
+	// delete secondary keys
+	err = s.deleteRunSecondaryKeys(ctx, &run, repositoryID)
```
itaiad200, line 440: Why are we deleting these?

pkg/actions/service.go
```diff
@@ -260,29 +331,50 @@ func (s *Service) saveRunManifestObjectStore(ctx context.Context, manifest RunMa
...
-		}
-		return nil, nil
-	})
-	return err
+	return nil
+}
+
+func (s *KVService) storeRun(ctx context.Context, run *RunResultData, repoID string) error {
+	runKey := getRunPath(repoID, run.RunId)
+	err := s.Store.SetMsg(ctx, runKey, run)
```
itaiad200, line 358: I think the order here is wrong. 

If you first set the Run and then the secondary indexes, it's much harder to check if you're in inconsistent state. For example, you created the Run, but failed to create the indexes. Getting all runs will return this specific run. To validate that the state is consistent, you would have to look for all secondary indexes, which some are ok to be missing - like with branch index for tags.

If you set the secondary indexes first, the inconsistent state is to have it pointing to a non-existing Run. I think it's easier to validate because (a) the index is pointing at it (b) in most flows you would need to read it regardless.

pkg/actions/service.go
```diff
@@ -326,117 +418,117 @@ func buildRunManifestFromTasks(record graveler.HookRecord, tasks [][]*Task) RunM
...
-			repositoryID, runID, hookRunID)
-		if err != nil {
-			return nil, fmt.Errorf("get task result: %w", err)
+func (s *KVService) GetRunResult(ctx context.Context, repositoryID string, runID string) (*RunResult, error) {
+	runKey := getRunPath(repositoryID, runID)
+	m := RunResultData{}
+	_, err := s.Store.GetMsg(ctx, runKey, &m)
+	if err != nil {
+		if errors.Is(err, kv.ErrNotFound) {
+			err = fmt.Errorf("%s, %w", err, ErrNotFound) // Wrap error for compatibility with DBService
```
itaiad200, line 502: I'm not so sure what's the type of the returned error here. 
```suggestion
			err = fmt.Errorf("%s: %w", err, ErrNotFound) // Wrap error for compatibility with DBService
```

pkg/actions/service.go
```diff
@@ -70,28 +82,87 @@ func (r *TaskResult) LogPath() string {
...
+}
+
+func getRunByBranchPath(repoID, branchID string) string {
+	return kv.FormatPath(actionsPrefix, repoID, branchPrefix, branchID)
+}
 
-func NewService(ctx context.Context, db db.Database, source Source, writer OutputWriter, stats stats.Collector, runHooks bool) *Service {
+func getRunByCommitPath(repoID, commitID string) string {
+	return kv.FormatPath(actionsPrefix, repoID, commitPrefix, commitID)
+}
```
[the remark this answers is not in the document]
itaiad200: A suggestion to treat this as a REST API, longer but more readable paths. I think it would also help us expanding the API as we continue to improve the actions service.

1. Run path: 
	`actions/repos/<repoID>/runs/<runID>`
2. Tasks per Run path:
	`actions/repos/<repoID>/runs/<runID>/tasks/<hookRunID>`
3. Run by Branch path:
	`actions/repos/<repoID>/branches/<branchID>/runs/<runID>`
4. Run by Commit path:
	`actions/repos/<repoID>/commits/<commitID>/runs/<runID>`

pkg/api/auth_middleware_test.go
```diff
@@ -12,8 +12,16 @@ import (
 	"github.com/treeverse/lakefs/pkg/auth/model"
 )
 
-func TestAuthMiddleware(t *testing.T) {
```
itaiad200, line 15: How was the `AuthMiddleware` impacted by this PR? 
The downside for testing everything twice is the runtime of the tests. It's not a big deal though..

pkg/api/db_controller_test.go
```diff
@@ -0,0 +1,2218 @@
+package api_test
```
itaiad200, line 1: Copying this amount of code, even if it's a test code, would lead to a huge maintenance overhead and possible bugs where new test code is only written in one of these. Maybe an env var to switch between the two DB implementations would save us the trouble?

pkg/api/serve_test.go
```diff
@@ -187,32 +203,49 @@ func setupClientWithAdminAndWalkerFactory(t testing.TB, factory catalog.WalkerFa
...
+			kvEnabled: true,
+		},
 	}
-	expectedErrMsg := api.ErrInvalidAPIEndpoint.Error()
-	errMsg := resp.JSONDefault.Message
-	if errMsg != expectedErrMsg {
-		t.Fatalf("client response error message: %s, expected: %s", errMsg, expectedErrMsg)
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			handler, _ := setupHandler(t, tt.kvEnabled)
```
itaiad200, line 221: `kvEnabled` value is meaningless for this test, so why should we test this?

pkg/kv/store_message.go
```diff
@@ -51,62 +51,133 @@ func (s *StoreMessage) DeleteMsg(ctx context.Context, path string) error {
...
 }
 
 type MessageEntry struct {
 	Key   string
 	Value protoreflect.ProtoMessage
 }
 
-func NewMessageIterator(ctx context.Context, store Store, msgType protoreflect.MessageType, prefix string) (*MessageIterator, error) {
+func NewMessageEntry(msg protoreflect.MessageType) *MessageEntry {
+	return &MessageEntry{
```
itaiad200, line 64: It's strange that a `NewMessageEntry` returns a  `MessageEntry` with uninitialzed fields.

pkg/kv/store_message.go
```diff
@@ -51,62 +51,133 @@ func (s *StoreMessage) DeleteMsg(ctx context.Context, path string) error {
...
+		Key:   "",
+		Value: msg.New().Interface(),
+	}
+}
+
+type MessageIterator interface {
+	Next() bool
+	Entry(key *string, value *protoreflect.ProtoMessage) error
```
itaiad200, line 72: What does `Entry(...)` do? please document.

pkg/kv/store_message.go
```diff
@@ -51,62 +51,133 @@ func (s *StoreMessage) DeleteMsg(ctx context.Context, path string) error {
...
-func (m *MessageIterator) Next() bool {
+func (m *PrimaryIterator) Next() bool {
 	if m.err != nil {
 		return false
 	}
 	return m.itr.Next()
 }
 
-func (m *MessageIterator) Entry() *MessageEntry {
+func (m *PrimaryIterator) Entry(key *string, value *protoreflect.ProtoMessage) error {
```
itaiad200, line 99: Now I get it. I think we should be aligned with the original store implementation:
```suggestion
func (m *PrimaryIterator) Entry(value *protoreflect.ProtoMessage) (*MessageEntry, error) {
```

pkg/kv/store_message.go
```diff
@@ -51,62 +51,133 @@ func (s *StoreMessage) DeleteMsg(ctx context.Context, path string) error {
...
-		return nil
+		return ErrNotFound
 	}
-	msg := m.msgType.New().Interface()
-	err := proto.Unmarshal(entry.Value, msg)
-	if err != nil {
-		m.err = fmt.Errorf("unmarshal proto data for key %s: %w", string(entry.Key), err)
-		return nil
+	if value != nil {
+		err := proto.Unmarshal(entry.Value, *value)
```
itaiad200, line 105: You're passing the value type of `value`. Is this on purpose ?

pkg/kv/store_message.go
```diff
@@ -51,62 +51,133 @@ func (s *StoreMessage) DeleteMsg(ctx context.Context, path string) error {
...
 
-func (m *MessageIterator) Close() {
+func (m *PrimaryIterator) Close() {
 	m.itr.Close()
 }
+
+// SecondaryIterator MessageIterator implementation for secondary key
+// The iterator iterates over the given prefix, extracts the primary key value from secondary key and then returns
+// the proto message and primary key
+type SecondaryIterator struct {
```
itaiad200, line 130: If we'll use proto for secondary indexes, we won't have to create special iterators for them.

pkg/kv/store_message.go
```diff
@@ -51,62 +51,133 @@ func (s *StoreMessage) DeleteMsg(ctx context.Context, path string) error {
...
+
+func (s *SecondaryIterator) Entry(key *string, value *protoreflect.ProtoMessage) error {
+	secondary := s.itr.Entry()
+	if secondary == nil {
+		return ErrNotFound
+	}
+
+	if value != nil {
+		data, err := s.store.Get(s.ctx, secondary.Value)
+		if err != nil {
```
itaiad200, line 160: Notice that ErrNotFound here needs to be ignored. Otherwise you may always find yourself in a failure loop.
Inconsistencies may occur in the KV world and the user (e.g. the action service) needs to be aware of that and handle gracefully to secure the operation of the service in all possible states.

pkg/kv/store_message_test.go
```diff
@@ -19,8 +19,7 @@ const modelPrefix = "tm"
```
itaiad200, line 19: I think we're missing tests for the new Iterators and all the possible states that the store may be at in any given time.

pkg/actions/service_test.go
```diff
@@ -33,108 +44,44 @@ func NewActionStatsMockCollector() ActionStatsMockCollector {
...
 
-			username, pass, ok := r.BasicAuth()
-			require.True(t, ok)
-			require.Equal(t, "some_username", username)
-			require.Equal(t, "some_password", pass)
+func GetKVService(t *testing.T, ctx context.Context, source actions.Source, writer actions.OutputWriter, stats stats.Collector, runHooks bool) actions.Service {
+	t.Helper()
+	kvStore := kvtest.GetStore(ctx, t)
+	return actions.NewService(ctx, kv.StoreMessage{Store: kvStore}, source, writer, stats, runHooks)
+}
```
itaiad200, line 66: Great stuff!! If only Github would know to recognize that indentations are not a change..

itaiad200 (commented):
Thanks for this impactful change, a lot of great stuff inside.

I think we should be very careful with secondary indexes and take into account that every key we set may be the last before the app crashes. We should be able to write services on top of the KV that are model to handle all KV states.

pkg/actions/kv_run_results_iterator.go
```diff
@@ -0,0 +1,73 @@
...
+	} else {
+		it, err = kv.NewPrimaryIterator(ctx, store.Store, key)
+		if err != nil {
+			return nil
+		}
+	}
+
+	return &KVRunResultIterator{
+		it:  it,
+		ctx: ctx,
```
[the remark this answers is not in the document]
N-o-Z: You are right - it's a leftover from previous implementation

pkg/actions/kv_run_results_iterator.go
```diff
@@ -0,0 +1,73 @@
...
+		key = getRunPath(repositoryID, prefix)
+		secondary = false
+	}
+
+	var it kv.MessageIterator
+	if secondary {
+		it, err = kv.NewSecondaryIterator(ctx, store.Store, key)
+		if err != nil {
+			return nil
+		}
```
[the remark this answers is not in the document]
N-o-Z: Fixed

pkg/actions/kv_run_results_iterator.go
```diff
@@ -0,0 +1,73 @@
...
+func (i *KVRunResultIterator) Next() bool {
+	return i.it.Next()
+}
+
+func (i *KVRunResultIterator) Value() *RunResult {
+	if i.Err() != nil {
+		return nil
+	}
+	var v proto.Message = &RunResultData{}
+	err := i.it.Entry(nil, &v)
```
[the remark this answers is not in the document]
N-o-Z: v is of type proto.Message and we are passing a reference to it.

pkg/actions/kv_run_results_iterator.go
```diff
@@ -0,0 +1,73 @@
...
+	if err != nil {
+		logging.Default().WithError(err).Errorf("value failed")
+	}
+	return runResultFromProto(v.(*RunResultData))
```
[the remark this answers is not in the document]
N-o-Z: Fixed

pkg/actions/service.go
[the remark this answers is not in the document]
N-o-Z: 1-2: It's not a good idea to nest the paths of different data models - it will complicate scanning
3-4: Done

pkg/actions/service.go
```diff
@@ -260,29 +331,50 @@ func (s *Service) saveRunManifestObjectStore(ctx context.Context, manifest RunMa
...
+
+	// Save secondary index by BranchID
+	bk := kv.FormatPath(getRunByBranchPath(repoID, run.BranchId), run.RunId)
+	err = s.Store.SetIf(ctx, []byte(bk), []byte(runKey), nil)
+	if err != nil {
+		return fmt.Errorf("save secondary index by branch (key %s): %w", bk, err)
+	}
+
+	// Save secondary index by CommitID
+	ck := kv.FormatPath(getRunByCommitPath(repoID, run.CommitId), run.RunId)
```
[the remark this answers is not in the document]
N-o-Z: Right, added a check for both BranchID and CommitID, if they are empty the secondary index will not be added.
This means that when filtering runs by commit ID we wont see any Tag events - which makes sense in my opinion

pkg/actions/service.go
N-o-Z: Since we are updating the commit id - the secondary index is no longer valid and we need to recreate it

pkg/actions/service.go
N-o-Z: Agree

pkg/api/auth_middleware_test.go
N-o-Z: Most of the tests in the api package are not impacted by the changes. I introduced the kv flavor to tests using the hookHandler to make sure nothing got accidentally broken.

pkg/api/db_controller_test.go
N-o-Z: Currently we don't support env var for unit testing. If we want to do that we will have to:
1. Change current behavior to use the env var instead of running everything in 2 falvors
2. Add another job in the pipeline to run the unit tests with this env var activated.
I think its too much effort for a temporary need.
On this specific module there were far too many changes to make to support KV enabled mode, so I thought it was better to just duplicate the test module. If you think it's worth the work I can change this module to use the same conventions we used so far for unit testing with KV

pkg/api/serve_test.go
N-o-Z: I automatically added KV flavor for all test that required the hooksHandler.
Removed

pkg/kv/store_message.go
N-o-Z: What we wanted to do here is to avoid unnecessary object creations and pass the responsibility to the user. This way, the user provides us with the pointer to the object and we de-serialize the data to it.
If we already return the data on the func argument - what do you propose we return in the *MessageEntry?

pkg/actions/service.go
N-o-Z: This can happen whenever a call to UpdateCommitID is performed unnecessarily. This check simply saves a lot of time both updating the storage and DB. I don't see the harm of keeping it

pkg/actions/service.go
N-o-Z: No real way for this to happen currently - I'll remove this check

pkg/kv/store_message.go
N-o-Z: The struct is no longer relevant after the refactoring

cmd/lakefs/cmd/import.go
```diff
@@ -121,15 +122,34 @@ func runImport(cmd *cobra.Command, args []string) (statusCode int) {
...
+			cfg.GetActionsEnabled(),
+		)
+	} else {
+		actionsService = actions.NewDBService(
+			ctx,
+			dbPool,
+			catalog.NewActionsSource(c),
+			catalog.NewActionsOutputWriter(c.BlockAdapter),
+			bufferedCollector,
+			cfg.GetActionsEnabled(),
```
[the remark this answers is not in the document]
N-o-Z: I think it will add too much logic that will eventually be deleted.

pkg/actions/kv_run_results_iterator.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/treeverse/lakefs/pkg/kv"
+)
+
+type KVRunResultIterator struct {
+	it  kv.MessageIterator
+	ctx context.Context
+}
+
+func NewKVRunResultIterator(ctx context.Context, store kv.StoreMessage, repositoryID, branchID, commitID, prefix string) *KVRunResultIterator {
```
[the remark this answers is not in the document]
N-o-Z: Made several changes as there was an error in the logic. We now accept 2 arguments "prefix" and "after"
prefix - bounds the key range on which we iterate over
after - used for pagination to provide the next set of entries

pkg/actions/service.go
N-o-Z: Please note that this is a change of behavior from DB service implementation when UPDATE returns no rows affected. As result I must remove some test logic in service_test so we won't fail on it

pkg/kv/store_message.go
N-o-Z: I understand your comment about defining a protobuf for secondary indexes but I think we will still have to create a special iterator for secondary indices since the Entry functionality is different.
Maybe we can discuss it further?

pkg/kv/store_message.go
N-o-Z: What do you think the expected behavior should be in this case? Should we iterate on secondary indices until a valid one is found? Should we return the ErrNotFound and let the caller handle this special case, should we return nil?

pkg/actions/service.go
```diff
@@ -260,29 +331,50 @@ func (s *Service) saveRunManifestObjectStore(ctx context.Context, manifest RunMa
...
+
+func (s *KVService) storeRun(ctx context.Context, run *RunResultData, repoID string) error {
+	runKey := getRunPath(repoID, run.RunId)
+	err := s.Store.SetMsg(ctx, runKey, run)
+	if err != nil {
+		return fmt.Errorf("save run result (runKey %s): %w", runKey, err)
+	}
+
+	// Save secondary index by BranchID
+	bk := kv.FormatPath(getRunByBranchPath(repoID, run.BranchId), run.RunId)
```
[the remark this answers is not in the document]
N-o-Z: Resolved

pkg/kv/store_message_test.go
N-o-Z: Added

pkg/kv/scan.go
```diff
@@ -43,8 +43,12 @@ func (b *PrefixIterator) Close() {
...
 
 // ScanPrefix returns an iterator on store that scan the set of keys that start with prefix
-func ScanPrefix(ctx context.Context, store Store, prefix []byte) (EntriesIterator, error) {
-	iter, err := store.Scan(ctx, prefix)
+func ScanPrefix(ctx context.Context, store Store, prefix, after []byte) (EntriesIterator, error) {
+	start := prefix
+	if len(after) > 0 {
+		start = after
+	}
+	iter, err := store.Scan(ctx, start)
```
itaiad200, line 51: From the implementation I understand that `after` is a full key and not a relative one under prefix. A documentation of this is missing. Also - should we validate it?

pkg/kv/store_message.go
```diff
@@ -51,95 +52,91 @@ func (s *StoreMessage) DeleteMsg(ctx context.Context, path string) error {
...
-		Key:   "",
-		Value: msg.New().Interface(),
-	}
-}
-
 type MessageIterator interface {
 	Next() bool
-	Entry(key *string, value *protoreflect.ProtoMessage) error
+	// Entry Receives key and value pointers from user and populates them with the iterator entry values
+	// The function will try to populate the func arguments which are not nil
```
itaiad200, line 67: Outdated? It receives nothing..

pkg/kv/store_message.go
```diff
@@ -51,95 +52,91 @@ func (s *StoreMessage) DeleteMsg(ctx context.Context, path string) error {
...
-	if value != nil {
-		err := proto.Unmarshal(entry.Value, *value)
-		if err != nil {
-			return fmt.Errorf("unmarshal proto data for key %s: %w", string(entry.Key), err)
-		}
-	}
-	if key != nil {
-		*key = string(entry.Key)
+	value := i.msgType.New().Interface()
+	err := proto.Unmarshal(entry.Value, value)
```
itaiad200, line 102: Shouldn't we move this logic to `Next`? 
The reason I'm asking is that the user calls `Next()` and gets true. Now he calls `Entry` and gets a `nil` with a new error. The iterator usage we're accustomed to is that when `Next()` returns `true` you would **always** have a value to read. This is how we assemble the iterators across the board.
To have the same functionality here, you would have to move the parsing code to `Next` and store the value.

pkg/kv/store_message.go
```diff
@@ -51,95 +52,91 @@ func (s *StoreMessage) DeleteMsg(ctx context.Context, path string) error {
...
+	if err != nil {
+		i.err = fmt.Errorf("unmarshal proto data for key %s: %w", entry.Key, err)
+		return nil
 	}
-	return nil
+	return &MessageEntry{
+		Key:   string(entry.Key),
+		Value: value}
```
itaiad200, line 109: lint

pkg/kv/store_message.go
```diff
@@ -149,26 +146,45 @@ func (s *SecondaryIterator) Next() bool {
...
+		secondary = s.itr.Entry()
+		if secondary == nil {
+			return nil
+		}
+		next = secondary.Value.(*SecondaryIndex)
+	}
+	if err != nil {
+		s.err = fmt.Errorf("getting value from key (primary key %s): %w", next.PrimaryKey, err)
+		return nil
+	}
```
itaiad200, line 177: I understand what you're trying to do here, but I think there's a bug since for a successful `Get` from the `store`, you would continue to loop when you should exit the for loop.

itaiad200: And like my previous comment, this logic should move to `next`.

pkg/graveler/hooks_handler.go
```diff
@@ -22,6 +22,8 @@ const (
 	EventTypePostCreateBranch EventType = "post-create-branch"
 	EventTypePreDeleteBranch  EventType = "pre-delete-branch"
 	EventTypePostDeleteBranch EventType = "post-delete-branch"
+
+	RunIDTimeLayout = "20060102150405"
```
itaiad200, line 26: How is this layout decreasing with time?

pkg/actions/kv_run_results_iterator.go
```diff
@@ -3,71 +3,77 @@ package actions
...
 type KVRunResultIterator struct {
 	it  kv.MessageIterator
-	ctx context.Context
+	err error
 }
 
-func NewKVRunResultIterator(ctx context.Context, store kv.StoreMessage, repositoryID, branchID, commitID, prefix string) *KVRunResultIterator {
-	var key string
-	var err error
+func NewKVRunResultIterator(ctx context.Context, store kv.StoreMessage, repositoryID, branchID, commitID, after string) (*KVRunResultIterator, error) {
```
itaiad200, line 14: Are we allowing a mix branch + commitID ? Maybe we should assert on that?

pkg/actions/kv_run_results_iterator.go
```diff
@@ -3,71 +3,77 @@ package actions
...
-			return nil
+			return nil, err
 		}
 	} else {
-		it, err = kv.NewPrimaryIterator(ctx, store.Store, key)
+		it, err = kv.NewPrimaryIterator(ctx, store.Store, (&RunResultData{}).ProtoReflect().Type(), prefix, after)
 		if err != nil {
-			return nil
+			return nil, err
 		}
```
itaiad200, line 44: Very elegant 🧐

esti/webhook_server_test.go
```diff
@@ -103,6 +103,7 @@ func hookHandlerFunc(respCh chan hookResponse) func(http.ResponseWriter, *http.R
 			return
 		}
 
+		time.Sleep(1 * time.Second) // Added sleep to differentiate between event timestamps
```
itaiad200, line 106: I hate it and understand why we need it.
The funny thing is: the test may still fail due to the same reason..

esti/hooks_test.go
```diff
@@ -18,7 +19,18 @@ import (
 )
 
 //go:embed action_files/*.yaml
-var actions embed.FS
+var actionsPath embed.FS
+
+var (
+	resultVal = make([]*webhookEventInfo, 0)
+	mu        sync.RWMutex
+)
```
itaiad200, line 27: You just added an `esti` package level mutex. Can you create a simple struct for it to avoid misuse?

pkg/actions/service_test.go
```diff
@@ -290,12 +290,6 @@ func TestDisableHooksRun(t *testing.T) {
 				t.Fatalf("Run() failed with err=%s", err)
 			}
 
-			// update commit using post event record
-			err = actionsService.UpdateCommitID(ctx, record.RepositoryID.String(), record.StorageNamespace.String(), record.RunID, "commit1")
```
itaiad200: Why did you decide to remove it?

pkg/actions/kv_task_results_iterator.go
```diff
@@ -3,48 +3,54 @@ package actions
...
+	err error
 }
 
-func NewKVTaskResultIterator(ctx context.Context, store kv.StoreMessage, repositoryID, runID, after string) *KVTaskResultIterator {
-	key := kv.FormatPath(getTasksPath(repositoryID, runID), after)
-	it, err := store.Scan(ctx, key)
+func NewKVTaskResultIterator(ctx context.Context, store kv.StoreMessage, repositoryID, runID, after string) (*KVTaskResultIterator, error) {
+	prefix := GetTasksPath(repositoryID, runID)
+	if after != "" {
+		after = kv.FormatPath(prefix, after)
```
itaiad200, line 17: Now that changes the behaviour you had in some different iterator where `after` was the full key. here you treat it as relative.

pkg/actions/service.go
```diff
@@ -132,20 +138,32 @@ func protoFromTaskResult(m *TaskResult) *TaskResultData {
 	}
 }
 
-func getTasksPath(repoID, runID string) string {
-	return kv.FormatPath(actionsPrefix, repoID, taskPrefix, runID)
+func GetBaseActionsPath(repoID string) string {
```
itaiad200, line 141: Sorry for the late comment. Basically the go convention is:
```suggestion
func BaseActionsPath(repoID string) string {
```

Go hates `Get`/`Set` prefixes.

pkg/actions/service.go
```diff
@@ -5,30 +5,36 @@ package actions
...
-	runPrefix     = "run"
-	taskPrefix    = "task"
-	branchPrefix  = "runByBranch"
-	commitPrefix  = "runByCommit"
+	reposPrefix   = "repos"
+	RunsPrefix    = "runs"
+	tasksPrefix   = "tasks"
+	branchPrefix  = "branches"
+	commitPrefix  = "commits"
+	unixYear3000  = 32500915200000
```
itaiad200, line 31:
```suggestion
	unixYear3000Milli  = 32500915200000
```

pkg/actions/service.go
```diff
@@ -586,6 +579,12 @@ func (s *KVService) PostDeleteBranchHook(_ context.Context, record graveler.Hook
 	s.asyncRun(record)
 }
 
+func (s *KVService) NewRunID() string {
+	entropy := ulid.Monotonic(rand.Reader, 0)
+	tm := time.UnixMilli(unixYear3000 - time.Now().UnixMilli()).UTC()
+	return ulid.MustNew(ulid.Timestamp(tm), entropy).String()
+}
```
itaiad200, line 586: I think I get the ** unixYear3000** semantics but not the entropy part.
Would you mind explaining everything in a comment? It's so easy to change this later on and break the actions listing.

itaiad200 (changes requested):
Looks great! still have some questions around the iterators but we're very close 🦾

pkg/actions/service.go
N-o-Z: Changed it back to Unix seconds

pkg/actions/kv_task_results_iterator.go
N-o-Z: After is a full key only in internal iterators - PrefixIterator
Run/Task iterators get the after from outside, user is not aware of full key - each pkg iterator will need to take the 'after' which will be a know (exposed) identifier and create the full key according to its logic

pkg/actions/service_test.go
N-o-Z: Removed due to previous changes in behavior which were decided. Added it back

pkg/actions/kv_run_results_iterator.go
N-o-Z: Current behavior in DB service is to first check branch and then commit - if branch and commit are provided only branch is considered, we do not error in that case.
I'm not sure we want to change this behavior - maybe we should do the validation on the controller side?

pkg/graveler/hooks_handler.go
N-o-Z: This is the layout that was hardcoded in graveler's NewRunID function, I just put it in a const. This is used only by the DB service

pkg/kv/scan.go
N-o-Z: We should probably sanitize the value when getting it from the user, I will add it and the documentation

cmd/lakefs/cmd/run.go
```diff
@@ -192,17 +179,41 @@ var runCmd = &cobra.Command{
...
+			multipartsTracker = multiparts.NewDBTracker(dbPool)
+			actionsService = actions.NewDBService(
+				ctx,
+				dbPool,
+				catalog.NewActionsSource(c),
+				catalog.NewActionsOutputWriter(c.BlockAdapter),
+				bufferedCollector,
+				cfg.GetActionsEnabled(),
+			)
+		}
```
itai-david, line 212: Just thinking out loud:
The following pattern repeats in various locations:
```
if dbParam.KVEnabled {
    OpenKV
    CreateSomeServiceWithKV
} else {
    CreateSomeServiceWithDB
}
```
It is (IMHO) hard to maintain, as it will keep growing with more modules `joining` club KV
Moreover, in case of changes - e.g. an addiional parameter is required to fix some bug etc - it will require multiple changes in multiple locations. There is, of course, the inevitable mega-modification of removing all DB support, once we are done implementing KV on all modules, and that, too, is going to be quite of an headache, considering the multiple locations that will require code change.
Long story short - maybe the `if dbParam.KVEnabled` decision should be moved to `actions.NewService` hiding  and call either `NewDBService`  or `NewSomethingElseService` and hiding the implementation details (Having a feature flag which controls the DB type) from the calling modules - `run` and `import` in that case, but can be more places for other modules.
This can aslo support possible future changes, similar to the move to KV, that may happen in the future.

pkg/actions/kv_run_results_iterator.go
itaiad200: Still - I think that a function accepting arguments should handle them, or assert if the input isn't valid (like a branch & commit combo).

pkg/kv/scan.go
```diff
@@ -43,6 +43,7 @@ func (b *PrefixIterator) Close() {
 }
 
 // ScanPrefix returns an iterator on store that scan the set of keys that start with prefix
+// after is the full key name for which to start the scan from
 func ScanPrefix(ctx context.Context, store Store, prefix, after []byte) (EntriesIterator, error) {
```
itaiad200, line 47: The behaviour of `after<prefix` is an empty iterator, is this expected?

pkg/actions/kv_task_results_iterator.go
itaiad200: Thanks for adding documentation for it.

pkg/kv/store_message.go
itaiad200: Oh right, sneaky you

pkg/kv/scan.go
N-o-Z: I believe this is the behavior we want to have. The other option is to validate and return an error, but I prefer this option less

cmd/lakefs/cmd/run.go
N-o-Z: The "NewKVService" requires creating a KV store and ensuring it is closed when we are done with it. This will require to create and pass a KVstore **even if** it is not required for each service we want to create.
For now there are only 2 places we preform the flow. I suggest to defer it until we find this to be a real issue, also keep in mind this is a temporary code.

pkg/actions/kv_run_results_iterator.go
N-o-Z: I can change that - but we need to understand this is a change in behavior

esti/hooks_test.go
```diff
@@ -29,18 +46,33 @@ func TestHooksSuccess(t *testing.T) {
...
+
 	})
 	t.Run("create delete branch test", func(t *testing.T) {
 		testCreateDeleteBranch(t, ctx, repo)
 	})
 	t.Run("create delete tag test", func(t *testing.T) {
 		testCreateDeleteTag(t, ctx, repo)
 	})
+
+	t.Log("check runs are sorted in descending order")
```
itaiad200, line 69: Thanks for adding this check, we didn't realize how important and tricky this might be until this PR.

pkg/actions/kv_run_results_iterator.go
itaiad200: The new verification will do for now.

pkg/kv/postgres/store.go
```diff
@@ -32,7 +32,7 @@ const (
 	DefaultTableName = "kv"
 	paramTableName   = "lakefskv_table"
 
-	// Changing the below value means repartitioning and probably a migration.
+	// DefaultPartitions Changing the below value means repartitioning and probably a migration.
```
itaiad200, line 35: I wonder why your linter is stricter than mine..

itaiad200 (commented):
Looks good, working to make the tests run in 2 modes in a side-branch

pkg/kv/postgres/store.go
N-o-Z: ¯\_(ツ)_/¯

pkg/kv/scan.go
itaiad200: I think this is a confusing situation. IMO `after<prefix` should behave like `after=""`

pkg/actions/service.go
```diff
@@ -260,29 +349,53 @@ func (s *Service) saveRunManifestObjectStore(ctx context.Context, manifest RunMa
...
+func (s *KVService) saveRunManifestDB(ctx context.Context, repositoryID graveler.RepositoryID, manifest RunManifest) error {
+	// insert run information
+	run := manifest.Run
+	err := s.storeRun(ctx, protoFromRunResult(&run), repositoryID.String())
+	if err != nil {
+		return err
+	}
+
+	// insert each task information
+	for i := range manifest.HooksRun {
```
itaiad200, line 361: Shouldn't the `hook_runs` be stored before the `run` itself?
Otherwise we'll have partial information if we'll save only the `run` without the `hooks`.

itaiad200 (changes requested):
Thanks - 2 more open issues:
1. `after<prefix`
2. Hooks run storage should precede run storage.

PR #3423, opened 2022-05-25
Simplify kv message store by embed the store

nopcoder:
Embed kv store into message store.
It simplifies with the message store as it will continue to function as a store and enrich more message type operations.


N-o-Z (approved):
Looks great! Thanks!

PR #3425, opened 2022-05-25
KV auth models

### LINKED ISSUE #3349: Create KV Auth models

eden-ohana:
closes #3349


pkg/auth/model/model.go
```diff
@@ -6,6 +6,8 @@ import (
 	"errors"
 	"time"
 
+	"google.golang.org/protobuf/types/known/timestamppb"
+
```
N-o-Z, line 10: Remove the redundant newline?

pkg/auth/model/model.proto
```diff
@@ -0,0 +1,53 @@
...
+
+import "google/protobuf/timestamp.proto";
+
+package io.treeverse.lakefs.auth.model;
+
+message UserData {
+    int64 ID = 1;
+    google.protobuf.Timestamp CreatedAt = 2;
+    string Username = 3;
+    // FriendlyName, if set, is a shorter name for the user than
```
N-o-Z, line 12: I don't think it's really necessary to document this parameter again.
I would add documentation on top of each message stating which go struct it models

pkg/auth/model/model.proto
```diff
@@ -0,0 +1,53 @@
+syntax = "proto3";
+option go_package = "github.com/treeverse/lakefs/auth/model";
+
+import "google/protobuf/timestamp.proto";
+
+package io.treeverse.lakefs.auth.model;
+
+message UserData {
+    int64 ID = 1;
```
itaiad200, line 9: Why do we need ID in the KV world?
I think all queries are DisplayName/Username based, no?

pkg/auth/model/model.go
```diff
@@ -122,3 +124,114 @@ type CredentialKeys struct {
 	AccessKeyID     string `json:"access_key_id"`
 	SecretAccessKey string `json:"secret_access_key"`
 }
+
+func userFromProto(pb *UserData) *User {
```
itaiad200, line 128: missing tests for all of the below conversions

itaiad200 (changes requested):

pkg/auth/model/model.proto
```diff
@@ -0,0 +1,53 @@
+syntax = "proto3";
+option go_package = "github.com/treeverse/lakefs/auth/model";
+
+import "google/protobuf/timestamp.proto";
+
+package io.treeverse.lakefs.auth.model;
+
+message UserData {
```
N-o-Z, line 8: The convention in proto files is CamelCase for messages and snake_case for variable names

pkg/auth/model/model.proto
```diff
@@ -0,0 +1,53 @@
...
+    int64 ID = 1;
+    google.protobuf.Timestamp CreatedAt = 2;
+    string Username = 3;
+    // FriendlyName, if set, is a shorter name for the user than
+    // Username.  Unlike Username it does not identify the user (it
+    // might not be unique); use it in the user's GUI rather than in
+    // backend code.
+    string FriendlyName = 4;
+    string Email = 5;
+    bytes EncryptedPassword = 6;
```
N-o-Z, line 18: What about source? Do we need to save this information?

pkg/auth/model/model.proto
```diff
@@ -0,0 +1,53 @@
...
+}
+
+message GroupData {
+    int64 ID = 1;
```
N-o-Z, line 22: For each go struct that holds an 'int' value - since it is an arch specific type we need to decid whether it is int32 or int64 and define it explicitly in the struct - otherwise we might encountered unexpected behavior

pkg/auth/model/model.proto
```diff
@@ -0,0 +1,53 @@
...
+    google.protobuf.Timestamp CreatedAt = 2;
+    string DisplayName = 3;
+}
+
+message PolicyData {
+    int64 ID = 1;
```
N-o-Z, line 28: Same as above comment

pkg/auth/model/model.proto
```diff
@@ -0,0 +1,53 @@
...
+    google.protobuf.Timestamp CreatedAt = 2;
+    string DisplayName = 3;
+    StatementsData Statements = 4;
```
N-o-Z, line 31: Are we saving StatmentData separately or as part of other data models? We need to decide how to save this data in the DB

N-o-Z (changes requested):
Thanks Eden!
Added some comment

pkg/auth/model/model.go
```diff
@@ -122,3 +124,114 @@ type CredentialKeys struct {
...
+		FriendlyName:      &pb.FriendlyName,
+		Email:             &pb.Email,
+		EncryptedPassword: pb.EncryptedPassword,
+	}
+}
+
+func protoFromUser(u *User) *UserData {
+	return &UserData{
+		ID:                u.ID,
+		CreatedAt:         timestamppb.New(u.CreatedAt),
```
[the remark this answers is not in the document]
N-o-Z: No - and when we created the data we need to use time.UTC()

N-o-Z (changes requested):
@eden-ohana we should also add the path definitions for all auth entities in DB so we can later use it in the implementation and migrations PRs in parallel

pkg/auth/model/model.proto
eden-ohana: we need it for the conversion between the proto and the original, we will probably not use the ID while in proto model

pkg/auth/model/model.go
eden-ohana: done

pkg/auth/model/model.proto
eden-ohana: Done

pkg/auth/model/model.go
```diff
@@ -6,6 +6,8 @@ import (
 	"errors"
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/model/model.proto
eden-ohana: done, thanks for the catch

pkg/auth/model/model.proto
eden-ohana: it is saved as a part of the policy model, but because it is not a regular type we need to convert it too

pkg/auth/model/model.proto
N-o-Z: The problem is that int is not a specific type, it can mean several types and depends on the system. Think about a situation where you save an int64 ID into the database on a 64 arch system, you export it and then try to import it in a 32bit arch -> this could potentially overflow. We should take this opportunity to set specific types in the structs

pkg/auth/model/model.proto
itaiad200: So I suggest we should remove it now? I don't see a reason for maintaining it when there are 0 paths using that field in the KV world and the DB world doesn't use this type.

pkg/auth/model/model.go
```diff
@@ -109,16 +126,165 @@ func (s *Statements) Scan(src interface{}) error {
 	return json.Unmarshal(data, s)
 }
 
-type Credential struct {
-	AccessKeyID                   string    `db:"access_key_id"`
-	SecretAccessKey               string    `db:"-" json:"-"`
-	SecretAccessKeyEncryptedBytes []byte    `db:"secret_access_key" json:"-"`
-	IssuedDate                    time.Time `db:"issued_date"`
-	UserID                        int64     `db:"user_id"`
+const authPrefix = "auth"
```
itaiad200, line 129: Document, prefix can mean a lot of things in general `model.go` file.

pkg/auth/model/model.go
```diff
@@ -109,16 +126,165 @@ func (s *Statements) Scan(src interface{}) error {
...
+
+//TODO - delete no lint comments when function is used
+//nolint
+func kvUserPath(Username string) string {
```
itaiad200, line 133:
```suggestion
func kvUserPath(username string) string {
```

pkg/auth/model/model.go
```diff
@@ -109,16 +126,165 @@ func (s *Statements) Scan(src interface{}) error {
...
+		Username:          u.Username,
+		FriendlyName:      *u.FriendlyName,
+		Email:             *u.Email,
+		EncryptedPassword: u.EncryptedPassword,
+		Source:            u.Source,
+	}
+}
+
+//nolint
+func kvGroupPath(DisplayName string) string {
```
itaiad200, line 164:
```suggestion
func kvGroupPath(displayName string) string {
```

pkg/auth/model/model.go
```diff
@@ -109,16 +126,165 @@ func (s *Statements) Scan(src interface{}) error {
...
+func protoFromGroup(g *Group) *GroupData {
+	return &GroupData{
+		Id:          int64(g.ID),
+		CreatedAt:   timestamppb.New(g.CreatedAt),
+		DisplayName: g.DisplayName,
+	}
+}
+
+//nolint
+func kvPolicyPath(DisplayName string) string {
```
itaiad200, line 187:
```suggestion
func kvPolicyPath(displayName string) string {
```

pkg/auth/model/model.go
```diff
@@ -109,16 +126,165 @@ func (s *Statements) Scan(src interface{}) error {
...
+	return &Policy{
+		ID:          int(pb.Id),
+		CreatedAt:   pb.CreatedAt.AsTime(),
+		DisplayName: pb.DisplayName,
+		Statement:   *statementsFromProto(pb.Statements),
+	}
+}
+
+//nolint
+func protoFromPolicy(g *Policy) *PolicyData {
```
itaiad200, line 202:
```suggestion
func protoFromPolicy(p *Policy) *PolicyData {
```

pkg/auth/model/model.go
```diff
@@ -109,16 +126,165 @@ func (s *Statements) Scan(src interface{}) error {
...
+	return &PolicyData{
+		Id:          int64(g.ID),
+		CreatedAt:   timestamppb.New(g.CreatedAt),
+		DisplayName: g.DisplayName,
+		Statements:  protoFromStatements(&g.Statement),
+	}
+}
+
+//nolint
+func kvCredentialPath(UserName string, AccessKeyID string) string {
```
itaiad200, line 212:
```suggestion
func kvCredentialPath(userName string, accessKeyID string) string {
```

pkg/auth/model/model.go
```diff
@@ -109,16 +126,165 @@ func (s *Statements) Scan(src interface{}) error {
...
+	for i := range *s {
+		statements[i] = protoFromStatement(&x[i])
+	}
+	return &StatementsData{
+		Statement: statements,
+	}
+}
+
+//nolint
+func kvUserToGroup(UserName string, DisplayName string) string {
```
itaiad200, line 278:
```suggestion
func kvUserToGroup(userName string, displayName string) string {
```

itaiad200: Display name of the group? It's confusion to use just the property name without the type when there are multiple types with the same property.

itaiad200: Here and elsewhere.

pkg/auth/model/model.go
```diff
@@ -109,16 +126,165 @@ func (s *Statements) Scan(src interface{}) error {
...
+	return kv.FormatPath(authPrefix, "groups", DisplayName, "users", UserName)
+}
+
+//nolint
+func kvPolicyToUser(DisplayName string, UserName string) string {
```
itaiad200, line 283:
```suggestion
func kvPolicyToUser(policyDisplayName string, userName string) string {
```

pkg/auth/model/model.go
```diff
@@ -109,16 +126,165 @@ func (s *Statements) Scan(src interface{}) error {
...
+	return kv.FormatPath(authPrefix, "users", UserName, "policies", DisplayName)
+}
+
+//nolint
+func kvPolicyToGroup(PolicyDisplayName string, GroupDisplayName string) string {
```
itaiad200, line 288:
```suggestion
func kvPolicyToGroup(policyDisplayName string, groupDisplayName string) string {
```

pkg/auth/model/model.proto
itaiad200: Actually, we decided something else in the [discovery](https://github.com/treeverse/lakeFS/pull/3405/files), so nm

pkg/auth/model/model.go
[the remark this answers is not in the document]
itaiad200: I don't understand. timestamps are created in many places in lakeFS, several of those in the auth package. Are we expecting to pass the responsibility of using the right TZ to every line of code creating timestamp to keep proto conversion correct? I think this layer of golang<>proto conversion needs to handle that, otherwise we'll be chasing our tail with TZ bugs.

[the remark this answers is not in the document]
N-o-Z: protobuf's timestamppb currently doesn't hold localization data, so I don't see a way of getting around it.

pkg/auth/model/model.go
eden-ohana: done

pkg/auth/model/model.go
eden-ohana: done

pkg/auth/model/model.go
eden-ohana: done

pkg/auth/model/model.go
eden-ohana: done

pkg/auth/model/model.go
eden-ohana: done

pkg/auth/model/model.go
eden-ohana: done

pkg/auth/model/model.go
eden-ohana: done

pkg/auth/model/model.go
eden-ohana: done

pkg/auth/model/model.go
```diff
@@ -89,9 +75,38 @@ type Statement struct {
...
-)
+type Credential struct {
+	AccessKeyID                   string    `db:"access_key_id"`
+	SecretAccessKey               string    `db:"-" json:"-"`
+	SecretAccessKeyEncryptedBytes []byte    `db:"secret_access_key" json:"-"`
+	IssuedDate                    time.Time `db:"issued_date"`
+	UserID                        int64     `db:"user_id"`
+}
+
+// For JSON serialization:
```
N-o-Z, line 86: Comment should have the following format 'CredentialKeys ...'

pkg/auth/model/model.go
```diff
@@ -111,96 +126,125 @@ func (s *Statements) Scan(src interface{}) error {
...
+//nolint
+func kvUserPath(userName string) string {
+	return kv.FormatPath(authPrefix, "users", userName)
 }
 
+//nolint
 func userFromProto(pb *UserData) *User {
 	return &User{
-		ID:                pb.ID,
+		ID:                int64(binary.LittleEndian.Uint64(pb.Id)),
```
N-o-Z, line 141: About this solution:
I'm not sure it works properly, and it unfortunately does not solve the problem I've mentioned before with the overflow.
I think the solution required here will have to include a schema change which will modify ID column to string / []byte and generate new IDs for all existing entities

pkg/auth/model/model.go
N-o-Z: @eden-ohana still not fixed :(

N-o-Z (changes requested):
@eden-ohana There's still an issue with the ID type which I think requires a wider solution

pkg/auth/model/model.go
eden-ohana: but this will happen in the migration

N-o-Z: For this to happen in the migration we need to create new structs for the KV models which will be used after the migration. Currently we are using the same structs which have the 'int' type for the ID fields - so the migration doesn't really help us here. What we need is to either do that or modify the current structs which requires DB schema version migration

nopcoder (commented):
@eden-ohana thanks for the models.  Need to remember the global IAM implementation currently is based on the current auth service and this PR changes the service. Have a BaseUser (or any Base) is confusing - even in the same interface, there are versions which return base and accept non-base.  It is common to use the same model and specify that in this case the identification is not use or not set if you wish to.
But I'll like to know the rational of having this change - the auth models were supposed to address the storage system (internal), so I didn't expect to see any change from the external one.

itaiad200:
> @eden-ohana thanks for the models. Need to remember the global IAM implementation currently is based on the current auth service and this PR changes the service. Have a BaseUser (or any Base) is confusing - even in the same interface, there are versions which return base and accept non-base. It is common to use the same model and specify that in this case the identification is not use or not set if you wish to. But I'll like to know the rational of having this change - the auth models were supposed to address the storage system (internal), so I didn't expect to see any change from the external one.

The global IAM implementation is unchanged. Notice that the external auth API doesn't implement the `auth.Service` API that was changed, the `APIAuthService` who calls that external service implements it. So changes were made to that code to supports the new models, but changes to the external service are unnecessary at this point.
I can expand on the rational for having 3 types instead of the single type we have today. Up until now, we had a single model with an `int` as the ID that was created by `postgres`. Moving away from `postgres` and not having a way to create serial IDs, we decided on a random `string` for ID. So we need at least 2 models for each type, based on the ID type. To reduce code duplication I extracted all the common properties to a `Base*` struct and embedded it in the appropriate types. That's why we have 3. This is all temporary, once `postgres` is history we can return to a single type. The idea of passing the `Base*` models in the API is to make it clearer that the ID is created in the service, I think it's clearer that way but I can change that too.

pkg/api/controller.go
```diff
@@ -187,7 +187,12 @@ func (c *Controller) Login(w http.ResponseWriter, r *http.Request, body LoginJSO
 
 	// user.Username will be different from username/access_key_id on
 	// LDAP login.  Use the stored value.
-	tokenString, err := GenerateJWTLogin(secret, user.ID, loginTime, expires)
+	n, err := auth.UserIDToInt64(user.ID)
```
N-o-Z, line 190: We should already advance to use the KV models which have a string ID and modify all the necessary code. We should use the DB models only in the transition to write/read from DB

pkg/api/controller.go
```diff
@@ -396,7 +401,7 @@ func (c *Controller) CreateGroup(w http.ResponseWriter, r *http.Request, body Cr
 	ctx := r.Context()
 	c.LogAction(ctx, "create_group")
 
-	g := &model.Group{
+	g := &model.BaseGroup{
```
N-o-Z, line 404: Same for all the base uses - We should use the KV models directly

pkg/auth/authenticator.go
```diff
@@ -99,7 +108,7 @@ func (ba *BuiltinAuthenticator) AuthenticateUser(ctx context.Context, username,
 	if subtle.ConstantTimeCompare([]byte(password), []byte(cred.SecretAccessKey)) != 1 {
 		return InvalidUserID, ErrInvalidSecretAccessKey
 	}
-	return cred.UserID, nil
+	return UserIDToInt64(cred.UserID)
```
N-o-Z, line 111: Same for all AuthenticateUser implementation

pkg/auth/authenticator.go
```diff
@@ -219,20 +228,19 @@ func (la *LDAPAuthenticator) AuthenticateUser(ctx context.Context, username, pas
...
 
-	user = &model.User{
+	newUser := &model.BaseUser{
 		CreatedAt:    time.Now(),
 		Username:     dn,
 		FriendlyName: &username,
 		Source:       "ldap",
 	}
-	id, err := la.AuthService.CreateUser(ctx, user)
+	id, err := la.AuthService.CreateUser(ctx, newUser)
```
N-o-Z, line 243: Same here - we need to change the client structs

pkg/gateway/middleware.go
```diff
@@ -54,7 +54,14 @@ func AuthenticationHandler(authService auth.GatewayService, next http.Handler) h
...
 		}
-		user, err := authService.GetUserByID(ctx, creds.UserID)
+		base, bitSize := 10, 64
+		intCredsID, err := strconv.ParseInt(creds.UserID, base, bitSize)
+		if err != nil {
+			logger.WithError(err).Warn("error parsing user ID")
+			_ = o.EncodeError(w, req, getAPIErrOrDefault(err, gatewayerrors.ErrAccessDenied))
+			return
+		}
+		user, err := authService.GetUserByID(ctx, intCredsID)
```
N-o-Z, line 64: Same as previous comments

itaiad200 (approved):

nopcoder:
> > @eden-ohana thanks for the models. Need to remember the global IAM implementation currently is based on the current auth service and this PR changes the service. Have a BaseUser (or any Base) is confusing - even in the same interface, there are versions which return base and accept non-base. It is common to use the same model and specify that in this case the identification is not use or not set if you wish to. But I'll like to know the rational of having this change - the auth models were supposed to address the storage system (internal), so I didn't expect to see any change from the external one.
> 
> The global IAM implementation is unchanged. Notice that the external auth API doesn't implement the `auth.Service` API that was changed, the `APIAuthService` who calls that external service implements it. So changes were made to that code to supports the new models, but changes to the external service are unnecessary at this point. I can expand on the rational for having 3 types instead of the single type we have today. Up until now, we had a single model with an `int` as the ID that was created by `postgres`. Moving away from `postgres` and not having a way to create serial IDs, we decided on a random `string` for ID. So we need at least 2 models for each type, based on the ID type. To reduce code duplication I extracted all the common properties to a `Base*` struct and embedded it in the appropriate types. That's why we have 3. This is all temporary, once `postgres` is history we can return to a single type. The idea of passing the `Base*` models in the API is to make it clearer that the ID is created in the service, I think it's clearer that way but I can change that too.

Thanks @itaiad200 for clarifying it to me.

nopcoder (changes requested):

pkg/auth/cache.go
```diff
@@ -9,13 +9,13 @@ import (
...
 type UserSetFn func() (*model.User, error)
-type UserPoliciesSetFn func() ([]*model.Policy, error)
+type UserPoliciesSetFn func() ([]*model.BasePolicy, error)
 
 type Cache interface {
 	GetCredential(accessKeyID string, setFn CredentialSetFn) (*model.Credential, error)
 	GetUser(username string, setFn UserSetFn) (*model.User, error)
-	GetUserByID(userID int64, setFn UserSetFn) (*model.User, error)
-	GetUserPolicies(userID string, setFn UserPoliciesSetFn) ([]*model.Policy, error)
+	GetUserByID(string string, setFn UserSetFn) (*model.User, error)
```
nopcoder, line 17: string string -> s? or something meaningful, except string :)

itaiad200: What's more meaningful than a `string string`? ;)

nopcoder (approved):

N-o-Z (approved):
Looks good, thanks!

PR #3427, opened 2022-05-25
Feature: kv actions migration 3347

N-o-Z:


pkg/actions/db_service.go
```diff
@@ -402,3 +414,134 @@ func (s *DBService) NewRunID() string {
...
+	// Create header
+	err := je.Encode(kv.Header{
+		LakeFSVersion:   version.Version,
+		PackageName:     packageName,
+		DBSchemaVersion: kv.InitialMigrateVersion,
+		CreatedAt:       time.Now().UTC(),
+	})
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
itaiad200: You can use it in all similar places.

pkg/actions/db_service.go
```diff
@@ -402,3 +414,134 @@ func (s *DBService) NewRunID() string {
...
+			Key:          key,
+			Value:        value,
+		})
+		if err != nil {
+			return err
+		}
+
+		// Add secondary keys branch
+		if m.BranchID != "" {
+			secKey := []byte(RunByBranchPath(m.RepoID, m.BranchID, newRunID))
```
itaiad200, line 475: I think we should leverage the code that exists in the Actions KV implementation to handle all of this, instead of copying it.

pkg/actions/kv_migrate_test.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/treeverse/lakefs/pkg/kv/postgres"
+	"github.com/treeverse/lakefs/pkg/testutil"
+)
+
+const (
+	migrateTestRepo   = "migrateTestRepo"
+	actionsSampleSize = 100
+)
+
+func TestMigrate(t *testing.T) {
```
itaiad200, line 27: Awesome test.

esti/migrate_test.go
```diff
@@ -121,20 +138,86 @@ func createAndUploadParts(t *testing.T, logger logging.Logger, resp *s3.CreateMu
...
 	fullContent := append(preContentBytes, partsConcat...)
 	require.Equal(t, fullContent, getResp.Body)
 }
+
+// Actions Tests
+
+func testPreMigrateActions(t *testing.T) {
+	// Create action data before migration
+	ctx, _, repo := setupTest(t)
+	state.Actions.Repo = repo
```
itaiad200, line 164: I know it's annoying but since `Repo` is being handled differently in the migration struct and path, I think you should test more than a single repo here.

esti/migrate_test.go
```diff
@@ -121,20 +138,86 @@ func createAndUploadParts(t *testing.T, logger logging.Logger, resp *s3.CreateMu
...
+			*ar.Hooks[i].EndTime = task.EndTime.UTC()
+		}
+		state.Actions.Runs = append(state.Actions.Runs, ar)
+	}
+}
+
+func testPostMigrateActions(t *testing.T) {
+	// Validate info using storage logs
+	ctx, _, _ := setupTest(t)
+	runs, err := client.ListRepositoryRunsWithResponse(ctx, state.Actions.Repo, &api.ListRepositoryRunsParams{})
```
itaiad200, line 199: Can you trigger more actions from the Post test and see that the results are saved correctly with the old runs? For example, the ordering might get tricky.

itaiad200 (changes requested):
Another brick in the wall 🧱

pkg/actions/db_service.go
```diff
@@ -402,3 +414,134 @@ func (s *DBService) NewRunID() string {
...
+
+	// Dump RunResults
+	rows, err := d.Query(ctx, "SELECT * FROM actions_runs ORDER BY run_id DESC")
+	if err != nil {
+		return err
+	}
+	defer rows.Close()
+	rowScanner := pgxscan.NewRowScanner(rows)
+	for rows.Next() {
+		newRunID := graveler.NewRunID()
```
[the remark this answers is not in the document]
N-o-Z: Done

esti/migrate_test.go
N-o-Z: I guess we can add it to this test - it will be quite a headache. Trying to understand what is the benefit of adding it to this test specifically? What do we want to check here?

pkg/actions/db_service.go
N-o-Z: I'm struggling with trying to reuse the code in the service - do you have any suggestions?

esti/migrate_test.go
N-o-Z: As discussed, added this to the actions migrate unit test

pkg/actions/db_service.go
```diff
@@ -402,3 +443,185 @@ func (s *DBService) NewRunID() string {
...
+	var g multierror.Group
+	for i := 0; i < runWorkers; i++ {
+		// Run worker reads runs from DB and writes respected entries to fd
+		g.Go(func() error {
+			return runWorker(runChan)
+		})
+		// tasks worker scans tasks per run ID and starts the task go routines which write the respected entries to fd
+		g.Go(func() error {
+			return tasksWorker(ctx, d, tasksChan)
+		})
```
itaiad200, line 590: I think it could be simplified if we had a single type that would handle a run and all its hooks. Right now it's hard to choose the right amount of workers of each type. You choose to create equal amount, but is the work really distributed evenly?

esti/migrate_test.go
itaiad200: Couldn't find it 🤷

itaiad200 (approved):
Approved with some comments, please either fix, discuss or open an issue before merging.
Thanks!

pkg/actions/db_service.go
N-o-Z: There's a question of the ratio of total runs vs total tasks, which I'm not sure there a singular truth to. I played with it a little with the worker counts and my conclusion was that the impact is not substantial with the given dataset (ratio of 10 tasks per run)

esti/migrate_test.go
N-o-Z: TestMigrate in pkg/actions/kv_migrate_test.go

pkg/kv/import.go
```diff
@@ -21,33 +27,25 @@ type Header struct {
...
-			return fmt.Errorf("empty file: %w", ErrInvalidFormat)
-		} else {
-			return fmt.Errorf("decoding header: %w", err)
-		}
-	}
-	// Decode does not return error onm missing data / incompatible format
-	if header == (Header{}) {
-		return fmt.Errorf("bad header format: %w", ErrInvalidFormat)
-	}
+type safeDecoder struct {
```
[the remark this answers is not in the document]
N-o-Z: json decoder is not thread safe - and trying to perform Decode concurrently is prone to errors

pkg/actions/db_service.go
```diff
@@ -8,17 +8,43 @@ import (
...
+//nolint:gochecknoinits
+func init() {
+	kvpg.RegisterMigrate(packageName, Migrate, []string{"actions_run_hooks", "actions_runs"})
+}
+
+const (
+	runWorkers       = 10
+	taskWorkers      = 10
+	migrateQueueSize = 100
+)
```
itai-david, line 39: Question - any specific reason to use these numbers?

pkg/actions/kv_migrate_test.go
```diff
@@ -25,30 +26,40 @@ const (
...
+	b.Run("Benchmark-2500", func(b *testing.B) {
+		benchmarkMigrate(2500, b)
+	})
+	b.Run("Benchmark-25000", func(b *testing.B) {
+		benchmarkMigrate(25000, b)
+	})
+}
+
+// benchmarkMigrate - use this test to benchmark the migration time of actions. Default dataset size is small as to not
+func benchmarkMigrate(runCount int, b *testing.B) {
```
itaiad200, line 42: This is very useful. Thank you.

pkg/actions/db_service.go
N-o-Z: No specific reason - mainly trial and error. I noticed that more than 10 workers do not really make any significant performance change and the decision for queue size was to make it at least an order of magnitude bigger than num of workers but not too big

pkg/actions/db_service.go
```diff
@@ -402,3 +428,185 @@ func (s *DBService) NewRunID() string {
...
+		}
+		runTime, err := time.Parse(graveler.RunIDTimeLayout, m.RunResult.RunID[:len(graveler.RunIDTimeLayout)])
+		if err != nil {
+			return err
+		}
+		newRunID := newRunIDFromTime(runTime)
+		oldRunID := m.RunResult.RunID
+		m.RunResult.RunID = newRunID
+		runChan <- m
+		t := &migrateTask{
```
itai-david, line 598: Is there any added value to have the tasks migration dependant on the corresponding run migration?
What I'm asking is, if we had 2 separate iterations, one over `action_runs` and the other over `actions_run_hooks` (regardless of `run_id`), having each trigger a job for each row, will it make any difference? (Other than the need to recalculate `newRunID` potentially many times)
The reason I'm asking is that if it is the same, each of these loops will look pretty similar to the `mulriparts.Migrate` loop and that makes me think of a generic `Migrate` that gets the specifics as paramaters (DB table name, type for marshaling etc.) and performs the "same" work for all
I guess this is probably not the right place to make such a change, but weas it considered?

N-o-Z: That's a great point. At the beginning the new run ID was created independently from the original ID so we had to save the tasks in the context of the run. We can definitely do it as you suggest and check if there's an improvement on performance. Regarding your second thought - I think this could be quite more tricky since the migration operation doesn't involve just transposing the data from the DB world to the KV. It also requires creating secondary indices and the conversion of some of the data (as in the example of actions and auth).

Issue #3399, opened 2022-05-20
kv hash key partitioning


nopcoder:
Enabling better underlaying performance can be achieved by adding another level, we can partition the data.
Postgres table can be partitioned by a key - to enable that, we can add an optional hash-key while setting and scanning data in our kv storage.

This issue was created to address:
- API changes to enable operations using hash-key
- Update the Postgres implementation to support partitioning

itaiad200:
Putting in writing the discussion @nopcoder And I had with @ozkatz:

The API should not be changed. It is written in a general way on purpose to allow many implementations to the kv store interface. We want to partition on Postgres since the current huge non-partitioned table will quickly become the bottleneck. 
The partitioning must take into account the main entity type with many changes and where performance matters => entries under a staging token. Since Keys are opaque to the kv store, it must decide on a parameter to partition by - which most likely be the prefix of the key. To ensure all entries under a staging token ends up in the same partition, we must write it at the very beginning of the key. e.g. instead of `refstore/repo/my-repo/branch/my-branch/stagingTokens/my-staging-token/path/to/my/file` do `stagingTokens/my-staging-token/repo/my-repo/branch/my-branch/path/to/my/file/`.

Given that we're currently not facing this problem (only multipart implementation is completed), and assuming that we agree on the above, this becomes more of a principal than an actual task to be done. I'll close this one if no objections are raised.

arielshaqed:
I pretty much disagree with [all of the above](https://github.com/treeverse/lakeFS/issues/3399#issuecomment-1134552176)

The API without separate hash and ordered keys is undoubtedly the simplest.  However not by much, and OTOH it will be very hard to scale.

First, note that _almost every key needs a prefix space_.  By not supporting it in the KV API, you're just moving the problem of handling prefixes over to the application.  Which is actually a harder problem than it seems if you don't want to limit the bytes allowed in your prefix.  So simplest might be to decide to go with `prefix`**:**`key`, but that means you don't get to use colons in your prefix.  

But more importantly, the current API does not match some useful real-life systems, and will not attain full performance.  Examples of high-value implementation targets:

* DynamoDB requires a hash and an ordered key.  Taking the first few bytes of the key as a hash will only work if `scan` becomes a complex operation.  Failing that, it is not that nice.
[...]

PR #3430, opened 2022-05-28
Update KV proposal with partition key

itaiad200:
Design for #3399

### DIFF
--- a/design/open/metadata_kv/index.md
+++ b/design/open/metadata_kv/index.md
@@ -285,6 +285,7 @@ Guidelines:
 - Partition keys are also a namespace for the key. i.e. the combination of (partitionKey,key) is unique in the KV database,
 but (key) is not guaranteed to be unique.
 - Although possible in some implementations, the interface will not support `dropPartition`. 
+  A later addition of `dropPartition` could speed up commits by iterating over keys in a goroutine after the commit is done and not have to worry about iterator invalidation.
 - New partition creation is implicit to keep the API free of additional `NewPartition` functionality.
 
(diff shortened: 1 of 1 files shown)


design/open/metadata_kv/index.md
```diff
@@ -265,6 +265,28 @@ This takes time proportional to the amount of changes to be committed and is unb
...
+The KV store implementation is in charge of managing the partitioned storage. The user the KV store
+is responsible for choosing the appropriate partitions. For example, the number of keys used for authentication
+& authorization is relatively small, so using the same 'auth' partition for all auth entities is appropriate.
+However, the number of entries (keys) under a single staging token is possibly the size of the entire repository,
+so creating a partition for each staging token is the right strategy.
+
+Guidelines:
+- Partition keys are also a namespace for the key. i.e. the combination of (partitionKey,key) is unique in the KV database,
+but (key) is not guaranteed to be unique.
+- Although possible in some implementations, the interface will not support `dropPartition`.
```
arielshaqed, line 287: Not required... but it will speed up commits on some KVs.  Obviously on PostgreSQL. Butalso on DynamoDB because now you can drop partitions by iterating over keys in a goroutine after the commit is done and not have to worry about iterator invalidation. If there are no single-key deletes, you can Scan 1000 keys at a time in your iterator and give correct results. This speedup due to allowable batching should work with most KVs. 

Iterators become faster this way.

itaiad200: Thx, added this clarification.

PR #3431, opened 2022-05-29
viper learn how to read ui code snippet from environment

nopcoder:
Have viper load UI snippet from environment value - string type.


nopcoder:
Not required, closing, as I prefer less code as possible.

PR #3434, opened 2022-05-29
Fix clean cookie on logout

guy-har:


nopcoder (approved):
Good catch! can move the code to a single 'create cookie' func to share the same properties for the next time.

PR #3444, opened 2022-05-30
KV partition implementation

### LINKED ISSUE #3399: kv hash key partitioning

itaiad200:
closes #3399

### DIFF
--- a/pkg/kv/kvtest/store.go
+++ b/pkg/kv/kvtest/store.go
@@ -447,6 +447,19 @@ func testStoreMissingArgument(t *testing.T, ms MakeStore) {
 			}
 		}
 	})
+
+	t.Run("Scan", func(t *testing.T) {
+		start := uniqueKey("test-missing-argument")
+		for _, pk := range [][]byte{nil, {}} {
+			it, err := store.Scan(ctx, pk, start)
+			if !errors.Is(err, kv.ErrMissingPartitionKey) {
+				t.Errorf("Scan using empty partition key - err=%v, expected %s", err, kv.ErrMissingPartitionKey)
+			}
+			if it != nil {
+				t.Errorf("Scan using empty partition key - expected nil iterator: %s", it)
+			}
+		}
+	})
 }
 
 func testScanPrefix(t *testing.T, ms MakeStore) {
--- a/pkg/kv/kvtest/store.go
+++ b/pkg/kv/kvtest/store.go
@@ -459,7 +459,7 @@ func testScanPrefix(t *testing.T, ms MakeStore) {
 		sampleData   []kv.Entry
 		samplePrefix []byte
 	)
-	const sampleItems = 2
+	const sampleItems = 100
 	const numberOfSets = 3
 
 	for i := 0; i < numberOfSets; i++ {
--- a/pkg/kv/kvtest/store.go
+++ b/pkg/kv/kvtest/store.go
@@ -18,16 +18,21 @@ type MakeStore func(t testing.TB, ctx context.Context) kv.Store
 
 var runTestID = nanoid.MustGenerate("abcdef1234567890", 8) //nolint:gomnd
 
+const (
+	testPartitionKey       = "test"
+	testUnusedPartitionKey = "unused"
+)
+
 func uniqueKey(k string) []byte {
 	return []byte(runTestID + "-" + k)
 }
 
-func setupSampleData(t *testing.T, ctx context.Context, store kv.Store, prefix string, items int) []kv.Entry {
+func setupSampleData(t *testing.T, ctx context.Context, store kv.Store, partition, prefix string, items int) []kv.Entry {
 	t.Helper()
 	entries := make([]kv.Entry, 0, items)
 	for i := 0; i < items; i++ {
 		entry := sampleEntry(prefix, i)
-		err := store.Set(ctx, entry.Key, entry.Value)
+		err := store.Set(ctx, []byte(partition), entry.Key, entry.Value)
 		if err != nil {
 			t.Fatalf("failed to setup data with '%s': %s", entry, err)
 		}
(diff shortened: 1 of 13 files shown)


pkg/kv/kvtest/store.go
```diff
@@ -146,14 +163,26 @@ func testStoreSetGet(t *testing.T, ms MakeStore) {
...
+		// get a missing key
+		partitionKeyNotExists := uniqueKey("partition-key-not-exists")
+		val3, err := store.Get(ctx, partitionKeyNotExists, testKey)
+		if !errors.Is(err, kv.ErrNotFound) {
+			t.Fatalf("get key='%s' err=%s, expected not found", partitionKeyNotExists, err)
+		}
+		if val3 != nil {
+			t.Fatalf("get key='%s' value='%s', expected nil", partitionKeyNotExists, val3)
+		}
+	})
```
itai-david, line 185: How about adding a key with a (valid and exist) partition key and trying to get it with another (also valid and exist) partition key?

pkg/kv/mem/store.go
```diff
@@ -135,30 +170,31 @@ func (e *EntriesIterator) Next() bool {
...
-		Key:   []byte(key),
-		Value: value,
+		PartitionKey: partitionKeyFromCombinedKey(combinedKey),
+		Key:          keyFromCombinedKey(combinedKey),
+		Value:        value,
 	}
 	// set start to the next item - nil as end indicator
 	if idx+1 < len(e.store.keys) {
-		e.start = []byte(e.store.keys[idx+1])
+		e.start = e.store.keys[idx+1]
```
itai-david, line 195: If the last key in a partition is reached, `Next()` continues to the next partition (right?). Is that the desired behavior? I believe the postgress implementation behaves differentley

pkg/kv/mem/store.go
```diff
@@ -92,40 +118,49 @@ func (s *Store) SetIf(_ context.Context, key, value []byte, valuePredicate kv.Pr
...
 		s.keys = append(s.keys[:idx], s.keys[idx+1:]...)
 	}
-	delete(s.m, string(key))
+	delete(s.m, combinedKey)
 	return nil
 }
 
-func (s *Store) Scan(_ context.Context, start []byte) (kv.EntriesIterator, error) {
+func (s *Store) Scan(_ context.Context, partitionKey, start []byte) (kv.EntriesIterator, error) {
+	combinedKey := combinedKey(partitionKey, start)
```
itai-david, line 159: I think that in case of `partitionKey = ""` we will scan all keys with no partition key, or all keys with partition key starting with lower case. All partition keys starting with upper case will be skipped as they preced `_`

pkg/kv/postgres/store.go
```diff
@@ -172,27 +212,37 @@ func (s *Store) SetIf(ctx context.Context, key, value []byte, valuePredicate kv.
...
-	if start == nil {
-		rows, err = s.Pool.Query(ctx, `SELECT key,value FROM `+s.Params.SanitizedTableName+` ORDER BY key`)
-	} else {
-		rows, err = s.Pool.Query(ctx, `SELECT key,value FROM `+s.Params.SanitizedTableName+` WHERE key >= $1 ORDER BY key`, start)
+
+	switch {
+	case start == nil && len(partitionKey) == 0:
+		rows, err = s.Pool.Query(ctx, `SELECT partition_key,key,value FROM `+s.Params.SanitizedTableName+` ORDER BY key`)
+	case start != nil && len(partitionKey) == 0:
+		rows, err = s.Pool.Query(ctx, `SELECT partition_key,key,value FROM `+s.Params.SanitizedTableName+` WHERE key >= $1 ORDER BY key`, start)
```
itai-david, line 239: Is scanning without specifying a partition key (scanning all partitions) a real use case?
The reason I'm asking is that DynamoDB does not support it, and in order to get it done we will have to explicitly iterate over all partitions.

itai-david (commented):
Thanks for this implementation - it aligns perfectly with my work on DynamoDB and should I think it can be adapted seamlessly.
A couple of questions/thoughts below.

pkg/kv/kvtest/store.go
itai-david: Yep, I think that covers it.

pkg/kv/mem/store.go
itaiad200: It's a test-only implementation (mem) and I'm assuming we won't have partitions with `_` in them. If we will have them, we can have a different separator (like `_*&^%$`).

pkg/kv/mem/store.go
```diff
@@ -39,23 +41,45 @@ func (d *Driver) Open(_ context.Context, _ string) (kv.Store, error) {
...
+
+func keyFromCombinedKey(combinedKey string) []byte {
+	//nolint:gomnd
+	return []byte(strings.SplitN(combinedKey, "_", 2)[1])
+}
+
+func partitionKeyFromCombinedKey(combinedKey string) []byte {
+	//nolint:gomnd
+	return []byte(strings.SplitN(combinedKey, "_", 2)[0])
+}
```
itai-david, line 56: I believe the usage of `'_'` as a separator can be easily broken, as nothing prevent a [artitionKey to contain `'_'`.
I know that `DynamoDB` doesn't prevent the use of `'_'` as part of ket, partition or value, so not sure what should be the correct behavior here.
Nevertheless, I don't think it is a blocker for this PR, certainley not at that point, but we might want to address it in a followup issue.

itai-david (approved):
Thanks for addressing all the comments.
I added another comment regarding the usage of `'_'` as a separator in `combinedKey`, but I don't think it is a blocker. I think it would be best to open an issue to address this later

Issue #3445, opened 2022-05-31
DynamoDB KV Support - Support Multiparts and Actions


itai-david:
Verify that Multiparts module (supports KV from KV1) and Actions (added KV support as part of KV2) runs smoothly with KV over DynamoDB.
DoD: Multiparts and Actions tests over KV pass with DynamoDB. Benchmarks for KV over DynamoDB for Multiparts and Actions. (Ideally) Benchmarks show no significant degradation from SQL and KV over Postgres

N-o-Z:
Initial benchmark of dynamoDB on multiparts - seems like we have a **serious** performance issue. branch: dynamodb-benchmark-tests
Example of Create test:
```
BenchmarkMultipartsTrackerCreate/db-16 	    4424	    291174 ns/op
BenchmarkMultipartsTrackerCreate/kv_mem
BenchmarkMultipartsTrackerCreate/kv_mem-16         	 1000000	      1875 ns/op
BenchmarkMultipartsTrackerCreate/kv_postgres
BenchmarkMultipartsTrackerCreate/kv_postgres-16    	    2112	    553008 ns/op
BenchmarkMultipartsTrackerCreate/kv_dynamodb
BenchmarkMultipartsTrackerCreate/kv_dynamodb-16    	       1	94941901099 ns/op
```

I'll continue with investigation - hoping this is a container/configuration issue? :pray:

N-o-Z:
After further investigation found a possible reason:
dynamoDB local is actually SQLite under the hood and explain the slowness https://stackoverflow.com/questions/28142895/dynamodb-local-terrible-performance

Decided to use dynamoDB prod for benchmark tests

N-o-Z:
Ran multiparts tests with dynamoDB and most of actions tests.
There's an issue with creating a big data set which takes a lot of time and the test times out after 10 minutes (TestRunResultsIterator). Need to understand if this is reasonable.
For future reference - when requesting kv store in unit test use testutil.GetStore
To run unit tests with dynamoDB - simply change the GetStore code to:
```
func GetStore(ctx context.Context, t testing.TB) kv.Store {
	t.Helper()
	return testutil.GetDynamoDBProd(ctx, t)
}
```

Issue #3443, opened 2022-05-30
DynamoDB KV Support - Migrate from postgres


itai-david:
Migrate from currently used postgres SQL DB (Not KV over postgress) to DynamoDB KV

One possibilityt is to use the exported data created as part of KV over postgress/mem support and import it

DoD: A migration process from Postges SQL to KV over DynamoDB, similar to what we have for `multiparts`. This process should include all currently supported modules (multiparts, actions, auth (?)).

itai-david:
Do we need to migrate from KV over Postgres to KV over DynamoDB?

N-o-Z:
Closing this issue as it was decided not to support direct migration to anything other than PGSQL.

Issue #3442, opened 2022-05-30
DynamoDB KV Support - Integrate with Production DynamoDB


itai-david:
DoD: A running instance/cluster of DynamoDB. KV DynamoDB implementation can connect to that DB and run properly.

Part of this task is to figure out various aspects of production dynamoDB usage:
* read/write capacities
* DyamoDB cluster structure
* AWS parameters (TBD?)

Issue #3441, opened 2022-05-30
DynamoDB KV Support - Integrate with Esti tests


itai-david:
DoD: A running Est system test with KV over DynamoDB
Currently we have an [Esti test running lakeFS wtigh S3 and KV](https://github.com/treeverse/lakeFS/blob/9d18e253e79fbed0c89fabcf08c8020addc498b5/.github/workflows/esti.yaml#L793). This task is about creating and running a copy of this test with KV over DynamoDB (Ideally, production DynamoDB)

PR #3446, opened 2022-05-31
KV Store implementation over DynamoDB

itai-david:
Initial implementation of KV over DynamoDB.
Currently covers all KV Store implementation and testing over local KV.
Still to be done:
* Esti tests and integrating it as part of our CI/CD (#3441)
* Working with a production DynamoDB (currently only local testing DB is supported) (#3442)
* Migrating from postgres (#3443)
* Support the new store with the using modules (#3445)


pkg/kv/dynamodb/store.go
```diff
@@ -0,0 +1,320 @@
...
+
+//nolint:gochecknoinits
+func init() {
+	kv.Register(DriverName, &Driver{})
+}
+
+func (d *Driver) Open(ctx context.Context, name string) (kv.Store, error) {
+	sess := session.Must(session.NewSessionWithOptions(session.Options{
+		SharedConfigState: session.SharedConfigEnable,
+	}))
```
itaiad200, line 58: Let's reuse the session we create for the blockstore. Unless you expect the credentials to be passed twice, they should be handled in a single place.

pkg/kv/dynamodb/store.go
```diff
@@ -0,0 +1,320 @@
...
+	ReadCapacityUnits  = 1000
+	WriteCapacityUnits = 1000
+)
+
+//nolint:gochecknoinits
+func init() {
+	kv.Register(DriverName, &Driver{})
+}
+
+func (d *Driver) Open(ctx context.Context, name string) (kv.Store, error) {
```
itaiad200, line 55: I think you're missing documentation for what `name` is for DynamoDB.

pkg/kv/dynamodb/store.go
```diff
@@ -0,0 +1,320 @@
...
+				KeyType:       aws.String("RANGE"),
+			},
+		},
+		ProvisionedThroughput: &dynamodb.ProvisionedThroughput{
+			ReadCapacityUnits:  aws.Int64(ReadCapacityUnits),
+			WriteCapacityUnits: aws.Int64(WriteCapacityUnits),
+		},
+	})
+	if err != nil {
+		if _, ok := err.(*dynamodb.ResourceInUseException); !ok {
```
itaiad200, line 119: What's special about this error?

pkg/kv/dynamodb/store.go
```diff
@@ -0,0 +1,320 @@
...
+	err          error
+	store        *Store
+	queryResult  *dynamodb.QueryOutput
+	currEntryIdx int
+	partKey      string
+	startKey     string
+}
+
+type DynKVItem struct {
+	PartKey   string
```
itaiad200, line 32: What is this?

pkg/kv/dynamodb/store.go
```diff
@@ -0,0 +1,320 @@
...
+	DriverName       = "dynamodb"
+	DefaultTableName = "kvstore"
+
+	PartKey   = "PartKey"
+	ItemKey   = "ItemKey"
+	ItemValue = "ItemValue"
+
+	// TBD: Which values to use?
+	ReadCapacityUnits  = 1000
+	WriteCapacityUnits = 1000
```
itaiad200, line 47: Why aren't these part of the `dsn`?

pkg/kv/dynamodb/store.go
itaiad200: Short for partition, got it.

pkg/kv/dynamodb/store.go
```diff
@@ -0,0 +1,320 @@
...
+func (s *Store) Get(ctx context.Context, key []byte) (*kv.ValueWithPredicate, error) {
+	if len(key) == 0 {
+		return nil, kv.ErrMissingKey
+	}
+
+	result, err := s.svc.GetItemWithContext(ctx, &dynamodb.GetItemInput{
+		TableName: aws.String(s.params.TableName),
+		Key:       s.bytesKeyToDynamoKey(key),
+	})
+
```
itaiad200, line 146: delete empty line

pkg/kv/dynamodb/store.go
```diff
@@ -0,0 +1,320 @@
...
+	item := DynKVItem{
+		PartKey:   s.getPartitionKey(key),
+		ItemKey:   string(key),
+		ItemValue: string(value),
+	}
+
+	marshaledItem, err := dynamodbattribute.MarshalMap(item)
+	if err != nil {
+		return fmt.Errorf("%s: %w", err, kv.ErrOperationFailed)
```
itaiad200, line 191: Is every error from any store a `kv.ErrOperationFailed`? What's the point of having this error then?

pkg/kv/dynamodb/store.go
```diff
@@ -0,0 +1,320 @@
...
+	if len(key) == 0 {
+		return kv.ErrMissingKey
+	}
+
+	_, err := s.svc.DeleteItemWithContext(ctx, &dynamodb.DeleteItemInput{
+		TableName: aws.String(s.params.TableName),
+		Key:       s.bytesKeyToDynamoKey(key),
+	})
+
+	if err != nil {
```
itaiad200, line 230: Does DynamoDB return ErrNotFound if the key is missing?

pkg/kv/dynamodb/store.go
```diff
@@ -0,0 +1,320 @@
...
+	if len(scanKey) > 0 {
+		keyConditionExpression += " AND " + ItemKey + " >= :fromkey"
+		expressionAttributeValues[":fromkey"] = &dynamodb.AttributeValue{
+			S: aws.String(string(scanKey)),
+		}
+	}
+	queryInput := &dynamodb.QueryInput{
+		TableName:                 aws.String(s.params.TableName),
+		KeyConditionExpression:    aws.String(keyConditionExpression),
+		ExpressionAttributeValues: expressionAttributeValues,
```
itaiad200, line 256: Why are both of them needed?

pkg/kv/dynamodb/store.go
```diff
@@ -0,0 +1,320 @@
...
+		ConsistentRead:            aws.Bool(true),
```
itaiad200, line 257: What's this?

pkg/kv/dynamodb/store.go
```diff
@@ -0,0 +1,320 @@
...
+		ScanIndexForward:          aws.Bool(true),
+		ExclusiveStartKey:         exclusiveStartKey,
+	}
+	queryOutput, err := s.svc.QueryWithContext(ctx, queryInput)
+	if err != nil {
+		return nil, fmt.Errorf("%s: %w (start=%v)", err, kv.ErrOperationFailed, string(scanKey))
```
itaiad200, line 263: That's a wrong format of go errors.

pkg/kv/dynamodb/store.go
```diff
@@ -0,0 +1,320 @@
...
+
+func (s *Store) Close() {}
+
+func (e *EntriesIterator) Next() bool {
+	if e.err != nil {
+		return false
+	}
+
+	if e.currEntryIdx == int(*e.queryResult.Count) {
+		if e.queryResult.LastEvaluatedKey == nil {
```
itaiad200, line 284: how can this happen?

itaiad200 (changes requested):
Thanks! Some questions and minor improvements.

pkg/kv/dynamodb/store.go
itai-david: Looks like a wrong C/P. Renamed to dns and added documentation

pkg/kv/dynamodb/store.go
itai-david: This is the error DynamoDB returns if the table already exists. As such, it can be ignored and the creation con be considered successful

pkg/kv/dynamodb/store.go
itai-david: I'm not sure I fully understand your question, so I will try to explain from my POV:
Errors that can be well indetified from **our** `KV` aspect, such as missing key, missing value or failed predicate, has a specialized well defined. Other errors, mainly these coming from the underlying engine/DB - we wrap the error with the general `ErrOperationFailed`.
In the specific case you commented above, this is a marshaling error that may be a result of changing the DynamoDB table structure, without  modifying our code accordingly, or the other way around (maybe due to versions mismatch or similar). This is just one example and there are probably more, but I think the common base for all these is to wrap various failures, mainley the ones that are underlying layer based, with "our own" `kv` error type.
Please let me know if this answers your question

pkg/kv/dynamodb/store.go
itai-david: Nope. `DeleteItem` (and `DeleteItemWithContext` for that matter) is idempotent and as such does not fails if the key (and, thus, the item) does not exist. That aligns with our `KV` interface

pkg/kv/dynamodb/store.go
itai-david: `KeyConditionExpression` is the scan term, `PartitionaLey = :partitionkey AND ItemKey >= :fromkey` where `:partitionkey` and `:fromkey` are attributes values, specified in `ExpressionAttributeValues`, in json format following dynamodb definitions.
I'm not sure if it is possible to skip the usage of `ExpressionAttributeValues` and specify the values in the `KeyConditionExpression` itself but I spend sometime on it and couldn't get it to work (at least not for `awscli`).
This format also aligns with all the examples from `aws`

pkg/kv/dynamodb/store.go
itai-david: This is to indicate strongly consistent reads - all write operations prior to this operation will be reflected, as we required for scan.
Otherwise, `DynamoDB` only guarantees `eventually consistent` which we cannot use

pkg/kv/dynamodb/store.go
itai-david: `Query` (`QueryWithContext`) will return `Count` items. If there are more items that satisfy the `KeyConditionExpression` the `LastEvaluatedKey` is used as a `ContinuationToken` which can be used for paging.
With each "page", I'm using the `currEntryIdx` to loop over the objects in that page. When `currEntryIdx` equals `Count` the end of the page is reached. If this is the last page the above condition will apply, signifying the last item for that partition is reached

pkg/kv/dynamodb/store.go
```diff
@@ -0,0 +1,320 @@
...
+	}
+	queryInput := &dynamodb.QueryInput{
+		TableName:                 aws.String(s.params.TableName),
+		KeyConditionExpression:    aws.String(keyConditionExpression),
+		ExpressionAttributeValues: expressionAttributeValues,
+		ConsistentRead:            aws.Bool(true),
+		ScanIndexForward:          aws.Bool(true),
+		ExclusiveStartKey:         exclusiveStartKey,
+	}
+	queryOutput, err := s.svc.QueryWithContext(ctx, queryInput)
```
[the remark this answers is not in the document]
itai-david: The number of items is indeterminate and is dictated by the items size. `DynamoDB` returns up to ~1MB of data, regardless of the number of items.
In addition the number of items can be limited, using a specified value (which we currently not using).
AFAICT, our tests **does not cover** pagination, but I tested it manually

pkg/kv/dynamodb/store.go
itai-david: Good enough reason to rename it. Done

pkg/kv/dynamodb/store.go
itai-david: Relevant credentials from env are read and overriden by aws conf files (if any).
In addition, added another layer to allow specific test params (fake access and secret keys) without affecting the env

pkg/kv/dynamodb/store.go
itai-david: They are now :-)

pkg/kv/dynamodb/store.go
itai-david: Done

pkg/kv/dynamodb/store.go
itai-david: Done. Moved the `%w` to the end of the fromatting string

pkg/kv/dynamodb/store.go
```diff
@@ -0,0 +1,322 @@
...
+		return fmt.Errorf("%s: %w", err, kv.ErrOperationFailed)
+	}
+	return nil
+}
+
+func (s *Store) Scan(ctx context.Context, start []byte) (kv.EntriesIterator, error) {
+	return s.scanInternal(ctx, start, nil)
+}
+
+func (s *Store) scanInternal(ctx context.Context, scanKey []byte, exclusiveStartKey map[string]*dynamodb.AttributeValue) (*EntriesIterator, error) {
```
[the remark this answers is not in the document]
itai-david: Is that crucial for you? I truly find the concatenation, using `+`, more readable

pkg/kv/dynamodb/store.go
itai-david: I see what you mean, but I'm not sure this is the right scope for that.
I think we will need a better definition of errors we return, and how we handle errors from the underlying infrastructure(s).
Note that the more specific errors you mentioned, such as `ErrMissingArgument` are identifiable by our code, while (at least most of) the ones that are simply wrapped by the very generic `ErrOperationFailed`, are a result of error returned from beneath, and it will require to add our own error handling code for these. Moreover, I'm not sure how actionable it is, as we (at least I) haven't fully gotten into all the errors that can happen.
One last point - the same behavior (more or less) is present in the other stores implementation, and if we revise this one, we will need to align all of them.
How about we handle it in a separate issue, for better error handling and reporting?

pkg/kv/dynamodb/main_test.go
```diff
@@ -1,23 +1,49 @@
...
+		TableName:          testUniqueTableName(),
+		ReadCapacityUnits:  100,
+		WriteCapacityUnits: 100,
+		Endpoint:           databaseURI,
+		AwsRegion:          "us-east-1",
+		AwsAccessKeyID:     "fakeMyKeyId",
+		AwsSecretAccessKey: "fakeSecretAccessKey",
+	}
+
+	dsnBytes, err := json.Marshal(testParams)
```
itaiad200, line 40: I'm pretty sure that's not what @nopcoder & @N-o-Z had in mind when they thought of the `dsn`. I believe that it would be super hard for the user to pass json configuration in a single string.

I understand why you did that though, there currently isn't an option to pass configuration to the KV implementation in any other way.

pkg/kv/dynamodb/store.go
```diff
@@ -77,14 +100,17 @@ func (d *Driver) Open(ctx context.Context, dsn string) (kv.Store, error) {
...
-	// At some point the partition key will probably depend on the key.
-	// Meanwhile, it is simply the table name (i.e. everything is under the same partition)
-	return s.params.TableName
+func parseDsn(dsn string) (*Params, error) {
+	params := &Params{
+		TableName:          DefaultTableName,
+		ReadCapacityUnits:  DefaultReadCapacityUnits,
+		WriteCapacityUnits: DefaultWriteCapacityUnits,
+	}
+	err := json.Unmarshal([]byte(dsn), params)
```
itaiad200, line 109: Does it keep the values if the parsing succeeded with those keys missing? I learned something new..

pkg/kv/dynamodb/store.go
```diff
@@ -55,18 +67,29 @@ func init() {
 // Open - opens and returns a KV store over DynamoDB. This function creates the DB session
 // and sets up the KV table. dsn is a string with the DynamoDB endpoint
 func (d *Driver) Open(ctx context.Context, dsn string) (kv.Store, error) {
+	// TODO: Get table name from env
+	params, err := parseDsn(dsn)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %s", kv.ErrDriverConfiguration, err)
```
itaiad200, line 73: When the caller gets this error back, what do you expect the error type to be? More precisely, what would you like to happen (I think you can only have one): `errors.Is(err, kv.ErrDriverConfiguration) == true` or `errors.Is(err, json.ParsingFailure) == true` as an example.

pkg/kv/dynamodb/store.go
```diff
@@ -29,22 +31,32 @@ type EntriesIterator struct {
...
+
+type Params struct {
+	TableName          string
+	ReadCapacityUnits  int64
+	WriteCapacityUnits int64
+
+	Endpoint           string
+	AwsRegion          string
+	AwsAccessKeyID     string
+	AwsSecretAccessKey string
```
itaiad200, line 47: Let's document these as they are tricky.

pkg/testutil/dynamodb.go
```diff
@@ -53,19 +54,8 @@ func RunLocalDynamoDBInstance(port string) (string, func(), error) {
...
-	if os.Getenv("AWS_ACCESS_KEY_ID") == "" {
-		os.Setenv("AWS_ACCESS_KEY_ID", "fakeMyKeyId")
-	}
-	if os.Getenv("AWS_SECRET_ACCESS_KEY") == "" {
-		os.Setenv("AWS_SECRET_ACCESS_KEY", "fakeSecretAccessKey")
-	}
-
 	// Disgusting, but I have to wait for the dynamodb container to be ready
-	time.Sleep(2 * time.Second)
+	time.Sleep(WaitForContainerSec * time.Second)
```
itaiad200, line 58: Let's probe the container healthcheck instead. Not only this is "Disgusting", it is also flaky and does not guarantee that the container is up and healthy.

pkg/kv/dynamodb/store.go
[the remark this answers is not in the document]
itaiad200: Our tests can cover pagination by design since it's a database internal implementation feature. If they do not cover it, too much code is kept uncovered. We should either add\change the store tests to fit it, or write something designated to DynamoDB.

itaiad200 (changes requested):

pkg/kv/dynamodb/store.go
```diff
@@ -0,0 +1,361 @@
...
+}
+
+const (
+	DriverName       = "dynamodb"
+	DefaultTableName = "kvstore"
```
N-o-Z, line 52: Lets keep the convention as in the postgres implementation and set it to "kv".
Also - does it make sense to move this definition to the store interface and have all implementations use it instead?

pkg/kv/dynamodb/store.go
itai-david: :star_struck:

pkg/kv/dynamodb/store.go
```diff
@@ -0,0 +1,361 @@
...
+		ScanIndexForward:          aws.Bool(true),
+		ExclusiveStartKey:         exclusiveStartKey,
+	}
+	queryOutput, err := s.svc.QueryWithContext(ctx, queryInput)
+	if err != nil {
+		return nil, fmt.Errorf("%s (start=%v): %w ", err, string(scanKey), kv.ErrOperationFailed)
+	}
+
+	return &EntriesIterator{
+		store:        s,
```
N-o-Z, line 308: Save the context - and use it in the tmpIter as well?

N-o-Z (commented):
@itai-david Looks awesome, can't wait to play with this implementation. I have just a couple of comments

pkg/kv/dynamodb/store.go
itai-david: No can do. DynamoDB requires 3 chars minimum for table names :man_shrugging: 
Maybe we should align the other way around?

N-o-Z: Sounds reasonable

pkg/kv/dynamodb/main_test.go
itai-david: The reason I went with that solution is, as you said, there is no `config` function, or configuration object to the store. It is pretty much the same for `postgres` where the dsn is actually the connection string (@N-o-Z please correctl me if I'm wrong), except for the `json` nature of it.
Anyhow, we can provide the parameters via env vars, via some `config` function (additional func in `Store` interface?) or leave it as it is, having the dsn and its options better documented in ortder to ease any future use of it.
We can also address it later, in a separate issue, as I believe there is going to be some discussion over it.
WDYT?

pkg/kv/dynamodb/store.go
itai-david: Done

pkg/kv/dynamodb/store.go
```diff
@@ -0,0 +1,361 @@
...
+		PartitionKey: {
+			S: aws.String(string(partitionKey)),
+		},
+		ItemKey: {
+			S: aws.String(string(key)),
+		},
+	}
+}
+
+func (s *Store) Get(ctx context.Context, partitionKey, key []byte) (*kv.ValueWithPredicate, error) {
```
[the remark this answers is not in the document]
itai-david: I'm not sure I follow. Can you please elaborate?

pkg/kv/dynamodb/store.go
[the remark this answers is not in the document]
itaiad200: Not crucial.

pkg/testutil/dynamodb.go
itai-david: Done

pkg/kv/dynamodb/store.go
itai-david: Excellent point. Done

pkg/kv/dynamodb/store.go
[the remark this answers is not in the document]
itai-david: I tried to resize our scan test to breach the pagination of ~1MB and it seems to take ages.
In addition, I don't thing this approach leads to a scalable test, as today it is 1MB but who knows what bears the future
I thought of leveraging the option to specify a limit on the number of entries per each page and so create a real, sustainable pagination.
It can be added as another param to the dsn. WDYT?

pkg/kv/dynamodb/store.go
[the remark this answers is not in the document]
N-o-Z: At the time we, upon error we logged the failed key value in the StoreMessage implementation. This was changed and the responsibility moved to the underlying store implementation. This means that each driver is responsible to log on which key the operation failed. You can see the example in the postgres implementation

pkg/kv/dynamodb/store.go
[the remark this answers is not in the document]
itai-david: Done.
Added paging size parameter and used it to force paging in tests.
Also found and fixed a bug in the delete while scan with paging, where deleting the entire last page, while scanning the former page, will lead to a crash, so +1 for that comment!

itaiad200 (approved):
Approved - thanks! Please open an issue for the configuration.

pkg/testutil/dynamodb.go
```diff
@@ -51,22 +49,21 @@ func RunLocalDynamoDBInstance(port string) (string, func(), error) {
...
-	// dummy aws env, in case not configured
-	if os.Getenv("AWS_DEFAULT_REGION") == "" {
-		os.Setenv("AWS_DEFAULT_REGION", "us-west-1")
-	}
-	if os.Getenv("AWS_ACCESS_KEY_ID") == "" {
-		os.Setenv("AWS_ACCESS_KEY_ID", "fakeMyKeyId")
-	}
-	if os.Getenv("AWS_SECRET_ACCESS_KEY") == "" {
-		os.Setenv("AWS_SECRET_ACCESS_KEY", "fakeSecretAccessKey")
+	err = dockerPool.Retry(func() error {
```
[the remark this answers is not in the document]
itai-david: It has a 1 minute default, so I think it is sufficient.
The container start is very quick - I had no prblems when I used the 2 sec sleep, so 1 minute should be enough

pkg/kv/dynamodb/main_test.go
itai-david: Opened #3474

N-o-Z:
@itai-david Are we sure we want to include this in the changelog? Seems ineffectual to users for now

Issue #3127, opened 2022-03-27
Add garbage collection report at end of run


arielshaqed:
When a GC job has finished, save the following data under the repository's storage namespace:
| Description                                                                                                                                                | Format  | Location                                                       |
|------------------------------------------------------------------------------------------------------------------------------------------------------------|---------|----------------------------------------------------------------|
| Basic information about the run, including:<br>- # of deleted objects<br>- The GC rules used for this jobs.                                                | JSON    | _lakefs/logs/gc/summary/dt=YYYY-MM-DD-HH-mm-SS/summary.json    |
| A set of parquet files describing all the ~deleted objects~ expired_addresses, including the columns:<br>- key (logical key in lakeFS)<br>- physical_address                   | Parquet | _lakefs/logs/gc/expired_addresses/dt=YYYY-MM-DD-HH-mm-SS/        |
| A list of all the commits participating in this GC job, including the columns:<br>- commit ID<br>- commit timestamp.<br>- whether it was expired or active | Parquet | _lakefs/logs/gc/summary/dt=YYYY-MM-DD-HH-mm-SS/commits.parquet |
| A list of all deleted objects                   | Parquet | _lakefs/logs/gc/deleted_objects/dt=YYYY-MM-DD-HH-mm-SS/deleted.parquet       |

Taken from the [product requirement for GC](#439)

Update:
When we implemented the report, we found the difference between expired_addresses, which are files that should be deleted, and the files that were actually deleted. We decided to report both, in case there will be a difference between them.

talSofer:
Before we start working on this we should review [this](format https://github.com/treeverse/demo-lakefs/issues/439#issuecomment-1073244157) comment from Oz and discuss with @johnnyaug on the final path

johnnyaug:
@talSofer @arielshaqed updated to new product requirement.

lynnro314:
Closed with #3450. Added a new issue https://github.com/treeverse/lakeFS/issues/3527 to add the commit timestamp.

PR #3450, opened 2022-06-02
Add gc report

lynnro314:
closes part of #3127 
Added reports at the end of gc run, reporting the deleted objects, relevant commits and the gc rules. 
Missing commit timestamp in the commits summary.

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
@@ -6,6 +6,8 @@ import io.treeverse.clients.LakeFSContext.{
   LAKEFS_CONF_API_SECRET_KEY_KEY,
   LAKEFS_CONF_API_URL_KEY
 }
+
+import org.apache.hadoop.fs._
 import org.apache.hadoop.conf.Configuration
 import org.apache.spark.broadcast.Broadcast
 import org.apache.spark.rdd.RDD
@@ -19,6 +21,9 @@ import com.amazonaws.services.s3.model
 import java.net.URI
 import collection.JavaConverters._
 import java.time.format.DateTimeFormatter
+import org.json4s.native.JsonMethods._
+import org.json4s._
+import org.json4s.JsonDSL._
 
 /** Interface to build an S3 client.  The object
  *  io.treeverse.clients.conditional.S3ClientBuilder -- conditionally
@@ -305,8 +310,6 @@ object GarbageCollector {
       System.exit(1)
     }
 
-    println("! this is with test code")
-
     val spark = SparkSession.builder().appName("GarbageCollector").getOrCreate()
 
     val repo = args(0)
@@ -326,6 +329,8 @@ object GarbageCollector {
     val accessKey = hc.get(LAKEFS_CONF_API_ACCESS_KEY_KEY)
     val secretKey = hc.get(LAKEFS_CONF_API_SECRET_KEY_KEY)
     val apiClient = new ApiClient(apiURL, accessKey, secretKey)
+
+    val gcRules = apiClient.getGarbageCollectionRules(repo)
 
     val res = apiClient.prepareGarbageCollectionCommits(repo, previousRunID)
     val runID = res.getRunId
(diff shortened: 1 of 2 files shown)


talSofer:
@lynnro314 can you please add an example report to the pr description? It will make it easier to review

Jonathan-Rosenberg (approved):
Cool 😺 !

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -446,4 +463,30 @@ object GarbageCollector {
 
     bulkRemove(df, MaxBulkSize, bucket, region, awsRetries, snPrefix, hcValues).toDF("addresses")
   }
+
+  private def writeParquetReport(
+      dstRoot: String,
+      df: DataFrame,
+      time: String,
+      suffix: String = ""
```
arielshaqed, line 471: Nit: The location of `df` makes it a bit hard to read calls to the function.  I think I might prefer
```suggestion
  private def writeParquetReport(
      df: DataFrame,
      dstRoot: String,
      time: String,
      suffix: String = ""
```

arielshaqed: (Also the same in `writeJsonSummary`, of course...).

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -446,4 +463,30 @@ object GarbageCollector {
...
+  }
+
+  private def writeJsonSummary(
+      dstRoot: String,
+      numDeletedObjects: Long,
+      gcRules: String,
+      hcValues: Broadcast[ConfMap],
+      time: String
+  ) = {
+    val dstPath = new Path(dstRoot + s"/dt=${time}/summary.json")
```
arielshaqed, line 484:
```suggestion
    val dstPath = new Path(s"${dstRoot}/dt=${time}/summary.json")
```

Also, I really like interpolating values into strings like this -- can we do it in the paths above, please, for consistency?

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -446,4 +463,30 @@ object GarbageCollector {
...
+    val dstFS = dstPath.getFileSystem(configurationFromValues(hcValues))
+    val jsonSummary = JObject("gc_rules" -> gcRules, "num_deleted_objects" -> numDeletedObjects)
```
arielshaqed, line 486: This is a bit strange: we just got `gcRules`, a value that is explicitly JSONable.  We transformed it into a string and then include this stringified value inside our real JSON.  The problem is that JSON-in-JSON is pretty awful to read (every quote is backslashed, newlines added for formatting before less readable, etc.).  Can we instead just stuff gcRules in some JSONified form here?

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -446,4 +463,30 @@ object GarbageCollector {
...
+
+    val stream = dstFS.create(dstPath)
+    stream.writeChars(compact(render(jsonSummary)))
```
arielshaqed, line 489: This leaks `stream` if it throws an exception.  It should catch exceptions and abort the stream on failure.

arielshaqed (approved):
Thanks!

lynnro314:
> @lynnro314 can you please add an example report to the pr description? It will make it easier to review

Sure.
1. Parquet files describing all the deleted objects:
image

2. A list of all the commits participating in this GC job:
image

3. Basic information about the run, including: # of deleted objects, the GC rules used for this jobs:
image
The rules:
image

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -353,13 +361,21 @@ object GarbageCollector {
...
+    writeParquetReport(expiredAddresses, reportExpiredDst, time)
+    writeJsonSummary(reportLogsDst, removed.count(), gcRules, hcValues, time)
+
     removed
       .withColumn("run_id", lit(runID))
       .write
       .partitionBy("run_id")
       .mode(SaveMode.Overwrite)
-      .parquet(gcAddressesLocation + ".deleted")
+      .parquet(s"${storageNamespace}/_lakefs/logs/gc/deleted_objects/${time}/deleted.parquet")
```
talSofer, line 378: Can you reuse the writeParquetReport to writing the deleted objects report? you can adjust it to receive optional partitionedBy param etc

talSofer (approved):
LGTM, thanks! 
Added one comment that you may want to consider addressing

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
lynnro314: I thought about that, but didn't find how I can do just `df.write.parquet(dstPath)` in that case..

PR #3452, opened 2022-06-02
Feature/support OIDC auth

johnnyaug:
## Support authentication with OIDC - step 1

Allows the user to configure an external OIDC provider. Once done, the login page will include a link to sign-in using the provider. When a user logs in with the provider, a corresponding lakeFS user is created.

Permissions (authorization) for the created user are still managed internally in lakeFS.


pkg/ddl/000038_oidc_user.down.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+DROP INDEX email_unique_idx;
+ALTER TABLE auth_users
+DROP COLUMN IF EXISTS email,
+    DROP COLUMN IF EXISTS encrypted_password;
```
arielshaqed, line 5: Doesn't this kill all users on `migrate down`?

pkg/ddl/000038_oidc_user.down.sql
```diff
@@ -0,0 +1,6 @@
+BEGIN;
+DROP INDEX email_unique_idx;
+ALTER TABLE auth_users
```
arielshaqed, line 3: Doesn't seem like this drops what it should.  In general, I am not sure we should drop *anything*; instead just make "migrate up" idempotent, then you can migrate down losing **nothing** and migrate back up when you're ready.

webui/src/pages/auth/login.jsx
```diff
@@ -51,10 +51,16 @@ const LoginForm = () => {
...
-                            <Button variant="link" className="reset-pwd" onClick={()=> {router.push("/auth/resetpassword")}}>Reset password</Button>
-                            : ""
-                        }
+                        <div className={"mt-2 mb-1"}>
+                            { showResetPwd ?
+                                <Button variant="link" className={"text-secondary mt-2"}  onClick={()=> {router.push("/auth/resetpassword")}}>Reset password</Button>
+                                : ""
+                            }
+                            { oidcEnabled ?
+                                <Button variant="link" className="text-secondary mt-2" onClick={()=> {window.location = "/oidc/login"}}>Sign in with OIDC provider</Button>
```
arielshaqed, line 60: Consider `s/OIDC Provider/SSO Provider`.  Users don't need to know they're using OIDC, they're just using their GMail.

arielshaqed (changes requested):
Thanks!  Looks really nice, not requesting anything major.

pkg/ddl/000038_oidc_user.down.sql
johnnyaug: oops! made a mess here

pkg/api/oidc_login_handler.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+		scheme := "http"
+		if r.TLS != nil {
+			scheme = "https"
+		}
+		u := url.URL{
+			Scheme: scheme,
+			Host:   r.Host,
+			Path:   BaseURL + "/oidc/callback",
+		}
```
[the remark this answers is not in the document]
johnnyaug: You mean dropping the host from the redirect URL? 
OAuth requires providing an absolute URL for this parameter.

docs/reference/oidc.md
```diff
@@ -0,0 +1,47 @@
+---
```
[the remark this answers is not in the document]
johnnyaug: Our latest docs always represent the features of our main branch - so I think the answer is yes.

pkg/api/auth_middleware.go
```diff
@@ -105,6 +123,58 @@ func checkSecurityRequirements(r *http.Request, securityRequirements openapi3.Se
 	return nil, nil
 }
 
+// userFromOIDC returns a user from an existing OIDC session.
+// If the user doesn't exist on the lakeFS side, it is created.
+// This function does not make any calls to an external provider.
```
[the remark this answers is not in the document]
johnnyaug: :blush:

cmd/lakefs/cmd/run.go
```diff
@@ -244,6 +246,25 @@ var runCmd = &cobra.Command{
 		if err != nil {
 			logger.WithError(err).Fatal("Emailer has not been properly configured, check the values in sender field")
 		}
+		oidcConfig := cfg.GetAuthOIDCConfiguration()
+		var oauthConfig *oauth2.Config
+		var oidcProvider *oidc.Provider
+		if oidcConfig != nil && oidcConfig.URL != "" {
```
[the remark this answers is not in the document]
johnnyaug: Added an "enabled" field.

pkg/api/oidc_login_handler.go
```diff
@@ -0,0 +1,61 @@
...
+	IdTokenClaimsSessionKey = "id_token_claims"
+	StateSessionKey         = "state"
+)
+
+// NewOIDCLoginPageHandler returns a handler to redirect the user the OIDC provider's login page.
+func NewOIDCLoginPageHandler(sessionStore sessions.Store, oauthConfig *oauth2.Config) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		state, err := generateRandomState()
+		if err != nil {
+			writeError(w, http.StatusInternalServerError, err.Error())
```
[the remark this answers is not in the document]
johnnyaug: Done

pkg/api/oidc_login_handler.go
```diff
@@ -0,0 +1,61 @@
...
+			return
+		}
+
+		session, err := sessionStore.Get(r, OIDCAuthSessionName)
+		if err != nil {
+			writeError(w, http.StatusInternalServerError, err.Error())
```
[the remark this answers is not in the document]
johnnyaug: Done

webui/src/pages/auth/login.jsx
johnnyaug: Done

pkg/api/oidc_login_handler.go
```diff
@@ -0,0 +1,67 @@
...
+	"golang.org/x/oauth2"
+)
+
+const (
+	OIDCAuthSessionName = "auth_session"
+
+	IDTokenClaimsSessionKey = "id_token_claims"
+	StateSessionKey         = "state"
+
+	stateByteSize = 32
```
arielshaqed, line 20: Is there a recommendation for this?  The best I could find was [this on login.gov](https://developers.login.gov/oidc/):

> **state**
> A unique value at least 22 characters in length used for maintaining state between the request and the callback. This value will be returned to the client on a successful authorization.

Admittedly this is quite a weak recommendation.

22 characters from a set of (say) 64 different characters is 6*22=132 bites, or 32.5 bytes.  But I would be happier to specify this in terms of characters, and use 22 or some recommended value.  nanoid is your friend for doing this sort of thing.

pkg/api/oidc_login_handler.go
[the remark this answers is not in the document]
arielshaqed: :-(

If we had the request URL `u` we could still `u.Resolve(BaseURL + "/oidc/callback")`, but I can understand that this might not be readily available.  (If you *can* get it onto `r`, then of course that would be better!)

pkg/ddl/000038_oidc_user.up.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+
+ALTER TABLE auth_users
+    ADD COLUMN IF NOT EXISTS external_id VARCHAR(255) UNIQUE NULL;
```
arielshaqed, line 4: I don't think we use the `NULL` (non-)constraint in our DDL.  I had to look up what it means, so I slightly prefer that you _remove_ it from here.

arielshaqed (approved):
Thanks!

cmd/lakefs/cmd/run.go
```diff
@@ -244,6 +246,25 @@ var runCmd = &cobra.Command{
...
+		var oidcProvider *oidc.Provider
+		if oidcConfig != nil && oidcConfig.Enabled {
+			oidcProvider, err = oidc.NewProvider(
+				cmd.Context(),
+				oidcConfig.URL,
+			)
+			if err != nil {
+				logger.WithError(err).Fatal("Failed to initialize OIDC provider")
+			}
+			cfg.GetBlockstoreDefaultNamespacePrefix()
```
ortz, line 260: is this relevant here?

cmd/lakefs/cmd/run.go
```diff
@@ -244,6 +246,25 @@ var runCmd = &cobra.Command{
...
+			oauthConfig = &oauth2.Config{
```
ortz, line 261: shouldn't we add the RedirectURL also?

pkg/api/controller.go
```diff
@@ -170,10 +175,47 @@ func (c *Controller) Logout(w http.ResponseWriter, _ *http.Request) {
...
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, err)
+		return
+	}
+	session.Values = nil
+	err = session.Save(r, w)
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, err)
+		return
+	}
```
ortz, line 188: what happens if OIDC is not configured?
shouldn't this return an error?

cmd/lakefs/cmd/run.go
```diff
@@ -14,6 +14,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/coreos/go-oidc"
```
nopcoder, line 17: This is v2 - I think you want to use `github.com/coreos/go-oidc/v3/oidc`

go.mod
```diff
@@ -81,6 +81,14 @@ require (
 	golang.org/x/time v0.0.0-20220224211638-0e9765cccd65
 )
 
+require (
+	github.com/coreos/go-oidc v2.2.1+incompatible // indirect
```
nopcoder, line 85: looks like you need to run `go mod tidy`

pkg/api/oidc_login_handler.go
johnnyaug: Done

pkg/ddl/000038_oidc_user.up.sql
johnnyaug: done

cmd/lakefs/cmd/run.go
johnnyaug: Since the redirect URL also contains the host, I construct it ad-hoc in every login.

pkg/api/controller.go
johnnyaug: This is the global logout, not specific to OIDC.
We could perform this part only if OIDC is configured - but maybe it's better to clear the session even if OIDC is disabled.

cmd/lakefs/cmd/run.go
johnnyaug: oops, removing

go.mod
johnnyaug: done

cmd/lakefs/cmd/run.go
johnnyaug: done

pkg/api/controller.go
```diff
@@ -170,11 +175,46 @@ func (c *Controller) Logout(w http.ResponseWriter, _ *http.Request) {
...
+	http.Redirect(w, r, "/", http.StatusTemporaryRedirect)
+}
+
 func (c *Controller) Login(w http.ResponseWriter, r *http.Request, body LoginJSONRequestBody) {
+	err := doOIDCLogout(w, r, c.sessionStore)
+	if err != nil {
+		c.Logger.WithError(err).Error("failed to perform OIDC logout")
+	}
```
nopcoder, line 217: Think we need to change one of two places. We can ignore the error here as loging a session in a system that doesn't have a session will fail. Or ignore it as we already just log it.

pkg/auth/oidc/authenticator.go
```diff
@@ -0,0 +1,44 @@
...
+// The ID token is verified to be valid, and its Claims are then returned.
+func (a *Authenticator) GetIDTokenClaims(ctx context.Context, code string) (Claims, error) {
+	token, err := a.oauthConfig.Exchange(ctx, code)
+	if err != nil {
+		return nil, err
+	}
+	rawIDToken, ok := token.Extra("id_token").(string)
+	if !ok {
+		return nil, err
+	}
```
nopcoder, line 31: There is no err here

pkg/ddl/000038_oidc_user.up.sql
```diff
@@ -0,0 +1,5 @@
+BEGIN;
+
+ALTER TABLE auth_users
+    ADD COLUMN IF NOT EXISTS external_id VARCHAR(255) UNIQUE;
```
nopcoder, line 4: 1. Any reason to use varchar and not text?
2. Missing drop column in the ...down.sql.

nopcoder (approved):
Some minor issues. Two suggestions. Very glad we have all the rest. Very cool!!

cmd/lakefs/cmd/run.go
```diff
@@ -254,6 +256,24 @@ var runCmd = &cobra.Command{
 		if err != nil {
 			logger.WithError(err).Fatal("Emailer has not been properly configured, check the values in sender field")
 		}
+		oidcConfig := cfg.GetAuthOIDCConfiguration()
+		var oauthConfig *oauth2.Config
+		var oidcProvider *oidc.Provider
+		if oidcConfig != nil && oidcConfig.Enabled {
```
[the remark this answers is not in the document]
johnnyaug: Done

pkg/api/auth_middleware.go
```diff
@@ -89,6 +106,8 @@ func checkSecurityRequirements(r *http.Request, securityRequirements openapi3.Se
 					continue
 				}
 				user, err = userByToken(ctx, logger, authService, jwtCookie.Value)
+			case "oidc_auth":
+				user, err = userFromOIDC(ctx, logger, authService, session, oidcConfig)
```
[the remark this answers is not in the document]
johnnyaug: Done: also added (temporary) migration from the old cookie, so that users are not logged out.

pkg/api/auth_middleware.go
```diff
@@ -105,6 +124,71 @@ func checkSecurityRequirements(r *http.Request, securityRequirements openapi3.Se
...
+	}
+	externalID, ok := idTokenClaims["sub"].(string)
+	if !ok {
+		logger.WithField("sub", idTokenClaims["sub"]).Error("Failed type assertion for sub claim")
+		return nil, ErrAuthenticatingRequest
+	}
+	friendlyName := ""
+	if oidcConfig.FriendlyNameClaimName != "" {
+		friendlyName, _ = idTokenClaims[oidcConfig.FriendlyNameClaimName].(string)
+	}
```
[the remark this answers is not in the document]
johnnyaug: Not doing - adds more complexity

pkg/api/controller.go
```diff
@@ -3401,7 +3443,10 @@ func NewController(
 	auditChecker AuditChecker,
 	logger logging.Logger,
 	emailer *email.Emailer,
+	oidcAuthenticator *oidc.Authenticator,
+	sessionStore sessions.Store,
 ) *Controller {
+	gob.Register(oidc.Claims{})
```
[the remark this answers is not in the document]
johnnyaug: Synced f2f: not really an option here since internally gorilla sessions have a `map[interface{}]interface{}` which is not support by the JSON encoder.

pkg/api/oidc_login_handler.go
```diff
@@ -0,0 +1,76 @@
...
+		session.Values[StateSessionKey] = state
+		if err := session.Save(r, w); err != nil {
+			logger.Errorf("failed to save oidc session: %w", err)
+			writeError(w, http.StatusInternalServerError, "Failed to redirect to login page")
+			return
+		}
+		scheme := "http"
+		if r.TLS != nil {
+			scheme = "https"
+		}
```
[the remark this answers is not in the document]
johnnyaug: Done

pkg/api/oidc_login_handler.go
```diff
@@ -0,0 +1,76 @@
...
+
+		session, err := sessionStore.Get(r, OIDCAuthSessionName)
+		if err != nil {
+			logger.Errorf("failed to get oidc session: %w", err)
+			writeError(w, http.StatusInternalServerError, "Failed to redirect to login page")
+			return
+		}
+		session.Values[StateSessionKey] = state
+		if err := session.Save(r, w); err != nil {
+			logger.Errorf("failed to save oidc session: %w", err)
```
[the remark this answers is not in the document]
johnnyaug: Done

pkg/api/oidc_login_handler.go
```diff
@@ -0,0 +1,76 @@
...
+		state, err := nanoid.New(stateLength)
+		if err != nil {
+			logger.Errorf("failed to generate state for oidc: %w", err)
+			writeError(w, http.StatusInternalServerError, "Failed to redirect to login page")
+			return
+		}
+
+		session, err := sessionStore.Get(r, OIDCAuthSessionName)
+		if err != nil {
+			logger.Errorf("failed to get oidc session: %w", err)
```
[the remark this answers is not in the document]
johnnyaug: Done

pkg/api/oidc_login_handler.go
```diff
@@ -0,0 +1,76 @@
...
+			Value:    "",
+			Domain:   cookieDomain,
+			Path:     "/",
+			HttpOnly: true,
+			Expires:  time.Unix(0, 0),
+			SameSite: http.SameSiteStrictMode,
+		})
+		state, err := nanoid.New(stateLength)
+		if err != nil {
+			logger.Errorf("failed to generate state for oidc: %w", err)
```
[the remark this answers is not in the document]
johnnyaug: Done

pkg/auth/oidc/authenticator.go
johnnyaug: Oops! Done

pkg/ddl/000038_oidc_user.up.sql
johnnyaug: 1. Changed to TEXT
2. According to @arielshaqed, living the down migration empty.

pkg/api/controller.go
johnnyaug: Fixed: now ignoring the error from `sessionStore.Get` when it makes sense.

cmd/lakefs/cmd/run.go
```diff
@@ -254,6 +256,25 @@ var runCmd = &cobra.Command{
...
+				cmd.Context(),
+				oidcConfig.URL,
+			)
+			if err != nil {
+				logger.WithError(err).Fatal("Failed to initialize OIDC provider")
+			}
+			oauthConfig = &oauth2.Config{
+				ClientID:     oidcConfig.ClientID,
+				ClientSecret: oidcConfig.ClientSecret,
+				RedirectURL:  oidcConfig.CallbackBaseURL + api.BaseURL + "/oidc/callback",
```
nopcoder, line 273: Need to trim leading "/" from callback base url so we will not end with "//"

pkg/api/auth_middleware.go
```diff
@@ -52,9 +59,44 @@ func AuthMiddleware(logger logging.Logger, swagger *openapi3.Swagger, authentica
 	}
 }
 
+// Deprecated
+// TODO(johnnyaug) remove this a week after released
+func migrateFromLegacyCookie(r *http.Request, w http.ResponseWriter, logger logging.Logger, sessionStore sessions.Store) {
```
nopcoder, line 64: Need to use code doc - start with the method name

pkg/api/auth_middleware.go
```diff
@@ -83,12 +125,31 @@ func checkSecurityRequirements(r *http.Request, securityRequirements openapi3.Se
...
+				var internalAuthSession *sessions.Session
+				internalAuthSession, _ = sessionStore.Get(r, InternalAuthSessionName)
+				token := ""
+				if internalAuthSession != nil {
+					token, _ = internalAuthSession.Values[TokenSessionKeyName].(string)
+				}
+				if token == "" {
+					migrateFromLegacyCookie(r, w, logger, sessionStore)
+				}
+				internalAuthSession, err = sessionStore.Get(r, InternalAuthSessionName)
```
nopcoder, line 136: We need to clear the cookie as you did - but reading the session after saving it - I don't think it will work.
When you "save" a cookie - you write a header. When you "read" a cookie, it is from the request header that we already parsed.
I maybe wrong as the code can implement some kind of workaround - just wanted to verify with you how it was implemented.

pkg/api/controller.go
```diff
@@ -160,21 +165,48 @@ func (c *Controller) DeleteObjects(w http.ResponseWriter, r *http.Request, body
...
+	err = session.Save(r, w)
+	if err != nil {
+		writeError(w, http.StatusInternalServerError, err.Error())
+		return
+	}
+	http.Redirect(w, r, "/", http.StatusTemporaryRedirect)
+}
+
 func (c *Controller) Login(w http.ResponseWriter, r *http.Request, body LoginJSONRequestBody) {
+	err := clearSession(w, r, c.sessionStore, OIDCAuthSessionName)
```
nopcoder, line 206: need to clear the old one - not the new one while login

pkg/api/controller.go
```diff
@@ -160,21 +165,48 @@ func (c *Controller) DeleteObjects(w http.ResponseWriter, r *http.Request, body
...
 	writeResponse(w, http.StatusOK, nil)
 }
 
+// OauthCallback gets a code generated by an OIDC provider.
+// It exchanges the code for an id token, and saves the claims from the ID token on a session.
+func (c *Controller) OauthCallback(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	session, _ := c.sessionStore.Get(r, OIDCAuthSessionName)
+	if r.URL.Query().Get("state") != session.Values[StateSessionKey] {
+		writeError(w, http.StatusBadRequest, "Invalid state parameter.")
```
nopcoder, line 188:
```suggestion
		writeError(w, http.StatusBadRequest, "Invalid state parameter")
```

pkg/api/oidc_login_handler.go
```diff
@@ -0,0 +1,43 @@
...
+			logger.WithError(err).Error("failed to generate state for oidc")
+			writeError(w, http.StatusInternalServerError, "Failed to redirect to login page")
+			return
+		}
+
+		session, _ := sessionStore.Get(r, OIDCAuthSessionName)
+		session.Values[StateSessionKey] = state
+		if err := session.Save(r, w); err != nil {
+			logger.WithError(err).Error("failed to save oidc session")
+			writeError(w, http.StatusInternalServerError, "Failed to redirect to login page")
```
nopcoder, line 38: If possible, change something small in the writeError in one of the two, so we can identify the issue without going into the logs. No need to expose more information if not needed, just a way to distinct between the two.

nopcoder (approved):
Added some comments, the one I was worried about is counting on save state follow up by get. All the rest looks great, thanks!

pkg/api/controller.go
johnnyaug: synced f2f: clearing the OIDC session here, and setting the "internal login" one.

PR #3471, opened 2022-06-07
Use a single actions service

itaiad200:


pkg/actions/store.go
```diff
@@ -0,0 +1,297 @@
+package actions
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/treeverse/lakefs/pkg/db"
```
N-o-Z, line 8: Redundant new lines

pkg/actions/store.go
```diff
@@ -0,0 +1,297 @@
+package actions
```
N-o-Z, line 1: I would still split implementations into 2 files so we can just delete the whole file when we finish the KV migration

pkg/actions/service.go
```diff
@@ -37,7 +36,8 @@ var (
 )
 
 type KVService struct {
```
N-o-Z, line 38: "Service"?

N-o-Z (changes requested):
Very small refactoring... XD
Added comments - seems like tests are failing as well

pkg/actions/service.go
itaiad200: It's the interface, changed the name to something else.

cmd/lakefs/cmd/run.go
```diff
@@ -180,7 +180,9 @@ var runCmd = &cobra.Command{
 		bufferedCollector.CollectMetadata(metadata)
 
 		var multipartsTracker multiparts.Tracker
```
N-o-Z, line 182: Suggest to join the var declarations

cmd/lakefs/cmd/run.go
```diff
@@ -191,26 +193,26 @@ var runCmd = &cobra.Command{
...
 			multipartsTracker = multiparts.NewTracker(storeMessage)
-			actionsService = actions.NewKVService(
-				ctx,
-				storeMessage,
-				catalog.NewActionsSource(c),
-				catalog.NewActionsOutputWriter(c.BlockAdapter),
-				bufferedCollector,
-				cfg.GetActionsEnabled(),
-			)
+
```
N-o-Z, line 196: Redundant new lines?

cmd/lakefs/cmd/run.go
```diff
@@ -191,26 +193,26 @@ var runCmd = &cobra.Command{
...
 			multipartsTracker = multiparts.NewDBTracker(dbPool)
-			actionsService = actions.NewDBService(
-				ctx,
-				dbPool,
-				catalog.NewActionsSource(c),
-				catalog.NewActionsOutputWriter(c.BlockAdapter),
-				bufferedCollector,
-				cfg.GetActionsEnabled(),
-			)
+
```
N-o-Z, line 201: same

pkg/loadtest/local_load_test.go
```diff
@@ -9,6 +9,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/treeverse/lakefs/pkg/kv"
+
```
N-o-Z, line 13: Newline

pkg/actions/id.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"time"
+
+	"github.com/rs/xid"
+
+	gonanoid "github.com/matoous/go-nanoid/v2"
+	"github.com/treeverse/lakefs/pkg/graveler"
+)
+
+type IDGenerator interface {
```
N-o-Z, line 12: Do we want to move the NewHookRunID functionality also under this interface?

pkg/actions/id.go
```diff
@@ -0,0 +1,33 @@
+package actions
+
+import (
+	"time"
+
+	"github.com/rs/xid"
+
```
N-o-Z, line 7: newline

N-o-Z (changes requested):
Looks really great, I added some more comments and suggestions

cmd/lakefs/cmd/run.go
itaiad200: Done.

cmd/lakefs/cmd/run.go
itaiad200: Done.

pkg/actions/id.go
itaiad200: I think not because it's the same for both implementations.

pkg/loadtest/local_load_test.go
itaiad200: Done.

pkg/actions/service.go
```diff
@@ -36,8 +35,12 @@ var (
 	ErrNilValue = errors.New("nil value")
 )
 
-type KVService struct {
-	Store    kv.StoreMessage
+// MultiStorageService is an implementation of actions.Service that saves
+// the run data to the blockstore and to the actions.Store (which is a
+// fancy name for a DB - kv style or postgres directly)
+type MultiStorageService struct {
```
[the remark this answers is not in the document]
itaiad200: I hate naming.

N-o-Z (approved):
Looks good :ok_hand:
Thanks for addressing all the comments

PR #3475, opened 2022-06-08
add reproducibility doc

peacing:
new explanation of the reproducibility concept.


docs/quickstart/reproducibility.md
```diff
@@ -0,0 +1,45 @@
...
+In a lakeFS repository, we are capable of taking many commits over the data, making many points in time reproducible. 
+
+![Commit History]({{ site.baseurl }}/assets/img/reproduce-commit-history.png)
+
+In the repository shown above, a new commit is taken each time a model training script is run, and the commit message includes the specific run number. 
+
+If we wanted to re-run the model training script and reproduce the exact same results for a historical run, say run #435, we could copy the commit ID associated with the run and read the data into a dataframe like so:
+
+```bash
+df = spark.read.parquet("s3://example/296e54fbee5e176f3f4f4aeb7e087f9d57515750e8c3d033b8b841778613cb23”)
```
talSofer, line 40: To read an object version at the time of a certain commit, users need to use commit as a ref, and replace the branch name with the commit id in the object patf. e.g.: 
val df2 = spark.read.parquet("s3a://my-repo/e2673a54a27d1a23bf4f71d13aec921bbe663c03211d96e518975a77347725ed/users.parquet")

docs/quickstart/reproducibility.md
```diff
@@ -0,0 +1,45 @@
...
+df = spark.read.parquet(‘s3://example/main/”)
+```
+
+In a lakeFS repository, we are capable of taking many commits over the data, making many points in time reproducible. 
+
+![Commit History]({{ site.baseurl }}/assets/img/reproduce-commit-history.png)
+
+In the repository shown above, a new commit is taken each time a model training script is run, and the commit message includes the specific run number. 
+
+If we wanted to re-run the model training script and reproduce the exact same results for a historical run, say run #435, we could copy the commit ID associated with the run and read the data into a dataframe like so:
```
talSofer, line 37: The ML model example is the easiest to follow, but please consider adding another example that will be applicable for data engineers type of users.

talSofer (changes requested):
Great writeup, and a very important piece thanks!
The example need to change, let me know your thoughts

docs/quickstart/reproducibility.md
```diff
@@ -0,0 +1,45 @@
...
+In comparison, lakeFS exposes a git-like interface to data that makes it possible to keep track of more than just the current state of data. This makes reproducing its state at any point in time straightforward.
+
+## Achieving Reproducibility with lakeFS
+
+To make data reproducible, we recommend taking a new commit of your lakeFS repository every time the data in it changes. As long as there’s a commit taken, the process to reproduce a given state is as simple as reading the data from a path that includes the unique `commit_id` generated for each commit.
+
+To read data at it’s current state, we can use a static path containing the repository and branch names. To give an example, if you have a repository named `example` with a branch named `main`, reading the latest state of this data into a Spark Dataframe is always:
+
+```bash
+df = spark.read.parquet(‘s3://example/main/”)
```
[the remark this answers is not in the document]
peacing: I tried uploading a parquet object to a lakeFS repo and using the above code to read it into a dataframe and it did work! Ok to leave as is?

docs/quickstart/reproducibility.md
peacing: Similarly, this also does work without the parquet suffix. I prefer not to reference specific files.. ok with you?

talSofer: No.. I think this is very confusing.. why do you prefer not to give an example that interacts with a specific object?

docs/quickstart/reproducibility.md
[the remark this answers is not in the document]
talSofer: I tried with more things under / and it's not working.. I think we should change the example

docs/quickstart/reproducibility.md
peacing: in reality, most times you want to read in a dataset it is a collection like "all orders at company X" and it involves many files that are all in the same format. 

Ideally you should never really be interacting with an individual file. it makes lakeFS seem limiting when you are pointing to and reading specific files.

talSofer: Good point, I see what you mean. 
Here is what I suggest - 
Can you use a path to an explicit data set? e.g. 
df = spark.read.parquet("s3://example/296e54fbee5e176f3f4f4aeb7e087f9d57515750e8c3d033b8b841778613cb23/orders”)

I also validated it works. My point is that it is confusing to point to a branch root. WDYT?

peacing: I like it! I don't think we've shown an example like that before and it's a good one.

talSofer (approved):
Thanks!

PR #3477, opened 2022-06-09
Fix bug: marshalling of db permission structs

johnnyaug:


itaiad200 (approved):
Yikes. Thanks for finding this bug and fixing it quickly!

We need to have tests in place to catch these next time (but not part of this PR). Should I open an issue for it?

johnnyaug:
@itaiad200 I think so, yes

PR #3484, opened 2022-06-09
Fix returned error for db Get method

guy-har:


nopcoder:
@guy-har can I close? this file no longer exists.

PR #3492, opened 2022-06-12
KV auth implemantaion

### LINKED ISSUE #3344: KV Auth Implementation
Implement Auth service using KV store
* **Without Migration**
Depend on #3348

eden-ohana:
closes #3344


pkg/auth/model/model.go
```diff
@@ -172,14 +177,11 @@ func (s *Statements) Scan(src interface{}) error {
 // authPrefix - key prefix in the KV model
 const authPrefix = "auth"
 
-//TODO - delete no lint comments when function is used
-//nolint
-func kvUserPath(userName string) string {
+func KvUserPath(userName string) string {
```
itaiad200, line 180:
```suggestion
func KVUserPath(userName string) string {
```

Makefile
```diff
@@ -235,6 +235,7 @@ proto: ## Build proto (Protocol Buffers) files
 	$(PROTOC) --proto_path=pkg/kv --go_out=pkg/kv --go_opt=paths=source_relative secondary_index.proto
 	$(PROTOC) --proto_path=pkg/gateway/multiparts --go_out=pkg/gateway/multiparts --go_opt=paths=source_relative multipart.proto
 	$(PROTOC) --proto_path=pkg/actions --go_out=pkg/actions --go_opt=paths=source_relative actions.proto
+	$(PROTOC) --proto_path=pkg/auth/model --go_out=pkg/auth/model --go_opt=paths=source_relative model.proto
```
itaiad200, line 238: It looks like it contains changes from the previous PR too. Commit message ("draft for the models") also suggests it. Are you on the latest `master`?

pkg/auth/model/model.go
```diff
@@ -193,26 +195,23 @@ func userFromProto(pb *UserData) *User {
...
+		FriendlyName:      swag.StringValue(u.FriendlyName),
+		Email:             swag.StringValue(u.Email),
 		EncryptedPassword: u.EncryptedPassword,
 		Source:            u.Source,
 	}
 }
 
-//nolint
-func kvGroupPath(displayName string) string {
+func KvGroupPath(displayName string) string {
```
itaiad200, line 210:
```suggestion
func KVGroupPath(displayName string) string {
```

pkg/auth/model/model.go
```diff
@@ -222,45 +221,40 @@ func groupFromProto(pb *GroupData) *Group {
...
 		CreatedAt:   timestamppb.New(p.CreatedAt),
 		DisplayName: p.DisplayName,
 		Statements:  protoFromStatements(&p.Statement),
+		Id:          []byte(id),
 	}
 }
 
-//nolint
-func kvCredentialPath(userName string, accessKeyID string) string {
+func KvCredentialPath(userName string, accessKeyID string) string {
```
itaiad200, line 253:
```suggestion
func KVCredentialPath(userName string, accessKeyID string) string {
```

pkg/auth/model/model.go
```diff
@@ -301,38 +292,31 @@ func protoFromStatement(s *Statement) *StatementData {
...
 	}
-	return &StatementsData{
-		Statement: statements,
-	}
+	return statements
 }
 
-//nolint
-func kvUserToGroup(userName string, groupDisplayName string) string {
+func KvUserToGroup(userName string, groupDisplayName string) string {
```
itaiad200, line 312:
```suggestion
func KVUserToGroup(userName string, groupDisplayName string) string {
```

pkg/auth/model/model.go
```diff
@@ -374,3 +377,25 @@ func ConvertGroup(g *DBGroup) *Group {
...
+func ConvertPolicyDataList(policies []proto.Message) []*BasePolicy {
+	res := make([]*BasePolicy, 0, len(policies))
+	for _, p := range policies {
+		a := p.(*PolicyData)
+		res = append(res, PolicyFromProto(a))
+	}
+	return res
+}
+
+func KvTokenPath(tokenID string) string {
```
itaiad200, line 399:
```suggestion
func KVTokenPath(tokenID string) string {
```

itaiad200 (changes requested):
Very initial review - will continue tomorrow.

pkg/auth/service.go
```diff
@@ -474,299 +473,275 @@ func (s *DBAuthService) ListEffectivePolicies(ctx context.Context, username stri
...
+			return nil, nil, err
+		}
+		for _, group := range groups {
+			path := model.KvUserToGroup(username, group.DisplayName)
+			m := model.UserData{}
+			_, err := s.store.GetMsg(ctx, model.PartitionKey, path, &m)
+			if err != nil && !errors.Is(err, kv.ErrNotFound) {
+				return nil, nil, err
+			}
+			if !errors.Is(err, kv.ErrNotFound) {
```
itaiad200, line 621: Why do you check it twice?

pkg/auth/service.go
```diff
@@ -474,299 +473,275 @@ func (s *DBAuthService) ListEffectivePolicies(ctx context.Context, username stri
...
-				AND auth_users.display_name like $3
-			ORDER BY auth_groups.display_name
-			LIMIT $4`,
-			groupDisplayName, params.After, fmt.Sprint(params.Prefix, "%"), params.Amount+1)
+
+	hasMoreGroups := true
+	nextPage := ""
+	count := 0
+	userGroups := make([]*model.Group, 0)
+	for hasMoreGroups && count < params.Amount {
```
itaiad200, line 609: The default amount is 0, no?

pkg/auth/service.go
```diff
@@ -474,299 +473,275 @@ func (s *DBAuthService) ListEffectivePolicies(ctx context.Context, username stri
...
-			WHERE
-				auth_groups.display_name = $1
-				AND auth_users.display_name > $2
-				AND auth_users.display_name like $3
-			ORDER BY auth_groups.display_name
-			LIMIT $4`,
-			groupDisplayName, params.After, fmt.Sprint(params.Prefix, "%"), params.Amount+1)
+
+	hasMoreGroups := true
+	nextPage := ""
```
itaiad200, line 606: nextPage should be declared above his usage, after the for loop

pkg/auth/service.go
```diff
@@ -474,299 +473,275 @@ func (s *DBAuthService) ListEffectivePolicies(ctx context.Context, username stri
...
+			hasMoreGroups = false
 		}
-		if len(users) == params.Amount+1 {
-			// we have more pages
-			users = users[0:params.Amount]
-			p.Amount = params.Amount
-			p.NextPageToken = users[len(users)-1].Username
-			return &res{users, p}, nil
+		params = &model.PaginationParams{
+			After: paginator.NextPageToken,
```
itaiad200, line 630: What about the other properties of `params`? Does it make sense to fetch for the default amount when much less was requested?

pkg/auth/service.go
```diff
@@ -474,299 +473,275 @@ func (s *DBAuthService) ListEffectivePolicies(ctx context.Context, username stri
...
+				return err
 			}
 		}
+		if err := model.ValidateArn(stmt.Resource); err != nil {
+			return err
+		}
+		if err := model.ValidateStatementEffect(stmt.Effect); err != nil {
+			return err
+		}
+	}
```
itaiad200, line 670: Maybe extract validation to a common place?

pkg/auth/service.go
```diff
@@ -474,299 +473,275 @@ func (s *DBAuthService) ListEffectivePolicies(ctx context.Context, username stri
...
-func (s *DBAuthService) GetPolicy(ctx context.Context, policyDisplayName string) (*model.BasePolicy, error) {
-	policy, err := s.db.Transact(ctx, func(tx db.Tx) (interface{}, error) {
-		return getPolicy(tx, policyDisplayName)
-	}, db.ReadOnly())
+func (s *KVAuthService) GetPolicy(ctx context.Context, policyDisplayName string) (*model.BasePolicy, error) {
+	policyKey := model.PolicyPath(policyDisplayName)
+	p := model.PolicyData{}
+	_, err := s.store.GetMsg(ctx, model.PartitionKey, policyKey, &p)
 	if err != nil {
 		return nil, err
```
itaiad200, line 686: converting ErrNotFound

pkg/auth/service.go
```diff
@@ -778,152 +753,148 @@ func (s *DBAuthService) AddCredentials(ctx context.Context, username, accessKeyI
...
-				AccessKeyID:                   accessKeyID,
-				SecretAccessKey:               secretAccessKey,
-				SecretAccessKeyEncryptedBytes: encryptedKey,
-				IssuedDate:                    now,
-			},
-			UserID: user.ID,
-		}
+	user, err := s.GetUser(ctx, username)
+	if err != nil {
+		return nil, fmt.Errorf("%s: %w", username, err)
```
itaiad200, line 758: wrong formatting

pkg/auth/service.go
```diff
@@ -778,152 +753,148 @@ func (s *DBAuthService) AddCredentials(ctx context.Context, username, accessKeyI
...
+		BaseCredential: model.BaseCredential{
+			AccessKeyID:                   accessKeyID,
+			SecretAccessKey:               secretAccessKey,
+			SecretAccessKeyEncryptedBytes: encryptedKey,
+			IssuedDate:                    now,
+		},
+		UserID: user.ID,
+	}
+	credentialsKey := model.KvCredentialPath(user.Username, c.AccessKeyID)
+	err = s.store.SetMsg(ctx, model.PartitionKey, credentialsKey, model.ProtoFromCredential(c))
```
itaiad200, line 771: `SetMsgIf` with nil? to avoid having a key overwritten..

pkg/auth/service.go
```diff
@@ -778,152 +753,148 @@ func (s *DBAuthService) AddCredentials(ctx context.Context, username, accessKeyI
...
-		if err != nil {
-			return nil, err
-		}
-		return credentials, nil
-	})
+func (s *KVAuthService) GetCredentialsForUser(ctx context.Context, username, accessKeyID string) (*model.Credential, error) {
+	if _, err := s.GetUser(ctx, username); err != nil {
+		return nil, fmt.Errorf("%s: %w", username, err)
+	}
+	creadantialsKey := model.KvCredentialPath(username, accessKeyID)
```
itaiad200, line 838: typo

pkg/auth/service.go
```diff
@@ -778,152 +753,148 @@ func (s *DBAuthService) AddCredentials(ctx context.Context, username, accessKeyI
...
+		user, ok := entry.Value.(*model.UserData)
+		if !ok {
+			return nil, fmt.Errorf("list users: %w", err)
+		}
+		c := model.CredentialData{}
+		credentialsKey := model.KvCredentialPath(user.Username, accessKeyID)
+		_, err := s.store.GetMsg(ctx, model.PartitionKey, credentialsKey, &c)
+		if err != nil && errors.Is(err, kv.ErrNotFound) {
 			return nil, err
+		} else if !errors.Is(err, kv.ErrNotFound) {
```
itaiad200, line 870: why do you check it again?

pkg/auth/service.go
```diff
@@ -1027,7 +998,7 @@ func (s *DBAuthService) Authorize(ctx context.Context, req *AuthorizationRequest
 	return &AuthorizationResponse{Allowed: true}, nil
 }
 
-func (s *DBAuthService) ClaimTokenIDOnce(ctx context.Context, tokenID string, expiresAt int64) error {
+func (s *KVAuthService) ClaimTokenIDOnce(ctx context.Context, tokenID string, expiresAt int64) error {
 	tokenExpiresAt := time.Unix(expiresAt, 0)
```
itaiad200, line 1002: Looks unrelated to kv

pkg/auth/service.go
```diff
@@ -1040,19 +1011,39 @@ func (s *DBAuthService) ClaimTokenIDOnce(ctx context.Context, tokenID string, ex
...
+	m := model.TokenData{TokenId: tokenID, ExpiredAt: timestamppb.New(tokenExpiresAt)}
+	err := s.store.SetMsg(ctx, model.PartitionKey, tokenPath, &m)
 	if err != nil {
 		return false, err
 	}
-	canUseToken := res.RowsAffected() == 1
+
 	// cleanup old tokens
-	_, err = s.db.Exec(ctx, `DELETE FROM auth_expired_tokens WHERE token_expires_at < $1`, time.Now())
+	t := model.TokenData{}
```
itaiad200, line 1023: This is where we differ from DB. I don't think this is the appropriate place for O(n) cleanups. DB cleanups are O(1) and therefore can be done here.

pkg/auth/service.go
```diff
@@ -1040,19 +1011,39 @@ func (s *DBAuthService) ClaimTokenIDOnce(ctx context.Context, tokenID string, ex
...
+			if err != nil {
+				s.log.WithError(err).Error("delete expired tokens")
+			}
+		}
+	}
+	if err = itr.Err(); err != nil {
+		return true, err
 	}
-	return canUseToken, nil
+	return true, nil
```
itaiad200, line 1046: Is it always true?

Makefile
eden-ohana: On the auth models pr we didn't add it, correct me if I'm wrong. I didn't see makefile in that PR.

pkg/auth/model/model.go
eden-ohana: done

pkg/auth/model/model.go
eden-ohana: done

pkg/auth/model/model.go
eden-ohana: done

pkg/auth/model/model.go
eden-ohana: done

pkg/auth/service.go
```diff
@@ -12,20 +12,25 @@ import (
 	"strings"
 	"time"
 
+	"google.golang.org/protobuf/types/known/timestamppb"
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -125,340 +130,334 @@ func fieldByTag(t reflect.Type, key, tag string) string {
...
-	return model.ConvertUser(user), nil
-}
+	count := 0
 
-func getGroup(tx db.Tx, groupDisplayName string) (*model.DBGroup, error) {
-	group := &model.DBGroup{}
-	err := tx.Get(group, `SELECT * FROM auth_groups WHERE display_name = $1`, groupDisplayName)
-	if err != nil {
-		return nil, err
+	entries := make([]proto.Message, 0)
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -125,340 +130,334 @@ func fieldByTag(t reflect.Type, key, tag string) string {
...
+	var lastKey string
+	for it.Next() && count < amount {
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -125,340 +130,334 @@ func fieldByTag(t reflect.Type, key, tag string) string {
...
+	userPath := model.KvUserPath(username)
+
+	// delete policy attached to user
+	policiesKey := model.PolicyPath("")
+	it, err := kv.NewSecondaryIterator(ctx, s.store.Store, (&model.PolicyData{}).ProtoReflect().Type(), model.PartitionKey, policiesKey, "")
+	for it.Next() {
+		entry := it.Entry()
+		policy := entry.Value.(*model.PolicyData)
+		if err = s.DetachPolicyFromUser(ctx, policy.DisplayName, username); err != nil {
+			return fmt.Errorf("%s: %w", policy.DisplayName, err)
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -125,340 +130,334 @@ func fieldByTag(t reflect.Type, key, tag string) string {
...
+		}
+	}
+	if err := itr.Err(); err != nil {
+		return err
+	}
+
+	// delete user
+	err = s.store.DeleteMsg(ctx, model.PartitionKey, userPath)
+	if err != nil {
+		return fmt.Errorf("delete user (usesrKey %s): %w", userPath, err)
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -125,340 +130,334 @@ func fieldByTag(t reflect.Type, key, tag string) string {
...
 	return s.cache.GetUser(username, func() (*model.User, error) {
-		user, err := s.db.Transact(ctx, func(tx db.Tx) (interface{}, error) {
-			return getUser(tx, username)
-		}, db.ReadOnly())
+		userKey := model.KvUserPath(username)
+		m := model.UserData{}
+		_, err := s.store.GetMsg(ctx, model.PartitionKey, userKey, &m)
 		if err != nil {
+			if errors.Is(err, kv.ErrNotFound) {
+				err = fmt.Errorf("%s: %w", err, ErrNotFound) // Wrap error for compatibility with DBService
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -125,340 +130,334 @@ func fieldByTag(t reflect.Type, key, tag string) string {
...
-	})
+func (s *KVAuthService) DeleteUser(ctx context.Context, username string) error {
+	if _, err := s.GetUser(ctx, username); err != nil {
+		return fmt.Errorf("%s: %w", username, err)
+	}
+	userPath := model.KvUserPath(username)
+
+	// delete policy attached to user
+	policiesKey := model.PolicyPath("")
+	it, err := kv.NewSecondaryIterator(ctx, s.store.Store, (&model.PolicyData{}).ProtoReflect().Type(), model.PartitionKey, policiesKey, "")
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -125,340 +130,334 @@ func fieldByTag(t reflect.Type, key, tag string) string {
...
+			return fmt.Errorf("%s: %w", policy.DisplayName, err)
+		}
+	}
+	if err := it.Err(); err != nil {
+		return err
+	}
+
+	// delete user membership of group
+	groupKey := model.KvGroupPath("")
+	itr, err := kv.NewPrimaryIterator(ctx, s.store.Store, (&model.GroupData{}).ProtoReflect().Type(), model.PartitionKey, groupKey, "")
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -125,340 +130,334 @@ func fieldByTag(t reflect.Type, key, tag string) string {
...
-func (s *DBAuthService) GetUserByEmail(ctx context.Context, email string) (*model.User, error) {
-	user, err := s.db.Transact(ctx, func(tx db.Tx) (interface{}, error) {
-		return getUserByEmail(tx, email)
-	}, db.ReadOnly())
+func (s *KVAuthService) GetUserByEmail(ctx context.Context, email string) (*model.User, error) {
+	m := &model.UserData{}
+	itr, err := s.store.Scan(ctx, m.ProtoReflect().Type(), model.PartitionKey, model.KvUserPath(""))
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("sacn users users: %w", err)
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -125,340 +130,334 @@ func fieldByTag(t reflect.Type, key, tag string) string {
...
-		return user.(*model.User), nil
+		return model.UserFromProto(&m), nil
 	})
 }
 
-func (s *DBAuthService) GetUserByEmail(ctx context.Context, email string) (*model.User, error) {
-	user, err := s.db.Transact(ctx, func(tx db.Tx) (interface{}, error) {
-		return getUserByEmail(tx, email)
-	}, db.ReadOnly())
+func (s *KVAuthService) GetUserByEmail(ctx context.Context, email string) (*model.User, error) {
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -125,340 +130,334 @@ func fieldByTag(t reflect.Type, key, tag string) string {
...
-			if err != nil {
-				return nil, err
-			}
-			return user, nil
-		}, db.ReadOnly())
+		m := &model.UserData{}
+		itr, err := s.store.Scan(ctx, m.ProtoReflect().Type(), model.PartitionKey, model.KvUserPath(""))
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("sacn users users: %w", err)
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
eden-ohana: opened issue #3500

pkg/auth/service.go
eden-ohana: It's an interface method, so we need to implement it

pkg/auth/service.go
eden-ohana: It is true if the save of the token in the DB succeeded

pkg/auth/service.go
```diff
@@ -125,340 +130,334 @@ func fieldByTag(t reflect.Type, key, tag string) string {
...
-func deleteOrNotFound(tx db.Tx, stmt string, args ...interface{}) error {
-	res, err := tx.Exec(stmt, args...)
-	if err != nil {
-		return err
-	}
-	numRows := res.RowsAffected()
-	if numRows == 0 {
-		return ErrNotFound
+	p := &model.Paginator{}
+	if params != nil && params.Amount >= 0 && len(entries) == params.Amount {
```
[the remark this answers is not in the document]
eden-ohana: 1. done
2. if params.amout == -1, we need to list all (the convention on the tests)
3. I don't use it, maybe not the latest version.

pkg/auth/service.go
```diff
@@ -125,340 +130,334 @@ func fieldByTag(t reflect.Type, key, tag string) string {
...
+func (s *KVAuthService) encryptSecret(secretAccessKey string) ([]byte, error) {
 	encrypted, err := s.secretStore.Encrypt([]byte(secretAccessKey))
 	if err != nil {
 		return nil, err
 	}
 	return encrypted, nil
 }
 
-func (s *DBAuthService) SecretStore() crypt.SecretStore {
+func (s *KVAuthService) SecretStore() crypt.SecretStore {
```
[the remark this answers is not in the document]
eden-ohana: It's an interface method, so we need to implement it

pkg/auth/service.go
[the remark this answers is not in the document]
eden-ohana: the prefix was wrong, I changed it
those functions already exist but I'm not using them to make to implementation clearer

pkg/auth/service.go
```diff
@@ -125,340 +130,334 @@ func fieldByTag(t reflect.Type, key, tag string) string {
...
+		return err
+	}
+
+	// delete user membership of group
+	groupKey := model.KvGroupPath("")
+	itr, err := kv.NewPrimaryIterator(ctx, s.store.Store, (&model.GroupData{}).ProtoReflect().Type(), model.PartitionKey, groupKey, "")
+	for itr.Next() {
+		entry := itr.Entry()
+		group := entry.Value.(*model.GroupData)
+		if err = s.RemoveUserFromGroup(ctx, username, group.DisplayName); err != nil {
```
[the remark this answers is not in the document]
eden-ohana: Nothing will happen because deleteMsg on the KV interface doesn't return error if not exist

pkg/auth/service.go
```diff
@@ -125,340 +130,334 @@ func fieldByTag(t reflect.Type, key, tag string) string {
...
-			Where(sq.And{
-				sq.Eq{"user_display_name": username},
-				sq.Like{"display_name": fmt.Sprint(params.Prefix, "%")},
-			}))
-
-	if slice == nil {
-		return nil, paginator, err
+	if _, err := s.GetPolicy(ctx, policyDisplayName); err != nil {
+		return fmt.Errorf("%s: %w", policyDisplayName, err)
 	}
```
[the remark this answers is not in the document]
eden-ohana: I can't think of one, the flow of deleting an entity first removes all it's secondary indexes

pkg/auth/service.go
```diff
@@ -125,340 +130,334 @@ func fieldByTag(t reflect.Type, key, tag string) string {
...
-	for _, p := range policies {
-		res = append(res, &p.BasePolicy)
+
+	pu := model.PolicyToUser(policyDisplayName, username)
+	err := s.store.DeleteMsg(ctx, model.PartitionKey, pu)
+	if err != nil {
+		return fmt.Errorf("policy detachment to user: (key %s): %w", pu, err)
 	}
-	return res, paginator, nil
+	return err
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -125,340 +130,334 @@ func fieldByTag(t reflect.Type, key, tag string) string {
...
 }
 
-func (s *DBAuthService) ListEffectivePolicies(ctx context.Context, username string, params *model.PaginationParams) ([]*model.BasePolicy, *model.Paginator, error) {
+func (s *KVAuthService) ListUserPolicies(ctx context.Context, username string, params *model.PaginationParams) ([]*model.BasePolicy, *model.Paginator, error) {
+	var policy model.PolicyData
+	userPolicyKey := model.PolicyToUser("", username)
```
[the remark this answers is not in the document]
eden-ohana: It was true but confusing so I changes the parameters order in the function

pkg/auth/service.go
```diff
@@ -474,299 +473,275 @@ func (s *DBAuthService) ListEffectivePolicies(ctx context.Context, username stri
...
-		Where(sq.And{
-			sq.Eq{"auth_groups.display_name": groupDisplayName},
-			sq.Like{"auth_policies.display_name": fmt.Sprint(params.Prefix, "%")},
-		})
-	slice, paginator, err := ListPaged(ctx, s.db, reflect.TypeOf(policy), params, "display_name",
-		psql.Select("*").FromSelect(query, "p"))
-	if err != nil {
+func (s *KVAuthService) ListGroupPolicies(ctx context.Context, groupDisplayName string, params *model.PaginationParams) ([]*model.BasePolicy, *model.Paginator, error) {
+	var policy model.PolicyData
+	groupPolicyKey := model.PolicyToGroup("", groupDisplayName)
```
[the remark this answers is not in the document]
eden-ohana: It was true but confusing so I changes the parameters order in the function

pkg/auth/service.go
```diff
@@ -474,299 +473,275 @@ func (s *DBAuthService) ListEffectivePolicies(ctx context.Context, username stri
...
+func (s *KVAuthService) CreateGroup(ctx context.Context, group *model.BaseGroup) error {
+	if err := model.ValidateAuthEntityID(group.DisplayName); err != nil {
+		return err
+	}
+
+	groupKey := model.KvGroupPath(group.DisplayName)
+	id := uuid.New().String()
+	groupWithID := model.Group{ID: id, BaseGroup: *group}
+
+	err := s.store.SetMsg(ctx, model.PartitionKey, groupKey, model.ProtoFromGroup(&groupWithID))
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
eden-ohana: done

pkg/auth/service.go
eden-ohana: I don't understand why?

pkg/auth/service.go
eden-ohana: done

pkg/auth/service.go
eden-ohana: done

pkg/auth/service.go
eden-ohana: Because If the credentials doesn't exist I don't want to exit I want to continue until I found the user that has these credentials

pkg/auth/service.go
eden-ohana: done

pkg/auth/service.go
eden-ohana: Irrelevant, not the latest version.

pkg/auth/service.go
eden-ohana: In order to not exit if the user is not member of that group

pkg/auth/service.go
eden-ohana: Irrelevant, not the latest version.

cmd/lakefs/cmd/setup.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/treeverse/lakefs/pkg/kv"
+
```
N-o-Z, line 9: Redundant new line

cmd/lakefs/cmd/superuser.go
```diff
@@ -6,6 +6,8 @@ import (
 	"os"
 	"time"
 
+	"github.com/treeverse/lakefs/pkg/kv"
+
```
N-o-Z, line 10: Redundant new line

cmd/lakefs/cmd/superuser.go
```diff
@@ -24,6 +26,7 @@ var superuserCmd = &cobra.Command{
 		cfg := loadConfig()
 		ctx := cmd.Context()
 		dbPool := db.BuildDatabaseConnection(ctx, cfg.GetDatabaseParams())
+		dbParams := cfg.GetDatabaseParams()
```
N-o-Z, line 29: You can define dbParams before dbPool and then pass it in the BuildDatabaseConnection

cmd/lakefs/cmd/superuser.go
```diff
@@ -42,7 +45,19 @@ var superuserCmd = &cobra.Command{
 			os.Exit(1)
 		}
 
-		authService := auth.NewDBAuthService(dbPool, crypt.NewSecretStore(cfg.GetAuthEncryptionSecret()), cfg.GetAuthCacheConfig(), nil)
+		var authService auth.Service
+		if dbParams.KVEnabled {
```
N-o-Z, line 49: I think that after the 3rd time of this code repeating itself - it might be a good time to create a factory for NewAuthService (consider the solution with the store that @itaiad200 implemented in actions pkg)

pkg/auth/model/model.go
```diff
@@ -222,45 +219,40 @@ func groupFromProto(pb *GroupData) *Group {
...
 		CreatedAt:   timestamppb.New(g.CreatedAt),
 		DisplayName: g.DisplayName,
 	}
 }
 
-//nolint
-func PolicyPath(DisplayName string) string {
-	return kv.FormatPath(authPrefix, "policies", DisplayName)
+func KVPolicyPath(displayName string) string {
+	return kv.FormatPath(authPrefix, "policies", displayName)
```
N-o-Z, line 231: Use consts for entity names

pkg/auth/model/model.go
```diff
@@ -222,45 +219,40 @@ func groupFromProto(pb *GroupData) *Group {
...
 }
 
-//nolint
-func protoFromPolicy(p *BasePolicy) *PolicyData {
+func ProtoFromPolicy(p *BasePolicy, id string) *PolicyData {
 	return &PolicyData{
 		CreatedAt:   timestamppb.New(p.CreatedAt),
 		DisplayName: p.DisplayName,
 		Statements:  protoFromStatements(&p.Statement),
+		Id:          []byte(id),
```
N-o-Z, line 247: As discussed before - if display name is a unique key - why do we need Id?

pkg/auth/model/model.go
```diff
@@ -353,6 +344,16 @@ func ConvertCredList(creds []*DBCredential) []*Credential {
 	return res
 }
 
+func ConvertCredDataList(creds []proto.Message) []*Credential {
+	res := make([]*Credential, 0, len(creds))
+	for _, c := range creds {
+		a := c.(*CredentialData)
```
N-o-Z, line 350: Did you check if we can omit the SecretAccessKey from CredentialData and leave only the encrypted value?

pkg/auth/service.go
```diff
@@ -125,340 +129,366 @@ func fieldByTag(t reflect.Type, key, tag string) string {
...
-func ListPaged(ctx context.Context, db db.Querier, retType reflect.Type, params *model.PaginationParams, tokenColumnName string, queryBuilder sq.SelectBuilder) (*reflect.Value, *model.Paginator, error) {
-	ptrType := reflect.PtrTo(retType)
-	tokenField := fieldByTag(retType, "db", tokenColumnName)
-	if tokenField == "" {
-		return nil, nil, fmt.Errorf("[I] no field %s: %w", tokenColumnName, ErrNoField)
-	}
-	slice := reflect.MakeSlice(reflect.SliceOf(ptrType), 0, 0)
-	queryBuilder = queryBuilder.OrderBy(tokenColumnName)
-	amount := 0
+func (s *KVAuthService) ListKVPaged(ctx context.Context, protoType protoreflect.MessageType, params *model.PaginationParams, prefix string, secondary bool) ([]proto.Message, *model.Paginator, error) {
```
N-o-Z, line 132: Batch const declarations together

pkg/loadtest/local_load_test.go
```diff
@@ -59,20 +59,33 @@ func (m *nullCollector) SetInstallationID(_ string) {}
...
 }
 
-func GetKVService(t *testing.T, ctx context.Context, source actions.Source, writer actions.OutputWriter, stats stats.Collector, runHooks bool) actions.Service {
+func GetKVActionsService(t *testing.T, ctx context.Context, source actions.Source, writer actions.OutputWriter, stats stats.Collector, runHooks bool) actions.Service {
 	t.Helper()
 	kvStore := kvtest.GetStore(ctx, t)
 	return actions.NewService(ctx, actions.NewActionsKVStore(kv.StoreMessage{Store: kvStore}), source, writer, &actions.DecreasingIDGenerator{}, stats, runHooks)
 }
 
+func GetDBAuthService(t *testing.T) auth.Service {
```
N-o-Z, line 76: Maybe we can split it to two versions of the load test (KV/DB) and then create an init function that will return the services according to the given configuration

N-o-Z (changes requested):
@eden-ohana Made a first look of the code changes.
Looks great, I've looked mostly at the service code since you mentioned tests are not complete.
Thanks!

pkg/auth/service.go
```diff
@@ -805,11 +877,14 @@ func (s *KVAuthService) AttachPolicyToGroup(ctx context.Context, policyDisplayNa
...
-	policyKey := model.PolicyPath(policyDisplayName)
-	pg := model.PolicyToGroup(policyDisplayName, groupDisplayName)
+	policyKey := model.KVPolicyPath(policyDisplayName)
+	pg := model.KVPolicyToGroup(groupDisplayName, policyDisplayName)
 
-	err := s.store.SetMsg(ctx, model.PartitionKey, pg, &kv.SecondaryIndex{PrimaryKey: []byte(policyKey)})
+	err := s.store.SetMsgIf(ctx, model.PartitionKey, pg, &kv.SecondaryIndex{PrimaryKey: []byte(policyKey)}, nil)
 	if err != nil {
+		if errors.Is(err, kv.ErrPredicateFailed) {
+			err = ErrAlreadyExists
```
itaiad200, line 886: Is this how the DB implementation behave for recreation of the attachment ?

pkg/auth/service.go
```diff
@@ -835,43 +910,48 @@ func (s *KVAuthService) GetCredentialsForUser(ctx context.Context, username, acc
...
-	defer itr.Close()
-
-	for itr.Next() {
-		if itr.Err() != nil {
-			return nil, itr.Err()
+	return s.cache.GetCredential(accessKeyID, func() (*model.Credential, error) {
+		m := &model.UserData{}
+		itr, err := s.store.Scan(ctx, m.ProtoReflect().Type(), model.PartitionKey, model.KVUserPath(""))
+		if err != nil {
+			return nil, fmt.Errorf("sacn users: %w", err)
```
itaiad200, line 931:
```suggestion
			return nil, fmt.Errorf("scan users: %w", err)
```

pkg/auth/service.go
itaiad200: Basically because go errors are not debug log and extending this behaviour throughout the different layers will result with errors like: `<username>: <username>: <username>: <username>: some db error`

pkg/auth/service.go
```diff
@@ -835,43 +910,48 @@ func (s *KVAuthService) GetCredentialsForUser(ctx context.Context, username, acc
...
-		credentialsKey := model.KvCredentialPath(user.Username, accessKeyID)
-		_, err := s.store.GetMsg(ctx, model.PartitionKey, credentialsKey, &c)
-		if err != nil && errors.Is(err, kv.ErrNotFound) {
-			return nil, err
-		} else if !errors.Is(err, kv.ErrNotFound) {
-			return model.CredentialFromProto(&c), err
+		defer itr.Close()
+
+		for itr.Next() {
+			if itr.Err() != nil {
```
itaiad200, line 936: `itr.Next()==true` cannot result with an error

pkg/auth/service.go
```diff
@@ -835,43 +910,48 @@ func (s *KVAuthService) GetCredentialsForUser(ctx context.Context, username, acc
...
+				return nil, itr.Err()
+			}
+			entry := itr.Entry()
+			user, ok := entry.Value.(*model.UserData)
+			if !ok {
+				return nil, fmt.Errorf("list users: %w", err)
```
itaiad200, line 942: Where's the error here?

itaiad200: I mean, `err` might be nil

pkg/auth/service.go
```diff
@@ -835,43 +910,48 @@ func (s *KVAuthService) GetCredentialsForUser(ctx context.Context, username, acc
...
+			}
+			c := model.CredentialData{}
+			credentialsKey := model.KVCredentialPath(user.Username, accessKeyID)
+			_, err := s.store.GetMsg(ctx, model.PartitionKey, credentialsKey, &c)
+			if err != nil && !errors.Is(err, kv.ErrNotFound) {
+				return nil, err
+			} else if !errors.Is(err, kv.ErrNotFound) {
```
itaiad200, line 949:
```suggestion
			} else if err==nil {
```

pkg/auth/service.go
itaiad200: Extracting common code to a shared function is better.

pkg/auth/service.go
```diff
@@ -1012,37 +1092,13 @@ func (s *KVAuthService) ClaimTokenIDOnce(ctx context.Context, tokenID string, ex
...
 // markTokenSingleUse returns true if token is valid for single use
 func (s *KVAuthService) markTokenSingleUse(ctx context.Context, tokenID string, tokenExpiresAt time.Time) (bool, error) {
-	tokenPath := model.KvTokenPath(tokenID)
+	tokenPath := model.KVTokenPath(tokenID)
 	m := model.TokenData{TokenId: tokenID, ExpiredAt: timestamppb.New(tokenExpiresAt)}
-	err := s.store.SetMsg(ctx, model.PartitionKey, tokenPath, &m)
+	err := s.store.SetMsgIf(ctx, model.PartitionKey, tokenPath, &m, nil)
 	if err != nil {
 		return false, err
 	}
```
itaiad200, line 1100: See the DB behaviour here. If it failed do to a predicate - we should return `false, nil`. If it didn't fail, return `true, nil`.

itaiad200 (changes requested):

pkg/auth/model/model.go
```diff
@@ -193,26 +193,23 @@ func userFromProto(pb *UserData) *User {
...
+		Email:             swag.StringValue(u.Email),
 		EncryptedPassword: u.EncryptedPassword,
 		Source:            u.Source,
 	}
 }
 
-//nolint
-func kvGroupPath(displayName string) string {
+func KVGroupPath(displayName string) string {
 	return kv.FormatPath(authPrefix, "groups", displayName)
```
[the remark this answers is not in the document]
eden-ohana: done

cmd/lakefs/cmd/setup.go
eden-ohana: done

cmd/lakefs/cmd/superuser.go
eden-ohana: done

cmd/lakefs/cmd/superuser.go
eden-ohana: done

pkg/auth/model/model.go
```diff
@@ -177,7 +175,7 @@ func (s *Statements) Scan(src interface{}) error {
 // authPrefix - key prefix in the KV model
 const authPrefix = "auth"
```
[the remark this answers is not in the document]
eden-ohana: removed

pkg/auth/model/model.go
```diff
@@ -396,6 +394,6 @@ func ConvertPolicyDataList(policies []proto.Message) []*BasePolicy {
 	return res
 }
 
-func KvTokenPath(tokenID string) string {
+func KVTokenPath(tokenID string) string {
```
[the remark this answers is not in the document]
eden-ohana: done

cmd/lakefs/cmd/superuser.go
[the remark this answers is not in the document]
eden-ohana: done

cmd/lakefs/cmd/setup.go
```diff
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/treeverse/lakefs/pkg/kv"
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/api/serve_test.go
```diff
@@ -95,14 +95,22 @@ func setupHandlerWithWalkerFactory(t testing.TB, factory catalog.WalkerFactory,
 	// wire actions
 	var actionsStore actions.Store
 	var idGen actions.IDGenerator
+	var authService auth.Service
 
 	if kvEnabled {
 		kvStore := kvtest.GetStore(ctx, t)
 		actionsStore = actions.NewActionsKVStore(kv.StoreMessage{Store: kvStore})
 		idGen = &actions.DecreasingIDGenerator{}
+		storeMessage := kv.StoreMessage{Store: kvStore}
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -219,69 +203,82 @@ func (s *KVAuthService) SecretStore() crypt.SecretStore {
...
 	id := uuid.New().String()
 	userWithID := model.User{ID: id, BaseUser: *user}
 
-	err := s.store.SetMsg(ctx, model.PartitionKey, userKey, model.ProtoFromUser(&userWithID))
+	err := s.store.SetMsgIf(ctx, model.PartitionKey, userKey, model.ProtoFromUser(&userWithID), nil)
 	if err != nil {
+		if errors.Is(err, kv.ErrPredicateFailed) {
+			err = ErrAlreadyExists
+		}
 		return "", fmt.Errorf("save user (usesrKey %s): %w", userKey, err)
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -219,69 +203,82 @@ func (s *KVAuthService) SecretStore() crypt.SecretStore {
...
 		group := entry.Value.(*model.GroupData)
 		if err = s.RemoveUserFromGroup(ctx, username, group.DisplayName); err != nil {
 			return fmt.Errorf("%s: %w", username, err)
 		}
 	}
-	if err := itr.Err(); err != nil {
+	if err = itr.Err(); err != nil {
 		return err
 	}
+	defer itr.Close()
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -125,340 +129,366 @@ func fieldByTag(t reflect.Type, key, tag string) string {
...
-	if err != nil {
-		return nil, err
-	}
-	return user.(*model.User), nil
+func (s *KVAuthService) GetUserByEmail(ctx context.Context, email string) (*model.User, error) {
+	return s.cache.GetUserByID(email, func() (*model.User, error) {
+		m := &model.UserData{}
+		itr, err := s.store.Scan(ctx, m.ProtoReflect().Type(), model.PartitionKey, model.KVUserPath(""))
+		if err != nil {
+			return nil, fmt.Errorf("sacn users: %w", err)
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -290,35 +287,37 @@ func (s *KVAuthService) GetUser(ctx context.Context, username string) (*model.Us
...
 }
 
 func (s *KVAuthService) GetUserByID(ctx context.Context, userID string) (*model.User, error) {
 	return s.cache.GetUserByID(userID, func() (*model.User, error) {
 		m := &model.UserData{}
-		itr, err := s.store.Scan(ctx, m.ProtoReflect().Type(), model.PartitionKey, model.KvUserPath(""))
+		itr, err := s.store.Scan(ctx, m.ProtoReflect().Type(), model.PartitionKey, model.KVUserPath(""))
 		if err != nil {
-			return nil, fmt.Errorf("sacn users users: %w", err)
+			return nil, fmt.Errorf("sacn users: %w", err)
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -406,55 +408,84 @@ func (s *KVAuthService) ListUserPolicies(ctx context.Context, username string, p
...
+		for _, policy := range policies {
+			policiesSet[policy.DisplayName] = policy
+			policiesArr = append(policiesArr, policy.DisplayName)
+		}
+		after = userPaginator.NextPageToken
+		hasMoreUserPolicy = userPaginator.NextPageToken != ""
+	}
+
+	hasMoreGroup := true
+	hasMoreGroupPolicy := true
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
eden-ohana: done

pkg/auth/service.go
```diff
@@ -474,456 +539,479 @@ func (s *DBAuthService) ListEffectivePolicies(ctx context.Context, username stri
...
-			ORDER BY auth_groups.display_name
-			LIMIT $4`,
-			groupDisplayName, params.After, fmt.Sprint(params.Prefix, "%"), params.Amount+1)
+	if params.Amount < 0 || params.Amount > maxPage {
+		params.Amount = maxPage
+	}
+
+	hasMoreGroups := true
+	var userGroups []*model.Group
+	var resPaginator *model.Paginator
```
itai-david, line 694: Uninitialized pointer. Leads to segmantation fault later, when accessed

itai-david (changes requested):

pkg/auth/service.go
```diff
@@ -474,456 +539,479 @@ func (s *DBAuthService) ListEffectivePolicies(ctx context.Context, username stri
...
-	slice, paginator, err := ListPaged(ctx, s.db, reflect.TypeOf(group), params, "display_name",
-		psql.Select("*").
-			From("auth_groups").
-			Where(sq.Like{"display_name": fmt.Sprint(params.Prefix, "%")}))
-	if err != nil {
+func (s *KVAuthService) ListGroups(ctx context.Context, params *model.PaginationParams) ([]*model.Group, *model.Paginator, error) {
+	var group model.GroupData
+	groupKey := model.KVGroupPath("")
+
+	msgs, paginator, err := s.ListKVPaged(ctx, (&group).ProtoReflect().Type(), params, groupKey, false)
```
itai-david, line 640: ListKVPaged uses `PrefixIterator` with `groupKey` as a prefix.
The value assigned to `groupKey` is `groups/` which is also the prefix to all user membership in group entries - `groups/<DisplayName>/users/<Username>` - and all group <--> policy relations - `groups/<Displayname>/policies/<DisplayName>`
The result slice contains a lot more than only groups
One option is to change the keys schema so it would be easy to distinguish between the entities (i.e. keys that cannot be prefixes of each other).
Another (ugly) option is to filter out irrelevant result as you iterate - not sure how trivial it is.

itai-david (changes requested):

pkg/auth/service.go
itai-david: I believe the same goes to all similar cases, e.g. `ListUsers` and the correlated user <--> relations

pkg/auth/model/model.go
```diff
@@ -308,15 +292,15 @@ func protoFromStatements(s *Statements) []*StatementData {
...
 }
 
 func KVPolicyToUser(userName string, policyDisplayName string) string {
-	return kv.FormatPath(authPrefix, "users", userName, "policies", policyDisplayName)
+	return kv.FormatPath("users", userName, "policies", policyDisplayName)
 }
 
 func KVPolicyToGroup(groupDisplayName string, policyDisplayName string) string {
-	return kv.FormatPath(authPrefix, "groups", groupDisplayName, "policies", policyDisplayName)
+	return kv.FormatPath("groups", groupDisplayName, "policies", policyDisplayName)
```
N-o-Z, line 303: 1. Use consts for path elements
2. Paths to different objects / secondary prefixes should not have a common subpath
3. Please put all Path functions together and use the same naming convention e.g.: "PolicyGroupPath"
4. Remove KV prefix from function names - it's redundant

pkg/auth/service.go
```diff
@@ -125,340 +129,401 @@ func fieldByTag(t reflect.Type, key, tag string) string {
...
+	encrypted, err := s.Encrypt([]byte(secretAccessKey))
 	if err != nil {
 		return nil, err
 	}
 	return encrypted, nil
 }
 
-func (s *DBAuthService) SecretStore() crypt.SecretStore {
-	return s.secretStore
+func (s *KVAuthService) ConvertCredDataList(creds []proto.Message) []*model.Credential {
```
N-o-Z, line 208: In my opinion, this and CredentialFromProto should be in model.go and be package functions

pkg/auth/service.go
eden-ohana: done

pkg/auth/service.go
eden-ohana: done

pkg/auth/model/model.go
eden-ohana: done

cmd/lakefs/cmd/setup.go
```diff
@@ -50,7 +52,19 @@ var setupCmd = &cobra.Command{
...
-		authService := auth.NewDBAuthService(dbPool, crypt.NewSecretStore(cfg.GetAuthEncryptionSecret()), cfg.GetAuthCacheConfig(), logging.Default().WithField("service", "auth_service"))
+		var authService auth.Service
+		if dbParams.KVEnabled {
+			kvStore, err := kv.Open(ctx, dbParams.Type, dbParams.ConnectionString)
+			if err != nil {
+				fmt.Printf("failed to open KV store: %s\n", err)
+				os.Exit(1)
+			}
+			storeMessage := kv.StoreMessage{Store: kvStore}
+			authService = auth.NewKVAuthService(storeMessage, crypt.NewSecretStore(cfg.GetAuthEncryptionSecret()), cfg.GetAuthCacheConfig(), logging.Default().WithField("service", "auth_service"))
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -219,69 +203,82 @@ func (s *KVAuthService) SecretStore() crypt.SecretStore {
...
 func (s *KVAuthService) GetUser(ctx context.Context, username string) (*model.User, error) {
 	return s.cache.GetUser(username, func() (*model.User, error) {
-		userKey := model.KvUserPath(username)
+		userKey := model.KVUserPath(username)
 		m := model.UserData{}
 		_, err := s.store.GetMsg(ctx, model.PartitionKey, userKey, &m)
 		if err != nil {
 			if errors.Is(err, kv.ErrNotFound) {
-				err = fmt.Errorf("%s: %w", err, ErrNotFound) // Wrap error for compatibility with DBService
+				err = ErrNotFound // Wrap error for compatibility with DBService
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -219,69 +203,82 @@ func (s *KVAuthService) SecretStore() crypt.SecretStore {
...
+	groupKey := model.KVGroupPath("")
 	itr, err := kv.NewPrimaryIterator(ctx, s.store.Store, (&model.GroupData{}).ProtoReflect().Type(), model.PartitionKey, groupKey, "")
+	if err != nil {
+		return err
+	}
+
 	for itr.Next() {
 		entry := itr.Entry()
 		group := entry.Value.(*model.GroupData)
 		if err = s.RemoveUserFromGroup(ctx, username, group.DisplayName); err != nil {
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -406,55 +408,84 @@ func (s *KVAuthService) ListUserPolicies(ctx context.Context, username string, p
...
-	//}
-	keys := make([]*model.BasePolicy, len(set))
-	i := 0
-	for k := range set {
-		keys[i] = k
-		i++
-	}
-	return keys, nil, nil
+	if _, err := s.GetUser(ctx, username); err != nil {
+		return nil, nil, fmt.Errorf("%s: %w", username, err)
```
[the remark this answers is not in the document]
eden-ohana: done, it was ok because the get user doesn't wrap the error, but I changed it so it will be consistent

pkg/auth/service.go
```diff
@@ -406,55 +408,84 @@ func (s *KVAuthService) ListUserPolicies(ctx context.Context, username string, p
...
+		policies, userPaginator, err := s.ListUserPolicies(ctx, username, &model.PaginationParams{
+			After:  after,
+			Amount: amount,
+		})
+		if err != nil {
+			return nil, nil, fmt.Errorf("list user policies: %w", err)
+		}
+		for _, policy := range policies {
+			policiesSet[policy.DisplayName] = policy
+			policiesArr = append(policiesArr, policy.DisplayName)
```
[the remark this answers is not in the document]
eden-ohana: Deleted. It isn't needed. To sort a map in go you first need to iterate over the map and create an array from all the keys. I just wanted to save another iteration over the map and do it while building the set.

pkg/auth/service.go
```diff
@@ -406,55 +408,84 @@ func (s *KVAuthService) ListUserPolicies(ctx context.Context, username string, p
...
+		after = userPaginator.NextPageToken
+		hasMoreUserPolicy = userPaginator.NextPageToken != ""
+	}
+
+	hasMoreGroup := true
+	hasMoreGroupPolicy := true
+	after = ""
+	for hasMoreGroup {
+		// get membership groups to user
+		groups, groupPaginator, err := s.ListGroups(ctx, &model.PaginationParams{
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
eden-ohana: return false err

https://github.com/treeverse/lakeFS/blob/2024646a5460f40a337593bf34784c51e6400893/pkg/auth/service.go#L1047

pkg/auth/service.go
eden-ohana: done

pkg/auth/service.go
eden-ohana: Casting error, I changed the description.

pkg/auth/service.go
eden-ohana: done

pkg/auth/service.go
```diff
@@ -706,8 +771,12 @@ func (s *KVAuthService) DeletePolicy(ctx context.Context, policyDisplayName stri
...
-	groupKey := model.KvGroupPath("")
+	groupKey := model.KVGroupPath("")
 	it, err = kv.NewPrimaryIterator(ctx, s.store.Store, (&model.UserData{}).ProtoReflect().Type(), model.PartitionKey, groupKey, "")
+	if err != nil {
+		return err
+	}
+	defer it.Close()
 	for it.Next() {
 		entry := it.Entry()
 		group := entry.Value.(*model.GroupData)
```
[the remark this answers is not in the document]
eden-ohana: Do you mean the get-call? I extracted it to another function without the get to not duplicate code with detach policy

pkg/auth/service.go
```diff
@@ -692,11 +753,15 @@ func (s *KVAuthService) DeletePolicy(ctx context.Context, policyDisplayName stri
...
-	usersKey := model.KvUserPath("")
+	usersKey := model.KVUserPath("")
 	it, err := kv.NewPrimaryIterator(ctx, s.store.Store, (&model.UserData{}).ProtoReflect().Type(), model.PartitionKey, usersKey, "")
+	if err != nil {
+		return err
+	}
+	defer it.Close()
 	for it.Next() {
 		entry := it.Entry()
 		user := entry.Value.(*model.UserData)
```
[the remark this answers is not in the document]
eden-ohana: done, deleted from help-function without another call.

pkg/auth/service.go
```diff
@@ -406,55 +408,84 @@ func (s *KVAuthService) ListUserPolicies(ctx context.Context, username string, p
...
+		groups, groupPaginator, err := s.ListGroups(ctx, &model.PaginationParams{
+			After:  after,
+			Amount: amount,
+		})
+		if err != nil {
+			return nil, nil, err
+		}
+		for _, group := range groups {
+			// get policies attracted to group
+			for hasMoreGroupPolicy {
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -406,55 +408,84 @@ func (s *KVAuthService) ListUserPolicies(ctx context.Context, username string, p
...
+				groupPolicies, groupPoliciesPaginator, err := s.ListGroupPolicies(ctx, group.DisplayName, &model.PaginationParams{
+					After:  after,
+					Amount: amount,
+				})
+				if err != nil {
+					return nil, nil, fmt.Errorf("list group policies: %w", err)
+				}
+				for _, policy := range groupPolicies {
+					policiesSet[policy.DisplayName] = policy
+					policiesArr = append(policiesArr, policy.DisplayName)
```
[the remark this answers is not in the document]
eden-ohana: deleted

pkg/auth/service.go
```diff
@@ -406,55 +408,84 @@ func (s *KVAuthService) ListUserPolicies(ctx context.Context, username string, p
...
+				}
+				after = groupPoliciesPaginator.NextPageToken
+				hasMoreGroupPolicy = groupPoliciesPaginator.NextPageToken != ""
+			}
+		}
+		after = groupPaginator.NextPageToken
+		hasMoreGroup = groupPaginator.NextPageToken != ""
+	}
+
+	if params.Amount < 0 || params.Amount > maxPage {
```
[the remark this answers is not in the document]
eden-ohana: params.Amout==0  return zero amount
params.Amout < 0 return all (maxPage) - the testing convention
params.Amout > maxPage return the max (maxPage)

pkg/auth/service.go
```diff
@@ -601,48 +650,46 @@ func (s *KVAuthService) ListUserGroups(ctx context.Context, username string, par
 	if _, err := s.GetUser(ctx, username); err != nil {
 		return nil, nil, fmt.Errorf("%s: %w", username, err)
 	}
+	if params.Amount < 0 || params.Amount > maxPage {
```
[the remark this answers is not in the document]
eden-ohana: returns 0 (no need to limit)

pkg/auth/service.go
```diff
@@ -601,48 +650,46 @@ func (s *KVAuthService) ListUserGroups(ctx context.Context, username string, par
...
+	var resPaginator *model.Paginator
+	for hasMoreGroups && len(userGroups) <= params.Amount {
+		groups, paginator, err := s.ListGroups(ctx, &model.PaginationParams{Prefix: params.Prefix, After: params.After, Amount: maxPage})
 		if err != nil {
 			return nil, nil, err
 		}
 		for _, group := range groups {
-			path := model.KvUserToGroup(username, group.DisplayName)
+			if len(userGroups) == params.Amount {
+				resPaginator.NextPageToken = group.DisplayName
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -489,52 +520,64 @@ func (s *KVAuthService) CreateGroup(ctx context.Context, group *model.BaseGroup)
...
+			err = ErrAlreadyExists
+		}
 		return fmt.Errorf("save group (groupKey %s): %w", groupKey, err)
 	}
 	return err
 }
 
 func (s *KVAuthService) DeleteGroup(ctx context.Context, groupDisplayName string) error {
-	if _, err := s.GetGroup(ctx, groupDisplayName); err != nil {
+	var err error
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -381,25 +381,34 @@ func (s *KVAuthService) AttachPolicyToUser(ctx context.Context, policyDisplayNam
...
-		return fmt.Errorf("%s: %w", username, err)
+		return err
 	}
 	if _, err := s.GetPolicy(ctx, policyDisplayName); err != nil {
-		return fmt.Errorf("detaching policy: %w", err)
+		return err
 	}
 
-	pu := model.KVPolicyToUser(username, policyDisplayName)
+	pu := model.UserPolicyPath(username, policyDisplayName)
```
itaiad200, line 401:
```suggestion
	return s.DetachPolicyFromUserNoValidation(ctx, policyDisplayName, username)
```

pkg/auth/service.go
```diff
@@ -489,52 +520,64 @@ func (s *KVAuthService) CreateGroup(ctx context.Context, group *model.BaseGroup)
...
+		return err
+	}
+	defer it.Close()
 	for it.Next() {
 		entry := it.Entry()
 		user := entry.Value.(*model.UserData)
 		if err = s.RemoveUserFromGroup(ctx, user.Username, groupDisplayName); err != nil {
 			return fmt.Errorf("%s: %w", user.Username, err)
 		}
 	}
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -489,52 +520,64 @@ func (s *KVAuthService) CreateGroup(ctx context.Context, group *model.BaseGroup)
...
 		entry := itr.Entry()
 		policy := entry.Value.(*model.PolicyData)
 		if err = s.DetachPolicyFromGroup(ctx, policy.DisplayName, groupDisplayName); err != nil {
 			return fmt.Errorf("%s: %w", policy.DisplayName, err)
 		}
 	}
-	if err := itr.Err(); err != nil {
+	if err = itr.Err(); err != nil {
 		return err
 	}
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
eden-ohana: done

pkg/auth/service.go
```diff
@@ -406,55 +408,84 @@ func (s *KVAuthService) ListUserPolicies(ctx context.Context, username string, p
...
+				resPaginator.Amount = len(resPolicies)
+				return resPolicies, resPaginator, nil
+			}
+			resPolicies = append(resPolicies, policiesSet[p])
+		}
+	}
+	return resPolicies, resPaginator, nil
 }
 
 func (s *KVAuthService) ListEffectivePolicies(ctx context.Context, username string, params *model.PaginationParams) ([]*model.BasePolicy, *model.Paginator, error) {
```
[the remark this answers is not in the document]
eden-ohana: done

cmd/lakefs/cmd/superuser.go
eden-ohana: we decided not to do it

pkg/auth/service.go
itai-david: Great. Thanks

pkg/auth/service.go
itai-david: :+1:

itaiad200 (approved):
It's been an eventful & fruitful PR under short timelines and with high author & reviewers standards. I see that all of my blocking comments resolved. From my side, if the tests are happy, anything else can be a technical debt (with an attached issue). 

LGTM 👯 🕺 💃

pkg/auth/model/model.go
```diff
@@ -7,16 +7,62 @@ import (
...
 )
 
 const (
-	StatementEffectAllow = "allow"
-	StatementEffectDeny  = "deny"
+	StatementEffectAllow   = "allow"
+	StatementEffectDeny    = "deny"
+	PartitionKey           = "auth"
+	groupsPrefix           = "groups"
+	groupsUsersPrefix      = "groups_users"
```
N-o-Z, line 24: This is great, but I think we still have a problem with prefix scanning.
Maybe change to "gPolicies, "gUsers", "uPolicies" "uCredentials"

pkg/auth/service.go
```diff
@@ -471,478 +518,494 @@ func (s *DBAuthService) ListEffectivePolicies(ctx context.Context, username stri
...
+	// delete policy attachment to group
+	groupKey := model.GroupPath("")
+	it, err = kv.NewPrimaryIterator(ctx, s.store.Store, (&model.UserData{}).ProtoReflect().Type(), model.PartitionKey, groupKey, "")
+	if err != nil {
+		return err
+	}
+	defer it.Close()
+	for it.Next() {
+		entry := it.Entry()
+		group := entry.Value.(*model.GroupData)
```
itai-david, line 801: it initialized with type `UserData` and attempted to be accessed as `GroupData`. This wiil crash

itai-david (approved):
Thanks for fixing the issues. I added another comment on another issue in `KVAuthService.DeletePolicy`.
Approving as not to block

pkg/auth/model/model.go
eden-ohana: done

pkg/auth/service.go
eden-ohana: done

pkg/auth/service.go
```diff
@@ -144,31 +143,25 @@ func (s *KVAuthService) ListKVPaged(ctx context.Context, protoType protoreflect.
...
-	entries := make([]proto.Message, 0)
-	var lastKey string
-	for it.Next() && count < amount {
+	var entries []proto.Message
+	p := &model.Paginator{}
+	for len(entries) <= amount && it.Next() {
 		entry := it.Entry()
 		value := entry.Value
+		if len(entries) == amount {
+			p.NextPageToken = entry.Key
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -406,55 +408,84 @@ func (s *KVAuthService) ListUserPolicies(ctx context.Context, username string, p
...
+	if params.Amount < 0 || params.Amount > maxPage {
+		params.Amount = maxPage
+	}
+	sort.Strings(policiesArr)
+	var resPolicies []*model.BasePolicy
+	var resPaginator *model.Paginator
+	for _, p := range policiesArr {
+		if p > after {
+			if len(resPolicies) == params.Amount {
+				resPaginator.NextPageToken = p
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/auth/service.go
```diff
@@ -406,55 +408,84 @@ func (s *KVAuthService) ListUserPolicies(ctx context.Context, username string, p
...
+	}
+
+	if params.Amount < 0 || params.Amount > maxPage {
+		params.Amount = maxPage
+	}
+	sort.Strings(policiesArr)
+	var resPolicies []*model.BasePolicy
+	var resPaginator *model.Paginator
+	for _, p := range policiesArr {
+		if p > after {
```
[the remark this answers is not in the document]
eden-ohana: done

N-o-Z (approved):
Looks Good

PR #3496, opened 2022-06-12
KV Improvement: Remove base prefix from packages

### LINKED ISSUE #3458: Remove base prefixes from packages
Since we are using partition keys - the package base prefix is redundant

N-o-Z:
Closes #3458

### DIFF
--- a/pkg/actions/service.go
+++ b/pkg/actions/service.go
@@ -20,14 +20,13 @@ import (
 )
 
 const (
-	packageName   = "actions"
-	actionsPrefix = "actions"
-	PartitionKey  = "actions"
-	reposPrefix   = "repos"
-	RunsPrefix    = "runs"
-	tasksPrefix   = "tasks"
-	branchPrefix  = "branches"
-	commitPrefix  = "commits"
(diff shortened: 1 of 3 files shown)


itaiad200:
I wonder what happens in postgres where we have a constant number of partitions: `Partition = hash(partition_key) % 100`.
What happens if you try listing a partition, let's say by doing `Scan("multiparts", "")`, will the entire has partition returned, or just the items under `multiparts`. I couldn't find a concrete answer, need to run a small test to check it.

N-o-Z:
> I wonder what happens in postgres where we have a constant number of partitions: `Partition = hash(partition_key) % 100`. What happens if you try listing a partition, let's say by doing `Scan("multiparts", "")`, will the entire has partition returned, or just the items under `multiparts`. I couldn't find a concrete answer, need to run a small test to check it.

So I did a little check and it seems that when scanning a partition key, just the items under the given partition_key are returned and not the entire partition - so I guess we are good

itaiad200 (approved):
Thanks for checking the postgres implementation.

Issue #3411, opened 2022-05-22
Repository Template: lakeFS + Spark + Metastore


johnnyaug:
### User story

Data engineers new to lakeFS want to try out one of the main use cases (dev env, parallel pipelines, cross-collection consistency, etc) for their existing Spark-based stack.

### Pain

To get started using a new lakeFS installation with Spark, the following manual steps are typically required:

1. Create a repository and configure it with a storage namespace
2. Bring in existing data (whether using import, copy, ingest, upload(?), etc)
3. Connect Spark using the installation's S3 Gateway *or* configure Spark to use lakeFSFS with this installation
4. Understand how to use lakeFS structured URIs in the existing code base *or more likely* configure metastore for this indirection to have existing code run on a separate branch/commit (i.e. _I want `SELECT * FROM events;` to use the events table that exists in my `experiment-oz-temp` branch, not main or production!_)
5. Now, integrate Hive Metastore/Glue Data Catalog with lakeFS to be able to have a schema for my branch that contains all the same tables that I have in prod.

### Solution

Create a repository template that walks the user through these steps, tailored to their installation, automating as much of the process as possible.

### Repository Templates

When a user chooses to create a repository from the `Spark Quickstart` template, they get a wizard that guides them through configuring the repository, importing existing data and setting up Spark and HMS.

This will include:

1. New "empty state" for repository list screen (see dev#147)
2. New drop down button next to "Create repository" button to also show menu of templates (see dev#147).
3. The template's wizard (see below)

### lakeFS + Spark

When the user chooses this template, show them a wizard, with the steps described in [this Figma](https://www.figma.com/file/haD9y599LzW6LvsYBI2xWU/Spark-use-case?node-id=21%3A1489).

Once the user finishes the wizard and configures their Spark environment, they are redirected to a new repository with a README.md containing a summary of the things we did (or skipped..) from the wizard

1. importing data into this repository
2. connecting a spark setup to this environment

For the first deliverable, metastore commands are not automated via hooks. Instead, the README will also include:

3. instructions on which `lakectl metastore` commands to run and when.

### Considerations

* Optimize for **usability, simplicity and clarity**! One effective sentence in the README file is worth a thousand cool automations.
* Wizard steps and components should be modular and composable: we may need to reuse them in future repository types and probably also as standalone features. For example, importing data will probably be a step in multiple wizards, and is also a feature on its own.

Jonathan-Rosenberg:
[...]

PR #3499, opened 2022-06-13
Wizard MVP proposal

Jonathan-Rosenberg:
This is a proposal for the time to value wizard MVP.
For further info please refer to #3411, #3373, and the [Figma design](https://www.figma.com/file/haD9y599LzW6LvsYBI2xWU/Spark-use-case?node-id=31%3A200).


design/open/ttv-wizard-mvp.md
```diff
@@ -0,0 +1,271 @@
...
+    }
+    ```
+8. The wizard will present each configuration in a different snippet view for the users to copy and paste into their configuration files.
+9. The wizard will generate a README file as follows:
+    ```markdown
+    1. Created a repository "spark-repo" and branch "main".
+    2. Imported data from <S3 location>.
+    3. Generated the following configurations:
+       <Spark configurations with [“<secret1>”, “<secret2>”...] hidden>
+    4. Instructions to configure Hive Metastore with lakeFS.
```
johnnyaug, line 115: And, which `lakectl metastore` commands to run when creating/deleting/merging a branch/tag.

johnnyaug (approved):
Boom

talSofer (commented):
Thanks! 
Can you please add a section that talks about where the templates are stored? 

(forgot to hit submit yesterday...)

design/open/ttv-wizard-mvp.md
Jonathan-Rosenberg: Point one will generate configurations that are exactly the same as the Spark configurations generated...

design/open/ttv-wizard-mvp.md
```diff
@@ -0,0 +1,271 @@
...
+## System Overview
+
+![RouterFS with lakeFS URI](diagrams/Wizard.png)
+[(excalidraw file)](diagrams/wizard.excalidraw)
+
+### Wizard UI Component
+
+The wizard UI component is responsible for the user’s Spark onboarding process. The process is as follows:
+1. Create a repository (named as the user wishes) and a ‘main’ branch in it.
+2. Import the user’s data to ‘main’ and display a progress bar (which will show a link to required permissions). Only available on cloud-deployed lakeFS.
```
arielshaqed, line 24: Why only cloud-deployed?  I suspect it will only add work to enforce this

design/open/ttv-wizard-mvp.md
```diff
@@ -0,0 +1,271 @@
...
+   1. Asks the user to enter the lakeFS endpoint (by default it will be https://`location.host`)
+   2. Asks the user for the runtime environment of Spark: one of [Databricks, AWS EMR, spark-submit] ([Figma design](https://www.figma.com/file/haD9y599LzW6LvsYBI2xWU/Spark-use-case?node-id=31%3A200))
+4. Create a README file with all actions performed and commit it as a first commit to the ‘main’ branch of the created repository. If the Spark configuration step was performed, additional Metastore guidelines will be provided in the README.
+
+### Templating Service
+
+The [templating service](https://github.com/treeverse/lakeFS/pull/3373) is responsible for fetching, authenticating, and expanding the required templates and returning them to the client.  
+**Process**:  
+1. Get the S3 templates (locations should be specified in the incoming request). The files must be valid [`html/template`](https://pkg.go.dev/html/template) or [`text/template`](https://pkg.go.dev/text/template) parsable template texts.
```
arielshaqed, line 34: Not sure incoming request should entirely specify the location.  Would be happier if a set prefix were configured, and the incoming request could only select below that prefix.  Otherwise admins will have a very tough time ensuring the access is safe.

The proposed template service never allows access by arbitrary path precisely for this reason.  It protects access to templates on lakefs in some way; if you want to do the same but on S3, we will need to design similar protections for files from S3.

design/open/ttv-wizard-mvp.md
```diff
@@ -0,0 +1,271 @@
...
+        "value": "<service name>"
+    }
+    ```
+3. Templating service - status 500 - no access to provided template location
+    ```json
+    {
+        "class": "templating_service",
+        "name": "no_access",
+        "value": "<service name>"
+    }
```
arielshaqed, line 254: Don't we send status on our monitoring outputs as field values, rather than as metric names?

design/open/ttv-wizard-mvp.md
```diff
@@ -0,0 +1,271 @@
...
+    ```
+5. Templating service - status 4xx - general
+    ```json
+    {
+        "class": "templating_service",
+        "name": "4xx",
+        "value": "<service name>"
+    }
+    ```
```
arielshaqed, line 271: Don't the existing metrics on web endpoints already cover this?

design/open/ttv-wizard-mvp.md
```diff
@@ -0,0 +1,271 @@
...
+- The generated README will not include the import step.
+
+### Sad flow - No credentials' generation permissions
+
+1. The user clicks on ‘Create Repository’, then clicks on ‘Spark Quickstart’
+2. The wizard starts by showing an input textbox for the users to enter the repo name. The user types ‘spark-repo’
+3. The wizard creates a repo named ‘spark-repo’ and sets a default ‘main’ branch.
+4. The wizard asks the users if they want to import existing data to lakeFS. The user skips this step using the skip button.
+5. The wizard asks the user: “How are you using Spark”, and offers three alternatives: Databricks, AWS EMR, and spark-submit. The user chooses ‘Databricks’. It also asks the user for their lakeFS endpoint (and will show a default placeholder pointing to the current URL). The user types the endpoint.
+6. The wizard sends a GET request to the templating service with a query string of the format:
```
arielshaqed, line 186: Optionally, we might use the templating service as the source for the "how are you using this" data.  It's some JSON, that needs to be generated somewhere, and may have some fields to specify.  Looks like another use-case for templates.

arielshaqed (approved):
NEAT!

How do we render the README?  I assume this requires an additional feature.

design/open/ttv-wizard-mvp.md
```diff
@@ -0,0 +1,330 @@
...
+   -> All three templates will be requested (each in a separate request).
+   1. Asks the user to enter the lakeFS endpoint (by default it will be https://`location.host`)
+   2. [Figma Design](https://www.figma.com/file/haD9y599LzW6LvsYBI2xWU/Spark-use-case?node-id=31%3A200)
+4. Generates a README file with all actions performed, and will commit it as a first commit (after the import) to the ‘main’ branch of the created repository.
+
+### Templating Service
+
+The [templating service](https://github.com/treeverse/lakeFS/pull/3373) is responsible for fetching, authenticating, and expanding the required templates and returning them to the client.  
+**Process**:  
+1. Get the template (the location should be specified in the incoming request). The file must be a valid [`html/template`](https://pkg.go.dev/html/template) or [`text/template`](https://pkg.go.dev/text/template) parsable template text.
```
arielshaqed, line 35: How do we distinguish the two?  I suggest the extension before `.tt`, say `PATH.something.tt` generates `PATH.something` and `PATH.html.tt` generates `PATH.html`.

design/open/ttv-wizard-mvp.md
```diff
@@ -0,0 +1,330 @@
...
+</configuration>
+```
+
+### Happy flow - All steps
+
+1. The user clicks on ‘Create Repository’, then clicks on ‘Spark Quickstart’
+2. The wizard starts by showing an input textbox for the users to type the repo name. The user types ‘spark-repo’
+3. The wizard creates a repo named ‘spark-repo’ and sets a default ‘main’ branch.
+4. The wizard asks the users if they want to import existing data to lakeFS. The user specifies the location of the bucket (after they validated that the lakeFS role has the right permissions and that the bucket has the correct policy), and clicks ‘OK’.
+   * An object counter will show the progress of the import process and will signal once it’s over.
```
arielshaqed, line 124: I would really like this to link to the import design.  Or, if we need to change import, let's call it out and remain with one *good* implementation of *the same* feature.

Here, if a counter makes sense during the wizard, it also makes sense during a plain import.

design/open/ttv-wizard-mvp.md
```diff
@@ -0,0 +1,330 @@
...
+2. Use the configured template functions to validate the user’s permissions to perform the required actions, and to generate credentials on the fly.
+3. Expand the template with the config file and query string params, and return it with the correct `Content-Type` header (inferred from the template).
+
+### Wizard Templates
+
+The following templates will be saved within the lakeFS binary and be directed by the Wizard component to be expanded:
+* SparkEMR.conf.tt
+* SparkDatabricks.conf.tt
+* SparkXML.conf.tt
+* README.md.tt
```
arielshaqed, line 45: How is the generated `README.md` rendered?

design/open/ttv-wizard-mvp.md
```diff
@@ -0,0 +1,330 @@
...
+    {
+        "class": "templating_service",
+        "name": "4xx",
+        "value": "<service name>"
+    }
+    ```
+
+### BI Metrics
+
+Sent directly from the GUI Wizard
```
arielshaqed, line 297: I think that all of these should identify which wizard it was.  Initially we expect to have only one, but later we expect to have more!

Similarly, we might report which Spark configs were generated.

arielshaqed (commented):
Thanks!

Still open issues:

* Rendering the README
* Monitoring: can we use existing metrics rather than define new ones?

design/open/ttv-wizard-mvp.md
Jonathan-Rosenberg: Cool

Jonathan-Rosenberg:
@arielshaqed, regarding monitoring, if we can leverage existing metrics, we shall do that, if not we'll make amendments to get the knowledge we need. Anyway, this is an implementation detail.

design/open/ttv-wizard-mvp.md
Jonathan-Rosenberg: They identify it as `spark_wizard`, did you mean something else?

design/open/ttv-wizard-mvp.md
Jonathan-Rosenberg: We agreed that it will not be rendered at first.

design/open/ttv-wizard-mvp.md
Jonathan-Rosenberg: We don't need to change the import. We don't use it as it is, just leverage it's functionality, not the exact behavior.
It's not the same feature.
There is a counter in the plain import.

design/open/ttv-wizard-mvp.md
Jonathan-Rosenberg: possibly. I just listed what we **want**, not what we need to **implement**.

design/open/ttv-wizard-mvp.md
itaiad200: The UI-Import is an assembly of 5-6 different API calls. To get the behaviour you need here you'll need to replace 1-2 API calls, but it's all pretty straight forward.

design/open/ttv-wizard-mvp.md
```diff
@@ -0,0 +1,330 @@
...
+1. The user clicks on ‘Create Repository’, then clicks on ‘Spark Quickstart’
+2. The wizard starts by showing an input textbox for the users to type the repo name. The user types ‘spark-repo’
+3. The wizard creates a repo named ‘spark-repo’ and sets a default ‘main’ branch.
+4. The wizard asks the users if they want to import existing data to lakeFS. The user specifies the location of the bucket (after they validated that the lakeFS role has the right permissions and that the bucket has the correct policy), and clicks ‘OK’.
+   * An object counter will show the progress of the import process and will signal once it’s over.
+5. The wizard asks the user for their lakeFS endpoint (and will show a default placeholder pointing to the current URL).
+6. The wizard will send a GET request to the templating service with a query string of the format:  
+    ```
+    ?lakefs_url=https://my-lakefs.io&template_location=databricksConfig.props.tt
+    ```
```
itaiad200, line 129: Why can't the service infer the lakeFS url from the request itself? The UI knows the address of the server, right?
For trickier installation, where the address depends on the caller flow (i.e. with/without VPC), supplying the address gets harder for the user itself.

design/open/ttv-wizard-mvp.md
```diff
@@ -0,0 +1,330 @@
...
+7. The templating service will fetch the templates from the provided locations, expand the `.querystring.lakefs_url` parameter for both templates and return a response of the format:
+    ```properties
+    spark.hadoop.fs.s3a.impl=org.apache.hadoop.fs.s3a.S3AFileSystem
+    spark.hadoop.fs.s3a.access_key=ACCESSKEYDONTTELL
+    spark.hadoop.fs.s3a.secret_key=SECRETKEYDONTTELL
```
itaiad200, line 134: This kinds of break our promise when we create credentials: `This is the only time where you can see your credentials - make sure to copy them......`. I don't think that we shouldn't do it, only that we should probably rephrase that sentence.

design/open/ttv-wizard-mvp.md
```diff
@@ -0,0 +1,330 @@
...
+1. Create a repository (named as the user wishes) and a ‘main’ branch in it.
+2. Import the user’s data to ‘main’ and display a progress bar (which will show a link to required permissions). Only available on cloud-deployed lakeFS.
+3. Generate Spark configurations: the wizard will return Spark configurations for the users to use in their Spark core-site.xml file, [databricks key-value format](https://docs.databricks.com/clusters/configure.html#spark-configuration), or [EMR JSON](https://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-configure-apps.html) ([core-site](https://docs.aws.amazon.com/emr/latest/ReleaseGuide/emr-configure-apps.html))
+   -> All three templates will be requested (each in a separate request).
+   1. Asks the user to enter the lakeFS endpoint (by default it will be https://`location.host`)
+   2. [Figma Design](https://www.figma.com/file/haD9y599LzW6LvsYBI2xWU/Spark-use-case?node-id=31%3A200)
+4. Generates a README file with all actions performed, and will commit it as a first commit (after the import) to the ‘main’ branch of the created repository.
+
+### Templating Service
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: a

design/open/ttv-wizard-mvp.md
Jonathan-Rosenberg: The intention was that the URL will be set by default by the Wizard GUI (inferred from the URL it's working at), and we'll let the users change it if they want to.

> (and will show a default placeholder pointing to the current URL).

design/open/ttv-wizard-mvp.md
Jonathan-Rosenberg: Good point.
Will add that

design/open/ttv-wizard-mvp.md
```diff
@@ -0,0 +1,330 @@
...
+    spark.hadoop.fs.s3a.path.style.access=true
+    ```
+   each returned template will return with a `Content-Type` header describing the way it should be presented.
+8. The wizard will present each configuration in a different snippet view for the users to copy and paste into their configuration files.
+9. The wizard will send a GET request to the templating service with a query string of the format:
+    ```
+    ?lakefs_url=https://my-lakefs.io&template_location=README.md.tt
+    ```
+   It will do so to generate a README.md file.
+10. The returned README file will describe the steps taken, the configurations generated but without secrets and some commands to explain how to connect Hive Metastore to lakeFS:
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: The Wizard's intention is to create a repo ready to use. Each run will create a different repo (we'll not allow overriding repos) thus a wizard can run once on each repo, making the README specify the steps taken while creating that repo.

PR #3502, opened 2022-06-13
Allow dynamodb cloud usage

### LINKED ISSUE #3442: DynamoDB KV Support - Integrate with Production DynamoDB

itaiad200:
closes #3442

### DIFF
--- a/pkg/kv/dynamodb/store.go
+++ b/pkg/kv/dynamodb/store.go
@@ -4,12 +4,14 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/dynamodb"
 	"github.com/aws/aws-sdk-go/service/dynamodb/dynamodbattribute"
+	"github.com/cenkalti/backoff/v4"
 	"github.com/treeverse/lakefs/pkg/kv"
 )
 
@@ -34,7 +36,7 @@ type EntriesIterator struct {
 type DynKVItem struct {
 	PartitionKey string
 	ItemKey      string
-	ItemValue    string
+	ItemValue    []byte
 }
 
 // Params struct holds all the configuration parameters that can be used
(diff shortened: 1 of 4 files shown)


N-o-Z (approved):
Overall look good. One comment though - you need to modify the predicate handling as well and change it to byte slice (blob)

pkg/kv/dynamodb/store.go
```diff
@@ -165,11 +173,33 @@ func setupKeyValueDatabase(ctx context.Context, svc *dynamodb.DynamoDB, params *
 		},
 	})
 	if err != nil {
-		if _, ok := err.(*dynamodb.ResourceInUseException); !ok {
-			return err
+		if _, ok := err.(*dynamodb.ResourceInUseException); ok {
```
N-o-Z, line 176: if err != nil and err is ResourceInUseException (table already created - as in the case of open twice) we should not wait for table ACTIVE state

N-o-Z (changes requested):

pkg/kv/dynamodb/store.go
```diff
@@ -192,8 +222,9 @@ func (s *Store) Get(ctx context.Context, partitionKey, key []byte) (*kv.ValueWit
 	}
 
 	result, err := s.svc.GetItemWithContext(ctx, &dynamodb.GetItemInput{
-		TableName: aws.String(s.params.TableName),
-		Key:       s.bytesKeyToDynamoKey(partitionKey, key),
+		TableName:      aws.String(s.params.TableName),
+		Key:            s.bytesKeyToDynamoKey(partitionKey, key),
+		ConsistentRead: aws.Bool(true),
```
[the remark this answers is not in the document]
itaiad200: Yap - there's no other option other for lakeFS than being consistent. lakeFS provides ACID guarantees and not being consistent while reading from any DB breaks that promises. For example: You committed by advancing the branch to the next staging token. Some read after that still reads the previous staging token after the commit already happened. Now you write a file under a stale staging token and it's forever gone.

pkg/kv/dynamodb/store.go
itaiad200: Which is what I believe I did, no?

PR #3503, opened 2022-06-14
Auth KV Migration - WIP

itai-david:
* Based on branch https://github.com/treeverse/lakeFS/tree/3344/kv-auth-implementation
* Export function for each relevant table in dv_service.go
* There is possible refactoring option of the export functions, at least to 2 common functions, but as each function has its nuances, it is not quite trivial. In addition, as this will make it harder to debug (IMO) I decided to leave it as is, at least for KV2 milestone
* Relevant migration test not included yet


pkg/auth/db_service.go
```diff
@@ -867,3 +882,304 @@ func (s *DBAuthService) markTokenSingleUse(ctx context.Context, tokenID string,
...
+		}
+		policyID2Name[dbPolicy.ID] = dbPolicy.DisplayName
+	}
+	return policyID2Name, nil
+}
+
+func exportUserGroups(ctx context.Context, d *pgxpool.Pool, je *json.Encoder, usersDetails UserIDToDetails, groupsNames IDToName) error {
+	type relation struct {
+		UserID  int64 `db:"user_id"`
+		GroupID int   `db:"group_id"`
```
itaiad200, line 1050: Are you sure this is the right type? Also - why not using the types directly from the db package (assuming they exist)?

pkg/auth/db_service.go
```diff
@@ -867,3 +882,304 @@ func (s *DBAuthService) markTokenSingleUse(ctx context.Context, tokenID string,
...
+	scanner := pgxscan.NewRowScanner(rows)
+	for rows.Next() {
+		up := relation{}
+		err = scanner.Scan(&up)
+		if err != nil {
+			return err
+		}
+		username := usersDetails[up.UserID].name
+		key := model.KVPolicyToUser(policiesNames[up.PolicyID], username)
+		secIndex := kv.SecondaryIndex{PrimaryKey: []byte(model.KVUserPath(username))}
```
itaiad200, line 1101: I think the key is of the form `users\<user>\policies\<policy>`, right?
So it should point to the policy and not the user then

pkg/auth/db_service.go
```diff
@@ -867,3 +882,304 @@ func (s *DBAuthService) markTokenSingleUse(ctx context.Context, tokenID string,
...
+	}
+	defer rows.Close()
+	scanner := pgxscan.NewRowScanner(rows)
+	for rows.Next() {
+		ug := relation{}
+		err = scanner.Scan(&ug)
+		if err != nil {
+			return err
+		}
+		username := usersDetails[ug.UserID].name
```
itaiad200, line 1064: Here and elsewhere: I think you should be careful when reading from all maps. Default values are not your friends and they will not return an error but spam the kv and probably end up in a non functional kv state.

itaiad200 (changes requested):

pkg/auth/db_service.go
itai-david: Right. Fixed

pkg/auth/db_service.go
itai-david: These are the types used in the corresponding DB objects (User and Group, in that case).
A specialised type for the relation itself does not exist, as the relation is never read as a standalone, only as part of a `join`

pkg/auth/db_service.go
```diff
@@ -4,22 +4,37 @@ package auth
...
 	"fmt"
+	"io"
 	"reflect"
+	"strconv"
 	"time"
 
+	"github.com/google/uuid"
+	"github.com/jackc/pgx/v4/pgxpool"
+	"google.golang.org/protobuf/proto"
+
```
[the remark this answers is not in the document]
itai-david: Removed

pkg/auth/db_service.go
```diff
@@ -867,3 +882,304 @@ func (s *DBAuthService) markTokenSingleUse(ctx context.Context, tokenID string,
...
+type IDToName map[int]string
+
+func exportUsers(ctx context.Context, d *pgxpool.Pool, je *json.Encoder) (UserIDToDetails, error) {
+	rows, err := d.Query(ctx, "SELECT * FROM auth_users ORDER BY created_at ASC")
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+	userID2Details := make(UserIDToDetails)
```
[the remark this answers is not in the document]
itai-david: No special reason to use it, I just like this annotation. Anyways, changed that

pkg/auth/db_service.go
itai-david: Excellent point. Fixed

pkg/auth/db_service.go
```diff
@@ -4,22 +4,47 @@ package auth
...
 	"github.com/treeverse/lakefs/pkg/logging"
+	"github.com/treeverse/lakefs/pkg/version"
 	"golang.org/x/crypto/bcrypt"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
+//nolint:gochecknoinits
+func init() {
+	tablesToDrop := []string{
```
N-o-Z, line 35: auth_installation_metadata?

pkg/auth/db_service.go
```diff
@@ -837,3 +868,372 @@ func (s *DBAuthService) markTokenSingleUse(ctx context.Context, tokenID string,
 	}
 	return canUseToken, nil
 }
+
+var ErrExportedEntNotFound = errors.New("previously exported entitiy not found")
+
+func Migrate(ctx context.Context, d *pgxpool.Pool, writer io.Writer) error {
```
N-o-Z, line 874: Missing installation metadata?

pkg/auth/db_service.go
```diff
@@ -837,3 +868,372 @@ func (s *DBAuthService) markTokenSingleUse(ctx context.Context, tokenID string,
...
+		if err = je.Encode(kv.Entry{
+			PartitionKey: []byte(model.PartitionKey),
+			Key:          []byte(key),
+			Value:        value,
+		}); err != nil {
+			return nil, err
+		}
+
+		userIDToDetails[dbUser.ID] = userDetails{
+			kvID: kvUser.ID,
```
N-o-Z, line 966: In the migration process - one of the things we need to do is to create a new UUID using a common source

pkg/auth/db_service.go
```diff
@@ -837,3 +868,372 @@ func (s *DBAuthService) markTokenSingleUse(ctx context.Context, tokenID string,
...
+	}
+
+	if err = exportExpiredTokens(ctx, d, je); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+type userDetails struct {
```
N-o-Z, line 924: Move type declarations to the top of the module

pkg/auth/db_service.go
N-o-Z: We need to decide if user is identified by its ID or name. If username is a unique identifier we need to remove support of user ID. Else, I think secondary key of user should rely on its ID and not its name

pkg/auth/kv_migrate_test.go
```diff
@@ -0,0 +1,271 @@
...
+
+	testutil.MustDo(t, "Import migrated auth data", kv.Import(ctx, &buf, kvStore))
+	kvAuthService := auth.NewKVAuthService(kv.StoreMessage{Store: kvStore}, crypt.NewSecretStore([]byte("someSecret")), authparams.ServiceCache{
+		Enabled: false,
+	}, logging.Default())
+
+	verifyMigrationResults(t, ctx, dbAuthService, kvAuthService)
+}
+
+const MaxCredsPerUser = 4
```
N-o-Z, line 45: consolidate const declarations and move to the top of the module

pkg/auth/kv_migrate_test.go
```diff
@@ -0,0 +1,271 @@
...
+const NumUsers = 20
+const NumGroups = 4
+
+func prepareTestData(t *testing.T, ctx context.Context, svc auth.Service) {
```
N-o-Z, line 49: installation_metadata?

N-o-Z (changes requested):
Thanks @itai-david, Overall - looks great, I added some preliminary comments. Will continue tomorrow morning

pkg/auth/db_service.go
itai-david: Currently `MetadataManager` over `KV` is not implemented yet, so migrating it will be meaningless and cannot be tested and verified.
I think we should exclude this from KV2 and open a different issue for that - WDYT?
@itaiad200 - What's your take on that?

pkg/auth/db_service.go
```diff
@@ -837,3 +868,372 @@ func (s *DBAuthService) markTokenSingleUse(ctx context.Context, tokenID string,
 	}
 	return canUseToken, nil
 }
+
+var ErrExportedEntNotFound = errors.New("previously exported entitiy not found")
```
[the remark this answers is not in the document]
itai-david: Done

pkg/auth/db_service.go
itai-david: Same as above - pending `MetadataManager` support of `KV`

pkg/auth/db_service.go
itai-david: Moreover - though we can already migrate the table to KV (and have it unused at the moment) dropping it at the end of a successful migration will turn the system unusable

pkg/auth/db_service.go
N-o-Z: Thanks, I just noticed it is standalone and not part of the auth service itself

pkg/auth/kv_migrate_test.go
itai-david: Pendind `MetadataManager` support of `KV`

pkg/auth/db_service.go
N-o-Z: I just noticed it's not related to the auth service so disregarded all comments about it. We do need to add it to our plan (I think we missed it)

pkg/auth/db_service.go
```diff
@@ -308,10 +318,16 @@ func (s *DBAuthService) ListUserPolicies(ctx context.Context, username string, p
...
 	if slice == nil {
 		return nil, paginator, err
 	}
-	return slice.Interface().([]*model.BasePolicy), paginator, nil
+	policies := slice.Interface().([]*model.DBPolicy)
+	res := make([]*model.BasePolicy, 0, len(policies))
+	for _, p := range policies {
+		res = append(res, &p.BasePolicy)
+	}
+	return res, paginator, nil
```
itaiad200, line 341: How is this related?

pkg/auth/db_service.go
itai-david: Done

pkg/auth/kv_migrate_test.go
itai-david: Done

itaiad200 (approved):
LGTM. Thanks for applying all the fixes! 
Please open a ticket to add an esti migration test for the auth package (although the other tests sort of cover it by design, but still there is probably some areas to cover with a dedicated auth system test)

pkg/auth/db_service.go
```diff
@@ -837,3 +868,372 @@ func (s *DBAuthService) markTokenSingleUse(ctx context.Context, tokenID string,
...
+	userIDToDetails := make(UserIDToDetails)
+	scanner := pgxscan.NewRowScanner(rows)
+	for rows.Next() {
+		dbUser := model.DBUser{}
+		err = scanner.Scan(&dbUser)
+		if err != nil {
+			return nil, err
+		}
+		kvUser := &model.User{
+			ID:       uuid.New().String(),
```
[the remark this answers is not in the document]
itai-david: We can go with `xid`, to gain sort capabilities. However , this also involves the `KVAuthService`, as it uses uuid to generate IDs for newly created entities.
I believe that at this point of KV2, we should leave it as is and open a separate issue for that - WDYT?

pkg/auth/db_service.go
itai-david: Seeanswer for `xid` below - let's handle it ina different issue

pkg/auth/db_service.go
itai-david: This is a bug in `ListUserPolicies` that was not detected as nothing triggers this fuynction. I bumped into it with my migration test, that does use it and had to fix it to get the migration test legit.
Should have done it in a separate commit, but it slipped my mind

itai-david:
> LGTM. Thanks for applying all the fixes! Please open a ticket to add an esti migration test for the auth package (although the other tests sort of cover it by design, but still there is probably some areas to cover with a dedicated auth system test)

Thank you @itaiad200 
Opend #3517 to follow up on `Esti` migration test

pkg/auth/kv_migrate_test.go
itai-david: #3520

pkg/auth/db_service.go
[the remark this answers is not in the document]
N-o-Z: Whether we use xid or uuid, we need a function (e.g. NewID) that both the service and migrate function use to generate IDs for models

pkg/auth/db_service.go
```diff
@@ -30,6 +30,17 @@ import (
...
+type userDetails struct {
+	name string
+	kvID string
+}
+
+type UserIDToDetails map[int64]userDetails
+
+type IDToName map[int]string
+
+var ErrExportedEntNotFound = errors.New("previously exported entitiy not found")
```
N-o-Z, line 42:
```suggestion
var ErrExportedEntNotFound = errors.New("previously exported entity not found")
```

eden-ohana: done

N-o-Z (approved):
Approved after last fixes

PR #3506, opened 2022-06-14
move tutorials to use cases

peacing:
move use case tutorials to new "use case" section. move exist use case section to "using lakefs"


adipolak (approved):
looks great!

PR #3508, opened 2022-06-14
Dynamodb benchmark tests and bug fixes

N-o-Z:


pkg/kv/dynamodb/store.go
```diff
@@ -372,6 +372,14 @@ func (s *Store) scanInternal(ctx context.Context, partitionKey, scanKey []byte,
 
 func (s *Store) Close() {}
 
+// DropTable used internally for testing purposes
+func (s *Store) DropTable() error {
```
itai-david, line 376: Can I accidentally use it to drop a production table?
E.g. if I write a test and use the wrong table name, can it be disastrous?

N-o-Z: Yes, but that is because of 2 things:
1. Because of the account we use to run the tests
2. Because of the permission you have on your local env
We can discuss this more - but I think we should take it somewhere else

itai-david (approved):

PR #3512, opened 2022-06-15
Adding a configuration for setting audit logs level

idanovo:


pkg/httputil/logging.go
```diff
@@ -71,18 +75,33 @@ func DebugLoggingMiddleware(requestIDHeaderName string, fields logging.Fields) f
...
-			logging.FromContext(r.Context()).WithFields(logging.Fields{
+			loggingFields := logging.Fields{
 				"took":        time.Since(startTime),
 				"status_code": writer.StatusCode,
 				"sent_bytes":  writer.ResponseSize,
-			}).Debug("HTTP call ended")
+			}
+			endLogMsg := "HTTP call ended"
+
+			switch strings.ToLower(middlewareLogLevel) {
```
nopcoder, line 85: consider exposing https://github.com/sirupsen/logrus/blob/v1.8.1/entry.go#L291

pkg/httputil/logging.go
```diff
@@ -71,18 +75,33 @@ func DebugLoggingMiddleware(requestIDHeaderName string, fields logging.Fields) f
...
+			case "panic":
+				logging.FromContext(r.Context()).WithFields(loggingFields).Panic(endLogMsg)
+			}
 		})
 	}
 }
 
 func LoggingMiddleware(requestIDHeaderName string, fields logging.Fields, traceRequestHeaders bool) func(next http.Handler) http.Handler {
-	if logging.Level() == "trace" {
+	loggingMiddlewareLevel := viper.GetString(config.LoggingAuditLogLevel)
```
nopcoder, line 102: Usually, we extract the configuration value from the config structure. The config structure also expose a method in this case it can also parse the string and return the logging level

pkg/httputil/logging.go
```diff
@@ -63,6 +66,7 @@ func DebugLoggingMiddleware(requestIDHeaderName string, fields logging.Fields) f
 				logging.MethodFieldKey:    r.Method,
 				logging.HostFieldKey:      r.Host,
 				logging.RequestIDFieldKey: reqID,
+				logging.LoggerName:        "MiddlewareLog",
```
nopcoder, line 69: 1. Usually we have property called "service".
2. How is this property is used? Because we didn't really create another logger and having a name can be confusing.

nopcoder (changes requested):

pkg/httputil/logging.go
```diff
@@ -13,6 +14,7 @@ type contextKey string
 
 const (
 	RequestIDContextKey contextKey = "request_id"
+	AuditLogEndMessage  string     = "HTTP call ended"
```
nopcoder, line 17: 1. No need the type - constant will get the type while using it.
2. You can update the message to include `API call information` or something like that if it make more sense.
```suggestion
	AuditLogEndMessage     = "HTTP call ended"
```

pkg/logging/dummy.go
```diff
@@ -52,6 +52,10 @@ func (d DummyLogger) Panic(args ...interface{}) {
 
 }
 
+func (d DummyLogger) Log(level string, args ...interface{}) {
```
nopcoder, line 55: Think it is time to have also level as part of the logging abstraction so we will not work with strings.

pkg/httputil/logging.go
nopcoder: Maybe "audit": "api"

pkg/httputil/logging.go
```diff
@@ -63,6 +65,7 @@ func DebugLoggingMiddleware(requestIDHeaderName string, fields logging.Fields) f
 				logging.MethodFieldKey:    r.Method,
 				logging.HostFieldKey:      r.Host,
 				logging.RequestIDFieldKey: reqID,
+				logging.LogSource:         "MiddlewareLogger",
```
nopcoder, line 68: Consider "audit" with value "api"

nopcoder (approved):
Couple of minor suggestions - looks good.

pkg/httputil/logging.go
idanovo: @nopcoder this change leads to [this](https://github.com/dominikh/go-tools/issues/866)

ortz (approved):
Nice!
Thank you!

PR #3516, opened 2022-06-15
Docs/use case rollback v2

peacing:
added more context for "data sync" and added why rollbacks section


vinodhini8694 (approved):
This looks good. The Atlassian example works great too!

PR #3522, opened 2022-06-16
Show content diff for conflicted files

### LINKED ISSUE #3269: Simple text file content conflict comparison under "Compare" tab
It seems like "show changes" is allowed only when there's no conflict.
Seeing the diff would be extremely viable for conflicts..

itaiad200:
closes #3269

### DIFF
--- a/webui/src/lib/components/repository/ObjectsDiff.jsx
+++ b/webui/src/lib/components/repository/ObjectsDiff.jsx
@@ -14,6 +14,7 @@ export const ObjectsDiff = ({diffType, repoId, leftRef, rightRef, path}) => {
     let right;
     switch (diffType) {
         case 'changed':
+        case 'conflict':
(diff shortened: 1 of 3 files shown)


webui/src/lib/components/repository/ObjectsDiff.jsx
```diff
@@ -14,6 +14,7 @@ export const ObjectsDiff = ({diffType, repoId, leftRef, rightRef, path}) => {
     let right;
     switch (diffType) {
         case 'changed':
+        case 'conflict':
```
talSofer, line 17: Can you please document the decision for how content-based diff will behave in case of a conflict? like a summary of the discussion you had over the github ticket.

talSofer (approved):
LGTM, thanks!

Added a comment about documentation, not blocking but worth addressing

eden-ohana (approved):

Issue #3493, opened 2022-06-12
ListRepositoryRuns: add validations in controller level


N-o-Z:
https://github.com/treeverse/lakeFS/pull/3427#discussion_r892799808

Allow either BranchID or CommitID but not both

PR #3524, opened 2022-06-17
KV milestone tidiness 

itaiad200:
Some small changes:
1. Fixing #3493 
2. unexport local functions
3. Fixed Runs + Tasks kv iterators to fail on entry parsing during `Next()` call.

### DIFF
--- a/pkg/actions/kv_run_results_iterator.go
+++ b/pkg/actions/kv_run_results_iterator.go
@@ -3,6 +3,7 @@ package actions
 import (
 	"context"
 	"errors"
+	"fmt"
 
 	"github.com/treeverse/lakefs/pkg/kv"
 )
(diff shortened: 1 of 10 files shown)


pkg/actions/kv_run_results_iterator.go
```diff
@@ -62,18 +64,24 @@ func (i *KVRunResultIterator) Next() bool {
 	if i.Err() != nil {
 		return false
 	}
-	return i.it.Next()
+	if !i.it.Next() {
+		i.entry = nil
+		return false
+	}
+	e := i.it.Entry()
+	if e == nil {
```
itai-david, line 72: How can this happen?
Is it only as a result of a bug, or is there any valid scenario where we may populate a nil value?
I mean, given the preceding `i.it.Next()` return `true`

itai-david (approved):
Thanks for these fixes
I have commented with one question, regarding the nature of moving the handking of a `nil` value from `Value()` to `Next()` (the 3rd point from your description above)
Other than that LG :+1:

pkg/actions/kv_run_results_iterator.go
itaiad200: Yeah, I have the same feeling that this might be a bit too defensive. It's not the only place though and I kept the existing behaviour as is. Maybe @N-o-Z can comment on that?

N-o-Z: We do not allow nil value in the Store - so this should never happen

Issue #3482, opened 2022-06-09
Revert confilct returns 500 response


itaiad200:
Conflicts can occur during a revert, for example when a file was changed since the commit-to-revert. We should be able to return a conflict response and not a 500 response.

```
./lakectl branch revert lakefs://repo1/main 6e30e5980abc2569604a5c797018eba5caa39b04411f6c969156e4a15fe8b21f
Branch: lakefs://repo1/main
Are you sure you want to revert the effect of commits 6e30e5980abc2569604a5c797018eba5caa39b04411f6c969156e4a15fe8b21f: y
conflict found
500 Internal Server Error
```

PR #3538, opened 2022-06-22
return code 409 instead of 500 for revert conflict

itai-david:
`handleAPIError` mapped `graveler.ErrConflictFound` to the default `http.StatusInternalServerError (500)` instead of the expected `http.StatusConflict (409)`
Fixed that and added an `Esti` test to create a conflict, attempt a revert and verify the status code

### DIFF
--- a/esti/rollback_test.go
+++ b/esti/rollback_test.go
@@ -2,7 +2,6 @@ package esti
 
 import (
 	"fmt"
-	"net/http"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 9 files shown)


itai-david:
Closes #3482

esti/rollback_test.go
```diff
@@ -244,3 +245,49 @@ func TestRevert(t *testing.T) {
 	body := string(getObjResp.Body)
 	require.Equal(t, objContent2, body, fmt.Sprintf("path: %s, expected: %s, actual:%s", objPath2, objContent2, body))
 }
+
+func TestRevertConflict(t *testing.T) {
```
itaiad200, line 249: minor: Could be a controller level test, it would reduce the runtime.

itai-david: I'm all in favour of **adding** a controller test, but do you think the `Esti` test should be removed?
As I see it, these E2E tests are the important ones, in terms of CI/CD, as they really cover a true use-case.
WDYT?

pkg/api/controller.go
```diff
@@ -1751,7 +1751,8 @@ func handleAPIError(w http.ResponseWriter, err error) bool {
 		errors.Is(err, actions.ErrParamConflict):
 		writeError(w, http.StatusBadRequest, err)
 
-	case errors.Is(err, graveler.ErrNotUnique):
+	case errors.Is(err, graveler.ErrNotUnique),
```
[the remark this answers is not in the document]
itai-david: `graveler.ErrNotUnique` is returned when attempting to create `repo`, `branch` or `tag` that already exist. Looking at `http` error codes, it does seem that `StatusConflict` is the most appropriate code to use. It looks like the usage of both `graveler.ErrNotUnique` and `graveler.ErrConflictFound`, and their mapping to `http.StatusConflict`, are in place

esti/rollback_test.go
itai-david: Added `controller` UT and remove `Esti` test

itai-david:
`swagger.yml` added documentration of `409` response for `revert`

itai-david:
> New responses need to be added to the swagger file too.

`swagger.yml` added documentration of `409` response for `revert`. Thanks for pointing this out

itaiad200 (approved):
Thanks for taking the extra mile with the esti -> controller tests. LGTM

PR #3562, opened 2022-06-23
Entirely remove Mermaid diagram from Markdown text

arielshaqed:
Turns out Jekyll (mistakenly!?!) parses only the first line of an HTML
comment as a comment, then inlines the rest.  So Mermaid text in comment
is actually displayed.

Remove the comment entirely.  Place the Mermaid diagram text into a separate
file so as not to lose it.


talSofer (approved):
Thank you!

PR #3563, opened 2022-06-23
OIDC improvements

johnnyaug:
* allow query params on authorization endpoint
* forward query params from login page to authorization endpoint
* add validation on ID token claims

### DIFF
--- a/pkg/api/auth_middleware.go
+++ b/pkg/api/auth_middleware.go
@@ -190,6 +190,7 @@ func userFromOIDC(ctx context.Context, logger logging.Logger, authService auth.S
 				"claim_name":     claimName,
 				"actual_value":   actualValue,
 				"expected_value": expectedValue,
+				"missing":        !ok,
 			}).Error("Authentication failed on validating ID token claims")
 			return nil, ErrAuthenticatingRequest
 		}
--- a/pkg/api/auth_middleware.go
+++ b/pkg/api/auth_middleware.go
@@ -182,6 +182,19 @@ func userFromOIDC(ctx context.Context, logger logging.Logger, authService auth.S
 		logger.WithField("sub", idTokenClaims["sub"]).Error("Failed type assertion for sub claim")
 		return nil, ErrAuthenticatingRequest
 	}
+
+	for claimName, expectedValue := range oidcConfig.ValidateIDTokenClaims {
+		actualValue, ok := idTokenClaims[claimName]
+		if !ok || actualValue != expectedValue {
+			logger.WithFields(logging.Fields{
+				"claim_name":     claimName,
+				"actual_value":   actualValue,
+				"expected_value": expectedValue,
+			}).Error("Authentication failed on validating ID token claims")
+			return nil, ErrAuthenticatingRequest
+		}
+	}
+
 	friendlyName := ""
 	if oidcConfig.FriendlyNameClaimName != "" {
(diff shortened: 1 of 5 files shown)


pkg/api/auth_middleware.go
```diff
@@ -182,6 +182,19 @@ func userFromOIDC(ctx context.Context, logger logging.Logger, authService auth.S
...
 	}
+
+	for claimName, expectedValue := range oidcConfig.ValidateIDTokenClaims {
+		actualValue, ok := idTokenClaims[claimName]
+		if !ok || actualValue != expectedValue {
+			logger.WithFields(logging.Fields{
+				"claim_name":     claimName,
+				"actual_value":   actualValue,
+				"expected_value": expectedValue,
+			}).Error("Authentication failed on validating ID token claims")
```
arielshaqed, line 193:
```suggestion
		if !ok || actualValue != expectedValue {
			logger.WithFields(logging.Fields{
				"claim_name":     claimName,
				"actual_value":   actualValue,
				"expected_value": expectedValue,
				"missing":        !ok,
			}).Error("Authentication failed on validating ID token claims")
```

pkg/api/oidc_login_handler.go
```diff
@@ -38,6 +39,16 @@ func NewOIDCLoginPageHandler(sessionStore sessions.Store, oauthConfig *oauth2.Co
 			writeError(w, http.StatusInternalServerError, "Failed to save auth session")
 			return
 		}
-		http.Redirect(w, r, oauthConfig.AuthCodeURL(state), http.StatusTemporaryRedirect)
+		opts := make([]oauth2.AuthCodeOption, 0, len(r.URL.Query()))
+		for k, v := range r.URL.Query() {
+			if len(v) < 1 {
```
arielshaqed, line 44: Is this the same as:
```suggestion
			if v == "" {
```
?

pkg/api/oidc_login_handler.go
```diff
@@ -38,6 +39,16 @@ func NewOIDCLoginPageHandler(sessionStore sessions.Store, oauthConfig *oauth2.Co
...
+				continue
+			}
+			opts = append(opts, oauth2.SetAuthURLParam(k, v[0]))
```
arielshaqed, line 47: I don't know enough about OIDC, but this scares me: the user supplies parameters that we pass directly to the OIDC server without understanding what they do.  We will override _some_ of these on l. 49-41... but do we know that we override _everything_ required?

I would feel safer if we only allowed whitelisted options here.

arielshaqed (commented):
Thanks!

I don't understand the OIDC flow well enough to approve.  Specifically: can you show me what to read to understand what auth URL params are, and what I can do on the server by setting them?

pkg/api/auth_middleware.go
johnnyaug: done

pkg/api/oidc_login_handler.go
johnnyaug: `v` in this context is an array.

pkg/api/oidc_login_handler.go
arielshaqed: After f2fing I understand that these parameters go (_only_) into the redirection URL returned from the OIDC provider.  Since there is nothing secret about the generated URL, there is no attack vector here.

arielshaqed (approved):
THANKS!

PR #3564, opened 2022-06-23
Refactor/delegate emailer to authservice

johnnyaug:


api/authorization.yml
```diff
@@ -156,6 +156,8 @@ components:
           format: byte
         external_id:
           type: string
+        invite:
```
arielshaqed, line 159: Please document this field. It is not clear to me whether this is imperative or descriptive.

pkg/api/login_handler.go
```diff
@@ -19,17 +19,15 @@ type LoginResponseData struct {
...
+	DefaultResetPasswordExpiration = 20 * time.Minute
 	// Deprecated: use gorilla managed session
 	JWTCookieName           = "access_token"
 	InternalAuthSessionName = "internal_auth_session"
 	TokenSessionKeyName     = "token"
 	OIDCAuthSessionName     = "oidc_auth_session"
 
-	LoginAudience         = "login"
-	ResetPasswordAudience = "reset_password"
+	LoginAudience = "login"
```
arielshaqed, line 30: The fact that the reset audience definition had to move is probably a sign that the login audience should also move over. (Not this PR, just an observation)

arielshaqed (approved):
Neat! Looks like a much better place for this logic.
Thanks...

arielshaqed:
I am curious: I was under the impression that we would be removing email-based login in favour of OIDC.  If so, then why are we improving this code when we could be working towards removing it?

api/authorization.yml
johnnyaug: Done

johnnyaug:
Regarding @arielshaqed's comment, we synced f2f, writing here to document: While we may want to phase out email-based login, this wasn't yet decided upon.

PR #3566, opened 2022-06-25
KV3 - graveler - Discovery and Planning

### LINKED ISSUE #3531: KV3 Milestone - Discovery and Planning
KV3 is all about `graveler` porting to `KV`. This ticket is for investigating and defining the required changes, and to bring up possible risks.
It needs to cover `graveler` move to `KV`, migration of current data from `SQL` and testing aspects

References:
[lakeFS on KV Design](https://github.com/treeverse/lakeFS/blob/b3204edad00f88f8eb98524ad940fde96e02ab0a/design/open/metadata_kv/index.md) - special attention to the graveler part
[lakeFS on KV Execution Plan](https://github.com/treeverse/lakeFS/blob/7b1fa03115891d43d89dde1d75ce778b11469dd1/design/open/lakefs-kv-execution-plan.md) - note the Per Package section
[lakeFS on KV Testing Plan](https://github.com/treeverse/lakeFS/blob/e5bdae22651d20458b6460491f75df18e64b0d1f/design/open/lakefs-on-kv-testing-plan.md) - summary of testing coverage
[KV Migration](https://github.com/treeverse/lakeFS/blob/e5bdae22651d20458b6460491f75df18e64b0d1f/design/accepted/kv_migration.md) - migration guidelines to follow

**DoD**
An approved execution plan for `graveler` over `KV` with derived tasks

itai-david:
closes #3531


itai-david:
Create the following issues:
#3567
#3568 
#3569 
#3570 
#3571 
#3572 
#3573 
#3574 
#3575

design/open/kv3-execution-plan.md
```diff
@@ -0,0 +1,52 @@
...
+`graveler_branches` is currently being protected by a lock, during commit actions, and in order to support lockless commits with `KV`, the [following design](https://github.com/treeverse/lakeFS/blob/b3204edad00f88f8eb98524ad940fde96e02ab0a/design/open/metadata_kv/index.md#graveler-metadata---branches-and-staged-writes) is proposed. However, as a first step we can implement it using a naive table lock, just to get it going, or better, a `branch-key` based locking, to prevent staging during commits</br>
+`graveler_staging_kv` creates entries based on the current branch `staging_token` and is protected from creating entries during a commit under the same lock. This should also be supported lockless-ly, as referred to by the above proposal, and can also be solved with naive locking, as a first step</br>
+Note: All the above tables (along with the late `auth_users`) are also used by `pkg/diagnostics` for diagnostics collection (RO). 
+
+## Iterators
+The following iterators are implemented and used by `pkg/graveler`:
+  * `branch_iterator` - iterates branches by repository. Has an option to order by ID (trivial) or by CommitID (Used by GC. How do we support that?)
```
itaiad200, line 15: I don't understand why anyone would like to sort anything based on a random generated string (commit ID) - maybe it's not really needed?

design/open/kv3-execution-plan.md
```diff
@@ -0,0 +1,52 @@
...
+  * `commit_iterator` - iterates a commit ancestry by running BFS. Not an `iterator` in the `KV` sense. Should be supported as is with `KV` `Get`
+  * `commit_ordered_iterator` - iterates commits of a repo, ordered by commit ID (regular `KV` `Scan` supports that). However - has an option to iterate over not-first-parent commits only (e.g. commits that are not a first parent of any other commit). Probably could (and should) be implemented by a specific `Next()` for that case
```
itaiad200, line 17: Needs more details. The list of `not-first-parent commits` can always change since you may introduce a new child to any commit at any time. I wonder how do we use this list? And why would the ordering of CommitID matter?

I find it hard to come up with an iterator on top of the kv implementation that would cover these requirements.

design/open/kv3-execution-plan.md
```diff
@@ -0,0 +1,52 @@
...
+  * `repository_iterator`, `tag_iterator` - simple iterators ordered by ID. `KV` standard `Scan` should be sufficient
+  * `staging/iterator` - iterates over a specific `staging_token` (as prefix) by order of key. Should be good with `KV` `Scan`
+
+## Deleting a Repository
+Unlike other entities' delete, which are standalone, deleting a repository results in deleting all branches, commits and tags correlated to that repository. While being protected by a transaction for `SQL` we need to handle failures for `KV`. One option that comes in mind is to flag the repository as `"being_deleted"` (maybe a `KV` entry for `being_deleted` repos, that can be scanned), clean everything, and if successful remove the flag. Another option is to remove the repository entry first, and deploy a background GC that scans for branches/commits/tags that are related to a non-exist repository and delete them. Other ideas?
```
itaiad200, line 22: Interesting. The problem with flagging the repository is that now you would have to check it with every call. For example, trying to get a Tag would first need to check that flag which is stored somewhere else.
I think the best we can do here is to delete it in an order that would leave a consistent state in every possible failure. One way could be: Tags, Staging Tokens (as in workspaces), Branches (all except for the default), update the default branch commit to "NotACommit", Commits and finally the default branch and the repo.

design/open/kv3-execution-plan.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Dump/Restore Refs
+Uses `graveler` ref-store for dumping a whole repository (branches, commits and tags) and restoring it to a bare repository. As there are no direct DB accesses, this should go seamlessly, but definitely need to be verified</br>
+Moreover, as these are quite exhaustive operations, may be used as part of performance verification
+
+## Suggested Key Schema
+Note: This is an initial suggestion that seems to support the required iterators. It tries to encapsulate the immutable parts of identifying an object (e.g. a branch can be identified by its name, but not by a commit, as it (a) changes and (b) a commit can be shared by branches. a repo and a tag has one:many relation, so a tag key contains its repo etc.) It is totally open to discussion and improvements.
+* Repository - `repo/<REPO_NAME>`
+* Branch     - `repo/<REPO_NAME>/branch<BRANCH_NAME>`
```
itaiad200, line 34: Missing a `/`?

design/open/kv3-execution-plan.md
```diff
@@ -0,0 +1,52 @@
...
+* Commit     - `repo/<REPO_NAME>/commit/<COMMIT_ID>`
+* Tag        - `repo/<REPO_NAME>/tag/<TAG_ID>`
+* Staged Obj - `repo/<REPO_NAME>/staging_token/<STAGING_TOKEN>/key/<OBJ_KEY>`
```
itaiad200, line 37: I pray for 0 double slash bugs 🙏

design/open/kv3-execution-plan.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Testing
+`pkg/graveler/ref` and `pkg/graveler/staging` currently has a **very high** test coverage. This should make it easier to trust the transition to `KV`. However, one notable aspect that is the absence of `benchmarks`. As `graveler` will introduce new algorithms to support lockless commits, there is a risk of performance degradation which need to be identified
+
+## Dump/Restore Refs
+Uses `graveler` ref-store for dumping a whole repository (branches, commits and tags) and restoring it to a bare repository. As there are no direct DB accesses, this should go seamlessly, but definitely need to be verified</br>
+Moreover, as these are quite exhaustive operations, may be used as part of performance verification
+
+## Suggested Key Schema
+Note: This is an initial suggestion that seems to support the required iterators. It tries to encapsulate the immutable parts of identifying an object (e.g. a branch can be identified by its name, but not by a commit, as it (a) changes and (b) a commit can be shared by branches. a repo and a tag has one:many relation, so a tag key contains its repo etc.) It is totally open to discussion and improvements.
```
itaiad200, line 32: I don't think it covers the special iterators above..

design/open/kv3-execution-plan.md
```diff
@@ -0,0 +1,52 @@
...
+* Repository - `repo/<REPO_NAME>`
+* Branch     - `repo/<REPO_NAME>/branch<BRANCH_NAME>`
+* Commit     - `repo/<REPO_NAME>/commit/<COMMIT_ID>`
+* Tag        - `repo/<REPO_NAME>/tag/<TAG_ID>`
+* Staged Obj - `repo/<REPO_NAME>/staging_token/<STAGING_TOKEN>/key/<OBJ_KEY>`
+
+# Execution Plan
+* Agree on keys schema (see here after) - [#3567](https://github.com/treeverse/lakeFS/issues/3567)
+* Supporting `KV` along side `SQL` (same as was done for previous modules)  
+  * Step 1: Naive implementation with locking - `lakeFS` on KV can start with empty DB and run all graveler operations correctly, although performance may be degraded and exhaustive operations (concurrent commits and uploads, anything else?) should not be considered - [#3568](https://github.com/treeverse/lakeFS/issues/3568)
```
itaiad200, line 42: This step alone is quite big. It means reimplementing Graveler in a single step.. I think we should support an hybrid mode, i.e. if it makes sense, move just the tags, then just the staging areas, then commits, finally branches.

design/open/kv3-execution-plan.md
```diff
@@ -0,0 +1,52 @@
+
+# General
+## Tables
+Graveler uses 5 DB tables:
+* `graveler_repositories`,`graveler_branches`,`graveler_commits`,`graveler_tags` - used by `pkg/graveler/ref`
+* `graveler_staging_kv` - used by `pkg/graveler/staging`
+
+`graveler_repositories`, `graveler_commits` and `graveler_tags` are immutable and does not update entries, and so should be straight forward to implement over our `KV` (`graveler_commits` actually performs update for commit generation, but this is only valid for `refs-restore` which can only take place on a `bare` repo and so it is of no concern)</br>
+`graveler_branches` is currently being protected by a lock, during commit actions, and in order to support lockless commits with `KV`, the [following design](https://github.com/treeverse/lakeFS/blob/b3204edad00f88f8eb98524ad940fde96e02ab0a/design/open/metadata_kv/index.md#graveler-metadata---branches-and-staged-writes) is proposed. However, as a first step we can implement it using a naive table lock, just to get it going, or better, a `branch-key` based locking, to prevent staging during commits</br>
```
N-o-Z, line 9: I agree with this approach.
Generally I think the lock-less feature is a post KV feature

design/open/kv3-execution-plan.md
```diff
@@ -0,0 +1,52 @@
...
+  * `staging/iterator` - iterates over a specific `staging_token` (as prefix) by order of key. Should be good with `KV` `Scan`
+
+## Deleting a Repository
+Unlike other entities' delete, which are standalone, deleting a repository results in deleting all branches, commits and tags correlated to that repository. While being protected by a transaction for `SQL` we need to handle failures for `KV`. One option that comes in mind is to flag the repository as `"being_deleted"` (maybe a `KV` entry for `being_deleted` repos, that can be scanned), clean everything, and if successful remove the flag. Another option is to remove the repository entry first, and deploy a background GC that scans for branches/commits/tags that are related to a non-exist repository and delete them. Other ideas?
+
+## Testing
+`pkg/graveler/ref` and `pkg/graveler/staging` currently has a **very high** test coverage. This should make it easier to trust the transition to `KV`. However, one notable aspect that is the absence of `benchmarks`. As `graveler` will introduce new algorithms to support lockless commits, there is a risk of performance degradation which need to be identified
```
N-o-Z, line 25: Since you already stated that lock-less commits are out of scope for KV, I think this part is irrelevant for this doc

design/open/kv3-execution-plan.md
N-o-Z: I would change the naming convention to plural - 'branches', 'commits' etc. Consistent with other packages

design/open/kv3-execution-plan.md
```diff
@@ -0,0 +1,52 @@
...
+* Branch     - `repo/<REPO_NAME>/branch<BRANCH_NAME>`
+* Commit     - `repo/<REPO_NAME>/commit/<COMMIT_ID>`
+* Tag        - `repo/<REPO_NAME>/tag/<TAG_ID>`
+* Staged Obj - `repo/<REPO_NAME>/staging_token/<STAGING_TOKEN>/key/<OBJ_KEY>`
+
+# Execution Plan
+* Agree on keys schema (see here after) - [#3567](https://github.com/treeverse/lakeFS/issues/3567)
+* Supporting `KV` along side `SQL` (same as was done for previous modules)  
+  * Step 1: Naive implementation with locking - `lakeFS` on KV can start with empty DB and run all graveler operations correctly, although performance may be degraded and exhaustive operations (concurrent commits and uploads, anything else?) should not be considered - [#3568](https://github.com/treeverse/lakeFS/issues/3568)
+  * Step 2: Lockless Commits - as described above - [#3569](https://github.com/treeverse/lakeFS/issues/3569)
```
N-o-Z, line 43: I've seems to misunderstand the previous section - I really think lock free commits are not part of the KV scope. I believe it's best to first focus on the transition to KV with the current implementation and only after that working on the lock free feature

design/open/kv3-execution-plan.md
```diff
@@ -0,0 +1,52 @@
...
+## Suggested Key Schema
+Note: This is an initial suggestion that seems to support the required iterators. It tries to encapsulate the immutable parts of identifying an object (e.g. a branch can be identified by its name, but not by a commit, as it (a) changes and (b) a commit can be shared by branches. a repo and a tag has one:many relation, so a tag key contains its repo etc.) It is totally open to discussion and improvements.
+* Repository - `repo/<REPO_NAME>`
+* Branch     - `repo/<REPO_NAME>/branch<BRANCH_NAME>`
+* Commit     - `repo/<REPO_NAME>/commit/<COMMIT_ID>`
+* Tag        - `repo/<REPO_NAME>/tag/<TAG_ID>`
+* Staged Obj - `repo/<REPO_NAME>/staging_token/<STAGING_TOKEN>/key/<OBJ_KEY>`
+
+# Execution Plan
+* Agree on keys schema (see here after) - [#3567](https://github.com/treeverse/lakeFS/issues/3567)
```
N-o-Z, line 40: What I'm missing is the data modelling for KV store. It can either be part of https://github.com/treeverse/lakeFS/issues/3567 or as a new issue. But it is required as part of the common code base that we all work over. In auth we didn't do it well enough and as a result there were many changes during implementation and technical debts left.

N-o-Z (changes requested):
Thanks @itai-david, 
I have 2 main concerns:
1. Missing data models
2. Adding the lock free commits as part of the KV milestone (maybe this can be discussed further)
Other than that - it seems like a very thorough plan!

design/open/kv3-execution-plan.md
arielshaqed: If a repository (probably with a prefix) is a partition key, then you can delete a repository by deleting everything inside the partition.  And you can keep a list of "repositories being deleted" (or an object inside the partition of each repository) and keep retrying to delete those repositories.

arielshaqed (commented):
Thanks!

My chief concern is that the GC calls will continue to work with reasonable performance.  I would like to understand the proposed iterators better to be sure of this...

arielshaqed:
(Also cool that this is really happening!)

design/open/kv3-execution-plan.md
```diff
@@ -0,0 +1,52 @@
...
+## Dump/Restore Refs
+Uses `graveler` ref-store for dumping a whole repository (branches, commits and tags) and restoring it to a bare repository. As there are no direct DB accesses, this should go seamlessly, but definitely need to be verified</br>
+Moreover, as these are quite exhaustive operations, may be used as part of performance verification
+
+## Suggested Key Schema
+Note: This is an initial suggestion that seems to support the required iterators. It tries to encapsulate the immutable parts of identifying an object (e.g. a branch can be identified by its name, but not by a commit, as it (a) changes and (b) a commit can be shared by branches. a repo and a tag has one:many relation, so a tag key contains its repo etc.) It is totally open to discussion and improvements.
+* Repository - `repo/<REPO_NAME>`
+* Branch     - `repo/<REPO_NAME>/branch<BRANCH_NAME>`
+* Commit     - `repo/<REPO_NAME>/commit/<COMMIT_ID>`
+* Tag        - `repo/<REPO_NAME>/tag/<TAG_ID>`
```
eden-ohana, line 36: tag path should save its commit ID

design/open/kv3-execution-plan.md
itaiad200, line 9: After reading it once again with the help of @arielshaqed and @N-o-Z comments, I think I have a better idea of what you were aiming for here. A `naive table lock` means that we would be using the same locking technique we have for the current postgres DB, right? Well - that cannot be achieved with KV. It's not part of the KV interface and adding it to the current 3 implementations we have is harder than just straightforward use the mechanisms described in the original proposal.

arielshaqed: I think by now we are arguing about different words to describe the exact same thing.  I would strongly encourage that we simply stop using the word "lock", CAS-type operations ("SetIf") like those used here are at a lower level and easier to use directly.  The term "lock" is overloaded and has many different meaning.  (Note that "lock-free" does **not** talk about locks, intentionally, so we can continue to use that term.)

Let's be concrete.  I call the following "the na&iuml;ve way" to perform a protected action on a branch:

1. Get the original branch state `o`.
1. (Maybe perform some actions, they might all be lost in the last step.)
1. Compute a new branch state `n`.
1. SetIf the branch state to `n` IF it is still `o`.
   * (If setting failed and some other actions were performed, maybe clean them up.)

This type of process is sometimes loosely called "optimistic locking".

design/open/kv3-execution-plan.md
johnnyaug: @itai-david, sure: this iterator is then used as input for a `GCStartingPointIterator`. This iterator assumes the branch heads are sorted by commit ID. This way, it can go over branch heads side by side with other commit IDs that are also sorted, and compare them.

itai-david: Thank you @johnnyaug.
@itaiad200 @arielshaqed - I looked at the GC code and it looks like we either need to support that with our `Graveler over KV` impl, or we will need to change the way GC traverses commits.
I think the first can be supported by maintaining a `repo/<REPO_NAME>/commit/<COMMIT_ID>/branch/<BRANCH_NAME>`. These keys will have to be constantly removed and added as commits take place, but we do have all the info at the time of the commit, so I don't think it is too difficult.
WDYT?

design/open/kv3-execution-plan.md
itai-david: Out of KV3 scope, or out of KV scope all together?
I think we need a wider opinions range here
@eden-ohana @itaiad200 - WDYT?

design/open/kv3-execution-plan.md
itai-david: Looks like we're aligned on this one

design/open/kv3-execution-plan.md
itai-david: These commits are used side by side with the `branch-by-commitID` list, as explained by @johnnyaug, hence the order by `commitID` for that list too.
regarding the `not-first-parent` attribute, its usage and the fact it is subtle to changes - is it any different than the way it is used today, over `SQL`? I see that the decision of whether a commit is `non-first-parent` or not is made at the time of query execution, but this also can be changed by the time the GC algorithm get there. Am I wrong?
I guess I miss some context on that algorithm - @johnnyaug, can you please help here?

design/open/kv3-execution-plan.md
itaiad200: Right @arielshaqed, but now every time you try to get an object from that repo, you would have to query that list.

design/open/kv3-execution-plan.md
itai-david: > I find it hard to come up with an iterator on top of the kv implementation that would cover these requirements.

@itaiad200 - I absolutely agree with that last sentence. Maybe we can maintain a live set of `not-first-parent` commits and modify it as commits occur. i.e. each new commit is by default `not-first-parent` and for each further operation, which makes it a `first-parent`, we **try** to remove that commit from the list (possibly already removed, hence the **try**) (**Question**: Can a commit **become** `not-first-parent` if it is already a `first-parent`?)
I can think of a couple of ways to  implement this list, each optnio with its own drawbacks:
* We can use some `not_first_parent/<COMMIT_ID>`. This will require require searching the list for each commit, or
* We can "flag" the commit key itself i.e. commit `commit/<COMMIT_ID>` is accompanied by `commit/<COMMIT_ID>/not_first_parent` (not sure this works well with our scan logic, right?)
What other options do we have?

design/open/kv3-execution-plan.md
itai-david: I like the usage of repository as partition. It makes sense in the way that each scan operation (in that case, for deleting everything) is partition bounded eitherways, plus, I think the locallity of partition (supported by `DynamoDB`, as an example) seems to be working in our benefits with this one (be it performance-wise or GDPR-wise).
We will need to maintain the repository entry itself outside of the partition, for repositories listing, correct?
I think we can "mark" a repository as `being-deleted` by simply remove its entry from the repositories list, and once successful, start removing everything in the relevant partition
Any objection we go with that?

design/open/kv3-execution-plan.md
itai-david: > In the end, this doesn't need to be a consistent list: when you find a commit that is not a first parent, you can safely return it, regardless of whether the resulting list as a whole makes sense.

Thank you for this input @johnnyaug.
What happens if I found a `not-first-parent` commit, return it as a starting point, and before the GC contniues, a commit is created over that commit, making it a `first parent`? Is that possible?

design/open/kv3-execution-plan.md
itai-david: @arielshaqed I'm not sure I fully understand how a `Bloom Filter` can help here, but I guess this requires a bit of redesign to the GC commits gathering alogrithm (?)
If that is so, I think it should be out of scope for this KV milestone - I think we should "force" our graveler over KV impl to provide all the functionality currently supported by graveler over SQL, in terms of correctness, even in the painful price of performance regression, and handle any significant algorithmic change in another scope. WDYT?
For that purpose, I think defining a `commitId` to `branchName(s)` mapping (as a key?) can provide what we need, at least as a first step (regardless of the (very) wrong conception that commits "have" or "belong to" branch(s))

design/open/kv3-execution-plan.md
```diff
@@ -0,0 +1,52 @@
...
+`graveler_staging_kv` creates entries based on the current branch `staging_token` and is protected from creating entries during a commit under the same lock. This should also be supported lockless-ly, as referred to by the above proposal, and can also be solved with naive locking, as a first step</br>
+Note: All the above tables (along with the late `auth_users`) are also used by `pkg/diagnostics` for diagnostics collection (RO). 
+
+## Iterators
+The following iterators are implemented and used by `pkg/graveler`:
+  * `branch_iterator` - iterates branches by repository. Has an option to order by ID (trivial) or by CommitID (Used by GC. How do we support that?)
+  * `commit_iterator` - iterates a commit ancestry by running BFS. Not an `iterator` in the `KV` sense. Should be supported as is with `KV` `Get`
```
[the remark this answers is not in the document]
itai-david: I believe the `commit_iterator` is used solely for the by `Log`, so indeed this is the case. What I meant to say is that this is not a real `iterator` in the `KV` sense, but rather an internal implementation that uses `Get` functionality iteratively. IMHO this is a trivial case which we should not change at this scope.
The GC uses another iterator `commit_ordered_iterator` which we are discussing [here](https://github.com/treeverse/lakeFS/pull/3566#discussion_r906716400)

design/open/kv3-execution-plan.md
itai-david: @itaiad200 - Yup. Fixed
@N-o-Z - done
@eden-ohana - Good catch, thank you. Fixed

design/open/kv3-execution-plan.md
itai-david: @eden-ohana - agree. Added under the new scema
@itaiad200 - double slash is the obvious. I added an option for triple slashes... My new suggestion removes the redundant keywords such as `key` and `branch` from anywhere except from the key prefix

design/open/kv3-execution-plan.md
itai-david: Correct - we are still discussing it. I believe we should support these by additional keys, and strive to remove them as an improvements iteration out of the scope of this milestone (and probably out of the scope of KV all together, as suggeated by @N-o-Z )

design/open/kv3-execution-plan.md
```diff
@@ -0,0 +1,52 @@
...
+
+`graveler_repositories`, `graveler_commits` and `graveler_tags` are immutable and does not update entries, and so should be straight forward to implement over our `KV` (`graveler_commits` actually performs update for commit generation, but this is only valid for `refs-restore` which can only take place on a `bare` repo and so it is of no concern)</br>
+`graveler_branches` is currently being protected by a lock, during commit actions, and in order to support lockless commits with `KV`, the [following design](https://github.com/treeverse/lakeFS/blob/b3204edad00f88f8eb98524ad940fde96e02ab0a/design/open/metadata_kv/index.md#graveler-metadata---branches-and-staged-writes) is proposed. However, as a first step we can implement it using a naive table lock, just to get it going, or better, a `branch-key` based locking, to prevent staging during commits</br>
+`graveler_staging_kv` creates entries based on the current branch `staging_token` and is protected from creating entries during a commit under the same lock. This should also be supported lockless-ly, as referred to by the above proposal, and can also be solved with naive locking, as a first step</br>
```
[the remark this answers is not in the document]
itai-david: Rephrased

design/open/kv3-execution-plan.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Dump/Restore Refs
+Uses `graveler` ref-store for dumping a whole repository (branches, commits and tags) and restoring it to a bare repository. As there are no direct DB accesses, this should go seamlessly, but definitely need to be verified</br>
+Moreover, as these are quite exhaustive operations, may be used as part of performance verification
+
+## Suggested Key Schema
+Note: This is an initial suggestion that seems to support the required iterators. It tries to encapsulate the immutable parts of identifying an object (e.g. a branch can be identified by its name, but not by a commit, as it (a) changes and (b) a commit can be shared by branches. a repo and a tag has one:many relation, so a tag key contains its repo etc.) It is totally open to discussion and improvements.
+* Repository - `repo/<REPO_NAME>`
+* Branch     - `repo/<REPO_NAME>/branch<BRANCH_NAME>`
+* Commit     - `repo/<REPO_NAME>/commit/<COMMIT_ID>`
```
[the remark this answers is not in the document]
itai-david: Correct. These are the same. I tried to maintain the `SQL` names, just for simplicity of this doc

design/open/kv3-execution-plan.md
itai-david: As part of the key?
It will be in the value

design/open/kv3-execution-plan.md
itai-david: Actually, what I had in mind for **naive and temporary locking** is locking of some common resource, on a logical level, not related to `KV`.
Both staging and commit operations (on the same branch) mutually block each other (this is the case, right? any other operatios need to be blocked), by locking a common resource (mutex?), just for the correctness of things.
We can reduce the contention by having NumLocks (say 2^10) such locks, and each operatation locks branchID%NumLocks, letting several operations (up to NumLocks, not sure how to calculate the average) run in parallel.
Any reason why that should not work as a **naive, temporary yet correct, solution**?

design/open/kv3-execution-plan.md
itaiad200: @arielshaqed and I discussed it a bit more and I think we now understand why it's needed for `GCStartingPointIterator`. We can have the same commitID order for a branch iterator on top of the kv implementation, by simply fetching all branches and sorting them in memory. It will not require to introduce any new indexing in KV. A counter argument would be memory consumption, but it's used by an algorithm that already stores all commits in memory, so there should not be any degradation.

design/open/kv3-execution-plan.md
itaiad200: Let's be more specific - what are the things we think can degradate? There might be paths we would like to benchmark now. Graveler is pretty wide so if you have any specific paths you think we need to benchmark let's discuss.

design/open/kv3-execution-plan.md
N-o-Z: Please note that with current path definitions - and as @eden-ohana mentioned, there's a risk of accidentally seeping into different entities when scanning on repositories

itai-david: @N-o-Z - that was fixed with the recent update to the keys

itai-david:
Having a discussion with @itaiad200 @N-o-Z, we came to the following understandings. Unless anything is missing, wrong or unacceptable, I will update the doc to reflect these, so we can move forward:
* **`GC`, `not-first-parent` and `branch iterator by commit ID`** - The following GC algorithm will remain unchanged, and `graveler over KV` will supported the 2 above requirement by in memory processing, i.e. `not-first-parent` will fetch all commits, using standard `KV` means (`scan`) and filtering out all not `not-first-parent` commits, in memory. `branc iterator by commit ID` will fetch all branches using `KV`, and will sort them by commit ID in memory. Will add a task to revisit this and further discuss this out of the scope of KV3 milestone.
* **Commits, staging and lock-free** - we will follow @arielshaqed [suggestion](https://github.com/treeverse/lakeFS/pull/3566#discussion_r907073940) to prevent commits overriding each other by using `KV`'s `SetIf` on the branch, making sure no racing commit is already finished and being overrided. In case a commit fails to finish due to another commit "beating" it to the finish line - a retry should be attmpted. This implementation will require to support multiple staging tokens, as proposed by [lakeFS on KV design proposal](https://github.com/treeverse/lakeFS/blob/b3204edad00f88f8eb98524ad940fde96e02ab0a/design/open/metadata_kv/index.md#committer-flow). Lock-free commits will handled out of thye scope of KV3 milestone
* **Using of partitions** * - Partitions, and partition keys, will be used to distinguish different logical aspects of graveler.
  * A general partition - `graveler` - will contain the repositories, to allow iterating them
  * A partition per each repository - `<REPO_NAME>` - will contain all repository related data, except for staged objects. This data includes branches, commits and tags. This will allow easy access to all the repository data and our ability to traverse it preceisley (in case of future cleaning algorithm, to support failed deletions, or alike)
  * A partition per each staging token - `<STAGING_TOKEN>` - will contain all the staged obejcts under this token. Will allow a beeter load balancing for multiple staging tokens handling. Note that the `<STAGING_TOKEN>` contains the `<REPO_NAME>` and `<BRANCH_NAME>`
* **Repository Deletion** * - A separate task to further investigate that. Currently, 2 possibilities seem valid:
  * Option 1 - It does seem that most (if not all) `graveler`'s operations, perfrom a repository existence check by issuing a call to `GetRepository`, which perfroms a DB access. If that is indeed the case, we can easily "mark" a repository as deleted by removing the repository entry from the common partition `graveler`, thus making it unreachable. We can later iterate **all** entries under this repository partition and delete them one by one, preceding the branch entry deletion by iteratively deleting all relelvant staging partitions. Failures can be handled by a background garbage collector (which, BTW, can handle the inital deletion too), but that is only for space saving purposes and should have no effect on the deleted repository itself, once the repository entry is deleted. A point to consider - what happens if a repository entry is deleted, the data in the relevant repository partition fails to delete, and a new repository with the same name is created? Can we solve that with adding a guid to the repository and use it as the patition key?
  * Option 2 - As suggested by @itaiad200 - performing the deletion in a way that preserves consistency with each possible failure - e.g. staging partitions need to be deleted before the branch, as not to leave any hanging in case of failure, commits cannot be deleted before branches, as this may leave a branch pointing to a deleted commit, the default branch cannot be deleted until the repository is deleted, etc. Part of the discovery should define the correct order
* **Execution Plan** - The initial task of basic `graveler` implementation over `KV`, is estimeted to be too big to be well contained as a single task. Instead, we will break it to smaller pieces, by initially implementing the `KV graveler` skeleton, which will delegate all calls to the `DB graveler`. We will later implement each functionality separately
* **Benchmarks** - A decision is to be made regarding the required benchmarks - which scenarios do we really want to test. One example is multiple parallel commits, that will definitely fail each other until all retries will succeed - while this is an exhaustive scenario, which stress the performance to an edge, it might not be as interesting as a "real life" scenario. The bottom line is we can predict that such a scenario will show performance degradation for our `not-lock-free` commit algoritm, but it is probably not interesting. Another example is a single long commit, with multiple staging operations taking place at the same time. This scenario is much more common in real life usage and a performance regression in this scenario is something we better discover sooner than later.

design/open/kv3-execution-plan.md
itai-david: Added to the doc - support of `iteration by commitID` by sorting branches in memory

design/open/kv3-execution-plan.md
itai-david: Added to doc - implementatingan initial skeleton with no real KV access (delegating all DB related calls to `DB gravlere`, as we discussed) and deriving followup implementation tasks

design/open/kv3-execution-plan.md
itai-david: Created #3586 to fully support lock-free commits, out of KV3 scope

design/open/kv3-execution-plan.md
itai-david: Added to the doc and to #3571 to comments we discussed earlier, regarding which scenarios are interesting and should be benchmarked

design/open/kv3-execution-plan.md
itai-david: Data modeling added to doc. As graveler pretty much uses the SQL DB as KVStore, there is not much of a difference, excpet for a few things we will need to add

design/open/kv3-execution-plan.md
```diff
@@ -47,17 +73,21 @@ Moreover, as these are quite exhaustive operations, may be used as part of perfo
 * `branch_iterator`, sorted by branch name - trivial. sorted by commitID - TBD (still being discussed)
 * Repository Deletion - as suggested, maintaining all repository relevant entries under a unique partition will make it easy to delete as an entire partition. The repository key itself is outside of the partition (in a general common partition, to support listing) and will be deleted first. Once deleted, the entire partition will become unavailable, making the data state consistent either ways
 
+## Update Commits, staging and lock-free
+we will follow @arielshaqed suggestion to prevent commits overriding each other by using `KV`'s `SetIf` on the branch update, making sure no racing commit is already finished and being overridden. In case a commit fails to finish due to another commit "beating" it to the finish line - a retry should be attempted. This implementation will require to support multiple staging tokens, as proposed by [lakeFS on KV design proposal](https://github.com/treeverse/lakeFS/blob/b3204edad00f88f8eb98524ad940fde96e02ab0a/design/open/metadata_kv/index.md#committer-flow). Lock-free commits will be handled out of the scope of KV3 milestone
```
itaiad200, line 77: Actually Ariel's suggestion was the `Lock-free commits`

design/open/kv3-execution-plan.md
```diff
@@ -47,17 +73,21 @@ Moreover, as these are quite exhaustive operations, may be used as part of perfo
...
-  * Step 2: Lock-free Commits - as described above - [#3569](https://github.com/treeverse/lakeFS/issues/3569)
-  * Step 3: Decide and implement a working solution for **Deleting a Repository** - [#3570](https://github.com/treeverse/lakeFS/issues/3570)
-  (Steps 2 and 3 are independent)
+* Supporting `KV` along side `SQL` (same as was done for previous modules)
+  * Step 1: Building a initial `KV graveler` and a `DB graveler`. `KV graveler` should delegate all calls to `DB gravler` (no `KV` usage at all for step) in order to allow partial `KV` support.
+  [#3568](https://github.com/treeverse/lakeFS/issues/3568) was modified to reflect that, and should result with continuation tasks for the `KV` support steps within
+  * Step 2: Basic implementation of `KV` usage within the `KV graveler` from the previous item. Tasks to be defined as part of [#3568](https://github.com/treeverse/lakeFS/issues/3568)
```
itaiad200, line 84: Why not define them now? They're the most important part of this plan!

itaiad200 (changes requested):
Almost there

design/open/kv3-execution-plan.md
```diff
@@ -8,21 +8,45 @@ Graveler uses 5 DB tables:
...
+
+## Entities
+### Repository
+A repository is identified by a unique name (type `RepositoryID`, a `string`) and contains its storage name space (type `StorageNamespace`, a string), default branch ID (type `BranchID`, a string) and its creation date (type `Time`).
+Repositories are iterable, and so all repositories should exist in a common partition (`graveler`)
+As this design suggests, among other things, that each repository should be a represented in a dedicated partition and that all other repository related entities will exist under this partition, and since we would like to use the existence/inexistence of a repository entry, to identify a valid/deleted repository, respectively, the repository entry should also include a random `guid`, which, when concatenated with the repo name, will be the partition key for that repository entities. That way it will be easy to distinguish between a partition related to "aRepo" that was already deleted and a partition related to a new "aRepo" that currently exist, as their partition names differ by the `guid` (`aRepo-guid1` vs `eRepo-guid2`).
```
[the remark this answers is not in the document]
itai-david: You perform `GetRepository` by `aRepo`. that will be translated to `kv.Get(partition = "graveler", key = "repos/aRepo")
Upon success, you will get a `Repository` object that has, among other things, a `my_guid` field set to "guid2"
This value will be autogenerated upon repository creation and will be initially used to create the repository with the default branch and initial commit
Does this make sense?

design/open/kv3-execution-plan.md
```diff
@@ -8,21 +8,45 @@ Graveler uses 5 DB tables:
...
+**Update:** It looks like most (if not all) `graveler`'s operations, perform a repository existence check by issuing a call to `GetRepository`, which performs a DB access. If that is indeed the case, we can easily "mark" a repository as deleted by removing the repository entry from the common partition `graveler`, thus making it unreachable. We can later iterate all entries under this repository partition and delete them one by one, preceding the branch entry deletion by iteratively deleting all relevant staging partitions. Failures can be handled by a background garbage collector (which, BTW, can handle the initial deletion too), but that is only for space saving purposes and should have no effect on the deleted repository itself, once the repository entry is deleted. A point to consider - what happens if a repository entry is deleted, the data in the relevant repository partition fails to delete, and a new repository with the same name is created? I think this is solvable by adding a guid to the repository and use it as the partition key (as suggested in the `Repository` entity above)
```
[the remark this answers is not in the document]
itai-david: Fixed

itai-david:
Added #3590 (out of KV3 scope) for failed repository deletion cleanup

itai-david:
> Approved, just need to break down `step 2` to concrete tasks.

Thank you
Created followup implementation tasks:
#3591 - for repository support (except for deletion, which is handled in #3590)
#3592 - for branches support, including by-commit-id iterator
#3593 - tags implementation
#3594 - commit implementation - storing and supporting a commit in our `KV` store - not about the commit flow
#3595 - staging implementation - supporting a staged object in `KV` - not the staging flow

design/open/kv3-execution-plan.md
itai-david: Done

design/open/kv3-execution-plan.md
itai-david: Eventually figured it out. By Ariel's suggestion, I was referring to [this comment](https://github.com/treeverse/lakeFS/pull/3566#discussion_r907073940) and you were referring to #3091
Modified the doc to clarify that

design/open/kv3-execution-plan.md
```diff
@@ -0,0 +1,98 @@
...
+Having that said, it is still essential to verify and agree that each data access essentially implies fetch of the repository entry, in order to (a)verify its existence and (b) get the correct `guid` in order to build the partition key
+### Branch
+A branch is identified by its unique name **within** a repository, hance the combination `<REPO_NAME>`+`<BRANCH_NAME>` uniquely identifies a branch in the system. Since all repository entities (except for the aforementioned Repository entity itself) belong to the repository partition, the Branch will be identified by its unique name within that partition.
+Other than that, a branch contains its current commit ID (type `CommitID`, a string) and its current staging token (type `StagingToken` a string). 
+In order to support the `commit flow`, proposed in [lakeFS on KV design](https://github.com/treeverse/lakeFS/blob/b3204edad00f88f8eb98524ad940fde96e02ab0a/design/open/metadata_kv/index.md#committer-flow), a branch will also have the propsed `sealed_token` field - an array of `StagingToken`s, as specified in the design doc
```
N-o-Z, line 22:
```suggestion
In order to support the `commit flow`, proposed in [lakeFS on KV design](https://github.com/treeverse/lakeFS/blob/b3204edad00f88f8eb98524ad940fde96e02ab0a/design/open/metadata_kv/index.md#committer-flow), a branch will also have the propsed `sealed_tokens` field - an array of `StagingToken`s, as specified in the design doc
```

N-o-Z (changes requested):
Thanks @itai-david. It's really hard to understand the data models for each entity as a textual description. I think it will be much more readable if you just model it

design/open/kv3-execution-plan.md
arielshaqed: I think that we might be overworking this.  The fact that something is efficiently possible on PostgreSQL does not mean that we have to do it efficiently in the KV world and in the same way.

Also please note that the proposal to add parameters to `Next()` simply makes it _not an iterator_.  I would argue that the iterator abstraction adds nothing here.

We will not be able to manage this list of "not first-parent commits" correctly and efficiently.  The question is really, do we need to?  We will probably be scanning all commits, or all commits up to a previously-processed set of "safe" commits, on each GC.  We could process not-first commits "just" by scanning all of these.  (I am starting to think that the commits logic for GC should _also_ move off of lakeFS and into the Spark cluster.  @johnnyaug WDYT?)

design/open/kv3-execution-plan.md
```diff
@@ -0,0 +1,98 @@
...
+  * Step 1: Building a initial `KV graveler` and a `DB graveler`. `KV graveler` should delegate all calls to `DB gravler` (no `KV` usage at all for this step) in order to allow `KV` implementation by the block - [#3568](https://github.com/treeverse/lakeFS/issues/3568)
+  * Step 2: Basic implementation of `KV` usage within the `KV graveler` from the previous item:
+    * Repository support (except for deletion) - [#3591](https://github.com/treeverse/lakeFS/issues/3591)
+    * Branch support - [#3592](https://github.com/treeverse/lakeFS/issues/3592)
+    * Tags support - [#3593](https://github.com/treeverse/lakeFS/issues/3593)
+    * Commits support (not including optimistic locking commit flow) - [#3594](https://github.com/treeverse/lakeFS/issues/3594)
+    * Staging support - [#3595](https://github.com/treeverse/lakeFS/issues/3595)
+  * Step 3: Optimistic Lock Commits, as described above - a working commit algorithm that prevents commits from overriding each other, by using `KV`'s `SetIf` to set the branch's commit, and failing (and retrying) to commit if `SetIf` fails, as described above. [#3569](https://github.com/treeverse/lakeFS/issues/3569)
```
arielshaqed, line 90: Consider retrying commits on the client side.  It will make it easier to provide different retry strategies for different clients.

design/open/kv3-execution-plan.md
```diff
@@ -0,0 +1,98 @@
...
+  * Step 4: Decide and implement a working solution for **Deleting a Repository** - [#3570](https://github.com/treeverse/lakeFS/issues/3570)
+  (Steps 3 and 4 are independent)
+* Benchmarks for common operations - to be executed manually, on both `SQL` and `KV` to verify performance are satisfactory - [#3571](https://github.com/treeverse/lakeFS/issues/3571)
+* Migration of existing `SQL` data to `KV`. Most of the tables are trivial to migrate. I believe `graveler_branches` migration should also be trivial, with no specific consideration for `sealed_tokens` but that needs to be verified - [#3572](https://github.com/treeverse/lakeFS/issues/3572)
+  * KV Migration Test is, obviously, included
+* Add `graveler` migration to `Esti` Migration Tests - [#3573](https://github.com/treeverse/lakeFS/issues/3573)
+* Optional - Dump/Restore Refs tests for `SQL` and `KV` - [#3574](https://github.com/treeverse/lakeFS/issues/3574)
+* Optional - `pkg/disagnostics` support of KV. Should be trivial - [#3575](https://github.com/treeverse/lakeFS/issues/3575)
```
arielshaqed, line 98:
```suggestion
* Optional - `pkg/diagnostics` support of KV. Should be trivial - [#3575](https://github.com/treeverse/lakeFS/issues/3575)
```

design/open/kv3-execution-plan.md
```diff
@@ -0,0 +1,98 @@
...
+Lock-free commits, as described in [Change KV storage proposal to guarantee progress](https://github.com/treeverse/lakeFS/pull/3091), will be handled out of the scope of KV3 milestone
+
+# Execution Plan
+* Agree on keys schema (see here after) - [#3567](https://github.com/treeverse/lakeFS/issues/3567)
+* Supporting `KV` along side `SQL` (same as was done for previous modules)
+  * Step 1: Building a initial `KV graveler` and a `DB graveler`. `KV graveler` should delegate all calls to `DB gravler` (no `KV` usage at all for this step) in order to allow `KV` implementation by the block - [#3568](https://github.com/treeverse/lakeFS/issues/3568)
```
arielshaqed, line 83:
```suggestion
  * Step 1: Building a initial `KV graveler` and a `DB graveler`. `KV graveler` should delegate all calls to `DB graveler` (no `KV` usage at all for this step) in order to allow `KV` implementation by the block - [#3568](https://github.com/treeverse/lakeFS/issues/3568)
```

arielshaqed (commented):
Thanks!

This PR mentions GUIDs, which AFAIK is just another term for UUID.  I cannot overstate how much I dislike these: they are a really long representation, hard for me to identify and distinguish by sight, support _versions_ that are pretty much useless, _require_ a particular [text format which is meaningless for the version commonly used](https://en.wikipedia.org/wiki/Universally_unique_identifier#Format), and cannot be configured to decrease or increase collision resistance.

I would much prefer to use a nanoid. :-)

design/open/kv3-execution-plan.md
johnnyaug: I'm totally open to it - the not-first-parent thing is a means to an end. We just need to start scanning the tree somewhere.

design/open/kv3-execution-plan.md
itai-david: @arielshaqed - are you referring to explicit retries by the user, or implicit retries by the (various) client? I believe it is the first?

design/open/kv3-execution-plan.md
itai-david: I'm aiming at preserving the current behavior with KV support, as a first step, and improve that later, based on our needs. I think introducing the KV aspect is enough of a load already, and would like to stay away from redesigning our GC as a side effect.
Even though it will not be as efficient, I think it can work, at least as a first step. Moreover, since it is the GC, I think we can loosen up the efficiency requirements, again - at least as a first step.
@arielshaqed @johnnyaug - WDYT?

itai-david: @arielshaqed - either ways we are not changing the interface of `Next()`

itai-david:
> Thanks!
> 
> This PR mentions GUIDs, which AFAIK is just another term for UUID. I cannot overstate how much I dislike these: they are a really long representation, hard for me to identify and distinguish by sight, support _versions_ that are pretty much useless, _require_ a particular [text format which is meaningless for the version commonly used](https://en.wikipedia.org/wiki/Universally_unique_identifier#Format), and cannot be configured to decrease or increase collision resistance.
> 
> I would much prefer to use a nanoid. :-)

Thanks @arielshaqed 
In general, I have nothing against using `nanoid` (to be sincere, I have nothing against `uuid` either). What about `uuid`s we already use in our staging token - do you want to change these too?

itai-david:
> Thanks @itai-david. It's really hard to understand the data models for each entity as a textual description. I think it will be much more readable if you just model it

@N-o-Z the thing is that other than the new `guid` field for the repository, and the `sealed_tokens` array for the branch - the model is the same as used by the current graveler. The current entities does not contain any `DB` related fields that need to change (as we saw in `auth`). In fact, `graveler` already uses the `<ENTITY>Record` types, which are essentially the same wrappers we implemented in `auth` (the data entity plus its ID in DB). As such, "coding" this model into the doc is pretty much copying the code, which feels redundant.
WDYT? Do you still think it is essential?

design/open/kv3-execution-plan.md
itaiad200: @itai-david I understood that we will be reading everything to memory, applying the same logic that was used in Postgres, thus implementing the same iterators as today. If so, what are the remaining open questions?

itai-david:
@N-o-Z - added the agreed code snippets. Please re-review

design/open/kv3-execution-plan.md
```diff
@@ -14,20 +14,61 @@ Note: All the above tables (along with the late `auth_users`) are also used by `
...
+Having that said, it is still essential to verify and agree that each data access essentially implies fetch of the repository entry, in order to (a)verify its existence and (b) get the correct `UniqueID` in order to build the partition key
+```go
+type RepositoryToken string
+
+type Repository struct { 
+  StorageNamespace  StorageNamespace  `db:"storage_namespace" json:"storage_namespace"` 
+  CreationDate      time.Time         `db:"creation_date" json:"creation_date"`
+  DefaultBranchID   BranchID          `db:"default_branch" json:"default_branch"`
```
N-o-Z, line 25: Need to consider using pointers for both DefaultBranchID and RepositoryToken as we need to support bare repositories

itai-david: Bare repos has the same default branch logic as non-bare repos (i.e. they "hold" the default branch ID == name, upon creation). The difference is the branch object is not really created. I think it would be better to leave it as is, and apply the same logic

PR #3576, opened 2022-06-25
Make logout separate from API

johnnyaug:
Breaking change: if (for some reason) you're using the API with cookie-based sessions, the logout endpoint is now not a part of the lakeFS OpenAPI spec.

## Explanation

What this endpoint does is clear the cookie-based session. Cookie sessions are intended for use by the UI, so it makes more sense for it to be a page, to which we redirect the user when they click the logout button.


pkg/config/template.go
```diff
@@ -90,8 +90,9 @@ type configuration struct {
 			Endpoint string
 			Token    string
 		}
-		LDAP *LDAP
-		OIDC OIDC
+		LDAP              *LDAP
+		OIDC              OIDC
+		LogoutRedirectURL string `mapstructure:"logout_redirect_url" validate:"logout_redirect_url"`
```
arielshaqed, line 95:
```suggestion
		LogoutRedirectURL string `mapstructure:"logout_redirect_url"`
```

The _only_ supported validation is `validate:"required"`.  I think that this will be silently ignored.  (Reflection is a really nice way of hiding stuff...)

arielshaqed (changes requested):
Worried about the redirect, and about some tag in config that I do not understand.

pkg/config/template.go
johnnyaug: oops

johnnyaug:
Fix: removed option for user to enter redirect URL.

arielshaqed (approved):
Thanks!

PR #3554, opened 2022-06-23
GC: Precisely explain what objects are retained

arielshaqed:

Issue #3107, opened 2022-03-23
Docs: unify "Branching model" and "Object model" sections


johnnyaug:
_[Object model](https://docs.lakefs.io/understand/object-model.html)_ all but contains _[Branching model](https://docs.lakefs.io/understand/branching-model.html)_.

Both explain the basic entities of lakeFS - while object models dives deeper into syntax and advanced concepts.

PR #3577, opened 2022-06-27
New documentation structure and content changes

arouyan:
As per the agreed new structure, I moved pages to new folders (+ created redirects), created new index pages, and proofread the content of all the existing pages

### DIFF
--- a/cmd/lakefs-loadtest/cmd/entry.go
+++ b/cmd/lakefs-loadtest/cmd/entry.go
@@ -16,6 +16,8 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/treeverse/lakefs/pkg/catalog"
 	"github.com/treeverse/lakefs/pkg/config"
+	"github.com/treeverse/lakefs/pkg/kv"
+	"github.com/treeverse/lakefs/pkg/logging"
 	"github.com/treeverse/lakefs/pkg/uri"
 )
 
@@ -62,10 +64,24 @@ var entryCmd = &cobra.Command{
 		if err != nil {
 			fmt.Printf("invalid config: %s\n", err)
 		}
+
+		var storeMessage *kv.StoreMessage
+		dbParams := conf.GetDatabaseParams()
+		if dbParams.KVEnabled {
+			kvParams := conf.GetKVParams()
+			kvStore, err := kv.Open(ctx, dbParams.Type, kvParams)
+			if err != nil {
+				logging.Default().WithError(err).Fatal("failed to open KV store")
+			}
+			defer kvStore.Close()
+			storeMessage = &kv.StoreMessage{Store: kvStore}
+		}
+
 		c, err := catalog.New(ctx, catalog.Config{
-			Config: conf,
-			DB:     database,
-			LockDB: lockDB,
+			Config:  conf,
+			DB:      database,
+			LockDB:  lockDB,
+			KVStore: storeMessage,
 		})
 		if err != nil {
 			fmt.Printf("Cannot create catalog: %s\n", err)
--- a/cmd/lakefs/cmd/import.go
+++ b/cmd/lakefs/cmd/import.go
@@ -85,17 +85,6 @@ func runImport(cmd *cobra.Command, args []string) (statusCode int) {
 	}
 	logger := logging.FromContext(ctx)
 
-	catalogCfg := catalog.Config{
-		Config: cfg,
-		DB:     dbPool,
-	}
-	c, err := catalog.New(ctx, catalogCfg)
-	if err != nil {
-		fmt.Printf("Failed to create c: %s\n", err)
-		return 1
-	}
-	defer func() { _ = c.Close() }()
-
 	u := uri.Must(uri.Parse(args[0]))
 	if !u.IsRepository() {
 		fmt.Printf("Invalid 'repository': %s\n", uri.ErrInvalidRefURI)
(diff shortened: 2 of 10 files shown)


docs/guides/index.md
```diff
@@ -0,0 +1,36 @@
...
+
+This guide shows which metrics lakeFS exposes, a selection of example queries, and example of a Grafana dashboard, and how to make it all work.
+
+[Learn more](https://docs.lakefs.io/reference/monitor.html)
+
+## Authentication and authorization
+
+lakeFS authenticates users from a built-in authentication database, or optionally from a configured LDAP server. Explore how authentication and authorization work in lakeFS.
+
+[Learn more](https://docs.lakefs.io/reference/authentication.html)
```
arielshaqed, line 23: I don't understand, I think this PR also **deletes** this file?!

docs/guides/index.md
```diff
@@ -0,0 +1,36 @@
...
+
+## Monitoring with Prometheus
+
+This guide shows which metrics lakeFS exposes, a selection of example queries, and example of a Grafana dashboard, and how to make it all work.
+
+[Learn more](https://docs.lakefs.io/reference/monitor.html)
+
+## Authentication and authorization
+
+lakeFS authenticates users from a built-in authentication database, or optionally from a configured LDAP server. Explore how authentication and authorization work in lakeFS.
```
arielshaqed, line 21: I guess @johnnyaug this is more text that needs to be updated to discuss OIDC.

docs/guides/index.md
```diff
@@ -0,0 +1,36 @@
...
+
+[Learn more](https://docs.lakefs.io/reference/authentication.html)
+
+## Upgrading lakeFS
+
+Upgrading lakeFS usually requires re-deploying with the latest image or downloading the latest version, if you’re using the binary.
+
+[Learn more](https://docs.lakefs.io/reference/upgrade.html)
```
arielshaqed, line 29: I don't understand this either.  I think this PR **renames** this file and will break this link.

docs/under the hood/garbage-collection.md
```diff
@@ -36,60 +38,17 @@ That is, it is hard-deleted only after the retention period has ended for all re
 Example GC rules for a repository:
 ```json
 {
-  "default_retention_days": 14,
+  "default_retention_days": 21,
```
arielshaqed, line 41: Why this change?

Also, this deletes #3554 and its two fixes #3561, #3562.  I assume that this is by accident!  Note that it is hard to perform a huge change on documentation when it is concurrently being updated by others.

arielshaqed (changes requested):
Nit: Do we need this empty file docs/reference/untitled?

I am somewhat worried about this PR, probably more so because it is so large.  It would be much easier to review and to test and verify if the separate commits were reworked to have semantic content (e.g. a single "upgrade" commit), or at least if commit titles were a bit more informative.

Specific issues:
* Renames a whole bunch of files and seems to be somewhat inconsistent; I think I found 2 examples of links that will break.
* All links changed to be absolute.  This breaks the docs version selector (@nopcoder can give more details).
* Deletes the improved explanation of garbage collection (see inline comments)
* Might ignore _other_ documentation updates that may have occurred.

Nits:
* Every "Learn more" link is missing a period at the end of the sentence.  I would probably follow the instructions in @adipolak 's [how to write great documentation](https://docs.google.com/document/d/1vif05nDiKZYPwYdINLoq4BJ6j7bBIQ6ZA2dt7McuzyY/edit#heading=h.n67ifx7rxp8w) documentation for "how to create a list", or just keep the convention of punctuation at the end of each sentence.
* You can (and probably should) link to the `.md` versions of files rather than their `.html` versions.  It all processes to the same HTML, but giving the correct filename makes it easier (for me) to find the correct file to edit.

docs/under the hood/garbage-collection.md
arouyan: I've been editing files locally but using Github Desktop, I made sure to pull updates regularly and right before pushing changes. Perhaps something didn't work. I guess the best way to go forward would be to reject my change - I'm happy to go over the file later on (for editing/proofreading)

arouyan:
> Nit: Do we need this empty file docs/reference/untitled?
> 
> I am somewhat worried about this PR, probably more so because it is so large. It would be much easier to review and to test and verify if the separate commits were reworked to have semantic content (e.g. a single "upgrade" commit), or at least if commit titles were a bit more informative.
> 
> Specific issues:
> 
> * Renames a whole bunch of files and seems to be somewhat inconsistent; I think I found 2 examples of links that will break.
> * All links changed to be absolute.  This breaks the docs version selector (@nopcoder can give more details).
> * Deletes the improved explanation of garbage collection (see inline comments)
> * Might ignore _other_ documentation updates that may have occurred.
>

Hi - thank you for pointing these issues out. I now see that combining it all into a massive pull request made things difficult - if there's a way to break the changes I applied into smaller parts, I'm happy to do so. 

> Nits:
> 
> * Every "Learn more" link is missing a period at the end of the sentence.  I would probably follow the instructions in @adipolak 's [how to write great documentation](https://docs.google.com/document/d/1vif05nDiKZYPwYdINLoq4BJ6j7bBIQ6ZA2dt7McuzyY/edit#heading=h.n67ifx7rxp8w) documentation for "how to create a list", or just keep the convention of punctuation at the end of each sentence.

I can correct this. Just to clarify, you mean "Learn more." instead of "Learn more"?

> * You can (and probably should) link to the `.md` versions of files rather than their `.html` versions.  It all processes to the same HTML, but giving the correct filename makes it easier (for me) to find the correct file to edit.

I've been told to link to html versions, but I can of course change that (this is a problem that ocurrs in index pages mainly since these are the ones I created from scratch).

johnnyaug (changes requested):
Comments regarding the content:
1. We now have both "Getting started" and "Quickstart". The names are very similar and they contain very different conent.
2. Neither of the items under "Getting started" actually help me get started with lakeFS.
3. Under "Getting started" we have Using lakeFS - which is also a category of its own below, which is strange.
4. The "Use cases" category contains very similar content to "Using lakeFS" category below.
5. No reason to have both "Under the Hood" and "Understand lakeFS", what's the difference?
6. Actually, I wouldn't say that any of the items under "Under the Hood" are actually describing anything under the hood, except "object model" which is almost the same as "branching model" and [needs to be unified with it](https://github.com/treeverse/lakeFS/issues/3107). 
7. Migrating away from lakeFS appears twice.
8. I think everything under "Guides" can move under "Set up".

PR #3578, opened 2022-06-27
Fix/remove ids from auth models 3521

N-o-Z:


pkg/auth/model/model.go
```diff
@@ -101,52 +101,35 @@ type BaseUser struct {
...
 }
 
-type User struct {
-	ID string
-	BaseUser
-}
-
 type DBUser struct {
 	ID int64 `db:"id"`
 	BaseUser
```
itaiad200, line 106: I suggest a rename just like with `Group` and `Policy`
```suggestion
	User
```

itaiad200 (approved):
LGTM with some comments. Please wait for the CN team to review it too

pkg/auth/model/model.go
```diff
@@ -217,23 +201,19 @@ func (s *Statements) Scan(src interface{}) error {
...
-			Source:            pb.Source,
-		},
+func UserFromProto(pb *UserData) *BaseUser {
+	return &BaseUser{
+		CreatedAt:         pb.CreatedAt.AsTime(),
+		Username:          pb.Username,
+		FriendlyName:      &pb.FriendlyName,
+		Email:             &pb.Email,
+		EncryptedPassword: pb.EncryptedPassword,
+		Source:            pb.Source,
```
ortz, line 211: The property ExternalID is missing here compared to the BaseUser

N-o-Z: Nice catch! :)

pkg/auth/authenticator.go
```diff
@@ -218,7 +218,7 @@ func (la *LDAPAuthenticator) AuthenticateUser(ctx context.Context, username, pas
 	user, err := la.AuthService.GetUser(ctx, dn)
 	if err == nil {
 		logger.WithField("user", fmt.Sprintf("%+v", user)).Debug("Got existing user")
-		return user.ID, nil
+		return user.Username, nil
```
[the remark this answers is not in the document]
N-o-Z: @ortz You are right, we need to change the way we hold the user credentials to keep the username instead/ in addition to the user id and work with username only in the authenticators

N-o-Z:
@guy-har  Would appreciate your review on this

guy-har (changes requested):
Great change :)
Requesting changes due to bug in `middleware.go`
One thing we should be aware of is that updating to this version will log out all users that are currently logged in (maybe, we should just remember to write it down in the changelog).

guy-har (approved):

N-o-Z:
@guy-har  Thanks! Though I still need to understand why TestMigrate fails (probably because of caching?)

PR #3579, opened 2022-06-27
Avoiding of printing DB password on connecting

idanovo:


johnnyaug (commented):
Yep, that was needed.  I would add the other non secret details as well, like user, host and port.

PR #3583, opened 2022-06-28
KV configuration block

### LINKED ISSUE #3505: Allow configuration block for any KV implementation
We currently use a single dsn string to fit them all. That doesn't serve us anymore because:
1. Not all KV implementations configuration are easily mapped to a single string.
2. There are cases where we need more than 1, for example during migration.

itaiad200:
closes #3505

### DIFF
--- a/cmd/lakefs/cmd/import.go
+++ b/cmd/lakefs/cmd/import.go
@@ -125,7 +125,8 @@ func runImport(cmd *cobra.Command, args []string) (statusCode int) {
 	var idGen actions.IDGenerator
 	var actionsStore actions.Store
 	if dbParams.KVEnabled {
-		kvStore, err := kv.Open(ctx, dbParams.Type, dbParams.ConnectionString)
+		kvparams := cfg.GetKVParams()
+		kvStore, err := kv.Open(ctx, dbParams.Type, kvparams)
 		if err != nil {
 			logger.WithError(err).Fatal("failed to open KV store")
 		}
--- a/pkg/kv/params/database.go
+++ b/pkg/kv/params/database.go
@@ -1,0 +1,38 @@
+package params
+
+type KV struct {
+	Postgres *Postgres
+	DynamoDB *DynamoDB
+}
+
+type Postgres struct {
+	ConnectionString string
+}
+
+type DynamoDB struct {
+	// The name of the DynamoDB table to be used as KV
+	TableName string
+
+	// Table provisioned throughput parameters, as described in
+	// https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Limits.html
+	ReadCapacityUnits  int64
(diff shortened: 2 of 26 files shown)


pkg/kv/params/database.go
```diff
@@ -0,0 +1,38 @@
+package params
+
+type KV struct {
+	Postgres *Postgres
+	DynamoDB *DynamoDB
+}
```
itai-david, line 6: Very elegant :+1:

pkg/kv/store.go
```diff
@@ -146,14 +148,14 @@ func UnregisterAllDrivers() {
 
 // Open lookup driver with 'name' and return Store based on 'dsn' (data source name).
 // Failed with ErrUnknownDriver in case 'name' is not registered
-func Open(ctx context.Context, name, dsn string) (Store, error) {
+func Open(ctx context.Context, name string, params kvparams.KV) (Store, error) {
 	driversMu.RLock()
 	d, ok := drivers[name]
 	driversMu.RUnlock()
```
itai-david, line 154: I know it's not part of your PR, but it caught my eye
```suggestion
	defer driversMu.RUnlock()
```

cmd/lakefs/cmd/import.go
```diff
@@ -125,7 +125,8 @@ func runImport(cmd *cobra.Command, args []string) (statusCode int) {
 	var idGen actions.IDGenerator
 	var actionsStore actions.Store
 	if dbParams.KVEnabled {
-		kvStore, err := kv.Open(ctx, dbParams.Type, dbParams.ConnectionString)
+		kvparams := cfg.GetKVParams()
+		kvStore, err := kv.Open(ctx, dbParams.Type, kvparams)
```
itai-david, line 129: Just a thought:
How about the option to send the entire `cfg` to `kv.Open` and it perfrom the `cfg.GetKVParams()`?
On a logical level it means that KV is "aware" of the entire `config` package and that it (`cfg`) has a designated section (`kvparams`) for its ('KV') use, which makes sense, and that the `cmd` package, does not need to be aware of `kvparams` which also makes sense (especially if and when `kvparams` will change)

itai-david (approved):
Looks great - clear and simple.
One thought regarding the right location to perform the `kvparams` extraction from `cfg`. That's just an idea and up to you whether to consider it or not. Not a change request

cmd/lakefs/cmd/import.go
itaiad200: I think it's better that the wiring happens and a single place. Also `kv` doesn't have a dependency on the `config` pkg, it does have a dependency on its inner `params` package. That's the same pattern the `db` package uses and without clear advantages to any solution, I rather keep it consistent.

Issue #3548, opened 2022-06-23
TTV: Extract import functionality


Jonathan-Rosenberg:
Extract the import's functionality in order to use it the Import step

Story: https://github.com/treeverse/lakeFS/issues/3539

PR #3589, opened 2022-06-29
TTV: GUI- extract import functionality

Jonathan-Rosenberg:
Extract the import functionality to be used in the TTV wizard.

Closes: #3548

### DIFF
--- a/webui/src/pages/repositories/repository/objects.jsx
+++ b/webui/src/pages/repositories/repository/objects.jsx
@@ -106,7 +106,6 @@ const ImportButton = ({ config, repo, reference, path, onDone, onClick, variant 
         }
         try {
             await runImport(updateStateFromImport,
-                initialState.isSourceValid,
                 `${destRef.current.value}`,
                 commitMsgRef.current.value,
                 sourceRef.current.value,
--- a/webui/src/pages/repositories/repository/objects.jsx
+++ b/webui/src/pages/repositories/repository/objects.jsx
@@ -106,7 +106,7 @@ const ImportButton = ({ config, repo, reference, path, onDone, onClick, variant 
         }
         try {
             await runImport(updateStateFromImport,
-                `${destRef.current.value}`,
+                destRef.current.value,
                 commitMsgRef.current.value,
                 sourceRef.current.value,
                 importBranch,
--- a/webui/src/pages/repositories/repository/objects.jsx
+++ b/webui/src/pages/repositories/repository/objects.jsx
@@ -3,23 +3,16 @@ import React, {useRef, useState} from "react";
 import {UploadIcon} from "@primer/octicons-react";
 import {RepositoryPageLayout} from "../../../lib/components/repository/layout";
 import RefDropdown from "../../../lib/components/repository/refDropdown";
-import {
-    ActionGroup,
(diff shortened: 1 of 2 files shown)


eden-ohana (approved):
LGTM

webui/src/pages/repositories/repository/objects.jsx
```diff
@@ -107,52 +101,19 @@ const ImportButton = ({ config, repo, reference, path, onDone, onClick, variant
...
-            onDone()
+            await runImport(updateStateFromImport,
+                `${destRef.current.value}`,
+                commitMsgRef.current.value,
+                sourceRef.current.value,
+                importBranch,
+                repo.id,
+                reference.id
+            );
+            onDone();
```
arielshaqed, line 116: (Not introduced in this PR but possibly relevant...)

Is `onDone` a synchronous call, or should this have been `await onDone()`?  In particular, if it's asynchronous, then IIRC both this _and the previous version_ lose exceptions that may occur while it runs.

arielshaqed (approved):
Thanks!

webui/src/pages/repositories/repository/objects.jsx
```diff
@@ -107,52 +101,19 @@ const ImportButton = ({ config, repo, reference, path, onDone, onClick, variant
...
-                    setImportState({...initialState, error})
-                    throw error
-                }
-            }
-
-            await commits.commit(repo.id, importBranchResp.id, commitMsg, {}, metarange.id)
-            setImportState({inProgress: false, error: null, done: true, numObj: sum, isSourceValid: importState.isSourceValid})
-            onDone()
+            await runImport(updateStateFromImport,
+                `${destRef.current.value}`,
```
arielshaqed, line 109:
```suggestion
                destRef.current.value,
```

webui/src/pages/repositories/repository/objects.jsx
```diff
@@ -107,52 +101,19 @@ const ImportButton = ({ config, repo, reference, path, onDone, onClick, variant
...
-                } else {
-                    setImportState({...initialState, error})
-                    throw error
-                }
-            }
-
-            await commits.commit(repo.id, importBranchResp.id, commitMsg, {}, metarange.id)
-            setImportState({inProgress: false, error: null, done: true, numObj: sum, isSourceValid: importState.isSourceValid})
-            onDone()
+            await runImport(updateStateFromImport,
```
arielshaqed, line 108: What happens if I reload during an import?

arielshaqed (commented):
Thanks!...

webui/src/pages/repositories/repository/objects.jsx
Jonathan-Rosenberg: then only the ranges that were created in the previous calls will exist. If no metarange was created, those ranges are gone.

webui/src/pages/repositories/repository/objects.jsx
Jonathan-Rosenberg: `onDone` is a method that's passed from outside.. Don't think we should enforce a wait on it.
Nonetheless, the passed `onDone` function is just a state update function (toggles a boolean)

Issue #3520, opened 2022-06-15
`auth.MetadataManager` KV support


itai-david:
Follow up task from #3492 

* implement KV support for metadata
  * need to define the key format
* add to auth migration function and kv_migration tests
* drop the table at the end of migration

itai-david:
For the current DB implementation, we are saving the DB metadata itself - `postgresql`(or `pg aurora`) version and various DB settings such as TZ, work memory and rds info. This is easily supported by the `pg_settings` table.
Do we want/need something similar for `KV`?
I think it can wait for a later step, but still, need to be decided

PR #3598, opened 2022-07-03
Add dynamoDB Esti test

### LINKED ISSUE #3441: DynamoDB KV Support - Integrate with Esti tests

itaiad200:
closes #3441


itaiad200:
I guess we must complete #3520 for this new test to pass.

webui/package-lock.json
```diff
@@ -14,14 +14,15 @@
         "bootstrap": "^4.6.0",
```
N-o-Z, line 14: I think these got in due to the npm update, should be removed?

esti/ops/docker-compose-dynamodb.yaml
```diff
@@ -0,0 +1,88 @@
+version: '3'
```
N-o-Z, line 1: We are running esti tests on local dynamodb instance? Not sure it will hold when all packages are migrated...

N-o-Z (changes requested):
Looks good! Though the database config params are starting to be confusing.

esti/ops/docker-compose-dynamodb.yaml
itaiad200: In terms of what? Throughput? Consistency guarantees? Shouldn't we wait to figure it out?

webui/package-lock.json
itaiad200: Build is failing without it 🤷

esti/ops/docker-compose-dynamodb.yaml
itaiad200: I'm also reluctant to pay for the real thing for our CI if it's not absolutely a must

N-o-Z (approved):
LGTM! 😎

itaiad200:
> Looks good! Though the database config params are starting to be confusing.

They didn't change in this PR, just used for the system tests. I guess we can drop some values until we'll understand it better.

N-o-Z:
> > Looks good! Though the database config params are starting to be confusing.
> 
> They didn't change in this PR, just used for the system tests. I guess we can drop some values until we'll understand it better.

I didn't imply they changed - I just meant the mix of the legacy env vars and the new configuration structure can lead to some confusion. We might want to align it at some point

Issue #3540, opened 2022-06-22
TTV: Operative Templating service


Jonathan-Rosenberg:
Definition of done:
Templating service is functional and returns an expanded template according to the design.

[Template Server design](https://github.com/treeverse/lakeFS/blob/master/design/open/template-server/template-server.md)
[TTV design](https://github.com/treeverse/lakeFS/blob/master/design/open/ttv-wizard-mvp.md)

arielshaqed:
Done by #3600 and #3606.

PR #3600, opened 2022-07-03
Add a template server

arielshaqed:
See design at [template-server.md](https://github.com/treeverse/lakeFS/blob/master/design/open/template-server/template-server.md).

Has only the minimal functions required to build the MVP.

Implements #3540 .


arielshaqed:
# Guide to reviewing

Apologies for a large PR, but it actually may be easier to review than it seems at first.

* Please review by commits:
  - c935081de771ff383746eab44a5639f0ab03720c is the templater.
  - 7aa8754274ca3812a529d834258ae1190760c2ab hooks it up to the API controller.
  - 9786efaa339b0889c77684976e20d3effe38cb45 adds API tests.
  - Other commits are small.
* Ignore all generated files: everything under `java/` and `python/`, as well as the copy of `swagger.yml`.
* Ignore lint and similar errors that may appear on commits halfway through; there are 2 commits to fix those (and you can just rely on the fact that the linter action is green).

pkg/templater/expander.go
```diff
@@ -0,0 +1,157 @@
...
+	// Data is directly visible to templates for expansion, with no
+	// authorization required.
+	Data *UncontrolledData
+}
+
+// Expander is a template that may be expanded as requested by users.
+type Expander interface {
+	// Prepare checks whether the template expansion will succeed.  Call
+	// it before Expand to ensure nothing is written when expansion fails.
+	Prepare(params *Params) error
```
Jonathan-Rosenberg, line 61: Why do you need this two step operation?
Can't you use `Expand` to try and `Execute` to a temporary `bytes.Buffer{}`, return error if it doesn't succeed, and write to the given Writer if it does succeed?

pkg/templater/expander.go
```diff
@@ -0,0 +1,157 @@
...
+		if e == nil {
+			// Negative cache
+			return nil, ErrTemplateNotFound
+		}
+		return e, nil
+	}
+
+	// Compute path
+	p := path.Join(embeddedContentPrefix, name)
+	if !strings.HasPrefix(p, embeddedContentPrefix+"/") {
```
Jonathan-Rosenberg, line 136: cool

pkg/templater/expander.go
arielshaqed: We could, but I am not sure that it would be better than this.  One disadvantage: I worry that the buffer might be large.  As an example, suppose someone abused template to serve a large object (maybe even static, say an image).  Then we would hold it in process memory, which is not fun for the process.

Are there advantages other than simpler implementation?

pkg/templater/expander.go
arielshaqed: Not important for the embedded FS (which has _nowhere_ to go above its intended root), but obviously relevant as soon as we hook it up to somewhere else!

arielshaqed (commented):
Thanks!

I think we should really decide on 2-phase vs. 1-phase processing.  What's better about 1-phase, other than simpler implementation?

pkg/templater/funcmap.go
```diff
@@ -0,0 +1,42 @@
...
+	"fmt"
+	"reflect"
+	"text/template"
+)
+
+var ErrBadFuncRet = errors.New("wrong number of return values")
+
+// WrapFuncMapWithData wraps a funcMap to a FuncMap that passes data as a
+// first element when calling each element.
+func WrapFuncMapWithData(funcMap template.FuncMap, data interface{}) template.FuncMap {
```
Jonathan-Rosenberg, line 14: Trying to figure out why is it necessary.
The output of this function is an array of functions (FuncMap) that contains functions that hold some data (that would be the controlled data in the implementation), and that appends that "hold-data" to the given argument-data, then calls the underlying function on the total-data.
So a number of questions:
1. Is my understanding correct?
2. Why not initialize the template structure with the original FuncMap, and pass the total-data (controlled + uncontrolled) to Execute()?
3. If it's still necessary to wrap the functions, can we wrap those functions and values in functions instead of reflections to make it more understandable (and probably maintainable)?

By "wrapping with functions" I mean something like:
```go
ret[k] = func(args ...interface{}) (interface{}, error) {
    argVals := make([]interface{}, 0, len(args)+1)
    argVals = append(argVals, data)
    argVals = append(argVals, args...)
    return v(argVals)
}
```
(holding the controlled data and the underlying function)

api/swagger.yml
```diff
@@ -3516,3 +3516,35 @@ paths:
...
+              type: string
+      responses:
+        200:
+          description: expanded template
+          content:
+            "*/*":
+              schema:
+                format: binary
+        401:
+          $ref: "#/components/responses/Unauthorized"
```
Jonathan-Rosenberg, line 3546: What about 403 Forbidden?
The user might be authenticated but not authorized to fetch the template or use the requested function

Jonathan-Rosenberg (changes requested):
Really awesome!!!

Not requesting changes but rather some explanations (no such radio button yet)...

pkg/templater/expander.go
Jonathan-Rosenberg: I do think that simpler implementation (thus better maintainability) is important enough, but the concerns you're raising are more important.
Regardless you can still collapse it to a single step process and do the validation step as a first step of the expansion... I think that it's not the role of the client to check if the template is expandable or not.

api/swagger.yml
arielshaqed: For some reason we *never* specify 403 in this file.  @nopcoder do you know why?  In any case, I'm not blocking on this.

pkg/templater/expander.go
arielshaqed: Sounds good!  I shall rework the logic from `service.Expand` into a single call here.

pkg/templater/funcmap.go
arielshaqed: By points...

(1) Yes, this is what this function does.

(2) Execute() is somewhat broken: I could find no mechanism to pass side-data into functions.  The 2nd arg `data` does *not* work for this, because any exported members of it are visible not only to functions but also to the template being expanded.  And I am afraid of that access in templates, I have less control over them and they are a DSL that is difficult to analyze.  In particular note the template function `call`, and consider how much damage a user allowed to create new templates will be able to do with the auth service, if that is supplied on `data`.

(3) I am not sure why wrapping functions manually is less error-prone than performing it automatically.  I prefer to pay the added price of having to understand reflection _now_ but **once**, to having to check each definition of a function wrapper.  Also I don't think that your example will compile without redefining `v` to take all arguments as `interface{}`.  The way to do it would probably be to specify explicitly the types of all arguments to `v`, and just copy the argument list into the call to `v`.

arielshaqed (commented):
Thanks!

arielshaqed:
You're still requesting changes, so asking for a re-review :-)

templates/content/spark.conf.tt
```diff
@@ -0,0 +1,7 @@
+# SAMPLE TEMPLATE: minimal Spark lakeFSFS configuration
+
+spark.hadoop.fs.lakefs.impl=io.lakefs.LakeFSFileSystem
+{{with $creds := new_credentials}}
+spark.hadoop.fs.lakefs.access_key={{$creds.Key}}
+spark.hadoop.fs.lakefs.secret_key={{$creds.Secret}}
+{{end}}
```
Jonathan-Rosenberg, line 7:
```suggestion
spark.hadoop.fs.s3a.impl=org.apache.hadoop.fs.s3a.S3AFileSystem
{{with $creds := new_credentials}}
spark.hadoop.fs.s3a.access_key={{$creds.Key}}
spark.hadoop.fs.s3a.secret_key={{$creds.Secret}}
{{end}}
spark.hadoop.fs.s3a.endpoint={{ .querystring.lakefs_url }}
spark.hadoop.fs.s3a.path.style.access=true
```

Jonathan-Rosenberg (approved):

arielshaqed:
> Thanks! I think that the template itself should be changed...

Did that, which uncovered a **BUG** in OpenAPI code generation (and that the previous code never really worked).  Updated version includes the requested template _and_ works.

PTAL...

Jonathan-Rosenberg (approved):

arielshaqed:
Thanks!

PR #3601, opened 2022-07-04
Auth module in Esti migration test

itai-david:
Pre migration - creating users in various groups, attempting various operations and expecting success/failure based on the group permission and the operation (basic sanity)
Post migration - verifying that all the above is migrated successfully by retrying the same operation and expecting the same results. In addition, adding policy to a previously added group (created over `DB`) and expecting correct behavior

### DIFF
--- a/esti/migrate_test.go
+++ b/esti/migrate_test.go
@@ -148,7 +148,7 @@ func preMigrateTests(t *testing.T) {
 	// all pre tests execution
 	t.Run("TestPreMigrateMultipart", testPreMigrateMultipart)
 	t.Run("TestPreMigrateActions", testPreMigrateActions)
-	t.Run("TestPreMigrateAuth", TestPreMigrateAuth)
+	t.Run("TestPreMigrateAuth", testPreMigrateAuth)
 
 	saveStateInLakeFS(t)
 }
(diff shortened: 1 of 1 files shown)


esti/migrate_test.go
```diff
@@ -253,3 +336,209 @@ func testPostMigrateActions(t *testing.T) {
...
+	require.NoError(t, err, "Admin failed while adding policy to group")
+	require.Equal(t, http.StatusCreated, respAddPolicy.StatusCode(), "Admin unexpectedly failed to add policy to group")
+
+	uid := "test-user"
+	customCreds := createUserWithCredentialsInGroup(t, ctx, uid, authCustomGroupName)
+	verifyUserPermissions(t, ctx, repo, "customUser", customCreds, customPermissions)
+
+	// Hardening relevant test data for post-migrate
+	state.Auth.Repo = repo
+	state.Auth.ViewerUser.Username = "testViewer"
```
itaiad200, line 389: You don't have to pass const data to the state, you can simply use the same values from the post test

itaiad200 (approved):
Thanks for adding this meticulous test, I feel much better with our KV Auth implementation now that we have it 💪

esti/migrate_test.go
itai-david: Moreover, the `username` is not really use after the user creation (in `PreTest`) as the users are identified by the credentials. Removed the field all together

PR #3606, opened 2022-07-04
Add phase-dependent template function contenttype

arielshaqed:
* Make functions aware of phase ("prepare" or "expand").
* Add function to set Content-Type -- during "prepare" only.

  (Once expansion has started, writing headers will fail!)
* Fix `new_credentials` to create credentials just once.

  (Previously they were created once during "prepare" and once during "expand").

### DIFF
--- a/pkg/templater/expander.go
+++ b/pkg/templater/expander.go
@@ -43,7 +43,7 @@ type ControlledParams struct {
 	Phase Phase
 	Auth  AuthService
 	// Headers are the HTTP response headers.  They may only be modified
-	// during PhasePrepate.
+	// during PhasePrepare.
 	Header http.Header
 	// User is the user expanding the template.
 	User *auth_model.User
--- a/pkg/templater/expander.go
+++ b/pkg/templater/expander.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io"
 	"io/fs"
+	"net/http"
 	"path"
 	"strings"
 	"text/template"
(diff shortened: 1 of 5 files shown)


pkg/templater/expander.go
```diff
@@ -30,11 +31,25 @@ type AuthService interface {
 	auth.CredentialsCreator
 }
 
+type Phase string
+
+const (
+	PhasePrepare Phase = "prepare"
+	PhaseExpand  Phase = "expand"
```
Jonathan-Rosenberg, line 38: Still relevant?
I think that we agreed that the two steps will be squashed into one, no?

pkg/templater/service.go
```diff
@@ -5,12 +5,12 @@ import (
...
 
 	"context"
-	"io"
 	"io/fs"
+	"net/http"
 )
 
 type Service interface {
-	Expand(ctx context.Context, w io.Writer, user *model.User, templateName string, query map[string]string) error
+	Expand(ctx context.Context, w http.ResponseWriter, user *model.User, templateName string, query map[string]string) error
```
Jonathan-Rosenberg, line 13: Why?

arielshaqed: `contenttype` will set a header :-)

pkg/templater/expander.go
```diff
@@ -30,11 +31,25 @@ type AuthService interface {
...
+)
+
 type ControlledParams struct {
-	Ctx  context.Context
-	Auth AuthService
+	Ctx   context.Context
+	Phase Phase
+	Auth  AuthService
+	// Headers are the HTTP response headers.  They may only be modified
+	// during PhasePrepate.
```
Jonathan-Rosenberg, line 46:
```suggestion
	// during PhasePrepare.
```

pkg/templater/funcs.go
```diff
@@ -45,19 +49,48 @@ func newCredentials(params *ControlledParams) (*Credentials, error) {
 		return nil, fmt.Errorf("create credentials for %+v: %w", params.User, ErrNotAuthorized)
 	}
 
+	var (
+		generatedCredsI interface{}
+		ok              bool
+		generatedCreds  map[string]*Credentials
+	)
```
Jonathan-Rosenberg, line 56: AFAIK, this variable declaration is not idiomatic to Go. Why do we need to declare `generatedCredsI` and `ok`?
Is it because of the conversion in line 63?

Jonathan-Rosenberg (approved):

pkg/templater/funcs.go
arielshaqed: Yes :-(

arielshaqed: Paging @nopcoder: is there a better way??!?

arielshaqed:
Rebased onto new trunk; will pull once tests pass!

PR #3607, opened 2022-07-04
TTV: GUI Wizard component

### LINKED ISSUE #3543: TTV: Wizard component
Define the Wizard component's structure and properties.
The Wizard should allow its steps to pass information.
Steps should be able to read values that were passed from previously executed steps.

Story: #3539

Jonathan-Rosenberg:
This PR introduces the Wizard component.
I used (mainly) the [react-step-wizard](https://www.npmjs.com/package/react-step-wizard) package to implement the component.

### Description:
- `progressBar`: Should the wizard display a progress bar?
- `isSkipButton`: Should the Wizard include a skip button?
- `isGoBackButton`: Should the Wizard display a go back button?
- `onComplete`: A function to call once the wizard has completed.
- `canProceed`: A value that tells the wizard if it should enable the user to proceed to the next step (the steps themselves can and will control it).
- `onNextStep`: A function that gets called once a step has passed.
- `children`: Will include the steps that the wizard should run through.

### Note
Some of the implementation might change, but this is good enough to start with.

Closes #3543

### DIFF
--- a/webui/src/pages/repositories/wizard/wizard.jsx
+++ b/webui/src/pages/repositories/wizard/wizard.jsx
@@ -16,8 +16,8 @@ const transitionDefaults = {
 const Wizard = ({
                     hashEnabled=false,
                     transitions=transitionDefaults,
-                    progressBar = false,
-                    skipButton = false,
+                    showProgressBar = false,
+                    showSkipButton = false,
                     onComplete = () => {},
                     canProceed = true,
                     onNextStep = () => {},
@@ -46,7 +46,7 @@ const Wizard = ({
                         <StepWizard
                             onStepChange={onStepChange}
                             transitions={transitions}
-                            nav={progressBar && <WizardNav />}
+                            nav={showProgressBar && <WizardNav />}
                             instance={setInstance}
                             isHashEnabled={hashEnabled}
                         >
@@ -59,7 +59,7 @@ const Wizard = ({
             {<WizardController
                 stepWizard={state.stepWizard}
                 canProceed={canProceed}
-                skipButton={skipButton}
+                skipButton={showSkipButton}
                 onComplete={onComplete}/>}
 
         </Container>
--- a/webui/src/pages/repositories/wizard/wizard.jsx
+++ b/webui/src/pages/repositories/wizard/wizard.jsx
@@ -18,6 +18,7 @@ const Wizard = ({
                     transitions=transitionDefaults,
                     progressBar = false,
                     isGoBackButton = false,
+                    isSkipButton = false,
                     onComplete=(any) => any,
                     canProceed = true,
                     onNextStep = () => {},
@@ -58,6 +59,7 @@ const Wizard = ({
             {<WizardController
                 stepWizard={state.stepWizard}
                 canProceed={canProceed}
+                isSkipButton={isSkipButton}
                 isGoBackButton={isGoBackButton}
                 onComplete={onComplete}/>}
 
--- a/webui/src/pages/repositories/wizard/wizard.jsx
+++ b/webui/src/pages/repositories/wizard/wizard.jsx
@@ -18,7 +18,7 @@ const Wizard = ({
                     transitions=transitionDefaults,
                     progressBar = false,
                     skipButton = false,
-                    onComplete=(any) => any,
+                    onComplete = () => {},
                     canProceed = true,
                     onNextStep = () => {},
                     children,
@@ -27,9 +27,10 @@ const Wizard = ({
     const [state, setState] = useState({
         stepWizard: {},
     });
+    const [refresh, setRefresh] = useState(false);
 
     const onStepChange = () => {
-        setState({...state});
+        setRefresh(!refresh);
         onNextStep();
     };
 
@@ -72,9 +73,6 @@ const WizardNav = ({totalSteps, currentStep}) => {
 }
 
 const WizardController = ({stepWizard, canProceed, skipButton = false, onComplete}) => {
-    const finalStep = () => {
-        onComplete();
-    }
     return (
         <Container>
             <Row className={'justify-content-center'}>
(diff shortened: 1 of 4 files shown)


webui/src/pages/repositories/wizard/wizard.jsx
```diff
@@ -0,0 +1,106 @@
...
+    exitRight: "",
+    exitLeft: "",
+    intro: "",
+}
+
+const Wizard = ({
+                    hashEnabled=false,
+                    transitions=transitionDefaults,
+                    progressBar = false,
+                    isGoBackButton = false,
```
arielshaqed, line 20:
```suggestion
                    goBackButton = false,
```
or
```suggestion
                    canGoBack = false,
```
?

AFAIU from reading the code, this says whether to _add_ a "go back button", _not_ that this _is" a "go back button".  So it's like `progressBar` IIUC.

Similarly for `isSkipButton`; I'd expect `canSkip` or `skipButton`, depending on whether you want to talk about the UI or the property of the flow.

webui/src/pages/repositories/wizard/wizard.jsx
```diff
@@ -0,0 +1,106 @@
...
+                    onNextStep = () => {},
+                    children
+                }) => {
+
+    const [state, setState] = useState({
+        stepWizard: {},
+    });
+
+    const onStepChange = () => {
+        setState({...state});
```
arielshaqed, line 33: Why copy the contents of `state`, if `onNextStep` cannot access it?

webui/src/pages/repositories/wizard/wizard.jsx
```diff
@@ -0,0 +1,106 @@
...
+const Wizard = ({
+                    hashEnabled=false,
+                    transitions=transitionDefaults,
+                    progressBar = false,
+                    isGoBackButton = false,
+                    isSkipButton = false,
+                    onComplete=(any) => any,
+                    canProceed = true,
+                    onNextStep = () => {},
+                    children
```
arielshaqed, line 25:
```suggestion
                    children,
```
?

webui/src/pages/repositories/wizard/wizard.jsx
```diff
@@ -0,0 +1,106 @@
...
+                            <button className='btn btn-secondary btn-block' onClick={stepWizard.nextStep}>Skip</button>
+                        </Col>
+                        :
+                        null
+                }
+                <Col className={"col-2 mb-2 mt-2"}>
+                    {
+                        stepWizard.currentStep < stepWizard.totalSteps ?
+                            <div>
+                                <button className='btn btn-primary btn-block' disabled={!canProceed} onClick={stepWizard.nextStep}>Next Step</button>
```
arielshaqed, line 95:
```suggestion
                                <button className='btn btn-primary btn-block' disabled={!canProceed} onClick={stepWizard.nextStep}>Next</button>
```
or similar.

I don't think "next **step**" is helpful to users, that looks like imposing wizard terminology on them.

webui/src/pages/repositories/wizard/wizard.jsx
```diff
@@ -0,0 +1,106 @@
...
+                            </div>
+                            :
+                            <button className='btn btn-success btn-block' onClick={finalStep} disabled={!canProceed}>Finish</button>
+                    }
+                </Col>
```
arielshaqed, line 100: Confused by this layout.  There are 2 or 3 buttons.  The first, if present, goes in one column.  The other 2 both go into a single column, separated by a colon.  What is the intended effect of this?  I think it will wrap strangely on a narrow display, and always display strangely (e.g., unequal spacing).

arielshaqed (changes requested):
Looks great!  Have  a few questions about language and layout.

webui/src/pages/repositories/wizard/wizard.jsx
Jonathan-Rosenberg: Product's request - @johnnyaug

webui/src/pages/repositories/wizard/wizard.jsx
Jonathan-Rosenberg: It's for the skip button to take less space than the other ones.
I think that the `<div>` there is a left over from before. I'll get rid of it.

The two buttons you're referring to are never displayed together: if `stepWizard.currentStep < stepWizard.totalSteps` is true, the first button will be displayed in this column, if not, the other one will.

The difference between the column defined in line 85 and the column that's defined in line 91 is that the former might not appear at all and latter will always appear.

webui/src/pages/repositories/wizard/wizard.jsx
Jonathan-Rosenberg: This is done to re-render the wizard controller (the buttons below that let you pass through the steps). 
If we won't do this, the `currentStep` property (and probably other ones as well) will not get updated and passed to the `WizardController` properly (via the `stepWizard={state.stepWizard}` property)

webui/src/pages/repositories/wizard/wizard.jsx
```diff
@@ -0,0 +1,101 @@
...
+    exitRight: "",
+    exitLeft: "",
+    intro: "",
+}
+
+const Wizard = ({
+                    hashEnabled=false,
+                    transitions=transitionDefaults,
+                    progressBar = false,
+                    skipButton = false,
```
johnnyaug, line 20:
```suggestion
                    isShowProgressBar = false,
                    isShowSkipButton = false,
```

johnnyaug (approved):
Nice! See small comments

webui/src/pages/repositories/wizard/wizard.jsx
Jonathan-Rosenberg: This is yet insufficient because I need to keep the stepWizard instance somewhere where it is kept updated. 
If I use the refresh method, and in `setInstance` I assign the step wizard instance to a local variable, it will be run over when the component will re-render, thus resulting with an empty step wizard instance at any given time.
Sounds a bit confusing, I know, but we can go over it.

johnnyaug: I didn't mean that you should assign it to a local variable - still assign it to the state.
But here, use a separate refresh state.

Jonathan-Rosenberg: Ok cool

webui/src/pages/repositories/wizard/wizard.jsx
Jonathan-Rosenberg:
```suggestion
                    showProgressBar = false,
                    showSkipButton = false,
```
?

johnnyaug: Was going for the "is" prefix with booleans, but I can live with that.

arielshaqed:
THANKS!

PR #3610, opened 2022-07-04
KV Store: Staging Manager Implementation

N-o-Z:


pkg/kv/mem/store.go
```diff
@@ -43,17 +43,17 @@ func (d *Driver) Open(_ context.Context, _ kvparams.KV) (kv.Store, error) {
...
 
 func combinedKey(partitionKey, key []byte) string {
-	return fmt.Sprintf("%s_%s", partitionKey, key)
+	return kv.FormatPath(string(partitionKey), string(key))
 }
 
 func keyFromCombinedKey(combinedKey string) []byte {
 	//nolint:gomnd
-	return []byte(strings.SplitN(combinedKey, "_", 2)[1])
+	return []byte(strings.SplitN(combinedKey, kv.PathDelimiter, 2)[1])
```
itaiad200, line 51: Isn't `PathDelimiter` a simple `/`? What if it's in the partitionKey as well?

pkg/kv/kvtest/test_model.proto
```diff
@@ -6,7 +6,7 @@ import "google/protobuf/timestamp.proto";
 package io.treeverse.lakefs.kvtest;
 
 message TestModel {
-    string name = 1;
+    bytes name = 1;
     string another_string = 2;
```
itaiad200, line 10:
```suggestion
    string just_a_string = 2;
```

pkg/kv/iterators.go
```diff
@@ -0,0 +1,275 @@
+package kv
```
itaiad200, line 1: Refactoring code by moving it between files in a PR with other changes, makes it almost impossible to understand if any previous code was changed.

pkg/kv/iterators.go
```diff
@@ -0,0 +1,275 @@
...
+
+func (si *SkipFirstIterator) Err() error {
+	return si.it.Err()
+}
+
+func (si *SkipFirstIterator) Close() {
+	si.it.Close()
+}
+
+type PartitionIterator struct {
```
itaiad200, line 206: What does it do? Please document

pkg/kv/iterators.go
```diff
@@ -0,0 +1,275 @@
...
+	ctx          context.Context
+	store        Store
+	msgType      protoreflect.MessageType
+	itr          EntriesIterator
+	partitionKey string
+	value        *MessageEntry
+	err          error
+}
+
+// NewPartitionIterator initiates the staging iterator with a batchSize
```
itaiad200, line 216: staging iterator?

pkg/kv/iterators.go
```diff
@@ -0,0 +1,275 @@
...
+func NewPartitionIterator(ctx context.Context, store Store, msgType protoreflect.MessageType, partitionKey string) (*PartitionIterator, error) {
+	itr, err := ScanPrefix(ctx, store, []byte(partitionKey), []byte(""), []byte(""))
```
itaiad200, line 218: Why `ScanPrefix` - you always pass an empty prefix.

pkg/kv/iterators.go
```diff
@@ -0,0 +1,275 @@
...
+	if p.Err() != nil {
+		return false
+	}
+	p.value = nil
+	if !p.itr.Next() {
+		return false
+	}
+	entry := p.itr.Entry()
+	if entry == nil {
+		p.err = ErrNotFound
```
itaiad200, line 241: I don't think that's the proper error here - you weren't looking for something.

pkg/kv/iterators.go
```diff
@@ -0,0 +1,275 @@
...
+		return false
+	}
+	value := p.msgType.New().Interface()
+	err := proto.Unmarshal(entry.Value, value)
+	if err != nil {
+		p.err = fmt.Errorf("unmarshal proto data for key %s: %w", entry.Key, err)
+		return false
+	}
+	p.value = &MessageEntry{
+		Key:   string(entry.Key),
```
itaiad200, line 251: The key shouldn't be `string` too.

pkg/kv/iterators.go
```diff
@@ -0,0 +1,275 @@
...
+		Value: value,
+	}
+	return true
+}
+
+func (p *PartitionIterator) SeekGE(key []byte) {
+	p.itr.Close() // Close previous before creating new iterator
```
itaiad200, line 258: Should we check for errors or silence them?

pkg/kv/iterators.go
itaiad200, line 206: I think we're missing tests for this iterator.

pkg/gateway/testutil/gateway_setup.go
```diff
@@ -39,6 +39,7 @@ func GetBasicHandler(t *testing.T, authService *FakeAuthService, databaseURI str
 		T:          t,
 	}
 	viper.Set(config.BlockstoreTypeKey, block.BlockstoreTypeMem)
+	viper.Set("database.kv_enabled", false)
```
itaiad200, line 42: Here too? Why?

pkg/graveler/graveler.proto
```diff
@@ -50,3 +50,9 @@ message BranchProtectionBlockedActions {
 message BranchProtectionRules {
   map<string, BranchProtectionBlockedActions> branch_pattern_to_blocked_actions = 1;
 }
+
+message StagedObject {
```
itaiad200, line 54: Maybe
```suggestion
message WorkspaceEntry {
```
or 
```suggestion
message UncommittedEntry {
```

pkg/graveler/staging/manager.go
```diff
@@ -4,112 +4,106 @@ import (
...
 
-func (p *Manager) Get(ctx context.Context, st graveler.StagingToken, key graveler.Key) (*graveler.Value, error) {
-	res, err := p.db.Transact(ctx, func(tx db.Tx) (interface{}, error) {
-		value := &graveler.Value{}
-		err := tx.Get(value, "SELECT identity, data FROM graveler_staging_kv WHERE staging_token=$1 AND key=$2", st, key)
-		return value, err
-	}, p.txOpts(db.ReadOnly())...)
-	if errors.Is(err, db.ErrNotFound) {
-		return nil, graveler.ErrNotFound
+func protoFromValue(key []byte, m *graveler.Value) *graveler.StagedObject {
```
itaiad200, line 24: What does `m` stand for?

pkg/graveler/staging/manager.go
```diff
@@ -4,112 +4,106 @@ import (
...
-				return nil, graveler.ErrPreconditionFailed
-			}
-			return res, err
+
+	var err error
+	pb := protoFromValue(key, value)
+	if overwrite {
+		err = m.store.SetMsg(ctx, string(st), key, pb)
+	} else {
+		err = m.store.SetMsgIf(ctx, string(st), key, pb, nil)
```
itaiad200, line 68: I think it's the other way around? overwrite should make sure that a value is there before writing?

pkg/graveler/staging/manager.go
```diff
@@ -4,112 +4,106 @@ import (
...
+	}
 	return err
 }
 
-func (p *Manager) DropKey(ctx context.Context, st graveler.StagingToken, key graveler.Key) error {
-	_, err := p.db.Transact(ctx, func(tx db.Tx) (interface{}, error) {
-		return tx.Exec("DELETE FROM graveler_staging_kv WHERE staging_token=$1 AND key=$2", st, key)
-	}, p.txOpts()...)
+func (m *Manager) DropKey(ctx context.Context, st graveler.StagingToken, key graveler.Key) error {
+	// Changed behavior from DB - we will not get an error on delete non-existent
```
itaiad200, line 77: Correctness maintained from the lakeFS side because the metadata state stays the same. But consumers of this layer might expect the previous error so we need to check that this change of behaviour is accepted by them.

pkg/graveler/staging/manager.go
```diff
@@ -4,112 +4,106 @@ import (
...
-	upperBound := graveler.UpperBoundForPrefix(prefix)
-	builder := sq.Delete("graveler_staging_kv").Where(sq.Eq{"staging_token": st}).Where("key >= ?::bytea", prefix)
-	_, err := p.db.Transact(ctx, func(tx db.Tx) (interface{}, error) {
-		if upperBound != nil {
-			builder = builder.Where("key < ?::bytea", upperBound)
-		}
-		query, args, err := builder.PlaceholderFormat(sq.Dollar).ToSql()
+func (m *Manager) DropByPrefix(ctx context.Context, st graveler.StagingToken, prefix graveler.Key) error {
+	// Wish we had 'drop partition'...
+	// Simple implementation
```
itaiad200, line 95: Link to the issue where we promise to solve this better. Also - I'm not sure we want to block on it. Another option would be to flag this staging token and delete it in the background.

pkg/graveler/staging/manager.go
```diff
@@ -4,112 +4,106 @@ import (
...
+	if data.Identity == nil {
 		return nil, nil
 	}
-	return value, nil
+	return valueFromProto(data), nil
 }
 
-func (p *Manager) Set(ctx context.Context, st graveler.StagingToken, key graveler.Key, value *graveler.Value, overwrite bool) error {
+func (m *Manager) Set(ctx context.Context, st graveler.StagingToken, key graveler.Key, value *graveler.Value, overwrite bool) error {
+	// Tombstone handling
```
itaiad200, line 56: That's not the `WriterFlow` described in the proposal - who's validating that the stagingToken did not change?

itaiad200 (changes requested):
Awesome to see millestone#3 in actions. Generations of lakeFS developers dreamed of this very moment.

pkg/graveler/staging/manager_test.go
```diff
@@ -7,19 +7,47 @@ import (
...
+		return ctx, staging.NewManager(kvStore)
+	}
 	conn, _ := testutil.GetDB(t, databaseURI)
-	return context.Background(), staging.NewManager(conn)
+	return context.Background(), staging.NewDBManager(conn)
 }
 
 func TestSetGet(t *testing.T) {
-	ctx, s := newTestStagingManager(t)
+	t.Run("TestDBSetGet", func(t *testing.T) {
```
itai-david, line 41: How about a general function that receives a testing function `func(t *testing.T, kvEnabled bool)`, and then put all the tests in array and iterate it from 2 function - one for `DB` and one for `KV`?
This will have a single point to delete later, when we abandon the `DB` support.
Take a look at [controller_test.go](https://github.com/treeverse/lakeFS/blob/master/pkg/api/controller_test.go)

itai-david (changes requested):
Thank you @N-o-Z 
A couple of comments/suggestion/requests inside

esti/migrate_test.go
```diff
@@ -187,6 +188,10 @@ func saveStateInLakeFS(t *testing.T) {
 	completedParts := uploadMultipartParts(t, logger, resp, parts, 0)
 	_, err = uploadMultipartComplete(svc, resp, completedParts)
 	require.NoError(t, err, "writing state file")
+	_, err = client.CommitWithResponse(ctx, migrateStateRepoName, migrateStateBranch, &api.CommitParams{}, api.CommitJSONRequestBody{
+		Message: "Save state file",
+	})
+	require.NoError(t, err, "commit state file")
```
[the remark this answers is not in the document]
N-o-Z: We could push it to GCP  :stuck_out_tongue_winking_eye:

pkg/api/serve_test.go
```diff
@@ -80,6 +80,7 @@ func setupHandlerWithWalkerFactory(t testing.TB, factory catalog.WalkerFactory,
 	ctx := context.Background()
 	conn, handlerDatabaseURI := testutil.GetDB(t, databaseURI, opts...)
 	viper.Set(config.BlockstoreTypeKey, block.BlockstoreTypeMem)
+	viper.Set("database.kv_enabled", false)
```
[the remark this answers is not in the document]
N-o-Z: Because TestKVDisabled fails after a kv enabled test due to persistent configuration

pkg/kv/mem/store.go
N-o-Z: I think we can defer this conversation since we're going to re-implement mem soon. This is just so the tests will pass

pkg/kv/iterators.go
N-o-Z: Code was not moved - file name was refactored. We have this issue with every KV package we implement unfortunately

pkg/gateway/testutil/gateway_setup.go
N-o-Z: See previous answer

pkg/graveler/graveler.proto
N-o-Z: StagedEntry works?

pkg/graveler/staging/manager.go
N-o-Z: @itai-david  is right and this is also the current behavior in DB

pkg/graveler/staging/manager.go
N-o-Z: This is DropByPrefix - not Drop so we can't flag the whole staging token. For Drop we might be able to do something like that, but I think we need the graveler scope for that and not staging manager. Opened https://github.com/treeverse/lakeFS/issues/3628

pkg/graveler/staging/iterator.go
```diff
@@ -3,104 +3,74 @@ package staging
...
-	if s.err != nil {
+	if s.Err() != nil {
 		return false
 	}
-	s.initPhase = false
-	s.idxInBuffer++
-	if s.idxInBuffer < len(s.buffer) {
-		return true
+	if !s.itr.Next() {
+		s.entry = nil
```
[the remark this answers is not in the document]
N-o-Z: We never set an error in the iterators if it was simply exhausted. Also this is the same behavior as the DB implementation

pkg/graveler/staging/manager_test.go
N-o-Z: This was considered but it has a downside which in my opinion is worse. It is not possible to run single tests this way

pkg/kv/iterators.go
N-o-Z: We are missing unit tests for all kv iterators - let's open a new issue for that

pkg/graveler/staging/manager.go
itaiad200: So we either implement that logic in `graveler` in this PR or open an issue for it.

pkg/graveler/staging/manager_test.go
itai-david: So even better - we can still have the general function `func actuallyRunTheTest` that receives the test func above and runs it once for DB and once for KV, and instead of putting all test functions in an array, leave the `func Test...` as is , but call the general testing `actuallyRunTheTest` providing it with the tests to run.
That way you can both run each `Test...` individually, and you minimized code to `runWithDB, runWithKV` to a single place
Makes sense?

N-o-Z:
https://github.com/treeverse/lakeFS/pull/3610#discussion_r916200632
This is the exact implementation in DB - this is not tombstone, this is just the removal of a key.

pkg/graveler/staging/manager_test.go
N-o-Z: Not sure I completely follow but, take a look the changes and tell me if this is what you have in mind. Keep in mind we need the test names to be descriptive to differentiate between "DB" and "KV" tests

pkg/api/serve_test.go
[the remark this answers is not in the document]
N-o-Z: Added inline comment

itaiad200 (approved):
LGTM - maybe just document the race we found for dropping the key.

itai-david (approved):

PR #3616, opened 2022-07-05
Change the email unique index creation to be idempotent

### LINKED ISSUE #3174: Non-idempotent code in 000036_users_email_password.down.sql
This line https://github.com/treeverse/lakeFS/blob/0ca69c4f0f7550e8af4e513a305a6c8732ae5d3a/pkg/ddl/000036_users_email_password.up.sql#L7 always attempts to create an index, so is not idempotent.

(I think) this should not cause any issues during normal operation, but it does violate the principles of what migration files should look like -- and will cause issues for some more advanced scenarios that the existing "migrate" commands cannot address.

ortz:
Closes #3174


arielshaqed (approved):
Thanks!

PR #3617, opened 2022-07-05
Allow OIDC as default login

johnnyaug:
Configure lakeFS to redirect to OIDC login automatically.


arielshaqed (approved):
NEAT!

If OIDC default is set, (how?) can a user still use username+password to login?  And, will it make sense to keep a cookie on the browser for how to try logging in the next time?

johnnyaug:
@arielshaqed - your comment is very important. I changed it so that if you explicitly browse to "/auth/login" you get the old, internal login. If you are redirected from another page - we redirect to OIDC login. PTAL.

arielshaqed:
> @arielshaqed - your comment is very important. I changed it so that if you explicitly browse to "/auth/login" you get the old, internal login. If you are redirected from another page - we redirect to OIDC login. PTAL.

Still looks great!  If I want to use username+password then I have to type in `.../auth/login`, or is there some link on the OIDC login page that I can use?

johnnyaug:
@arielshaqed, thanks! The OIDC page is at the control of the external provider - the user may choose to add a link there to "/auth/login".

PR #3621, opened 2022-07-06
metadata manager over KV

### LINKED ISSUE #3520: `auth.MetadataManager` KV support

itai-david:
closes #3520


pkg/auth/metadata.go
```diff
@@ -46,108 +43,83 @@ func generateInstallationID(installationID string) string {
 	return installationID
 }
 
-func insertOrGetInstallationID(tx db.Tx, installationID string) (string, error) {
-	res, err := tx.Exec(`INSERT INTO auth_installation_metadata (key_name, key_value)
-			VALUES ($1,$2)
-			ON CONFLICT DO NOTHING`,
-		InstallationIDKeyName, installationID)
+func (m *KVMetadataManager) insertOrGetInstallationID(ctx context.Context, installationID string) (string, error) {
+	err := m.store.SetIf(ctx, []byte(model.PartitionKey), []byte(installationID), []byte(installationID), nil)
```
N-o-Z, line 47: Maybe we should define a proto model for this, so we can support future changes as needed?

pkg/auth/metadata.go
```diff
@@ -46,108 +43,83 @@ func generateInstallationID(installationID string) string {
...
-func (d *DBMetadataManager) Write(ctx context.Context) (map[string]string, error) {
+func (m *KVMetadataManager) Write(ctx context.Context) (map[string]string, error) {
 	metadata := make(map[string]string)
-	metadata["lakefs_version"] = d.version
+	metadata["lakefs_version"] = m.version
 	metadata["golang_version"] = runtime.Version()
 	metadata["architecture"] = runtime.GOARCH
 	metadata["os"] = runtime.GOOS
-	dbMeta, err := d.db.Metadata(ctx)
+	// dbMeta, err := d.db.Metadata(ctx)
```
N-o-Z, line 103: Leftovers?

pkg/auth/metadata.go
```diff
@@ -7,11 +7,13 @@ import (
 	"time"
```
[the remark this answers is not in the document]
N-o-Z: BTW, after looking at it - I don't really understand why this is part of the auth pkg

N-o-Z (changes requested):
Thanks! Couple of comments

pkg/auth/metadata.go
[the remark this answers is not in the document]
itaiad200: Race conditions between lakeFS instances is one reason for not switching.

pkg/auth/metadata.go
itaiad200, line 47: The key and the value are equals? How would you `Get` it if the installationID change?

itaiad200: I think you meant:
```suggestion
	err := m.store.SetIf(ctx, []byte(model.PartitionKey), []byte(InstallationIDKeyName), []byte(installationID), nil)
```

pkg/auth/metadata.go
```diff
@@ -46,108 +43,83 @@ func generateInstallationID(installationID string) string {
...
 
-func (d *DBMetadataManager) Write(ctx context.Context) (map[string]string, error) {
+func (m *KVMetadataManager) Write(ctx context.Context) (map[string]string, error) {
 	metadata := make(map[string]string)
-	metadata["lakefs_version"] = d.version
+	metadata["lakefs_version"] = m.version
 	metadata["golang_version"] = runtime.Version()
 	metadata["architecture"] = runtime.GOARCH
 	metadata["os"] = runtime.GOOS
-	dbMeta, err := d.db.Metadata(ctx)
```
itaiad200, line 127: Maybe we need another function to our kv interface? `func Metadata() map[string]string`

It's ok to leave this out as a technical debt, as long as we understand the implications. I would consult with @NoyDavidson2 on what metadata values we stopped collecting to understand the impact.

pkg/auth/db_metadata.go
```diff
@@ -0,0 +1,180 @@
...
+	defer rows.Close()
+	scanner := pgxscan.NewRowScanner(rows)
+	for rows.Next() {
+		entry := metadataEntry{}
+		err := scanner.Scan(&entry)
+		if err != nil {
+			return err
+		}
+		if _, exist := dbMeta[entry.KeyName]; exist {
+			// DB metadata - skip
```
itaiad200, line 167: What's the harm of exporting it too?

itaiad200 (changes requested):
Neat implementation! Please sync Noy into this to make sure we don't mess up the stats too much. Also - it requires manual testing to verify that they are collected properly, but that is tricky to avoid spamming the logs (I believe Noy can be helpful with this too).

pkg/auth/metadata.go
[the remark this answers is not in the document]
N-o-Z: I didn't understand your comment - can you please explain?

pkg/auth/db_metadata.go
itai-david: No harm. Just preserves some irrelevant (IMHO) data in our system.
When lakeFS starts it rewrites the metadata, updating existing keys, and so, if you upgraded postgress, the new version will be reflected by the metadata.
If you switch to another technology, this value will not be updated so you will have irrelevant/outdated metadata stored
WDYT?

pkg/auth/metadata.go
itai-david: 1. Will talk to Noy and update here
2. Nevertheless, I believe that having a relevant `Metadata()` function, is not only a good idea, but somewhat essential. However, we should design something more flexible than what we have today, as our KV is only the mediation layer, and the real data should be undlying-store-specific - I will open an issue for that, once I confirm the needs with Noy

pkg/auth/metadata.go
itai-david: Yicks :open_mouth: 
How did  that even passed all tests...?

pkg/auth/metadata.go
itai-david: Yes and no. Actually left this to mark that we possibly want to implement something similar for KV.
Anyhow - removed

pkg/auth/metadata.go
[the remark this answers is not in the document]
itai-david: I don't see any advantage of moving the entire metadata under a single key. It is also possible (if we will add store specific metadata) that we have different keys for different setups, which will be more natural to support with a simple
string->string mapping, as it is.
Don't you think?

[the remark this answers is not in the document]
itaiad200: @N-o-Z I meant that replicas of the same lakeFS installations may set this concurrently and override keys in comparison to the current implementation where the keys are separated.

itaiad200 (approved):
Approved, please validate:
1. BI can get along without postgres metadata for now.
2. Stats are sent with `kv_enabled` as expected. Unfortunately this can't\isn't tested properly.

N-o-Z (approved):

PR #3632, opened 2022-07-07
Fix/kv mem store implementation 3622

N-o-Z:


itaiad200:
@N-o-Z should we review this? It looks like it contains many of the changes that are part of #3610

pkg/kv/mem/store.go
```diff
@@ -37,23 +37,12 @@ func init() {
 
 func (d *Driver) Open(_ context.Context, _ kvparams.KV) (kv.Store, error) {
 	return &Store{
-		m:    make(map[string][]byte),
-		keys: []string{},
+		m: make(map[string]map[string]kv.Entry, 0),
```
itai-david, line 40: Looks like a comment explaining the first is the partition and the second is the key could be helpful.
Or better, IMHO, define a partition type that is a key->val mapping and then `m` which is a key->partition mapping.

pkg/kv/mem/store.go
```diff
@@ -66,14 +55,14 @@ func (s *Store) Get(_ context.Context, partitionKey, key []byte) (*kv.ValueWithP
 	s.mu.RLock()
 	defer s.mu.RUnlock()
 
-	combinedKey := combinedKey(partitionKey, key)
-	value, ok := s.m[combinedKey]
+	sKey := encodeKey(key)
+	value, ok := s.m[string(partitionKey)][sKey]
```
itai-david, line 59: What is the reason to allow plain strings for partition keys?
Why not encoding it too?

itai-david (commented):
I reviewed only the last commit - 09d118f - as it seems the rest are related to not branching from master (?)
Anyhow, some comments and questions inside.
I'm not marking as `Request changes` as not  to block you during Sunday

N-o-Z:
> @N-o-Z should we review this? It looks like it contains many of the changes that are part of #3610

It's based over the staging manager changes. The relevant code is in the last commit

pkg/kv/mem/store.go
N-o-Z: Since we control the partition key, we can take advantage of that and assume it can be converted to string

pkg/kv/mem/store.go
```diff
@@ -146,72 +133,58 @@ func (s *Store) Delete(_ context.Context, partitionKey, key []byte) error {
...
-	combinedKey := e.store.keys[idx]
-	value := e.store.m[combinedKey]
 
-	partition := partitionKeyFromCombinedKey(combinedKey)
-	if !bytes.Equal(partition, e.partition) {
-		e.start = ""
-		e.entry = nil
-		return false
+	l := make([]*kv.Entry, 0)
+	if _, ok := e.store.m[e.partition]; ok {
```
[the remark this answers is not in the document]
N-o-Z: Yes - on each next call we cannot assume anything about the dataset. We need to collect everything, sort the keys and continue from where we left off last time

pkg/kv/mem/store.go
```diff
@@ -146,72 +133,58 @@ func (s *Store) Delete(_ context.Context, partitionKey, key []byte) error {
...
+		for _, v := range e.store.m[e.partition] {
```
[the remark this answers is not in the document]
N-o-Z: I like your idea :cowboy_hat_face:

itaiad200:
> > @N-o-Z should we review this? It looks like it contains many of the changes that are part of #3610
> 
> It's based over the staging manager changes. The relevant code is in the last commit

In that case, you can't commit it until you merge the staging PR, right? If so, it's much better to open a PR with the destination branch being the dependant branch and not master.

N-o-Z:
> > > @N-o-Z should we review this? It looks like it contains many of the changes that are part of #3610
> > 
> > 
> > It's based over the staging manager changes. The relevant code is in the last commit
> 
> In that case, you can't commit it until you merge the staging PR, right? If so, it's much better to open a PR with the destination branch being the dependant branch and not master.

Yes, this will merge only after the staging manager PR. Also you are correct, I'm still working on my Github workflow 😃

pkg/kv/mem/store.go
```diff
@@ -122,18 +114,19 @@ func (s *Store) SetIf(_ context.Context, partitionKey, key, value []byte, valueP
...
+	curr, currOK := s.m[string(partitionKey)][sKey]
 	if valuePredicate == nil {
 		if currOK {
 			return fmt.Errorf("key=%v: %w", key, kv.ErrPredicateFailed)
 		}
-		s.insertNewKey(combinedKey)
-	} else if !bytes.Equal(valuePredicate.([]byte), curr) {
-		return fmt.Errorf("%w: key=%v", kv.ErrPredicateFailed, combinedKey)
+	} else if !bytes.Equal(valuePredicate.([]byte), curr.Value) {
+		return fmt.Errorf("%w: partition=%s, key=%v, encoding=%s", kv.ErrPredicateFailed, partitionKey, key, sKey)
```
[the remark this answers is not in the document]
N-o-Z: See previous comment

pkg/kv/mem/store.go
```diff
@@ -66,14 +61,14 @@ func (s *Store) Get(_ context.Context, partitionKey, key []byte) (*kv.ValueWithP
 	s.mu.RLock()
 	defer s.mu.RUnlock()
 
-	combinedKey := combinedKey(partitionKey, key)
-	value, ok := s.m[combinedKey]
+	sKey := encodeKey(key)
+	value, ok := s.m[string(partitionKey)][sKey]
```
itaiad200, line 65: if `s.m` is missing the `partitionKey`, you'll get the default value of `PartitionMap` which is `nil` and your code will panic. I wonder how did no test find this. I think we should add a test for non existing partition.

pkg/kv/mem/store.go
```diff
@@ -122,18 +114,19 @@ func (s *Store) SetIf(_ context.Context, partitionKey, key, value []byte, valueP
 	}
 	s.mu.Lock()
 	defer s.mu.Unlock()
-	combinedKey := combinedKey(partitionKey, key)
 
-	curr, currOK := s.m[combinedKey]
+	sKey := encodeKey(key)
+	curr, currOK := s.m[string(partitionKey)][sKey]
```
itaiad200, line 119: See previous comment about unfound `partitionKey`.

pkg/kv/mem/store.go
```diff
@@ -14,21 +14,27 @@ import (
...
-	partition []byte
+	start     []byte
+	partition string
 	store     *Store
 }
 
-const DriverName = "mem"
+const (
+	DriverName  = "mem"
+	EmptyResult = 0
```
itaiad200, line 36:
```suggestion
	emptyResult = 0
```

pkg/kv/mem/store.go
```diff
@@ -146,72 +139,56 @@ func (s *Store) Delete(_ context.Context, partitionKey, key []byte) error {
...
+				entry := v
+				l = append(l, &entry)
+			}
+		}
 	}
-	// set start to the next item - nil as end indicator
-	if idx+1 < len(e.store.keys) {
-		e.start = e.store.keys[idx+1]
+	sort.Slice(l, func(i, j int) bool { return bytes.Compare(l[i].Key, l[j].Key) < 0 })
+	if len(l) <= EmptyResult+1 { // only 1 key >= start, set start to nil, so to indicate this is the last iteration
```
itaiad200, line 183:
```suggestion
	if len(l) <= EmptyResult+1 { // at most 1 key >= start, set start to nil, so to indicate next call to return false immediately.
```

pkg/kv/mem/store.go
```diff
@@ -146,72 +139,56 @@ func (s *Store) Delete(_ context.Context, partitionKey, key []byte) error {
...
+		e.start = nil
+		if len(l) == EmptyResult { // No results
+			return false
+		}
```
itaiad200, line 187: I think this is more readable - per your consideration
```suggestion 
	if len(l) == 0 {
		e.start = nil
		return false
	} 
	if len(l) == 1 {
		e.start = nil
		e.entry = l[0]
		return true
	}
	sort.Slice(l, func(i, j int) bool { return bytes.Compare(l[i].Key, l[j].Key) < 0 })
	e.start = l[1]
	e.entry = l[0]
	return true
```

itaiad200 (changes requested):
Thanks for this quick win PR 💪 
Single blocking comment: Can we also reuse the mem implementation in the tests where `postgres` is now used?

N-o-Z:
> Thanks for this quick win PR muscle Single blocking comment: Can we also reuse the mem implementation in the tests where `postgres` is now used?

Yes! I will modify the tests

pkg/kv/mem/store.go
N-o-Z: It doesn't panic, this specific scenario is tested, and this is similar to https://github.com/treeverse/lakeFS/pull/3632#discussion_r917385528 comment

pkg/kv/mem/store.go
N-o-Z: Same as above

N-o-Z:
@itaiad200 Please note the replies regarding the 'panic'

pkg/graveler/staging/manager_test.go
```diff
@@ -7,12 +7,13 @@ import (
 	"fmt"
 	"testing"
 
+	"github.com/treeverse/lakefs/pkg/kv"
+
+	"github.com/treeverse/lakefs/pkg/kv/kvtest"
+
```
itaiad200, line 13: delete lines

itaiad200 (approved):
LGTM

Issue #2622, opened 2021-10-27
Fix Documentation for K8s deployment


iddoavn:
Pretty silly but can be time consuming to troubleshoot. 

When you follow: [https://docs.lakefs.io/deploy/k8s.html](https://docs.lakefs.io/deploy/k8s.html)

There are extra spaces under the first box of s3 that will cause an error:

```
iddoavneri@Iddos-MacBook-Pro ~ % helm install example-lakefs lakefs/lakefs -f conf-values.yaml
Error: INSTALLATION FAILED: YAML parse error on lakefs/templates/configmap.yaml: error converting YAML to JSON: yaml: line 8: could not find expected ':'
```

Instead one should use:
```
secrets:
 # replace DATABASE_CONNECTION_STRING with the connection string of the database you created in a previous step.
 # e.g. postgres://postgres:Password@my-lakefs-db.rds.amazonaws.com:5432/lakefs
 databaseConnectionString: postgres://postgres:MYPASSWORD@iddo-lakefs-db-dev.random.us-east-2.rds.amazonaws.com:5432/iddo-lakefs-db-dev
 # replace this with a randomly-generated string
 authEncryptSecretKey: RandomGeneratedString
lakefsConfig: |
blockstore:
  type: s3
  s3:
    region: us-east-2
    credentials:
      access_key_id: AKIA6HHRMQLEXAMPLE
      secret_access_key: UBfIYdDA7CEmHhTzxuvyEXAMPLE
```

PR #3634, opened 2022-07-09
tried to solve extra line issue in lakeFS/docs/deploy/k8s.md

Harsh3363:
I tried to resolve the docs/deploy/k8s.md file and solve issue #2622


arielshaqed (changes requested):
Hi @Harsh3363 ,

This does not appear to resolve #2622 (which also appears to refer to S3 rather than to Azure).  I am also not sure that it yields the desired YAML.

Could you explain some more about the change?

@iddoavn can you please confirm?  @peacing to review (and tell me I'm wrong!) since you're on the issue and may have context.

PR #3642, opened 2022-07-11
Fix & Reenable TestRunResultsIterator

### LINKED ISSUE #3623: Fix and reenable TestRunResultsIterator
It's flaky and we need to fix it and unskip it

itaiad200:
closes #3623

### DIFF
--- a/pkg/actions/kv_run_results_iterator_test.go
+++ b/pkg/actions/kv_run_results_iterator_test.go
@@ -21,23 +21,14 @@ const (
 	testByCommit       = "testCommit1"
 	testMissingPrimary = "branchPartialPrimary"
 	IndexOutOfRange    = "zzz"
+	withHooks          = 100
 )
-
-var keyMap = make(map[string]int, 0)
-var keyList []string
-
-func TestRSGF(t *testing.T) {
-	for i := 0; i < 50; i++ {
-		t.Run(strconv.Itoa(i), func(t *testing.T) {
-			TestRunResultsIterator(t)
-		})
-	}
-}
 
 func TestRunResultsIterator(t *testing.T) {
 	ctx := context.Background()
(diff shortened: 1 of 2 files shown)


pkg/actions/kv_run_results_iterator_test.go
```diff
@@ -213,12 +223,23 @@ func createTestData(t *testing.T, ctx context.Context, kvStore kv.StoreMessage)
...
-	keyList[msgIdx] = key
+	keyList = append(keyList, key)
+
+	// reverse the list to make it sorted in ascending order
+	reverse(keyList)
+
+	return keyMap, keyList
+}
+
+func reverse(s []string) {
```
N-o-Z, line 241: You know the size of the list, you can create an array and insert the items in reverse order, I think this func is redundant

N-o-Z (changes requested):
Thanks - one small comment

pkg/actions/kv_run_results_iterator_test.go
```diff
@@ -20,16 +21,14 @@ const (
 	testByCommit       = "testCommit1"
 	testMissingPrimary = "branchPartialPrimary"
 	IndexOutOfRange    = "zzz"
+	withHooks          = 100
```
N-o-Z, line 24: What does withHooks mean?

itaiad200: rephrased.

pkg/actions/kv_run_results_iterator_test.go
itaiad200: It is, but with all the mappings it makes handling `keyList` & `keyMap` so much harder that I wish to avoid another off-by-one mistake. Reversing is much easier than inserting in reverse order.

PR #3645, opened 2022-07-11
KV Store: Add StagingManager Migration

N-o-Z:


pkg/graveler/staging/db_manager.go
```diff
@@ -113,3 +137,81 @@ func (p *DBManager) txOpts(opts ...db.TxOpt) []db.TxOpt {
...
+			}
+			if err = encoder.Encode(kv.Entry{
+				PartitionKey: []byte(stagingToken),
+				Key:          record.Key,
+				Value:        data,
+			}); err != nil {
+				return err
+			}
+		}
+		rows.Close()
```
itai-david, line 169: `defer rows.close()` right after the query and err check

itai-david (approved):
Minor comment to move `rows.Close()` under defer.
Other than that, looks solid!

pkg/graveler/staging/db_manager.go
N-o-Z: @itai-david  Thanks for the review!! Please note that the rows variable is created inside a for loop, so the common practice is to use close and not defer.
We can export this logic to function, but I think it's safe to say that the migration process will not cause a resource leak in the server (it is executed separately and we don't handle panics)

pkg/graveler/staging/db_manager.go
```diff
@@ -113,3 +137,81 @@ func (p *DBManager) txOpts(opts ...db.TxOpt) []db.TxOpt {
 	}
 	return append(o, opts...)
 }
+
+// runWorker writing staging token data to fd
+func runWorker(ctx context.Context, d *pgxpool.Pool, jobChan <-chan string) error {
+	record := &graveler.ValueRecord{}
+	for t := range jobChan {
```
itaiad200, line 144: The data skew here worries me. It's not uncommon to have very few branches of lakeFS, with only one the branches containing hundred of thousands of objects. Paralle5lizing per branch won't really help in these situation, we would have to parallelize within a staging token.

This comment is an optimization one for sure, we can flag it as premature and open an issue. I believe we should also recommend the users to commit their changes before the migration to speed it up dramatically.

pkg/graveler/staging/kv_migrate_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+type stagingTestRecord struct {
+	stagingToken graveler.StagingToken
+	records      []graveler.ValueRecord
+}
+
+func TestMigrate(t *testing.T) {
+	ctx := context.Background()
+	conn, _ := testutil.GetDB(t, databaseURI)
+	store, err := kv.Open(context.Background(), kvpg.DriverName, kvparams.KV{Postgres: &kvparams.Postgres{ConnectionString: databaseURI}})
```
itaiad200, line 27: Can we use mem instead?

pkg/graveler/staging/kv_migrate_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+func createMigrateTestData(t *testing.T, ctx context.Context, mgr graveler.StagingManager, size int) []stagingTestRecord {
+	t.Helper()
+	d := make([]stagingTestRecord, 0)
+	for i := 0; i < size; i++ {
+		tokenData := stagingTestRecord{
+			stagingToken: graveler.StagingToken("test_token_" + strconv.Itoa(i)),
+			records:      make([]graveler.ValueRecord, 0),
+		}
+		for j := 0; j < rand.Intn(size); j++ {
```
itaiad200, line 72: Trying to avoid more flakiness: Will a random 0 that creates no entries results with a failing test?

pkg/graveler/staging/kv_migrate_test.go
```diff
@@ -0,0 +1,90 @@
...
+
+func genRandomBytes(t *testing.T, size int) (blk []byte) {
+	t.Helper()
+	blk = make([]byte, size)
+	_, err := rand.Read(blk)
+	testutil.MustDo(t, "Generate random bytes", err)
+	return
+}
+
+func createMigrateTestData(t *testing.T, ctx context.Context, mgr graveler.StagingManager, size int) []stagingTestRecord {
```
itaiad200, line 64: Can we also add tombstones to the test?

itaiad200 (commented):
Approved with some comments, thanks!

itaiad200 (approved):
Approved with some comments, thanks!

pkg/graveler/staging/db_manager.go
N-o-Z: I think it's a very good point on the commit.
I was using the 'natural' division of labor for the initial concurrency and was deliberating about this exact point. I agree we should open it as an improvement, we just need to see how we ensure we assign the right amount of workers for each workload.

pkg/graveler/staging/kv_migrate_test.go
N-o-Z: Migrate tests explicitly work with PG

pkg/graveler/staging/kv_migrate_test.go
N-o-Z: The point here was to allow a test of a staging token with no records. Moreover - maybe we should explicitly have one with no records and not rely on randomness

Issue #3570, opened 2022-06-25
KV Graveler Implementation - Repository Delete


itai-david:
Delete a repository derives deleetion of all correlated entities (branches, commits and tags). How do we handle failures during this procedure?
See #3566 for further details

PR #3646, opened 2022-07-12
3591 kv graveler repository

itai-david:


pkg/graveler/ref/db_manager.go
```diff
@@ -502,3 +526,64 @@ func (m *DBManager) addGenerationToNodes(nodes map[graveler.CommitID]*CommitNode
 		nodesCommitIDs = nextIterationNodes
 	}
 }
+
+// rWorker writes a repository, and all repository related entities, to fd
+func rWorker(rChan <-chan *graveler.RepositoryRecord) error {
```
itaiad200, line 531: The worker does so little that it's almost not worth parallelizing on it, unless this is a preparation for other entities under the worker like branches, etc.

pkg/graveler/ref/db_manager.go
```diff
@@ -502,3 +526,64 @@ func (m *DBManager) addGenerationToNodes(nodes map[graveler.CommitID]*CommitNode
...
+	if err != nil {
+		return err
+	}
+	defer rows.Close()
+	rowScanner := pgxscan.NewRowScanner(rows)
+	for rows.Next() {
+		r := new(graveler.RepositoryRecord)
+		err = rowScanner.Scan(r)
+		if err != nil {
+			return err
```
itaiad200, line 582: I think the channel is not closed here, which can result with ungraceful exit of the workers.

pkg/graveler/ref/manager.go
```diff
@@ -47,23 +47,75 @@ func NewKVRefManager(executor batch.Batcher, kvStore kv.StoreMessage, db db.Data
...
+	}
+	return nil
 }
 
 func (m *KVManager) CreateRepository(ctx context.Context, repositoryID graveler.RepositoryID, repository graveler.Repository, token graveler.StagingToken) error {
-	return m.db.CreateRepository(ctx, repositoryID, repository, token)
+	firstCommit := graveler.NewCommit()
+	firstCommit.Message = graveler.FirstCommitMsg
+	firstCommit.Generation = 1
+	// commitID := m.addressProvider.ContentAddress(firstCommit)
```
itaiad200, line 81: Why is it commented out?

pkg/graveler/ref/manager.go
```diff
@@ -47,23 +47,75 @@ func NewKVRefManager(executor batch.Batcher, kvStore kv.StoreMessage, db db.Data
...
 
 func (m *KVManager) ListRepositories(ctx context.Context) (graveler.RepositoryIterator, error) {
-	return m.db.ListRepositories(ctx)
+	return NewKVRepositoryIterator(ctx, &m.kvStore)
 }
 
 func (m *KVManager) DeleteRepository(ctx context.Context, repositoryID graveler.RepositoryID) error {
-	return m.db.DeleteRepository(ctx, repositoryID)
+	// TODO: delete me
+	// temp code to align with DB manager. Delete once https://github.com/treeverse/lakeFS/issues/3640 is done
```
itaiad200, line 111: 👍

pkg/graveler/ref/repository_iterator.go
```diff
@@ -4,99 +4,87 @@ import (
...
-		return
+	if !ri.it.Next() {
+		ri.value = nil
+		return false
 	}
-	if len(ri.buf) > 0 {
-		return
+	e := ri.it.Entry()
+	if e == nil {
+		ri.err = graveler.ErrNotFound
```
itaiad200, line 50: I don't think `graveler.ErrNotFound` is appropriate here

pkg/graveler/ref/repository_iterator.go
```diff
@@ -4,99 +4,87 @@ import (
...
-			ORDER BY id ASC
-			LIMIT $2`, ri.offset, ri.fetchSize)
-	if ri.err != nil {
-		return
-	}
-	if len(ri.buf) < ri.fetchSize {
-		ri.state = iteratorStateDone
+	repo, ok := e.Value.(*graveler.RepositoryData)
+	if repo == nil || !ok {
+		ri.err = graveler.ErrNotFound
```
itaiad200, line 56: Same here

pkg/graveler/ref/repository_iterator.go
```diff
@@ -4,99 +4,87 @@ import (
...
-	if errors.Is(ri.err, ErrIteratorClosed) {
+func (ri *KVRepositoryIterator) SeekGE(id graveler.RepositoryID) {
+	if errors.Is(ri.Err(), kv.ErrClosedEntries) {
 		return
 	}
-	ri.offset = string(id)
-	ri.buf = ri.buf[:0]
+	ri.it.Close()
+	ri.it, ri.err = kv.NewPrimaryIterator(ri.ctx, ri.store, (&graveler.RepositoryData{}).ProtoReflect().Type(), graveler.RepoPartition(), []byte(graveler.RepoPath("")), kv.IteratorOptionsFrom([]byte(graveler.RepoPath(id))))
 	ri.value = nil
```
itaiad200, line 70: Does `SeekGE` require a `Next()` call before reading the values? Because if it doesn't, next value is nil.

pkg/graveler/ref/repository_iterator_test.go
```diff
@@ -25,7 +25,7 @@ func TestRepositoryIterator(t *testing.T) {
 	}
 
 	t.Run("listing all repos", func(t *testing.T) {
-		iter := ref.NewRepositoryIterator(context.Background(), db, 3)
+		iter := ref.NewDBRepositoryIterator(context.Background(), db, 3)
```
eden-ohana, line 28: why is the kv repository iterator not tested?

pkg/graveler/graveler.go
```diff
@@ -802,7 +802,7 @@ func NewKVGraveler(branchLocker BranchLocker, committedManager CommittedManager,
 }
 
 func (g *KVGraveler) GetRepository(ctx context.Context, repositoryID RepositoryID) (*Repository, error) {
-	return g.db.GetRepository(ctx, repositoryID)
+	return g.RefManager.GetRepository(ctx, repositoryID)
 }
 
 func (g *KVGraveler) CreateRepository(ctx context.Context, repositoryID RepositoryID, storageNamespace StorageNamespace, branchID BranchID) (*Repository, error) {
```
eden-ohana, line 808: why use the DB and not its kv ref manager?

pkg/graveler/ref/manager.go
```diff
@@ -47,23 +47,75 @@ func NewKVRefManager(executor batch.Batcher, kvStore kv.StoreMessage, db db.Data
...
+	if err != nil {
+		return err
+	}
+
+	commitID, err := m.AddCommit(ctx, repositoryID, firstCommit)
+	if err != nil {
+		return err
+	}
+
+	err = m.CreateBranch(ctx, repositoryID, repository.DefaultBranchID, graveler.Branch{CommitID: commitID, StagingToken: token})
```
eden-ohana, line 93: so we decided that if we create a repository, succeed in creating the repository and fail in the commit or branch and then try to create another repository with the same name it will fail?

pkg/graveler/graveler.go
eden-ohana: on all the other functions create, list repository

pkg/graveler/ref/repository_iterator.go
```diff
@@ -4,99 +4,87 @@ import (
...
-		db:        db,
-		ctx:       ctx,
-		fetchSize: fetchSize,
-		buf:       make([]*graveler.RepositoryRecord, 0, fetchSize),
-	}
+type KVRepositoryIterator struct {
+	ctx   context.Context
+	it    kv.MessageIterator
+	err   error
+	value *graveler.RepositoryRecord
```
eden-ohana, line 23: I chose to use repository data instead of repository record (on the commit), then on the kv iterator you don't need to convert, on the DB iterator you do need, but once we delete the DB the repository record will not be needed anymore. WDYT?

pkg/graveler/ref/repository_iterator.go
eden-ohana: yes it requires

pkg/graveler/ref/manager.go
```diff
@@ -47,23 +47,75 @@ func NewKVRefManager(executor batch.Batcher, kvStore kv.StoreMessage, db db.Data
...
+	if err != nil {
+		return err
```
N-o-Z, line 95: Try to delete commit if create branch failed?

pkg/graveler/ref/manager.go
```diff
@@ -47,23 +47,75 @@ func NewKVRefManager(executor batch.Batcher, kvStore kv.StoreMessage, db db.Data
...
+	if err != nil {
+		if errors.Is(err, kv.ErrNotFound) {
+			err = graveler.ErrRepositoryNotFound
+		}
+		return nil, err
+	}
+	return graveler.RepoFromProto(&repo).Repository, nil
+}
+
+func (m *KVManager) createBareRepository(ctx context.Context, repositoryID graveler.RepositoryID, repository graveler.Repository) error {
```
N-o-Z, line 61: Seems like we don't really need this internal method anymore - we can just use the public interface

pkg/graveler/ref/manager.go
```diff
@@ -47,23 +47,75 @@ func NewKVRefManager(executor batch.Batcher, kvStore kv.StoreMessage, db db.Data
...
 func (m *KVManager) ListRepositories(ctx context.Context) (graveler.RepositoryIterator, error) {
-	return m.db.ListRepositories(ctx)
+	return NewKVRepositoryIterator(ctx, &m.kvStore)
 }
 
 func (m *KVManager) DeleteRepository(ctx context.Context, repositoryID graveler.RepositoryID) error {
-	return m.db.DeleteRepository(ctx, repositoryID)
+	// TODO: delete me
+	// temp code to align with DB manager. Delete once https://github.com/treeverse/lakeFS/issues/3640 is done
+	_, err := m.GetRepository(ctx, repositoryID)
```
N-o-Z, line 112: If err != nil - either err == kv.ErrNotFound ==> return graveler.ErrRepositoryNotFound
else return err

pkg/graveler/ref/repository_iterator.go
```diff
@@ -4,99 +4,87 @@ import (
...
 	}
+
+	ri.value = graveler.RepoFromProto(repo)
+	return true
 }
 
-func (ri *RepositoryIterator) SeekGE(id graveler.RepositoryID) {
-	if errors.Is(ri.err, ErrIteratorClosed) {
+func (ri *KVRepositoryIterator) SeekGE(id graveler.RepositoryID) {
+	if errors.Is(ri.Err(), kv.ErrClosedEntries) {
```
N-o-Z, line 65: Question: Should we ignore seek on any iterator err, or continue with the same behavior as in the DB implementation?

pkg/graveler/ref/repository_iterator.go
N-o-Z: graveler.ErrReadingFromStore?

N-o-Z (changes requested):
@itai-david Thanks for this PR and the work to enable the Migration Test in Esti.
As requested I did not review the migration part - but we should be mindful on future failures until this is properly implemented. Some comments

pkg/graveler/ref/repository_iterator.go
eden-ohana: me and @niro discussed and decided to keep it with the record for now until we delete the DB and then maybe we'll change it

pkg/graveler/ref/manager.go
itai-david: `AddCommit` does not  receive a `commitID` as a parameter, so in order to use it here, I had to remove it.
This can be returned when the `Commit` entitiy support is implemented, possibly in a similar way to what we have in the DB ref manager, where an internal `addCommit` is receiveing a commit and its ID and performs the actual DB update, while the public `AddCommit` uses it.
I will add a comment to indicate that

itai-david: Rethinking it, this `commitID` is used for the purpose of creating the default branch before the commit itself, which doesn't make any sense, right?
Am I missing something?
I'm removing this line altogether, as I switched the order of creation of `branch` and `commit`

pkg/graveler/ref/repository_iterator.go
itai-david: Done

pkg/graveler/ref/repository_iterator.go
itai-david: Done

pkg/graveler/ref/repository_iterator_test.go
itai-david: Yicks.
Done - thanks

pkg/graveler/ref/repository_iterator.go
itai-david: :+1:

pkg/graveler/ref/manager.go
itai-david: How about the repo? Try to delete that too?
How do we handle failure in the delete?
What about failures such as crash after repo is created?

pkg/graveler/ref/manager.go
itai-david: I'm keeping this on hold at the moment, as I would like to discuss a possible solution with `repositoryStatus`
If that solution is a go, we would still need the internal `createBareRepository`, as the public `CreateBareRepository` will set the repository status right after creation, unlike `CreateRepository`

pkg/graveler/ref/manager_test.go
```diff
@@ -770,7 +770,8 @@ func TestManager_GetCommitByPrefix(t *testing.T) {
 	}
 }
 
-func TestManager_FillGenerations(t *testing.T) {
+// TODO: temporarily disabled due to dependency in GetRepository
```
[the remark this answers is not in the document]
itai-david: Yes we can!!!
#3656

pkg/graveler/ref/repository_iterator.go
itai-david: That is a good question.
I would vouche for keeping the same behavior, at least for the phase when we supporting both.
If you believe a change of behavior is beneficial, we can open an issue to handle it once transition is complete, and SQL support is abandoned.
WDYT?

pkg/graveler/ref/manager.go
itai-david: Done

pkg/graveler/graveler.go
itai-david: Right. Done

pkg/graveler/ref/db_manager.go
itai-david: Done

pkg/graveler/ref/manager.go
itai-david: Absolutely not!
I still need to figure out the correct (or, should I say, best) error handling in that case.
Please refer to my proposition on regarding the `repositroy_status`

pkg/graveler/ref/manager.go
```diff
@@ -47,23 +47,77 @@ func NewKVRefManager(executor batch.Batcher, kvStore kv.StoreMessage, db db.Data
...
+	if err != nil {
+		return err
+	}
+
+	err = m.CreateBranch(ctx, repositoryID, repository.DefaultBranchID, graveler.Branch{CommitID: commitID, StagingToken: token})
+	if err != nil {
+		return err
+	}
+
+	return nil
```
itaiad200, line 97:
```suggestion
	return m.CreateBranch(ctx, repositoryID, repository.DefaultBranchID, graveler.Branch{CommitID: commitID, StagingToken: token})
```

pkg/graveler/ref/manager.go
```diff
@@ -47,23 +47,77 @@ func NewKVRefManager(executor batch.Batcher, kvStore kv.StoreMessage, db db.Data
...
+	// temp code to align with DB manager. Delete once https://github.com/treeverse/lakeFS/issues/3640 is done
+	_, err := m.GetRepository(ctx, repositoryID)
+	if errors.Is(err, kv.ErrNotFound) {
+		return graveler.ErrRepositoryNotFound
+	}
+	if err != nil {
+		return err
+	}
+	// END TODO: delete me
+
```
itaiad200, line 119: We're also missing branches, commits, tags deletion, right?

pkg/graveler/ref/manager.go
N-o-Z: only the repo was created ~= bare repo => legal state, the user can then attempt to create the default branch manually
repo and commit created but no branch => dangling commit

If we fail the cleanup - then I guess we don't really have anything else to do...

pkg/graveler/ref/repository_iterator.go
N-o-Z: That's a good idea, we have several areas in the code that behave the same way. We should address all of them in the issue

pkg/graveler/ref/repository_iterator_test.go
itaiad200: And if we can also reuse the same test code for the db iterator 💪

itaiad200 (approved):

pkg/graveler/graveler.proto
```diff
@@ -5,6 +5,13 @@ import "google/protobuf/timestamp.proto";
 
 package io.treeverse.lakefs.graveler;
 
+message RepositoryData {
```
N-o-Z, line 8: I noticed that you didn't add the UUID to the repository data. Did we decide not to go with this solution eventually?

N-o-Z (changes requested):
@itai-david Remind me if we already had this discussion, did we remove the requirement for the repository UUID?

pkg/graveler/ref/manager.go
itai-david: Correct. This PR totally ignores the repository deletion process as it is handled in #3570

pkg/graveler/graveler.proto
itai-david: As we agreed:
* Added `func RepoPartition(repoID RepositoryID, _ Repository) string` to return the repository relevant partition key. Currently it returns the repo ID but eventually it will return some unique ID to identify the specific instance of the repository with the ID
* Decission on repository creation/deletion error handling and state-machine is extracted to another issue

N-o-Z (approved):
Thanks!

PR #3647, opened 2022-07-12
Add GUI templates expansion client

arielshaqed:
Returns the _text_ of the expanded response.  This may need to be changed to
just the plain response -- e.g. perhaps for downloading, or to avoid double
expansion in memory?

Tested by applying the attached diffs and _seeing_ a template expansion on the main page!
[test-expand-template.diffs.gz](https://github.com/treeverse/lakeFS/files/9091151/test-expand-template.diffs.gz)


Jonathan-Rosenberg (changes requested):
Cool cool cool cool
minor possible change

Jonathan-Rosenberg (approved):

PR #3654, opened 2022-07-13
Make GC read the expired addresses csv from Azure blob

### LINKED ISSUE #3545: Make GC write the expired addresses csv to Azure blob (via azure fs)
- Discovery: Which scheme we should translate Azure objects URIs to (wasbs, wasb, adl, etc).
Need to make sure that the documented azure filesystem Spark properties work with the scheme we choose. (e.g., spark.hadoop.fs.azure.account.key.<storage_account_name>.blob.core.windows.net=*********** works with wasbs) 
- Discovery: how do we recognize a URI of an object stored in Azure blob. (blockstore type, by gc configuration, by scheme, etc.)
- Implement scheme translation

### DOD

- GC writes the expired addresses csv to Azure blob (via azure fs)

lynnro314:
closes #3545

Added check of storage type and in case the storage type is azure - added schema conversion to abfs, in order to read and write with spark from azure blob. 
Used schema abfs, which is the [Hadoop Filesystem driver](https://docs.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-introduction-abfs-uri) that is compatible with Azure Data Lake Storage Gen2.
Tested locally with local lakeFS installation, once with storage type _data lake gen2_, and once with storage type _blob_.


clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -27,8 +32,22 @@ private object ApiClient {
...
+
+      /** regex to catch url of type: https://StorageAccountName.blob.core.windows.net/Container[/BlobName] and convert it to abfs url */
+      val regex = "https?://([^.]+)(\\.[^/]+)(?:/([^/]+)(.+))?"
+      val pattern = Pattern.compile(regex)
+      val matcher = pattern.matcher(uri.toString())
+      if (matcher.find()) {
+        return new URI(
+          s"abfs://${matcher.group(3)}@${matcher.group(1)}.dfs.core.windows.net${matcher.group(4)}"
+        )
+      }
```
arielshaqed, line 45: This approach assumes that Azure URLs have a particular form.  But a quick search seemed to reveal [the host can be modified](https://docs.microsoft.com/en-us/azure/storage/blobs/storage-custom-domain-name?tabs=azure-portal).  I do not thing it will work in that case.

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -27,8 +32,22 @@ private object ApiClient {
...
               uri.getFragment
              )
-    else
+    } else if (storageType == StorageTypeAzure) {
+
+      /** regex to catch url of type: https://StorageAccountName.blob.core.windows.net/Container[/BlobName] and convert it to abfs url */
+      val regex = "https?://([^.]+)(\\.[^/]+)(?:/([^/]+)(.+))?"
+      val pattern = Pattern.compile(regex)
+      val matcher = pattern.matcher(uri.toString())
+      if (matcher.find()) {
```
arielshaqed, line 41: Can we use [`Regex.replaceFirstIn`](https://www.scala-lang.org/api/2.12.5/scala/util/matching/Regex.html#replaceFirstIn(target:CharSequence,replacement:String):String)?  I'd be disappointed if we _cannot_...

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -27,8 +32,22 @@ private object ApiClient {
               uri.getQuery,
               uri.getFragment
              )
-    else
+    } else if (storageType == StorageTypeAzure) {
+
+      /** regex to catch url of type: https://StorageAccountName.blob.core.windows.net/Container[/BlobName] and convert it to abfs url */
+      val regex = "https?://([^.]+)(\\.[^/]+)(?:/([^/]+)(.+))?"
```
arielshaqed, line 38: These regexes ("regexen") would be easier to understand ("grok") if they did _not_ include the periods inside the matches.  Then you'd have to add the periods back in in the replacement string, but it would probably be easier for me to understand.

arielshaqed, line 38: Another way to do it, and one that I would prefer: Parse the URL.  Now you have a `host` part (that you can split on the first dot to get groups 1 and 2), followed by a path (that you can split on the first slash to get groups 3 and 4).  It should be more readable.

Context: JWZ said:
> Some people, when confronted with a problem, think "I know, I'll use regular expressions." Now they have two problems.

(Who's JWZ?  JWZ wrote Netscape Navigator (which became Netscape Communicator, which became Firefox).  Before that he wrote XEmacs (which died so that Emacs could continue to live).  Go to https://www.jwz.org/hacks/ and search for `gdb-highlight.el`.)

arielshaqed (changes requested):
Thanks!

I'm worried that the not all Azure Blobstore URLs will look like this code assumes they do.  If the block adaptor is Azure then the storage namespace might not be under windows.net; we should make an effort if we can to extract a good abfs URL there, too.

Note that this code is AFAIR _also_ used to access metaranges and ranges -- not just the CSV file generated by lakeFS.

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
talSofer: @lynnro314 to verify if lakeFS supports this URL modification, we should try to create a lakeFS repository with a storage namespace and see if lakeFS support that at all. If this doesn't work I think that we are good to go because GC only deletes objects that lakeFS wrote to a repo storage namespace. @arielshaqed WDYT?

(The suspect that may fail this this https://github.com/treeverse/lakeFS/blob/2358e8dd7f72af67844413f7177b7e3c94ee1c71/pkg/block/azure/adapter.go#L175 line of code we've looked at together)

talSofer (changes requested):
Thanks @lynnro314! 

Can you please add the following to the pr description:
- What it means that you tested with gen2 and blob so that we can reproduce testing easily later on? 
- A usage example for how did you ran GC (including the azure details and no secrets of course)
- The Spark configurations that abfs requires

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
arielshaqed: Yes, but please bear in mind:

1. If _we_ cannot configure it to work, that does not mean a greater Azure expert might not be able to configure it to work :-)
1. It might be fixed in lakeFS someday, we might not know (or I will know it was fixed but forget this PR), and then GC will stop working.

So I agree that if it is hard to do correctly right now we do not have to do it, but if it is easy to do then I would rather we tried.

arielshaqed (approved):
Neat, thanks!

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -13,22 +14,40 @@ import java.util.concurrent.TimeUnit
...
+                    )
+    } else if (storageType == StorageTypeAzure) {
+
+      /** get the host and path from url of type: https://StorageAccountName.blob.core.windows.net/Container[/BlobName],
+       *  extract the storage account, container and blob path, and use them in abfs url
+       */
+      val storageAccountName = uri.getHost.split('.')(0)
+      val pathSplit = uri.getPath.split("/", 3)
+      val container = pathSplit(1)
+      val blobPath = pathSplit(2)
```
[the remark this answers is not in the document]
lynnro314: Thanks :)

clients/spark/build.sbt
```diff
@@ -104,7 +105,7 @@ lazy val spark3Type =
 
 // EMR-6.5.0 beta, managed GC
 lazy val spark312Type =
-  new BuildType("312-hadoop3", scala212Version, "3.1.2", "0.10.11", "3.2.1", "hadoop3", "hadoop2-2.0.1")
+  new BuildType("312-hadoop3", scala212Version, "3.1.2", "0.10.11", "3.2.1", "hadoop3", "hadoop3-2.0.0")
```
talSofer, line 108: What does hadoop3-2.0.0 mean? is this change related to this review or fixes an existing issue?

talSofer (approved):
Cool! thank you!

One minor comment, can you edit 
> Tested locally with local lakeFS installation, once with storage-namespace of type data lake gen2, and once with storage storage-namespace of type blob.

to specify the concrete URL structures please? I mean, without prior knowledge I can't understand how to test with both type myself

PR #3657, opened 2022-07-14
Proposal: declarative views

ozkatz:
This is pending a few user interviews but based on prior user conversations - spanning 3 different use cases:

1. manage inputs for ML and DS experiments at the data consumer side (vs commits that are on the producer side)
2. Support use cases where users might want to [access the same branch across repos](https://lakefs.slack.com/archives/C02CV7MUV4G/p1657291958714669)
3. could possibly allow a server-side router-fs-like solution where data is composed from external sources

### DIFF
--- a/design/open/declarative-views.md
+++ b/design/open/declarative-views.md
@@ -1,0 +1,70 @@
+# Proposal for declarative scoped repositories
+
+
+### Goals
+
+Make it easy for users to compose data coming from different sources (lakeFS repositories & references, external object store locations) as input for processing jobs and data science experiments.
+
+Today, simply using tags and attaching job/notebook URI to a lakeFS commit allows for reproducibility, but has a couple of limitations:
+
+1. We explicitly tie all input data to a single repository. Usually the scope of the repository is determined by the writer and not the reader - this means that input data could possibly come from multiple repos and requires tracking the inputs in some other place.
+1. The inverse is also true - a given repository could contain a lot more data than what was actually required for a run/experiment/transformation/etc.. To ensure reproducibility in that case, we need to either somehow know what input is actually _used_ by that run, or ensure the lakeFS commit is never garbage collected entirely. Both hard to manage.
+1. In some cases, we know a process is deterministic, so all we want to do is map the input and the transformation without actually saving the output. This is especially useful when we run multiple experiments with the same input data and different parameters but storing all outputs is expensive
+
+
+### Design
+
+Provide a layer similar to a Dockerfile - compose a "virtual repository" or view in a declarative, layered manner.
+A set of API endpoints will be used to create and manage these "images".
+
+Here's what a `Lakefile` could look like:
+
+```dockerfile
+# not required, but we can use an entire ref as a "base layer"
+FROM lakefs://repo1/tag1
+
+# Or implicitly use
+FROM scratch
+
+# OR inherit from another image
+FROM lakefs://image:v123
+
+# Compose additional inputs
+ADD "lakefs://repo2/commit123/path/to/collection/"                 "path/to/collection/"
+ADD "lakefs://repo2/commit456/path/to/collection/date=11072022/"   "path/to/collection/date=11072022/"
+ADD "lakefs://repo4/tag123/pictures/cats/"                         "pictures/animals/cats/"
+
+# Optionally, if the submitting user has a `fs:ImportFromStorage` permission we can also support:
+ADD "s3://my-bucket/some-path/"                                    "external/some-path/"
+
+# Let's also support labels as used by Dockerfiles! this is useful to point to external locations, map dependencies and other metadata
+LABEL "author=oz.katz"
+LABEL "notebook_url=https://dbc-123.cloud.databricks.com/?o=12345#notebook/45678"
+LABEL "spark_version=3.3.2"
+LABEL "spark_terraform=https://github.com/org/terraform/tree/c54624/spark/"
+```
+
+Along with this format, we also define a set of endpoints:
+
+* `POST /api/v1/images/<image_tag>` - accepts a `Lakefile` and creates a lakeFS "image" for the given image tag specified. Returns the URI and metadata for the image.
+* `GET /api/v1/images/<image_tag>` - returns URI and metadata for the given image by its tag
+* `GET /api/v1/images` - list images
+* `DELETE /api/v1/images/<image_tag>` - delete an image
+
+A set of IAM permissions:
+
+* `fs:ReadImage` - Get information about an image
+* `fs:CreateImage` - Create a new image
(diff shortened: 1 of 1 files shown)


arielshaqed (commented):
Neat!  This is a great way to resolve pretty much the same issues that boto-router and routerFS handle.  I think that it would be (much) better and easier to use.

I am missing the semantics for looking things up.  Is this going to be an overlay FS?

Even more, I am missing the semantics for _writing_ stuff.  Docker solves this by effectively _ignoring_ the issue, IIRC the overlays behave as though they were read-only and you end up writing a completely separate layer on top of them all.  (You cannot see what really happened, because the image itself is readonly!)

Additionally: if I'm not mistaken, this looks like a completely different service than lakeFS: it's basically an Overlay Object Store.  Separating it from lakeFS means that you don't have to answer tricky questions about writing (hard!), branching (harder!!), committing (hardest!!!) and merging (even hardester!).

design/open/declarative-views.md
```diff
@@ -0,0 +1,70 @@
...
+
+
+### Goals
+
+Make it easy for users to compose data coming from different sources (lakeFS repositories & references, external object store locations) as input for processing jobs and data science experiments.
+
+Today, simply using tags and attaching job/notebook URI to a lakeFS commit allows for reproducibility, but has a couple of limitations:
+
+1. We explicitly tie all input data to a single repository. Usually the scope of the repository is determined by the writer and not the reader - this means that input data could possibly come from multiple repos and requires tracking the inputs in some other place.
+1. The inverse is also true - a given repository could contain a lot more data than what was actually required for a run/experiment/transformation/etc.. To ensure reproducibility in that case, we need to either somehow know what input is actually _used_ by that run, or ensure the lakeFS commit is never garbage collected entirely. Both hard to manage.
```
johnnyaug, line 11: What's the concern here? That GC deletes data I need for my run? Or is there something else that I'm missing.

design/open/declarative-views.md
```diff
@@ -0,0 +1,70 @@
...
+1. In some cases, we know a process is deterministic, so all we want to do is map the input and the transformation without actually saving the output. This is especially useful when we run multiple experiments with the same input data and different parameters but storing all outputs is expensive
```
johnnyaug, line 12: Not sure I followed this one. Is the problem here that we save the same output many times? Or that there are a lot of possible results and I don't necessarily want to save all of them? Something else?

johnnyaug (commented):
Very interesting read, I love this idea. Some notes:
1. I think allowing external sources is a must. Moreover, it also has a great potential on its own: an image can be a seed for a real repository. I can combine multiple data sources into a single lakeFS repository, and get versioning on top of all my sources without copying anything.
2. My biggest question is what do I actually do with the image once it's created? Are "views" visible as repositories in the UI? How can I access them?
3. Did not entirely understand the Goals section - added some requests for clarification.

nopcoder:
Like the idea, here are some thoughts I had while reading this one (No order or connection between the items):

- Views like images can be static and live - having live views can help integrations or Boto like use cases. Very similar to the idea to handle the "Boto" routing on the server side.
- The suggested idea can be relevant as data tool, not just part of lakeFS. We can copy data and build data, while in the case of lakeFS we will have it with zero copy.
- Instead of _Lakefile_ we can have a manifest like JSON file that can be processed as part of repository creation. Having a fast way to create repository with initial state. Include an option to create an immutable repository, and/or branchless repository (like git without master branch), and we can leverage the current lakeFS capabilities without introducing a new entity. The repository can be accessed without branch information, and it will include the view we built.

design/open/declarative-views.md
```diff
@@ -0,0 +1,70 @@
...
+* `fs:CreateImage` - Create a new image
+* `fs:AttachStorageNamespace` - This already exists, but should probably be used since an image should be stored somewhere..
+* `fs:ImportFromStorage` - Also exists, could be used if we wish to support mapping external object store paths
+* `fs:DeleteImage` - Delete an image
+
+
+### Open Questions
+
+1. Do we want to support external locations not from lakeFS? This could be achieved (albeit with some more work by the user) by defining a repository that imports an external location and then using that instead.
+1. Syntax - does the `Dockerfile` analogy make sense? is there a better/more intuitive way to achieve this declaratively?
```
ortz, line 66: I think that the intuitive way for DS to experiment with data is using Notebooks.
Can't we try to achieve the same using a similar approach?
My concern is that users will find this approach unintuitive (managing the `Lakefile` in some text editor, switching to a terminal and running `lakectl` manually, compared to Notebook where they don't switch context)

I think that this kind of file is great for automated processes where you still want to have access to multiple repositories, branches, and technologies and can be used for systems like CI.
If that's the case, maybe adding functionalities like `ENTRYPOINT` and `RUN` would make sense.

design/open/declarative-views.md
```diff
@@ -0,0 +1,70 @@
...
+1. Using images as input - what does it look like in practice? how are they accessed using the different clients (s3gw, HadoopFS, lakeFS objects API and SDKs...)
```
ortz, line 67: I think the term itself is a bit confusing, eventually, we'll maintain another reference, it might be a cross-repo one, or even cross tech, but it's still pointers and metadata.

design/open/declarative-views.md
```diff
@@ -0,0 +1,70 @@
...
+1. How do we handle images when doing Garbage collection?
+1. How do we handle non-immutable refs? Not just external storage but also branch HEADs and anything based on HEADs (i.e. `main~10`). Warn about it on creation? Disallow?
```
ortz, line 69: I wouldn't block it, but, I'd print a warning, this file could be used for a sort of Continous Integration system where clients that are working with multiple technologies in their data stack, would benefit from this approach to test things changed.

design/open/declarative-views.md
```diff
@@ -0,0 +1,70 @@
...
+
+# Or implicitly use
+FROM scratch
+
+# OR inherit from another image
+FROM lakefs://image:v123
+
+# Compose additional inputs
+ADD "lakefs://repo2/commit123/path/to/collection/"                 "path/to/collection/"
+ADD "lakefs://repo2/commit456/path/to/collection/date=11072022/"   "path/to/collection/date=11072022/"
```
itaiad200, line 34: Will this `ADD` override any previous data from `lakefs://repo2/commit123/path/to/collection/date=11072022/`

design/open/declarative-views.md
itaiad200, line 69: Another option is to treat it as a onetime reference. Translate `main~10` during image creation to the actual commit and stick to it. Won't work for the HEAD of the branch though.

nopcoder (approved):

PR #3661, opened 2022-07-17
repository operation error handling design

itai-david:
Design proposal for adding a `state` and a `unique_identifier` to `Repository`, to correctly handle `Repository` deletion and errors during creation and deletion.

Please review and add your opinion on the 3 open question at the end


design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
+# Respository Operations Error Handling
+
+## Abstract
+As part of porting our DB from SQL to KV Store we are facing certain limitations forced by the later. One of the main differences between the to is the support of transactions. While SQL DBs enjoy the benefit of atomic transactional operations, and so can bind several DB operations and guarantee the completion, or rollback, of the entire transaction, KV Stores does not support this attribute and each operation has to stand on its own.
```
N-o-Z, line 4:
```suggestion
As part of porting our DB from SQL to KV Store we are facing certain limitations forced by the latter. One of the main differences between the two is the support of transactions. While SQL DBs enjoy the benefit of atomic transactional operations, and therefore can bind several DB operations and guarantee the completion, or rollback, of the entire transaction, KV Stores does not support this functionality and each operation has to stand on its own.
```

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+This document purpose is to describe the way a `Repository` is currently being created and deleted by `graveler`, describe the challenges it poses when using it over KV Store and propose possible solutions
+
+## Challenges
+### CreateRepository
+`CreateRepository` operation create a `Repository` entry, a default `Branch` entry and first `Commit`. All DB operations are executed under a single transaction and so, a failure at a later step (e.g. `Commit` or `Branch` creation) is rolled back and the DB remains consistent and clean (i.e. No `Repository` DB Entry without correlated default `Branch` and first `Commit`).
+With KV in mind, the transaction protection is absent and the above transaction is translated to 3 stand-alone operations. A failure in a later step does not derive an automatic cleanup of the previously successful steps and so a failure in creating the default-branch will leave the DB with a Repository entry with no default-branch associated. This Repository is unusable on one hand, and cannot be recreated on the other, as there is a "valid" repository entry in the KV Store
```
N-o-Z, line 10: This is also true for the commit created

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+
+### CreateBareRepository
+`CreateBareRepository` is brought here as it uses the same logic to create the `Repository` entry. The `Repository` is created with neither a default `Branch` nor an initial `Commit`. At first sight, it seems that failure in `CreateRepositry`, after the `Repository` entry is created, can be treated as a successful `CreateBareRepository` but this is not the case, as `CreateBareRepository` is a plumbing command that is meant to be used alongside `RestoreRefs`. Creating a `bare Repository` is not part of the common usage of `lakeFS` and should be treated as such
```
N-o-Z, line 13: As I see it - when creating bare repository its state should be init as in the normal flow. Since this is a plumbing operation, we can work around the state to perform the load refs and if the load fails move it to a failed state

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+
+### DeleteRepository
+`DeleteRepository` operation cleans a `Repository`, and all its correlated entities, from the DB. That is, all `Branch`, `Commit` and `Tag` entries correlated to that repository are deleted in a single transaction. Any failure during the above, fails the entire transaction.
+When using KV, as the transaction protection is absent, each deletion can fail independently, making the operation partially done, or in any unknown state. This can be problematic in various aspects, as the DB needs to maintain consistency - an entity (e.g. a `Branch`) that is not correlated to a valid `Repository` cannot be allowed to be accessible. On a less severe aspect, having it lying in our KV as an unreachable object, is also not ideal, although it does not compromise the correctness
```
N-o-Z, line 17: For this solution to be completed, we need to add states for branch, commit and tag.
This will also pave our way to long running hooks.

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Solution
+The proposed solution introduces 2 new elements to solve the above, in terms of DB consistency a.k.a **Correctness**:
+* A `repository_state` that will allow to correctly differ `Repository` in use, from an useless `Repository` due to error, deletion etc
+* A `unique_identifier` for each `Repository` that will allow to differ it from other (previously created and possibly deleted) `Repository` that had the same `RepositoryID`. This will allow to correlate entities to the correct `Repository` entity, in case of `RepositoryID` collision (due to entities left from a previously deleted `Repository` with the same `RepositoryID`)
+In addition:
+* A garbage cleaning mechanism will remove zombie entities, and maintain our KV Store clean and as compact as possible. This GC will identify useless `Repository` entities and will reattempt to delete all the correlated entities, which are now easily recognizable with the addition of the new `unique_identifier`. Note that this GC is not a part of maintaining the DB consistency
```
N-o-Z, line 28: Should be applicable to all graveler entities

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+These unused `Repository` entries, which now can be easily detected, can later be cleaned by a future garbage cleaner, to maintain the KV cleanliness and to reduce its size.
+The new `repository_state` attribute can have one of four values: `initial`, `active`, `deleted` or `failed`
+* Upon creation, a `Repository` entry will have an `initial` state, to indicate it is not usable yet. It will also indicate that this `Repository` is soon to be valid for usage, so other threads/processes treat it as such
+  * A failure in the next creation steps - default `Branch`, initial `Commit` or any other future step, will transfer the `Repository` to state `failed`, if possible. If that cannot be achieved (e.g. the creating process crashed, or KV connectivity is lost) the `Repository` can stay in state `initial` as it is still unused
+  * If during creation, a `Repository` with identical ID (name) is found, the creation of the new `Repository` is stopped before any update the the `KV` is done. This is similar to the current behavior and will be done regardless of the existing `Repository` status
```
N-o-Z, line 36:
```suggestion
  * If during creation, a `Repository` with identical ID (name) is found, the creation of the new `Repository` is stopped before any update to the `KV` is done. This is similar to the current behavior and will be done regardless of the existing `Repository` status
```

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+* When retrieving a `Repository` it will only be retrieved if it is in `active` status. Otherwise an error will be returned.
+  * If the `Repository` entry does not exist, an `ErrRepositoryNotFound` will be returned
```
N-o-Z, line 38: I think we should just return ErrRepositoryNotFound for any state other than active

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+  * If the `Repository` entry exists and is not in `active` state - **TBD #1** - should we return `ErrRepositoryNotFound` here too? Maybe a new error - `ErrRepositoryDeletionInProgress`-ish?
+* Deleting a `Repository` will be done by first setting the `Repository` state to `deleted`, making it immediately unavailable. Once the `Repository` is set to deleted all its correlated entities can be removed safely, as none of them should be accessible. In case of a failure at any step, the `Repository` and all its entities remain unreachable due to the `deleted` status.
```
N-o-Z, line 40: Why do we actually need the deleted state? Once we move it to failed state - it is inaccessible and eligible for deletion

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+  * A failure to set the `Repository` state to `deleted`, will result in no change to `KV` and so the data remains consistent. The error from the failed update operation will be returned
+  * Once all correlated entities were successfully deleted, the `Repository` entry itself will can be deleted too
```
N-o-Z, line 42:
```suggestion
  * Once all the associated entities were successfully deleted, the `Repository` entry itself can be deleted too
```

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+  * Any failure to delete any entity during the deletion process, still leaves the DB in consistent state as the `Repository` entity, and by that its correlated entities, are unreachable due to the `deleted` state
```
N-o-Z, line 43: 😃

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+
+![Repository State Transitions](diagrams/repository-state-transitions.png)
+[(excalidraw file)](diagrams/repository-state-transitions.excalidraw)
+
+### Repository Unique Identifier
+Each `Repository` will have a `unique_identifier` (To be discussed later) which can differ it from other repositories with the same **Name** (`RepositoryID`). This will provide us with the ability to differ object correlated with different repositories with the same **Name**, from different points in time. Each repository will have a designated KV partition, identified by its **Name** and `unique_identifier` and all correlated objects will be handled in this partition, making it possible to identify objects correlated only to an `active` `Repository`. Moreover, as the `unique_identifier` is an attribute of the `Repository`, in order to create the partition name, the `Repository` entity must be acquired, and so non-exist or unusable `Repository` entries will make unreachable partitions and, in turn, unreachable entities, making it safe to use any reachable entity
```
N-o-Z, line 49: It also allows us to have two repositories with the same name, while one is being deleted and the other being created

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+  * If failed - try to set the `Repository` entry state to `failed` and return the error from the KV layer
+  * If cannot set the `Repository` state to `failed` it is still consistent, as it is still in `initial` state
+* Set the `Repository` state to `active`
+  * If failed - try to set the `Repository` entry state to `failed` and return the error from the KV layer
+  * If cannot set the `Repository` state to `failed` it is still consistent, as it is still in `initial` state
+* Success
+
+### CreateBareRepository
+Just the repository is created so this is pretty trivial
+* Create a `Repository` entity with its given `RepositoryID`, randomly generated `UniqueID` and `State` = `active`, under the common `graveler` partition
```
N-o-Z, line 73: See previous comment - creating a bare repository and setting it to active exposes us to race conditions and is not really necessary in this scenario

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+* Scan through the rest of the partition and delete all objects. That is `Commit`s and `Tag`s
+  * With any failure, return the error
+* If all deletions above are successful, delete the `Repository` itself
+  * If failed return the error
+
+### ListRepositories
+* List repositories should only return `active` entities. That can be achieved by identifying and skipping `Repository` entries with status other than `active`
+
+### Other Operations During Repository Creation/Deletion
+* All operations that require a valid `Repository` should start with `GetRepository`. This will prevent the usage of `Repository` in mid creation (with SQL DB it is impossible to reach this entity)
```
N-o-Z, line 102: 👍🏽

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+* If a `Repository` is being deleted when it is assumed to exists (e.g. `CreateBranch` operation that starts when the `Repository` exists and `active`, then the `Repository` is removed and only then the `Branch` is created in the KV) will leave the created entity unreachable, as the `Repository` does no longer exist (or is not `active` anymore) and so, though the newly created entity exists, it exists as unreachable garbage. This is similar of an entity being created, than the `Repository` is deleted and the entity removed with it
```
N-o-Z, line 103: This is different since, the repository delete won't clean up the branch, and we will not have anyway of knowing this branch is unreachable.
We might want to perfect the 'GC' so that it checks for this scenarios as well

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Garbage Collection - Nice to Have
```
N-o-Z, line 105: The 'GC' design and logic should be in the Graveler scope and address all entities

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+The logic is very simple, given the solution above is implemented:
+* Scan through all the `Repository` entities in the common `graveler` partition
+* For each `Repository`:
+  * If it is in `active` state - ignore
+  * If it is in `deleted` or `failed` state - delete it applying the same steps as described above for `DeleteRepository`
+  * If it is in `initial` state - decide if it should move to `failed` and if so, set to `failed` and try to delete, as described above
+Run periodically (**TBD** How often?) as a background worker and upon error just skip the `Repository`.
+
+## Open Questions
+1. What error value should we return for a `GetRepository` when the repository is in `deleted` or `failed` state?
```
N-o-Z, line 116: CreateRepository doesn't have to rely on GetRepository result, it can have an internal get which returns the repository if it exists in the DB to check the state or have a designated getRepoState. In any case - to the end user it should always be `ErrRepositoryNotFound`

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+  * `ErrRepositoryNotFound`? (might be a problem as a user may attempt to create such a `Repository` and fail).
+  * Some new `ErrResourceBusy`/`ErrRepositoryDeletionInProgress`?
+2. `Repository`'s `unique_identifier` - nanoid? xid? `RepositoryID`+creation-timestamp?
```
N-o-Z, line 119: I think this is an implementation details that is out of the scope of the design

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+  * Crucial to decide now as this is a major part of the implementation
+3. What is a sufficient timeout to decide an `initial` `Repository` is `failed`?
```
N-o-Z, line 121: I think we can be generous in this case

N-o-Z (changes requested):
Nicely done!
I have a couple of suggestion regarding the design, but my main points are:

- Suggest to remove the retry mechanism
- Reduce number of states (remove delete)
- 'GC' as a graveler service

I think we need to extend the states idea to the rest of the graveler entities for the following reasons:

- Support transaction flows (e.g. branch delete - need to delete all tokens (staging + sealed)
- Pave way to long running hooks

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+
+### GetRepository
+* Get the `Repository` with the given `RepositoryID` from the common `graveler` partition
+  * If not found return `graveler.ErrRepositoryNotFound`
+  * If found and status equals `active` - return the `Repository` entity - Success
+  * Otherwise, return error (according to the state? **TBD #1**)
+
+### DeleteRepository
+* Get the `Repository`, using the `GetRepository` flow above 
+  * In case of error return the error
```
itaiad200, line 84: Even if that `status` isn't `active`? Should the `Delete` flow follow the `Get` flow on these cases too?

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+* Set the `Repository` state to `deleted`
+  * If failed return the error
+* Generate the partition key for the `Repository`, from its `RepositoryID` and `unique_identifier`
```
itaiad200, line 87: Generate?

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+* Scan through the branches in the partition partition and for each `Branch`:
```
itaiad200, line 88: Should all of this happen during the `DeleteRepository` request? Without `DeletePartition` in `kv.store` this can take a long time.

design/open/repository-operations-error-handling.md
itaiad200: I think we should classify the garbage by severity and likelihood. If we keep the staging area for deleted branch then the likelihood is 100% and the amount of entries in kv is unbounded -> we should definitely treat that. If during  the request of a repo deletion someone creates a branch, the chances of that happening is low: it is very uncommon and requests needs to happen in a specific order. Also it will also create a single non-reachable garbage entry.
That being said, if the entity is reachable and can cause inconsistencies/failures for some flows, then we should prioritize accordingly.

design/open/repository-operations-error-handling.md
itaiad200, line 105: Can we name it differently? We don't want to confuse with lakeFS GC.

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+* If a `Repository` is being deleted when it is assumed to exists (e.g. `CreateBranch` operation that starts when the `Repository` exists and `active`, then the `Repository` is removed and only then the `Branch` is created in the KV) will leave the created entity unreachable, as the `Repository` does no longer exist (or is not `active` anymore) and so, though the newly created entity exists, it exists as unreachable garbage. This is similar of an entity being created, than the `Repository` is deleted and the entity removed with it
+
+## Garbage Collection - Nice to Have
+The above proposed solution provides the correctness needed for graveler. In addition, a dedicated garbage collection can help and maintain the KV Store cleaner and, as a result, smaller.
+The logic is very simple, given the solution above is implemented:
+* Scan through all the `Repository` entities in the common `graveler` partition
+* For each `Repository`:
+  * If it is in `active` state - ignore
+  * If it is in `deleted` or `failed` state - delete it applying the same steps as described above for `DeleteRepository`
```
itaiad200, line 111: But `deleted` is handled in the main flow of the `DeleteRepository`, won't it collide?

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+  * If it is in `initial` state - decide if it should move to `failed` and if so, set to `failed` and try to delete, as described above
+Run periodically (**TBD** How often?) as a background worker and upon error just skip the `Repository`.
+
+## Open Questions
+1. What error value should we return for a `GetRepository` when the repository is in `deleted` or `failed` state?
+  * `ErrRepositoryNotFound`? (might be a problem as a user may attempt to create such a `Repository` and fail).
+  * Some new `ErrResourceBusy`/`ErrRepositoryDeletionInProgress`?
```
itaiad200, line 118: 👍

design/open/repository-operations-error-handling.md
itaiad200: 2 minutes?

itaiad200 (changes requested):

design/open/repository-operations-error-handling.md
N-o-Z: One possible example - once repository is active user can try to create branches, which can conflict with the restore refs

design/open/repository-operations-error-handling.md
N-o-Z: I actually think this makes it simpler and prevents issues where the user can perform operations on the bare repository before / during RestoreRefs

design/open/repository-operations-error-handling.md
itaiad200: For creating branch you need a source ref (commit/branch) which doesn't exist in a bare repo.

design/open/repository-operations-error-handling.md
itai-david: @N-o-Z - what you are suggesting is a behavorial change that deserves a discovery and a design of its own. I don't think this change is something that can be done as a side effect of the current lets-make-KV-work-properly design. Things to consider, for example, what happens if I create a bare  repository, and only intened to perform `RestoreRefs` at a later time? Will the repository "declared" `failed` after a timeout?
Since `CreateBareRepository` is a single DB operation, it is not really affected (in terms of consistency) by the transition to KV and as far as I can see it, creating it with `inital` state rather than `active` is mainly a semantic difference, rather than operational.
If you do think it is meaningful and worth changing, let's fork this discussion to another issue. WDYT?

design/open/repository-operations-error-handling.md
itai-david: From a repository POV it does, as cleaning a repository includes cleaning all the entities in that repository.
If you are refering to clean an erroneous entities within an active repository - this should be based on error handling of these entities within the repository and should have its own discovery/design
Implementing a GC worker for graveler, as part of repositories cleanup, can later be expanded to handle other entities, should we decide to

design/open/repository-operations-error-handling.md
itai-david: Sounds legit to me.

design/open/repository-operations-error-handling.md
itai-david: The only thing I had in mind is the option to return a more targeted information to the user, to allow different handling. For example if `DeleteRepository` returns `ErrorNotFound` the user can try and create it, if it's `ErrDeleteInProgress` the user might wait and then try again, and for any other error the user can retry immediately or alike.

design/open/repository-operations-error-handling.md
itai-david: That would be impossible, as (currently) the 2 repositories uses the same key in KV, for the repository entries themselves. That is necessary as the `unique_identifier` is not part of the `Repository` path in KV, as it will not allow direct access.

I had a thought of creating a new key for each deleted/failed repository. Something like `/deleted_repos/<REPO_ID>/<UNIQUE_IDENTIFIER>`, or even a dedicated "deleted" partition, with /<REPOS_ID>_<UNIQ_ID> keys, just as a place holders for cleanup, instead of the /reps/<REPO_ID>.
This will allow multiple instances of the same `RepositoryID`, but only one active under `/repos/<REPOS_ID>`, and will support an immediate "recreation" of a deleted/failed repository, as the /repos/<REPOS_ID> key will be removed
However, that adds another level of complication, as in case of failure, there are 2 more KV actions to perform (addition of a key to the "deleted" partition and removal of the `/repo/<REPO_ID>` key from the common `graveler` partition, and since we got here due to (possibly) a KV failure to start with, this does not solve anything and is a best effort optimization at best

design/open/repository-operations-error-handling.md
itai-david: Agreed, if and when we apply some state logic and error handling to the entities themselves.
For the scope of this design (state and error handling for repositories) all entities are cleaned as part of the repository cleanup.
As the GC itself is not part of this design, I'm forking it to a new issue #3683

design/open/repository-operations-error-handling.md
itai-david: `CreateRepository` does not rely on `GetRepository`. This open question is regarding the error to return for `GetRepository` in general. As designed, the repository might be exist in `deleted` or `failed` state, which implies it neither exist, nor going to be "fixed", but the entry currently exist which prevent it from being recreated.
Just wondering if `ErrRepositoryNotFound` is sufficient

design/open/repository-operations-error-handling.md
itai-david: 2 minutes is generous enough IMHO

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+* Create a `Repository` entity with its given `RepositoryID`, randomly generated `UniqueID` and `State` = `initial`, under the common `graveler` partition
+  * If failed return the error (`graveler.ErrNotUnique` will be returned in case a `Repository` entity with the same `RepositoryID` exists) - entity was not created - no harm done.
+* Generate the partition key for the new `Repository`, from its `RepositoryID` and `unique_identifier`
+* Create the initial `Commit` with a `CommitID` (same as it is done today) and store it with the partition key generated above
+  * If failed - try to set the `Repository` entry state to `failed` and return the error from the KV layer
+  * If cannot set the `Repository` state to `failed` it is still consistent, as it is still in `initial` state
+* Create the default `Branch` set to the initial `CommitID` and store it with the partition key generated above
+  * If failed - try to set the `Repository` entry state to `failed` and return the error from the KV layer
+  * If cannot set the `Repository` state to `failed` it is still consistent, as it is still in `initial` state
+* Set the `Repository` state to `active`
```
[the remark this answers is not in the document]
itai-david: Done

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Cleaning Up an `initial` Repository
+A `Repository` entity with `initial` status is either being created (which will soon change the status to `active`), or already failed to create without changing status to `failed` (due to, for example, a crashed process). We will decide a `Repository` is failed if it is in `initial` state, and a sufficient time from its creation has passed. In that case, we will change the `Repository` state to `failed` and make him a candidate for deletion.
+This can be checked every time the `Repository` is accessed (trivial) and by a designated cleanup procedure (part of the above mentioned GC, or a designated scan at startup).
+ 
+##  Flows
+### CreateRepository
+* Create a `Repository` entity with its given `RepositoryID`, randomly generated `UniqueID` and `State` = `initial`, under the common `graveler` partition
+  * If failed return the error (`graveler.ErrNotUnique` will be returned in case a `Repository` entity with the same `RepositoryID` exists) - entity was not created - no harm done.
```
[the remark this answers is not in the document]
itai-david: Correct - fixed. Thanks for pointing this out

design/open/repository-operations-error-handling.md
```diff
@@ -0,0 +1,121 @@
...
+* Generate the partition key for the new `Repository`, from its `RepositoryID` and `unique_identifier`
+* Create the initial `Commit` with a `CommitID` (same as it is done today) and store it with the partition key generated above
+  * If failed - try to set the `Repository` entry state to `failed` and return the error from the KV layer
+  * If cannot set the `Repository` state to `failed` it is still consistent, as it is still in `initial` state
```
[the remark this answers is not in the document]
itai-david: @itaiad200 - which cleanup protocol are tou refering too? The set to `failed`? It is done to allow this repo to be cleaned in the next cleanup routine (whatever it may be), without waiting for the timeout (from TBD #3) to expire.
Other than that, no cleanup actions are done as part of `CreateRepository`

design/open/repository-operations-error-handling.md
itai-david: If the status isn't active,the first step of `GetRepository` will return an error which will stop the deletion on that spot.
Does it make sense?

itai-david: Oh, I see what you mean. Rephrased

design/open/repository-operations-error-handling.md
itai-david: Great idea!
Few options I have in mind:
* TrashCleaner
* Cleanser
* WasteManager - my favorite
All can either be preceded by `Graveler`, or not
WDYT? Any other ideas?

design/open/repository-operations-error-handling.md
itai-david: Actually it may. When writing this, I had the garbage collection in mind as an optional addition, to handle leftovers due to previous failures. In that case, I didn't want the `DeleteRepository` leave all the repos contents as garbage forever.
Once garbage collector is implemented (I belive not as part KV3) the Scan and Delete logic can move there, leaving `DeleteRepository` with just setting the repository as deleted

itai-david: Added a note to clarify that

itai-david:
> Nicely done! I have a couple of suggestion regarding the design, but my main points are:
> 
> * Suggest to remove the retry mechanism
> * Reduce number of states (remove delete)
> * 'GC' as a graveler service
> 
> I think we need to extend the states idea to the rest of the graveler entities for the following reasons:
> 
> * Support transaction flows (e.g. branch delete - need to delete all tokens (staging + sealed)
> * Pave way to long running hooks

@N-o-Z:
* Not sure which retry mechanism you are refering to. Can you please elaborate?
* Though removing the `delete` state is possible, I think it should stay as a semantically more appropriate state. WDYT?
* `GC` (or athever name it will end with) design issue created - #3683 
* States and Error handling for other entities is out of this scope - created #3684 for Discovery and Design

design/open/repository-operations-error-handling.md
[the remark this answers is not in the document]
itaiad200: Exactly - it's an optimization that we should introduce at a later stage, since we have to handle stuck `initial` state now anyhow.

design/open/repository-operations-error-handling.md
N-o-Z: Referencing the previous comment: What difference does it make for the user if the repository is currently being deleted? For all things considered this repo doesn't exist and the user should be able to create a new one with the same name immediately after calling delete

design/open/repository-operations-error-handling.md
N-o-Z: meta_axolotler

design/open/repository-operations-error-handling.md
```diff
@@ -1,13 +1,13 @@
 # Respository Operations Error Handling
```
N-o-Z, line 1: I like the delete repo flow, I think it covers everything.What I like less is the introduction of new terminology. I would prefer we drop the 'disposeRepository'

design/open/repository-operations-error-handling.md
```diff
@@ -27,73 +27,95 @@ The proposed solution introduces 2 new elements to solve the above, in terms of
...
-  * If failed - try to set the `Repository` entry state to `failed` and return the error from the KV layer
-  * If cannot set the `Repository` state to `failed` it is still consistent, as it is still in `initial` state
-* Create the default `Branch` set to the initial `CommitID` and store it with the partition key generated above
-  * If failed - try to set the `Repository` entry state to `failed` and return the error from the KV layer
-  * If cannot set the `Repository` state to `failed` it is still consistent, as it is still in `initial` state
-* Set the `Repository` state to `active`
-  * If failed - try to set the `Repository` entry state to `failed` and return the error from the KV layer
-  * If cannot set the `Repository` state to `failed` it is still consistent, as it is still in `initial` state
-* Success
+### GetRepository - **Step 1**
```
N-o-Z, line 52: In both step1 and step2, I think we should be proactive on failure, and try to clean up after ourselves

N-o-Z (approved):
@itai-david Thanks for the hard work on this 🦾 
The new diagrams are a big improvement!
Please see comments

design/open/repository-operations-error-handling.md
itai-david: `disposeRepository` was an intermediate idea, I was playing with and is no longer a part of this document. I forgot to remove the diagram file. Will be deleted

itai-david: Removed

design/open/repository-operations-error-handling.md
itai-david: Proactive cleaning (or cleaning at all) is not a requirement. This is refelcted in Step 1 where I propose (as derived from the conversation thread with Oz) the minimal work to get correctness in place.
Regarding proactive cleaning in case of failure - it (a) makes the failures longer - instead of quick error response to the user, (b) has no real benefit as you can still fail during the cleaning itself, which leads to (c) additional error handling code to handle a failure to cleanup after failure. Having all cleanup tasks in one place (i.e. designated Cleaner) will simplify the code as we will have one place to handle errors during cleanup

itai-david:
> @itai-david Thanks for the hard work on this mechanical_arm The new diagrams are a big improvement! Please see comments

Thank you @N-o-Z 
Please see answers to your comments. Redundant diagrams removed

design/open/repository-operations-error-handling.md
N-o-Z: We are talking specifically about create repository flow, so point (a) is irrelevant as we are just talking about deleting the initial commit.
point (b) it can, but we can still try
point (c) it can be a best effort and we don't have to handle errors at this point
The biggest problem I have with not trying to clean up is the fact that in step(1) if we fail on create branch, we will have a dangling commit that no cleanup process will be able to find

itai-david: Your point about the dangling commit (or a dangling branch too, as a failure not neccessarily means it was not created) is right, but the assumption is a repository is not an entity that is created too often, and it will fail even less frequently.

itaiad200 (approved):
LGTM

PR #3666, opened 2022-07-17
Proposal: Graveler stateful entities

N-o-Z:


design/open/graveler_stateful_entities.md
```diff
@@ -0,0 +1,61 @@
...
+All entities which currently support pre/post actions on creation and deletion must support states. These include:
+* Branches
+* Commits
+* Tags
+
+All entities which require support of transactional / concurrent / rollback operations:
+* Repositories
+* Branches
+
+## Supported states
```
itaiad200: State is an application level feature which is bound to the usages of the specific entity. Since every entity has a different usage patterns, I think that trying to align all entities to use the same list of states is as hard as writing the same code to handle all entities.

design/open/graveler_stateful_entities.md
```diff
@@ -0,0 +1,69 @@
...
+  * Once all tokens are deleted, safely delete the branch. 
+  * If token deletion fails, branch can remain marked `FAILED` state, for future deletion
+
+_Create Repository_  
+* Create repository in `INIT` state. 
+* Created default branch and initial commit. 
+  * If successful move repository state to `ACTIVE`
+  * If default branch creation failed - move repository to `FAILED` state (and possibly the commit), to be deleted in a future time.
+
+#### Long Running Hooks Example:
```
itaiad200, line 63: I think we should do this the other way around - Gather the requirements for long running hooks, perform a discovery on current/future features to be added to lakeFS, have a design and then decide what we need from graveler.

itaiad200 (commented):
I love the initiative and the way it tries to generalize the for from #3661 to fit any graveler entity, not just repos. As I see it, currently this design will mostly help with the removal of entities like branches and I think there are other ways to achieve the same for non-critical flows like removing entries from a partition. I do see us using something like `state` for all hooks-enabled graveler entities and I think we should design that first.

design/open/graveler_stateful_entities.md
```diff
@@ -0,0 +1,69 @@
...
+
+Terminal state - Inaccessible to the user, and a candidate for deletion
+
+## Error Handling
+
+Since error handling is defined the specific use case it will not be discussed under the current document, but generally 
+we will have 2 types of error handling:
+1. Setting entity to `FAILED` state for future deletion
+2. Performing a rollback operation for all the actions 
+3. Retry mechanism
```
arielshaqed, line 47: 3 types here :-)

IMO errors (everywhere) should be split into a _message part_ and a _disposition part_.

The message part is intended for operators and users.  So it is detailed, and might be split into "visible on the backend" and "returned to the user" bits.  Because of this it is useless for programs.

The disposition part is intended for the calling program.  It is extremely **undetailed**, and consists of multiple property fields that says what to do in various cases.  In fact, the only property that is really needed is "ephemeral": is there any point in retrying this operation if nothing else has changed?  So for instance "branch not found" is _not_ ephemeral: if nothing changes then there is no point in looking for the branch again.  However "staging token changed while writing object to branch" _is_ ephemeral: if you try (everything) again then you might succeed!

The reason for talking about "ephemeral error" rather than "retrying" (at the lowest level) is that retries are determined by the client as much as they are by the server.  So a client can add a retry policy, but retry only on ephemeral failures.  It can still have its retry policy controlling everything else :-)

design/open/graveler_stateful_entities.md
```diff
@@ -0,0 +1,69 @@
...
+* Set Branch state to `FAILED`. 
+* Perform staging token and sealed token deletion. 
+  * Once all tokens are deleted, safely delete the branch. 
+  * If token deletion fails, branch can remain marked `FAILED` state, for future deletion
+
+_Create Repository_  
+* Create repository in `INIT` state. 
+* Created default branch and initial commit. 
+  * If successful move repository state to `ACTIVE`
+  * If default branch creation failed - move repository to `FAILED` state (and possibly the commit), to be deleted in a future time.
```
arielshaqed, line 61: I think you need to create the initial commit before you can create the default branch.

For this to work each of these must have a unique ID and has to use a `SetIf` operation, right?  I.e. create repository in state `INIT` with a unique ID IF it is `nil` (does not exist), create an initial commit, create branch IF it does not exist (otherwise CREATE | DELETE | CREATE races), make respository failed IF it is the same repository and still in state `INIT`.

design/open/graveler_stateful_entities.md
```diff
@@ -0,0 +1,69 @@
...
+
```
arielshaqed, line 62: I think some more detail on how to delete a repository is missing here.  Especially if the repository is in state `FAILED`, and how to handle concurrent CREATE and DELETE operations.

design/open/graveler_stateful_entities.md
```diff
@@ -0,0 +1,69 @@
...
+#### Long Running Hooks Example:
+
+_Create Tag_  
+* Create Tag in `INIT` state
+* Run pre-create actions 
+  * On pre-create actions successful move tag to `ACTIVE` state
+  * In case of failure / timeout - delete tag.
```
arielshaqed, line 69: Not sure why we sometimes delete and sometimes move to state `FAILED`.

arielshaqed (commented):
Wow, nice going: This abstraction makes things so much easier to talk about!

I think that we have to think some more about concurrent operations: what to do in state FAILED, how to ensure concurrent CREATE | DELETE | CREATE | DELETE works, etc.  Unique object IDs may help here, perhaps unifying the two proposals will allow us to write complete flows and make things _less_ confusing?

design/open/graveler_stateful_entities.md
```diff
@@ -0,0 +1,69 @@
...
+
+All entities which require support of transactional / concurrent / rollback operations:
+* Repositories
+* Branches
+
+## Supported States
+
+### `INIT`
+
+The entity was created, in DB but not accessible to the user. Will move to `ACTIVE` state until initialization is completed.
```
itai-david, line 31:
```suggestion
The entity was created, in DB but not accessible to the user. Will move to `ACTIVE` state once initialization is completed. 
```

itai-david (commented):
Is this a draft?

Issue #3571, opened 2022-06-25
KV Graveler Benchmarks


itai-david:
Currently, there are no benchmarks for graveler. This can make it difficult to identify perfromance regression due to the move to KV.
Need to implement benchmarks and run on both SQL and KV.
The benchmarks should include **at least** a represneting set of operations for `pkg/graveler/ref` and `pkg/graveler/staging`, as these are the DB accessing packages

**DoD**
A representing benchmark that runs on SQL and KV and a conclusion regarding the performance quality of graveler over KV

itai-david:
**Note:**
A decision is to be made regarding the required benchmarks - which scenarios do we really want to test. One example is multiple parallel commits, that will definitely fail each other until all retries will succeed - while this is an exhaustive scenario, which stress the performance to an edge, it might not be as interesting as a "real life" scenario. The bottom line is we can predict that such a scenario will show performance degradation for our not-lock-free commit algoritm, but it is probably not interesting. Another example is a single long commit, with multiple staging operations taking place at the same time. This scenario is much more common in real life usage and a performance regression in this scenario is something we better discover sooner than later.

itaiad200:
**Test plan**
Perform the following operations on a lakeFS running locally with postgres container. Compare KV and DB results for:
1. Write 100K records.
2. Read 100K records.
3. Commit 100K records.
4. Listing 100K records.
5. Commit every 5 seconds while writing to the same branch.

itaiad200:
DB write:

Historgram (ms):
1	0
2	0
5	0
7	0
10	0
15	0
25	0
50	0
75	0
100	0
250	213
350	221
500	221
750	243
1000	9345
5000	10000
min	188
max	2242
total	10000

DB Read (all entries in staging):
1	0
2	0
5	0
7	0
10	0
15	0
25	1
50	21
75	334
100	2869
250	95060
350	99732
500	102049
750	102870
1000	102928
5000	103028
min	24
max	1228
total	103028

itaiad200:
KV Write:

Historgram (ms):
1	0
2	0
5	0
7	0
10	0
15	0
25	0
50	12
75	25
100	51
250	986
350	2612
500	5770
750	8953
1000	9805
5000	10000
min	29
max	2289
total	10000

KV Read:
Historgram (ms):
1	0
2	0
5	0
7	0
10	0
15	1
25	368
50	1508
75	1558
100	11316
250	44783
350	60504
500	76788
750	91387
1000	97953
[...]

PR #3669, opened 2022-07-17
KV Store: Ref Manager Branch Implementation

### LINKED ISSUE #3592: KV Graveler Implementation - Branch
Implement all branch operations over KV : CreateBranch, SetBranch, DeleteBranch and GetBranch
Implement branch_iterator without and with ordering by CommitID
See #3566 for further description of the said iterators

DoD
Implementation of branch operations over 'KV'.
KV graveler uses KV for these operation (in oppose to delegate them to the DB graveler)

N-o-Z:
Closes #3592 

- Please note there are still some open issues due to concerns raised in https://github.com/treeverse/lakeFS/pull/3646 and https://github.com/treeverse/lakeFS/pull/3666#pullrequestreview-1041106669


pkg/graveler/ref/branch_iterator.go
```diff
@@ -2,143 +2,168 @@ package ref
...
+func (b *BranchByCommitIterator) Err() error {
+	return nil
 }
 
-func (ri *BranchIterator) getOrderBy() string {
-	if ri.orderByCommitID {
-		return "commit_id"
-	}
-	return "id"
+func (b *BranchByCommitIterator) Close() {
```
itaiad200, line 168: You make it possible to read from a closed iterator.

pkg/graveler/ref/branch_iterator_test.go
```diff
@@ -44,7 +47,191 @@ func TestBranchIterator(t *testing.T) {
...
+		}
+		iter.Close()
+
+		if diffs := deep.Equal(ids, []graveler.BranchID{"aa", "b", "c", "d", "e", "main"}); diffs != nil {
+			t.Fatalf("got wrong list of branch IDs")
+		}
+	})
+}
+
+// TODO: Need repository PR
```
itaiad200, line 101: Still relevant?

pkg/graveler/ref/branch_iterator_test.go
```diff
@@ -44,7 +47,191 @@ func TestBranchIterator(t *testing.T) {
...
+		}
+		iter.Close()
+
+		if diffs := deep.Equal(ids, []graveler.CommitID{"a", "aa", "b", "c", "d", "e", "main"}); diffs != nil {
+			t.Fatalf("got wrong list of IDs: %v", diffs)
+		}
+	})
+
+	t.Run("listing branches using prefix", func(t *testing.T) {
+		iter, err := ref.NewBranchSimpleIterator(ctx, &kvStore, repo)
```
itaiad200, line 233: This is not `byCommitIteratorTest`.

pkg/graveler/ref/branch_iterator_test.go
```diff
@@ -62,7 +249,8 @@ func TestBranchIterator(t *testing.T) {
 	})
 
 	t.Run("listing branches SeekGE", func(t *testing.T) {
-		iter := ref.NewBranchIterator(ctx, db, "repo1", 3)
+		iter, err := ref.NewBranchSimpleIterator(ctx, &kvStore, repo)
```
itaiad200, line 252: This is not `byCommitIteratorTest`.

pkg/graveler/ref/branch_iterator_test.go
```diff
@@ -2,16 +2,19 @@ package ref_test
 
 import (
 	"context"
+	"fmt"
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/require"
+
```
itaiad200, line 10: Fix imports.

pkg/graveler/retention/garbage_collection_manager.go
```diff
@@ -145,7 +145,7 @@ func (m *GarbageCollectionManager) SaveGarbageCollectionCommits(ctx context.Cont
 		refManager:   m.refManager,
 		repositoryID: repositoryID,
 	}
-	branchIterator := ref.NewBranchIterator(ctx, m.db, repositoryID, 1000, ref.WithOrderByCommitID()) //nolint: gomnd
+	branchIterator := ref.NewDBBranchIterator(ctx, m.db, repositoryID, 1000, ref.WithOrderByCommitID()) //nolint: gomnd
```
itaiad200, line 148: This direct initilization of a DB implementation now makes `GarbageCollectionManager` the `DBGarbageCollectionManager`. Instead you can pass the `GarbageCollectionManager` a `branchIteratorFactory` that accepts params and returns the appropriate implementation (DB/KV) iterator.

itaiad200 (changes requested):
Thanks for the very hard work on this 💎

itaiad200 (changes requested):
Thanks for the very hard work on this 💎

pkg/graveler/ref/branch_iterator.go
```diff
@@ -2,143 +2,168 @@ package ref
...
-			FROM graveler_branches
-			WHERE repository_id = $1
-			AND id `+offsetCondition+` $2
-			ORDER BY `+ri.getOrderBy()+`  ASC
-			LIMIT $3`, ri.repositoryID, ri.offset, ri.fetchSize)
-	if err != nil {
-		ri.err = err
-		return
+func (bi *BranchSimpleIterator) Close() {
+	bi.err = ErrIteratorClosed
```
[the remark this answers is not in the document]
N-o-Z: I'm not sure - if we already closed the iterator - should we consider the inner iterator's error at all?

pkg/graveler/ref/branch_iterator.go
```diff
@@ -2,143 +2,168 @@ package ref
...
-	offset          string
-	fetchSize       int
-	err             error
-	state           iteratorState
-	orderByCommitID bool
+// CompareFunc type used for sorting in InMemIterator, it is a strictly bigger comparison function required for the
+// sort.Slice algorithm, implementors need to decide how to handle equal values
+type CompareFunc func(i, j int) bool
+
+type BranchIterator interface {
```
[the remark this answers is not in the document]
N-o-Z: You are correct

pkg/graveler/ref/manager.go
```diff
@@ -37,6 +37,31 @@ type CommitNode struct {
...
+	}
+	for _, st := range pb.SealedTokens {
+		branch.SealedTokens = append(branch.SealedTokens, graveler.StagingToken(st))
+	}
+	return branch
+}
+
+func protoFromBranch(branchID graveler.BranchID, b *graveler.Branch) *graveler.BranchData {
+	branch := &graveler.BranchData{
+		Id:           branchID.String(),
```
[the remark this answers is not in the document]
N-o-Z: Nope - preferably we would like to add it, and remove records entirely. But we decided it to look into it and take action later.

pkg/graveler/ref/manager.go
```diff
@@ -123,35 +160,103 @@ func (m *KVManager) ResolveRawRef(ctx context.Context, repositoryID graveler.Rep
 	return ResolveRawRef(ctx, m, m.addressProvider, repositoryID, raw)
 }
 
+func (m *KVManager) getBranchWithPredicate(ctx context.Context, repo *graveler.RepositoryRecord, branchID graveler.BranchID) (kv.Predicate, *graveler.Branch, error) {
```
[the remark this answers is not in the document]
N-o-Z: Just out of curiosity: any specific reason to switch the fields?

pkg/graveler/ref/manager.go
```diff
@@ -37,6 +37,31 @@ type CommitNode struct {
 	generation     int
 }
 
+func branchFromProto(pb *graveler.BranchData) *graveler.Branch {
+	branch := &graveler.Branch{
+		CommitID:     graveler.CommitID(pb.CommitId),
+		StagingToken: graveler.StagingToken(pb.StagingToken),
+		SealedTokens: make([]graveler.StagingToken, 0),
```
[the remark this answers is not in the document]
N-o-Z: You can't initialize a list without a size AFAIK

pkg/graveler/model.go
```diff
@@ -54,3 +55,18 @@ func ProtoFromRepo(repo *RepositoryRecord) *RepositoryData {
...
+	}
+}
+
+func ProtoFromStagedEntry(key []byte, v *Value) *StagedEntry {
+	return &StagedEntry{
+		Key:      key,
+		Identity: v.Identity,
+		Data:     v.Data,
+	}
+}
```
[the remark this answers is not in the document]
N-o-Z: It is, I will update the StagedEntry name to StagedEntryData, the thing is Value is used both for committed and uncommitted data so I can't change the name

pkg/graveler/graveler.go
```diff
@@ -862,15 +875,126 @@ func generateStagingToken(repositoryID RepositoryID, branchID BranchID) StagingT
...
+		iter, err := g.StagingManager.List(ctx, b.StagingToken, ListingDefaultBatchSize)
+		if err != nil {
+			return err
+		}
+		defer iter.Close()
+		if iter.Next() {
+			return ErrConflictFound
+		}
+		for _, st := range b.SealedTokens {
+			err = func() error {
```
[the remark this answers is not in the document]
N-o-Z: Poor attempt at optimization... 😄

pkg/graveler/retention/garbage_collection_manager.go
N-o-Z: I currently ignored GC completely - I will address it

pkg/graveler/graveler.go
```diff
@@ -862,15 +875,127 @@ func generateStagingToken(repositoryID RepositoryID, branchID BranchID) StagingT
...
+
+	reference, err := g.Dereference(ctx, repositoryID, ref)
+	if err != nil {
+		return nil, fmt.Errorf("source reference '%s': %w", ref, err)
+	}
+	if reference.ResolvedBranchModifier == ResolvedBranchModifierStaging {
+		return nil, fmt.Errorf("source reference '%s': %w", ref, ErrCreateBranchNoCommit)
+	}
+
+	_, err = g.RefManager.GetBranch(ctx, repositoryID, branchID)
```
itai-david, line 891: `GetBranch` hides a call to `GetRepository`, which should already be covered by the call to `GetRepository` at the beginning of the function. In addition, `CreateBrnach` below also calls `GetRepository` and I'm not sure how `Dereference` above works, but it might add another call (?)
I think we should implement kind of `getBranchInternal`, `createBranchInternal` etc, that receive, in addition to the current respective functions parameters, the `Repository` object retrieved by the first (and ideally only) call to `GetRepository`

pkg/graveler/graveler.go
```diff
@@ -1006,11 +1131,90 @@ func (g *KVGraveler) Log(ctx context.Context, repositoryID RepositoryID, commitI
...
+	return wg.Wait().ErrorOrNil()
 }
 
 func (g *KVGraveler) DeleteBranch(ctx context.Context, repositoryID RepositoryID, branchID BranchID) error {
-	return g.db.DeleteBranch(ctx, repositoryID, branchID)
+	var (
+		preRunID         string
+		storageNamespace StorageNamespace
+		commitID         CommitID
+	)
```
itai-david, line 1161: Why not simply use short declaration right before these are needed?

pkg/graveler/ref/branch_iterator_test.go
```diff
@@ -94,3 +96,191 @@ func TestBranchIterator(t *testing.T) {
...
+		if diffs := deep.Equal(ids, []graveler.CommitID{"aa", "b", "c", "d", "e", "main"}); diffs != nil {
+			t.Fatalf("got wrong list of branch IDs")
+		}
+
+		require.False(t, iter.Next())
+		require.Nil(t, iter.Value())
+		require.ErrorIs(t, iter.Err(), ref.ErrIteratorClosed)
+	})
+}
```
itai-david, line 286: Might be interesting to benchmark `BranchByCommitIterator` against the old `BranchIterator` with `orderByCommitID = true`, just to get a sense of how painful the sort is.
If it is not that significant we can peacefully go on, but if it is too painful, we might want to consider different approach.
Either ways, I think this insight can be useful

pkg/graveler/ref/manager.go
```diff
@@ -46,6 +71,18 @@ func NewKVRefManager(executor batch.Batcher, kvStore kv.StoreMessage, db db.Data
 	}
 }
 
+// getRepository returns RepositoryRecord
+func (m *KVManager) getRepository(ctx context.Context, repositoryID graveler.RepositoryID) (*graveler.RepositoryRecord, error) {
```
itai-david, line 75:
```suggestion
func (m *KVManager) getRepositoryRec(ctx context.Context, repositoryID graveler.RepositoryID) (*graveler.RepositoryRecord, error) {
```
It is quite confiusing

pkg/graveler/ref/manager.go
```diff
@@ -123,35 +160,103 @@ func (m *KVManager) ResolveRawRef(ctx context.Context, repositoryID graveler.Rep
...
+	return pred, branchFromProto(&data), nil
+}
+
 func (m *KVManager) GetBranch(ctx context.Context, repositoryID graveler.RepositoryID, branchID graveler.BranchID) (*graveler.Branch, error) {
-	return m.db.GetBranch(ctx, repositoryID, branchID)
+	repo, err := m.getRepository(ctx, repositoryID)
+	if err != nil {
+		return nil, err
+	}
+	_, branch, err := m.getBranchWithPredicate(ctx, repo, branchID)
```
itai-david, line 181: Nice reuse of the `repo` above. Love it :heart:

pkg/graveler/ref/tag_iterator.go
```diff
@@ -8,26 +8,26 @@ import (
...
-	it     kv.MessageIterator
-	err    error
-	value  *graveler.TagRecord
-	repoID graveler.RepositoryID
-	store  kv.Store
-	ctx    context.Context
+	it    kv.MessageIterator
+	err   error
+	value *graveler.TagRecord
+	repo  *graveler.RepositoryRecord
```
itai-david, line 14:
```suggestion
	partitionKey string
```
I don't see any use for the repository itself, other than getting the partition key

itai-david (changes requested):
Thank you @N-o-Z - this is definitely not a trivial one
Please see some comments and suggestions inside
My main concern is the multiple implicit calls to `GetRepository` from `KVGraveler` in `CreateBranch`, `DeleteBrnach` and `UpdateBranch`

Thanks

pkg/graveler/graveler.proto
```diff
@@ -15,6 +15,8 @@ message RepositoryData {
 message BranchData {
   string id = 1;
   string commit_id = 2;
+  string staging_token = 3;
+  repeated string sealed_tokens = 4;
```
itaiad200, line 19: I think the order of the `sealed_tokens` is worth a documentation. Is the first/last token the newest one?

N-o-Z: I was just thinking about it - I think that for the commit logic, the best way will be to have it sorted by oldest first. This way, for any new change we can simply override the current changes

pkg/graveler/graveler.go
```diff
@@ -862,15 +879,124 @@ func generateStagingToken(repositoryID RepositoryID, branchID BranchID) StagingT
...
+		// TODO(Guys) return error only on conflicts, currently returns error for any changes on staging
+		if err = g.checkEmptyToken(ctx, b.StagingToken); err != nil {
+			return nil, err
+		}
+		for _, st := range b.SealedTokens {
+			err = g.checkEmptyToken(ctx, st)
+			if err != nil {
+				return nil, err // First token with entries found, break
+			}
+		}
```
itaiad200, line 986: There's a bug here which is really easy to overlook, so let's add a test for it.
Imagine the following scenario: 
- `StagingToken:token1`
- `SealedTokens:[token2]`
- Staging area items:
	  - 	`token1: ["key1":tombstone]`, `token2: ["key1":"some-value"]`
	  - 	Imagine that the commit doesn't have `key1`

The update will fail, but staging area is actually empty.

pkg/graveler/ref/tag_iterator.go
N-o-Z: 👍🏽

pkg/graveler/ref/branch_iterator_test.go
N-o-Z: It would - though we will need to be very careful testing, the benchmark should include the creation + iteration flows since the sorting happens during the creation process.
Added comment in: https://github.com/treeverse/lakeFS/issues/3571

pkg/graveler/graveler.go
N-o-Z: Dereference does not call GetRepository but essentially you are right. Unfortunately this is temp constriction as we don't want to bloat the interface. When the DB implementation is removed we can change the interface so that we could save some Get calls

itai-david: Sounds legit
Do we have an issue to follow up on thins?

N-o-Z: Opened: https://github.com/treeverse/lakeFS/issues/3695

N-o-Z:
@itai-david @itaiad200  Please note, I added a bug fix to the BranchIterator sorted by commit implementation with SeekGE

pkg/graveler/graveler.go
itai-david: Thanks

itai-david (approved):
I'm good with the recent changes
Thank you @N-o-Z

pkg/catalog/catalog.go
```diff
@@ -457,15 +455,15 @@ func (c *Catalog) ListBranches(ctx context.Context, repository string, prefix st
...
-	if afterBranch < prefixBranch {
-		it.SeekGE(prefixBranch)
+	if after < prefix {
+		it.SeekGE(prefix)
 	} else {
-		it.SeekGE(afterBranch)
+		it.SeekGE(after)
 	}
 	var branches []*Branch
 	for it.Next() {
```
itaiad200, line 464: The iterator may continue past the predefined `prefix`, is that expected?

pkg/graveler/graveler.go
```diff
@@ -297,6 +299,8 @@ type CommitRecord struct {
 type Branch struct {
 	CommitID     CommitID
 	StagingToken StagingToken
+	// SealedTokens - Staging tokens are appended to the back so that iteration of tokens will be from oldest to newest
```
itaiad200, line 302: We're writing once to the sealed tokens from a single location and reading many times. It might be worth considering to make the one time operation more complicated than the many times operations.

pkg/graveler/graveler.go
```diff
@@ -1006,11 +1132,85 @@ func (g *KVGraveler) Log(ctx context.Context, repositoryID RepositoryID, commitI
...
+	for _, st := range tokens {
+		st := st // Pinning
+		wg.Go(func() error {
+			if err := g.StagingManager.Drop(ctx, st); err != nil && !errors.Is(err, ErrNotFound) {
+				return err
+			}
+			return nil
+		})
+	}
+	return wg.Wait().ErrorOrNil()
```
itaiad200, line 1154: A smoother way of dropping a branch workspace, would be to replace the staging token (or drop just that) and erase all its sealed tokens (while dropping in the background). `Drop` isn't an atomic operation, and even after we succeed we'll end up with empty sealed tokens which is meaningless.

itaiad200: I see it's only being used during `DeleteBranch`, since the `Branch` entry is on its way to be removed, my previous comment is of less significance.

pkg/graveler/graveler.go
```diff
@@ -1176,7 +1414,36 @@ func (g *KVGraveler) LoadCommits(ctx context.Context, repositoryID RepositoryID,
 }
 
 func (g *KVGraveler) LoadBranches(ctx context.Context, repositoryID RepositoryID, metaRangeID MetaRangeID) error {
-	return g.db.LoadBranches(ctx, repositoryID, metaRangeID)
+	repo, err := g.GetRepository(ctx, repositoryID)
+	if err != nil {
+		return err
+	}
+	// TODO (niro): Replace by KV committedManager
```
itaiad200, line 1421: 🤔 
Why does committed manager differ between KV and DB?

pkg/graveler/graveler.go
```diff
@@ -1196,7 +1463,28 @@ func (g *KVGraveler) DumpCommits(ctx context.Context, repositoryID RepositoryID)
...
+	iter, err := g.RefManager.ListBranches(ctx, repositoryID)
+	if err != nil {
+		return nil, err
+	}
+	defer iter.Close()
+	schema, err := serializeSchemaDefinition(&BranchData{})
+	if err != nil {
+		return nil, err
+	}
+	// TODO (niro): Replace by KV committedManager
```
itaiad200, line 1479: Same here

pkg/graveler/graveler.proto
```diff
@@ -58,7 +60,7 @@ message BranchProtectionRules {
   map<string, BranchProtectionBlockedActions> branch_pattern_to_blocked_actions = 1;
 }
 
-message StagedEntry {
+message StagedEntryData {
```
itaiad200, line 63: Why the renaming? It's more confusing IMO

pkg/graveler/ref/branch_iterator.go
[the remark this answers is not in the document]
itaiad200: The bigger question is: Should we address iterators error after we closed it?

pkg/graveler/ref/branch_iterator.go
```diff
@@ -2,143 +2,166 @@ package ref
...
 	}
+	return bi.itr.Err()
+}
+
+func (bi *BranchSimpleIterator) Close() {
+	bi.err = ErrIteratorClosed
+	bi.itr.Close()
+}
+
+type BranchByCommitIterator struct {
```
itaiad200, line 95: Missing documentation

pkg/graveler/ref/branch_iterator.go
```diff
@@ -2,143 +2,166 @@ package ref
...
-	offset          string
-	fetchSize       int
-	err             error
-	state           iteratorState
-	orderByCommitID bool
+// CompareFunc type used for sorting in InMemIterator, it is a strictly bigger comparison function required for the
+// sort.Slice algorithm, implementors need to decide how to handle equal values
+type CompareFunc func(i, j int) bool
+
+type BranchSimpleIterator struct {
```
itaiad200, line 15: Missing documenetation

pkg/graveler/ref/branch_iterator.go
```diff
@@ -2,143 +2,166 @@ package ref
...
-			WHERE repository_id = $1
-			AND id `+offsetCondition+` $2
-			ORDER BY `+ri.getOrderBy()+`  ASC
-			LIMIT $3`, ri.repositoryID, ri.offset, ri.fetchSize)
+func NewBranchByCommitIterator(ctx context.Context, store *kv.StoreMessage, repo *graveler.RepositoryRecord) (*BranchByCommitIterator, error) {
+	bi := &BranchByCommitIterator{
+		ctx:    ctx,
+		values: make([]*graveler.BranchRecord, 0),
+	}
+	itr, err := kv.NewPrimaryIterator(ctx, store.Store, (&graveler.BranchData{}).ProtoReflect().Type(), graveler.RepoPartition(repo), []byte(""), kv.IteratorOptionsFrom([]byte("")))
```
itaiad200, line 112: Are you trying to get all the branches? I think you're missing the branch prefix..
I believe it would be much simpler to use the `BranchSimpleIterator` here, it's basically what it does..

pkg/graveler/ref/branch_iterator.go
```diff
@@ -2,143 +2,166 @@ package ref
...
+		bi.values = append(bi.values, &graveler.BranchRecord{
+			BranchID: graveler.BranchID(value.Id),
+			Branch:   branchFromProto(value),
+		})
+	}
+	if itr.Err() != nil {
+		return nil, err
 	}
+
+	sort.Slice(bi.values, bi.SortByCommitID)
```
itaiad200, line 131: 🤩

pkg/graveler/ref/branch_iterator_test.go
```diff
@@ -94,3 +96,191 @@ func TestBranchIterator(t *testing.T) {
...
+			t.Fatalf("unexpected error: %v", iter.Err())
+		}
+		iter.Close()
+
+		if diffs := deep.Equal(ids, []graveler.BranchID{"a", "aa", "b", "c", "d", "e", "main"}); diffs != nil {
+			t.Fatalf("got wrong list of IDs: %v", diffs)
+		}
+	})
+
+	t.Run("listing branches using prefix", func(t *testing.T) {
```
itaiad200, line 137: Where's the prefix? looks identical to the next test

pkg/graveler/ref/branch_iterator_test.go
```diff
@@ -94,3 +96,191 @@ func TestBranchIterator(t *testing.T) {
...
+			t.Fatalf("unexpected error: %v", iter.Err())
+		}
+		iter.Close()
+
+		if diffs := deep.Equal(ids, []graveler.CommitID{"a", "aa", "b", "c", "d", "e", "main"}); diffs != nil {
+			t.Fatalf("got wrong list of IDs: %v", diffs)
+		}
+	})
+
+	t.Run("listing branches using prefix", func(t *testing.T) {
```
itaiad200, line 230: Same here

pkg/graveler/ref/manager.go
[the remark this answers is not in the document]
itaiad200: By order of importance I guess, Branch is more useful than the predicate (I always use that, the predicate is sometimes use)

pkg/graveler/ref/manager.go
```diff
@@ -123,35 +160,105 @@ func (m *KVManager) ResolveRawRef(ctx context.Context, repositoryID graveler.Rep
...
+	return pred, branchFromProto(&data), nil
+}
+
 func (m *KVManager) GetBranch(ctx context.Context, repositoryID graveler.RepositoryID, branchID graveler.BranchID) (*graveler.Branch, error) {
-	return m.db.GetBranch(ctx, repositoryID, branchID)
+	repo, err := m.getRepositoryRec(ctx, repositoryID)
+	if err != nil {
+		return nil, err
+	}
+	_, branch, err := m.getBranchWithPredicate(ctx, repo, branchID)
```
itaiad200, line 181: for example 😄

pkg/graveler/ref/manager.go
```diff
@@ -123,35 +160,105 @@ func (m *KVManager) ResolveRawRef(ctx context.Context, repositoryID graveler.Rep
...
+	}
+	pred, b, err := m.getBranchWithPredicate(ctx, repo, branchID)
+	if err != nil {
+		return err
+	}
+	newBranch, err := f(b)
+	if err != nil {
+		return err
+	}
+	return m.kvStore.SetMsgIf(ctx, graveler.RepoPartition(repo), []byte(graveler.BranchPath(branchID)), protoFromBranch(branchID, newBranch), pred)
```
itaiad200, line 218: Nice

pkg/graveler/ref/manager.go
```diff
@@ -123,35 +160,105 @@ func (m *KVManager) ResolveRawRef(ctx context.Context, repositoryID graveler.Rep
...
+	}
+	// Simulate same behavior as was in DB implementation
+	data := graveler.BranchData{}
+	_, err = m.kvStore.GetMsg(ctx, graveler.RepoPartition(repo), []byte(graveler.BranchPath(branchID)), &data)
+	if err != nil {
+		if errors.Is(err, kv.ErrNotFound) {
+			err = graveler.ErrBranchNotFound
+		}
+		return err
+	}
```
itaiad200, line 234: `m.GetBranch`

pkg/graveler/retention/garbage_collection_manager.go
```diff
@@ -145,7 +145,10 @@ func (m *GarbageCollectionManager) SaveGarbageCollectionCommits(ctx context.Cont
 		refManager:   m.refManager,
 		repositoryID: repositoryID,
 	}
-	branchIterator := ref.NewBranchIterator(ctx, m.db, repositoryID, 1000, ref.WithOrderByCommitID()) //nolint: gomnd
+	branchIterator, err := m.refManager.GCBranchIterator(ctx, repositoryID) //nolint: gomnd
```
itaiad200, line 148: Do we still need to `nolint`?

pkg/graveler/testutil/fakes.go
```diff
@@ -707,10 +711,10 @@ func (m *FakeBranchIterator) Next() bool {
 	return m.Index < len(m.Data)
 }
 
-func (m *FakeBranchIterator) SeekGE(id graveler.BranchID) {
+func (m *FakeBranchIterator) SeekGE(id string) {
 	m.Index = len(m.Data)
 	for i, item := range m.Data {
-		if item.BranchID >= id {
+		if item.BranchID.String() >= id {
```
itaiad200, line 717: One more reason to keep it `BranchID`

pkg/graveler/ref/manager_test.go
```diff
@@ -189,149 +192,225 @@ func TestManager_GetBranch(t *testing.T) {
...
-			t.Fatalf("unexpected commit for branch2: %s - expected: c3", b.CommitID)
-		}
+			if b.CommitID != "c3" {
+				t.Fatalf("unexpected commit for branch2: %s - expected: c3", b.CommitID)
+			}
+		})
+	}
+}
+
+func TestManager_BranchUpdate(t *testing.T) {
```
itaiad200, line 280: 💪

itaiad200 (changes requested):

pkg/graveler/graveler.go
```diff
@@ -862,15 +879,124 @@ func generateStagingToken(repositoryID RepositoryID, branchID BranchID) StagingT
...
+
+// checkEmptyToken Checks whether token contains etries. Returns error if List operation fails or token contains entries, False otherwise
+func (g *KVGraveler) checkEmptyToken(ctx context.Context, stagingToken StagingToken) error {
+	iter, err := g.StagingManager.List(ctx, stagingToken, 1)
+	if err != nil {
+		return err
+	}
+	defer iter.Close()
+	if iter.Next() {
+		return ErrConflictFound
```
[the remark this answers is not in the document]
N-o-Z: Will handle on the staging area PR

pkg/graveler/graveler.go
N-o-Z: It's just a TODO for me to validate the current flow as all parts come togehter

pkg/graveler/graveler.go
N-o-Z: It's just a TODO for me to validate the current flow as all parts come together

pkg/graveler/ref/manager.go
N-o-Z: I didn't understand the comment

N-o-Z: Now I get it :)

PR #3672, opened 2022-07-17
Fix lint error and testing migrate test failures

nopcoder:


pkg/db/kv_migration_test.go
```diff
@@ -17,7 +17,9 @@ func TestKVMigration(t *testing.T) {
 	ctx := context.Background()
 	dbParams := params.Database{Driver: "pgx", ConnectionString: databaseURI, KVEnabled: true, DropTables: true}
 	kvStore, err := kv.Open(ctx, kvpg.DriverName, kvparams.KV{Postgres: &kvparams.Postgres{ConnectionString: dbParams.ConnectionString}})
+	require.NoError(t, err)
```
itaiad200, line 20: @nopcoder uses `require`. Now I've seen everything 😄

pkg/config/struct_keys_test.go
```diff
@@ -60,7 +60,7 @@ func TestStructKeys_SimpleTagged(t *testing.T) {
 	type s struct {
 		A int `test:"Aaa"`
 		B int `toast:"bee"`
-		c int `test:"ccc" toast:"sea"`
```
itaiad200, line 63: I think this is intentional.

pkg/config/struct_keys_test.go
```diff
@@ -234,7 +234,7 @@ func TestValidateMissingRequired_SimpleTagged(t *testing.T) {
 	type s struct {
 		A int `test:"Aaa" validate:"required"`
 		B int `toast:"bee" validate:"required"`
-		c int `test:"ccc" toast:"sea" validate:"required"`
+		C int `test:"ccc" toast:"sea" validate:"required"`
```
itaiad200, line 237: Here too.

esti/import_test.go
```diff
@@ -46,31 +29,30 @@ func TestImport(t *testing.T) {
 	case block.BlockstoreTypeAzure:
 		importPath = azureImportPath
 	default:
-		// import isn't supported for non-production block adapters
-		t.Skip()
+		t.Skip("import isn't supported for non-production block adapters")
```
itaiad200, line 32: 👍

itaiad200 (approved):

pkg/db/kv_migration_test.go
nopcoder: ha :D used the same code the test implementor used - still prefer the simple code.

pkg/config/struct_keys_test.go
nopcoder: Do you think it was to see if we can extract the metadata from the struct for private fields - because usually the linter things we are not using the field.

nopcoder: @arielshaqed this is by design? if so I'll revert and leave a comment

PR #3679, opened 2022-07-18
Break combined iterator to accept many iterator and return tombstones

itaiad200:


pkg/graveler/combined_iterator.go
```diff
@@ -67,35 +77,26 @@ func (c *CombinedIterator) advanceInnerIterators() bool {
...
+	valA := c.iterA.Value()
+	valB := c.iterB.Value()
+	switch {
+	case valA == nil && valB == nil:
+		c.p = c.iterA
+		return false
+	case valA == nil:
+		c.p = c.iterB
+	case valB == nil:
+		c.p = c.iterA
```
nopcoder, line 93: We need to skip the nil and still process the iterators based on key ordering. So in case one of the iterators points to a nil value we need to skip until we get to a value or marked as ended by setting it to nil.

nopcoder: ^ can ignore I thought it was the value that hold nil for tomstone

pkg/graveler/combined_iterator.go
```diff
@@ -122,3 +123,37 @@ func (c *CombinedIterator) Close() {
 	c.iterA.Close()
 	c.iterB.Close()
 }
+
+// FilterTombstoneIterator wraps a value iterator and filters out tombstones.
+type FilterTombstoneIterator struct {
+	iter ValueIterator
+}
+
```
nopcoder, line 131: Consider just retype ValueIterator - it will have the same implementation, just without the 'iter' as a member

nopcoder (approved):
lgtm!

pkg/graveler/combined_iterator.go
itaiad200: synced f2f - cannot do :)

pkg/graveler/combined_iterator.go
```diff
@@ -2,8 +2,6 @@ package graveler
 
 import (
 	"bytes"
-
-	"github.com/treeverse/lakefs/pkg/logging"
 )
 
 // CombinedIterator iterates over two listing iterators,
```
N-o-Z, line 7: I think documentation is crucial here. Moreover it still says "CombinedIterator iterates over two listing iterators", while this is "technically" true, it is still confusing

itaiad200: Solved in #3688

PR #3680, opened 2022-07-18
fixed diff-viewer version by using a known fork

ozkatz:


nopcoder (approved):
Great, this package didn't update and doesn't look like it will soon :/

PR #3697, opened 2022-07-20
Don't double-encode query params in WebUI templates client

### LINKED ISSUE #3681: TTV: Templating service returns un-encoded URL
When sending to the templating service a URL as a parameter to be returned in a template, like `http://localhost:8000`, the templating service returns it in a wrong encoding: `http%3A%2F%2Flocalhost%3A8000`.

arielshaqed:
Tested using the diff'ed client code from #3647. 

Fixes #3681.

### DIFF
--- a/webui/src/lib/api/index.js
+++ b/webui/src/lib/api/index.js
@@ -891,7 +891,7 @@ class Templates {
             urlParams.set(k, v);
         }
         const response = await apiRequest(
-            encodeURI(`/templates/${templateLocation}?${urlParams.toString()}`),
+            `/templates/${encodeURI(templateLocation)}?${urlParams.toString()}`,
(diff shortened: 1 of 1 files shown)


webui/src/lib/api/index.js
```diff
@@ -891,7 +891,7 @@ class Templates {
             urlParams.set(k, v);
         }
         const response = await apiRequest(
-            encodeURI(`/templates/${templateLocation}?${urlParams.toString()}`),
+            `/templates/${encodeURI(templateLocation)}?${urlParams.toString()}`,
```
Jonathan-Rosenberg, line 894: nice

Jonathan-Rosenberg (approved):

PR #3702, opened 2022-07-20
API: return conflict status code on already resource already exists

nopcoder:
API: return status code 409 (Conflict) when relevant


nopcoder:
Close - updated by pull-request #4350

PR #3703, opened 2022-07-20
Remove ambiguity around username

johnnyaug:
1. Add description to id field in API.
2. Add description to user ID field in lakectl.
3. Add "username" field to Auth API and deprecate "name".
4. (Bonus) reorder flags on lakectl auth commands so that pagination flags are at the bottom.

### DIFF
--- a/docs/reference/commands.md
+++ b/docs/reference/commands.md
@@ -422,7 +422,7 @@ lakectl auth groups create [flags]
 
 ```
   -h, --help        help for create
-      --id string   group identifier
+      --id string   Group identifier
 ```
 
 
@@ -440,7 +440,7 @@ lakectl auth groups delete [flags]
 
 ```
   -h, --help        help for delete
-      --id string   group identifier
+      --id string   Group identifier
 ```
 
 
@@ -513,8 +513,8 @@ lakectl auth groups members add [flags]
 
 ```
   -h, --help          help for add
-      --id string     group identifier
-      --user string   Username (email for password-based users). Defaults to current user.
+      --id string     Group identifier
+      --user string   Username (email for password-based users, default: current user)
 ```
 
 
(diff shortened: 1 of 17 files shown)


docs/reference/commands.md
```diff
@@ -865,7 +865,7 @@ lakectl auth users credentials create [flags]
 
 ```
   -h, --help        help for create
-      --id string   user identifier (default: current user)
+      --id string   Username (email for password-based users). Defaults to current user.
```
nopcoder, line 868: There are 3 places in this PR that we started adding '.' at the end of the help line of the argument. we need to align.

pkg/auth/service.go
```diff
@@ -1273,9 +1281,13 @@ func (a *APIAuthService) ListUsers(ctx context.Context, params *model.Pagination
 	results := resp.JSON200.Results
 	users := make([]*model.User, len(results))
 	for i, r := range results {
+		username := r.Username
+		if username == "" {
+			username = swag.StringValue(r.Name)
```
nopcoder, line 1286: We would like to return the Name as Username even if it is empty, as we will make it required. Unless we plan on the next PR remove this requirement.

pkg/auth/service.go
```diff
@@ -1439,9 +1451,13 @@ func (a *APIAuthService) ListGroupUsers(ctx context.Context, groupDisplayName st
 	members := make([]*model.User, len(resp.JSON200.Results))
 
 	for i, r := range resp.JSON200.Results {
+		username := r.Username
+		if username == "" {
+			username = swag.StringValue(r.Name)
```
nopcoder, line 1456: We would like to return the Name as Username even if it is empty, as we will make it required

nopcoder (changes requested):

pkg/auth/service.go
johnnyaug: The username is in the Auth service API definition - here is the lakeFS OpenAPI definition - no Name field.

nopcoder (approved):

PR #3732, opened 2022-07-23
Implement KV graveler Set & Delete

### LINKED ISSUE #3629: Graveler KV Writer Flow
Follow the writer flow described in the [kv proposal](https://github.com/treeverse/lakeFS/blob/master/design/open/metadata_kv/index.md) once the staging and branches are ported to the kv store.

itaiad200:
closes #3629

### DIFF
--- a/pkg/graveler/graveler.go
+++ b/pkg/graveler/graveler.go
@@ -1294,7 +1294,77 @@ func (g *KVGraveler) Get(ctx context.Context, repositoryID RepositoryID, ref Ref
 }
 
 func (g *KVGraveler) Set(ctx context.Context, repositoryID RepositoryID, branchID BranchID, key Key, value Value, writeConditions ...WriteConditionOption) error {
-	return g.db.Set(ctx, repositoryID, branchID, key, value, writeConditions...)
+	isProtected, err := g.protectedBranchesManager.IsBlocked(ctx, repositoryID, branchID, BranchProtectionBlockedAction_STAGING_WRITE)
+	if err != nil {
+		return err
+	}
+	if isProtected {
+		return ErrWriteToProtectedBranch
+	}
+
+	setFunc := func(branch *Branch) error {
+		writeCondition := &WriteCondition{}
+		for _, cond := range writeConditions {
+			cond(writeCondition)
+		}
+
+		if writeCondition.IfAbsent {
+			// check if the given key exist in the branch first
+			_, err := g.Get(ctx, repositoryID, Ref(branchID), key)
+			if err == nil {
+				// we got a key here already!
+				return ErrPreconditionFailed
+			}
+			if !errors.Is(err, ErrNotFound) {
+				// another error occurred!
+				return err
+			}
+		}
+		if err = g.StagingManager.Set(ctx, branch.StagingToken, key, &value, !writeCondition.IfAbsent); err != nil {
+			return err
+		}
+		return nil
+	}
+
+	return g.safeBranchWrite(ctx, g.log.WithField("key", key).WithField("operation", "set"), repositoryID, branchID, setFunc)
+}
+
+// safeBranchWrite is a helper function that wraps a branch write operation with validation that the staging token
+// didn't change while writing to the branch.
+func (g *KVGraveler) safeBranchWrite(ctx context.Context, log logging.Logger, repositoryID RepositoryID, branchID BranchID, branchOperation func(branch *Branch) error) error {
+	// setTries is the number of times to repeat the set operation if the staging token changed
+	const setTries = 3
+
+	var try int
+	for try = 0; try < setTries; try++ {
+		branch, err := g.GetBranch(ctx, repositoryID, branchID)
+		if err != nil {
+			return err
+		}
+		startToken := branch.StagingToken
+
+		err = branchOperation(branch)
+		// Checking if the token has changed.
+		// If it changed, we need to write the changes to the branch's new staging token
+		branch, err = g.GetBranch(ctx, repositoryID, branchID)
+		if err != nil {
+			return err
+		}
+		endToken := branch.StagingToken
+		if startToken == endToken {
+			break
+		} else {
+			// we got a new token, try again
+			log.WithField("try", try+1).
+				WithField("startToken", startToken).
+				WithField("endToken", endToken).
+				Info("Retrying Set")
+		}
+	}
+	if try == setTries {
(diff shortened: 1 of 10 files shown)


pkg/graveler/graveler.go
```diff
@@ -1294,11 +1296,141 @@ func (g *KVGraveler) Get(ctx context.Context, repositoryID RepositoryID, ref Ref
...
+	}
+
+	setFunc := func(branch *Branch) error {
+		writeCondition := &WriteCondition{}
+		for _, cond := range writeConditions {
+			cond(writeCondition)
+		}
+
+		if !writeCondition.IfAbsent {
+			return g.StagingManager.Set(ctx, branch.StagingToken, key, &value, !writeCondition.IfAbsent)
```
N-o-Z, line 1314: What does it matter which overwrite value are we passing? It's no longer used

pkg/graveler/staging/manager.go
```diff
@@ -37,25 +37,36 @@ func (m *Manager) Get(ctx context.Context, st graveler.StagingToken, key gravele
...
 		}
+	} else {
+		oldValue = graveler.StagedEntryFromProto(oldValueProto)
+	}
+	updatedValue, err := updateFunc(oldValue)
+	if err != nil {
+		return err
 	}
-	return err
+	return m.store.SetMsgIf(ctx, string(st), key, graveler.ProtoFromStagedEntry(key, updatedValue), pred)
```
N-o-Z, line 69:
```suggestion
	return m.store.SetMsgIf(ctx, st.String(), key, graveler.ProtoFromStagedEntry(key, updatedValue), pred)
```

pkg/graveler/graveler.go
N-o-Z: I think this is hiding a bug

pkg/graveler/graveler.go
```diff
@@ -1294,11 +1296,141 @@ func (g *KVGraveler) Get(ctx context.Context, repositoryID RepositoryID, ref Ref
...
+			// we got a key here already!
+			return ErrPreconditionFailed
+		}
+		if !errors.Is(err, ErrNotFound) {
+			// another error occurred!
+			return err
+		}
+
+		return g.StagingManager.Update(ctx, branch.StagingToken, key, func(v *Value) (*Value, error) {
+			if v == nil || v.Identity == nil {
```
N-o-Z, line 1329: Not really sure why we need the updateFunc, seems like the staging manager Update is being used only in Set

pkg/graveler/graveler.go
```diff
@@ -1294,11 +1296,141 @@ func (g *KVGraveler) Get(ctx context.Context, repositoryID RepositoryID, ref Ref
...
+	}
+
+	return g.safeBranchWrite(ctx, g.log.WithField("key", key).WithField("operation", "set"), repositoryID, branchID, setFunc)
+}
+
+// safeBranchWrite is a helper function that wraps a branch write operation with validation that the staging token
+// didn't change while writing to the branch.
+func (g *KVGraveler) safeBranchWrite(ctx context.Context, log logging.Logger, repositoryID RepositoryID, branchID BranchID, stagingOperation func(branch *Branch) error) error {
+	// setTries is the number of times to repeat the set operation if the staging token changed
+	const setTries = 3
```
N-o-Z, line 1344: Maybe we should export the retry mechanism?

pkg/graveler/graveler.go
```diff
@@ -1308,6 +1440,7 @@ func (g *KVGraveler) listStagingArea(ctx context.Context, b *Branch) (ValueItera
 		return nil, ErrNotFound
 	}
 	it, err := g.StagingManager.List(ctx, b.StagingToken, 1)
+
 	if err != nil {
```
N-o-Z, line 1444:
```suggestion
	it, err := g.StagingManager.List(ctx, b.StagingToken, 1)
	if err != nil {
```

pkg/graveler/graveler.go
N-o-Z: Even though this is an overwrite flow it still should be under the safeBranchWrite flow

N-o-Z (changes requested):
Some questions, some comments. Also I would love to see some tests on Set and Update flows - as it's currently not tested at all

pkg/graveler/graveler.go
itaiad200: a. Right - it's no longer needed. However since I must pass something - I rather pass the right value..
b. It's under the `setFunc` hence under `safeBranchWrite`

pkg/graveler/graveler.go
itaiad200: Would you rather I'll refactor the `Set` func? I don't want to force other users to pass a function if they don't need to.

pkg/graveler/graveler.go
itaiad200: The only other usage I see for it is the `Commit` and it can use it from here too, no?

pkg/graveler/staging/manager.go
itaiad200: Used a partition func like everywhere else.

pkg/graveler/graveler.go
```diff
@@ -1294,11 +1296,141 @@ func (g *KVGraveler) Get(ctx context.Context, repositoryID RepositoryID, ref Ref
...
+		return err
+	}
+
+	deleteEntry := func(branch *Branch) error {
+		commit, err := g.RefManager.GetCommit(ctx, repositoryID, branch.CommitID)
+		if err != nil {
+			return err
+		}
+
+		// check key in committed - do we need tombstone?
```
N-o-Z, line 1400: I think we can remove this comment

pkg/graveler/staging/manager_test.go
```diff
@@ -63,9 +63,52 @@ func TestStagingManager(t *testing.T) {
...
+				testUpdate(t)
+			})
+		}
 	}
 }
 
+func testUpdate(t *testing.T) {
+	ctx, s := newTestStagingManager(t, true)
+
+	key := "a/b/c/my-key-1234"
```
N-o-Z, line 77: image

N-o-Z (approved):
@itaiad200 Looks sharp!
I'll add the graveler Set tests to the testing gaps

itaiad200:
> 

Added tests, if (when) checks pass I'll merge

PR #3733, opened 2022-07-23
Make GC hard-delete expired objects from Azure blob storage

### LINKED ISSUE #3546: Azure GC: Make GC hard-delete expired objects
- Discovery: how to delete expired objects (Azure fs Vs. Azure client)
- Discovery: which auth hadoop configurations are needed for 1. personal account token 2. service principle 
- Implement objects deletion. 

 
### DOD 

- GC hard-deletes expired objects
- GC job is working e2e on Azure

talSofer:
Closes #3546 

### In this review
* Use Azure SDK to bulk delete expired objects from Azure blob. 
* Add sbt merge strategy setting that discards changes done to the assembled META-INF directory (based on https://stackoverflow.com/questions/46287789/running-an-uber-jar-from-sbt-assembly-results-in-error-could-not-find-or-load-m)

### Not in this review
Currently, objects are deleted, but the status of the deleteBlobs operation is not captured, resulting in a misleading GC end-of-run report. This will be handled separately in https://github.com/treeverse/lakeFS/issues/3734

### How was this tested 
Tested Manually with lakeFS connected to Azure, watching expired objects getting removed. 

### Note to reviewers 
* Its recommended to review by commits. 
* ~The current version of the code does not compile with Spark 2.4.7 and Scala 2.11 due to https://stackoverflow.com/a/43664779/11553804~

### DIFF
--- a/clients/spark/build.sbt
+++ b/clients/spark/build.sbt
@@ -117,7 +117,23 @@ lazy val examples312 = generateExamplesProject(spark312Type).dependsOn(core312)
 lazy val root = (project in file(".")).aggregate(core2, core3, core312, examples2, examples3, examples312)
 
 lazy val assemblySettings = Seq(
-  assembly / assemblyMergeStrategy := (_ => MergeStrategy.first),
+  assembly / assemblyMergeStrategy := {
+    case PathList("META-INF", xs @ _*) =>
+      (xs map {_.toLowerCase}) match {
+        case ("manifest.mf" :: Nil) | ("index.list" :: Nil) | ("dependencies" :: Nil) =>
+          MergeStrategy.discard
+        case ps @ (x :: xs) if ps.last.endsWith(".sf") || ps.last.endsWith(".dsa") =>
+          MergeStrategy.discard
+        case "plexus" :: xs =>
+          MergeStrategy.discard
+        case "services" :: xs =>
+          MergeStrategy.filterDistinctLines
+        case ("spring.schemas" :: Nil) | ("spring.handlers" :: Nil) =>
+          MergeStrategy.filterDistinctLines
+        case _ => MergeStrategy.first
+      }
+    case _ => MergeStrategy.first
+  },
   assembly / assemblyShadeRules := Seq(
     ShadeRule.rename("org.apache.http.**" -> "org.apache.httpShaded@1").inAll,
     ShadeRule.rename("com.google.protobuf.**" -> "shadeproto.@1").inAll,
--- a/clients/spark/build.sbt
+++ b/clients/spark/build.sbt
@@ -119,6 +119,8 @@ lazy val examples312 = generateExamplesProject(spark312Type).dependsOn(core312)
 
 lazy val root = (project in file(".")).aggregate(core2, core3, core312, examples2, examples3, examples312)
 
+// We are using the default sbt assembly merge strategy https://github.com/sbt/sbt-assembly#merge-strategy with a change
+// to the general case: use MergeStrategy.first instead of MergeStrategy.deduplicate.
 lazy val assemblySettings = Seq(
   assembly / assemblyMergeStrategy := {
     case PathList("META-INF", xs @ _*) =>
--- a/clients/spark/build.sbt
+++ b/clients/spark/build.sbt
@@ -60,6 +60,9 @@ def generateCoreProject(buildType: BuildType) =
         // version, but ask to use whatever is provided so we do not
         // override what it selects.
         "com.amazonaws" % "aws-java-sdk-bundle" % "1.12.194" % "provided",
+        "com.azure" % "azure-core" % "1.10.0",
+        "com.azure" % "azure-storage-blob" % "12.9.0",
+        "com.azure" % "azure-storage-blob-batch" % "12.7.0",
         // Snappy is JNI :-(.  However it does claim to work with
         // ClassLoaders, and (even more importantly!) using a preloaded JNI
         // version will probably continue to work because the C language API
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
@@ -3,7 +3,10 @@ package io.treeverse.clients
 import com.google.common.cache.CacheBuilder
 import io.lakefs.clients.api
 import io.lakefs.clients.api.{ConfigApi, RetentionApi}
-import io.lakefs.clients.api.model.{GarbageCollectionPrepareRequest, GarbageCollectionPrepareResponse}
+import io.lakefs.clients.api.model.{
+  GarbageCollectionPrepareRequest,
+  GarbageCollectionPrepareResponse
+}
 import io.treeverse.clients.StorageClientType.StorageClientType
 import io.treeverse.clients.StorageUtils.{StorageTypeAzure, StorageTypeS3}
 
@@ -82,7 +85,7 @@ class ApiClient(apiUrl: String, accessKey: String, secretKey: String) {
               .normalize()
               .toString
           case StorageClientType.SDKClient => repo.getStorageNamespace
-          case _                              => throw new IllegalArgumentException
+          case _                           => throw new IllegalArgumentException
         }
         storageNamespace
       })
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
@@ -2,24 +2,19 @@ package io.treeverse.clients
 
 import com.google.common.cache.CacheBuilder
 import io.lakefs.clients.api
-import io.lakefs.clients.api.RetentionApi
-import io.lakefs.clients.api.ConfigApi
-import io.lakefs.clients.api.model.{
-  GarbageCollectionPrepareRequest,
-  GarbageCollectionPrepareResponse
-}
-import io.treeverse.clients.StorageAccessType.StorageAccessType
+import io.lakefs.clients.api.{ConfigApi, RetentionApi}
+import io.lakefs.clients.api.model.{GarbageCollectionPrepareRequest, GarbageCollectionPrepareResponse}
+import io.treeverse.clients.StorageClientType.StorageClientType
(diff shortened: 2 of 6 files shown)


clients/spark/build.sbt
```diff
@@ -60,6 +60,9 @@ def generateCoreProject(buildType: BuildType) =
         // version, but ask to use whatever is provided so we do not
         // override what it selects.
         "com.amazonaws" % "aws-java-sdk-bundle" % "1.12.194" % "provided",
+        "com.azure" % "azure-core" % "1.10.0",
+        "com.azure" % "azure-storage-blob" % "12.9.0",
+        "com.azure" % "azure-storage-blob-batch" % "12.7.0",
```
arielshaqed, line 65: Assuming these will be "provided" as part of your work on fixing dependencies :-).  (I reckon if they are "provided", we can just `--package` them into `spark-submit` in the worst case.  I hope...)

clients/spark/build.sbt
```diff
@@ -117,7 +120,23 @@ lazy val examples312 = generateExamplesProject(spark312Type).dependsOn(core312)
...
+        case "plexus" :: xs =>
+          MergeStrategy.discard
+        case "services" :: xs =>
+          MergeStrategy.filterDistinctLines
+        case ("spring.schemas" :: Nil) | ("spring.handlers" :: Nil) =>
+          MergeStrategy.filterDistinctLines
+        case _ => MergeStrategy.first
+      }
+    case _ => MergeStrategy.first
+  },
```
arielshaqed, line 139: Can you add documentation for what this does, or where to find its documentation?  I _think_ that it deletes package signing.  If so, are we sure it's OK?

Also, if we say "provided" above do we still need this?

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -4,23 +4,27 @@ import com.google.common.cache.CacheBuilder
 import io.lakefs.clients.api
 import io.lakefs.clients.api.RetentionApi
 import io.lakefs.clients.api.ConfigApi
-import io.lakefs.clients.api.model.{
-  GarbageCollectionPrepareRequest,
-  GarbageCollectionPrepareResponse
-}
+import io.lakefs.clients.api.model.{GarbageCollectionPrepareRequest, GarbageCollectionPrepareResponse}
+import io.treeverse.clients.StorageAccessType.StorageAccessType
+import io.treeverse.clients.StorageUtils.{StorageTypeAzure, StorageTypeS3}
```
arielshaqed, line 9: I don't understand these lines: I thought we were in `package io.treeverse.clients`?!  (Perhaps "organize imports" in your IDE can fix this...)

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -4,23 +4,27 @@ import com.google.common.cache.CacheBuilder
...
 
 import java.net.URI
 import java.util.concurrent.TimeUnit
 import java.util.concurrent.Callable
 
+// The different types of storage access the metadata client uses.
+object StorageAccessType extends Enumeration {
+  type StorageAccessType = Value
+
+  val DirectAccess, HadoopFS = Value
```
arielshaqed, line 19: Do these get a nice string representation?

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -67,16 +71,22 @@ class ApiClient(apiUrl: String, accessKey: String, secretKey: String) {
...
-          .translateURI(URI.create(repo.getStorageNamespace), getBlockstoreType())
-          .normalize()
-          .toString
+        val storageNamespace =  accessType match {
+          case StorageAccessType.HadoopFS  =>
+            ApiClient
+              .translateURI(URI.create(repo.getStorageNamespace), getBlockstoreType())
+              .normalize()
+              .toString
+          case StorageAccessType.DirectAccess  => repo.getStorageNamespace
```
arielshaqed, line 86: I am not sure what the difference is between these two "access types".  Also, "direct access" in `lakectl` is to do essentially what lakeFSFS does, so this name is confusing.

clients/spark/core/src/main/scala/io/treeverse/clients/StorageUtils.scala
```diff
@@ -0,0 +1,37 @@
+package io.treeverse.clients
+
+import java.net.URI
+
+object StorageUtils {
+  val StorageTypeS3 = "s3"
+  val StorageTypeAzure = "azure"
+  val SchemeDelimiter = "://"
```
arielshaqed, line 8: I think this last constant is a bit ugly.  It will never change: we're using URLs (and if we ever stopped using URLs a lot more would change).  We could use a URI object or some (non-`java.net`) URL object to construct it, or just inline `://` to URL construction.

clients/spark/core/src/main/scala/io/treeverse/clients/StorageUtils.scala
```diff
@@ -0,0 +1,37 @@
...
+
+  object AzureBlob {
+    val StorageAccountKeyPropertyPattern = "fs.azure.account.key.<storageAccountName>.dfs.core.windows.net"
+    val StorageAccNamePlaceHolder = "<storageAccountName>"
+    // https://docs.microsoft.com/en-us/dotnet/api/overview/azure/storage.blobs.batch-readme#key-concepts
+    // Note that there is no official java SDK documentation of the max batch size, therefore assuming the above.
```
arielshaqed, line 14: Excellent documentation!  Thanks...

clients/spark/core/src/main/scala/io/treeverse/clients/StorageUtils.scala
```diff
@@ -0,0 +1,37 @@
...
+
+    def uriToStorageAccountName(storageNsURI: URI): String ={
+      storageNsURI.getHost.split('.')(0)
+    }
+  }
+
+  object S3 {
+    val S3MaxBulkSize = 1000
+    val S3NumRetries = 1000
+  }
```
arielshaqed, line 35: This should be next to `AzureBlob`.

arielshaqed (commented):
Neat!

Not approving until it moves out of draft status.

clients/spark/build.sbt
Jonathan-Rosenberg: This is the default merge strategy of the sbt-assembly plugin with our changes. You can read about it [here](https://github.com/sbt/sbt-assembly#merge-strategy).
Provided doesn't solve the issue...

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
talSofer: The idea is to distinguish between the following cases:
1. Needed a storage namespace in a format that some hadoop file system can work with (e.g. should start with s3a for s3aFileSystem, or be of the form "abfs://${container}@${storageAccountName}.dfs.core.windows.net/${blobPath}" for AzureBlobFileSystem - I call this HadoopFS access type
2. Needed a storage namespace in a format used by SDK clients to access the storage directly - I call this direct access type. 

I'm very open to other naming suggestions :) 

Maybe I can use something related to the type of client that uses it? such as StorageClientType with HadoopFS, SDKclient? WDYT?

clients/spark/core/src/main/scala/io/treeverse/clients/StorageUtils.scala
talSofer: Not sure that I understand what you mean, can you please clarify? It is in the same hierarchy now too right?

clients/spark/build.sbt
talSofer: How can we rely on marking those as provided? I don't expect that every user who runs Spark on Azure will have those packages available.

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
arielshaqed: ?

clients/spark/core/src/main/scala/io/treeverse/clients/StorageUtils.scala
arielshaqed: Sorry, thought indentation was different.

arielshaqed (approved):
This is very nice!  Please note that most comments from the previous round remain unaddressed; you should be able to resolve them all as inlined comments or by adding docs to the code.

talSofer:
> Please note that most comments from the previous round remain unaddressed;

@arielshaqed sure, I'm planning to address everything :) I wanted to get your answers to the open questions I have before doing that. we re-request a review after making the changes of course.

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
talSofer: @arielshaqed WDYT about this?

clients/spark/core/src/main/scala/io/treeverse/clients/StorageUtils.scala
talSofer: Thanks, I Inlined :// to URL construction

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
talSofer: Changed this to StorageClientType: SDKClient, HadoopFS

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
talSofer: optimize imports didn't do the trick :(

See answer number 2 here https://stackoverflow.com/a/67364323/11553804
> Just like every identifiers in scala, you need to import to access identifiers defined in another package or object

clients/spark/build.sbt
talSofer: Added the link to docs. 

> I think that it deletes package signing.

My understanding is that it preserves the signing we define in case of a conflict.

Issue #3713, opened 2022-07-21
Repository Creation and Deletion - Step 1 - Repository State and Unique Identifier


itai-david:
Step 1 as described in #3661

N-o-Z:
As part of this change, I want to change the behavior of deleting a non existing repository.
The current behavior is to return an error and I believe we should change it to return success.
I think this is the correct behavior for most flows that require deleting an object in the system, this should also enable us to implement the suggested deletion flow. @itai-david @itaiad200 @eden-ohana  @nopcoder  your thoughts?

N-o-Z:
New view when repository is under deletion:
![image](https://user-images.githubusercontent.com/3787958/182826704-29586bbb-96f4-485d-be78-490e2c3dd111.png)

PR #3735, opened 2022-07-24
3713 repository state and unique identifier

itai-david:
Introducing `RepositoryState` - A repository is either `active` - can be used and accessed normally, or it is `deleting` which means it is not accessible for any use, except for a delete retry
Repository creation was changed to create the initial commit and default branch prior to the repository itself, guaranteeing that once the reposiory is created it is created completely. There is a chance, of course, that due to a failure the commit or branch are left hanging, unreachable, in the DB
Repository deletion was modified to set the `RepositoryState` to `deleting`, making the repository inaccessible and, in fact, deleted. This state update is followed by a deletion of all repository`s entities and eventually the repository itself. It is possible that this will be a long operation and may involve some retries. For the entire operation, the `RepositoryID` cannot be reused, although the repository "does not exist"
Get/list respositories was modified to only show `active` repositories

In order to achieve the above, this PR includes some refactoring to the ref/manager:
Some new repository (private) get functions introduced, to support getting a `deleted` repository, without breaking the `GetRepository` interface that only returns an active repository. Added some functionality to return a Repository or a RepositoryRecord, with or without predicate, to supprt different needs by different functions
Branch and Commit creation logic was moved to private createBranch and addCommit to break the dependency of an existing repository and support the precreation of these entities as part of a respoitory creation. The interface functions `CreateBranch` and `AddCommit` use these private functions


itai-david:
Closes #3713

itai-david:
Covered by #3841

PR #3741, opened 2022-07-24
Proposal: pull request

nopcoder:
Initial mechanism to have manuall approved pull-request - introduce changes into a branch.


N-o-Z (commented):
Very exciting!
Do we need explicit authorization for PR approval?
Should we allow merge (without force) when no reviewers are assigned?
PR update == commit / revert / rebase / etc. It means we have to add a hook trigger on all flows? Will the PR merge trigger a secondary hook call (pre/post merge) in addition to the PR hook?

design/open/pull-request.md
```diff
@@ -0,0 +1,28 @@
...
+- Automation using actions can verify / provide feedback on pull-request
+
+### How it will work
+
+Introduce a new resource called pull-request at the repository level.
+The pull-request will hold owner (the user who created the PR), assignee, reviewers the source branch/reference and the target branch.
+Each reviewer can set approved/reject (or clear his response) on the PR.
+PR can be merge operation is enabled when at least one reviewer approves or when no reviewers are assigned.
+PR can be closed at any time.
+PR will not be updated when new commits into the source branch.
```
itaiad200, line 19:
```suggestion
PR will not be updated with changes to the source or destination branches, like updates to objects, commits, merges, etc. 
```

design/open/pull-request.md
```diff
@@ -0,0 +1,28 @@
...
+Merge PR will perform a merge operation and merge the PR as merged. If a conflict or an error occurs while merging, it will be returned as a response to the request.
```
itaiad200, line 20: That's tricky. I might be reviewing a PR that has no chances to be merged since there are conflicts.

design/open/pull-request.md
```diff
@@ -0,0 +1,28 @@
...
+When a repository is deleted, all pull-requests will be deleted with it.
+
```
itaiad200, line 22: How would branch protection rules be updated to support pull requests?

design/open/pull-request.md
```diff
@@ -0,0 +1,28 @@
+# Pull-request Proposal
+
+Enable a way to submit changes to a branch by creating a request to merge changes.
+The pull-request captures the merge operation we request to apply.
+
+### Goals
+
+- Capture request to merge a branch or specific reference
+- Mechanism to review, discuss and approve changes before applying to a branch
```
itaiad200, line 9: I don't think `"discuss"` is something that is supported by the below proposal.

itaiad200, line 9: Just merges for now, right?
```suggestion
- Mechanism to review, discuss and approve merges before applying to a branch
```

design/open/pull-request.md
```diff
@@ -0,0 +1,28 @@
...
+- Automation using actions can verify / provide feedback on pull-request
```
itaiad200, line 10: I think we need to elaborate on it. `pull_request_open` trigger with actions that must pass?

design/open/pull-request.md
nopcoder: We can enable client side request to run dry merge - but currelty we don't know if a merge operation will have conflicts until we do the actual merge.

I think it rases another question - does a PR represent the merge itself or the request?
If we have a pending request to merge a commit or a branch, any future request to merge this PR should point to the same PR? Can we have multiple PRs requesting the same thing open in parallel? (will move this one to issues to discuss)

design/open/pull-request.md
nopcoder: Current branch protection enable merge only changes - as PR performs just merge we can use it to push changes into a branch just by merge/PRs.

Additional support can be added to the branch protection by allow only through PRs. And more options like - only approved, with specific checks and etc.

design/open/pull-request.md
nopcoder: Added discussion section that will include zero or more comments. Each comment will be added to the end of the discussion.

nopcoder:
Very exciting!
> I can't wait I think it is the next step for enabling better workflows using lakeFSt

Do we need explicit authorization for PR approval?
> Same as in merge - currently we check create commit on target branch

Should we allow merge (without force) when no reviewers are assigned?
> Can open to discussion - I think we can add future 'commit as admin' option. Or enforce at leave X approval for merge PRs on branch part of branch protection.

PR update == commit / revert / rebase / etc. It means we have to add a hook trigger on all flows? Will the PR merge trigger a secondary hook call (pre/post merge) in addition to the PR hook?
> The current proposal does not include pull-request update as part of branch update. We can include this functionality (open to discussion). But I don't think it should effect the actions we currently triggering and the pull-request event that we will trigger as part of this feature.
The user will be able to configure what will run as part of the pull-request step and if still there is a pre-merge hook - it can still block the pull-request 'Merge' operation.

design/open/pull-request.md
nopcoder: Not going to change the support with pre-merge - it will trigger when you merge the PR as it is equal to a merge operation.
The PR trigger 'pull-request' action after internal merge operation is done without conflicts.
The new event event will enable check the changes before approving the merge.

design/open/pull-request.md
nopcoder: Added some description on the action - currently I think that an action 'pull-request' that will be triggered on creation and by request is enough. Until we support updates.

design/open/pull-request.md
```diff
@@ -6,23 +6,30 @@ The pull-request captures the merge operation we request to apply.
...
+New comments are appended to the end of the discussion and will include a time stamp, commenter and the text with the comment.
 PR can be merge operation is enabled when at least one reviewer approves or when no reviewers are assigned.
 PR can be closed at any time.
-PR will not be updated when new commits into the source branch.
+PR will not be updated with changes to the source or destination branches, like updates to objects, commits, merges, etc. 
 Merge PR will perform a merge operation and merge the PR as merged. If a conflict or an error occurs while merging, it will be returned as a response to the request.
 When a repository is deleted, all pull-requests will be deleted with it.
 
+Actions - no change in current support with pre/post merge. These actions will be triggered when the PR is merged.
+A new event called 'pull-request' will be triggered when is created (or by request). These action will be triggered if no conflict found on an internal commit that will be visible to the action in order to verify the PR data.
```
itaiad200, line 25: What does `by request` mean? Can I rerun an action of a PR? Can I request a trigger for PR that did not run in the PR?

design/open/pull-request.md
```diff
@@ -6,23 +6,30 @@ The pull-request captures the merge operation we request to apply.
...
+
+
 ### Other features to consider
 
 - PR auto-update with latest branch commits. PR will include the base commit and all the changes until the latest commit on branch.
 - Multiple reviews with text from each reviewer. Can be even per level of commit the current PR addresses.
 - New hook can be trigger when PR is updated. Action triggered by PR will be considered as reviewers and the status is approved/rejects based on a successful run.
+- Do we need to support merge dry run to enable UI indication for conflicts?
```
itaiad200, line 33: This can be done automatically, simply diff the 2 and notify on conflict. btw I think that the UI will present the diff and will show conflicts anyhow

design/open/pull-request.md
nopcoder: A request to re-run the action as the target branch can hold/update over time. We will need to enable re-running the action as it verify the merge with the target at a point in time.

itaiad200 (approved):

design/open/pull-request.md
nopcoder: Make sense I'll remove this item, thanks.

design/open/pull-request.md
```diff
@@ -0,0 +1,35 @@
...
+
+### Goals
+
+- Capture request to merge a branch or specific reference
+- Mechanism to review, discuss and merge changes before applying to a branch
+- Automation using actions can verify / provide feedback on pull-request
+
+### How it will work
+
+Introduce a new resource called pull-request at the repository level.
```
N-o-Z, line 14: What is the expected behavior if src or dst were deleted after the PR was opened?

N-o-Z (approved):
Thanks, when do we start? 😄

design/open/pull-request.md
nopcoder: Good point - I'll update the proposal to act only on active branches. When you delete a branch the associated pull-requests should be closed.

N-o-Z: If a Branch is deleted, do we need to go over the list of PR and check if it is associated to any? Or do we close the PR only on a failed Merge attempt?
If we want to couple it, maybe we should put the PRs under the dst branch? This way If a src branch is deleted, the PR will be deleted as part of it.
Also, I think we can ignore deletion of src branch and handle the PR as an independent from it.

nopcoder:
Resolve open items. Main change includes:
1. PR is per source/target branch
2. PR is not for a specific commit
3. PR actions will be triggered on PR creation and on every commit to the source branch.

design/open/pull-request.md
nopcoder: We will require to close all PRs related to the branch - as a source or destination - the idea is to represent a valid request.

Failed merge will keep the PR open and notify the requester that the PR failed to merge.

PR #3743, opened 2022-07-24
TTV: Wizard - Spark templates and Syntax highlighting

Jonathan-Rosenberg:
Added the 3 different versions of templates:
1. Spark submit
2. Core-site.xml
3. Databricks

Also added a syntax highlighting to display the returned configurations properly.

## Visuals:
Screen Shot 2022-07-24 at 17 12 44
Screen Shot 2022-07-24 at 17 12 50
Screen Shot 2022-07-24 at 17 12 57


templates/content/spark.core.site.conf.tt
```diff
@@ -0,0 +1,25 @@
...
+      <name>fs.s3a.impl</name>
+      <value>org.apache.hadoop.fs.s3a.S3AFileSystem</value>
+  </property>
+  <property>
+      <name>fs.s3a.endpoint</name>
+      <value>{{ .Query.lakefs_url }}</value>
+  </property>
+{{- with $creds := new_credentials "spark"}}
+  <property>
+      <name>fs.s3a.access.key</name>
```
johnnyaug, line 17: Wondering if we want to use per-bucket configuration here, with the repository name.
Pros: less intrusive.
Cons: may bot work with older Hadoop versions.

Leaning towards yes. WDYT?

johnnyaug (approved):
🥇 
See one question

talSofer (commented):
Love the design! 
What about EMR? does it require the core site?

templates/content/spark.core.site.conf.tt
Jonathan-Rosenberg: Thought about it, yet as you said, if the users have a Hadoop version older than 2.8, the configurations won't work at all.
Do you have details about the Hadoop API versions users use?

Jonathan-Rosenberg:
@talSofer basically you run a spark job using spark-submit in EMR, so we got that covered

talSofer:
> @talSofer basically you run a spark job using spark-submit in EMR, so we got that covered

Right, forgot this detail - thanks :)

templates/content/spark.core.site.conf.tt
johnnyaug: I don't have this information. However hadoop-aws:2.7.7 is 4 years old. Not sure we have to support it, leaving it for your consideration. If we get rejections in the future we can adapt.

PR #3746, opened 2022-07-24
Update Azure db connection string in configurations example

talSofer:


docs/reference/configuration.md
```diff
@@ -262,7 +262,7 @@ logging:
   output: "-"
 
 database:
-  connection_string: "postgres://user:pass@lakefs.rds.amazonaws.com:5432/postgres"
+  connection_string: "postgres://azureuser@lakefs-db:pass@lakefs-db.postgres.database.azure.com/postgres"
```
talSofer, line 265:
```suggestion
  connection_string: "postgres://user@lakefs-db:pass@lakefs-db.postgres.database.azure.com/postgres"
```

docs/reference/configuration.md
```diff
@@ -262,7 +262,7 @@ logging:
   output: "-"
 
 database:
-  connection_string: "postgres://user:pass@lakefs.rds.amazonaws.com:5432/postgres"
+  connection_string: "postgres://user@lakefs-db:pass@lakefs-db.postgres.database.azure.com/postgres"
```
nopcoder, line 265: I don't think this syntax is supported. https://www.postgresql.org/docs/current/libpq-connect.html
check 34.1.1.2. Connection URIs

nopcoder (changes requested):

docs/reference/configuration.md
talSofer: I have just setup an instance using this syntax :) that's why I'm suggesting this change

talSofer: Maybe there is a difference between Postgres versions?

johnnyaug (approved):

docs/reference/configuration.md
talSofer: @nopcoder WDYT?

nopcoder: Don't think we need to update the documentation - the update example uses a user name called 'user@lakefs-db' which is a specific example.

talSofer:
Cocnluded to discard this pr because username is 'user@lakefs-db' in this case
see https://github.com/treeverse/lakeFS/pull/3746#discussion_r932083766

PR #3750, opened 2022-07-25
KV add commit to branch head

nopcoder:
Add commit to branch head.
The KV implementation will not skip branch update in case the commit is already exists. It will create the commit (if needed) and set it as the last commit done by the branch.
Any change during create commit and the branch update will result in conflict (as done in update branch)

### DIFF
--- a/pkg/graveler/graveler.go
+++ b/pkg/graveler/graveler.go
@@ -1628,6 +1628,20 @@ func (g *KVGraveler) AddCommitToBranchHead(ctx context.Context, repositoryID Rep
 		if branch.CommitID != parentCommitID {
 			return nil, ErrCommitNotHeadBranch
 		}
+
+		// report conflict if there are changes pending on the branch
+		if len(branch.SealedTokens) > 0 {
+			return nil, ErrConflictFound
+		}
+		iterator, err := g.StagingManager.List(ctx, branch.StagingToken, 1)
+		if err != nil {
+			return nil, fmt.Errorf("check branch empty: %w", err)
+		}
+		defer iterator.Close()
+		if iterator.Next() {
+			return nil, ErrConflictFound
+		}
+
 		return &Branch{
 			CommitID:     commitID,
 			StagingToken: branch.StagingToken,
--- a/pkg/graveler/graveler.go
+++ b/pkg/graveler/graveler.go
@@ -1618,27 +1618,25 @@ func (g *KVGraveler) AddCommitToBranchHead(ctx context.Context, repositoryID Rep
 		return "", fmt.Errorf("%w: %s", ErrMetaRangeNotFound, commit.MetaRangeID)
 	}
 
-	// add commit and/or use the commit id of the one we found
+	// add commit to our ref manager
 	commitID, err := g.RefManager.AddCommit(ctx, repositoryID, commit)
-	if err != nil && !errors.Is(err, kv.ErrPredicateFailed) {
+	if err != nil {
 		return "", fmt.Errorf("adding commit: %w", err)
 	}
 
+	// update branch with commit after verify:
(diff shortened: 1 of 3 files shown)


pkg/graveler/graveler.go
```diff
@@ -1597,45 +1597,50 @@ func CommitExists(ctx context.Context, repositoryID RepositoryID, commitID Commi
...
+
+	err = g.RefManager.BranchUpdate(ctx, repositoryID, branchID, func(branch *Branch) (*Branch, error) {
+		if branch.CommitID != parentCommitID {
+			return nil, ErrCommitNotHeadBranch
 		}
-		return commitID, nil
+		return &Branch{
+			CommitID:     commitID,
+			StagingToken: branch.StagingToken,
+			SealedTokens: branch.SealedTokens,
```
itaiad200, line 1634: Shouldn't we check that these are not empty?

itaiad200 (changes requested):
I think you missed a critical check (staging + sealed are not empty) - please add a test for that too.

pkg/graveler/graveler.go
nopcoder: You are right, I missed the part I removed the empty check - there is no way to lock the branch. So checking the sealed tokens will work, but the listing can be an issue.

pkg/graveler/graveler.go
```diff
@@ -1597,45 +1597,64 @@ func CommitExists(ctx context.Context, repositoryID RepositoryID, commitID Commi
...
+	if err != nil {
+		return "", fmt.Errorf("commit missing meta range %s: %w", commit.MetaRangeID, err)
+	}
+	if !ok {
+		return "", fmt.Errorf("%w: %s", ErrMetaRangeNotFound, commit.MetaRangeID)
+	}
+
+	// add commit and/or use the commit id of the one we found
+	commitID, err := g.RefManager.AddCommit(ctx, repositoryID, commit)
+	if err != nil && !errors.Is(err, kv.ErrPredicateFailed) {
```
N-o-Z, line 1623: We don't return ErrPredicateFailed from AddCommit anymore

pkg/graveler/graveler.go
```diff
@@ -1597,45 +1597,64 @@ func CommitExists(ctx context.Context, repositoryID RepositoryID, commitID Commi
...
 		}
 
-		// check if commit already exists.
-		commitID := CommitID(ident.NewHexAddressProvider().ContentAddress(commit))
-		if exists, err := CommitExists(ctx, repositoryID, commitID, g.RefManager); err != nil {
-			return nil, err
-		} else if exists {
-			return commitID, nil
+		// report conflict if there are changes pending on the branch
+		if len(branch.SealedTokens) > 0 {
```
N-o-Z, line 1633: Not necessarily changes - since the sum of the staging area might be empty. Maybe use stagingEmpty instead?

pkg/graveler/graveler.go
```diff
@@ -1597,45 +1597,64 @@ func CommitExists(ctx context.Context, repositoryID RepositoryID, commitID Commi
...
+			return nil, ErrConflictFound
 		}
-
-		commitID, err = g.addCommitNoLock(ctx, repositoryID, commit)
+		iterator, err := g.StagingManager.List(ctx, branch.StagingToken, 1)
```
N-o-Z, line 1636: To list the whole staging area - you need listStagingArea, but I think stagingEmpty should cover your needs

N-o-Z (changes requested):
Looks good, some comments on staging area validation

N-o-Z (approved):
💣

itaiad200 (approved):

PR #3752, opened 2022-07-25
Merge KV Implementation

### LINKED ISSUE #3721: KV Graveler: Implement Merge

itaiad200:
closes #3721

### DIFF
--- a/pkg/graveler/graveler.go
+++ b/pkg/graveler/graveler.go
@@ -985,18 +985,19 @@ func (g *KVGraveler) UpdateBranch(ctx context.Context, repositoryID RepositoryID
 	}
 	newBranch := &Branch{}
 
-	err = g.RefManager.BranchUpdate(ctx, repositoryID, branchID, func(b *Branch) (*Branch, error) {
-		newBranch.CommitID = reference.CommitID
-		newBranch.StagingToken = b.StagingToken
-
+	err = g.RefManager.BranchUpdate(ctx, repositoryID, branchID, func(curBranch *Branch) (*Branch, error) {
 		// TODO(Guys) return error only on conflicts, currently returns error for any changes on staging
-		empty, err := g.stagingEmpty(ctx, repositoryID, repo, b)
+		empty, err := g.stagingEmpty(ctx, repositoryID, repo, curBranch)
 		if err != nil {
 			return nil, err
 		}
 		if !empty {
 			return nil, ErrConflictFound
 		}
+
+		newBranch.CommitID = reference.CommitID
+		newBranch.StagingToken = curBranch.StagingToken
+		newBranch.SealedTokens = curBranch.SealedTokens
 		return newBranch, nil
 	})
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/graveler/graveler.go
```diff
@@ -985,18 +985,19 @@ func (g *KVGraveler) UpdateBranch(ctx context.Context, repositoryID RepositoryID
...
-		empty, err := g.stagingEmpty(ctx, repositoryID, repo, b)
+		empty, err := g.stagingEmpty(ctx, repositoryID, repo, curBranch)
 		if err != nil {
 			return nil, err
 		}
 		if !empty {
 			return nil, ErrConflictFound
 		}
+
+		newBranch.CommitID = reference.CommitID
```
N-o-Z, line 998: Is there a way to perform deep copy? This can potentially save us bugs and maintenance of this code

pkg/graveler/graveler.go
```diff
@@ -1879,7 +1880,116 @@ func (g *KVGraveler) Revert(ctx context.Context, repositoryID RepositoryID, bran
...
+			return nil, fmt.Errorf("check if staging empty: %w", err)
+		}
+		if !empty {
+			return nil, fmt.Errorf("%s: %w", destination, ErrDirtyBranch)
+		}
+		fromCommit, toCommit, baseCommit, err := g.getCommitsForMerge(ctx, repositoryID, source, Ref(destination))
+		if err != nil {
+			return nil, err
+		}
+		g.log.WithFields(logging.Fields{
```
N-o-Z, line 1911: We should log the operation at the Merge level so that we get a logging of all attempts regardless of failures

pkg/graveler/graveler.go
```diff
@@ -1879,7 +1880,116 @@ func (g *KVGraveler) Revert(ctx context.Context, repositoryID RepositoryID, bran
...
+		mergeStrategy := MergeStrategyNone
+		if strategy == MergeStrategyDestWins {
+			mergeStrategy = MergeStrategyDest
+		}
+		if strategy == MergeStrategySrcWins {
+			mergeStrategy = MergeStrategySource
+		}
+		metaRangeID, err := g.CommittedManager.Merge(ctx, storageNamespace, toCommit.MetaRangeID, fromCommit.MetaRangeID, baseCommit.MetaRangeID, mergeStrategy)
+		if err != nil {
+			if !errors.Is(err, ErrUserVisible) {
```
N-o-Z, line 1928: Not sure what the original purpose was - but this Err is checked but never raised. Maybe @arielshaqed can elaborate more about this error

N-o-Z (approved):
@itaiad200 👍🏽  with 2 comments

pkg/graveler/graveler.go
arielshaqed: graveler/errors.go wraps numerous user-visible errors around this error, and AFAIK we _do_ return those errors.  These are errors that get pushed to the user, so must not be further wrapped with debug info.  Part of fixing an old bug where we would drop an inordinate amount of information, some of it potentially sensitive, into errors reported to the user.

N-o-Z: Thanks - I missed the wrapping - BTW is the TODO in errors.go still relevant?

PR #3755, opened 2022-07-26
KV Store: Graveler implement Reset

N-o-Z:
Decided to split the PRs eventually

### DIFF
--- a/pkg/graveler/graveler.go
+++ b/pkg/graveler/graveler.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 	"time"
 
 	"github.com/google/uuid"
@@ -931,6 +930,7 @@ func (g *KVGraveler) CreateBranch(ctx context.Context, repositoryID RepositoryID
 	newBranch := Branch{
 		CommitID:     reference.CommitID,
 		StagingToken: generateStagingToken(repositoryID, branchID),
+		SealedTokens: make([]StagingToken, 0),
 	}
 	storageNamespace := repo.StorageNamespace
 	preRunID := g.hooks.NewRunID()
@@ -1656,7 +1656,7 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
 		}
 		err = g.RefManager.SetBranch(ctx, repositoryID, branchID, Branch{
 			CommitID:     newCommit,
-			StagingToken: newStagingToken(repositoryID, branchID),
+			StagingToken: generateStagingToken(repositoryID, branchID),
 		})
 		if err != nil {
 			return "", fmt.Errorf("set branch commit %s: %w", newCommit, err)
@@ -1696,11 +1696,6 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
 			Error("Post-commit hook failed")
 	}
 	return newCommitID, nil
-}
-
-func newStagingToken(repositoryID RepositoryID, branchID BranchID) StagingToken {
-	v := strings.Join([]string{repositoryID.String(), branchID.String(), uuid.New().String()}, "-")
-	return StagingToken(v)
 }
 
 func validateCommitParent(ctx context.Context, repositoryID RepositoryID, commit Commit, manager RefManager) (CommitID, error) {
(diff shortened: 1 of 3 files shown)


pkg/graveler/graveler.go
```diff
@@ -1851,8 +1846,51 @@ func (g *KVGraveler) stagingEmpty(ctx context.Context, repositoryID RepositoryID
...
+	if err != nil {
+		return err
+	}
+	if isProtected {
+		return ErrWriteToProtectedBranch
+	}
+	currBranch, err := g.RefManager.GetBranch(ctx, repositoryID, branchID)
+	if err != nil {
+		return err
+	}
```
itaiad200, line 1876: Use the branch given to you by the `BranchUpdate` func. The value here may be outdated.
```suggestion
```

pkg/graveler/graveler.go
```diff
@@ -1851,8 +1846,51 @@ func (g *KVGraveler) stagingEmpty(ctx context.Context, repositoryID RepositoryID
...
+
+	err = g.RefManager.BranchUpdate(ctx, repositoryID, branchID, func(branch *Branch) (*Branch, error) {
+		if currBranch.StagingToken != branch.StagingToken { // Check staging token hasn't changed after get with predicate
+			return nil, ErrConflictFound
+		}
```
itaiad200, line 1881: `BranchUpdate` got your back! This check is redundant

pkg/graveler/graveler.go
```diff
@@ -1851,8 +1846,51 @@ func (g *KVGraveler) stagingEmpty(ctx context.Context, repositoryID RepositoryID
...
+		// Zero tokens and try to set branch
+		branch.StagingToken = generateStagingToken(repositoryID, branchID)
+		branch.SealedTokens = make([]StagingToken, 0)
+		return branch, nil
+	})
+
```
itaiad200, line 1887: remove line

pkg/graveler/graveler.go
itaiad200: Just save the old tokens in a list

pkg/graveler/graveler.go
```diff
@@ -1851,8 +1846,51 @@ func (g *KVGraveler) stagingEmpty(ctx context.Context, repositoryID RepositoryID
...
 
+// dropStaging deletes all staging area entries of a given branch from store
+// We do not wait for result as this is an asynchronous operation
+func (g *KVGraveler) dropStaging(ctx context.Context, branch *Branch) {
+	tokens := branch.SealedTokens
+	tokens = append(tokens, branch.StagingToken)
+	for _, st := range tokens {
+		token := st // pinning
+		go func() {
+			err := g.StagingManager.Drop(ctx, token)
```
itaiad200, line 1857: I think the error should be returned, why shadow it from the caller?

itaiad200 (changes requested):
Thanks - dropped some comments. Also - I know we don't have tests before, please consider adding a single sanity test to the graveler `Reset` functionality.

pkg/graveler/graveler.go
N-o-Z: Once we update the branch for all things considered we've reset the branch. Waiting for the error to be returned will prolong the operation and return an error which is not interesting for the user

pkg/graveler/graveler.go
```diff
@@ -1851,8 +1846,45 @@ func (g *KVGraveler) stagingEmpty(ctx context.Context, repositoryID RepositoryID
...
 
+// dropStaging deletes all staging area entries of a given branch from store
+// We do not wait for result as this is an asynchronous operation
+func (g *KVGraveler) dropTokens(ctx context.Context, tokens []StagingToken) {
+	for _, st := range tokens {
+		token := st // pinning
+		go func() {
+			err := g.StagingManager.Drop(ctx, token)
+			if err != nil {
+				logging.Default().WithError(err).WithField("staging_token", token).Error("failed to drop staging token")
```
itaiad200, line 1857:
```suggestion
				logging.Default().WithError(err).WithField("staging_token", token).Error("Failed to drop staging token")
```

N-o-Z:
> Thanks - dropped some comments. Also - I know we don't have tests before, please consider adding a single sanity test to the graveler `Reset` functionality.

@itaiad200 I don't really know how we can unit test this in the graveler context. The whole branch update and drop logics are mocked.
It is covered though by system tests

N-o-Z:
Found a bug in Get flow

pkg/graveler/graveler.go
```diff
@@ -1335,14 +1335,17 @@ func (g *KVGraveler) Get(ctx context.Context, repositoryID RepositoryID, ref Ref
...
+		if err != nil {
+			if !errors.Is(err, ErrNotFound) {
+				return nil, err
+			} // proceed to committed
+		} else { // In staging
+			if value == nil {
+				// tombstone - the entry was deleted on the branch => doesn't exist
+				return nil, ErrNotFound
+			}
+			return value, nil
```
itaiad200, line 1347: Great catch!!

I think it deserves its own PR. Also, there are tests for `Get` flow in Graveler. Can we add/change them to fit this scenario too?

N-o-Z: The tests were there, they were just outdated. Unfortunately I suspect we have many such tests where the flow is not tested correctly due to mock configurations. We should take action once we remove the DB support to refactor the tests and clean up the mocks

itaiad200 (approved):
LGTM

Issue #3410, opened 2022-05-22
Display README.md content if it is present


johnnyaug:
Similarly to GitHub, if a file called _README.md_ is present at a branch's root, render and show it in the UI.

The content should be markdown-rendered and displayed only when viewing the repository root.

arielshaqed:
Done by #3761, nice one!

PR #3761, opened 2022-07-26
Display readme file if available

Jonathan-Rosenberg:
Display a readme file if it exists under the root of the current reference.
## Visual
Screen Shot 2022-07-26 at 20 09 26

### DIFF
--- a/webui/src/pages/repositories/repository/objects.jsx
+++ b/webui/src/pages/repositories/repository/objects.jsx
@@ -302,7 +302,7 @@ const ReadmeContainer = ({repo, reference}) => {
     });
 
     if (loading || error) {
-        return <></>
+        return <></>;
     }
 
     return (
--- a/webui/src/pages/repositories/repository/objects.jsx
+++ b/webui/src/pages/repositories/repository/objects.jsx
@@ -388,7 +388,7 @@ const ObjectsBrowser = ({config, configError}) => {
                 </ActionGroup>
             </ActionsBar>
 
-            <Box sx={{display: 'flex', flexDirection: 'column', gap:'10px'}}>
+            <Box sx={{display: 'flex', flexDirection: 'column', gap: '10px', mb: '30px'}}>
                 <TreeContainer
                     config={config}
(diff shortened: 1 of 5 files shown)


webui/src/pages/repositories/repository/objects.jsx
```diff
@@ -291,6 +296,31 @@ const TreeContainer = ({
     );
 }
 
+const READMEContainer = ({repo, reference}) => {
+    const {response, error, loading} = useAPI(async () => {
+        return await objects.get(repo.id, reference.id, 'README.md');
+    });
+
+    if (loading || error) {
+        return <></>
```
arielshaqed, line 305: nit:
```suggestion
        return <></>;
```

arielshaqed, line 305: Ideally we would display any non-404 errors: if we got here and there _is_ a README.md, the user might be expecting to see it!

arielshaqed (approved):
# NEAT!

I'm assuming remark and react-markdown are well-regarded packages.

I would be happy if we could lock their versions down a bit more: these packages take actual user data and display it to other users, which might be rather more than we've had so far in the GUI.  @Jonathan-Rosenberg + @nopcoder WDYT -- am I just being paranoid here?

webui/src/pages/repositories/repository/objects.jsx
johnnyaug: null also works

webui/src/pages/repositories/repository/objects.jsx
```diff
@@ -291,6 +296,31 @@ const TreeContainer = ({
     );
 }
 
+const READMEContainer = ({repo, reference}) => {
```
johnnyaug, line 299:
```suggestion
const ReadmeContainer = ({repo, reference}) => {
```

johnnyaug (approved):
Awesome!

johnnyaug:
Solves: #3410

PR #3765, opened 2022-07-27
Bug: Fix KV Graveler Get

N-o-Z:
Fixed test to cover this scenario as well


nopcoder (commented):
LGTM, I'll open an issue to handle the tombstone as a value vs null

nopcoder (approved):

PR #3778, opened 2022-07-27
[GC] Changed 'region' parameter to be optional

### LINKED ISSUE #3673: Make "region" on optional parameter for GC
Azure blob client does not require a region, but currently, Garbage collection requires it.

lynnro314:
Closes #3673
Changed 'region' parameter to be optional, and also removed the region parameter from gc tests in azure.


clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -301,17 +301,17 @@ object GarbageCollector {
   }
 
   def main(args: Array[String]) {
-    if (args.length != 2) {
```
talSofer, line 304: I would change this to a check by storage type instead, and exit with an error in case the arguments are invalid

talSofer (changes requested):
Thanks!
See my comment, I believe that we should fail the job in case of invalid arguments. let me know what you think

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
lynnro314: In order to get the storage type (for now) we need to initiate lakeFS client.. Do you think we should do it and fail the job?

lynnro314: Changing it since there is no use of the arguments before initiating the client.

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -301,20 +302,33 @@ object GarbageCollector {
...
+    if (args.length < 1) {
```
talSofer, line 305: I find this confusing... as a user running GC on azure, I will be confused in the following case:
arguments: empty, and the application tells me that I need to use an optional "region", but it is not really optional for Azure 

Can we add a condition special to Azure in an else if at line 326 and remove this one?

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -14,6 +14,7 @@ import org.apache.spark.sql.functions._
 import org.apache.spark.sql.types.{StringType, StructField, StructType}
 import org.apache.spark.sql.{SparkSession, _}
 import com.amazonaws.services.s3.AmazonS3
+import io.treeverse.clients.StorageUtils._
```
arielshaqed, line 17: This is a bit scary, because it _looks_ like it could bring in an unlimited number of strange words.  In fact it isn't -- which also means it's not that useful.  I would _prefer_ (not required in any way, we have no style guide!) to remove this and use a `StorageUtils.` prefix wherever we need it.

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -301,20 +302,33 @@ object GarbageCollector {
...
+    val secretKey = hc.get(LAKEFS_CONF_API_SECRET_KEY_KEY)
+    val apiClient = new ApiClient(apiURL, accessKey, secretKey)
+    val storageType = apiClient.getBlockstoreType()
+
+    if (storageType == StorageTypeS3 && args.length != 2) {
+      Console.err.println(
+        "Usage: ... <repo_name> <region>"
+      )
+      System.exit(1)
+    }
```
arielshaqed, line 326: This does not verify `args.length == 1` when `storageType != StorageTypeS3`.  There are several ways forward: we could [slice](https://www.scala-lang.org/api/2.13.3/scala/Array.html#slice(from:Int,until:Int):Array[A]) off the first element each time we use it (into a new Array) and then verify that nothing is left.

Or we could change the CLI and use a real flags library.  I would prefer this option.  And it will take us less time if we keep the current PR and just open an issue to clean things up.

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -301,20 +302,33 @@ object GarbageCollector {
...
 
     val repo = args(0)
-    val region = args(1)
+    val region = if (args.length == 2) args(1) else ""
```
arielshaqed, line 329:
```suggestion
    val region = if (args.length == 2) args(1) else null
```
`null` is a perfectly good `String` on the JVM, and it is more clearly a "missing value" than the empty string `""`.

arielshaqed (approved):
Neat!

Approved because this is good, and urgent, and my preferred resolution to the major comment is just to open an issue.

THANKS!

talSofer (approved):
Thanks!

PR #3783, opened 2022-07-27
KV Store: Graveler naive delete repository

N-o-Z:


pkg/graveler/graveler.go
```diff
@@ -892,7 +895,71 @@ func (g *KVGraveler) WriteMetaRangeByIterator(ctx context.Context, repositoryID
...
+	defer itr.Close()
+	var wg multierror.Group
+	for itr.Next() {
+		commit := itr.Value()
+		wg.Go(func() error {
+			return g.RefManager.RemoveCommit(ctx, repositoryID, commit.CommitID)
+		})
+	}
+	return wg.Wait().ErrorOrNil()
+}
```
itaiad200, line 944: I'm unsure that creating (#Tags + #Commits + #Branches) goroutines is a good idea. You may be needing workers to make sure you're not creating a 1M routines for large repositories.

itaiad200 (approved):
Added a single comment, let's make sure we control the load for the deletion

pkg/graveler/graveler.go
N-o-Z: @itaiad200  Do you think it's better to just serialize the calls for now? This is just temporary solution, which obviously we will not release KV with

itaiad200: For a temporary solution it's fine for now.

PR #3789, opened 2022-07-28
Add allocs and mutex to pprof available profiles

idanovo:


nopcoder (approved):
I think we can skip write our own router

PR #3794, opened 2022-07-29
kv implementation optimistic lock commit

### LINKED ISSUE #3569: KV Graveler Implementation - Optimistic Lock Commits
Please refer to [Design proposal - lakeFS on KV](https://github.com/treeverse/lakeFS/blob/b3204edad00f88f8eb98524ad940fde96e02ab0a/design/open/metadata_kv/index.md#graveler-metadata---branches-and-staged-writes) and #3566 for further details

**DoD**
Working implementation of Optimistic Lock Commits over KV, with relevant tests

### LINKED ISSUE #3727: KV Graver: Enable skipped Esti tests on completed implementation

eden-ohana:
closes #3569 
closes #3727


pkg/graveler/errors.go
```diff
@@ -53,6 +53,7 @@ var (
 	ErrDeleteDefaultBranch          = wrapError(ErrUserVisible, "cannot delete repository default branch")
 	ErrCommitMetaRangeDirtyBranch   = wrapError(ErrUserVisible, "cannot use source MetaRange on a branch with uncommitted changes")
 	ErrTooManyStagingWriteTries     = errors.New("too many tries to set to staging area")
+	ErrTooManyBranchUpdateTries     = errors.New("too many tries to update branch")
```
nopcoder, line 56: I think it is time to create one error (ErrTooManyTries) and use fmt.Errorf to tell the rest

pkg/graveler/graveler.go
```diff
@@ -1571,6 +1571,29 @@ func (g *KVGraveler) listStagingArea(ctx context.Context, b *Branch) (ValueItera
...
+				it.Close()
+			}
+			return nil, err
+		}
+		itrs = append(itrs, it)
+	}
+
+	if len(itrs) == 0 {
+		return nil, nil
+	}
```
nopcoder, line 1590: 1. We can check it before the 'for' loop.
2. Would expect to get an iterator with no values and not nil if no error was found

pkg/graveler/graveler.go
```diff
@@ -1571,6 +1571,29 @@ func (g *KVGraveler) listStagingArea(ctx context.Context, b *Branch) (ValueItera
 	return NewCombinedIterator(itrs...), nil
 }
 
+func (g *KVGraveler) listSealedTokens(ctx context.Context, b *Branch) (ValueIterator, error) {
+	var itrs []ValueIterator
```
nopcoder, line 1575: can allocate the slice size based on the length of sealed tokens (at least the capacity)

pkg/graveler/graveler_test.go
```diff
@@ -1239,14 +1271,16 @@ func testGravelerCommit(t *testing.T, kvEnabled bool) {
...
-				Values:      values,
+				Values:      tt.values,
 				MetaRangeID: expectedRangeID,
 			}
 			if tt.args.sourceMetarange != nil {
 				expectedAppliedData = testutil.AppliedData{}
 			}
-			if diff := deep.Equal(tt.fields.CommittedManager.AppliedData, expectedAppliedData); diff != nil {
-				t.Errorf("unexpected apply data %s", diff)
+			if kvEnabled || tt.name != "valid commit with staging and sealed" { // TODO - delete when removing the DB
```
nopcoder, line 1280: It looks like we are running this test for 'db' but no compare the results. Prefer to use t.Skip and specify the case and not the name of the test.
For example looks like when !kvEnabled && tt.values != values we would like to skip.

pkg/graveler/graveler.go
```diff
@@ -1606,29 +1629,50 @@ func (g *KVGraveler) List(ctx context.Context, repositoryID RepositoryID, ref Re
...
+		return "", fmt.Errorf("get repository: %w", err)
+	}
+	storageNamespace = repo.StorageNamespace
+
+	branch, err := g.RefManager.GetBranch(ctx, repositoryID, branchID)
+	if err != nil {
+		return "", fmt.Errorf("get branch: %w", err)
+	}
+
+	if params.SourceMetaRange != nil {
```
nopcoder, line 1655: I think this check should be part of the branch update as we like to verify that the branch state doesn't change during this one or verify that the branch information read as part of UpdateBranch wasn't modified as we checked empty staging on specific tokens.

pkg/graveler/graveler.go
```diff
@@ -1606,29 +1629,50 @@ func (g *KVGraveler) List(ctx context.Context, repositoryID RepositoryID, ref Re
...
-		if err != nil {
-			return "", fmt.Errorf("get branch: %w", err)
-		}
+	err = g.RefManager.BranchUpdate(ctx, repositoryID, branchID, func(branch *Branch) (*Branch, error) {
+		tokens := make([]StagingToken, len(branch.SealedTokens)+1)
+		copy(tokens, branch.SealedTokens)
+		tokens[len(tokens)-1] = branch.StagingToken
+
+		branch.StagingToken = generateStagingToken(repositoryID, branchID)
+		branch.SealedTokens = tokens
```
nopcoder, line 1671: Any reason to clone the data?

```suggestion
		branch.SealedTokens = append(branch.SealedTokens, branch.StagingToken)
		branch.StagingToken = generateStagingToken(repositoryID, branchID)
```

pkg/graveler/graveler.go
```diff
@@ -1684,53 +1718,52 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
 		if params.SourceMetaRange != nil {
 			commit.MetaRangeID = *params.SourceMetaRange
 		} else {
-			changes, err := g.StagingManager.List(ctx, branch.StagingToken, ListingMaxBatchSize)
+			changes, err := g.listSealedTokens(ctx, branch)
```
nopcoder, line 1721: My concern here is that if the sealed tokens modified - we can stop trying to commit or do the optimization and remove from the sealed tokens just the tokens we set to commit.
We are using 'retryBranchUpdate' after we updated the branch and we do not checking the specific sealed tokens we set to commit and just list all the latest.

pkg/graveler/graveler.go
```diff
@@ -1745,6 +1778,24 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
 	return newCommitID, nil
 }
 
+func (g *KVGraveler) retryBranchUpdate(ctx context.Context, repositoryID RepositoryID, branchID BranchID, f BranchUpdateFunc) error {
```
nopcoder, line 1781: 1. I think we have more than one backoff/retry packages in the project we can use
2. We should sleep and check context between in order to break if needed and wait between retries

pkg/graveler/graveler.go
```diff
@@ -1745,6 +1778,24 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
...
+		// TODO - if the branch commit id hasn't changed, update the fields instead of fail
+		err := g.RefManager.BranchUpdate(ctx, repositoryID, branchID, f)
+		if errors.Is(err, kv.ErrPredicateFailed) {
+			continue
+		}
+		return err
+	}
+	if try == setTries {
+		return ErrTooManyBranchUpdateTries
+	}
```
nopcoder, line 1795: The loop continues just of `ErrPredicateFailed` - so the only option after the loop is `ErrTooManyBranchUpdateTries` we can simplify the loop.

nopcoder (changes requested):

pkg/graveler/graveler.go
```diff
@@ -1745,6 +1778,24 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
 	return newCommitID, nil
 }
 
+func (g *KVGraveler) retryBranchUpdate(ctx context.Context, repositoryID RepositoryID, branchID BranchID, f BranchUpdateFunc) error {
+	const setTries = 3
+
+	var try int
+	for try = 0; try < setTries; try++ {
+		// TODO - if the branch commit id hasn't changed, update the fields instead of fail
```
itaiad200, line 1786: Is this the next commit algorithm? If so, please link to the relevant github issue.

pkg/graveler/graveler.go
```diff
@@ -1745,6 +1778,24 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
...
+		err := g.RefManager.BranchUpdate(ctx, repositoryID, branchID, f)
+		if errors.Is(err, kv.ErrPredicateFailed) {
```
itaiad200, line 1788: I think we should add a log when we retry

pkg/graveler/graveler.go
```diff
@@ -1642,16 +1686,6 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
 			commit.Parents = CommitParents{branch.CommitID}
 		}
 
-		if params.SourceMetaRange != nil {
```
itaiad200, line 1645: You have to check it here too, checks outside the lock are optimizations at best.

itaiad200: You also need to know if the staging area is empty for "regular" commit - since we don't allow empty commits by default. 
(If an underlying layer is taking care of that, add a comment where it's done)

pkg/graveler/graveler.go
itaiad200: It doesn't really matter. Even if some other commit flow succeeded to add another `stagingToken` after you added yours and now you read it, our commit flow can still succeed if no one updated the branch in between.

pkg/graveler/graveler.go
```diff
@@ -1684,53 +1718,52 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
...
 		}
-		err = g.RefManager.SetBranch(ctx, repositoryID, branchID, Branch{
-			CommitID:     newCommit,
-			StagingToken: generateStagingToken(repositoryID, branchID),
-		})
-		if err != nil {
-			return "", fmt.Errorf("set branch commit %s: %w", newCommit, err)
+
+		for _, st := range branch.SealedTokens {
+			err = g.StagingManager.Drop(ctx, st)
```
itaiad200, line 1740: Very dangerous!
You're dropping the sealed tokens without knowing if the commit succeeded! The commit may fail and you'll lose uncommitted data.

itaiad200: It can also be done in the background..

pkg/graveler/graveler.go
```diff
@@ -1684,53 +1718,52 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
...
 
 			commit.MetaRangeID, _, err = g.CommittedManager.Commit(ctx, storageNamespace, branchMetaRangeID, changes)
 			if err != nil {
-				return "", fmt.Errorf("commit: %w", err)
+				return nil, fmt.Errorf("commit: %w", err)
 			}
 		}
 
 		// add commit
 		newCommit, err := g.RefManager.AddCommit(ctx, repositoryID, commit)
```
itaiad200, line 1734:
```suggestion
		newCommitID, err = g.RefManager.AddCommit(ctx, repositoryID, commit)
```

pkg/graveler/graveler.go
```diff
@@ -1684,53 +1718,52 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
...
-				"branch_id":     branchID,
-				"commit_id":     branch.CommitID,
-				"message":       params.Message,
-				"staging_token": branch.StagingToken,
-			}).Error("Failed to drop staging data")
-		}
-		return newCommit, nil
+
+		newCommitID = newCommit
+		branch.CommitID = newCommit
```
itaiad200, line 1753: I think you need to remove the sealedTokens here

pkg/graveler/graveler_test.go
```diff
@@ -1030,6 +1030,9 @@ func testGravelerCommit(t *testing.T, kvEnabled bool) {
 	expectedCommitID := graveler.CommitID("expectedCommitId")
```
itaiad200, line 1030: The failing flows are also really important to validate: a sealed token was added and a staging token was replaced -> you need to make sure that the branch is in the expected state.

itaiad200 (changes requested):

pkg/graveler/graveler.go
```diff
@@ -1571,6 +1571,29 @@ func (g *KVGraveler) listStagingArea(ctx context.Context, b *Branch) (ValueItera
 	return NewCombinedIterator(itrs...), nil
```
itaiad200, line 1571: I think you can also replace the DB graveler call in `SetHooksHandler` and remove the DB graveler from here entirely.

pkg/graveler/errors.go
eden-ohana: done

pkg/graveler/graveler.go
eden-ohana: done

pkg/graveler/graveler.go
eden-ohana: done

pkg/graveler/graveler.go
eden-ohana: done

pkg/graveler/graveler.go
eden-ohana: done

pkg/graveler/graveler.go
eden-ohana: done

pkg/graveler/graveler.go
eden-ohana: done

pkg/graveler/graveler.go
eden-ohana: done

pkg/graveler/graveler.go
eden-ohana: done

pkg/graveler/graveler.go
eden-ohana: done

pkg/graveler/graveler.go
eden-ohana: done

pkg/graveler/graveler_test.go
eden-ohana: done

pkg/graveler/graveler.go
```diff
@@ -1571,6 +1571,29 @@ func (g *KVGraveler) listStagingArea(ctx context.Context, b *Branch) (ValueItera
...
 }
 
+func (g *KVGraveler) listSealedTokens(ctx context.Context, b *Branch) (ValueIterator, error) {
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/graveler/graveler.go
eden-ohana: changed this function

pkg/graveler/graveler.go
eden-ohana: it's not working if the list is empty

pkg/graveler/graveler_test.go
eden-ohana: I'll add it after you'll add the mocks

pkg/graveler/graveler.go
eden-ohana: done

pkg/catalog/catalog.go
```diff
@@ -212,7 +212,7 @@ func New(ctx context.Context, cfg Config) (*Catalog, error) {
 		settingManager := settings.NewManager(refManager, branchLocker, adapter, cfg.Config.GetCommittedBlockStoragePrefix())
 		protectedBranchesManager = branch.NewProtectionManager(settingManager)
 		stagingManager = staging.NewManager(*cfg.KVStore)
-		gStore = graveler.NewKVGraveler(branchLocker, committedManager, stagingManager, refManager, gcManager, protectedBranchesManager)
```
itaiad200, line 215: 🤩

pkg/graveler/graveler.go
```diff
@@ -1557,9 +1556,18 @@ func (g *KVGraveler) listStagingArea(ctx context.Context, b *Branch) (ValueItera
 		return it, nil
 	}
 
-	itrs := []ValueIterator{it}
+	itrs, err := g.listSealedTokens(ctx, b)
+	if err != nil {
+		return nil, err
+	}
+	itrs = append(itrs, it)
```
itaiad200, line 1563: `it` should be the first iterator. If our tests didn't catch this, it's very bad 😭

pkg/graveler/graveler.go
```diff
@@ -1606,29 +1614,59 @@ func (g *KVGraveler) List(ctx context.Context, repositoryID RepositoryID, ref Re
...
+		if params.SourceMetaRange != nil {
+			empty, err := g.stagingEmpty(ctx, repositoryID, repo, branch)
+			if err != nil {
+				return nil, fmt.Errorf("checking empty branch: %w", err)
+			}
+			if !empty {
+				return nil, ErrCommitMetaRangeDirtyBranch
+			}
 		}
+		branch.SealedTokens = append(branch.SealedTokens, branch.StagingToken)
```
itaiad200, line 1661: Wrong order

pkg/graveler/graveler.go
```diff
@@ -1606,29 +1614,59 @@ func (g *KVGraveler) List(ctx context.Context, repositoryID RepositoryID, ref Re
...
+			return "", ErrCommitMetaRangeDirtyBranch
 		}
-		storageNamespace = repo.StorageNamespace
+	}
 
-		branch, err := g.db.GetBranch(ctx, repositoryID, branchID)
-		if err != nil {
-			return "", fmt.Errorf("get branch: %w", err)
+	err = g.RefManager.BranchUpdate(ctx, repositoryID, branchID, func(branch *Branch) (*Branch, error) {
+		if params.SourceMetaRange != nil {
```
itaiad200, line 1652: I think you also need to check for empty changes for regular commit

pkg/graveler/graveler.go
```diff
@@ -1684,31 +1712,43 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
...
 		} else {
-			changes, err := g.StagingManager.List(ctx, branch.StagingToken, ListingMaxBatchSize)
+			itrs, err := g.listSealedTokens(ctx, branch)
 			if err != nil {
-				return "", fmt.Errorf("staging list: %w", err)
+				return nil, err
+			}
+			if len(itrs) == 0 { // staging empty
+				return nil, ErrNotFound
+			}
```
itaiad200, line 1721: This is a funky situation, I would log it..

itaiad200: Also - I don't think the error is appropriate.

pkg/graveler/graveler.go
```diff
@@ -1684,31 +1712,43 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
...
+			if len(itrs) == 1 {
+				commit.MetaRangeID, _, err = g.CommittedManager.Commit(ctx, storageNamespace, branchMetaRangeID, itrs[0])
+			} else {
+				changes := NewCombinedIterator(itrs...)
+				defer changes.Close()
+				commit.MetaRangeID, _, err = g.CommittedManager.Commit(ctx, storageNamespace, branchMetaRangeID, changes)
 			}
```
itaiad200, line 1728: Maybe refactor the combined iterator to accept a single iterator and avoid code branches here?

pkg/graveler/graveler.go
```diff
@@ -1745,6 +1781,40 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
 	return newCommitID, nil
 }
 
+func (g *KVGraveler) retryBranchUpdate(ctx context.Context, repositoryID RepositoryID, branchID BranchID, f BranchUpdateFunc) error {
+	const (
+		maxInterval = 5
```
itaiad200, line 1786: units

pkg/graveler/graveler.go
```diff
@@ -1745,6 +1781,40 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
...
+		maxElapsed  = 30
```
itaiad200, line 1787: units

itaiad200 (changes requested):

pkg/graveler/graveler.go
eden-ohana: done

pkg/graveler/graveler.go
```diff
@@ -1745,6 +1781,40 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
...
+
+	var try int
+	err := backoff.Retry(func() error {
+		// TODO(3586) - if the branch commit id hasn't changed, update the fields instead of fail
+		err := g.RefManager.BranchUpdate(ctx, repositoryID, branchID, f)
+		if err != nil {
+			if !errors.Is(err, kv.ErrPredicateFailed) {
+				return backoff.Permanent(err)
+			}
+			if try < setTries {
```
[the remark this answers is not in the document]
eden-ohana: yes.
done, thanks!

pkg/graveler/graveler.go
eden-ohana: deleted, I don't understand what to unite.

pkg/graveler/graveler.go
eden-ohana: It is checked by commit operation on the commit manager.
This check is for non-empty changes for the source meta range.

pkg/graveler/graveler.go
eden-ohana: Not sure it should be a part of this pr, I can open an issue

pkg/graveler/graveler.go
eden-ohana: I also added another test to check it

pkg/graveler/graveler.go
eden-ohana: I deleted it, it is checked on the commit manager

pkg/graveler/graveler.go
eden-ohana: graveler.go 1724

pkg/graveler/graveler.go
itaiad200: Still - I think we should be avoiding adding another staging token the commit cannot succeed. Adding a staging token increases the latency of all read operations.
I think it should look something like:
```suggestion
			empty, err := g.stagingEmpty(ctx, repositoryID, repo, branch)
			if err != nil {
				return nil, fmt.Errorf("checking empty branch: %w", err)
			}
			if params.SourceMetaRange == nil && empty{
				return nil, ErrTypeThatWeUseForThis
			}
			if params.SourceMetaRange != nil && !empty {
				return nil, ErrCommitMetaRangeDirtyBranch
			}
```

pkg/graveler/graveler.go
itaiad200: Not done for `params.SourceMetaRange != nil`

itaiad200 (changes requested):

pkg/graveler/graveler.go
```diff
@@ -1745,6 +1779,38 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
...
+	bo.MaxInterval = maxInterval * time.Second
+
+	var try int
+	err := backoff.Retry(func() error {
+		// TODO(3586) - if the branch commit id hasn't changed, update the fields instead of fail
+		err := g.RefManager.BranchUpdate(ctx, repositoryID, branchID, f)
+		if err != nil {
+			if !errors.Is(err, kv.ErrPredicateFailed) {
+				return backoff.Permanent(err)
+			}
```
nopcoder, line 1797: 1. For readability - we can move this check before err != nil
2. The try < setTries can be check with err != nil together

pkg/graveler/graveler.go
```diff
@@ -1745,6 +1779,38 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
...
+			if try < setTries-1 {
+				g.log.WithField("try", try+1).
+					WithField("branchID", branchID).
+					Info("Retrying update branch")
+				try += 1
```
nopcoder, line 1802: By moving this one to the start of the func we can prevent all the -1 +1 while using try

pkg/graveler/graveler.go
```diff
@@ -1684,31 +1712,41 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
...
-				return "", fmt.Errorf("staging list: %w", err)
+				return nil, err
+			}
+			if len(itrs) == 1 {
+				commit.MetaRangeID, _, err = g.CommittedManager.Commit(ctx, storageNamespace, branchMetaRangeID, itrs[0])
+			} else {
+				changes := NewCombinedIterator(itrs...)
+				defer changes.Close()
+				// returns err if the commit is empty (no changes)
+				commit.MetaRangeID, _, err = g.CommittedManager.Commit(ctx, storageNamespace, branchMetaRangeID, changes)
```
nopcoder, line 1725: 1. can we consolidate - no need to have a specific case for one iterator the NewCombinedIterator should optimize it.
2. Looks like the call to Close on the listSealedTokens and combined iterator is not aligned

pkg/graveler/graveler.go
```diff
@@ -1606,29 +1614,59 @@ func (g *KVGraveler) List(ctx context.Context, repositoryID RepositoryID, ref Re
...
 		}
+		branch.SealedTokens = append([]StagingToken{branch.StagingToken}, branch.SealedTokens...)
+		branch.StagingToken = generateStagingToken(repositoryID, branchID)
+		return branch, nil
+	})
+	if err != nil {
+		return "", err
+	}
 
+	err = g.retryBranchUpdate(ctx, repositoryID, branchID, func(branch *Branch) (*Branch, error) {
```
nopcoder, line 1669: How do we make sure that the sealed tokens we try to commit are the one we just updated in the previous update?

nopcoder (changes requested):

pkg/graveler/graveler.go
```diff
@@ -1482,7 +1481,7 @@ func (g *KVGraveler) safeBranchWrite(ctx context.Context, log logging.Logger, re
 		}
 	}
 	if try == setTries {
-		return ErrTooManyStagingWriteTries
+		return fmt.Errorf("set staging area: %w", ErrTooManyTries)
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/graveler/graveler.go
```diff
@@ -1557,9 +1556,18 @@ func (g *KVGraveler) listStagingArea(ctx context.Context, b *Branch) (ValueItera
 		return it, nil
 	}
 
-	itrs := []ValueIterator{it}
+	itrs, err := g.listSealedTokens(ctx, b)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/graveler/graveler.go
eden-ohana: done

pkg/graveler/graveler.go
eden-ohana: It was because of another comment of Itai to check it before the branchUpdate as well, but we decided to delete it. It is checked in the branchUpdate too.

pkg/graveler/graveler.go
```diff
@@ -1745,6 +1779,38 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
...
+	const (
+		maxInterval = 5
+		setTries    = 3
+	)
+	bo := backoff.NewExponentialBackOff()
+	bo.MaxInterval = maxInterval * time.Second
+
+	var try int
+	err := backoff.Retry(func() error {
+		// TODO(3586) - if the branch commit id hasn't changed, update the fields instead of fail
```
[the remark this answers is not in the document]
eden-ohana: https://github.com/treeverse/lakeFS/blob/b8bf1b28086358e960669de41a0e88af671584e3/design/accepted/metadata_kv/index.md?plain=1#L200

I know, but it will be assigned later.

pkg/graveler/graveler.go
eden-ohana: done

pkg/graveler/graveler.go
eden-ohana: we want to lon only from the second time when we retry

pkg/graveler/graveler.go
itaiad200: You don't need to validate that.

itaiad200:
@eden-ohana Followup task: #3833

pkg/graveler/graveler.go
```diff
@@ -1694,40 +1713,53 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
...
 		} else {
-			changes, err := g.StagingManager.List(ctx, branch.StagingToken, ListingMaxBatchSize)
+			itrs, err := g.listSealedTokens(ctx, branch)
 			if err != nil {
-				return "", fmt.Errorf("staging list: %w", err)
+				return nil, err
 			}
-			defer changes.Close()
 
+			changes := NewCombinedIterator(itrs...)
```
itaiad200, line 1737: You can panic here

itaiad200 (approved):

pkg/graveler/graveler.go
```diff
@@ -1932,6 +1995,16 @@ func (g *KVGraveler) stagingEmpty(ctx context.Context, repositoryID RepositoryID
 	return true, nil
 }
 
+func (g *KVGraveler) sealedEmpty(ctx context.Context, repositoryID RepositoryID, repo *Repository, branch *Branch) (bool, error) {
+	itrs, err := g.listSealedTokens(ctx, branch)
+	if err != nil {
+		return false, err
+	}
+	it := NewCombinedIterator(itrs...)
```
nopcoder, line 2003: make sure there is len(iters)>0 for panic - consider returning an empty value iterator - where nothing is returned and no fear of panic in runtime

nopcoder (approved):
Added two things to validate and aprroved

pkg/graveler/graveler.go
```diff
@@ -1694,40 +1713,56 @@ func (g *KVGraveler) Commit(ctx context.Context, repositoryID RepositoryID, bran
...
+			}
+			if len(itrs) == 0 {
+				return nil, ErrNoChanges
 			}
-			defer changes.Close()
 
+			changes := NewCombinedIterator(itrs...)
+			defer changes.Close()
+			// returns err if the commit is empty (no changes)
 			commit.MetaRangeID, _, err = g.CommittedManager.Commit(ctx, storageNamespace, branchMetaRangeID, changes)
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/graveler/graveler.go
eden-ohana: done

PR #3803, opened 2022-07-31
Collect kv store type in metadata manager

### LINKED ISSUE #3633: Add KV Driver Type to metadata
Follow up on #3520
Add a metadata entry with the KVs underlying DB type. This can be taken from the configuration and does not need a store specific implementation

itaiad200:
closes #3633


nopcoder (approved):
nice and simple!

PR #3804, opened 2022-07-31
Modify ResetKey to work on the same staging token

### LINKED ISSUE #3788: KV ResetKey should not introduce new stagingToken
It's simply too pricey. Instead, if needed, write the committed value to the current staging token.

itaiad200:
closes #3788


N-o-Z (approved):
Looks good! 💪🏽

Issue #3769, opened 2022-07-27
Graveler testing improvements


itaiad200:
It's really hard to work with Graveler tests. It relies on custom mocks which make it hard to simulate real scenarios. We end up with unit-tests that do not provide real coverage for flows in the system.

itaiad200:
The P0 flows that must be tested for the KV (IMO):
- Set/Delete/Get entry
- Reset / ResetKey / ResetPrefix
- Any operation that changes the branch HEAD:
	- Commit
	- Merge
	- AddCommitToBranchHead
	- Revert
	- UpdateBranch

PR #3816, opened 2022-08-01
Use gomock to test graveler

itaiad200:
part of #3769 

Used `gomock` to simulate all of Graveler dependencies and see if I can really test `Graveler.Get` as an example. I find it much more easy to use than the self-written fakes that we have. I hope that we will move more and more Graveler tests to `gomock`, especially the new flows we added with the new KV implementation.

### DIFF
--- a/pkg/graveler/graveler_v2_test.go
+++ b/pkg/graveler/graveler_v2_test.go
@@ -1,0 +1,174 @@
+package graveler_test
+
+import (
+	"context"
+	"testing"
+	"time"
+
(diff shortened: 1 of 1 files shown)


eden-ohana (approved):
Cool! Is there an option to mock the same input with different values?

itaiad200:
> Cool! Is there an option to mock the same input with different values?

Yap - you can do it by the order of the calls or pass a function to apply any logic of your liking.

PR #3817, opened 2022-08-01
KV Store: Add repository state

### LINKED ISSUE #3813: KV Graveler: Add repository unique identifier and states
First part of https://github.com/treeverse/lakeFS/issues/3713

N-o-Z:
Added state, uid and modified create flow. Delete flow will be updated in https://github.com/treeverse/lakeFS/issues/3713
closes #3813


pkg/graveler/ref/manager.go
```diff
@@ -84,15 +103,19 @@ func (m *KVManager) getRepositoryRec(ctx context.Context, repositoryID graveler.
...
 		return nil, err
 	}
-	return graveler.RepoFromProto(&repo).Repository, nil
+
+	switch repo.State {
+	case graveler.RepositoryStateActive:
+		return repo, nil
+	case graveler.RepositoryStateDeleted:
+		return nil, graveler.ErrRepositoryInDeletion
+	default:
```
itaiad200, line 116: We have 1 more state. What about it?

pkg/graveler/ref/manager.go
```diff
@@ -140,10 +174,7 @@ func (m *KVManager) ListRepositories(ctx context.Context) (graveler.RepositoryIt
...
 	// TODO: delete me
 	// temp code to align with DB manager. Delete once https://github.com/treeverse/lakeFS/issues/3640 is done
-	_, err := m.GetRepository(ctx, repositoryID)
-	if errors.Is(err, kv.ErrNotFound) {
-		return graveler.ErrRepositoryNotFound
-	}
+	_, err := m.getRepository(ctx, repositoryID)
 	if err != nil {
 		return err
 	}
```
itaiad200, line 180: what is the below TODO?

pkg/graveler/ref/manager.go
```diff
@@ -111,7 +135,7 @@ func (m *KVManager) createBareRepository(ctx context.Context, repositoryID grave
 	return nil
 }
 
-func (m *KVManager) CreateRepository(ctx context.Context, repositoryID graveler.RepositoryID, repository graveler.Repository, token graveler.StagingToken) error {
+func (m *KVManager) CreateRepository(ctx context.Context, repositoryID graveler.RepositoryID, repository graveler.Repository) error {
```
itaiad200, line 138: We said we'll create the repo last...

pkg/graveler/ref/manager.go
```diff
@@ -71,9 +72,27 @@ func NewKVRefManager(executor batch.Batcher, kvStore kv.StoreMessage, addressPro
 	}
 }
 
+func (m *KVManager) updateRepoState(ctx context.Context, repo *graveler.RepositoryRecord, state graveler.RepositoryState) error {
```
itaiad200, line 75: I think some of these usages should be `SetIf`

itaiad200 (changes requested):
Thanks! Simple and clean implementation.
I do think we need to add tests since the repo creation and deletion are becoming more complicated & fragile.

pkg/graveler/ref/manager.go
N-o-Z: Currently we return the default error for state other than in-deletion and active.
This can happen for example when creating repository the second time while it is already being created, the chances of that are very small, thus I prefer not creating a new error type for it

pkg/graveler/ref/manager.go
N-o-Z: Delete flow will be addressed in https://github.com/treeverse/lakeFS/issues/3713

pkg/graveler/ref/manager.go
N-o-Z: For now, I don't see a scenario for SetIf, it will also require us to have another Get or pass the predicate to this function.
I suggest handling this when we have a valid scenario for SetIf

N-o-Z:
Will add test for state once we complete implementation of repository delete

pkg/graveler/ref/kv_migrate_test.go
```diff
@@ -168,6 +167,9 @@ func verifyMigrationResults(t *testing.T, ctx context.Context, kvMgr, dbMgr grav
 		dbr, err := dbMgr.GetRepository(ctx, repo.RepositoryID)
 		require.NoError(t, err)
 		dbr.CreationDate = dbr.CreationDate.UTC()
+		// Fill DB missing fields
+		dbr.State = graveler.RepositoryState_ACTIVE
+		dbr.InstanceUID = kvr.InstanceUID
```
itaiad200, line 172: Can we check it's not empty?

pkg/graveler/graveler.go
```diff
@@ -191,6 +192,18 @@ type Repository struct {
 	StorageNamespace StorageNamespace `db:"storage_namespace"`
 	CreationDate     time.Time        `db:"creation_date"`
 	DefaultBranchID  BranchID         `db:"default_branch"`
+	State            RepositoryState
```
itaiad200, line 195: I think this is the first occurrence that we use the protobuf type in the golang struct. Why did you remove the golang type?

pkg/graveler/graveler.go
```diff
@@ -842,26 +855,27 @@ func (g *KVGraveler) GetRepository(ctx context.Context, repositoryID RepositoryI
...
 
 func (g *KVGraveler) CreateRepository(ctx context.Context, repositoryID RepositoryID, storageNamespace StorageNamespace, branchID BranchID) (*Repository, error) {
-	repo := Repository{
-		StorageNamespace: storageNamespace,
-		CreationDate:     time.Now(),
-		DefaultBranchID:  branchID,
+	_, err := g.RefManager.GetRepository(ctx, repositoryID)
+	if err != nil && !errors.Is(err, ErrRepositoryNotFound) {
+		return nil, err
 	}
```
itaiad200, line 861: What's the purpose of this test if it can happen anyhow during creation? I think we should expect `kv.ErrPredicateFailed` from the creation and act on that.

pkg/graveler/graveler.go
```diff
@@ -842,26 +855,27 @@ func (g *KVGraveler) GetRepository(ctx context.Context, repositoryID RepositoryI
...
 
 func (g *KVGraveler) CreateBareRepository(ctx context.Context, repositoryID RepositoryID, storageNamespace StorageNamespace, defaultBranchID BranchID) (*Repository, error) {
-	repo := Repository{
-		StorageNamespace: storageNamespace,
-		CreationDate:     time.Now(),
-		DefaultBranchID:  defaultBranchID,
+	_, err := g.RefManager.GetRepository(ctx, repositoryID)
+	if err != nil && !errors.Is(err, ErrRepositoryNotFound) {
+		return nil, err
 	}
```
itaiad200, line 875: Same here.

pkg/graveler/graveler.go
```diff
@@ -191,6 +192,18 @@ type Repository struct {
...
 	CreationDate     time.Time        `db:"creation_date"`
 	DefaultBranchID  BranchID         `db:"default_branch"`
+	State            RepositoryState
+	InstanceUID      string
+}
+
+func NewRepository(storageNamespace StorageNamespace, defaultBranchID BranchID) Repository {
+	return Repository{
+		StorageNamespace: storageNamespace,
+		CreationDate:     time.Now().UTC(),
```
itaiad200, line 202: I don't think UTC matters here.

pkg/graveler/graveler.go
```diff
@@ -191,6 +192,18 @@ type Repository struct {
 	StorageNamespace StorageNamespace `db:"storage_namespace"`
 	CreationDate     time.Time        `db:"creation_date"`
 	DefaultBranchID  BranchID         `db:"default_branch"`
+	State            RepositoryState
+	InstanceUID      string
```
itaiad200, line 196: We need documenetation for these non-straightforward properties.

itaiad200 (approved):
Few comments to fix but nothing blocking, thanks!

pkg/graveler/graveler.go
N-o-Z: No need for redundant declaration of of enums - it will require unnecessary maintenance when adding / removing states. We only duplicate proto structs to keep backwards compatibility with DB. Optimally we would never have needed to create golang structs and perform the conversions all the time

pkg/graveler/graveler.go
N-o-Z: It matters - we should not keep time in localized format since protobuf doesn't save time localization information

itaiad200: from `timestamp.pb.go`:
```
func New(t time.Time) *Timestamp {
	return &Timestamp{Seconds: int64(t.Unix()), Nanos: int32(t.Nanosecond())}
}
```

so protobuf's library got your back :)

N-o-Z: It returns a protobuf timestamp object, we are using the golang variation

itaiad200: Right - I thought you were worried about the conversation (I can't think of any other reason to use UTC here)

pkg/graveler/graveler.go
N-o-Z: I think the convention we're trying to hold is always to treat time in UTC. @nopcoder correct me if I'm wrong

N-o-Z: Done

pkg/graveler/ref/kv_migrate_test.go
N-o-Z: Done

pkg/graveler/graveler.go
N-o-Z: We want to provide the user with informative error in case of different repository states.
For example, if repository is "in deletion". The user will know to either wait for deletion to complete and retry or call delete again

PR #3827, opened 2022-08-02
Add preparation step before updating branch HEAD

### LINKED ISSUE #3628: KV Graveler: staging manager find a better solution for Drop
Optimally we would have a drop partition functionality, since that is not a standard for KV stores we might want to implement a process where we flag the staging token in graveler and create a background process to delete all keys for that token

### LINKED ISSUE #3771: Changing the branch HEAD while tokens have entries lead to inconsistent behavior
Operations like `UpdateBranch`, `AddCommitToBranchHead`, `Revert` & `Merge` relies on `stagingEmpty` that checks that all layers of staging area combined doesn't introduce changes to a branch before changing the HEAD. However, once these operation end and change the branch HEAD, they don't necessarily make changes to the tokens which might lead to inconsistencies. 

**Example**:
Branch A: 
- Committed: `foo.txt`: `val1`. 
- Staging token empty.
- No sealed tokens.

Branch B:
- Committed: No entries. 
- Staging token `foo.txt`: `tombstone`.
- No sealed tokens.

Merge from Branch A (source) to Branch B (destination) is allowed. B's staging area is empty since there's a non-effective deletion. Once the merge is done, the state of Branch B is:
- Committed: `foo.txt`: `val1`. 
- Staging token `foo.txt`: `tombstone`.
- No sealed tokens.
That will result with no entries when reading from `B`.

itaiad200:
closes #3771 
closes #3628

### DIFF
--- a/pkg/graveler/graveler.go
+++ b/pkg/graveler/graveler.go
@@ -773,6 +773,10 @@ type StagingManager interface {
 	// Drop clears the given staging area
 	Drop(ctx context.Context, st StagingToken) error
 
+	// DropAsync clears the given staging area eventually. Keys may still exist under the token
+	// for a short period of time after the request was made.
+	DropAsync(ctx context.Context, st StagingToken) error
+
 	// DropByPrefix drops all keys starting with the given prefix, from the given staging area
 	DropByPrefix(ctx context.Context, st StagingToken, prefix Key) error
 }
@@ -1009,11 +1013,16 @@ func (g *KVGraveler) UpdateBranch(ctx context.Context, repositoryID RepositoryID
 	if reference.ResolvedBranchModifier == ResolvedBranchModifierStaging {
 		return nil, fmt.Errorf("reference '%s': %w", ref, ErrDereferenceCommitWithStaging)
 	}
-	newBranch := Branch{}
 
-	err = g.RefManager.BranchUpdate(ctx, repositoryID, branchID, func(curBranch *Branch) (*Branch, error) {
+	if err := g.prepareForCommitIDUpdate(ctx, repositoryID, branchID, repo); err != nil {
+		return nil, err
+	}
+
+	var tokensToDrop []StagingToken
+	var newBranch *Branch
+	err = g.RefManager.BranchUpdate(ctx, repositoryID, branchID, func(currBranch *Branch) (*Branch, error) {
 		// TODO(Guys) return error only on conflicts, currently returns error for any changes on staging
-		empty, err := g.stagingEmpty(ctx, repositoryID, repo, curBranch)
+		empty, err := g.sealedEmpty(ctx, repositoryID, repo, currBranch)
 		if err != nil {
 			return nil, err
 		}
@@ -1021,9 +1030,11 @@ func (g *KVGraveler) UpdateBranch(ctx context.Context, repositoryID RepositoryID
 			return nil, ErrConflictFound
 		}
 
-		newBranch = *curBranch
-		newBranch.CommitID = reference.CommitID
-		return &newBranch, nil
+		tokensToDrop = currBranch.SealedTokens
+		currBranch.SealedTokens = []StagingToken{}
+		currBranch.CommitID = reference.CommitID
+		newBranch = currBranch
+		return currBranch, nil
 	})
 	if err != nil {
 		if errors.Is(err, kv.ErrPredicateFailed) {
@@ -1031,7 +1042,36 @@ func (g *KVGraveler) UpdateBranch(ctx context.Context, repositoryID RepositoryID
 		}
 		return nil, err
 	}
-	return &newBranch, nil
+	g.dropTokens(ctx, tokensToDrop...)
+
+	return newBranch, nil
+}
+
+// prepareForCommitIDUpdate will check that the staging area is empty,
+// before adding a new staging token. It is best to use it before changing
+// the branch HEAD as a preparation for deleting the staging area.
+// see issue #3771 for more information on the algorithm used.
+func (g *KVGraveler) prepareForCommitIDUpdate(ctx context.Context, repositoryID RepositoryID, branchID BranchID, repo *Repository) error {
+	err := g.RefManager.BranchUpdate(ctx, repositoryID, branchID, func(currBranch *Branch) (*Branch, error) {
+		empty, err := g.stagingEmpty(ctx, repositoryID, repo, currBranch)
+		if err != nil {
+			return nil, err
+		}
+		if !empty {
+			return nil, ErrConflictFound
+		}
+
+		currBranch.SealedTokens = append([]StagingToken{currBranch.StagingToken}, currBranch.SealedTokens...)
+		currBranch.StagingToken = GenerateStagingToken(repositoryID, branchID)
+		return currBranch, nil
+	})
+	if err != nil {
+		if errors.Is(err, kv.ErrPredicateFailed) {
+			err = ErrConflictFound
+		}
+		return err
+	}
+	return nil
 }
 
 func (g *KVGraveler) GetBranch(ctx context.Context, repositoryID RepositoryID, branchID BranchID) (*Branch, error) {
@@ -1552,6 +1592,28 @@ func (g *KVGraveler) sealedTokensIterator(ctx context.Context, b *Branch) (Value
 	return changes, nil
 }
 
+func (g *KVGraveler) listSealedArea(ctx context.Context, b *Branch) (ValueIterator, error) {
+	if len(b.SealedTokens) == 0 {
+		return nil, nil
+	}
+
+	var itrs []ValueIterator
+	for _, st := range b.SealedTokens {
+		it, err := g.StagingManager.List(ctx, st, 1)
+		if err != nil {
+			for _, it = range itrs {
+				it.Close()
+			}
+			return nil, err
+		}
+		itrs = append(itrs, it)
+	}
+	if len(itrs) == 1 {
+		return itrs[0], nil
+	}
+	return NewCombinedIterator(itrs...), nil
+}
+
 func (g *KVGraveler) List(ctx context.Context, repositoryID RepositoryID, ref Ref) (ValueIterator, error) {
 	repo, err := g.RefManager.GetRepository(ctx, repositoryID)
 	if err != nil {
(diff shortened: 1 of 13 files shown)


pkg/graveler/graveler.go
```diff
@@ -1984,13 +2032,13 @@ func (g *KVGraveler) stagingEmpty(ctx context.Context, repositoryID RepositoryID
 
 // checkEmpty - staging iterator is not considered empty IFF it contains any non-tombstone entry
 // or a tombstone entry exists for a key which is already committed
-func (g *KVGraveler) checkEmpty(ctx context.Context, repositoryID RepositoryID, repo *Repository, branch *Branch, itr ValueIterator) (bool, error) {
+func (g *KVGraveler) checkEmpty(ctx context.Context, repositoryID RepositoryID, repo *Repository, branch *Branch, changesIt ValueIterator) (bool, error) {
```
N-o-Z, line 2035: Can use the new and improved
 diff uncommitted iterator to save on code duplication

pkg/graveler/graveler_v2_test.go
```diff
@@ -2,11 +2,14 @@ package graveler_test
 
 import (
 	"context"
+	"errors"
 	"testing"
 	"time"
 
 	"github.com/golang/mock/gomock"
 	"github.com/stretchr/testify/require"
+
```
N-o-Z, line 11: Redundant new line

pkg/graveler/graveler.go
```diff
@@ -2014,15 +2065,13 @@ func (g *KVGraveler) sealedEmpty(ctx context.Context, repositoryID RepositoryID,
 }
 
 // dropStaging deletes all staging area entries of a given branch from store
```
N-o-Z, line 2067:
```suggestion
// dropTokens deletes all staging area entries of a given branch from store
```

pkg/graveler/staging/manager.go
```diff
@@ -100,3 +113,38 @@ func (m *Manager) DropByPrefix(ctx context.Context, st graveler.StagingToken, pr
 
 	return nil
 }
+
+func (m *Manager) cleanupLoop(ctx context.Context) {
```
N-o-Z, line 117: Add a sleep in the for loop?
Also, If we want an event loop for managers, I think we should do something more generic that can be reused

pkg/graveler/staging/manager.go
```diff
@@ -100,3 +113,38 @@ func (m *Manager) DropByPrefix(ctx context.Context, st graveler.StagingToken, pr
...
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-m.wakeup:
```
N-o-Z, line 122: When is it false?

pkg/graveler/staging/manager.go
```diff
@@ -10,15 +10,24 @@ import (
...
 
 type Manager struct {
-	store kv.StoreMessage
-	log   logging.Logger
+	store  kv.StoreMessage
+	log    logging.Logger
+	wakeup chan bool
+
+	// notifiers are being called with every successful cleanup cycle
+	notifiers []chan bool
```
N-o-Z, line 18: Who are the notifiers - where is this being used?

pkg/graveler/staging/manager.go
```diff
@@ -10,15 +10,24 @@ import (
...
 }
 
-func NewManager(store kv.StoreMessage) *Manager {
-	return &Manager{
-		store: store,
-		log:   logging.Default().WithField("service_name", "staging_manager"),
+func NewManager(ctx context.Context, store kv.StoreMessage, notifiers ...chan bool) *Manager {
```
nopcoder, line 21: Notifiers are for testing and to make the implementation of this manager simple I'll just add a method that can set callback func to notify on drop completed.
This way we can set a function that will close a channel in the case of our test and we can use it for logging if we like in non-test.

N-o-Z (changes requested):
Very nice changes on the prepare flow.
Some comments and suggestions.
I have some reservations regarding the AsyncDrop mechanism - not sure it belongs in the scope of this PR.
Also I think the event loop should be defined in a more generalized way

pkg/graveler/staging/manager.go
```diff
@@ -10,15 +10,24 @@ import (
...
+	const wakupChanCapacity = 100
+	m := &Manager{
+		store:     store,
+		log:       logging.Default().WithField("service_name", "staging_manager"),
+		wakeup:    make(chan bool, wakupChanCapacity),
```
nopcoder, line 26: I think sync.Cond will fit this one better. we like to signal clean and not trigger it based on the number of times we called drop.

nopcoder, line 26:
```suggestion
		wakeup:    make(chan bool, wakeupChanCapacity),
```

nopcoder (changes requested):

pkg/graveler/graveler.go
```diff
@@ -1070,29 +1074,65 @@ func (g *KVGraveler) UpdateBranch(ctx context.Context, repositoryID RepositoryID
...
+	g.dropTokens(ctx, tokensToDrop...)
+
+	return newBranch, nil
+}
+
+// prepareForCommitIDUpdate will check that the staging area is empty,
+// before adding a new staging token. It is best to use it before changing
+// the branch HEAD as a preparation for deleting the staging area.
+// see issue #3771 for more information on the algorithm used.
+func (g *KVGraveler) prepareForCommitIDUpdate(ctx context.Context, repositoryID RepositoryID, branchID BranchID, repo *Repository) error {
```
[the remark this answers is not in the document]
itaiad200: It's currently not needed as it will always be followed up by another `BranchUpdate`, so using the returned branch from here could result with errors.

pkg/graveler/graveler.go
```diff
@@ -2005,6 +2053,9 @@ func (g *KVGraveler) checkEmpty(ctx context.Context, repositoryID RepositoryID,
 }
 
 func (g *KVGraveler) sealedEmpty(ctx context.Context, repositoryID RepositoryID, repo *Repository, branch *Branch) (bool, error) {
+	if len(branch.SealedTokens) == 0 {
```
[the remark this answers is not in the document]
itaiad200: Done.

pkg/graveler/graveler.go
itaiad200: Done.

pkg/graveler/graveler.go
itaiad200: Done.

pkg/graveler/graveler.go
```diff
@@ -2014,15 +2065,13 @@ func (g *KVGraveler) sealedEmpty(ctx context.Context, repositoryID RepositoryID,
...
-func (g *KVGraveler) dropTokens(_ context.Context, tokens ...StagingToken) {
+func (g *KVGraveler) dropTokens(ctx context.Context, tokens ...StagingToken) {
 	for _, st := range tokens {
-		go func(token StagingToken) {
-			err := g.StagingManager.Drop(context.Background(), token)
-			if err != nil {
-				logging.Default().WithError(err).WithField("staging_token", token).Error("failed to drop staging token")
-			}
-		}(st)
+		token := st // pinning
```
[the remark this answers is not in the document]
itaiad200: Done.

pkg/graveler/graveler_v2_test.go
itaiad200: Done.

pkg/graveler/staging/manager.go
itaiad200: Done.

pkg/graveler/staging/manager.go
itaiad200: 1. We can generalize, if we need to, from the second manager cleanup loop. Trying to accommodate for future use cases is hard.
2. Why adding a sleep?

pkg/graveler/staging/manager.go
itaiad200: synced f2f, doesn't fit since we don't want to split the shutdown and cleanup flows (sync.Condition isn't waitable in a select statement)

pkg/graveler/staging/manager.go
itaiad200: You must have a value for a channel and I hate `struct{}` :)

pkg/graveler/graveler.go
nopcoder: think it is still relevant - I see the comment is 'dropStaging'

pkg/graveler/staging/manager.go
N-o-Z: Creating an event loop with a switch case shouldn't be too difficult. You want to add a sleep to allow context switching - if I got what you want to do right, then you don't really need to check wakeup continuously (i.e. its okay to check the value every couple of seconds)
BTW, consider using a buffer of 'events' and switch casing on event type instead of a specific boolean parameter for a specific use case. You can then rename the function to eventListener and reuse it for any other event based actions

itaiad200: Done (fixed)

N-o-Z (approved):

nopcoder (approved):
LGTM with small comment

pkg/graveler/staging/manager.go
```diff
@@ -80,11 +92,15 @@ func (m *Manager) List(ctx context.Context, st graveler.StagingToken, _ int) (gr
...
-	// Wish we had 'drop partition'... https://github.com/treeverse/lakeFS/issues/3628
-	// Simple implementation
 	return m.DropByPrefix(ctx, st, []byte(""))
 }
 
+func (m *Manager) DropAsync(ctx context.Context, st graveler.StagingToken) error {
+	err := m.store.Store.Set(ctx, []byte(graveler.CleanupTokensPartition()), []byte(st), []byte("stub-value"))
+	m.wakeup <- true
+	return err
+}
```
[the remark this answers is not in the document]
itaiad200: I want to wakeup regardless of error, worst case we'll scan for nothing

pkg/graveler/staging/manager.go
```diff
@@ -10,15 +10,27 @@ import (
...
+	m := &Manager{
+		store:  store,
+		log:    logging.Default().WithField("service_name", "staging_manager"),
+		wakeup: make(chan bool, wakeupChanCapacity),
 	}
+	go m.cleanupLoop(ctx)
+	return m
+}
+
+func (m *Manager) AddCallback(cleanupCallback func()) {
```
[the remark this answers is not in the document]
itaiad200: Done.

PR #3828, opened 2022-08-02
fix abuse client transport and print generator name

nopcoder:


cmd/lakectl/cmd/abuse.go
```diff
@@ -64,7 +64,7 @@ var abuseRandomReadsCmd = &cobra.Command{
 		}
 		Fmt("read a total of %d keys from key file\n", len(keys))
 
-		generator := stress.NewGenerator(parallelism, stress.WithSignalHandlersFor(os.Interrupt, syscall.SIGTERM))
+		generator := stress.NewGenerator("read", parallelism, stress.WithSignalHandlersFor(os.Interrupt, syscall.SIGTERM))
```
idanovo, line 67: const for actions?

cmd/lakectl/cmd/abuse.go
```diff
@@ -344,7 +343,7 @@ func init() {
 	abuseCmd.AddCommand(abuseCreateBranchesCmd)
 	abuseCreateBranchesCmd.Flags().String("branch-prefix", "abuse-", "prefix to create branches under")
 	abuseCreateBranchesCmd.Flags().Bool("clean-only", false, "only clean up past runs")
-	abuseCreateBranchesCmd.Flags().Int("amount", 1000000, "amount of things to do")
+	abuseCreateBranchesCmd.Flags().Int("amount", 100000, "amount of things to do")
```
idanovo, line 346: nice
I would use const
Do you think the default values for the other commands(read, write, commit, list) here are OK?

idanovo (approved):
Two minor comments

cmd/lakectl/cmd/abuse.go
nopcoder: Here I think const is not required as it is like print out message with `Printf`

cmd/lakectl/cmd/abuse.go
nopcoder: 1. Will change to `const`
2. Will revert to 1m - I wanted to see the end

PR #3829, opened 2022-08-02
KV Store: Implement settings manager

### LINKED ISSUE #3785: KV Settings Manager
SettingsManager relies on BranchLocker to keep the object store settings file consistent. 
Since we no longer have a `BranchLocker` in the kv world, we need to replace the settings manager.

See previous related discussions: #2491 #2498

N-o-Z:
Excluding implementation of migrate functionality.

- Closes #3785


pkg/graveler/settings/manager.go
```diff
@@ -8,49 +8,73 @@ import (
...
+	return kv.FormatPath(settingsPrefix, key)
+}
+
+// TODO (niro): Remove interface once DB implementation is deleted
+type Manager interface {
+	Save(context.Context, graveler.RepositoryID, string, proto.Message) error
+	GetLatest(context.Context, graveler.RepositoryID, string, proto.Message) (proto.Message, error)
+	Get(context.Context, graveler.RepositoryID, string, proto.Message) (proto.Message, error)
+	Update(context.Context, graveler.RepositoryID, string, proto.Message, updateFunc) error
+	WithCache(ache cache.Cache)
```
itaiad200, line 48:
```suggestion
	WithCache(cache cache.Cache)
```

pkg/graveler/settings/manager.go
```diff
@@ -71,20 +95,55 @@ func (m *Manager) Save(ctx context.Context, repositoryID graveler.RepositoryID,
 	if err != nil {
 		return err
 	}
-	err = m.blockAdapter.Put(ctx, block.ObjectPointer{
+
+	data := Settings{
+		Identifier: fmt.Sprintf(settingsUniqueRelativeKey, m.committedBlockStoragePrefix, key, xid.NewWithTime(time.Now())),
```
itaiad200, line 100: That's the default behaviour
```suggestion
		Identifier: fmt.Sprintf(settingsUniqueRelativeKey, m.committedBlockStoragePrefix, key, xid.New()),
```

pkg/graveler/settings/manager.go
```diff
@@ -71,20 +95,55 @@ func (m *Manager) Save(ctx context.Context, repositoryID graveler.RepositoryID,
...
-	}, int64(len(messageBytes)), bytes.NewReader(messageBytes), block.PutOpts{})
+	}
+	// Save to  block store
+	err = m.blockAdapter.Put(ctx, obj, int64(len(messageBytes)), bytes.NewReader(messageBytes), block.PutOpts{})
+	if err != nil {
+		return err
+	}
+
+	// Write key in KV Store
+	err = m.store.SetMsg(ctx, graveler.RepoPartition(&graveler.RepositoryRecord{RepositoryID: repositoryID, Repository: repo}), []byte(settingsPath(key)), &data)
```
itaiad200, line 114: If succesful, you should delete the old object from the store, it could accumulate pretty fast..

pkg/graveler/settings/manager.go
```diff
@@ -131,34 +186,73 @@ func (m *Manager) getFromStore(ctx context.Context, repositoryID graveler.Reposi
...
+	err = update(setting)
+	if err != nil {
+		return err
+	}
+	messageBytes, err := proto.Marshal(setting)
+	if err != nil {
+		return err
+	}
+	oldIdentifier := data.Identifier // Save old identifier for deletion
+	data.Identifier = fmt.Sprintf(settingsUniqueRelativeKey, m.committedBlockStoragePrefix, key, xid.NewWithTime(time.Now()))
```
itaiad200, line 223: To avoid inconsistencies, let's put this repeating line in a func

pkg/graveler/settings/manager.go
```diff
@@ -131,34 +186,73 @@ func (m *Manager) getFromStore(ctx context.Context, repositoryID graveler.Reposi
...
+	}
+
+	// Write key in KV Store
+	err = m.store.SetMsgIf(ctx, graveler.RepoPartition(&graveler.RepositoryRecord{RepositoryID: repositoryID, Repository: repo}), []byte(settingsPath(key)), &data, pred)
+	if err != nil {
+		// Delete file from blockstore
+		_ = m.blockAdapter.Remove(ctx, obj)
+		if errors.Is(err, kv.ErrPredicateFailed) {
+			err = fmt.Errorf("save setting in progress: %w", graveler.ErrPreconditionFailed)
 		}
```
itaiad200, line 242: I think this error should be repeated. This error is due to an implementation detail, retrying here should mask this away from the user

itaiad200 (changes requested):
🆒

pkg/graveler/settings/manager.go
N-o-Z: Do you suggest update should be retried?

nopcoder (changes requested):
Couple of small things - I think we can keep the complete settings inside KV and pay the extra loading of the data from block storage while we migrate.

N-o-Z:
> Couple of small things - I think we can keep the complete settings inside KV and pay the extra loading of the data from block storage while we migrate.

I also agree. @itaiad200  What do you say? should we give it a go?
@johnnyaug small question, does the GC consume the repo settings file? If so, does it read it directly from the store or does it use an API call?

N-o-Z:
> > Couple of small things - I think we can keep the complete settings inside KV and pay the extra loading of the data from block storage while we migrate.
> 
> I also agree. @itaiad200 What do you say? should we give it a go? @johnnyaug small question, does the GC consume the repo settings file? If so, does it read it directly from the store or does it use an API call?

Will reimplement based on a DB solution. I will implement the migration process on the same PR as well to see that there are no blocking issues with it

johnnyaug:
> small question, does the GC consume the repo settings file? If so, does it read it directly from the store or does it use an API call?

@N-o-Z , It does use the repository settings, and it does so using the API.

N-o-Z:
> > small question, does the GC consume the repo settings file? If so, does it read it directly from the store or does it use an API call?
> 
> @N-o-Z , It does use the repository settings, and it does so using the API.

Thanks - perfect!

N-o-Z:
DB only implementation: https://github.com/treeverse/lakeFS/pull/3848

PR #3841, opened 2022-08-04
KV Store: graveler delete repository step 1

### LINKED ISSUE #3713: Repository Creation and Deletion - Step 1 - Repository State and Unique Identifier

N-o-Z:
- Closes #3713


pkg/graveler/ref/manager.go
```diff
@@ -7,6 +7,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/treeverse/lakefs/pkg/logging"
+
```
itaiad200, line 11: delete line

pkg/graveler/ref/manager.go
```diff
@@ -163,16 +165,61 @@ func (m *KVManager) ListRepositories(ctx context.Context) (graveler.RepositoryIt
...
+	repo, err := m.getRepositoryRec(ctx, repositoryID)
 	if err != nil {
 		return err
 	}
-	// END TODO: delete me
 
-	return m.kvStore.DeleteMsg(ctx, graveler.RepositoriesPartition(), []byte(graveler.RepoPath(repositoryID)))
+	// Set repository state to deleted and then perform background delete.
+	if repo.State != graveler.RepositoryState_IN_DELETION {
+		err = m.updateRepoState(ctx, repo, graveler.RepositoryState_IN_DELETION)
```
itaiad200, line 215: I can think of a scenario where one thread deletes the state and then another one creates it with state `RepositoryState_IN_DELETION`. I think `updateRepoState` should use `SetIf`

pkg/graveler/ref/manager.go
```diff
@@ -163,16 +165,61 @@ func (m *KVManager) ListRepositories(ctx context.Context) (graveler.RepositoryIt
...
+	return m.kvStore.SetMsg(ctx, graveler.RepositoriesPartition(), []byte(graveler.RepoPath(repo.RepositoryID)), graveler.ProtoFromRepo(repo))
+}
+
+func (m *KVManager) deleteRepository(ctx context.Context, repo *graveler.RepositoryRecord) error {
+	// ctx := context.Background() TODO (niro): When running this async create a new context and remove ctx from signature
+	logger := logging.FromContext(ctx)
+	// Scan through repository partition and delete entities, using SecondaryIndex struct since we don't really need the data, just the key
+	// Failures on deleting repository entities don't constitute as repository deletion failure
+	partition := graveler.RepoPartition(repo)
+	it, err := kv.NewPartitionIterator(ctx, m.kvStore.Store, (&kv.SecondaryIndex{}).ProtoReflect().Type(), partition)
```
itaiad200, line 179: Good old `m.kvStore.Store.Scan` is even easier if you only care about the keys.

pkg/graveler/ref/manager.go
```diff
@@ -163,16 +165,61 @@ func (m *KVManager) ListRepositories(ctx context.Context) (graveler.RepositoryIt
...
+	if err != nil {
```
itaiad200, line 180: You should exit here.

pkg/graveler/ref/manager.go
```diff
@@ -163,16 +165,61 @@ func (m *KVManager) ListRepositories(ctx context.Context) (graveler.RepositoryIt
...
+		v := it.Entry()
+		err = m.kvStore.DeleteMsg(ctx, partition, v.Key)
+		if err != nil && !errors.Is(err, kv.ErrNotFound) {
+			logger.WithField("repository", repo.RepositoryID).
+				WithField("key", string(v.Key)).
+				Error("Failed to delete repository entity")
+		}
+	}
+	if it.Err() != nil {
+		if err != nil {
```
itaiad200, line 196: Did you mean to use this `if` here?

pkg/graveler/ref/manager.go
```diff
@@ -163,16 +165,61 @@ func (m *KVManager) ListRepositories(ctx context.Context) (graveler.RepositoryIt
...
+
+func (m *KVManager) deleteRepository(ctx context.Context, repo *graveler.RepositoryRecord) error {
+	// ctx := context.Background() TODO (niro): When running this async create a new context and remove ctx from signature
+	logger := logging.FromContext(ctx)
+	// Scan through repository partition and delete entities, using SecondaryIndex struct since we don't really need the data, just the key
+	// Failures on deleting repository entities don't constitute as repository deletion failure
+	partition := graveler.RepoPartition(repo)
+	it, err := kv.NewPartitionIterator(ctx, m.kvStore.Store, (&kv.SecondaryIndex{}).ProtoReflect().Type(), partition)
+	if err != nil {
+		logger.WithField("repository", repo.RepositoryID).
```
itaiad200, line 181: You can start this function with `logger := logger.WithField("repository", repo.RepositoryID)` and use `logger` everywhere instead of creating the same field again and again

pkg/graveler/ref/manager_test.go
```diff
@@ -150,6 +169,17 @@ func TestManager_DeleteRepository(t *testing.T) {
...
 				t.Fatalf("expected ErrRepositoryNotFound, got: %v", err)
 			}
+
+			// Create after delete
+			testutil.Must(t, tt.refManager.CreateRepository(ctx, repoID, graveler.Repository{
+				StorageNamespace: "s3://foo",
+				CreationDate:     time.Now(),
+				DefaultBranchID:  "weird-branch",
+			}))
+			_, err = tt.refManager.GetRepository(context.Background(), "example-repo")
```
itaiad200, line 179: I would validate the returned repo, at least that the unique identifier is different that the one of the deleted repo.

webui/src/pages/repositories/repository/error.jsx
```diff
@@ -0,0 +1,37 @@
...
+                    return router.push('/repositories')
+                }}>
+                    <TrashIcon/> Delete Repository
+                </Button>
+            </div>
+        </Alert>
+    );
+};
+
+export const RepoError = ({error}) => {
```
itaiad200, line 32: Nice one!

webui/src/pages/repositories/repository/error.jsx
```diff
@@ -0,0 +1,37 @@
...
+    if (error instanceof RepositoryDeletionError) {
+        return RepositoryInDeletionContainer(error.repoId);
+    }
+    return Error(error)
+}
```
itaiad200, line 37: newline

itaiad200 (changes requested):
Few comments that require fixing but other than that looks great!
I would add the new screens as images to the discussion in the PR - much easier to review UI that way..

webui/src/pages/repositories/repository/error.jsx
itaiad200: However it might be hard for the next developer to understand when to use `Error` and when to use `RepoError`. 
Consider switching the usage, let everyone call the simple `Error` and there perform this check and initiate `RepositoryInDeletionContainer` for `RepositoryDeletionError`.

N-o-Z: We wanted to bound the usage of RepoError to the repository modules so that it will be accessible only in this context. Also this error is being used only on specific flows

pkg/graveler/ref/manager.go
N-o-Z: Unless I'm missing something - repo is always created in ACTIVE and the only scenario of updateRepoState currently is for delete. In this scenario I will always want to set it to deleted regardless of previous state. I suggest we defer this for when we will actually see the need for setIf. I'm afraid this will just introduce uninteresting bugs

pkg/graveler/ref/manager.go
N-o-Z: This is exactly what I get from PartitionIterator - I prefer using it over a more low level call

pkg/graveler/ref/manager.go
N-o-Z: I disagree. The most important part is the deletion of the repository itself. I prefer returning a successful removal of the repository and leaving garbage in the DB than failing the operation. WDYT?

pkg/graveler/ref/manager.go
N-o-Z: No I did not :)
Thanks

N-o-Z:
Updated repository in deletion window:
image

pkg/graveler/ref/manager_test.go
N-o-Z: Creation of instance ID is in the graveler scope not in ref manager scope - unless you want to change that, this check is irrelevant

pkg/graveler/ref/manager.go
itaiad200: The `Delete` below might happen before some other flow updates the repo state to `RepositoryState_IN_DELETION`. Then you returned to that state, although there's nothing more to do.

pkg/graveler/ref/manager.go
itaiad200: I think this is an unexpected error and hence should not be ignored.

itaiad200: RepositoryState_IN_DELETION

pkg/graveler/ref/manager.go
itaiad200: Again 😿

itaiad200 (approved):

PR #3843, opened 2022-08-04
Batch kv ref manager requests

### LINKED ISSUE #3842: Batch KVRef manager requests
Just like we do with the DBManager. Batch requests and reduce the number of needed connections to postgres dramatically.

itaiad200:
closes #3842


nopcoder (approved):
LGTM - good catch and speed boost

Issue #3686, opened 2022-07-19
"Test lakeFS multipart upload with Hadoop S3A" is flaky


itaiad200:
Example: https://github.com/treeverse/lakeFS/runs/7395107778?check_suite_focus=true#logs

Looks like it's always a network issue:
```
log4j:WARN See http://logging.apache.org/log4j/1.2/faq.html#noconfig for more info.
[error] (run-main-0) com.amazonaws.AmazonClientException: Unable to upload part: Unable to execute HTTP request: Broken pipe (Write failed)
[error] com.amazonaws.AmazonClientException: Unable to upload part: Unable to execute HTTP request: Broken pipe (Write failed)
[error] 	at com.amazonaws.services.s3.transfer.internal.UploadMonitor.collectPartETags(UploadMonitor.java:256)
[error] 	at com.amazonaws.services.s3.transfer.internal.UploadMonitor.completeMultipartUpload(UploadMonitor.java:236)
```

arielshaqed:
Plan: start by skipping the test.  It is old and not particularly informative.

arielshaqed:
> Plan: start by skipping the test. It is old and not particularly informative.

Updated: Reduce network access and total size.  This is not a stress test, we only need to exercise MPU to ensure that it works.

arielshaqed:
_Should_ have been fixed by #3850.  Closing, but _please_ feel free to reopen if it happens again.

PR #3850, opened 2022-08-07
Reduce S3A MPU test runtime and size-of-data

arielshaqed:
* Smaller file: easier upload.  (Still 10MiB, which is above the 5MiB
  theshold for multipart upload!).
* Bump write sizes 16x: fewer, faster network operations.

_Should_ reduce occurrences of #3686.

### DIFF
--- a/test/spark/s3a-multipart/src/main/scala/S3AMultipart.scala
+++ b/test/spark/s3a-multipart/src/main/scala/S3AMultipart.scala
@@ -13,8 +13,8 @@ object S3AMultipart extends App {
 
   override def main(args: Array[String]) {
     val partSize = 5 << 20 // Must be >= 5MiB on AWS S3.
-    val fileSize = 4 * partSize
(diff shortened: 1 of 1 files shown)


Jonathan-Rosenberg:
How was this tested?

arielshaqed:
> How was this tested? Or is it just a hypothesis that will be tested as we go?

Unfortunately deflaking is hard to test.  I just ran tests on the PR, and will do so again.  But because flaking out is self-correlated (not a stationary process, it has good times and bad times), it is not possible to test that the tests are sufficiently less flaky.

Jonathan-Rosenberg (approved):

arielshaqed:
Thanks!

Running another test; if it works _again_ then it's **definitely** (_maybe_) less flaky, I hope... and I shall merge.

arielshaqed:
Passed 3 times, what could possibly go wrong?!

PR #3855, opened 2022-08-07
Templating service: Log expand template

Jonathan-Rosenberg:


arielshaqed (approved):
Yikes, was sure we report it from the controller infrastructure and missed that one.

Thanks!

PR #3856, opened 2022-08-07
Remove postgres specific database collection information from diagnostics

nopcoder:
Updated `lakefs diagnostics` to work with kv.
Removed all specific postgres database related information collection.


pkg/api/serve_test.go
```diff
@@ -10,6 +10,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/treeverse/lakefs/pkg/stats"
+
```
N-o-Z, line 14:
```suggestion
	"github.com/treeverse/lakefs/pkg/stats"
```

cmd/lakefs/cmd/run.go
```diff
@@ -171,26 +171,28 @@ var runCmd = &cobra.Command{
 			idGen = &actions.DecreasingIDGenerator{}
 
 			// init authentication
-			if !cfg.IsAuthTypeAPI() {
+			if authService == nil {
```
N-o-Z, line 174: Just a thought, this could easily go wrong if someone accidentally adds an initialization code for auth service on the wrong line of code. Maybe we should introduce an authServiceType to the configuration (obviously not for this PR)

N-o-Z (approved):
Looks good, only one comment

nopcoder:
Thanks @N-o-Z for the review - updated the with single initialization code block for the auth service. Let me know if this makes the code cleaner and easy to read.

N-o-Z (approved):
LGTM

cmd/lakefs/cmd/run.go
```diff
@@ -141,22 +139,8 @@ var runCmd = &cobra.Command{
...
-				cfg.GetAuthAPIEndpoint(),
-				cfg.GetAuthAPIToken(),
-				crypt.NewSecretStore(cfg.GetAuthEncryptionSecret()),
-				cfg.GetAuthCacheConfig(), nil, apiEmailer)
-			if err != nil {
-				logger.WithError(err).Fatal("failed to create authentication service")
-			}
-		}
 
+		idGen := &actions.DecreasingIDGenerator{}
```
N-o-Z, line 143: This is your bug 😄

N-o-Z (commented):
The issue is that you removed the use of the increasing ID generator

PR #3861, opened 2022-08-08
Try running dynamodb without postgres

itaiad200:


N-o-Z:
@itaiad200 FYI - can't assign you as a reviewer since you opened the PR

cmd/lakefs/cmd/run.go
```diff
@@ -112,27 +113,60 @@ var runCmd = &cobra.Command{
...
+			kvParams = cfg.GetKVParams()
+			kvStore, err = kv.Open(ctx, kvParams)
+			if err != nil {
+				logger.WithError(err).Fatal("Failed to open KV store")
+			}
+			defer kvStore.Close()
 
+			// Check if migration required only on postgres
+			migrationRequired := false
+			if dbParams.Type == kvpg.DriverName {
```
itaiad200, line 134: Does this mean that every DynamoDB installation will work even if no migration was made?

cmd/lakefs/cmd/setup.go
```diff
@@ -25,10 +26,23 @@ var setupCmd = &cobra.Command{
...
+		)
+		if dbParams.KVEnabled {
+			kvParams = cfg.GetKVParams()
+			migrator = kv.NewDatabaseMigrator(kvParams)
+		} else {
+			dbPool = db.BuildDatabaseConnection(ctx, dbParams)
+			defer dbPool.Close()
+
+			migrator = db.NewDatabaseMigrator(dbParams)
+
```
itaiad200, line 44: remove line

esti/migrate_test.go
```diff
@@ -10,6 +10,8 @@ import (
 	"testing"
 	"time"
 
+	kvpg "github.com/treeverse/lakefs/pkg/kv/postgres"
+
```
itaiad200, line 14: goimports

esti/migrate_test.go
```diff
@@ -143,6 +145,10 @@ var (
 )
 
 func TestMigrate(t *testing.T) {
+	// skip test if not on postgres
+	if viper.GetString("database_type") != kvpg.DriverName {
```
itaiad200, line 149: This value doesn't exist in the pre-test, and I don't think it should either..

pkg/kv/params/database.go
```diff
@@ -1,6 +1,7 @@
 package params
 
 type KV struct {
+	Type     string
```
itaiad200, line 4: Nice one.

pkg/kv/store.go
```diff
@@ -14,7 +14,7 @@ import (
 const (
 	InitialMigrateVersion = 1
 	PathDelimiter         = "/"
-	DBSchemaVersionKey    = "kv" + PathDelimiter + "schema" + PathDelimiter + "version"
+	dbSchemaVersionKey    = "kv" + PathDelimiter + "schema" + PathDelimiter + "version"
```
itaiad200, line 17: We have a function for this concatenation now.

cmd/lakefs/cmd/run.go
itaiad200: After reading the code of `ValidateSchemaVersion` there are few things I don't understand. Where can I go to to understand more about the migration process?

pkg/kv/migration.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+
+	case err != nil:
+		return fmt.Errorf("get KV schema version: %w", err)
+
+	case kvVersion < InitialMigrateVersion:
+		if migrationRequired {
+			return fmt.Errorf("migration required, for more information see https://docs.lakefs.io/deploying-aws/upgrade.html: %w", err)
+		} else {
+			return fmt.Errorf("missing KV schema version (%d): %w", kvVersion, err)
```
itaiad200, line 57: Why does this imply that the version is missing?

pkg/kv/migration.go
```diff
@@ -0,0 +1,61 @@
...
+	params kvparams.KV
+}
+
+func NewDatabaseMigrator(params kvparams.KV) *DatabaseMigrator {
+	return &DatabaseMigrator{
+		params: params,
+	}
+}
+
+// Migrate TODO (niro): Currently just set up the KV DB Schema version. Need to create a migration flow for KV
```
itaiad200, line 26: I don't understand what migration is supposed to be run here.

N-o-Z: Currently I'm keeping the same conventions as in DB. When running setup - it runs the migration flow to reach the most updated DB schema (even though the tables are empty). If and when we reach the point where KV migrations are needed we need to revisit this code and decide how to handle it.

cmd/lakefs/cmd/run.go
N-o-Z: This code handles the specific checks for migration from DB to KV. It will change on post KV migration.
It basically checks that if we are using postgres - that we do it only after we migrated the data or alternatively start with a fresh install.
Since we do not support migration directly to dynamodb, it is up to the user to make sure he doesn't try to upgrade a previous lakeFS version without going through the proper migration process via Postgres. This will be documented in the upgrade guide

pkg/kv/migration.go
N-o-Z: This is a situation that should never happen. It means the there is no old DB data that requires migration and that the kv version is corrupted (either migration or setup occurred -> version should be >= 1, otherwise no entry should exist)

cmd/lakefs/cmd/run.go
```diff
@@ -112,27 +113,60 @@ var runCmd = &cobra.Command{
...
+				dbPool = db.BuildDatabaseConnection(ctx, dbParams)
+				defer dbPool.Close()
+				_, _, err := db.MigrateVersion(ctx, dbPool, dbParams)
+				if err == nil {
+					migrationRequired = true
+				} else if !errors.Is(err, migrate.ErrNilVersion) {
+					logger.WithError(err).Fatal("Failed to get schema version")
+				}
```
nopcoder, line 142: What will happen if we start lakefs with kv enabled with blank postgress - it will fail?

esti/system_test.go
```diff
@@ -225,14 +226,15 @@ func listRepositories(t *testing.T, ctx context.Context) []api.Repository {
...
+	if !viper.GetBool("database_kv_enabled") ||
+		viper.GetString("database_type") != kvpg.DriverName {
+		t.Skip("PG KV not enabled")
 	}
 
 	// connect to database and verify that kv table exists
 	dbURI := viper.GetString("database_connection_string")
 	if dbURI == "" {
-		t.Fatal("lakefs database connection string environment variable is missing")
+		t.Skip("lakefs database connection string environment variable is missing")
```
nopcoder, line 237: The DB URI in postgres is required - why not fail?

nopcoder (changes requested):

cmd/lakefs/cmd/run.go
N-o-Z: No - since we will get ErrNilValue - so we won't go into this if clause

cmd/lakefs/cmd/run.go
```diff
@@ -112,27 +113,60 @@ var runCmd = &cobra.Command{
...
+			}
+			err = kv.ValidateSchemaVersion(ctx, kvStore, migrationRequired)
+			if err != nil {
+				logger.WithError(err).Fatal("Failure on schema validation")
```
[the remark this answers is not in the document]
N-o-Z: Fixed it the other way :)

esti/migrate_test.go
```diff
@@ -143,6 +144,10 @@ var (
 )
 
 func TestMigrate(t *testing.T) {
+	// skip test if not on postgres
+	if viper.GetViper().GetBool("database_kv_enabled") && viper.GetString("database_type") != kvpg.DriverName {
```
N-o-Z, line 148:
```suggestion
	if viper.GetBool("database_kv_enabled") && viper.GetString("database_type") != kvpg.DriverName {
```

nopcoder (approved):
lgtm

PR #3873, opened 2022-08-10
test: handle esti storage namespace format to handle root scheme

nopcoder:
Format unique storage namespace to handle local://


N-o-Z (approved):
LGTM!

PR #3875, opened 2022-08-10
Prevent lakefs setup and superuser command while using auth API

nopcoder:
While configured to work with auth API, there is no sense in running setup or superuser command as the remote service is responsible of the setup.
The current commands setup a local auth database data which is not relevant while working with the API.


guy-har (approved):
LGTM

PR #3880, opened 2022-08-10
KV Store: Add postgres config section in KV params

N-o-Z:
First step: add postgres section to KV params.

Post KV - remove database section parameters entirely

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -290,10 +290,10 @@ func (c *Config) Validate() error {
 
 func (c *Config) GetDatabaseParams() dbparams.Database {
 	return dbparams.Database{
-		ConnectionString:      c.values.Database.ConnectionString.SecureValue(),
-		MaxOpenConnections:    c.values.Database.MaxOpenConnections,
-		MaxIdleConnections:    c.values.Database.MaxIdleConnections,
-		ConnectionMaxLifetime: c.values.Database.ConnectionMaxLifetime,
+		ConnectionString:      c.values.Database.DeprecatedConnectionString.SecureValue(),
+		MaxOpenConnections:    c.values.Database.DeprecatedMaxOpenConnections,
+		MaxIdleConnections:    c.values.Database.DeprecatedMaxIdleConnections,
+		ConnectionMaxLifetime: c.values.Database.DeprecatedConnectionMaxLifetime,
 		Type:                  c.values.Database.Type,
 		KVEnabled:             c.values.Database.KVEnabled,
 		DropTables:            c.values.Database.DropTables,
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -74,7 +74,7 @@ const (
 	DefaultLakefsEmailBaseURL      = "http://localhost:8000"
 
 	DefaultDynamoDBTableName = "kvstore"
-	// TBD: Which values to use for DynamoDB tables?
+	// TODO (niro): Which values to use for DynamoDB tables?
 	DefaultDynamoDBReadCapacityUnits  = 1000
 	DefaultDynamoDBWriteCapacityUnits = 1000
 )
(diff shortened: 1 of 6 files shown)


pkg/config/template.go
```diff
@@ -73,17 +73,29 @@ type configuration struct {
...
-		MaxOpenConnections    int32         `mapstructure:"max_open_connections"`
-		MaxIdleConnections    int32         `mapstructure:"max_idle_connections"`
+		// Deprecated: use Postgres struct
+		ConnectionString SecureString `mapstructure:"connection_string"`
+		// Deprecated: use Postgres struct
+		MaxOpenConnections int32 `mapstructure:"max_open_connections"`
+		// Deprecated: use Postgres struct
+		MaxIdleConnections int32 `mapstructure:"max_idle_connections"`
+		// Deprecated: use Postgres struct
 		ConnectionMaxLifetime time.Duration `mapstructure:"connection_max_lifetime"`
```
itaiad200, line 83:
```suggestion
		// Deprecated: use Postgres struct
		DeprecatedConnectionString SecureString `mapstructure:"connection_string"`
		// Deprecated: use Postgres struct
		DeprecatedMaxOpenConnections int32 `mapstructure:"max_open_connections"`
		// Deprecated: use Postgres struct
		DeprecatedMaxIdleConnections int32 `mapstructure:"max_idle_connections"`
		// Deprecated: use Postgres struct
		DeprecatedConnectionMaxLifetime time.Duration `mapstructure:"connection_max_lifetime"`
```

pkg/config/config.go
```diff
@@ -301,12 +301,20 @@ func (c *Config) GetDatabaseParams() dbparams.Database {
...
-		},
-		DynamoDB: &kvparams.DynamoDB{
+	}
+	switch {
+	case c.values.Database.Postgres != nil:
+		p.Postgres = &kvparams.Postgres{
+			ConnectionString:      c.values.Database.Postgres.ConnectionString.SecureValue(),
+			MaxIdleConnections:    c.values.Database.Postgres.MaxIdleConnections,
+			MaxOpenConnections:    c.values.Database.Postgres.MaxOpenConnections,
+			ConnectionMaxLifetime: c.values.Database.Postgres.ConnectionMaxLifetime,
```
itaiad200, line 313: Do you wish to drop support for the old values as well and make it a breaking change? Because most users will continue to use them

pkg/config/template.go
```diff
@@ -73,17 +73,29 @@ type configuration struct {
...
+
 		// KVEnabled Development flag to switch between postgres DB and KV store implementations
 		KVEnabled bool `mapstructure:"kv_enabled"`
 		// DropTables Development flag to delete tables after successful migration to KV
 		DropTables bool `mapstructure:"drop_tables"`
 		// Type  Name of the KV Store driver DB implementation which is available according to the kv package Drivers function
 		Type string `mapstructure:"type"`
 
+		Postgres *struct {
+			ConnectionString      SecureString  `mapstructure:"connection_string"`
```
itaiad200, line 93: We also need to modify the documentation for these new values.

pkg/config/config.go
N-o-Z: I don't see a way we keep them in this structure. Once we decided to create sections per driver implementation - it makes no sense to keep these on the root level. I suggest this will be added to the migration procedure. WDYT?

pkg/config/template.go
N-o-Z: https://github.com/treeverse/lakeFS/pull/3881

pkg/config/config.go
itaiad200: I see a way around it. you can do something like:
```suggestion
	connectionStr := c.values.Database.Postgres.ConnectionString.SecureValue()
	if len(connectionStr) == 0 && len(c.values.Database.DeprecatedConnectionString.SecureValue()) == 0 {
		connectionStr = c.values.Database.DeprecatedConnectionString.SecureValue()
		log.Warning("Using deprecated connection string, use DATABASE_POSTGRES_CONNECTION_STRING instead. Support will be dropped soon.")
	}
	p.Postgres.ConnectionString = connectionStr
```

itaiad200 (approved):

itaiad200:
Don't merge this directly to master

N-o-Z:
> Reminder: Don't merge this directly to master for now

@itaiad200 Why not?

PR #3885, opened 2022-08-11
KV: make sure iterators aren't opened to be closed

### LINKED ISSUE #3884: KV Iterators are created twice
When SeekGE is being used after creating the iterator, the first creation of the iterator is meaningless and adds an overhead of a KV list call

itaiad200:
closes #3884

### DIFF
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -764,7 +764,10 @@ func (c *Catalog) ListEntries(ctx context.Context, repository string, reference 
 	it := NewEntryListingIterator(NewValueToEntryIterator(iter), prefixPath, delimiterPath)
 	defer it.Close()
 
-	it.SeekGE(afterPath)
+	if afterPath != "" {
+		it.SeekGE(afterPath)
+	}
+
 	var entries []*DBEntry
 	for it.Next() {
 		v := it.Value()
--- a/pkg/catalog/prefix_iterator.go
+++ b/pkg/catalog/prefix_iterator.go
@@ -13,9 +13,7 @@ type prefixIterator struct {
 }
 
 func NewPrefixIterator(it EntryIterator, prefix Path) EntryIterator {
-	if len(prefix) > 0 {
-		it.SeekGE(prefix)
-	}
+	it.SeekGE(prefix)
 	return &prefixIterator{
 		prefix: prefix.String(),
 		it:     it,
--- a/pkg/catalog/prefix_iterator.go
+++ b/pkg/catalog/prefix_iterator.go
@@ -13,7 +13,9 @@ type prefixIterator struct {
 }
 
 func NewPrefixIterator(it EntryIterator, prefix Path) EntryIterator {
-	it.SeekGE(prefix)
+	if len(prefix) > 0 {
+		it.SeekGE(prefix)
+	}
 	return &prefixIterator{
 		prefix: prefix.String(),
 		it:     it,
--- a/pkg/kv/iterators.go
+++ b/pkg/kv/iterators.go
@@ -253,17 +253,16 @@ type PartitionIterator struct {
 }
 
 func NewPartitionIterator(ctx context.Context, store Store, msgType protoreflect.MessageType, partitionKey string) (*PartitionIterator, error) {
-	itr, err := store.Scan(ctx, []byte(partitionKey), []byte(""))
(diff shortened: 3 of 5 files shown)


pkg/kv/iterators.go
```diff
@@ -253,24 +253,29 @@ type PartitionIterator struct {
 }
 
 func NewPartitionIterator(ctx context.Context, store Store, msgType protoreflect.MessageType, partitionKey string) (*PartitionIterator, error) {
```
N-o-Z, line 255: In that case - so why do we need to return an error?

pkg/kv/iterators.go
```diff
@@ -253,24 +253,29 @@ type PartitionIterator struct {
...
 	return &PartitionIterator{
 		ctx:          ctx,
 		store:        store,
 		msgType:      msgType,
-		itr:          itr,
 		partitionKey: partitionKey,
 	}, nil
 }
 
+func getNewIter(ctx context.Context, store Store, partitionKey string, start []byte) (EntriesIterator, error) {
```
N-o-Z, line 264: getNewEntriesIter?

N-o-Z (changes requested):
Few comments

pkg/catalog/catalog.go
```diff
@@ -764,7 +764,10 @@ func (c *Catalog) ListEntries(ctx context.Context, repository string, reference
 	it := NewEntryListingIterator(NewValueToEntryIterator(iter), prefixPath, delimiterPath)
 	defer it.Close()
 
-	it.SeekGE(afterPath)
+	if afterPath != "" {
+		it.SeekGE(afterPath)
+	}
```
nopcoder, line 769: out of iterator optimization :)

pkg/catalog/prefix_iterator.go
```diff
@@ -13,7 +13,9 @@ type prefixIterator struct {
 }
 
 func NewPrefixIterator(it EntryIterator, prefix Path) EntryIterator {
-	it.SeekGE(prefix)
+	if len(prefix) > 0 {
+		it.SeekGE(prefix)
+	}
```
nopcoder, line 18: Think this one can cause a bug - we assume we accept a fresh 'it' and we do not need to call SeekGE, but we must

pkg/kv/iterators.go
nopcoder, line 264: Not sure we need a function that calls a function - if it was a member to capture the use of: p.{ctx, store, paritionKey} it was a different story.

pkg/kv/iterators.go
```diff
@@ -253,24 +253,29 @@ type PartitionIterator struct {
...
 	if p.Err() != nil {
 		return false
 	}
 	p.value = nil
+	if p.itr == nil {
+		p.itr, p.err = getNewIter(p.ctx, p.store, p.partitionKey, []byte(""))
+		if p.Err() != nil {
+			return false
+		}
+	}
```
nopcoder, line 278: In the case 'p.itr' is nil it means we failed to initialize or seek to a key - we should render an error or return false.

nopcoder (changes requested):

pkg/catalog/prefix_iterator.go
itaiad200: Agreed, fixed on the caller side.

pkg/kv/iterators.go
```diff
@@ -252,25 +252,26 @@ type PartitionIterator struct {
 	err          error
 }
 
-func NewPartitionIterator(ctx context.Context, store Store, msgType protoreflect.MessageType, partitionKey string) (*PartitionIterator, error) {
-	itr, err := store.Scan(ctx, []byte(partitionKey), []byte(""))
-	if err != nil {
-		return nil, err
-	}
+func NewPartitionIterator(ctx context.Context, store Store, msgType protoreflect.MessageType, partitionKey string) *PartitionIterator {
```
N-o-Z, line 255: I think the issue with SeekGE is relevant also for KVOrderedCommitIterator

N-o-Z (approved):
See additional comment

pkg/kv/iterators.go
itaiad200: That's a GC iterator, right? I don't assume it happens that often

PR #3898, opened 2022-08-14
KV Store: Fix setup and DB config init

N-o-Z:
Move dynamodb configuration defaults to driver. Add check for nil configuration on drivers. Fix bug in setup cmd. Add checks for migration


nopcoder (approved):

pkg/kv/postgres/store.go
```diff
@@ -62,16 +62,19 @@ func normalizeDBParams(p *kvparams.Postgres) {
...
 }
 
-func (d *Driver) Open(ctx context.Context, kvparams kvparams.KV) (kv.Store, error) {
+func (d *Driver) Open(ctx context.Context, kvParams kvparams.KV) (kv.Store, error) {
 	// TODO(barak): should we handle Open reuse the same store based on name
-	normalizeDBParams(kvparams.Postgres)
-	config, err := pgxpool.ParseConfig(kvparams.Postgres.ConnectionString)
+	if kvParams.Postgres == nil {
+		return nil, kv.ErrDriverConfiguration
+	}
```
nopcoder, line 69: 1. missing the same for dynamodb
2.
```suggestion
	if kvparams.Postgres == nil {
		return nil, fmt.Errorf("%w: missing %s settings", kv.ErrDriverConfiguration, DriverName)
	}

```

nopcoder (approved):
one comment related to error handling

cmd/lakefs/cmd/run.go
```diff
@@ -131,7 +131,7 @@ var runCmd = &cobra.Command{
 
 			// Check if migration required only on postgres
 			migrationRequired := false
-			if dbParams.Type == kvpg.DriverName {
+			if dbParams.Type == kvpg.DriverName && len(dbParams.ConnectionString) > 0 {
```
itai-david, line 134: In case ConnectionString is missing, what is the desired behavior?

pkg/kv/dynamodb/store.go
```diff
@@ -45,18 +45,40 @@ const (
...
+		p.ReadCapacityUnits = DefaultDynamoDBReadCapacityUnits
+	}
+
+	if p.WriteCapacityUnits == 0 {
+		p.WriteCapacityUnits = DefaultDynamoDBWriteCapacityUnits
+	}
+}
+
 // Open - opens and returns a KV store over DynamoDB. This function creates the DB session
 // and sets up the KV table. dsn is a string with the DynamoDB endpoint
```
itai-david, line 75: `dsn` looks irrelevant

itai-david, line 75:
```suggestion
// and sets up the KV table.
```
`dsn` seems irrelevant

pkg/kv/postgres/store.go
```diff
@@ -62,16 +62,20 @@ func normalizeDBParams(p *kvparams.Postgres) {
 	}
 }
 
-func (d *Driver) Open(ctx context.Context, kvparams kvparams.KV) (kv.Store, error) {
+func (d *Driver) Open(ctx context.Context, kvParams kvparams.KV) (kv.Store, error) {
 	// TODO(barak): should we handle Open reuse the same store based on name
-	normalizeDBParams(kvparams.Postgres)
-	config, err := pgxpool.ParseConfig(kvparams.Postgres.ConnectionString)
+	if kvparams.Postgres == nil {
```
itai-david, line 67:
```suggestion
	if kvParams.Postgres == nil {
```
???

itai-david (commented):
Hi.
Some comments/questions inside. Nothing blocking, as this is already approved
Please take a look at the last one as I think there's a problem there
Thanks

N-o-Z:
> Hi. Some comments/questions inside. Nothing blocking, as this is already approved Please take a look at the last one as I think there's a problem there Thanks

@itai-david  Thanks for the review,
Last night I noticed there were some missing scope for this PR and I started working on it again. I will fix the issues and resubmit for review!

cmd/lakefs/cmd/run.go
N-o-Z: Basically our only indication that we have a previous postgres DB is the connection string. If it doesn't exist we assume migration is either completed or not needed

N-o-Z:
@nopcoder  @itai-david  Please review again.
Found more places to fix + verified manually all lakefs entry point work with only dynamodb configured.
Opened an issue to add testing to the lakefs cmds (https://github.com/treeverse/lakeFS/issues/3910)

cmd/lakefs/cmd/migrate.go
```diff
@@ -56,14 +69,18 @@ var gotoCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		cfg := loadConfig()
 		ctx := cmd.Context()
+		dbParams := cfg.GetDatabaseParams()
+		if dbParams.KVEnabled {
+			fmt.Printf("Unsupported command for KV migration version\n")
+			os.Exit(1)
+		}
```
itai-david, line 76: Why is that unsupported?
I understand that the first version with KV cannot have a goto version, but is it correct for KV in general?

N-o-Z: Exactly what you said - this is currently unsupported. Once we define what it means to perform db migration in the KV world and if we even support the goto command - we will change this accordingly

cmd/lakefs/cmd/setup.go
```diff
@@ -32,14 +32,24 @@ var setupCmd = &cobra.Command{
...
+		if len(dbParams.Type) > 0 && len(dbParams.ConnectionString) > 0 { // Conflicting configuration
+			fmt.Printf("Conflicting database parameters, connection_string should be defined for the specific driver. Do you need to go through migration?\n")
+			os.Exit(1)
+		}
 		var (
 			dbPool   db.Database
 			kvParams params.KV
 			kvStore  kv.Store
 			migrator db.Migrator
+			err      error
```
itai-david, line 44: Not crucial and mainley a matter of personal preference, but I would remove the `err` declaration from here and use the short declaration in both sections below.
For me, as a code reader, declaring a `var` implies that it has some wider relevancy, like the `migrator` that is initialized withinh the `if/else` block and is used afterwards, while the `error` is only used locally and whatever value it gets, is meaningless immediately after the check.
Again, that's my personal opinion :)

N-o-Z: It's an issue since doing that Go treats the internal kvStore as a new parameter in the initilization and passes a nil pointer afterwards

itai-david (approved):
Thanks for the recent changes
Some non-crucial comments within - up to you if to handle or not. Other than that LG

cmd/lakefs/cmd/migrate.go
```diff
@@ -21,6 +23,17 @@ var versionCmd = &cobra.Command{
 		cfg := loadConfig()
 		ctx := cmd.Context()
 		dbParams := cfg.GetDatabaseParams()
+		if dbParams.KVEnabled {
+			kvStore, err := kv.Open(ctx, cfg.GetKVParams())
+			if err != nil {
+				fmt.Printf("Failed to open KV store: %s\n", err)
+				return
+			}
+			if kv.ValidateSchemaVersion(ctx, kvStore, false) == nil {
```
nopcoder, line 32: in case of an error
```go
defer kvStore.Close
```

pkg/kv/dynamodb/store.go
```diff
@@ -45,18 +45,40 @@ const (
...
+	DefaultDynamoDBReadCapacityUnits  = 1000
+	DefaultDynamoDBWriteCapacityUnits = 1000
 )
 
 //nolint:gochecknoinits
 func init() {
 	kv.Register(DriverName, &Driver{})
 }
 
+func normalizeDBParams(p *kvparams.DynamoDB) {
```
nopcoder, line 60: will have conflicts I guess

pkg/kv/migration.go
```diff
@@ -53,10 +56,11 @@ func ValidateSchemaVersion(ctx context.Context, store Store, migrationRequired b
 
 	case kvVersion < InitialMigrateVersion:
 		if migrationRequired {
-			return fmt.Errorf("migration required, for more information see https://docs.lakefs.io/deploying-aws/upgrade.html: %w", err)
+			return fmt.Errorf("migration required, for more information see https://docs.lakefs.io/reference/upgrade.html : %w", ErrInvalidSchemaVersion)
 		} else {
-			return fmt.Errorf("missing KV schema version (%d): %w", kvVersion, err)
+			return fmt.Errorf("(scehma version %d): %w", kvVersion, ErrInvalidSchemaVersion)
```
nopcoder, line 61: can we start the error with something without '('

nopcoder (commented):
Think we can leave the cases we like to educate the user about the previous connection string. We can mark the field as deprected in this release and document it.
In the release we will drop support the configuration loader will fail to load the configuration.

Issue #3906, opened 2022-08-15
Allow to disable serving of embedded UI


eladlachmi:
Currently, lakefs serves UI static resources (`html`, `js`, `css`, etc.) embedded in the binary. This is great for an all-in-one solution and remains the default behavior. 
However, for more advanced use cases, users may want to disable this functionality.
A few example use cases:
- CLI-only access
- API-only access
- In-house developed custom UI
- Integration with self-service developer portals

This is in line with other popular open-source projects. For example, Vault has the UI disabled by default with the option to enable it, if needed. ([reference](https://www.vaultproject.io/docs/configuration/ui#activating-the-vault-ui))
Another example is RabbitMQ, which requires enabling a plugin to get the built-in UI. ([reference](https://www.rabbitmq.com/management.html))

PR #3907, opened 2022-08-15
flag to enable/disable serving of UI

eladlachmi:
Related issue #3906

### DIFF
--- a/pkg/api/serve.go
+++ b/pkg/api/serve.go
@@ -116,7 +116,11 @@ func Serve(
 		r.Mount("/oidc/login", NewOIDCLoginPageHandler(oidcConfig, sessionStore, oauthConfig, logger))
 	}
 	r.Mount("/logout", NewLogoutHandler(sessionStore, logger, cfg.GetAuthLogoutRedirectURL()))
-	r.Mount("/", NewUIHandler(gatewayDomains, snippets))
+
+	// Disable serving of embedded static assets when using an external client
+	if cfg.GetUIEnabled() {
+		r.Mount("/", NewUIHandler(gatewayDomains, snippets))
+	}
 	return r
 }
 
--- a/pkg/api/serve.go
+++ b/pkg/api/serve.go
@@ -117,12 +117,17 @@ func Serve(
 	}
 	r.Mount("/logout", NewLogoutHandler(sessionStore, logger, cfg.GetAuthLogoutRedirectURL()))
 
-	// Disable serving of embedded static assets
+	// Configuration flag to control if the embedded UI is served
+	// or not and assign the correct handler for each case.
 	var rootHandler http.Handler
 	if cfg.GetUIEnabled() {
+		// Handler which serves the embedded UI
+		// as well as handles erroneous S3 gateway requests
+		// and returns a compatible response
 		rootHandler = NewUIHandler(gatewayDomains, snippets)
 	} else {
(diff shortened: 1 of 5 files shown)


pkg/config/template.go
```diff
@@ -224,6 +224,8 @@ type configuration struct {
 		LakefsBaseURL      string        `mapstructure:"lakefs_base_url"`
 	}
 	UI struct {
+		// Enable serving of embedded UI
```
nopcoder, line 227: Just becase code doc in Go requires to start with the name of the field/func
```suggestion
		// Enabled - control serving of embedded UI
```

pkg/config/config.go
```diff
@@ -187,6 +189,8 @@ const (
 	DynamoDBTableNameKey          = "database.dynamodb.table_name"
 	DynamoDBReadCapacityUnitsKey  = "database.dynamodb.read_capacity_units"
 	DynamoDBWriteCapacityUnitsKey = "database.dynamodb.write_capacity_units"
+
+	UIEnabled = "ui.enabled"
```
nopcoder, line 193:
```suggestion
	UIEnabledKey = "ui.enabled"
```

pkg/api/serve.go
```diff
@@ -116,7 +116,11 @@ func Serve(
 		r.Mount("/oidc/login", NewOIDCLoginPageHandler(oidcConfig, sessionStore, oauthConfig, logger))
 	}
 	r.Mount("/logout", NewLogoutHandler(sessionStore, logger, cfg.GetAuthLogoutRedirectURL()))
-	r.Mount("/", NewUIHandler(gatewayDomains, snippets))
+
+	// Disable serving of embedded static assets when using an external client
+	if cfg.GetUIEnabled() {
+		r.Mount("/", NewUIHandler(gatewayDomains, snippets))
+	}
```
nopcoder, line 123: 1. We need to handle the default route in case the request arrives to one of the gateway domains - trace how the call uses `gatewayDomains`
2. Suggestion - configuration parameter 'ui.disable' which will make it easy on the defaults - as for `false` we don't need to add code to set the default and nothing to keep in memory by default.

nopcoder (changes requested):

pkg/config/template.go
eladlachmi: Fixed

pkg/api/serve.go
eladlachmi: 1. Fixed to accommodate this case
2. `ui.enabled` was used for consistency with other config structs

nopcoder (changes requested):
- Added comment about the comment.
- We need to document this flag in the documentation reference.

pkg/api/serve.go
```diff
@@ -116,7 +116,17 @@ func Serve(
 		r.Mount("/oidc/login", NewOIDCLoginPageHandler(oidcConfig, sessionStore, oauthConfig, logger))
 	}
 	r.Mount("/logout", NewLogoutHandler(sessionStore, logger, cfg.GetAuthLogoutRedirectURL()))
-	r.Mount("/", NewUIHandler(gatewayDomains, snippets))
+
+	// Disable serving of embedded static assets
```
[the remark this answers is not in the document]
eladlachmi: Updated to clarify

nopcoder (approved):
lgtm

PR #3911, opened 2022-08-16
KV pg primary key and read complete kv configuration

nopcoder:
- Fix KV store open on 'setup'
- KV params read will pass all configured parameters
- KV Postgres use PRIMARY KEY instead of UNIQUE constraint
- Close kv store after migrate

### DIFF
--- a/pkg/kv/postgres/store.go
+++ b/pkg/kv/postgres/store.go
@@ -63,13 +63,11 @@ func normalizeDBParams(p *kvparams.Postgres) {
 }
 
 func (d *Driver) Open(ctx context.Context, kvParams kvparams.KV) (kv.Store, error) {
-	// TODO(barak): should we handle Open reuse the same store based on name
-	if kvParams.Postgres == nil {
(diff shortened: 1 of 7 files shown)


N-o-Z (approved):
Looks good. You have conflicts on master

pkg/kv/postgres/store.go
```diff
@@ -64,6 +64,9 @@ func normalizeDBParams(p *kvparams.Postgres) {
 
 func (d *Driver) Open(ctx context.Context, kvparams kvparams.KV) (kv.Store, error) {
 	// TODO(barak): should we handle Open reuse the same store based on name
+	if kvparams.Postgres == nil {
+		return nil, fmt.Errorf("%w: missing %s settings", kv.ErrDriverConfiguration, DriverName)
```
itaiad200, line 68:
```suggestion
		return nil, fmt.Errorf("missing %s settings: %w", DriverName, kv.ErrDriverConfiguration)
```

PR #3912, opened 2022-08-16
KV Docs: update upgrade and configuration documentation

N-o-Z:


docs/reference/configuration.md
```diff
@@ -52,10 +52,10 @@ This reference uses `.` to denote the nesting of values.
     + `database.postgres.connection_max_lifetime` `(duration : 5m)` - Sets the maximum amount of time a connection may be reused `(valid units: ns|us|ms|s|m|h)` 
   + `database.dynamodb` - Configuration section when using `database.type="dynamodb"`
     + `database.dynamodb.table_name` `(string : "kvstore")` - Table used to store the data
-    + `database.dynamodb.read_capacity_units` `(int : 1000)` - Throughput parameter for read capacity
-    + `database.dynamodb.write_capacity_units` `(int : 1000)` - Throughput parameter for write capacity
+    + `database.dynamodb.read_capacity_units` `(int : optional)` - Throughput parameter for read capacity
+    + `database.dynamodb.write_capacity_units` `(int : optional)` - Throughput parameter for write capacity
```
itai-david, line 56: Not sure, but isnt the number in parentheses shows the default?

docs/reference/upgrade.md
```diff
@@ -22,12 +22,14 @@ Starting with version <TBD.X.X>, lakeFS has transitioned from using a PostgreSQL
...
 
-> **Note:**  
-> **Users using OS environment variables for database configuration must define the `connection_string` explicitly or as environment variable before proceeding with the migration.**  
-> 
-> Prior to the migration procedure it is strongly recommended to perform the following steps:
-> * Commit all uncommitted data on branches
-> * Create a snapshot of your database
+> **IMPORTANT: Pre Migrate Requirements**  
+> * **Users using OS environment variables for database configuration must define the `connection_string` explicitly or as environment variable before proceeding with the migration.**  
+> * **Database storage free capacity of at least twice the amount of the currently used capacity**
```
itai-david, line 27: This sentence is a bit vague:
Which DB needs the free space? Is it the original DB that needs the extra space?

itai-david: In addition, why is it `twice`?

itai-david (changes requested):
Thanks @N-o-Z 
Please refer to my comment on the `free capactiy` below

docs/reference/configuration.md
N-o-Z: Both default and optional/required you can see examples for it in `auth.encrypt.secret_key` `auth.ldap.server_endpoint` `auth.ldap.bind_dn`...
If you think this is confusing we should change this for all the document

docs/reference/upgrade.md
N-o-Z: Removed redundant.
We don't define the user which DB he should use. The user can either use the current one or create a new one.

itaiad200 (approved):

docs/reference/upgrade.md
```diff
@@ -22,12 +22,14 @@ Starting with version <TBD.X.X>, lakeFS has transitioned from using a PostgreSQL
...
+> * It is strongly recommended to perform these additional steps:
+>   * Commit all uncommitted data on branches
+>   * Create a snapshot of your database
+> * By default, old database tables are not being deleted by the migration process, and should be removed manually after a successful migration.
```
nopcoder, line 31: I don't think we should instruct that these tables should be removed manually after successful migration - maybe supply a script that will do the work or future version should remove them. WDYT?

nopcoder (approved):
Approved, with one question related to drop tables.

docs/reference/configuration.md
itai-david: I agree. If we specify a default-value, it implies that this parameter is optional to the user, right?
I would refrain from using `optional` all together, and just use `required` to show there is no default, and the configuration must include the value, or, the default-value, in case it is optional.

docs/reference/upgrade.md
itai-david: I don't have any better value. I was just wondering if the 2x is really require or it is an overkill. If a user has huge system, with an enormous DB - is requiring an additional 2x (Did I understand it correctly? It is **additional** twice the capacity?) a realistic demand?

docs/reference/upgrade.md
N-o-Z: Providing a script will require them to provide it with the old connection string. I think this could be quite confusing to some users. I would like to hear some more opinions about it

itai-david (approved):

docs/reference/upgrade.md
N-o-Z: It's twice the currently used capacity, not the allocated capacity. For example, if the user is currently uses 2GB out of 10GB, he does not need to allocate more storage.
On the other hand if he uses 9GB out of 10, he will need to allocate 9GB more (assuming that he's going to keep using the server and his data will grow - it makes sense to me no?)

itai-david: So, in that case I would remove the "twice", as it says (at least as I understood it) that the free capacity should be at least twice than the used one. In that case it would imply an additional 17GB to have 18GB free

N-o-Z: Yes - that's what I meant (miscalculation), additional 18GB. If the user has already 9 GB of data it means that in the end of the migration he will have at least 18 GB of data therefore increasing capacity to 19GB makes no sense

PR #3913, opened 2022-08-17
scan with paging for kv over pg

itai-david:
Implementing a configurable paging for scans with KV over PG
Default page size is 1000


itai-david:
Testing the effect of the page size using
```
$ lakectl abuse random-write --amount 100000 --prefix abuse-for-list/ lakefs://load-test/main

$ time lakectl fs ls lakefs://load-test/main/abuse-for-list/ | wc -l 
```

With the following result:
SQL (KV disabled):
```
100000

real	0m3.423s
user	0m1.591s
sys	0m1.155s
```

KV over PG with page size = 1000:
```
100000

real	0m4.145s
user	0m1.595s
sys	0m1.251s
```

KV over PG with page size = 300:
```
100000

real	0m3.984s
user	0m1.601s
sys	0m1.187s


```

KV over PG with page size = 100:
```
100000

real	0m4.038s
user	0m1.615s
sys	0m1.182s
```

KV over PG with page size = 10:
```
100000

real	0m4.662s
user	0m1.631s
sys	0m1.152s
```

KV over PG with page size = 3:
```
100000

real	0m6.532s
user	0m1.753s
sys	0m1.238s
```

pkg/config/config.go
```diff
@@ -311,6 +311,7 @@ func (c *Config) GetKVParams() kvparams.KV {
 			MaxIdleConnections:    c.values.Database.Postgres.MaxIdleConnections,
 			MaxOpenConnections:    c.values.Database.Postgres.MaxOpenConnections,
 			ConnectionMaxLifetime: c.values.Database.Postgres.ConnectionMaxLifetime,
+			ScanPageSize:          c.values.Database.Postgres.ScanPageSize,
```
itaiad200, line 314: I don't think this value needs to be configurable. The user had little understanding of what it does, how to configure it and it's much likelier to make a mistake when using it.

pkg/kv/postgres/store.go
```diff
@@ -41,6 +44,7 @@ const (
 	DefaultMaxOpenConnections    = 25
 	DefaultMaxIdleConnections    = 25
 	DefaultConnectionMaxLifetime = 5 * time.Minute
+	DefaultScanPageSize          = 300
```
itaiad200, line 47: Actually (assuming you tested using a local instance of postgres with 0 network latency), I believe it was tested before with the similar db iterator and 1000 was the optimal value.

itaiad200 (changes requested):
some comments around the value configuration and default value, other than that looks good

pkg/config/config.go
itai-david: What I had in mind is us fine tuning this value in production envs, via this configuration. We can leave it configurable and not document it, having it "hiden" from the common user. WDYT?

pkg/kv/postgres/store.go
itai-david: So, `1000` should be the default?
I'm not really opinionated about it, as my test show really similar results and as you said - it was local and not necessarily representing a production system

pkg/kv/postgres/store.go
```diff
@@ -260,22 +271,32 @@ func (s *Store) Scan(ctx context.Context, partitionKey, start []byte) (kv.Entrie
...
+		rows, err = s.Pool.Query(ctx, `SELECT partition_key,key,value FROM `+s.Params.SanitizedTableName+` WHERE partition_key=$1 AND key `+compareOp+` $2 ORDER BY key LIMIT $3`, partitionKey, start, s.Params.ScanPageSize)
 	}
 
 	if err != nil {
 		return nil, fmt.Errorf("%s: %w (start=%v)", err, kv.ErrOperationFailed, start)
 	}
 	return &EntriesIterator{
-		rows: rows,
+		ctx:   ctx,
+		rows:  rows,
```
itaiad200, line 298: I think the most important part is not fixed here. The connection to postgres is being held by the rows.
Look at `loadBuffer` of `DBIterator` to understand the differences.

pkg/config/config.go
itaiad200: I think this is too internal to be fine-tuned by the user. If anything it's easy to misconfigure this and cause huge latencies. We can always add this later if needed, remove a configuration value is something that is hard to deprecate.

itaiad200 (changes requested):

pkg/config/config.go
itai-david: Done - removed

pkg/kv/postgres/store.go
itai-david: Done - back to 1000

pkg/kv/postgres/store.go
itai-david: Rows are now scanned and "converted" to entries as part of `scanInternal`. This allows `rows.Close`, and the implied connection release, before iterations start.
Hop that what you meant.

Ran the same BM. Results are similar - again, it is a local PG instance, so no real network latency
```
time ./lakectl fs ls lakefs://load-test/main/abuse-for-list/ | wc -l 
100000

real	0m4.028s
user	0m1.633s
sys	0m1.046s
```

pkg/kv/postgres/store.go
```diff
@@ -309,54 +317,38 @@ func (e *EntriesIterator) Next() bool {
...
-		return false
-	}
-	e.entry = &ent
+
 	return true
 }
 
 func (e *EntriesIterator) Entry() *kv.Entry {
-	return e.entry
+	return &e.entries[e.currEntryIdx]
```
itaiad200, line 342: Can this panic? Isn't `nil` a better option?

itaiad200 (approved):
2 minor comments
LGTM

pkg/kv/postgres/store.go
itai-david: Good point - it certainly can.
Fixed

PR #3920, opened 2022-08-17
improve commit log by objects performance

nopcoder:
- lookup object instead searching diff
- cache object lookup per commit while searching the log


guy-har (approved):
LGTM

PR #3949, opened 2022-08-22
enforce link between prs to issues

talSofer:


johnnyaug (approved):
LGTM. Will this be re-rerun when the link is added post failure?

talSofer:
> LGTM. Will this be re-rerun when the link is added post failure?

@johnnyaug No, we should re-run manually like we do with enforce label. Makes sense?

PR #3965, opened 2022-08-22
Prefetch and parallelize commits compare

### LINKED ISSUE #3964: Prefetch commits in Graveler ListCommits
ListCommits is capable of scanning thousands of commits, but it does so serially.
It's better to prefetch the commit, metarange and range to parallelize the download from S3 which becomes the bottleneck.

itaiad200:
closes #3964


guy-har (approved):
Looks good
Thanks!

PR #3878, opened 2022-08-10
Docs: Add KV documentation

N-o-Z:
Modified documentation:

- upgrade doc
- sizing guide DB storage and Mem **only**

PR #3968, opened 2022-08-23
Docs/lakefs setup with dynamodb

### LINKED ISSUE #3895: KV Docs: lakeFS setup with DynamoDB
- Docs base PR: https://github.com/treeverse/lakeFS/pull/3878

Create a PR over the base PR and add page on KV setup with DynamoDB

itai-david:
Closes #3895 

First go - Updated lakeFS deployement on AWS with DynamoDB

### DIFF
--- a/docs/deploy/aws.md
+++ b/docs/deploy/aws.md
@@ -21,12 +21,7 @@ Expected deployment time: 25 min
 
 ## Preparing the Database for the Key Value Store
 lakeFS uses a key-value store to synchronize actions on your repositories. Out of the box, this key-value store can rely on DynamoDB or a PostgreSQL DB. As lakeFS open source, you can also write your own implementation and use any other DB
-The following two sections explain how to setup either DynamoDB or PostgreSQL as key-value backing DB
-
-### DynamoDB on AWS
-DynamoDB on AWS does not require any specific preparation, other than properly configuring lakeFS to use it and valid AWS credentials. Please refer to `database.dynamodb ` section in the [configuration reference](../reference/configuration.md#reference) for complete configuration options.
-AWS credentials, for DynamoDB, can also be provided via environment variables, as described in the [configuration reference](../reference/configuration.md#using-environment-variables)
-Please refer to [AWS documentation](https://aws.amazon.com/dynamodb/getting-started/) for further information on DynamoDB
+The following two sections explain how to setup either PostgreSQL or DynamoDB as key-value backing DB
 
 ### Creating PostgreSQL Database on AWS RDS
 We will show you how to create a database on AWS RDS but you can use any PostgreSQL database as long as it's accessible by your lakeFS installation.
@@ -41,29 +36,14 @@ If you already have a database, take note of the connection string and skip to t
 
 3. Make sure your security group rules allow you to connect to the database instance.
 
+### DynamoDB on AWS
+DynamoDB on AWS does not require any specific preparation, other than properly configuring lakeFS to use it and valid AWS credentials. Please refer to `database.dynamodb ` section in the [configuration reference](../reference/configuration.md#reference) for complete configuration options.
+AWS credentials, for DynamoDB, can also be provided via environment variables, as described in the [configuration reference](../reference/configuration.md#using-environment-variables)
+Please refer to [AWS documentation](https://aws.amazon.com/dynamodb/getting-started/) for further information on DynamoDB
+
 ## Installation Options
 
 ### On EC2
-#### With DynamoDB
-1. Save the following configuration file as `config.yaml`:
-
-   ```yaml
-   ---
-   database:
-     type: "dynamodb"
-     dynamodb: 
-       table_name: "[DYNAMODB_TABLE_NAME]"
-       aws_region: "[DYNAMODB_AWS_REGION]"
-       aws_access_key_id: "[AWS_ACCESS_KEY_ID]"
-       aws_secret_access_key: "[AWS_SECRET_ACCESS_KEY]"
-   auth:
-     encrypt:
-       # replace this with a randomly-generated string:
-       secret_key: "[ENCRYPTION_SECRET_KEY]"
-   blockstore:
-     type: s3
-     s3:
-       region: us-east-1 # optional, fallback in case discover from bucket is not supported
 
 #### With PostgreSQL
 1. Save the following configuration file as `config.yaml`:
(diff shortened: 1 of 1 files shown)


itai-david:
closes #3895

docs/deploy/aws.md
```diff
@@ -19,8 +19,16 @@ Expected deployment time: 25 min
 
 {% include_relative includes/prerequisites.md %}
 
-## Creating the Database on AWS RDS
-lakeFS requires a PostgreSQL database to synchronize actions on your repositories.
+## Preparing the Database for the Key Value Store
+lakeFS uses a key-value store to synchronize actions on your repositories. Out of the box, this key-value store can rely on DynamoDB or a PostgreSQL DB. As lakeFS open source, you can also write your own implementation and use any other DB
+The following two sections explain how to setup either DynamoDB or PostgreSQL as key-value backing DB
+
+### DynamoDB on AWS
```
itaiad200, line 26: I think we should put this second as it is less battle proven than postgres.

docs/deploy/aws.md
```diff
@@ -36,12 +44,36 @@ If you already have a database, take note of the connection string and skip to t
...
    ```yaml
    ---
    database:
-     connection_string: "[DATABASE_CONNECTION_STRING]"
+     type: "dynamodb"
+     dynamodb: 
+       table_name: "[DYNAMODB_TABLE_NAME]"
+       aws_region: "[DYNAMODB_AWS_REGION]"
+       aws_access_key_id: "[AWS_ACCESS_KEY_ID]"
+       aws_secret_access_key: "[AWS_SECRET_ACCESS_KEY]"
```
itaiad200, line 58: We don't need these 3, just like you don't provide them for the S3 blockstore.

docs/deploy/aws.md
```diff
@@ -60,14 +92,30 @@ If you already have a database, take note of the connection string and skip to t
...
-command to demonstrate starting lakeFS using Docker:
-
+To support container-based environments like AWS ECS, lakeFS can be configured using environment variables. Here are a couple of `docker run` 
+commands to demonstrate starting lakeFS using Docker:
+#### With PostgreSQL
+```sh
+docker run \
+  --name lakefs \
+  -p 8000:8000 \
+  -e LAKEFS_DATABASE_KV_ENABLED="true"
```
itaiad200, line 102: This won't be needed by the time it's merged.

docs/deploy/aws.md
```diff
@@ -60,14 +92,30 @@ If you already have a database, take note of the connection string and skip to t
...
+  -e LAKEFS_DATABASE_TYPE: "dynamodb" \
+  -e LAKEFS_DATABASE_DYNAMODB_AWS_ACCESS_KEY_ID="[AWS_ACCESS_KEY_ID]" \
+  -e LAKEFS_DATABASE_DYNAMODB_AWS_SECRET_ACCESS_KEY="[AWS_SECRET_ACCESS_KEY]" \
+  -e LAKEFS_DATABASE_DYNAMODB_AWS_REGION="[DYNAMODB_AWS_REGION]" \
```
itaiad200, line 106: Same here.

docs/deploy/aws.md
```diff
@@ -60,14 +92,30 @@ If you already have a database, take note of the connection string and skip to t
...
-
+To support container-based environments like AWS ECS, lakeFS can be configured using environment variables. Here are a couple of `docker run` 
+commands to demonstrate starting lakeFS using Docker:
+#### With PostgreSQL
+```sh
+docker run \
+  --name lakefs \
+  -p 8000:8000 \
+  -e LAKEFS_DATABASE_KV_ENABLED="true"
+  -e LAKEFS_DATABASE_TYPE: "dynamodb" \
```
itaiad200, line 103: Where's the table name? That one is actually important.

docs/deploy/aws.md
```diff
@@ -60,14 +92,30 @@ If you already have a database, take note of the connection string and skip to t
...
+  -e LAKEFS_BLOCKSTORE_TYPE="s3" \
+  treeverse/lakefs:latest run
+```
+#### With PostgreSQL
 ```sh
 docker run \
   --name lakefs \
   -p 8000:8000 \
-  -e LAKEFS_DATABASE_CONNECTION_STRING="[DATABASE_CONNECTION_STRING]" \
+  -e LAKEFS_DATABASE_KV_ENABLED="true"
```
itaiad200, line 116: Unnecessary

docs/reference/database-migration.md
```diff
@@ -0,0 +1,17 @@
...
+has_children: false
+---
+
+# lakeFS Database Migrate
+{: .no_toc }
+
+**Note:** Feature in development
+{: .note }
+
+The lakeFS database migration tool simplifies switching from one database implementation to another.
```
itaiad200, line 16: I would add an example to make it clear.
```suggestion
The lakeFS database migration tool simplifies switching from one database implementation to another, e.g. from Postgres to DynamoDB.
```

itaiad200 (changes requested):
See comments inline. I don't think we should merge it to master, it also includes many irrelevant commits.

N-o-Z (changes requested):
@itai-david As described in the original issue (https://github.com/treeverse/lakeFS/issues/3895)
This PR should be to https://github.com/treeverse/lakeFS/pull/3878 and not master

docs/deploy/aws.md
itai-david: Makes total sense. Also switched the order of the configuration/docker snippets to reflect the same order

docs/deploy/aws.md
itai-david: Removed

docs/deploy/aws.md
itai-david: Correct - removed

docs/deploy/aws.md
itai-david: Added.
Although, we do have a default table name, which is workable

docs/deploy/aws.md
itai-david: Removed

docs/deploy/aws.md
itai-david: Removed

docs/reference/database-migration.md
itai-david: This one is coming from the migration PR. It is no longer part of this PR as I changed the base

itai-david:
> See comments inline. I don't think we should merge it to master, it also includes many irrelevant commits.

@itaiad - merge target branch modified to reflect the original base. Thanks for pointing this out

docs/deploy/aws.md
```diff
@@ -19,8 +19,11 @@ Expected deployment time: 25 min
 
 {% include_relative includes/prerequisites.md %}
 
-## Creating the Database on AWS RDS
-lakeFS requires a PostgreSQL database to synchronize actions on your repositories.
+## Preparing the Database for the Key Value Store
+lakeFS uses a key-value store to synchronize actions on your repositories. Out of the box, this key-value store can rely on DynamoDB or a PostgreSQL DB. As lakeFS open source, you can also write your own implementation and use any other DB
```
N-o-Z, line 23: Maybe generalize it more and not mention the current supported drivers explicitly, so we won't have to maintain this documentation everytime we add a new driver

N-o-Z (approved):
Looks good! One suggestion

docs/deploy/aws.md
itai-david: I guess that if we do add other drivers, we will need to update this doc one way or another, and at least add another configuration section, similar to the 2 we have below.
I think it is best to be as explicit as possible with manuals, so I would vote to leave the specification and expand it in the future (if needed). WDYT?

itaiad200 (approved):

PR #3971, opened 2022-08-23
log commits with optional limit

### LINKED ISSUE #3970: Looking for single commit log record performance
Request for commit log can take a log time when searching for a single change.
Searching for a single commit log by passing object and amount=1 can take long time as the underlaying API
will continue searching for a second commit for 'has more' indication.

nopcoder:
In cases that 'has more' indicator is not required, passing limit=true to log commit request will enable faster response time.
Cases for searching for a objects or prefix - limit the search to the 'amount' that requested will perform faster and no additional search will be done for 'has more' indication.

Fix #3970

### DIFF
--- a/pkg/api/controller_test.go
+++ b/pkg/api/controller_test.go
@@ -21,8 +21,6 @@ import (
 	"text/template"
 	"time"
 
-	"github.com/go-openapi/swag"
-
 	"github.com/go-test/deep"
 	nanoid "github.com/matoous/go-nanoid/v2"
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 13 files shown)


pkg/api/controller_test.go
```diff
@@ -21,6 +21,8 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/go-openapi/swag"
+
```
itaiad200, line 25: empty line

itaiad200 (approved):
Looks great, few phrasing comments

nopcoder:
> Looks great, few phrasing comments

Thanks! apply and fixed the code.

PR #3974, opened 2022-08-23
Update CHANGELOG before release

nopcoder:


guy-har (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.70.4 - 2022-08-23
+
+What's new:
+- Improve commit log performance for single match by adding limit flag (#3970)
```
arielshaqed, line 6: This adds a flag, so requires a minor version bump.

(Still **NOT** a semver violation, because we're pre-1.0.0, and all the places behave as though shifted by one place...).

arielshaqed (commented):
Not major, but maybe worth an announcement.  In order to be able to use this, we need to release a 0.71.0 as it adds a flag 🎏 to the Swagger API.

Issue #3989, opened 2022-08-25
GC: improve API client initialization readability 


talSofer:
Address the review comments in:
* https://github.com/treeverse/lakeFS/pull/3983#discussion_r954521057
* https://github.com/treeverse/lakeFS/pull/3983#discussion_r954523403
* https://github.com/treeverse/lakeFS/pull/3983#discussion_r954522220
* https://github.com/treeverse/lakeFS/pull/3983#discussion_r954638686
* https://github.com/treeverse/lakeFS/pull/3998#discussion_r956664204

PR #3998, opened 2022-08-25
re-use http client in garbage collection to reduce number of open connections

### LINKED ISSUE #3995: Re-use ApiClient in metadata client
The metadata client creates a new ApiClient for almost every call.  This actually triggers a _worst_ practice: [OkHttpClient.kt says](https://github.com/square/okhttp/blob/b1ad9fcf892c9bdfa8714f595b1b2e7c15314c67/okhttp/src/jvmMain/kotlin/okhttp3/OkHttpClient.kt#L52-L59):
```kotlin
/*
 * ## OkHttpClients Should Be Shared
 *
 * OkHttp performs best when you create a single `OkHttpClient` instance and reuse it for all of
 * your HTTP calls. This is because each client holds its own connection pool and thread pools.
 * Reusing connections and threads reduces latency and saves memory. Conversely, creating a client
 * for each request wastes resources on idle pools.
 *
 * Use `new OkHttpClient()` to create a shared instance with the default settings:
 */
```

The ApiClient **generated code** actually [creates a new client builder](https://cs.github.com/treeverse/lakeFS/blob/1ca7676d07d49ae75607e10e0628a4683f99bf71/clients/java/src/main/java/io/lakefs/clients/api/ApiClient.java#L125) every time we construct it:
```java
        OkHttpClient.Builder builder = new OkHttpClient.Builder();
        builder.addNetworkInterceptor(getProgressInterceptor());
```

This can exhaust file descriptors on the lakeFS server.  So don't do that :-)

talSofer:
Fixes #3995

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
@@ -91,6 +91,7 @@ class ApiClient(
     readTimeoutSec: String = ""
 ) {
   val client = ApiClient.getApiClient(apiUrl, accessKey, secretKey, connectionTimeoutSec, readTimeoutSec)
+
   private val repositoriesApi = new api.RepositoriesApi(client)
   private val commitsApi = new api.CommitsApi(client)
   private val metadataApi = new api.MetadataApi(client)
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
@@ -1,6 +1,5 @@
 package io.treeverse.clients
 
-import com.google.common.cache.{CacheBuilder, CacheLoader, Cache, LoadingCache}
 import io.lakefs.clients.api
 import io.lakefs.clients.api.{ConfigApi, RetentionApi}
 import io.lakefs.clients.api.model.{
@@ -9,6 +8,8 @@ import io.lakefs.clients.api.model.{
 }
 import io.treeverse.clients.StorageClientType.StorageClientType
 import io.treeverse.clients.StorageUtils.{StorageTypeAzure, StorageTypeS3}
+
+import com.google.common.cache.{Cache, CacheBuilder, CacheLoader, LoadingCache}
 
 import java.net.URI
 import java.util.concurrent.{Callable, TimeUnit}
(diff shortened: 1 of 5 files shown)


johnnyaug (approved):
Nice!

Jonathan-Rosenberg (dismissed):
Awesome!
Turns out that design principles are still valid today!
We should actually validate that dependency injection is implemented all around the scala/spark/java area if you ask me...

arielshaqed:
> Awesome! Turns out that design principles are still valid today! We should actually validate that dependency injection is implemented all around the scala/spark/java area if you ask me...

Actually this disappoints me: HTTP clients are supposed to re-use connections for you.  OkHttp is different, and the generated ApiClient code doesn't realize this.

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -22,6 +22,41 @@ object StorageClientType extends Enumeration {
...
+      case Some(client) => client
+      case None => {
+        val FROM_SEC_TO_MILLISEC = 1000
+
+        val client = new api.ApiClient
+        client.setUsername(accessKey)
+        client.setPassword(secretKey)
+        client.setBasePath(apiUrl.stripSuffix("/"))
+        if (connectionTimeoutSec != null && !connectionTimeoutSec.isEmpty) {
+          val connectionTimeoutMillisec = connectionTimeoutSec.toInt * FROM_SEC_TO_MILLISEC
```
arielshaqed, line 46: (Not introduced in this PR, and to fix later:)  This might have a nasty failure mode.  When the configuration option is malformatted, the `String.toInt` call here is (probably) the first time that we discover it.  So this configuration error is reported as a stack trace from somewhere inside the driver, rather than during startup.  It will look a bit like this:
```
  at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
  at java.base/java.lang.Integer.parseInt(Integer.java:652)
  at java.base/java.lang.Integer.parseInt(Integer.java:770)
[...]
```
The error message will not say which configuration parameter was wrong -- read or connect timeouts; the user will have to open this file to determine that.

So flags should be parsed during startup.  I think we really want to fix this because it could be painful for users.

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -22,6 +22,41 @@ object StorageClientType extends Enumeration {
 
 private object ApiClient {
 
+  val clients = collection.mutable.Map[String, api.ApiClient]()
```
arielshaqed, line 25:
```suggestion
  // TODO(talsofer): Replace with LoadingCache.
  val clients = collection.mutable.Map[String, api.ApiClient]()
```
or
```suggestion
  // TODO(ariels): Replace with LoadingCache.
  val clients = collection.mutable.Map[String, api.ApiClient]()
```

arielshaqed (approved):
Looks great, thanks!  Nothing for this round, small request that is anyway planned for next PR.

I am approving because we pair programmed on the small bit that I added.  But if we have time, I'd rather someone else on the team also went over it and approved.

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
talSofer: Thanks! I added this to https://github.com/treeverse/lakeFS/issues/3989

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
talSofer: done

johnnyaug (approved):
Yessss!

LGTM

PR #4001, opened 2022-08-26
Updating failure due to migration message with link to docs

itai-david:


pkg/kv/migration.go
```diff
@@ -45,7 +45,9 @@ func ValidateSchemaVersion(ctx context.Context, store Store, migrationRequired b
 	switch {
 	case errors.Is(err, ErrNotFound):
 		if migrationRequired {
-			return fmt.Errorf("missing KV schema version: %w", err)
+			// TODO: We can update this message to point straight to the KV migration section, once the TBD is removed.
```
N-o-Z, line 48: I think we can simplify the code here:
```
	kvVersion, err := GetDBSchemaVersion(ctx, store)
	if errors.Is(err, ErrNotFound) && !migrationRequired {
		logging.Default().Info("No KV Schema version, setup required")
		return nil
	} else if err != nil {
		return fmt.Errorf("get KV schema version: %w", err)
	}

	if kvVersion < InitialMigrateVersion {
		if migrationRequired {
			return fmt.Errorf("migration required, for more information see https://docs.lakefs.io/reference/upgrade.html : %w", ErrInvalidSchemaVersion)
		} else {
			return fmt.Errorf("(scehma version %d): %w", kvVersion, ErrInvalidSchemaVersion)
		}
	}

	logging.Default().WithField("version", kvVersion).Info("KV valid")
	return nil
}
```

N-o-Z (approved):
Approved with a code improvement suggestion

pkg/kv/migration.go
nopcoder: I even simplify it by dropping the `else` x2

itaiad200:
Let's merge this to master and not the `docs/lakefs-kv-migration`.

itai-david:
> Let's merge this to master and not the `docs/lakefs-kv-migration`.

Yeah... that would be problematic as it branched from the `docs/lakefs-kv-migration` branch and contains some of its modifications. I will create a new PR on top of `master` and close this one

itai-david:
See #4074 instead

PR #4006, opened 2022-08-28
KV DynamoDB metrics

nopcoder:


pkg/kv/dynamodb/store.go
```diff
@@ -195,12 +195,15 @@ func (s *Store) Get(ctx context.Context, partitionKey, key []byte) (*kv.ValueWit
 		return nil, kv.ErrMissingKey
 	}
 
+	start := time.Now()
 	result, err := s.svc.GetItemWithContext(ctx, &dynamodb.GetItemInput{
 		TableName:      aws.String(s.params.TableName),
 		Key:            s.bytesKeyToDynamoKey(partitionKey, key),
 		ConsistentRead: aws.Bool(true),
 	})
+	dynamoRequestDuration.WithLabelValues("GetItem").Observe(time.Since(start).Seconds())
```
itaiad200, line 204: Why is the consumed capacity not counted for in this flow?

itaiad200 (approved):
LGTM
Just a single question

pkg/kv/dynamodb/store.go
nopcoder: Missed it - thanks!

itaiad200 (approved):

pkg/kv/dynamodb/store.go
```diff
@@ -351,6 +382,9 @@ func (s *Store) DropTable() error {
 	_, err := s.svc.DeleteTable(&dynamodb.DeleteTableInput{
 		TableName: &s.params.TableName,
 	})
+	if err != nil {
+		dynamoFailures.WithLabelValues("DeleteTable").Inc()
```
N-o-Z, line 386: This is an internal function for testing. Not sure we need to collect metrics on it.

N-o-Z (commented):
@nopcoder Thanks!
Do we need to add documentation for it?

pkg/kv/dynamodb/store.go
nopcoder: true, will remove it

nopcoder:
> @nopcoder Thanks!
> Do we need to add documentation for it?

good point, ill add pr for the documentation

PR #4009, opened 2022-08-28
Metadata client: Keep (generated) API clients in a Guava Cache

### LINKED ISSUE #4007: Use Guava (Google) cache in metadata client api.ApiClient cache
In #3998 we cache api.ApiClients.  We do this using a regular mutable Map.  Use a Guava cache anyway (it's already used elsewhere in the code).

arielshaqed:
This uses a regular Guava (Google Commons) cache to hold api.ApiClients.
Using LoadingCache here would force us to hold the secret key as part of the
cache key, which is both pointless and risks further spreading secret
material.

(I did this in #3998 because of working too quickly.  This is just a minor refactoring of that PR.)

Fixes #4007.

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
@@ -38,24 +38,27 @@ private object ApiClient {
       secretKey: String,
       connectionTimeoutSec: String = "",
       readTimeoutSec: String = ""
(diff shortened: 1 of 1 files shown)


arielshaqed:
Reviewer tip: "Hide whitespace" in the settings box to see how few changes there really are!

talSofer (approved):
Thanks!

arielshaqed:
Thanks!  Pushing to reuse-http-client, continuing there on next issue.

PR #4011, opened 2022-08-28
Cache top-level ApiClient not low-level generated api.ApiClient

arielshaqed:
The storage namespace cache is part of the top-level ApiClient.  If we share
only the _generated_ api.ApiClient then the storage namespace cache is much
less useful.


talSofer (approved):
LGTM!

talSofer:
I expect this change to reduce the number of connections even more

arielshaqed:
Thanks!

Not sure it will reduce # of connections (we were already sharing HTTP clients by base URL).  But it *will* reduce the number of calls to GetRepository and to GetStorageNamespace by ~100x or so.

PR #4015, opened 2022-08-28
Proposal: lakeFS hard-delete

### LINKED ISSUE #1933: Hard-delete objects that were never committed
## User story
In general, objects should be removed from the storage when there are no valid pointers from lakeFS to them.
In particular, objects that were removed from the staging area without ever being committed should be hard-deleted from the underlying storage.

## Requirements

In case an object exists only in the staging area of a branch (does not belong to any commit) - hard-delete it immediately when it has no lakeFS entries that use it as a physical address. This may happen when:

* The object is deleted in lakeFS.
* The object is overwritten in lakeFS.
* All uncommitted changes on the branch are reverted.
* The branch is deleted.
* The repository is deleted.

_Important:_ the implementation should take into account that lakeFS sometimes uses the same object for multiple entries (AKA deduplication).

## Proposal
The requirements can be kept by the following:
* When removing staged data (by delete or by revert) delete the data from underlying storage as well (*this could be done only if deduplication works only on committed data)
* When removing a branch delete all staged data from underlying storage.
* TBD - handle Objects that where overwritten within staging

N-o-Z:
Closes #1933 (but not really...)

Design proposals document for in-house hard-delete in lakeFS

The


N-o-Z:
More or less hit a brick wall with this 😅 
Please see the summary for all the proposal attempts, please share if you have any ideas

design/open/hard-delete.md
```diff
@@ -0,0 +1,202 @@
...
+
+A background job will be responsible for scanning the references prefix, and handling the references according to the state.
+
+
+### Background delete job (pseudo-code)
+
+1. Scan `reference` prefix (can use after prefix for load balancing)
+2. For each `physical_address` read all entries
+      1. If found state == 'committed' in any entry
+         1. Delete all keys for `phisycal_address`, by order of state: deleted -> staged -> committed
```
arielshaqed, line 154: This seems worrying:
1. I commit an object at physical path `abcd` during staging stg1.  So I have `committed` at `references/abcd/stg1`.
1. Concurrently:
   * Background delete runs and deletes this entry (the intent is to keep `abcd`, it was committed).
   * A copy operation stages `abcd` into staging stg2
   After both of these, I can have `staged` at `references/abcd/stg2` and nothing (no key!) at `references/abcd/stg1`.
1. Delete this object.  Now `references/abcd/stg2` is `deleted`.
1. The next background delete deletes `abcd`, despite it being referenced by a commit.

arielshaqed (changes requested):
I think we need some more detail on the precise sequence of staging.  I think _delete object_ is actually the hardest operation to get right.

design/open/hard-delete.md
N-o-Z: > Background delete runs and deletes this entry (the intent is to keep abcd, it was committed).
This will not happen since we will not delete references which have at least one entry with state `committed`

Additionally, something that was discussed and was omitted from the design (will add it), is that the background delete operation is time based. i.e. a candidate for delete is a physical address which all of its references are in state `deleted` and the last modified timestamp is bigger than 'x' (for example a day).
This ensures us that this physical address is no longer being in use or have a chance to be referenced again.

design/open/hard-delete.md
```diff
@@ -0,0 +1,202 @@
...
+
+Staging manager will track physical addresses of staged objects and the staging tokens referencing them.
+We will introduce a new object to the database:
+
+      key    = <repo>/references/<physical_address>/<staging_token>  
+      value  = state enum (staged|deleted|comitted)
+
+On Upload object to a branch, we will add a key in the references path with the physical address and current staging token and 
+mark it as staged.   
+On Delete object from a staging area, we will update the reference key with value `deleted`
```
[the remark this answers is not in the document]
N-o-Z: I think the reference counting is inferred with this solution, it is an aggregation of all the references of a physical address which are not in state `deleted`

design/open/hard-delete.md
```diff
@@ -0,0 +1,202 @@
...
+### Blockers
+
+1. Commit flow is not atomic:
+   1. Start write on staged object
+   2. Start Commit
+   3. Commit flow writes current object to committed tables
+   4. Write to object is completed - changing the data after it was already committed
+
+### Opens
+1. Solve the blocker - how to prevent data modification after commit?
```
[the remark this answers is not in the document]
N-o-Z: I think that's just a different perspective to the same problem 😄

design/open/hard-delete.md
```diff
@@ -0,0 +1,202 @@
...
+   user or by a dedicated GC process
+
+#### Delete Object
+
+1. Read object's metadata
+2. Reference counter exists:
+   1. if counter == 1
+      1. Hard-delete object
+   2. if counter > 1
+      1. Decrement counter
```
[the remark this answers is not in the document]
N-o-Z: True - another option:

1. Decrement counter
2. if counter == 0 -> hard delete

In all other flows - read counter first, if counter < 1 - treat file as deleted. 
WDYT?

design/open/hard-delete.md
```diff
@@ -0,0 +1,202 @@
...
+
+Several approaches were considered and described in this document. All proposal suffer from unresolved design issues, and
+so no specific approach was selected as a solution.
+
+### Base assumptions
+
+The following assumptions are required in all the suggested proposals:
+1. Copy operation requires to be changed and implemented is full copy.
+2. lakeFS is not responsible for retention of data outside the repository path (data ingest)
+3. An online solution is not bulletproof, and will require a complementary external solution to handle edge cases
```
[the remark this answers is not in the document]
N-o-Z: Added

design/open/hard-delete.md
```diff
@@ -0,0 +1,202 @@
...
+As result, uncommitted objects are all objects on a path of staging tokens which were not yet committed.  
+When deleting a branch, each object under the branch's current staging token can be deleted.
+
+### Blockers
+
+1. Commit flow is not atomic:
+   1. Start write on staged object
+   2. Start Commit
+   3. Commit flow writes current object to committed tables
+   4. Write to object is completed - changing the data after it was already committed
```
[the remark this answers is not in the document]
N-o-Z: The problem is that the set operation will happen regardless if the upload was succeeded the first time. It means that we write to the staging token (in this case - overwrite the object on the physical address). When the writer discovers the staging token has changed, we simply retry the operation with the new staging token. The new data will be written to both the old token and the new token. This is something we couldn't quite find a way to deal with in the new KV design.

design/open/hard-delete.md
```diff
@@ -0,0 +1,202 @@
...
+Due to the basic concept of handling the staging data in the objects store, this proposal suffers from a significant performance 
+degradation in one of lakeFS's principal flows - **listing**. Listing is used in flows such as: list branch objects, getting objects and committing.
+We tested listing of ~240000 **staged objects** using `lakeFS ls` and comparing it with `aws s3 ls` with the following results:
+> ./lakectl fs ls lakefs://niro-bench/main/guy-repo/itai-export/medium/ >   6.42s user 0.78s system 44% cpu 16.089 total
+
+> aws s3 ls s3://guy-repo/itai-export/medium/ > /dev/null  56.82s user 2.34s system 74% cpu 1:19.79 total
+
+### Design
+
+Objects will be stored in a path relative to the branch and staging token. For example: file `x/y.z` will be uploaded to path
```
eden-ohana, line 39: Maybe it's possible to save it on the KV store instead under the same prefix, it can help make the flow commit atomic

design/open/hard-delete.md
```diff
@@ -0,0 +1,202 @@
...
+
+### Background delete job (pseudo-code)
+
+1. Scan `reference` prefix (can use after prefix for load balancing)
+2. For each `physical_address` read all entries
+      1. If found state == 'committed' in any entry
+         1. Delete all keys for `phisycal_address`, by order of state: deleted -> staged -> committed
+      2. If state == 'deleted' in all entries
+         1. Hard-delete object
+         2. Delete all keys for `phisycal_address`
```
itaiad200, line 157:
```suggestion
         2. Delete all keys for `physical_address`
```

design/open/hard-delete.md
```diff
@@ -0,0 +1,202 @@
...
+Store the reference counter in the blob's metadata and use `set blob metadata` API with the supported conditional header on the ETag to perform an If-Match operation and update
+the reference counter
+
+
+#### Google Store
+
+Store the reference counter in the blob's metadata, use the `patch object` API with the `ifMetagenerationMatch` conditional.
+
+
+## 3. Tracking references in staging manager
```
itaiad200, line 133: Why is base assumption 1 valid here? I don't think we need to change the way we copy in this solution.

design/open/hard-delete.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Motivation
+
+Uncommitted data which is no longer referenced (due to branch deletion, reset branch etc.) is not being deleted by lakeFS.
+This may result excessive storage usage and possible compliance issues.
+To solve this two approaches were suggested:
+1. A batch operation performed as part of an external process (GC)
+2. An online solution inside lakeFS
+
+**This document details the latter.**
```
itai-david, line 11: I might be missing some context here, but what are the reason to not consider an offline GC?

design/open/hard-delete.md
```diff
@@ -0,0 +1,202 @@
...
+         1. Hard-delete object
+         2. Delete all keys for `phisycal_address`
+
+### Key Update Order of Precedence
+
+1. `commited` state takes precedence over all (terminal state - overrides any other value) - uses **Set**
+2. `deleted` state can only be done on entries which are not `committed` and uses **SetIf**
+3. `staged` state can only be done on entries which are not `committed` and uses **SetIf** 
+
+### Flows
```
itai-david, line 165: How about branch deletion, when committed objects are left unreferenced?
Does the ref count cover it too?

itai-david (commented):
Great coverage of these options and mainley, and most importantly, of their imperfections.
Some questions inside

design/open/hard-delete.md
N-o-Z: Offline GC is also considered. We want to exhaust the options of implementing an online solution.
This document discusses only the online solutions

design/open/hard-delete.md
N-o-Z: Saving it in the KV store will require changing the way we manage uncommitted data. Currently entries are saved per staging token. What you suggest requires saving them per physical address. This will break the current staging flows, and will incur a lot of overhead when reading from staging area

design/open/hard-delete.md
N-o-Z: Yes, on branch deletion, and as part of the staging token drop, we scan over token entries,
if entry was tombstone - we delete the reference entry, otherwise we modify the reference state to deleted

N-o-Z: This is the same as reset branch

design/open/hard-delete.md
```diff
@@ -0,0 +1,202 @@
...
+degradation in one of lakeFS's principal flows - **listing**. Listing is used in flows such as: list branch objects, getting objects and committing.
+We tested listing of ~240000 **staged objects** using `lakeFS ls` and comparing it with `aws s3 ls` with the following results:
+> ./lakectl fs ls lakefs://niro-bench/main/guy-repo/itai-export/medium/ >   6.42s user 0.78s system 44% cpu 16.089 total
+
+> aws s3 ls s3://guy-repo/itai-export/medium/ > /dev/null  56.82s user 2.34s system 74% cpu 1:19.79 total
+
+### Design
+
+Objects will be stored in a path relative to the branch and staging token. For example: file `x/y.z` will be uploaded to path
+`<bucket-name>/<repo-name>/<branch-name>/<staging-token>/x/y.z`.
```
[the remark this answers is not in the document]
N-o-Z: I couldn't come up with any sufficient solution to this problem. The only option I thought of is adding a metadata entry on the object which marks it as deleted

design/open/hard-delete.md
N-o-Z: Yes, deeper examining of the copy object flow show's this assumption is irrelevant for option 3

design/open/hard-delete.md
```diff
@@ -0,0 +1,206 @@
...
+#### Delete Object
+
+1. If object staged
+   1. Read reference key
+   2. If not `committed` 
+      1. Change reference state to `deleted` and update `last_modified`
+
+#### Commit
+
+1. Mark all entries in staging area as `committed`
```
itaiad200, line 196: Are they ever deleted from here? We cannot reference all physical addresses in the kvstore

design/open/hard-delete.md
```diff
@@ -0,0 +1,206 @@
...
+
+Assume this API uses a physical address in the repo namespace
+1. Add reference entry to the database
+2. Add staging entry
+
+#### Delete Object
+
+1. If object staged
+   1. Read reference key
+   2. If not `committed`
```
itaiad200, line 191: How can it be committed?

design/open/hard-delete.md
itaiad200, line 196: That takes some time and not easily revertable.

design/open/hard-delete.md
N-o-Z: race between Commit and StageObject

design/open/hard-delete.md
itai-david: I'm not sure I follow on this one - I'm asking about a branch with all data commited - no objects at staging are for that matter, and say a certain object is only referenced by this branch. What happens when the branch is deleted? The (now) unreferenced object is not accessible via any of staging tokens, as these are gone

design/open/hard-delete.md
N-o-Z: I need to check further but it might be possible to perform the marking during the iteration over the changes as part of the commit process - this should save us some time

design/open/hard-delete.md
```diff
@@ -0,0 +1,206 @@
...
+1. Object path is in the repo namespace:
+   1. Add reference entry to the database 
+2. Object path is outside the repo namespace:
+   1. lakeFS will not handle its retention
+3. Add staged entry
+
+#### LinkPhysicalAddress
+
+Assume this API uses a physical address in the repo namespace
+1. Add reference entry to the database
```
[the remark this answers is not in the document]
N-o-Z: In addition to the address we will provide on GetPhysicalAddress a validation token which will have also an expiry time (corresponding with the retention policy of the garbage collection). On GetPhysicalAddress we will create a reference entry with state deleted.
LinkPhysicalAddress will check the token validity, if token is valid - create entry and add reference for physical address with state `staged`.
As part of the changes - I'm starting to think we should drop support altogether for StageObject

design/open/hard-delete.md
N-o-Z: We do not track committed data references, once data is committed the reference keys will eventually be deleted (only the reference keys, not the data) by the background process. Data which is committed is referenced by its commit, not its branch. By design we do not deleted these objects. This proposal discusses only the garbage collection on uncommitted data. For committed data we have the external GC process retention policy. I hope I answered what was asked

itai-david: Actually you did. Thanks

design/open/hard-delete.md
itaiad200: It's a different partition. You would be replacing 2 iterations over the staging area with a single iterations that does 2 things. I don't think it would optimize the runtime.

itaiad200: Or the ability to revert easily.

design/open/hard-delete.md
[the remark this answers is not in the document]
arielshaqed: I think this is an important issue to resolve.  (I know it's hard because AFAIR we've had several rounds on it.)

Adding metadata doesn't work because then you need to get actual data for the object to know how to handle it.  E.g. think of listing, suddenly it has to HEAD every object in the listing.  So your *only* options are to use [data available in ListObjectsV2](https://docs.aws.amazon.com/AmazonS3/latest/API/API_ListObjectsV2.html#API_ListObjectsV2_ResponseSyntax), where every object is represented by an [Object](https://docs.aws.amazon.com/AmazonS3/latest/API/API_Object.html).  AFAICT your only chance there is an ETag -- if you declared some sequence of bytes "The Deleted Object" then you could compute its ETag and try to upload _that_ whenever you wanted to delete an object.  I am **not a fan** of having fixed "forbidden" strings, however -- they're exactly the kind of thing people like me end up wanting to store inside actual objects.

arielshaqed (commented):
Thanks!  Historically being able to represent deleted objects in staging has been the breaker for many similar proposals.  I think that we should definitively resolve  the representation of deleted objects before proceeding.

That said, I am not sure why "online hard-delete" has to depend on "no-KV staging contents" -- an alternate way forward might be to find a formulation that does not depend on keeping the staging catalogue on S3.

design/open/hard-delete.md
```diff
@@ -0,0 +1,257 @@
...
+3. Otherwise, modify reference state to `deleted` and update `last_modified`
+
+## 4. Staging Token States
+
+Track state on the staging token instead of objects - but keep tracking references (without state)
+
+    <repo>/staging/deleted/<staging_token>
+    <repo>/staging/committed/<commit_id> - list(staging tokens)
+
+### Flows
```
itai-david, line 225: Where is the Commit flow?

design/open/hard-delete.md
```diff
@@ -0,0 +1,257 @@
...
+
+* Upload will add reference to physical address as before
+* Delete object will not remove the reference
+* References to objects are kept as long as staging token is still 'active'
```
itai-david, line 229: Please define `active`

design/open/hard-delete.md
```diff
@@ -0,0 +1,257 @@
...
+
+When resetting a branch we throw all the uncommitted branch data, this operation happens asynchronously via `AsyncDrop`.
+We can leverage this asynchronous job to also check and perform the hard-delete
+1. Scan deleted staging token entries
+2. If entry is `tombstone` remove entry reference key
+3. Otherwise, modify reference state to `deleted` and update `last_modified`
+
+## 4. Staging Token States
+
+Track state on the staging token instead of objects - but keep tracking references (without state)
```
itai-david, line 220: Not sure I understand - now we only keep a reference entry, for staged?

design/open/hard-delete.md
```diff
@@ -0,0 +1,257 @@
...
+
+#### CopyObject
+
+**Open:** If and how to support copy object
+
+### Background delete job
+
+* Scan committed tokens
+    * For each committed object - remove all references of object
+    * "Move staging token to deleted" (all objects in committed staging tokens that were not actually committed are candidates for deletion)
```
itai-david, line 252: Which staging token?

itai-david (changes requested):
Thank you @N-o-Z for uploading suggestion #4
As the general vibe is that this (#4) is the solution we are currently aiming at, can you please add a bit more details? Make it more workable, towards implementation?
Thanks

N-o-Z:
> Thank you @N-o-Z for uploading suggestion #4 As the general vibe is that this (#4) is the solution we are currently aiming at, can you please add a bit more details? Make it more workable, towards implementation? Thanks

@itai-david  Thank you for the very valuable input. Please see the updated proposal and please let me know how else this can be improved

design/open/hard-delete.md
```diff
@@ -0,0 +1,307 @@
...
+
+For each staging token of a deleted / reset branches a deleted entry will be added
+
+    <repo>/staging/deleted/<staging_token>
+
+For each commit we will add the staging tokens to a committed entry.
+Adding a reference is done using `set if` and a retry mechanism to ensure consistency.
+
+    key    = <repo>/references/<physical_address>  
+    value  = list(staging tokens)
```
itai-david, line 234: Looks like the same as for reference

design/open/hard-delete.md
```diff
@@ -0,0 +1,307 @@
...
+      2. Add reference
+      3. Add entry in staging area
+
+#### CopyObject
+
+The new flow ensures that we do not add a reference for a committed object after we removed all it's references as part of 
+the background delete job, and preventing the accidental deletion of committed objects.
+
+1. While retry not exhausted
+   1. Read reference list for address, if list is empty
```
itai-david, line 281: "List is empty" - does that include a key not found, or ONLY key found with empty list?

design/open/hard-delete.md
```diff
@@ -0,0 +1,307 @@
...
+
+#### Reset / Delete Branch
+
+1. Perform reset / delete flow in the same manner as today
+2. On successful execution - create entries for all staging tokens under the `deleted` prefix
+
+### Background delete job
+
+* Scan committed tokens
+    * For each committed object - remove all references of object
```
itai-david, line 299: That's the one with the staging token list, right?

    key    = <repo>/references/<physical_address>  
    value  = list(staging tokens)

design/open/hard-delete.md
```diff
@@ -0,0 +1,307 @@
...
+1. While retry not exhausted
+   1. Read reference list for address, if list is empty
+      1. Assume address is committed or not managed by lakeFS
+   2. Else - add branch staging token to reference list and perform `set if`
+      1. If predicate failed - retry
+
+#### Commit
+
+1. Perform commit flow in the same manner as today
+2. On successful commit - create entry using `commit_id` and ordered list of tokens under the `committed` prefix
```
itai-david, line 289: Do we still clean staging entries from the DB? If so, how can we scan commited objects for each token, as part of the BG delete?

itai-david (commented):
Once again, thanks @N-o-Z for this hard work. Definitely a challenging one
I'm still missing some pieces in the BG delete procedure described at the bottom of #4.
I think it my be best to add some "illustration" (textual) of the process. I mean, the state of the KV after a variety of actions, to include all the new entries you introduce, and than display how it is used by the BG delete.
Is that to much to add? I think it can add a lot of value.

design/open/hard-delete.md
N-o-Z: We will still be cleaning the staging entries from the DB, but it will have to be done in a different way. Either as part of or triggered by the background job. I wasn't sure if to add this as it borders with implementation details

design/open/hard-delete.md
N-o-Z: Sorry, there was a typo here - it refers to the `staging/committed' key. Fixed

design/open/hard-delete.md
N-o-Z: BTW it's mentioned in the background job flow (Delete staging area)

design/open/hard-delete.md
N-o-Z: Updated the background job and the copy flow to be more clear

design/open/hard-delete.md
N-o-Z: Yes - rephrased

arielshaqed (commented):
Thanks!

IIUC we are only considering option 4 (and maybe 5).  Can we remove the other options (and rephrase 4 without 3)?  I think it would help clarify the proposal.

design/open/hard-delete.md
```diff
@@ -0,0 +1,400 @@
...
+- The S3 gateway `put` operation with the copy support will use the underlying adapter to copy the data. Will require from each adapter to support/emulate a 'copy' operation.
+* (optional) Enable 'move' API as alternative to 'copy' we have seen that moving objects from one location to another by copy+delete of metadata will enable easy support for lakeFSFS move. This can be implemented by marking the entry on staging as 'locked'. During the move operation - lock+copy+delete+unlock operation. In case of any failure, as we don't have transactions, we may keep two entries, or keep locked object without ever delete its physical address.
+
+TODO(Barak): list more cases and how it will be address in this solution.
+
+### Flows
+
+#### Move operation
+1. Get entry
+2. "Lock" entry (`SetIf`)
```
itaiad200, line 341: What is "Lock"? How is it implemented, how is it released? How is it treated by other operations?

design/open/hard-delete.md
```diff
@@ -0,0 +1,400 @@
...
+
+- Make sure we don't have the same physical address by sign links we return in our API. The logical path will be part of the signature that will be part of the physical address we generate. This will enable us to block and use of a physical address outside the repo/branch/entry.
+- The S3 gateway `put` operation with the copy support will use the underlying adapter to copy the data. Will require from each adapter to support/emulate a 'copy' operation.
+* (optional) Enable 'move' API as alternative to 'copy' we have seen that moving objects from one location to another by copy+delete of metadata will enable easy support for lakeFSFS move. This can be implemented by marking the entry on staging as 'locked'. During the move operation - lock+copy+delete+unlock operation. In case of any failure, as we don't have transactions, we may keep two entries, or keep locked object without ever delete its physical address.
+
+TODO(Barak): list more cases and how it will be address in this solution.
+
+### Flows
+
+#### Move operation
```
itaiad200, line 339: If it's optional can we start without?

design/open/hard-delete.md
```diff
@@ -0,0 +1,400 @@
...
+4. Delete old entry
+5. "Unlock" entry (new entry) 
+
+#### Upload Object
+
+1. Get entry
+2. "Lock" entry if exists (override scenario) (`SetIf`)
+3. Write blob
+4. Add staged entry (or update)
+5. Delete physical address if previously existed
```
itaiad200, line 352: If the physical path is predetermined by the logical path and the staging token, why do we need to delete anything?

design/open/hard-delete.md
```diff
@@ -0,0 +1,400 @@
...
+#### 3. Concurrent Move
+Concurrent move should be blocked as it may cause multiple pointers for the same physical address.
+The most straightforward way to do so is to rely on the absence of the lock.
+The problem with this approach is that permanently locked addresses (which is a protection mechanism in case of catastrophic failures)
+can never be moved -which leads us to the stale lock problem.
+
+#### 4. Stale lock problem
+We use the lock mechanism to protect from race scenarios on Upload (override), Move and Delete. As such, an entry with a stale
+lock will prevent us from performing any of these operations.
+We can overcome it with a time based lock - but this might present additional challenges to the proposed solution.
```
itaiad200, line 400: How can we not have it? Can we live with a system that some entries are locked for good?

design/open/hard-delete.md
```diff
@@ -0,0 +1,400 @@
...
+4. Add staged entry (or update)
+5. Delete physical address if previously existed
+
+#### Delete Object
+1. Get entry
+2. "Lock" entry (`SetIf`) to protect from delete - move race
+3. Delete staging entry
+4. Hard delete object
+
+### Races and issues
```
itaiad200, line 360: Can we not support move and avoid all races? 👿

itaiad200 (changes requested):

design/open/hard-delete.md
N-o-Z: IMHO this is not something that we can live with. But - we need to take into account that using a time based lock might introduce new problems

design/open/hard-delete.md
N-o-Z: It is not - it was never suggested to couple the logical and physical path in this proposal (#5). We only ensure we have a single logical address for each physical address

design/open/hard-delete.md
N-o-Z: AFAIU this is a requirement. The move operation was created to support a main flow in our Spark client which currently is being performed on the metadata level (by using copy + delete).
As per @arielshaqed, this is something that we do not want to support by a full copy due to the performance implications

design/open/hard-delete.md
N-o-Z: See previous comment regarding Move

design/open/hard-delete.md
itaiad200: So let's break it down then

design/open/hard-delete.md
N-o-Z: Added more information about the lock. Operations which involve the locking mechanism are explicitly addressed.Operations that are not mentioned are agnostic to the lock. (If you think I missed any please tell)

design/open/hard-delete.md
N-o-Z: As I see it - we should first try to solve issue #1 (Commit - Move Race), before taking the time to solve this one.
Seems like the commit move race is our blocker.

design/open/hard-delete.md
```diff
@@ -0,0 +1,334 @@
...
+
+For each staging token of a deleted / reset branches a deleted entry will be added
+
+    <repo>/staging/deleted/<staging_token>
+
+For each commit we will add the staging tokens to a committed entry.
+Adding a reference is done using `set if` and a retry mechanism to ensure consistency.
+
+    key    = <repo>/staging/committed/<physical_address>  
+    value  = list(staging tokens)
```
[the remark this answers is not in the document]
N-o-Z: Eventually committed references will be deleted by the background job

design/open/hard-delete.md
arielshaqed: Time based locks are inherently unsafe unless your infrastructure has fairly strict guarantees on clocks.  We need to ensure that at least AWS + EKS provide such a guarantee.

design/open/hard-delete.md
arielshaqed: Note that Spark lakeFSFS and Hadoop S3A do not exactly require a "move".  We currently use LinkPhysicalAddress followed by a delete.  So it isn't atomic, but at least it requires only metadata operations.

arielshaqed (commented):
Thanks!

IIUC we've ruled out options 1-4.  Can we perhaps get rid of them?  It would make it easier (for me) to focus. 🔍

design/open/hard-delete.md
```diff
@@ -0,0 +1,334 @@
...
+* Delete object will not remove the reference
+* References to objects are kept as long as staging token is not deleted or committed.
+
+#### Delete Object
+
+Deleted object will remain the same. We do not delete references in this flow.
+
+#### Stage Object
+Allow stage object operations only on addresses outside the repository namespace
+lakeFS will not manage the retention of these objects
```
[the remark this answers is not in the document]
N-o-Z: The link you provided and what I saw myself (though not savvy with our Java code) is for renameObject (move).
For that purpose we should use the CopyObject instead.
Short answer: Yes - if and when these changes are implemented, the old client's renameObject will no longer work

N-o-Z:
> Thanks!
> 
> IIUC we've ruled out options 1-4. Can we perhaps get rid of them? It would make it easier (for me) to focus. mag

Thanks @arielshaqed 
I feel that as long as we didn't settle on any proposal, we should keep all of the proposals information. In case this will be eventually rejected we will benefit from having documentation of all the attempted solutions for future reference.

wengchenyang1:
> > Thanks!
> > IIUC we've ruled out options 1-4. Can we perhaps get rid of them? It would make it easier (for me) to focus. mag
> 
> Thanks @arielshaqed I feel that as long as we didn't settle on any proposal, we should keep all of the proposals information. In case this will be eventually rejected we will benefit from having documentation of all the attempted solutions for future reference.

Hello @N-o-Z , may I ask for the progress of proposal 5? We face compliance issue because some files are not hard deleted on s3.

N-o-Z:
@wengchenyang1,
All the suggested proposals in this document have inherent faults which make non of them a viable solution. We've decided for the time being to work in other directions for this problem.
We''re currently working on designing an offline solution similar to the current GC process we have.
I'll make sure to tag you once the PR is opened

itaiad200:
@N-o-Z can we merge this proposal to rejected?

N-o-Z:
> @N-o-Z can we merge this proposal to rejected?

Yes - lets do that

itaiad200 (approved):
Although this was rejected, the hard work on this suggestion proved us that an online consistent, atomic & maintainable solution to this problem is not feasible. This output is better than implementing a design that cannot work :)

N-o-Z:
> Although this was rejected, the hard work on this suggestion proved us that an online consistent, atomic & maintainable solution to this problem is not feasible. This output is better than implementing a design that cannot work :)

@itaiad, I still want to believe that in the future we will be able to find a working solution for this problem

PR #4017, opened 2022-08-28
create api configurations class

### LINKED ISSUE #3989: GC: improve API client initialization readability

talSofer:
Closes #3989

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
@@ -81,10 +81,10 @@ private object ApiClient {
   }
 }
 
-class APIConfigurations(
-    var apiUrl: String,
-    var accessKey: String,
-    var secretKey: String,
+case class APIConfigurations(
+    val apiUrl: String,
+    val accessKey: String,
+    val secretKey: String,
     val connectionTimeoutSec: String = "",
     val readTimeoutSec: String = ""
 ) {
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
@@ -40,12 +40,12 @@ private object ApiClient {
     ClientKey(conf.apiUrl, conf.accessKey),
     new Callable[ApiClient] {
       def call() = new ApiClient(
-        new APIConfigurations(conf.apiUrl,
-                              conf.accessKey,
-                              conf.secretKey,
-                              conf.connectionTimeoutSec,
-                              conf.readTimeoutSec
-                             )
+        APIConfigurations(conf.apiUrl,
+                          conf.accessKey,
+                          conf.secretKey,
+                          conf.connectionTimeoutSec,
+                          conf.readTimeoutSec
+                         )
       )
     }
   )
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
@@ -82,23 +82,22 @@ private object ApiClient {
 }
 
 case class APIConfigurations(
-    val apiUrl: String,
-    val accessKey: String,
-    val secretKey: String,
-    val connectionTimeoutSec: String = "",
-    val readTimeoutSec: String = ""
+    apiUrl: String,
+    accessKey: String,
+    secretKey: String,
+    connectionTimeoutSec: String = "",
+    readTimeoutSec: String = ""
 ) {
   val FROM_SEC_TO_MILLISEC = 1000
 
-  var connectionTimeoutMillisec: Int = TIMEOUT_NOT_SET
-  var readTimeoutMillisec: Int = TIMEOUT_NOT_SET
+  val connectionTimeoutMillisec: Int = stringAsMillisec(connectionTimeoutSec)
+  val readTimeoutMillisec: Int = stringAsMillisec(readTimeoutSec)
 
-  if (connectionTimeoutSec != null && !connectionTimeoutSec.isEmpty) {
-    this.connectionTimeoutMillisec = connectionTimeoutSec.toInt * FROM_SEC_TO_MILLISEC
-  }
-
-  if (readTimeoutSec != null && !readTimeoutSec.isEmpty) {
-    this.readTimeoutMillisec = readTimeoutSec.toInt * FROM_SEC_TO_MILLISEC
+  def stringAsMillisec(s: String): Int = {
+    if (s != null && !s.isEmpty)
+      s.toInt * FROM_SEC_TO_MILLISEC
+    else
+      TIMEOUT_NOT_SET
   }
 }
 
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
@@ -8,8 +8,8 @@ import io.lakefs.clients.api.model.{
 }
 import io.treeverse.clients.StorageClientType.StorageClientType
 import io.treeverse.clients.StorageUtils.{StorageTypeAzure, StorageTypeS3}
-
 import com.google.common.cache.{Cache, CacheBuilder, CacheLoader, LoadingCache}
+import io.treeverse.clients.ApiClient.TIMEOUT_NOT_SET
 
 import java.net.URI
 import java.util.concurrent.{Callable, TimeUnit}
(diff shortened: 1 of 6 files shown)


clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -35,16 +36,17 @@ private object ApiClient {
...
+    ClientKey(conf.apiUrl, conf.accessKey),
     new Callable[ApiClient] {
-      def call() = new ApiClient(apiUrl, accessKey, secretKey, connectionTimeoutSec, readTimeoutSec)
+      def call() = new ApiClient(
+        new APIConfigurations(conf.apiUrl,
+                              conf.accessKey,
+                              conf.secretKey,
+                              conf.connectionTimeoutSec,
+                              conf.readTimeoutSec
+                             )
```
arielshaqed, line 48: Isn't this just a copy of APIConfigurations?

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -79,29 +81,40 @@ private object ApiClient {
...
-class ApiClient private (
-    apiUrl: String,
-    accessKey: String,
-    secretKey: String,
-    connectionTimeoutSec: String = "",
-    readTimeoutSec: String = ""
+class APIConfigurations(
+    var apiUrl: String,
+    var accessKey: String,
+    var secretKey: String,
```
arielshaqed, line 87: Why are these mutable?

arielshaqed (approved):
Thanks!

@johnnyaug may have something to say about this, suggest asking for a review.

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
talSofer: by mistake, of course, thanks :)

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
arielshaqed: More details: all members of `conf` are immutable, it is a value class.  So there is no value (😀) in copying it.

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -79,29 +81,40 @@ private object ApiClient {
...
+    val connectionTimeoutSec: String = "",
+    val readTimeoutSec: String = ""
 ) {
   val FROM_SEC_TO_MILLISEC = 1000
 
-  val client = new api.ApiClient
-  client.setUsername(accessKey)
-  client.setPassword(secretKey)
-  client.setBasePath(apiUrl.stripSuffix("/"))
+  var connectionTimeoutMillisec: Int = TIMEOUT_NOT_SET
```
arielshaqed, line 93: More Scala-ish (functional, immutable):

```suggestion
  val connectionTimeoutMillisec: Int =
    if (connectionTimeoutSec != null && !connectionTimeoutSec.isEmpty)
      connectionTimeoutSec.toInt * FROM_SEC_TO_MILLISEC
    else
      TIMEOUT_NOT_SET
```
and the same for readTimeoutSec.

Since it repeats twice, you might

```scala
def stringAsMillisec(s: String): Int {
  return if (s != null && s != "") s.toInt * FROM_SEC_TO_MILLISEC else TIMEOUT_NOT_SET
}
```

and use that.

arielshaqed (changes requested):
Thanks!

As we discussed off-review yesterday, I would *much* rather that we did either one of these:
* Use a proper duration parser, so I can configure a timeout to be something like `1m10s`.  `java.time.Duration.parse("PT" + s)` will probably do this.
* Add a proper duration parser later, and rename the _currently used_ configuration options to end `_secs` or `seconds`, to make configuration clear.  The variables that carry these unitless numbers should also have units in their names.  Getting the wrong value is too easy.

About default timeout values: assuming we always get these from a Hadoop configuration.  So we might use [`org.hadoop.Configuration.get(name String, defaultValue String)`](https://hadoop.apache.org/docs/r2.7.1/api/org/apache/hadoop/conf/Configuration.html#get(java.lang.String,%20java.lang.String)) to get-with-default, and get rid of the entire setting logic.  WDYT?

I prefer to keep

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
talSofer: done * 2

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
talSofer: Changed APIConfigurations into a case class

> Thanks!
> 
> As we discussed off-review yesterday, I would _much_ rather that we did either one of these:
> 
> * Use a proper duration parser, so I can configure a timeout to be something like `1m10s`.  `java.time.Duration.parse("PT" + s)` will probably do this.
> * Add a proper duration parser later, and rename the _currently used_ configuration options to end `_secs` or `seconds`, to make configuration clear.  The variables that carry these unitless numbers should also have units in their names.  Getting the wrong value is too easy.
> 
> About default timeout values: assuming we always get these from a Hadoop configuration. So we might use [`org.hadoop.Configuration.get(name String, defaultValue String)`](https://hadoop.apache.org/docs/r2.7.1/api/org/apache/hadoop/conf/Configuration.html#get(java.lang.String,%20java.lang.String)) to get-with-default, and get rid of the entire setting logic. WDYT?
> 
> I prefer to keep

Thanks! 

I changed the property name as suggested in option number 2. 
as for the default timeout values, since these are defaults of the API client, I'm avoiding hard-coding it somewhere else. that is, we are not getting them from hadoop configurations by default, makes sense?

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -35,16 +36,17 @@ private object ApiClient {
...
     new Callable[ApiClient] {
-      def call() = new ApiClient(apiUrl, accessKey, secretKey, connectionTimeoutSec, readTimeoutSec)
+      def call() = new ApiClient(
+        APIConfigurations(conf.apiUrl,
+                          conf.accessKey,
+                          conf.secretKey,
+                          conf.connectionTimeoutSec,
+                          conf.readTimeoutSec
+                         )
+      )
```
arielshaqed, line 49: I guess I am asking if we can just do this:
```suggestion
      def call() = new ApiClient(conf)
```

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSContext.scala
```diff
@@ -15,8 +15,8 @@ object LakeFSContext {
   val LAKEFS_CONF_API_URL_KEY = "lakefs.api.url"
   val LAKEFS_CONF_API_ACCESS_KEY_KEY = "lakefs.api.access_key"
   val LAKEFS_CONF_API_SECRET_KEY_KEY = "lakefs.api.secret_key"
-  val LAKEFS_CONF_API_CONNECTION_TIMEOUT_KEY = "lakefs.api.connection.timeout"
-  val LAKEFS_CONF_API_READ_TIMEOUT_KEY = "lakefs.api.read.timeout"
+  val LAKEFS_CONF_API_CONNECTION_TIMEOUT_SEC_KEY = "lakefs.api.connection.timeout.seconds"
+  val LAKEFS_CONF_API_READ_TIMEOUT_SEC_KEY = "lakefs.api.read.timeout.seconds"
```
arielshaqed, line 19:
```suggestion
  val LAKEFS_CONF_API_CONNECTION_TIMEOUT_SEC_KEY = "lakefs.api.connection.timeout_seconds"
  val LAKEFS_CONF_API_READ_TIMEOUT_SEC_KEY = "lakefs.api.read.timeout_seconds"
```

We'll end up using `lakefs.api.connection.timeout` and `lakefs.api.read.timeout`, and having something that looks like a field in them will be confusing.

arielshaqed (approved):
THANKS!

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSContext.scala
talSofer: done

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
talSofer: I've considered it and am trying to remember why it didn't work..! :) 
I'm going ahead with the pr without addressing this comment for time reasons, will open an issue to track this improvement.

PR #4022, opened 2022-08-29
Fix GC test timeout config keys

### LINKED ISSUE #4021: Fix GC test timeout config keys

arielshaqed:
Fixes #4021.


talSofer (approved):
Thanks! how was this tested?

arielshaqed:
> Thanks! how was this tested?

Good question :-(.  No easy way to test this (that's why it broke in the first place), I think that this is one of the ways in which Hadoop config is broken.

If you prefer I can push a dummy branch to CI and run it with *zero* timeouts, to see that everything breaks.  I'd rather not delay this PR for that, so pulling it now.

THANKS!

PR #4035, opened 2022-08-29
Delete repositories at the end of the test

idanovo:


esti/system_test.go
```diff
@@ -118,6 +126,13 @@ func createRepository(ctx context.Context, t *testing.T, name string, repoStorag
 		"create repository '%s', storage '%s'", name, repoStorage)
 }
 
+func deleteRepository(ctx context.Context, t *testing.T, repositoryName string) {
+	resp, err := client.DeleteRepositoryWithResponse(ctx, repositoryName)
+	require.NoErrorf(t, err, "failed to delete repository '%s'", repositoryName)
```
N-o-Z, line 131: I would not fail the test if the repository doesn't exist

esti/system_test.go
```diff
@@ -73,6 +73,14 @@ func setupTest(t *testing.T) (context.Context, logging.Logger, string) {
 	return ctx, logger, repo
 }
 
+func tearDownTest(t *testing.T, repoName string) {
+	ctx := context.Background()
```
N-o-Z, line 77: We should have a flag / env variable that should be disabled by default and we can toggle to perform the teardown. We are purposely not cleaning up tests for debugging

esti/system_test.go
```diff
@@ -73,6 +73,14 @@ func setupTest(t *testing.T) (context.Context, logging.Logger, string) {
...
+	name := makeRepositoryName(t.Name())
+	logger := logger.WithField("testName", name)
+	deleteRepository(ctx, t, repoName)
```
N-o-Z, line 80: It's important to keep in mind that deleting the repository, the data uploaded to the bucket will still exist. Ultimately if you use the same lakeFS instance for testing over a long period - this will spam and bloat your bucket

esti/migrate_test.go
```diff
@@ -169,13 +169,16 @@ func preMigrateTests(t *testing.T) {
 }
 
 func postMigrateTests(t *testing.T) {
+	defer deleteRepository(context.Background(), t, migrateStateRepoName)
```
N-o-Z, line 172: I have some concerns using the deleteRepository function directly.
What if I want to keep the repository for debugging purposes?

esti/s3_gateway_test.go
```diff
@@ -112,7 +114,11 @@ func TestS3UploadAndDownload(t *testing.T) {
 
 func TestS3CopyObject(t *testing.T) {
 	ctx, _, repo := setupTest(t)
-	destRepo := createRepositoryByName(ctx, t, "tests3copyobjectdest")
+	defer tearDownTest(t, repo)
+
+	destRepoName := "tests3copyobjectdest"
+	destRepo := createRepositoryByName(ctx, t, destRepoName)
+	defer deleteRepository(context.Background(), t, destRepoName)
```
N-o-Z, line 121: Same as previous

esti/migrate_test.go
idanovo: if It's for debugging, you can comment out the deletions. The problem is that if I do not delete those repos, the next run will fail on an existing installation

esti/system_test.go
```diff
@@ -73,11 +73,9 @@ func setupTest(t *testing.T) (context.Context, logging.Logger, string) {
 	return ctx, logger, repo
 }
 
-func tearDownTest(t *testing.T, repoName string) {
+func tearDownTest(repoName string) {
 	ctx := context.Background()
-	name := makeRepositoryName(t.Name())
-	logger := logger.WithField("testName", name)
-	deleteRepository(ctx, t, repoName)
+	deleteRepositoryIfAskedTo(ctx, repoName)
```
N-o-Z, line 78: Not a blocker, but I recommend the env var to be for teardown which will be checked both in tearDownTest and in deleteRepositoryIfAskedTo

idanovo: why double checking?

N-o-Z: Simply because in the future we might want to add more logic to the teardown function and this flag will be relevant for it. Optimally the check for this flag should be only on this function. But since you explicitly use the delete repository function we need to add the check also there.
But again, not blocking

PR #4038, opened 2022-08-30
Fix kv ordered commit iterator close panic

### LINKED ISSUE #4046: lakefs panic in kv commit iterator close during commit listing
While KVOrderedCommitIterator SeekGE fails it will set `nil` into a member iterator.
The practice of calling `defer Close()` on after successful NewKVOrderedCommitIterator will call the Close() implementation while the point is set to nil.

nopcoder:
Fix #4046


itaiad200:
It looks like it fixed the panic!!!
I ran the Esti tests again to validate it.

itaiad200 (approved):
LGTM

PR #4057, opened 2022-08-31
[Alg] Reuse ranges of contended merges to speed up concurrent merges

arielshaqed:


arielshaqed:
I asked DALL-E to draw this.  Here's
> A cyborg reusing ranges to speed up concurrent merges

DALL-E drawing of "A cyborg reusing ranges to speed up concurrent merges" shows a cyborg doing something with 2 targets and meaningless Latin-character text"

design/open/reuse-contented-merge-metaranges/reuse-contented-merge-metaranges.md
```diff
@@ -0,0 +1,128 @@
...
+
+  This failure is due to losing a race against a concurrent merge or commit.
+  It must either be retried or abandoned (and, presumably, retried later).
+
+Naturally only race failures are of interest.  When a merge fails because of
+a merge, it must be retried.  However we can restart it using _the generated
+merged metarange_ as source and the original destination as the base!  (See
+"Correctness" below for why this yields the correct result.)
+
+The performance gain occurs during the retry: We expect the new source to
```
eden-ohana, line 70: To gain performance improvement the assumption here is that the two merges don't change the same ranges, need to make sure that this is the case

design/open/reuse-contented-merge-metaranges/reuse-contented-merge-metaranges.md
```diff
@@ -0,0 +1,128 @@
...
+| base | src | dst/base' | result/src' | dst' | final result | comment                        |
+|------|-----|-----------|-------------|------|--------------|--------------------------------|
+| A    | B   | A         | B           | A    | B            |                                |
+| A    | A   | B         | B           | B    | B            |                                |
+| A    | A   | A         | A           | B    | B            | (Only) dst changed during race |
+| A    | B   | C         | conflict    | --   | conflict     | Conflicts never proceed        |
+| A    | B   | A         | B           | C    | conflict     | Conflict with updated dst      |
+| A    | A   | B         | B           | C    | C            |                                |
+
+
```
eden-ohana, line 124: I think there is a correction issue when using the conflict resolution flag. The example uses one file and the letter represents a different version of it.

Two concurrent merges

1. Source wins in case of conflict
Base	A
Source B
Dest C
Res B

1. Dest wins in case of conflict
Base A
Source D
Dest C
Res C

The first one finish first so the second one will run with the result metarange - C and the dest wins flag:
 
Base C
Source C
Dest B
Res B

eden-ohana (commented):
Great idea!

nopcoder (approved):

design/open/reuse-contented-merge-metaranges/reuse-contented-merge-metaranges.md
arielshaqed: You're right: this will be true for some but not all merges.  With our emerging focus on merging as the basic tool for handling multi-object file formats (Spark outputs, probably Iceberg, maybe Delta too), I would expect ranges to separate if there are repeated racing merges -- simply because large multi-objects will split off into their range soon enough :-)

Adding words to that effect.

design/open/reuse-contented-merge-metaranges/reuse-contented-merge-metaranges.md
arielshaqed: I think conflict resolution also works here, actually.  You just need to interpret the new base and source.

Think of this as doing _exactly_ what you would do during manual conflict resolution after losing a race:
* First you merge destination HEAD into your source.  This uses the _reverse_ conflict resolution strategy.
* Now you merge the result back to your destination HEAD.

arielshaqed (commented):
Thanks!

arielshaqed:
@eden-ohana PTAL :-)

eden-ohana (approved):
Thanks!

arielshaqed:
## THANKS!

Pulling by force because docs don't run tests 🤷🏽

PR #4062, opened 2022-08-31
Fix: Pyramid delete before open

### LINKED ISSUE #4061: Pyramid: Fix delete file before open
A race condition, where a file that is currently being opened, can be selected for eviction and deleted.
In this scenario, we will get an error while trying to open the file.

The solution is to add a lock mechanism, that requires delete function to wait for all users of the file to release to lock before deleting it

N-o-Z:
- Closes #4061


pkg/pyramid/tier_fs.go
```diff
@@ -289,7 +286,10 @@ func (tfs *TierFS) openWithLock(ctx context.Context, fileRef localFileRef) (*os.
 			"fullpath":  fileRef.fullPath,
 		}).Trace("try to lock for open")
 	}
-	_, err := tfs.keyLock.Compute(fileRef.filename, func() (interface{}, error) {
+
+	rcb := tfs.fileLocker.lock(fileRef.fsRelativePath)
+	defer rcb() // release callback
```
itaiad200, line 291:
```suggestion
	releaseFileLock := tfs.fileLocker.lock(fileRef.fsRelativePath)
	defer releaseFileLock()
```

itaiad200 (approved):

pkg/pyramid/file_locker.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"sync"
+
+	"github.com/treeverse/lakefs/pkg/pyramid/params"
+)
+
+// fileLocker allows file locking to avoid race conditions between cache rejection/eviction
+// and the file usage by TierFS
+type fileLocker struct {
+	wgMap *sync.Map
```
nopcoder, line 12: 1. Consider using map+mutex (usually it will work faster)
2. Using a map with mutex will remove the extra allocation of waitgroup in case there is already a waitgroup on the path

nopcoder (changes requested):

N-o-Z:
@itaiad200 @nopcoder  Please see new code

pkg/pyramid/file_tracker.go
```diff
@@ -0,0 +1,50 @@
...
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	ref := 1
+
+	if val, ok := t.refMap[string(path)]; ok {
+		ref = val + 1
+		t.refMap[string(path)] = ref
+	} else {
+		t.refMap[string(path)] = ref
+	}
```
nopcoder, line 34:
```suggestion
	if val, ok := t.refMap[string(path)]; ok {
		ref += val
	}
	t.refMap[string(path)] = ref
```

pkg/pyramid/file_tracker.go
```diff
@@ -0,0 +1,50 @@
...
+	refMap map[string]int
+	mu     sync.Mutex
+}
+
+type decCallback func(int)
+
+func newFileTracker() *fileTracker {
+	return &fileTracker{
+		refMap: map[string]int{},
+		mu:     sync.Mutex{},
```
nopcoder, line 20: not needed

pkg/pyramid/file_tracker.go
```diff
@@ -0,0 +1,50 @@
...
+
+	if val, ok := t.refMap[string(path)]; ok {
+		ref = val + 1
+		t.refMap[string(path)] = ref
+	} else {
+		t.refMap[string(path)] = ref
+	}
+}
+
+func (t *fileTracker) dec(path params.RelativePath, fn decCallback) {
```
nopcoder, line 37: I think we can remove the 'fn' argument here - if dec returns true/false or count - we can defer call to 'dec' and delete the file.

pkg/pyramid/tier_fs.go
```diff
@@ -114,7 +115,12 @@ func (tfs *TierFS) removeFromLocal(rPath params.RelativePath, filesize int64) {
 	// This will be called by the cache eviction mechanism during entry insert.
 	// We don't want to wait while the file is being removed from the local disk.
 	evictionHistograms.WithLabelValues(tfs.fsName).Observe(float64(filesize))
-	go tfs.removeFromLocalInternal(rPath)
+	// Decrement the reference count on the file. When the ref count reaches -1, The file will be deleted
+	tfs.fileTracker.dec(rPath, func(ref int) {
```
nopcoder, line 119: I think we need a 'delete' marker on the map - we like that for every time we called open we inc/dec in case the file wasn't found - it means that evict delete it.
if we like a better visibility I think the tracker should understand when to perform the callback or make sure that every time you call dec you also check and delete the file.

Something like:
tracker := NewTracker(deleteCallback) // callback to the implementation on how to delete file from local

closer := tracker.Open(path); // return a callback 
defer closer() // closer will take the lock remove the ref count when 0 and remove from map will check if a delete flag is mark and call the callback

tracker.Delete(path) // mark the file for delete or delete by keeping the lock

pkg/pyramid/file_tracker.go
```diff
@@ -0,0 +1,50 @@
...
+	defer t.mu.Unlock()
+	ref := 0
+
+	if val, ok := t.refMap[string(path)]; ok {
+		ref = val - 1
+		t.refMap[string(path)] = ref
+	}
+	if ref == -1 {
+		delete(t.refMap, string(path))
+	}
```
itai-david, line 48: 1. Why is -1 the point of eviction and not 0?
2. What is the meaning of `dec` an non existing path?

pkg/pyramid/tier_fs.go
```diff
@@ -114,7 +115,12 @@ func (tfs *TierFS) removeFromLocal(rPath params.RelativePath, filesize int64) {
...
+		if ref == -1 {
+			tfs.removeFromLocalInternal(rPath)
```
itai-david, line 121: What happens if `ref` returns as `-1` and before the file is removed, it is acquired again, by another thread.
I think it will still be removed, as the mutex protects the ref-count, but nothing prevents ref increase, after the `-1` is returned and `removeFromLocalInternal` is fired

itai-david (changes requested):
Can you please add some tests for this? Apart from helping in verifying this fix, it can also help in understanding what flows/use-cases are relevant, and by that to really understand the problem.

Another thing, I am not sure this fix really protect against deleting the file by one thread, from under the legs of another. Please see comment inside

pkg/pyramid/tier_fs.go
N-o-Z: This cannot happen because removeFromLocalInternal happens in the scope of dec where we have a mutex lock on the map.
What will happen is that on dec the entry will be deleted from the map, and the file deleted from storage. Only after that the lock will be released - the inc will happen and a new entry in the map will be created for the file

pkg/pyramid/file_tracker.go
N-o-Z: 0 Doesn't mean the file was flagged for eviction - it just means that nobody is currently trying to open it.
When we trigger the eviction function it performs a decrement on the file without a prior increment - which will eventually after all the open requests completed will cause it to reach -1. This is the indication that the file was requested for deletion and only then the logic will trigger

N-o-Z:
Please see revised code

itaiad200 (changes requested):
Blocking this PR until Sunday

pkg/pyramid/file_tracker.go
nopcoder: can remove the mu initialization it is a value type no need initialize it with the same value

pkg/pyramid/file_tracker.go
```diff
@@ -0,0 +1,73 @@
...
+
+func (t *fileTracker) close(path params.RelativePath) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	if val, ok := t.refMap[string(path)]; ok {
+		val.ref--
+		if val.ref == 0 && val.deleted {
+			t.innerDelete(path)
+		}
```
nopcoder, line 55: 1. when val.ref == 0 - we need to delete it from the map
2. when we delete it from the map and val.deleted true - we need to invoke the callback

pkg/pyramid/file_tracker.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+func (t *fileTracker) Delete(path params.RelativePath) {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+	if val, ok := t.refMap[string(path)]; ok {
+		val.deleted = true
+		if val.ref == 0 {
+			t.innerDelete(path)
+		}
```
nopcoder, line 71: On Delete - when path in map - mark it 'deleted' = true.
When path not in map - delete the file.
All while map is locked.

pkg/pyramid/tier_fs.go
```diff
@@ -256,19 +261,12 @@ func (tfs *TierFS) openFile(ctx context.Context, fileRef localFileRef, fh *os.Fi
 	}
 
 	if !tfs.eviction.Store(fileRef.fsRelativePath, stat.Size()) {
```
nopcoder, line 263: OnReject is called also in case method returns false?

nopcoder (changes requested):
Looks better - comments related on how we handle delete

pkg/pyramid/tier_fs.go
N-o-Z: Both OnEvict and OnReject CB are configured to run the same OnEvict method

nopcoder (approved):
lgtm!

pkg/pyramid/file_tracker_test.go
```diff
@@ -0,0 +1,55 @@
...
+	// Close one reference and ensure file still not deleted
+	closer()
+	require.True(t, testFileMap[string(file1)])
+
+	// Close last reference and ensure file was deleted
+	closer()
+	require.False(t, testFileMap[string(file1)])
+	require.True(t, testFileMap[string(file2)])
+
+	// Close delete file - sanity
```
itai-david, line 41:
```suggestion
	// Close deleted file - sanity
```

pkg/pyramid/file_tracker_test.go
```diff
@@ -0,0 +1,55 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/treeverse/lakefs/pkg/pyramid"
+	"github.com/treeverse/lakefs/pkg/pyramid/params"
+)
+
+var testFileMap map[string]bool
+
+func TestTracker(t *testing.T) {
```
itai-david, line 13: Awesome :star_struck:

itai-david (approved):
Thanks for the recent changers. Excellent test

pkg/pyramid/tier_fs.go
itaiad200: Right - but for a rejected file you will call `tfs.fileTracker.Delete(rPath)` twice. Here and in the `OnReject` callback (which calls `tfs.removeFromLocal`)

pkg/pyramid/tier_fs.go
```diff
@@ -128,10 +130,13 @@ func (tfs *TierFS) removeFromLocalInternal(rPath params.RelativePath) {
 		return
 	}
 
-	if err := tfs.syncDir.deleteDirRecIfEmpty(path.Dir(string(rPath))); err != nil {
-		tfs.logger.WithError(err).Error("Failed deleting empty dir")
-		errorsTotal.WithLabelValues(tfs.fsName, "DirRemoval")
-	}
+	// Delete Dir async
+	go func() {
```
itaiad200, line 134: Why are we changing it to async?

itaiad200 (changes requested):
See comment and question inside. 
Also - how was this tested? I don't think our current tests cover cache being filled up and many concurrent calls to the same paths.

pkg/pyramid/tier_fs.go
N-o-Z: removeFromLocalInternal was called asynchronously to not delay the flow on the delete dir operation. Now that it is called out of context, I've modified only the call to delete dir to be perfomed async

pkg/pyramid/tier_fs.go
N-o-Z: From what I see onReject is called only in processItems (NewCache, Clear)
This flow is for Set

N-o-Z:
> See comment and question inside. Also - how was this tested? I don't think our current tests cover cache being filled up and many concurrent calls to the same paths.

I added a unit test for the file tracker. It is true that we currently don't have test that perform concurrent operations but the unit test should cover this specific scenario

pkg/pyramid/tier_fs.go
itaiad200: Now sure I understand. `onReject` is called whenever a file isn't being stored (decided by their policy) - so here we would definitely call `Delete` twice for a rejected file.

itaiad200 (changes requested):

pkg/pyramid/tier_fs.go
N-o-Z: ristrettoEviction Store calls cache Set which in turn calls SetWithTTL.
In the SetWithTTL flow there are several places where the function return false without calling onReject CB.

itaiad200: Right - `onReject` only gets called when the eviction policy decides to reject. All other rejections are not called.

itaiad200 (approved):
Tested together on a real scenario - works amazing!

PR #4066, opened 2022-08-31
Add list of potential wins for LakeFSOutputCommitter

### LINKED ISSUE #4064: Add list of potential wins for LakeFSOutputCommitter

arielshaqed:
Fixes #4064.

### DIFF
--- a/design/open/spark-outputcommitter/committer.md
+++ b/design/open/spark-outputcommitter/committer.md
@@ -15,6 +15,30 @@ In this document:
 
 # Why <a href="#user-content-why" id="user-content-why">#</a>
 
+## Summary
+
+The LakeFSOutputCommitter will be:
+
+* **Faster.** Because it uses native lakeFS features we can avoid the long
+  troubled history of Spark writes on object stores.
+
+  * 3x fewer API calls than when using the FileOutputCommitter v1 algorithm
+	(which seems the most commonly used by our users).  Many of these are
+	entirely useless statObject calls.
+  * 2x fewer API calls that when using the FileOutputCommitter v2 algorithm
+    (which seems less used, and is often not as safe as the v1 algorithm).
+  * Around the same number of API calls as we would achieve if we supported
+    the new magic committer.  Many of these are entirely useless statObject
+    call.
+* **Atomic.** All objects and partitions of a file appear at once as a
+  committed object on the target branch. `_SUCCESS` files are still
+  generated, but only as a courtesy to AirFlow and similar triggers.  A
+  process that examines the target branch will never see a partial state.
+* **Integrated.** A full Spark write is a native lakeFS commit, a
+  first-class object in the lakeFS ecosystem.  It has a commit log (useful
+  for lineage and structured metadata tagging), it can be reverted and
+  merged, it has clear semantics for garbage collection, etc.
+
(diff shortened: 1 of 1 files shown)


adipolak:
Thanks @arielshaqed .

**Some pain points related to S3 that I wish this PR will resolve are:**

1. S3 Consistency used to be a pain and was solved with *strong read after write* for individual files which made EMRFS obsolete. but lack of atomic directories operations is a pain – copying a large set of parquet files (10k) and having multiple readers and writers at the same time can be a challenge when looking at the directories/repositories as a whole and not a file-by-file consistency. Hence, atomic operations on an s3 directory are desired for working with intense data applications and will solve many inconsistency issues with data.
Another example -  Having many readers to table with 1K parquet files and one update – update is not an atomic action, and readers might read the wrong data.

**more issues with S3, that could be nice to resolve are:**

2. AWS will often return HTTP 503 Slow Down errors for request rates that are much, much lower than their advertised limits.
3. You are supposed to code clients with backoffs/retries adequate to absorb arbitrary levels of HTTP 503 Slow Down until they finish scaling (which is entirely unobservable and could be minutes, hours, or days).

I understand that this PR will not solve 2+3, but it has the potential to drastically reduce metadata ops during writes which can improve the experience and number of errors or "slow down" exceptions.


Thank you again for looking into it.

design/open/spark-outputcommitter/committer.md
```diff
@@ -187,8 +211,17 @@ the same paths but keeping things safe regardless.
 [^2]: 	Whenever there is a conflict, we want the task branch (which will become
     the "latest writer" after a successful HOC commit) to win.
 
+# Potential wins
+
+* Explicit requests _not_ to write 3 times and give better behaviour than
+  the default FileOutputCommitter from Spark have appeared on our Slack
+  [#data-architecture-discussion][slack-dont-write-thrice].
+* Multiple users have requested "overwrite" save mode.
+* Multiple users have requested multi-writer support.
```
Jonathan-Rosenberg, line 220: I would specify how the LakeFSOutputCommitter would support that (basically mentioning that some writes would fail if multiple writes concurrently occur?)

arielshaqed: Actually that would depend on [SaveMode](https://spark.apache.org/docs/3.2.1/api/java/index.html?org/apache/spark/sql/SaveMode.html) I think:

* For append, everything is sort-of concatenated (I'm not sure that Spark actually does this, specifically as far as multiple values for the same key are concerned).
* For ErrorIfExists, all but one ("the first") writer should error out -- even in the face of concurrent writes.
* For ignore, the write should silently do nothing.
* For overwrite, "latest" writer wins.

I'd prefer not to do the research right now :-/

Jonathan-Rosenberg (approved):

design/open/spark-outputcommitter/committer.md
```diff
@@ -187,8 +211,17 @@ the same paths but keeping things safe regardless.
...
+* Note by a developer that Spark performs many lakeFS API operations when
```
talSofer, line 221: and also it makes things more predictable from the lakeFS server perspective. with FOC we know what amount of API calls to expect and their types.

talSofer (approved):

PR #4078, opened 2022-09-01
Fix retry branch update

### LINKED ISSUE #4076: retryBranchUpdate fails on the last (3rd) retry

eden-ohana:
closes #4076

### DIFF
--- a/pkg/graveler/graveler.go
+++ b/pkg/graveler/graveler.go
@@ -1743,11 +1743,11 @@ func (g *KVGraveler) retryBranchUpdate(ctx context.Context, repository *Reposito
 		if err != nil {
 			return backoff.Permanent(err)
 		}
-		if try > setTries {
-			return backoff.Permanent(fmt.Errorf("update branch: %w", ErrTooManyTries))
-		}
 		return nil
 	}, bo)
+	if try > setTries {
+		return backoff.Permanent(fmt.Errorf("update branch: %w", ErrTooManyTries))
+	}
 	return err
 }
 
--- a/pkg/graveler/graveler.go
+++ b/pkg/graveler/graveler.go
@@ -1740,11 +1740,14 @@ func (g *KVGraveler) retryBranchUpdate(ctx context.Context, repository *Reposito
 			try += 1
 			return err
 		}
+		if err != nil {
+			return backoff.Permanent(err)
+		}
+		if try > setTries {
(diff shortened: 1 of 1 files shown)


pkg/graveler/graveler.go
```diff
@@ -1742,7 +1742,7 @@ func (g *KVGraveler) retryBranchUpdate(ctx context.Context, repository *Reposito
 		}
 		return nil
 	}, bo)
-	if try == setTries {
+	if try > setTries {
```
N-o-Z, line 1745: @eden-ohana  There's still a bug on the if statement above this line (1736)
If `err != nil` and `try >= setTries` we return nil - which is incorrect
We should return the error if err != nil regardless of `try`

N-o-Z (changes requested):
I think there's still a bug in this flow. Please see comment

pkg/graveler/graveler.go
eden-ohana: The branch update will returns nil, and then the retry branch update will return an error because of line 1745. It's returning nil to exit the backoff.retry if you return an error it keeps retrying.

N-o-Z: `if try == setTries` and `err != nil` we return nil - This is a bug

eden-ohana: done

pkg/graveler/graveler.go
```diff
@@ -1740,10 +1740,13 @@ func (g *KVGraveler) retryBranchUpdate(ctx context.Context, repository *Reposito
...
 		}
+		if err != nil {
+			return backoff.Permanent(err)
+		}
 		return nil
 	}, bo)
-	if try == setTries {
-		return fmt.Errorf("update branch: %w", ErrTooManyTries)
+	if try > setTries {
+		return backoff.Permanent(fmt.Errorf("update branch: %w", ErrTooManyTries))
```
N-o-Z, line 1749: No need to wrap this with backoff.Permanent

eden-ohana: oops

pkg/graveler/graveler.go
eden-ohana: added changes to make it work

N-o-Z (approved):
Looks great, thanks!

PR #4084, opened 2022-09-01
Fix: KV Migration no space left on device

### LINKED ISSUE #4083: KV: Migration stuck indefinitely in case postgres no storage left on device
In case we run out of storage space during migration - the import flow which reads from file into the db is stuck due to all the workers exiting on errors before all the data was read from file.
The chan buffer used to allocate tasks to workers gets full and we are waiting for workers to take tasks from the buffer indefinitely.

N-o-Z:
- Closes #4083

### DIFF
--- a/pkg/kv/import.go
+++ b/pkg/kv/import.go
@@ -44,53 +44,59 @@ func (e *SafeEncoder) Encode(v interface{}) error {
 }
 
 func consumer(ctx context.Context, cancel context.CancelFunc, jobChan <-chan *Entry, store Store) error {
-	for e := range jobChan {
-		if len(e.PartitionKey) == 0 {
-			cancel()
-			return fmt.Errorf("bad entry partition key: %w", ErrInvalidFormat)
-		}
-		if len(e.Key) == 0 {
-			cancel()
-			return fmt.Errorf("bad entry key: %w", ErrInvalidFormat)
-		}
-		if e.Value == nil {
-			cancel()
-			return fmt.Errorf("bad entry value: %w", ErrInvalidFormat)
-		}
-		err := store.SetIf(ctx, e.PartitionKey, e.Key, e.Value, nil)
-		if err != nil {
-			cancel()
-			return fmt.Errorf("import (partition key: %s, key: %s): %w", e.PartitionKey, e.Key, err)
+	for {
+		select {
+		case e, more := <-jobChan:
+			if !more {
+				return nil
+			}
+			if len(e.PartitionKey) == 0 {
+				cancel()
+				return fmt.Errorf("bad entry partition key: %w", ErrInvalidFormat)
+			}
+			if len(e.Key) == 0 {
+				cancel()
+				return fmt.Errorf("bad entry key: %w", ErrInvalidFormat)
+			}
+			if e.Value == nil {
+				cancel()
+				return fmt.Errorf("bad entry value: %w", ErrInvalidFormat)
+			}
+			err := store.SetIf(ctx, e.PartitionKey, e.Key, e.Value, nil)
+			if err != nil {
+				cancel()
+				return fmt.Errorf("import (partition key: %s, key: %s): %w", e.PartitionKey, e.Key, err)
+			}
+		case <-ctx.Done():
+			return nil
 		}
 	}
-	return nil
 }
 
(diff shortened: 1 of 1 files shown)


pkg/kv/import.go
```diff
@@ -43,25 +43,58 @@ func (e *SafeEncoder) Encode(v interface{}) error {
...
+				// Decode does not return error on missing data / incompatible format
+				if err != nil {
+					cancel()
+					return fmt.Errorf("decoding entry: %w", err)
+				}
+				if i%100_000 == 0 {
+					log.Infof("Migrated %d entries", i)
+				}
+				jobChan <- entry
+			}
```
itai-david, line 93:
```suggestion
			} else {
			    time.Sleep(1 * time.Millisecond)
			}
```

Both avoids busy waiting and gives the channel a chance to release some entries. 1ms is not necessarily the best value

pkg/kv/import.go
```diff
@@ -43,25 +43,58 @@ func (e *SafeEncoder) Encode(v interface{}) error {
...
 		}
 		if len(e.Key) == 0 {
+			cancel()
 			return fmt.Errorf("bad entry key: %w", ErrInvalidFormat)
 		}
 		if e.Value == nil {
+			cancel()
 			return fmt.Errorf("bad entry value: %w", ErrInvalidFormat)
 		}
 		err := store.SetIf(ctx, e.PartitionKey, e.Key, e.Value, nil)
```
itai-david, line 60: Maybe we can check for `ctx.Done()` here too, prevent redundant DB writes when we now we already fail?

itai-david (approved):
Thanks @N-o-Z - looks excellent

Can we had tests to verify that fix? Would that be easy to achieve?

In addition, 2 suggestions below. Up to you if to accept ort not.
Other than that - Looks Great :rocket:

pkg/kv/import.go
```diff
@@ -43,25 +43,58 @@ func (e *SafeEncoder) Encode(v interface{}) error {
...
 			return fmt.Errorf("bad entry partition key: %w", ErrInvalidFormat)
 		}
 		if len(e.Key) == 0 {
+			cancel()
 			return fmt.Errorf("bad entry key: %w", ErrInvalidFormat)
 		}
 		if e.Value == nil {
+			cancel()
 			return fmt.Errorf("bad entry value: %w", ErrInvalidFormat)
 		}
```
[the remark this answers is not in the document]
N-o-Z: Added

pkg/kv/import.go
N-o-Z: Since there are 10 consumers and 1 producer I see it highly unlikely that we will reach busy wait here. Unless you have another perspective?

N-o-Z:
> Thanks @N-o-Z - looks excellent
> 
> Can we had tests to verify that fix? Would that be easy to achieve?
> 
> In addition, 2 suggestions below. Up to you if to accept ort not. Other than that - Looks Great rocket

Thanks for the review - I added a check for the ctx Done on the consumer too. I tested this fix manually - but I cannot see how we can add tests to it in the current timeframe. This setup requires to create a customized docker pool.

pkg/kv/import.go
itaiad200: Channel prevents "busy waiting". Whether it's a writer blocked on a full channel, or a reader blocked on an empty channel - it's not busy waiting

itaiad200, line 93: I think there's a non-blocking non-busy-waiting way of achieving the same
```suggestion
	for {
		  i++
		  entry := new(Entry)
		  err := jd.Decode(entry)
		  if errors.Is(err, io.EOF) {
			  close(jobChan)
			  return nil
		  }
		  // Decode does not return error on missing data / incompatible format
		  if err != nil {
			  cancel()
			  return fmt.Errorf("decoding entry: %w", err)
		  }
		  if i%100_000 == 0 {
			  log.Infof("Migrated %d entries", i)
		  }
		select {
		case <-ctx.Done():
			return nil
		case jobChan <- entry:
		        // nothing to do
```

itaiad200 (changes requested):

pkg/kv/import.go
itaiad200: Excuse my formatting

N-o-Z:
Added requested changes

itaiad200 (approved):
Looks great!

Issue #3586, opened 2022-06-29
KV Graveler Implementation - Lock-Free Commits


itaidavid:
The real deal - full blown lock-free commits implementation, as designed in 
(https://github.com/treeverse/lakeFS/blob/b3204edad00f88f8eb98524ad940fde96e02ab0a/design/open/metadata_kv/index.md#graveler-metadata---branches-and-staged-writes) and discussed in https://github.com/treeverse/lakeFS/pull/3566
Remove the commits exclusion, introduced by #3569

PR #4089, opened 2022-09-03
optimistic locking and transactionless exaplained

itai-david:


docs/reference/database-migration.md
```diff
@@ -27,4 +27,31 @@ Under the hood, our KV implementation relies on a backing DB, which persists the
...
+In order to overcome this obstacle we introduced a per-`Repository`-partition, which holds all repository related objects (the `Branch` and `Commit` in this scenario). The partition ket can only be derived from the `Repository` object it self. In addition we reversed the order of objects creations, where the `Commit` and the `Branch` are created, under the `Repository`'s partition key, and the `Repository` is created last. The `Repository`, and as a result its partition, will be accessible only after a successful creation of all 3 entities, and every failure might leave some dangling objects, but the consistency remains in place, as these objects are not reachable.
```
N-o-Z, line 56:
```suggestion
To mitigate this we introduced a per-`Repository`-partition, which holds all repository related objects (the `Branch` and `Commit` in this scenario). The partition key can only be derived from the specific`Repository` instance itself. In addition we first create the `Repository` objects, the `Commit` and the `Branch`, under the `Repository`'s partition key, and then the `Repository` is created. The `Repository` and its objects will be accessible only after a successful creation of all 3 entities. A failure in this flow might leave some dangling objects, but consistency is maintained.
```

N-o-Z (commented):
Nice work! some suggestions

docs/reference/database-migration.md
```diff
@@ -27,4 +27,31 @@ Under the hood, our KV implementation relies on a backing DB, which persists the
...
\ No newline at end of file
+For further reading, please refer to our [KV Design](https://github.com/treeverse/lakeFS/blob/master/design/accepted/metadata_kv/index.md)
+
+### Optimistic Locking with KV
+
+One important key difference between SQL databases and Key Value Store is the ability to lock resources. While this is a common practice with relational databases, Key Value stores lack this ability. Let us take a look at a common lakeFS flow, a `Commit`. During this flow, several ref store operations are performed:
+* All relevant (`Branch` correlated) uncommitted objects are collected and marked as committed 
+* A new `Commit` object is created
+* The relevant `Branch` is updated to point to the new commit
+The `Commit` flow includes multiple database accesses and modifications, and is very sensitive to concurrent executions: If 2 `Commit` flows run in parallel, we cannot control which one will update the `Branch` last and we cannot guarantee correctness of the data. `lakeFS` with PostgreSQL solved it simply by locking the `Branch` for the entire `Commit` operation, which prevented concurrent execution of such flows.
```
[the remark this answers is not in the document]
arielshaqed: Perhaps we can rephrase positively? Something like
```suggestion
The `Commit` flow includes multiple database accesses and modifications, and is very sensitive to concurrent executions: If 2 `Commit` flows run in parallel, we must guarantee correctness of the data. `lakeFS` with PostgreSQL simply locks the `Branch` for the entire `Commit` operation, preventing concurrent execution of such flows.
```
And then explaining what kv does makes it clear it's an alternative.

docs/reference/database-migration.md
```diff
@@ -27,4 +27,31 @@ Under the hood, our KV implementation relies on a backing DB, which persists the
...
+2. Both `Commits` interfere with each other, and both will fail, as  depicted in the following scenario:
+  * `Commit` A sets the `StagingToken` to val1 and samples the `Branch`,
+  * `Commit` B sets the `StagingToken` to val2 and samples the `Branch`,
+  * `Commit` A finishes, tries to update the `Branch` and fails due to the recent modification by `Commit` B,
+  * `Commit` A is retired, setting `StagingToken` to val3 and samples the `Branch`
+  * `Commit` B finishes, tries to update the `Branch` and fails due to the recent modification by `Commit` A,
+  * And so on...
+At some point, as retries will be exhausted, commits will succeed, if not earlier, but this was definitely something that had to be taken into consideration when designing our KV Store. You can read more on the Commit Flow in the [dedicated section in the KV Design](https://github.com/treeverse/lakeFS/blob/master/design/accepted/metadata_kv/index.md#graveler-metadata---branches-and-staged-writes)
```
arielshaqed, line 48: When phrased like this it is very worrying, and makes it sound as though workflows that attempt to commit periodically will never work.
At the very least I would explain that commits are idempotent, so as soon as one succeeds we advance, and other commits only need to commit work started after that successful commit.
Otherwise starvation would be inevitable.

arielshaqed (commented):
Great stuff! It does the hard job of explaining to a non-expert what kv is all about. I think it will help people understand the new system. 
I am not sure that this is the right page for it. Think of an engineer who is trying to migrate to kv. This should be the least of their worries! There is no action item here, it's an explanation of how everything works and why it's not trivial. I might want to read it if I had specific worries or confusion, but that's no reason to make me think I have to understand it to be able to upgrade to kv. 
I think it deserves its own page, maybe "kv in brief". And migration should link to it "for further details".
Also we might want to add why it is *better*, not just how it reaches feature parity.

itaiad200:
I think this deserves a rebase, it has some unrelated changes and conflicting files.

docs/reference/database-migration.md
```diff
@@ -27,4 +27,31 @@ Under the hood, our KV implementation relies on a backing DB, which persists the
...
+The `Commit` flow includes multiple database accesses and modifications, and is very sensitive to concurrent executions: If 2 `Commit` flows run in parallel, we cannot control which one will update the `Branch` last and we cannot guarantee correctness of the data. `lakeFS` with PostgreSQL solved it simply by locking the `Branch` for the entire `Commit` operation, which prevented concurrent execution of such flows.
+Now, with KV Store replacing the SQL database, this easy solution is no longer available. Instead, we implemented an [Optimistic Locking](https://en.wikipedia.org/wiki/Optimistic_concurrency_control) algorithm, which leverages the KV Store `Compare-and-Set` (`CaS`) functionality to remember the `Branch` state at the beginning of the `Commit` flow, and updating the branch at the end, only if it remains unchanged, using `CaS`. Before sampling the `Branch` another update, of the `Branch`'s `StagingToken`, is made, and so, if another `Commit` is already running in parallel, it will fail the final step, of updating the `Branch`, as it was already changed. If 2 commits do run in parallel there are only 2 possible outcomes:
```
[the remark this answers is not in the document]
itai-david: Fixed

docs/reference/database-migration.md
itai-david: @arielshaqed - AFAICT, starvation is possible (probably not inevitable).
The best thing I could come up with is the remove the last bullet "and so on..." which implies endlessness, and change the next line to:
"The first commit to succeed, releases this mutual interference, and the next commits can continue uninterrupted."

WDYT?
I wuold love to adopt any better suggestion

CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
 # Changelog
 
+## v0.80.0 - 2022-08-31
+This release requires running database migration.
+The lakeFS service will not run if the migration version isn't compatible with the binary.
+Before running the new version you will be required to run migrate, with the new version. 
+Please refer to this [upgrade documentation](https://docs.lakefs.io/reference/upgrade.html##lakefs-0800-or-greater-kv-migration) for more information on the specific migration to KV
+
+This is the first lakeFS version over Key-Value Store
```
arielshaqed, line 9:
```suggestion
⚠️  This is the first lakeFS version over Key-Value Store ⚠️ 
```

docs/reference/database-migration.md
```diff
@@ -0,0 +1,58 @@
...
+With no transaction in KV Store, if for example the `Branch` creation fails, it will leave the `Repository` without an initial `Branch` (or a `Branch` at all), yet the `Repository` will be accessible. Trying to delete the `Repository` as a response to `Branch` creation failure is ony a partial solution as this operation can fail as well.
+To mitigate this we introduced a per-`Repository`-partition, which holds all repository related objects (the `Branch` and `Commit` in this scenario). The partition key can only be derived from the specific`Repository` instance itself. In addition we first create the `Repository` objects, the `Commit` and the `Branch`, under the `Repository`'s partition key, and then the `Repository` is created. The `Repository` and its objects will be accessible only after a successful creation of all 3 entities. A failure in this flow might leave some dangling objects, but consistency is maintained.
+The amount of such dangling objects is not expected to be significant, and we plan to implement a cleaning algorithm to keep our KV Store neat and clean
```
arielshaqed, line 58: Consider removing this.  It depends on the workload.  Workloads that create many temporary objects where they write, or that write and then overwrite existing objects, will generate dangling objects.  It is hard for us to know what users run.

arielshaqed (commented):
Great measurements (and explanations too, of course)!

docs/reference/database-migration.md
arielshaqed: Either leave it alone ("eventually a commit succeeds"), or point to #3586.  The current makes me feel really bad -- it gives little hope, but also doesn't say we know how to fix it when that becomes necessary.  So I'd prefer to go in either direction.

docs/reference/database-migration.md
itai-david: @arielshaqed - this one is specifically reagrding a failure with repository creation, in which we can have a dangling branch and commit, at most. It also relies on the (maybe wrong) assumptions that repository creation is not something that users do periodically. 
Is that assumption wrong?
Do you still think we are better off without this last sentence?
Thanks

docs/reference/database-migration.md
itai-david: @arielshaqed - rephrased as we eariler discessed

itai-david:
@arielshaqed @N-o-Z - Thank you for your feedback
Please review the updated version
Thank you

docs/understand/kv-in-a-nutshell.md
```diff
@@ -0,0 +1,51 @@
...
+
+In order to store ref store objects (that is `Repositories`, `Branches`, `Commits`, `Tags`, and `Uncommitted Objects`), lakeFS implements another layer over the generic KV Store, which supports serialization and deserialization of these objects as [protobuf](https://en.wikipedia.org/wiki/Protocol_Buffers). As this layer relies on the generic interface of the KV Store layer, it is totally agnostic to whichever store implementation is in use, gaining our users the maximal flexibility
+
+For further reading, please refer to our [KV Design](https://github.com/treeverse/lakeFS/blob/master/design/accepted/metadata_kv/index.md)
+
+## Optimistic Locking with KV
+
+One important key difference between SQL databases and Key Value Store is the ability to lock resources. While this is a common practice with relational databases, Key Value stores lack this ability. Let us take a look at a common lakeFS flow, `Commit`, during which several ref store operations are performed:
```
N-o-Z, line 27: They don't lack the ability but it is a less common practice I think

docs/understand/kv-in-a-nutshell.md
```diff
@@ -0,0 +1,51 @@
...
+
+Under the hood, our KV implementation relies on a backing DB, which persists the data. Theoretically, it could be any type of database and out of the box, we already implemented drivers for [DynamoDB](https://en.wikipedia.org/wiki/Amazon_DynamoDB), for AWS users, and [PostgreSQL](https://en.wikipedia.org/wiki/PostgreSQL), using its relational nature to store a KV Store. More databases will be supported in the future, and lakeFS users and contributors can develop their own driver to use their own favorite database. For experimenting purposes, an in-memory KV store can be used, though it obviously lack the persistency aspect
+
+In order to store ref store objects (that is `Repositories`, `Branches`, `Commits`, `Tags`, and `Uncommitted Objects`), lakeFS implements another layer over the generic KV Store, which supports serialization and deserialization of these objects as [protobuf](https://en.wikipedia.org/wiki/Protocol_Buffers). As this layer relies on the generic interface of the KV Store layer, it is totally agnostic to whichever store implementation is in use, gaining our users the maximal flexibility
```
N-o-Z, line 21: This is the first time in the document referring to "ref store", up until now we discussed only a key-value store. Might want to clarify the terminology

docs/understand/kv-in-a-nutshell.md
```diff
@@ -0,0 +1,51 @@
...
+* A new `Commit` object is created
+* The relevant `Branch` is updated to point to the new commit
+
+The `Commit` flow includes multiple database accesses and modifications, and is very sensitive to concurrent executions: If 2 `Commit` flows run in parallel, we must guarantee correctness of the data. `lakeFS` with PostgreSQL simply locks the `Branch` for the entire `Commit` operation, preventing concurrent execution of such flows.
+Now, with KV Store replacing the SQL database, this easy solution is no longer available. Instead, we implemented an [Optimistic Locking](https://en.wikipedia.org/wiki/Optimistic_concurrency_control) algorithm, which leverages the KV Store `Compare-And-Set` (`CAS`) functionality to remember the `Branch` state at the beginning of the `Commit` flow, and updating the branch at the end, only if it remains unchanged, using `CAS`, with the former `Branch` state, used as a comparison criteria. If the sampled `Branch` state and the current state differ, it could only mean that another, later, `Commit` is in progress, causing the current `Commit` to fail, and give the later `Commit` a chance to complete.
```
N-o-Z, line 33: The second part of this paragraph could use some refactoring. I think referring to a 'later' commit might be a bit confusing. might want to talk about commit A and B instead?

docs/understand/kv-in-a-nutshell.md
```diff
@@ -0,0 +1,51 @@
...
+With no transaction in KV Store, if for example the `Branch` creation fails, it will leave the `Repository` without an initial `Branch` (or a `Branch` at all), yet the `Repository` will be accessible. Trying to delete the `Repository` as a response to `Branch` creation failure is ony a partial solution as this operation can fail as well.
+To mitigate this we introduced a per-`Repository`-partition, which holds all repository related objects (the `Branch` and `Commit` in this scenario). The partition key can only be derived from the specific`Repository` instance itself. In addition we first create the `Repository` objects, the `Commit` and the `Branch`, under the `Repository`'s partition key, and then the `Repository` is created. The `Repository` and its objects will be accessible only after a successful creation of all 3 entities. A failure in this flow might leave some dangling objects, but consistency is maintained.
+The amount of such dangling objects is not expected to be significant, and we plan to implement a cleaning algorithm to keep our KV Store neat and clean
```
N-o-Z, line 51: Maybe add a summary at the end - and finish with a more positive note 😄

N-o-Z (changes requested):
Looks much better, what I feel missing is a bottom line. Suggest adding a summary at the end, maybe listing changes / pros-cons in bullets or something similar

docs/understand/kv-in-a-nutshell.md
itai-david: I believe KV Stores (at least DynamoDB) are connectionless, and as such cannot support locking (I think a lock should be bound to a connection, otherwise , in case of disconnection, no one will be able to unlock it)

docs/understand/kv-in-a-nutshell.md
itai-david: ref-store rephrased as `metadata` to avoid any confusion. I would rather not add an explanation of what is ref-store, in that scope

docs/understand/kv-in-a-nutshell.md
itai-david: The commit A and B are used in the step-by-step example right after. The above is a general depction of a scenario that is later specified in greater detail

docs/reference/database-migration.md
arielshaqed: Thanks for the explanation!  You're right (of course).

docs/understand/kv-in-a-nutshell.md
```diff
@@ -0,0 +1,51 @@
...
```
arielshaqed, line 51: Nits:
```suggestion
The number of such dangling objects is not expected to be significant, and we plan to implement a cleaning algorithm to keep our KV Store neat and clean.
```
("Objects" are numerable.)

arielshaqed (approved):
THANKS!

docs/understand/kv-in-a-nutshell.md
N-o-Z: I think we shouldn't state something as a fact: "While this is a common practice with relational databases, Key Value stores lack this ability" if we are uncertain of it

itai-david: You convinced me. Fixed

docs/understand/kv-in-a-nutshell.md
itai-david: I'm not sure I agree with this approach. This is a technical page and I tried to describe design challenges we were facing, and the solutions we came up with. I don't think it needs to end "happily", as this doc is not coming to convivce anyone of anything, IMO - it comes to provide an explanation
WDYT?

N-o-Z: I'll try to rephrase:
This document contains a lot of technical information and is very verbose. At the end of the document - the reader might be confused as to what exactly are the changes that were described in the document.
So I feel it would benefit the reader to get a concise summary of the changes in the form or bullets or otherwise, at the beginning of the document or at the end

N-o-Z (approved):
@itai-david approved, but please take a look at the last comment

docs/understand/kv-in-a-nutshell.md
itai-david: Added

PR #4093, opened 2022-09-04
Simplify kv import

### LINKED ISSUE #4103: KV simplify import implementation

nopcoder:
- Reader not as part of go routine
- Reader check for returned error and return

Close #4103

### DIFF
--- a/pkg/kv/import.go
+++ b/pkg/kv/import.go
@@ -116,10 +116,10 @@ func Import(ctx context.Context, reader io.Reader, store Store) error {
 			return nil
 		})
 	}
-	err := importReader(ctx, log, entryChan, jd)
+	readerErr := importReader(ctx, log, entryChan, jd)
 	close(entryChan)
 
-	merr := g.Wait()
-	merr = multierror.Append(merr, err)
-	return merr.ErrorOrNil()
+	err := g.Wait()
+	err = multierror.Append(err, readerErr)
+	return err.ErrorOrNil()
 }
--- a/pkg/kv/import.go
+++ b/pkg/kv/import.go
@@ -43,50 +43,27 @@ func (e *SafeEncoder) Encode(v interface{}) error {
 	return nil
 }
 
-func consumer(ctx context.Context, cancel context.CancelFunc, jobChan <-chan *Entry, store Store) error {
-	for {
-		select {
-		case e, more := <-jobChan:
-			if !more {
-				return nil
-			}
-			if len(e.PartitionKey) == 0 {
-				cancel()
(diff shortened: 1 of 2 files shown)


pkg/kv/import.go
```diff
@@ -127,16 +103,35 @@ func Import(ctx context.Context, reader io.Reader, store Store) error {
...
+	err := importReader(ctx, log, entryChan, jd)
+	close(entryChan)
+
+	wg.Wait()
 
-	return g.Wait().ErrorOrNil()
+	var merr *multierror.Error
+	merr = multierror.Append(merr, err)
+	merr = multierror.Append(merr, workerErr...)
+	return merr.ErrorOrNil()
```
N-o-Z, line 136: multiErr better?

pkg/kv/import.go
```diff
@@ -127,16 +103,35 @@ func Import(ctx context.Context, reader io.Reader, store Store) error {
 	})
 	log.Info("Processing file")
 
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
N-o-Z, line 107: Why do we need this?

N-o-Z (approved):
Looks great!
1 question, 1 comment

pkg/kv/import.go
nopcoder: This is the same type used by `m.Go` - we can revert part of my change to use it and simplify it a bit more

pkg/kv/import.go
nopcoder: prevent resource leak

pkg/kv/import.go
N-o-Z: I was referring to the variable name

pkg/kv/import.go
nopcoder: https://pkg.go.dev/context#WithCancel
with the example as good resource

pkg/kv/import.go
```diff
@@ -127,16 +99,27 @@ func Import(ctx context.Context, reader io.Reader, store Store) error {
...
 		})
 	}
-	g.Go(func() error {
-		return producer(cctx, cancel, log, entryChan, jd)
-	})
+	err := importReader(ctx, log, entryChan, jd)
+	close(entryChan)
 
-	return g.Wait().ErrorOrNil()
+	merr := g.Wait()
```
N-o-Z, line 122: Can you just change the variable name

N-o-Z (approved):
Thanks

pkg/kv/import.go
nopcoder: Updated

PR #4106, opened 2022-09-04
Cleanup env before esti run

idanovo:


esti/main_test.go
```diff
@@ -54,13 +69,204 @@ var (
...
+	}
+	return err
+}
+
 func TestMain(m *testing.M) {
 	systemTests := flag.Bool("system-tests", false, "Run system tests")
 	useLocalCredentials := flag.Bool("use-local-credentials", false, "Generate local API key during `lakefs setup'")
 	adminAccessKeyID := flag.String("admin-access-key-id", DefaultAdminAccessKeyId, "lakeFS Admin access key ID")
 	adminSecretAccessKey := flag.String("admin-secret-access-key", DefaultAdminSecretAccessKey, "lakeFS Admin secret access key")
+	cleanupEnv := flag.Bool("cleanup-env-pre-run", false, "Clean repositories, groups, users and polices before sunning esti tests")
```
N-o-Z, line 264:
```suggestion
	cleanupEnv := flag.Bool("cleanup-env-pre-run", false, "Clean repositories, groups, users and polices before running esti tests")
```

esti/main_test.go
```diff
@@ -86,6 +292,15 @@ func TestMain(m *testing.M) {
 	logger, client, svc = testutil.SetupTestingEnv(&params)
 
 	var err error
+	if *cleanupEnv {
```
N-o-Z, line 295: I think it makes more sense to put the call to `envCleanup` in `SetupTestingEnv` inside the else block of `if setupLakeFS `

esti/main_test.go
```diff
@@ -15,7 +18,10 @@ import (
 	"github.com/treeverse/lakefs/pkg/testutil"
 )
 
-type Booleans []bool
+type (
+	Booleans   []bool
+	arrayFlags []string
```
N-o-Z, line 23: Maybe I've missed it - but why do you need to wrap it in a type with dedicated functions?
I couldn't find any user for Set, or String - which interface is this suppose to implement?

N-o-Z (changes requested):
I like the direction you took. This looks great!
Couple of comments

esti/main_test.go
idanovo: Not really TBH

esti/main_test.go
idanovo: 1. to clean up my env, we create one only after this if-else statement.
2. If someone uses the setup lakeFS env, he shouldn't use the cleanup flag too, it doesn't relevant for him.

N-o-Z: So we should protect against using the ESTI_SETUP_LAKEFS env var and the cleanup flag. But I think it will be more logical to just put it in the setup flow

esti/main_test.go
idanovo: Yes, the `flag.Var` function, gets a "value" interface as its first parameter and this interface should implement 	`String() string` and `Set(string) error` functions

pkg/testutil/setup.go
```diff
@@ -31,7 +31,7 @@ type SetupTestingEnvParams struct {
 	AdminSecretAccessKey string
 }
 
-func SetupTestingEnv(params *SetupTestingEnvParams) (logging.Logger, api.ClientWithResponsesInterface, *s3.S3) {
+func SetupTestingEnv(params *SetupTestingEnvParams) (logging.Logger, api.ClientWithResponsesInterface, *s3.S3, bool) {
```
N-o-Z, line 34: You don't need to return the value - you can just get it from viper after the setup

N-o-Z (changes requested):
One comment

pkg/testutil/setup.go
idanovo: I know, I thought it will make more sense to get it once
will change if you think it make more sense to get it from viper

N-o-Z: I think so, since you already have viper and it saves an unnecessary return parameter

esti/main_test.go
```diff
@@ -85,7 +291,17 @@ func TestMain(m *testing.M) {
 
 	logger, client, svc = testutil.SetupTestingEnv(&params)
 
+	setupLakeFS := viper.GetBool("setup_lakefs")
 	var err error
```
N-o-Z, line 295: Should be directly above usage

N-o-Z (approved):
LGTM!

One comment

PR #4109, opened 2022-09-05
checking itr type for nil

itai-david:
Tested manually by running in debug and forcing an error from SeekGE.

### DIFF
--- a/pkg/kv/iterators.go
+++ b/pkg/kv/iterators.go
@@ -299,7 +299,7 @@ func (p *PartitionIterator) Next() bool {
 func (p *PartitionIterator) SeekGE(key []byte) {
 	if p.Err() == nil {
 		if !p.itrClosed {
-			p.itr.Close() // Close previous before creating new iterator
+			p.Close() // Close previous before creating new iterator
 		}
 		p.itr, p.err = p.store.Scan(p.ctx, []byte(p.partitionKey), key)
 		p.itrClosed = p.err != nil
--- a/pkg/kv/iterators.go
+++ b/pkg/kv/iterators.go
@@ -5,7 +5,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	reflect "reflect"
 
 	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/reflect/protoreflect"
@@ -247,6 +246,7 @@ type PartitionIterator struct {
 	ctx          context.Context
 	store        Store
 	msgType      protoreflect.MessageType
+	itrClosed    bool
 	itr          EntriesIterator
 	partitionKey string
 	value        *MessageEntry
@@ -258,6 +258,7 @@ func NewPartitionIterator(ctx context.Context, store Store, msgType protoreflect
 		ctx:          ctx,
 		store:        store,
 		msgType:      msgType,
+		itrClosed:    true,
 		partitionKey: partitionKey,
 	}
 }
@@ -270,9 +271,11 @@ func (p *PartitionIterator) Next() bool {
 	if p.itr == nil {
 		p.itr, p.err = p.store.Scan(p.ctx, []byte(p.partitionKey), []byte(""))
 		if p.Err() != nil {
+			p.itrClosed = true
 			return false
 		}
 	}
+	p.itrClosed = false
 	if !p.itr.Next() {
 		return false
 	}
@@ -300,6 +303,9 @@ func (p *PartitionIterator) SeekGE(key []byte) {
 			p.itr.Close() // Close previous before creating new iterator
 		}
 		p.itr, p.err = p.store.Scan(p.ctx, []byte(p.partitionKey), key)
+		if p.err != nil {
+			p.itrClosed = true
+		}
 	}
 }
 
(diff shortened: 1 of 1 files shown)


N-o-Z (approved):
I think this is good enough for now - but we should try to find a more sustainable solution for the underlying nil != interface nil  problem

nopcoder (changes requested):
Reflect usually cost - can we manage this state by bool indicator?

pkg/kv/iterators.go
```diff
@@ -269,9 +271,11 @@ func (p *PartitionIterator) Next() bool {
 	if p.itr == nil {
```
nopcoder, line 271: check the bool here

pkg/kv/iterators.go
```diff
@@ -269,9 +271,11 @@ func (p *PartitionIterator) Next() bool {
...
 		p.itr, p.err = p.store.Scan(p.ctx, []byte(p.partitionKey), []byte(""))
 		if p.Err() != nil {
```
nopcoder, line 273: check p.err = no need to call Err in this case

pkg/kv/iterators.go
```diff
@@ -269,9 +271,11 @@ func (p *PartitionIterator) Next() bool {
...
+			p.itrClosed = true
```
nopcoder, line 274: already closed at this point - need to turn it off after the return

pkg/kv/iterators.go
```diff
@@ -269,9 +271,11 @@ func (p *PartitionIterator) Next() bool {
...
 			return false
 		}
 	}
+	p.itrClosed = false
```
nopcoder, line 278: can be pushed up the scope

pkg/kv/iterators.go
```diff
@@ -299,6 +303,9 @@ func (p *PartitionIterator) SeekGE(key []byte) {
 			p.itr.Close() // Close previous before creating new iterator
```
nopcoder, line 303: above this line - a nil check - need to check the bool

pkg/kv/iterators.go
```diff
@@ -299,6 +303,9 @@ func (p *PartitionIterator) SeekGE(key []byte) {
...
 		}
 		p.itr, p.err = p.store.Scan(p.ctx, []byte(p.partitionKey), key)
+		if p.err != nil {
+			p.itrClosed = true
+		}
```
nopcoder, line 308:
```suggestion
		p.itrClosed = p.err != nil
```

pkg/kv/iterators.go
```diff
@@ -317,8 +324,8 @@ func (p *PartitionIterator) Err() error {
 }
 
 func (p *PartitionIterator) Close() {
-	// check itr is set, can be null in case seek fails
-	if p.itr != nil {
+	// Check itr is set, can be null in case seek fails
+	if !p.itrClosed && p.itr != nil {
 		p.itr.Close()
 	}
```
nopcoder, line 330: prefer to manage just one state - don't want to reference a option to nil
```suggestion
	// Check itr is set, can be null in case seek fails
	if !p.itrClosed {
		p.itr.Close()
		p.iterClosed = true
	}
```

nopcoder (changes requested):

pkg/kv/iterators.go
itai-david: Why not check both?

pkg/kv/iterators.go
itai-david: @nopcoder Can you please explain?

pkg/kv/iterators.go
itai-david: Done

pkg/kv/iterators.go
itai-david: Done

pkg/kv/iterators.go
itai-david: Done

pkg/kv/iterators.go
itai-david: Done

pkg/kv/iterators.go
itai-david: Done

pkg/kv/iterators.go
```diff
@@ -295,10 +298,11 @@ func (p *PartitionIterator) Next() bool {
 
 func (p *PartitionIterator) SeekGE(key []byte) {
 	if p.Err() == nil {
-		if p.itr != nil {
+		if !p.itrClosed {
 			p.itr.Close() // Close previous before creating new iterator
```
nopcoder, line 302: can you set itrClosed to true here - to call Close()

pkg/kv/iterators.go
```diff
@@ -317,8 +321,9 @@ func (p *PartitionIterator) Err() error {
 }
 
 func (p *PartitionIterator) Close() {
-	// check itr is set, can be null in case seek fails
-	if p.itr != nil {
+	// Check itr is set, can be null in case seek fails
+	if !p.itrClosed {
+		p.itrClosed = true
```
nopcoder, line 326: can you place it after the call to Close

nopcoder (approved):
lgtm

PR #4110, opened 2022-09-05
flatMap range contents directly

arielshaqed:
Don't materialize them in memory, just shove them out to Spark.


johnnyaug (approved):
Nice!

arielshaqed:
Reverted [a69e437](https://github.com/treeverse/lakeFS/pull/4110/commits/a69e437e6798d3757d7f7dafa36122a982d01f44), [54bdef0](https://github.com/treeverse/lakeFS/pull/4110/commits/54bdef0b7638c9b7d79d2a3c0a687af1e2960098) as Spark cannot plan with Iterator types.  (But that also means the conversion to Seq isn't so bad, it has to happen anyway.)

Will pull if tests pass, direct flatMap should be the bigger win here anyway.

arielshaqed:
@johnnyaug PTAL :-/

arielshaqed:
Thanks!

PR #4111, opened 2022-09-05
test: remove unused println from test code

nopcoder:
- remove print values in test code
- remove values from test code


guy-har (approved):
🙏

PR #4112, opened 2022-09-05
Test: Create KV test driver with callback. Add test for nil interface

N-o-Z:


pkg/kv/testdriver/store.go
```diff
@@ -0,0 +1,112 @@
+package testdriver
```
itaiad200, line 1: I think you reimplemented the code of mockgen, it gives you exactly what you need here.

itaiad200 (changes requested):
I suggest giving mockgen a spin before creating our own mock-with-callbacks implementations.

itaiad200 (changes requested):
I suggest giving mockgen a spin before creating our own mock-with-callbacks implementations.

pkg/kv/kvtest/store.go
```diff
@@ -869,5 +870,5 @@ func GetStore(ctx context.Context, t testing.TB) kv.Store {
 	t.Cleanup(func() {
 		store.Close()
 	})
-	return store
+	return testdriver.NewTestStore(store)
```
eden-ohana, line 873: why not test the actual store?

pkg/kv/iterators_test.go
```diff
@@ -0,0 +1,49 @@
...
+	err := sm.SetMsg(ctx, modelPartitionKey, setModel.Name, setModel)
+	if err != nil {
+		t.Fatal("failed to set model", err)
+	}
+
+	itr := kv.NewPartitionIterator(ctx, store, (&kvtest.TestModel{}).ProtoReflect().Type(), modelPartitionKey)
+	require.NotNil(t, itr)
+	itr.Close()
+
+	itr = kv.NewPartitionIterator(ctx, store, (&kvtest.TestModel{}).ProtoReflect().Type(), modelPartitionKey)
```
eden-ohana, line 34: duplication?

N-o-Z:
- Resolved in #4118

PR #4117, opened 2022-09-05
s3 block adapter: enable skip verify certificate of storage endpoint

### LINKED ISSUE #4153: S3 adapter https certificate check disable for testing
Working with custom s3 storage system, you sometimes need to work with self signed certificates.
In order to simplify testing working with such storage or for local development it will be good to enable skip certificate verification.
This option should be marked for test only and be used only for this test case.

nopcoder:
Fix #4153


arielshaqed (changes requested):
I am puzzled by this feature. Enabling it explicitly violates the documentation of the crypto/tls package:

        This should be used only for
	// testing or in combination with VerifyConnection or VerifyPeerCertificate.
	InsecureSkipVerify bool

If users wish to use an unsigned transport, they should use a plaintext (http) url. Using https and not verifying the cert disables the ability of an organisation to verify encryption: it *looks* encrypted but is not.

Blocking as we will need to document why this is OK, or our users will have a hard time certifying lakeFS.

nopcoder:
> I am puzzled by this feature. Enabling it explicitly violates the documentation of the crypto/tls package

It is for testing / local development when the storage endpoint uses https with self signed certificate.

nopcoder:
Will have more information on an associated issue to describe the use-case

nopcoder:
Updated the configuration and parameter name to include 'test only'.

arielshaqed (approved):
Thanks!  This naturally resolves all concerns that I have, and explicitly follows the guidance in the crypto/tls godoc.

PR #4118, opened 2022-09-06
Improve how we handle kv iterator interface

### LINKED ISSUE #4164: Improve how kv iterator internal iterator state
KV iterator that handle internal iterator handle internal call to store that can be failed.
Calling Close or Err on invalid state of the iterator can cause an access violation.

nopcoder:
- Align how repository iterator use of MessageIterator handle when it is closed
- Align how tag iterator use of MessageIterator handle when it is closed
- kv/PartitionIterator Err should check itr closed flag and not interface
- Remove interface compare in store test

Close #4164

### DIFF
--- a/pkg/graveler/ref/commit_ordered_iterator.go
+++ b/pkg/graveler/ref/commit_ordered_iterator.go
@@ -18,30 +18,6 @@ type KVOrderedCommitIterator struct {
 	firstParents       map[string]bool
 }
 
-// getAllFirstParents returns a set of all commits that are not a first parent of other commit in the given repository.
-func getAllFirstParents(ctx context.Context, store *kv.StoreMessage, repo *graveler.RepositoryRecord) (map[string]bool, error) {
-	it, err := kv.NewPrimaryIterator(ctx, store.Store, (&graveler.CommitData{}).ProtoReflect().Type(),
-		graveler.RepoPartition(repo),
-		[]byte(graveler.CommitPath("")), kv.IteratorOptionsFrom([]byte("")))
-	if err != nil {
-		return nil, err
-	}
-	defer it.Close()
-	firstParents := make(map[string]bool)
-	for it.Next() {
-		entry := it.Entry()
-		commit := entry.Value.(*graveler.CommitData)
-		if len(commit.Parents) > 0 {
-			if graveler.CommitVersion(commit.Version) < graveler.CommitVersionParentSwitch && len(commit.Parents) > 1 {
-				firstParents[commit.Parents[1]] = true
-			} else {
-				firstParents[commit.Parents[0]] = true
-			}
-		}
-	}
-	return firstParents, nil
-}
-
 // NewKVOrderedCommitIterator returns an iterator over all commits in the given repository.
 // Ordering is based on the Commit ID value.
 // WithOnlyAncestryLeaves causes the iterator to return only commits which are not the first parent of any other commit.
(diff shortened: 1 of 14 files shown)


pkg/graveler/ref/commit_ordered_iterator.go
```diff
@@ -113,14 +89,42 @@ func (i *KVOrderedCommitIterator) Value() *graveler.CommitRecord {
...
 		return i.it.Err()
 	}
-	return i.err
+	return nil
 }
 
 func (i *KVOrderedCommitIterator) Close() {
 	if i.it != nil {
 		i.it.Close()
+		i.it = nil
```
N-o-Z, line 104: We basically want that for all iterators, no?

nopcoder: The ones that holds internal pointer to implementation - I'll map the rest - I think I went over the pkg/kv in this pr.

N-o-Z (changes requested):
Thanks for adding this code.
I would love to see some tests that cover the additional logic

nopcoder:
@N-o-Z added specific tests that would fail before the above change.

pkg/graveler/ref/commit_ordered_iterator_test.go
```diff
@@ -8,6 +8,12 @@ import (
 	"testing"
 	"time"
 
+	"github.com/treeverse/lakefs/pkg/kv"
+
```
N-o-Z, line 12: Fix imports

pkg/graveler/ref/repository_iterator_test.go
```diff
@@ -5,6 +5,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/golang/mock/gomock"
+	"github.com/treeverse/lakefs/pkg/kv"
+	"github.com/treeverse/lakefs/pkg/kv/mock"
+
```
N-o-Z, line 11: New line

pkg/graveler/ref/tag_iterator_test.go
```diff
@@ -5,6 +5,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/golang/mock/gomock"
+	"github.com/treeverse/lakefs/pkg/kv"
+	"github.com/treeverse/lakefs/pkg/kv/mock"
+
```
N-o-Z, line 11: New line

pkg/kv/kvtest/store.go
```diff
@@ -462,12 +462,9 @@ func testStoreMissingArgument(t *testing.T, ms MakeStore) {
 	t.Run("Scan", func(t *testing.T) {
 		start := uniqueKey("test-missing-argument")
 		for _, pk := range [][]byte{nil, {}} {
-			it, err := store.Scan(ctx, pk, start)
+			_, err := store.Scan(ctx, pk, start)
 			if !errors.Is(err, kv.ErrMissingPartitionKey) {
-				t.Errorf("Scan using empty partition key - err=%v, expected %s", err, kv.ErrMissingPartitionKey)
-			}
-			if it != nil {
```
N-o-Z, line 469: Why did we remove this check?

pkg/kv/iterators_test.go
```diff
@@ -0,0 +1,29 @@
...
+	"context"
+	"testing"
+
+	"github.com/golang/mock/gomock"
+	"github.com/treeverse/lakefs/pkg/graveler"
+	"github.com/treeverse/lakefs/pkg/kv"
+	"github.com/treeverse/lakefs/pkg/kv/mock"
+)
+
+func TestPartitionIterator_ClosedBehaviour(t *testing.T) {
```
N-o-Z, line 13: If I understand correctly - this test doesn't reproduce the original issue?

N-o-Z (changes requested):
Thanks, see comments

pkg/graveler/ref/commit_ordered_iterator_test.go
nopcoder: Thanks - I need to fix this in my editor settings.

pkg/kv/kvtest/store.go
nopcoder: - Didn't want to check interface value with nil
- Assumed we test only valid protocol - in this case, when we call the function and there is an error, we can't assume anything about the iterator value.

pkg/kv/iterators_test.go
nopcoder: Wanted test what will happen when getting an error inside SeekGE calling to store's Scan func.
Wanted to address the following:
- When calling close we check the flag and not the pointer.
- Close mark that iterator was closed - prevent crash in calling multiple times to Close.
- Err implementation return the iterator's error based on flag.

nopcoder:
> Thanks, see comments

Thanks - I'll add another test to test SeekGE failed in case of PartitionIterator to cover the previous case.

pkg/kv/iterators_test.go
```diff
@@ -27,3 +28,16 @@ func TestPartitionIterator_ClosedBehaviour(t *testing.T) {
...
+	ctx := context.Background()
+	ctrl := gomock.NewController(t)
+	store := mock.NewMockStore(ctrl)
+	var entIt *mock.MockEntriesIterator
+	entItErr := errors.New("failed to scan")
+	store.EXPECT().Scan(ctx, gomock.Any(), gomock.Any()).Return(entIt, entItErr).Times(1)
+
+	it := kv.NewPartitionIterator(ctx, store, (&graveler.StagedEntryData{}).ProtoReflect().Type(), "partitionKey")
+	it.SeekGE([]byte("test"))
+	it.Close() // verify we don't crash after see failedcx
```
N-o-Z, line 42: see failedcx?

N-o-Z: Ahh I think I understand now, it's German for "the failed seek" 😆

nopcoder: fixed the comment

N-o-Z (approved):
lgtm, thanks!

pkg/kv/iterators_test.go
nopcoder: my the hands write something, while the head is thinking about the next line.

Issue #4123, opened 2022-09-06
GC: add debug helpers


johnnyaug:
## New GC flags
* `lakefs.debug.gc.max_commit_iso_datetime`: allow to run GC only on commits before this date
* `lakefs.debug.gc.no_delete`: run the GC without actually deleting anything

## Metadata client feature
* Allow creating a dataframe out of the repository's entire storage namespace

PR #4124, opened 2022-09-06
gc debug flags

johnnyaug:
Solves #4123

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
@@ -177,11 +177,7 @@ class ApiClient private (conf: APIConfigurations) {
     storageConfig.getBlockstoreType()
   }
 
-  /** Query lakeFS for a URL to the metarange of commitID of repoName and
-   *  translate that URL to use an appropriate Hadoop FileSystem.
-   */
-  def getMetaRangeURL(repoName: String, commitID: String): String = {
-    val commit = commitsApi.getCommit(repoName, commitID)
+  def getMetaRangeURL(repoName: String, commit: api.model.Commit): String = {
     val metaRangeID = commit.getMetaRangeId
     if (metaRangeID != "") {
       val metaRange = metadataApi.getMetaRange(repoName, metaRangeID)
@@ -192,6 +188,18 @@ class ApiClient private (conf: APIConfigurations) {
         .normalize()
         .toString
     } else ""
+  }
+
+  def getCommit(repoName: String, commitID: String): api.model.Commit = {
+    commitsApi.getCommit(repoName, commitID)
+  }
+  
+  /** Query lakeFS for a URL to the metarange of commitID of repoName and
+   *  translate that URL to use an appropriate Hadoop FileSystem.
+   */
+  def getMetaRangeURL(repoName: String, commitID: String): String = {
+    val commit = commitsApi.getCommit(repoName, commitID)
+    getMetaRangeURL(repoName, commit)
   }
 
   /** Query lakeFS for a URL to the range of rangeID of repoName and
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
@@ -6,7 +6,9 @@ import io.treeverse.clients.LakeFSContext.{
   LAKEFS_CONF_API_SECRET_KEY_KEY,
   LAKEFS_CONF_API_URL_KEY,
   LAKEFS_CONF_API_CONNECTION_TIMEOUT_SEC_KEY,
-  LAKEFS_CONF_API_READ_TIMEOUT_SEC_KEY
+  LAKEFS_CONF_API_READ_TIMEOUT_SEC_KEY,
+  LAKEFS_CONF_DEBUG_GC_NO_DELETE_KEY,
+  LAKEFS_CONF_DEBUG_GC_MAX_COMMIT_DATE_KEY
 }
 import org.apache.hadoop.fs._
 import org.apache.hadoop.conf.Configuration
@@ -23,6 +25,9 @@ import java.time.format.DateTimeFormatter
 import org.json4s.native.JsonMethods._
 import org.json4s._
 import org.json4s.JsonDSL._
+import org.apache.commons.lang3.StringUtils
+import java.time.LocalDate
+import java.time.ZoneId
 
 /** Interface to build an S3 client.  The object
  *  io.treeverse.clients.conditional.S3ClientBuilder -- conditionally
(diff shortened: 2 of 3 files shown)


clients/spark/.gitignore
```diff
@@ -1,3 +1,7 @@
 target/
 gen/
 .bsp/
+.bloop/
+.bsp/
+.metals/
+**/metals.sbt
```
arielshaqed, line 7: Yikes!  We _definitely_ don't want those.

Maybe move that to clients/.gitignore?

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -194,6 +190,18 @@ class ApiClient private (conf: APIConfigurations) {
...
+  }
+  
+  /** Query lakeFS for a URL to the metarange of commitID of repoName and
+   *  translate that URL to use an appropriate Hadoop FileSystem.
+   */
+  def getMetaRangeURL(repoName: String, commitID: String): String = {
+    val commit = commitsApi.getCommit(repoName, commitID)
+    getMetaRangeURL(repoName, commit)
+  }
+
```
arielshaqed, line 204: I prefer to keep overloads together.  Something like this, perhaps...

```suggestion
  /** Query lakeFS for a URL to the metarange of commitID of repoName and
   *  translate that URL to use an appropriate Hadoop FileSystem.
   */
  def getMetaRangeURL(repoName: String, commitID: String): String = {
    val commit = commitsApi.getCommit(repoName, commitID)
    getMetaRangeURL(repoName, commit)
  }

  def getCommit(repoName: String, commitID: String): api.model.Commit = {
    commitsApi.getCommit(repoName, commitID)
  }
```

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -23,6 +25,9 @@ import java.time.format.DateTimeFormatter
 import org.json4s.native.JsonMethods._
 import org.json4s._
 import org.json4s.JsonDSL._
+import org.apache.commons.lang3.StringUtils
```
arielshaqed, line 28: Why do we need this?  Scala provides its own StringUtils, _and_ makes it implicit on Strings:
```scala
val s = "the quick brown fox"
// val s: String = the quick brown fox

s startsWith("the qu")
// val res0: Boolean = true
```
(You don't need a dot to apply a method to an object in Scala, so the second line is either more readable or a bit scary...)

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -76,14 +82,22 @@ object GarbageCollector {
...
   ): Set[(String, Array[Byte], Array[Byte])] = {
-    val location = ApiClient
-      .get(apiConf)
-      .getMetaRangeURL(repo, commitID)
+    val conf = configurationFromValues(hcValues)
+    val apiClient = ApiClient.get(apiConf)
+    val commit = apiClient.getCommit(repo, commitID)
+    if (StringUtils.isNotBlank(conf.get(LAKEFS_CONF_DEBUG_GC_MAX_COMMIT_DATE_KEY))) {
+      val maxDate = LocalDate.parse(conf.get(LAKEFS_CONF_DEBUG_GC_MAX_COMMIT_DATE_KEY), DateTimeFormatter.ISO_DATE)
+      val maxDateEpoch = maxDate.atStartOfDay(ZoneId.of("UTC")).toInstant.toEpochMilli() / 1000
```
arielshaqed, line 90: Why not just parse a full ISO8601 datetime?  Then I can get a more accurate cutoff...

arielshaqed, line 90: This performs the conversion at every call to getRangeTuples.  I worry that it behave unpleasantly if I give it a bad date format.  Can we instead parse _once_ at startup, and then package up a time or millis-since-epoch or whatever into the executors?

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -386,17 +400,19 @@ object GarbageCollector {
     if (!storageNSForSdkClient.endsWith("/")) {
       storageNSForSdkClient += "/"
     }
+    
+    var removed : DataFrame = spark.emptyDataFrame
 
-    val removed =
-      remove(storageNSForSdkClient,
+    if (!hc.getBoolean(LAKEFS_CONF_DEBUG_GC_NO_DELETE_KEY, false)) {
```
arielshaqed, line 406: This line contains three negatives: `! ... NO_DELETE ... false`!  Maybe have the key name in positive form with default "true"?

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -386,17 +400,19 @@ object GarbageCollector {
     if (!storageNSForSdkClient.endsWith("/")) {
       storageNSForSdkClient += "/"
     }
+    
+    var removed : DataFrame = spark.emptyDataFrame
```
arielshaqed, line 404: More Scalaish (everything stays immutable):
```suggestion
    val removed = if (...) spark.emptyDataFrame else remove(...)
```

arielshaqed (approved):
Neat, and _very_ useful!

Some minor comments about having a config key with a negative name, and also about Scala style.  Nothing blocking or even major.

Thanks!

clients/spark/.gitignore
johnnyaug: Done

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
johnnyaug: Done

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
johnnyaug: Removed

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
johnnyaug: Done

johnnyaug: Done - passing it as configuration

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
johnnyaug: Don't want to change the configuration name, which I think is good, for the sake of simplifying the code.
However, after implementing your ternary suggestion, we are now left with a double negative.

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
johnnyaug: Done

arielshaqed (approved):
THANKS!

arielshaqed:
Can you correct the config key names on the issue, please?

PR #4125, opened 2022-09-06
Metadata Client: DataFrame out of entire storage namespace

### LINKED ISSUE #4123: GC: add debug helpers

johnnyaug:
Fixes #4123


clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSContext.scala
```diff
@@ -18,16 +18,20 @@ object LakeFSContext {
...
+  val LAKEFS_CONF_JOB_STORAGE_NAMESPACE_KEY = "lakefs.job.storage_namespace"
   val LAKEFS_CONF_JOB_COMMIT_ID_KEY = "lakefs.job.commit_id"
 
   def newRDD(
       sc: SparkContext,
       repoName: String,
-      commitID: String
+      storageNamespace: String,
+      commitID: String,
+      inputFormat: Class[_ <: LakeFSBaseInputFormat]
```
arielshaqed, line 29: This name is a bit confusing.  Can we rename, maybe `inputFormatClass`?

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSContext.scala
```diff
@@ -62,6 +73,37 @@ object LakeFSContext {
...
+    newDF(spark, repoName, "", "", classOf[LakeFSRepositoryInputFormat])
+  }
+
+  def newDF(
+      spark: SparkSession,
+      repoName: String,
+      commitID: String
+  ): DataFrame = {
+    newDF(spark, repoName, "", commitID, classOf[LakeFSCommitInputFormat])
+  }
```
arielshaqed, line 100: I think that this may be easier to read as a single function with default argument
values:

```suggestion
  def newDF(
      spark: SparkSession,
      repoName: String,
      commitID: String = ""
  ): DataFrame = {
    newDF(spark, repoName, "", commitID, classOf[LakeFSCommitInputFormat])
  }
```

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSContext.scala
```diff
@@ -62,6 +73,37 @@ object LakeFSContext {
...
+
+  def newRDD(
```
arielshaqed, line 102: Why do we need to create RDDs _and_ DataFrames?

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSInputFormat.scala
```diff
@@ -113,21 +188,67 @@ class LakeFSInputFormat extends InputFormat[Array[Byte], WithIdentifier[Entry]]
...
+    res
   }.asJava
+}
 
-  override def createRecordReader(
-      split: InputSplit,
-      context: TaskAttemptContext
-  ): RecordReader[Array[Byte], WithIdentifier[Entry]] = {
-    new EntryRecordReader(Entry.messageCompanion)
+class LakeFSRepositoryInputFormat extends LakeFSBaseInputFormat {
```
arielshaqed, line 207: AFAICT this reads all metaranges of a repository.  So perhaps LakeFSAllMetarangesInputFormat?  (The current name is quite hard for me, it sounds more like a DF of all the repositories defined on the installation...).

arielshaqed (changes requested):
Thanks!

Nothing major, but requesting changes because I want to re-read it after we make some of these changes.  It's a tough PR for me to review, sorry.

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSInputFormat.scala
```diff
@@ -113,21 +188,67 @@ class LakeFSInputFormat extends InputFormat[Array[Byte], WithIdentifier[Entry]]
...
+        conf.get(LAKEFS_CONF_API_SECRET_KEY_KEY),
+        conf.get(LAKEFS_CONF_API_CONNECTION_TIMEOUT_SEC_KEY),
+        conf.get(LAKEFS_CONF_API_READ_TIMEOUT_SEC_KEY)
+      )
+    )
+    if (StringUtils.isBlank(storageNamespace)) {
+      storageNamespace = apiClient.getStorageNamespace(repoName, StorageClientType.HadoopFS)
+    }
+    var nsURI = URI.create(storageNamespace)
+    nsURI = new URI(nsURI.getScheme,
```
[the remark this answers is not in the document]
johnnyaug: If the input format is initialized with a blank namespace, we use the apiClient to get it - which already does the translate. Otherwise, the user should supply it in the correct format.

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSContext.scala
johnnyaug: It is used in List.scala- not necessarily with good justification (also - you made me realize I had a bug here in the method signature, now fixes).

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSInputFormat.scala
johnnyaug: Done

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSInputFormat.scala
```diff
@@ -113,21 +188,67 @@ class LakeFSInputFormat extends InputFormat[Array[Byte], WithIdentifier[Entry]]
...
-      context: TaskAttemptContext
-  ): RecordReader[Array[Byte], WithIdentifier[Entry]] = {
-    new EntryRecordReader(Entry.messageCompanion)
+class LakeFSRepositoryInputFormat extends LakeFSBaseInputFormat {
+  import LakeFSInputFormat._
+  override def getSplits(job: JobContext): java.util.List[InputSplit] = {
+    val conf = job.getConfiguration
+    val repoName = conf.get(LAKEFS_CONF_JOB_REPO_NAME_KEY)
+    var storageNamespace = conf.get(LAKEFS_CONF_JOB_STORAGE_NAMESPACE_KEY)
+    if (StringUtils.isBlank(repoName) && StringUtils.isBlank(storageNamespace)) {
```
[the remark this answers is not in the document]
arielshaqed: Yuck.

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSInputFormat.scala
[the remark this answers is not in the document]
arielshaqed: Ouch.  Is there anywhere to document this field (I know about the Actual Docs Page, and we probably need to document there, but nobody reads that...)?

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSInputFormat.scala
```diff
@@ -1,74 +1,145 @@
...
+    sb.append("rangeID: ")
+    sb.append(new String(rangeID))
+    sb.append("\n")
+    sb.append("path: ")
+    sb.append(path)
+    sb.append("\n")
+    sb.append("range: ")
+    sb.append(range)
+    sb.toString
+  }
```
arielshaqed, line 93:
```suggestion
  override def toString = s"GravelerSplit: rangeID: ${rangeID}\npath: ${path}\nrange: ${range}"
```
Maybe also add a tab to indent it nicely?
```suggestion
  override def toString = s"GravelerSplit:\trangeID: ${rangeID}\t\tpath: ${path}\n\t\trange: ${range}"
```

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSInputFormat.scala
```diff
@@ -1,74 +1,145 @@
...
 
 class WithIdentifier[Proto <: GeneratedMessage with scalapb.Message[Proto]](
     val id: Array[Byte],
-    val message: Proto
+    val message: Proto,
+    val rangeID: String
 ) {}
 
+object EntryRecordReader {
+  val logger: Logger = LoggerFactory.getLogger(getClass.toString)
```
arielshaqed, line 103: By putting it on a companion object, you get a weird class name:
```scala
scala> object X { val className = getClass.toString }
object X
scala> X.className
val res0: String = class X$

scala> class X { val className = getClass.toString }
scala> (new X).className
val res2: String = class X
```

Why not just put it inside the class?

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSInputFormat.scala
```diff
@@ -1,74 +1,145 @@
...
+
     val fs = gravelerSplit.path.getFileSystem(context.getConfiguration)
     fs.copyToLocalFile(gravelerSplit.path, new Path(localFile.getAbsolutePath))
     // TODO(johnnyaug) should we cache this?
     val sstableReader =
       new SSTableReader(localFile.getAbsolutePath, companion)
+    val props = sstableReader.getProperties()
+    logger.debug(s"Props: $props")
+    if (gravelerSplit.range == null) {
+      if (new String(props.get("type").get) != "ranges" || props.get("entity").nonEmpty) {
```
arielshaqed, line 125: Not sure I understand what this is doing.  In particular, when does this happen?  If it happens, didn't something go badly wrong -- something we'd want to report?

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSInputFormat.scala
```diff
@@ -113,21 +192,67 @@ class LakeFSInputFormat extends InputFormat[Array[Byte], WithIdentifier[Entry]]
...
+    nsURI = new URI(nsURI.getScheme,
+                    nsURI.getUserInfo,
+                    nsURI.getHost,
+                    nsURI.getPort,
+                    nsURI.getPath + (if (nsURI.getPath.endsWith("/")) "_lakefs" else "/_lakefs"),
+                    null,
+                    null
+                   )
+    val fs = FileSystem.get(nsURI, conf)
+    val it = fs.listFiles(new Path(nsURI.toString), false)
```
arielshaqed, line 244: Don't we need to scan just the ranges subprefix, and not the entire storage namespace?  (Or does that happen elsewhere?  If so, then `nsURI` is a really confusing name...)

arielshaqed (changes requested):
Neat!

I really like the way this extends the InputFormat to handle entire ranges.

I have major 2 worries, both should be easy to fix:

* I got confused on l.244 how we manage to scan just the ranges files.  I'd like some clarity  there, possibly renaming some variables will be enough.
* It is quite hard to use this correctly: we need to specify _either_ a commit ID _or_ a storage namespace prefix.  We do test to see that at least one is used, but I am not sure what  happens when we select both.  (I _think_ that should be an error.)

Thanks!

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSInputFormat.scala
johnnyaug: Bad name, sorry. Changing.

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSInputFormat.scala
johnnyaug: This prefix will also include metaranges, and may potentially include "ref-dumps" which represent other entities. This check filters out only ranges.

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSInputFormat.scala
johnnyaug: If it's inside the class, it's created together with every instance. Thoughts?

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSInputFormat.scala
johnnyaug: (sort of) done

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSInputFormat.scala
```diff
@@ -1,74 +1,133 @@
...
-class GravelerSplit(var range: RangeData, var path: Path) extends InputSplit with Writable {
-  def this() = this(null, null)
+object GravelerSplit {
+  val logger: Logger = LoggerFactory.getLogger(getClass.toString)
+}
+class GravelerSplit(var range: RangeData, var path: Path, var rangeID: String, var byteSize: Long)
+    extends InputSplit
+    with Writable {
+  import GravelerSplit._
+  def this() = this(null, null, null, 0L)
```
arielshaqed, line 31: I don't understand this (sorry, bad pun, but the point is valid).  Scala already has a `this` operator, and this makes it really hard to understand to what `this` refers.  Also it does not make sense for most purposes.  Could you document it, or change the name?

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSInputFormat.scala
arielshaqed: I am not sure whether LoggerFactory actually creates a new instance.  And there are not that many RecordReaders in existence.  Your way gives a strange classname, so I would rather avoid it.  (And I don't know how to access the Class object of the class from the companion object, sorry.)

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSInputFormat.scala
```diff
@@ -1,74 +1,133 @@
...
+  val logger: Logger = LoggerFactory.getLogger(getClass.toString)
+}
 class EntryRecordReader[Proto <: GeneratedMessage with scalapb.Message[Proto]](
     companion: GeneratedMessageCompanion[Proto]
 ) extends RecordReader[Array[Byte], WithIdentifier[Proto]] {
+  import EntryRecordReader._
   var it: SSTableIterator[Proto] = _
   var item: Item[Proto] = _
-
+  var gravelerSplit: GravelerSplit = _
```
arielshaqed, line 99: AFAICS only `rangeID` is actually needed outside of `initialize`.  Could we keep just that on the recordreader?

arielshaqed (approved):
THANKS!  Would still be happier to see standard logging (using the correct classname).

PR #4126, opened 2022-09-06
Update installing.md

### LINKED ISSUE #2614: Call out documentations limitation when running docker-compose
I believe most users will start (at least as first) to run the "quick start" way with docker compose as called out here:
[https://docs.lakefs.io/quickstart/installing.html](https://docs.lakefs.io/quickstart/installing.html)

Then, you might want to point to a s3 bucket instead of your local or change some other configuration and will try and follow this:

[https://docs.lakefs.io/reference/configuration.html](https://docs.lakefs.io/reference/configuration.html)

The challenge is that if you run in the simple quick way i.e.:

`curl [https://compose.lakefs.io](https://compose.lakefs.io/) | docker-compose -f - up`

The configuration files referenced are not going to take effect. 

Instead, you can download the file, i.e.

[https://compose.lakefs.io](https://compose.lakefs.io/)

This will download a download.yml locally. Then you can edit that file:
for example, specifically if you like to use a remote s3 bucket change:

```
- LAKEFS_BLOCKSTORE_TYPE=${LAKEFS_BLOCKSTORE_TYPE:-local}

...

- LAKEFS_BLOCKSTORE_S3_CREDENTIALS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID:-}

- LAKEFS_BLOCKSTORE_S3_CREDENTIALS_ACCESS_SECRET_KEY=${AWS_SECRET_ACCESS_KEY:-}
```

to:

```
- LAKEFS_BLOCKSTORE_TYPE=${LAKEFS_BLOCKSTORE_TYPE:-s3}

...

- LAKEFS_BLOCKSTORE_S3_CREDENTIALS_ACCESS_KEY_ID=###

- LAKEFS_BLOCKSTORE_S3_CREDENTIALS_ACCESS_SECRET_KEY=###
```

Next, you can run docker-compose with a parameter for the new file (as opposed to downloading the file again):

`docker-compose -f ModifiedYMLFile.yml up`

IMHO I don't actually think we should document this in the documentation. I do think it is worth calling out in the reference that if you use docker-compose, you should make sure to overwrite the file.

iddoavn:
Added documentation for https://github.com/treeverse/lakeFS/issues/2614 after seeing users running into this issue.

Fix #2614

### DIFF
--- a/docs/quickstart/installing.md
+++ b/docs/quickstart/installing.md
@@ -42,7 +42,7 @@ You can try lakeFS:
 To modify the local configuration file, for example, in order to use your local lakeFS environment against S3 storage as opposed to the local storage, download the configuration file https://compose.lakefs.io, modify it and then run the container with the modified copy:
 
 ```bash
-docker-compose -f ModifiedYMLFile.yml up
+docker-compose -f modified-docker-compose.yml up
 ```
 
 For example, to run against S3 instead of local storage, change:
--- a/docs/quickstart/installing.md
+++ b/docs/quickstart/installing.md
@@ -37,7 +37,7 @@ You can try lakeFS:
 1. With docker-compose [on Windows](more_quickstart_options.md#docker-on-windows).
 1. By [running the binary directly](more_quickstart_options.md#using-the-binary).
 
-## Modifying the local configuration file
+## Modifying the Compose file
 
 To modify the local configuration file, for example, in order to use your local lakeFS environment against S3 storage as opposed to the local storage, download the configuration file https://compose.lakefs.io, modify it and then run the container with the modified copy:
 
--- a/docs/quickstart/installing.md
+++ b/docs/quickstart/installing.md
@@ -37,35 +37,39 @@ You can try lakeFS:
 1. With docker-compose [on Windows](more_quickstart_options.md#docker-on-windows).
 1. By [running the binary directly](more_quickstart_options.md#using-the-binary).
 
-## Modifying the Compose file
+## Modifying the local deployment
 
-To modify the local configuration file, for example, in order to use your local lakeFS environment against S3 storage as opposed to the local storage, download the configuration file https://compose.lakefs.io, modify it and then run the container with the modified copy:
+To modify the local deployment, for example, in order to use your local lakeFS against S3 storage (as opposed to the local storage), you can either:
+1. Pass or expose the variables using this syntax:
 
-```bash
-docker-compose -f modified-docker-compose.yml up
-```
+   ```bash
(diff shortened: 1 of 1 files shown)


docs/quickstart/installing.md
```diff
@@ -37,6 +37,36 @@ You can try lakeFS:
 1. With docker-compose [on Windows](more_quickstart_options.md#docker-on-windows).
 1. By [running the binary directly](more_quickstart_options.md#using-the-binary).
 
+## Modifying the local configuration file
+
+To modify the local configuration file, for example, in order to use your local lakeFS environment against S3 storage as opposed to the local storage, download the configuration file https://compose.lakefs.io, modify it and then run the container with the modified copy:
+
+```bash
+docker-compose -f ModifiedYMLFile.yml up
```
nopcoder, line 45: 1. We use the syntax `${LAKEFS_BLOCKSTORE_TYPE:-local}` in order to pass environment variables to the docker command without the need to download and edit it. So passing/expose the BLOCKSTORE/KEY/SECRET variables will do the job.
2. I think we need to add a note about the fact that lakeFS will try to identify repository that uses different storange namespace and prevert running - unless we remove the previous run and not just stop/start.

docs/quickstart/installing.md
```diff
@@ -37,6 +37,36 @@ You can try lakeFS:
 1. With docker-compose [on Windows](more_quickstart_options.md#docker-on-windows).
 1. By [running the binary directly](more_quickstart_options.md#using-the-binary).
 
+## Modifying the local configuration file
```
johnnyaug, line 40:
```suggestion
## Modifying the Compose file
```
(also below)

docs/quickstart/installing.md
johnnyaug, line 45:
```suggestion
docker-compose -f modified-docker-compose.yml up
```

iddoavn: @nopcoder ,
Would you say instead:
To modify the local deployment, for example, in order to use your local lakeFS environment against S3 storage as opposed to the local storage, you can either pass variables using the syntax ${LAKEFS_BLOCKSTORE_TYPE:-local}  or download the configuration file https://compose.lakefs.io, modify it and then run the container with the modified copy:

```bash
docker-compose -f ModifiedYMLFile.yml up
Member

And of course add your note that between runs you should remove the previous docker installation and not just stop/start it.

arielshaqed:
Thanks, @nopcoder and @johnnyaug! Iddo got the real Docker and usability experts on this one...

docs/quickstart/installing.md
nopcoder: Yes, just specify the names of the variable, not the complete bash syntax of the name+default

PR #4127, opened 2022-09-07
proposal - lakefs kv tool

itai-david:
First version - some open questions at the bottom

### DIFF
--- a/design/open/kv-query-tool.md
+++ b/design/open/kv-query-tool.md
@@ -6,6 +6,8 @@ Our newly introduced KV Store stores the data in the form of [`Partition Key`, `
 This data is not easily interpretable for human viewer, whereas our former SQL DB stored the data (mostly) as human readable data, making it easier for developers to look into the DB, and debug various situations. It also provided the same ease to insert data or alter existing data in the DB for experimentation and, in some cases, as a fix
 With our new KV Store this became less trivial and this proposal comes to offer a solution for that
 
+**Important Assumption:** This tool is meant to be an internal tool for us (`lakefs` developers) and as such it heavily relies on the KV internal knowledge we have (keys construction, hard-coded and generated partitions, etc)
+
 ## Introducing `lakefs kv`
 
 The `lakefs` server contains all the logic needed to correctly read and update the KV Store. It generates objects into byte arrays with their protobuf encoding, on one hand, and rebuilds the objects from byte arrays with protobuf on the other, and, it holds all the logic to generate the relevant keys (e.g. `partition key` from a repository object, branch `key` from the branch object etc.)
@@ -14,32 +16,69 @@ We will elevate this logic and into a new command that will provide the ability 
 lakefs kv <cmd> <args>
 ```
 
-### Get/Scan/Set
+## Implementation Steps
 
-* `lakefs kv get <partition> <key>` </br>
-  Given a `partition key` and a `key` (both as strings), returns the corresponding object (json format)
+These steps describe the order and importance of the functionality we would like to get from `lakefs kv` and does not refer to specific commands format. The commands specified are for illustration purposes. The specifics of the commands format and their arguments are discussed in later section
 
-* `lakefs kv scan <partition> [<key>]` </br>
-  Given a `partition key` and an optional `key`, returns all object in that partition that keys are equal or greater to the specified `key` (no `key` will return all objects in the partition). Results will be formatted as a json array
+### Step 1 - Get/Scan
 
-* `lakefs kv set <partition> <key> <value>` </br>
-  Given a `partition key`, a `key` and a json formatted object, sets the value for the given `key` in the given `partition`, as an encoded protobuf for the given object
+* `lakefs kv get <args>`
+* `lakefs kv scan <args>`
 
-### Delete (?)
-* `lakefs kv delete <partition> <key>` </br>
-  Given a `partition key` and a `key`, deletes the corresponding entry from the KV Store
-  **Question:** Do we really want this?
+These 2 are `lakefs kv`'s MVP. They put the **query** in "**query**ing tool". Given some kind of identifier, these commands returns an object, or scans through a range of objects, outputting their details. The identifier can be a (`partition`,`key`) pair, or an object-type and its identifier (as used by lakefs, e.g. RepositoryID, BranchID etc.), depends on the approach we select - to be discussed later
 
-### The `<key>` is the Key
-Our KV implementation creates different formatted `path`s to different object types, having a hint for the type encoded within the `path`. The specified `key` (for each operation) should match this `path` format, and as such, it hints for the object type.
-The querying tool will use a `key-format` to `object` translation functionality where the given key will be matched
-This will be elevated to use the correct object to decode either the read protobug (for Get/Scan), or given json (for set), to an actual object, and decode it to json/protobuf respectively
-Upon process start, all `path` formatters will register with their corresponding object type, creating a format->object mapping, to be used.
-This will make it easy to add new object types in the future, and register them as well.
-**TBD:** How to implement and enforce this requirement?
+### Step 2 - REPL
 
-## Open Questions
-* How can we implement the registration of the `path` function to a `path` -> object translator? I think, that as a first step, going over all the currently known object types and registering them, will be sufficient
-* How about partitions other than the trivial "auth" and "graveler"? Is it enough to leave it to the user to query for a repository, and calculate its `partition key`? Or should we add some ad-hoc commands for that (e.g. `lakefs kv get-repo-key <repo-id>`)?
+An interactive CLI prompt
+```
+$ lakefs kv repl
+
+kv>
+```
+The possible commands are no other than the ones supported by `lakefs kv`, i.e. Get & Scan, and if step 3 is to be implemented, Set and Delete
+
+The benefits of this approach is that unlike the `lakefs kv XXX` (where XXX is `get` or `set`), which runs as a dedicated process and terminates once input is generated, the `repl` command can execute multiple commands in the same process, saving the need to recreate data structures and connect to KV, as well as reducing the chance of failures due to errors in the above
+
+### Step 3 - Set/Delete - If the need arises
+
+* `lakefs kv set <args>`
+* `lakefs kv delete <args>`
+
+These commands can be added in case the need arises. The first is to alter/add a KV store entry, and the other is for removing an entry. These are very error prone operations that can easily result in data corruption, and so we will (a) not implement these right away, but only if they are really needed and (b) protect these with, at least, a requirement for command confirmation: `"Are you sure you want to delete entry? [yes/N]" etc.
+
+## Different Approaches, Different Interfaces, Different Problems
+
+### Full Knowledge
+
+With this approach, `lakefs kv` is implemented with all the knowledge available. Knowledge like: how to generate a repository partition from the repository ID and object, how to generate a branch key from the repository and branch IDs, the fact that staged objects' keys cannot be provided as strings etc.
+The commands in that case will be very specific, having the user specifying which entity he is working with. Essentially, this will be a CLI interface to various services, using KV (graveler, auth, etc.)
+* `lakefs kv get-repository <REPO_ID>`
+* `lakefs kv scan-staging <STAGING_TOKEN>` etc.
+With this approach, `lakefs kv` will not have functionalities such as `get-staged-object` by key, or any other operation we know is not possible or does not make any sense
+
+Pros:
+* Implementation is pretty straight forward - we actually already have everything implemented, and only need to make it accessible via `lakefs kv` tool
+
+Cons:
+* Lack of flexibility - adding a new entity, requires to implement new commands (bring in the knowledge) in the `lakefs kv` tool
+* Not really a "KV Querying" but more of a "services querying" tool. This might bring bugs from the existing logic to mask KV problems, or make them harder to identify
+
+### Minimal Knowledge
+
+With this approach, `lakefs kv` will have the minimal required knowledge of objects and their structures. It will present a generic KV-like interface that works with `partition`, `key` and `entries`.
+* `lakefs kv get <partition> <key>`
+* `lakefs kv scan <partition> [<key>]`
+This implementation will acquire the required data dynamically at runtime, by having entities register their a key parsing/matching function, an object structure and any other needed data, I cannot think of at the moment. This approach relies on the fact that the `key` structure "hides" information about the entity type, and so `lakefs kv` can leverage that in a sort of reflection-like way, to access the correct entity and construct the corresponding object, in order to decode the entry data correctly and  display it
+
+Pros:
+* High level of flexibility - in order to introduce a new entity type, all one needs to do is to implement a correct object and a matcher/parser func, and register it. `lakefs kv` code has zero knowledge of the entities and all this knowledge is coming from the entities implementation
+* `lakefs kv` interface remains unchanged - the commands are generic and are ready to handle newly introduced entities
+* Accessing KV table the way it was meant to be accessed, which is quite what we are aiming for
+
+Cons:
+* This general approach might not hold, has not all entities can be treated the same - e.g. a repository object can be found, with his ID, under a hard-coded well known partition, but a branch need a repository in order to get its partition
+  * Can be bypassed by counting on the user to provide the required partition correctly. i.e. in order to get a branch, one need to know the repository it belongs too (makes sense) get this repository using `lakefs kv`, generating the repository partition, and use the generated partition in to get the branch. This solution relies on the assumption that a user is an internal user (dev)
+  * We can ease it up a bit by providing very specific ad-hoc commands such as `lakefs kv get-repo-partition`, to make the user life easier
+* Non printable keys, such as staged object keys, may contain non printable chars. It will be impossible to use a `lakefs kv get` and specify a key for that. I think we can overcome this by blocking this specific functionality (get) and only allow `scan` for staging tokens. This will break the generality a bit, but I don't see any better option
 
 
--- a/design/open/kv-query-tool.md
+++ b/design/open/kv-query-tool.md
@@ -1,0 +1,45 @@
+# KV Store Querying Tool
+
+## Scope
+
+Our newly introduced KV Store stores the data in the form of [`Partition Key`, `Key`, `Value`] triplets. The `partition key` and `key` fields are strings stored as byte arrays and the `value` is an object (various types) serialized as protobuf and stored as byte array.
+This data is not easily interpretable for human viewer, whereas our former SQL DB stored the data (mostly) as human readable data, making it easier for developers to look into the DB, and debug various situations. It also provided the same ease to insert data or alter existing data in the DB for experimentation and, in some cases, as a fix
+With our new KV Store this became less trivial and this proposal comes to offer a solution for that
+
+## Introducing `lakefs kv`
+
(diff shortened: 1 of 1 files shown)


design/open/kv-query-tool.md
```diff
@@ -0,0 +1,45 @@
+# KV Store Querying Tool
+
+## Scope
+
+Our newly introduced KV Store stores the data in the form of [`Partition Key`, `Key`, `Value`] triplets. The `partition key` and `key` fields are strings stored as byte arrays and the `value` is an object (various types) serialized as protobuf and stored as byte array.
```
N-o-Z, line 5: Key is byte slice, not string

design/open/kv-query-tool.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Introducing `lakefs kv`
+
+The `lakefs` server contains all the logic needed to correctly read and update the KV Store. It generates objects into byte arrays with their protobuf encoding, on one hand, and rebuilds the objects from byte arrays with protobuf on the other, and, it holds all the logic to generate the relevant keys (e.g. `partition key` from a repository object, branch `key` from the branch object etc.)
+We will elevate this logic and into a new command that will provide the ability to read data items, scan keys and update values, using a new `lakefs` command - `kv`
+```
+lakefs kv <cmd> <args>
+```
+
+### Get/Scan/Set
```
N-o-Z, line 17: In my opinion supporting Set and Delete are risky and not necessary. In the very rare occasion where an entry needs to be modified it should be done directly on the DB and by a lakeFS engineer.
Any situation that requires modifying multiple key (via scan and set) for example, is irrelevant since it will probably already be a catastrophe

design/open/kv-query-tool.md
```diff
@@ -0,0 +1,45 @@
...
+
+* `lakefs kv set <partition> <key> <value>` </br>
+  Given a `partition key`, a `key` and a json formatted object, sets the value for the given `key` in the given `partition`, as an encoded protobuf for the given object
+
+### Delete (?)
+* `lakefs kv delete <partition> <key>` </br>
+  Given a `partition key` and a `key`, deletes the corresponding entry from the KV Store
+  **Question:** Do we really want this?
+
+### The `<key>` is the Key
```
N-o-Z, line 33: We do not expose paths to the user and in some cases it will be impossible to determine it without internal knowledge.
For example: how does the user infer the path of a branch under a repository which uses the repository UUID?

design/open/kv-query-tool.md
```diff
@@ -0,0 +1,45 @@
...
+Our KV implementation creates different formatted `path`s to different object types, having a hint for the type encoded within the `path`. The specified `key` (for each operation) should match this `path` format, and as such, it hints for the object type.
+The querying tool will use a `key-format` to `object` translation functionality where the given key will be matched
+This will be elevated to use the correct object to decode either the read protobug (for Get/Scan), or given json (for set), to an actual object, and decode it to json/protobuf respectively
```
N-o-Z, line 36: 'protobuf' 😄

N-o-Z (commented):
The suggestion is very interesting, 2 questions I had in mind.
1. Is this tool meant to be used internally or for lakeFS users as well?
If this is internal, I would implement it as an internal tool not provided as part of the lakeFS server.
Then, we might allow Set and Delete functionality with extra care.
Otherwise if this a tool for users as well, I think we need to take extreme caution not to expose any functionality which will accidentally cause the user to lose data.
2. How do we deal with scan on a prefix that spans over several object types?

design/open/kv-query-tool.md
itaiad200: I think `partition_key` too

design/open/kv-query-tool.md
itaiad200: I tend to agree that P0 would be Scan/Get (or even just Scan) and the rest can be added when needed.

design/open/kv-query-tool.md
itaiad200: Yap, repository UUID is a killer. Maybe we can try to abstract it away from the user, i.e. perform another call to get the repository UUID when given the repository path?

design/open/kv-query-tool.md
```diff
@@ -0,0 +1,45 @@
...
+
+The `lakefs` server contains all the logic needed to correctly read and update the KV Store. It generates objects into byte arrays with their protobuf encoding, on one hand, and rebuilds the objects from byte arrays with protobuf on the other, and, it holds all the logic to generate the relevant keys (e.g. `partition key` from a repository object, branch `key` from the branch object etc.)
+We will elevate this logic and into a new command that will provide the ability to read data items, scan keys and update values, using a new `lakefs` command - `kv`
+```
+lakefs kv <cmd> <args>
+```
+
+### Get/Scan/Set
+
+* `lakefs kv get <partition> <key>` </br>
```
itaiad200, line 19: Why `lakefs` and not `lakectl`?

design/open/kv-query-tool.md
```diff
@@ -0,0 +1,45 @@
...
+### Delete (?)
+* `lakefs kv delete <partition> <key>` </br>
+  Given a `partition key` and a `key`, deletes the corresponding entry from the KV Store
+  **Question:** Do we really want this?
+
+### The `<key>` is the Key
+Our KV implementation creates different formatted `path`s to different object types, having a hint for the type encoded within the `path`. The specified `key` (for each operation) should match this `path` format, and as such, it hints for the object type.
+The querying tool will use a `key-format` to `object` translation functionality where the given key will be matched
+This will be elevated to use the correct object to decode either the read protobug (for Get/Scan), or given json (for set), to an actual object, and decode it to json/protobuf respectively
+Upon process start, all `path` formatters will register with their corresponding object type, creating a format->object mapping, to be used.
```
itaiad200, line 37: Which process? `lakefs`?

design/open/kv-query-tool.md
```diff
@@ -0,0 +1,45 @@
...
+This will make it easy to add new object types in the future, and register them as well.
+**TBD:** How to implement and enforce this requirement?
```
itaiad200, line 39: I don't think that we do, it's a plumbing command so we don't have to be strict about anything.

itaiad200 (changes requested):
This is an awesome first step. I think we can overcome many of the obstacle if we decide:
1. Should be a standalone, lakeFS or lakectl command. If the latter, what other configuration is needed?
2. It's a plumbing command, it can assume the user is very familiar with lakeFS and it can require some additional configuration.

design/open/kv-query-tool.md
N-o-Z: We can always represent `partition_key` as a string, we can't give that guarantee for `key` (e.g. staging entry)

design/open/kv-query-tool.md
eden-ohana: To save two calls you can use:
```repo scan <repo-name>``` -  without the key, it will translate it to the result and returns a table with the keys and all repository entities like when you're querying directly from the DB.

**Review by eden-ohana — COMMENTED** [review]
I don't think this tool needs to be visible to the users. Maybe add it as a script. worth checking if current users querying their DB, if not I'm not sure we should expose this tool to them.

File: design/open/kv-query-tool.md (hunk 1 above)
Comment (itai-david) [reply design/open/kv-query-tool.md:5]: That's right for the keys (partition and key) that `lakefs` send to the KV store. But somewher in the process, lakefs translates strings into byte slices. I want to use the same logic for the querying tool.
Eventually, a `CLI`-ish tool, will have to use readable strings, at least at the `I` part. WDYT?

Comment (N-o-Z) [reply design/open/kv-query-tool.md:5]: Keep in mind that `keys` in contrast to `partition_keys` are by design (graveler constraint) byte slices - we can't assume they have any encoding or have a valid string representation

File: design/open/kv-query-tool.md (hunk 2 above)
Comment (itai-david) [reply design/open/kv-query-tool.md:17]: I'm OK with Scan/Get as a first phase.
I do think that altering data will be proven useful, at least for testing and development attempts.
We can also "protect" it by requiring an `Are you sure?` repsponse, etc. - WDYT?

Will update the doc to reflect the prioritization changes

Comment (N-o-Z) [reply design/open/kv-query-tool.md:17]: Definitely requires a prompt, also I'm for making it an internal dev tool.

File: design/open/kv-query-tool.md (hunk 3 above)
Comment (itai-david) [reply design/open/kv-query-tool.md:33]: 1. AFAIU, this tool is not meant to be exposed to users. I cannot prevent the agile user from looking at the code, understanding the logic and use it, but this tool is meant to help **us** understanding and debugging a system. I can see a scenario where we ask/suggest a user to run several operations, in order to help us get more info of his system state, but I believe it will have to be lead by one of us, so the knowledge is here
2. Adding ad-hoc commands will require to introduce specific knowledge into this logic (i.e. `repo-scan` will require to understand this is a repository object and parse it as such. While this is definitely an option, it is not easily scalable (e.g. introducing new object we would like to scan, like StagingToken). I think that as a first step we should consider limiting oursleves to a KV-like interface

**Comment by itai-david** [comment]
> The suggestion is very interesting, 2 questions I had in mind.
> 
> 1. Is this tool meant to be used internally or for lakeFS users as well?
>    If this is internal, I would implement it as an internal tool not provided as part of the lakeFS server.
>    Then, we might allow Set and Delete functionality with extra care.
>    Otherwise if this a tool for users as well, I think we need to take extreme caution not to expose any functionality which will accidentally cause the user to lose data.
> 2. How do we deal with scan on a prefix that spans over several object types?

@N-o-Z 
1. It is meant to be used solely by us for debugging, experimenting and maybe, in rare situation, to fix issues (no realistic scenario comes to my mind at the moment)
2. My idea is to derive the object type from the path (key) by implementing a `parse` or `match` functions that will do the opposite of the `XXXPath` functions - e.g. a path with `branches` prefix leads to a branch object. `commits` prefix to a commit and so on. My main purpose is to map all these matchers/parsers to an object and make it dynamically scalable, as new object types, if needed, can register their parser and be handled

File: design/open/kv-query-tool.md (hunk 5 above)
Comment (itai-david) [reply design/open/kv-query-tool.md:19]: `lakefs` already has all the logic to access the KV and to manage it's objects.
I think that `lakectl`, as a client, should stay as thin as possible and I don't it should be exposed to very internal details such as KV interface.
Moreover, if we do decide, now or due to some future need, to add ad-hoc commands, such as `scan repository <repository ID>` (as suggested by @eden-ohana), we can leverage graveler to achieve that. This will make `lakectl` even more exposed to `lkaefs` internals
Implementing the client logic in `lakectl` and sending to `lakefs` as a server, seems rather an overkill with API additions, and I don't think we want to get there.

File: design/open/kv-query-tool.md (hunk 6 above)
Comment (itai-david) [reply design/open/kv-query-tool.md:37]: Correct. `lakefs kv` which, at step 1 is a single command process that starts, builds its internal data structures (mapping of registered parsers/matchers to objects etc) and connects to KV, runs the specified (currently single, but not necessarily) operation on KV, generates textual output and terminates. At step x > 1, @nopcoder suggested a REPL, that will save the processing of bringing up the internal datastructures and KV connection, and will act as a CLI itself:
```
lakefs kv repl

kv>
```

**Comment by itai-david** [comment]
> I don't think this tool needs to be visible to the users. Maybe add it as a script. worth checking if current users querying their DB, if not I'm not sure we should expose this tool to them.

@eden-ohana - totally agree. This is not meant to be visible to users. The option to have it wrapped in a script, that we provide, in order to do allow the use by users in certain cases, is magnificent

File: design/open/kv-query-tool.md (hunk 1 above)
Comment (itai-david) [reply design/open/kv-query-tool.md:5]: @N-o-Z - I believe you are specifically refering to the staging object path, that may contain an unreadable chars.
In that case, or any other case we cannot guarantee displayable encoding, I cannot see how can we access/display these from a human-usable, string based, CLI tool
We might want to skip the `key` part here, and only allow scan. This will display some kind of key - not neccessarily readable, and the formatted data
This will require the tool to specifically handle known caveats, such as this, and is not generally scalable, but apparently we will have to address it

File: design/open/kv-query-tool.md (hunk 2 above)
Comment (itai-david) [reply design/open/kv-query-tool.md:17]: Agree. The entire tool is aimed to be internal

File: design/open/kv-query-tool.md (hunk 7 above)
Comment (itai-david) [reply design/open/kv-query-tool.md:39]: I'm asking regarding enforcing implementation details - how do I enforce that new objects introduced to lakefs, will comply with the requirement to register a parser/matcher with the `lakefs kv` tool, so it can interpret a key, such as `/someNewType/id123` as a NewType object, and display it correctly.
This is also not a must, but will probably require some maintaining if and when these types are introduced.

File: design/open/kv-query-tool.md (hunk 3 above)
Comment (itai-david) [reply design/open/kv-query-tool.md:33]: Since this is internal tool, we can expect the user (us) to have the required knowledge to operate it correctly, e.g, get the repository best on the well-known `graveler` partition and either `scan` or the repopsitoryID, and from there, building the repository's partition based on the InstanceUID (that we will need to expose)
Alternatively, we can implement a `get repo-partition`, or `repo-scan` to easily overcome this, although this breaks the generality I had in mind in the first place. I guess it's a question we need to ask

File: design/open/kv-query-tool.md (hunk 1 above)
Comment (itai-david) [reply design/open/kv-query-tool.md:5]: Modified doc to reflect the challenge presented by non printable keys, and offer optional bypassing

File: design/open/kv-query-tool.md (hunk 2 above)
Comment (itai-david) [reply design/open/kv-query-tool.md:17]: Added to the document

**Comment by itai-david** [comment]
@itaiad200 @N-o-Z @eden-ohana 
Thanks a lot for these excellent comments. It got me aware of some ideas I did not think of, thinking of some aspects I was not aware of, and got me to come up with some bypasses and ways to overcome it.
Document was updated according to your inputs
Please rereview
Thanks :pray:

File: design/open/kv-query-tool.md [hunk 8]
```diff
@@ -14,32 +16,69 @@ We will elevate this logic and into a new command that will provide the ability
...
+
+These commands can be added in case the need arises. The first is to alter/add a KV store entry, and the other is for removing an entry. These are very error prone operations that can easily result in data corruption, and so we will (a) not implement these right away, but only if they are really needed and (b) protect these with, at least, a requirement for command confirmation: `"Are you sure you want to delete entry? [yes/N]" etc.
+
+## Different Approaches, Different Interfaces, Different Problems
+
+### Full Knowledge
+
+With this approach, `lakefs kv` is implemented with all the knowledge available. Knowledge like: how to generate a repository partition from the repository ID and object, how to generate a branch key from the repository and branch IDs, the fact that staged objects' keys cannot be provided as strings etc.
+The commands in that case will be very specific, having the user specifying which entity he is working with. Essentially, this will be a CLI interface to various services, using KV (graveler, auth, etc.)
+* `lakefs kv get-repository <REPO_ID>`
```
itaiad200, line 55: How is it different than `lakectl repo get <REPO_ID>`?

design/open/kv-query-tool.md
```diff
@@ -14,32 +16,69 @@ We will elevate this logic and into a new command that will provide the ability
...
+* `lakefs kv scan-staging <STAGING_TOKEN>` etc.
```
itaiad200, line 56:
```
lakectl diff lakefs://example-repo/example-branch
Show uncommitted changes in example-branch.
```

design/open/kv-query-tool.md
```diff
@@ -14,32 +16,69 @@ We will elevate this logic and into a new command that will provide the ability
...
+### Minimal Knowledge
 
-### Delete (?)
-* `lakefs kv delete <partition> <key>` </br>
-  Given a `partition key` and a `key`, deletes the corresponding entry from the KV Store
-  **Question:** Do we really want this?
+With this approach, `lakefs kv` will have the minimal required knowledge of objects and their structures. It will present a generic KV-like interface that works with `partition`, `key` and `entries`.
+* `lakefs kv get <partition> <key>`
+* `lakefs kv scan <partition> [<key>]`
+This implementation will acquire the required data dynamically at runtime, by having entities register their a key parsing/matching function, an object structure and any other needed data, I cannot think of at the moment. This approach relies on the fact that the `key` structure "hides" information about the entity type, and so `lakefs kv` can leverage that in a sort of reflection-like way, to access the correct entity and construct the corresponding object, in order to decode the entry data correctly and  display it
```
itaiad200, line 71: I don't think reflection is needed. Like you wrote, it's more of a regex matching keys problem, with each matcher linking to a protobuf schema.

itaiad200 (approved):

design/open/kv-query-tool.md
```diff
@@ -14,32 +16,69 @@ We will elevate this logic and into a new command that will provide the ability
...
+Cons:
+* This general approach might not hold, has not all entities can be treated the same - e.g. a repository object can be found, with his ID, under a hard-coded well known partition, but a branch need a repository in order to get its partition
+  * Can be bypassed by counting on the user to provide the required partition correctly. i.e. in order to get a branch, one need to know the repository it belongs too (makes sense) get this repository using `lakefs kv`, generating the repository partition, and use the generated partition in to get the branch. This solution relies on the assumption that a user is an internal user (dev)
+  * We can ease it up a bit by providing very specific ad-hoc commands such as `lakefs kv get-repo-partition`, to make the user life easier
+* Non printable keys, such as staged object keys, may contain non printable chars. It will be impossible to use a `lakefs kv get` and specify a key for that. I think we can overcome this by blocking this specific functionality (get) and only allow `scan` for staging tokens. This will break the generality a bit, but I don't see any better option
```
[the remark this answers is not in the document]
itai-david: That too is an option :)

design/open/kv-query-tool.md
itai-david: That was exactly my intention, hence the reflection-like

design/open/kv-query-tool.md
itai-david: First, `lakectl repo get` involves an API call, but I don't think that what you meant.
`lakectl repo get` will get the repository from graveler's ref manager (via catalog, via controller, via API ofcourse), and will get it processed and chewed. For instance, if the repository state is `RepositoryState_IN_DELETION`, `lakectl repo get` will get an error rather than a repositry. This is not my intention for `lakefs kv` - it will return the repository as it extracted from the KV, with the minimal processing of unmarshalling. That way, `lkaefs kv` will be useful in debugging KV issues, as it will not be amsked by other logic. Does it make sense?

design/open/kv-query-tool.md
itai-david: ?

design/open/kv-query-tool.md
```diff
@@ -14,32 +16,69 @@ We will elevate this logic and into a new command that will provide the ability
...
+### Step 3 - Set/Delete - If the need arises
+
+* `lakefs kv set <args>`
+* `lakefs kv delete <args>`
+
+These commands can be added in case the need arises. The first is to alter/add a KV store entry, and the other is for removing an entry. These are very error prone operations that can easily result in data corruption, and so we will (a) not implement these right away, but only if they are really needed and (b) protect these with, at least, a requirement for command confirmation: `"Are you sure you want to delete entry? [yes/N]" etc.
+
+## Different Approaches, Different Interfaces, Different Problems
+
+### Full Knowledge
```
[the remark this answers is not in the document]
itai-david: I agree - I'm not a fan of this approach either.
Would much like to go with the second approach, of generic `get`, `scan`
Given you last comment of "worse case we'll present garbage..." it makes it easier to deal with the impossible non-printable keys

N-o-Z (approved):
lgtm, thanks!

PR #4129, opened 2022-09-07
Retry cloud id fetch

itaiad200:
Need to build image and check in the playground before releasing
closes https://github.com/treeverse/lakeFS/issues/4132

### DIFF
--- a/pkg/cloud/aws/metadata.go
+++ b/pkg/cloud/aws/metadata.go
@@ -55,12 +55,13 @@ func (m *MetadataProvider) GetMetadata() map[string]string {
 	err = backoff.Retry(func() error {
 		identity, err = stsClient.GetCallerIdentity(&sts.GetCallerIdentityInput{})
 		if err != nil {
-			m.logger.WithError(err).Warn("Failed to get AWS account ID for BI")
+			m.logger.WithError(err).Warn("Tried to to get AWS account ID for BI")
 			return err
 		}
 		return nil
 	}, bo)
 	if err != nil {
+		m.logger.WithError(err).Warn("Failed to to get AWS account ID for BI")
 		return nil
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/cloud/aws/metadata.go
```diff
@@ -42,11 +43,28 @@ func (m *MetadataProvider) GetMetadata() map[string]string {
...
+		maxElapsedTime = 3 * time.Second
+	)
+	bo := backoff.NewExponentialBackOff()
+	bo.MaxInterval = maxInterval
+	bo.MaxElapsedTime = maxElapsedTime
+	var identity *sts.GetCallerIdentityOutput
+	err = backoff.Retry(func() error {
+		identity, err = stsClient.GetCallerIdentity(&sts.GetCallerIdentityInput{})
+		if err != nil {
+			m.logger.WithError(err).Warn("Failed to get AWS account ID for BI")
```
nopcoder, line 58: can we write here 'Try to get' and outside the retyr 'Failed to get'

pkg/cloud/aws/metadata.go
```diff
@@ -42,11 +43,28 @@ func (m *MetadataProvider) GetMetadata() map[string]string {
 	sess.ClientConfig(s3.ServiceName)
 
 	stsClient := sts.New(sess)
-	identity, err := stsClient.GetCallerIdentity(&sts.GetCallerIdentityInput{})
+
+	const (
+		maxInterval    = 200 * time.Millisecond
+		maxElapsedTime = 3 * time.Second
```
nopcoder, line 49: total 3 seconds made the difference?

nopcoder (approved):
suggested to keep the old log to understand when we failed

pkg/cloud/aws/metadata.go
itaiad200: It's not about the time, it's more about the tries. In the playground we'll only start lakeFS if it (`aws sts get-caller-identity`) succeeded once before. But since AWS IAM is eventual consistent, there is a chance that it will fail from here. Retrying it here at least 15 times makes it much more reliable to get at least one successful response.

itaiad200:
Tested in demo-staging and it worked as expected

PR #4131, opened 2022-09-07
Avoid writing binary to logs

### LINKED ISSUE #4130: lakeFS logs write unparsed binary data in the kv store

itaiad200:
closes #4130

### DIFF
--- a/pkg/kv/dynamodb/store.go
+++ b/pkg/kv/dynamodb/store.go
@@ -206,7 +206,7 @@ func (s *Store) Get(ctx context.Context, partitionKey, key []byte) (*kv.ValueWit
 	dynamoRequestDuration.WithLabelValues(operation).Observe(time.Since(start).Seconds())
 	if err != nil {
 		dynamoFailures.WithLabelValues(operation).Inc()
-		return nil, fmt.Errorf("get item: %s (key=%v): %w", err, string(key), kv.ErrOperationFailed)
+		return nil, fmt.Errorf("get item: %w", err)
 	}
 	if result.ConsumedCapacity != nil {
 		dynamoConsumedCapacity.WithLabelValues(operation).Add(*result.ConsumedCapacity.CapacityUnits)
@@ -219,7 +219,7 @@ func (s *Store) Get(ctx context.Context, partitionKey, key []byte) (*kv.ValueWit
 	var item DynKVItem
 	err = dynamodbattribute.UnmarshalMap(result.Item, &item)
 	if err != nil {
-		return nil, fmt.Errorf("unmarshal map: %s (key=%v): %w", err, string(key), kv.ErrOperationFailed)
+		return nil, fmt.Errorf("unmarshal map: %w", err)
 	}
 
 	return &kv.ValueWithPredicate{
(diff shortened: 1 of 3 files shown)


pkg/kv/dynamodb/store.go
```diff
@@ -219,7 +219,7 @@ func (s *Store) Get(ctx context.Context, partitionKey, key []byte) (*kv.ValueWit
 	var item DynKVItem
```
N-o-Z, line 219: You missed one at 216

itai-david, line 219: I think you forgot one on line 216
(Can't add a comment on the exact line, as it is uncahnged)

itaiad200: Yikes. Thanks.

pkg/kv/dynamodb/store.go
```diff
@@ -206,7 +206,7 @@ func (s *Store) Get(ctx context.Context, partitionKey, key []byte) (*kv.ValueWit
 	dynamoRequestDuration.WithLabelValues(operation).Observe(time.Since(start).Seconds())
 	if err != nil {
 		dynamoFailures.WithLabelValues(operation).Inc()
-		return nil, fmt.Errorf("get item: %s (key=%v): %w", err, string(key), kv.ErrOperationFailed)
+		return nil, fmt.Errorf("get item: %w", err)
```
itai-david, line 209: I think this one (and all the following that converts `key` into string, actually logs the key in a pretty readable format. Are you sure we want to discard these?

itaiad200:
> I understand where this is coming from, but I still think it might benefit from having this information. Can we maybe print it to a trace log instead?

The HTTP logging middleware prints this information. With the proper wrapping along the way, we would know what db request failed. Notice that the keys are binary, you can't assume encoding. An example of a log line:
```
"context canceled: operation failed (key=[99 100 52 97 100 100 99 55 54 51 55 48 57 100 57 100 56 98 99 49  52 56 99 53 56 50 53 100 48 101 101 48 54 56 51 101 54 97 51 49 102 97 99 48 53 55 57 111 109 109 105 116 115 47 98 55 54 49 102 50 49 50 100 100 48 100 102 54 98 49 53 54])"

```

itai-david (commented):
Suggestion/Opinion : I think the postgres store is the one printing unreadable byte arrays, as it uses the `key` as is. In dynamodb store, these are converted to strings and are mostly readable. I think we should consider leaving these keys and use the same conversion for the keys in the other stores (postgres, mem), as this info might make it a lot easier to focus on any store related failure

N-o-Z (approved):
Lgtm

itaiad200:
> Suggestion/Opinion : I think the postgres store is the one printing unreadable byte arrays, as it uses the `key` as is. In dynamodb store, these are converted to strings and are mostly readable. I think we should consider leaving these keys and use the same conversion for the keys in the other stores (postgres, mem), as this info might make it a lot easier to focus on any store related failure

How are they converted to strings? If you mean DynamoDB UI, then it's base64 encoded and that's not helpful to debug anything:
image

PR #4133, opened 2022-09-07
Always run Esti

arielshaqed:
Skipping Esti for docs-only PRs does not work: the PR must be pushed by force.  So instead of doing the wrong thing quickly, waste some time and do the right thing.


johnnyaug (commented):
Until next time

johnnyaug (approved):

Issue #4108, opened 2022-09-05
Crash in `PartitionIterator.Close()`


itai-david:
Found on production
`itr` field is checked for nil value but as it is an interface its type should be checked too
Need to reproduce this as part of #3636 

```
2022/09/04 22:27:12 http: panic serving 10.0.41.133:32934: runtime error: invalid memory address or nil pointer dereference
goroutine 99962570 [running]:
net/http.(*conn).serve.func1()
	/usr/local/go/src/net/http/server.go:1802 +0xb9
panic({0x17af000, 0x34bf510})
	/usr/local/go/src/runtime/panic.go:1047 +0x266
github.com/treeverse/lakefs/pkg/kv/postgres.(*EntriesIterator).Close(0x7f51089b6108)
	/build/pkg/kv/postgres/store.go:352 +0x12
github.com/treeverse/lakefs/pkg/kv.(*PartitionIterator).Close(...)
	/build/pkg/kv/iterators.go:322
github.com/treeverse/lakefs/pkg/graveler/staging.(*Iterator).Close(0xc01a5b46b0)
	/build/pkg/graveler/staging/iterator.go:73 +0x2b
github.com/treeverse/lakefs/pkg/graveler.(*CombinedIterator).Close(0xc01b53eed0)
	/build/pkg/graveler/combined_iterator.go:130 +0x29
github.com/treeverse/lakefs/pkg/graveler.(*FilterTombstoneIterator).Close(0xc00e436a00)
	/build/pkg/graveler/combined_iterator.go:165 +0x22
github.com/treeverse/lakefs/pkg/catalog.(*valueEntryIterator).Close(0x143bf71)
	/build/pkg/catalog/value_entry_iterator.go:59 +0x22
github.com/treeverse/lakefs/pkg/catalog.(*prefixIterator).Close(0x143eb26)
	/build/pkg/catalog/prefix_iterator.go:58 +0x22
github.com/treeverse/lakefs/pkg/catalog.(*entryListingIterator).Close(0x143c07b)
	/build/pkg/catalog/entry_listing_iterator.go:53 +0x22
github.com/treeverse/lakefs/pkg/catalog.(*Catalog).ListEntries(0xc000113140, {0x28a5030, 0xc01c308cf0}, {0xc017198dd9, 0x11}, {0xc017198df0, 0x28}, {0xc01c6415f0, 0x85}, {0xc017198e38, ...}, ...)
[...]

PR #4136, opened 2022-09-08
kv iterators unit tests

### LINKED ISSUE #3636: Add unit tests for KV iterators

eden-ohana:
closes #3636

### DIFF
--- a/pkg/kv/iterators_test.go
+++ b/pkg/kv/iterators_test.go
@@ -1,0 +1,307 @@
+package kv_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/go-test/deep"
+	"github.com/stretchr/testify/require"
+	"github.com/treeverse/lakefs/pkg/kv"
+	"github.com/treeverse/lakefs/pkg/kv/kvtest"
+)
+
+const (
+	firstPartitionKey  = "m"
+	secondPartitionKey = "ma"
+	thirdPartitionKey  = "mb"
+)
+
+func TestIterators(t *testing.T) {
+	ctx := context.Background()
+	store := kvtest.GetStore(ctx, t)
+	sm := kv.StoreMessage{
+		Store: store,
+	}
+	m := []kvtest.TestModel{
+		{Name: []byte("a")},
+		{Name: []byte("aa")},
+		{Name: []byte("b")},
+		{Name: []byte("c")},
+		{Name: []byte("d")},
+	}
+
+	// prepare data
+	for i := 0; i < len(m); i++ {
+		err := sm.SetMsg(ctx, firstPartitionKey, m[i].Name, &m[i])
+		if err != nil {
+			t.Fatal("failed to set model", err)
+		}
+
+		err = sm.SetMsg(ctx, secondPartitionKey, m[i].Name, &m[i])
+		if err != nil {
+			t.Fatal("failed to set model", err)
+		}
+	}
+
+	err := sm.SetMsg(ctx, thirdPartitionKey, []byte("e"), &kvtest.TestModel{Name: []byte("notin")})
+	if err != nil {
+		t.Fatal("failed to set model", err)
+	}
+
+	t.Run("partition iterator listing all values of partition", func(t *testing.T) {
+		itr := kv.NewPartitionIterator(ctx, store, (&kvtest.TestModel{}).ProtoReflect().Type(), firstPartitionKey)
+		require.NotNil(t, itr)
+		names := make([]string, 0)
+		for itr.Next() {
+			e := itr.Entry()
+			model, ok := e.Value.(*kvtest.TestModel)
+			if !ok {
+				t.Fatalf("cannot read from store")
+			}
+			names = append(names, string(model.Name))
+		}
+		if itr.Err() != nil {
+			t.Fatalf("unexpected error: %v", itr.Err())
+		}
+		itr.Close()
+
+		if diffs := deep.Equal(names, []string{"a", "aa", "b", "c", "d"}); diffs != nil {
+			t.Fatalf("got wrong list of names: %v", diffs)
+		}
+	})
+
+	t.Run("partition iterator listing values SeekGE", func(t *testing.T) {
+		itr := kv.NewPartitionIterator(ctx, store, (&kvtest.TestModel{}).ProtoReflect().Type(), secondPartitionKey)
+		require.NotNil(t, itr)
+		itr.SeekGE([]byte("b"))
+		names := make([]string, 0)
+		for itr.Next() {
+			e := itr.Entry()
+			model, ok := e.Value.(*kvtest.TestModel)
+			if !ok {
+				t.Fatalf("cannot read from store")
+			}
+			names = append(names, string(model.Name))
+		}
+		if itr.Err() != nil {
+			t.Fatalf("unexpected error: %v", itr.Err())
+		}
+
+		if diffs := deep.Equal(names, []string{"b", "c", "d"}); diffs != nil {
+			t.Fatalf("got wrong list of names: %v", diffs)
+		}
+	})
+
+	t.Run("primary iterator listing all values of partition", func(t *testing.T) {
+		itr, err := kv.NewPrimaryIterator(ctx, store, (&kvtest.TestModel{}).ProtoReflect().Type(),
+			firstPartitionKey, []byte(""), kv.IteratorOptionsFrom([]byte("")))
+		require.NoError(t, err)
+		require.NotNil(t, itr)
+		names := make([]string, 0)
+		for itr.Next() {
+			e := itr.Entry()
+			model, ok := e.Value.(*kvtest.TestModel)
+			if !ok {
+				t.Fatalf("cannot read from store")
+			}
+			names = append(names, string(model.Name))
+		}
+		if itr.Err() != nil {
+			t.Fatalf("unexpected error: %v", itr.Err())
+		}
+		itr.Close()
+
+		if diffs := deep.Equal(names, []string{"a", "aa", "b", "c", "d"}); diffs != nil {
+			t.Fatalf("got wrong list of names: %v", diffs)
+		}
+	})
+
+	t.Run("primary iterator listing with prefix", func(t *testing.T) {
+		itr, err := kv.NewPrimaryIterator(ctx, store, (&kvtest.TestModel{}).ProtoReflect().Type(),
+			secondPartitionKey, []byte("a"), kv.IteratorOptionsFrom([]byte("")))
+		require.NoError(t, err)
+		require.NotNil(t, itr)
+		names := make([]string, 0)
+		for itr.Next() {
+			e := itr.Entry()
+			model, ok := e.Value.(*kvtest.TestModel)
+			if !ok {
+				t.Fatalf("cannot read from store")
+			}
+			names = append(names, string(model.Name))
+		}
+		if itr.Err() != nil {
+			t.Fatalf("unexpected error: %v", itr.Err())
+		}
+		itr.Close()
+
+		if diffs := deep.Equal(names, []string{"a", "aa"}); diffs != nil {
+			t.Fatalf("got wrong list of names: %v", diffs)
+		}
+	})
+
+	t.Run("primary iterator listing empty value", func(t *testing.T) {
+		itr, err := kv.NewPrimaryIterator(ctx, store, (&kvtest.TestModel{}).ProtoReflect().Type(),
+			secondPartitionKey, []byte(""), kv.IteratorOptionsAfter([]byte("d")))
+		require.NoError(t, err)
+		require.NotNil(t, itr)
+		names := make([]string, 0)
+
+		for itr.Next() {
+			e := itr.Entry()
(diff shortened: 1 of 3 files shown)


pkg/kv/iterators_test.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+
+	err := sm.SetMsg(ctx, thirdPartitionKey, []byte("e"), &kvtest.TestModel{Name: []byte("notin")})
+	if err != nil {
+		t.Fatal("failed to set model", err)
+	}
+
+	t.Run("partition iterator listing all values of partition", func(t *testing.T) {
+		itr := kv.NewPartitionIterator(ctx, store, (&kvtest.TestModel{}).ProtoReflect().Type(), firstPartitionKey)
+		require.NotNil(t, itr)
```
nopcoder, line 53: Using 'testify/require' here why in prepare data we use 'if' - can we align in the test scope.

nopcoder (changes requested):
Note about using mix of testify and non-testify.
Also I think we should test the iterators in case on error - our last issues were related to calling SeekGE and Close while the first failed.

eden-ohana:
> Note about using mix of testify and non-testify. Also I think we should test the iterators in case on error - our last issues were related to calling SeekGE and Close while the first failed.

Added more scenarios, the only thing I couldn't simulate is iterating over an empty key - the store won't let me write it.

pkg/kv/iterators_test.go
```diff
@@ -0,0 +1,377 @@
...
+
+const (
+	firstPartitionKey  = "m"
+	secondPartitionKey = "ma"
+	thirdPartitionKey  = "mb"
+)
+
+func TestIterators(t *testing.T) {
+	ctx := context.Background()
+	store := kvtest.GetStore(ctx, t)
```
itai-david, line 21: Can we add other store drivers too?
The iterators rely heavily on the different `Scan` and `Next` implementations by the different stores
In fact, the panic we fixed in #4108 is a result of the way the postgres driver implements `Scan` (which is pretty similar to dynamodb, but totally different from the mem kv)

pkg/kv/iterators_test.go
```diff
@@ -0,0 +1,377 @@
...
+		}
+		itr.Close()
+
+		if diffs := deep.Equal(names, []string{"b", "c", "d"}); diffs != nil {
+			t.Fatalf("got wrong list of names: %v", diffs)
+		}
+	})
+
+	t.Run("partition iterator failed SeekGE", func(t *testing.T) {
+		itr := kv.NewPartitionIterator(ctx, store, (&kvtest.TestModel{}).ProtoReflect().Type(), "")
```
itai-david, line 108: Can we also add `SeekGE` failures when the partition key exists and the key itself is not found?
The 2 failure types behave differently on postgres and dynamodb drivers

itai-david (changes requested):
Thanks for these much needed tests

My main concern is that it only tests the mem store, while the different store implementations (by the different drivers) may affect the iterators significantly.
Maybe it is possible to call this test suite from each driver `store_test`, like we did with the rest of the kv tests? If so, I think it will be ideal.

Another point worth mentioning - failures due to empty partition key or key, are not the same as failures due to partition key or key that does not exist. The former is caught at the driver level and does not really reaches the underlying DB

pkg/kv/iterators_test.go
```diff
@@ -86,17 +97,42 @@ func TestIterators(t *testing.T) {
 		if itr.Err() != nil {
 			t.Fatalf("unexpected error: %v", itr.Err())
 		}
+		itr.Close()
```
nopcoder, line 100: Can use defer and call it after kv.NewPartitionIterator error check? Relevant for all tests in this file.

pkg/kv/iterators_test.go
```diff
@@ -86,17 +97,42 @@ func TestIterators(t *testing.T) {
...
+			model, ok := e.Value.(*kvtest.TestModel)
+			if !ok {
+				t.Fatalf("cannot read from store")
+			}
+			names = append(names, string(model.Name))
+		}
+		itr.Close()
+
+		if !errors.Is(itr.Err(), kv.ErrMissingPartitionKey) {
+			t.Fatalf("expected error: got %v", itr.Err())
```
nopcoder, line 125:
```suggestion
			t.Fatalf("expected error: %s, got %v", kv.ErrMissingPartitionKey, itr.Err())
```

pkg/kv/iterators_test.go
```diff
@@ -86,17 +97,42 @@ func TestIterators(t *testing.T) {
...
 			t.Fatalf("unexpected error: %v", itr.Err())
 		}
+		itr.Close()
 
 		if diffs := deep.Equal(names, []string{"b", "c", "d"}); diffs != nil {
 			t.Fatalf("got wrong list of names: %v", diffs)
 		}
 	})
 
+	t.Run("partition iterator failed SeekGE", func(t *testing.T) {
```
nopcoder, line 107: The fail here is - not found - the suggested test was related to the internal error as the ScanGE in some of implementations can fail. But not with mem driver.

pkg/kv/iterators_test.go
```diff
@@ -276,17 +321,41 @@ func TestIterators(t *testing.T) {
...
+			model, ok := e.Value.(*kvtest.TestModel)
+			if !ok {
+				t.Fatalf("cannot read from store")
+			}
+			names = append(names, string(model.Name))
+		}
+		itr.Close()
+
+		if !errors.Is(itr.Err(), kv.ErrMissingKey) {
+			t.Fatalf("expected error: got %v", itr.Err())
```
nopcoder, line 349:
```suggestion
			t.Fatalf("expected error: %s, got %v", kv.ErrMissingKey, itr.Err())
```

nopcoder (changes requested):
Think the iterator testing can be move to the 'store' testing - this way the same tests will run on all implementations. It will test the expected iterator align.

pkg/kv/iterators_test.go
eden-ohana: done

pkg/kv/iterators_test.go
eden-ohana: done

pkg/kv/iterators_test.go
eden-ohana: done

pkg/kv/iterators_test.go
eden-ohana: done

pkg/kv/iterators_test.go
eden-ohana: Not sure I understand the case, if the partition exist and the key doesn't the seek will not fail but return nil.

itai-david: `SeekGE` calls `store.Scan`. Both `postgres/store` and `dynamodb/store` has a `Scan` function that returns a `kv.EntriesIterator` interface. The implementation of this `Scan` calls an `internalScan` function that returns a pointer to a `EntriesIterator` struct. Now, if the failure occurrs in `Scan` (e,g `postgres/store` with empty `partitionKey`) it will return `nil`, but if the failiure is in `InternalScan`, `nil` will be returned to `Scan` which will return am interface pointing to a `nil` object, which is exactly what happend in #4108. Failure due to empty `partitionKey` will not cause the same panic.

pkg/kv/kvtest/store.go
```diff
@@ -60,6 +60,9 @@ func TestDriver(t *testing.T, name string, params kvparams.KV) {
 	t.Run("ScanPrefix", func(t *testing.T) { testScanPrefix(t, ms) })
 	t.Run("DeleteWhileIterating", func(t *testing.T) { testDeleteWhileIterPrefix(t, ms) })
 	t.Run("DeleteWhileIteratingSamePrefix", func(t *testing.T) { testDeleteWhileIterSamePrefix(t, ms) })
+	t.Run("PartitionIterator", func(t *testing.T) { testPartitionIterator(t, ms) })
+	t.Run("PrimaryIterator", func(t *testing.T) { testPrimaryIterator(t, ms) })
+	t.Run("SecondaryIterator", func(t *testing.T) { testSecondaryIterator(t, ms) })
```
itai-david, line 65: :fire: :+1: :rocket: 
Thank you

pkg/kv/iterators_test.go
eden-ohana: Got it. To implement it we need a mock for the store. 
It will be implemented on https://github.com/treeverse/lakeFS/pull/4112, not apart of the unit tests

pkg/kv/kvtest/iterators.go
```diff
@@ -0,0 +1,428 @@
...
+			names = append(names, string(model.Name))
+		}
+		if itr.Err() != nil {
+			t.Fatalf("unexpected error: %v", itr.Err())
+		}
+
+		if diffs := deep.Equal(names, []string{}); diffs != nil {
+			t.Fatalf("got wrong list of names: %v", diffs)
+		}
+	})
```
nopcoder, line 427: Can you explain the difference and why in one test we get an error while on the other we get no values?
Maybe it is just because of the test name are very similar I expected the same behavior.

pkg/kv/kvtest/iterators.go
```diff
@@ -0,0 +1,428 @@
...
+				t.Fatalf("cannot read from store")
+			}
+		}
+
+		if !errors.Is(itr.Err(), kv.ErrMissingKey) {
+			t.Fatalf("expected error: %s, got %v", kv.ErrMissingKey, itr.Err())
+		}
+	})
+
+	t.Run("primary not found", func(t *testing.T) {
```
nopcoder, line 404: primary?

pkg/kv/kvtest/iterators.go
```diff
@@ -0,0 +1,428 @@
...
+
+	t.Run("entry not found", func(t *testing.T) {
+		itr, err := kv.NewSecondaryIterator(ctx, store, (&TestModel{}).ProtoReflect().Type(),
+			thirdPartitionKey, []byte(""), []byte(""))
+		if err != nil {
+			t.Fatalf("failed to create secondary iterator: %v", err)
+		}
+		defer itr.Close()
+
+		for itr.Next() {
```
nopcoder, line 391: in case we expect an error we should test Next() returns false

nopcoder (changes requested):

pkg/kv/kvtest/iterators.go
eden-ohana: yes, secondary iterates over (key1, value1) -> (value1, value2) and returns value 2. entry not found means key1 not found, primary not found means value1 not found. changed the names for it to be more clear.

pkg/kv/kvtest/iterators.go
```diff
@@ -0,0 +1,428 @@
...
+	})
+
+	t.Run("failed SeekGE partition not found", func(t *testing.T) {
+		itr := kv.NewPartitionIterator(ctx, store, (&TestModel{}).ProtoReflect().Type(), "")
+		if itr == nil {
+			t.Fatalf("failed to create partition iterator")
+		}
+		defer itr.Close()
+		itr.SeekGE([]byte("d"))
+		for itr.Next() {
```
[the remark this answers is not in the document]
eden-ohana: done

pkg/kv/kvtest/iterators.go
eden-ohana: It returns false because it will exit the loop

pkg/kv/kvtest/iterators.go
```diff
@@ -98,12 +98,9 @@ func testPartitionIterator(t *testing.T, ms MakeStore) {
...
 		itr.SeekGE([]byte("d"))
-		for itr.Next() {
-			e := itr.Entry()
-			_, ok := e.Value.(*TestModel)
-			if !ok {
-				t.Fatalf("cannot read from store")
-			}
+
+		if itr.Next() {
+			t.Fatalf("seekGE error expected to be false")
```
nopcoder, line 103:
```suggestion
			t.Fatalf("key not found expected to be false")
```

pkg/kv/kvtest/iterators.go
```diff
@@ -0,0 +1,425 @@
...
+
+	t.Run("value as key not found", func(t *testing.T) {
+		itr, err := kv.NewSecondaryIterator(ctx, store, (&TestModel{}).ProtoReflect().Type(),
+			thirdPartitionKey, []byte(""), []byte("e"))
+		if err != nil {
+			t.Fatalf("failed to create secondary iterator: %v", err)
+		}
+		defer itr.Close()
+		names := make([]string, 0)
+		for itr.Next() {
```
nopcoder, line 409: - should verify that next returns false.
- names diff can be removed as we not suppose to go into a loop here.
-

pkg/kv/kvtest/iterators.go
nopcoder: True, so we need to verify it returns false and delete a code that never should run.

nopcoder (changes requested):

itai-david (approved):
LGTM. Thank you

nopcoder (approved):

PR #4137, opened 2022-09-08
Add pgxpool metrics to kv/postgres implementation

### LINKED ISSUE #4138: KV postgres information about pgxpool
The current KV Postgres implementation uses `pgxpool` for database connection pool.
The package expose metrics usage that an be exposed as Prometheus metrics.
Using these metrics we can help tune and investigate performance issues.

nopcoder:
Add https://pkg.go.dev/github.com/jackc/pgx/v4/pgxpool#Stat to prometheus metrics we collect
Name | Description
-- | --
pgxpool_acquire_count | Cumulative count of successful acquires from the pool.
pgxpool_acquire_duration_ns | Total duration of all successful acquires from the pool in nanoseconds.
pgxpool_acquired_conns | Number of currently acquired connections in the pool.
pgxpool_canceled_acquire_count | Cumulative count of acquires from the pool that were canceled by a context.
pgxpool_constructing_conns | Number of conns with construction in progress in the pool.
pgxpool_empty_acquire | Cumulative count of successful acquires from the pool that waited for a resource to be released or constructed because the pool was empty.
pgxpool_idle_conns | Number of currently idle conns in the pool.
pgxpool_max_conns | Maximum size of the pool.
pgxpool_total_conns | Total number of resources currently in the pool. The value is the sum of ConstructingConns, AcquiredConns, and IdleConns.

Close #4138


nopcoder:
@itaiad200 wait with the review there is a uniqueness issue when we open more than one store to the same database while having one set of metrics. Need to think to think how to handle this case or make sure we do not instance more than one.

nopcoder:
- Updated the code to enable configurable metrics for postgres.
- Updated the test not to open two stores in parallel

nopcoder:
@itaiad200 ready for ready

pkg/kv/params/database.go
```diff
@@ -10,12 +10,27 @@ type KV struct {
 	DynamoDB *DynamoDB
 }
 
+// EnableMetrics returns a copy of params with Metrics enabled for Postgres.
+// This is used only when we like to get parameters for the main run of lakeFS
+// as the is only one metrics set we expose
```
N-o-Z, line 15: Rephrase?

pkg/kv/postgres/migrate.go
```diff
@@ -20,8 +20,10 @@ import (
 	"github.com/treeverse/lakefs/pkg/logging"
```
N-o-Z, line 20: We don't need this module anymore

docs/reference/monitor.md
```diff
@@ -48,6 +48,15 @@ In addition, lakeFS exposes the following metrics to help monitor your deploymen
 | dynamo_request_duration_seconds  | Time spent doing DynamoDB requests                          | **operation**: DynamoDB operation name
```
N-o-Z, line 48: We need to update the sizing guide metrics section for both dynamodb and postgres

pkg/kv/params/database.go
```diff
@@ -10,12 +10,27 @@ type KV struct {
...
+func (p KV) EnableMetrics() KV {
```
N-o-Z, line 16: I don't think this should be here, it's specific logic for postgres.
We should define the parameter as bool pointer in the config.go struct.
On GetKVParams we should check if it's nil and assign it the default value

pkg/kv/params/database.go
```diff
@@ -10,12 +10,27 @@ type KV struct {
...
+	return p
+}
+
 type Postgres struct {
 	ConnectionString      string
 	MaxOpenConnections    int32
 	MaxIdleConnections    int32
 	ConnectionMaxLifetime time.Duration
 	ScanPageSize          int32
+	Metrics               bool
```
N-o-Z, line 33: Add the new param to the configuration doc

N-o-Z (changes requested):
Thanks Barak,
I've done a partial review of the code

pkg/kv/params/database.go
nopcoder: Currently is it not configurable - having it as part of the postgres configuration is to enable us to control on which pgxpool we like to have the metrics enabled.

pkg/kv/params/database.go
nopcoder: Didn't want the metrics we expose to be optional or configurable in the application configuration.
Needed a way to configure the metrics registration for a specific pgx pool.
While the process is running there are cases, like while performing 'setup', that we open more than one kv store.
In this case we need to make sure that the initial connection database pool will be the only one that reports these metrics.

pkg/kv/postgres/migrate.go
nopcoder: logging? it is using in this file.

N-o-Z: postgres/migrate.go

N-o-Z (approved):
Thanks!

pkg/kv/postgres/store.go
```diff
@@ -7,10 +7,12 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/IBM/pgxpoolprometheus"
 	"github.com/georgysavva/scany/pgxscan"
 	"github.com/jackc/pgconn"
 	"github.com/jackc/pgx/v4"
 	"github.com/jackc/pgx/v4/pgxpool"
+	"github.com/prometheus/client_golang/prometheus"
```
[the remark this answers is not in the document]
nopcoder: pgxpoolprometheus uses prometheus api to define a new Collector. It uses the go promethus api, so we need both.

docs/reference/monitor.md
nopcoder: The specific metrics added for pxpool currently helps monitoring connection usage. It can help sizing, but it is not like the usage metrics we have for dynamodb.
Postgres performance monitoring can be does externally, not like the internal pool we monitor here.

PR #4139, opened 2022-09-08
Refactor GC Spark job with explicit partitions and fewer shuffles

### LINKED ISSUE #4135: Refactor GC Spark job with explicit partitions and fewer shuffles
# Why

Prototyping performed by @johnnyaug reveals that Spark can perform operations very similar to GC in _much less_ time, if we write temporary Parquet files.

# What

Parquet should have no real advantage over internal Hadoop partitions.  But... you do get a repartitioning step around the keys.  So refactor GC to repartition everything around keys at each step -- but make each step simpler, to allow Spark to do its work.

# How

(Pseudocode!)

- Get lists of commits to _keep_ and to _expire_.
- flatMap both list of commits, expand each metarange to its ranges.
  - We now have 2 DFs: rangeIDsToKeep, rangeIDsToExpire.
- Repartition both DFs by rangeID (the only column...) _using the same partitioner_.
- (Maybe distinctify each DF before proceeding.  But try first without, which might be faster!)
- Anti JOIN (even semi-outer, if we can) the two DFs.  (_They're already partitioned to make this faster!)
  - Result: non-intersecting (distinct!) DF effectiveRangeIDsToExpire that removes _en masse_ all range IDs that will be entirely kept. 
- flatMap both DFs, expand each rangeID to its physical addresses.
  - We now have 2 DFs: physToKeep, physToExpire.
- Repartition both DFs by physical address (the only column...) _using the same partitioner_.
- (Maybe distinctify each DF before proceeding.  But try first without, which might be faster!)
- Anti JOIN (even semi-outer, if we can) the two DFs `expire -- keep`.  (_They're already partitioned to make this faster!)
  - Result: DF of physical addresses that are in expiring commits but _not_ in any commit to keep.
- Delete those addresses :-)
-

(_updated_ to keep the range ids anti join rather than a symmetrical differences. That would not keep the full difference as a curvy antijoin!)

arielshaqed:
Fixes #4135.

### DIFF
--- a/clients/spark/build.sbt
+++ b/clients/spark/build.sbt
@@ -73,7 +73,9 @@ def generateCoreProject(buildType: BuildType) =
         "org.scalatest" %% "scalatest" % "3.2.9" % "test",
         "com.dimafeng" %% "testcontainers-scala-scalatest" % "0.40.10" % "test",
         "com.lihaoyi" %% "upickle" % "1.4.0" % "test",
-        "com.lihaoyi" %% "os-lib" % "0.7.8" % "test"
+        "com.lihaoyi" %% "os-lib" % "0.7.8" % "test",
+        // Test with an up-to-date fasterxml.
+         "com.fasterxml.jackson.module" %% "jackson-module-scala" % "2.12.2" % "test"
       ),
       Test / logBuffered := false,
       // Uncomment to get (very) full stacktraces in test:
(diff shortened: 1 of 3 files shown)


talSofer (commented):
Great tests! thank you

arielshaqed:
This should be reviewable by commit.  Doing that will let you read the business logic separately from the component tests.  There's enough in the component testing to make it interesting (we were not able to have even what little component tests this adds beforehand, the refactoring helped do that too...).

arielshaqed:
Pushed as #4139 .

PR #4143, opened 2022-09-08
Add delete operation for GC rules

### LINKED ISSUE #4142: Allow deleting GC rules
Once the user has created GC rules, it is not possible to delete them.

Add this option to the API, lakectl and a delete button to the UI.

johnnyaug:
Fixes #4142.

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -2272,6 +2272,23 @@ func (c *Controller) SetGarbageCollectionRules(w http.ResponseWriter, r *http.Re
 	writeResponse(w, http.StatusNoContent, nil)
 }
 
+func (c *Controller) DeleteGarbageCollectionRules(w http.ResponseWriter, r *http.Request, repository string) {
+	if !c.authorize(w, r, permissions.Node{
+		Permission: permissions.Permission{
+			Action:   permissions.SetGarbageCollectionRulesAction,
+			Resource: permissions.RepoArn(repository),
+		},
+	}) {
+		return
+	}
+	ctx := r.Context()
+	err := c.Catalog.SetGarbageCollectionRules(ctx, repository, nil)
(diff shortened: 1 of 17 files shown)


pkg/api/controller.go
```diff
@@ -2272,6 +2272,23 @@ func (c *Controller) SetGarbageCollectionRules(w http.ResponseWriter, r *http.Re
...
+	if !c.authorize(w, r, permissions.Node{
+		Permission: permissions.Permission{
+			Action:   permissions.SetGarbageCollectionRulesAction,
+			Resource: permissions.RepoArn(repository),
+		},
+	}) {
+		return
+	}
+	ctx := r.Context()
+	err := c.Catalog.SetGarbageCollectionRules(ctx, repository, nil)
```
talSofer, line 2285: What does it translate to? is it saving an empty json?

pkg/graveler/retention/garbage_collection_manager.go
```diff
@@ -84,6 +84,9 @@ func (m *GarbageCollectionManager) GetRules(ctx context.Context, storageNamespac
 	if err != nil {
```
talSofer, line 84: I would document that this is when no gc rules exits and line 87 is when gc rules were deleted

talSofer (approved):
Thanks!! added minor commnets

pkg/api/controller.go
johnnyaug: It saves an empty file (not an empty json)

pkg/graveler/retention/garbage_collection_manager.go
johnnyaug: Adding a comment that an empty file means no GC rules. I don't think it matters whether they have been deleted or otherwise.

PR #4148, opened 2022-09-09
Remove code leftovers: id translator used by recorded

### LINKED ISSUE #4152: Remove unused upload ID translator from the code
Remove code leftover from the time there was a traffic recorder and we needed to control the upload ID in record/playback.

nopcoder:
ID translator used by adapter layer to solve recorded playback testing.
Recoded use and code no longer used and the code can be dropped.

Fix #4152


arielshaqed (approved):
Words cannot express how much I like this PR!

PR #4154, opened 2022-09-11
Lookup commit when possible while commit prefix search

### LINKED ISSUE #4155: Optimize GetCommitByPrefix for full commit ID lookup
Get commit by prefix can use get commit when the prefix length is a full commit ID.
Some of the entry points in Graveler accepts a reference as argument.
While we parse the reference (Dereference) we lookup the commit information using a prefix search, in a lot of cases - get branch, commit log and etc. We are getting to the GetCommitByPrefix using a full commit ID retried from Graveler.
In this case, a simple get by direct ID will perform better than using a scan / use iterator for lookup.

nopcoder:
Fix #4155

### DIFF
--- a/pkg/graveler/ref/manager.go
+++ b/pkg/graveler/ref/manager.go
@@ -26,6 +26,9 @@ const MaxBatchDelay = time.Millisecond * 3
 
 const BatchUpdateSQLSize = 10000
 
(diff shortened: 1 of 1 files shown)


eden-ohana (approved):
LGTM

PR #4156, opened 2022-09-11
Design: GC on Parquet

### LINKED ISSUE #4157: Research: use Parquet format as the input for GC
Garbage collection in lakeFS is essentially an anti-join between the lists of "expired" and "active" addresses.
Check whether performing this operation on Parquet files improves performance, and map the risk and considerations.

johnnyaug:
Resolves #4157

### DIFF
--- a/design/open/gc-on-parquet.md
+++ b/design/open/gc-on-parquet.md
@@ -14,7 +14,8 @@ Parquet is more than a file format. It's a breathing open-source project aiming 
 #### 1. Sync
 Ranges of the repository are translated into Parquet files. These are stored in a new metadata location under the repository's storage namespace.
 
-#### 2. Sanity: verify all the ranges required for this run are present, and that the number of entries in each range is correct.
+#### 2. Sanity
+Verify all the ranges required for this run are present, and that the number of entries in each range is correct.
 
 #### 3. Start
 Do an anti-join extracting all the addresses that need to be deleted.
--- a/design/open/gc-on-parquet.md
+++ b/design/open/gc-on-parquet.md
@@ -1,0 +1,24 @@
+## Garbage collection using Parquet format
+
+### Motivation
+
+Garbage collection in lakeFS is struggling at large scale.
+
+### Why Parquet?
+
+Parquet is more than a file format. It's a breathing open-source project aiming to provide exactly what we're looking for here: efficient data retrieval. GC is essentially a huge anti-join, and it may be very difficult to achieve a similar optimization in our dedicated format.
+
+
+### Steps
+
+#### 1. Sync
+Ranges of the repository are translated into Parquet files. These are stored in a new metadata location under the repository's storage namespace.
+
+#### 2. Sanity: verify all the ranges required for this run are present, and that the number of entries in each range is correct.
+
+#### 3. Start
+Do an anti-join extracting all the addresses that need to be deleted.
+
+### Considerations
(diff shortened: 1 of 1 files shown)


design/open/gc-on-parquet.md
```diff
@@ -0,0 +1,25 @@
...
+#### 1. Sync
+Ranges of the repository are translated into Parquet files. These are stored in a new metadata location under the repository's storage namespace.
+
+#### 2. Sanity
+Verify all the ranges required for this run are present, and that the number of entries in each range is correct.
+
+#### 3. Start
+Do an anti-join extracting all the addresses that need to be deleted.
+
+### Considerations
```
talSofer, line 23: Another thing that comes to mind is that this solution won't be reusable by GC uncommitted, since it is specific to lakeFS metadata processing.

johnnyaug: Not sure I understand - why is it more/less usable then using a dedicated format?

design/open/gc-on-parquet.md
```diff
@@ -0,0 +1,25 @@
...
+Parquet is more than a file format. It's a breathing open-source project aiming to provide exactly what we're looking for here: efficient data retrieval. GC is essentially a huge anti-join, and it may be very difficult to achieve a similar optimization in our dedicated format.
+
+
+### Steps
+
+#### 1. Sync
+Ranges of the repository are translated into Parquet files. These are stored in a new metadata location under the repository's storage namespace.
+
+#### 2. Sanity
+Verify all the ranges required for this run are present, and that the number of entries in each range is correct.
```
arielshaqed, line 18: Not sure why this is needed.  If it's a parquet from a range, it's good.  (If we don't trust it, why trust anything else, really?)

design/open/gc-on-parquet.md
```diff
@@ -0,0 +1,25 @@
...
+
+#### 3. Start
+Do an anti-join extracting all the addresses that need to be deleted.
+
+### Considerations
+1. Storage: this requires using more of the user's storage. I don't think that's an issue since metadata is usually very small compared to the data size. In the future we can also optimize the set of ranges we actually bring there.
```
arielshaqed, line 24: This will essentially double storage used for metadata.  (Pretty much all uncompressible data in a range is the random part of the object address, so a column store cannot do much more than the RocksDB / LevelDB file format.)

design/open/gc-on-parquet.md
```diff
@@ -0,0 +1,25 @@
...
+1. Risk: we are using a copy of the metadata instead of the actual one. We need to be careful not to miss anything, since missing data may potentially cause innocent objects to be deleted.  This is a risk that any GC solution needs to take into account.
```
arielshaqed, line 25: Additional considerations:

* Increased ops burden: we now have more objects to handle.  What happens if an object is missing?  Sure, we can fix it -- that's a burden.

arielshaqed (approved):
Thanks!

There is a quantitative product question that we can ask here: would users be fine with doubling their metadata usage?  I suspect that the answer is yes, but perhaps we can ask?

design/open/gc-on-parquet.md
johnnyaug: At the very least, I want to validate that all the ranges are _there_.
Missing a range is a disaster. 
Since we're creating it using our InputFormat, I want to double-check that we have the correct number of addresses. Again, missing an address is also a disaster.

PR #4168, opened 2022-09-13
Updated instructions for installation on Google Cloud Run

### LINKED ISSUE #4177: Google Cloud Run instructions are wrong in docs
Google Cloud Run instructions are wrong in docs

kesarwam:
Updated instructions for installation on Google Cloud Run.

Fix #4177


iddoavn (approved):

docs/deploy/gcp.md
```diff
@@ -59,21 +59,71 @@ For example, if you install lakeFS on GKE, you need to deploy the SQL Auth Proxy
...
+   spec:
+     template:
+       metadata:
+         annotations:
+           autoscaling.knative.dev/maxScale: '5'
+           run.googleapis.com/client-name: cloud-console
+           run.googleapis.com/cloudsql-instances: [POSTGRES_INSTANCE_NAME]
+       spec:
+         containers:
+         - image: gcr.io/lakefs/treeverse/lakefs@sha256:18be2f7e86f7572160909b7dcae2e50be5704cb174eeaf4fb61274def3536009
```
nopcoder, line 80: This is internal repository in our cloud - we need to address using lakeFS under Google Cloud Run as it will not accept the one we use from docker hub.
So this one will not work for the end-user - unless they will push our image to their repository and use their address here.

docs/deploy/gcp.md
```diff
@@ -59,21 +59,71 @@ For example, if you install lakeFS on GKE, you need to deploy the SQL Auth Proxy
...
+           ports:
+           - name: http1
+             containerPort: 8000
+           env:
+           - name: LAKEFS_BLOCKSTORE_TYPE
+             value: gs
+           - name: LAKEFS_BLOCKSTORE_GS_CREDENTIALS_JSON
+             value: '[YOUR SERVICE ACCOUNT JSON STRING]'
```
nopcoder, line 88: Use secrets or other mechanism to pas the credentials - but not environment vars.

docs/deploy/gcp.md
```diff
@@ -59,21 +59,71 @@ For example, if you install lakeFS on GKE, you need to deploy the SQL Auth Proxy
...
+           - name: LAKEFS_DATABASE_CONNECTION_STRING
+             value: [DATABASE_CONNECTION_STRING]
+           - name: LAKEFS_STATS_ENABLED
+             value: 'false'
```
nopcoder, line 92: Please remove this one

docs/deploy/gcp.md
```diff
@@ -59,21 +59,71 @@ For example, if you install lakeFS on GKE, you need to deploy the SQL Auth Proxy
...
+           - name: INSTANCE_CONNECTION_NAME
+             value: [POSTGRES_INSTANCE_NAME]
+           - name: CLOUD_SQL_CONNECTION_NAME
+             value: [POSTGRES_INSTANCE_NAME]
+           - name: LAKEFS_AUTH_ENCRYPT_SECRET_KEY
+             valueFrom:
+               secretKeyRef:
+                 key: '2'
+                 name: [ENCRYPTION_SECRET_KEY]
```
nopcoder, line 101: You need to explain how this one will work - we reference a secret here but, we need to add it first

docs/deploy/gcp.md
```diff
@@ -59,21 +59,71 @@ For example, if you install lakeFS on GKE, you need to deploy the SQL Auth Proxy
...
+
+   <img src="{{ site.baseurl }}/assets/img/gcp_1.png" alt="Google Cloud Run UI" />
+
+1. Enter "allUsers" in "New principals" column, select "Cloud Run Invoker" role and save:
+
+   <img src="{{ site.baseurl }}/assets/img/gcp_2.png" alt="Add principals and roles" />
+
+1. Select service name to go to service details and click on the URL to go to lakeFS UI:
+
+   <img src="{{ site.baseurl }}/assets/img/gcp_3.png" alt="lakeFS URL" />
```
nopcoder, line 126: 1. Need to explain what we are doing not just what to do.
2. We explain how to do the above operations using `gcloud` - we should instruct the user also the following using gcloud and not switching to the UI.

nopcoder (changes requested):

docs/deploy/gcp.md
kesarwam: @nopcoder What instructions do customers need to push lakeFS image to their repository and use their address here? I reviewed GCP Container Registry. I can view our images but can't create new image to figure this out.

Can we make our image public? I saw this info under Settings in Container Registry UI:
"You can make the images under a Container Registry host accessible to the public. This allows users to pull images without authentication. The project will still be charged for Cloud Storage network usage."

docs/deploy/gcp.md
kesarwam: @nopcoder
Should it be like this:
           - name: LAKEFS_BLOCKSTORE_GS_CREDENTIALS_JSON
             valueFrom:
               secretKeyRef:
                 name: [YOUR_SERVICE_ACCOUNT_JSON_STRING_SECRET_KEY]

Also, do we have to provide gcloud CLI command to create secrets or should we leave up to customer's GCP admin to handle this?

docs/deploy/gcp.md
kesarwam: @nopcoder 
Do we have to provide gcloud CLI command to create secrets or should we leave up to customer's GCP admin to handle this?

You had provided me installation instructions for Google Cloud Run. What does `key: '2'` mean here? Is it version # for the key.

docs/deploy/gcp.md
kesarwam: @nopcoder 
You had provided me installation instructions for Google Cloud Run. I am just documenting it. Do you have `gcloud` command for this?

docs/deploy/gcp.md
nopcoder: Our docker image is public and published under DockerHub. But Google Cloud Run requires to use GCP Container Registry. I pushed one of the lakeFS docker images manually to the GCP registry in order to use it. It is not a valid solution to make the specific version public - we need to have a release process to publish to such image or instruct to use how to push one of the current images to google. (first option is preferred).

docs/deploy/gcp.md
nopcoder: @kesarwam This is the syntax (missing 'key' property). But, you will need to instruct (documentation or commands) to add the value of the service account as secret. So, the name as the value of '[YOUR_SERVICE_ACCOUNT_JSON_STRING_SECRET_KEY]' is not right. We like to pull it from the secrets.

docs/deploy/gcp.md
nopcoder: @kesarwam I prefer full instructions that works, we do not require the use to have the same names and conventions but it will be a lot easy to follow instructions from start to end that will work. This is also why I had a comment about switching to the UI to perform operations that probably can be done from 'gcloud' cli.

I provided you a way to deploy my lakeFS as a service - The key (2) is my second version of the secret I used to deploy. Key is required based on the docs.

docs/deploy/gcp.md
nopcoder: @kesarwam The instructions I gave you can't go directly to documentation, if you remember I specified when we talked what the database connection is environment and should be secret. And that I've uploaded an image to google to make it work. Documenting the process need to generalize and give instructions that explain how the same process works for all.

If you have instruction and they will work for other people, they are good instructions - the above key, image, account will work only on our environment as these are the instructions I gave you.

nopcoder (changes requested):
Added my answers

nopcoder:
@kesarwam is this still relevant? can close?

PR #4180, opened 2022-09-14
lakefs migrate without kv migrations

### LINKED ISSUE #4179: lakeFS migrate command should be enabled for kv
In order to support data migration for future KV changes, we need to enable 'lakefs migrate' commands.
This is the first version of kv and no active migrations, but the command should be available.

nopcoder:
Fix #4179

### DIFF
--- a/cmd/lakefs/cmd/migrate.go
+++ b/cmd/lakefs/cmd/migrate.go
@@ -1,11 +1,13 @@
 package cmd
 
 import (
+	"context"
+	"errors"
 	"fmt"
+	"os"
 
 	"github.com/spf13/cobra"
 	"github.com/treeverse/lakefs/pkg/kv"
-	"github.com/treeverse/lakefs/pkg/logging"
 )
 
 // migrateCmd represents the migrate command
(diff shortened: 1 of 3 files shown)


cmd/lakefs/cmd/migrate.go
```diff
@@ -15,23 +18,41 @@ var versionCmd = &cobra.Command{
...
+
+		ctx := cmd.Context()
+		kvParams := cfg.GetKVParams()
+		kvStore, err := kv.Open(ctx, kvParams)
+		if err != nil {
+			logger.WithError(err).Fatal("Failed to open KV store")
+		}
+		defer kvStore.Close()
+
+		version, err := kv.GetDBSchemaVersion(ctx, kvStore)
```
N-o-Z, line 32: If it's a new installation before setup - maybe we should return a different error on 'not found', such as 'KV not initialized' or something similar?

N-o-Z: Perhaps use ValidateSchemaVersion

N-o-Z (changes requested):
Thanks, please see comments

cmd/lakefs/cmd/migrate.go
nopcoder: Updated base don modified ValidateSchema.
Will do followup PR to remove additional DB which we no longer use.
And enable retry pg connect, found it missing in the new kv implementation.

cmd/lakefs/cmd/migrate.go
```diff
@@ -18,32 +20,49 @@ var versionCmd = &cobra.Command{
...
-			return
-
-		}
-		logger.WithField("schema_version", version).Info("DB schema version")
+		version := MustValidateSchemaVersion(ctx, kvStore)
 		fmt.Printf("Database schema version: %d\n", version)
 	},
 }
 
+func MustValidateSchemaVersion(ctx context.Context, kvStore kv.Store) int {
```
N-o-Z, line 38: Does it need to be public?

cmd/lakefs/cmd/migrate.go
```diff
@@ -18,32 +20,49 @@ var versionCmd = &cobra.Command{
...
-		logger := logging.Default()
 		cfg := loadConfig()
-
-		ctx := cmd.Context()
 		kvParams := cfg.GetKVParams()
+		ctx := cmd.Context()
 		kvStore, err := kv.Open(ctx, kvParams)
 		if err != nil {
-			logger.WithError(err).Fatal("Failed to open KV store")
+			_, _ = fmt.Fprintf(os.Stderr, "Failed to open KV store: %s\n", err)
```
N-o-Z, line 28: Isn't it the same as log.Fatal?

N-o-Z (approved):
Approved with comments

cmd/lakefs/cmd/migrate.go
nopcoder: Removed the use of logging for this command.
The logger fatal will print out to the log + exit(1).
The Fprintf will write to std err without and added a call to os.Exit after.
The log output format and file/stderr is based on the lakefs configuration, while the Fprint is not.

cmd/lakefs/cmd/migrate.go
nopcoder: No - its my default. No one imports cmd package.

nopcoder: Updated to be private

Issue #4182, opened 2022-09-14
[Everything Bagel] lakefs-setup and lakefs containers fail 


rmoff:
Running the latest version of the [Everything Bagel](https://github.com/treeverse/lakeFS/blob/master/deployments/compose/docker-compose.yml) compose file. 

The `lakefs-setup` container fails: 

```
╰─$ docker-compose logs lakefs-setup
lakefs-setup  | time="2022-09-14T16:44:04Z" level=info msg="Config loaded" func=cmd/lakefs/cmd.initConfig file="cmd/root.go:103" fields.file=/home/lakefs/.lakefs.yaml file="cmd/root.go:103" phase=startup
[...]

PR #4183, opened 2022-09-14
Fix broken Everything Bagel (pin lakefs image version)

rmoff:
ref: #4182

This change pins the lakefs docker image used in Everything Bagel so that it continues to work. 

The 0.80 version of the image expects different environment variables (`LAKEFS_DATABASE_TYPE` / `LAKEFS_DATABASE_POSTGRES_CONNECTION_STRING`) which is a subsequent issue to fix assuming Everything Bagel is to keep track with the latest release.

### DIFF
--- a/deployments/compose/docker-compose.yml
+++ b/deployments/compose/docker-compose.yml
@@ -2,7 +2,7 @@ version: "3"
 services:
 
   lakefs-setup:
-    image: treeverse/lakefs:latest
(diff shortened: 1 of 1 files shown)


nopcoder (approved):
Thanks @rmoff, this is great - as far as I remember there is a pending PR that should align this one with the latest changes soon. Having this fix should help our curernt users.

PR #4188, opened 2022-09-15
Add lakefs cmd local config flag

### LINKED ISSUE #4186: lakefs run flag to configure local default parameters
Add a flag to the `lakefs run` command that will enable default local configuration variables running with local blockstore and KV driver.
This will make it easy to use the docker run command without providing any additional parameters

N-o-Z:
Closes #4186


N-o-Z:
@nopcoder Not such an elegant solution - feel free to suggest/push improvements

nopcoder (approved):

PR #4190, opened 2022-09-16
Spark client: wrap API calls to lakeFS with retries

### LINKED ISSUE #4189: GC: improve ApiClient resiliency
Currently, https://github.com/treeverse/lakeFS/blob/master/clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala GC is using has no retry mechanism for Api calls it makes to the lakeFS server. This makes the application rely on Spark’s retry mechanism causing the failure of a whole stage when a single UDF fails instead of immediately retrying the call at the API client level. This means that sometimes a stage that takes a few hours fails due to a single API call failure (Epcor saw that ~30 API calls are batched to a single stage).

talSofer:
Closes #4189

Using https://failsafe.dev/ 

### How was this tested

Tested manually on a repo with objects to be garbage collected and GC finished successfully.


clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -198,28 +219,48 @@ class ApiClient private (conf: APIConfigurations) {
...
   // Instances of case classes are compared by structure and not by reference https://docs.scala-lang.org/tour/case-classes.html.
   case class StorageNamespaceCacheKey(
       repoName: String,
       storageClientType: StorageClientType
   )
 }
+
+class RequestRetryWrapper(val numRetries: Int = 5) {
+    private val retryPolicy : Policy[Any] = RetryPolicy.builder()
+      .withDelay(Duration.ofMillis(500))
```
johnnyaug, line 258: Is there a reason not to use exponential backoff? Worried about bombing the server.

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -2,17 +2,18 @@ package io.treeverse.clients
...
 import io.treeverse.clients.StorageUtils.{StorageTypeAzure, StorageTypeS3}
 import com.google.common.cache.{Cache, CacheBuilder, CacheLoader, LoadingCache}
+import dev.failsafe.function.CheckedSupplier
 import io.treeverse.clients.ApiClient.TIMEOUT_NOT_SET
 
 import java.net.URI
 import java.util.concurrent.{Callable, TimeUnit}
+import java.time.Duration
+import collection.JavaConverters._
+import dev.failsafe.{Failsafe, Policy, RetryPolicy}
```
arielshaqed, line 16: Nit: Let's keep all the `failsafe` imports together if possible!  Also, if we have a blank line at l. 13/11 then let's keep standard library (`java.*`, `collection.*`) below it and everything else above it.

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -139,8 +142,11 @@ class ApiClient private (conf: APIConfigurations) {
       })
 
   def keyToStorageNamespace(key: StorageNamespaceCacheKey): String = {
-    val repo = repositoriesApi.getRepository(key.repoName)
 
+    val getRepo = new dev.failsafe.function.CheckedSupplier[Repository]() {
+      def get(): Repository = repositoriesApi.getRepository(key.repoName)
+    }
+    val repo = retryWrapper.wrapWithRetry(getRepo)
```
johnnyaug, line 149:
```suggestion
    val repo = retryWrapper.wrapWithRetry(() => repositoriesApi.getRepository(key.repoName))
```

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -161,30 +167,45 @@ class ApiClient private (conf: APIConfigurations) {
...
-      repoName,
-      new GarbageCollectionPrepareRequest().previousRunId(previousRunID)
-    )
+    val prepareGcCommits = new dev.failsafe.function.CheckedSupplier[GarbageCollectionPrepareResponse]() {
+      def get(): GarbageCollectionPrepareResponse = retentionApi.prepareGarbageCollectionCommits(
+        repoName,
+        new GarbageCollectionPrepareRequest().previousRunId(previousRunID)
+      )
+    }
+    retryWrapper.wrapWithRetry(prepareGcCommits)
```
johnnyaug, line 176:
```suggestion
    retryWrapper.wrapWithRetry(() =>
      retentionApi.prepareGarbageCollectionCommits(
        repoName,
        new GarbageCollectionPrepareRequest().previousRunId(previousRunID)
      )
    )
```
etc etc.

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
arielshaqed, line 258: * I think that exponential backoff makes sense here rather than this linear/constant backoff, at least for the massive calls like `getRange`.  Otherwise I worry that a Spark cluster with a lot of processing power can still overload a weak lakeFS cluster.
* Does failsafe automatically add jitter to its calls?  If not then we should ask it to add jitter.  Otherwise all retries from a cluster can happen almost exactly at once!  (If the package offers other guidance, then of course we should go with what the experts say...).
* It offers a total of 2s backoff, which is not a lot for a Spark job that runs for very many minutes.  I'd be perfectly happy with waiting up to 20 seconds (say) before deciding that this job has failed).
* OTOH, this makes the read timeout configuration _really hard to use_!  If I configure a read timeout of 10s, then it waits in total for 5*10s + 4*0.5s = 52s.  That will not be very clear to users.  Two possible fixes (others are possible!) might be to expose all parameters, or have 2 parameters `read_timeout` and `total_read_timeout` and somehow tell failsafe to handle retries that take up to `total_read_timeout`.

clients/spark/core/src/test/scala/io/treeverse/clients/ApiClientSpec.scala
```diff
@@ -32,17 +38,136 @@ class ApiClientSpec extends AnyFunSpec with Matchers {
...
+
+  override def beforeEach(): Unit = {
+    super.beforeEach()
+    retryWrapper = spy(new RequestRetryWrapper)
+    assert(retryWrapper.numRetries == DefaultMaxNumRetries)
+    dummyMethodInvoker = spy(new DummyMethodInvoker)
+  }
+
+  class DummyMethodInvoker {
+    @throws[Exception]
```
arielshaqed, line 81: I don't understand why we need this?  AFAICT `someMethod` does _not_ ever throw `Exception`.

clients/spark/core/src/test/scala/io/treeverse/clients/ApiClientSpec.scala
```diff
@@ -32,17 +38,136 @@ class ApiClientSpec extends AnyFunSpec with Matchers {
...
+        .doThrow(classOf[Exception])
+        .doThrow(classOf[Exception])
+        .doThrow(classOf[Exception])
+        .doThrow(classOf[ApiException])
+        .when(dummyMethodInvoker)
+        .someMethod()
+
+      assertThrows[FailsafeException] {
+        retryWrapper.wrapWithRetry(dummyMethodInvoker.createCheckedSupplierForSomeMethod())
+      }
```
arielshaqed, line 111: IIUC this says that this retry loses all information about which checked exception was actually thrown!  It threw `ApiException`, and we received `FailsafeException`.  But if that happens, how do we debug a failure that gave us `ApiException`s?

clients/spark/core/src/test/scala/io/treeverse/clients/ApiClientSpec.scala
```diff
@@ -32,17 +38,136 @@ class ApiClientSpec extends AnyFunSpec with Matchers {
...
+        .doThrow(classOf[NullPointerException])
+        .doCallRealMethod()
+        .when(dummyMethodInvoker)
+        .someMethod()
+
+      // test
+      retryWrapper.wrapWithRetry(dummyMethodInvoker.createCheckedSupplierForSomeMethod())
+
+      // assert
+      verify(dummyMethodInvoker, times(5)).someMethod()
```
arielshaqed, line 150: DefaultMaxNumRetries?

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -198,28 +219,48 @@ class ApiClient private (conf: APIConfigurations) {
...
 }
+
+class RequestRetryWrapper(val numRetries: Int = 5) {
+    private val retryPolicy : Policy[Any] = RetryPolicy.builder()
+      .withDelay(Duration.ofMillis(500))
+      .withMaxRetries(numRetries)
+      .build()
+
+  def wrapWithRetry[T](fn: CheckedSupplier[T]): T = {
+      Failsafe.`with`(Seq(retryPolicy).asJava).get(fn)
```
johnnyaug, line 263: This creates a new `FailsafeExecutor` per request. Not sure it's needed. You can create a `FailsafeExecutor[Any]` once (for example, in the ApiClient's companion object).

Then you just call it like:
```
    val repo = executor.get(() => repositoriesApi.getRepository(key.repoName))
```

clients/spark/core/src/test/scala/io/treeverse/clients/ApiClientSpec.scala
```diff
@@ -32,17 +38,136 @@ class ApiClientSpec extends AnyFunSpec with Matchers {
...
+      // test
+      retryWrapper.wrapWithRetry(dummyMethodInvoker.createCheckedSupplierForSomeMethod())
+
+      // assert
+      verify(dummyMethodInvoker, times(3)).someMethod()
+    }
+
+    it("should use custom max number of retries") {
+      retryWrapper = new RequestRetryWrapper(3)
+      assert(retryWrapper.numRetries == 3)
```
arielshaqed, line 170: Nit: This does not test the number of retries, it tests the constructor copies its parameter into some field.

arielshaqed (changes requested):
Thanks -- this looks like great code that uses a neat library!

Requested changes: I am not sure about the retry policy used; see the too-long comment on its definition.

clients/spark/core/src/test/scala/io/treeverse/clients/ApiClientSpec.scala
```diff
@@ -32,17 +38,136 @@ class ApiClientSpec extends AnyFunSpec with Matchers {
...
+        )
+        translate("https://account.example.net/container") should be(
+          "abfs://container@account.dfs.core.windows.net/"
+        )
       }
     }
   }
 }
+
+class RequestRetryWrapperSpec extends AnyFunSpec with Matchers with BeforeAndAfterEach {
```
johnnyaug, line 65: While I like how this test is implemented, it basically checks the functionality of an external library.
I think the test should be of the ApiClient class.

johnnyaug (changes requested):
Nice! Very much needed, and it looks like a good choice of library.
I have some concerns about memory usage, and about the tests.

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
talSofer: Unfortunately, this won't compile with the Spark 2.4.7 + scala  2.11.12 combo

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
talSofer: same here

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
Jonathan-Rosenberg: In IntelliJ: control + option + O

clients/spark/core/src/test/scala/io/treeverse/clients/ApiClientSpec.scala
talSofer: I have not yet addressed this one, but I will add ApiClient tests too. I think that it is valuable to keep these tests too because it demonstrates how we expect the library to work. WDYT?

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -198,28 +221,70 @@ class ApiClient private (conf: APIConfigurations) {
...
+class RequestRetryWrapper(
+    val readTimeout: Int,
+    val maxDurationSeconds: Double = -1,
+    val maxNumRetries: Int = 5
+) {
+  val UnsetMaxDuration = -1
+
+  var maxDuration = maxDurationSeconds
+  if (maxDuration == UnsetMaxDuration) {
+    maxDuration = maxNumRetries * readTimeout + 7.1 * 1
```
talSofer, line 267: @arielshaqed please let me know if this is what you meant by basic backoff or should replace 1 with something else. Thank you!

clients/spark/core/src/test/scala/io/treeverse/clients/ApiClientSpec.scala
johnnyaug: Frankly, I don't see the value in it. The code should (and does) explain itself.

johnnyaug (approved):
Thanks!

clients/spark/core/src/test/scala/io/treeverse/clients/ApiClientSpec.scala
talSofer: I removed the unit tests. Invested this morning in trying to think of unit tests that could be useful, but did not have ideas. @johnnyaug will be happy to hear ideas if you have some!

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
```diff
@@ -198,28 +221,70 @@ class ApiClient private (conf: APIConfigurations) {
...
+  }
+
+  private val retryPolicy: Policy[Any] = RetryPolicy
+    .builder()
+    .withBackoff(1, 20, ChronoUnit.SECONDS)
+    .withJitter(.25)
+    .withMaxRetries(maxNumRetries)
+    .withMaxDuration(Duration.ofSeconds(maxDuration.asInstanceOf[Number].longValue))
```
arielshaqed, line 275: Why not
```suggestion
    .withMaxDuration(Duration.ofSeconds(maxDuration.asInstanceOf[Long]))
```
?

But I'd probably use milliseconds anyway, otherwise 1.9s is treated as 1s and 0.5s is treated as 0s.

arielshaqed (approved):
Thanks!  This definitely fixes the bug!

Please consider changing the timeouts handling to use millisecond precision and not whole-second precision.

clients/spark/core/src/main/scala/io/treeverse/clients/ApiClient.scala
talSofer: done thanks

PR #4191, opened 2022-09-17
lakefs kv cmd

### LINKED ISSUE #4121: lakeFS tool to query kvstore

itai-david:
closes #4121 

`lakefs kv get` and `lakefs kv scan` implemented.
There is a generic value -> proto -> json converter, that allows each module (auth, graveler, etc.) to associate different types with different path format, and so each value is formatted correctly according to the path is is brought from. This will also support future addtion of types without the need to change this tool
It still falls short with staged objects, as their value is also a proto message which needs another translation

Only tested manually at this point


pkg/kv/kv_object_format.go
```diff
@@ -0,0 +1,54 @@
+package kv
+
+import (
+	"google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+var matchers = make(map[string]protoreflect.ProtoMessage)
```
nopcoder, line 9: Can we use a simple list of prefixes (just for a start) and not regex?

pkg/kv/kv_object_format.go
```diff
@@ -0,0 +1,54 @@
...
+var defaultMsg protoreflect.ProtoMessage = nil
```
nopcoder, line 10: I don't think we need a default msg - if we can't find a prefix to decode the proto we can decode the message dynamically. Print out for each field the code/type/value.

pkg/kv/kv_object_format.go
```diff
@@ -0,0 +1,54 @@
...
+
+func ToPrettyString(path string, protoStr []byte) (string, error) {
+	msg, err := resolvePath(path)
+	if err != nil {
+		return "", err
+	}
+
+	if msg == nil {
+		return string(protoStr), nil
+	}
```
nopcoder, line 47: Maybe we should have encoded version and metadata as part of the value to make this one easy. But for now I think with a little code we can just dump the field codes, types and values we find.

cmd/lakefs/cmd/kv.go
```diff
@@ -0,0 +1,103 @@
...
+		logger.WithField("version", version.Version).Info("lakeFS kv")
+
+		kvParams := cfg.GetKVParams()
+		kvStore, err := kv.Open(ctx, kvParams)
+		if err != nil {
+			logger.WithError(err).Fatal("Failed to open KV store")
+		}
+		defer kvStore.Close()
+
+		val, err := kvStore.Get(ctx, []byte(args[0]), []byte(args[1]))
```
nopcoder, line 43: name args[N]

```suggestion
		partitionKey := args[0]
		key := args[1]
		val, err := kvStore.Get(ctx, []byte(partitionKey), []byte(key))
```

cmd/lakefs/cmd/kv.go
```diff
@@ -0,0 +1,103 @@
...
+	Use:    "get <partition key> <key>",
+	Short:  "Return the value for the given key under the given partition",
+	Hidden: true,
+	Args:   cobra.ExactArgs(GetCmdNumArgs),
+	Run: func(cmd *cobra.Command, args []string) {
+		logger := logging.Default()
+		cfg := loadConfig()
+
+		ctx := cmd.Context()
+		logger.WithField("version", version.Version).Info("lakeFS kv")
```
nopcoder, line 34: Can be removed - we have a command that print it out if needed.

cmd/lakefs/cmd/kv.go
```diff
@@ -0,0 +1,103 @@
...
+		val, err := kvStore.Get(ctx, []byte(args[0]), []byte(args[1]))
+		if err != nil {
+			logger.WithError(err).Fatal("Failed to get value")
+		}
+		prettyVal, err := kv.ToPrettyString(args[1], val.Value)
+		if err != nil {
+			logger.WithError(err).Fatal("Failed to build object from KV value")
+		}
+
+		fmt.Printf("%s:\n%s\n", args[1], prettyVal)
```
nopcoder, line 52: I think the an interactive command should interact with stdout and stderr.
Having an error in the log for get operation that gets key not found is a valid flow which I expect to see in the console and not look into the logs.

cmd/lakefs/cmd/kv.go
```diff
@@ -0,0 +1,103 @@
...
+		if err != nil {
+			logger.WithError(err).Fatal("Failed to open KV store")
+		}
+		defer kvStore.Close()
+
+		val, err := kvStore.Get(ctx, []byte(args[0]), []byte(args[1]))
+		if err != nil {
+			logger.WithError(err).Fatal("Failed to get value")
+		}
+		prettyVal, err := kv.ToPrettyString(args[1], val.Value)
```
nopcoder, line 47: If we named the format and fetch it here - without calling a function that will format it for us, it will enable us to dump and later use this information.

cmd/lakefs/cmd/kv.go
nopcoder, line 52: Having the output in an easy to read format - or at leave option to get it in json - will enable new command to process data we fetched

cmd/lakefs/cmd/kv.go
```diff
@@ -0,0 +1,103 @@
...
+			logger.WithError(err).Fatal("Failed to build object from KV value")
+		}
+
+		fmt.Printf("%s:\n%s\n", args[1], prettyVal)
+	},
+}
+
+var kvScanCmd = &cobra.Command{
+	Use:    "scan <partition key> [<key>]",
+	Short:  "Scan through keys and values under the given partition. An optional key cna be specified as a starting point (inclusive)",
```
nopcoder, line 58: cna -> can

cmd/lakefs/cmd/kv.go
```diff
@@ -0,0 +1,103 @@
...
+		if err != nil {
+			logger.WithError(err).Fatal("Failed to build object from KV value")
+		}
+
+		fmt.Printf("%s:\n%s\n", args[1], prettyVal)
+	},
+}
+
+var kvScanCmd = &cobra.Command{
+	Use:    "scan <partition key> [<key>]",
```
nopcoder, line 57: 1. key is prefix - we should enable one
2. we should enable 'limit' to a number of entries to return
3. we should enable 'until' to match a prefix that if matched we should stop - the prefix is also the way we identify the type of the value. alternative will be to have a command that will print out all the message types and pass it as an option to scan so as long as we process a message from that kind we will continue the scan.

cmd/lakefs/cmd/kv.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+		defer kvStore.Close()
+
+		val, err := kvStore.Get(ctx, []byte(args[0]), []byte(args[1]))
+		if err != nil {
+			logger.WithError(err).Fatal("Failed to get value")
+		}
+		prettyVal, err := kv.ToPrettyString(args[1], val.Value)
+		if err != nil {
+			logger.WithError(err).Fatal("Failed to build object from KV value")
```
nopcoder, line 49: Prefer to print out the error - with special case for not found - and return.
Fatal is not a graceful exit.

cmd/lakefs/cmd/kv.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+		defer kvStore.Close()
+
+		var start []byte = nil
+		if len(args) > ScanCmdMinArgs {
+			start = []byte(args[1])
+		}
+		iter, err := kvStore.Scan(ctx, []byte(args[0]), start)
+		if err != nil {
+			logger.WithError(err).Fatal("Scan failed")
```
nopcoder, line 81: Relevant for couple of cases in this pr - we should print out the error and exit gracefully by return as we like to call out defer functions.

cmd/lakefs/cmd/kv.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+		defer iter.Close()
+
+		for iter.Next() {
```
nopcoder, line 85: missing error check after Next() returns false

nopcoder (changes requested):

pkg/auth/metadata.go
```diff
@@ -16,6 +16,12 @@ const (
 	SetupTimestampKeyName = "setup_timestamp"
 )
 
+//nolint:gochecknoinits
+func init() {
+	kv.RegisterType(InstallationIDKeyName, nil)
+	kv.RegisterType(SetupTimestampKeyName, nil)
```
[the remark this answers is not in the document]
itai-david: I hope you meant lakefs,right?
Eventually, if I want the `kv` tool to be able to parse the value from KV correctly, it has to be familiar with the type - am I wrong?

pkg/kv/store.go
```diff
@@ -37,6 +40,11 @@ func FormatPath(p ...string) string {
 	return strings.Join(p, string(PathDelimiter))
 }
 
+func MatchPath(path string, pathRegexp string) bool {
+	match, err := regexp.MatchString(PathBeginRegexp+pathRegexp, path)
+	return err == nil && match
+}
```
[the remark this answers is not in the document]
itai-david: 1. No good reason. Moved to `kv_object_format`, where it is actually used, and changed to private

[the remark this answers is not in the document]
itai-david: 2. Avoid building regex - as part of the matcer func, or all together?

pkg/kv/kv_object_format.go
itai-david: That would prevent us from identifying everything that is more complex than pref+ID, such as secondary indexes (user-groups, group-policies, user-credentials)
I think it will also require specific handling for const paths (installaion ID, installation timestamp)
We need to decide the risk of using these regexes, over the benefits they bring. I'm specifically aiming at future types we will need to introduce. WDYT?

pkg/kv/kv_object_format.go
```diff
@@ -0,0 +1,54 @@
...
+
+// The pb message type to parse a value, in case the path does not meet
+// any regex of the above. This is done mainly to support paths which
+// have no rules and may cause a conflict with another regex
+// Note that multiple calls ti RegisterDefaultType will override each other
+func RegisterDefaultType(pb protoreflect.ProtoMessage) {
+	defaultMsg = pb
+}
+
+func resolvePath(path string) (protoreflect.ProtoMessage, error) {
```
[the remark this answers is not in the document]
itai-david: Renamed to `resolveKVPathToMsgType`

cmd/lakefs/cmd/kv.go
itai-david: Right... Done

cmd/lakefs/cmd/kv.go
itai-david: Removed

pkg/kv/kv_object_format.go
itai-david: 1. I'm not sure I follow - don't we need to have at least some info on the msg type? Otherwise, how can the different fields be differentiate. Can you refer me to some example on how it can be done?
2. Why not use a default type? There is currently only one path we cannot identify correctly

pkg/kv/kv_object_format.go
itai-david: Not sure I follow - here, I identify a field that is not really a pb msg, and I dump it as a string. Is that differernt from what you suggest?

cmd/lakefs/cmd/kv.go
itai-david: Fixed. Thanks

pkg/kv/store.go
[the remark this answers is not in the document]
itai-david: Done

pkg/kv/kv_object_format.go
itai-david: Done - removed all regexes as the type can be identified by the prefix

cmd/lakefs/cmd/kv.go
itai-david: Fixed

cmd/lakefs/cmd/kv.go
itai-david: Fixed

cmd/lakefs/cmd/kv.go
itai-david: Fixed

cmd/lakefs/cmd/kv.go
itai-david: Done

cmd/lakefs/cmd/kv.go
itai-david: `ToPrettyString` change to `ToKvObject` returning an object to support json. Output is later printed as json

cmd/lakefs/cmd/kv.go
itai-david: Changed to json

pkg/kv/store.go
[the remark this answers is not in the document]
itai-david: Moved back to using regex as actions package cannot be supported otherwise.
`Task` and `Run` kv entries uses the same prefix, which contains the repoID

cmd/lakefs/cmd/kv.go
itai-david: Done - added `limit` and `until` flags

pkg/kv/kv_object_format.go
itai-david: Update - regexes are back due to current inability to define actions prefix

pkg/auth/metadata.go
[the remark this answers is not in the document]
nopcoder: As I see it - we should be able to describe all message types using the right partition/key patterns. I understand that the current state prevent this from happening, and I think it can cause conflcits in case we render, for example, staging token and branch information which both use custom paritition keys with path that can be any object that cloud be match a branch name.

Without adding any optimization I've used the following to register all the types:

```go
var recordsMatches []MatchRecord

// RegisterType - Register a pb message type to parse the data, according to a path regex
// All objects which match the path regex will be parsed as that type
// A nil type parses the value as a plain string
func RegisterType(partitionPattern, pathPattern string, mt protoreflect.MessageType) {
	recordsMatches = append(recordsMatches, MatchRecord{
		PartitionPattern: partitionPattern,
		PathPattern:      pathPattern,
		MessageType:      mt,
	})
	sort.Slice(recordsMatches, func(i, j int) bool {
		p1 := strings.TrimPrefix(recordsMatches[i].PartitionPattern, "*")
		p2 := strings.TrimPrefix(recordsMatches[j].PartitionPattern, "*")
		if p1 == p2 {
			pp1 := strings.TrimPrefix(recordsMatches[i].PathPattern, "*")
			pp2 := strings.TrimPrefix(recordsMatches[j].PathPattern, "*")
			return pp1 > pp2
		}
		return p1 > p2
	})
}
```

Used the following to create a record (previously kv object):

```go
func NewRecord(partition, path string, rawValue []byte) (*Record, error) {
	var mt protoreflect.MessageType
	for _, r := range recordsMatches {
		partitionMatch, err := filepath.Match(r.PartitionPattern, partition)
		if err != nil {
			return nil, err
		}
		if !partitionMatch {
			continue
		}

		pathMatch, err := filepath.Match(r.PathPattern, path)
		if err != nil {
			return nil, err
		}
		if pathMatch {
			mt = r.MessageType
			break
		}
	}

	var value interface{}
	if mt == nil {
		value = rawValue
	} else {
		msg := mt.New().Interface()
		err := proto.Unmarshal(rawValue, msg)
		if err != nil {
			return nil, err
		}
		value = msg
	}

	return &Record{
		Partition: partition,
		Key:       path,
		Value:     value,
	}, nil
}
```

Having each message type register using pattern matching:

```go
	kv.RegisterType(PartitionKey, "repos/*/tasks/*", (&TaskResultData{}).ProtoReflect().Type())
	kv.RegisterType(PartitionKey, "repos/*/runs/*", (&RunResultData{}).ProtoReflect().Type())
	kv.RegisterType(PartitionKey, "repos/*/branches/*/*", (&kv.SecondaryIndex{}).ProtoReflect().Type())
	kv.RegisterType(PartitionKey, "repos/*/commits/*/*", (&kv.SecondaryIndex{}).ProtoReflect().Type())
	//...
	kv.RegisterType(PartitionKey, "users/*", (&UserData{}).ProtoReflect().Type())
	kv.RegisterType(PartitionKey, "policies/*", (&PolicyData{}).ProtoReflect().Type())
	kv.RegisterType(PartitionKey, "groups/*", (&GroupData{}).ProtoReflect().Type())
	kv.RegisterType(PartitionKey, "uCredentials/*/credentials/*", (&CredentialData{}).ProtoReflect().Type())
	kv.RegisterType(PartitionKey, "gUsers/*/users/*", (&kv.SecondaryIndex{}).ProtoReflect().Type())
	kv.RegisterType(PartitionKey, "gPolicies/*/policies/*", (&kv.SecondaryIndex{}).ProtoReflect().Type())
	kv.RegisterType(PartitionKey, "uPolicies/*/policies/*", (&kv.SecondaryIndex{}).ProtoReflect().Type())
	kv.RegisterType(PartitionKey, "expiredTokens/*", (&TokenData{}).ProtoReflect().Type())
	kv.RegisterType(PartitionKey, "installation_metadata/*", nil)
	//...
	kv.RegisterType(gravelerPartition, "repos/*", (&RepositoryData{}).ProtoReflect().Type())
	kv.RegisterType("*", "branches/*", (&BranchData{}).ProtoReflect().Type())
	kv.RegisterType("*", "commits/*", (&CommitData{}).ProtoReflect().Type())
	kv.RegisterType("*", "tags/*", (&TagData{}).ProtoReflect().Type())
	kv.RegisterType("*", "*", (&StagedEntryData{}).ProtoReflect().Type())
```

The above support lookup based on partition key which make the result expected - not based on the un-ordered map.
Used filepath match to enabled fast matching - not match regex pattern at runtime which is very slow.
Used strings - but we can format the path using `kv.FormatPath`.
Don't use 'default message' as we have fallback by using '*' '*.

nopcoder (changes requested):

cmd/lakefs/cmd/kv.go
itai-david: Renamed to `path`

cmd/lakefs/cmd/kv.go
itai-david: Right. Removed check from within loop

cmd/lakefs/cmd/kv.go
```diff
@@ -0,0 +1,192 @@
...
+	Short:  "Return the value for the given key under the given partition",
+	Hidden: true,
+	Args:   cobra.ExactArgs(GetCmdNumArgs),
+	Run: func(cmd *cobra.Command, args []string) {
+		cfg := loadConfig()
+
+		pretty, err := cmd.Flags().GetBool("pretty")
+		if err != nil {
+			os.Exit(1)
+		}
```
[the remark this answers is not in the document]
itai-david: It is handled by `cobra` pkg, and currently prints the following:
`invalid argument "abc123" for "--pretty" flag: strconv.ParseBool: parsing "abc123": invalid syntax`
Do you think we need to further handle this?

cmd/lakefs/cmd/kv.go
```diff
@@ -0,0 +1,192 @@
...
+var kvCmd = &cobra.Command{
+	Use:    "kv",
+	Short:  "Inspect lakeFS' Key-Value Store",
+	Hidden: true,
+}
+
+var kvGetCmd = &cobra.Command{
+	Use:    "get <partition key> <key>",
+	Short:  "Return the value for the given key under the given partition",
+	Hidden: true,
```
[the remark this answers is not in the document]
itai-david: 1. I believe it is meant to be a plumbing command and in that case should be hidden
2. Removed `hidden` from internal kv commands

cmd/lakefs/cmd/kv.go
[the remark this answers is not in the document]
itai-david: Changed to RunE and return the error

cmd/lakefs/cmd/kv.go
```diff
@@ -0,0 +1,192 @@
...
+		pretty, err := cmd.Flags().GetBool("pretty")
+		if err != nil {
+			os.Exit(1)
+		}
+
+		ctx := cmd.Context()
+		kvParams, err := cfg.GetKVParams()
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "KV params: %s\n", err)
+			os.Exit(1)
```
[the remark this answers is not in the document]
itai-david: Changed to RunE and return the error

cmd/lakefs/cmd/kv.go
```diff
@@ -0,0 +1,192 @@
...
+		defer kvStore.Close()
+
+		iter, err := kvStore.Scan(ctx, []byte(partitionKey), start)
+		if err != nil {
+			logMsg := "Failed to scan partition '" + partitionKey + "'"
+			if start != nil {
+				logMsg += " with start key '" + string(start) + "'"
+			}
+			logMsg += ": " + err.Error() + "\n"
+			fmt.Fprint(os.Stderr, logMsg)
```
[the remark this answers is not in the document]
itai-david: Removed. A general failed + error is returned to rootCmd

pkg/auth/metadata.go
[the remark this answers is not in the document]
itai-david: Thanks for this suggestion. Decided to take it as is. Regex usage is removed all together

pkg/kv/kv_object_format.go
```diff
@@ -0,0 +1,71 @@
+package kv
```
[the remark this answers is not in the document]
itai-david: Done

pkg/kv/kv_object_format.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"regexp"
+
+	"google.golang.org/protobuf/proto"
+	"google.golang.org/protobuf/reflect/protoreflect"
+)
+
+var matchers = make(map[string]protoreflect.ProtoMessage)
```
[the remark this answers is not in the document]
itai-david: Done as part of adopting your code suggestion

pkg/kv/kv_object_format.go
```diff
@@ -0,0 +1,71 @@
...
+var defaultMsg protoreflect.ProtoMessage = nil
+
+type KvObject struct {
+	Partition string
+	Key       string
+	Value     interface{}
+}
```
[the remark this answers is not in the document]
itai-david: Done - added

cmd/lakefs/cmd/kv.go
```diff
@@ -0,0 +1,192 @@
...
+		if err != nil {
+			os.Exit(1)
+		}
+		pretty, err := cmd.Flags().GetBool("pretty")
+		if err != nil {
+			os.Exit(1)
+		}
+
+		partitionKey := args[0]
+		var start []byte = nil
```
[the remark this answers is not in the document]
itai-david: Done - removed

cmd/lakefs/cmd/kv.go
```diff
@@ -0,0 +1,192 @@
...
+			fmt.Fprintf(os.Stderr, "Scan operation ended with error: %s", iter.Err())
+			os.Exit(1)
+		}
+
+		var kvObjsJSON []byte
+		if pretty {
+			kvObjsJSON, err = json.MarshalIndent(kvObjs, "", "  ")
+		} else {
+			kvObjsJSON, err = json.Marshal(kvObjs)
+		}
```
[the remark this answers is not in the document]
itai-david: Done

pkg/kv/kv_object_format.go
```diff
@@ -0,0 +1,71 @@
...
+
+// Register a pb message type to parse the data, according to a path regex
+// All objects which match the path regex will be parsed as that type
+// A nil type parses the value as a plain string
+func RegisterType(pathRegexp string, pb protoreflect.ProtoMessage) {
+	matchers[pathRegexp] = pb
```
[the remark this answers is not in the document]
itai-david: Done. Panic added to relevant section of new `RegsiterType`

pkg/kv/kv_object_format.go
nopcoder: My problem is that the default type solve one problem which is indication for the next one...
We set the default to staging object proto - which means we don't have a pattern that match this type.
It looks like based on the pattern of partition and/or path we should be able to identify the type, so the fallback will be the raw value and not proto message.
There is a solution to dynamically dump any protobuf message, but as I see it now it is not required.

pkg/kv/kv_object_format.go
nopcoder: You can ignore. I thought there `msg == nil` is unknown proto we need to decode - but it means not a proto value.

cmd/lakefs/cmd/kv.go
```diff
@@ -0,0 +1,167 @@
...
+		for iter.Next() {
+			entry := iter.Entry()
+			if len(until) > 0 && string(entry.Key) > until {
+				break
+			}
+			kvObj, err := kv.NewRecord(partitionKey, string(entry.Key), entry.Value)
+			if err != nil {
+				return fmt.Errorf("KV record from value: %w", err)
+			}
+			encoder := json.NewEncoder(os.Stdout)
```
nopcoder, line 138: We should have a new encoder once outside the loop

nopcoder (changes requested):
Added one comment about the json encoder.
There is one issue related to the implementation I suggested - the match path stops at '/' - need to verify that all keys that we are mapping does not include '/' as part of the value or if we are using path as prefix - we will need to address it.
It means maybe switching back to regex to match possible values with '/'. :/ sorry.

cmd/lakefs/cmd/kv.go
itai-david: Correct - done

itai-david:
> Added one comment about the json encoder. There is one issue related to the implementation I suggested - the match path stops at '/' - need to verify that all keys that we are mapping does not include '/' as part of the value or if we are using path as prefix - we will need to address it. It means maybe switching back to regex to match possible values with '/'. :/ sorry.

@nopcoder - I believe the only place we allow a path separator as part of the key (other than the spearators used to build the path literals) is with `StagedObjects` keys. I don't think that the regular expression solves it, but it was solved using the "defultType".
As both the repository and the key for `StagedObject` can be anything (with '/' allowed as part of the key), it pretty much means that everything that cannot be classified as any other type, is a potential `StageObject`. I think we should add a "defaultType" that will map to `StagedObject` anyhting that is not mapped to anything else.
WDYT?

nopcoder (changes requested):
Suggestion on how to handle the partition/path matching 
 https://github.com/treeverse/lakeFS/pull/4496/files

nopcoder (approved):

PR #4192, opened 2022-09-18
Improve Graveler Delete by checking staging first and return when possible

### LINKED ISSUE #4193: Improve Graveler delete prevent access to committed when possible
Improve Graveler Delete operation by returning as soon possible.
Checking staging first and returning if needed will perform faster than checking both committed and uncommitted all the time.

nopcoder:
Improve Graveler Delete operation by returning as soon as possible.

Close #4193

### DIFF
--- a/pkg/graveler/graveler.go
+++ b/pkg/graveler/graveler.go
@@ -1451,47 +1451,42 @@ func (g *KVGraveler) Delete(ctx context.Context, repository *RepositoryRecord, b
 		return ErrWriteToProtectedBranch
 	}
 
-	deleteEntry := func(branch *Branch) error {
-		commit, err := g.RefManager.GetCommit(ctx, repository, branch.CommitID)
-		if err != nil {
-			return err
-		}
-
-		// check key in committed - do we need tombstone?
-		foundInCommitted := false
-		_, err = g.CommittedManager.Get(ctx, repository.StorageNamespace, commit.MetaRangeID, key)
-		if err != nil {
-			if !errors.Is(err, ErrNotFound) {
-				// unknown error
-				return fmt.Errorf("reading from committed: %w", err)
-			}
-		} else {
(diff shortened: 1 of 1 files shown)


pkg/graveler/graveler.go
```diff
@@ -1451,47 +1451,43 @@ func (g *KVGraveler) Delete(ctx context.Context, repository *RepositoryRecord, b
...
-			foundInCommitted = true
-		}
 
-		foundInStaging := true
-		val, err := g.getFromStagingArea(ctx, branch, key)
-		if err != nil {
-			if !errors.Is(err, ErrNotFound) {
-				// unknown error
-				return fmt.Errorf("reading from staging: %w", err)
+			// check staging for entry or tombstone
```
itaiad200: Can be performed after the staging read.

pkg/graveler/graveler.go
```diff
@@ -1451,47 +1451,42 @@ func (g *KVGraveler) Delete(ctx context.Context, repository *RepositoryRecord, b
...
+			case err != nil:
+				if !errors.Is(err, ErrNotFound) {
+					// unknown error
+					return fmt.Errorf("reading from staging: %w", err)
+				}
+				// entry not found in staging - continue to committed check
+			case val == nil:
+				// found tombstone in staging, return ErrNotFound
+				return ErrNotFound
+			default:
```
itaiad200, line 1468: Consider doing it explicit for readability:
```suggestion
			case val != nil:
```

itaiad200 (approved):

pkg/graveler/graveler.go
nopcoder: Done - thanks

pkg/graveler/graveler.go
nopcoder: prefer to catch all at this point

PR #4194, opened 2022-09-18
[gc] Speed up by improving SSTable parser and improving partitioning

arielshaqed:
1. Replace SparkSQL antijoin computations with a more Hadoopish Map/Reduce style that computes first differences of range IDs and then of object addresses.  This enables explicit control over partitioning.  Use hash-based partitioning (of both range IDs and object addresses!) to significantly reduce shuffling.
1. Speed up SSTable parser by replacing the use of ByteBuffers with plain byte arrays.  (This is _not_ a major speedup unfortunately, however by giving *Java* GC more visibility it _might_ improve memory usage.

This is _much_ faster (no precise timing yet, but a large repo got GCed in <2h with this code and never finished with the existing code, so this might be 10x?).


arielshaqed:
# Hints for reviewers

Apologies for this long PR.  It's actually >= 2 separate PRs, that cannot be easily be separated.  This is my fault.

But you _can_ review it in smaller parts.

* GarbageCollector.scala.  Here we replaced the Spark-SQL distinct + antijoin operations with explicit Map/Reduce operations.  This allows precise control over partitioning, so we select hash-based partitioning and explicitly request the number of partitions.
* GarbageCollectorSpec.scala.  These are Spark component tests for GarbageCollector.  This verifies some behaviour of the Spark code by running it on an actual Spark (local) cluster.  **PLEASE consider whether I should add any tests here!**
* SSTableReader, BlockParser, IndexedBytes, and their tests.  Move from parsing on ByteBuffers (directly memory-mapped from the file and/or Snappy decompressor) to Array[Byte] (just plain Java byte arrays).  While ByteBuffers should be faster, in practice they aren't -- probably because we don't do enough with them.
* DumpSSTable.  A benchmark for the SSTable reader.
* run-gc-test.  Since GarbageCollector lets us select number of partitions, drop the number of partitions used.  This does not appear to speed up the test significantly 😿 .

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -22,293 +20,191 @@ import java.time.ZoneOffset
...
-      apiConf: APIConfigurations,
+  def minus(
+      left: Dataset[String],
+      right: Dataset[String],
+      partitioner: Partitioner
+  ): Dataset[String] = {
+    import spark.implicits._
+    def mark(f: Int) = (p: Any) => p match { case (t: String) => (t, f) }
+    val both = left.map(mark(1)).union(right.map(mark(2)))
+    val reduced = both.rdd.reduceByKey(partitioner, _ | _)
```
talSofer, line 131: for my understanding, what is the reducer function? I can't understand by reading it. 
I expect it to generate 1 for items that are only part of left, but what would it generate for the rest?

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSContext.scala
```diff
@@ -20,11 +20,16 @@ object LakeFSContext {
   val LAKEFS_CONF_JOB_REPO_NAME_KEY = "lakefs.job.repo_name"
   val LAKEFS_CONF_JOB_STORAGE_NAMESPACE_KEY = "lakefs.job.storage_namespace"
   val LAKEFS_CONF_JOB_COMMIT_ID_KEY = "lakefs.job.commit_id"
+  val LAKEFS_CONF_GC_NUM_RANGE_PARTITIONS = "lakefs.gc.range.num_partitions"
+  val LAKEFS_CONF_GC_NUM_ADDRESS_PARTITIONS = "lakefs.gc.address.num_partitions"
```
talSofer, line 24: I suggest that we document it as a configuration option but we can do it under a different pr.

talSofer (commented):
Thanks Ariel! 
I reviewed the GC part which looks great. and added some questions. posting this without approving because I have not reviewed the reader piece yet. 

A general question, do we know why things weren't partitioned properly in our previous code? or it is a mix of many things?

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSContext.scala
arielshaqed: You're right, of course!  I'll start that PR (don't want to delay this one).

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
arielshaqed: `_ | _` is Scala for "the bitwise OR function".  So it returns 1 for keys which have only been seen on the left, 2 for keys which have only been seen on the right, and 3 (== 1|2) for keys which have been seen on both sides.  On the next action we filter to keep only 1's -- keys which have only been seen on the left.  What can I say, this is a great time for [bit twiddling](http://catb.org/jargon/html/B/bit-twiddling.html).

Adding a name for `_ | _`, which I hope will make things understandable.

arielshaqed (commented):
Thanks for the review!

The previous version didn't offer very wide parallelism anywhere: a single CSV file is a single partition, therefore mapping its commit ID to metarange IDs still yields a single partition, and splitting each into metarange into ranges is _still_ a single partition. Adding various "repartition" calls does increase parallelism, but does nothing for performance if you need to shuffle.

So seeing small number of partitions and a lot of shuffling when repartitioning without consideration to _what_ we're doing gives bad stats (low parallelism, a lot of shuffling if repartitioning).

The classic SparkSQL method of fixing this is by adding hints that say how to repartition each column.  By now we're at about the same code complexity of just coding a reduction.  If I want the critical part of the job to have a particular shape, I may as well write it out explicitly.

Note that GarbageCollector.scala is actually (much) _shorter_ in this version.  The PR adds code mainly because of tests.

clients/spark/core/src/test/scala/io/treeverse/clients/GarbageCollectorSpec.scala
```diff
@@ -0,0 +1,152 @@
...
+  def getRangeAddresses(rangeID: String, repo: String): Iterator[String] = {
+    verifyRepo(repo)
+    ranges(rangeID).iterator
+  }
+}
+
+class GarbageCollectorSpec extends AnyFunSpec with Matchers with SparkSessionSetup {
+  describe("Spark") {
+    it("should perform Gauss summation") {
+      withSparkSession(spark => {
```
johnnyaug, line 34: Why is this test relevant?

clients/spark/core/src/test/scala/io/treeverse/clients/GarbageCollectorSpec.scala
```diff
@@ -0,0 +1,152 @@
...
+        withSparkSession(spark => {
+          import spark.implicits._
+          val sc = spark.sparkContext
+          val gc = new GarbageCollector(getter, new ConfigMapper(spark.sparkContext.broadcast(Array.empty)))
+
+          val partitioner = new HashPartitioner(3)
+          val threes = sc.parallelize(0 to 100 by 3).map(_.toString).toDS
+          val sevens = sc.parallelize(0 to 100 by 7).map(_.toString).toDS
+          val m = gc.minus(threes, sevens, partitioner).map(_.toInt)
+          compareDS(m, sc.parallelize((0 to 100 by 3).filter(x => x % 7 != 0)).toDS)
```
johnnyaug, line 69: Cool!

johnnyaug (approved):
Amazing work! 
Love the naming changes and the tests. Leaving some minor comments.

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -42,273 +93,126 @@ trait S3ClientBuilder extends Serializable {
...
-  /** @return a serializable summary of values in hc starting with prefix.
-   */
-  def getHadoopConfigurationValues(hc: Configuration, prefixes: String*): ConfMap =
-    hc.iterator.asScala
-      .filter(c => prefixes.exists(c.getKey.startsWith))
-      .map(entry => (entry.getKey, entry.getValue))
-      .toList
-      .asInstanceOf[ConfMap]
+class GarbageCollector(val rangeGetter: RangeGetter, configMap: ConfigMapper) extends Serializable {
+  @transient lazy val spark = SparkSession.active
```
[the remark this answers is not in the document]
arielshaqed: Thanks!

It's a real annoyance: otherwise it's really hard to `import spark.implicits._`, and _nothing works_.

clients/spark/core/src/main/scala/io/treeverse/jpebble/BlockReadable.scala
```diff
@@ -22,14 +21,27 @@ trait BlockReadable {
...
-  val size = in.size() // Compute once, the file should anyway be immutable!
-
-  override def length: Long = size
-
-  override def readBlock(offset: Long, size: Long) =
-    // TODO(ariels): Cache return values - readonly mmaps are reusable
-    IndexedBytes.create(in.map(FileChannel.MapMode.READ_ONLY, offset, size))
+class BlockReadableFile(private val in: RandomAccessFile) extends BlockReadable with Closeable {
+  if (in == null) {
+    throw new NullPointerException("null file")
```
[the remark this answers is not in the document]
arielshaqed: Done!  I just removed it entirely, it's a pointless check anyway.

clients/spark/core/src/main/scala/io/treeverse/jpebble/IndexedBytes.scala
```diff
@@ -1,67 +1,92 @@
...
-  def toByteBuffer: ByteBuffer = throw new UnsupportedOperationException(
-    "toByteBuffer not implemented"
-  )
 }
 
-/** Iterator over a ByteByffer.
+/** Iterator over a BufferIndexedBytes
  */
-class ByteBufferIterator(private val buf: ByteBufferIndexedBytes) extends Iterator[Byte] {
+class BufferIterator(private val buf: BufferIndexedBytes) extends Iterator[Byte] {
```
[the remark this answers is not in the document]
arielshaqed: ByteBufferIndexedBytesSpec tests this as part of "iterator".

clients/spark/core/src/test/scala/io/treeverse/clients/GarbageCollectorSpec.scala
arielshaqed: It demonstrates how to test using Spark.  Otherwise it's really hard to understand what an easy form is.  Opened #4211 to do it soon-ish.

arielshaqed (commented):
Thanks!

Waiting for review by @Jonathan-Rosenberg before pulling.

Jonathan-Rosenberg (approved):
Went through the parser part. LGTM.

arielshaqed:
Pulling once tests pass (silly failure).

PR #4196, opened 2022-09-19
Extended metrics

### LINKED ISSUE #454: Use testing short flag to skip playback tests
Using `go test -short` will skip the gateway playback testing.
It will enable faster basic test run without download and play of playback data.

### LINKED ISSUE #4265: Extended analytics event
Extended analytics enables reporting more information on each event report.
Only when relevant the following values, hashed except 'client', will be reported:

1. Repository - hashed value of the repository name
2. Ref - hashed value of the tag/commit/branch name 
3. SourceRef - hashed value of source tag/commit/branch 
4. UserID - hashed value of the user ID
5. Client - user agent of the client performing the operation

This feature is turned **off** by default.

nopcoder:
- default is off.
- sends client user-agent as id, repository, ref, source ref and username.
- all extended values are hashed, except client id.
- lakectl, webui and sdk now include specific user-agent with build time version information.

Close #4265

### DIFF
--- a/clients/java/src/main/java/io/lakefs/clients/api/ApiClient.java
+++ b/clients/java/src/main/java/io/lakefs/clients/api/ApiClient.java
@@ -131,7 +131,7 @@ public class ApiClient {
         json = new JSON();
 
         // Set default User-Agent.
-        setUserAgent("OpenAPI-Generator/0.1.0-SNAPSHOT/java");
+        setUserAgent("lakefs-java-sdk/0.1.0-SNAPSHOT");
 
         authentications = new HashMap<String, Authentication>();
     }
--- a/clients/python/lakefs_client/api_client.py
+++ b/clients/python/lakefs_client/api_client.py
@@ -77,7 +77,7 @@ class ApiClient(object):
             self.default_headers[header_name] = header_value
         self.cookie = cookie
         # Set default User-Agent.
-        self.user_agent = 'OpenAPI-Generator/0.1.0-SNAPSHOT/python'
+        self.user_agent = 'lakefs-python-sdk/0.1.0-SNAPSHOT'
 
     def __enter__(self):
         return self
--- a/cmd/lakefs/cmd/run.go
+++ b/cmd/lakefs/cmd/run.go
@@ -167,8 +167,10 @@ var runCmd = &cobra.Command{
 			logger.WithField("adapter_type", blockstoreType).
 				Error("Block adapter NOT SUPPORTED for production use")
 		}
+
 		metadata := stats.NewMetadata(ctx, logger, blockstoreType, authMetadataManager, cloudMetadataProvider)
-		bufferedCollector := stats.NewBufferedCollector(metadata.InstallationID, cfg)
+		bufferedCollector := stats.NewBufferedCollector(metadata.InstallationID, cfg, stats.WithLogger(logger))
+
 		// init block store
 		blockStore, err := factory.BuildBlockAdapter(ctx, bufferedCollector, cfg)
 		if err != nil {
--- a/cmd/lakefs/cmd/run.go
+++ b/cmd/lakefs/cmd/run.go
@@ -313,7 +313,7 @@ var runCmd = &cobra.Command{
 		bufferedCollector.Run(ctx)
 		defer bufferedCollector.Close()
 
-		bufferedCollector.CollectEvent("global", "run")
+		bufferedCollector.CollectEvent(stats.Event{Class: "global", Name: "run"})
 
 		logging.Default().WithField("listen_address", cfg.GetListenAddress()).Info("starting HTTP server")
 		server := &http.Server{
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -102,7 +102,7 @@ func (c *Controller) GetAuthCapabilities(w http.ResponseWriter, _ *http.Request)
 
 func (c *Controller) DeleteObjects(w http.ResponseWriter, r *http.Request, body DeleteObjectsJSONRequestBody, repository string, branch string) {
 	ctx := r.Context()
-	c.LogAction(ctx, "delete_objects", r.UserAgent(), repository, branch, "")
+	c.LogAction(ctx, "delete_objects", r, repository, branch, "")
 
 	// limit check
 	if len(body.Paths) > DefaultMaxDeleteObjects {
@@ -242,7 +242,7 @@ func (c *Controller) GetPhysicalAddress(w http.ResponseWriter, r *http.Request, 
 		return
 	}
 	ctx := r.Context()
-	c.LogAction(ctx, "generate_physical_address", r.UserAgent(), repository, branch, "")
+	c.LogAction(ctx, "generate_physical_address", r, repository, branch, "")
 
 	repo, err := c.Catalog.GetRepository(ctx, repository)
 	if errors.Is(err, catalog.ErrNotFound) {
@@ -299,7 +299,7 @@ func (c *Controller) LinkPhysicalAddress(w http.ResponseWriter, r *http.Request,
 	}
 
 	ctx := r.Context()
-	c.LogAction(ctx, "stage_object", r.UserAgent(), repository, branch, "")
+	c.LogAction(ctx, "stage_object", r, repository, branch, "")
 
 	repo, err := c.Catalog.GetRepository(ctx, repository)
 	if errors.Is(err, catalog.ErrNotFound) {
@@ -380,7 +380,7 @@ func (c *Controller) ListGroups(w http.ResponseWriter, r *http.Request, params L
 	}
 
 	ctx := r.Context()
-	c.LogAction(ctx, "list_groups", r.UserAgent(), "", "", "")
+	c.LogAction(ctx, "list_groups", r, "", "", "")
 	groups, paginator, err := c.Auth.ListGroups(ctx, &model.PaginationParams{
 		After:  paginationAfter(params.After),
 		Prefix: paginationPrefix(params.Prefix),
@@ -418,7 +418,7 @@ func (c *Controller) CreateGroup(w http.ResponseWriter, r *http.Request, body Cr
 		return
 	}
 	ctx := r.Context()
-	c.LogAction(ctx, "create_group", r.UserAgent(), "", "", "")
+	c.LogAction(ctx, "create_group", r, "", "", "")
 
 	g := &model.Group{
 		CreatedAt:   time.Now().UTC(),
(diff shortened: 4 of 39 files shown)


clients/java/src/main/java/io/lakefs/clients/api/ApiClient.java
```diff
@@ -131,7 +131,7 @@ private void init() {
         json = new JSON();
 
         // Set default User-Agent.
-        setUserAgent("OpenAPI-Generator/0.1.0-SNAPSHOT/java");
+        setUserAgent("lakefs-java-sdk/dev-f1d18ae9.with.local.changes");
```
itaiad200, line 134: Should we check this in? IIUC this will now change with every build.

clients/python/lakefs_client/api_client.py
```diff
@@ -77,7 +77,7 @@ def __init__(self, configuration=None, header_name=None, header_value=None,
             self.default_headers[header_name] = header_value
         self.cookie = cookie
         # Set default User-Agent.
-        self.user_agent = 'OpenAPI-Generator/0.1.0-SNAPSHOT/python'
+        self.user_agent = 'lakefs-python-sdk/dev-f1d18ae9.with.local.changes'
```
itaiad200, line 80: Same here

webui/vite.config.js
```diff
@@ -1,11 +1,18 @@
 import reactRefresh from '@vitejs/plugin-react-refresh';
 import eslintPlugin from 'vite-plugin-eslint';
-
+import replace from '@rollup/plugin-replace';
 
 // https://vitejs.dev/config/
 export default ({ command }) => {
   const baseConfig = {
     plugins: [
+      replace({
```
itaiad200, line 9: Sorcery

itaiad200 (changes requested):
Partial review, will continue tomorrow.

clients/java/src/main/java/io/lakefs/clients/api/ApiClient.java
nopcoder: You are right I'll keep just the VERSION value - we will have 'dev' and every publish we will have the lakefs version without the git sha.

webui/vite.config.js
nopcoder: Had to find a quick way to embed this information inside our app.

pkg/api/controller.go
```diff
@@ -2196,7 +2196,7 @@ func (c *Controller) GetCommit(w http.ResponseWriter, r *http.Request, repositor
 		return
 	}
 	ctx := r.Context()
-	c.LogAction(ctx, "get_commit")
+	c.LogAction(ctx, "get_commit", r.UserAgent(), repository, commitID, "")
```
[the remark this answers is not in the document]
NoyDavidson2: AFAIC it's not at the moment. 
Might be in the future - if we collect commit id in other actions like create commit and not just at the get commit point.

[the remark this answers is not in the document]
nopcoder: @itaiad200 it is also two for other operations - diff, merge and create branch.

cmd/lakefs/cmd/run.go
```diff
@@ -92,6 +94,18 @@ func newLDAPAuthenticator(cfg *config.LDAP, service auth.Service) *auth.LDAPAuth
...
+
+func (c *ConsoleSender) SendEvent(ctx context.Context, installationID, processID string, m []stats.Metric) error {
+	fmt.Printf("SendEvent - installationID: %s, processID: %s, Metric: %s\n", installationID, processID, spew.Sdump(m))
+	return nil
+}
+
+func (c *ConsoleSender) UpdateMetadata(ctx context.Context, m stats.Metadata) error {
+	fmt.Printf("UpdateMetadata - %s\n", spew.Sdump(m))
+	return nil
+}
```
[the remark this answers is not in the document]
nopcoder: Will drop the code - it helpful in the development time of this feature.

pkg/api/controller.go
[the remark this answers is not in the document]
NoyDavidson2: @nopcoder create branch only gets 'branch' as ref according to the table you added on the issue associated #454

@itaiad200 I agree about the commit ID at this point, it's not part of the requirements, let's consider not using it in this constellation.
I share your concern about multiple values with different meaning for this field (ref/branch/commit ID).
Going back to the original requirement - we wanted to collect branch, and then came up the question - what branch do we keep for events like 'merge'? we do need both source and destination. (similar case with diff)
This is why came up the idea of additional field 'source ref', but now I'm not so sure keeping multiple fields for these unique cases seems reasonable.

I don't want to block this PR, I think we should stick to the original plan and add branch field with branch values **only** at this point, going forward if we see fit we can add the others in some form.

As for the 'merge' events I'm thinking maybe we can keep the **source branch** as 'branch' value and create another table for merges with a foreign key field, on the 'events' table side, referring to 'merge'(installation id, time, source, destination) table. On the 'merge' table we can keep any other information about the merge if there is. 
Makes sense? WDYT?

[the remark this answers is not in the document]
itaiad200: I see many BI collections that use the `ref` as is. It's more than the `CommitID` being probably redundant and will hard to deduct any meaningful insights from it. It also masks important branch information. However I understand the limitations - `Ref` interpolation is happening in Catalog/Graveler, and not in the Controller where BI is collected. I'm fine with collecting it as it's in this PR, as long as @NoyDavidson2 is onboard with collecting `Refs` which is a fancy word for either Branch/Commit/Traverse of the commit tree (`main^2~3`)

pkg/api/controller.go
```diff
@@ -2731,7 +2731,7 @@ func (c *Controller) logCommitsHelper(w http.ResponseWriter, r *http.Request, re
 		return
 	}
 	ctx := r.Context()
-	c.LogAction(ctx, "get_branch_commit_log")
+	c.LogAction(ctx, "get_branch_commit_log", r.UserAgent(), repository, ref, "")
```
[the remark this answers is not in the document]
nopcoder: We can update this one if we agree to also run 'update' on the current database after we deploy

pkg/gateway/operations/listbuckets.go
```diff
@@ -15,12 +15,13 @@ func (controller *ListBuckets) RequiredPermissions(_ *http.Request) (permissions
...
 			Action:   permissions.ListRepositoriesAction,
-			Resource: "*"},
+			Resource: "*",
+		},
 	}, nil
 }
 
 func (controller *ListBuckets) Handle(w http.ResponseWriter, req *http.Request, o *AuthorizedOperation) {
-	o.Incr("list_repos")
+	o.Incr("list_repos", o.Principal, "", "")
```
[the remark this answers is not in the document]
nopcoder: For each handler 'o' got different type - there are at least 3.

pkg/api/controller.go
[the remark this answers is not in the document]
nopcoder: 1. Got the linked issue wrong. The number is from a different repo - will probably need to replace it with non link or a different issue.
2. There is no point in linking another table with extra properties for each event, unless we create different tables for different types as the relation is 1:1 + it will require adding unique ID to the event to bind the extra information.
3. Wanted to make it clear, as @itaiad200 said - almost all places that we accept 'ref' and not 'branch' can report commit id (event short), branch, tag or relative commit which will make the different IDs to work with.
If we plan to use just branch names - at the point we report the metric we can limit the branch name just to operations that accepts branch name. It in case of 'merge' the target branch name. But in case of 'diff' we don't have any promise we compare two branches.

pkg/stats/collector.go
```diff
@@ -324,19 +371,10 @@ func (s *BufferedCollector) sendRuntimeStats() {
...
-	} else {
-		sender = NewDummySender()
-	}
-	return uuid.Must(uuid.NewUUID()).String(),
-		[]BufferedCollectorOpts{
-			WithSender(sender),
-			WithFlushInterval(c.GetStatsFlushInterval()),
-		}
+	v := xxhash.Sum64String(s)
+	return fmt.Sprintf("%02x", v)
```
itaiad200, line 379: 2 hex digits gives you 256 options which is considered safe usage for 16 unique examples. I don't think it's enough for branches, let alone commits which can accumulate quite heavily.

itaiad200 (approved):
See single comment before merging

pkg/stats/collector.go
nopcoder: This is full hex string with leading zero for 2 digit padding it doesn't cut the value

pkg/api/controller.go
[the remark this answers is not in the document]
NoyDavidson2: Update - @nopcoder and I have spoke this morning and decided to release this feature the way it is described here. After working with the data we will decide if and how we can make it better as we go.

webui/src/lib/api/index.js
```diff
@@ -144,7 +140,7 @@ class Auth {
         const response = await fetch(`${API_ENDPOINT}/auth/password`, {
             headers: new Headers(defaultAPIHeaders),
             method: 'POST',
-            body: json({token: token, newPassword: newPassword, email: email})
+            body: JSON.stringify({token: token, newPassword: newPassword, email: email})
```
itaiad200, line 143: Is `JSON.stringify(data, null, "")` equal to `JSON.stringify(data)` ?

pkg/stats/collector.go
```diff
@@ -245,6 +244,9 @@ func (s *BufferedCollector) isCtxCancelled() bool {
 }
 
 func (s *BufferedCollector) Run(ctx context.Context) {
+	if atomic.LoadInt32(&s.runCalled) != 0 {
```
itaiad200, line 247: Why would run be called more than once?

itaiad200: Also consider using `Sync.Once`.

pkg/httputil/client.go
```diff
@@ -0,0 +1,13 @@
+package httputil
+
+import "net/http"
+
+// GetRequestLakeFSClient get lakeFS client identifier from request.
+//   It extracts the data from X-Lakefs-Client header and fallback to the user-agent
+func GetRequestLakeFSClient(r *http.Request) string {
+	id := r.Header.Get("X-Lakefs-Client")
+	if id == "" {
+		id = r.UserAgent()
```
itaiad200, line 10: Won't it spam the DB? The cardinality of user-agent is huge

itaiad200 (approved):

webui/src/lib/api/index.js
nopcoder: Yea, same.

pkg/stats/collector.go
nopcoder: Will do. I also think it is bit over-eng - but will address it in a different issue.

nopcoder: sync.Once can help making sure `Run` will perform per incese. Currently the counter also used a marker for 'Close' to indicate we do not close/wait for resources we allocated on Run.

pkg/httputil/client.go
nopcoder: I consider 'event' information as raw data - we can filter out the data. Using the user-agent as the client identify will help us also process previous/older clients using the lakefs.

PR #4205, opened 2022-09-20
Delete objects as part of Graveler

### LINKED ISSUE #4203: Use bulk-delete API for recursive delete in lakeFSFS
Measurements indicate that this can speed up the `cleanupJob` operation on `FileOutputCommitter` (which uses recursive delete) by 7x.

nopcoder:
Optimize multiple object delete operation as part of gateway and open-api, by adding DeleteBatch operation that will perform the multiple delete as part of Graveler.

Notes:
- That this changes removed the 204 status code from delete objects api, in case of successful or any error it will return 200 and the objects error list will include the specific error

Fix #4203

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -106,15 +106,16 @@ func (c *Controller) DeleteObjects(w http.ResponseWriter, r *http.Request, body 
 
 	// limit check
 	if len(body.Paths) > DefaultMaxDeleteObjects {
-		writeError(w, http.StatusInternalServerError, fmt.Errorf("%w, max paths is set to %d",
-			ErrRequestSizeExceeded, DefaultMaxDeleteObjects))
+		err := fmt.Errorf("%w, max paths is set to %d", ErrRequestSizeExceeded, DefaultMaxDeleteObjects)
+		writeError(w, http.StatusInternalServerError, err)
 		return
 	}
 
-	// delete all the files and collect responses
-	var errs []ObjectError
+	// errs used to collect errors as part of the response, can't be nil
+	errs := make([]ObjectError, 0)
+	// check if we authorize to delete each object, prepare a list of paths we can delete
+	var pathsToDelete []string
 	for _, objectPath := range body.Paths {
-		// authorize this object deletion
 		if !c.authorize(w, r, permissions.Node{
 			Permission: permissions.Permission{
 				Action:   permissions.DeleteObjectAction,
@@ -126,14 +127,24 @@ func (c *Controller) DeleteObjects(w http.ResponseWriter, r *http.Request, body 
 				StatusCode: http.StatusUnauthorized,
 				Message:    http.StatusText(http.StatusUnauthorized),
 			})
-			continue
+		} else {
+			pathsToDelete = append(pathsToDelete, objectPath)
 		}
+	}
 
+	// batch delete the entries we allow to delete
+	delErr := c.Catalog.DeleteEntries(ctx, repository, branch, pathsToDelete)
+	delErrs := graveler.NewMapDeleteErrors(delErr)
+	for _, objectPath := range pathsToDelete {
+		// set err to the specific error when possible
+		err := delErrs[objectPath]
+		if err == nil {
+			err = delErr
+		}
 		lg := c.Logger.WithField("path", objectPath)
-		err := c.Catalog.DeleteEntry(ctx, repository, branch, objectPath)
 		switch {
-		case errors.Is(err, catalog.ErrNotFound):
-			lg.Debug("tried to delete a non-existent object")
+		case errors.Is(err, catalog.ErrNotFound), errors.Is(err, graveler.ErrNotFound):
+			lg.WithError(err).Debug("tried to delete a non-existent object")
 		case errors.Is(err, graveler.ErrWriteToProtectedBranch):
 			errs = append(errs, ObjectError{
 				Path:       StringPtr(objectPath),
@@ -156,12 +167,7 @@ func (c *Controller) DeleteObjects(w http.ResponseWriter, r *http.Request, body 
 			lg.Debug("object set for deletion")
 		}
 	}
-	// no content in case there are no errors
-	if len(errs) == 0 {
-		writeResponse(w, http.StatusNoContent, nil)
-		return
-	}
-	// status ok with list of errors
+
 	response := ObjectErrorList{
 		Errors: errs,
 	}
--- a/pkg/api/controller_test.go
+++ b/pkg/api/controller_test.go
@@ -21,6 +21,8 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/davecgh/go-spew/spew"
+
 	"github.com/go-test/deep"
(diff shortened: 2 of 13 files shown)


pkg/graveler/graveler.go
```diff
@@ -1453,40 +1453,71 @@ func (g *KVGraveler) Delete(ctx context.Context, repository *RepositoryRecord, b
...
+func (g *KVGraveler) DeleteBatch(ctx context.Context, repository *RepositoryRecord, branchID BranchID, keys []Key) error {
+	isProtected, err := g.protectedBranchesManager.IsBlocked(ctx, repository, branchID, BranchProtectionBlockedAction_STAGING_WRITE)
+	if err != nil {
+		return err
+	}
+	if isProtected {
+		return ErrWriteToProtectedBranch
+	}
+	if len(keys) > DeleteKeysMaxSize {
+		return fmt.Errorf("%w keys length passsed %d", ErrInvalidValue, DeleteKeysMaxSize)
```
itaiad200, line 1471:
```suggestion
		return fmt.Errorf("keys length (%d) passed the maximum allowed(%d): %w", len(keys), DeleteKeysMaxSize, ErrInvalidValue)
```

pkg/graveler/graveler.go
```diff
@@ -2775,3 +2806,20 @@ func NewRepoInstanceID() string {
 	tm := time.Now().UTC()
 	return xid.NewWithTime(tm).String()
 }
+
+// NewMapDeleteErrors map multi error holding DeleteError to a map of object key -> error
+func NewMapDeleteErrors(err error) map[string]error {
```
itaiad200, line 2811: Can we move it to the `errors.go` file?

pkg/api/controller_test.go
```diff
@@ -21,6 +21,8 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/davecgh/go-spew/spew"
+
```
itaiad200, line 25: goimports

pkg/gateway/operations/deleteobjects.go
```diff
@@ -56,43 +69,107 @@ func (controller *DeleteObjects) Handle(w http.ResponseWriter, req *http.Request
...
+	}
+
+	// construct response - concat what we had so far with delete results
 	if len(errs) > 0 {
-		resp.Error = errs
+		resp.Error = append(errs, resp.Error...)
 	}
-	if !decodedXML.Quiet && len(responses) > 0 {
-		resp.Deleted = responses
+	if len(responses) > 0 {
```
itaiad200, line 108: Where is responses being used? I'm confused

itaiad200 (changes requested):
Great progress!

pkg/api/controller.go
```diff
@@ -106,15 +106,16 @@ func (c *Controller) DeleteObjects(w http.ResponseWriter, r *http.Request, body
...
-			ErrRequestSizeExceeded, DefaultMaxDeleteObjects))
+		err := fmt.Errorf("%w, max paths is set to %d", ErrRequestSizeExceeded, DefaultMaxDeleteObjects)
+		writeError(w, http.StatusInternalServerError, err)
 		return
 	}
 
-	// delete all the files and collect responses
-	var errs []ObjectError
+	// errs used to collect errors as part of the response, can't be nil
+	errs := make([]ObjectError, 0)
```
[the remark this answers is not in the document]
nopcoder: this is the current behavior we have through the S3 gateway and the api. 
Each item will be processed and return an error for the specific path.
The S3 gateway also enables different paths that means different branches for graveler.

pkg/gateway/operations/deleteobjects.go
```diff
@@ -28,9 +30,20 @@ func (controller *DeleteObjects) Handle(w http.ResponseWriter, req *http.Request
 		_ = o.EncodeError(w, req, gerrors.Codes.ToAPIErr(gerrors.ErrBadRequest))
 		return
 	}
+
+	// TODO(barak): limit request size to DefaultMaxDeleteObjects
```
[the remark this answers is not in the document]
nopcoder: yes, I'll open an issue and remove the comment

pkg/gateway/operations/deleteobjects.go
nopcoder: Fixed - leftovers from the refactor.

pkg/graveler/graveler.go
```diff
@@ -1453,40 +1453,71 @@ func (g *KVGraveler) Delete(ctx context.Context, repository *RepositoryRecord, b
...
+		return err
+	}
+	if isProtected {
+		return ErrWriteToProtectedBranch
+	}
+	if len(keys) > DeleteKeysMaxSize {
+		return fmt.Errorf("%w keys length passsed %d", ErrInvalidValue, DeleteKeysMaxSize)
+	}
+	var m *multierror.Error
+	err = g.safeBranchWrite(ctx, g.log.WithField("operation", "delete_keys"),
```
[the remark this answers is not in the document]
nopcoder: 1. Didn't want to introduce parallel work yet. Maybe in followup PR and/or more testing to see if we still have performance issue with this one.
2. Agree about the internal fetch commit information - we can cache this information once.

pkg/api/controller.go
```diff
@@ -126,14 +127,24 @@ func (c *Controller) DeleteObjects(w http.ResponseWriter, r *http.Request, body
...
-			continue
+		} else {
+			pathsToDelete = append(pathsToDelete, objectPath)
 		}
+	}
 
+	// batch delete the entries we allow to delete
+	delErr := c.Catalog.DeleteEntries(ctx, repository, branch, pathsToDelete)
+	delErrs := graveler.NewMapDeleteErrors(delErr)
+	for _, objectPath := range pathsToDelete {
```
itai-david, line 138: Why are you iterating the pathsToDelete, rather than the actual errors?
What should happen if all deletions in batch are successful?

pkg/graveler/graveler.go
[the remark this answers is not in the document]
itaiad200: 3. For every entry you are:
- Reading from every single token in the staging area once.
- Reading from the Committed area once.

Assuming many of the keys are closer together:
- Listing the staging area would get many of the staging entries in a single call per token.
- Same goes for the committed area, although the performance gain is negligible assuming the ranges are eventually cached.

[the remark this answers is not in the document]
itaiad200: We can start without it and see if the current improvements are enough. If they aren't, we can introduce these followup improvements next.

pkg/api/controller.go
nopcoder: At the first part of the function we verified authorization - each file that we do not have access we added an error or added to 'pathsToDelete'.
The list 'pathsToDelete' represents all the files we need to delete (errors will include why we can't delete the rest).
Part of the protocol we need to report (optional - quiet flag) which files we deleted or not.
So going over the files it is to either report a successful delete or the specific/general error why it is not.

itaiad200 (approved):
LGTM

PR #4208, opened 2022-09-20
Setup to migrate auth-api installations too

### LINKED ISSUE #4207: lakeFS setup should migrate up AuthAPI installations too

itaiad200:
closes #4207

### DIFF
--- a/cmd/lakefs/cmd/setup.go
+++ b/cmd/lakefs/cmd/setup.go
@@ -22,11 +22,6 @@ var setupCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		cfg := loadConfig()
 
-		if cfg.IsAuthTypeAPI() {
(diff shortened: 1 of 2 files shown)


cmd/lakefs/cmd/setup.go
```diff
@@ -37,6 +37,11 @@ var setupCmd = &cobra.Command{
 			os.Exit(1)
 		}
 
+		if cfg.IsAuthTypeAPI() {
```
ortz, line 40: there's a conditional 15 lines above that exists if it's using the AuthTypeAPI

ortz (changes requested):
thanks! one comment :)

nopcoder (changes requested):
Thanks looks I blocked setup completely when auth API is used. When I should just block the admin setup.
Just remove the block in lakefs setup command.

nopcoder (approved):

ortz (approved):

PR #4204, opened 2022-09-20
[lakeFSFS] Use bulk deletes in recursive FileSystem delete

arielshaqed:
Fixes #4203.

PR #4209, opened 2022-09-20
Integrate server into bulk delete lakeFSFS

arielshaqed:
DO NOT MERGE, this merges #4204 and #4205 to see if integration tests are any faster.


arielshaqed:
No go.  No good.

PR #4212, opened 2022-09-21
Change Esti S3 endpoint client to use path style format

idanovo:


pkg/testutil/setup.go
```diff
@@ -111,6 +111,7 @@ func SetupTestingEnv(params *SetupTestingEnvParams) (logging.Logger, api.ClientW
 			WithRegion("us-east-1").
 			WithEndpoint(s3Endpoint).
 			WithDisableSSL(true).
+			WithS3ForcePathStyle(true).
```
itaiad200, line 114: Does this mean we're not covering the host base style in Esti no more?

esti/delete_objects_test.go
```diff
@@ -42,7 +42,52 @@ func TestDeleteObjects(t *testing.T) {
...
+	listOut, err = pathStyleSvc.ListObjects(&s3.ListObjectsInput{
+		Bucket: aws.String(repo),
+		Prefix: aws.String(mainBranch + "/"),
+	})
+
+	assert.NoError(t, err)
+	assert.Len(t, listOut.Contents, 0)
+}
+
+func TestDeleteObjectsHostStyleSvc(t *testing.T) {
```
itaiad200, line 54: Instead of copying the entire test which may lead to future bugs, you can write it as a function that accepts the S3 client and uses that throughout the test. Then simply trigger that function with the appropriate client when needed

esti/multipart_test.go
```diff
@@ -54,12 +54,53 @@ func TestMultipartUpload(t *testing.T) {
 	require.Equal(t, partsConcat, getResp.Body, "uploaded object did not match")
 }
 
-func uploadMultipartParts(t *testing.T, logger logging.Logger, resp *s3.CreateMultipartUploadOutput, parts [][]byte, firstIndex int) []*s3.CompletedPart {
+func TestMultipartUploadHostStyleClient(t *testing.T) {
```
itaiad200, line 57: Why is this test unique to host-style? Shouldn't we test it for path-style too?

pkg/testutil/setup.go
```diff
@@ -106,19 +106,32 @@ func SetupTestingEnv(params *SetupTestingEnvParams) (logging.Logger, api.ClientW
...
+		aws.NewConfig().
+			WithRegion("us-east-1").
+			WithEndpoint(s3Endpoint).
+			WithDisableSSL(true).
+			WithCredentials(credentials.NewCredentials(
+				&credentials.StaticProvider{
+					Value: credentials.Value{
+						AccessKeyID:     viper.GetString("access_key_id"),
+						SecretAccessKey: viper.GetString("secret_access_key"),
+					}})))
```
itaiad200, line 132: Let's extract common code to a single location and not repeat that when it's not mandatory.

pkg/testutil/setup.go
```diff
@@ -31,7 +31,7 @@ type SetupTestingEnvParams struct {
 	AdminSecretAccessKey string
 }
 
-func SetupTestingEnv(params *SetupTestingEnvParams) (logging.Logger, api.ClientWithResponsesInterface, *s3.S3) {
+func SetupTestingEnv(params *SetupTestingEnvParams) (logging.Logger, api.ClientWithResponsesInterface, *s3.S3, *s3.S3) {
```
itaiad200, line 34: Usually when the return args list gets long, especially when there are some with equal type, it's common to name the return variables for clarity
```suggestion
func SetupTestingEnv(params *SetupTestingEnvParams) (log logging.Logger, apiClient api.ClientWithResponsesInterface, pathStyleSvc *s3.S3, hostStyleSvc *s3.S3) {
```

esti/migrate_test.go
```diff
@@ -190,7 +190,7 @@ func saveStateInLakeFS(t *testing.T) {
 	_ = createRepositoryByName(ctx, t, migrateStateRepoName)
 
 	// file is big - so we better use multipart writing here.
```
itaiad200, line 192: Feel free to delete this test, it's no longer needed.

itaiad200 (changes requested):
Thanks for adding host-style tests! Added some comments around reusing tests/code rather than duplicating them

esti/multipart_test.go
idanovo: we do

esti/migrate_test.go
idanovo: What about the other migrate tests?

nopcoder (changes requested):
From the PR it looks like delete and multipart are using the the s3 client to test the gateway.
I think we can get from viper 'S3ForcePathStyle' value and run the same tests for all environments.

idanovo:
> From the PR it looks like delete and multipart are using the the s3 client to test the gateway. I think we can get from viper 'S3ForcePathStyle' value and run the same tests for all environments.

@nopcoder but we want to test it once with a host-base client and once using a path-base client, don't we?

nopcoder:
> @nopcoder but we want to test it once with a host-base client and once using a path-base client, don't we?

True, when you pull the value from viper you can run the system tests when it is 'off' and other tests when it is 'on'.

arielshaqed (approved):
Neat solution!  Also easier to parallelize in Esti :-)

nopcoder (approved):
LGTM

itaiad200 (approved):
LGTM

PR #4220, opened 2022-09-21
Parallel commits compare

itaiad200:
closes https://github.com/treeverse/lakeFS/issues/3964


pkg/catalog/catalog.go
```diff
@@ -974,6 +979,24 @@ func (c *Catalog) GetCommit(ctx context.Context, repositoryID string, reference
 	return catalogCommitLog, nil
 }
 
+// wrapper for simplelru.LRU that is safe to access from multiple go routines
+type safeCache struct {
```
nopcoder, line 983: Can be removed and use 'github.com/hnlq715/golang-lru' - the current lru is the underlaying storage.

pkg/catalog/catalog.go
```diff
@@ -199,12 +205,13 @@ func New(ctx context.Context, cfg Config) (*Catalog, error) {
 	protectedBranchesManager := branch.NewProtectionManager(settingManager)
 	stagingManager := staging.NewManager(ctx, *cfg.KVStore)
 	gStore := graveler.NewKVGraveler(committedManager, stagingManager, refManager, gcManager, protectedBranchesManager)
-
+	workpool := pond.New(sharedWorkers, sharedWorkers*workersToBufferRatio)
```
N-o-Z, line 208: Might want to initialize the pool with the parent context
```suggestion
	workpool := pond.New(sharedWorkers, sharedWorkers*workersToBufferRatio, pond.Context(ctx))
```
Also, did you try to play around with the resize strategy and see which provides the best performance?

pkg/catalog/catalog.go
```diff
@@ -122,6 +126,8 @@ const (
 	ListTagsLimitMax         = 1000
 	DiffLimitMax             = 1000
 	ListEntriesLimitMax      = 10000
+	sharedWorkers            = 15
+	workersToBufferRatio     = 3
```
N-o-Z, line 130: I don't really understand why we need to do the math here and not just define the actual buffer size we want to use

pkg/catalog/catalog.go
```diff
@@ -1548,6 +1655,7 @@ func (c *Catalog) Close() error {
 			_ = multierror.Append(errs, err)
 		}
 	}
+	c.workpool.StopAndWaitFor(1 * time.Second)
```
N-o-Z, line 1658: Seems like 1 second might be a bit strict

N-o-Z (commented):
Thanks Itai!
I have read through this, and though I don't have any substantial comments - it seems like the ListCommits function became very long, and at points very complicated to understand. I think that if there's a way to break this into smaller logical units it would be much easier to read

itaiad200:
> Thanks Itai! I have read through this, and though I don't have any substantial comments - it seems like the ListCommits function became very long, and at points very complicated to understand. I think that if there's a way to break this into smaller logical units it would be much easier to read

I agree. Improved it some more. Breaking it for path / non-paths log-commits usage should make the code clearer at least for the common flow (non-paths). LMKWYT

itaiad200:
Tested it locally by adding 200ms delay to block adapter read and zeroing out the cache. For 200 commits we currently log all of them in 78 seconds. This change reduces that number to 6 seconds.

pkg/catalog/catalog.go
```diff
@@ -114,6 +117,7 @@ type Catalog struct {
 	log           logging.Logger
 	walkerFactory WalkerFactory
 	managers      []io.Closer
+	workpool      *pond.WorkerPool
```
itai-david, line 120:
```suggestion
	workPool      *pond.WorkerPool
```

pkg/catalog/catalog.go
```diff
@@ -200,11 +206,14 @@ func New(ctx context.Context, cfg Config) (*Catalog, error) {
 	stagingManager := staging.NewManager(ctx, *cfg.KVStore)
 	gStore := graveler.NewKVGraveler(committedManager, stagingManager, refManager, gcManager, protectedBranchesManager)
 
+	// The size of the workpool is determined by the number of workers and the number of desired pending tasks for each worker.
+	workpool := pond.New(sharedWorkers, sharedWorkers*pendingTasksPerWorker, pond.Context(ctx))
```
itai-david, line 210:
```suggestion
	workPool := pond.New(sharedWorkers, sharedWorkers*pendingTasksPerWorker, pond.Context(ctx))
```

pkg/catalog/catalog.go
```diff
@@ -200,11 +206,14 @@ func New(ctx context.Context, cfg Config) (*Catalog, error) {
...
 	return &Catalog{
 		BlockAdapter:  tierFSParams.Adapter,
 		Store:         gStore,
 		log:           logging.Default().WithField("service_name", "entry_catalog"),
 		walkerFactory: cfg.WalkerFactory,
+		workpool:      workpool,
```
itai-david, line 216:
```suggestion
		workPool:      workPool,
```

pkg/catalog/catalog.go
N-o-Z: What about the resize strategy? Have you looked into it?

pkg/catalog/catalog.go
N-o-Z: Ping

N-o-Z (commented):
Some comments haven't been addressed yet

pkg/catalog/catalog.go
```diff
@@ -1015,58 +1024,151 @@ func (c *Catalog) ListCommits(ctx context.Context, repositoryID string, branch s
...
+		if foundAllCommits(params, commits) {
+			// All results returned until the last commit found
+			// and the number of commits found is as expected.
+			// we have what we need. return commits, true
 			break
 		}
 	}
-	if err := it.Err(); err != nil {
-		return nil, false, err
+
```
itai-david, line 1146: Needed `it.Err()` check after `it.Next()` returns false

pkg/catalog/catalog.go
```diff
@@ -1015,58 +1024,151 @@ func (c *Catalog) ListCommits(ctx context.Context, repositoryID string, branch s
...
+		for ; it.Next(); num++ {
+			numm := num
+			vall := it.Value()
+			if done.Load() {
+				break
 			}
-		}
+			workerGroup.Submit(func() error {
+				if done.Load() {
+					out <- compareJobResult{num: numm}
```
itai-david, line 1063: Why do we need that last result? is `done.Load()` mean you are already draining?

pkg/catalog/catalog.go
```diff
@@ -1075,10 +1177,57 @@ func (c *Catalog) ListCommits(ctx context.Context, repositoryID string, branch s
 	return commits, hasMore, nil
 }
 
+func drain(done *atomic.Bool, compareJobResults chan compareJobResult) {
+	done.Store(true)
+
```
itai-david, line 1182: `close(compareJobResults)` - somewhere here? Not sure it is **required**

pkg/api/controller_test.go
```diff
@@ -240,65 +241,127 @@ func TestController_LogCommitsHandler(t *testing.T) {
...
+
+			commitsLog := resp.JSON200.Results
+			if len(commitsLog) != tt.expectedCommits {
+				t.Fatalf("Log %d commits, expected %d", len(commitsLog), tt.expectedCommits)
+			}
+		})
+	}
+}
+
+func TestController_LogCommitsParallelHandler(t *testing.T) {
```
itai-david, line 320: Does this verify that `LogCommits` is parallelized by the new code, or is it not the purpose of that test?
If it does, can you please explain how it is verified?
Thanks

itai-david (changes requested):
A very challenging one - both to implement and to review :-)

`listCommitsWithPaths` is somewhat complex and challenging tyo follow. I think it deserves an explanation text of what exaclty is done by each part of the `mgmtGroup` and why. Will definitely make it easier to understand it in the future.
Other than that some minor comments/requests, but mainley questions

pkg/catalog/catalog.go
itaiad200: Changed to 5 seconds.

pkg/api/controller_test.go
itai-david: Thanks :pray:

itai-david (approved):
Thanks for the answers and fixes. LGTM

pkg/catalog/catalog.go
itaiad200: Sorry, missed that. What do you mean by resize strategy? `pond` can scale up to the given maximum number of workers. It's not just a matter of performance, we should also consider starvation for other requests too. I assumed that capping it to 15 would allow serving several (up to 15) LogCommits concurrent requests, while not hurting performance of other requests. What measurements would you like me to perform?

N-o-Z: See documentation at: https://github.com/alitto/pond

> 
> - Strategy: Configures the strategy used to resize the pool when backpressure is detected. You can create a custom strategy by implementing the pond.ResizingStrategy interface or choose one of the 3 presets:
>   - Eager: maximizes responsiveness at the expense of higher resource usage, which can reduce throughput under certain conditions. This strategy is meant for worker pools that will operate at a small percentage of their capacity most of the time and may occasionally receive bursts of tasks. This is the default strategy.
>   - Balanced: tries to find a balance between responsiveness and throughput. It's suitable for general purpose worker pools or those that will operate close to 50% of their capacity most of the time.
>   - Lazy: maximizes throughput at the expense of responsiveness. This strategy is meant for worker pools that will operate close to their max. capacity most of the time.

itaiad200: Thanks for the explanation - I did miss that.
The default strategy is `Eager` and I think that it suits LogCommits usages best of all 3.

nopcoder (commented):
Coded the suggested changes in draft/branch chore/parallelize-list-commit-2-alt.
- Used a goroutine+channel+heap to have ordered results for the commit log. It was more clear to me how to process the worker's jobs this way.
- Added more tests with predefineed commits log - compare the expected results.

N-o-Z (approved):
LGTM! Thanks for addressing all the comments!

itaiad200:
Tested locally again after all changes applied, the above numbers are still valid. Thanks all for the improvements and the review! Pulling..

Issue #2344, opened 2021-08-10
Add project description to lakeFS’s python package.


guy-har:
We should have some description in the [lakeFS’s python package](https://pypi.org/project/lakefs-client/) referencing the [lakeFS docs](https://docs.lakefs.io/integrations/python.html) and the [lakeFS pydocs](https://pydocs.lakefs.io/).

vaibhavk1992:
Hi Team,
I am planning to work on this issue.Please tag me to it. I will be touch over slack for further communication.

PR #4230, opened 2022-09-22
Adding template support to python client build

itai-david:
Adding `setup.mustache` and configuring `Makefile` to use it and to not delete it.
Done to allow to override of the `long_description` field to allow fix of #2344 

We should consider upgrading our OpenAPI Generator!!


itai-david:
Code was merged as part of #4260

PR #4239, opened 2022-09-22
Garbage Collector- validate that the written report DataFrame isn't empty

### LINKED ISSUE #4238: Running GC in dry run mode always fails
When running GC in dry-run mode it will always fail because it tries to write a DataFrame with a report of the actual deleted objects to the bucket, but this is not possible when the DataFrame is empty.
Exception:
`Cannot use all columns for partition columns`

Jonathan-Rosenberg:
Check that the written DataFrame is not empty before writing it, so that it won't throw an exception.

### How was this tested?
I ran the GC in dry-run mode against a repo with ~850K objects to be deleted. It used to fail on that repo but now it finishes successfully.

Fixes #4238

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
@@ -15,9 +15,12 @@ import java.time.format.DateTimeFormatter
 import org.json4s.native.JsonMethods._
 import org.json4s._
 import org.json4s.JsonDSL._
+
 import java.time.LocalDateTime
 import java.time.ZoneOffset
 import io.lakefs.clients.api.model.GarbageCollectionPrepareResponse
+import org.apache.spark.sql.types.{IntegerType, StringType, StructField, StructType}
+
 import java.util.UUID
 
 class ConfigMapper(val hcValues: Broadcast[Array[(String, String)]]) extends Serializable {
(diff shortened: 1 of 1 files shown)


clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -340,24 +340,36 @@ object GarbageCollector {
...
-    removed
-      .withColumn("run_id", lit(runID))
-      .write
-      .partitionBy("run_id")
-      .mode(SaveMode.Overwrite)
-      .parquet(
-        concatToGCLogsPrefix(storageNSForHadoopFS, s"deleted_objects/$time/deleted.parquet")
-      )
-
-    spark.close()
```
arielshaqed, line 360: No longer closing?

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -340,24 +340,36 @@ object GarbageCollector {
...
+    if(!removed.isEmpty) {
```
arielshaqed, line 363: But: this finishes running silently if it succeeded.  So any program downstream that needs to process the list of deleted objects now has to have a special-case for a GC run that succeeded without deleting anything.

I would very much prefer a consistent output, of an empty object.  There do exist empty Parquet files.

arielshaqed (changes requested):
Thanks!  This definitely solves the issue, but it introduces an inconsistency that will be hard for automated programs to follow.

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
Jonathan-Rosenberg: closes in line 345

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
Jonathan-Rosenberg: @arielshaqed

arielshaqed (approved):
Neat, thanks!
Mon blocking, but can the String be non-nullable? That will better reflect what it really is.

arielshaqed:
Also, I think this is worth releasing today any which way. It's a serious noisy to usability!

PR #4249, opened 2022-09-25
UI remove unknown webkit-border from css

nopcoder:
During build:

```Invalid property name '-webkit-border-' at 13579:4.```

It should match one of the webkit css properties - removing as it is currently ignored.


eladlachmi (approved):
LGTM

PR #4250, opened 2022-09-25
Use path style for minio client in Esti

idanovo:


nopcoder:
Talked online - I don't think this one is required, as the error that brought us to change the test can be caused from misconfiguration.

PR #4252, opened 2022-09-25
Fix logging.IsTracing should check default logger level

### LINKED ISSUE #4251: Conditional logging.IsTracing doesn't work
The logging package IsTracing doesn't reflect the logging package tracing level.
Calling logging.SetLevel("trace") does not effect the return value from 'IsTracing' so any conditional code we use to print trace message does not execute.

nopcoder:
Fix #4251

### DIFF
--- a/pkg/logging/logger.go
+++ b/pkg/logging/logger.go
@@ -271,7 +271,7 @@ func (l logrusEntryWrapper) Logf(level logrus.Level, format string, args ...inte
 }
 
 func (*logrusEntryWrapper) IsTracing() bool {
-	return logrus.IsLevelEnabled(logrus.TraceLevel)
+	return defaultLogger.IsLevelEnabled(logrus.TraceLevel)
 }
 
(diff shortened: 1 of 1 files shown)


Jonathan-Rosenberg (approved):
Thanks!

pkg/logging/logger.go
```diff
@@ -271,7 +271,7 @@ func (l logrusEntryWrapper) Logf(level logrus.Level, format string, args ...inte
 }
 
 func (*logrusEntryWrapper) IsTracing() bool {
-	return logrus.IsLevelEnabled(logrus.TraceLevel)
+	return defaultLogger.IsLevelEnabled(logrus.TraceLevel)
```
arielshaqed, line 274: Neat!  I remember this line having a long and worrying history :-/

nopcoder: now I'm worried from the amount of logging while we trace.

Issue #4223, opened 2022-09-21
Migrate lakeFS webui to TypeScript


eladlachmi:
Migrate the lakeFS webui to TypeScript.
This effort will include setting up the npm project for using TypeScript and migrating existing code to TS, so we don't need to start off with a lax `tsconfig`, which will require additional refactoring efforts in the future.

As the code base grows, TS's value will hugely outweigh its overhead.
It will also make it easier to contribute to the webui of lakeFS confidently.

Feedback is very welcome!

PR #4253, opened 2022-09-25
TS Conversion: Step 1 - tooling

eladlachmi:
Step 1 of #4223 

To make the entire effort more review-able, I'm breaking it down to multiple steps.
Each step will be more manageable for review vs. a huge changeset.


eladlachmi:
Checked both dev server and production builds and both work as they did previously.

webui/.eslintrc
```diff
@@ -0,0 +1,29 @@
...
+      "jest": true,
+      "es6": true
+    },
+    "settings": {
+      "react": {
+        "version": "17.0"
+      }
+    },
+    "rules": {
+      "eqeqeq": "warn",
```
arielshaqed, line 17: :-(

webui/src/lib/components/controls.jsx
```diff
@@ -25,18 +25,18 @@ import {
 
 const defaultDebounceMs = 300;
 
-export function debounce(func, wait, immediate) {
+export const debounce = (func, wait, immediate) => {
```
arielshaqed, line 28: IIRC this rename is subtly different with regard to where `debounce` is considered defined.  Is this change intentional?

webui/src/lib/components/controls.jsx
```diff
@@ -25,18 +25,18 @@ import {
...
     let timeout;
     return function() {
-        let context = this, args = arguments;
+        let args = arguments;
         let later = function() {
             timeout = null;
-            if (!immediate) func.apply(context, args);
+            if (!immediate) func.apply(null, args);
```
arielshaqed, line 34: This changes the semantics by a _lot_, now AFAIR `func` cannot be a method call!

arielshaqed (changes requested):
Thanks!  This is going to be such a brilliant step forward!

How was this tested?  And... how do we plan to test the followup PRs?

I'm a bit concerned about the behavioural change introduced regarding `debounce`, and an answer there is the only blocker.

webui/package.json
```diff
@@ -30,36 +31,27 @@
...
     "babel-eslint": "^10.1.0",
     "eslint-plugin-react": "^7.23.2",
-    "jest": "^27.3.1",
+    "jest": "^29.0.3",
     "react-test-renderer": "^17.0.2",
+    "ts-jest": "^29.0.2",
+    "ts-node": "^10.9.1",
+    "typescript": "^4.8.3",
+    "typesync": "^0.9.2",
     "vite": "^2.3.7",
```
nopcoder, line 54: consider moving to vide 3

webui/src/lib/components/controls.jsx
nopcoder, line 28: Also, I saw you added lodash, which also include an implementation for debounce.

webui/.eslintrc
eladlachmi: It was a direct copy of what was defined in `package.json`. I don't like including these things in `package.json` because I feel it's a mixing of concerns (and `package.json` is over-loaded as is), so I moved it as-is to `.eslintrc`.
Having said that, I agree a `warn` is not enough for this rule. However, I need to check that it doesn't break existing code and refactor if it does.

webui/src/lib/components/controls.jsx
eladlachmi: The original function made a copy of `this`, which is a no-no these days. Instead of making a copy of `this` and passing it along to `apply`, an arrow function "naturally" conserves the context of the module in which it resides.
That was the idea. Regardless, I mean to replace this debounce implementation with one from `lodash`, which is more battle-tested IMHO.

eladlachmi: I agree. I just wrote that two comments ago :)

webui/package.json
eladlachmi: Thanks for the feedback. I'm not sure what (if any) the breaking changes are. I'll look into it.

webui/src/lib/components/controls.jsx
arielshaqed: Not a JavaScript expert, but this is at odds with [this (random) StackOverflow answer](https://stackoverflow.com/a/33640254/192263).  Also MDN matches that answer and does not specify special behaviour for when `thisArg` is `null`.

To my understanding the new version is subtly different from the old one.  It's just that both will be subtly surprising in different ways when called from an object.  I'd rather not change it and just go for `_.debounce` (for some value of `_`), but I won't block on this.  Instead, we should just go over the actual current uses of `debounce` and see that none care about `this`.

arielshaqed (approved):
Thanks!

But please go over uses of debounce, it now calls `func` with `this === null` :-/

eladlachmi:
@arielshaqed I've done exactly that. Debounce is currently used for the repo search input. If you test it out, it makes the header twitchy as-is, which is also not great. I think that replacing the home-grown implementation with the `lodash` one is a win for all use cases. However, since the current functionality is identical to the functionality before this commit, I suggest we include it in the next commit/step. WDYT?

eladlachmi:
@arielshaqed To your more general question of how this was tested and how I plan to test this moving forward - I'm trying to make the most minimal changes possible in anything that's not directly related to TypeScript.
As long as things compile, we can debate typeing, if we're strict enough or not, and if we're using a consistent paradigm, but it's not going to break at runtime.

Where I must make changes that are not only in types, I try to define the blast radius and test the UI manually.
Where I'm unsure, I ask around to make sure I didn't miss any use cases.

My next long term goal, after TypeScript, will be E2E testing, which will help with this considerably.

PR #4254, opened 2022-09-25
GC: two runnable steps proposal

### LINKED ISSUE #4255: GC - intermediate results
Allow examination of intermediate results before deleting the objects, and a way to proceed with the previously generated results.

Jonathan-Rosenberg:
This proposal includes breaking a GC run into two runnable steps:
1. **Marking run**- collecting the addresses to be removed.
2. **Sweeping run**- removing the objects that were mapped by the collected addresses.

Closes #4255

### DIFF
--- a/design/open/gc-two-steps-separation.md
+++ b/design/open/gc-two-steps-separation.md
@@ -42,6 +42,10 @@ If the `lakefs.gc.only_sweep` is enabled, and `lakefs.gc.mark_id` is not provide
 GC will use the `lakefs.gc.mark_id` to write intermediate results to `STORAGE_NAMESPACE/_lakefs/retention/gc/addresses/mark_id=MARK_ID/`,
 and will do a complete run (marking and sweeping).
 
+#### Only Mark and Only Sweep
+
+The operation will fail.
+
 #### No flags are provided
 
 Complete run of GC, including marking and sweeping (while writing intermediate results to `STORAGE_NAMESPACE/_lakefs/retention/gc/addresses/mark_id=RANDOM_VALUE/`).
--- a/design/open/gc-two-steps-separation.md
+++ b/design/open/gc-two-steps-separation.md
@@ -11,38 +11,38 @@ The `lakefs.debug.gc.no_delete` flag lets users run GC and generate intermediate
 yet GC doesn't continue over those results, and retrieving them manually is quite tedious.  
 Both problems can be solved by introducing two flags and changing the `lakefs.debug.gc.no_delete`:
 
-1. `lakefs.debug.gc.no_delete` -> `lakefs.gc.only_mark` (Boolean): This flag instructs the GC operation to only mark, i.e. collect, the addresses that are intended to be deleted.
-2. `lakefs.gc.only_sweep` (Boolean): The GC operation will only sweep, i.e. delete, the marked addresses based on the provided mark ID.
+1. `lakefs.debug.gc.no_delete` -> `lakefs.gc.do_mark` (Boolean: default = true): This flag instructs the GC operation to mark, i.e. collect, the addresses that are intended to be deleted.
+2. `lakefs.gc.do_sweep` (Boolean: default = true): The GC operation sweep, i.e. delete, the marked addresses.
 3. `lakefs.gc.mark_id` (String): This flag specifies which ID the GC will use to generate intermediate results and write outputs to.
 
 ## Flags permutations
 
 #### Only Mark
 
-In the case of `lakefs.gc.only_mark`, the GC will run, collect the addresses to be deleted, and write them to:
+In the case of `lakefs.gc.do_sweep=false`, the GC will run, collect the addresses to be deleted, and write them to:
 `STORAGE_NAMESPACE/_lakefs/retention/gc/addresses/mark_id=RANDOM_VALUE/`.
 
 #### Only Mark and Mark ID
 
-If the `lakefs.gc.only_mark` is enabled, and a `lakefs.gc.mark_id` is provided,
+If `lakefs.gc.do_sweep=false`, and a `lakefs.gc.mark_id` is provided,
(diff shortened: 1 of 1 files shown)


design/open/gc-two-steps-separation.md
```diff
@@ -0,0 +1,71 @@
...
+
+## How?
+
+The `lakefs.debug.gc.no_delete` flag lets users run GC and generate intermediate results,
+yet GC doesn't continue over those results, and retrieving them manually is quite tedious.  
+Both problems can be solved by introducing two flags and changing the `lakefs.debug.gc.no_delete`:
+
+1. `lakefs.debug.gc.no_delete` -> `lakefs.gc.only_mark` (Boolean): This flag instructs the GC operation to only mark, i.e. collect, the addresses that are intended to be deleted.
+2. `lakefs.gc.only_sweep` (Boolean): The GC operation will only sweep, i.e. delete, the marked addresses based on the provided mark ID.
+3. `lakefs.gc.mark_id` (String): This flag specifies which ID the GC will use to generate intermediate results and write outputs to.
```
talSofer, line 16: @ozkatz had a great idea to use Terraform terminology here: 
only mark -> plan 
only sweep -> apply with a plan id 
mark id -> plan id

@Jonathan-Rosenberg WDYT?

design/open/gc-two-steps-separation.md
```diff
@@ -0,0 +1,71 @@
...
+
+## Flags permutations
+
+#### Only Mark
+
+In the case of `lakefs.gc.only_mark`, the GC will run, collect the addresses to be deleted, and write them to:
+`STORAGE_NAMESPACE/_lakefs/retention/gc/addresses/mark_id=RANDOM_VALUE/`.
+
+#### Only Mark and Mark ID
```
talSofer, line 25: I think that we should fail GC in case this permutation is used, why would we want to override a previous GC plan run (I'm using the new terminology)? 
thoughts?

design/open/gc-two-steps-separation.md
```diff
@@ -0,0 +1,71 @@
...
+#### No flags are provided
+
+Complete run of GC, including marking and sweeping (while writing intermediate results to `STORAGE_NAMESPACE/_lakefs/retention/gc/addresses/mark_id=RANDOM_VALUE/`).
+
+### Note
+
+The use of the `lakefs.gc.mark_id` flag is under the responsibility and management of the users.
+It should be made clear to users that using it sequentially with the same mark ID may override results.
+
+## Current behavior (and why we don't need the `run_id=...` with addresses)
```
talSofer, line 58: Thanks for including this, it was my next question

design/open/gc-two-steps-separation.md
```diff
@@ -0,0 +1,71 @@
...
+There is no need to use run IDs here. It will be possible to limit the work of GC by using Run IDs, but this will be done with **commits** rather than addresses.
+Mark ID seems to make more sense than run ID.
+
+## Unrelated but important
+
+lakeFS generates a run ID and collects the commits that take part in it.
+Between this step and the actual deletion, many things can go wrong, so the GC should write a successful run,
+with the return run ID, in a different location than the one lakeFS refers to,
+i.e. the place where it has written the commits (`/_lakefs/retention/gc/commits/run_id=RUN_ID/commits.csv`).  
+It involves changes to both the GC and lakeFS that should be pulled before version 1.0.0 of lakeFS is released.
```
talSofer, line 70: I'm not sure that I understand this section, can you please clarify?

design/open/gc-two-steps-separation.md
```diff
@@ -0,0 +1,71 @@
...
+
```
talSofer, line 71: We need to think about how and when to make this feature available in lakeFS cloud, and about its user experience. Since cloud users can't control the GC run command, this will require additional effort. I suggest discussing this separately. Can you please open an issue for it? 
cc @ozkatz

talSofer (approved):
Thanks @Jonathan-Rosenberg, this looks great! 

I'm approving but will appreciate it if you consider the comments before merging.

design/open/gc-two-steps-separation.md
Jonathan-Rosenberg: I actually think that we should GC terminology with the GC...

design/open/gc-two-steps-separation.md
Jonathan-Rosenberg: This is the case where we only want to mark the addresses to be deleted with a specific mark ID so we can examine the results according to it.
Why should it fail here?
As explained in the [Note](https://github.com/treeverse/lakeFS/pull/4254/files#diff-5e7e580bed0e5e9fdb5a1a66fab9a832d7229cbf7ed34596b0a5f22c86ff6eb4R53) below, the users who orchestrate the GC run will need to ensure that they don't override the previous mark/plan or do override it if they wish...

arielshaqed: It is a very useful flag combination.  I can use it for automation or testing or whatever.

design/open/gc-two-steps-separation.md
talSofer: Ok, I get it. makes sense

arielshaqed: I prefer mark/sweep which is established terminology for this.

Neither mark/sweep nor plan/apply make it entirely clear that the second stage will delete stuff.  But sweep is clearer, that's another point in its favour.

design/open/gc-two-steps-separation.md
talSofer: I'm ok with it, thanks

design/open/gc-two-steps-separation.md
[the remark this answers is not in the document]
Jonathan-Rosenberg: cool
I'll change it

PR #4259, opened 2022-09-27
added jupyter notebook password in everything bagel docker-compose.yaml

vinodhini8694:
evrything bagel docker-compose.yaml doesn't include password for jupyter notebook service. Added the password as a comment. 

@nopcoder should we rather remove the password altogether?


nopcoder (approved):
If there is a way to skip the password go for it.

PR #4266, opened 2022-09-28
Bulk-delete directory renames

arielshaqed:


arielshaqed:
Tests pass, but we need to decide whether we care about the change to semantics: directory renames continue to be non-atomic, but in a whole new way.

arielshaqed:
> Tests pass, but we need to decide whether we care about the change to semantics: directory renames continue to be non-atomic, but in a whole new way.

... so back to draft this PR goes, until we decide whether we actually want it!

arielshaqed:
This PR reliably shortens one test run from 221-224 seconds to 215-220 seconds.

```
spark-shell ... \
	-c spark.hadoop.mapred.output.committer.class=org.apache.hadoop.mapred.FileOutputCommitter \
	-c spark.hadoop.fs.lakefs.impl=io.lakefs.LakeFSFileSystem \
	-c spark.hadoop.fs.lakefs.secret.key=... \
	-c spark.hadoop.fs.s3a.access.key=... \
	-c spark.hadoop.fs.s3a.bucket.test-1.secret.key=... \
	-c spark.hadoop.spark.sql.sources.outputCommitterClass=org.apache.hadoop.mapreduce.lib.output.FileOutputCommitter \
	-c spark.sql.sources.commitProtocolClass=org.apache.spark.sql.execution.datasources.SQLHadoopMapReduceCommitProtocol \
	-c spark.sql.shuffle.partitions=4000 \
	-c spark.hadoop.mapreduce.fileoutputcommitter.algorithm.version=1 \
	-c spark.hadoop.fs.s3a.path.style.access=true \
	-c spark.hadoop.fs.s3a.secret.key=... \
	-c spark.hadoop.fs.lakefs.access.key=... \
	-c spark.hadoop.fs.lakefs.endpoint=https://orgorg.us-east-1.lakefscloud.ninja/api/v1 \
	-c spark.hadoop.fs.s3a.bucket.test-1.endpoint=https://orgorg.us-east-1.lakefscloud.ninja
```
and then run
```scala
val before = System.nanoTime(); reviews.sample(0.002).repartition(125).write.mode(SaveMode.Overwrite).parquet("lakefs://test-1/test-overwrite2/amazon-reviews-repartition-sample/"); (System.nanoTime() - before).toFloat / 1e9
```

arielshaqed:
We shall probably _not_ be doing this.  But let's keep the branch!

PR #4267, opened 2022-09-28
GC Separation: Enable "mark only" mode

### LINKED ISSUE #4272: GC - intermediate results
Write intermediate results (object addresses to be removed) of a GC run for further examination before deleting the objects it gathered.

Jonathan-Rosenberg:
This PR includes the option to only mark the addresses to be deleted without actually deleting them.
Basically, this is "dry mode"...

Closes #4272


arielshaqed (changes requested):
Neat!

Please document in this PR or open an issue to make sure we don't forget it.

Also, I'd like to keep the old key as another option for a while yet:suppose I have an existing script that sets that key. With this PR it now deletes 🤦🏽. So we should either earn and not delete, or error out if the old key is set to _any_ value.  I tend towards the first option, it did say "debug"  so not many will use it. But some change here is required, sorry.

arielshaqed (approved):
Neat, THANKS!

PR #4274, opened 2022-09-28
GC Separation: Add "run_id" to a "mark"'s metadata 

### LINKED ISSUE #4272: GC - intermediate results

Jonathan-Rosenberg:
This is a way to link between a mark to a run_id.

The addresses will be written to:
* `STORAGE_NAMESPACE/_lakefs/retention/gc/addresses/mark_id=MARK_ID/`

The mark's run metadata (such as its associated run id) will be written to:
* `STORAGE_NAMESPACE/_lakefs/retention/gc/addresses/MARK_ID.meta/`
**as JSON**

The deleted objects report summary will be written to:
`STORAGE_NAMESPACE/_lakefs/logs/gc/deleted_objects/DATE/deleted.parquet/mark_id=MARK_ID/run_id=RUN_ID/`

### How was this tested?
Manually 

Closes #4272

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
@@ -345,7 +345,7 @@ object GarbageCollector {
                storageNSForSdkClient,
                gcAddressesLocation,
                expiredAddresses,
-               runID,
+               markId,
                region,
                storageType,
                schema
@@ -354,7 +354,7 @@ object GarbageCollector {
     val commitsDF = gc.getCommitsDF(runID, gcCommitsLocation)
     writeReports(storageNSForHadoopFS,
                  gcRules,
-                 runID,
+                 markId,
                  commitsDF,
                  expiredAddresses,
                  removed,
@@ -367,7 +367,7 @@ object GarbageCollector {
   private def writeReports(
       storageNSForHadoopFS: String,
       gcRules: String,
-      runID: String,
+      markId: String,
       commitsDF: DataFrame,
       expiredAddresses: DataFrame,
       removed: DataFrame,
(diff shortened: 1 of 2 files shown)


clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -385,15 +389,26 @@ object GarbageCollector {
...
     writeParquetReport(expiredAddresses, reportExpiredDst, time)
     writeJsonSummary(configMapper, reportLogsDst, removed.count(), gcRules, time)
 
+    import spark.implicits._
+    val metadataArray = addressesMarkMetadataDF
+      .select(RUN_ID_KEY)
+      .map(_.getString(0))
+      .collect
+
+    val runID = metadataArray(0)
```
talSofer, line 405: Can you please help me understand the following: 
- I'm not sure I understand the relationship between mark and run ids: is it one-to-many in the sense that a markId can be associated with multiple run ids? 
- Why would this runID be different than the id in line 294? 
Thanks!



Why would the runId associated with a markId be

talSofer (changes requested):
Thanks @Jonathan-Rosenberg! 
Requested changes because there are some open questions I see. I might not understand this correctly so waiting for your comments :)

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
Jonathan-Rosenberg: * Since a mark ID can be provided by the user, multiple run IDs might have the same mark ID (1-n relationship). This is possible if the user wants to overwrite the previous mark ID.
* The run ID in line 294 is generated if a "mark" had occurred (it's generated in the prepareCommits API call in lakeFS). If this is a "sweep-only" run, then we should fetch the run ID from the metadata of the associated mark ID. Having said that, the next PR (spoiler alert) will pass the run ID to `writeReports`

PR #4283, opened 2022-09-29
GC Separation: Enable "sweep only" mode

### LINKED ISSUE #4275: GC - Separate stages: Implement sweep-only mode
Implement Sweep-only mode, i.e. sweep the objects according to the provided mark_id (if not provided-> fail).

Implement according to the [design](https://github.com/treeverse/lakeFS/blob/master/design/accepted/gc-two-steps-separation.md).

Jonathan-Rosenberg:
This PR includes the changes needed to implement the "Sweep-only" mode, that is, given a "mark ID", GC will delete all of the objects according to a previous run addresses output.
* If no mark ID is given, GC will fail, i.e. running GC and providing `spark.hadoop.lakefs.gc.do_mark=false` and **not** providing a `spark.hadoop.lakefs.gc.mark_id=...` value:
Screen Shot 2022-09-29 at 16 57 23

After a "mark-only" mode, with `MY_MARK` as the mark ID, we'll result in the following:
Screen Shot 2022-09-29 at 18 01 52

After running GC in a "sweep-only", with `MY_MARK` as the mark ID, the summary will show that the deleted objects were:
Screen Shot 2022-09-29 at 18 56 47

This PR is presenting the option of using one of the steps of the GC (sweep or mark) without the other. **However, GC can also be run the same way we did before**. The only difference is that a mark ID will be generated for the mark step (if we didn't provide one) and the paths of the addresses will change to `.../_lakefs/retention/gc/addresses/mark_id=<mark id>/`, a mark metadata will be written to `.../_lakefs/retention/gc/addresses/<mark id>.meta/`, and the report of deleted objects would be found at `.../_lakefs/logs/gc/deleted_objects/<utc date>/deleted.parquet/mark_id=<mark id>/run_id=<run id>/`

Closes #4275

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
@@ -345,7 +345,7 @@ object GarbageCollector {
                storageNSForSdkClient,
                gcAddressesLocation,
                expiredAddresses,
-               runID,
+               markId,
                region,
                storageType,
                schema
@@ -354,7 +354,7 @@ object GarbageCollector {
     val commitsDF = gc.getCommitsDF(runID, gcCommitsLocation)
     writeReports(storageNSForHadoopFS,
                  gcRules,
-                 runID,
+                 markId,
                  commitsDF,
                  expiredAddresses,
                  removed,
@@ -367,7 +367,7 @@ object GarbageCollector {
   private def writeReports(
       storageNSForHadoopFS: String,
       gcRules: String,
-      runID: String,
+      markId: String,
       commitsDF: DataFrame,
       expiredAddresses: DataFrame,
       removed: DataFrame,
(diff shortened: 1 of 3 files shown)


clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -475,23 +517,55 @@ object GarbageCollector {
...
+    markMetadataDF
+      .select(RUN_ID_KEY, COMMITS_LOCATION_KEY)
+      .map(row => {
+        var arr = Array[String]()
+        for (i <- 0 until row.length) {
+          arr = arr :+ row.getString(i)
+        }
+        arr
+      })
+      .first()
```
talSofer, line 540: What does this part do?

talSofer (approved):
Thank you @Jonathan-Rosenberg! 

I think that we need to mention in the pr description that from this pr and on the GC usage changes and add an example usage for how to run what's used to be a full GC run (now is mark + sweep).

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
Jonathan-Rosenberg: It transforms the first Row of the DataFrame into Array[String]

Jonathan-Rosenberg:
@talSofer thanks!
Added the changes in the way that GC runs in the description.
I think that different run configuration options should be added to the documentation.

PR #4293, opened 2022-10-02
Correctly report status code and operation ID on S3 gateway

arielshaqed:
Previously we incorrectly reported only status 200.  Also some unknown
operations are misreported as "", meaning Grafana makes up a long
incomprehensible list of labels for them.

### DIFF
--- a/pkg/gateway/handler.go
+++ b/pkg/gateway/handler.go
@@ -126,7 +126,7 @@ func NewHandler(region string, catalog catalog.Interface, multipartsTracker mult
 func (h *handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	setDefaultContentType(w, req)
 	o := req.Context().Value(ContextKeyOperation).(*operations.Operation)
-	operationHandler := h.operationHandlers[o.OperationID]
+	operationHandler := h.operationHandlers[o.FooOperationID]
 	if operationHandler == nil {
 		// TODO(johnnyaug): consider other status code or add text with unknown gateway operation
 		w.WriteHeader(http.StatusNotFound)
--- a/pkg/gateway/handler.go
+++ b/pkg/gateway/handler.go
@@ -126,7 +126,7 @@ func NewHandler(region string, catalog catalog.Interface, multipartsTracker mult
 func (h *handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	setDefaultContentType(w, req)
 	o := req.Context().Value(ContextKeyOperation).(*operations.Operation)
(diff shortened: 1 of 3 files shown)


pkg/gateway/handler.go
```diff
@@ -126,7 +126,7 @@ func NewHandler(region string, catalog catalog.Interface, multipartsTracker mult
 func (h *handler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 	setDefaultContentType(w, req)
 	o := req.Context().Value(ContextKeyOperation).(*operations.Operation)
-	operationHandler := h.operationHandlers[o.OperationID]
+	operationHandler := h.operationHandlers[o.FooOperationID]
```
nopcoder, line 129: can we revert this change?

pkg/gateway/middleware.go
```diff
@@ -127,7 +127,8 @@ func DurationHandler(next http.Handler) http.Handler {
 		o := ctx.Value(ContextKeyOperation).(*operations.Operation)
 		start := time.Now()
 		mrw := httputil.NewMetricResponseWriter(w)
-		next.ServeHTTP(w, req)
+		next.ServeHTTP(mrw, req)
+		operationID := o.OperationID
```
nopcoder, line 131: can revert this one too

pkg/gateway/middleware.go
```diff
@@ -175,7 +176,7 @@ func OperationLookupHandler(next http.Handler) http.Handler {
 		ctx := req.Context()
 		o := ctx.Value(ContextKeyOperation).(*operations.Operation)
 		repoID := ctx.Value(ContextKeyRepositoryID).(string)
-		var operationID operations.OperationID
+		operationID := operations.OperationIDOperationNotFound
```
nopcoder, line 179: I think we need to update the structure (o.OperationID) at this point as we do return in case of an error.

nopcoder (changes requested):

pkg/gateway/handler.go
arielshaqed: YIKES!

pkg/gateway/middleware.go
arielshaqed: removed `operationID` entirely :-)

arielshaqed (commented):
Thanks!

PTAL...

pkg/gateway/middleware.go
nopcoder: Still we don't use this one, based on diff

nopcoder (approved):
Just one unused local var I think we can remove, but looks good - thanks!

PR #4298, opened 2022-10-02
Add skip flag to skip Esti tests

idanovo:


nopcoder (changes requested):
Can we use the same syntax as '-run' - if go will add this capability they will keep the same syntax probably and it will be easy to migrate.

esti/main_test.go
```diff
@@ -309,6 +313,13 @@ func TestMain(m *testing.M) {
...
 	defer func() { _ = server.s.Close() }()
 
+	testsToSkipMap = arrayFlagsToMap(testsToSkip)
 	logger.Info("Setup succeeded, running the tests")
 	os.Exit(m.Run())
 }
+
+func SkipTestIfAskedTo(t *testing.T) {
+	if testsToSkipMap[t.Name()] {
+		t.Skip(fmt.Printf("Skip on test: %s", t.Name()))
```
nopcoder, line 323: just `SkipNow()` as we don't give any good explanation :)

nopcoder (changes requested):

nopcoder (changes requested):

esti/main_test.go
```diff
@@ -308,7 +312,14 @@ func TestMain(m *testing.M) {
 	}
 
 	defer func() { _ = server.s.Close() }()
+	testsToSkipRegex, _ = regexp.Compile(*testsToSkip)
```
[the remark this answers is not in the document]
idanovo: In case we have an error I should fail the run or just run all the tests?

esti/main_test.go
```diff
@@ -308,7 +312,17 @@ func TestMain(m *testing.M) {
 	}
 
 	defer func() { _ = server.s.Close() }()
+	testsToSkipRegex, err = regexp.Compile(*testsToSkip)
+	if err != nil {
+		log.Fatal(err)
```
nopcoder, line 317:
```suggestion
		log.Fatalf("Skip pattern '%s' failed to compile:", *testsToSkip, err)
```

esti/main_test.go
```diff
@@ -308,7 +312,17 @@ func TestMain(m *testing.M) {
 	}
 
 	defer func() { _ = server.s.Close() }()
+	testsToSkipRegex, err = regexp.Compile(*testsToSkip)
```
nopcoder, line 315: if *testsToSkip != "" {
}

esti/main_test.go
```diff
@@ -308,7 +312,17 @@ func TestMain(m *testing.M) {
...
+	if err != nil {
+		log.Fatal(err)
+	}
 
 	logger.Info("Setup succeeded, running the tests")
 	os.Exit(m.Run())
 }
+
+func SkipTestIfAskedTo(t *testing.T) {
+	if testsToSkipRegex.String() != "" && testsToSkipRegex.MatchString(t.Name()) {
```
nopcoder, line 325: testsToSkipRegex
```suggestion
	if testsToSkipRegex != nil && testsToSkipRegex.MatchString(t.Name()) {
```

nopcoder (changes requested):

nopcoder (approved):
LGTM! hope we will have it as part of the next Go release

PR #4301, opened 2022-10-02
Fix UI render error inside repo error

### LINKED ISSUE #4302: UI fail to load on full path to non existing repository path
Based on browser console, there is an error with white screen rendered while entering an address of non-existing repository.

Example: http://localhost:8000/repositories/no-repo/objects

```
Failed to load resource: the server responded with a status of 404 (Not Found)
react-dom.production.min.js:216 TypeError: Cannot read properties of undefined (reading 'toString')
    at fr (controls.jsx:78:63)
    at ha (error.jsx:36:12)
    at ns (react-dom.production.min.js:157:137)
    at Ul (react-dom.production.min.js:267:460)
    at Cc (react-dom.production.min.js:250:347)
    at yc (react-dom.production.min.js:250:268)
    at Sc (react-dom.production.min.js:250:128)
    at Tc (react-dom.production.min.js:243:163)
    at react-dom.production.min.js:123:115
    at e.unstable_runWithPriority (scheduler.production.min.js:18:343)
```

nopcoder:
Fix #4302

- Simplify error naming by using this.constructor.name
- Error and RepositoryInDeletionContainer fix return as control by
  passing error value

### DIFF
--- a/webui/src/lib/api/index.js
+++ b/webui/src/lib/api/index.js
@@ -80,21 +80,21 @@ const apiRequest = async (uri, requestData = {}, additionalHeaders = {}) => {
 export class NotFoundError extends Error {
     constructor(message) {
         super(message)
-        this.name = "NotFoundError"
+        this.name = this.constructor.name;
     }
 }
 
 export class BadRequestError extends Error {
     constructor(message) {
         super(message)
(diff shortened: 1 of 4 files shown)


eladlachmi (approved):
LGTM

Issue #4290, opened 2022-10-02
GC: add backup and restore capability to objects marked as expired


talSofer:
Add the following capabilities: 
- Ability to read list of files from a dry run and copy it to external location (backup)
- Ability to read list of files from a dry run and copy it from external location (restore)

PR #4318, opened 2022-10-04
Add GC backup and restore utility

talSofer:
Closes https://github.com/treeverse/lakeFS/issues/4290

### How was this tested

Tested manually with s3 and Azure blob

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
@@ -1,0 +1,121 @@
+package io.treeverse.clients
+
+import scala.util.control.Breaks._
+import org.apache.hadoop.tools.DistCp
+import org.apache.hadoop.fs.{FileSystem, Path}
+import org.apache.spark.sql.{DataFrame, SparkSession}
+
+import java.net.URI
+
+class GCBackupAndRestore {}
+
+/**
+Example usage:
+S3 -
+  spark-submit --class io.treeverse.clients.GCBackupAndRestore \
+  -c spark.hadoop.fs.s3a.access.key=#### \
+  -c spark.hadoop.fs.s3a.secret.key=#### \
+  lakefs-spark-client-312-hadoop3-assembly-0.4.0.jar \
+  s3a://storage_ns/_lakefs/retention/gc/addresses/mark_id=1/part-00057-36bb66f0-6a38-47b4-86cb-0e8718dfacfe.c000.snappy.parquet s3://storage_ns s3://external_location/gc_backup s3
+
+Azure -
+
+ */
+object GCBackupAndRestore {
+  lazy val spark = SparkSession.builder().appName("GCBackupAndRestore").getOrCreate()
+
+  /** Validates that args are the following:
+   *  1. address of parquet that includes the relative paths of files to backup or restore, created by a gc run
+   *  2. src: the namespace to backup or restore objects from/to (i.e. repo storage namespace, or an external location compatibly)
+   *  3. backup/restore destination: the namespace to backup or restore objects from/to (i.e. an external location or repo storage namespace compatibly)
+   *  4. Object storage type: "s3" or "azure"
+   */
+  def validateArgs(args: Array[String]): Unit = {
+    if (args.length != 4) {
+      Console.err.println(
+        "Usage: ... <objects list location> <src namespace> <dst namespace> <storage type>"
+      )
+      System.exit(1)
+    }
+
+    val objectsListPath = args(0)
+    if (!objectsListPath.endsWith("parquet")) {
+      Console.err.println(
+        "<objects list path> must be of type parquet"
+      )
+      System.exit(1)
+    }
+
+    val storageType = args(3)
+    if (!storageType.equals(StorageUtils.StorageTypeS3) && !storageType.equals(StorageUtils.StorageTypeAzure)) {
+      Console.err.println(
+        "Invalid <storage type>, must be either \"s3\" or \"azure\""
+      )
+      System.exit(1)
+    }
+  }
+
+  def constructAbsoluteObjectPaths(objectsRelativePathsDF: DataFrame, srcNamespace: String, storageType: String): DataFrame = {
+    var storageNSForHadoopFS = ApiClient
+      .translateURI(URI.create(srcNamespace), storageType)
+      .normalize()
+      .toString
+    if (!storageNSForHadoopFS.endsWith("/")) {
+      storageNSForHadoopFS += "/"
+    }
+    import spark.implicits._
+    val objectsRelativePathsSeq = objectsRelativePathsDF
+      .select("address")
+      .map(_.getString(0))
+      .collect()
+      .toSeq
+
+    StorageUtils.concatKeysToStorageNamespace(objectsRelativePathsSeq, storageNSForHadoopFS, storageType).toDF()
+  }
+
+  def main(args: Array[String]): Unit = {
+    val hc = spark.sparkContext.hadoopConfiguration
+    //TODO: should I broadcast hadoop conf? not sure if distCp does it automatically?
+
+    validateArgs(args)
+    val relativeAddressesLocation = args(0)
+    val srcNamespace = args(1)
+    val dstNamespace = args(2)
+    val storageType = args(3)
+    val relativeAddressesLocationForHadoopFs = ApiClient.translateURI(URI.create(relativeAddressesLocation), storageType).toString
+    val dstNamespaceForHadoopFs = ApiClient.translateURI(URI.create(dstNamespace), storageType).toString
+    print("translated dstNamespace: " + dstNamespaceForHadoopFs + "\n")
+
+    val objectsRelativePathsDF = spark.read.parquet(relativeAddressesLocationForHadoopFs)
+    val objectsAbsolutePathsDF = constructAbsoluteObjectPaths(objectsRelativePathsDF, srcNamespace, storageType)
+    val absoluteAddressesLocation = dstNamespaceForHadoopFs + "/absolute_addresses/"
+    print("absoluteAddressesLocation: " + absoluteAddressesLocation + "\n")
+    // This application uses distCp to copy files. distCp can copy a list of files in a given input text file. therefore,
+    // we write the absolute file paths into a text file rather than a parquet.
+    objectsAbsolutePathsDF
+      .repartition(1)
+      .write
+      .text(absoluteAddressesLocation)
+
+    // Spark writes two files under absoluteAddressesLocation, a _SUCCESS file and the actual txt file that has dynamic
+    // name. iterate the files in the absoluteAddressesLocation and find the path of the txt files that includes the
+    // list of absolute addresses.
+    val fs = FileSystem.get(URI.create(absoluteAddressesLocation), hc)
+    val dirIterator = fs.listFiles(new Path(absoluteAddressesLocation), false)
+    var absoluteAddressesTextFilePath = ""
+    breakable(
+      while (dirIterator.hasNext) {
+        val curFile = dirIterator.next()
+        val curPath = curFile.getPath.toString
+        if (curPath.endsWith("txt")) {
+          absoluteAddressesTextFilePath = curPath
+          break
+        }
+      }
+    )
+    print("txtFilePath: " + absoluteAddressesTextFilePath + "\n")
+
+    // -f option copies the files listed in the file after the -f option
+    DistCp.main(Array("-f", absoluteAddressesTextFilePath, dstNamespaceForHadoopFs))
+  }
+}
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
@@ -1,5 +1,7 @@
 package io.treeverse.clients
 
+import org.apache.hadoop.conf.Configuration
+
 import scala.util.control.Breaks._
 import org.apache.hadoop.tools.DistCp
 import org.apache.hadoop.fs.{FileSystem, Path}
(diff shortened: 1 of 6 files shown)


clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
```diff
@@ -0,0 +1,180 @@
...
+      System.exit(1)
+    }
+
+    val objectsListPath = args(0)
+    if (!objectsListPath.endsWith("parquet")) {
+      Console.err.println(
+        "<objects list path> must be of type parquet"
+      )
+      System.exit(1)
+    }
```
arielshaqed, line 40: Is this the only file suffix ever used for Parquet?  And with no case changes?  Otherwise it's a needless limitation on the command.

If the goal is only to accept files from one run_id then let's specify just the run_id.  Otherwise let's not force the type from the name.

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  def constructAbsoluteObjectPaths(
+      objectsRelativePathsDF: DataFrame,
+      srcNamespace: String,
+      storageType: String
+  ): DataFrame = {
```
arielshaqed, line 58: Can we use DataSets?  Those have types and may let us avoid `.getString(0)` ugliness.

If not, please document the shape of each DataFrame (e.g. "single column of relative paths").

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
```diff
@@ -0,0 +1,180 @@
...
+    var storageNSForHadoopFS = ApiClient
```
arielshaqed, line 59: "Hadoop FS" is almost overloaded to mean both "a FileSystem in org.apache.hadoop" and "HDFS".  I would be happier to call it "FS".

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
```diff
@@ -0,0 +1,180 @@
...
+      .translateURI(URI.create(srcNamespace), storageType)
+      .normalize()
+      .toString
+    if (!storageNSForHadoopFS.endsWith("/")) {
+      storageNSForHadoopFS += "/"
+    }
+    import spark.implicits._
```
arielshaqed, line 66: I think we can have this at the top level, right after l.18.  It's basically "make Spark work".

arielshaqed (changes requested):
Thanks!

Only really required comment is not to `collect()` everything to main memory.

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
```diff
@@ -0,0 +1,180 @@
...
+    var res: Array[String] = Array()
+    for (prop <- hadoopProps) {
+      val formattedProp = "-D" + prop._1 + "=" + prop._2
+      res = res :+ formattedProp
+    }
+    // -f option copies the files listed in the file after the -f option
+    res = res :+ "-f"
+    res = res :+ absoluteAddressesTextFilePath
+    res = res :+ dstNamespaceForHadoopFs
+    res
```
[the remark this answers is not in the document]
talSofer: cool, thx

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  // Construct a DistCp command of the form:
+  // `hadoop distcp -D<per storage credentials> -f absoluteAddressesTextFilePath dstNamespaceForHadoopFs`
+  // example for command format https://docs.lakefs.io/integrations/distcp.html
+  def constructDistCpCommand(
+      hadoopProps: Array[(String, String)],
+      absoluteAddressesTextFilePath: String,
+      dstNamespaceForHadoopFs: String
+  ): Array[String] = {
```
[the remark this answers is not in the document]
talSofer: Is this because the items are ordered?

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
```diff
@@ -0,0 +1,180 @@
...
+      while (dirIterator.hasNext) {
+        val curFile = dirIterator.next()
+        val curPath = curFile.getPath.toString
+        if (curPath.endsWith("txt")) {
+          textFilePath = curPath
+          break
+        }
+      }
+    )
+    textFilePath
```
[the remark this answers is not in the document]
talSofer: while this looks great, i'm not sure how to make Hadoop's RemoteIterator work with map function. let me know if you understand how to do that

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
```diff
@@ -0,0 +1,180 @@
...
+  }
+
+  def main(args: Array[String]): Unit = {
+    validateArgs(args)
```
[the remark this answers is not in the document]
talSofer: I prefer not to because then I need to validate the ars length again and I'm losing the point in the validate args function. I can get rid of it if that makes things more readable

clients/spark/core/src/main/scala/io/treeverse/clients/HadoopUtils.scala
```diff
@@ -0,0 +1,14 @@
+package io.treeverse.clients
+
+import org.apache.hadoop.conf.Configuration
+import scala.collection.JavaConverters._
+
+object HadoopUtils {
+
```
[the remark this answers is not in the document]
talSofer: done

clients/spark/core/src/main/scala/io/treeverse/clients/HadoopUtils.scala
```diff
@@ -0,0 +1,14 @@
...
+  def getHadoopConfigurationValues(hc: Configuration, prefixes: String*) = {
+    hc.iterator.asScala
+      .filter(c => prefixes.exists(c.getKey.startsWith))
+      .map(entry => (entry.getKey, entry.getValue))
+      .toArray
+  }
```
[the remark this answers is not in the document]
talSofer: done

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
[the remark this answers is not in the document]
talSofer: The distCp command required an array of strings so I'm not sure that there is value in changing it

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
```diff
@@ -0,0 +1,180 @@
...
+      Console.err.println(
+        "Usage: ... <objects list location> <src namespace> <dst namespace> <storage type>"
+      )
+      System.exit(1)
+    }
+
+    val storageType = args(3)
+    if (
+      !storageType
+        .equals(StorageUtils.StorageTypeS3) && !storageType.equals(StorageUtils.StorageTypeAzure)
```
arielshaqed, line 38:
```suggestion
      storageType != StorageUtils.StorageTypeS3 && storageType != StorageUtils.StorageTypeAzure)
```

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
```diff
@@ -0,0 +1,180 @@
...
+    ) {
+      Console.err.println(
+        "Invalid <storage type>, must be either \"s3\" or \"azure\""
+      )
+      System.exit(1)
+    }
+  }
```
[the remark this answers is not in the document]
arielshaqed: Not sure I like it.  I deploy a client, a year later Treeverse add GC support for GCS, and now my perfectly good client refuses to run.

If we must have this, I'd be happier to include gcs in there as well... even though it will fail later.

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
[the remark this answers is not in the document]
arielshaqed: Yuck, it's a fake iterator.  A quick search reveals a couple of gists, [this one](https://gist.github.com/vsimko/04fd6d289bac4b7c66a7e870c2ffc718) looks reasonable.

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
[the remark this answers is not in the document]
arielshaqed: OK.

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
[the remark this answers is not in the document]
arielshaqed: If we're anyway creating it directly in a functional style, you're right that we may as well specify the result type we want.
Thanks!

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
```diff
@@ -0,0 +1,180 @@
...
+      .normalize()
+      .toString
+    if (!storageNSForFS.endsWith("/")) {
+      storageNSForFS += "/"
+    }
+
+    objectsRelativePathsDF
+      .select("address")
+      .as[String]
+      .flatMap(x => StorageUtils.concatKeysToStorageNamespace(Seq(x), storageNSForFS, storageType))
```
arielshaqed, line 63: Can we just

```suggestion
      .map(x => StorageUtils.concatKeysToStorageNamespace(x, storageNSForFS, storageType))
```

?

arielshaqed (approved):
Thanks, nice!

How was this tested?  (Probably best to repeat the tests after getting rid of DF.collect...).

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
talSofer: Unfortunately not... StorageUtils.concatKeysToStorageNamespace returns Seq[String]

talSofer:
> How was this tested? (Probably best to repeat the tests after getting rid of DF.collect...). 

@arielshaqed I repeated the tests grid and will now test with more load.

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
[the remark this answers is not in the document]
talSofer: Thanks, I prefer to keep it as is, I think that the wrapper decreases readability in this case.

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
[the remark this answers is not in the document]
talSofer: I will remove it and let it fail in another storage type validation. I don't like the option of adding GCS at this point

clients/spark/core/src/main/scala/io/treeverse/clients/StorageUtils.scala
```diff
@@ -26,10 +40,45 @@ object StorageUtils {
...
+      val addSuffixSlash =
+        if (storageNamespace.endsWith("/")) storageNamespace else storageNamespace.concat("/")
+      val snPrefix =
+        if (addSuffixSlash.startsWith("/")) addSuffixSlash.substring(1) else addSuffixSlash
+
+      if (keys.isEmpty) return Seq.empty
+      keys.map(x => snPrefix.concat(x))
+    }
+
+    def concatKeysToStorageNamespacePrefix(
```
johnnyaug, line 70: If I understand correctly, the difference between this function and the one above is that this one trims the "s3://<bucket name>" prefix. The names are confusing - not sure why it's called "concatKeysToStorageNamespace**Prefix**".

Not sure it even requires a separate function. I would add this ability as an option to the function.

clients/spark/core/src/main/scala/io/treeverse/clients/StorageUtils.scala
```diff
@@ -1,11 +1,24 @@
 package io.treeverse.clients
 
 import java.net.URI
+import java.nio.charset.Charset
 
 object StorageUtils {
   val StorageTypeS3 = "s3"
   val StorageTypeAzure = "azure"
 
+  def concatKeysToStorageNamespace(
```
johnnyaug, line 10: It sounds like this should be an abstract method / interface method - and the other static objects in this file should be implementations.

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
```diff
@@ -0,0 +1,165 @@
...
+import scala.util.control.Breaks._
+import org.apache.hadoop.tools.DistCp
+import org.apache.hadoop.fs.{FileSystem, Path}
+import org.apache.spark.sql.{DataFrame, Dataset, SparkSession}
+
+import java.net.URI
+
+class GCBackupAndRestore {}
+
+object GCBackupAndRestore {
```
johnnyaug, line 14: Documentation?

johnnyaug (commented):
I think this feature is important for users to gain trust in GC. 
However, I see this code as a DistCp wrapper: it takes a list addresses, manipulates it and runs distcp on the result. I would consider a more generic approach not necessarily tying to Spark.
Spark is not really utilized here before the DistCp part. Maybe a script that runs the distcp binary is enough?

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
```diff
@@ -0,0 +1,165 @@
...
+  val S3AccessKeyName = "fs.s3a.access.key"
+  val S3SecretKeyName = "fs.s3a.secret.key"
+  val AzureStorageAccountKeyName = "fs.azure.account.key"
+  lazy val spark = SparkSession.builder().appName("GCBackupAndRestore").getOrCreate()
+  import spark.implicits._
+
+  def constructAbsoluteObjectPaths(
+      objectsRelativePathsDF: DataFrame,
```
johnnyaug, line 22: These are physical addresses, right?
So where do I get this list in the first place? How do I know which files I want to back up?

talSofer: Right, our docs (that I'll add separately) will guide you on what to look for. 
To backup: you will need to find the addresses created by a mark_only gc run you've conducted before the backup. 
To restore: you will need to find the addresses created by a mark_only gc run you've swept the objects it pointed to, this is what you would want to restore.

talSofer:
> Spark is not really utilized here before the DistCp part. Maybe a script that runs the distcp binary is enough?

@johnnyaug Thanks for the feedback! 
I think that Spark is one convenient way to write a distCp wrapper and I feel comfortable with asking users to run a spark job to use this utility for the following reasons: GC is a spark job, meaning, if we can run GC we can run this utility. Also, distCp should run on some cluster so users will need to set up an environment for it. IMO it is more convenient to have it wrapped with a Spark job. WDYT?

clients/spark/core/src/main/scala/io/treeverse/clients/StorageUtils.scala
talSofer: Unfortunately I don't have time to address this comment this week, will open an issue to track it :)

clients/spark/core/src/main/scala/io/treeverse/clients/GCBackupAndRestore.scala
talSofer: Unfortunately I don't have time to address this comment this week, will open an issue to track it :)

PR #4324, opened 2022-10-06
fix username not shown for OIDC users

### LINKED ISSUE #4214: Username is not enriched in users page while using OIDC
Username is not enriched when browsing the users page with OIDC.

image

eladlachmi:
Resolves #4214

### DIFF
--- a/webui/src/pages/auth/users/user/policies.jsx
+++ b/webui/src/pages/auth/users/user/policies.jsx
@@ -1,7 +1,7 @@
 import React from "react";
 
 import {AuthLayout} from "../../../../lib/components/auth/layout";
-import { UserHeaderWithContext } from "./userHeaderWithContext";
+import {UserHeaderWithContext} from "./userHeaderWithContext";
 import {
     ActionGroup,
     ActionsBar,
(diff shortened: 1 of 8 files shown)


webui/src/pages/auth/users/user/policies.jsx
```diff
@@ -1,7 +1,7 @@
 import React from "react";
 
 import {AuthLayout} from "../../../../lib/components/auth/layout";
-import {UserHeader} from "../../../../lib/components/auth/nav";
+import { UserHeaderWithContext } from "./userHeaderWithContext";
```
nopcoder, line 4: Align the format "{ something }" vs "{something}" - maybe we should have something like prettier

nopcoder (changes requested):
Small code and style comments

webui/src/pages/auth/users/user/policies.jsx
eladlachmi: Yes, we should and we will :)

nopcoder (approved):
LGTM!

PR #4325, opened 2022-10-06
Enforce context cancellation errors in kvstore

### LINKED ISSUE #4291: DyanmoDB wraps context cancellations errors
DynamoDB client has its own type of context cancellation error, which masks the actual context error. It makes error handling from the caller side difficult to handle as it requires understanding of the specific implementation of the kv store.

itaiad200:
closes #4291

### DIFF
--- a/pkg/kv/dynamodb/store.go
+++ b/pkg/kv/dynamodb/store.go
@@ -2,10 +2,12 @@ package dynamodb
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/aws/credentials"
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/dynamodb"
@@ -207,7 +209,7 @@ func (s *Store) Get(ctx context.Context, partitionKey, key []byte) (*kv.ValueWit
 	dynamoRequestDuration.WithLabelValues(operation).Observe(time.Since(start).Seconds())
 	if err != nil {
 		dynamoFailures.WithLabelValues(operation).Inc()
-		return nil, fmt.Errorf("get item: %w", err)
+		return nil, fmt.Errorf("get item: %w", handleClientError(err))
 	}
 	if result.ConsumedCapacity != nil {
 		dynamoConsumedCapacity.WithLabelValues(operation).Add(*result.ConsumedCapacity.CapacityUnits)
@@ -283,8 +285,9 @@ func (s *Store) setWithOptionalPredicate(ctx context.Context, partitionKey, key,
 		if _, ok := err.(*dynamodb.ConditionalCheckFailedException); ok && usePredicate {
 			return kv.ErrPredicateFailed
 		}
+
 		dynamoFailures.WithLabelValues(operation).Inc()
-		return fmt.Errorf("put item: %w", err)
+		return fmt.Errorf("put item: %w", handleClientError(err))
 	}
 	if resp.ConsumedCapacity != nil {
 		dynamoConsumedCapacity.WithLabelValues(operation).Add(*resp.ConsumedCapacity.CapacityUnits)
@@ -310,7 +313,7 @@ func (s *Store) Delete(ctx context.Context, partitionKey, key []byte) error {
 	dynamoRequestDuration.WithLabelValues(operation).Observe(time.Since(start).Seconds())
 	if err != nil {
 		dynamoFailures.WithLabelValues(operation).Inc()
-		return fmt.Errorf("delete item: %w", err)
+		return fmt.Errorf("delete item: %w", handleClientError(err))
 	}
 	if resp.ConsumedCapacity != nil {
 		dynamoConsumedCapacity.WithLabelValues(operation).Add(*resp.ConsumedCapacity.CapacityUnits)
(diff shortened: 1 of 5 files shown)


pkg/kv/dynamodb/store.go
```diff
@@ -376,6 +379,14 @@ func (s *Store) scanInternal(ctx context.Context, partitionKey, scanKey []byte,
 	}, nil
 }
```
nopcoder, line 381: alternative - if it is awserr have a struct with this error embed or another type. implement the go standard Unwrap signature that will return the original error.
This will remove the cancel error check here and support other errors we like to check using errors.Is.

pkg/kv/local/iterator.go
```diff
@@ -39,6 +42,10 @@ func (e *EntriesIterator) Next() bool {
 	if e.err != nil {
 		return false
 	}
+	if e.ctx.Err() != nil {
```
nopcoder, line 45: not recommended to check the ctx Err. In this specific case there is a switch case that we can add ctx done check and break/err as needed.

pkg/kv/local/store.go
```diff
@@ -43,6 +43,9 @@ func (s *Store) Get(ctx context.Context, partitionKey, key []byte) (*kv.ValueWit
 
 	var value []byte
 	err := s.db.View(func(txn *badger.Txn) error {
+		if ctx.Err() != nil {
```
nopcoder, line 46: all the operations on badger transaction support passing context ready.

nopcoder (changes requested):
a lot of ctx.Err checks can be removed or moved to a select statement. suggested to wrap the awserr to get the underlying error as any other go error.

pkg/kv/mem/store.go
```diff
@@ -111,6 +119,9 @@ func (s *Store) SetIf(_ context.Context, partitionKey, key, value []byte, valueP
 	}
 	s.mu.Lock()
 	defer s.mu.Unlock()
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
```
itai-david, line 124:
```suggestion
	if ctx.Err() != nil {
		return ctx.Err()
	}
	s.mu.Lock()
	defer s.mu.Unlock()
```

Don't think we should (possibly) wait for the lock just to find out ctx is cancelled.
It will also keep it consistent with the rest of the functions in this file and prevent future quesiotns

pkg/kv/mem/store.go
```diff
@@ -48,16 +49,19 @@ func encodeKey(key []byte) string {
...
+func (s *Store) Get(ctx context.Context, partitionKey, key []byte) (*kv.ValueWithPredicate, error) {
 	if len(partitionKey) == 0 {
 		return nil, kv.ErrMissingPartitionKey
 	}
 	if len(key) == 0 {
 		return nil, kv.ErrMissingKey
 	}
+	if ctx.Err() != nil {
+		return nil, ctx.Err()
+	}
```
itai-david, line 61: Just a thought - probably not for this PR - maybe we should refactopr all these checks to a `verifyXXX` function? It can serve other (possibly future) drivers too. WDYT?

itai-david (approved):
Minor change request in mem-store
Approving as not to block

pkg/kv/local/store.go
itaiad200: Unfortunately it isn't.

pkg/kv/local/iterator.go
itaiad200: It's a switch, not a select. I can move the `e.ctx.Err() != nil` check there but not `<-e.ctx.Done()`

pkg/kv/dynamodb/store.go
itaiad200: `awserr.RequestError` doesn't have `Unwrap`, just `OrigErr`

pkg/kv/mem/store.go
itaiad200: Done.

pkg/kv/dynamodb/store.go
nopcoder: Yes I know - we need to switch to the new Go AWS SDK :/
What I had in mind to make the handle more generic by:

```go
type AWSError struct {
	Err awserr.Error
}

func (e *AWSError) Error() string {
	return e.Err.Error()
}

func (e *AWSError) Code() string {
	return e.Err.Code()
}

func (e *AWSError) Message() string {
	return e.Err.Message()
}

func (e *AWSError) OrigErr() error {
	return e.Err.OrigErr()
}

func (e *AWSError) Unwrap() error {
	return e.Err.OrigErr()
}

func handleClientError(err error) error {
	var awsError awserr.Error
	if errors.As(err, &awsError) {
		return &AWSError{awsError}
	}
	return err
}
```

pkg/kv/local/store.go
nopcoder: Don't think we should handle the context cancellation here - as part of long iteration it is acceptable, but single get I think it will just bloat the code.

itaiad200: I agree - but we have acceptance tests we must pass. I think it's reasonable to expect a cancellation error from the `store` when the context was cancelled before the operation. If we remove it from here we can't expect it from other stores.

pkg/kv/dynamodb/store.go
itaiad200: Do you suggest to wrap the client and convert all returned errors? Otherwise I can't think of a generic way which doesn't go thru wraping every returned error explicitly.

pkg/kv/local/store.go
nopcoder: We can't except context cancellation from all our underlaying implementation - for example mem implementation doesn't use context.
Not all functionallity should check and return ctx error.

nopcoder (commented):
Approved,

nopcoder (approved):
Approved - but I am not sure we need to bloat all code with ctx checks for error - this is not the way.

itaiad200:
> > Approved - but I am not sure we need to bloat all code with ctx checks for error - this is not the way.
> 
> I think you mean that we shouldn't have acceptance tests for context cancellation, which is the heart of this PR. Let's reiterate on that next week. Not merging for now.

After f2f sync - I've relaxed the requirements. If an error is returned for cancellable operation, it must be of type `context.Cancelled`. This will enable the stores to use the context properly and only if needed.

PR #4330, opened 2022-10-07
Fix KV local path using tilde (~) doesn't expand

### LINKED ISSUE #4233: KV local storage default uses tilda part of the db path
Using ~ in the configuration of kv local directory location doesn't expend.
The process create a new directory under the current working one ./~/... and write the files under it.

nopcoder:
Fix #4233


N-o-Z (approved):
LGTM

PR #4333, opened 2022-10-08
GitHub actions skip tests for docs changes

nopcoder:
The idea here is to have a workflow that will report the same steps we marked as required by branch protection, but skipped by ignore.
In our case workflows we don't want to run in case of documentation change -  Esti, test and spark workflows

### DIFF
--- a/.github/workflows/docs-skip.yaml
+++ b/.github/workflows/docs-skip.yaml
@@ -1,0 +1,91 @@
+name: Docs skip checks
+
+# Workflow to skip steps required for branch protection but not required for documentation changes.
+#
+# Add the job there with 'skip-' prefix and the step will just echo it is not required. 
+# The 'name' of the step MUST match the original step we like to skip
+#
+# Example:
+#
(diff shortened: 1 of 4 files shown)


.github/workflows/docs-skip.yaml
```diff
@@ -0,0 +1,91 @@
...
+#
+# We will add a skip step here -
+#
+# skip-test-something:
+#    name: Testing something
+#    steps:
+#      - run: 'echo "Not required"'
+#
+
+on:
```
N-o-Z, line 25: I think I understand this workflow but I didn't quite get why do we need the paths-ignore here?

N-o-Z (commented):
Thanks! One comment

.github/workflows/docs-skip.yaml
nopcoder: @N-o-Z thanks, my bad the above will not work as I don't have a way to have include only on path changes or when other workflow is skipped. Thanks for your time. I'll close this one for now.

PR #4336, opened 2022-10-08
Fix create repository API should return status created

### LINKED ISSUE #4335: CreateRepository API bare option returns 200
Create repository API returns 201 (StatusCreated) on success.
Passing bare option returns 200, it should return 201 as specified in the spec.

nopcoder:
Fix #4335


N-o-Z (approved):
👍🏽

PR #4343, opened 2022-10-09
Fix multiparts KV migration for null content-type

### LINKED ISSUE #4342: Nil content_type in kv migration fails the package migration
[v0.53.0](https://github.com/treeverse/lakeFS/releases/tag/v0.53.0) Introduced new [content-type and metadata](https://github.com/treeverse/lakeFS/issues/2296) to multiparts upload. It made old multiparts upload to have `content-type` nil which is not expected in the server side (should be plain `string`). This bug was long overlooked since no one ever uses stale multiparts uploads. Then came the kv migration which migrated the entire DB schemas over Postgres. It tried to read that old entry and failed. We should skip those old multiparts entries during migration.

```
func="pkg/logging.(*logrusEntryWrapper).Infof" file="build/pkg/logging/logger.go:246" TempDir=/tmp/kv_migrate_4110592530
Failed to setup DB: 1 error occurred:
   * failed migration on package multiparts: scany: scan row into struct fields: can't scan into dest[5]: cannot scan null into *string
```

itaiad200:
fix #4342 

After all changes are approved, I'll create a release from this branch and tag it as 0.80.2

### DIFF
--- a/pkg/gateway/multiparts/db_tracker.go
+++ b/pkg/gateway/multiparts/db_tracker.go
@@ -119,7 +119,8 @@ func Migrate(ctx context.Context, d *pgxpool.Pool, _ blockparams.AdapterConfig, 
 		return err
 	}
 
-	rows, err := d.Query(ctx, "SELECT * FROM gateway_multiparts")
+	// skipping null content_type due to https://github.com/treeverse/lakeFS/issues/4342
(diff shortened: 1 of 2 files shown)


johnnyaug (approved):

N-o-Z (approved):
LGTM!

PR #4347, opened 2022-10-09
Stats flush by size

### LINKED ISSUE #4346: Statistics flush by size
Collecting a large number of unique statistics information can cause a large post that can reach a limit.
Currently we post statistics on internal, we should enable posting the data also based on size to prevent large data post.

nopcoder:
- flush stats by size
- document stats.flush_interval and flush_size

Fix #4346

### DIFF
--- a/docs/reference/configuration.md
+++ b/docs/reference/configuration.md
@@ -168,6 +168,8 @@ This reference uses `.` to denote the nesting of values.
 * `gateways.s3.region` `(string : "us-east-1")` - AWS region we're pretending to be in, it should match the region configuration used in AWS SDK clients
 * `gateways.s3.fallback_url` `(string)` - If specified, requests with a non-existing repository will be forwarded to this URL. This can be useful for using lakeFS side-by-side with S3, with the URL pointing at an [S3Proxy](https://github.com/gaul/s3proxy) instance.
 * `stats.enabled` `(bool : true)` - Whether to periodically collect anonymous usage statistics
+* `stats.flush_interval` `(duration : 30s)` - Interval used to post anonymous statistics collected
+* `stats.flush_size` `(int : 100)` - A size of anonymous statistics collected in which we post
 * `security.audit_check_interval` `(duration : 24h)` - Duration in which we check for security audit.
 * `ui.enable` `(bool: true)` - Whether to server the embedded UI from the binary  
 {: .ref-list }
--- a/pkg/config/defaults.go
+++ b/pkg/config/defaults.go
@@ -71,6 +71,7 @@ const (
 	DefaultStatsEnabled       = true
 	DefaultStatsAddr          = "https://stats.treeverse.io"
 	DefaultStatsFlushInterval = time.Second * 30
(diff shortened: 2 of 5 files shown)


pkg/config/defaults.go
```diff
@@ -71,6 +71,7 @@ const (
 	DefaultStatsEnabled       = true
 	DefaultStatsAddr          = "https://stats.treeverse.io"
 	DefaultStatsFlushInterval = time.Second * 30
+	DefaultStatsFlushSize     = 1000
```
ortz, line 74: shouldn't this be aligned with the documentation? (either setting this one to 100, or aligning the configuration's default value to 1000)

pkg/stats/collector.go
```diff
@@ -251,18 +255,21 @@ func (s *BufferedCollector) Run(ctx context.Context) {
 	go func() {
 		for {
 			select {
-			case w := <-s.writes: // collect events
+			case w := <-s.writes:
+				// collect events, and flush if needed by size
 				s.incr(w)
+				if len(s.cache) >= s.flushSize {
+					s.flush()
+				}
```
ortz, line 263: maybe I'm missing that, but where does the cache cleanup? I couldn't find the reference in the `flush` function

ortz (changes requested):
Nice!
Overall looks good, I do have a couple of questions/requests though

pkg/stats/collector.go
nopcoder: In `flush()` implementation `s.cache = make(keyIndex)` we reset the cache.

ortz: 🤦 sorry, missed that

ortz (approved):

docs/reference/configuration.md
```diff
@@ -168,6 +168,8 @@ This reference uses `.` to denote the nesting of values.
 * `gateways.s3.region` `(string : "us-east-1")` - AWS region we're pretending to be in, it should match the region configuration used in AWS SDK clients
 * `gateways.s3.fallback_url` `(string)` - If specified, requests with a non-existing repository will be forwarded to this URL. This can be useful for using lakeFS side-by-side with S3, with the URL pointing at an [S3Proxy](https://github.com/gaul/s3proxy) instance.
 * `stats.enabled` `(bool : true)` - Whether to periodically collect anonymous usage statistics
+* `stats.flush_interval` `(duration : 30s)` - Interval used to post anonymous statistics collected
+* `stats.flush_size` `(int : 100)` - A size of anonymous statistics collected in which we post
```
N-o-Z, line 172: Size in what? Records / MBs?

nopcoder: Records - each event include a key (composite of the events fields) and count

nopcoder: Added "records" to the doc line

N-o-Z (approved):
Thanks for the update!

nopcoder:
> Thanks for the update!

Thanks for the review!

PR #4353, opened 2022-10-11
Correctly prefix "after" params in lakeFS auth service

arielshaqed:
Contains tests _only_ for users-by-prefix.  Might be useful to test for everything-by-prefix, each call is subtly different.


nopcoder (approved):
LGTM! Thanks Ariel.

arielshaqed:
An API issue with amount==1 prevents pagination (there is also a **bug**
generating has-next in that case, but even if we fixed that the API issue
would still prevent clients from being able to page).  So don't test that.

Opened #4356 to handle that API issue.

arielshaqed:
Thanks, @nopcoder !

Pulling once tests pass...

PR #4357, opened 2022-10-12
lakeFS Hadoop FileSystem: make Hadoop a provided dependency

### LINKED ISSUE #3956: lakeFSFS to Hadoop 3: add and pass spec tests

johnnyaug:
And run all tests on Hadoop 2 and 3.

Jar was tested on Databricks versions:
- 7.3 LTS (Spark 3.0.1)
- 10.4 (Spark 3.2.1)
- 11.2 (Spark 3.3.0)

Resolves #3956.


clients/hadoopfs/pom.xml
```diff
@@ -52,10 +52,26 @@ To export to S3:
 
     <profiles>
         <profile>
-            <id>contract-tests</id>
+            <id>tests-hadoop2</id>
             <properties>
-                <exclude.tests>nothing-to-exclude</exclude.tests>
+                <exclude.tests>io/lakefs/contract/hadoop3/**</exclude.tests>
```
Jonathan-Rosenberg, line 57: hadoop3?

arielshaqed (commented):
Neat!

I'm surprised it works, perhaps some recently removed dependencies helped! Anyway, some things to check:
1. DataBricks, at least lts. The actual dependencies provided there are different and imo proprietary... So hard to predict what will happen (I'm optimistic but paranoid).
2. IIRC we have some code thatbtries to steal an s3 client from s3a, or generate a new one if it cannot. The former is (much) better in terms of configuration. Can we verify that this doesn't change the control flow? (Hadoop 3 s3a might allow stealing the s3 client using a supported method, that would be even better!)

clients/hadoopfs/pom.xml
```diff
@@ -222,23 +240,36 @@ To export to S3:
...
+            <scope>provided</scope>
         </dependency>
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-common</artifactId>
             <scope>test</scope>
             <type>test-jar</type>
-            <version>2.7.7</version>
+            <version>${hadoop.version}</version>
         </dependency>
```
Jonathan-Rosenberg, line 261: I think we can keep only the `provided` one as it can also be used for tests
https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html#Dependency_Scope

Jonathan-Rosenberg (approved):
Looks good apart from @arielshaqed comment

clients/hadoopfs/pom.xml
johnnyaug: This one is needed for the "test-jar" type.

clients/hadoopfs/src/test/java/io/lakefs/contract/TestLakeFSFileSystemContract.java
```diff
@@ -13,23 +15,28 @@
...
+
+  @Before
   public void setUp() throws Exception {
     Configuration conf = new Configuration();
 
     fs = LakeFSTestUtils.createTestFileSystem(conf);
 
-    pathPrefix = conf.get(TEST_FS_LAKEFS_NAME) + "/main";
-    super.setUp();
+    pathPrefix = conf.get(TEST_FS_LAKEFS_NAME) + "/main/";
```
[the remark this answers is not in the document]
johnnyaug: Now using the @Before annotation.

johnnyaug:
@arielshaqed, thanks for the review. Regarding your comments:
1. I have tested the basic functionality (reading and writing a dataframe) in DataBricks. Do you have ideas for a more thorough testing?

2. I'm assuming you're referring to the code in MetadataClient.java, where we try to get the object metadata in various ways. I've tested it and it works as expected*. In any case, I don't think we create an instance of the S3 client in our code.

* With Hadoop <3.2, it uses the first fallback - getting S3 client from the FileSystem. With Hadoop >= 3.3, it actually uses the abstraction of FileSystem to directly get the metadata, which is the first thing we try.

clients/hadoopfs/pom.xml
```diff
@@ -52,10 +52,26 @@ To export to S3:
 
     <profiles>
         <profile>
-            <id>contract-tests</id>
+            <id>tests-hadoop2</id>
```
arielshaqed, line 55: I kinda _liked_ the word "contract" in there.  Unless it's incorrect, can we keep it please?

clients/hadoopfs/src/test/java/io/lakefs/contract/TestLakeFSFileSystemContract.java
```diff
@@ -65,19 +63,18 @@ public void testRenameDirectoryAsExistingDirectory() throws Exception {
...
         fs.exists(path("/test/new/newdir/subdir/file2")));
   }
 
   @Override
   public void testWorkingDirectory() throws Exception {
     // TODO make this test green and remove override
   }
-
-}
+ }
```
arielshaqed, line 80: Nit: trailing newline seems to have been dropped :-(

clients/hadoopfs/src/test/java/io/lakefs/contract/hadoop3/TestLakeFSFileSystemContractHadoop3.java
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    public void testRenameChildDirForbidden() {
+    }
+
+    public void testLSRootDir() throws Throwable {
+    }
+
+    public void testListStatusRootDir() throws Throwable {
+    }
```
arielshaqed, line 36: Do we need to drop all of these tests?  Otherwise I'm not sure we really support Hadoop 3, if behaviour has changed and we cannot support the new required behaviour...?!

clients/hadoopfs/src/test/resources/contract/lakefs.xml
```diff
@@ -106,7 +106,10 @@
         <name>fs.contract.rename-overwrites-dest</name>
         <value>true</value>
     </property>
-
+    <property>
+        <name>fs.contract.create-visibility-delayed</name>
```
arielshaqed, line 110: Maybe worth a comment?  IIUC it says objects are not visible after `out = fs.create()`, only after `out.close()`.  Which makes sense, but is not clear.

arielshaqed (commented):
Neat!

Not sure why we cancel so many of the tests.  If behaviour has changed and these tests should pass in Hadoop 3 but do not, then I am worried :-/

clients/hadoopfs/src/test/java/io/lakefs/contract/hadoop3/TestLakeFSFileSystemContractHadoop3.java
johnnyaug: The behavior wasn't changed, it's just that new tests were added in Hadoop 3. I will open an issue to address these, but after going over them my educated guess is that most of them test behaviors that are not supported by lakeFS. For example, the root directory in lakeFS behaves differently than in S3.

clients/hadoopfs/src/test/resources/contract/lakefs.xml
johnnyaug: Done

clients/hadoopfs/src/test/java/io/lakefs/contract/TestLakeFSFileSystemContract.java
johnnyaug: Fixed

clients/hadoopfs/pom.xml
```diff
@@ -222,23 +240,36 @@ To export to S3:
...
             <scope>test</scope>
             <type>test-jar</type>
-            <version>2.7.7</version>
+            <version>${hadoop.version}</version>
         </dependency>
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-aws</artifactId>
-            <version>2.7.7</version>
+            <version>${hadoop.version}</version>
```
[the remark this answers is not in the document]
arielshaqed: ^^^ this ^^^

clients/hadoopfs/src/main/java/io/lakefs/MetadataClient.java
```diff
@@ -23,7 +22,7 @@ public class MetadataClient {
 
     public MetadataClient(FileSystem fs) {
         if (fs == null) {
-            throw new NullArgumentException("fs");
+            throw new java.lang.IllegalArgumentException();
```
arielshaqed, line 25: This loses information (that `fs == null`).  Can you either revert or add that information in the message?

clients/hadoopfs/src/test/java/io/lakefs/contract/hadoop3/TestLakeFSFileSystemContractHadoop3.java
arielshaqed: OK, so we need to handle all these cases - even if some or all of them are just to say "lakeFS isn't like this", these functions are the right place to say it.  Thanks for opening the issue!

arielshaqed (approved):
THANKS!

Don't forget: we *must* update documentation to say to provide hadoop-aws, and also put it in the CHANGELOG.

PR #4359, opened 2022-10-12
Update and create policy use the user's provided ID

### LINKED ISSUE #4358: Update policy should use the provided policy ID
- UI on request to update policy the policy requested (not the one inside the json body) should be the one we like to update.
- Server should verify on update policy to match the requested ID to update to the information in the request body.
- UI on create policy we should prefer the policy ID we wanted to create (not the one we edit in the json body).

nopcoder:
- update policy api can return not found
- update policy api should match id with policy's id

Fix #4358


N-o-Z (approved):
Thanks!

Issue #4368, opened 2022-10-13
Change using lakeFS to Data Lifecycle Management


adipolak:
Change the title of using lakefs segment into data lifecycle management, with an introduction to data lifecycle management flow.

PR #4369, opened 2022-10-13
Change using lakeFS to Data Lifecycle Management #4368

adipolak:
Renamed using lakefs to data lifecycle management, including referencing link to make sure not to break any existing link. using_lakefs now directs to this section

Change using lakeFS to Data Lifecycle Management #4368

### DIFF
--- a/docs/data_lifecycle_management/ci.md
+++ b/docs/data_lifecycle_management/ci.md
@@ -1,7 +1,7 @@
 ---
 layout: default
 title: During Deployment
-parent: Using lakeFS
+parent: Data Lifecycle Management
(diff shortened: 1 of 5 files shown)


ozkatz (approved):
Thanks!!

johnnyaug:
@adipolak , this change causes strange behavior for old links:
https://docs.lakefs.io/using_lakefs/production.html still works, but due to a bug in our doc publishing process - note that it leads to the **old** version of the docs (where the section title is still "Using lakeFS").
Specifically, we still have links to "/using_lakefs/" pages in some places in the docs.

PR #4371, opened 2022-10-13
lakeFS Hadoop Filesystem - Hadoop 3 packaging support

### LINKED ISSUE #3957: lakeFSFS to Hadoop 3: package
- Discovery: Consider the right structure of the software: different jars by hadoop version (similar to our Spark client), or something else. 
- Implement packaging

Jonathan-Rosenberg:
* Added support for Hadoop 2, 3-3.2.1, and 3.2.2+ packaging.
All package options have a common path (`src/main/java`) from which they package classes, and in addition, a specified location per package is added. For example, the Hadoop 3 profile states that in addition to the common packaged files, all of the files under `src/hadoop3/java` will be packed as well.
To choose which version to package, all you need to do is to add one of the following profiles to the `mvn` command:
  * `hadoop2` - packaging with Hadoop 2 dependencies.
  * `hadoop3` - packaging with Hadoop 3 dependencies.
  * `hadoop321` - packaging with Hadoop 3.2.1 dependencies. 

* Changed the ESTI tests to package Hadoop 2 and use it (as it did so far). **Testing other Hadoop versions using ESTI's workflow should be completed on a different PR**.
* Changed the "Publish Hadoop lakeFS Filesystem" workflow to receive a Hadoop version [2, 3.2.1, 3] and package by it. 

### Workflow
Changed the bucket name to a private bucket, and ran it with the three Hadoop versions.
### Compatibility
Ran the Sonnets app on Databricks using the three Hadoop versions that were published in the previous step (the Hadoop versions on DB were 2.7.4 and 3.3.1 respectively).

---
### P.S.
The need for differentiation between versions 3.0.0-3.2.1 and 3.2.2+ arises from [this](https://issues.apache.org/jira/browse/SPARK-33212) JIRA ticket

Closes #3957


talSofer:
@Jonathan-Rosenberg can you please add a comment that captures why we've closed this issue for now?

Jonathan-Rosenberg:
@talSofer this PR was closed and reopened as #4389

PR #4375, opened 2022-10-16
suggestion changes to chore/parallelize-list-commit-2

nopcoder:


itaiad200:
Resolved conflicts for checks to run

pkg/catalog/catalog.go
```diff
@@ -1043,90 +1048,103 @@ func (c *Catalog) listCommitsWithPaths(ctx context.Context, repository *graveler
...
-					return nil
+		defer close(resultCh)
+		// read and sort by heap the result to results channel
+		current := 0
+		for result := range outCh {
+			heap.Push(&jobsHeap, result)
+			for len(jobsHeap) > 0 && jobsHeap[0].order == current {
+				job := heap.Pop(&jobsHeap).(*commitLogJob)
+				if job.log != nil {
+					resultCh <- job.log
```
itaiad200, line 1115: I think you can hang here forever if the reader (`foundAllCommits`) `break` and stopped reading from the channel.

itaiad200 (changes requested):

pkg/catalog/catalog.go
```diff
@@ -1043,90 +1048,103 @@ func (c *Catalog) listCommitsWithPaths(ctx context.Context, repository *graveler
...
-				_, foundAll := findAllCommits(results, params, createdTasks)
-				if foundAll {
-					drain(done, out)
-					return nil
+		defer close(resultCh)
+		// read and sort by heap the result to results channel
+		current := 0
+		for result := range outCh {
+			heap.Push(&jobsHeap, result)
+			for len(jobsHeap) > 0 && jobsHeap[0].order == current {
```
itaiad200, line 1112: You might get stuck forever here, the `results` channel is not ordered. There's no guarantee that `current` will be at the top of the queue.

pkg/catalog/catalog.go
```diff
@@ -1181,51 +1199,34 @@ func logCommitsResult(commits []*CommitLog, params LogParams) ([]*CommitLog, boo
...
-	for i := 0; i < len(results); i++ {
-		log, ok := results[i]
-		if !ok {
-			// we're missing a result - not done yet
-			return nil, false
-		}
-		if log != nil {
-			commits = append(commits, log)
-		}
+type commitLogJobHeap []*commitLogJob
```
itaiad200, line 1207: Document it's a min heap

pkg/catalog/catalog.go
nopcoder: added drain loop after we process the results

itaiad200 (approved):

PR #4382, opened 2022-10-18
Proposal: Uncommitted GC

N-o-Z:


design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,114 @@
...
+
+Optimized run uses the previous GC run output, to perform a partial scan of the branch to remove uncommitted garbage.
+
+##### Step 1. Analyze Data and Perform Cleanup for old entries (GC client)
+
+1. Run PrepareUncommittedForGC
+2. Read addresses from branch's new commits (all commits up to the last GC run commit id) -> `lakeFS DF`
+3. Read addresses from branch `GC commit` -> `lakeFS DF`
+4. Subtract results `lakeFS DF` from previous run's `GC commit`
+5. The result is a list of files that can be safely removed
```
itaiad200, line 97: Noice

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,114 @@
...
+8. Finally, save the current run's `GC commit` the last read partition and newest commit id
+
+#### Flow 2: Optimized Run
+
+Optimized run uses the previous GC run output, to perform a partial scan of the branch to remove uncommitted garbage.
+
+##### Step 1. Analyze Data and Perform Cleanup for old entries (GC client)
+
+1. Run PrepareUncommittedForGC
+2. Read addresses from branch's new commits (all commits up to the last GC run commit id) -> `lakeFS DF`
```
itaiad200, line 94:
```suggestion
2. Read addresses from branch's new commits (all commits from to the last GC run commit id) -> `lakeFS DF`
```

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,114 @@
...
+3. Read addresses from branch `GC commit` -> `lakeFS DF`
+4. Subtract results `lakeFS DF` from previous run's `GC commit`
+5. The result is a list of files that can be safely removed
+
+>**Note:** This step handles cases of objects that were uncommitted during previous GC run and are now deleted
+
+##### Step 2. Analyze Data and Perform Cleanup for new entries (GC client)
+1. Read all objects on branch path up to the previous run's last read partition (can be done in parallel by 'partition') -> `Branch DF`
```
itaiad200, line 102:
```suggestion
1. Read all objects on branch path up from the previous run's last read partition (can be done in parallel by 'partition') -> `Branch DF`
```

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,114 @@
...
+2. Subtract `lakeFS DF` from `Branch DF`
+3. Filter files newer than < TOKEN_EXPIRY_TIME > and special paths
+4. The remainder is a list of files which can be safely removed
+5. Finally, save the current run's `GC commit` the last read partition and newest commit id
+
+## Limitations
```
itaiad200, line 108: It's also a dedicated flow per branch:
1. How do I discover branches that were deleted?
2. How do I skip old branches that were cleaned up and not used since?
3. Branch discovery (I assume `listBranches` should suffice..?)

itaiad200 (changes requested):
I think you're on to something, great work! See my comments inline.

ozkatz:
Great ideas @N-o-Z! This is a great direction to explore.
I'm wondering:
1. Why are we scoping each "partition" with a branch? Isn't it simpler to simply have all partitions live under the storage namespace directly? 
2. The concept of partitions is really nice! Perhaps we can even start with a simpler implementation that isn't incremental? because partitions are more-or-less bounded in size, you've also solved prior design's limitation with object listing that couldn't be parallelized! do a prefix listing once, then have the executors divvy up these prefixes (with little to no skew since they are all pretty much the same size!) 

still O(n) but at least it doesn't have to execute serially..

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Required changes by lakeFS
+
+The following are necessary changes in lakeFS in order to implement this proposal successfully.
+
+#### Objects Path Conventions
+
+1. Repository objects will be stored under the prefix `<bucket_name>/repos/<repo_uid>`
+2. Branch objects will be stored under the repo prefix with the path `branches/<branch_id>/`
+3. Each lakeFS instance will create a unique prefix partition (serialized) under the branch path to
```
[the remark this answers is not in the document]
N-o-Z: Effectively it will look the same.
Each lakeFS instance will maintain it's own partitions appending a unique suffix to the partition name, in the following manner:
`<sortable_ascending_serialized_uid>_<lakefs_instance_uid>`
Each instance will manage an exclusive partition  - on container restart the lakeFS instance will created a new partition according to the above conventions.
The current idea is to track partition size in memory - this means a restarted instance does not have context on the existing partitions, effectively leaving partitions partially allocated. If you think this is not sufficient we might want to think of an additional solution on how to track the partitions

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,114 @@
...
+The following are necessary changes in lakeFS in order to implement this proposal successfully.
+
+#### Objects Path Conventions
+
+1. Repository objects will be stored under the prefix `<bucket_name>/repos/<repo_uid>`
+2. Branch objects will be stored under the repo prefix with the path `branches/<branch_id>/`
+3. Each lakeFS instance will create a unique prefix partition (serialized) under the branch path to 
+store the branch objects.  
+The serialized partition prefix will allow partial scans of the bucket when running the optimized GC
+4. lakeFS will track the count of objects uploaded to the prefix and create a new one every < TBD > objects uploaded
```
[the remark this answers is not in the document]
N-o-Z: As mentioned above this is an upper bound, which might not be fulfilled in the case of the instance restart / shutdown

design/open/uncomitted-gc.md
N-o-Z: It's actually down to the previous run's last read partition, since we are utilizing the adapter's list property to retrieve the entries sorted alphabetically

design/open/uncomitted-gc.md
N-o-Z: Since log commits returns commits from newest to last, shouldn't it be "down to"?

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,114 @@
...
+The above-mentioned API will enable doing so for all branch objects (committed and uncommitted).
+
+### GC Flows
+
+#### Flow 1: Clean Run
+
+1. Run PrepareUncommittedForGC
+2. Read all addresses from branch commits -> `lakeFS DF`
+3. Read all addresses from branch `GC commit` -> `lakeFS DF`
+4. Read all objects on branch path (can be done in parallel by 'partition') -> `Branch DF`
```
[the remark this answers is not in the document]
N-o-Z: Correct - rephrased

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,114 @@
...
+5. Subtract results `lakeFS DF` from `Branch DF`
+6. Filter files newer than < TOKEN_EXPIRY_TIME > and special paths
+7. The remainder is a list of files which can be safely removed
+8. Finally, save the current run's `GC commit` the last read partition and newest commit id
```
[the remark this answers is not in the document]
N-o-Z: Per branch - updated

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,114 @@
...
+3. "Move/Rename" operation will need to use the new version of CopyObject instead of StageObject.
+
+With the above changes, bucket data is partitioned by repositories, branches and finally a bounded partition.
+This allows running the GC process at branch level, reduces the data scan size and allows parallelization.
+
+#### PrepareUncommittedForGC
+
+A new API which will create meta-ranges and ranges using the given branch uncommitted data. These files
+will be saved in a designated path used by the GC client to list branch's uncommitted objects.  
+For the purpose of this document we'll call this the `GC commit` (feel free to suggest a better term :) )
```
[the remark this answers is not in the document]
N-o-Z: I think using this term might confuse between the uncommitted object themselves and the special metarange+ranges which describe the staging area

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,114 @@
...
+
+Reading data from lakeFS will be similar to the current GC process - using the SSTable reader to quickly list branch objects.
+The above-mentioned API will enable doing so for all branch objects (committed and uncommitted).
+
+### GC Flows
+
+#### Flow 1: Clean Run
+
+1. Run PrepareUncommittedForGC
+2. Read all addresses from branch commits -> `lakeFS DF`
```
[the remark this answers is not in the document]
N-o-Z: That's an interesting idea. I'm not certain we have currently this information without walking over the entire commit tree.
We could if needed maybe add to the branch information the "root" commit om which it was created, but taking into account this is only for the initial GC run - I'm not sure it warrants it. WDYT?

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,114 @@
...
+## Design
+
+### Required changes by lakeFS
+
+The following are necessary changes in lakeFS in order to implement this proposal successfully.
+
+#### Objects Path Conventions
+
+1. Repository objects will be stored under the prefix `<bucket_name>/repos/<repo_uid>`
+2. Branch objects will be stored under the repo prefix with the path `branches/<branch_id>/`
```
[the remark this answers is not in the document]
N-o-Z: Rephrased - let me know if it clarifies it - or that I should add a terminology section

design/open/uncomitted-gc.md
N-o-Z: Answer 3: Using list branches should suffice
Answer 1: Listing the `branches` prefix and comparing with the above should solve the issue

Answer 2: This might be a topic for "phase 2" or optimizations effort. But consider the following assumptions:
1. No new partitions were created on branch
2. No new commits on the branch
3. Uncommitted list equals previous run's uncommitted
It is simply the process of listing the last branch partition and comparing it with the branch's current uncommitted objects which should be a _relatively_ lightweight operation

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,114 @@
...
+The following are necessary changes in lakeFS in order to implement this proposal successfully.
+
+#### Objects Path Conventions
+
+1. Repository objects will be stored under the prefix `<storage_namespace>/repos/<repo_uid>/`
+2. Branch objects will be stored under the repo prefix with the path `branches/<branch_id>/`
+3. Each lakeFS instance will create a unique prefix partition (serialized) under the branch path to 
+store the branch objects.  
+The serialized partition prefix will allow partial scans of the bucket when running the optimized GC
+4. lakeFS will track the count of objects uploaded to the prefix and create a new one every < TBD > objects uploaded
```
itai-david, line 33: Create a new what? Prefix (i.e. unique prefix?)

N-o-Z:
> Great ideas @N-o-Z! This is a great direction to explore. I'm wondering:
> 
> 1. Why are we scoping each "partition" with a branch? Isn't it simpler to simply have all partitions live under the storage namespace directly?
> 2. The concept of partitions is really nice! Perhaps we can even start with a simpler implementation that isn't incremental? because partitions are more-or-less bounded in size, you've also solved prior design's limitation with object listing that couldn't be parallelized! do a prefix listing once, then have the executors divvy up these prefixes (with little to no skew since they are all pretty much the same size!)
> 
> still O(n) but at least it doesn't have to execute serially..

Thank you @ozkatz 

1. scoping the "partitions" to branches allows us to scope our views of the committed and uncommitted objects to the branch level as well. It in fact allows us to run the GC process on the branch level and enables the concurrency. If we created the partitions on the repository scope (namespace), we would have to go over all committed and uncommitted data for the entire namespace in order to ascertain which objects can be deleted.
2. We can definitely start with the clean flow, and add the optimized flow later.
Though we need to take several things into mind:
    1. Even when parallelized, testing S3 listing of a path with ~500,000 objects on a databrick notebook running on the given configuration took around 25 seconds:
image
taking for example a repo with around 1 billion objects (not so far fetched as I understand), even if it is perfectly partitioned into bulks of 500,000 objects, the entire listing might take an unreasonable amount of time
    2. While relying only on the clean flow this becomes is a continually growing problem, both the bucket size and commits are unbounded and eventually we will reach an unmanageable size

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,114 @@
...
+
+Reading data from lakeFS will be similar to the current GC process - using the SSTable reader to quickly list branch objects.
+The above-mentioned API will enable doing so for all branch objects (committed and uncommitted).
+
+### GC Flows
+
+#### Flow 1: Clean Run
+
+1. Run PrepareUncommittedForGC
+2. Read all addresses from branch commits -> `lakeFS DF`
```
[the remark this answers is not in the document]
N-o-Z: Data Frame - since eventually this design is designated to be implemented over a Spark job, I found using this terminology the most sensible

design/open/uncomitted-gc.md
N-o-Z: Rephrased - let me know if it's clearer

design/open/uncomitted-gc.md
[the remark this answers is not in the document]
N-o-Z: Updated document

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,119 @@
...
+#### Reading data from lakeFS
+
+Reading data from lakeFS will be similar to the current GC process - using the SSTable reader to quickly list branch objects.
+The above-mentioned API will enable doing so for all branch objects (committed and uncommitted).
+
+### GC Flows
+
+The following describe the GC process run flows in the branch scope:
+
+#### Flow 1: Clean Run
```
itai-david, line 81: If I understand correctly, it`s a part of our current GC, right? Deleted-uncommited and expired-committed will be deleted in the same run?

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,119 @@
...
+1. Copy object in the same branch will work the same - creating a new staging entry using the existing entry information.
+2. For objects that are not part of the branch, use the underlying adapter copy operation.
+3. "Move/Rename" operation will need to use the new version of CopyObject instead of StageObject.
+
+With the above changes, bucket data is partitioned by repositories, branches and finally a bounded partition.
+This allows running the GC process at branch level, reduces the data scan size and allows parallelization.
+
+#### PrepareUncommittedForGC
+
+A new API which will create meta-ranges and ranges using the given branch uncommitted data. These files
```
itai-david, line 68: So, this is kinda "fake" commit with all uncomitted-undeleted objects, correct?
Pretty cool

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,119 @@
...
+4. The remainder is a list of files which can be safely removed
+5. Finally, save the current run's `GC commit` the last read partition and newest commit id in a designated location on the branch path
+
+## Limitations
+
+* Since this solution relies on the new repository structure, it is not backwards compatible. Therefore, another solution will be required for existing 
+repositories
+* This solution partitions the branch data into segments that allow parallel listing of branch objects. In cases where
+the branch path is extremely large, it might still take a lot of time to perform a clean run.
+* For GC to work optimally, it must be executed on a timely basis. Long periods between runs might result in excessively long runs.
```
itai-david, line 119: Can we improve it by limiting GC cycles (time limit or number of objects to handle)?

design/open/uncomitted-gc.md
itai-david: It is. Thanks

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,114 @@
...
+#### PrepareUncommittedForGC
+
+A new API which will create meta-ranges and ranges using the given branch uncommitted data. These files
+will be saved in a designated path used by the GC client to list branch's uncommitted objects.  
+For the purpose of this document we'll call this the `GC commit` (feel free to suggest a better term :) )
+
+#### Reading data from lakeFS
+
+Reading data from lakeFS will be similar to the current GC process - using the SSTable reader to quickly list branch objects.
+The above-mentioned API will enable doing so for all branch objects (committed and uncommitted).
```
arielshaqed, line 75: What are the advantages of doing this over directly providing an API to list the branch objects?  If the goal is "merely" to get a static listing, could we instead provide an API to push a new staging token and return the previous list of staging tokens?  Then the GC Spark job could use another (new) API call list all objects on that list of staging tokens.

Otherwise we end up with a strange unused metarange. 🤷🏽

design/open/uncomitted-gc.md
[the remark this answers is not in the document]
arielshaqed: I don't think that that works: these are all addresses to keep, they might occur only on another branch (e.g. if copied from another branch by staging them to this one).

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,114 @@
...
+### GC Flows
+
+#### Flow 1: Clean Run
+
+1. Run PrepareUncommittedForGC
+2. Read all addresses from branch commits -> `lakeFS DF`
+3. Read all addresses from branch `GC commit` -> `lakeFS DF`
+4. Read all objects on branch path (can be done in parallel by 'partition') -> `Branch DF`
+5. Subtract results `lakeFS DF` from `Branch DF`
+6. Filter files newer than < TOKEN_EXPIRY_TIME > and special paths
```
arielshaqed, line 88:
```suggestion
6. Remove files newer than < TOKEN_EXPIRY_TIME > and special paths
```
("Filter" tends to mean both "keep" and "remove", and usually "keep", for strange reasons...)

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,114 @@
...
+##### Step 2. Analyze Data and Perform Cleanup for new entries (GC client)
+1. Read all objects on branch path up to the previous run's last read partition (can be done in parallel by 'partition') -> `Branch DF`
+2. Subtract `lakeFS DF` from `Branch DF`
+3. Filter files newer than < TOKEN_EXPIRY_TIME > and special paths
+4. The remainder is a list of files which can be safely removed
+5. Finally, save the current run's `GC commit` the last read partition and newest commit id
+
+## Limitations
+
+* Since this solution relies on the new repository structure, it is not backwards compatible. Therefore, another solution will be required for existing
```
arielshaqed, line 115: Can we have a slower (one-time) run to cleanup old repositories, once?

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,114 @@
...
+8. Finally, save the current run's `GC commit` the last read partition and newest commit id
+
+#### Flow 2: Optimized Run
+
+Optimized run uses the previous GC run output, to perform a partial scan of the branch to remove uncommitted garbage.
+
+##### Step 1. Analyze Data and Perform Cleanup for old entries (GC client)
+
+1. Run PrepareUncommittedForGC
+2. Read addresses from branch's new commits (all commits up to the last GC run commit id) -> `lakeFS DF`
```
arielshaqed: Where do we store this "run ID"?

design/open/uncomitted-gc.md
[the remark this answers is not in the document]
arielshaqed: It still requires serializing all uploads.

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Required changes by lakeFS
+
+The following are necessary changes in lakeFS in order to implement this proposal successfully.
+
+#### Objects Path Conventions
+
+1. Repository objects will be stored under the prefix `<storage_namespace>/repos/<repo_uid>/`
+2. Branch objects will be stored under the repo prefix with the path `branches/<branch_id>/`
+3. Each lakeFS instance will create a unique prefix partition (serialized) under the branch path to
```
arielshaqed: Not sure what these partitions are: Staging tokens?  Are different instances allowed to scan and use other instances' partitions?

Can you add some motivation for this?  It raises some questions.

* How "ascending" does this need to be?  Would a timestamp, which is not really ascending but generally satisfies this, be good enough?
* How "unique" does this need to be?

arielshaqed (commented):
Thanks!  It looks really exciting, but I didn't entirely understand what "partitions" are.

design/open/uncomitted-gc.md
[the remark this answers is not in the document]
johnnyaug: It's not a commit, and it's also not parallel to a commit because it contains only a subset of the objects. I'd go with Itai's suggestion.

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,119 @@
...
+4. The lakeFS instance will track the count of objects uploaded to the prefix and create a new partition every < TBD > objects uploaded
+
+The full object path for an object `file1` uploaded to repository `my-repo` on branch `test-branch` will be as follows:  
+    
+    <lakeFS_bucket>/repos/my-repo/branches/test-branch/<lakeFS_instance_partition_uid>/<file1_uid>
+
+#### StageObject
+
+1. Allowed only for address outside the repo namespace
+2. Prevent race between staging object and GC job
```
johnnyaug, line 45: What kind of race? Is this the token thing?

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,119 @@
...
+### GC Flows
+
+The following describe the GC process run flows in the branch scope:
+
+#### Flow 1: Clean Run
+
+1. Run PrepareUncommittedForGC
+2. Read all addresses from branch commits -> `lakeFS DF`
+3. Read all addresses from branch `GC commit` -> `lakeFS DF`
+4. Read all objects on branch path directly from object store (can be done in parallel by 'partition') -> `Branch DF`
```
johnnyaug, line 86: Is this a listing on the object store? In what sense is it parallel?

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,119 @@
...
+
+#### Objects Path Conventions
+
+1. Repository objects will be stored under the prefix `<storage_namespace>/repos/<repo_uid>/`
+2. For each repository branch, objects will be stored under the repo prefix with the path `branches/<branch_id>/`
+3. Each lakeFS instance will create a unique prefix partition under the branch path. This will be used by the specific
+lakeFS instance to store the branch's objects. This prefix will be composed of two parts:
+   1. Lexicographically sortable, descending time based serialization
+   2. A unique identifier for the lakeFS instance
+      `<sortable_descending_serialized_uid>_<lakefs_instance_uid>`
```
johnnyaug, line 34: Why is the instance ID a part of the partition?

design/open/uncomitted-gc.md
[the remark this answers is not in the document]
N-o-Z: @arielshaqed I'm not really clear about what you were referring to when you said "I don't think that that works".
Did you mean the presented flow, or my reply to Itai?

design/open/uncomitted-gc.md
N-o-Z: I believe it should be incorporated into the same ecosystem. Whether it is the same job, or a different one is to be considered. I'm not sure deciding this now is a requirement for this proposal

design/open/uncomitted-gc.md
N-o-Z: Credit to @itaiad200  😄

design/open/uncomitted-gc.md
N-o-Z: This will be a bit difficult IMO. Since object creation date is loosely coupled with commit date it might be very complicated to achieve

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,119 @@
...
+2. For objects that are not part of the branch, use the underlying adapter copy operation.
+3. "Move/Rename" operation will need to use the new version of CopyObject instead of StageObject.
+
+With the above changes, bucket data is partitioned by repositories, branches and finally a bounded partition.
+This allows running the GC process at branch level, reduces the data scan size and allows parallelization.
+
+#### PrepareUncommittedForGC
+
+A new API which will create meta-ranges and ranges using the given branch uncommitted data. These files
+will be saved in a designated path used by the GC client to list branch's uncommitted objects.
```
nopcoder, line 69: - Prefer to have a new folder for each request in order to have history, support parallel request, prevent data delete or keeping old files around. And/or we should keep a state of the current request, progress, completion and errors.
- A risk here will be that a single instance is responsible to extract all the staging information. The request can take a lot of time, it is handled by single instance that can go down and we will need to start from scratch.
- From the work Yoni did, we should consider write this data in a parquet format.

design/open/uncomitted-gc.md
N-o-Z: I think we won't have any option but to do that :). I'm just not sure this should be a part of this proposal. WDYT?

design/open/uncomitted-gc.md
N-o-Z: Added:

> 8. Finally, save the current run's `GC commit` the last read partition and newest commit id in a designated location on the branch path

design/open/uncomitted-gc.md
nopcoder, line 68: Just double-checking because the design talked about branch level - this is one set of files per repository?

design/open/uncomitted-gc.md
N-o-Z: By not allowing StageObject on paths inside the repo namespace, we prevent a race where an un-referenced address is being staged using StageObject while it is being deleted by GC

design/open/uncomitted-gc.md
N-o-Z: Since the objects on a given branch a divided into partition prefixes we can use workers to list objects by partition and aggregate the result

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,119 @@
...
+## Design
+
+### Required changes by lakeFS
+
+The following are necessary changes in lakeFS in order to implement this proposal successfully.
+
+#### Objects Path Conventions
+
+1. Repository objects will be stored under the prefix `<storage_namespace>/repos/<repo_uid>/`
+2. For each repository branch, objects will be stored under the repo prefix with the path `branches/<branch_id>/`
```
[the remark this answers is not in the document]
N-o-Z: This allows us to scope the GC process to the commits and staging area that are related to the branch. Otherwise, we will have to read all the information from all commits and all staging areas to understand what we need to delete.

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Design
+
+### Required changes by lakeFS
+
+The following are necessary changes in lakeFS in order to implement this proposal successfully.
+
+#### Objects Path Conventions
+
+1. Repository objects will be stored under the prefix `<storage_namespace>/repos/<repo_uid>/`
```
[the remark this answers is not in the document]
N-o-Z: You are right - but we need to take into consideration scenarios where a repository is deleted and created again with the same name. Although it has the same name, the unique identifier is different and we need to take that into consideration.
Perhaps you have a better suggestion on how to do this?

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,119 @@
...
+>**Note:** This step handles cases of objects that were uncommitted during previous GC run and are now deleted
+
+##### Step 2. Analyze Data and Perform Cleanup for new entries (GC client)
+1. Read all objects on branch path down to the previous run's last read partition (can be done in parallel by 'partition') -> `Branch DF`
+2. Subtract `lakeFS DF` from `Branch DF`
+3. Filter files newer than < TOKEN_EXPIRY_TIME > and special paths
+4. The remainder is a list of files which can be safely removed
+5. Finally, save the current run's `GC commit` the last read partition and newest commit id in a designated location on the branch path
+
+## Limitations
```
nopcoder, line 113: I think that the solution should include - how we delete uncommitted data at the 3 levels:
1. delete repository
1. delete branch
1. delete object

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,119 @@
...
+## Design
+
+### Required changes by lakeFS
+
+The following are necessary changes in lakeFS in order to implement this proposal successfully.
+
+#### Objects Path Conventions
+
+1. Repository objects will be stored under the prefix `<storage_namespace>/repos/<repo_uid>/`
+2. For each repository branch, objects will be stored under the repo prefix with the path `branches/<branch_id>/`
```
nopcoder, line 29: We should address create/delete/create of the same branch by using unique id also in the branch level.

nopcoder (changes requested):
Comments in the body - think this is the way, just need to extend the description to how we handle delete of uncommitted data also above the branch level. (branch delete and repository delete).

design/open/uncomitted-gc.md
johnnyaug, line 45:
```suggestion
The StageObject operation will only be allowed on addresses outside the repository's storage namespace. This way, objects added using this operation are never collected.
```

N-o-Z:
Thank you everyone for the most invaluable input! 🙏🏽 
We've decided to try and modify this proposal to remove the use of branch prefixes and use a time based partition only.
Please hold off any new comments until I'm able to address the current ones and modify to proposal.
I'll update once I've made all the required changes - Thanks!

talSofer (commented):
@N-o-Z thanks for this great proposal! 
I'm posting my comments since they are unrelated to the changes you are planning to the document.

design/open/uncomitted-gc.md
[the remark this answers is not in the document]
itaiad200: Even today, you can't* create a repository in a storage namespace that serves as a storage namespace for some other repo. So it doesn't matter if the repo is deleted, created by an installation that is no longer active, etc, you can't recreate the repo in the same storage namespace unless it was cleaned.

* Repo creation checks for the dummy object in the root of the namespace. You can argue that it's not safe enough, but I think we can rely on it for the time being.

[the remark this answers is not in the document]
johnnyaug: I didn't understand what we are trying to mitigate here.

N-o-Z:
Please see new revision of proposal

- Removed the branch partitioning and made the relevant adjustments
- Addressed the PR comments that were relevant after the change

Please give your feedback and let me know if I missed any previous comment

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,162 @@
...
+
+### 1. Listing namespace objects
+
+For large repositories, object listing is a very time-consuming operation - therefore we need to find a way to optimize it.
+The suggested method is to split the repository structure into fixed size (upper bounded) partitions.
+These partitions can then be scanned independently using multiple workers.
+In addition, taking advantage of the common property of the listing operation, which lists objects in a lexicographical order, we can create the partitions in a manner which
+enables additional optimizations on the GC process (read further for details).
+
+![Repository Structure](diagrams/uncommitted-gc-repo-struct.png)
```
johnnyaug, line 35: Awesome explanation! Thank you

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,162 @@
...
+For the purpose of this document we'll call this the `BranchUncommittedMetarange`
+
+### GC Flows
+
+The following describe the GC process run flows on a repository:
+
+#### Flow 1: Clean Run
+
+1. Listing namespace objects
+   1. Read all objects directly from object store (can be done in parallel using the partitions) -> `Store DF`
```
arielshaqed, line 100: This is a really cool idea.  But we should detail a bit more how we parallelize this: since storage partitions are time-based, it will be quite hard to know how many parallel tasks to generate, and also quite hard to ensure that each parallel task is large enough.  The issue is that I do not know how many objects are created every minute, so I have no idea how to select start points for each task.

One way to do it might be to use _only_ size-based storage partitions, and then add an API call that gives the estimated size of each storage partition and the first and last storage partitions.  That makes parallelizing the read in the Spark client easier.

johnnyaug, line 100:
```suggestion
   1. List all objects directly from object store (can be done in parallel using the partitions) -> `Store DF`
```

design/open/uncomitted-gc.md
arielshaqed: In a separate proposal is great, just let's have it :-)

design/open/uncomitted-gc.md
arielshaqed: This is an ops issue.  I might _prefer_ occasional long runs, or common short runs.

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,162 @@
...
+   3. Filter files in special paths
+   4. The remainder is a list of files which can be safely removed
+3. Save run information under the GC path  
+   See previous for steps
+
+## Limitations
+
+* Since this solution relies on the new repository structure, it is not backwards compatible. Therefore, another solution will be required for existing 
+repositories
+* For GC to work optimally, it must be executed on a timely basis. Long periods between runs might result in excessively long runs.
```
arielshaqed, line 162: Can we add some minimal required performance?  It will help both our implementation and users' sizing.

arielshaqed (commented):
Neat!

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,162 @@
...
+2. For objects that are not part of the branch, use the underlying adapter copy operation.
+
+#### Move/RenameObject
+Clients working through the S3 Gateway can use the CopyObject + DeleteObject to perform a Rename or Move operation.
+For clients using the OpenAPI this could have been done using StageObject + DeleteObject.
+To continue support of this operation, introduce a new API to rename an object which will be scoped to a single branch.
+
+#### PrepareUncommittedForGC
+
+A new API which will create meta-ranges and ranges for a given branch using its uncommitted data. These files
```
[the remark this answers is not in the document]
N-o-Z: I'm really not sure which is more efficient.
When performing this operation per branch, we enable distribution of the workload to multiple lakeFS instances.
If we perform this globally, we lose this optimization. WDYT?

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,162 @@
...
+### Required changes by lakeFS
+
+The following are necessary changes in lakeFS in order to implement this proposal successfully.
+
+#### Objects Path Conventions
+
+Uncommitted GC must scan the bucket in order to find objects that are not referenced by lakeFS.
+To optimize this process, suggest the following changes:
+
+1. Divide the repository namespace into time-and-size based partitions.
```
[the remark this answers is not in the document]
N-o-Z: Initially I wanted to use "buckets" 😄 partitions was the next best thing (similar to DB partitioning). I guess we can change it to slices or prefixes if it's really that confusing

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,162 @@
...
+##### Step 1. Analyze Data and Perform Cleanup for old entries (GC client)
+
+1. Read previous run's information
+   1. Previous `Uncommitted DF`
+   2. Last read partition
+   3. Last run's timestamp
+2. Listing of lakeFS repository uncommitted objects  
+    See previous for steps
+3. Listing of lakeFS repository committed objects (optimized)
+   1. Read addresses from branch's new commits (all new commits down to the last GC run timestamp) -> `Committed DF`
```
johnnyaug, line 136: It can be problematic to rely on the timestamp of the commit, because:
1. There can be clock differences.
2. There is a plumbing command that allows to manually set the commit date.

If we go down that road, you should take a margin of safety, and include commits from some time before the last GC run - to avoid race conditions.

johnnyaug (commented):
Looks great!

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,162 @@
...
+      1. Read all addresses from Repository commits -> `Committed DF`
+4. Find candidates for deletion
+   1. Subtract committed data from all objects (`Store DF` - `Committed DF`)
+   2. Subtract uncommitted data from all objects (`Store DF` - `Uncommitted DF`)
+   3. Filter files in special paths
+   4. The remainder is a list of files which can be safely removed
+5. Save run information under the GC path
+    1. The current run's `Uncommitted DF` (as a parquet file)
+    2. The last read partition
+    3. The GC run start timestamp
```
itaiad200, line 121: I think we need to structure the new objects in the GC prefix in the storage for alignment. Can be a followup task too.

itaiad200 (approved):
This is a great design! I think you tackled all the blockers and we can continue with most things in the execution plan (i.e. followup discoveries)

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,162 @@
...
+   1. Subtract `Committed DF` from `Store DF`
+   2. Subtract current `Uncommitted DF` from `Store DF`
+   3. Filter files in special paths
+   4. The remainder is a list of files which can be safely removed
+3. Save run information under the GC path  
+   See previous for steps
+
+## Limitations
+
+* Since this solution relies on the new repository structure, it is not backwards compatible. Therefore, another solution will be required for existing
```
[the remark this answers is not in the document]
N-o-Z: Special handling in the sense that the read from old lakeFS installation cannot use the partition optimization when performing a full scan and will require reading from the root.
Basically after the upgrade we'll need to perform the following for a full scan:
1. Scan by partitions - for new data created after upgrade
2. Scan root objects - for data created before upgrade

Current solution doesn't address reading objects directly from root - but it should be relatively easy to manage

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,162 @@
...
+
+1. GetPhysicalAddress to return a validation token along with the address (or embedded as part of the address).
+2. The token will be valid for a specified amount of time and for a single use.
+3. lakeFS will need to track issued tokens/addresses, and delete them when tokens are expired/used
+4. LinkPhysicalAddress to verify token valid before creating an entry.
+    1. Doing so will allow us to use this time interval to filter objects that might have been uploaded and waiting for
+       the link API and avoid them being deleted by the GC process.
+    2. Objects that were uploaded to a physical address issued by the API and were not linked before the token expired will
+       eventually be deleted by the GC job.
+>**Note:** These changes also solves an existing bug where an issued physical address which was committed can be overwritten by Link
```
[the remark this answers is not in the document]
N-o-Z: Done

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,162 @@
...
+   1. Read all objects directly from object store (can be done in parallel using the partitions) -> `Store DF`
+   2. Skip partitions that are newer than < TOKEN_EXPIRY_TIME >
+2. Listing of lakeFS repository uncommitted objects
+   1. Mark uncommitted data
+      1. List branches
+      2. Run _PrepareUncommittedForGC_ on all branches
+   2. Get all uncommitted data addresses
+      1. For each branch
+         1. Read all addresses from branch `BranchUncommittedMetarange` -> `Uncommitted DF`
+    >**Note:** To avoid possible bug, uncommitted listing must happen before listing of committed data
```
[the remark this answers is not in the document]
N-o-Z: Fixed

design/open/uncomitted-gc.md
N-o-Z: We're already taking a margin on the GetPhysicalAddress token expiration time - for certainty we can say that GC will not process partitions created in the last 24 hours (though I prefer not stating specific numbers in the proposal). WDYT?

Regarding the plumbing command - can you tell me what it is used for?

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,162 @@
...
+
+#### S3 Gateway CopyObject
+
+1. Copy object in the same branch will work the same - creating a new staging entry using the existing entry information.
+2. For objects that are not part of the branch, use the underlying adapter copy operation.
+
+#### Move/RenameObject
+Clients working through the S3 Gateway can use the CopyObject + DeleteObject to perform a Rename or Move operation.
+For clients using the OpenAPI this could have been done using StageObject + DeleteObject.
+To continue support of this operation, introduce a new API to rename an object which will be scoped to a single branch.
```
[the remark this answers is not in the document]
N-o-Z: It is still possible to use CopyObject + DeleteObject, it will be also possible to use RenameObject.
It will no longer be possible to use StageObject as done in lakeFSFS (renameObject).

In any case you are correct (and I also did not state :) ) that the RenameObject API is atomic

design/open/uncomitted-gc.md
N-o-Z: Rephrased

design/open/uncomitted-gc.md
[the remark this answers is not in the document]
itaiad200: I don't understand how can the kv interface support an atomic rename.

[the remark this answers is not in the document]
N-o-Z: I'll rephrase - since the parenthesis were confusing.
**RenameObject API is not atomic**

design/open/uncomitted-gc.md
N-o-Z: Added a discovery task:
https://github.com/treeverse/lakeFS/pull/4382

design/open/uncomitted-gc.md
N-o-Z: It will be very difficult to align the data we're supposed to read from the namespace, with the commits we read without making the slices time-based. We added a upper bound to the slice size so that we can efficiently divide the work.

design/open/uncomitted-gc.md
```diff
@@ -161,4 +161,13 @@ Optimized run uses the previous GC run output, to perform a partial scan of the
 * Since this solution relies on the new repository structure, it is not backwards compatible. Therefore, another solution will be required for existing 
 repositories
 * Even with the given optimizations, the GC process is still very much dependent on the amount of changes that were made on the repository
-since the last GC run.
\ No newline at end of file
+since the last GC run.
+
+## Performance Requirements
```
itaiad200, line 166: I'm missing the actual requirement below. It is sort of an estimation of a single task of the uncommitted GC. How long will the entire GC run? What's the repo status (like number of commits, branches, objects to delete, etc.)

nopcoder (approved):

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,173 @@
...
+## Design
+
+Garbage collection of uncommitted data will be performed using the same principles of the current GC process.
+The basis for this is a GC Client (i.e. _Spark_ job) consuming objects information from both lakeFS and directly from the underlying object storage,
+and using this information to determine which objects be deleted from the namespace.
+
+The GC process is composed of 3 main parts:
+1. Listing namespace objects
+2. Listing of lakeFS repository committed objects
+3. Listing of lakeFS repository uncommitted objects
```
Jonathan-Rosenberg, line 23: An object that's in case 3 means that there's an active branch that holds the uncommitted object, or is it any uncommitted object under that repository ("dangling" or not)?

design/open/uncomitted-gc.md
N-o-Z: Done

design/open/uncomitted-gc.md
N-o-Z: Done

design/open/uncomitted-gc.md
```diff
@@ -0,0 +1,178 @@
...
+since the last GC run.
+
+## Performance Requirements
+
+The heaviest operation during the GC process, is the namespace listing. And while we added the above optimizations to mitigate
+this process, the fact remains - we still need to scan the entire namespace (in the Clean Run mode).
+Performing tests against an AWS S3 bucket using a Databricks notebook on a m4.large cluster, we've observed listing of 
+~500,000 objects takes approximately 25 seconds.
+We can estimate that on a repository with ~1B objects, using ~500K object size slices, and using 10 workers - listing will take
+around 1.5 hours.
```
itai-david, line 180: How do we handle failures in such a long run - say, a crash after an hour of execution?
Does the entire process need to be rerun?
Maybe we can parallelize the GC itself, instead of just the listing?

itai-david (approved):
Thanks @N-o-Z 
Added a question regarding the impact of an error in such a long process, but other than that LGTM.

design/open/uncomitted-gc.md
```diff
@@ -153,8 +150,16 @@ Optimized run uses the previous GC run output, to perform a partial scan of the
...
    4. The remainder is a list of files which can be safely removed
-3. Save run information under the GC path  
-   See previous for steps
+3. Save run data (see: [GC Saved Information](#gc-saved-information))
+
+### GC Saved Information
+
+For each GC run, save the following information using the GC run id as detailed in this [proposal](https://github.com/treeverse/cloud-controlplane/blob/main/design/accepted/gc-with-run-id.md):
+1. Save metaranges in `_lakefs/gc/run_id/metadata/`
+2. Save `Uncommitted DF` in `_lakefs/gc/run_id/uncommitted.parquet`
```
itaiad200, line 159: Correct me if I'm wrong @treeverse/ecosystem, but parquet outputs are multiple files (per column) and it's normally saved under a prefix of it's own for better separation, right?
So a better path would be `_lakefs/gc/run_id/uncommitted/entries.parquet`

design/open/uncomitted-gc.md
N-o-Z: GC job will be parallelized, but we can't rely on failed runs. To get a correct picture of the namespace the process must run successfully and without errors.
Relying on partial data, and failed jobs will increase risk of deleting committed data - which is something that absolutely must not happen

N-o-Z:
Moved proposal to "accepted" folder. 
Please take the time to give final notes - before I merge this PR

design/open/uncomitted-gc.md
johnnyaug: The plumbing command is used today for testing Garbage Collection.

design/accepted/gc_plus/uncomitted-gc.md
```diff
@@ -0,0 +1,201 @@
...
+5. The sorted slices will enable partial scans of the bucket when running the optimized GC.
+
+#### StageObject
+
+The StageObject operation will only be allowed on addresses outside the repository's storage namespace. This way, objects added using this operation are never collected by GC.
+
+#### [Get/Link]PhysicalAddress 
+
+1. GetPhysicalAddress to return a validation token along with the address (or embedded as part of the address).
+2. The token will be valid for a specified amount of time and for a single use.
```
johnnyaug, line 69: Why does it matter that the token is valid for a single use?

N-o-Z: See: https://github.com/treeverse/lakeFS/issues/4438

design/open/uncomitted-gc.md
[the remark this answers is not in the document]
itaiad200: If it's not atomic, can we add and use copy instead?

design/accepted/gc_plus/uncomitted-gc.md
```diff
@@ -0,0 +1,203 @@
...
+#### Move/RenameObject
+Clients working through the S3 Gateway can use the CopyObject + DeleteObject to perform a Rename or Move operation.
+For clients using the OpenAPI this could have been done using StageObject + DeleteObject.
+To continue support of this operation, introduce a new API to rename an object which will be scoped to a single branch.
+
+#### PrepareUncommittedForGC
+
+A new API which will create meta-ranges and ranges for a given branch using its uncommitted data. These files
+will be saved in a designated path used by the GC client to list branch's uncommitted objects.  
+For the purpose of this document we'll call this the `BranchUncommittedMetarange`
```
itaiad200, line 94: I think there's a bug.
I start preparing the uncommitted objects. During that time a Copy is happening on the same branch, making it a metadata operation. The copy copies from the path `zz` to the path `aa`. The preparation of the uncommitted started before `aa` was created, but by the time it reaches `zz` it was already deleted. Now I don't see neither of them in the output file and the physical object may be deleted by the GC.

itaiad200 (changes requested):
Found a bug :(

design/accepted/gc_plus/uncomitted-gc.md
N-o-Z: Thanks! Addressed in proposal

design/accepted/gc_plus/uncomitted-gc.md
```diff
@@ -79,19 +79,26 @@ objects added using this operation are never collected by GC.
 
 #### S3 Gateway CopyObject
 
-1. Copy object in the same branch will work the same - creating a new staging entry using the existing entry information.
-2. For objects that are not part of the branch, use the underlying adapter copy operation.
+When performing a shallow copy - track copied objects in ref-store.
+GC will read the copied objects information from the ref-store, and will add them to the list of uncommitted.
```
itaiad200, line 83: The timing here is critical - it MUST read the new table after it iterated on the staging area of all branches.

design/accepted/gc_plus/uncomitted-gc.md
```diff
@@ -79,19 +79,26 @@ objects added using this operation are never collected by GC.
...
+lakeFS will periodically clear the copied list according to timestamp.
+
+1. Copy of staged object in the same branch will perform a shallow copy as described above
+2. All other copy operations will use the underlying adapter copy operation.
 
 #### Move/RenameObject
 Clients working through the S3 Gateway can use the CopyObject + DeleteObject to perform a Rename or Move operation.
 For clients using the OpenAPI this could have been done using StageObject + DeleteObject.
 To continue support of this operation, introduce a new API to rename an object which will be scoped to a single branch.
+Rename will add copied objects in ref-store similarly to CopyObject
```
itaiad200, line 93: So it needs to maintain that new table too, right?

design/open/uncomitted-gc.md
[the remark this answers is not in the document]
itaiad200: @N-o-Z  Still relevant - If we have a way to logically do a rename with copy and delete, why do we need a non-atomic rename in the API?

itaiad200 (approved):
Thanks for addressing the comments, I added a few more but wish not to further block.

design/open/uncomitted-gc.md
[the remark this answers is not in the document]
N-o-Z: You have a logical way to do it using the S3 Gateway, no way of doing it using the OpenAPI

design/accepted/gc_plus/uncomitted-gc.md
N-o-Z: Explained in PrepareUncommittedForGC - added additional note to emphasize

design/accepted/gc_plus/uncomitted-gc.md
N-o-Z: Added

design/accepted/gc_plus/uncomitted-gc.md
```diff
@@ -0,0 +1,215 @@
...
+Objects that are found in 1 and are not in 2 or 3 can be safely deleted by the Garbage Collector.
+
+### 1. Listing namespace objects
+
+For large repositories, object listing is a very time-consuming operation - therefore we need to find a way to optimize it.
+The suggested method is to split the repository structure into fixed size (upper bounded) slices.
+These slices can then be scanned independently using multiple workers.
+In addition, taking advantage of the common property of the listing operation, which lists objects in a lexicographical order, we can create the slices in a manner which
+enables additional optimizations on the GC process (read further for details).
+To resolve indexing issues with existing repositories that have a flat filesystem, suggested to create these slices under a `data` prefix.
```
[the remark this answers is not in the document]
itaiad200: we'll need to revert https://github.com/treeverse/lakeFS/pull/4530 if we change the proposal.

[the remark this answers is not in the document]
itaiad200: Also @arielshaqed - not using prefixes that are linearly decreasing will limit our ability to run gradual uncommitted GC. Read the gradual flow to understand the reasoning on why it's critical to have prefixes for easy filtering and skip listing of the entire storage namespace time after time.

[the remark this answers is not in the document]
arielshaqed: Thanks!

design/accepted/gc_plus/uncomitted-gc.md
arielshaqed: Also on AWS awslabs/amazon-eks-ami#249 is a recurring issue in AMIs, and a linked issue has someone with a 7 minute time skew on EKS.

Clocks should be synced, but I think we will exact too strong a penalty for failure here. We should add:

1. Some safety mechanism.

   For instance:
   * If when we scan we discover prefixes from the distant future (>1 minute?) then error out and mark some metric for alerting.
   *  Create an additional marker object in every partition once an hour, and scan what the other cluster members put in _theirs_.  This gives an estimate of time skew, which can be monitored.
3. Operational guidance (strongly worded documentation).

arielshaqed (commented):
Neat!

I'm only worried about the operational aspect of maintaining synchronized times across the cluster of lakeFS servers.  But I believe that we can observe the time skew automatically, even as part of time-based partitioning!

Given that we're talking about deleting files, we should be very sure that we help operators control time sync.  Ideally fail-shut and refuse to GC uncommitted or even work at all when time sync is not good enough.  (I'd talking about minutes-level time sync here, of course.  NTP gives you milliseconds-level time sync.  But virtualization is notorious for screwing up time sync occasionally.)

design/accepted/gc_plus/uncomitted-gc.md
```diff
@@ -0,0 +1,215 @@
...
+
+![Repository Structure](uncommitted-gc-repo-struct.png)
+
+### 2. Listing of lakeFS repository committed objects
+
+Similar to the way GC works today, use repository meta-ranges and ranges to read all committed objects in the repository. 
+
+### 3. Listing of lakeFS repository uncommitted objects
+
+Expose a new API in lakeFS which writes repository uncommitted objects information into a parquet file in a dedicated path
```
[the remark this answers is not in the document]
N-o-Z: I agree this is an implementation detail and will remove the format specific details from the proposal

design/accepted/gc_plus/uncomitted-gc.md
```diff
@@ -0,0 +1,215 @@
...
+    1. Doing so will allow us to use this time interval to filter objects that might have been uploaded and waiting for
+       the link API and avoid them being deleted by the GC process.
+    2. Objects that were uploaded to a physical address issued by the API and were not linked before the token expired will
+       eventually be deleted by the GC job.
+>**Note:** These changes will also solves the following [issue](https://github.com/treeverse/lakeFS/issues/4438)
+
+#### Track copied objects in ref-store
+
+lakeFS will track copy operations of uncommitted objects and store them in the ref-store for a limited duration.
+GC will use this information as part of the uncommitted data to avoid a race between the GC job and rename operation.
```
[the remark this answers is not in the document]
N-o-Z: We are tracking copy operations but the purpose is to avoid a race in rename.
Consider the following:
1. GC starts reading branch staging area 
2. Rename operation create a copy of staged entry - which doesn't not get listed due to listing order
3. Rename operation deletes original entry before GC is able to scan it as part of the listing of staging area

As a result, the physical address will not be listed in uncommitted (or committed) and will be candidate for deletion.
Reading from the copy table - allows exempting copied objects from the delete list and avoid this race.

The copied entries lifespan can be minutes or can be removed as part of a branch operation (commit/reset/delete branch)

In the next GC iteration - these addresses will either be in committed, still in staging or deleted and will be handled accordingly

design/accepted/gc_plus/uncomitted-gc.md
```diff
@@ -0,0 +1,215 @@
...
+* Even with the given optimizations, the GC process is still very much dependent on the amount of changes that were made on the repository
+since the last GC run.
+
+## Performance Requirements
+
+TLDR; [Bottom Line](#minimal-performance-bottom-line)
+
+The heaviest operation during the GC process, is the namespace listing. And while we added the above optimizations to mitigate
+this process, the fact remains - we still need to scan the entire namespace (in the Clean Run mode).
+Performing tests against an AWS S3 bucket using a Databricks notebook on a m4.large cluster, we've observed listing of
```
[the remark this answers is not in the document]
N-o-Z: Checked with an xlarge cluster and max workers 10.
Did some more tests and created a udf which reads a specific partition and returns the objects information.
For 1,000,000 objects divided equally into 100 partitions I was able to load the information into a DF within 30 seconds.
The results looks pretty promising - and I'm sure we can also improve on that.
I will update the proposal with final results this week

PR #4388, opened 2022-10-18
Configure jobBranch and taskBranch

arielshaqed:
* Replace "workBranch" with "jobBranch" and "taskBranch":
  - jobBranch branches from the output branch and is a base for all tasks on a job.
  - taskBranch branches from jobBranch and is for one task attempt.  It can be deleted when the task aborts _without_ affecting jobBranch.
* Create branches in setup{Job,Task}.


lynnro314 (approved):
LGTM, thanks!

PR #4389, opened 2022-10-18
lakeFS Hadoop Filesystem - Hadoop 3 packaging

### LINKED ISSUE #3957: lakeFSFS to Hadoop 3: package

Jonathan-Rosenberg:
* Added support for Hadoop 2, 3-3.2.1, and 3.2.2+ packaging.
All package options have a common path (`src/main/java`) from which they package classes, and in addition, a specified location per package is added. For example, the Hadoop 3 profile states that in addition to the common packaged files, all of the files under `src/hadoop3/java` will be packed as well.
To choose which version to package, all you need to do is to add one of the following profiles to the `mvn` command:
  * `hadoop2` - packaging with Hadoop 2 dependencies.
  * `hadoop3` - packaging with Hadoop 3 dependencies.
  * `hadoop321` - packaging with Hadoop 3.2.1 dependencies. 

* Changed the ESTI tests to package Hadoop 2 and use it (as it did so far). **Testing other Hadoop versions using ESTI's workflow should be completed on a different PR**.
* Changed the "Publish Hadoop lakeFS Filesystem" workflow to receive a Hadoop version [2, 3.2.1, 3] and package by it.
Screen Shot 2022-10-18 at 14 24 29

### Workflow
Changed the bucket name to a private bucket, and ran it with the three Hadoop versions.
### Compatibility
Ran the Sonnets app on Databricks using the three Hadoop versions that were published in the previous step (the Hadoop versions on DB were 2.7.4 and 3.3.1 respectively).

---
### P.S.
The need for differentiation between versions 3.0.0-3.2.1 and 3.2.2+ arises from [this](https://issues.apache.org/jira/browse/SPARK-33212) JIRA ticket

Closes #3957


clients/hadoopfs/src/test/java/io/lakefs/contract/TestLakeFSFileSystemContract.java
```diff
@@ -4,6 +4,10 @@
 import org.apache.hadoop.fs.FileSystemContractBaseTest;
 import org.apache.hadoop.fs.Path;
 
+import org.junit.Before;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
```
johnnyaug, line 9: This change is effectively covered by #4357.

clients/hadoopfs/src/hadoop2/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-common</artifactId>
+            <scope>test</scope>
+            <type>test-jar</type>
+            <version>${hadoop.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-aws</artifactId>
+            <version>${hadoop.version}</version>
```
johnnyaug, line 48: This dependency will become provided in #4357.

clients/hadoopfs/src/hadoop3/pom-h321.xml
```diff
@@ -0,0 +1,70 @@
...
+        </dependency>
+        <dependency>
+            <groupId>commons-lang</groupId>
+            <artifactId>commons-lang</artifactId>
+            <version>2.6</version>
+        </dependency>
+
+        <dependency>
+            <groupId>org.apache.hadoop</groupId>
+            <artifactId>hadoop-common</artifactId>
```
johnnyaug, line 32: Can all the hadoop-* be in the parent pom? (instead of being in all versions)

clients/hadoopfs/src/hadoop3/pom-h321.xml
```diff
@@ -0,0 +1,70 @@
...
+            <scope>test</scope>
+            <type>test-jar</type>
+            <version>${hadoop.version}</version>
+        </dependency>
+    </dependencies>
+
+    <build>
```
johnnyaug, line 45: Can this go in the parent pom? (instead of being in all versions)

johnnyaug (commented):
While the code looks very good, I would like to express my opinion that creating this separation is an unnecessary complication at the moment:

1. There are no code changes between Hadoop versions.
2. All Hadoop-related dependencies are not included in the final JARs (either right now or in the [near future](#4357)).

Therefore, all three jars are essentially the same one, so I see no reason to burden to the user with the cognitive load of choosing the right one.

Jonathan-Rosenberg:
Thanks, @johnnyaug!

I agree that currently there are no code differences between Hadoop 2 and Hadoop 3 APIs, but I'm not that certain that there won't be such changes in the foreseeable future.
Adding the LakeFSOutputCommitter or any change in the current metadata client (as happened recently) might introduce a breaking change, resulting in three different Jars.
Personally, I wouldn't want to add a new code and then find out that it doesn't work, diverge from the main task and create the supporting infrastructure around it.
I also think that Spark users should know which Hadoop version they use and select the appropriate one from the list of three.

johnnyaug:
@Jonathan-Rosenberg, you raise a good point. We need the tests to run on the correct versions of Hadoop dependencies to ensure the code works. However, it doesn't require packaging different Jars at the end.

PR #4397, opened 2022-10-19
Docs: Add note to Airbyte doc broken integration

N-o-Z:
Added header note, until integration issue with lakeFS is fixed (reference https://github.com/airbytehq/airbyte/pull/18031)


nopcoder (changes requested):

nopcoder (approved):
LGTM, but it will look better if we fix the integration ;)

N-o-Z:
@nopcoder  Unfortunately it's not up to us. From what I saw the TTM (Time To Merge 😆 )  of a PR in Airbyte is around 3 weeks so it's best if we add this for now.

PR #4400, opened 2022-10-19
Use "provided" hadoop dependencies

### LINKED ISSUE #4399: Use "provided" hadoop-* dependencies in Spark metadata client
hadoop-* dependencies are not well-shaded.  By including them in our
assembly we make users' lives harder.  Make them "provided": they _will_ be
provided in any Spark installation that is in actual use (and if not, users
should include e.g. hadoop-aws explicitly).

arielshaqed:
Fixes #4399.

Tested by manually including hadoop-aws packages in Spark.


docs/reference/garbage-collection.md
```diff
@@ -158,6 +158,7 @@ You should specify the Uber-jar path instead of `<APPLICATION-JAR-PATH>` and run
 
   ```bash
 spark-submit --class io.treeverse.clients.GarbageCollector \
+  --packages org.apache.hadoop:hadoop-aws:2.7.7 \
```
talSofer, line 161: @arielshaqed is this the right hadoop-aws version for all three Spark clients?

Jonathan-Rosenberg (approved):

docs/reference/garbage-collection.md
arielshaqed: Not really, of course.  Spark 3 with Hadoop 3 uses some 3.x.y.  E.g. the 312-hadoop3 build is _compiled_ against Hadoop 3.2.1.

Good catch, fixed this version!  But _none_ of the guide really explains Hadoop 2 vs. Hadoop 3 very well.

arielshaqed (commented):
Thanks!

PTAL...

talSofer (approved):
Thanks! can you please update CHANGELOG to include this change before committing?

PR #4402, opened 2022-10-19
Move audit logs tag to the right place

idanovo:


pkg/httputil/logging.go
```diff
@@ -76,9 +75,10 @@ func DefaultLoggingMiddleware(requestIDHeaderName string, fields logging.Fields,
 			next.ServeHTTP(writer, r) // handle the request
 
 			loggingFields := logging.Fields{
-				"took":        time.Since(startTime),
-				"status_code": writer.StatusCode,
-				"sent_bytes":  writer.ResponseSize,
+				"took":           time.Since(startTime),
+				"status_code":    writer.StatusCode,
+				"sent_bytes":     writer.ResponseSize,
+				logging.LogAudit: "API",
```
nopcoder, line 81: I don't think we need this one, at least not here - the same middleware is used for logging s3 and api requests.
The field "service_name" value can be "s3_gateway" or "rest_api" to capture which instance of this middleware we report from.
If service_name is used more than this middleware we can mark this log with a different label to separate it from other logging lines - but it is not just API.

nopcoder (changes requested):

ortz:
@idanovo I'm not sure I understand how this differentiates between the different API calls. 
Mind providing more context?

idanovo:
> @idanovo I'm not sure I understand how this differentiates between the different API calls. Mind providing more context?

The purpose of this PR is to have the audit logs tag only for the logs that audit the calls the user does (rest/S3 gateway). 
Currently, we use this tag also for irrelevant logs.

Issue #4070, opened 2022-08-31
Docs: Explain metadata operations cost and affects better


itaiad200:
Explain better how diffs, merges and commits work. Explain how metadata operations are linear to the size of the change. Explain common usage patterns like getting a change summary, running multiple commits or merges on a branch, etc. and how it might impact performance. Something like [this section of the S3 documentation](https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance-design-patterns.html) could be very helpful for users going to production

PR #4403, opened 2022-10-19
[API breaking change] add prefix change indicator

### LINKED ISSUE #4068: Prefix Change indicator for Diff should change to reflect prefix change
Diff API: returning the same "change" indicator for prefixes as we do for objects is misleading. common prefixes should probably have a different indicator that means "there are changes under this prefix" so users understand they need to diff further.

eden-ohana:
closes #4068 
documentation will be added as part of #4070 

I'm open to better suggestions for a name to indicate prefix change

### DIFF
--- a/pkg/api/transform.go
+++ b/pkg/api/transform.go
@@ -15,7 +15,7 @@ func transformDifferenceTypeToString(d catalog.DifferenceType) string {
 	case catalog.DifferenceTypeConflict:
 		return "conflict"
 	case catalog.DifferenceTypePrefixChanged:
-		return "prefix changed"
+		return "changes under prefix"
 	default:
 		return ""
 	}
--- a/pkg/api/transform.go
+++ b/pkg/api/transform.go
@@ -14,6 +14,8 @@ func transformDifferenceTypeToString(d catalog.DifferenceType) string {
 		return "changed"
 	case catalog.DifferenceTypeConflict:
 		return "conflict"
+	case catalog.DifferenceTypePrefixChanged:
+		return "prefix changed"
 	default:
 		return ""
 	}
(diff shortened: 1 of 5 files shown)


itaiad200 (changes requested):
Thanks! Some questions & comments:
1. Does the UI support it out of the box? I think we should at least introduce new colouring to changed prefixes to differentiate them from changed objects.
2. We need some testing (new or modify existing) that validates the new behaviour.
3. I think we should make the documentation changes in this PR, #4070 is a whole different issue (although it involves docs too..).

pkg/api/transform.go
```diff
@@ -14,6 +14,8 @@ func transformDifferenceTypeToString(d catalog.DifferenceType) string {
 		return "changed"
 	case catalog.DifferenceTypeConflict:
 		return "conflict"
+	case catalog.DifferenceTypePrefixChanged:
+		return "prefix changed"
```
johnnyaug, line 18:
```suggestion
		return "changes under prefix"
```

johnnyaug:
@itaiad200 - please, no more colors in the diff screen. The UI shows a folder icon, indicating that this is a prefix and not a single object. I would argue that this enough.

eden-ohana:
@itaiad200 
1. Currently, The UI shows the same color for prefix as for change. I don't think we should add another color for it, maybe add text with "changes under this prefix" but I'm not sure it's necessary because it has a change color and folder icon.
2. Done
3. There isn't an explanation on the docs about the current diff behavior, on #4070 I'm adding a diff section on the docs that we can explain there in detail. unless you have an idea where to document it for now.

pkg/api/transform.go
eden-ohana: done

eden-ohana:
image

itaiad200 (approved):
LGTM
Please change the hover message on the directory symbol

PR #4414, opened 2022-10-20
Fix creation validation for % in entity names

### LINKED ISSUE #3708: Failure in loading a user with % in their name
1. Create an integration user with a name like idan%demo
2. Try to go click on the user that was created
Because % is a special character in URL, you will get an empty page

### LINKED ISSUE #3809: Creating a new policy with the same ID as an existing policy "merges" them
[Reported by a user](https://app.slack.com/client/T013V60QY06/C02CV7MUV4G/thread/C02CV7MUV4G-1658946800.540749).

> I tried to create a read policy but somehow instead of a new policy, the change is merged to an existing policy.
> [...]
> Thanks, I figured it out. I used GUI. I gave a new policy name, but in the JSON I copied, I **forget to change the policy id**, then lakeFS “smartly” merge it to existing policy instead of creating a new one.

(Emphasis mine)

eladlachmi:
Resolves #3708

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -3637,7 +3637,7 @@ func (c *Controller) authorize(w http.ResponseWriter, r *http.Request, perms per
 
 func (c *Controller) isNameValid(name string, nameType string) (bool, string) {
 	if strings.Contains(name, "%") {
-		return false, fmt.Sprintf("%s name cannot contain '%'", nameType)
+		return false, fmt.Sprintf("%s name cannot contain '%%'", nameType)
 	}
 	return true, ""
 }
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -867,7 +867,7 @@ func (c *Controller) generateResetPasswordToken(email string, duration time.Dura
 func (c *Controller) CreateUser(w http.ResponseWriter, r *http.Request, body CreateUserJSONRequestBody) {
 	invite := swag.BoolValue(body.InviteUser)
 	username := body.Id
-	
+
 	// Check that username is valid
 	valid, msg := c.isNameValid(username, "User")
 	if !valid {
@@ -3626,7 +3626,7 @@ func (c *Controller) authorize(w http.ResponseWriter, r *http.Request, perms per
 func (c *Controller) isNameValid(name string, nameType string) (bool, string) {
 	// URLs are % encoded. Allowing % signs in entity names would
 	// limit the ability to use these entity names in the URL for both
-	// client-side routing and API fetch requests 
+	// client-side routing and API fetch requests
 	if strings.Contains(name, "%") {
 		return false, fmt.Sprintf("%s name cannot contain '%%'", nameType)
 	}
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -710,18 +710,6 @@ func (c *Controller) CreatePolicy(w http.ResponseWriter, r *http.Request, body C
 		writeError(w, http.StatusBadRequest, msg)
 		return
 	}
-	
-	// Check that we don't already have a policy with this ID
-	_, gpErr := c.Auth.GetPolicy(ctx, body.Id)
-	// if err is nil, it means we got a policy, which means we can't create
-	// a new policy with this ID. Return a 409 Conflict http status code
-	// which means the request conflicts with the current state of the server
-	if gpErr == nil {
-		gpErr = auth.ErrAlreadyExists
-		if c.handleAPIError(ctx, w, gpErr) {
-			return
-		}
-	}
 
 	stmts := make(model.Statements, len(body.Statement))
 	for i, apiStatement := range body.Statement {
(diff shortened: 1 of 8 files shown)


pkg/api/controller.go
```diff
@@ -3600,3 +3634,10 @@ func (c *Controller) authorize(w http.ResponseWriter, r *http.Request, perms per
 	}
 	return true
 }
+
+func (c *Controller) isNameValid(name string, nameType string) (bool, string) {
+	if strings.Contains(name, "%") {
+		return false, fmt.Sprintf("%s name cannot contain '%'", nameType)
```
arielshaqed, line 3640: (Apart from the govet format string error...)

I think validation belongs in the auth service not in the controller.

Also llease comment here why a name with `%` is invalid.  I remember something about the React router component being broken.  Otherwise it is easy to read this code and assume we have some silly format-string vulnerability or worse.

pkg/api/controller.go
```diff
@@ -3600,3 +3634,10 @@ func (c *Controller) authorize(w http.ResponseWriter, r *http.Request, perms per
 	}
 	return true
 }
+
+func (c *Controller) isNameValid(name string, nameType string) (bool, string) {
```
arielshaqed, line 3638: Is the second arg intentionally unused?

pkg/api/controller.go
```diff
@@ -697,6 +704,25 @@ func (c *Controller) CreatePolicy(w http.ResponseWriter, r *http.Request, body C
...
+	_, gpErr := c.Auth.GetPolicy(ctx, body.Id)
+	// if err is nil, it means we got a policy, which means we can't create
+	// a new policy with this ID. Return a 409 Conflict http status code
+	// which means the request conflicts with the current state of the server
+	if gpErr == nil {
+		gpErr = auth.ErrAlreadyExists
+		if c.handleAPIError(ctx, w, gpErr) {
+			return
+		}
+	}
```
arielshaqed, line 724: AFAICT this is a race condition.  And one that can actually occur, e.g. if my GUI is stuck and I redo a similar operation from another window, when both are unstuck we can race here.

We should use some SetIf capability on the database, as part of the `c.Auth.WritePolicy` call.

pkg/api/controller.go
```diff
@@ -697,6 +704,25 @@ func (c *Controller) CreatePolicy(w http.ResponseWriter, r *http.Request, body C
...
+	
+	// Check that we don't already have a policy with this ID
+	_, gpErr := c.Auth.GetPolicy(ctx, body.Id)
+	// if err is nil, it means we got a policy, which means we can't create
+	// a new policy with this ID. Return a 409 Conflict http status code
+	// which means the request conflicts with the current state of the server
+	if gpErr == nil {
+		gpErr = auth.ErrAlreadyExists
+		if c.handleAPIError(ctx, w, gpErr) {
+			return
```
arielshaqed, line 722:
```suggestion
		if c.handleAPIError(ctx, w, auth.ErrAlreadyExists) {
			return
```
will probably be somewhat easier for me to analyze in future, as I don't need to scan below to see that `gpErr` is not re-used.

arielshaqed (changes requested):
Please retitle the PR to a descriptive name.  Otherwise it is much harder (for me) to figure out history and why things changed.

Requested change: avoid the race condition before setting a policy.  Please consider moving more business logic into the auth service, too.

THANKS!

pkg/api/controller.go
eladlachmi: I'm not sure I agree regarding the name validation, as I'd say it's a type of request/input validation. You could make the case for it being in the controller.   

I was also following the example of the existing email address validation, which is placed in the controller.
Regarding the duplicate policy name check, I agree that it's less input validation and more specific to the entity type, so I moved it to auth.

Added comment and fixed string formatting issue.

pkg/api/controller.go
eladlachmi: Unused? Not sure I follow.
`nameType` is used in the string formatting.

pkg/api/controller.go
eladlachmi: This code has been moved and refactored per your previous comment.

arielshaqed (approved):
Thanks!

A thought: IIRC the reason we disallow `%`s is some bug in react-router.  Are we tracking that bug somewhere?  If not, maybe the place to do it is in the comment explaining that we don't like % signs.

Because it's a _really_ unfortunate and strange exception.

webui/src/pages/auth/validation.ts
```diff
@@ -0,0 +1,30 @@
+type ValidationError = string;
+
+export const INVALID_USER_NAME_ERROR_MESSAGE: ValidationError =
+  "User name cannot include the % sign";
+export const INVALID_GROUP_NAME_ERROR_MESSAGE: ValidationError =
+  "Group name cannot include the % sign";
+export const INVALID_POLICY_ID_ERROR_MESSAGE: ValidationError =
+  "Policy ID cannot include the % sign";
```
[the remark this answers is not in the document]
eladlachmi: I understand, but this is a very common practice these days. Fewer HTTP requests mean less pressure on backends and a better and more responsive user experience.

PR #4418, opened 2022-10-20
lakectl fs download support

### LINKED ISSUE #4095: Add a download capability to lakectl
Support lakectl download command to support single/multiple files download to local directory.

```
# single file
lakectl fs download lakefs://example-repo/example-branch/prefix/doc.md

# multiple files
lakectl fs download --recursive lakefs://example-repo/example-branch/prefix/

# select destination
lakectl fs download --recursive lakefs://example-repo/example-branch/prefix/ mydata
```

nopcoder:
Close #4095

### DIFF
--- a/cmd/lakectl/cmd/fs.go
+++ b/cmd/lakectl/cmd/fs.go
@@ -439,10 +439,6 @@ var fsDownloadCmd = &cobra.Command{
 		var dest string
 		if len(args) > 1 {
 			dest = args[1]
-			// verify destination path exists
-			if len(dest) > 0 {
-				pathExists(dest)
-			}
 		}
 
 		// list the files
@@ -451,14 +447,14 @@ var fsDownloadCmd = &cobra.Command{
 		sourcePath := api.StringValue(pathURI.Path)
 
 		// recursive assume source is directory
-		if recursive && len(sourcePath) > 0 && !strings.HasSuffix(sourcePath, "/") {
-			sourcePath += "/"
+		if recursive && len(sourcePath) > 0 && !strings.HasSuffix(sourcePath, uri.PathSeparator) {
+			sourcePath += uri.PathSeparator
 		}
 
 		// prefix to remove from destination
 		prefix := filepath.Dir(sourcePath)
 		if prefix != "" {
-			prefix += "/"
+			prefix += uri.PathSeparator
 		}
 		ctx := cmd.Context()
 		// list objects to download
--- a/cmd/lakectl/cmd/fs.go
+++ b/cmd/lakectl/cmd/fs.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"mime"
@@ -33,6 +34,8 @@ const fsRecursiveTemplate = `Files: {{.Count}}
 Total Size: {{.Bytes}} bytes
 Human Total Size: {{.Bytes|human_bytes}}
 `
+
+var ErrRequestFailed = errors.New("request failed")
 
 var fsStatCmd = &cobra.Command{
 	Use:               "stat <path uri>",
@@ -410,12 +413,17 @@ var fsRmCmd = &cobra.Command{
 	},
 }
 
-const DefaultDownloadParallel = 6
+const (
+	fsDownloadCmdMinArgs = 1
+	fsDownloadCmdMaxArgs = 2
+
+	fsDownloadParallelDefault = 6
+)
 
 var fsDownloadCmd = &cobra.Command{
 	Use:   "download <path uri> [<destination path>]",
 	Short: "Download object(s) from a given repository path",
-	Args:  cobra.RangeArgs(1, 2),
+	Args:  cobra.RangeArgs(fsDownloadCmdMinArgs, fsDownloadCmdMaxArgs),
 	Run: func(cmd *cobra.Command, args []string) {
 		pathURI := MustParsePathURI("path", args[0])
 		flagSet := cmd.Flags()
@@ -440,41 +448,25 @@ var fsDownloadCmd = &cobra.Command{
 		// list the files
 		client := getClient()
 		downloadCh := make(chan string)
-		sourceURI := api.StringValue(pathURI.Path)
+		sourcePath := api.StringValue(pathURI.Path)
 
 		// recursive assume source is directory
-		if recursive && len(sourceURI) > 0 && !strings.HasSuffix(sourceURI, "/") {
-			sourceURI += "/"
+		if recursive && len(sourcePath) > 0 && !strings.HasSuffix(sourcePath, "/") {
+			sourcePath += "/"
 		}
+
+		ctx := cmd.Context()
+		// list objects to download
 		go func() {
 			defer close(downloadCh)
 			if recursive {
-				pfx := api.PaginationPrefix(sourceURI)
-				var from string
-				for {
-					params := &api.ListObjectsParams{
-						Prefix: &pfx,
-						After:  api.PaginationAfterPtr(from),
-					}
-					resp, err := client.ListObjectsWithResponse(cmd.Context(), pathURI.Repository, pathURI.Ref, params)
-					DieOnErrorOrUnexpectedStatusCode(resp, err, http.StatusOK)
-					for _, p := range resp.JSON200.Results {
-						downloadCh <- p.Path
-					}
-
-					pagination := resp.JSON200.Pagination
-					if !pagination.HasMore {
-						break
-					}
-					from = pagination.NextOffset
-				}
+				listRecursiveHelper(ctx, client, pathURI.Repository, pathURI.Ref, sourcePath, downloadCh)
 			} else {
 				downloadCh <- api.StringValue(pathURI.Path)
 			}
 		}()
 
 		// download in parallel
-		ctx := cmd.Context()
 		var (
 			wg         sync.WaitGroup
 			errCounter int64
@@ -490,7 +482,7 @@ var fsDownloadCmd = &cobra.Command{
 						Path:       &downloadPath,
 					}
 					// destination is without the source URI
-					dst := filepath.Join(dest, strings.TrimPrefix(downloadPath, sourceURI))
+					dst := filepath.Join(dest, strings.TrimPrefix(downloadPath, sourcePath))
 					err := downloadHelper(ctx, client, direct, src, dst)
 					if err == nil {
 						fmt.Printf("Download: %s to %s\n", src.String(), dst)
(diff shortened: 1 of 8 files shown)


cmd/lakectl/cmd/fs.go
```diff
@@ -409,6 +410,138 @@ var fsRmCmd = &cobra.Command{
 	},
 }
 
+const DefaultDownloadParallel = 6
```
itaiad200, line 413:
```suggestion
const defaultDownloadParallel = 6
```

cmd/lakectl/cmd/fs.go
```diff
@@ -409,6 +410,138 @@ var fsRmCmd = &cobra.Command{
...
+
+var fsDownloadCmd = &cobra.Command{
```
itaiad200, line 415: Can we add esti test to the new command?

cmd/lakectl/cmd/fs.go
```diff
@@ -409,6 +410,138 @@ var fsRmCmd = &cobra.Command{
...
+		err  error
+		body io.ReadCloser
+	)
+	if direct {
+		_, body, err = helpers.ClientDownload(ctx, client, src.Repository, src.Ref, *src.Path)
+	} else {
+		var resp *http.Response
+		resp, err = client.GetObject(ctx, src.Repository, src.Ref, &api.GetObjectParams{Path: *src.Path})
+		if resp != nil {
+			body = resp.Body
```
itaiad200, line 525: This might be the body of non-200 response - check the status code first.

itaiad200 (changes requested):
Thanks - mostly minor comments. Please add tests to lakectl for the new functionaility

cmd/lakectl/cmd/fs.go
```diff
@@ -409,6 +413,159 @@ var fsRmCmd = &cobra.Command{
...
+	Short: "Download object(s) from a given repository path",
+	Args:  cobra.RangeArgs(fsDownloadCmdMinArgs, fsDownloadCmdMaxArgs),
+	Run: func(cmd *cobra.Command, args []string) {
+		pathURI := MustParsePathURI("path", args[0])
+		flagSet := cmd.Flags()
+		direct := MustBool(flagSet.GetBool("direct"))
+		recursive := MustBool(flagSet.GetBool("recursive"))
+		parallel := MustInt(flagSet.GetInt("parallel"))
+
+		if parallel < 1 {
```
N-o-Z, line 434: Do we not want to bound it from above as well?

cmd/lakectl/cmd/fs.go
```diff
@@ -409,6 +413,159 @@ var fsRmCmd = &cobra.Command{
...
+					src := uri.URI{
+						Repository: pathURI.Repository,
+						Ref:        pathURI.Ref,
+						Path:       &downloadPath,
+					}
+					// destination is without the source URI
+					dst := filepath.Join(dest, strings.TrimPrefix(downloadPath, sourcePath))
+					err := downloadHelper(ctx, client, direct, src, dst)
+					if err == nil {
+						fmt.Printf("Download: %s to %s\n", src.String(), dst)
```
N-o-Z, line 488: Maybe we shouldn't print on every file - since this might spam the stdout?

N-o-Z: Preferably print a single 'download from path: ' on recursive flag
Suggest also to add a periodic output the prints:
'Downloaded x/y objects'

N-o-Z (changes requested):
@nopcoder  Thanks! 
Some user experience suggestions

cmd/lakectl/cmd/fs.go
```diff
@@ -409,6 +413,164 @@ var fsRmCmd = &cobra.Command{
...
+				pathExists(dest)
+			}
+		}
+
+		// list the files
+		client := getClient()
+		downloadCh := make(chan string)
+		sourcePath := api.StringValue(pathURI.Path)
+
+		// recursive assume source is directory
```
itai-david, line 453: Why not support a partial prefix too, the same way we do for la`lakectl fs ls`?

cmd/lakectl/cmd/fs.go
nopcoder: Prefer to limit just things that will not work and let the user abuse if needed. 
Checked the upload implementation - it is unbound (will need to fix it too) - you get a goroutine for each file. which means that if you do a recursive upload of a large file system, you will probably crash.

cmd/lakectl/cmd/fs.go
nopcoder: In this implementation is cloned the `aws s3 cp` behaviour - but I'll check how to use one of the progress packages we have to get line per worker that will print out the current uploaded file.

cmd/lakectl/cmd/fs.go
nopcoder: 1. why do you mean partial prefix?
2. fixed this implementation - had a bug with the prefix.  For single file - just download the file to the destination. For recursive - download all the directory structure from the source to the local destination. Same behaviour of common tools (and aws s3 cp).

itai-david: Partial prefix with `recursive` option should handle all prefixes ("dirs") that start with the prefix. Here's how it looks on `fs ls`:
```
$ ./lakectl fs ls lakefs://myrepo/main/somepre --recursive
object          2022-10-20 14:30:20 -0500 CDT    3.2 kB          somepref/file1
object          2022-10-20 14:30:20 -0500 CDT    3.2 kB          somepref/file2
object          2022-10-20 14:59:49 -0500 CDT    3.2 kB          somepref1/file3
object          2022-10-20 14:59:49 -0500 CDT    3.2 kB          somepref1/file4
```
So I believe one should expect the same objects list to be downloaded by `fs download`, given the same arguments and flags

nopcoder: Good point - both commands listing and copy try to replicate aws s3 behaviour.
When you list the prefix will list the files with the same prefix, as you described.
But, if you try to download these files using `aws s3 cp` with the `--recursive` flag it will not copy non folder structure to a folder structure.
I assume this is by design as we like to keep the folder structure on the bucket/repository on the download folder.

cmd/lakectl/cmd/fs.go
```diff
@@ -409,6 +413,164 @@ var fsRmCmd = &cobra.Command{
...
+	if err != nil {
+		return err
+	}
+	defer func(body io.ReadCloser) {
+		_ = body.Close()
+	}(body)
+
+	// create destination dir if needed
+	dir := filepath.Dir(dst)
+	_ = os.MkdirAll(dir, os.ModePerm)
```
itai-david, line 543: We allow an object and a prefix of the same name. I think it might be problematic, due to the behavior of [`os.MkDirAll`](https://github.com/golang/go/blob/9e7c5030d064395e24a25dd1a81722f5583defcf/src/os/path.go#L25), as the object will be created as a local file, failing a firther attempt to create it as a directory. 
```
$ ./lakectl fs ls lakefs://myrepo/main/external/
object          2022-10-21 12:09:30 -0500 CDT    3.2 kB          internal
common_prefix                                                    internal/

$ ./lakectl fs ls lakefs://myrepo/main/external/ --recursive
object          2022-10-21 12:09:30 -0500 CDT    3.2 kB          external/internal
object          2022-10-21 12:09:36 -0500 CDT    3.2 kB          external/internal/obj1
```

itai-david: Thinking of it thourhg, this is actually a problem with the different way prefixes are treated in `lakefs` (and probably object stores ?) vs the way OS handles dirs. Not sure what the solution should be

nopcoder: - about the fail cascade in case of write that fail, it can occur for example when we try to download to a place we can't write. all the lines will report an error that we failed.
- about the behavior of 'ls' and 'download' regarding local files and prefix - we captured aws S3 behavior.

cmd/lakectl/cmd/fs.go
```diff
@@ -409,6 +413,159 @@ var fsRmCmd = &cobra.Command{
...
+
+		if parallel < 1 {
+			DieFmt("Invalid value for parallel (%d), minimum is 1.\n", parallel)
+		}
+
+		// optional destination directory
+		var dest string
+		if len(args) > 1 {
+			dest = args[1]
+			// verify destination path exists
```
itai-david, line 442:
```suggestion
			// verify destination path is valid
```
The comment is misleading as this will only fail for invalid `dest` path - the result from `pathExists` is ignored, plus the parent paths are created later, as ppart of the download.

itai-david (changes requested):
Thanks @nopcoder .
Some comments/requests inside, most of which are pretty minor.
One thing I think we should address is the way we correlate store's `prefix` with OS's directory, which seems problematic, due to optional same 'prefix' and 'object' paths

nopcoder:
> Thanks @nopcoder . Some comments/requests inside, most of which are pretty minor. One thing I think we should address is the way we correlate store's `prefix` with OS's directory, which seems problematic, due to optional same 'prefix' and 'object' paths

About the destination directory - for listing:

```sh
$ lakectl fs ls --recursive lakefs://repo-cd8s7ib4lvk89u4ff420/main/
object          2022-10-21 00:55:21 +0300 IDT    1.0 kB          a/b/c
```

if we download `a` recursive to `repo` destination folder:

```sh
$ lakectl fs download --recursive lakefs://repo-cd8s7ib4lvk89u4ff420/main/a repo
Download: lakefs://repo-cd8s7ib4lvk89u4ff420/main/a/b/c to repo/b/c
```

we will get the expected result of the same structure as we had on the object store on local directory:

```sh
$ find repo
repo
repo/b
repo/b/c
```

Can you give me an example to test that doesn't give something as expected?

cmd/lakectl/cmd/fs.go
itai-david: Interesting. I can see how it aligns with `aws s3 cp` behavior
Why aren't we using `lakectl fs cp` for download  too?

itai-david (approved):
Thanks for addressing all questions/comments. LGTM

cmd/lakectl/cmd/fs.go
nopcoder: We don't have `cp` command - we had so far `upload` and `cat`. Think we can open a new issue to implement `cp` and deprecate all the rest.

image

cmd/lakectl/cmd/fs.go
nopcoder: @N-o-Z played with it a little - think we should align the progress between upload/download. The big issue for me was the fact that it was hard to test the output when periodic download progress was printed.
I liked the fact that we are like aws cp and you can log and grep the output.

N-o-Z (approved):
LGTM

itaiad200 (approved):

PR #4426, opened 2022-10-20
Doc improvement: lakeFS Hadoop FileSystem

johnnyaug:
1. Update version number.
2. Update the Hadoop FileSystem CHANGELOG.
3. Remove disclaimer about Hadoop version.
4. Proofread, remove confusing text about per-bucket configuration.

### DIFF
--- a/docs/integrations/spark.md
+++ b/docs/integrations/spark.md
@@ -35,7 +35,7 @@ There are two ways you can use lakeFS with Spark:
 ## Use the S3 gateway
 
 lakeFS has an S3-compatible endpoint. Simply point Spark to this endpoint to get started quickly.
-You will use S3-style URIs to access your data, e.g. `s3a://example-repo/example-branch/example-table`.
+You will access your data using S3-style URIs, e.g. `s3a://example-repo/example-branch/example-table`.
 
 ### Configuration
 {: .no_toc }
--- a/docs/integrations/spark.md
+++ b/docs/integrations/spark.md
@@ -187,7 +187,7 @@ The data is now created in lakeFS as new changes in your branch. You can now com
 
 ## Use the lakeFS-specific Hadoop FileSystem
 
- If you're using lakeFS on top of S3, this mode will enhance your application's performance.
+If you're using lakeFS on top of S3, this mode will enhance your application's performance.
 In this mode, Spark will read and write objects directly from S3, reducing the load on the lakeFS server.
 It will still access the lakeFS server for metadata operations.
 
--- a/docs/integrations/spark.md
+++ b/docs/integrations/spark.md
@@ -34,19 +34,13 @@ There are two ways you can use lakeFS with Spark:
 
 ## Use the S3 gateway
 
-To use this mode you configure the Spark application to use S3A using the S3-compatible
-endpoint that the lakeFS server provides. Accordingly, all data flows through the lakeFS
-server.
-
-Accessing data in lakeFS from Spark is the same as accessing S3 data from Spark. The only
-changes you need to consider are:
-1. Setting the configurations to access lakeFS,
-1. Accessing objects using the lakeFS S3 path convention.
+lakeFS has an S3-compatible endpoint. Simply point Spark to this endpoint to get started quickly.
+You will use S3-style URIs to access your data, e.g. `s3a://example-repo/example-branch/example-table`.
 
 ### Configuration
 {: .no_toc }
 
-To configure Spark to work with lakeFS, we set S3 Hadoop configuration to the lakeFS endpoint and credentials:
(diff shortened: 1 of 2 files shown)


docs/integrations/spark.md
```diff
@@ -191,157 +185,128 @@ df.write.partitionBy("example-column").parquet(s"s3a://${repo}/${branch}/output-
...
-reducing the load on the lakeFS server.
-
-Accessing data in lakeFS from Spark is the same as accessing S3 data from Spark. The only
-changes you need to make are:
-
-1. Configure Spark to access lakeFS for metadata and S3 or a compatible underlying object
-   store to access data,
-1. Use `lakefs://repo/ref/path/to/data` URIs to read and write data on lakeFS rather than
-   `s3a://...` URIs.
+reducing the load on the lakeFS server. It will still access the lakeFS server for metadata operations.
```
talSofer, line 192: What do you think about adding the privacy advantage?

docs/integrations/spark.md
```diff
@@ -191,157 +185,128 @@ df.write.partitionBy("example-column").parquet(s"s3a://${repo}/${branch}/output-
...
+
+   Add the package to your _spark-submit_ command:
+
+   ```
+   --packages io.lakefs:hadoop-lakefs-assembly:0.1.8
+   ```
+
+   The jar is also available on a public S3 location: `s3://treeverse-clients-us-east/hadoop/hadoop-lakefs-assembly-0.1.8.jar`
+
+2. Configure S3A filesystem to with your S3 credentials (**not** the lakeFS credentials).
```
talSofer, line 210:
```suggestion
2. Configure S3A filesystem with your S3 credentials (**not** the lakeFS credentials).
```

docs/integrations/spark.md
```diff
@@ -191,157 +185,128 @@ df.write.partitionBy("example-column").parquet(s"s3a://${repo}/${branch}/output-
...
-
-Refer to the Hadoop AWS guide [Configuring different S3 buckets with Per-Bucket
-Configuration](https://hadoop.apache.org/docs/current/hadoop-aws/tools/hadoop-aws/index.html#Configuring_different_S3_buckets_with_Per-Bucket_Configuration).
-
-There's no need for per-repo configurations in lakeFS when all repositories are on the same
-lakeFS server. If you need to access repositories located on *multiple* lakeFS servers,
-configure multiple prefixes. For instance, you might configure both `fs.lakefs.impl` and
-`fs.lakefs2.impl` to be `io.lakefs.LakeFSFileSystem`, place separate endpoints and credentials
-under `fs.lakefs.*` and `fs.lakefs2.*`, and access the two servers using `lakefs://...` and
-`lakefs2://...` URLs.
```
talSofer, line 342: I agree that the per-bucket config text is confusing, but thin that we should lose this part 

> If you need to access repositories located on *multiple* lakeFS servers,
configure multiple prefixes. For instance, you might configure both `fs.lakefs.impl` and
`fs.lakefs2.impl` to be `io.lakefs.LakeFSFileSystem`, place separate endpoints and credentials
under `fs.lakefs.*` and `fs.lakefs2.*`, and access the two servers using `lakefs://...` and
`lakefs2://...` URLs.

Maybe we can call it something like lakeFS filesystem configuration in a multi-instance environment. WDYT?

docs/integrations/spark.md
```diff
@@ -191,157 +185,128 @@ df.write.partitionBy("example-column").parquet(s"s3a://${repo}/${branch}/output-
 
 The data is now created in lakeFS as new changes in your branch. You can now commit these changes or revert them.
 
-## Access lakeFS using the lakeFS-specific Hadoop FileSystem
+## Use the lakeFS-specific Hadoop FileSystem
 
-To use this mode, you configure the Spark application to perform metadata operations on the
-lakeFS server and all data operations directly through the same underlying object store that
-lakeFS uses. The lakeFS FileSystem currently supports Spark with Hadoop Apache 2.7 using only the S3A Hadoop FileSystem
```
talSofer, line 198: > using only the S3A Hadoop FileSystem

I think we should keep this limitation and make it clear that lakefsfs only works with s3a

talSofer (changes requested):
Thanks! I like the changes a lot, it simplifies reading. 

blocking because I think that we should reconsider bringing back some of the omitted text

docs/integrations/spark.md
johnnyaug: Added a clarification that this mode is only relevant for S3 users.

docs/integrations/spark.md
johnnyaug: I think this belongs in other community channel, rather than in the docs.

docs/integrations/spark.md
johnnyaug: Added.

talSofer (approved):

PR #4434, opened 2022-10-23
Remove lakefs import command

### LINKED ISSUE #4323: Deprecate `lakefs import`
There are 3 ways to [import data to lakeFS](https://docs.lakefs.io/setup/import.html#zero-copy-import) without copying it.
1. `lakefs import ...`
1. `lakectl ingest ...`
1. UI Import.
While the lakefs import was the first & only option to import data to lakeFS, it is still tricky to configure - you need to setup the inventory, run the lakeFS binary with the same configuration as in your production containers therefore requiring admin access and more. It’s also only available for S3. My assumption that options 2&3 covers all import use-cases when it comes to scale and ease of use. I think it’s time that we deprecate the lakefs import for a more consistent import experience and the chance of simplifying onboarding flows.

nopcoder:
Close #4323


docs/reference/garbage-collection.md
```diff
@@ -265,7 +265,7 @@ repositories may require increasing a read timeout.  If you run into timeout err
    An object which is later deleted from `main` will always be present in the stale branch, preventing it from being hard-deleted.
    
 1. lakeFS will never delete objects outside your repository's storage namespace.
-   In particular, objects that were imported using `lakefs import` or `lakectl ingest` will not be affected by GC jobs.
+   In particular, objects that were imported using `lakectl ingest` will not be affected by GC jobs.
```
itaiad200, line 268: Name the other import - UI Import Wizard

itaiad200 (approved):
Bye-Bye

Issue #4435, opened 2022-10-23
Unify "not found" and other common errors under a common base error


arielshaqed:
There are multiple "not found" errors in the code base, and all code has to be aware of all of them in order to check for them.  For instance not finding an object can return `graveler.ErrNotFound` or `catalog.ErrNotFound`.  This causes bugs:

* From the agile development viewpoint, calling code has to be deeply aware of implementation in order to know _which_ errors to handle.
* From the software engineering viewpoint, something like "not found" is a useful common case (for instance, when looking for an object, such an error probably means that we should not retry), and we should abstract accordingly.
* This is a degradation, we used to have something like this.
* This caused #4427.  In the past, before we had this common error, we had multiple similar instances of missing ErrNotFound.

PR #4436, opened 2022-10-23
Check for graveler *and* catalog ErrNotFound

### LINKED ISSUE #4427: lakeFS returns code 500 when getting absent object
Hi,

My locally-deployed lakeFS instance (`treeverse/lakefs:0.82.0`) returns a code 500 when trying to get an object that does not exist through the S3 API. This breaks compatibility with other applications.

Per boto3's documentation:
> You need the relevant read object (or version) permission for this operation. For more information, see [Specifying Permissions in a Policy](https://docs.aws.amazon.com/AmazonS3/latest/dev/using-with-s3-actions.html). If the object you request does not exist, the error Amazon S3 returns depends on whether you also have the ``s3:ListBucket`` permission.
> * If you have the ``s3:ListBucket`` permission on the bucket, Amazon S3 will return an HTTP status code 404 ("no such key") error. 
> * If you don’t have the ``s3:ListBucket`` permission, Amazon S3 will return an HTTP status code 403 ("access denied") error.

Expected behaviour: 
* lakeFS returns a code 404 as I have all read and write access.

arielshaqed:
Fixes #4427.

Opened #4435 to refactor to help protect against similar recurrences.

### DIFF
--- a/esti/s3_gateway_test.go
+++ b/esti/s3_gateway_test.go
@@ -114,6 +114,69 @@ func TestS3UploadAndDownload(t *testing.T) {
 	}
 }
 
+func TestS3GetObject(t *testing.T) {
+	SkipTestIfAskedTo(t)
+
+	const (
+		contents = "the quick brown fox jumps over the lazy dog"
+		goodPath = "main/exists"
+		badPath  = "main/does/not/exist"
+	)
+
+	ctx, _, repo := setupTest(t)
+	defer tearDownTest(repo)
+
+	accessKeyID := viper.GetString("access_key_id")
+	secretAccessKey := viper.GetString("secret_access_key")
+	endpoint := viper.GetString("s3_endpoint")
+	endpointSecure := viper.GetBool("s3_endpoint_secure")
+
+	// Upload an object
+	creds := sigs[0].GetCredentials(accessKeyID, secretAccessKey, "")
+	client, err := minio.New(endpoint, &minio.Options{
+		Creds:  creds,
+		Secure: endpointSecure,
+	})
+	if err != nil {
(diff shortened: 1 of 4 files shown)


arielshaqed:
Tested manually:
```
❯ AWS_MAX_ATTEMPTS=1 aws --profile local-s3 --endpoint http://localhost:8000/ s3api get-object --bucket foo --key main/asdf -

An error occurred (NoSuchKey) when calling the GetObject operation: The specified key does not exist.
```
(also tested with a bad branch and a bad repo).

Will add separately unit tests for getobject and headobject.

esti/s3_gateway_test.go
```diff
@@ -114,6 +114,60 @@ func TestS3UploadAndDownload(t *testing.T) {
...
+		if string(got) != contents {
+			t.Errorf("Got contents \"%s\" but expected \"%s\"", string(got), contents)
+		}
+	})
+	t.Run("Doesn't exist", func(t *testing.T) {
+		_, err := client.GetObject(ctx, repo, goodPath, minio.GetObjectOptions{})
+		if err != nil {
+			t.Errorf("client.GetObject(%s, %s): %s", repo, goodPath, err)
+		}
+	})
```
Jonathan-Rosenberg, line 168: shouldn't this test check for the correct response if the path doesn't exist?

arielshaqed: It does, now :-)

Jonathan-Rosenberg (approved):
LGTM!

itaiad200 (approved):
Thanks for the quick fix!

arielshaqed:
Thanks!

PR #4439, opened 2022-10-23
Graveler cache repository information

### LINKED ISSUE #4425: Cache lakeFS repository object
It is currently being fetched by lakeFS for every request (batched for 3ms, still a lot). It's not editable. We should cache it for short time (<5 seconds) to reduce the amount of calls to the KV store.

nopcoder:
Close #4425 

Cache repository information in Graveler ref manager kv implementation.

### DIFF
--- a/pkg/api/controller_test.go
+++ b/pkg/api/controller_test.go
@@ -31,7 +31,6 @@ import (
 	"github.com/treeverse/lakefs/pkg/catalog"
 	"github.com/treeverse/lakefs/pkg/catalog/testutils"
 	"github.com/treeverse/lakefs/pkg/graveler"
-	"github.com/treeverse/lakefs/pkg/graveler/ref"
 	"github.com/treeverse/lakefs/pkg/httputil"
 	"github.com/treeverse/lakefs/pkg/stats"
 	"github.com/treeverse/lakefs/pkg/testutil"
@@ -919,23 +918,24 @@ func TestController_DeleteRepositoryHandler(t *testing.T) {
 	ctx := context.Background()
 
 	t.Run("delete repo success", func(t *testing.T) {
-		_, err := deps.catalog.CreateRepository(ctx, "my-new-repo", onBlock(deps, "foo1"), "main")
+		repo := testUniqueRepoName()
+		_, err := deps.catalog.CreateRepository(ctx, repo, onBlock(deps, "foo1"), "main")
 		testutil.Must(t, err)
 
-		resp, err := clt.DeleteRepositoryWithResponse(ctx, "my-new-repo")
+		resp, err := clt.DeleteRepositoryWithResponse(ctx, repo)
 		verifyResponseOK(t, resp, err)
 
-		// wait for cache expiry
-		time.Sleep(ref.RepositoryCacheExpiry + ref.RepositoryCacheJitter + time.Second)
-
-		_, err = deps.catalog.GetRepository(ctx, "my-new-repo")
-		if !errors.Is(err, catalog.ErrNotFound) {
-			t.Fatalf("expected repo to be gone, instead got error: %v", err)
+		// delete again to expect repository not found
+		resp, err = clt.DeleteRepositoryWithResponse(ctx, repo)
+		testutil.Must(t, err)
+		if resp.JSON404 == nil {
+			t.Fatalf("expected repository to be gone (404), instead got status: %s", resp.Status())
 		}
 	})
 
 	t.Run("delete repo doesnt exist", func(t *testing.T) {
-		resp, err := clt.DeleteRepositoryWithResponse(ctx, "my-other-repo")
+		repo := testUniqueRepoName()
+		resp, err := clt.DeleteRepositoryWithResponse(ctx, repo)
 		testutil.Must(t, err)
 		if resp.StatusCode() == http.StatusOK {
 			t.Fatalf("DeleteRepository should fail on non existing repository, got %d", resp.StatusCode())
--- a/pkg/api/controller_test.go
+++ b/pkg/api/controller_test.go
@@ -914,7 +914,7 @@ func TestController_CreateRepositoryHandler(t *testing.T) {
 }
 
 func TestController_DeleteRepositoryHandler(t *testing.T) {
-	clt, deps := setupClientWithAdmin(t)
+	clt, deps := setupClientWithAdmin(t, WithRepositoryCache(false))
 	ctx := context.Background()
 
 	t.Run("delete repo success", func(t *testing.T) {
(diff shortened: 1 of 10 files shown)


pkg/catalog/catalog.go
```diff
@@ -107,9 +107,10 @@ const (
 )
 
 type Config struct {
-	Config        *config.Config
-	KVStore       *kv.StoreMessage
-	WalkerFactory WalkerFactory
+	Config             *config.Config
+	KVStore            *kv.StoreMessage
+	WalkerFactory      WalkerFactory
+	UseRepositoryCache bool
```
N-o-Z, line 113: Why is this a configurable value? Where do you set it to true (other than tests)

N-o-Z (commented):
Cool addition,
though I didn't quite understand where we enable it in the server.
Also, not quite sure this should be configurable in catalog. If it was just for test enabling purposes, I think there are better options for that

nopcoder:
> Cool addition, though I didn't quite understand where we enable it in the server. Also, not quite sure this should be configurable in catalog. If it was just for test enabling purposes, I think there are better options for that

Here https://github.com/treeverse/lakeFS/pull/4439/files#diff-dd51eee5d8ad1d57c5e785105ca5a0556ed5e2119cf0173e111f992e80a4dcb9R84 we set the default which is true.
At first I've enable the configuration of with/without cache at the ref manager level for testing.
But, when we test the API in the controller's tests we verify that when we delete a repository, we will get not found on the following get request.
In order to support this test and the fact that we setup the test by configure a local handler that setup the our catalog as part of the implementation - I've added a configuration option at the catalog level.
There https://github.com/treeverse/lakeFS/pull/4439/files#diff-13e18d34ca6c711c206f908ee31c4ca09844bd36684bb628b22fa994e2b940eeR180 we also set the default to true for the catalog level.

pkg/catalog/catalog.go
nopcoder: Added comment explaining why we have a catalog level configuration that configure the ref manager's level.

N-o-Z: Why are we not just passing a cache object to the ref manager? That is our convention for managers for the most part

nopcoder: At this level for simplicity as we are going to add more caches, enabling it using boolean looks like a good first step.
Next we can configure each one using application level settings.

pkg/graveler/ref/manager_test.go
```diff
@@ -12,14 +12,39 @@ import (
...
+	"github.com/treeverse/lakefs/pkg/batch"
 	"github.com/treeverse/lakefs/pkg/graveler"
+	"github.com/treeverse/lakefs/pkg/graveler/ref"
 	"github.com/treeverse/lakefs/pkg/ident"
 	"github.com/treeverse/lakefs/pkg/kv"
+	"github.com/treeverse/lakefs/pkg/kv/mock"
 	"github.com/treeverse/lakefs/pkg/testutil"
 )
 
+// TestManager_GetRepositoryCache test get repository information while using cache. Match the number of times we
```
N-o-Z, line 27: Maybe we can add to the test another call after 5 seconds and verify that `Get` was called again?

pkg/api/controller_test.go
```diff
@@ -924,9 +925,12 @@ func TestController_DeleteRepositoryHandler(t *testing.T) {
 		resp, err := clt.DeleteRepositoryWithResponse(ctx, "my-new-repo")
 		verifyResponseOK(t, resp, err)
 
+		// wait for cache expiry
+		time.Sleep(ref.RepositoryCacheExpiry + ref.RepositoryCacheJitter + time.Second)
```
itaiad200, line 929: That's a lot of time to sleep in a single test. Let's make it configurable for the tests?

N-o-Z (approved):
Looks great!
One suggestion regarding test scenario.
One comment regarding the formatting changes - I think it would be much better in the future to do these changes in a separate issue, as it makes it much harder to concentrate on the actual changes.
Thanks!

pkg/graveler/ref/manager.go
```diff
@@ -58,46 +66,58 @@ func protoFromBranch(branchID graveler.BranchID, b *graveler.Branch) *graveler.B
...
+			return nil, err
+		}
+
+		switch repo.State {
+		case graveler.RepositoryState_ACTIVE:
+			return repo, nil
+		case graveler.RepositoryState_IN_DELETION:
+			return nil, graveler.ErrRepositoryInDeletion
+		default:
+			return nil, fmt.Errorf("invalid repository state (%d) rec: %w", repo.State, graveler.ErrInvalid)
```
itaiad200, line 114: Is a number for a state indicative in the returned error?

pkg/graveler/ref/manager_test.go
```diff
@@ -161,6 +186,9 @@ func TestManager_DeleteRepository(t *testing.T) {
 			t.Fatalf("unexpected error: %v", err)
 		}
 
+		// wait for cache expiry
+		time.Sleep(ref.RepositoryCacheExpiry + ref.RepositoryCacheJitter + time.Second)
```
itaiad200, line 190: So long....

pkg/api/controller_test.go
itaiad200: Also if delete repo isn't using the cache then why wait?

itaiad200 (approved):
Please avoid sleeping in tests

pkg/graveler/ref/manager_test.go
N-o-Z: Could be avoided if provided a unique repo name

pkg/api/controller_test.go
N-o-Z: I don't think it's that bad to add this sleep to only one test. Feels like adding a configuration just for the sake of a single test is a far worse option

nopcoder: As long as there is no way to control the cache time or enable/disable - will need to wait while working with GetRepository.
Can make DeleteRepository skip the cache mechanism and write the test using two DeleteRepository calls to remove the sleep here.

pkg/graveler/ref/manager.go
nopcoder: 1. Just made the format explict - it was a number
2. When we will get this error it means we have a bug and we should handle the new repository state here or add a new error UnknownState and drop this format.
3. 
Let me know what you think I will handle it in a different PR.

pkg/graveler/ref/manager_test.go
nopcoder: Can't be avoided - this is the test - delete + verify + create again.

itaiad200: We can have the cache configuration injected and configure it to 0 for tests.

nopcoder:
Configured cache parameters as part of ref manager to enable less sleep time while running tests.

PR #4442, opened 2022-10-23
Add kv metrics

### LINKED ISSUE #4440: KV store metrics
KV stores are not monitored due to lack of metrics. It's best if we wrap the store with metrics for better observability.

itaiad200:
Closes #4440

### DIFF
--- a/pkg/kv/store.go
+++ b/pkg/kv/store.go
@@ -37,7 +37,7 @@ func FormatPath(p ...string) string {
 	return strings.Join(p, PathDelimiter)
 }
 
-// Driver is the interface to access a kv database as a Store.
+// Driver is the interface to access a kv database as a store.
 // Each kv provider implements a Driver.
 type Driver interface {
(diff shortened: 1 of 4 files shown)


pkg/kv/store.go
```diff
@@ -146,7 +146,7 @@ func UnregisterAllDrivers() {
 	}
 }
 
-// Open lookup driver with 'name' and return Store based on 'dsn' (data source name).
+// Open lookup driver with 'name' and return store based on 'dsn' (data source name).
```
N-o-Z, line 149: If you already modified it - you can remove the DSN reference as it is irrelevant

N-o-Z (approved):
Very nice! 😎 

One minor comment

nopcoder (changes requested):

nopcoder (approved):
LGTM! thanks - small comment about the buckets - I think 30 and 60 are something we can remove, above 10 seconds for a request seems too much.

PR #4445, opened 2022-10-24
Add S3 gateway HeadBucket test

arielshaqed:


johnnyaug (approved):

arielshaqed:
Rebased back on top of trunk -- now with the previous PR #4441, which is required for this Esti to compile.

Will push once tests pass, there is no new code with that and It Works on My Machine&trade;

PR #4462, opened 2022-10-25
Fix: remove unused diff_type from diffRefs operation

### LINKED ISSUE #4461: diffRefs API has type, diff_type params but uses only type
diffRefs defines both:

* a "type" parameter, an unrestricted string;
* a "diff_type" parameter, a string enum of two_dot and three_dot.

But it ignores diff_type and uses only type, which should take one of the
enum values supposedly checked on diff_type.

## `diffRefs` Swagger definition

``` swagger
  /repositories/{repository}/refs/{leftRef}/diff/{rightRef}:
    parameters:
      - in: path
        name: repository
        required: true
        schema:
          type: string
      - in: path
        name: leftRef
        required: true
        schema:
          type: string
        description: a reference (could be either a branch or a commit ID)
      - in: path
        name: rightRef
        required: true
        schema:
          type: string
        description: a reference (could be either a branch or a commit ID) to compare against
      - $ref: "#/components/parameters/PaginationAfter"
      - $ref: "#/components/parameters/PaginationAmount"
      - $ref: "#/components/parameters/PaginationPrefix"
      - $ref: "#/components/parameters/PaginationDelimiter"
      - in: query
        name: type
        schema:
          type: string
      - in: query
        name: diff_type
        schema:
          type: string
          enum: [two_dot, three_dot]
          default: three_dot

    get:
      tags:
        - refs
      operationId: diffRefs
      summary: diff references
      responses:
        200:
          description: diff between refs
          content:
            application/json:
              schema:
                $ref: "#/components/schemas/DiffList"
        401:
          $ref: "#/components/responses/Unauthorized"
        404:
        default:
```

nopcoder:
The current code uses 'type' to control which diff type we like to perform.

Fix #4461

### DIFF
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -2563,10 +2563,6 @@ paths:
         name: type
         schema:
           type: string
-      - in: query
-        name: diff_type
-        schema:
-          type: string
           enum: [two_dot, three_dot]
           default: three_dot
 
--- a/clients/java/api/openapi.yaml
+++ b/clients/java/api/openapi.yaml
@@ -2792,13 +2792,6 @@ paths:
         name: type
         required: false
         schema:
-          type: string
-        style: form
-      - explode: true
-        in: query
-        name: diff_type
-        required: false
-        schema:
           default: three_dot
           enum:
           - two_dot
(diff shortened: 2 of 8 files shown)


itaiad200 (approved):
LGTM

PR #4467, opened 2022-10-25
Fix deprecated use of io/ioutil package

nopcoder:
Update code not to use deprecated `io/ioutil`.
+ Additional code format.

Required in order to move to Go >1.17

### DIFF
--- a/cmd/lakectl/cmd/dbt_util_test.go
+++ b/cmd/lakectl/cmd/dbt_util_test.go
@@ -143,8 +143,8 @@ func TestDbtLsToJson(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.Name, func(t *testing.T) {
 			dbtResources, err := cmd.DBTLsToJSON(tt.Args.ProjectRoot, tt.Args.ResourceType, tt.Args.SelectValues, tt.Args.Executor.ExecuteCommand)
-			if (err != nil) && !errors.As(err, &tt.ErrType) {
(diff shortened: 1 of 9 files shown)


cmd/lakectl/cmd/dbt_util_test.go
```diff
@@ -143,8 +143,8 @@ func TestDbtLsToJson(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.Name, func(t *testing.T) {
 			dbtResources, err := cmd.DBTLsToJSON(tt.Args.ProjectRoot, tt.Args.ResourceType, tt.Args.SelectValues, tt.Args.Executor.ExecuteCommand)
-			if (err != nil) && !errors.As(err, &tt.ErrType) {
-				t.Errorf("DBTLsToJSON() error = %v, expected ErrType: %v", err, tt.ErrType)
+			if err != nil && reflect.TypeOf(err) != reflect.TypeOf(tt.ErrType) {
```
itaiad200, line 146: Why does `errors` package not good here?

itaiad200 (approved):
LGTM

cmd/lakectl/cmd/dbt_util_test.go
nopcoder: The call for `errors.As` expect to have a type it should map `err` into - in this case the type for all tests is `error`.
I assumed the test writer wanted to match the error type as suggested by the name 'ErrType'.

PR #4473, opened 2022-10-26
Uncommitted GC execution plan

### LINKED ISSUE #1933: Hard-delete objects that were never committed

N-o-Z:
Closes #1933

### DIFF
--- a/design/open/gc-plus-execution-plan.md
+++ b/design/open/gc-plus-execution-plan.md
@@ -1,0 +1,80 @@
+# Uncommitted Garbage Collection - Execution Plan
+
+Uncommitted Garbage Collection [Proposal](link-to-proposal)
+
+## Overview
+
+Implementing uncommitted GC can be divided into 3 main changes:
+1. lakeFS additions and modifications
+2. Implementing uncommitted GC logic on client side
+3. Modifying existing GC
+
+All changes in lakeFS are independent changes, work on lakeFS changes can start immediately.
+Some GC changes are dependent on lakeFS changes, these are prioritized as part of the plan.  
+All tasks are tracked under the [GC+ label](https://github.com/treeverse/lakeFS/labels/GC%2B) in GitHub
+
+## Changes on lakeFS
+
+The following are a list of changes in lakeFS by order of priority, according the dependency constraints 
+
+1. Implement PrepareUncommittedForGC API
+   - Creates metaranges and ranges for a selected branch's staging area, and writes them to a designated GC path
+   - After this is done, we can implement new uncommitted logic in GC
+
+2. Modify [Get/Link]PhysicalAddress
+    - Issue validation token on GetPhysicalAddress with an expiry time and add it to valid token list
+    - LinkPhysicalAddress will validate token and remove it from valid token list
+
+3. Implement Move/RenameObject API
+    - Get entry from branch
+    - Stage new entry using old entry data and new path 
+    - Perform delete for old entry (write tombstone)
+
+4. Modify Gateway CopyObject API
+    - For copy operations on different branches, change CopyObject behavior to perform "full copy" using the underlying 
+   storage adapter's Copy method.
+
+5. Modify lakeFSFS renameObject method
+    - Use CopyObject + DeleteObject when working via Gateway
+    - Use RenameObject when working with OpenAPI
+
+6. lakeFS clients
+    - Any action required on clients?
+    - Do we need to make them version aware?
+    - Should we issue a notification to upgrade clients?
+
+7. Modify StageObject API
+    - Return error if given address is inside repository namespace
+
+8. Implement new repository structure
+    - Root prefix for new lakeFS data - `data/`
+    - Use slice naming conventions as defined in proposal to create object path
+    - Track object upload count
+    - Create new slice by time or count
+
+## Changes on GC
+
+- New logic in GC to support uncommitted garbage collection
+  - Create uncommitted metaranges
+  - Ingest uncommitted data from lakeFS
+  - Write uncommitted data to the object store (as part of the reports)
+  - Implement the optimized run deltas
+- Incorporate uncommitted changes into the current GC flow
+  - Implement Run ID proposal + [changes](https://github.com/treeverse/lakeFS/issues/4469)
+  - Add uncommitted options to GC rules
+  - Add additional metadata to the GC report (last scanned slice)
+  - Support mark and sweep for uncommitted 
+
+## Testing
+
+- Functionality - Basic GC functionality, check that committed GC logic not broken, uncommitted GC works as expected
+- Data Integrity - Ensure only objects that are eligible for deletion are hard deleted from repository
+- Performance - Verify uncommitted GC performance requirements are kept
+- Existing repository - test first run on an old repository, with and without objects in new repository structure. Ensure consecutive runs work as expected (in Optimized Run)
+
+## Milestone
+
+- Implement lakeFS changes
+- Implement GC changes - Clean Run with the minimal performance requirements met
+- Implement GC changes - Optimized Run
+- Integration testing
--- a/design/open/gc-plus-execution-plan.md
+++ b/design/open/gc-plus-execution-plan.md
@@ -56,7 +56,7 @@ The following are a list of changes in lakeFS by order of priority, according th
 
 - New logic in GC to support uncommitted garbage collection
   - Create uncommitted metaranges
-  - Ingest uncommitted data from lakeFS
+  - Read uncommitted data from lakeFS
   - Write uncommitted data to the object store (as part of the reports)
   - Implement the optimized run deltas
 - Incorporate uncommitted changes into the current GC flow
--- a/design/open/gc-plus-execution-plan.md
+++ b/design/open/gc-plus-execution-plan.md
@@ -61,7 +61,6 @@ The following are a list of changes in lakeFS by order of priority, according th
   - Implement the optimized run deltas
 - Incorporate uncommitted changes into the current GC flow
   - Implement Run ID proposal + [changes](https://github.com/treeverse/lakeFS/issues/4469)
-  - Add uncommitted options to GC rules
   - Add additional metadata to the GC report (last scanned slice)
   - Support mark and sweep for uncommitted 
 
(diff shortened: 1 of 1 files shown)


itaiad200 (changes requested):
Checking

design/open/gc-plus-execution-plan.md
```diff
@@ -0,0 +1,79 @@
...
+## Changes on GC
+
+- New logic in GC to support uncommitted garbage collection
+  - Create uncommitted metaranges
+  - Ingest uncommitted data from lakeFS
+  - Write uncommitted data to the object store (as part of the reports)
+  - Implement the optimized run deltas
+- Incorporate uncommitted changes into the current GC flow
+  - Implement Run ID proposal + [changes](https://github.com/treeverse/lakeFS/issues/4469)
+  - Add uncommitted options to GC rules
```
johnnyaug, line 64: Opinion: this should not be part of the GC rules. In my view, these are two different features. We chose to tie them together mainly for implementation reasons.

johnnyaug (commented):
Sounds like a solid plan!

design/open/gc-plus-execution-plan.md
```diff
@@ -0,0 +1,79 @@
...
+6. lakeFS clients
+    - Any action required on clients?
+    - Do we need to make them version aware?
+    - Should we issue a notification to upgrade clients?
+
+7. Modify StageObject API
+    - Return error if given address is inside repository namespace
+
+8. Implement new repository structure
+    - Root prefix for new lakeFS data - `data/`
```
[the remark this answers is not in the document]
N-o-Z: It's  an amendment in the proposal which mitigates handling old repository structures. See "1. Listing namespace objects" section in proposal

design/open/gc-plus-execution-plan.md
N-o-Z: I phrased this item as "Add" but it's more of a open/discovery question. We should understand what are the use cases and user requirements.
IMHO, uncommitted should run on each GC run, no filters, no exceptions and in that case, this means we shouldn't add any specific reference to uncommitted in the GC rules.
But I'm sure there are other opinions

design/open/gc-plus-execution-plan.md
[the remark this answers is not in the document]
N-o-Z: Updated the Path Conventions section for clarity

design/open/gc-plus-execution-plan.md
```diff
@@ -0,0 +1,79 @@
...
+
+The following are a list of changes in lakeFS by order of priority, according the dependency constraints 
+
+1. Implement PrepareUncommittedForGC API
+   - Creates metaranges and ranges for a selected branch's staging area, and writes them to a designated GC path
+   - After this is done, we can implement new uncommitted logic in GC
+
+2. Modify [Get/Link]PhysicalAddress
+    - Issue validation token on GetPhysicalAddress with an expiry time and add it to valid token list
+    - LinkPhysicalAddress will validate token and remove it from valid token list
```
itaiad200, line 26: This is also a client change, the API for it has changed..

design/open/gc-plus-execution-plan.md
```diff
@@ -0,0 +1,79 @@
...
+
+3. Implement Move/RenameObject API
+    - Get entry from branch
+    - Stage new entry using old entry data and new path 
+    - Perform delete for old entry (write tombstone)
```
itaiad200, line 31: Not sure I understand why we need it at all...

design/open/gc-plus-execution-plan.md
```diff
@@ -0,0 +1,79 @@
...
+
+4. Modify Gateway CopyObject API
+    - For copy operations on different branches, change CopyObject behavior to perform "full copy" using the underlying 
+   storage adapter's Copy method.
+
+5. Modify lakeFSFS renameObject method
+    - Use CopyObject + DeleteObject when working via Gateway
+    - Use RenameObject when working with OpenAPI
```
itaiad200, line 39: Same as above - why do we need it for uncommitted GC?

design/open/gc-plus-execution-plan.md
```diff
@@ -0,0 +1,79 @@
...
+    - Should we issue a notification to upgrade clients?
+
+7. Modify StageObject API
+    - Return error if given address is inside repository namespace
+
+8. Implement new repository structure
+    - Root prefix for new lakeFS data - `data/`
+    - Use slice naming conventions as defined in proposal to create object path
+    - Track object upload count
+    - Create new slice by time or count
```
itaiad200, line 53: and by lakeFS instance

design/open/gc-plus-execution-plan.md
```diff
@@ -0,0 +1,79 @@
...
+
+## Changes on GC
+
+- New logic in GC to support uncommitted garbage collection
+  - Create uncommitted metaranges
+  - Ingest uncommitted data from lakeFS
+  - Write uncommitted data to the object store (as part of the reports)
```
itaiad200, line 60: Now that I think about it again, why do we need to write the data twice? Can't the first write use the format the job actually need?

design/open/gc-plus-execution-plan.md
```diff
@@ -0,0 +1,79 @@
...
+- Functionality - Basic GC functionality, check that committed GC logic not broken, uncommitted GC works as expected
+- Data Integrity - Ensure only objects that are eligible for deletion are hard deleted from repository
+- Performance - Verify uncommitted GC performance requirements are kept
+- Existing repository - test first run on an old repository, with and without objects in new repository structure. Ensure consecutive runs work as expected (in Optimized Run)
+
+## Milestone
+
+- Implement Clean Run (including all lakeFS change) with minimal performance requirements met
+- Implement GC changes - Optimized Run + mark and sweep
+- Deployment to lakeFS Cloud
```
itaiad200, line 79: What tasks are included in each milestone?

design/open/gc-plus-execution-plan.md
itaiad200: Let's open a discovery issue and remove it from the execution plan then?

design/open/gc-plus-execution-plan.md
```diff
@@ -0,0 +1,79 @@
...
+  - Add uncommitted options to GC rules
+  - Add additional metadata to the GC report (last scanned slice)
+  - Support mark and sweep for uncommitted 
+
+## Testing
+
+- Functionality - Basic GC functionality, check that committed GC logic not broken, uncommitted GC works as expected
+- Data Integrity - Ensure only objects that are eligible for deletion are hard deleted from repository
+- Performance - Verify uncommitted GC performance requirements are kept
+- Existing repository - test first run on an old repository, with and without objects in new repository structure. Ensure consecutive runs work as expected (in Optimized Run)
```
itaiad200, line 73: Are these 4 separate tasks? I assume you can write a task to cover 1-3 and maybe even 4. Let's switch it then to list the actual task and not why they are covering.

itaiad200 (changes requested):
Thanks! I can almost see the uncommitted GC if I close my eyes 👀

design/open/gc-plus-execution-plan.md
N-o-Z: Not necessarily - if we embed the token into the path somehow the API doesn't have to change

design/open/gc-plus-execution-plan.md
N-o-Z: I think this comment belongs to the proposal PR (and I think there are also answers there)

design/open/gc-plus-execution-plan.md
N-o-Z: Same

design/open/gc-plus-execution-plan.md
N-o-Z: We are using the same method used today to get the committed objects via RangeGetter

design/open/gc-plus-execution-plan.md
N-o-Z: These are not tasks - these are testing areas and emphasis.

design/open/gc-plus-execution-plan.md
N-o-Z: Done

design/open/gc-plus-execution-plan.md
talSofer: > Deployment to lakeFS Cloud

The milestone is larger then that - it would be first to understand how to do that and answer multiple questions like - should we use the committed MGC infra or a different approach? what frequency? how do we get observability? etc

design/open/gc-plus-execution-plan.md
```diff
@@ -0,0 +1,78 @@
...
+
+8. Implement new repository structure
+    - Root prefix for new lakeFS data - `data/`
+    - Use slice naming conventions as defined in proposal to create object path
+    - Track object upload count
+    - Create new slice by time or count
+
+## Changes on GC
+
+- New logic in GC to support uncommitted garbage collection
```
talSofer, line 57: I just want to confirm that we acknowledge the need for a design discussion before we implement any change. I mean to answer the question of how to fit the uncommitted GC implementation into the existing GC implementation; are they two separate Spark apps? what are the implications on performance, observability, and stability of the products? how does the user runs each job, product-wise - can the jobs be separated from the user's perspective? etc.

PR #4483, opened 2022-10-26
Create branch & tag conflicting name validations

### LINKED ISSUE #2667: Do not allow a branch and a tag with the same name
Reasons:
1. It is never a good idea to have a branch and a tag with the same name.
2. Although git allows this, it also offers a way to refer to each of the branch/tag explicitly:
    `git checkout refs/tags/example` will checkout the tag `example` even if a branch by this name exists. lakeFS does not offer something similar, so you will basically never be able to access a tag if a branch with the name already exists.

Therefore, let's block this option.
Note: solving this should also take into account that there may already be branches/tags with the same name on existing installations.

### LINKED ISSUE #4468: Do not allow tag name that uses slash (/)
Using tag with slash will prevent the use of the tag as part object access path - the slash delimiter is used to separate between the repository, ref (branch, tag, commit) and the object path.
We should block block the create as done for branch name.

itaiad200:
closes #4468 
closes #2667

### DIFF
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -415,11 +415,10 @@ func (c *Catalog) CreateBranch(ctx context.Context, repositoryID string, branch 
 	if err != nil {
 		return nil, err
 	}
-	if _, err := c.Store.GetCommit(ctx, repository, graveler.CommitID(branchID)); err == nil {
-		return nil, fmt.Errorf("commit ID %s: %w", branchID, graveler.ErrConflictFound)
-	} else if !errors.Is(err, graveler.ErrNotFound) {
+	if err := c.checkCommitIDDuplication(ctx, repository, graveler.CommitID(branchID)); err != nil {
 		return nil, err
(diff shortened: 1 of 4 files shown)


pkg/catalog/catalog.go
```diff
@@ -415,6 +415,16 @@ func (c *Catalog) CreateBranch(ctx context.Context, repositoryID string, branch
 	if err != nil {
 		return nil, err
 	}
+	if _, err := c.Store.GetCommit(ctx, repository, graveler.CommitID(branchID)); err == nil {
```
N-o-Z, line 418: We can consolidate the duplicate code into a validation function.

N-o-Z (changes requested):
Looks good!
One comment

pkg/catalog/catalog.go
itaiad200: Done.

N-o-Z (approved):

PR #4495, opened 2022-10-27
Cache unfound settings too

### LINKED ISSUE #4494: Settings cache doesn't cache NotFound results
For repos without branch protections for example, every write request has to reach to the kv store to find nothing.

itaiad200:
closes #4494

### DIFF
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -19,7 +19,6 @@ import (
 	"github.com/treeverse/lakefs/pkg/batch"
 	"github.com/treeverse/lakefs/pkg/block"
 	"github.com/treeverse/lakefs/pkg/block/factory"
-	"github.com/treeverse/lakefs/pkg/cache"
 	"github.com/treeverse/lakefs/pkg/config"
 	"github.com/treeverse/lakefs/pkg/graveler"
 	"github.com/treeverse/lakefs/pkg/graveler/branch"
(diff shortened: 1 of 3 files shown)


pkg/catalog/catalog.go
```diff
@@ -204,6 +212,10 @@ func New(ctx context.Context, cfg Config) (*Catalog, error) {
 	refManager := ref.NewKVRefManager(executor, *cfg.KVStore, ident.NewHexAddressProvider(), nil)
 	gcManager := retention.NewGarbageCollectionManager(tierFSParams.Adapter, refManager, cfg.Config.GetCommittedBlockStoragePrefix())
 	settingManager := settings.NewManager(refManager, *cfg.KVStore)
+	if cfg.CacheFactory != nil {
```
nopcoder, line 215: Prefer the default to use the default cache and override the cache settings in the specific test.

nopcoder (changes requested):

pkg/catalog/catalog.go
itaiad200: Notice that there's no dependency injection to the `Catalog`. It creates its dependencies. Tests can't override the settings since they do not create the `SettingsManager`

pkg/catalog/catalog.go
```diff
@@ -110,6 +111,13 @@ type Config struct {
 	Config        *config.Config
 	KVStore       *kv.StoreMessage
 	WalkerFactory WalkerFactory
+	CacheFactory  CacheFactory
+}
+
+// CacheFactory provides an abstraction for creating cache.
+// Use in tests to override
+type CacheFactory interface {
```
eden-ohana, line 119: Why factory is needed? we can initialize the cache to NoCache

pkg/catalog/catalog.go
nopcoder: Prefer we keep passing options (WithStagingCache) that can override the defaults.
I also think that having Config here is a shortcut.

nopcoder (changes requested):
We wanted a way to control the cache parameters and not switch implementations for the tests.

pkg/catalog/catalog.go
itaiad200: Done.

nopcoder (approved):
Approved and going to open an issue to simplify the composition of catalog that will enable configuration without passing our Config structure.

PR #4497, opened 2022-10-27
Cache commit on Graveler ref manager

### LINKED ISSUE #4422: Cache Commits in Graveler
Commits are an immutable object. We can cache the recently used commits in the server and avoid the roundtrip to the DB whenever requested to fetch a commit. Even a small cache (~1000, which translates to <1M) should reduce the number of calls to the database.

nopcoder:
Close #4422

### DIFF
--- a/pkg/graveler/ref/manager.go
+++ b/pkg/graveler/ref/manager.go
@@ -33,7 +33,7 @@ const (
 	DefaultRepositoryCacheJitter = DefaultRepositoryCacheExpiry / 2
 
 	DefaultCommitCacheSize   = 1000
-	DefaultCommitCacheExpiry = 5 * time.Second
+	DefaultCommitCacheExpiry = 30 * time.Second
 	DefaultCommitCacheJitter = DefaultRepositoryCacheExpiry / 2
 )
 
@@ -42,6 +42,20 @@ type CacheConfig struct {
 	Expiry time.Duration
 	Jitter time.Duration
 }
+
+var (
+	DefaultRepositoryCacheConfig = &CacheConfig{
+		Size:   DefaultRepositoryCacheSize,
+		Expiry: DefaultRepositoryCacheExpiry,
+		Jitter: DefaultRepositoryCacheJitter,
+	}
+
+	DefaultCommitCacheConfig = &CacheConfig{
+		Size:   DefaultCommitCacheSize,
+		Expiry: DefaultCommitCacheExpiry,
+		Jitter: DefaultCommitCacheJitter,
+	}
+)
 
 type KVManager struct {
 	kvStore         *kv.StoreMessage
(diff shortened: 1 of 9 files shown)


pkg/graveler/ref/manager.go
```diff
@@ -31,19 +31,24 @@ const (
 	DefaultRepositoryCacheSize   = 1000
 	DefaultRepositoryCacheExpiry = 5 * time.Second
 	DefaultRepositoryCacheJitter = DefaultRepositoryCacheExpiry / 2
+
+	DefaultCommitCacheSize   = 1000
+	DefaultCommitCacheExpiry = 5 * time.Second
```
itaiad200, line 36: Let's increase it to 30 seconds? It's immutable anyhow..

pkg/graveler/ref/manager.go
```diff
@@ -85,11 +100,27 @@ func NewKVRefManager(executor batch.Batcher, kvStore kv.StoreMessage, addressPro
...
+		commitCacheConfig = &CacheConfig{
+			Size:   DefaultCommitCacheSize,
+			Expiry: DefaultCommitCacheExpiry,
+			Jitter: DefaultCommitCacheJitter,
+		}
+	}
+	commitCache := cache.NoCache
+	if commitCacheConfig.Size > 0 {
+		commitCache = cache.NewCache(commitCacheConfig.Size, commitCacheConfig.Expiry, cache.NewJitterFn(commitCacheConfig.Jitter))
+	}
```
itaiad200, line 116: Minor: Duplicated logic to shared location

pkg/graveler/ref/manager.go
```diff
@@ -456,23 +487,38 @@ func (m *KVManager) GetCommitByPrefix(ctx context.Context, repository *graveler.
 }
 
 func (m *KVManager) GetCommit(ctx context.Context, repository *graveler.RepositoryRecord, commitID graveler.CommitID) (*graveler.Commit, error) {
+	key := fmt.Sprintf("%s:%s", repository.RepositoryID, commitID)
+	v, err := m.repoCache.GetOrSet(key, func() (v interface{}, err error) {
+		return m.getCommitBatch(ctx, repository, commitID)
```
itaiad200, line 492: It won't cache NotFound commits. I can't remember if we expect this to happen a lot. The chances of a user pointing directly to a commit hash which wasn't created yet is not big. However, we accept refs which could be anything. If we currently check that `main` isn't a commit, I think we should avoid that constant call to the store.

itaiad200 (changes requested):
Great tests! Only blocking comment regarding unfound commit

eden-ohana (approved):

pkg/graveler/ref/manager.go
nopcoder: I think that full commit ID for non existing is something that probably will not be part of any main flow. Anything short or resolved from branch or tag should be an existing commit.
So filling the cache with non existing commit for a case that a user/app try to identify if commit exits I assume it will be something I don't want to return answer from the last 30sec.
Keeping it this way.

itaiad200: @nopcoder missed this one?

nopcoder: Will open an issue/pr to handle optimization of branch vs commit prefix.

pkg/graveler/ref/manager_test.go
```diff
@@ -33,15 +33,22 @@ func TestManager_GetRepositoryCache(t *testing.T) {
...
 	ctrl := gomock.NewController(t)
 	mockStore := mock.NewMockStore(ctrl)
-	storeMessage := kv.StoreMessage{Store: mockStore}
+	storeMessage := &kv.StoreMessage{Store: mockStore}
 	ctx := context.Background()
 	mockStore.EXPECT().Get(ctx, []byte("graveler"), []byte("repos/repo1")).Times(times).Return(&kv.ValueWithPredicate{}, nil)
-	repoCacheConfig := &ref.RepositoryCacheConfig{
+	cacheConfig := &ref.CacheConfig{
 		Size:   100,
 		Expiry: 2 * time.Second,
```
itaiad200, line 41: If we're going to `sleep` in tests, why not 20 milliseconds?

itaiad200 (approved):

PR #4501, opened 2022-10-30
Use go run for go:generate instead of download tools by makefile

nopcoder:
Remove download part of build process when using mockgen and oapi-gen to build code using go:generate.
By using `go run` the tool will be downloaded and cache once.


eden-ohana (approved):
Cool

PR #4512, opened 2022-10-30
GC: Refactor GC permissions name

### LINKED ISSUE #4500: GC+: Modify PrepareGarbageCollectionCommitsAction permission name
Suggest to modify `PrepareGarbageCollectionCommitsAction` to `PrepareGarbageCollectionAction` and reuse it for both the prepare commits and prepare uncommitted actions instead of creating two different permissions

N-o-Z:
Closes #4500 
@treeverse/ecosystem  FYI,

Not sure about the complete implications of this change - please give your opinion.
The other option is to create separate actions for prepare commits and prepare uncommitted - which makes a lot less sense in my opinion.


talSofer:
Thanks for making sure team @treeverse/ecosystem is in the loop on this one!

This change will break managed GC for the following reason - 
Each org in lakeFS cloud has a gc user associated with it, and this gc user has the following policy attached to it 
Screen Shot 2022-10-31 at 8 25 07
The policy includes everything the user needs to run GC. 

To your question, I also vote to creating separate permissions for prepare commits and prepare uncommitted because they are logically different.

N-o-Z:
> Thanks for making sure team @treeverse/ecosystem is in the loop on this one!
> 
> This change will break managed GC for the following reason - Each org in lakeFS cloud has a gc user associated with it, and this gc user has the following policy attached to it Screen Shot 2022-10-31 at 8 25 07 The policy includes everything the user needs to run GC.
> 
> To your question, I also vote to create separate permissions for prepare commits and prepare uncommitted because they are logically different. although you can rename but handle what I mentioned above

Thanks Talk for the input!

Obviously changing permission name means breaking the users who use these permissions. IMHO this change is needed. I think that the name given to the Action was inaccurate regardless, since this permission is de-facto for running GC and not for prepare commits - an operation that seemingly has no sense outside the context of the GC process. In addition adding another Action will burden the user with configuring additional permission for something that is going to be a single process

Jonathan-Rosenberg:
@N-o-Z, as you and @talSofer mentioned, this is a breaking change so it should be accompanied by a PR to change the soon-to-be broken permissions in the appropriate place (port it to existing lakeFS instances)

arielshaqed (commented):
> > Thanks for making sure team @treeverse/ecosystem is in the loop on this one!
[...]
> > To your question, I also vote to create separate permissions for prepare commits and prepare uncommitted because they are logically different. although you can rename but handle what I mentioned above
> 
> Thanks Tal for the input!
> 
> Obviously changing permission name means breaking the users who use these permissions. IMHO this change is needed. I think that the name given to the Action was inaccurate regardless, since this permission is de-facto for running GC and not for prepare commits - an operation that seemingly has no sense outside the context of the GC process. In addition adding another Action will burden the user with configuring additional permission for something that is going to be a single process

Regardless, as it stands this is a breaking change for any lakeFS user who has configured GC -- and _also_ for lakeFS cloud customers.  If we just want to be legalistic... It would require a new major version post-1.0.  Since we are pre-1.0, which is pretty much "anything goes" land, it would require "just" a new minor version and a strongly worded release note.

In the pre-KV world we required a migration for any change that broke existing installations.  I am blocking this change as it needs a similar migration.  Or instead allow _both_ spellings for this permission.

I am not sure how to do it post-KV, probably "just a script".  Pre-KV it would be a slightly nasty SQL statement that adds the new permission for any policy that included the old permission.  For instance [this](https://github.com/treeverse/lakeFS/blob/v0.69.0/pkg/ddl/000032_iam_add_attach_namespace.up.sql) was an extreme example :-)

N-o-Z:
> > > Thanks for making sure team @treeverse/ecosystem is in the loop on this one!
> > > [...]
> > > To your question, I also vote to create separate permissions for prepare commits and prepare uncommitted because they are logically different. although you can rename but handle what I mentioned above
> > 
> > 
> > Thanks Tal for the input!
> > Obviously changing permission name means breaking the users who use these permissions. IMHO this change is needed. I think that the name given to the Action was inaccurate regardless, since this permission is de-facto for running GC and not for prepare commits - an operation that seemingly has no sense outside the context of the GC process. In addition adding another Action will burden the user with configuring additional permission for something that is going to be a single process
> 
> Regardless, as it stands this is a breaking change for any lakeFS user who has configured GC -- and _also_ for lakeFS cloud customers. If we just want to be legalistic... It would require a new major version post-1.0. Since we are pre-1.0, which is pretty much "anything goes" land, it would require "just" a new minor version and a strongly worded release note.
> 
> In the pre-KV world we required a migration for any change that broke existing installations. I am blocking this change as it needs a similar migration. Or instead allow _both_ spellings for this permission.
> 
> I am not sure how to do it post-KV, probably "just a script". Pre-KV it would be a slightly nasty SQL statement that adds the new permission for any policy that included the old permission. For instance [this](https://github.com/treeverse/lakeFS/blob/v0.69.0/pkg/ddl/000032_iam_add_attach_namespace.up.sql) was an extreme example :-)

Thanks @arielshaqed,

As I understand, the only current standing argument not to change the permission name is the need for migration?
I'd like to know if you think there's another argument as to why we need separate permissions for operations that seems to be entirely internal workings of the GC process which the user should not be aware of.
As a user, I would not understand why I need to define 2 different permissions just to run the GC job.
If you also agree that this makes no sense - it might be advisable to do it as soon as possible, while the user base is still small and the implications will be less heavy.

arielshaqed:
[...]

> Thanks @arielshaqed,
> 
> As I understand, the only current standing argument not to change the permission name is the need for migration? I'd like to know if you think there's another argument as to why we need separate permissions for operations that seems to be entirely internal workings of the GC process which the user should not be aware of. As a user, I would not understand why I need to define 2 different permissions just to run the GC job. If you also agree that this makes no sense - it might be advisable to do it as soon as possible, while the user base is still small and the implications will be less heavy.

The only blocker as far as I'm concerned is migration.

About the rest: I have no strong opinions about naming here.  I have a preference *against* unifying permissions, but it is not blocking.

# Possible damage with this permission

Generally I like to have fine-grained permissions.  One way I think of permissions is not only through "what value does this permission add?" but especially through "what damage can someone with the permission do?".  Seen through this lens, the two permissions are very different:
* Prepare committed GC returns all commits.
* Prepare uncommitted GC returns a bunch of object references.
So they are qualitatively different.

# Commit IDs

AFAIU returning all commit IDs does no damage.  I can probably use timing to guess commit IDs merely by using the digest expansion feature to traverse the trie of commits.  I just try to fetch a prefix.  If I get a hit then I found a commit; otherwise I try to distinguish between "no commit with this prefix" and "multiple commits with this prefix" by timing (no commit IDs is 1 failing access, >1 commit ID is >=2 successful fetches -- should be possible to time from nearby).  I can repeat this some fixed number of times, and get a pretty good picture of all commits in time linear in the number of commits.

Once I have a commit, I have metadata access but still no way to access data.

# Objects

We should _not_ need to give users ListObjects permissions on the data bucket, not even for direct-access (lakeFSFS).  So there is literally no way a user can discover valid object paths without prepareUncommittedObjects or something.

If I have an object reference and ReadObject on the bucket, I have full data access to that object.  Even worse, uncommitted objects to be deleted might be exactly those with the data I would most like to protect!

So the two permissions do not appear equivalent.

# Why not 2 permissions?

My preference (again, non-blocking!) would be to keep 2 separate permissions.  Then put both or neither of them on each preconfigured policy.  But configure what each one does.

And yes, we still need an upgrade script to fix existing policies however we decide!

N-o-Z:
@arielshaqed 

AFAIU Non of the APIs return any information but the location of where to read the files containing the data created by the APIs.
To read the files from `_lakefs` you need the correct permissions that have nothing to do with the GC permissions.
According to what you say - anyone with permission to this path can read all the files generated by any GC run anytime and in addition can read the metaranges and ranges for the repository.

I can't say I agree with the justifications given, but since I feel there's a lot of push back - I'm going to close this issue

PR #4513, opened 2022-10-31
Make a working OutputCommitter

arielshaqed:
Tested by writing a 3-partition dataframe as text and as Parquet, nothing comprehensive.

* Use `outputPath` as "job" branch name for writing output.

  **Why?** Spark `FileFormatWriter` writes using _two_ JobIDs: it calls `setupJob`/`commitJob` on the first and `setupTask`/`commitTask` on the second.  So if we name the work branch by "its" JobID we will write all objects to the second and then try to merge the first.  That does not work.  Instead, use `outputPath` which is shared between them as the base for the branch.

  **How?** `outputPath` itself cannot be used as a branch name -- at the very least it will contain slashes.  Instead, use its SHA-256 as the branch name; to make it readable, append some prefix of the branch name with all non-alphanumerics replaced by `-`s.

  **What else?** This _will not work_ in the multiwriter case, because all writers will share the same working branch.
* Clean up DEBUG prints/logs.
* Merging:
  - Use merge strategy to merge back
  - Avoid empty commit/merge API errors: only commit or merge when there are diffs.
* Use timeouts in client.

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSClient.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSClient.java
@@ -31,7 +31,7 @@ public class LakeFSClient {
             throw new IOException("Missing lakeFS secret key");
         }
 
-        ApiClient apiClient = io.lakefs.clients.api.Configuration.getDefaultApiClient();
+        ApiClient apiClient = new ApiClient();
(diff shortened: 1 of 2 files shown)


lynnro314 (approved):
LGTM. Thanks!

arielshaqed:
Thanks!
Pulling :-)

PR #4529, opened 2022-10-31
hadoopfs/lakefs client header

### LINKED ISSUE #4524: Add X-Lakefs-Client header to requests by lakeFS client
Add the header to requests sent to the lakeFS server by the following clients:
1. HadoopFS
2. lakeFS metadata client - GC
3. lakeFS metadata client - Export
4. lakeFS metadata client - other

johnnyaug:
Manually tested with a locally-built jar.

Resolves #4524.

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSClient.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSClient.java
@@ -35,7 +35,7 @@ public class LakeFSClient {
             endpoint = endpoint.substring(0, endpoint.length() - 1);
         }
         apiClient.setBasePath(endpoint);
-
+        apiClient.addDefaultHeader("X-Lakefs-Client", "lakefs-hadoopfs/" + getClass().getPackage().getImplementationVersion());
         HttpBasicAuth basicAuth = (HttpBasicAuth) apiClient.getAuthentication(BASIC_AUTH);
         basicAuth.setUsername(accessKey);
         basicAuth.setPassword(secretKey);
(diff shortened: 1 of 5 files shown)


nopcoder (approved):
Looks good I don't think it will happen - but just in case: based on the docs it may return null so I suggest a simple wrapper to check and return "unknown" or other indicative value.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSClient.java
```diff
@@ -35,7 +35,7 @@ public LakeFSClient(String scheme, Configuration conf) throws IOException {
             endpoint = endpoint.substring(0, endpoint.length() - 1);
         }
         apiClient.setBasePath(endpoint);
-
+        apiClient.addDefaultHeader("X-Lakefs-Client", "lakefs-hadoopfs/" + getClass().getPackage().getImplementationVersion());
```
arielshaqed, line 38: This is a bit inconsistent with just "`metaclient`" in [the metadata client PR](https://github.com/treeverse/lakeFS/pull/4532/files#diff-be6ff7bdfc93e00e2e2acb398b667fce8d03ea393d285d3d41156e3963323653R124).  I probably slightly prefer _that_ one because I see no reason to repeat "lakefs" -- unless we expect others to manage their own client names, in which case I'd expect an or prefix there too...

Either way is fine, but I would rather have them both as similar as possible.

arielshaqed (approved):
Neat, thanks!  But please see note about being more consistent in how we name our clients.

PR #4530, opened 2022-10-31
Upload data new structure for uploaded data physical addresses

### LINKED ISSUE #4481: GC+: Implement new repository structure
Uncommitted GC must scan the bucket in order to find objects that are not referenced by lakeFS.
To optimize this process, suggest the following changes:

1. Divide the repository namespace into time-and-size based slices.
2. Slice name will be a time based, reverse sorted unique identifier.
3. LakeFS will create a new slice on a timely basis (for example: hourly) or when it has written < MAX_SLICE_SIZE > objects to the slice.
4. Each slice will be written by a single lakeFS instance in order to track slice size.
5. The sorted slices will enable partial scans of the bucket when running the optimized GC.

nopcoder:
Close #4481

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -71,10 +71,6 @@ type actionsHandler interface {
 	ListRunTaskResults(ctx context.Context, repositoryID string, runID string, after string) (actions.TaskResultIterator, error)
 }
 
-type UploadPathProvider interface {
-	NewPath() string
-}
-
 type Controller struct {
 	Config                *config.Config
 	Catalog               catalog.Interface
@@ -92,7 +88,7 @@ type Controller struct {
 	Templater             templater.Service
 	sessionStore          sessions.Store
 	oidcAuthenticator     *oidc.Authenticator
-	PathProvider          UploadPathProvider
+	PathProvider          upload.PathProvider
 }
 
 func (c *Controller) GetAuthCapabilities(w http.ResponseWriter, _ *http.Request) {
@@ -3523,7 +3519,7 @@ func NewController(
 	templater templater.Service,
 	oidcAuthenticator *oidc.Authenticator,
 	sessionStore sessions.Store,
-	pathProvider UploadPathProvider,
+	pathProvider upload.PathProvider,
 ) *Controller {
 	gob.Register(oidc.Claims{})
 	return &Controller{
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -21,7 +21,6 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/go-openapi/swag"
 	"github.com/gorilla/sessions"
-	nanoid "github.com/matoous/go-nanoid/v2"
 	"github.com/treeverse/lakefs/pkg/actions"
 	"github.com/treeverse/lakefs/pkg/auth"
 	"github.com/treeverse/lakefs/pkg/auth/email"
@@ -72,6 +71,10 @@ type actionsHandler interface {
 	ListRunTaskResults(ctx context.Context, repositoryID string, runID string, after string) (actions.TaskResultIterator, error)
 }
 
+type UploadPathProvider interface {
+	NewPath() string
+}
+
 type Controller struct {
 	Config                *config.Config
 	Catalog               catalog.Interface
@@ -89,6 +92,7 @@ type Controller struct {
 	Templater             templater.Service
 	sessionStore          sessions.Store
 	oidcAuthenticator     *oidc.Authenticator
+	PathProvider          UploadPathProvider
 }
 
 func (c *Controller) GetAuthCapabilities(w http.ResponseWriter, _ *http.Request) {
@@ -264,18 +268,8 @@ func (c *Controller) GetPhysicalAddress(w http.ResponseWriter, r *http.Request, 
 		return
 	}
 
-	// Generate a name.
-	name, err := nanoid.New()
-	if err != nil {
-		writeError(w, http.StatusInternalServerError, err)
-		return
-	}
-
-	tokenPart := ""
-	if token != nil {
-		tokenPart = *token + "/"
-	}
-	qk, err := block.ResolveNamespace(repo.StorageNamespace, fmt.Sprintf("data/%s%s", tokenPart, name), block.IdentifierTypeRelative)
+	address := c.PathProvider.NewPath()
+	qk, err := block.ResolveNamespace(repo.StorageNamespace, address, block.IdentifierTypeRelative)
 	if err != nil {
 		writeError(w, http.StatusInternalServerError, err)
 		return
@@ -2071,7 +2065,8 @@ func (c *Controller) UploadObject(w http.ResponseWriter, r *http.Request, reposi
 	}
 	defer func() { _ = file.Close() }()
 	contentType := handler.Header.Get("Content-Type")
-	blob, err := upload.WriteBlob(ctx, c.BlockAdapter, repo.StorageNamespace, file, handler.Size, block.PutOpts{StorageClass: params.StorageClass})
+	address := c.PathProvider.NewPath()
+	blob, err := upload.WriteBlob(ctx, c.BlockAdapter, repo.StorageNamespace, address, file, handler.Size, block.PutOpts{StorageClass: params.StorageClass})
 	if err != nil {
 		writeError(w, http.StatusInternalServerError, err)
 		return
@@ -3528,6 +3523,7 @@ func NewController(
 	templater templater.Service,
 	oidcAuthenticator *oidc.Authenticator,
 	sessionStore sessions.Store,
+	pathProvider UploadPathProvider,
 ) *Controller {
 	gob.Register(oidc.Claims{})
 	return &Controller{
@@ -3547,6 +3543,7 @@ func NewController(
 		Templater:             templater,
 		sessionStore:          sessionStore,
 		oidcAuthenticator:     oidcAuthenticator,
+		PathProvider:          pathProvider,
 	}
 }
 
--- a/pkg/api/serve_test.go
+++ b/pkg/api/serve_test.go
@@ -11,6 +11,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/treeverse/lakefs/pkg/upload"
+
 	"github.com/deepmap/oapi-codegen/pkg/securityprovider"
 	"github.com/spf13/viper"
 	"github.com/treeverse/lakefs/pkg/actions"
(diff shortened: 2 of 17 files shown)


pkg/upload/write_blob.go
```diff
@@ -16,11 +16,11 @@ type Blob struct {
...
 }
 
-func WriteBlob(ctx context.Context, adapter block.Adapter, bucketName string, body io.Reader, contentLength int64, opts block.PutOpts) (*Blob, error) {
+func WriteBlob(ctx context.Context, adapter block.Adapter, bucketName, address string, body io.Reader, contentLength int64, opts block.PutOpts) (*Blob, error) {
 	// handle the upload itself
 	hashReader := block.NewHashingReader(body, block.HashFunctionMD5, block.HashFunctionSHA256)
-	uid := uuid.New()
-	address := hex.EncodeToString(uid[:])
+	// uid := uuid.New()
+	// address := hex.EncodeToString(uid[:])
```
N-o-Z, line 23: Can be removed?

pkg/upload/path_provider.go
```diff
@@ -0,0 +1,117 @@
+package upload
```
N-o-Z, line 1: Very cool 😎

pkg/upload/path_provider.go
```diff
@@ -0,0 +1,117 @@
...
+	timestamp time.Time
+	// count make sure we will not produce more than 'count' IDs
+	count int
+	// partition current set and reset by size and time
+	partition string
+}
+
+const (
+	DefaultDataSize     = 50000
+	DefaultDataInterval = 5 * time.Minute
```
N-o-Z, line 26: I think this default interval is a bit small?
Maybe we should start with an hour?

pkg/upload/path_provider.go
```diff
@@ -0,0 +1,117 @@
...
+type PathProviderOption func(o *PathProviderConfig)
+
+func WithPathProviderSize(size int) PathProviderOption {
+	return func(cfg *PathProviderConfig) {
+		cfg.Size = size
+	}
+}
+
+func WithPathProviderInterval(d time.Duration) PathProviderOption {
+	return func(cfg *PathProviderConfig) {
```
N-o-Z, line 56: Perhaps we would want to provide some limitations on the given interval? Theoretically some one can pass 1ns

pkg/upload/path_provider_test.go
```diff
@@ -0,0 +1,106 @@
+package upload
+
+import (
+	"strings"
+	"testing"
+	"time"
+)
+
+func TestPathProvider(t *testing.T) {
```
N-o-Z, line 9: Maybe add a test that uses the defaults as well

pkg/gateway/operations/base.go
```diff
@@ -40,6 +40,10 @@ const (
 
 type ActionIncr func(action, repository, ref, userID string)
 
+type UploadPathProvider interface {
```
N-o-Z, line 43: Why is the UploadPathProvider interface defined under the gateway package?

pkg/api/serve_test.go
```diff
@@ -11,6 +11,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/treeverse/lakefs/pkg/upload"
+
```
N-o-Z, line 15: Redundant newline

pkg/api/controller.go
```diff
@@ -72,6 +71,10 @@ type actionsHandler interface {
 	ListRunTaskResults(ctx context.Context, repositoryID string, runID string, after string) (actions.TaskResultIterator, error)
 }
 
+type UploadPathProvider interface {
```
N-o-Z, line 74: Another interface definition?

pkg/gateway/operations/base.go
N-o-Z, line 43: Why is the UploadPathProvider interface defined under gateway?

N-o-Z (changes requested):
Awesome changes! 
Some questions and comments

pkg/upload/path_provider.go
nopcoder: True I prefer to not limiting configuration values and let the caller control and/or fail the app.
In this case the main consumer is the test code itself as the user currently doesn't control this value.
We can limit the user configuration input if needed.

pkg/upload/path_provider_test.go
nopcoder: Didn't want to wait 5min - 1h to testing the default nor creating 50k objects.
I think the system test for the GC can capture this one.

pkg/gateway/operations/base.go
nopcoder: Local interface that match the requirements from path provider.

pkg/api/controller.go
nopcoder: Yes - prefer to have local requirement as interface by the package that consume the implementation than having a global one.

pkg/gateway/operations/base.go
nopcoder: I can consolidate the interface definition - but using the above practice remove the dependency of `upload` package in gateway.
But because we already use the implementation directly - I accept and I will consolidate the interface usage.

pkg/upload/path_provider.go
```diff
@@ -0,0 +1,117 @@
...
+	// timestamp used to calculate the time based ID value
+	timestamp time.Time
+	// count make sure we will not produce more than 'count' IDs
+	count int
+	// partition current set and reset by size and time
+	partition string
+}
+
+const (
+	DefaultDataSize     = 50000
```
itaiad200, line 30:
```suggestion
	DefaultMaxSliceSize     = 50000
```

pkg/upload/path_provider.go
```diff
@@ -0,0 +1,117 @@
...
+		count:     0,
+		partition: newDescendingID(now).String(),
+	}
+}
+
+// NewPath returns a new upload path based on the current partition.
+func (i *PathProvider) NewPath() string {
+	part := i.partitionForID()
+	name := xid.New().String()
+	return path.Join(i.cfg.Prefix, part, name)
```
itaiad200, line 101: How do we avoid a double slash bug?

pkg/upload/path_provider.go
```diff
@@ -0,0 +1,117 @@
...
+
+func WithPathProviderInterval(d time.Duration) PathProviderOption {
+	return func(cfg *PathProviderConfig) {
+		cfg.Interval = d
+	}
+}
+
+func WithPathProviderPrefix(p string) PathProviderOption {
+	return func(cfg *PathProviderConfig) {
+		cfg.Prefix = p
```
itaiad200, line 68: Prefix needs sanitization

pkg/upload/path_provider.go
```diff
@@ -0,0 +1,117 @@
...
+	partition string
+}
+
+const (
+	DefaultDataSize     = 50000
+	DefaultDataInterval = 5 * time.Minute
+	DefaultDataPrefix   = "data"
+
+	// unixYear4000 epoch value for Saturday, January 1, 4000 12:00:00 AM
+	unixYear4000 = 64060588800
```
itaiad200, line 35: I would add a warning that changing this value is a breaking change. "4000 is too much, we should do 3000" is definitely something someone may want to change in some time..

pkg/upload/path_provider.go
```diff
@@ -0,0 +1,122 @@
...
+	part := i.partitionForID()
+	name := xid.New().String()
+	return path.Join(i.cfg.Prefix, part, name)
+}
+
+// partitionForID return the current partition to use. It will update the partition value before return in case we
+// crossed the size or internal (time) configured.
+func (i *PathPartitionProvider) partitionForID() string {
+	i.mu.Lock()
+	defer i.mu.Unlock()
```
itaiad200, line 108: Should we be so strict? IMO we can lose the lock and find a non locking although less accurate way to count the addresses..

pkg/upload/path_provider.go
```diff
@@ -0,0 +1,122 @@
...
+
+import (
+	"path"
+	"sync"
+	"time"
+
+	"github.com/rs/xid"
+)
+
+// PathProvider capture the requirements from PathPartitionProvider implementation to return a new path
```
itaiad200, line 11:
```suggestion
// PathProvider captures the requirements from PathPartitionProvider implementation to return a new path
```

itaiad200 (changes requested):
Looks great! Mostly worried about another double slash bug

pkg/upload/path_provider.go
nopcoder: https://go.dev/play/p/ae1iVUnL-nc

not worried about double slash

pkg/upload/path_provider.go
nopcoder: can you elaborate? this is a prefix in object store not local filesystem.

pkg/upload/path_provider.go
nopcoder: Can use atomic counter to have better performance - but I will still have race condition on read/write partition value.
Don't think it will have a bottleneck while uploading files from single instance of lakeFS.

pkg/upload/path_provider.go
itaiad200: Right - what if I pass `data/` instead of `data`?

pkg/upload/path_provider.go
```diff
@@ -22,18 +27,18 @@ type PathProvider struct {
...
 
 const (
-	DefaultDataSize     = 50000
-	DefaultDataInterval = 5 * time.Minute
+	DefaultMaxDataSize  = 50000
+	DefaultDataInterval = time.Hour
 	DefaultDataPrefix   = "data"
 
-	// unixYear4000 epoch value for Saturday, January 1, 4000 12:00:00 AM
+	// unixYear4000 epoch value for Saturday, January 1, 4000 12:00:00 AM. No not change the value as it is used to have reverse order for time based unique ID (xid).
```
itaiad200, line 34:
```suggestion
	// unixYear4000 epoch value for Saturday, January 1, 4000 12:00:00 AM. Changing this value is a breaking change as it is used to have reverse order for time based unique ID (xid).
```

itaiad200 (approved):

pkg/upload/path_provider.go
N-o-Z: @nopcoder Lets stick to a single convention - either `slice` or `partition`, I'm good with either one

nopcoder: Update the DefaultDataPartitionSize - dropped the slice and max. final.

N-o-Z (approved):
LGTM!

PR #4531, opened 2022-10-31
Cache action objects and listing

### LINKED ISSUE #4423: Limit blockstore/DB requests for Actions
It turns out that every lakeFS operation that could possibly trigger a hook, queries graveler for the actions files.
Action files by nature don't change very often. I think we need to consider some relaxation to the atomicity of action files changes, so that we won't have to perform X DB calls for every such operation, along side Y blockstore calls, if some action file is available.

itaiad200:
closes #4423

### DIFF
--- a/esti/hooks_test.go
+++ b/esti/hooks_test.go
@@ -277,7 +277,7 @@ func testCreateDeleteBranch(t *testing.T, ctx context.Context, repo string) {
 	appendRes(postCreateBranchEvent)
 	require.Equal(t, webhookEventInfo{
 		EventTime:    postCreateBranchEvent.EventTime,
-		SourceRef:    commitID,
+		SourceRef:    mainBranch,
 		EventType:    "post-create-branch",
 		ActionName:   "Test Post Create Branch",
 		HookID:       "test_webhook",
@@ -461,8 +461,8 @@ func parseAndUploadActions(t *testing.T, ctx context.Context, repo, branch strin
 		require.Equal(t, http.StatusCreated, resp.StatusCode())
 	}
 
-	// wait 10 seconds to let the actions cache expire.
-	time.Sleep(10 * time.Second)
+	// wait 8 seconds to let the actions cache expire.
+	time.Sleep(8 * time.Second)
 }
 
 type webhookEventInfo struct {
--- a/esti/hooks_test.go
+++ b/esti/hooks_test.go
@@ -277,7 +277,7 @@ func testCreateDeleteBranch(t *testing.T, ctx context.Context, repo string) {
 	appendRes(postCreateBranchEvent)
 	require.Equal(t, webhookEventInfo{
 		EventTime:    postCreateBranchEvent.EventTime,
-		SourceRef:    mainBranch,
+		SourceRef:    commitID,
 		EventType:    "post-create-branch",
 		ActionName:   "Test Post Create Branch",
 		HookID:       "test_webhook",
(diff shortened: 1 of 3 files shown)


esti/hooks_test.go
```diff
@@ -461,6 +460,9 @@ func parseAndUploadActions(t *testing.T, ctx context.Context, repo, branch strin
 		require.NoError(t, err)
 		require.Equal(t, http.StatusCreated, resp.StatusCode())
 	}
+
+	// wait 10 seconds to let the actions cache expire.
+	time.Sleep(10 * time.Second)
```
nopcoder, line 465: not doing good to the other work of reduce tests running time.

pkg/catalog/actions_source.go
```diff
@@ -4,24 +4,51 @@ import (
...
 
 func NewActionsSource(catalog *Catalog) *ActionsSource {
 	return &ActionsSource{
 		catalog: catalog,
+		cache:   cache.NewCache(DefaultActionsCacheSize, DefaultActionsCacheExpiry, cache.NewJitterFn(DefaultActionsCacheJitter)),
 	}
 }
 
 func (s *ActionsSource) List(ctx context.Context, record graveler.HookRecord) ([]string, error) {
+	key := fmt.Sprintf("list_%s_%s", record.RepositoryID.String(), record.SourceRef.String())
```
nopcoder, line 41: - Suggest using ':' as separator like we do for all cache keys.
- Don't think we require the prefix as the having <repo>:<ref> vs <repo>:<ref>:<name> will not collide.

pkg/graveler/graveler.go
```diff
@@ -979,7 +979,7 @@ func (g *KVGraveler) CreateBranch(ctx context.Context, repository *RepositoryRec
 		RunID:            postRunID,
 		StorageNamespace: storageNamespace,
 		EventType:        EventTypePostCreateBranch,
-		SourceRef:        ref,
+		SourceRef:        newBranch.CommitID.Ref(),
```
nopcoder, line 982: Just verify - `newBranch` will always resolve to a commit ID?

nopcoder (changes requested):
My main concern is that we may cache listing/object data based on branch name - do we always pass a commit id as the source for the action's source ref?

I think we should focus on using the same loaded action from the 'pre' action in the 'post' action - as it will have better performance and consistency.

pkg/graveler/graveler.go
itaiad200: Reverting back. We can load the actions from the source ref, the actions filtering is using the new branch ID.

esti/hooks_test.go
itaiad200: I agree, but it's a system test where we have cache we don't want to override.

pkg/catalog/actions_source.go
```diff
@@ -4,24 +4,51 @@ import (
...
+	names, err := s.cache.GetOrSet(key, func() (interface{}, error) {
```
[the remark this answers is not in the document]
itaiad200: Every action except `Commit` is using `commit.ID` which is immutable to load the action files .
We must use the branch to load action files for commit since the action files might have changed and we need to load them and run the latest version. My assumption is that the number of changes to the action files is low compared to changes to data files and with low contention, and a 5 seconds delay introduced by this cache is reasonable.

N-o-Z (approved):
Thanks for addressing the comments

nopcoder (approved):
LGTM. The sleep in the test code is the only thing that I prefer to find a way to remove if possible.

itaiad200:
> LGTM. The sleep in the test code is the only thing that I prefer to find a way to remove if possible.

It's Esti - the only way to control it is to have it as a lakeFS config and that seems like an overkill.

PR #4537, opened 2022-11-01
Added Enigma Use Case Reference

### LINKED ISSUE #4536: Should Add Use cases to docs
As we have new use cases published, we should add them to the docs.

iddoavn:
FIX #4536 

Adding Enigma's use case reference to our docs.


adipolak (approved):
LGTM!

PR #4539, opened 2022-11-01
Add video to run lakeFs locally

### LINKED ISSUE #4538: Differentiate Local installation vs local Run
We should differentiate between local installations and local runs on our quick start docs. 
Today we call running locally on a container "install" and it's a bit misleading.

iddoavn:
FIX #4538 

Added a video to running locally + changes addressing it as local run vs local install.

### DIFF
--- a/docs/quickstart/installing.md
+++ b/docs/quickstart/installing.md
@@ -8,7 +8,7 @@ has_children: false
 next: ["Create your first repository", "./repository.html"]
 ---
 
-# Install lakeFS
+# Run lakeFS Locally
 {: .no_toc }
 
 {% include learn_only.html %} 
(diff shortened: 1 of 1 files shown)


docs/quickstart/installing.md
```diff
@@ -1,14 +1,14 @@
...
+title: Run lakeFS Locally
+description: Running lakeFS locally is easy and quick. This section covers how to spin up lakeFS using Docker.
 parent: Quickstart
 nav_order: 10
 has_children: false
 next: ["Create your first repository", "./repository.html"]
 ---
 
-# Install lakeFS
+# Run lakeFS Locally
```
adipolak, line 11: It's better to use Install lakeFS locally instead of run, as from a developer point of view, the user essentially install it.

adipolak (changes requested):

kesarwam (approved):

adipolak (approved):
LGTM

PR #4540, opened 2022-11-02
Ensure output branch has no uncommitted files before starting

### LINKED ISSUE #4519: [LakeFSOutputCommitter] Verify output branch is clean before starting
When setting up to write verify that the output branch is clean (has no
diffs), fail if it is not.  It will be impossible to merge into a dirty
branch.

This should be configurable (default true).

arielshaqed:
The merge at the _end_ of the write will fail if there are uncommitted
files.  Don't bother even to start.

This is configurable using `committer.output_branch.ensure_clean`, default
true.

Fixes #4519.

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/Constants.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/Constants.java
@@ -9,9 +9,10 @@ public class Constants {
     public static final String SECRET_KEY_KEY_SUFFIX = "secret.key";
     public static final String ENDPOINT_KEY_SUFFIX = "endpoint";
     public static final String LIST_AMOUNT_KEY_SUFFIX = "list.amount";
+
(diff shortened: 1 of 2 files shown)


johnnyaug (approved):

lynnro314:
I'm not sure, why not to check it also in setupTask?

arielshaqed:
> I'm not sure, why not to check it also in setupTask?

Not sure, what would this would give us?

* `setupJob` runs, once, before any calls to `setupTask`.  So this change is enough to ensure the test.
* Even if we tested it later on, during `setupTask`, there would _still_ be a race.  I can test it every time, but the improvement is marginal and only in very special cases.

Please note that because of ef4b57ea9e933e4f351fa810a3b93260fc1cfa83 the task branches all use the same job branch that was used to call `setupJob`.

lynnro314 (approved):

arielshaqed:
THANKS!

PR #4545, opened 2022-11-02
Create success file when done

### LINKED ISSUE #4520: [LakeFSOutputCommitter] Create _SUCCESS file during commitJob
FileOutputCommitter creates such a file, and very many tools expect to see
it.  Create such a file.

This should be configurable (default true).

arielshaqed:
Create an object named _SUCCESS in the output path.  This is controlled by
option `lakefs.fs.committer.marksuccessfuljobs`, with these values:

* "true": Create success file.
* "false": Don't create the file.
* "auto" (the default): Do whatever FileOutputCommitter would do, by reading
   its option `mapreduce.fileoutputcommitter.marksuccessfuljobs` (default
   `true`).

So: by default it creates _SUCCESS or whatever FileOutputCommitter is
configured to do.  And it can be set explicitly.

Fixes #4520.

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/committer/DummyOutputCommitter.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/committer/DummyOutputCommitter.java
@@ -226,7 +226,10 @@ public class DummyOutputCommitter extends FileOutputCommitter {
         if (markSuccessfulJobs.equals("auto")) {
             return conf.getBoolean(FileOutputCommitter.SUCCESSFUL_JOB_OUTPUT_DIR_MARKER, true);
         }
-        return Boolean.parseBoolean(markSuccessfulJobs);
+        if (markSuccessfulJobs.equalsIgnoreCase("false")) {
+            return false;
+        }
+        return true;
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


clients/hadoopfs/src/main/java/io/lakefs/committer/DummyOutputCommitter.java
```diff
@@ -216,11 +221,29 @@ private void cleanupJob() throws IOException {
         }
     }
 
+    private boolean needsSuccessFile() {
+        String markSuccessfulJobs = FSConfiguration.get(conf, outputPath.toUri().getScheme(), Constants.OC_SUCCESSFUL_JOB_OUTPUT_DIR_MARKER, "auto");
+        if (markSuccessfulJobs.equals("auto")) {
+            return conf.getBoolean(FileOutputCommitter.SUCCESSFUL_JOB_OUTPUT_DIR_MARKER, true);
+        }
+        return Boolean.parseBoolean(markSuccessfulJobs);
```
Jonathan-Rosenberg, line 229: Maybe add a validation that `markSuccessfulJobs`'s value isn't something weird (neither `true` or `false`)

arielshaqed (commented):
Thanks!

Just set a default value (true), which is what Hadoop does :-)

lynnro314 (approved):

arielshaqed:
# :-)

PR #4548, opened 2022-11-02
GC+: Implement SaveGarbageCollectionUncommitted

### LINKED ISSUE #4547: Implement GC Save Uncommitted
This is part one of: https://github.com/treeverse/lakeFS/issues/4475

N-o-Z:
- Closes #4547


pkg/catalog/catalog.go
```diff
@@ -1711,6 +1717,99 @@ func (c *Catalog) PrepareExpiredCommits(ctx context.Context, repositoryID string
 	return c.Store.SaveGarbageCollectionCommits(ctx, repository, previousRunID)
 }
 
+func createParquetFile(entries []DBEntry) (string, error) {
+	// TODO: Modify using a run ID generator (https://github.com/treeverse/lakeFS/issues/4469)
+	filename := uuid.New().String()
```
itaiad200, line 1722: Write under `/tmp/`

pkg/catalog/catalog.go
```diff
@@ -1711,6 +1717,99 @@ func (c *Catalog) PrepareExpiredCommits(ctx context.Context, repositoryID string
 	return c.Store.SaveGarbageCollectionCommits(ctx, repository, previousRunID)
 }
 
+func createParquetFile(entries []DBEntry) (string, error) {
```
itaiad200, line 1720: Seems like a very straightforward parquet file writer, can we extract it to somewhere common?

pkg/graveler/graveler.proto
```diff
@@ -50,8 +50,9 @@ message GarbageCollectionRules {
 
 message GarbageCollectionRunMetadata {
   string run_id = 1;
-  string Commits_csv_location = 2;
-  string Address_location = 3;
+  string commits_csv_location = 2;
+  string address_location = 3;
+  string uncommitted_location = 4;
 }
```
itaiad200, line 56: It's hard to understand these fields without documentation. `address_location`? I know you just added 1 field but I think we need to add some info here.

pkg/graveler/graveler.go
```diff
@@ -31,6 +31,8 @@ const (
 
 	// BranchWriteMaxTries is the number of times to repeat the set operation if the staging token changed
 	BranchWriteMaxTries = 3
+
+	GCMaxUncommittedObjects = 1000000
```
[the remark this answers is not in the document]
itaiad200: The "ideal" file size for uploading anything to S3 is roughly 8-20MB

pkg/graveler/graveler.go
```diff
@@ -509,6 +511,13 @@ type VersionController interface {
 	// Note: Ancestors of previously expired commits may still be considered if they can be reached from a non-expired commit.
 	SaveGarbageCollectionCommits(ctx context.Context, repository *RepositoryRecord, previousRunID string) (garbageCollectionRunMetadata *GarbageCollectionRunMetadata, err error)
 
+	// GetGarbageCollectionUncommitted Returns a list of at most GCMaxUncommittedObjects uncommitted objects, and a continuation token in the form of GCUncommittedMark if needed.
+	GetGarbageCollectionUncommitted(ctx context.Context, repository *RepositoryRecord, mark GCUncommittedMark) ([]*ValueRecord, *GCUncommittedMark, error)
```
itaiad200, line 515: Needs an iterator. Passing a 1M objects here with a potential of 1000 bytes per entry is 1GB of memory for a single call.

itaiad200, line 515: Is the mark mandatory? If not, pass it as a pointer.

itaiad200: And then the mark is redundant. Simply read from the iterator as long as you want.

pkg/graveler/graveler.go
```diff
@@ -509,6 +511,13 @@ type VersionController interface {
...
+
+	// SaveGarbageCollectionUncommitted Uploads parquet file of uncommitted objects to the repository in the run ID path. Return the GC run's metadata
+	// including the run ID and the location of the saved parquet file
+	SaveGarbageCollectionUncommitted(ctx context.Context, repository *RepositoryRecord, filename, runID string) (*GarbageCollectionRunMetadata, error)
```
itaiad200, line 519: IIUC it uploads to the file to the object store, right? Why is it a graveler functionality? We have the storage namespace in the repo record, we can use that and uplod directly to the store.

pkg/graveler/graveler.go
```diff
@@ -1286,6 +1295,66 @@ func (g *KVGraveler) SaveGarbageCollectionCommits(ctx context.Context, repositor
...
+	var entries []*ValueRecord
+	bItr, err := g.ListBranches(ctx, repository)
+	if err != nil {
+		return nil, nil, err
+	}
+	defer bItr.Close()
+	bItr.SeekGE(mark.BranchID)
+
+	count := 0
+	for bItr.Next() {
```
itaiad200, line 1313: You start here a long operation, please make sure that the context is respected by the inner iterators.

pkg/graveler/graveler.go
```diff
@@ -1286,6 +1295,66 @@ func (g *KVGraveler) SaveGarbageCollectionCommits(ctx context.Context, repositor
...
+		b := bItr.Value()
+		itr, err := g.listStagingArea(ctx, b.Branch)
+		if err != nil {
+			bItr.Close()
+			return nil, nil, err
+		}
+		itr.SeekGE(mark.Key)
```
itaiad200, line 1320: That's a bug, not every entry iterator should `SeekGE`. Just the first one.

pkg/graveler/graveler.go
```diff
@@ -1286,6 +1295,66 @@ func (g *KVGraveler) SaveGarbageCollectionCommits(ctx context.Context, repositor
...
+	}
+	defer bItr.Close()
+	bItr.SeekGE(mark.BranchID)
+
+	count := 0
+	for bItr.Next() {
+		b := bItr.Value()
+		itr, err := g.listStagingArea(ctx, b.Branch)
+		if err != nil {
+			bItr.Close()
```
itaiad200, line 1317: Unnecessary - you deferred it several lines above.

pkg/graveler/graveler.go
```diff
@@ -1286,6 +1295,66 @@ func (g *KVGraveler) SaveGarbageCollectionCommits(ctx context.Context, repositor
...
+					Key:      entry.Key,
+				}
+				return entries, &contToken, nil
+			}
+			entries = append(entries, entry)
+			itr.Close()
+			if itr.Err() != nil {
+				return nil, nil, itr.Err()
+			}
+		}
```
itaiad200, line 1336: After an iterator return `false` from `Next` - you must check for errors.

pkg/graveler/graveler.go
```diff
@@ -1286,6 +1295,66 @@ func (g *KVGraveler) SaveGarbageCollectionCommits(ctx context.Context, repositor
...
+					BranchID: b.BranchID,
+					Key:      entry.Key,
+				}
+				return entries, &contToken, nil
+			}
+			entries = append(entries, entry)
+			itr.Close()
+			if itr.Err() != nil {
+				return nil, nil, itr.Err()
+			}
```
itaiad200, line 1335: Why do we do this inside the loop?

pkg/graveler/graveler.go
```diff
@@ -1286,6 +1295,66 @@ func (g *KVGraveler) SaveGarbageCollectionCommits(ctx context.Context, repositor
...
+		}
+		itr.Close()
+	}
+	return entries, nil, bItr.Err()
```
itaiad200, line 1339: We should return a nil result in case of an error.

pkg/graveler/graveler.go
```diff
@@ -1286,6 +1295,66 @@ func (g *KVGraveler) SaveGarbageCollectionCommits(ctx context.Context, repositor
...
+	}
+	uncommittedLocation, err := g.garbageCollectionManager.GetUncommittedLocation(runID, repository.StorageNamespace)
+	if err != nil {
+		return nil, err
+	}
+
+	return &GarbageCollectionRunMetadata{
+		RunId:               runID,
+		UncommittedLocation: uncommittedLocation,
+	}, err
```
itaiad200, line 1355:
```suggestion
	}, nil
```

pkg/catalog/catalog.go
```diff
@@ -1711,6 +1717,99 @@ func (c *Catalog) PrepareExpiredCommits(ctx context.Context, repositoryID string
...
+	mark := graveler.GCUncommittedMark{
+		BranchID: "",
+		Key:      []byte(""),
+	}
+	if contToken != "" {
+		err = json.NewDecoder(base64.NewDecoder(base64.StdEncoding, strings.NewReader(contToken))).Decode(&mark)
+		if err != nil {
+			return nil, nil, err
+		}
+	}
```
itaiad200, line 1783: I think it's more of the controller responsibility. The Catalog role is to pass the mark, the controller role is to pass it conveniently using HTTP. So I guess we can use the actual type and let the controller perform the conversions.

pkg/catalog/catalog.go
```diff
@@ -1711,6 +1717,99 @@ func (c *Catalog) PrepareExpiredCommits(ctx context.Context, repositoryID string
...
+	if err != nil {
+		return nil, nil, err
+	}
+	var catalogEntries []DBEntry
+	for _, value := range values {
+		entry, err := ValueToEntry(value.Value)
+		if err != nil {
+			return nil, nil, err
+		}
+		catalogEntries = append(catalogEntries, newCatalogEntryFromEntry(false, "", entry))
```
itaiad200, line 1798: We actually already have iterators for this type of conversions.

pkg/catalog/catalog.go
```diff
@@ -1711,6 +1717,99 @@ func (c *Catalog) PrepareExpiredCommits(ctx context.Context, repositoryID string
...
+	}
+
+	filename, err := createParquetFile(catalogEntries)
+	if err != nil {
+		return nil, nil, err
+	}
+	md, err := c.Store.SaveGarbageCollectionUncommitted(ctx, repository, filename, runID)
```
itaiad200, line 1805: What's `md`?

pkg/graveler/retention/garbage_collection_manager.go
```diff
@@ -17,9 +18,10 @@ import (
 )
 
 const (
-	configFileSuffixTemplate    = "%s/retention/gc/rules/config.json"
-	addressesFilePrefixTemplate = "%s/retention/gc/addresses/"
-	commitsFileSuffixTemplate   = "%s/retention/gc/commits/run_id=%s/commits.csv"
+	configFileSuffixTemplate      = "%s/retention/gc/rules/config.json"
+	addressesFilePrefixTemplate   = "%s/retention/gc/addresses/"
+	commitsFileSuffixTemplate     = "%s/retention/gc/commits/run_id=%s/commits.csv"
+	uncommittedFilePrefixTemplate = "%s/retention/gc/%s/uncommitted/"
```
itaiad200, line 24: Can't we use the same format as in committed?

pkg/graveler/retention/garbage_collection_manager.go
```diff
@@ -46,6 +48,41 @@ func (m *GarbageCollectionManager) GetAddressesLocation(sn graveler.StorageNames
...
+	fd, err := os.Open(filename)
+	if err != nil {
+		return err
+	}
+	stat, err := fd.Stat()
+	if err != nil {
+		return err
+	}
+
+	identifier := fmt.Sprintf("%s%s", location, filename)
```
itaiad200, line 75: No `/` to separate?

pkg/graveler/graveler.go
```diff
@@ -1286,6 +1295,66 @@ func (g *KVGraveler) SaveGarbageCollectionCommits(ctx context.Context, repositor
...
+	}
+	return entries, nil, bItr.Err()
+}
+
+func (g *KVGraveler) SaveGarbageCollectionUncommitted(ctx context.Context, repository *RepositoryRecord, filename, runID string) (*GarbageCollectionRunMetadata, error) {
+	err := g.garbageCollectionManager.SaveGarbageCollectionUncommitted(ctx, repository, filename, runID)
+	if err != nil {
+		return nil, err
+	}
+	uncommittedLocation, err := g.garbageCollectionManager.GetUncommittedLocation(runID, repository.StorageNamespace)
```
itaiad200, line 1347: Can't `uncommittedLocation ` be the output of `SaveGarbageCollectionUncommitted` and remove `GetUncommittedLocation` from the API?

pkg/catalog/catalog.go
```diff
@@ -1711,6 +1717,99 @@ func (c *Catalog) PrepareExpiredCommits(ctx context.Context, repositoryID string
...
+	encoder := base64.NewEncoder(base64.StdEncoding, &buf)
+	err := json.NewEncoder(encoder).Encode(mark)
+	if err != nil {
+		return nil, err
+	}
+	result := buf.String()
+	return &result, nil
+}
+
+func (c *Catalog) PrepareGCUncommitted(ctx context.Context, repositoryID, runID, contToken string) (*graveler.GarbageCollectionRunMetadata, *string, error) {
```
itaiad200, line 1763: This is a partial implementation according to the plan. The write of the `copies table` is missing. How would we paginate that?

pkg/graveler/retention/garbage_collection_manager_test.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/google/uuid"
+	"github.com/stretchr/testify/require"
+	"github.com/treeverse/lakefs/pkg/block"
+	"github.com/treeverse/lakefs/pkg/block/mem"
+	"github.com/treeverse/lakefs/pkg/graveler"
+	"github.com/treeverse/lakefs/pkg/graveler/retention"
+	"github.com/treeverse/lakefs/pkg/graveler/testutil"
+)
+
+func TestGarbageCollectionManager_GetUncommittedLocation(t *testing.T) {
```
itaiad200, line 20: Please make this a matrix of tests and add `/` shenanigans. `prefix` is one of (`test_prefix`, `test_prefix/`, `/ test_prefix/`), storage namespace is one of (`mem://test-namespace/my-repo`, `mem://test-namespace/my-repo/`, `mem://test-namespace/`, `mem://test-namespace`), etc.

pkg/graveler/retention/garbage_collection_manager_test.go
```diff
@@ -0,0 +1,88 @@
...
+	blockAdapter := mem.New()
+	refMgr := &testutil.RefsFake{}
+	prefix := "test_prefix"
+	runID := "my_test_runID"
```
itaiad200, line 24: can the runID contain a `/`?

pkg/graveler/graveler.go
itaiad200: Also the `mark.Key` is valid only if the branch equals to `mark.Branch`. Imagine the mark is `branch1,key200`, and the `branch1` was deleted. You will now start from `branch2,key200` although you didn't cover prior keys in `branch2` at all!

pkg/graveler/graveler_v2_test.go
```diff
@@ -547,3 +550,132 @@ func TestGravelerCommit_v2(t *testing.T) {
...
+		if offset > graveler.GCMaxUncommittedObjects {
+			total += 1
+		}
+	} else {
+		total = numBranches
+	}
+
+	for i := 0; i < total; i++ {
+		test.stagingManager.EXPECT().List(gomock.Any(), gomock.Any(), gomock.Any()).Return(testutils.NewFakeValueIterator(records[i+branchOffset]), nil)
+	}
```
itaiad200, line 667: 😱😱😱 
I guess that this is where you spent most of your time trying to find a bug. I think it will be hard for us to maintain. I suggest to simplify it or at the very least explain what was done here.

itaiad200:
Can you please add the followup tasks (those part of #4547) that are not part of this PR.

pkg/graveler/graveler.go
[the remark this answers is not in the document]
N-o-Z: What are we trying to optimize? the upload time to S3 or the download by the GC client? I think it should be the latter. AFAIU files up to 200 MB should to optimal.
Correct me if I'm wrong

pkg/catalog/catalog.go
N-o-Z: 2 options:
1. graveler GetGarbageCollectionUncommitted once we reach the end we can safely read the copy table as part of the data
2. In catalog perform another call in graveler to explicitly read the table

pkg/graveler/graveler.go
N-o-Z: GetUncommittedLocation would be used by the client to get the the previous run's uncommitted data

pkg/graveler/retention/garbage_collection_manager.go
N-o-Z: location is already delimited

pkg/catalog/catalog.go
N-o-Z: metadata

pkg/graveler/graveler.go
N-o-Z: Done

pkg/graveler/graveler.go
N-o-Z: Fixed

pkg/graveler/graveler.go
N-o-Z: 😨

pkg/graveler/graveler.go
johnnyaug: This is a strange method signature for graveler: it basically gets a file and saves it in S3. It doesn't really do anything GC-related.

pkg/graveler/graveler.go
johnnyaug: Agreed that this needs to be an iterator. Also, I wonder if the mark is returned to the client (I assume yes, because it is serialized and encoded) - but wondering about the motivation for requiring the client to make multiple paginated calls here.

If it's because we worry that the request will take too long, then this needs to be an asynchronous request - i.e. return a token with which the client can sample the server until the file is ready.

pkg/catalog/catalog.go
```diff
@@ -1711,6 +1719,99 @@ func (c *Catalog) PrepareExpiredCommits(ctx context.Context, repositoryID string
 	return c.Store.SaveGarbageCollectionCommits(ctx, repository, previousRunID)
 }
 
+func createParquetFile(entries []DBEntry) (string, error) {
```
johnnyaug, line 1722: I would want this to return a parquet reader. Together with the iterator approach suggested by Itai in another comment - we could stream the entries from the KV store to the final Parquet file without saving it in memory or in the file system.

johnnyaug (changes requested):
Great progress!

pkg/catalog/catalog.go
```diff
@@ -1717,50 +1718,116 @@ func (c *Catalog) PrepareExpiredCommits(ctx context.Context, repositoryID string
 	return c.Store.SaveGarbageCollectionCommits(ctx, repository, previousRunID)
 }
 
-func createParquetFile(entries []DBEntry) (string, error) {
-	// TODO: Modify using a run ID generator (https://github.com/treeverse/lakeFS/issues/4469)
-	filename := uuid.New().String()
-	fw, err := local.NewLocalFileWriter(filename)
+type GCUncommittedMark struct {
```
itaiad200, line 1721: We usually do `after` in the API. Here it's harder since you have 2 entities to skip. Can you document here (and later in the API) that the mark is the next key to write (and not the last key written)..?

pkg/catalog/catalog.go
```diff
@@ -1717,50 +1718,116 @@ func (c *Catalog) PrepareExpiredCommits(ctx context.Context, repositoryID string
...
-	encoder := base64.NewEncoder(base64.StdEncoding, &buf)
-	err := json.NewEncoder(encoder).Encode(mark)
+	err = pw.WriteStop()
 	if err != nil {
 		return nil, err
 	}
-	result := buf.String()
-	return &result, nil
+
+	// Finished reading all staging area - not continuation token
```
itaiad200, line 1825:
```suggestion
	// Finished reading all staging area - no continuation token
```

pkg/graveler/graveler.go
itaiad200: @N-o-Z think you missed this

pkg/graveler/graveler_v2_test.go
```diff
@@ -541,141 +513,12 @@ func TestGravelerCommit_v2(t *testing.T) {
...
-		val, err := test.sut.Commit(ctx, repository, branch1ID, graveler.CommitParams{})
+		val, err := test.Sut.Commit(ctx, repository, branch1ID, graveler.CommitParams{})
 
 		require.Error(t, err)
 		require.True(t, errors.Is(err, graveler.ErrTooManyTries))
 		require.Equal(t, val, graveler.CommitID(""))
 	})
 }
-
-func TestKVGraveler_GetGarbageCollectionUncommitted(t *testing.T) {
```
itaiad200: So no tests at all 😢 ?

pkg/catalog/catalog_test.go
```diff
@@ -510,3 +522,148 @@ func TestCatalog_ListEntries(t *testing.T) {
...
+		RepositoryID: repoID,
+		Repository: &graveler.Repository{
+			StorageNamespace: "mock-sn",
+			CreationDate:     time.Now(),
+			DefaultBranchID:  "main",
+		},
+	}
+)
+
+func TestKVGraveler_GetGarbageCollectionUncommitted(t *testing.T) {
```
itaiad200, line 538: NM found it 😃

pkg/graveler/graveler.go
itaiad200: Asynchronous requests are very hard to manage. Need to store the request and manage its state in a multi instances case. We normally push this responsibility to the client, AFAIR we still don't have a single handling of async requests, even in places where it's impossible to paginate. E.g a commit of 1B objects.

pkg/graveler/retention/garbage_collection_manager.go
itaiad200: I'll forever be the advocate for double/missing slash bugs.. Can we use `path.Join`?

itaiad200 (changes requested):
Thanks for addressing many of the comments. I added some more - great progress!

pkg/catalog/catalog.go
```diff
@@ -1717,50 +1718,116 @@ func (c *Catalog) PrepareExpiredCommits(ctx context.Context, repositoryID string
...
+					return nil, err
+				}
+				done, err := fileSizeExceeded()
+				if err != nil {
+					eIt.Close()
+					return nil, err
+				}
+				if done {
+					key := vIt.Value().Key
+					eIt.Close()
```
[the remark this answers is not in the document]
N-o-Z: The reason I'm using vIt - is to avoid a bug where the key is cast into path string (https://github.com/treeverse/lakeFS/issues/4554) - I don't believe this justifies a new iterator

pkg/catalog/catalog.go
```diff
@@ -1717,50 +1718,116 @@ func (c *Catalog) PrepareExpiredCommits(ctx context.Context, repositoryID string
...
+		b := bIt.Value()
+		vIt, err := c.Store.ListStaging(ctx, b.Branch)
+		if err != nil {
+			return nil, err
+		}
+		if b.BranchID == mark.BranchID {
+			vIt.SeekGE(mark.Key)
+		}
+		eIt := NewValueToEntryIterator(vIt)
+		for eIt.Next() {
```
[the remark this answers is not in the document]
N-o-Z: And I do it in line 1818

pkg/catalog/catalog.go
```diff
@@ -1717,50 +1718,116 @@ func (c *Catalog) PrepareExpiredCommits(ctx context.Context, repositoryID string
...
+		return nil, err
 	}
 	// Compression type
 	pw.CompressionType = parquet.CompressionCodec_GZIP
-	for _, entry := range entries {
-		if err = pw.Write(entry); err != nil {
-			return "", err
+
+	const GCMaxUncommittedFileSize = 20 * 1024 * 1024 // 20 MB
+	const periodicCheckSize = 100000
```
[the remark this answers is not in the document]
N-o-Z: Since we control the address string:
"data" prefix = 4 bytes
partition id = 20 bytes (xid)
object name = 20 bytes (xid)
timestamp (int64) = 8 bytes

Total per entry ~52 bytes 
Deviation should be around 5 MB

We should also take into consideration that Flushing + Stating the file is a relatively heavy process which will considerably lengthen the operation

WDYT?

pkg/catalog/catalog.go
```diff
@@ -1717,50 +1718,116 @@ func (c *Catalog) PrepareExpiredCommits(ctx context.Context, repositoryID string
...
 	}
-	return filename, nil
-}
-
-func markToString(mark *graveler.GCUncommittedMark) (*string, error) {
-	if mark == nil {
-		return nil, nil
+	defer bIt.Close()
+	bIt.SeekGE(mark.BranchID)
+	for bIt.Next() {
```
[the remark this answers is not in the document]
N-o-Z: line 1818

pkg/catalog/catalog.go
```diff
@@ -1717,50 +1718,116 @@ func (c *Catalog) PrepareExpiredCommits(ctx context.Context, repositoryID string
...
-func createParquetFile(entries []DBEntry) (string, error) {
-	// TODO: Modify using a run ID generator (https://github.com/treeverse/lakeFS/issues/4469)
-	filename := uuid.New().String()
-	fw, err := local.NewLocalFileWriter(filename)
+type GCUncommittedMark struct {
+	BranchID graveler.BranchID `json:"branch"`
+	Key      graveler.Key      `json:"key"`
+}
+
+func (m *GCUncommittedMark) String() (string, error) {
```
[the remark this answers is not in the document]
N-o-Z: I don't think Decode should be a struct method, will implement when required

pkg/graveler/graveler.go
N-o-Z: I prefer the garbage collection manager handling the uploading of files to its managed prefix. Catalog seems like a too far high level place to decide where to write GC information. I can expose GetUncommittedLocation through graveler but it feels like a stretch. You could argue the same thing about SaveGarbageCollectionCommits (basically all the logic can be done in graveler)

pkg/catalog/catalog.go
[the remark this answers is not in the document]
itaiad200: Notice that for staged entries we do not control the address. 
Either way, your calculation is more accurate than mine -> let's add it as a comment for future reference.

pkg/catalog/uncommitted_iterator.go
```diff
@@ -0,0 +1,127 @@
...
+	vItr, err := u.store.ListStaging(u.ctx, u.branchItr.Value().Branch)
+	if err != nil {
+		u.err = err
+		return false
+	}
+	u.entryItr = NewValueToEntryIterator(vItr)
+	return true
+}
+
+func (u *UncommittedIterator) next() bool {
```
itaiad200, line 48: It was hard for me to understand why this function is correct. Consider renaming and/or document it.

pkg/catalog/catalog.go
```diff
@@ -1838,11 +1834,8 @@ func (c *Catalog) PrepareGCUncommitted(ctx context.Context, repositoryID, runID
 		return nil, nil, err
 	}
 
-	if contToken == nil {
-		contToken = &GCUncommittedMark{
-			BranchID: "",
-			Key:      []byte(""),
-		}
+	if mark != nil && runID == nil {
+		return nil, nil, fmt.Errorf("must provide run ID with mark: %w", ErrConflictFound)
```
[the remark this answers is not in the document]
N-o-Z: Removed check

pkg/catalog/uncommitted_iterator.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+func (u *UncommittedIterator) SeekGE(branchID graveler.BranchID, id Path) {
+	if u.Err() != nil {
+		return
+	}
+	u.entry = nil
+	if branchID != u.branchItr.Value().BranchID {
+		u.branchItr.SeekGE(branchID)
+		if u.branchItr.Next() {
```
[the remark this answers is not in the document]
N-o-Z: You have to advance the branch iterator because you need to perform SeekGE on the value iterator, and to get the value iterator you need the branch value

pkg/catalog/uncommitted_iterator.go
```diff
@@ -0,0 +1,127 @@
...
+			u.entry.EntryRecord = u.entryItr.Value()
+			return true
+		} else {
+			u.entryItr.Close()
+			if u.entryItr.Err() != nil {
+				return false
+			}
+		}
+	}
+	return u.next()
```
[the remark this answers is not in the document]
N-o-Z: Reverting the suggestion - this is not the correct flow

pkg/catalog/uncommitted_iterator.go
```diff
@@ -0,0 +1,127 @@
...
+		return
+	}
+	u.entry = nil
+	if branchID != u.branchItr.Value().BranchID {
+		u.branchItr.SeekGE(branchID)
+		if u.branchItr.Next() {
+			u.nextStaging()
+		}
+	}
+	if u.nextStaging() {
```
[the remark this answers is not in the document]
N-o-Z: Fixed

N-o-Z:
@itaiad200  Added documentation to the iterator functions.
Let's sync up if things are still unclear

pkg/graveler/graveler.go
```diff
@@ -1526,6 +1553,11 @@ func (g *KVGraveler) deleteUnsafe(ctx context.Context, repository *RepositoryRec
 	return g.StagingManager.Set(ctx, branch.StagingToken, key, nil, true)
 }
 
+// ListStaging Exposing listStagingArea to catalog for PrepareGCUncommitted
+func (g *KVGraveler) ListStaging(ctx context.Context, branch *Branch) (ValueIterator, error) {
```
johnnyaug, line 1557: You can just expose the original method.

pkg/graveler/graveler.go
```diff
@@ -1286,6 +1293,26 @@ func (g *KVGraveler) SaveGarbageCollectionCommits(ctx context.Context, repositor
 	}, err
 }
 
+func (g *KVGraveler) SaveGarbageCollectionUncommitted(ctx context.Context, repository *RepositoryRecord, filename string, runID *string) (*GarbageCollectionRunMetadata, error) {
+	rID := g.garbageCollectionManager.NewID()
+	if runID != nil {
+		rID = *runID
+	}
```
johnnyaug, line 1300: Make runID a string and then:
```suggestion
	if runID == "" {
		runID = *g.garbageCollectionManager.NewID()
	}
```

pkg/catalog/catalog_test.go
```diff
@@ -510,3 +522,156 @@ func TestCatalog_ListEntries(t *testing.T) {
...
+				md, mark, err = c.PrepareGCUncommitted(ctx, repoID.String(), runID, mark)
+				require.NoError(t, err)
+				require.Equal(t, *runID, md.RunId)
+			}
+			verifyData(t, tt.numBranch, tt.numRecords, testFolder)
+		})
+	}
+}
+
+func createTestScenario(t *testing.T, numBranches, numRecords, expectedCalls int) (*testutil.GravelerTest, string) {
```
johnnyaug, line 597:
```suggestion
func createPrepareUncommittedTestScenario(t *testing.T, numBranches, numRecords, expectedCalls int) (*testutil.GravelerTest, string) {
```

pkg/catalog/catalog_test.go
```diff
@@ -510,3 +522,156 @@ func TestCatalog_ListEntries(t *testing.T) {
...
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			g, testFolder := createTestScenario(t, tt.numBranch, tt.numRecords, tt.expectedCalls)
+			c := &catalog.Catalog{
+				Store: g.Sut,
+			}
+			var (
+				runID *string
+				//md *graveler.GarbageCollectionRunMetadata
```
johnnyaug, line 579: /dd

pkg/catalog/uncommitted_iterator.go
```diff
@@ -0,0 +1,133 @@
...
+	entryItr  *valueEntryIterator
+	entry     *UncommittedRecord
+}
+
+type UncommittedRecord struct {
+	branchID graveler.BranchID
+	*EntryRecord
+}
+
+func NewUncommittedIterator(ctx context.Context, store Store, repository *graveler.RepositoryRecord) (*UncommittedIterator, error) {
```
johnnyaug, line 23: Do we have a test for this file?

pkg/graveler/retention/garbage_collection_manager_test.go
```diff
@@ -0,0 +1,88 @@
...
+	fd, err := os.Create(filename)
+	require.NoError(t, err)
+	defer fd.Close()
+	for i := 0; i < count; i++ {
+		_, err := fd.WriteString(fmt.Sprintf("%s_%d\n", testLine, i))
+		require.NoError(t, err)
+	}
+}
+
+func TestGarbageCollectionManager_SaveGarbageCollectionUncommitted(t *testing.T) {
```
johnnyaug, line 46: Again, I feel that "SaveGarbageCollectionUncommitted" is out of place. It doesn't really do anything related to GC, and this test proves it - you basically test that it saves a file under a given location.

johnnyaug (commented):
Looks very, very good. I'm still blocking because:
1. I think uncommitted iterator should be tested.
2. I think the API in GC manager can still be more coherent.

Also, please see my other minor comments.

pkg/catalog/catalog.go
```diff
@@ -1720,6 +1727,145 @@ func (c *Catalog) PrepareExpiredCommits(ctx context.Context, repositoryID string
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Finished reading all staging area - no continuation token
+	return nil, nil
+}
+
+// TODO (niro): Add check in controller that runID is provided if mark != nil
+func (c *Catalog) PrepareGCUncommitted(ctx context.Context, repositoryID string, runID *string, mark *GCUncommittedMark) (*graveler.GarbageCollectionRunMetadata, *GCUncommittedMark, error) {
```
nopcoder, line 1836: Suggestions:
1. Return PrepareGCUncommittedInfo or a single struct instead of multiple return variables.
2. RunID as string when empty is the same as passing nil

pkg/catalog/catalog.go
```diff
@@ -1720,6 +1727,145 @@ func (c *Catalog) PrepareExpiredCommits(ctx context.Context, repositoryID string
...
+		{Name: "repository", Value: repositoryID, Fn: graveler.ValidateRepositoryID},
+	}); err != nil {
+		return nil, nil, err
+	}
+	repository, err := c.getRepository(ctx, repositoryID)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	filename := path.Join(os.TempDir(), uuid.New().String())
```
nopcoder, line 1847: There is an API to generate temporary file

pkg/catalog/uncommitted_iterator.go
```diff
@@ -0,0 +1,133 @@
...
+		store:     store,
+		ctx:       ctx,
+		branchItr: bItr,
+	}, nil
+}
+
+// nextStaging reads the next branch staging area into entryItr
+func (u *UncommittedIterator) nextStaging() bool {
+	if u.entryItr != nil {
+		u.entryItr.Close()
```
nopcoder, line 38: verify that multiple calls to Close are handled by the iterator or use nil as a marker

pkg/catalog/uncommitted_iterator.go
```diff
@@ -0,0 +1,133 @@
...
+	u.entryItr = NewValueToEntryIterator(vItr)
+	return true
+}
+
+// next Sets iterators to provide the next entry. Handles dependency between branch and value iterators.
+// Sets value and returns true if next entry available - false otherwise
+func (u *UncommittedIterator) next() bool {
+	u.entry = nil // will stay nil as long as no new value found
+	for u.entry == nil {
+		if u.branchItr.Next() && u.nextStaging() {
```
nopcoder, line 54: revert the check to reduce code indent

pkg/catalog/uncommitted_iterator.go
```diff
@@ -0,0 +1,133 @@
...
+				u.entry = &UncommittedRecord{
+					branchID: u.branchItr.Value().BranchID,
+				}
+				u.entry.EntryRecord = u.entryItr.Value()
+				return true
+			}
+			u.entryItr.Close()
+			if u.entryItr.Err() != nil {
+				return false
+			}
```
nopcoder, line 65: The "protocol" state that we need to check the error first - as closing the iterator doesn't require to keep the error.

pkg/catalog/uncommitted_iterator.go
N-o-Z: As mentioned before - due to the size of the PR the tests for the iterator will be deferred to another task

pkg/catalog/uncommitted_iterator.go
```diff
@@ -0,0 +1,133 @@
...
+	if u.entryItr == nil {
+		if u.nextStaging() {
+			u.entryItr.SeekGE(id)
+		}
+	}
+}
+
+func (u *UncommittedIterator) Value() *UncommittedRecord {
+	return u.entry
+}
```
nopcoder, line 116: consider return nil in case err is not nil

pkg/graveler/retention/garbage_collection_manager.go
nopcoder: 1. Not sure that location must end with '/' based on the API we use - I think we should check if it ends with '/' and add if needed.
2. Using + (`location+filename`) looks and usually perform better over `fmt.Sprintf("%s%s", location, filename)`

pkg/graveler/graveler.go
N-o-Z: I prefer not exposing the original method. It was created for a specific reason which is unrelated to GC. Any additional logic that GC may require should be added to the new method.

pkg/graveler/retention/garbage_collection_manager.go
```diff
@@ -46,6 +53,41 @@ func (m *GarbageCollectionManager) GetAddressesLocation(sn graveler.StorageNames
...
+	err = m.blockAdapter.Put(ctx, block.ObjectPointer{
+		Identifier:     identifier,
+		IdentifierType: block.IdentifierTypeFull,
+	}, stat.Size(), fd, block.PutOpts{})
+	if err != nil {
+		return err
+	}
+	return nil
```
nopcoder, line 88:
```suggestion
	return err
```

nopcoder (changes requested):

pkg/catalog/catalog.go
```diff
@@ -1720,6 +1727,145 @@ func (c *Catalog) PrepareExpiredCommits(ctx context.Context, repositoryID string
...
+			return nil, err
+		}
+	}
+	if itr.Err() != nil {
+		return nil, itr.Err()
+	}
+	err = pw.WriteStop()
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
N-o-Z: No sure what is the concern in this instance. There is mo handler that needs to be closed here. If we have an error - the parquet writer can be thrown away - but the underlying fd is being closed with defer

pkg/catalog/catalog.go
```diff
@@ -1720,6 +1727,145 @@ func (c *Catalog) PrepareExpiredCommits(ctx context.Context, repositoryID string
...
+	result := buf.String()
+	return result, nil
+}
+
+type UncommittedParquetObject struct {
+	PhysicalAddress string `parquet:"name=physical_address, type=BYTE_ARRAY, convertedtype=UTF8, encoding=PLAIN_DICTIONARY"`
+	CreationDate    int64  `parquet:"name=creation_date, type=INT64, convertedtype=INT_64"`
+}
+
+func (c *Catalog) writeUncommittedLocal(ctx context.Context, repository *graveler.RepositoryRecord, fd *os.File, mark *GCUncommittedMark) (*GCUncommittedMark, error) {
```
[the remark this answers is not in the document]
N-o-Z: We need the file interface for Stat

pkg/graveler/graveler.go
N-o-Z: Talked with @nopcoder agreed that this will be handled in the controller

pkg/catalog/catalog.go
```diff
@@ -1744,38 +1742,38 @@ func (m *GCUncommittedMark) Encode() (string, error) {
...
+		// Calculation of size deviation by periodicCheckSize value
+		// "data" prefix = 4 bytes
+		// partition id = 20 bytes (xid)
+		// object name = 20 bytes (xid)
+		// timestamp (int64) = 8 bytes
+		//
+		// Total per entry ~52 bytes
+		// Deviation with periodicCheckSize = 100000 will be around 5 MB
+		periodicCheckSize = 100000
+	)
```
nopcoder, line 1768: - Consider moving these constants outside the func for better visibility.
- GCMaxUncommittedFileSize -> gcMaxUncommittedFileSize

pkg/catalog/catalog.go
```diff
@@ -1744,38 +1742,38 @@ func (m *GCUncommittedMark) Encode() (string, error) {
...
+
+	pw, err := writer.NewParquetWriterFromWriter(&w, new(UncommittedParquetObject), parallelNum) // TODO: Play with np count
 	if err != nil {
 		return nil, err
 	}
 	// Compression type
 	pw.CompressionType = parquet.CompressionCodec_GZIP
```
nopcoder, line 1775: nice comment

pkg/catalog/catalog.go
```diff
@@ -1946,3 +1946,26 @@ func newDifferenceFromEntryDiff(v *EntryDiff) (Difference, error) {
 	diff.Type, err = catalogDiffType(v.Type)
 	return diff, err
 }
+
+func NewUncommittedWriter(writer io.Writer) UncommittedWriter {
```
nopcoder, line 1950: I would return a pointer to an implementation in case we like assign the return value to an interface.

nopcoder (approved):
lgtm with minor comments

pkg/catalog/uncommitted_iterator.go
```diff
@@ -51,19 +51,19 @@ func (u *UncommittedIterator) nextStaging() bool {
...
-		} else {
+			u.entry.EntryRecord = u.entryItr.Value()
+			return true
+		}
+		err := u.entryItr.Err()
+		u.entryItr.Close()
+		if err != nil {
 			return false
 		}
 	}
```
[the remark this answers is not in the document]
N-o-Z: Not reachable but still required by linter

johnnyaug (approved):

itaiad200 (approved):
Approved - please add a followup task for the uncommitted iterator tests..

N-o-Z:
Please go over latest changes

PR #4549, opened 2022-11-02
New reposnse code - 499

### LINKED ISSUE #4370: Client cancelled request should not return 5XX response
If a request is terminated early due to cancellation from the client side, we should not treat that as server error. Nginx uses 499 responses for `client closed the connection`. The main incentive to change the response code is to distinguish real server errors from client cancellation. A justified reason to cancel a request is latency, which should be monitored by the appropriate metrics. 

Waiting for completion of #4291 .

itai-david:
close #4370 

Tested manually by running lakefs in debug, and 'pushing' context cancelled on various operations abd verifying the return code

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -63,6 +63,8 @@ const (
 	setupStateNotInitialized = "not_initialized"
 
 	DefaultMaxDeleteObjects = 1000
+
+	HttpStatusRequestCanceled = 499
 )
 
 type actionsHandler interface {
@@ -250,7 +252,7 @@ func (c *Controller) GetPhysicalAddress(w http.ResponseWriter, r *http.Request, 
 		return
 	}
 	if err != nil {
-		writeError(w, http.StatusInternalServerError, err)
+		writeCanceledOrServerError(w, err)
 		return
 	}
 
@@ -260,7 +262,7 @@ func (c *Controller) GetPhysicalAddress(w http.ResponseWriter, r *http.Request, 
 		return
 	}
 	if err != nil {
-		writeError(w, http.StatusInternalServerError, err)
+		writeCanceledOrServerError(w, err)
 		return
 	}
 
@@ -307,7 +309,7 @@ func (c *Controller) LinkPhysicalAddress(w http.ResponseWriter, r *http.Request,
 		return
 	}
 	if err != nil {
-		writeError(w, http.StatusInternalServerError, err)
+		writeCanceledOrServerError(w, err)
 		return
 	}
 	// write metadata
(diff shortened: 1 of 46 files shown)


itaiad200 (changes requested):
Looks great! Can we choose a few and test the new error code? I would suggest `StatObject`, `List`, `Commit`, `Merge`, `Diff` as a start.

pkg/api/controller.go
```diff
@@ -1776,6 +1778,9 @@ func (c *Controller) handleAPIError(ctx context.Context, w http.ResponseWriter,
 	case errors.Is(err, graveler.ErrTooManyTries):
 		writeError(w, http.StatusLocked, "Too many attempts, try again later")
 
+	case errors.Is(err, context.Canceled):
```
itaiad200, line 1781: Just realized something. We should differentiate client and server side cancellation errors. To return `HTTPStatusRequestCanceled` we need to check that `request.Context` was cancelled. Checking the return error for the same might cover all requests cancellation, but it might also cover other reasons for cancellations, some of them are related lakeFS logic, or external components (object store, DB, ...)

itai-david: @itaiad200 - this was actually brought up on the original issue, where you claimed (rightfuly, IMO) that a cancellation might also come from the server side (our logic, or other component).
I currently introduced a general "Request Canceled" without refering to which side the cancellation comes from.
If you want to distinguish the 2, it will require another new code.
How significant is the distinguish between the two? If I understand correctly, if the client cancels the request, it is not really interested in the result, as it is aware of the cancellation. If the client is actually receiving an unexpected "Request Canceled" it can only come from the server side (right? am I missing something?)

itaiad200: Right - if we cancelled the request it should not return `499. `500` is more appropriate. We need to check if the http request context was cancelled and return `499` only if it was.

nopcoder: I've checked the request's context for cancellation for this reason.
About the server side cancel concern - if we are using context.WithTimeout the error will be context deadline exceeded. So unless we have a specific code that cancel the context in the request level it should be same to assume that the cancellation is by the http handler on client closed request.

itaiad200: Not sure I follow @nopcoder. If the context was cancelled by some downstream component, like graveler or pgx client, then the context on the request will not be cancelled and the status code we return to the client isn't 499. The only reason we'll return 499 is if the `request.Context` was cancelled, and we don't control that, right?

itaiad200:
Continued in #4839

PR #4452, opened 2022-11-03
Python lakefs-client library does not have a license

arielshaqed:
Hi,

The lakefs-client Python library is great and works really well as an OpenAPI generated lib. It however does not have a license. We usually check for invalid license types to ensure we can work safely with libraries.

Is the absence of a license intentional?

Thanks!

### DIFF
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -3,6 +3,9 @@ openapi: "3.0.0"
 info:
   description: lakeFS HTTP API
   title: lakeFS API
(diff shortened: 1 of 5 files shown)


talSofer (approved):
Cool!

arielshaqed:
Thanks!
pulling... A weird GitHub API inconsistency allows an issue also to be a PR... but the connected issue enforcer does not know this.

PR #4419, opened 2022-10-20
"Next" button in docs

johnnyaug:
Use the page properties to create a "Next" button at the bottom

image

image

Resolves #4420.

PR #4550, opened 2022-11-03
fix the markdown files exclusion

ortz:
Go tests are running for markdown files by accident.
In order to avoid running tests for files ending with `.md`, we should exclude `**.md` and not `*.md`

See GitHub's documentation [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-including-paths)

Closes https://github.com/treeverse/lakeFS/issues/4551

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -5,7 +5,7 @@ on:
       - master
   pull_request:
     paths-ignore:
-      - "*.md"
+      - "**.md"
       - "docs/**"
       - "design/**"
 jobs:


itaiad200:
The problem is with the mandatory checks. They don't care that you excluded some files and that they won't run. Those checks are configured in the repo level and must be executed.

ortz:
> The problem is with the mandatory checks. They don't care that you excluded some files and that they won't run. Those checks are configured in the repo level and must pass for the merge to complete.

I don't think this is true, they must pass as long as they required.
For instance, take a look at [this PR](https://github.com/treeverse/lakeFS/pull/4419) which didn't run the Go tests, and this happened because the branch name started with `docs/`, where this prefix is excluded properly from the workflow.

itaiad200:
> > The problem is with the mandatory checks. They don't care that you excluded some files and that they won't run. Those checks are configured in the repo level and must pass for the merge to complete.
> 
> I don't think this is true, they must pass as long as they required. For instance, take a look at [this PR](https://github.com/treeverse/lakeFS/pull/4419) which didn't run the Go tests, and this happened because the branch name started with `docs/`, where this prefix is excluded properly from the workflow.

That example doesn't reveal if @johnnyaug forced merge, like the one I did [here](https://github.com/treeverse/lakeFS/pull/4537) for @iddoavn

Issue #2798, opened 2021-12-15
Document which OutputCommitter to use with Spark


ozkatz:
lakeFS Spark integration docs should recommend which OutputCommitter to use, as well as how to configure it.
The docs should probably take into account questions like:

- is the native HadoopFS being used, or the S3 Gateway?
- What is the underlying object store used?
- is it possible for multiple jobs/clusters to write to the same collections on the same branch?
- can we assume failed jobs will be followed by a "reset" action in lakeFS?

It's possible not all of these actually have any impact, but are worth considering.

arielshaqed:
Note that "the" FileOutputCommitter (sort-of default) actually has _two_ available versions, the old v1 and the newer v2 which is recommended **NOT** to be used -- but popular with some users.  So nontrivial recommendation.

johnnyaug:
@arielshaqed can we close this?

PR #4552, opened 2022-11-03
Document all configuration options and clarify lakeFS OC is beta

arielshaqed:


docs/integrations/spark.md
```diff
@@ -389,7 +397,32 @@ On Hadoop 2, configure:
 </configuration>
 ```
 
-(You can also set a default output committer using mapred.output.committer.class and possibly setting mapred.mapper.new-api=false.)
+The lakeFS output committer holds Hadoop working objects on temporary
+branches and merges them back to the output branch when done.  Objects
```
lynnro314, line 401: Consider:
```suggestion
The lakeFS output committer holds Hadoop working objects on temporary lakeFS 
branches and merges them back to the destination location when done.
```

lynnro314 (approved):
Yay, thanks!

docs/integrations/spark.md
```diff
@@ -389,7 +397,32 @@ On Hadoop 2, configure:
...
+appear atomically: it is not possible for another task to see the output
+path with any set of objects except the entire desired output set.
```
talSofer, line 403: A suggestion: replace the order of the how and what. the current version starts with how the foc works, but as a user I would rather understand WIFM first (atomicity).

docs/integrations/spark.md
```diff
@@ -369,7 +374,10 @@ df.write.partitionBy("example-column").parquet(s"lakefs://${repo}/${branch}/outp
 
 The data is now created in lakeFS as new changes in your branch. You can now commit these changes or revert them.
 
-## Using the new OutputCommitter
+## Using the lakeFS OutputCommitter
+
+This feature is currently **in beta**.  It provides atomic writes and
+increases write performance.
```
talSofer, line 380: Not part of this review, but are we planning to document the hadoop 3 way of configuring the foc?

docs/integrations/spark.md
```diff
@@ -24,17 +24,22 @@ Similarly, properties defining lakeFS credentials should be configured in secure
...
-## Two-tiered Spark support
+## Multi-tiered Spark support
 
-lakeFS support in Spark has two tiers:
+lakeFS support in Spark has three tiers:
 
 * Access lakeFS using the [S3A gateway](#access-lakefs-using-the-s3a-gateway).
 * Access lakeFS using the [lakeFS-specific Hadoop
   FileSystem](#access-lakefs-using-the-lakefs-specific-hadoop-filesystem).
+* (**BETA!**) Access lakeFS using lake lakeFS-specific Hadoop FileSystem and
```
talSofer, line 34: I like the fact that you are adding lakeFSFS + FOC as another support tier and not as a configuration option of lakeFSFS!

docs/integrations/spark.md
```diff
@@ -389,7 +397,32 @@ On Hadoop 2, configure:
...
 
-(You can also set a default output committer using mapred.output.committer.class and possibly setting mapred.mapper.new-api=false.)
+The lakeFS output committer holds Hadoop working objects on temporary
+branches and merges them back to the output branch when done.  Objects
+appear atomically: it is not possible for another task to see the output
+path with any set of objects except the entire desired output set.
+
+#### Additional options
+
+* For old Hadoop versions you might set a default output committer using
```
talSofer, line 407: Older than 2? what are we saying here, that the foc may work for older hadoop versions?

talSofer (approved):
Looks good, thanks! 

Added some (non-blocking) questions

docs/integrations/spark.md
arielshaqed: ** *For now!* **  

One day all this will be the default

docs/integrations/spark.md
arielshaqed: Of course.  This is not a current priority.  Also I don't yet know how much I can make it become a default configuration.

docs/integrations/spark.md
arielshaqed: Thanks!  Used your change and changed it further to focus even more on branches.

docs/integrations/spark.md
arielshaqed: Yup!  So I moved it up to the higher-level "using the lakeFS OC" header -- which already states up-front the advantages.

docs/integrations/spark.md
arielshaqed: I'm trying to say "if you're using mapred rather than mapreduce code".  Changed to talk about "mapred" directly.

arielshaqed (commented):
Thanks!

arielshaqed:
Thanks!

I'm linking #2798 to this, but making sure _not_ to close that issue.

PR #4553, opened 2022-11-03
Setting kv driver defaults as configuration

### LINKED ISSUE #4565: Running lakefs with configured local database creates ~ on working directory
Running lakeFS configured using local database, using empty 'local' scope:

```yaml
database:
  type: local
  local:
```

While running lakeFS, a directory "~/metadata" is created under the current working folder. (Note that "~" is not expanded to evaluated as home directory)

nopcoder:
Flag local-settings will enable the local storage and local kv, but the values used by each driver will be set by the configuration and not by each driver's configuration normalization.

The block adapter type will be a required field - the user will have to specify it or use "--local-settings" flag to set it as default.

This change will prevent processing of default as part of each driver and simplify the loading process done by our configuration without splining the default values into two more than one place. Having the configuration log the values will reflect the values we use by each driver.

Fix #4565

### DIFF
--- a/pkg/config/defaults.go
+++ b/pkg/config/defaults.go
@@ -7,8 +7,8 @@ import (
 )
 
 const (
-	DatabaseTypeKey     = "database.type"
-	DefaultDatabaseType = "local"
+	DatabaseTypeKey   = "database.type"
+	LocalDatabaseType = "local"
 
 	DatabaseLocalPathKey     = "database.local.path"
 	DefaultDatabaseLocalPath = "~/lakefs/metadata"
@@ -109,8 +109,8 @@ const (
 	LoggingFilesKeepKey     = "logging.files_keep"
 	LoggingAuditLogLevel    = "logging.audit_log_level"
 
-	AuthEncryptSecretKey        = "auth.encrypt.secret_key"            // #nosec
-	DefaultAuthEncryptSecretKey = "THIS_MUST_BE_CHANGED_IN_PRODUCTION" // #nosec
+	AuthEncryptSecretKey      = "auth.encrypt.secret_key"            // #nosec
+	LocalAuthEncryptSecretKey = "THIS_MUST_BE_CHANGED_IN_PRODUCTION" // #nosec
 
 	ActionsEnabledKey = "actions.enabled"
 
@@ -157,8 +157,8 @@ const (
 
 func setDefaults(local bool) {
 	if local {
-		viper.SetDefault(DatabaseTypeKey, DefaultDatabaseType)
-		viper.SetDefault(AuthEncryptSecretKey, DefaultAuthEncryptSecretKey)
+		viper.SetDefault(DatabaseTypeKey, LocalDatabaseType)
+		viper.SetDefault(AuthEncryptSecretKey, LocalAuthEncryptSecretKey)
 	}
 
 	viper.SetDefault(ListenAddressKey, DefaultListenAddr)
--- a/pkg/config/defaults.go
+++ b/pkg/config/defaults.go
@@ -10,8 +10,29 @@ const (
 	DatabaseTypeKey     = "database.type"
 	DefaultDatabaseType = "local"
 
-	DatabaseKVLocalPath        = "database.local.path"
-	DefaultDatabaseLocalKVPath = "~/lakefs/metadata"
+	DatabaseLocalPathKey     = "database.local.path"
+	DefaultDatabaseLocalPath = "~/lakefs/metadata"
+
+	DatabaseLocalPrefetchSizeKey     = "database.local.prefetch_size"
+	DefaultDatabaseLocalPrefetchSize = 256
+
+	DatabaseDynamodbTableNameKey     = "database.dynamodb.table_name"
+	DefaultDatabaseDynamodbTableName = "kvstore"
+
+	DatabaseDynamodbReadCapacityUnitsKey     = "database.dynamodb.read_capacity_units"
+	DefaultDatabaseDynamodbReadCapacityUnits = 1000
+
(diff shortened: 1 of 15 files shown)


pkg/config/defaults.go
```diff
@@ -134,17 +155,10 @@ const (
...
-	viper.SetDefault(BlockstoreLocalPathKey, DefaultBlockstoreLocalPath)
-	viper.SetDefault(AuthEncryptSecretKey, DefaultAuthEncryptSecretKey)
-	viper.SetDefault(BlockstoreTypeKey, DefaultBlockstoreType)
-}
-
 func setDefaults(local bool) {
 	if local {
-		setDefaultLocalConfig()
+		viper.SetDefault(DatabaseTypeKey, DefaultDatabaseType)
+		viper.SetDefault(AuthEncryptSecretKey, DefaultAuthEncryptSecretKey)
```
guy-har, line 161:
```suggestion
		viper.SetDefault(DatabaseTypeKey, LocalDefaultDatabaseType)
		viper.SetDefault(AuthEncryptSecretKey, LocalDefaultAuthEncryptSecretKey)
```
Just a suggestion, It was a bit hard for me to understand, 1. if they are defaults why set them only if local? 2. why are the other local defaults always set?

guy-har: Maybe even comment in the definition of these values are defaults for the case of --local-settings.

pkg/kv/postgres/store.go
```diff
@@ -42,37 +41,15 @@ const (
...
 	// DefaultPartitions Changing the below value means repartitioning and probably a migration.
 	// Change it only if you really know what you're doing.
-	DefaultPartitions = 100
-
-	DefaultMaxOpenConnections    = 25
-	DefaultMaxIdleConnections    = 25
-	DefaultConnectionMaxLifetime = 5 * time.Minute
-	DefaultScanPageSize          = 1000
+	DefaultPartitions   = 100
+	DefaultScanPageSize = 1000
```
guy-har, line 45: Are these defaults or constants? Didn't see where they could be configured

guy-har (approved):
LGTM
Thanks!

pkg/kv/postgres/store.go
nopcoder: DefaultPartitions also didn't found a way to configure it. The value used as constant. 
Currently the DefaultScanPageSize used by default - without any option to change, just from a test code.

pkg/config/defaults.go
nopcoder: - About the names of the constants - I kept the naming pattern.
- About the defaults - we are talking about the storage type?

nopcoder: I've pushed another commit - removed the local-settings flag.

guy-har (approved):

docs/quickstart/installing.md
```diff
@@ -42,29 +42,29 @@ You can try lakeFS:
...
 If you want to use the JSON file that contains your Google service account key instead of JSON string (as in the previous command) then go to the directory where JSON file is stored and run the command with local parameters:
    ```bash
-docker run --pull always -p 8000:8000 -v $PWD:/myfiles -e LAKEFS_BLOCKSTORE_TYPE='gs' -e LAKEFS_BLOCKSTORE_GS_CREDENTIALS_FILE='/myfiles/YourGoogleServiceAccountKey.json' treeverse/lakefs run --local-settings
+docker run --pull always -p 8000:8000 -v $PWD:/myfiles -e LAKEFS_BLOCKSTORE_TYPE='gs' -e LAKEFS_BLOCKSTORE_GS_CREDENTIALS_FILE='/myfiles/YourGoogleServiceAccountKey.json' treeverse/lakefs run
    ```
 This command will mount your present working directory (PWD) within the container and will read the JSON file from your PWD.
 
 
-Note using the ```--local-settings``` flag, metadata is being stored locally in the lakeFS container. Therefore, avoid using this flag for production usages.
+Note that by default the metadata is being stored locally in the container. Therefore, avoid using this flag for production usages.
```
N-o-Z, line 67: What flag?

N-o-Z (changes requested):
Need to fix the documentation.
Generally I think it is very risky to default to local settings, it allows running lakeFS without any configuration and potentially can result in misuse. That is originally why we added the local settings flag

nopcoder:
Adding checks to see if user use the default and warn about specific defaults: encryption, local storage and local metadata storage.
If the user set a value and not used the default - the warning will be skipped.

N-o-Z:
> Adding checks to see if user use the default and warn about specific defaults: encryption, local storage and local metadata storage. If the user set a value and not used the default - the warning will be skipped.

The visibility of the warnings is too low in my opinion. Automatic deployment certainly won't detect it, and seemingly the server will be up with no issues

docs/quickstart/installing.md
nopcoder: Fixing

nopcoder:
Accept the comments and feedback from @guy-har - reverting the last change to include --local-settings. Just kept the database defaults without normalization.

N-o-Z (changes requested):
Looks good.
2 suggestions:
1. Move the local blockstore default also under the local settings flag
2. Missing a warning when setting local on database type

nopcoder:
> Looks good. 2 suggestions:
> 
> 1. Move the local blockstore default also under the local settings flag
> 2. Missing a warning when setting local on database type

- Added database.type as required configuration key.
- Set blockstore type default by "--local-settings" - as we handle database.type.

N-o-Z (approved):
Thanks for addressing all the comments! Looks awesome!

N-o-Z:
@itaiad200 Added the uncommitted iterator - which added 200 more lines of code instead of 20. Since there's already a lot of content inside this PR, I prefer adding tests to it in a different PR

PR #4574, opened 2022-11-08
Add TS support with generics to router hooks

### LINKED ISSUE #4573: Add TS support to the `useRouter` and `useQuery` hooks
While converting other areas of the code base to TS, I had TS compatibility issues with the router hooks, specifically `useQuery`. Because we don't use `URLSearchParams` directly, we don't get its typings support. This forces you to use an explicit cast `(query as MyQueryStringType).<some property>`, which always feels wrong, or use an explicit `any`, which feels even worse.

Using TS generics improves the situation. The type is passed as a generic parameter to the call to `useQuery`. Using a generic type also has the benefit of declaratively defining the properties this component expects to be in the query string, i.e., its dependency on query string parameters.  

Example usage:
```JavaScript
const query = useQuery<TypeWithExpectedQueryStringParams>();
// This does not generate a TS error anymore and keeps us type-safe
const refId = query.refId;
```

eladlachmi:
Resolves #4573


webui/src/lib/hooks/router.ts
```diff
@@ -0,0 +1,52 @@
...
+  generatePath,
+} from "react-router-dom";
+
+export const useQuery = <T extends Partial<T>>(): T => {
+  const location = useLocation();
+  const result: Record<string, string> = {};
+  for (const [key, value] of new URLSearchParams(location.search).entries()) {
+    result[key] = value;
+  }
+  return result as T;
```
arielshaqed, line 14: I don't remember much JavaScript, but isn't this just
```suggestion
  return Object.fromEntries(new URLSearchParams(location.search).entries()) at T;
```
?

Also not sure about `T extends Partial<T>`; is this a way of restricting the requested type to be an object?

Also note that both versions continue the usual TypeScript lie about object keys.  They are not really strings, they can be numbers as well.  For instance:
```
>>> Object.fromEntries(new URLSearchParams("c=d&e=f&12=13").entries())
Object { 12: "13", "c": "d", e: "f" }
```
has a _number_ key 12, and depends on an implicit conversion to make it seem like it has a string key. 🤷🏽

arielshaqed (approved):
Very neat indeed!

webui/src/lib/hooks/router.ts
```diff
@@ -0,0 +1,52 @@
+import {
+  useHistory,
+  useLocation,
+  useParams,
+  generatePath,
+} from "react-router-dom";
+
+export const useQuery = <T extends Partial<T>>(): T => {
```
arielshaqed, line 8: Also note that this function lies about its return type: it never checks that has all keys for T.  So it's really
```suggestion
export const useQuery = <T>(): Partial<T> => {
```

eladlachmi: Good point! I agree that forcing the consumer to "`undefined` check" the result is a good thing.

webui/src/lib/hooks/router.ts
```diff
@@ -0,0 +1,52 @@
...
+
+interface URLDetails {
+  pathname: string;
+  params: Record<string, string>;
+  query: Record<string, string>;
+}
+
+type URLBuilderInput = URLDetails | string;
+
+export const buildURL = (url: URLBuilderInput): string => {
```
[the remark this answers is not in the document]
eladlachmi: TODO added

webui/src/lib/hooks/router.ts
eladlachmi: It does work, but the [URL spec](https://url.spec.whatwg.org/#interface-urlsearchparams) defines keys and values as strings. I'm guessing the automated conversion is done for convenience, but that's implementation-specific and doesn't _have_ to be the case in all browsers, even if currently it is 😊

arielshaqed: Yeah, I dug around the ECMAScript standard and JavaScript objects were standardized as a mess.  A JavaScript object can have keys of types symbol, number, or string.  Symbols are symbols because of their types, but a key is a number because if it is convertible to a number 🤦🏽 .  Welcome to JavaScript.

[#not-my-meme
welcome to JavaScript where the rules don't matter and datatypes are made up](https://www.memecreator.org/meme/welcome-to-javascript-where-rules-dont-matter-and-datatypes-are-made-up/)

StilLGTM, of course!

PR #4576, opened 2022-11-08
4575 upgrade vite 3.2.3

### LINKED ISSUE #4575: Upgrade Vite to v3.2.3
Requires node >= 16 to run smoothly. Shouldn't be an issue.

eladlachmi:
Upgraded Vite to v3.2.3 + updated plugins as required/recommended.
I've run through all the scenarios I can think of in dev mode, and everything looks to be working as expected.  
If you have any thoughts on things I should try, please let me know.  

I also fixed something in the loading of `RepositoryObjectsPage`. It's nearly unnoticeable when running in prod mode, but has been annoying me for a while with HMR in dev.

Resolves #4575

### DIFF
--- a/webui/package.json
+++ b/webui/package.json
@@ -31,6 +31,7 @@
     "validator": "^13.7.0"
   },
   "devDependencies": {
+    "@rollup/plugin-replace": "^4.0.0",
     "@types/bootstrap": "^4.6.2",
     "@types/jest": "^29.0.3",
     "@types/lodash": "^4.14.185",
@@ -42,8 +43,7 @@
     "@types/validator": "^13.7.7",
     "@typescript-eslint/eslint-plugin": "^5.38.0",
     "@typescript-eslint/parser": "^5.38.0",
-    "@rollup/plugin-replace": "^4.0.0",
-    "@vitejs/plugin-react-refresh": "^1.3.1",
+    "@vitejs/plugin-react": "^2.2.0",
     "babel-eslint": "^10.1.0",
     "eslint-plugin-react": "^7.23.2",
     "jest": "^29.0.3",
@@ -52,7 +52,7 @@
     "ts-node": "^10.9.1",
     "typescript": "^4.8.3",
     "typesync": "^0.9.2",
-    "vite": "^2.3.7",
-    "vite-plugin-eslint": "^1.1.2"
+    "vite": "^3.2.3",
+    "vite-plugin-eslint": "^1.8.1"
   }
 }
--- a/webui/package.json
+++ b/webui/package.json
@@ -6,7 +6,7 @@
   },
   "scripts": {
     "dev": "vite",
-    "build": "vite build",
+    "build": "cross-env NODE_OPTIONS=--max-old-space-size=4096 vite build",
     "serve": "vite preview",
     "test": "jest src",
     "typesync": "npx typesync"
@@ -48,6 +48,7 @@
     "@typescript-eslint/parser": "^5.38.0",
     "@vitejs/plugin-react": "^2.2.0",
     "babel-eslint": "^10.1.0",
+    "cross-env": "^7.0.3",
     "eslint-plugin-react": "^7.23.2",
     "jest": "^29.0.3",
     "react-test-renderer": "^17.0.2",
--- a/webui/package.json
+++ b/webui/package.json
@@ -1,6 +1,9 @@
 {
   "name": "lakefs-ui",
   "version": "0.1.0",
+  "engines": {
+    "node": ">= 16.16.0"
+  },
   "scripts": {
     "dev": "vite",
     "build": "vite build",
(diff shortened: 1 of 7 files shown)


webui/vite.config.js
```diff
@@ -1,6 +1,9 @@
-import reactRefresh from '@vitejs/plugin-react-refresh';
+import react from '@vitejs/plugin-react';
 import eslintPlugin from 'vite-plugin-eslint';
 import replace from '@rollup/plugin-replace';
+import dns from "dns";
+
+dns.setDefaultResultOrder('verbatim');
```
nopcoder, line 6: Not sure we need this one - the first request is the minimum nodejs version it requires and the fact that I don't have a problem dns resolved 'localhost' on local development.
Did you had an issue with the default dns resolution?

nopcoder (changes requested):
One comment about the DNS and looks like the current node crashes with the new code :/

johnnyaug (approved):
LGTM

webui/vite.config.js
eladlachmi: @nopcoder I didn't have an issue either until the Vite v3 upgrade. Without this Vite mounts on `127.0.0.1` and does not resolve `localhost` correctly (at least on my machine - node 16.16.0.)  Come to think of it, while checking this, I also fixed the default port (back to 3000), so I'll double-check the need for modifying the DNS resolution locally again to be sure it's needed.

I see that the GitHub agent is running node 16.17.1. I'll `nvm` that locally and see if it's a node version issue or a GitHub agent memory issue.

eladlachmi: Also, probably a good time to add a `.nvmrc` file to declare the preferred node version

nopcoder: From what I read from the docs the issue is in case your dns localhost resolve don't not point to 127.0.0.1 the request may go to a different server - think it is a rare case and I am fan of using defaults.

nopcoder: nvm is a specific tool - prefer to specify it in package.json using:
```json
 "engines" : { 
    "node" : ">=16.0.0"
  }
```

eladlachmi: Good point. I'll do that.

eladlachmi: @nopcoder Fixed + tests are now passing

nopcoder (changes requested):
We need to add `--max-old-space-size=4096` to the relevant code gen step in the Makefile in order not to communicate it on every workflow and slack.

nopcoder:
@eladlachmi one more thing related to the memory issue node have with vite:
1. Not crashing with my node v18 - so maybe there is a fix for the 16.x branch too
2. Check the option to disable the map files

eladlachmi:
> We need to add `--max-old-space-size=4096` to the relevant code gen step in the Makefile in order not to communicate it on every workflow and slack.

@nopcoder I think I'll take this even one step further and add it to the npm script for `build`
IMHO it's closest to the "source" of the issue

eladlachmi:
Removed `sourcemaps` from the production build. They aren't served anyway and just make the binary larger, since they are embedded. For the dev build I've set `sourcemaps` to `inline`, which improved the dev experience in Chrome devtools.

webui/package.json
```diff
@@ -6,7 +6,7 @@
   },
   "scripts": {
     "dev": "vite",
-    "build": "vite build",
+    "build": "cross-env NODE_OPTIONS=--max-old-space-size=4096 vite build",
```
nopcoder, line 9: Nice - didn't used cross-env before (don't like extra packages). My thought was adding it to the Makefile so no specific value is forced into my build. or in my case when I'm trying 'bun' to build the code.

webui/vite.config.js
```diff
@@ -47,7 +44,10 @@ export default ({ command }) => {
             secure: false
           }
         }
-      }
+      },
+      build: {
+        sourcemap: 'inline',
```
nopcoder, line 49: did you tried to turn it off to see if it remove the requirements for --max-old-space-size for the production build?

eladlachmi: @nopcoder The default is off. This is just for the dev build.
Again, even if it does help, it's temporary. I don't see the point in avoiding it. You're just kicking the can down the road. I'd rather handle it now, while I'm in the context than in two months when we'll have to have the entire discussion again 😉

webui/package.json
eladlachmi: @nopcoder I actually think the npm scripts is a great place to keep build-related node params. I think it's the right place for them. This is so inconsequential that I'm starting to feel a bit silly going on about it 😊

nopcoder (approved):
Approve in hope we will find a packaging tool based on non node.js soon.

eladlachmi:
> Approve in hope we will find a packaging tool based on non node.js soon.

@nopcoder It's still a ways off, but [this](https://rome.tools/blog/2022/11/08/rome-10/) looks promising

PR #4579, opened 2022-11-09
[GC] Validate bucket's correct region in AWS S3

### LINKED ISSUE #4583: GC: Use the correct region-bucket combination
If lakeFS is configured in region X, and a repository in that lakeFS installation is configured in region Y, and GC will be injected with X as the region to work on and will fail with:
Caused by: com.amazonaws.services.s3.model.AmazonS3Exception: The bucket is in this region: Y. Please use this region to retry the request

Jonathan-Rosenberg:
In case of a mismatch of the input region and the bucket's location, get the correct region and use it.
I kept the input region (even though it's not needed with this implementation) to keep it backward compatible usage-wise.
To use this functionality, you should add the `s3:GetBucketLocation` permission to your bucket's policy or to the role used to access the bucket, with the resource being the bucket itself.

## How was this tested?

1. Using a lakeFS Cloud installation located in `us-west-2`:
    1. Created a repo with a  bucket located in `ca-central-1`.
    2. Defined GC policy.
    3. Update the GC role to include the `s3:GetBucketLocation` action.
    4. Ran GC.
    5. Result:.
2. Using a lakeFS Cloud installation located in `us-west-2`:
    1. Created a repo with a  bucket located in `us-east-1`.
    2. Defined GC policy.
    3. Ran GC.
    4. Result: failure- the `s3:GetBucketLocation` action wasn't set.
3. Using a lakeFS Cloud installation located in `us-west-2`:
    1. Created a repo with a  bucket located in `us-west-2`.
    2. Defined GC policy.
    3. Ran GC.
    4. Result: success.

Closes #4583


clients/spark/core/src/main/hadoop2/scala/io/treeverse/clients/conditional/S3ClientBuilder.scala
```diff
@@ -28,16 +30,49 @@ object S3ClientBuilder extends io.treeverse.clients.S3ClientBuilder {
         )
       } else None
 
+    initializeClient(configuration, credentialsProvider, region, bucket)
+  }
+
+  private def initializeClient(
+      config: ClientConfiguration,
```
arielshaqed, line 37: Let's keep a single name, `configuration` or `config`: we read code more than we type it, and reading this function I wondered whether it was the same "configuration" from above.

clients/spark/core/src/main/hadoop2/scala/io/treeverse/clients/conditional/S3ClientBuilder.scala
```diff
@@ -28,16 +30,49 @@ object S3ClientBuilder extends io.treeverse.clients.S3ClientBuilder {
...
     val builderWithCredentials = credentialsProvider match {
       case Some(cp) => builder.withCredentials(cp)
       case None     => builder
     }
+    val client = builderWithCredentials.build
+
+    if (validateClientCorrectness(client, bucket)) {
+      return client
+    }
+    val bucketRegion = getAWSS3Region(client, bucket)
```
arielshaqed, line 55: Can we add a log here and before line 53?  It can really help me debug connection failures if I know which path this code took!

clients/spark/core/src/main/hadoop2/scala/io/treeverse/clients/conditional/S3ClientBuilder.scala
```diff
@@ -28,16 +30,49 @@ object S3ClientBuilder extends io.treeverse.clients.S3ClientBuilder {
...
+    builder.withRegion(bucketRegion).build
+  }
+
+  private def validateClientCorrectness(client: AmazonS3, bucket: String): Boolean = {
+    try {
+      client.headBucket(new HeadBucketRequest(bucket))
+      true
+    } catch {
+      case _: Exception => false
```
arielshaqed, line 64: This swallows all errors silently, which can make debugging harder.  Things that could make it a bit easier:

* Log the exception.
* Catch a more restricted set of exceptions, for example only AWS exceptions from `headBucket`.  Or report those differently.

Otherwise if I have the right region but my headBucket fails because of a sporadic issue and I have no permission to getBucketLocation, my GC fails (which is sort-of OK) but I don't understand why.

clients/spark/core/src/main/hadoop2/scala/io/treeverse/clients/conditional/S3ClientBuilder.scala
```diff
@@ -28,16 +30,49 @@ object S3ClientBuilder extends io.treeverse.clients.S3ClientBuilder {
...
+  }
 
-    builderWithCredentials.build
+  private def getAWSS3Region(client: AmazonS3, bucket: String): String = {
+    val bucketRegion = client.getBucketLocation(bucket)
+    val region = Region.fromValue(bucketRegion)
+    // The comparison `region.equals(Region.US_Standard))` is required due to AWS's backward compatibility:
+    // https://github.com/aws/aws-sdk-java/issues/1470.
+    // "us-east-1" was previously called "US Standard". This resulted in a return value of "US" when
+    // calling `client.getBucketLocation(bucket)`.
```
arielshaqed, line 74: Thanks, very helpful explanation!

[Best Java SDK is Amazon AWS v1 Java SDK.
Kim "best Korea" meme](https://imgflip.com/i/707w5o)

clients/spark/core/src/main/hadoop3/scala/io/treeverse/clients/conditional/S3ClientBuilder.scala
```diff
@@ -36,15 +38,50 @@ object S3ClientBuilder extends io.treeverse.clients.S3ClientBuilder {
         )
       } else None
 
+    initializeClient(configuration, credentialsProvider, region, bucket)
```
arielshaqed, line 41: This is a lot of code duplication between the Hadoop 2 and Hadoop 3 variations.  It suggests that We (in the sense of the royal "We", I'm ashamed to say) put the conditional API in the wrong place.  Can you please open an issue to refactor the API along more useful lines?  For instance we could have a method `prepareMostOfTheAWSClientBuilderButLeaveMeTheBitWhereISetTheCredentials` (name open to discussion) that is shared between all both variations.

arielshaqed (approved):
Extremely very neat!

We should document this before releasing to users: it is in effect a hidden configuration option (that happens to go through IAM).  That makes it easy to use, but it is hard to know what to do.

clients/spark/core/src/main/hadoop2/scala/io/treeverse/clients/conditional/S3ClientBuilder.scala
```diff
@@ -28,16 +30,49 @@ object S3ClientBuilder extends io.treeverse.clients.S3ClientBuilder {
...
+    val client = builderWithCredentials.build
+
+    if (validateClientCorrectness(client, bucket)) {
+      return client
+    }
+    val bucketRegion = getAWSS3Region(client, bucket)
+    builder.withRegion(bucketRegion).build
+  }
+
+  private def validateClientCorrectness(client: AmazonS3, bucket: String): Boolean = {
```
talSofer: I would use a more specific name, something like 
```suggestion
  private def validateClientAndBucketRegionsMatch(client: AmazonS3, bucket: String): Boolean = {
```

talSofer (approved):
Thanks @Jonathan-Rosenberg!
Can we add unit-tests?

clients/spark/core/src/main/hadoop2/scala/io/treeverse/clients/conditional/S3ClientBuilder.scala
Jonathan-Rosenberg: What kind of log?

clients/spark/core/src/test/scala/io/treeverse/clients/StorageUtilsSpec.scala
```diff
@@ -0,0 +1,146 @@
...
+      val headBucketRequest: RecordedRequest = server.takeRequest()
+      val getLocationRequest: RecordedRequest = server.takeRequest()
+      extractBucketFromRecordedRequest(headBucketRequest) should equal(BUCKET_NAME)
+      headBucketRequest.getMethod should equal(HttpMethod.HEAD.toString)
+      getLocationRequest.getMethod should equal(HttpMethod.GET.toString)
+    }
+  }
+
+  private def extractBucketFromRecordedRequest(request: RecordedRequest): String = {
+    val splicedRequestLine = request.getRequestLine.split('/')
```
talSofer, line 118:
```suggestion
    val splitedRequestLine = request.getRequestLine.split('/')
```
Is that what you meant?

talSofer (approved):
Nice work on unit tests!! thanks for spending time on it

clients/spark/core/src/test/scala/io/treeverse/clients/StorageUtilsSpec.scala
Jonathan-Rosenberg:
```suggestion
    val splitRequestLine = request.getRequestLine.split('/')
```
It should actually be this... Probably fell into IntelliJ's spelling correction trap...
Thanks!

clients/spark/build.sbt
```diff
@@ -72,7 +75,9 @@ def generateCoreProject(buildType: BuildType) =
         "org.scalactic" %% "scalactic" % "3.2.9",
         "dev.failsafe" % "failsafe" % "3.2.4",
         "org.apache.hadoop" % "hadoop-distcp" % buildType.hadoopVersion,
-        "org.scalatestplus" %% "mockito-3-4" % "3.2.9.0" % "test",
+        "org.scalatestplus" %% "mockito-4-6" % "3.2.14.0" % "test",
+        // https://mvnrepository.com/artifact/com.squareup.okhttp3/mockwebserver
+        "com.squareup.okhttp3" % "mockwebserver" % "4.10.0" % Test,
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: According to [this](https://www.scala-sbt.org/release/docs/Scopes.html#Scoping+by+the+configuration+axis), `"test"` is an Ivy configuration, while `Test` is based on it and is the newer version of dependency configurations.
To make it uniform I'll change it to `"test"`

clients/spark/core/src/main/scala/io/treeverse/clients/StorageUtils.scala
```diff
@@ -80,5 +87,71 @@ object StorageUtils {
...
+    ): AmazonS3 = {
+
+      require(awsS3ClientBuilder != null)
+      require(bucket.nonEmpty)
+
+      var client =
+        initializeS3Client(configuration, credentialsProvider, awsS3ClientBuilder, region)
+
+      if (!validateClientAndBucketRegionsMatch(client, bucket)) {
+        val bucketRegion = getAWSS3Region(client, bucket)
```
arielshaqed, line 106: Continuing previous comment thread about logging here: I would like this line to contain a log explaining that there was a mismatch and that we got the region from S3.  It should probably follow the general strategy for branch logs: what we did and why (input parameters) and what was the result.  So e.g.
```scala
logger.info(s"Bucket ${bucket} not in region ${region}, discovered it in region ${bucketRegion}")
```

clients/spark/core/src/main/scala/io/treeverse/clients/StorageUtils.scala
```diff
@@ -80,5 +87,71 @@ object StorageUtils {
...
     }
+
+    def createAndValidateS3Client(
+        configuration: ClientConfiguration,
+        credentialsProvider: Option[AWSCredentialsProvider],
+        awsS3ClientBuilder: AmazonS3ClientBuilder,
+        region: String,
+        bucket: String
+    ): AmazonS3 = {
+
```
arielshaqed, line 98: Nit: no blank lines at tops and bottoms of functions.  (No, we do not have a standard for this, I just like to squash more lines onto my screen...)

```suggestion
```

clients/spark/core/src/test/scala/io/treeverse/clients/StorageUtilsSpec.scala
```diff
@@ -0,0 +1,145 @@
...
+  private val US_WEST_2 = "us-west-2"
+  private val BUCKET_NAME = "bucket"
+
+  before {
+    server = new MockWebServer
+    server.start()
+    awsS3ClientBuilder = AmazonS3ClientBuilder.standard().withPathStyleAccessEnabled(true)
+    val baseUrl = server.url("/")
+    clientConfiguration = generateS3ClientConfigurations(baseUrl)
+  }
```
arielshaqed, line 38: I don't understand this function (AFAIU `before` is just a function that takes a function written in a weird way) in conjunction with all the `var`s above.  If we had this block:
```scala
  private val awsS3ClientBuilder  = AmazonS3ClientBuilder.standard().withPathStyleAccessEnabled(true)
  private val server = new MockWebServer
  private val clientConfiguration = {
    val baseUrl = server.url("/)
    generateS3ClientConfigurations(baseUrl)
  }
```
then we could have here just
```suggestion
  before {
    server.start()
  }
```

I believe that would be clearer, and _add_ immutability.

clients/spark/core/src/main/scala/io/treeverse/clients/StorageUtils.scala
```diff
@@ -80,5 +87,71 @@ object StorageUtils {
...
+        region: String
+    ): AmazonS3 = {
+      val builder = awsS3ClientBuilder
+        .withClientConfiguration(configuration)
+        .withRegion(region)
+      val builderWithCredentials = credentialsProvider match {
+        case Some(cp) => builder.withCredentials(cp)
+        case None     => builder
+      }
+      builderWithCredentials.build
```
arielshaqed, line 129: I think there is some Scala style that recommends using bracket-less calls _only_ for code with no side-effects.  `Builder.build` returns a `new` object, so (I think) this would be clearer as
```suggestion
      builderWithCredentials.build()
```

clients/spark/core/src/test/scala/io/treeverse/clients/StorageUtilsSpec.scala
```diff
@@ -0,0 +1,145 @@
...
+      val getLocationRequest: RecordedRequest = server.takeRequest()
+      extractBucketFromRecordedRequest(headBucketRequest) should equal(BUCKET_NAME)
+      headBucketRequest.getMethod should equal(HttpMethod.HEAD.toString)
+      getLocationRequest.getMethod should equal(HttpMethod.GET.toString)
+    }
+  }
+
+  private def extractBucketFromRecordedRequest(request: RecordedRequest): String = {
+    val splitRequestLine = request.getRequestLine.split('/')
+    splitRequestLine(splitRequestLine.length - 3)
```
arielshaqed, line 119: This will fail strangely if the request is _really_ bad.  Maybe wrap the failure?  (OTOH, it's not like it's a particularly *interesting* failure mode...)

arielshaqed (approved):
Neat, thanks!

clients/spark/core/src/test/scala/io/treeverse/clients/StorageUtilsSpec.scala
Jonathan-Rosenberg: `server.start` will fail if not re-initializing the mock server... It's the package's fault.
I'll reposition the others.
Thanks!

clients/spark/core/src/main/scala/io/treeverse/clients/StorageUtils.scala
Jonathan-Rosenberg: that was actually an IntelliJ suggestion 🤷‍♂️

clients/spark/core/src/test/scala/io/treeverse/clients/StorageUtilsSpec.scala
Jonathan-Rosenberg: the response is generated by the function `generateGetBucketLocationResponseWithRegion` so no worries about that.

PR #4580, opened 2022-11-09
created basic java test

lynnro314:


clients/hadoopfs/pom.xml
```diff
@@ -256,6 +256,13 @@ To export to S3:
             <version>3.12.0</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>spark-sql_2.12</artifactId>
+            <version>3.0.1</version>
+            <scope>provided</scope>
```
arielshaqed, line 263: I don't think that this code changes the client in any way, only the tests.  So I believe that the client still does not have a Spark dependency -- only the test.  Can we instead try this?

```suggestion
            <scope>test</scope>
```

clients/hadoopfs/src/test/java/io/lakefs/committer/LakeFSOutputCommitterTest.java
```diff
@@ -0,0 +1,155 @@
...
+    protected String s3Bucket;
+
+    protected SparkSession spark;
+
+    private static final DockerImageName MINIO = DockerImageName.parse("minio/minio:RELEASE.2021-06-07T21-40-51Z");
+    protected static final String S3_ACCESS_KEY_ID = "AKIArootkey";
+    protected static final String S3_SECRET_ACCESS_KEY = "secret/minio/key=";
+
+    @Rule
+    public final GenericContainer s3 = new GenericContainer(MINIO.toString()).
```
arielshaqed, line 72: Can we reduce visibility?
```suggestion
    protected final GenericContainer s3 = new GenericContainer(MINIO.toString()).
```

clients/hadoopfs/src/test/java/io/lakefs/committer/LakeFSOutputCommitterTest.java
```diff
@@ -0,0 +1,155 @@
...
+                .set("fs.lakefs.endpoint", "http://localhost:8000/api/v1")
+                .set("fs.s3a.impl", "org.apache.hadoop.fs.s3a.S3AFileSystem")
+                .set(org.apache.hadoop.fs.s3a.Constants.ACCESS_KEY, S3_ACCESS_KEY_ID)
+                .set(org.apache.hadoop.fs.s3a.Constants.SECRET_KEY, S3_SECRET_ACCESS_KEY)
+                .set(org.apache.hadoop.fs.s3a.Constants.ENDPOINT, s3Endpoint);
+        this.spark = SparkSession.builder().config(sparkConf).getOrCreate();
+    }
+
+
+    @Test
```
arielshaqed, line 122: We should disable this test until we include and wire up a lakeFS container: right now it fails CI. :-(

clients/hadoopfs/src/test/java/io/lakefs/committer/LakeFSOutputCommitterTest.java
```diff
@@ -0,0 +1,155 @@
...
+    public void testWriteTextfile() throws ApiException, IOException {
+        String output = "lakefs://example/main/some_df";
+
+        StructType structType = new StructType();
+        structType = structType.add("A", DataTypes.StringType, false);
```
arielshaqed, line 127: I think we could instead
```suggestion
        StructType structType = new StructType().add("A", DataTypes.StringType, false);
```

clients/hadoopfs/src/test/java/io/lakefs/committer/LakeFSOutputCommitterTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+        List<Row> nums = new ArrayList<Row>();
+        nums.add(RowFactory.create("value1"));
+        nums.add(RowFactory.create("value2"));
```
arielshaqed, line 131: Again, I probably prefer
```suggestion
        List<Row> nums = ImmutableList.of(
        RowFactory.create("value1"),
        RowFactory.create("value2"));
```
using ImmutableList from Guava (Google Commons).

clients/hadoopfs/src/test/java/io/lakefs/committer/LakeFSOutputCommitterTest.java
```diff
@@ -0,0 +1,155 @@
...
+        Dataset<Row> df = spark.createDataFrame(nums, structType);
```
arielshaqed, line 132: As an alternative, we might use `parallelize`.  But truly your call.

clients/hadoopfs/src/test/java/io/lakefs/committer/LakeFSOutputCommitterTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+        ArrayList<String> failed = new ArrayList<String>();
+
+        // TODO(Lynn): Add relevant file formats
+        ArrayList<String> fileFormats = new ArrayList<String>(Arrays.asList("text"));
```
arielshaqed, line 137: Why convert twice?

```suggestion
        List<String> fileFormats = Arrays.asList("text");
```

Or

```suggestion
        List<String> fileFormats = ImmutableList.of("text");
```

clients/hadoopfs/src/test/java/io/lakefs/committer/LakeFSOutputCommitterTest.java
```diff
@@ -0,0 +1,155 @@
...
+        for (String fileFormat : fileFormats) {
+            try {
+                // save data in selected format
+                String outputPath = String.format("%s.%s", output, fileFormat);
+                df.write().format(fileFormat).save(outputPath);
+
+                // read the data we just wrote
+                Dataset<Row> df_read = spark.read().format(fileFormat).load(outputPath);
+                Dataset<Row> diff = df.exceptAll(df_read);
+                Assert.assertTrue(diff.isEmpty());
```
arielshaqed, line 147: This tests one direction: that all rows written were actually in the input.  It is also interesting to verify that all rows in the input were actually written.  This code should do that too:
```suggestion
                Dataset<Row> diff = df.exceptAll(df_read);
                Assert.assertTrue("all rows written were read", diff.isEmpty());
                diff = df_read.exceptAll(df);
                Assert.assertTrue("all rows read were written", diff.isEmpty());
```
We might also stringify the first 20 rows of diff, so that if it fails we have some examples.  Something like this (but I did not test that it works!):
```suggestion
                Dataset<Row> diff = df.exceptAll(df_read);
                Assert.assertTrue(String.format("expected all rows written were read but go %s", diff.take(20).collect), diff.isEmpty());
                diff = df_read.exceptAll(df);
                Assert.assertTrue("...same...", diff.isEmpty());
```

clients/hadoopfs/src/test/java/io/lakefs/committer/LakeFSOutputCommitterTest.java
```diff
@@ -0,0 +1,155 @@
...
+            } catch (Exception e) {
+                failed.add(fileFormat);
```
arielshaqed, line 149: Can we instead either fail here by using assertNoThrows, or alternatively add the exception to the list of failures so that we can print it on failure, or at least log the failure to the JUnit output so that we can read _what_ failed?

arielshaqed (changes requested):
VERY neat, thanks!

A few small comments:
* Need to check both directions with exceptAll to get a real diff
* Improved error messages
* Please skip the test until it passes on CI.

THANKS!

clients/hadoopfs/src/test/java/io/lakefs/committer/LakeFSOutputCommitterTest.java
lynnro314: This must be public.

clients/hadoopfs/src/test/java/io/lakefs/committer/LakeFSOutputCommitterTest.java
arielshaqed: Ouch.  So what [StackOverflow, the source of knowledge](https://stackoverflow.com/a/17731324/192263) says is just to do nothing about and let the exception fail the error. Your way is better, though -- it tests _all_ formats and doesn't fail on the first one.

clients/hadoopfs/src/test/java/io/lakefs/committer/LakeFSOutputCommitterTest.java
```diff
@@ -0,0 +1,163 @@
...
+                diff = df_read.exceptAll(df);
+                Assert.assertTrue("all rows read were written", diff.isEmpty());
+
+            } catch (Exception e) {
+                logger.error(String.format("Format %s unexpected exception: %s", fileFormat, e));
+                failed.add(fileFormat);
+            }
+        }
+
+        Assert.assertEquals("Failed list is not empty: %s" , 0, failed.size());
```
arielshaqed, line 161: Using assertThat with an appropriate matcher, as [suggested here](https://stackoverflow.com/a/51279675/192263), might be both more readable _and_ give a better error message!

arielshaqed (approved):
Thanks!

PR #4590, opened 2022-11-10
Prepare garbage collection uncommitted API

### LINKED ISSUE #4561: Create PrepareUncommittedForGC swagger API
Create the swagger API for PrepareUncommittedForGC

This is part of: https://github.com/treeverse/lakeFS/issues/4475

nopcoder:
Close #4561


api/swagger.yml
```diff
@@ -981,6 +981,26 @@ components:
...
+      properties:
+        continuation_token:
+          type: string
+
+    PrepareGCUncommittedResponse:
+      type: object
+      properties:
+        run_id:
+          type: string
+        location:
```
N-o-Z, line 995: We might want to add the location of the copy object file list to this command as well.
Suggest to change the name to be more specific

pkg/api/controller.go
```diff
@@ -94,6 +95,91 @@ type Controller struct {
...
+			Action:   permissions.PrepareGarbageCollectionUncommittedAction,
+			Resource: permissions.RepoArn(repository),
+		},
+	}) {
+		return
+	}
+	ctx := r.Context()
+	c.LogAction(ctx, "prepare_garbage_collection_uncommitted", r, repository, "", "")
+
+	// TODO(barak): move this code the a Decode func in catalog
```
N-o-Z, line 133: Is this comment still relevant?

pkg/api/controller.go
```diff
@@ -2328,7 +2414,7 @@ func (c *Controller) PrepareGarbageCollectionCommits(w http.ResponseWriter, r *h
 	}
 	ctx := r.Context()
 	c.LogAction(ctx, "prepare_garbage_collection_commits", r, repository, "", "")
-	gcRUnMetadata, err := c.Catalog.PrepareExpiredCommits(ctx, repository, swag.StringValue(body.PreviousRunId))
+	gcRUnMetadata, err := c.Catalog.PrepareExpiredCommits(ctx, repository, StringValue(body.PreviousRunId))
```
N-o-Z, line 2417:
```suggestion
	gcRunMetadata, err := c.Catalog.PrepareExpiredCommits(ctx, repository, StringValue(body.PreviousRunId))
```

pkg/api/controller.go
```diff
@@ -94,6 +95,91 @@ type Controller struct {
...
+			WithFields(logging.Fields{"repository": repository, "continuation_token": continuationToken}).
+			Error("PrepareGCUncommitted failed")
+		c.handleAPIError(ctx, w, err)
+		return
+	}
+
+	// continuation token
+	var nextContinuationToken *string
+	if uncommittedInfo.Mark != nil {
+		nextMark, err := encodeGCUncommittedMark(uncommittedInfo.Mark)
```
N-o-Z, line 160: Should remove the encode method from mark

pkg/catalog/catalog.go
```diff
@@ -1871,8 +1857,11 @@ func (c *Catalog) PrepareGCUncommitted(ctx context.Context, repositoryID string,
 		return nil, err
 	}
 	defer func() {
-		fd.Close()
-		os.Remove(fd.Name())
+		_ = fd.Close()
+		err := os.Remove(fd.Name())
+		if err != nil {
+			c.log.WithField("filename", fd.Name()).Warn("Failed to delete temporary gc uncommitted data file")
```
N-o-Z, line 1863:
```suggestion
		if err := os.Remove(fd.Name()); err != nil {
			c.log.WithField("filename", fd.Name()).Warn("Failed to delete temporary gc uncommitted data file")
```

pkg/api/controller_test.go
```diff
@@ -3083,3 +3083,90 @@ func TestController_GetPhysicalAddress_(t *testing.T) {
...
+				ContinuationToken: continuationToken,
+			})
+			verifyResponseOK(t, resp, err)
+			if resp.JSON201.RunId == "" {
+				t.Errorf("PrepareGarbageCollectionUncommittedWithResponse empty RunID, value expected")
+			}
+			if resp.JSON201.Location == "" {
+				t.Errorf("PrepareGarbageCollectionUncommittedWithResponse empty Location, value expected")
+			}
+			token := api.StringValue(resp.JSON201.ContinuationToken)
```
N-o-Z, line 3134: We should validate that token == "" for items = 3, it is a bug if we get a different result

pkg/api/controller_test.go
```diff
@@ -3083,3 +3083,90 @@ func TestController_GetPhysicalAddress_(t *testing.T) {
...
+			if resp.JSON201.Location != "" {
+				t.Errorf("PrepareGarbageCollectionUncommittedWithResponse Location=%s, expected empty", resp.JSON201.Location)
+			}
+			token := api.StringValue(resp.JSON201.ContinuationToken)
+			if token == "" {
+				break
+			}
+		}
+	})
+}
```
N-o-Z, line 3172: Maybe add a check the ensures we get at least 1 cont token?

N-o-Z (changes requested):
Looks good!
Some comments / suggestions

api/swagger.yml
nopcoder: gc_uncommitted_location

pkg/api/controller.go
nopcoder: fixing

pkg/api/controller.go
nopcoder: Don't think there is Encode anymore

pkg/api/controller_test.go
nopcoder: Endless loop, that will cause timeout - but I'll add verification for both committed and uncommitted.

pkg/api/controller_test.go
nopcoder: Will depend on the catalog test for this one as generating 20MB using the API during tests will just slow the system down.
We will need to have this parameter configurable and test the token.

N-o-Z (approved):
LGTM!

Issue #4475, opened 2022-10-26
GC+: Implement PrepareUncommittedForGC API


N-o-Z:
A new API which will create meta-ranges and ranges for a given branch using its uncommitted data. These files
will be saved in a designated path used by the GC client to list branch's uncommitted objects. 

**Input:** branch_id, run_id

**Flow:**
- List repository uncommitted data
- List copied objects from ref-store
- Write data to parquet file (physical address + creation date)
- Upload parquet files to object store at: `_lakefs/retention/gc/uncommitted<run_id>/metadata/`

Breakdown of task:
[Implement GC Save Uncommitted](https://github.com/treeverse/lakeFS/issues/4547#top)
[Create PrepareUncommittedForGC swagger API](https://github.com/treeverse/lakeFS/issues/4561#top)

PR #4592, opened 2022-11-13
Initial GC+ client logic

### LINKED ISSUE #4488: GC+: Implement uncommitted GC main logic
Pre-requirements:
1. Uncommitted format
2. Committed format 

Create logic which lists object store into DF, and performs the minus operation with the committed and uncommitted data

Opens:
1. Define clearly how to list and partition data from the object store
2. How to split the work between workers

johnnyaug:
Resolves #4488.

This change introduces the most basic GC+ login on the client side.

### Description
1. List all addresses of data found in the storage namespace.
2. List all addresses found in Graveler ranges.
3. List all uncommitted entries.
4. Write a dataframe with the result of (1) - (2 + 3).

### Notes
1. The implementation currently assumes the uncommitted entries are listed in a Parquet file somewhere in the storage. Later, we will use the [dedicated API (see #4475).
6. We list the data from the storage without any parallelization. Of course, this is not scalable.
7. We always get all ranges (=committed metadata) from the storage namespace, without any optimization. This will be improved on the next milestone, when we implement the optimization part of the #4382.


clients/spark/core/src/main/scala/io/treeverse/gc/DataLister.scala
```diff
@@ -0,0 +1,25 @@
...
+ */
+trait DataLister {
+  def listData(spark: SparkSession, path: Path): DataFrame
+}
+
+class NaiveDataLister extends DataLister {
+  override def listData(spark: SparkSession, path: Path): DataFrame = {
+    import spark.implicits._
+    val fs = path.getFileSystem(spark.sparkContext.hadoopConfiguration)
+    val dataIt = fs.listFiles(path, false)
```
itaiad200, line 18: We need an exclusion list. For example `dummy` or anything under `_lakefs` should not be listed.

clients/spark/core/src/main/scala/io/treeverse/gc/DataLister.scala
```diff
@@ -0,0 +1,25 @@
...
+    val dataList = new ListBuffer[String]()
+    while (dataIt.hasNext) {
+      dataList += dataIt.next().getPath.toString
```
itaiad200, line 21: The creation time is very important too - it will be used to decide if an object is safe to delete.

clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
```diff
@@ -0,0 +1,30 @@
...
+    var normalizedStorageNamespace = storageNamespace
+    if (!normalizedStorageNamespace.endsWith("/")) {
+      normalizedStorageNamespace = "/"
+    }
+    val params =
+      LakeFSJobParams.forStorageNamespace(s"${normalizedStorageNamespace}", "uncommitted_gc")
+    var df = LakeFSContext.newDF(spark, params)
+    df = df
+      .select("address")
+      .withColumn("absolute_address", concat(lit(normalizedStorageNamespace), df("address")))
```
itaiad200, line 24: Is it always a naive `concat`? I think some addresses might be relative..

clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
```diff
@@ -0,0 +1,30 @@
...
+}
+
+class NaiveCommittedAddressLister extends CommittedAddressLister {
+  override def listCommittedAddresses(spark: SparkSession, storageNamespace: String): DataFrame = {
+    var normalizedStorageNamespace = storageNamespace
+    if (!normalizedStorageNamespace.endsWith("/")) {
+      normalizedStorageNamespace = "/"
+    }
+    val params =
+      LakeFSJobParams.forStorageNamespace(s"${normalizedStorageNamespace}", "uncommitted_gc")
```
itaiad200, line 20:
```suggestion
      LakeFSJobParams.forStorageNamespace(s"${normalizedStorageNamespace}", UNCOMMITTED_GC_SOURCE_NAME)
```

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -0,0 +1,56 @@
...
+    val dataLocation = getDataLocation(storageNamespace)
+    val uncommittedLocation = args(2) // TODO use actual uncommitted location
+    val outputLocation = args(1)
+
+    val dataPath = new Path(dataLocation)
+    val dataLister = new NaiveDataLister()
+
+    val dataDF = dataLister.listData(spark, dataPath)
+    val committedDF =
+      new NaiveCommittedAddressLister().listCommittedAddresses(spark, storageNamespace)
```
N-o-Z, line 50: Read of committed data must start after reading all uncommitted data (it should not be parallelized)

clients/spark/core/src/main/scala/io/treeverse/gc/DataLister.scala
johnnyaug: done

clients/spark/core/src/main/scala/io/treeverse/gc/DataLister.scala
johnnyaug: done

clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
johnnyaug: added a comment

clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
johnnyaug: done

PR #4596, opened 2022-11-14
LinkPhysicalAddress to support OutputCommitter

itaiad200:


arielshaqed (approved):
Obviously I really like this:

> Multiple entries on *different branches* may point to the same physical object, as long as they do that within the time frame.

Emphasis mine.  This is the ideal solution _for LakeFSOutputCommitter_... if you can support it :-)

itaiad200:
This failed miserably.
Me: "Why is timeframe not a strong enough guarantee? I assume that if an entity is committed on some branch, we will read it, thus avoid the deletion, no?"
@N-o-Z: "It's all about the timing, you can read it once it's committed on one branch and uncommitted on another, second time you read it it will be deleted on the second branch but we will not read the previous commit again (incremental)" (thus leading to it being deleted)

arielshaqed:
# 😿

But as discussed, completely understandable.

PR #4606, opened 2022-11-16
Python client map APIs use dynamically

### LINKED ISSUE #4604: Python LakeFSClient missing API tags
Swagger APIs separated by different tags - the import, retention and templates APIs are missing from the client.

nopcoder:
- Map each existing api to the name of the tag, assign it twice (with and without _api)
- If the attribute match python keyword skip the assignment (example 'import_api' will not be mapped as 'import' because it is a keyword)

Fix: added missing APIs - import, templates and retention
Break: the attribute 'health' will be mapped to 'healthcheck' and 'healthcheck_api'
- [ ] Document the above breaking change in the CHANGELOG

Fix #4604

### DIFF
--- a/clients/python/lakefs_client/client.py
+++ b/clients/python/lakefs_client/client.py
@@ -1,10 +1,11 @@
 from urllib3.util import parse_url, Url
 import sys
 import inspect
-import keyword
 
 import lakefs_client
 import lakefs_client.apis
+
+_API_CLASS_SUFFIX = "api"
 
 class _WrappedApiClient(lakefs_client.ApiClient):
     """ApiClient that fixes some weirdnesses."""
@@ -31,12 +32,11 @@ class LakeFSClient:
                                           header_value=header_value, cookie=cookie, pool_threads=pool_threads)
         for key, value in inspect.getmembers(sys.modules['lakefs_client.apis'], inspect.isclass):
             name = key.lower()
-            if not name.endswith('api'):
+            if not name.endswith(_API_CLASS_SUFFIX):
                 continue
             api_instance = value(self._api)
-            attr_name = name[0:len(name)-3]
-            if attr_name not in keyword.kwlist:
-                setattr(self, attr_name, api_instance)
+            attr_name = name[:-len(_API_CLASS_SUFFIX)]
+            setattr(self, attr_name, api_instance)
             setattr(self, attr_name + '_api', api_instance)
 
     @staticmethod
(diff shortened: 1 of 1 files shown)


clients/python/lakefs_client/client.py
```diff
@@ -27,18 +29,15 @@ def __init__(self, configuration=None, header_name=None, header_value=None, cook
...
-        self.refs = RefsApi(self._api)
-        self.repositories = RepositoriesApi(self._api)
-        self.staging = StagingApi(self._api)
-        self.tags = TagsApi(self._api)
+        for key, value in inspect.getmembers(sys.modules['lakefs_client.apis'], inspect.isclass):
+            name = key.lower()
+            if not name.endswith('api'):
+                continue
+            api_instance = value(self._api)
+            attr_name = name[0:len(name)-3]
```
arielshaqed, line 37: This is somewhat scary.  Can we instead use substitution?  Something like
```suggestion
            attr_name = re.sub('api$', '', name)
            if attr_name == name:
                continue
            api_instance = value(self._api)
```
might be better, as it makes it says "api" only once and avoids the magic number `3`.

clients/python/lakefs_client/client.py
```diff
@@ -27,18 +29,15 @@ def __init__(self, configuration=None, header_name=None, header_value=None, cook
...
+            if attr_name not in keyword.kwlist:
+                setattr(self, attr_name, api_instance)
```
arielshaqed, line 39: Why not
```suggestion
            setattr(self, attr_name, api_instance)
```
?

If it _is_ a keyword then you cannot write down the access... but it will still be an attribute and you can still use it with getattr or whatever.  I am not sure why we should prevent that.

arielshaqed (approved):
Nice, _many_ thanks!  I reckon this is great.

But because this is not even OpenAPI we will also need to document it.  Can you please open an issue to document how to get to APIs from the client?

Also...

>  Break: the attribute 'health' will be mapped to 'healthcheck' and 'healthcheck_api'

Can we manually add this back as an alias?  That's just nice-to-have, because obviously these are not very useful API calls -- but we cannot be sure that no-one is using them...

clients/python/lakefs_client/client.py
nopcoder: Simplified the code and used constant - I think it is better than having sub regex.
Only in Python 3.9 `removesuffix` was added :/

clients/python/lakefs_client/client.py
nopcoder: removed the keyword check - add them both.

nopcoder:
About the healthcheck is nice to have - but I don't see anyone using it in any way.

arielshaqed (commented):
> About the healthcheck is nice to have - but I don't see anyone using it in any way.

I doubt that we can know whether anyone is using an API.  We must _either_ document in the CHANGELOG that this changed, or manually add the "health" alias and preserve backwards compatibility.

nopcoder:
> I doubt that we can know whether anyone is using an API. We must _either_ document in the CHANGELOG that this changed, or manually add the "health" alias and preserve backwards compatibility.

Agree - we will document this breaking change in the CHANGELOG.

PR #4610, opened 2022-11-20
Parallelize listing of data in GC+ client code

### LINKED ISSUE #4599: Efficiently list entries from storage in GC+
Parallelize listing data objects from the storage in GC+ client code. Currently, we do it in the Spark driver, without any parallelization.

johnnyaug:
Resolves #4599.


N-o-Z (approved):
LGTM!

clients/spark/core/src/main/scala/io/treeverse/gc/DataLister.scala
```diff
@@ -25,3 +30,40 @@ class NaiveDataLister extends DataLister {
...
+      .map((status: FileStatus) =>
+        new FileDescriptor(status.getPath.getName, status.getModificationTime)
+      )
+      .toList
+  }
+
+  override def listData(configMapper: ConfigMapper, path: Path): DataFrame = {
+    import spark.implicits._
+    val fs = path.getFileSystem(spark.sparkContext.hadoopConfiguration)
+    val slices = listPath(configMapper, path.toString)
```
itaiad200, line 53: Shouldn't we exclude some paths?
1. Default `_lakefs`
2. `dummy`
3. Every other flat object that existed for previous repos.

johnnyaug: Filtering out is done outside the scope if this class.

PR #4622, opened 2022-11-21
Don't batch GetCommit calls during merges

arielshaqed:
During a merge, prevent batching GetCommit calls when there is a cache miss.

FindMergeBase performs one GetCommit call per commit traversed.  This is currently performed in series, so any batching immediate slows it down.


nopcoder (approved):
LGTM!

arielshaqed:
Thanks!

PR #4633, opened 2022-11-22
Fix uncommitted path for prepare gc

### LINKED ISSUE #4632: Uncommitted GC path contains `s3://` in the path
Root cause: Wrong concatenation of the full uri

itaiad200:
closes #4632

### DIFF
--- a/pkg/block/mem/adapter.go
+++ b/pkg/block/mem/adapter.go
@@ -80,10 +80,6 @@ func getKey(obj block.ObjectPointer) string {
 	return fmt.Sprintf("%s:%s", obj.StorageNamespace, obj.Identifier)
 }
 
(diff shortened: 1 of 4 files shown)


itaiad200:
Tested locally and the path looks ok now.

eden-ohana (approved):

N-o-Z (changes requested):
Please add the missing checks in the catalog unit test

N-o-Z (approved):
I understand we want to fix it fast so we won't block but I don't like this as a permanent solution.
Can we open an issue to handle this better later?

itaiad200:
Opened https://github.com/treeverse/lakeFS/issues/4649

PR #4639, opened 2022-11-23
Fix diff refs api to return prefix_changed

### LINKED ISSUE #4638: DiffRefs API returns unmapped diff type
- Diff refs API supports "prefix change" diff type that was not described in the OpenAPI spec.
- The value retuned in case of prefix change was "changes under prefix" while a snake case value that matches the reset of the api should be used.

nopcoder:
Fix #4638

### DIFF
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -343,7 +343,7 @@ components:
       properties:
         type:
           type: string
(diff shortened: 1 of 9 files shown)


itaiad200 (approved):
Thanks for the quick fix!

PR #4643, opened 2022-11-23
Fix unmatch-delete log line

### LINKED ISSUE #4642: Unmatch delete log tracing can panic when lakeFS is log level is TRACE

itaiad200:
closes #4642


N-o-Z (commented):
Is there no importance to the removed trace?
Do we want to just log the key?

itaiad200:
> Is there no importance to the removed trace? Do we want to just log the key?

Since we switched to kv, tombstones are written to the staging area regardless to a match committed entity. So by design we'll get unmatched delete -> this log line is useless.

N-o-Z (approved):
Thanks!

PR #4660, opened 2022-11-24
GC+: Prepare uncommitted ignore external addresses and tombstones

### LINKED ISSUE #4636: GC+: catalog panics when UncommittedIterator return tombstones
Catalog panics on https://github.com/treeverse/lakeFS/blob/e4c5a5db398677b813a771440688cb20108b9ef7/pkg/catalog/catalog.go#L1807
since it's trying to access properties of a nil Entry. UncommittedIterator should not list tombstones as they are in-fact not part of the uncommitted entries.

### LINKED ISSUE #4655: GC+: PrepareGCUncommitted to exclude external addresses
Modify code to ignore addresses outside of namespace

N-o-Z:
Closes #4636
Closed #4655

### DIFF
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -1768,6 +1768,12 @@ type UncommittedParquetObject struct {
 	CreationDate    int64  `parquet:"name=creation_date, type=INT64, convertedtype=INT_64"`
 }
 
+// isValidEntry Returns False if entry is tombstone or if address is outside of repository namespace
+func isValidEntry(entry *UncommittedRecord) bool {
+
+	return entry.Entry != nil && entry.Entry.AddressType == Entry_RELATIVE
+}
+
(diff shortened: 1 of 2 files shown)


pkg/catalog/catalog.go
```diff
@@ -1768,6 +1768,12 @@ type UncommittedParquetObject struct {
 	CreationDate    int64  `parquet:"name=creation_date, type=INT64, convertedtype=INT_64"`
 }
 
+// isValidEntry Returns False if entry is tombstone or if address is outside of repository namespace
+func isValidEntry(entry *UncommittedRecord) bool {
```
itaiad200, line 1772: That's a catalog-level function named `isValidEntry` which might be misused. Maybe
```suggestion
func shouldAddToUncommittedGC(entry *UncommittedRecord) bool {
```

itaiad200 (approved):
Thanks for adding tests! Approved with a single comment

pkg/catalog/catalog.go
N-o-Z: Removed function altogether

PR #4666, opened 2022-11-27
Use temporary workdir on job branch and copy out to commit task

arielshaqed:
Summary of flow:

* setupJob creates a job branch.

  Use the output path for job branch not the job ID on the context, as some
  Spark writers call setupJob on *one* job context and then run all tasks
  on *another* job context.
* setupTask does nothing!
* Tasks write to workDir `_temporary/<task_id>` on the job branch in the
  outputPath location.
* commitTask lists and copies files up from workDir to the intended
  location.
* commitJob deletes `_temporary/`, creates _SUCCESS if requested, commits
  and merges to output.

Tested manually with 4000 partitions.  It takes ~22 minutes to write 4000
(small!) partitions on a cloud (staging) deployment.

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/committer/LakeFSOutputCommitter.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/committer/LakeFSOutputCommitter.java
@@ -259,10 +259,10 @@ public class LakeFSOutputCommitter extends FileOutputCommitter {
 
             if (hasChanges(jobBranch)) {
                 CommitsApi commits = lakeFSClient.getCommits();
+                CommitCreation commitCreation = new CommitCreation()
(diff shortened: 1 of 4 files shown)


arielshaqed:
Note the added **unitest** is a unit test.  It tests _only_ some expected
behaviour of Bulk operations.  So Bulk will work correctly _if_ lakeFS
behaves in the expected way.  It is _not_ a component test and does _not_
verify that lakeFS behaves in the expected way.

talSofer (approved):
Very cool!

Added questions and suggestions.

arielshaqed (commented):
Thanks!

Resolving conflicts and pushing...

arielshaqed:
Thanks!

PR #4670, opened 2022-11-27
Document Uncommitted GC

### LINKED ISSUE #4612: Document GC+
Add to docs the "how-to" of running GC uncommitted including the known caveats of performance, new/old repos, etc.

itaiad200:
closes #4612

### DIFF
--- a/docs/howto/garbage-collection.md
+++ b/docs/howto/garbage-collection.md
@@ -330,7 +330,7 @@ This utility is a Spark application that uses [DistCp](https://hadoop.apache.org
 * There has been a change in the storage namespace structure relevant to lakeFS >= v0.84.0, which causes the [GC Backup & Restore job to break](https://github.com/treeverse/lakeFS/issues/4775).  Currently, we are working on a fix. 
 * GC Backup & Restore is available from version 0.5.2 of lakeFS Spark client.
 * It is compatible with Hadoop API versions 3.1.3 and higher.
-* Note that the utility is not fast due to DistCp performance limitations. You may prefer to backup your whole storage namespace with [AzCopy](https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-v10) / [aws cp](https://docs.aws.amazon.com/cli/latest/reference/s3/cp.html) / [rclone](https://rclone.org/).
+* Note that the utility is not fast due to DistCp performance limitations. You may prefer to backup your whole storage namespace with [AzCopy](https://learn.microsoft.com/en-us/azure/storage/common/storage-use-azcopy-v10) / [aws cp](https://docs.aws.amazon.com/cli/latest/reference/s3/cp.html) / [rclone](https://docs.lakefs.io/howto/copying.html#using-rclone).
 
 #### Job options
 {: .no_toc }
--- a/docs/howto/garbage-collection.md
+++ b/docs/howto/garbage-collection.md
@@ -266,10 +266,10 @@ repositories may require increasing a read timeout.  If you run into timeout err
    You should remove stale branches to prevent them from retaining old objects.
    For example, consider a branch that has been merged to `main` and has become stale.
    An object which is later deleted from `main` will always be present in the stale branch, preventing it from being hard-deleted.
-   
+
 1. lakeFS will never delete objects outside your repository's storage namespace.
    In particular, objects that were imported using `lakectl ingest` or `UI Import Wizard` will not be affected by GC jobs.
-   
+
 1. In cases where deleted objects are brought back to life while a GC job is running, said objects may or may not be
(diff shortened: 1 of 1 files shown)


docs/reference/garbage-collection.md
```diff
@@ -394,3 +393,72 @@ spark-submit --class io.treeverse.clients.GCBackupAndRestore \
   objects-to-restore-list-location backup-external-location storage-namespace azure
   ```
 </div>
+
+## Beta: Deleting uncommitted objects
+
+Deletion of objects that were never committed was always a difficulty for lakeFS, see
+[#1933](https://github.com/treeverse/lakeFS/issues/1933) for more details. Examples for 
+objects that will be collected as part of the uncommitted GC job:
```
johnnyaug, line 401:
```suggestion
Run this job to [delete objects that were never committed and have no valid pointers from lakeFS](https://github.com/treeverse/lakeFS/issues/1933). This may happen when:
```

docs/reference/garbage-collection.md
```diff
@@ -394,3 +393,72 @@ spark-submit --class io.treeverse.clients.GCBackupAndRestore \
...
+1. Objects that were uploaded to lakeFS and deleted.
+2. Objects that were uploaded to lakeFS and were overridden.
```
johnnyaug, line 403:
```suggestion
1. Objects are deleted or overwritten.
2. Changes are reverted from a branch before being committed.
3. A branch with uncommitted changes is deleted.
```

docs/reference/garbage-collection.md
```diff
@@ -394,3 +393,72 @@ spark-submit --class io.treeverse.clients.GCBackupAndRestore \
...
+
+The uncommitted GC will not clean:
+1. Committed objects. For committed objects cleanup see [above](#what-gets-collected)
```
johnnyaug, line 406:
```suggestion
1. Objects that are included in any commit. For committed objects cleanup see [above](#what-gets-collected)
```

docs/reference/garbage-collection.md
```diff
@@ -394,3 +393,71 @@ spark-submit --class io.treeverse.clients.GCBackupAndRestore \
...
+    <REPO_NAME> <TODO_OUTPUT_LOCATION>
+```
+
+#### Uncommitted GC job options
+
+Similar to the described [above](#gc-job-options).
+
+#### Backup and restore 
+
+<TODO> - will it work?
```
N-o-Z, line 428: It will work, but lets put a pin on that for the first milestone

docs/reference/garbage-collection.md
```diff
@@ -394,3 +393,71 @@ spark-submit --class io.treeverse.clients.GCBackupAndRestore \
...
+
+#### Limitations
+{: .no_toc }
+
+The uncommitted GC job has several limitations in its Beta version: 
+1. No writes to lakeFS during the execution of the job. Objects written to lakeFS
+during the job run may or may not be detected by the job. It's advised to avoid any
```
N-o-Z, line 435: I think we should be more strict than "advise". More on the lines of any non-read operation during execution of the beta GC process might result in unexpected behavior

docs/reference/garbage-collection.md
```diff
@@ -394,3 +393,71 @@ spark-submit --class io.treeverse.clients.GCBackupAndRestore \
...
+3. Scale may be limited, see performance results below.
+
+#### Next steps
+
+The uncommitted GC is under development, next releases will include:
+
+1. Incorporation of committed & uncommitted GC into a single job. We understand the friction
+of having 2 garbage collection jobs for a lakeFS installation and working to creating a
+single job for it.
+2. Support for non-S3 repositories.
```
N-o-Z, line 448: In my opinion this should be at the bottom of the list

N-o-Z (changes requested):
Looking good - I think we need to be a bit more cautionary in our language during the beta stage

docs/reference/garbage-collection.md
itaiad200: Just like mark&sweep - it's a safety mechanisms that users can rely on for safer delete. It can wait though.

docs/reference/garbage-collection.md
itaiad200: You got it.

docs/reference/garbage-collection.md
itaiad200: Done.

johnnyaug (approved):

docs/reference/garbage-collection.md
N-o-Z: So let's change the comment to 'be supported in the future'

N-o-Z (approved):
Approved - see comment

PR #4675, opened 2022-11-27
Add a configurable cutoff time for gc+

### LINKED ISSUE #4659: Configurable grace duration for deleted entities

itaiad200:
closes #4659

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSContext.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSContext.scala
@@ -67,8 +67,9 @@ object LakeFSContext {
   val LAKEFS_CONF_DEBUG_GC_MAX_COMMIT_EPOCH_SECONDS_KEY = "lakefs.debug.gc.max_commit_epoch_seconds"
   val LAKEFS_CONF_DEBUG_GC_REPRODUCE_RUN_ID_KEY = "lakefs.debug.gc.reproduce_run_id"
 
-  //  Objects that are written during the skip period are not collected
-  val LAKEFS_CONF_DEBUG_GC_UNCOMMITTED_SKIP_SECONDS_KEY = "lakefs.debug.gc.uncommitted_skip_seconds"
+  //  Objects that are written during this duration are not collected
+  val LAKEFS_CONF_DEBUG_GC_UNCOMMITTED_MIN_AGE_SECONDS_KEY =
+    "lakefs.debug.gc.uncommitted_min_age_seconds"
   val LAKEFS_CONF_GC_DO_MARK = "lakefs.gc.do_mark"
   val LAKEFS_CONF_GC_DO_SWEEP = "lakefs.gc.do_sweep"
   val LAKEFS_CONF_GC_MARK_ID = "lakefs.gc.mark_id"
@@ -79,6 +80,9 @@ object LakeFSContext {
   val COMMITS_LOCATION_KEY = "commits_location"
   val DEFAULT_LAKEFS_CONF_GC_NUM_RANGE_PARTITIONS = 50
   val DEFAULT_LAKEFS_CONF_GC_NUM_ADDRESS_PARTITIONS = 200
+
+  // By default, objects that are written in the last 6 hours are not collected
+  val DEFAULT_GC_UNCOMMITTED_MIN_AGE_SECONDS = 6 * 60 * 60
 
   def newRDD(
       sc: SparkContext,
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSContext.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSContext.scala
@@ -66,6 +66,9 @@ object LakeFSContext {
   val LAKEFS_CONF_DEBUG_GC_MAX_COMMIT_ISO_DATETIME_KEY = "lakefs.debug.gc.max_commit_iso_datetime"
   val LAKEFS_CONF_DEBUG_GC_MAX_COMMIT_EPOCH_SECONDS_KEY = "lakefs.debug.gc.max_commit_epoch_seconds"
   val LAKEFS_CONF_DEBUG_GC_REPRODUCE_RUN_ID_KEY = "lakefs.debug.gc.reproduce_run_id"
+
+  //  Objects that are written during the skip period are not collected
+  val LAKEFS_CONF_DEBUG_GC_UNCOMMITTED_SKIP_SECONDS_KEY = "lakefs.debug.gc.uncommitted_skip_seconds"
   val LAKEFS_CONF_GC_DO_MARK = "lakefs.gc.do_mark"
   val LAKEFS_CONF_GC_DO_SWEEP = "lakefs.gc.do_sweep"
   val LAKEFS_CONF_GC_MARK_ID = "lakefs.gc.mark_id"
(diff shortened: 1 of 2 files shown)


clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSContext.scala
```diff
@@ -66,6 +66,9 @@ object LakeFSContext {
   val LAKEFS_CONF_DEBUG_GC_MAX_COMMIT_ISO_DATETIME_KEY = "lakefs.debug.gc.max_commit_iso_datetime"
   val LAKEFS_CONF_DEBUG_GC_MAX_COMMIT_EPOCH_SECONDS_KEY = "lakefs.debug.gc.max_commit_epoch_seconds"
   val LAKEFS_CONF_DEBUG_GC_REPRODUCE_RUN_ID_KEY = "lakefs.debug.gc.reproduce_run_id"
+
+  //  Objects that are written during the skip period are not collected
+  val LAKEFS_CONF_DEBUG_GC_UNCOMMITTED_SKIP_SECONDS_KEY = "lakefs.debug.gc.uncommitted_skip_seconds"
```
N-o-Z, line 71: Did you choose this resolution for the tests sake?

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -48,7 +48,16 @@ object UncommittedGarbageCollector {
     val secretKey = hc.get(LAKEFS_CONF_API_SECRET_KEY_KEY)
     val connectionTimeout = hc.get(LAKEFS_CONF_API_CONNECTION_TIMEOUT_SEC_KEY)
     val readTimeout = hc.get(LAKEFS_CONF_API_READ_TIMEOUT_SEC_KEY)
+    val skipSecondsStr = hc.get(LAKEFS_CONF_DEBUG_GC_UNCOMMITTED_SKIP_SECONDS_KEY)
```
N-o-Z, line 51: I think we should have a minimal duration requirement - setting it too low may result in a serious bug.
It should be aligned somehow with the link physical address expiry time

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -74,9 +83,12 @@ object UncommittedGarbageCollector {
           // in case of no uncommitted entries
           spark.emptyDataFrame.withColumn("physical_address", lit(""))
         }
+
+      // TODO: should filter by time too
```
N-o-Z, line 87: You don't really need to filter committed and uncommitted in the client side. It's enough to remove it from the dataDF to ensure these won't be deleted

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -102,6 +114,9 @@ object UncommittedGarbageCollector {
     }
   }
 
+  // By default, objects that are written in the last 6 hours are not collected
+  private val DEFAULT_GC_UNCOMMITTED_SKIP_SECONDS = 6 * 60 * 60
```
N-o-Z, line 118: You can add it also to the LakeFSContext

N-o-Z (changes requested):
Thank you for this PR, some comments

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -64,7 +73,7 @@ object UncommittedGarbageCollector {
     }
 
     try {
-      val dataDF = listObjects(storageNamespace, DateUtils.addHours(new Date(), -6))
+      val dataDF = listObjects(storageNamespace, cuttoffTime)
```
johnnyaug, line 76:
```suggestion
      val dataDF = listObjects(storageNamespace, cutoffTime)
```

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSContext.scala
johnnyaug, line 71: is this key only for debug? Otherwise drop the "debug" word.
Also, how about "gc.uncommitted_min_age_seconds" as the name?

johnnyaug (approved):
LGTM - minor comments

clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSContext.scala
itaiad200: Yap, do you have anything else in mind?

itaiad200: I believe this should be used only for debugging. The default value should be tied to some server-internal configuration, like the validity duration of the token used for Get/LinkPhysicalAddress. Making this anything other than debug/plumbing is allowing an unaware user to potentially delete unlinked data.

Regarding the name - sounds better, changing.

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
itaiad200: The default value should be aligned, I agree. Do you see any value (in terms of "safety") in a minimum value that is bigger than a minute? We need to add system-tests and waiting more than that doesn't sound reasonable.

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
itaiad200: Done.

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
itaiad200: Yesterday I had a strong hunch that having the 3 DFs (physicals, committed, uncommitted) looking at different time frames may lead to undeleted objects in the optimized run. I can't come up with the example now. 
We must consider the grace period when running the uncommitted GC, i.e. next optimized run's time frame needs to start from the previous run beginning of the grace period. Does it make sense?

Removing these comments for now.

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
itaiad200: Done.

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
itaiad200: I don't think it's wise to let the users define the expiry for the tokens for non-testing purposes. For example, setting it to a low number in the server may lead to blocking all writes due to invalidity of all tokens.

Also, if the user is messing with arguments that are labeled as `debug` in a task that is deleting data, he can expect undefined behaviour from it.

N-o-Z: I didn't mean the users will be able to modify it in the server. My idea was to provide this parameter from the server. This way we'll be able to control it via the tests and validate it against the expiry token

itaiad200: How would you change it from the server in a way that you can and the users can't?

N-o-Z: In unit tests that should not be an issue. For system tests we could add an undocumented configuration parameter, but once it's in the server, we can always validate it against the token timeout and handle conflicting values. Leaving it in the client side it means we will have no way to identify this problem

itaiad200: Opened #4695 to tackle this.

N-o-Z (approved):
We need to somehow validate this parameter against the server to see that we are not using an invalid configuration

PR #4685, opened 2022-11-28
GC+: Fix list objects

### LINKED ISSUE #4683: GC+: Fix list objects

N-o-Z:
closes #4683

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
@@ -52,11 +52,11 @@ object UncommittedGarbageCollector {
     val oldDataPath = new Path(storageNamespace)
     // TODO (niro): implement parallel lister for old repositories (https://github.com/treeverse/lakeFS/issues/4620)
     val oldDataDF = new NaiveDataLister().listData(configMapper, oldDataPath)
-    dataDF.union(oldDataDF)
+    dataDF =
+      dataDF.union(oldDataDF).filter(dataDF("last_modified") < before.getTime).select("address")
 
-    dataDF = dataDF.filter(dataDF("last_modified") < before.getTime).select("address")
     // TODO (niro): Check if really needed and consider passing it as part of uncommittedGCRunInfo
-    dataDF = dataDF.filter(dataDF("address").startsWith("_lakefs"))
+    dataDF = dataDF.filter(!dataDF("address").startsWith("_lakefs"))
     dataDF
   }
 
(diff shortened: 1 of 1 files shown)


clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -52,11 +52,11 @@ object UncommittedGarbageCollector {
...
     // TODO (niro): implement parallel lister for old repositories (https://github.com/treeverse/lakeFS/issues/4620)
     val oldDataDF = new NaiveDataLister().listData(configMapper, oldDataPath)
-    dataDF.union(oldDataDF)
+    dataDF =
+      dataDF.union(oldDataDF).filter(dataDF("last_modified") < before.getTime).select("address")
 
-    dataDF = dataDF.filter(dataDF("last_modified") < before.getTime).select("address")
     // TODO (niro): Check if really needed and consider passing it as part of uncommittedGCRunInfo
-    dataDF = dataDF.filter(dataDF("address").startsWith("_lakefs"))
+    dataDF = dataDF.filter(!dataDF("address").startsWith("_lakefs"))
```
itaiad200, line 59: I might be misunderstanding something..
How is `_lakefs\` being listed at all? I thought the `oldDataDF` will only list objects which are directly under the storage namespace. If it lists `_lakefs\` why doesn't it list `data\` too?

N-o-Z: That's why I added the comment - anyway this is not the only issue here

itaiad200: We possibly list everything twice, I think we should address it now (can be in a separate PR though)

N-o-Z: We don't list everything twice - this is already tested

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -49,14 +50,12 @@ object UncommittedGarbageCollector {
       .select("address", "last_modified")
 
     // Read objects from namespace root, for old structured repositories
-    val oldDataPath = new Path(storageNamespace)
+
     // TODO (niro): implement parallel lister for old repositories (https://github.com/treeverse/lakeFS/issues/4620)
     val oldDataDF = new NaiveDataLister().listData(configMapper, oldDataPath)
-    dataDF.union(oldDataDF)
+    dataDF =
+      dataDF.union(oldDataDF).filter(dataDF("last_modified") < before.getTime).select("address")
```
itaiad200, line 57: Will the filtering work correctly? 
The `oldDataDF` isn't part of `dataDF` when the args are captured. I think the `oldDataDF` won't get filtered.

itaiad200 (changes requested):

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
N-o-Z: This is just to provide the column name - modified to use `col` instead

itaiad200 (approved):
Thanks!

PR #4687, opened 2022-11-28
Make "docker run" the main Quickstart option

### LINKED ISSUE #4688: Docs: make lakeFS Playground the main Quickstart option
And move Everything Bagel to the other methods.

johnnyaug:
And moved Everything Bagel to "more quickstart options".
Also, add Playground to "more quickstart options".

Resolves #4688


ozkatz:
I think we should actually stick with the current assumption (which I feel that has been somewhat validated at this point) around our user journey:
local docker run -> local docker run with cloud storage -> cloud installation.

I'd argue that the playground belongs in the "more quickstart options" as well, and in other assets that are meant for lower intent users (lakefs.io landning page, perhaps the index page and README...

PR #4693, opened 2022-11-29
Update "what is lakeFS" with "git for data"

michalwosk:
Added a reference to the term "git for data" in order to improve SEO on this term.
If this sounds too much like "marketing" feel free to change it a bit, but please keep the title as well as the mentions of the term "git for data".


adipolak (approved):
LGTM

Issue #4699, opened 2022-11-29
Support relative path in metadata client


N-o-Z:
Related to [#4698](https://github.com/treeverse/lakeFS/issues/4698)
Uncommitted GC reads object store and uncommitted data in relative path format. 
Committed data is returned as full path

Suggested to add functionality to request addresses in relative path format from metadata client.

N-o-Z:
@johnnyaug FYI

N-o-Z:
Not an issue

PR #4700, opened 2022-11-29
GC+: CommittedDataLister transform address column to relative path

### LINKED ISSUE #4698: Committed entries are marked for deletion
With the latest master I get:
- Committed entries:
`s3a://storagenamespace/data/gr7t8hlfcls7ethcf6k0/cdviuglfcls7ethcf6ng`
- Physical entries:
`data/gr7t8hlfcls7ethcf6k0/cdviuglfcls7ethcf6ng`
- uncommitted entries:
`data/gr57ka5fclsdflgskrhg/ce28r55fclsdflgskrj0`

So committed entries are not excluded and can be picked up by the GC.

N-o-Z:
Closes #4698 

Intermediate solution until https://github.com/treeverse/lakeFS/issues/4699 is implemented

Tested manually:

```
+---------------------------------------------------------------------+
|address                                                              |
+---------------------------------------------------------------------+
|s3a://storagenamespace/data/gr7t8hlfcls7ethcf6k0/cdviuglfcls7ethcf6ng|
|s3a://storagenamespace/data/gr7t8hlfcls7ethcf6k0/sdkfjlfkvlkvcbcz    |
+---------------------------------------------------------------------+

+----------------------------------------------+
|address                                       |
+----------------------------------------------+
|data/gr7t8hlfcls7ethcf6k0/cdviuglfcls7ethcf6ng|
|data/gr7t8hlfcls7ethcf6k0/sdkfjlfkvlkvcbcz    |
+----------------------------------------------+
```

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
@@ -27,7 +27,9 @@ class NaiveCommittedAddressLister extends CommittedAddressLister {
     // TODO push down a filter to the input format, to filter out absolute addresses!
     df = df
       // TODO (niro): Revert substring after https://github.com/treeverse/lakeFS/issues/4699
-      .select(df("absolute_address"), substring_index(col("absolute_address"), "/", -3) .as("address"))
+      .select(df("absolute_address"),
+              substring_index(col("absolute_address"), "/", -3).as("address")
+             )
       .distinct
     df
   }
--- a/clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
@@ -28,7 +28,7 @@ class NaiveCommittedAddressLister extends CommittedAddressLister {
     df = df
       // TODO (niro): Revert substring after https://github.com/treeverse/lakeFS/issues/4699
       .select(df("absolute_address"),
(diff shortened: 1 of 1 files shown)


clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
```diff
@@ -26,7 +26,10 @@ class NaiveCommittedAddressLister extends CommittedAddressLister {
       .withColumn("absolute_address", concat(lit(normalizedStorageNamespace), df("address")))
     // TODO push down a filter to the input format, to filter out absolute addresses!
     df = df
-      .select(df("absolute_address").as("address"))
+      // TODO (niro): Revert substring after https://github.com/treeverse/lakeFS/issues/4699
+      .select(df("absolute_address"),
+              substring_index(col("absolute_address"), "/", -3).as("address")
```
itaiad200, line 31: Why hardcoding `-3`? Storage namespace can be 
1. `s3://bucket/`
2. `s3://bucket/path/`
3. `s3://bucket/some/very/long/path/`
 
And I think we handle it, but we need to take into account storage namespace that does or doesn't end with a slash.

itaiad200 (changes requested):
Thanks - See single comment.

clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
itaiad200: We have the storage namespace, let's use it :)

itaiad200 (approved):
Thanks!

PR #4703, opened 2022-11-29
GC uncommitted skip dummy object from old data location

### LINKED ISSUE #4654: GC+: Exclude special paths from garbage collection
Exclude:

Data Listing:
-  `_lakefs` prefix

Committed DF
-  External addresses

nopcoder:
Close #4654

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
@@ -1,21 +1,17 @@
 package io.treeverse.gc
 
-import io.treeverse.clients.APIConfigurations
-import io.treeverse.clients.ApiClient
 import io.treeverse.clients.LakeFSContext._
-import io.treeverse.clients.StorageClientType
+import io.treeverse.clients._
+import org.apache.commons.lang3.time.DateUtils
 import org.apache.hadoop.fs.Path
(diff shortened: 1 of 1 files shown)


clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -92,39 +102,45 @@ object UncommittedGarbageCollector {
...
+        .listCommittedAddresses(spark, storageNamespace)
 
       addressesToDelete = dataDF
         .except(committedDF)
         .except(uncommittedDF)
+
       // TODO (niro): not working - need to find the most efficient way to save the first slice
-      val firstFile = dataDF.select(col("address")).first().toString()
-      firstSlice = firstFile.substring(0, firstFile.lastIndexOf("/"))
+      if (!dataDF.isEmpty) {
```
N-o-Z, line 125: This will still not work since dataDF contains both "old" and "new" address so basically the first row could consist of an "old" address which will not give us the first partition. We might want to return the slice from the listObjects method

nopcoder: I've just wrapped the code to prevent exception during extract from empty data-frame. No change to the logic inside.

N-o-Z (approved):
👍🏽 💪🏽

PR #4704, opened 2022-11-30
UGC DataLister improve list files by iterator

nopcoder:


clients/spark/core/src/main/scala/io/treeverse/gc/DataLister.scala
```diff
@@ -32,33 +31,39 @@ class NaiveDataLister extends DataLister {
...
   }
 
   override def listData(configMapper: ConfigMapper, path: Path): DataFrame = {
     import spark.implicits._
+
     val slices = listPath(configMapper, path)
-    val pathStr = path.toString
     val objectsUDF = udf((sliceId: String) => {
-      val slicePath = new Path(pathStr, sliceId)
+      val slicePath = new Path(path, sliceId)
```
N-o-Z, line 59: This fails in older versions sine Path is not serializable

N-o-Z (changes requested):
Reviewed only the DataLister code. 
Looks great - I wonder how much faster this improvement is compared to the previous code.
Add one comment

N-o-Z (approved):
Thanks!

johnnyaug (approved):

PR #4709, opened 2022-11-30
Fix committed addresses listing

### LINKED ISSUE #4708: GC+: Committed address has both absolute and relative columns

itaiad200:
Fix #4708


itaiad200:
Dup

PR #4715, opened 2022-11-30
GC+: Ignore external addresses in CommittedAddressLister

### LINKED ISSUE #4641: GC+: Committed and Uncommitted listing do not skip external addresses
Uncommitted GC will only cleanup objects in the storage namespace. Committed & uncommitted data frames should not include external (imported) addresses.

N-o-Z:
Closes #4641

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
@@ -24,8 +24,12 @@ class NaiveCommittedAddressLister extends CommittedAddressLister {
     var df = LakeFSContext.newDF(spark, params)
     df = df
       // TODO (optional): push down a filter to the input format, to filter out absolute addresses!
-      // filtering on != FULL on purpose - required for backwards compatability with entries prior to address_type
-      .filter(col("address_type") =!= AddressType.FULL.name)
+      .filter(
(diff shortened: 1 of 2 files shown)


johnnyaug (approved):
Awesome!

clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
```diff
@@ -22,14 +23,14 @@ class NaiveCommittedAddressLister extends CommittedAddressLister {
                                          )
     var df = LakeFSContext.newDF(spark, params)
     df = df
+      // TODO (optional): push down a filter to the input format, to filter out absolute addresses!
+      .filter(
+        (col("address_type") === AddressType.RELATIVE.name) ||
+          // Backwards compatability with entries prior to address_type
+          (col("address_type") === AddressType.BY_PREFIX_DEPRECATED.name &&
+            col("address_type").contains("://"))
```
itaiad200, line 31:
```suggestion
            !col("address_type").contains("://"))
```

itaiad200, line 31: Need to check that if you split by `://` you get at least 2 parts, and that the first one doesn't contain `/` or `:`.

itaiad200 (changes requested):

itaiad200 (approved):

PR #4717, opened 2022-11-30
4490/GC+ mark & sweep

### LINKED ISSUE #4490: GC+: Mark and Sweep
Enable mark and sweep functionality for uncommitted GC

eden-ohana:
closes #4490


clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -86,6 +90,16 @@ object UncommittedGarbageCollector {
     val cutoffTime = DateUtils.addSeconds(new Date(), -minAgeSeconds)
     val startTime = java.time.Clock.systemUTC.instant()
 
+    val shouldMark = hc.getBoolean(LAKEFS_CONF_GC_DO_MARK, true)
+    val shouldSweep = hc.getBoolean(LAKEFS_CONF_GC_DO_SWEEP, true)
```
itaiad200, line 94: Can the default be false? At least for the beta version..

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -86,6 +90,16 @@ object UncommittedGarbageCollector {
...
+    val markID = hc.get(LAKEFS_CONF_GC_MARK_ID, "")
+
+    GarbageCollector.validateRunModeConfigs(hc.getBoolean(LAKEFS_CONF_DEBUG_GC_NO_DELETE_KEY, false),
+      shouldMark,
+      shouldSweep,
+      hc.get(LAKEFS_CONF_GC_MARK_ID, "")
```
itaiad200, line 100: You got the markID before, didn't you?

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -86,6 +90,16 @@ object UncommittedGarbageCollector {
     val cutoffTime = DateUtils.addSeconds(new Date(), -minAgeSeconds)
     val startTime = java.time.Clock.systemUTC.instant()
 
+    val shouldMark = hc.getBoolean(LAKEFS_CONF_GC_DO_MARK, true)
+    val shouldSweep = hc.getBoolean(LAKEFS_CONF_GC_DO_SWEEP, true)
+    val markID = hc.get(LAKEFS_CONF_GC_MARK_ID, "")
```
itaiad200, line 95: Is the markID the previous runID?

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -157,14 +193,27 @@ object UncommittedGarbageCollector {
...
+    val reportDst = reportPath(storageNamespace, runID)
     writeJsonSummary(reportDst, runID, firstSlice, startTime, success, removed.count())
 
+    expiredAddresses.write.parquet(s"$reportDst/marked")
     removed.write.parquet(s"$reportDst/deleted")
   }
 
+  private def reportPath(storageNamespace: String, runID: String): String = {
+    val report = s"${storageNamespace}_lakefs/retention/gc/uncommitted/$runID"
+    report
```
itaiad200, line 208: Do I know Scala and this may work?
```suggestion
     s"${storageNamespace}_lakefs/retention/gc/uncommitted/$runID"
```

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -157,14 +193,27 @@ object UncommittedGarbageCollector {
       firstSlice: String,
       startTime: java.time.Instant,
       success: Boolean,
+      expiredAddresses: DataFrame,
       removed: DataFrame
```
itaiad200, line 197: What's the difference between these 2? Aren't they mutually exclusive?

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -95,42 +109,63 @@ object UncommittedGarbageCollector {
...
+            addressesToDelete = readMarkedAddresses(storageNamespace, markID)
+          }
+
+          val region = if (args.length == 2) args(1) else null
+          var storageNSForSdkClient = apiClient.getStorageNamespace(repo, StorageClientType.SDKClient)
+          if (!storageNSForSdkClient.endsWith("/")) {
+            storageNSForSdkClient += "/"
+          }
+          val hcValues = spark.sparkContext.broadcast(HadoopUtils.getHadoopConfigurationValues(hc, "fs.", "lakefs."))
+          val configMapper = new ConfigMapper(hcValues)
```
itaiad200, line 164: Is this logic similar to committed GC? can we extract it to somewhere common?

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -95,42 +109,63 @@ object UncommittedGarbageCollector {
...
+
+        // TODO (niro): not working - need to find the most efficient way to save the first slice
+        if (!dataDF.isEmpty) {
+          val firstFile = dataDF.select(col("address")).first().toString()
+          firstSlice = firstFile.substring(0, firstFile.lastIndexOf("/"))
+        }
+      }
+      removed = {
+        if (shouldSweep) {
+          if (markID != "") { // get the expired addresses from the mark id run
```
itaiad200, line 154: Can we log something here about the beginning of deletion of data?

itaiad200 (changes requested):
Thanks - seems like you nailed the last piece of the puzzle! See my comments..
How was this tested?

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
eden-ohana: done

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
eden-ohana: done

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
eden-ohana: It doesn't have to be the precious, but yes

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
eden-ohana: done

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
eden-ohana: done

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
eden-ohana: The expired address is the addresses we marked, and the removed ones are the addresses we sweep. They can be identical if no other mark id was given, or empty if we did only one operation.

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
eden-ohana: Added print, I see that we don't use logs

eden-ohana:
> Thanks - seems like you nailed the

It was tested locally with spark-submit

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -157,14 +197,26 @@ object UncommittedGarbageCollector {
       firstSlice: String,
       startTime: java.time.Instant,
       success: Boolean,
+      expiredAddresses: DataFrame,
```
N-o-Z, line 200: I think this is redundant - this is in fact the same DF as removed, there's no difference except the naming regardless of the GC operation we are doing.

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -157,14 +197,26 @@ object UncommittedGarbageCollector {
...
+    expiredAddresses.write.parquet(s"$reportDst/marked")
     removed.write.parquet(s"$reportDst/deleted")
   }
 
+  private def reportPath(storageNamespace: String, runID: String): String = {
+    s"${storageNamespace}_lakefs/retention/gc/uncommitted/$runID"
+  }
+
+  private def readMarkedAddresses(storageNamespace: String, markID: String): DataFrame = {
+    // TODO(eden) - check if this run succeeded
```
N-o-Z, line 215: Why is this a TODO? We already have everything we need to check it

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -157,14 +197,26 @@ object UncommittedGarbageCollector {
...
+    val path = reportPath(storageNamespace, markID)
+    spark.read.parquet(s"$path/marked")
+  }
+
   private def writeJsonSummary(
```
N-o-Z, line 220: How can we tell which run type was executed in this job? How can we verify it later?

N-o-Z (changes requested):
Thank you for adding this code - I find a lot of hidden assumptions that are not defined and we didn't discuss and should be well documented either in the code or as part of the design.

1. What are there rules of usage for mark and sweep, what assumptions do we make? Is there a specific order or sequence that is supported?
2. How do we know if the provided run ID for the sweep operation is the last run that was made by the GC, how do we know its a was a mark operation? how is GC supposed to behave when it's not the last job that ran?
3. How do we know what type of run was executed by the given runID? We could potentially be doing a lot of work for nothing

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -204,8 +208,7 @@ object UncommittedGarbageCollector {
   }
 
   private def reportPath(storageNamespace: String, runID: String): String = {
-    val report = s"${storageNamespace}_lakefs/retention/gc/uncommitted/$runID"
-    report
+    s"${storageNamespace}_lakefs/retention/gc/uncommitted/$runID"
```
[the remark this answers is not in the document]
N-o-Z: The {} is redundant in scala - suggest to remove it altogether

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -90,14 +90,15 @@ object UncommittedGarbageCollector {
     val cutoffTime = DateUtils.addSeconds(new Date(), -minAgeSeconds)
     val startTime = java.time.Clock.systemUTC.instant()
 
-    val shouldMark = hc.getBoolean(LAKEFS_CONF_GC_DO_MARK, true)
-    val shouldSweep = hc.getBoolean(LAKEFS_CONF_GC_DO_SWEEP, true)
+    val shouldMark = hc.getBoolean(LAKEFS_CONF_GC_DO_MARK, false)
```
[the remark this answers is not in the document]
N-o-Z: @itaiad200  initially I thought so too - but then I figured that if someone wants to perform only sweep and forgets to set mark to false it will result in an unwanted behavior

itaiad200:
> > Thanks - seems like you nailed the
> 
> It was tested locally with spark-submit

Cool! I can think of 4 scenarios that are worth verifying, but maybe there are more:
1. Just Mark.
2. Just Sweep.
3. Mark&Sweep.
4. No Mark & No Sweep.

Did we check all 4?

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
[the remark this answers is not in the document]
eden-ohana: done

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -154,15 +155,18 @@ object UncommittedGarbageCollector {
           if (markID != "") { // get the expired addresses from the mark id run
             addressesToDelete = readMarkedAddresses(storageNamespace, markID)
           }
+          println("deleting marked addresses: " + markID)
```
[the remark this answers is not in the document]
eden-ohana: opened https://github.com/treeverse/lakeFS/issues/4739

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
eden-ohana: done

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
eden-ohana: for now, I'm writing the JSON summary only for mark runs as we said we don't need the deletion info

itaiad200 (approved):
Thanks!

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -157,12 +200,26 @@ object UncommittedGarbageCollector {
...
+    expiredAddresses.show()
+    expiredAddresses.write.parquet(s"${reportDst}/deleted")
+  }
+
+  private def reportPath(storageNamespace: String, runID: String): String = {
+    s"${storageNamespace}_lakefs/retention/gc/uncommitted/${runID}"
+  }
 
-    removed.write.parquet(s"$reportDst/deleted")
+  private def readMarkedAddresses(storageNamespace: String, markID: String): DataFrame = {
```
N-o-Z, line 215: Please open an issue to add a unit test for this method

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -157,12 +200,26 @@ object UncommittedGarbageCollector {
...
+    val path = reportPath(storageNamespace, markID)
+    val markedRunSummary = spark.read.json(s"${path}/summary.json")
+    if (markedRunSummary.select("success").first() == false) {
```
N-o-Z, line 218: We should throw an exception here. We ignore the failed run status and the user assumes his sweep run completed successfully

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -153,9 +154,12 @@ object UncommittedGarbageCollector {
       removed = {
         if (shouldSweep) {
           if (markID != "") { // get the expired addresses from the mark id run
-            addressesToDelete = readMarkedAddresses(storageNamespace, markID)
+            markedAddresses = readMarkedAddresses(storageNamespace, markID)
```
N-o-Z, line 157: There's a bug here, if we run mark&sweep and pass a markID we will not deleted the items we marked.
We should assert that there's no MarkID if mark == true

N-o-Z: Also it's better to change the conditions to: if mark==true

N-o-Z (changes requested):
Thanks, please see comments

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
eden-ohana: Hi, it’s not a bug, if we pass a mark id it will delete it even if we also mark in this run
It's the same as the GC behavior

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
eden-ohana: It is caught in the try-catch of main

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -157,12 +200,26 @@ object UncommittedGarbageCollector {
...
       startTime: java.time.Instant,
       success: Boolean,
-      removed: DataFrame
+      expiredAddresses: DataFrame
   ): Unit = {
-    val reportDst = s"${storageNamespace}_lakefs/retention/gc/uncommitted/$runID"
-    writeJsonSummary(reportDst, runID, firstSlice, startTime, success, removed.count())
+    val reportDst = reportPath(storageNamespace, runID)
+    writeJsonSummary(reportDst, runID, firstSlice, startTime, success, expiredAddresses.count())
+    expiredAddresses.show()
```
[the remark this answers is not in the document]
eden-ohana: done

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
eden-ohana: done

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
N-o-Z: We are handling the addresses to delete differently from how the committed GC is working. I don't see any valid use case for passing mark ID with mark == true and I feel it might cause more confusion for us and the users. In what kind of scenario do you see the user pass both markID and performs mark and sweep?

eden-ohana: We discussed it and decided to block the option to the use to pass mark id when performing mark

N-o-Z (approved):
Great! Thanks for addressing the issues

PR #4719, opened 2022-11-30
Perform fewer API calls for exists

arielshaqed:
Make the "does not exist" path faster: with this change it typically takes 1 listObjects() API call instead of 1 statObject and 2 listObjects.  This is surprisingly *common*.  The trivial "exists" path now requires 1 listObject instead of 1 statObject call, which is more expensive.


arielshaqed:
LakeFSOutputCommitter project abandoned as we were unable to overcome these issues on current lakeFS.  Closing in favour of a new PR to put this on trunk.

PR #4728, opened 2022-12-02
Fixing the misconception that gateway is not scalable

### LINKED ISSUE #4727: Docs create a misconception that s3 gateway is not scalable
The docs describing the two tiered approach to spark make it seem like the client is better than the s3a gateway approach because of performance.

iddoavn:
Fix #4727 

The docs describing the two tiered approach to spark make it seem like the client is better than the s3a gateway approach because of performance.


adipolak (approved):
thanks for this - looks much better now!

PR #4733, opened 2022-12-04
Fix: LakeFSAllRangesInputFormat - Throw exception if namespace doesn't exist

N-o-Z:


clients/spark/core/src/main/scala/io/treeverse/clients/LakeFSInputFormat.scala
```diff
@@ -209,24 +209,17 @@ class LakeFSAllRangesInputFormat extends LakeFSBaseInputFormat {
...
-      null
-    )
-    val fs = FileSystem.get(lakeFSMetadataURI, conf)
-    val splits = new ListBuffer[InputSplit]()
-    val metadataPath = new Path(lakeFSMetadataURI.toString)
+    val namespaceURI = URI.create(storageNamespace)
+    val fs = FileSystem.get(namespaceURI, conf)
+    fs.getStatus(new Path(namespaceURI)) // Will throw an exception if namespace doesn't exist
+
+    val metadataURI = namespaceURI.resolve("_lakefs")
```
johnnyaug, line 216: I love that the code is much cleaner here!
IIRC, resolve will only get the desired result here if the storage namespace ends with a "/" here, so better double-check this. If it's the case, consider adding it if needed before creating the URI.

johnnyaug (approved):
Thanks for fixing this! And sorry for the delay in reviewing. See one comment.

PR #4743, opened 2022-12-06
Fix commit prefix is prefered over branch/tag

### LINKED ISSUE #4738: Commit ref takes precedence over branches for user-defined ref
I have a commit with ID `767cc4fb78e7424886245dff9217b66c` and a branch named `767`. Whenever I try to read objects using `767`, it will return objects from the commit and not the branch.
We recently changed the order of the ref lookup to look for the commit first; the reasoning behind it was that most internal ref lookups use commits and not branches. But for user given refs, it's more common to look for branches first. The ideal ordering of refParse to me would be:
1. `byCommit` 
2. `byBranch`
3. `byCommitPrefix`

itaiad200:
closes #4738

### DIFF
--- a/pkg/graveler/ref/resolve_ref.go
+++ b/pkg/graveler/ref/resolve_ref.go
@@ -25,7 +25,7 @@ func isAHash(part string) bool {
 }
 
 func isAFullCommitHash(part string) bool {
-	return len(part) == 64 && hashRegexp.MatchString(part)
+	return len(part) == commitIDStringLength && hashRegexp.MatchString(part)
 }
 
 // revResolve return the first resolve of 'rev' - by hash, branch or tag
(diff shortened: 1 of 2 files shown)


pkg/graveler/ref/resolve_ref.go
```diff
@@ -26,7 +26,33 @@ func isAHash(part string) bool {
 
 // revResolve return the first resolve of 'rev' - by hash, branch or tag
 func revResolve(ctx context.Context, store Store, addressProvider ident.AddressProvider, repository *graveler.RepositoryRecord, rev string) (*graveler.ResolvedRef, error) {
-	resolvers := []revResolverFunc{revResolveAHash, revResolveBranch, revResolveTag}
+	// looking for commit first - a full commit takes precedence over a branch or a tag
+	r, err := revResolveAHash(ctx, store, addressProvider, repository, rev)
```
nopcoder, line 30: Can we keep one set of func and just split the `revResolveAHash` to `revResolveAHash` and `revResolveAHashPrefix`?
This way we can place `revResolveAHash` first and `revResolveAHashPrefix` last.

nopcoder (changes requested):

nopcoder (approved):
Looks good the 64 probably will fail as it looks like a magic number - consider const.

PR #4748, opened 2022-12-06
Bump metadata client version in the docs

itaiad200:


talSofer (approved):
Pls merge after the client is released to s3 and mvn central

nopcoder (approved):

PR #4751, opened 2022-12-06
Store relative path in stage and link physical address

### LINKED ISSUE #4750: Staged and linked objects are referenced using absolute path
The API StageObject and LinkPhysicalAddress accepts full physical address and store it as full inside our catalog/stage.
The expected behaviour is when the link is generated by lakeFS. Pointing to the underlying repository storage namespace - lakefs should manage this data.

nopcoder:
Based on repository storage namespace, use relative path to object when data is kept under our managed storage.

Fix #4750

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -376,7 +376,7 @@ func (c *Controller) LinkPhysicalAddress(w http.ResponseWriter, r *http.Request,
 	}
 
 	writeTime := time.Now()
-	physicalAddress, addressType := physicalAddressFullToRelative(repo.StorageNamespace, StringValue(body.Staging.PhysicalAddress))
+	physicalAddress, addressType := normalizePhysicalAddress(repo.StorageNamespace, StringValue(body.Staging.PhysicalAddress))
 
 	// Because CreateEntry tracks staging on a database with atomic operations,
 	// _ignore_ the staging token here: no harm done even if a race was lost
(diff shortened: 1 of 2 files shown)


pkg/api/controller.go
```diff
@@ -418,6 +420,19 @@ func (c *Controller) LinkPhysicalAddress(w http.ResponseWriter, r *http.Request,
 	writeResponse(w, http.StatusOK, response)
 }
 
+// physicalAddressFullToRelative return relative address based on storage namespace if possible. If address doesn't match
+// the storage namespace prefix, the return address type is full.
+func physicalAddressFullToRelative(storageNamespace string, physicalAddress string) (string, catalog.AddressType) {
```
itaiad200, line 425:
```suggestion
func normalizePhysicalAddress(storageNamespace string, physicalAddress string) (string, catalog.AddressType) {
```

pkg/api/controller.go
```diff
@@ -376,14 +376,16 @@ func (c *Controller) LinkPhysicalAddress(w http.ResponseWriter, r *http.Request,
 	}
 
 	writeTime := time.Now()
+	physicalAddress, addressType := physicalAddressFullToRelative(repo.StorageNamespace, StringValue(body.Staging.PhysicalAddress))
+
 	// Because CreateEntry tracks staging on a database with atomic operations,
 	// _ignore_ the staging token here: no harm done even if a race was lost
 	// against a commit.
 	entryBuilder := catalog.NewDBEntryBuilder().
```
itaiad200, line 384: `NewDBEntryBuilder` 🤮

pkg/api/controller_test.go
```diff
@@ -2092,6 +2092,37 @@ func TestController_ObjectsStageObjectHandler(t *testing.T) {
...
 	})
 
+	t.Run("stage object in storage ns", func(t *testing.T) {
+		linkResp, err := clt.GetPhysicalAddressWithResponse(ctx, "repo1", "main", &api.GetPhysicalAddressParams{Path: "foo/bar2"})
+		verifyResponseOK(t, linkResp, err)
+		if linkResp.JSON200 == nil {
+			t.Fatalf("GetPhysicalAddress non 200 response - status code %d", linkResp.StatusCode())
+		}
+		const expectedSizeBytes = 38
+		resp, err := clt.StageObjectWithResponse(ctx, "repo1", "main", &api.StageObjectParams{Path: "foo/bar2"}, api.StageObjectJSONRequestBody{
```
itaiad200, line 2102: This test will soon break with the limitation of staging object in the namespace. Consider changing the call to Link

itaiad200 (approved):
LGTM

pkg/api/controller.go
nopcoder: Don't like builder pattern?

nopcoder: NewEntryBuilder -> different pr

pkg/api/controller_test.go
nopcoder: Update the code to use LinkPhysicalAddress

PR #4754, opened 2022-12-07
Uncommitted GC - collect full path if part of repo storage namespace

### LINKED ISSUE #4759: Uncommitted GC committed entries with abs path should be considered when part of storage namespace
In case the path is abs and point to our storage namespace, we should convert them into relative path to process like any other repository managed object.

nopcoder:
When entry metadata physical path points to an object under the repository storage namespace - we consider it as data we manage.

Fix #4759

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
@@ -33,9 +33,12 @@ class NaiveCommittedAddressLister extends CommittedAddressLister {
       else clientStorageNamespace + "/"
 
     var storageScheme = new Path(clientStorageNamespace).toUri.getScheme
-    if (storageScheme.nonEmpty) {
-      storageScheme += ":"
+    if (storageScheme.isEmpty) {
+      throw new IllegalArgumentException(
+        s"Invalid storage namespace - missing scheme $clientStorageNamespace"
+      )
     }
+    storageScheme += ":"
 
     filterAddresses(spark, df, normalizedClientStorageNamespace, storageScheme)
   }
@@ -57,7 +60,7 @@ class NaiveCommittedAddressLister extends CommittedAddressLister {
         // - non relative with no schema (relative)
         addrType.equals(AddressType.RELATIVE.name) ||
         addr.startsWith(normalizedClientStorageNamespace) ||
-        (storageScheme.nonEmpty && !addr.startsWith(storageScheme))
+        !addr.startsWith(storageScheme)
       })
       .map(row => {
         val (addrType, addr) = (row.getString(0), row.getString(1))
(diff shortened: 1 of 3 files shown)


clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
```diff
@@ -25,8 +25,7 @@ class NaiveCommittedAddressLister extends CommittedAddressLister {
       .filter(
         (col("address_type") === AddressType.RELATIVE.name) ||
           // Backwards compatability with entries prior to address_type
-          (col("address_type") === AddressType.BY_PREFIX_DEPRECATED.name &&
-            !col("address_type").contains("://"))
+          col("address").startsWith(normalizedStorageNamespace)
```
N-o-Z, line 28: We should modify the comment above.

N-o-Z (approved):
Approved with one comment

johnnyaug (approved):
LGTM

clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
nopcoder: True - I'll need to add code that will trim the storage prefix so the rest of the code will work the same with this physical address.

johnnyaug (approved):

clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
```diff
@@ -8,28 +8,50 @@ import io.treeverse.lakefs.catalog.Entry.AddressType
...
+      .map(row => {
+        val (addrType, addr) = (row.getString(0), row.getString(1))
+        if (addrType.equals(AddressType.RELATIVE.name)) {
+          (addr)
+        } else {
+          (addr.substring(normalizedStorageNamespace.length - 1))
+        }
+      })
+      .distinct()
+      .toDF("address")
```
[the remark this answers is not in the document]
nopcoder: I'll keep the efficient one as it will search only for relevant paths and just the prefix.

clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
```diff
@@ -1,35 +1,53 @@
...
+
+    val normalizedClientStorageNamespace =
+      if (clientStorageNamespace.endsWith("/")) clientStorageNamespace
+      else clientStorageNamespace + "/"
+
+    import spark.implicits._
     df
+      .select("address_type", "address")
+      .filter(row => {
+        val (addrType, addr) = (row.getString(0), row.getString(1))
```
N-o-Z, line 37: The current code ignores address_type BY_PREFIX_DEPRECATED which are not in full path format.
BY_PREFIX_DEPRECATED can have 3 type of addresses:
1. Relative path in namespace
2. Full path in namespace
3. Full path outside of namespace
We need to take all under consideration

N-o-Z (changes requested):
Thanks Barak - see my comment

clients/spark/core/src/main/scala/io/treeverse/gc/CommittedAddressLister.scala
nopcoder: Addressed the last path - type deprecated with relative path

nopcoder:
Added test for listing addresses of committed data.

N-o-Z (approved):
💣 🔥 💥

PR #4763, opened 2022-12-07
GC: delete absolute paths if they are under the storage namespace

### LINKED ISSUE #4762: GC: getRangeAddresses ignores all addresses with type != relative
getRangeAddresses function in GarbageCollector filters result by addressType.isRelative.
This in fact ignores the following addresses:
1. Addresses under the repository namespace which were added prior to existence of address_type field and their type is by_prefix_deprecated
2. Addresses that were added using stageObject/linkObject which are under the namespace but have a full path format

The implications of this bug is that potentially address that were expired and expected to be deleted are being filtered out

johnnyaug:
Fixes #4762.

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
@@ -17,7 +17,6 @@ import java.net.URI
 import java.time.format.DateTimeFormatter
 import java.time.{LocalDateTime, ZoneOffset}
 import java.util.UUID
-import org.apache.commons.lang3.StringUtils
 
 trait RangeGetter extends Serializable {
 
@@ -25,10 +24,9 @@ trait RangeGetter extends Serializable {
    */
   def getRangeIDs(commitID: String, repo: String): Iterator[String]
 
-  /** @return all object addresses in range rangeID on repo.
-   *  Absolute addresses are ignored, unless they are under the given storage namespace.
+  /** @return all object addresses in range rangeID on repo
    */
-  def getRangeAddresses(rangeID: String, repo: String, storageNS: String): Iterator[String]
+  def getRangeAddresses(rangeID: String, repo: String): Iterator[String]
 }
 
 class LakeFSRangeGetter(val apiConf: APIConfigurations, val configMapper: ConfigMapper)
@@ -52,18 +50,15 @@ class LakeFSRangeGetter(val apiConf: APIConfigurations, val configMapper: Config
         .map(o => new String(o.id))
   }
 
-  def getRangeAddresses(rangeID: String, repo: String, storageNS: String): Iterator[String] = {
+  def getRangeAddresses(rangeID: String, repo: String): Iterator[String] = {
     val location = ApiClient
       .get(apiConf)
       .getRangeURL(repo, rangeID)
     SSTableReader
       .forRange(configMapper.configuration, location)
       .newIterator()
-      .filter(e => e.message.addressType.isRelative || e.message.address.startsWith(storageNS))
-      .map(e =>
-        if (e.message.addressType.isRelative) e.message.address
-        else StringUtils.removeStart(e.message.address, storageNS)
-      )
+      .filter(_.message.addressType.isRelative)
+      .map(a => a.message.address)
   }
 }
 
@@ -141,7 +136,6 @@ class GarbageCollector(val rangeGetter: RangeGetter) extends Serializable {
       expiredRangeIDs: Dataset[String],
       keepRangeIDs: Dataset[String],
       repo: String,
-      storageNS: String,
       numRangePartitions: Int,
       numAddressPartitions: Int
   ): Dataset[String] = {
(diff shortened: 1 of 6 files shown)


clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -50,15 +52,18 @@ class LakeFSRangeGetter(val apiConf: APIConfigurations, val configMapper: Config
...
+  def getRangeAddresses(rangeID: String, repo: String, storageNS: String): Iterator[String] = {
     val location = ApiClient
       .get(apiConf)
       .getRangeURL(repo, rangeID)
     SSTableReader
       .forRange(configMapper.configuration, location)
       .newIterator()
-      .filter(_.message.addressType.isRelative)
-      .map(a => a.message.address)
+      .filter(e => e.message.addressType.isRelative || e.message.address.startsWith(storageNS))
```
N-o-Z, line 62: @johnnyaug Please note that this doesn't solve the issue for address type BY_PREFIX_DEPRECATED which is a relative path.
BY_PREFIX_DEPRECATED can theoretically have 3 type of addresses:

1. Relative path in namespace
2. Full path in namespace
3. Full path outside of namespace
You need to take all under consideration

@nopcoder is facing the same issue in https://github.com/treeverse/lakeFS/pull/4754

johnnyaug: I'm on with not solving it. It's been deprecated for a while now

nopcoder (approved):
LGTM - agree it will not handle the deprected type, but I don't think we have anyone running and migrating data from that point in time. And for GC vs uncommitted the source is the metadata so there will be no harm to the user data.

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -17,16 +17,18 @@ import java.net.URI
...
 
 trait RangeGetter extends Serializable {
 
   /** @return all rangeIDs in metarange of commitID on repo.
    */
   def getRangeIDs(commitID: String, repo: String): Iterator[String]
 
-  /** @return all object addresses in range rangeID on repo
+  /** @return all object addresses in range rangeID on repo.
+   *  Absolute addresses are ignored, unless they are under the given storage namespace.
```
talSofer, line 29: * I would add here that the ignored addresses are of imported objects, and an explanation of absolute within the namespace and how di they get there
* Semantically, this method is getRangeAddressesOfDeltableObjects consider reflecting that in naming

test/spark/gc-tests/test_scenarios.json
```diff
@@ -160,6 +160,7 @@
       }
     ],
     "file_deleted": false,
+    "direct_upload": true,
```
talSofer, line 163: Can you please edit the test description of those that now involve direct uploads to reflect that?

test/spark/run-gc-test.sh
```diff
@@ -138,11 +138,16 @@ prepare_for_gc() {
...
+  if [[ $LAKEFS_BLOCKSTORE_TYPE == "s3" ]]; then
+    direct_flag_for_non_deleted='--direct'
+  fi
+  run_lakectl fs upload ${direct_flag_for_non_deleted} "lakefs://${repo}/main/not_deleted_file3" -s /local/gc-tests/sample_file
   run_lakectl commit "lakefs://${repo}/main" -m "add three files not to be deleted" --epoch-time-seconds 0
 
   run_lakectl branch create "lakefs://${repo}/a${test_id}" -s "lakefs://${repo}/main"
-  run_lakectl fs upload "lakefs://${repo}/a${test_id}/file${test_id}" -s /local/gc-tests/sample_file
+  local direct_upload_flag=$("$(echo ${test_case} | jq -r '.direct_upload // false')" && [[ ${LAKEFS_BLOCKSTORE_TYPE} == "s3" ]] && echo '--direct' || echo '')
+   run_lakectl fs upload ${direct_upload_flag} "lakefs://${repo}/a${test_id}/file${test_id}" -s /local/gc-tests/sample_file
```
talSofer, line 150: Shouldn't "lakefs://${repo}/a${test_id}/file${test_id}" be a physical address on s3 within the storage namespace?

test/spark/run-gc-test.sh
```diff
@@ -138,11 +138,16 @@ prepare_for_gc() {
   run_lakectl repo create "lakefs://${repo}" "${STORAGE_NAMESPACE}/${repo}/"
   run_lakectl fs upload "lakefs://${repo}/main/not_deleted_file1" -s /local/gc-tests/sample_file
   run_lakectl fs upload "lakefs://${repo}/main/not_deleted_file2" -s /local/gc-tests/sample_file
-  run_lakectl fs upload "lakefs://${repo}/main/not_deleted_file3" -s /local/gc-tests/sample_file
+  local direct_flag_for_non_deleted=''
+  if [[ $LAKEFS_BLOCKSTORE_TYPE == "s3" ]]; then
+    direct_flag_for_non_deleted='--direct'
+  fi
+  run_lakectl fs upload ${direct_flag_for_non_deleted} "lakefs://${repo}/main/not_deleted_file3" -s /local/gc-tests/sample_file
```
talSofer, line 145: Shouldn't "lakefs://${repo}/main/not_deleted_file3" be a physical address on s3 outside the storage namespace?

talSofer (changes requested):
Thanks!
Added minor suggestions, blocking because of the question on the integration tests part

test/spark/run-gc-test.sh
johnnyaug: The "--direct" flag handles everything under the hood - it does a get+link physical address from the client side.

test/spark/gc-tests/test_scenarios.json
johnnyaug: I tried not to make big changes in the tests, so it really doesn't change the description - the text there still holds, I think.

talSofer: The text still holds, but if the test is failing there is no hint that this may be related to directly written objects

talSofer (approved):
Looks good, thanks!

PR #4764, opened 2022-12-07
Prepare uncommitted catalog use relative path for managed full path

### LINKED ISSUE #4758: Prepare Uncommitted GC include relative path for abs address inside storage namespace
Parquet files will include the address relative to storage namespace - in case we find abs path - we will trim the address in case it is part of the repo storage namespace.

nopcoder:
Fix #4758

### DIFF
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -1844,8 +1844,8 @@ func (c *Catalog) writeUncommittedLocal(ctx context.Context, repository *gravele
 	}
 
 	// Finished reading all staging area - no continuation token
-	hasMore := count > 0
-	return nil, hasMore, nil
+	hasData := count > 0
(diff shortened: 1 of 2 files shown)


pkg/catalog/catalog.go
```diff
@@ -1833,7 +1844,8 @@ func (c *Catalog) writeUncommittedLocal(ctx context.Context, repository *gravele
 	}
 
 	// Finished reading all staging area - no continuation token
-	return nil, hasData, nil
+	hasMore := count > 0
```
N-o-Z, line 1847: Suggest to change the param name back to hasData - this is an indication to whether we need to write anything to parquet, not if there's more data to read

N-o-Z (approved):
Thanks, LGTM only one comment

PR #4767, opened 2022-12-07
Fix: DataLister exception when trying to read a path with ':' in it

### LINKED ISSUE #4766: GC+: DataLister: Relative path in absolute URI: gc-migrate-lakefs-0-40-1:b40b9540-91f0-4add-96ba-205f17e03c57
When iterating over slices with the format `x:y` (this can happen on old link physical addresses syntax), we get an exception since HDFS assumes a schema and therefore treats the string as an absolute path

Link to known issue:
https://issues.apache.org/jira/browse/HDFS-14762

Trace:
```
	at io.treeverse.gc.UncommittedGarbageCollector.main(UncommittedGarbageCollector.scala)
Caused by: java.lang.IllegalArgumentException: java.net.URISyntaxException: Relative path in absolute URI: gc-migrate-lakefs-0-40-1:b40b9540-91f0-4add-96ba-205f17e03c57
	at org.apache.hadoop.fs.Path.initialize(Path.java:263)
	at org.apache.hadoop.fs.Path.<init>(Path.java:221)
	at org.apache.hadoop.fs.Path.<init>(Path.java:119)
	at io.treeverse.gc.ParallelDataLister.$anonfun$listData$1(DataLister.scala:55)
```

N-o-Z:
Closes #4766

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/gc/DataLister.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/gc/DataLister.scala
@@ -52,7 +52,9 @@ class ParallelDataLister extends DataLister with Serializable {
     val slices = listPath(configMapper, path)
     val objectsPath = path.toString // udf require serializable string and not Path
     val objectsUDF = udf((sliceId: String) => {
-      val slicePath = new Path(objectsPath, sliceId)
+      val p =
+        s"$objectsPath$sliceId" // WA for https://issues.apache.org/jira/browse/HDFS-14762 (object path is already normalized with trailing '/')
+      val slicePath = new Path(p)
       listPath(configMapper, slicePath).toSeq
         .map(s => (s.path, s.lastModified))
     })
--- a/clients/spark/core/src/main/scala/io/treeverse/gc/DataLister.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/gc/DataLister.scala
@@ -50,14 +50,11 @@ class ParallelDataLister extends DataLister with Serializable {
   override def listData(configMapper: ConfigMapper, path: Path): DataFrame = {
     import spark.implicits._
     val slices = listPath(configMapper, path)
-    var objectsPath = path.toString // udf require serializable string and not Path
+    val objectsPath = if (path.toString.endsWith("/")) path.toString else path.toString + "/"
+    // udf require serializable string and not Path
     val objectsUDF = udf((sliceId: String) => {
-      if (!objectsPath.endsWith("/")) {
-        objectsPath += "/"
-      }
-      val p =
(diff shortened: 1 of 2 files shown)


clients/spark/core/src/main/scala/io/treeverse/gc/DataLister.scala
```diff
@@ -50,9 +50,14 @@ class ParallelDataLister extends DataLister with Serializable {
   override def listData(configMapper: ConfigMapper, path: Path): DataFrame = {
     import spark.implicits._
     val slices = listPath(configMapper, path)
-    val objectsPath = path.toString // udf require serializable string and not Path
+    var objectsPath = path.toString // udf require serializable string and not Path
```
arielshaqed, line 53: Please keep it immutable (a `val`), by setting it to a conditional expression.

clients/spark/core/src/main/scala/io/treeverse/gc/DataLister.scala
```diff
@@ -50,9 +50,14 @@ class ParallelDataLister extends DataLister with Serializable {
...
     val objectsUDF = udf((sliceId: String) => {
-      val slicePath = new Path(objectsPath, sliceId)
+      if (!objectsPath.endsWith("/")) {
+        objectsPath += "/"
+      }
+      val p =
+        s"$objectsPath$sliceId" // WA for https://issues.apache.org/jira/browse/HDFS-14762
```
arielshaqed, line 59: Just making sure: sliceId is always relative, right?

Also, why not just add the strings here?

clients/spark/core/src/test/scala/io/treeverse/gc/DataListerSpec.scala
```diff
@@ -64,6 +64,26 @@ class ParallelDataListerSpec
...
+
+          val path = new Path(dataDir.toURI)
+          val configMapper = new ConfigMapper(
+            spark.sparkContext.broadcast(
+              HadoopUtils.getHadoopConfigurationValues(spark.sparkContext.hadoopConfiguration)
+            )
+          )
+          val df =
+            new ParallelDataLister().listData(configMapper, path).sort("base_address")
+          df.count() should be(1)
```
arielshaqed, line 84: Please check that we got the correct path here. I think it can be readable, and test a bit more.

arielshaqed (commented):
Neat! Not requesting or approving because I lack context. Comments are about Scalaishness and general robustness.

clients/spark/core/src/main/scala/io/treeverse/gc/DataLister.scala
N-o-Z: Done

nopcoder (approved):
lgtm - sad story for hadoop Path API

PR #4787, opened 2022-12-12
GC+: DataLister fix get first slice

### LINKED ISSUE #4785: GC+: Fix firstSlice

N-o-Z:
Closes #4785


N-o-Z:
@nopcoder  Refactored as discussed

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -49,6 +50,18 @@ object UncommittedGarbageCollector {
     )
     // Read objects from data path (new repository structure)
     var dataDF = new ParallelDataLister().listData(configMapper, dataPath)
+
+    // Save first slice from data prefix
+    if (!dataDF.isEmpty) {
+      // Need the before filter to to exclude slices that are not actually read
+      val slices = dataDF.filter(
+        col("last_modified") < before.getTime && !col("base_address").startsWith(repo)
+      )
```
eden-ohana, line 59: Why does it mean we haven't read it?

N-o-Z: We need to choose the first slice that we actually read data from. When we list the data - we provide a buffer time (before) which we filter all objects which have a creation date < before. If we select the first slice without this filter we might choose a slice that its data wasn't read at all. If we use this slice in the incremental run it will mean that we will stop reading data too early and miss objects that weren't read in the previous run

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -49,6 +50,18 @@ object UncommittedGarbageCollector {
     )
     // Read objects from data path (new repository structure)
     var dataDF = new ParallelDataLister().listData(configMapper, dataPath)
+
+    // Save first slice from data prefix
+    if (!dataDF.isEmpty) {
```
nopcoder, line 55: we don't need this check, we apply it after the filter

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -49,6 +50,18 @@ object UncommittedGarbageCollector {
...
+      // Need the before filter to to exclude slices that are not actually read
```
nopcoder, line 56: can we extract this implementation from liar objects. it seems until now this method did one thing and now it does two.

nopcoder (changes requested):

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -54,19 +54,30 @@ object UncommittedGarbageCollector {
...
+    dataDF = dataDF.union(oldDataDF).filter(col("last_modified") < before.getTime)
 
     dataDF
   }
 
+  def getFirstSlice(dataDF: DataFrame, repo: String): String = {
+    var firstSlice = ""
+    // Need the before filter to to exclude slices that are not actually read
+    val slices =
+      dataDF.filter(col("address").startsWith("data") && !col("base_address").startsWith(repo))
```
nopcoder, line 74: - starts with "data/"

nopcoder (approved):
Approved, just align the data check match "data/"

PR #4791, opened 2022-12-12
Reject lakeFS output committer

arielshaqed:
Not possible to attain reasonable performance on lakeFS KV.


arielshaqed:
Docs change, we don't need tests to pass where we're going...

PR #4799, opened 2022-12-12
GC report expired addresses also as text

### LINKED ISSUE #4800: GC report expired address as text to simplify the use with rclone
Reporting the expired address to delete in text format will enable simple instructions on how backup and restore

nopcoder:
Close #4800

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
@@ -35,7 +35,7 @@ class LakeFSRangeGetter(val apiConf: APIConfigurations, val configMapper: Config
   def getRangeIDs(commitID: String, repo: String): Iterator[String] = {
     val conf = configMapper.configuration
     val apiClient = ApiClient.get(apiConf)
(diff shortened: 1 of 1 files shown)


clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -447,28 +447,36 @@ object GarbageCollector {
...
       .partitionBy(MARK_ID_KEY)
       .mode(SaveMode.Overwrite)
       .parquet(gcAddressesLocation)
 
-    writeAddressesMarkMetadata(runID, markID, gcAddressesLocation, gcCommitsLocation)
-
     println("Expired addresses:")
     expiredAddresses.show()
+
+    // write expired addresses as text - write to the "[addresses path] + .text"
```
talSofer, line 471: Can you please document that this is needed for b&r with rclone?

talSofer (approved):
LGTM, thanks!

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
nopcoder: Done

PR #4805, opened 2022-12-13
Uncommitted GC print mark_id and summary

### LINKED ISSUE #4804: Uncommitted GC print out summary with mark_id
Summary print out and later loggin is required as the rengerated output by 'mark' should be referenced as mark_id by the 'sweep' step.

nopcoder:
Close #4804


N-o-Z (approved):
LGTM

PR #4806, opened 2022-12-13
GC+: Add exceptions and error handling

### LINKED ISSUE #4803: GC+: Add exceptions and error handling

N-o-Z:
Closes #4803


clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -78,21 +78,21 @@ object UncommittedGarbageCollector {
...
-      Console.out.printf("Please provide a mark ID (%s) for sweep-only mode. Exiting...\n",
-                         LAKEFS_CONF_GC_MARK_ID
-                        )
-      System.exit(2)
+      throw new ParameterValidationException(
+        s"Please provide a mark ID ($LAKEFS_CONF_GC_MARK_ID) for sweep-only mode. Exiting...\n"
+      )
     } else if (shouldMark && markID.nonEmpty) {
-      Console.out.println("Can't provide mark ID for mark mode. Exiting...")
+      throw new ParameterValidationException("Can't provide mark ID for mark mode. Exiting...")
```
nopcoder, line 95: can remove the elses

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -237,22 +228,25 @@ object UncommittedGarbageCollector {
     cachedAddresses.write.text(s"$reportDst/deleted.text")
   }
 
-  private def reportPath(storageNamespace: String, runID: String): String = {
+  private def runPath(storageNamespace: String, runID: String): String = {
```
nopcoder, line 231: Can we name it 'formatRunPath' or something that we don't think we are actually running something.

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -199,11 +195,6 @@ object UncommittedGarbageCollector {
           .toDF()
         removed.collect()
       }
-    } catch {
-      case e: Throwable =>
-        success = false
```
nopcoder, line 204: Nice - just initialize `success` with false and set it to true at the end of the 'try' scope.

nopcoder (changes requested):
Looks good - minor comments, my main concern before we make this one is the `lakefs.gc.do_mark=true` as it will change the behaviour we just documented does not align with the current GC process.

N-o-Z:
> Looks good - minor comments, my main concern before we make this one is the `lakefs.gc.do_mark=true` as it will change the behaviour we just documented does not align with the current GC process.

Aligned defaults with committed garbage collector for now.
Modified the documentation (which had a bug in it as well)

nopcoder (approved):
LGTM - thanks!

PR #4808, opened 2022-12-13
Write GC summary file using UTF-8

### LINKED ISSUE #4644: GC: writeJsonSummary writes file with interleaved null characters
The function uses `stream.writeChars` on a JObject which results in a an invalid string, which causes an exception when trying to read into json object afterwards

```
[, {, , ", , r, , u, , n, , _, , i, , d, , ", , :, , ", , 1, , 4, , 1, , 9, , f, , c, , e, , a, , -, , a, , 4, , 4, , 1, , -, , 4, , 9, , 7, , b, , -, , b, , 3, , 0, , 1, , -, , 2, , 4, , 6, , c, , 7, , f, , 3, , 6, , 0, , 1, , b, , a, , ", , ,, , ", +178 more]
```

Suggested solution to use `writeBytes` instead

arielshaqed:
Tested by adding a unit test for writeJsonSummary.

Fixes #4644.

### DIFF
--- a/clients/spark/core/src/test/scala/io/treeverse/clients/GarbageCollectorSpec.scala
+++ b/clients/spark/core/src/test/scala/io/treeverse/clients/GarbageCollectorSpec.scala
@@ -223,7 +223,6 @@ class GarbageCollectorJsonOutputSpec extends AnyFunSpec with Matchers with Spark
               .toSeq
             written.size should be(1)
             val actualBytes = Files.readAllBytes(Paths.get(written(0).toString))
-            Console.out.println("GOT: " + new String(actualBytes, "UTF-8"))
             // Explicitly verify that we received UTF-8 encoded data!
             val actual = JsonMethods.parse(new String(actualBytes, "UTF-8"))
             (actual \ "gc_rules") should be(JString(gcRules))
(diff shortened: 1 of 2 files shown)


clients/spark/core/src/test/scala/io/treeverse/clients/GarbageCollectorSpec.scala
```diff
@@ -1,12 +1,29 @@
...
+import org.json4s.native.JsonMethods
+
+import java.nio.file.{Files, Path, Paths}
+
+trait TempDirectory {
+  def withTempDirectory(testMethod: (Path) => Any) {
+    val tempDir = Files.createTempDirectory("test-gc")
+
+    testMethod(tempDir)
+    FileUtils.deleteDirectory(tempDir.toFile)
```
johnnyaug, line 24: What happens if an exception is thrown in the test? Will the file still be deleted?

arielshaqed: True.  Fixed to delete.

arielshaqed (commented):
Thanks, good catch!

Pulling once tests pass.

PR #4809, opened 2022-12-14
Check if DDB Table exist before Create to reduce excessive permissions requirements  

### LINKED ISSUE #4761: Working with existing DynamoDB table flow: requires excessive permissions
When working with DynamoDB, [some users](https://lakefs.slack.com/archives/C01678929QT/p1670342268895299) prefer to create the DynamoDB externally (with IAC for example). 

**Current Flow:** 

1. Try create table 
2. Check type: `if error.Type == ResourceInUseException`  then table exists and it's [ok dismiss error](https://github.com/treeverse/lakeFS/blob/master/pkg/kv/dynamodb/store.go#L133). 
3. `else:...` handle error.  

**The problem:** 

Even when working with an existing DDB table the code requires to have excessive permissions (CreateTable, DeleteTable and UpdateTable).   
If the user did not provide those extra permissions it will be a different error than `ResourceInUseException` which causes the flow to fail. 

**Suggested solutions:**  

Could be that user indicates in the config that it's an existing table and the code will skip `CreateTableWithContext`.  
Another option is to try and `DescribeTable` before trying to create it, the downside here is that it will cause initial error in cases the table doesn't exist. 
Another option could be check if the error is [Permission related](https://github.com/treeverse/lakeFS/blob/master/pkg/kv/dynamodb/store.go#L133) and try to describe the table, only if the table does not exist handle this as real error.

Isan-Rivkin:
Fixes #4761

### DIFF
--- a/pkg/kv/dynamodb/store.go
+++ b/pkg/kv/dynamodb/store.go
@@ -105,12 +105,12 @@ func isTableExist(ctx context.Context, svc *dynamodb.DynamoDB, table string) (bo
 		TableName: aws.String(table),
 	})
 	if err != nil {
-		_, ok := err.(*dynamodb.ResourceNotFoundException)
-		if ok {
+		if aerr, ok := err.(awserr.Error); ok && aerr.Code() == dynamodb.ErrCodeResourceNotFoundException {
 			return false, nil
 		}
+		return false, err
 	}
-	return err == nil && describeOutput != nil, err
+	return true, nil
 }
 
 // setupKeyValueDatabase setup everything required to enable kv over postgres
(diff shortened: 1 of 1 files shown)


pkg/kv/dynamodb/store.go
```diff
@@ -99,9 +99,28 @@ func (d *Driver) Open(ctx context.Context, kvParams kvparams.KV) (kv.Store, erro
...
+		TableName: aws.String(table),
+	})
+	if err != nil {
+		_, ok := err.(*dynamodb.ResourceNotFoundException)
+		if ok {
+			return false, nil
+		}
+	}
+	return err == nil && describeOutput != nil, err
+}
```
guy-har, line 114: nit
```suggestion
	if err != nil {
		if aerr, ok := err.(awserr.Error); ok && aerr.Code() == dynamodb.ErrCodeResourceNotFoundException {
			return false, nil
		}
		return false, err
	}
	return true, nil
}
```

guy-har (approved):
Looks good.
Great solution to the permission issue 😄

PR #4822, opened 2022-12-15
docs: re-introduce the hooks reference

ozkatz:
When restructuring the lakeFS documentation, the hooks reference page got lost.
This PR re-introduces the page, along with a few updates to reflect the current state of hooks.

### DIFF
--- a/docs/reference/hooks.md
+++ b/docs/reference/hooks.md
@@ -1,0 +1,344 @@
+---
+layout: default
+title: Hooks
+parent: Reference
+description: Hooks configuration and API reference
+nav_order: 30
+has_children: false
+redirect_from: ../hooks.html
+---
+# Hooks
+{: .no_toc }
+
+{% include toc.html %}
+
+Like other version control systems, lakeFS allows you to configure `Actions` to trigger when predefined events occur.
+
+Supported Events:
+
+| Event                | Description                                                                    |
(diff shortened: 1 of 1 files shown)


docs/reference/hooks.md
```diff
@@ -0,0 +1,344 @@
...
+          end
+          if props.pattern and not regexp.match(props.pattern, current_value) then
+            error("current value for commit metadata field " .. k .. " does not match pattern: " .. props.pattern .. " - got: " .. current_value)
+          end
+        end
+```
+
+For more examples and configuration samples, check out the [examples/hooks/](https://github.com/treeverse/lakeFS/tree/master/examples/hooks) directory in the lakeFS repository.
+
+For the full set of exposed Lua APIs, usage and reasoning is available on the [proposal document](https://github.com/treeverse/lakeFS/blob/master/design/open/zero-deployment-hooks.md)
```
itaiad200, line 341: Design docs usually go out of sync and are hard to track. 
The intention of users reading this section is high and they would prefer a single, clear searchable page they can view the script options, I think we should add the supported list in bullets here with short description.

itaiad200 (approved):
The phoenix wiki comes back again :) 
Added a comment for the new section

docs/reference/hooks.md
ozkatz: will do so in a separate PR.

PR #4825, opened 2022-12-15
Fix commit to keep base entry in case staged entry with matched identity

### LINKED ISSUE #3756: Staging behaviour is unclear when uploading the exact same objects
It is not allowed to commit to a branch with 0 changes to objects. However the term "empty staging area" is not clear when uploading the same objects again.
**Example**: lakeFS committed data with object `foo.bar`. No entries in the staging area. The same object `foo.bar` is uploaded under the same name. The list of uncommitted changes doesn't show the object and the UI commit button is disabled, which makes sense. However I was able to commit using `lakectl`. The commit diff view shows `no changes`.

This occurs in the DB (postgres) implementation of graveler, KV implementation isn't finished yet but will be able to resolve this.

nopcoder:
Fix #3756


itaiad200 (approved):
Much clearer now, thanks!

PR #4828, opened 2022-12-16
Fix auth's wildcard Match function

### LINKED ISSUE #1996: Malicious user can DoS lakeFS CPU for some IAM patterns
# End result

Some (fairly reasonable) IAM pattern configurations by admins may allow unprivileged users to  consume large amounts of CPU on the lakeFS server using short payloads.

# Conditions

*IF* an administrator sets an IAM rule with _multiple_ asterisks, a user can cause IAM rule verification to consume large amounts of CPU.

# Workaround

Avoid IAM path patterns with more than 2 asterisks.

# Details

The wildcard matcher used (the same is used in MinIO) runs inefficiently and can backtrack unnecessarily (Ω(n^k) for input of length n with k asterisks).

* If an admin constructs a pattern with multiple asterisks, a user can consume vast amounts of CPU by requesting well-chosen file names.  For instance this benchmark:
   ```go
   package wildcard

   import (
	"testing"
   )

   func rep(s string, n int) string {
	   ret := ""
   	   for i := 0; i < n; i++ {
		   ret += s
	   }
	   return ret
   }

   func BenchmarkMatch(b *testing.B) {
	   var (
		   pattern = "a*b*c*z"
		   str = rep("a", 1000) + rep("b", 1000) + rep("c", 1000)
	   )
	   for i := 0; i < b.N; i++ {
		   if Match(pattern, str) {
			   b.Errorf("expecteded %s NOT to match %s", str, pattern)
		   }
	   }
   }
   ```

   consumes 4 seconds for the check (i7 @ 1.30 GHz, aka "my laptop"), with input of length 3000.

nopcoder:
Address performance and malicious usage.
Fix #1996 

Added unit test code from MinIO for testing the functionality.

Used the following modified benchmark code supplied by @arielshaqed to test code for malicious usage:

```go
package wildcard

import (
	"strings"
	"testing"
)

func BenchmarkMatch(b *testing.B) {
	const pattern = "a*b*c*z"
	str := strings.Repeat("a", 1000) + strings.Repeat("b", 1000) + strings.Repeat("c", 1000)
	for i := 0; i < b.N; i++ {
		if Match(pattern, str) {
			b.Errorf("Expecteded '%s' NOT to match '%s'", str, pattern)
		}
	}
}

func BenchmarkMatchDuh(b *testing.B) {
	const pattern = "a*a*a*z"
	str := strings.Repeat("a", 2000)
	for i := 0; i < b.N; i++ {
		if Match(pattern, str) {
			b.Errorf("Expecteded '%s' NOT to match '%s'", str, pattern)
		}
	}
}
```

Test results of current implementation:
```
BenchmarkMatch-10                      1        4474831292 ns/op
BenchmarkMatchDuh-10                   1        11981327834 ns/op
```

Test results of the new implementation:
```
BenchmarkMatch-10                 176950              6804 ns/op
BenchmarkMatchDuh-10              264994              4511 ns/op
```

### DIFF
--- a/pkg/auth/wildcard/match.go
+++ b/pkg/auth/wildcard/match.go
@@ -1,53 +1,82 @@
-/*
- * MinIO Cloud Storage, (C) 2015, 2016 MinIO, Inc.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
 package wildcard
 
-// Match -  finds whether the text matches/satisfies the pattern string.
-// supports  '*' and '?' wildcards in the pattern string.
-// unlike path.Match(), considers a path as a flat name space while matching the pattern.
-// The difference is illustrated in the example here https://play.golang.org/p/Ega9qgD4Qz .
-func Match(pattern, name string) (matched bool) {
-	if pattern == "" {
-		return name == pattern
+import (
+	"unicode/utf8"
+)
+
+// Match reports whether name matches the shell pattern.
+// This is a strip down version of Go's `path.Match` https://pkg.go.dev/path#Match
+func Match(pattern, name string) bool {
+Pattern:
+	for len(pattern) > 0 {
+		var (
+			star  bool
+			chunk string
+		)
+		star, chunk, pattern = scanChunk(pattern)
+		if star && chunk == "" {
+			// Trailing * matches rest of string
+			return true
+		}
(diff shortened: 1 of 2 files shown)


arielshaqed (approved):
Nice one, thanks!

I had some difficulty understanding the algorithm at first (it doesn't backtrack or consider more than one possibility at any point!).  For future readers and especially future me here are my notes on how it works.  So very safe to skip this!

Call a "fixword" a _maximal_ portion of the pattern consisting only of regular characters and `?`s.  So a fixword has to begin after `*` or at the beginning of the string, and it has to end before `*` or at the end of the string.

Each fixword matches a fixed length of string.  Now a pattern is a list of fixwords separated by `*`s.  Consider a fixword that is _not_ preceded by a `*`; that's an easy match to find because it can only be at one place.  Consider a fixword that _is_ preceded by a `*`; if it matches at multiple places then it is always safe to match it at the first possible location: either the pattern ends after that fixword in which case there's only one possible location, or the pattern continues with `*`, in which case that `*` can "expand" to pick up all characters and the next match of the fixword.

Either way, it's safe, so everything works.

nopcoder:
@arielshaqed thank you, credit to the Go team - I'll add the above explanation to the code so other will enjoy it.

PR #4834, opened 2022-12-19
lakectl show commit

### LINKED ISSUE #3973: lakectl show doesn't accept commit uri
It accepts a repository uri and commit as an argument which is a misalignment with the rest of the system.

nopcoder:
Update `lakectl show` command to use subcommand `commit` instead of flag to align the the way we address commits.
Also fixes the URI format vs commit ID when specified as a flag.

Fix #3973


N-o-Z (commented):
@nopcoder  I don't really understand the change,

Looking at the code 'show' command it only handles the show commits command, while the Short description states it is used for several entities.
Also I would expect to see the full list of current supported entities (it causes ambiguity with the user)

Secondly I see the each entity has its own show command defined under the entity cmd (for example branch)

If we have a show command maybe we should consider moving all the other show commands

nopcoder:
@N-o-Z The associated issue describes the problem in `lakectl show` command which this PR fixes - using full URI to reference a commit.
I'll update the command 'show' general description to be more general.

About the command description and the current implementation - 
These are the current 'show' commands I found:
- lakectl auth policies show
- lakectl tag show
- lakectl branch show

We can have "alias" under `lakectl show` to display the above, but I don't think it is required.
Do you have other suggestion how to layout the commands when `commit` is not category but we supply `log` to list them and `show` to display the commit information.

nopcoder:
@N-o-Z updated the command description and the docs reference.

N-o-Z (approved):
Thanks

PR #4835, opened 2022-12-19
Fix github action use checkout v3

nopcoder:


itaiad200 (approved):
Is this the only place where we checkout?

nopcoder:
@itaiad200 with v2 that we left out

PR #4848, opened 2022-12-22
API upload object without using tmp file

### LINKED ISSUE #4849: API UploadObject uses tmp file for large objects
The current implementation uses http.Request's `FormFile` which parse multipart request and read the content we require to update.
The runtime implementation reads up until 32MB and uses temporary file for content bigger than that.
For lakeFS running a large number of request to upload large objects in parallel this can cause out of disk space and service degradation.

nopcoder:
Fix #4849

Multipart request read content directly to block adapter.
The issue is that the Go implementation uses local temporary file when accessing content size >32MB.
In order to support reading the data directly in this case we stream the request data into the adapter's `Put` request without specify the content size.

To support Put with reader without size (stream), updated the S3 to use of s3manager upload in case the size == -1.
All other adapters ignore the size in the request to upload.

Tested manually with file size ~170MB

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -2141,7 +2141,7 @@ func (c *Controller) UploadObject(w http.ResponseWriter, r *http.Request, reposi
 		writeError(w, r, http.StatusInternalServerError, err)
 		return
 	}
-	if !strings.HasPrefix(mt, "multipart/") {
+	if mt != "multipart/form-data" {
 		writeError(w, r, http.StatusInternalServerError, http.ErrNotMultipart)
 		return
 	}
@@ -2150,22 +2150,26 @@ func (c *Controller) UploadObject(w http.ResponseWriter, r *http.Request, reposi
 		writeError(w, r, http.StatusInternalServerError, http.ErrMissingBoundary)
 		return
 	}
+
 	reader := multipart.NewReader(r.Body, boundary)
 	var (
 		contentUploaded bool
 		contentType     string
 		blob            *upload.Blob
 	)
-	for {
+	for !contentUploaded {
 		part, err := reader.NextPart()
 		if err == io.EOF {
 			break
 		}
+		if err != nil {
+			writeError(w, r, http.StatusInternalServerError, err)
+			return
+		}
 		contentType = part.Header.Get("Content-Type")
 		partName := part.FormName()
-		// part is an io.Reader, deal with it
-		if !contentUploaded && partName == "content" {
-			// upload the first "content"
+		if partName == "content" {
+			// upload the first "content" and exit the loop
 			address := c.PathProvider.NewPath()
 			blob, err = upload.WriteBlob(ctx, c.BlockAdapter, repo.StorageNamespace, address, part, -1, block.PutOpts{StorageClass: params.StorageClass})
 			if err != nil {
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io"
 	"mime"
+	"mime/multipart"
 	"net/http"
 	"net/mail"
 	"net/url"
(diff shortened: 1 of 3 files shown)


pkg/api/controller.go
```diff
@@ -2134,21 +2135,50 @@ func (c *Controller) UploadObject(w http.ResponseWriter, r *http.Request, reposi
...
 	}
-	defer func() { _ = file.Close() }()
-	contentType := handler.Header.Get("Content-Type")
-	address := c.PathProvider.NewPath()
-	blob, err := upload.WriteBlob(ctx, c.BlockAdapter, repo.StorageNamespace, address, file, handler.Size, block.PutOpts{StorageClass: params.StorageClass})
-	if err != nil {
+	if !strings.HasPrefix(mt, "multipart/") {
+		writeError(w, r, http.StatusInternalServerError, http.ErrNotMultipart)
+		return
+	}
```
arielshaqed, line 2147: Not sure this works for _any_ multipart Content-Type, perhaps we should limit ourselves to a list of known types?  (Package http probably limits itself in some way, so if we do the same would we be safe?)

pkg/api/controller.go
```diff
@@ -2134,21 +2135,50 @@ func (c *Controller) UploadObject(w http.ResponseWriter, r *http.Request, reposi
...
+		if !contentUploaded && partName == "content" {
+			// upload the first "content"
+			address := c.PathProvider.NewPath()
+			blob, err = upload.WriteBlob(ctx, c.BlockAdapter, repo.StorageNamespace, address, part, -1, block.PutOpts{StorageClass: params.StorageClass})
+			if err != nil {
+				_ = part.Close()
+				writeError(w, r, http.StatusInternalServerError, err)
+				return
+			}
+			contentUploaded = true
```
arielshaqed, line 2176: Why not `break` here?  Nothing remains that could be useful (I think).

pkg/api/controller.go
```diff
@@ -2134,21 +2135,50 @@ func (c *Controller) UploadObject(w http.ResponseWriter, r *http.Request, reposi
...
+	var (
+		contentUploaded bool
+		contentType     string
+		blob            *upload.Blob
+	)
+	for {
+		part, err := reader.NextPart()
+		if err == io.EOF {
+			break
+		}
```
ortz, line 2163: what happens if `err != io.EOF && err != nil`?

pkg/api/controller.go
nopcoder: Will fix - wanted to complete the body processing - but I see that Close will do it for me!

pkg/api/controller.go
ortz, line 2147: IIUC, the previous implementation allowed non-multipart uploads to go through this API endpoint, isn't it a breaking change?

nopcoder: Good point - I'll limit it to "multipart/form-data". The Go's request implementation support `form-data` and optional `mixed`.

pkg/api/controller.go
nopcoder: Thanks - adding a check

pkg/api/controller.go
nopcoder: Didn't saw this kind of support - can you send me an example? I used the lakectl, python sdk code and UI

nopcoder:
> I think that this can change behaviour with some S3 "compatible" storage backends, since we're no longer uploading data with a length. We might need to add an option to turn this off.

The the backend will get the size as the s3manager upload operation works with a single put or multi-part upload.

pkg/api/controller.go
nopcoder: > Also, the current implementation will be dangerous if we ever add another part to UploadObject in swagger.yml. Let's add a comment on its definition _not_ to do that!

Currently we added the information in the swagger description: `Only a single file per upload which must be named "content".`

PR #4854, opened 2022-12-23
test: improve the client disconnect test code

nopcoder:
Use client with short response timeout to simulate client disconnect before response.

### DIFF
--- a/pkg/api/controller_test.go
+++ b/pkg/api/controller_test.go
@@ -3400,12 +3400,14 @@ func TestController_PrepareGarbageCollectionUncommitted(t *testing.T) {
 
 func TestController_ClientDisconnect(t *testing.T) {
 	handler, deps := setupHandlerWithWalkerFactory(t, store.NewFactory(nil))
+
+	// setup lakefs
 	server := setupServer(t, handler)
 	clt := setupClientByEndpoint(t, server.URL, "", "")
 	_ = setupCommPrefs(t, clt)
 	cred := createDefaultAdminUser(t, clt)
 
-	clt = setupClientByEndpoint(t, server.URL, cred.AccessKeyID, cred.SecretAccessKey)
+	// setup repository
 	ctx := context.Background()
 	_, err := deps.catalog.CreateRepository(ctx, "repo1", onBlock(deps, "repo1"), "main")
 	testutil.Must(t, err)
--- a/pkg/api/controller_test.go
+++ b/pkg/api/controller_test.go
@@ -8,8 +8,8 @@ import (
 	"fmt"
 	"io"
 	"math"
-	"math/rand"
 	"mime/multipart"
+	"net"
 	"net/http"
 	"net/http/httptest"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


pkg/api/controller_test.go
```diff
@@ -3404,45 +3404,50 @@ func TestController_ClientDisconnect(t *testing.T) {
...
+		Transport: &http.Transport{
+			ResponseHeaderTimeout: time.Nanosecond,
+			DialContext:           dialer.DialContext,
+			DisableKeepAlives:     true,
+			IdleConnTimeout:       90 * time.Second,
+			ExpectContinueTimeout: 1 * time.Second,
+			ForceAttemptHTTP2:     false,
+		},
+	}
+	clt = setupClientByEndpoint(t, server.URL, cred.AccessKeyID, cred.SecretAccessKey, api.WithHTTPClient(httpClient))
```
N-o-Z, line 3428: Duplicate line 3408

pkg/api/controller_test.go
```diff
@@ -3404,45 +3404,50 @@ func TestController_ClientDisconnect(t *testing.T) {
 	clt := setupClientByEndpoint(t, server.URL, "", "")
 	_ = setupCommPrefs(t, clt)
 	cred := createDefaultAdminUser(t, clt)
-	clt = setupClientByEndpoint(t, server.URL, cred.AccessKeyID, cred.SecretAccessKey)
 
+	clt = setupClientByEndpoint(t, server.URL, cred.AccessKeyID, cred.SecretAccessKey)
```
N-o-Z, line 3408: Not needed

pkg/api/controller_test.go
```diff
@@ -3404,45 +3404,50 @@ func TestController_ClientDisconnect(t *testing.T) {
...
-	go func() {
-		w, _ := mpw.CreateFormFile("content", "file.data")
-		r := rand.New(rand.NewSource(time.Now().UnixNano()))
-		_, _ = io.CopyN(w, r, 1024)
-		cancel()
-		_, _ = io.CopyN(w, r, 1024)
-		_ = mpw.Close()
-	}()
+	// prepare a client that will not wait for a response and timeout
+	dialer := net.Dialer{
```
N-o-Z, line 3414: Didn't quite understand from the code why the client will immediately time out if dialer timeout is 30 seconds

N-o-Z (changes requested):
Thanks, one comment and a question since I'm not familiar with the package

pkg/api/controller_test.go
nopcoder: The first `clt` (without creds) does setup lakefs.
The second `clt` (this one - with creds) create a repo.

pkg/api/controller_test.go
nopcoder: The dialer is used to connect as part of the transport - these are the default values.
The transport uses also the defaults, with the exception of `ResponseHeaderTimeout` it will wait for a nanoseconds for the response headers, which means it will timeout before we get the response.

pkg/api/controller_test.go
N-o-Z: Correct me if I'm wrong - but I don't see where clt is used between this line and line 3428

N-o-Z (approved):
LGTM!

PR #4855, opened 2022-12-23
test: makefile and minor improvements 

nopcoder:
- Makefile: go generate auth package once
- Makefile: remove GO_TEST_MODULES not required anymore
- Test AWS SigVerify: run parallel, reduce number of iterations
- Load test: reduce total load test time, optional progress bar, code cleanup

### DIFF
--- a/pkg/loadtest/loader.go
+++ b/pkg/loadtest/loader.go
@@ -40,7 +40,7 @@ type Config struct {
 	KeepRepo         bool
 	Credentials      model.Credential
 	ServerAddress    string
-	NoProgress       bool
+	ShowProgress     bool
 }
 
 var (
@@ -70,7 +70,7 @@ func (t *Loader) Run() error {
 		return err
 	}
 	stopCh := make(chan struct{})
-	if !t.Config.NoProgress {
+	if t.Config.ShowProgress {
 		progressBar(t.Config.Duration)
 	}
 	out := new(SimpleScenario).Play(t.Config.ServerAddress, repoName, stopCh)
--- a/pkg/loadtest/loader.go
+++ b/pkg/loadtest/loader.go
@@ -40,6 +40,7 @@ type Config struct {
 	KeepRepo         bool
 	Credentials      model.Credential
 	ServerAddress    string
+	NoProgress       bool
 }
 
 var (
@@ -69,7 +70,9 @@ func (t *Loader) Run() error {
 		return err
 	}
 	stopCh := make(chan struct{})
-	progressBar(t.Config.Duration)
+	if !t.Config.NoProgress {
+		progressBar(t.Config.Duration)
+	}
 	out := new(SimpleScenario).Play(t.Config.ServerAddress, repoName, stopCh)
 	errs := t.streamRequests(out)
 	hasErrors := t.doAttack()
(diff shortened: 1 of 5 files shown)


pkg/loadtest/loader.go
```diff
@@ -40,6 +40,7 @@ type Config struct {
 	KeepRepo         bool
 	Credentials      model.Credential
 	ServerAddress    string
+	NoProgress       bool
```
N-o-Z, line 43: Maybe we can try to avoid the double negative:

```suggestion
	Progress       bool
```

pkg/loadtest/loader.go
```diff
@@ -69,7 +70,9 @@ func (t *Loader) Run() error {
 		return err
 	}
 	stopCh := make(chan struct{})
-	progressBar(t.Config.Duration)
+	if !t.Config.NoProgress {
```
N-o-Z, line 73:
```suggestion
	if t.Config.Progress {
```

pkg/loadtest/local_load_test.go
```diff
@@ -28,145 +28,105 @@ import (
...
+	storeMessage := &kv.StoreMessage{Store: kvStore}
+	authService := auth.NewKVAuthService(storeMessage, crypt.NewSecretStore([]byte("some secret")), nil, authparams.ServiceCache{}, logging.Default().WithField("service", "auth"))
+	meta := auth.NewKVMetadataManager("local_load_test", conf.GetFixedInstallationID(), conf.GetDatabaseType(), kvStore)
+
+	blockstoreType := os.Getenv(testutil.EnvKeyUseBlockAdapter)
+	if blockstoreType == "" {
+		blockstoreType = "mem"
+	}
+
+	storeMessage = &kv.StoreMessage{Store: kvStore}
```
N-o-Z, line 60: Already initialized at line 51?

pkg/loadtest/local_load_test.go
```diff
@@ -28,145 +28,105 @@ import (
...
+	viper.Set("database.kv_enabled", true)
```
N-o-Z, line 61: Not needed

pkg/loadtest/local_load_test.go
```diff
@@ -28,145 +28,105 @@ import (
...
+		Config:  conf,
+		KVStore: storeMessage,
+	})
+	testutil.MustDo(t, "build catalog", err)
+
+	source := catalog.NewActionsSource(c)
+	outputWriter := catalog.NewActionsOutputWriter(c.BlockAdapter)
+
+	// wire actions
+	actionsService := actions.NewService(ctx, actions.NewActionsKVStore(kv.StoreMessage{Store: kvStore}), source, outputWriter, &actions.DecreasingIDGenerator{}, &stats.NullCollector{}, true)
```
N-o-Z, line 74: You already have to storeMessage

N-o-Z (changes requested):
Thanks - some comments

pkg/loadtest/loader.go
nopcoder: Updated the code - just note that I use NoProgress in the first place to have the default `false`.  But for this use-case I don't mind having the default false on show progress.

pkg/loadtest/local_load_test.go
nopcoder: Need to update the actions to use pointer or interface instead of a copy

N-o-Z (approved):
LGTM!

PR #4856, opened 2022-12-23
test: uncommit gc test mount host jars folder

nopcoder:
Fix issue with readonly mount jars folder in docker for running spark job


N-o-Z (approved):
Thanks,

Just a question, why did we decide to use the 301 client for the tests?

nopcoder:
because we use spark 3.2

N-o-Z:
> because we use spark 3.2
OK, so why not use the core-312 target?

PR #4858, opened 2022-12-25
Enable running ripsecrets

### LINKED ISSUE #4857: Enable committing sensitive information to lakeFS repository

nopcoder:
- comment code when needed
- list secrets to ignore at .secretsignore

Enables using [ripsecrets](https://github.com/sirwart/ripsecrets)

Adding pre commit git hook will prevent secrets to be committed.

```sh
ripsecrets --install-pre-commit
```

Fix #4857


ortz:
@nopcoder does this mean that every developer, that wish to protect its secrets from being committed, needs to install this tool?
have we explored the possibility and adding some logic to the post-push (pre-base or something like that), where GitHub will search for invalid information in the commit blob and act accordingly (fail the push if it finds any)

nopcoder:
> @nopcoder does this mean that every developer, that wish to protect its secrets from being committed, needs to install this tool? have we explored the possibility and adding some logic to the post-push (pre-base or something like that), where GitHub will search for invalid information in the commit blob and act accordingly (fail the push if it finds any)

Sure, this is optional - sent request to github to enable push protection which will cover.
Can consider this option as local option to protect your local commits.

ortz:
> > @nopcoder does this mean that every developer, that wish to protect its secrets from being committed, needs to install this tool? have we explored the possibility and adding some logic to the post-push (pre-base or something like that), where GitHub will search for invalid information in the commit blob and act accordingly (fail the push if it finds any)
> 
> Sure, this is optional - sent request to github to enable push protection which will cover. Can consider this option as local option to protect your local commits.

I agree that's a good solution for the developer, less for the organization as there's no enforcement.
Anyway, I'm approving as it only adds protection.

ortz (approved):

PR #4862, opened 2022-12-25
Fix CredentialFromProto can fail to decode secret

### LINKED ISSUE #4861: lakeFS panic on nil access after fail to decode credentials secret
After setup, change the `auth.encryption.secret_key`.
Trying to loading, lakeFS will fail to decode the key's secret and panic.

nopcoder:
Fix #4861


N-o-Z (approved):
Thanks! Looks good!

PR #4864, opened 2022-12-26
Add missing Auditing fields into logger context  

### LINKED ISSUE #4843: Missing fields for Auditing
# Context

- When looking at Audit / Debugging logs some logs contain user information while other don't in the log fields.
After inspecting the code it seems that all the logs that contain user field (no bug) are requests made via the gateway while the ones that missing the field are via the Open API. 

- Missing operation_id in the s3_gateway as well. 

# References 

- The Gateway [Auth middleware](https://github.com/treeverse/lakeFS/blob/master/pkg/gateway/middleware.go#L65) that attaches the user field to the logger. 
- The Open API [Auth middleware](https://github.com/treeverse/lakeFS/blob/master/pkg/api/auth_middleware.go#L41) when executed adds the user to the context but **NOT to the logger.** 

# Desired Outcome 

- Have the user info and operation_id attached in all generated logs.

Isan-Rivkin:
Fixes #4843


pkg/api/auth_middleware.go
```diff
@@ -62,7 +61,8 @@ func AuthMiddleware(logger logging.Logger, swagger *openapi3.Swagger, authentica
 				return
 			}
 			if user != nil {
-				r = r.WithContext(auth.WithUser(r.Context(), user))
+				ctx := logging.AddFields(r.Context(), logging.Fields{logging.UserFieldKey: user.Username})
+				r = r.WithContext(auth.WithUser(ctx, user))
```
arielshaqed, line 65: The S3 gateway also has [this line](https://github.com/treeverse/lakeFS/blob/0dec04af1aaa97d95a8441cee01fec966477c4ca/pkg/gateway/middleware.go#L67).  Do we need to add it here?  (Perhaps delete it over there?)

arielshaqed (commented):
Thanks, this looks great!  How was this tested?

arielshaqed (approved):

ortz (approved):

Isan-Rivkin:
@arielshaqed it was tested manually by spinning up the server, trying different API operations and comparing logs where `audit_logs=true` before those fields were missing, now they are there.

pkg/api/auth_middleware.go
Isan-Rivkin: I don't think the mentioned line should be here since it represents additional context related to the sig s3 gw operations.

PR #4868, opened 2022-12-26
fix syntax highlighting with a matching theme

ozkatz:


ozkatz:
For reference, the theme is auto-generated using the following command:

```sh
$ cd docs/
$ bundle exec rougify style igorpro > ./assets/css/pygments/igorpro.css
```

johnnyaug (approved):

PR #4871, opened 2022-12-26
Remove KV from Graveler types

nopcoder:
Cleanup naming as previously we had KV and DB implementation of these types.

In order to keep the PR small - didn't remove the KV naming from `actions` and `auth`.


eden-ohana (approved):
LGTM

PR #4874, opened 2022-12-27
add an endpoint for sending statistics (when enabled)

### LINKED ISSUE #4860: Add an API endpoint for experimental features
Add a new endpoint to collect stats about experimental features usage when stats are enabled. 

`POST statistics/`

talSofer:
closes #4860

This endpoint aims to simplify experiments and, when enabled, collect statistics about feature usage in their very first steps. even before a new API that provides a capability is defined. 

### How was this tested 
* Unit tests 
* Manually tested that the endpoint is reachable 
* Tested manually that events are sent to a mock server


api/swagger.yml
```diff
@@ -1138,6 +1138,23 @@ components:
...
 
+    UsageEvent:
+      type: object
+      properties:
+        class:
+          description: stats event class
+          type: string
+        name:
+          description: stats event name
+          type: string
```
arielshaqed, line 1149: I am not sure what these two are.  Perhaps an example might clarify this documentation?

api/swagger.yml
```diff
@@ -3783,3 +3800,29 @@ paths:
...
+      summary: report usage event
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/UsageEvent"
+      responses:
+        204:
+          description: reported succssfully
```
arielshaqed, line 3818: Nit:
```suggestion
          description: reported successfully
```

api/swagger.yml
```diff
@@ -3783,3 +3800,29 @@ paths:
...
+      tags:
+        - statistics
+      operationId: reportUsageEvent
+      summary: report usage event
+      requestBody:
+        required: true
+        content:
+          application/json:
+            schema:
+              $ref: "#/components/schemas/UsageEvent"
```
arielshaqed, line 3815: I would like to be able to report multiple events in one message.  Not entirely sure why, but most stats reporting APIs that I know of allow collecting this.  The schema change will just be to pass a list of UsageEvents.

pkg/api/controller.go
```diff
@@ -3635,6 +3636,57 @@ func (c *Controller) GetLakeFSVersion(w http.ResponseWriter, r *http.Request) {
...
+		writeError(w, r, http.StatusBadRequest, "invalid usage class: cannot use an empty string")
+		return
+	}
+
+	if usageName == "" {
+		writeError(w, r, http.StatusBadRequest, "invalid usage name: cannot use an empty string")
+		return
+	}
+
+	if usageCount <= 0 {
```
arielshaqed, line 3658:
```suggestion
	if usageCount < 0 {
```
In some cases it might be easier to allow 0... despite it lowering performance :-/

pkg/api/controller.go
```diff
@@ -3635,6 +3636,57 @@ func (c *Controller) GetLakeFSVersion(w http.ResponseWriter, r *http.Request) {
...
+		Name:       usageName,
+		Repository: "",
+		Ref:        "",
+		SourceRef:  "",
+		UserID:     user.Username,
+		Client:     client,
+	}
+	for i := 0; i < body.Count; i++ {
+		c.Collector.CollectEvent(ev)
+	}
```
arielshaqed, line 3675: This will be very inefficient.  Can we change the collector interface to allow counts?

pkg/api/controller.go
arielshaqed, line 3658:
```suggestion
	if usageCount < 0 {
```
In some cases it might be easier to allow 0... despite it lowering performance :-/

arielshaqed (approved):
Nice...

Please consider adding a prefix to the reported event class, perhaps `experimental/` or `external/`.  I am somewhat unhappy that it will be so easy for a user to report events that we _do_ care about.

pkg/api/controller_test.go
```diff
@@ -3463,3 +3463,89 @@ func TestController_ClientDisconnect(t *testing.T) {
...
+
+	tests := []struct {
+		name               string
+		usageClass         string
+		usageName          string
+		usageCount         int
+		expectedStatusCode int
+	}{
+		{
+			name:               "single-event",
```
nopcoder, line 3479: Use '_' or space (that go test will turn to _) instead of '-'.

pkg/api/controller_test.go
```diff
@@ -3463,3 +3463,89 @@ func TestController_ClientDisconnect(t *testing.T) {
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			resp, err := clt.ReportUsageEventWithResponse(ctx, api.ReportUsageEventJSONRequestBody{
+				Class: tt.usageClass,
+				Name:  tt.usageName,
+				Count: tt.usageCount,
+			})
+			if err != nil {
+				t.Fatal(err, "cannot perform report usage event")
```
nopcoder, line 3530:
```suggestion
				t.Fatalf("ReportUsageEven failed: %s", err)
```

pkg/api/controller_test.go
```diff
@@ -3463,3 +3463,89 @@ func TestController_ClientDisconnect(t *testing.T) {
...
+			}
+
+			if resp.StatusCode() != tt.expectedStatusCode {
+				t.Fatal(err, "expected status code %d, but got %d", tt.expectedStatusCode, resp.StatusCode())
```
nopcoder, line 3534: (No `err`)
```suggestion
				t.Fatalf("ReportUsageEven status code: %d, expected: %d", resp.StatusCode(), tt.expectedStatusCode)
```

pkg/api/controller_test.go
```diff
@@ -3463,3 +3463,89 @@ func TestController_ClientDisconnect(t *testing.T) {
...
+			}
+
+			// For successful requests, verify the expected events were sent
+			if tt.expectedStatusCode == http.StatusNoContent {
```
nopcoder, line 3538: You can keep this expression result as boolean and match it to the lookup you are running.
So, when status code NoContent is returned we match found==true. When status code is not something else found == false.
This way we will verify we do not post usage in case of an error.

pkg/api/controller_test.go
```diff
@@ -3463,3 +3463,89 @@ func TestController_ClientDisconnect(t *testing.T) {
...
+				var eventsOfClass []*stats.Event
+				for _, ev := range deps.collector.Events {
+					if ev.Class == tt.usageClass {
```
nopcoder, line 3541: The loop should go ver the event and count the number (no need to collect them) of events that match the test class and name.
So after the loop the number will match the test count for sucessfull code or zero for none.
(When collector will support event reporting with count, it will be a single match lookup, so it will simplify the loop)

pkg/api/controller_test.go
```diff
@@ -3463,3 +3463,89 @@ func TestController_ClientDisconnect(t *testing.T) {
...
+						eventsOfClass = append(eventsOfClass, ev)
+					}
+				}
+				if len(eventsOfClass) != tt.usageCount {
+					t.Fatal(err, "wrong number of events sent, expected=%d but actual=%d", tt.usageCount, len(eventsOfClass))
```
nopcoder, line 3546: - No `err`
- Use Fatalf to format - Fatal does not format %d and etc.

nopcoder (changes requested):

nopcoder:
> closes #4860
> 
> This endpoint aims to simplify experiments and, when enabled, collect statistics about feature usage in their very first steps. even before a new API that provides a capability is defined.

The original issue with include implementation detail (path) should be updated too.

talSofer:
FYI @NoyDavidson2, we are adding a direct way to send statistics to allow experimenting with features at their very first stages. This means that we see new event classes and names internally.  I'll follow up with you on the first use we are planning for it

pkg/api/controller.go
arielshaqed: Sure.  Say you add a feature to find all Deltas in my repo and you can do it through the existing API.  It makes sense for the GUI to report _how many_ Deltas it found.  Why should the API make you check that you found any Deltas before you're allowed to report how many you've found?

arielshaqed (commented):
Just some examples :-)

Still approving, but agree with @nopcoder 's comments about adding count to the collector interface - it could load the server side.

pkg/api/controller.go
talSofer: done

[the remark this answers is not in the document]
talSofer: done

pkg/api/controller_test.go
talSofer: Changed this, but the collector interface I created looks a bit different than what (I think you imagined)

api/swagger.yml
```diff
@@ -3783,3 +3810,29 @@ paths:
           $ref: "#/components/responses/NotFound"
         default:
           $ref: "#/components/responses/ServerError"
+
+  /statistics:
+    post:
+      tags:
+        - statistics
+      operationId: sendStatsEvents
+      summary: send stats events
```
nopcoder, line 3819: If it is not too much - can we call it postStatsEvents - old convention: usually send means it will happen now and post means it will be. (will have to address it also in the collector API)
but I think it will make the caller more aware when to expect it.

pkg/api/controller.go
```diff
@@ -3635,6 +3636,51 @@ func (c *Controller) GetLakeFSVersion(w http.ResponseWriter, r *http.Request) {
...
+
+		if statsEv.Name == "" {
+			writeError(w, r, http.StatusBadRequest, "invalid value: name is required")
+			return
+		}
+
+		if statsEv.Count < 0 {
+			writeError(w, r, http.StatusBadRequest, fmt.Sprintf("invalid usage count=%d", statsEv.Count))
+			return
+		}
```
nopcoder, line 3661: Think we should loop and validate the data first.
In the current implementation it means that part of the data will be handled and than we get an error.

pkg/api/controller.go
```diff
@@ -3635,6 +3636,51 @@ func (c *Controller) GetLakeFSVersion(w http.ResponseWriter, r *http.Request) {
...
+
+		client := httputil.GetRequestLakeFSClient(r)
```
nopcoder, line 3663: pull out of the loop - required once

pkg/api/controller.go
```diff
@@ -3635,6 +3636,51 @@ func (c *Controller) GetLakeFSVersion(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+
+		if statsEv.Name == "" {
+			writeError(w, r, http.StatusBadRequest, "invalid value: name is required")
+			return
+		}
+
+		if statsEv.Count < 0 {
+			writeError(w, r, http.StatusBadRequest, fmt.Sprintf("invalid usage count=%d", statsEv.Count))
```
nopcoder, line 3659: no need to specify the value the user posted
```suggestion
			writeError(w, r, http.StatusBadRequest, "invalid value: count must be a positive integer starting from 1")
```

pkg/api/controller_test.go
```diff
@@ -3463,3 +3463,127 @@ func TestController_ClientDisconnect(t *testing.T) {
...
+			}
+
+			// For successful requests, verify the expected events were sent
+			if tt.expectedStatusCode == http.StatusNoContent {
+				for _, sentEv := range tt.events {
+					var eventsOfClass int
+					for _, collectedEv := range deps.collector.Events {
+						if collectedEv.Class == sentEv.Class {
+							eventsOfClass++
+						}
```
nopcoder, line 3580: we need to add the count that we found in the map

pkg/api/controller_test.go
```diff
@@ -3463,3 +3463,127 @@ func TestController_ClientDisconnect(t *testing.T) {
...
+			if err != nil {
+				t.Fatalf("SendStatsEvents failed: %s", err)
+			}
+
+			if resp.StatusCode() != tt.expectedStatusCode {
+				t.Fatalf("SendStatsEvents status code: %d, expected: %d", resp.StatusCode(), tt.expectedStatusCode)
+			}
+
+			// For successful requests, verify the expected events were sent
+			if tt.expectedStatusCode == http.StatusNoContent {
```
nopcoder, line 3574: I think we can remove the check to validate that we didn't find any event that match the class/name.
It will require `expectedCount` that we will match the "count" of event we found.

pkg/api/controller_test.go
```diff
@@ -3463,3 +3463,127 @@ func TestController_ClientDisconnect(t *testing.T) {
...
+				for _, sentEv := range tt.events {
+					var eventsOfClass int
+					for _, collectedEv := range deps.collector.Events {
+						if collectedEv.Class == sentEv.Class {
```
nopcoder, line 3578: I would match the class and name to verify we post the right event

pkg/stats/collector.go
```diff
@@ -26,6 +26,7 @@ const (
 
 type Collector interface {
 	CollectEvent(ev Event)
+	CollectEvents(ev Event, count uint64)
```
nopcoder, line 29: 1. Why not use `int`?
2. Can we just update `Event` to include `Count`?

pkg/stats/collector.go
```diff
@@ -217,8 +223,8 @@ func (s *BufferedCollector) getInstallationID() string {
 	return s.installationID
 }
 
-func (s *BufferedCollector) incr(k Event) {
-	s.cache[k]++
+func (s *BufferedCollector) incr(k EventWithCount) {
```
nopcoder, line 226: As I see it we cloud update `Event` to include count. But we added `EventWithCount` - Any reason?
In this case I'll keep `incr` current signature that accepts `Event` and increment just 1 and add `add` or `update` method that will accept `EventWithCount`.

pkg/api/serve_test.go
```diff
@@ -57,10 +57,16 @@ type memCollector struct {
...
 }
 
-func (m *memCollector) CollectEvent(ev stats.Event) {
+func (m *memCollector) CollectEvents(ev stats.Event, count uint64) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
-	m.Events = append(m.Events, &ev)
+	for i := 0; i < int(count); i++ {
+		m.Events = append(m.Events, &ev)
+	}
```
nopcoder, line 65: This is the loop we had in the controller - I think it should be handled the same way - remove it in favor of reporting / update the event's count.

nopcoder (changes requested):

pkg/api/controller_test.go
talSofer: done

pkg/api/controller_test.go
talSofer: removed the condition

pkg/api/controller_test.go
talSofer: done

pkg/stats/collector.go
talSofer: done

api/swagger.yml
talSofer: done

pkg/api/controller.go
```diff
@@ -3656,11 +3656,13 @@ func (c *Controller) SendStatsEvents(w http.ResponseWriter, r *http.Request, bod
 		}
 
 		if statsEv.Count < 0 {
-			writeError(w, r, http.StatusBadRequest, fmt.Sprintf("invalid usage count=%d", statsEv.Count))
+			writeError(w, r, http.StatusBadRequest, fmt.Sprintf("invalid value: count must be a positive integer"))
```
arielshaqed, line 3659:
```suggestion
			writeError(w, r, http.StatusBadRequest, fmt.Sprintf("invalid value: count %d must be a non-negative integer", statsEv.Count))
```

arielshaqed (approved):
Thanks!

Note some scary "check failure" annotations now appear on pkg/api/serve.go; not sure why.

nopcoder (changes requested):

pkg/api/controller.go
```diff
@@ -3635,6 +3636,53 @@ func (c *Controller) GetLakeFSVersion(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+
+		if statsEv.Name == "" {
+			writeError(w, r, http.StatusBadRequest, "invalid value: name is required")
+			return
+		}
+
+		if statsEv.Count < 0 {
+			writeError(w, r, http.StatusBadRequest, fmt.Sprint("invalid value: count must be a non-negative integer"))
```
nopcoder, line 3659: You need to remove the call to Sprint - you do not format anything - waste of call

pkg/api/controller_test.go
```diff
@@ -3463,3 +3463,132 @@ func TestController_ClientDisconnect(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			resp, err := clt.PostStatsEventsWithResponse(ctx, api.PostStatsEventsJSONRequestBody{
+				Events: tt.events,
+			})
+			if err != nil {
+				t.Fatalf("SendStatsEvents failed: %s", err)
+			}
+
+			if resp.StatusCode() != tt.expectedStatusCode {
+				t.Fatalf("SendStatsEvents status code: %d, expected: %d", resp.StatusCode(), tt.expectedStatusCode)
```
nopcoder, line 3578:
```suggestion
				t.Fatalf("PostStatsEvents status code: %d, expected: %d", resp.StatusCode(), tt.expectedStatusCode)
```

pkg/api/controller_test.go
```diff
@@ -3463,3 +3463,132 @@ func TestController_ClientDisconnect(t *testing.T) {
...
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			resp, err := clt.PostStatsEventsWithResponse(ctx, api.PostStatsEventsJSONRequestBody{
+				Events: tt.events,
+			})
+			if err != nil {
+				t.Fatalf("SendStatsEvents failed: %s", err)
```
nopcoder, line 3574:
```suggestion
				t.Fatalf("PostStatsEvents failed: %s", err)
```

pkg/api/controller_test.go
```diff
@@ -3463,3 +3463,132 @@ func TestController_ClientDisconnect(t *testing.T) {
...
+
+			for i, sentEv := range tt.events {
+				var collectedEventsCount int
+				for _, collectedMetric := range deps.collector.Metrics {
+					if collectedMetric.Event.Class == sentEv.Class && collectedMetric.Event.Name == sentEv.Name {
+						collectedEventsCount += int(collectedMetric.Value)
+					}
+				}
+				if collectedEventsCount != tt.expectedEventCounts[i] {
+					t.Fatalf("SendStatsEvents events for class %s count: %d, expected: %d", sentEv.Class, collectedEventsCount, sentEv.Count)
```
nopcoder, line 3589:
```suggestion
					t.Fatalf("PostStatsEvents events for class %s count: %d, expected: %d", sentEv.Class, collectedEventsCount, sentEv.Count)
```

pkg/api/controller_test.go
```diff
@@ -3463,3 +3463,132 @@ func TestController_ClientDisconnect(t *testing.T) {
...
+					Class: "single_event_count_gt_1",
+					Name:  "name",
+					Count: 3,
+				},
+			},
+			expectedEventCounts: []int{3},
+			expectedStatusCode:  http.StatusNoContent,
+		},
+		{
+			name: "multiple_events",
```
nopcoder, line 3502: Multiple events can hold the same class/name - we need to test it.
In order to test it I think you want to describe the expected events count as a map and not an array - we like to count the stats collector after posting the event and match the map that sums the total of each event type (class/name) to the expected count.

nopcoder (changes requested):

pkg/api/controller_test.go
talSofer: Added tests and modified structures accordingly

nopcoder (approved):
LGTM!

PR #4875, opened 2022-12-27
KV scan batch size control

### LINKED ISSUE #4712: KV listing with a limit parameter reads too many items
kv.Store.Scan does not take a limit parameter "hint".  So it reads some
fairly large number of items from the underlying store.  For DynamoDB KV
this is apparently whatever number of items it takes to fill 1 MiB; for
PostgreSQL this is 1000 items.

So whenever the listObjects API is called with a small limit it fetches many
more objects than it should.  A very typical use-case in lakeFSFS is to call
listObjects for just 1 object!  This triggers the worst case, and is slow.

nopcoder:
Close #4712

Pass listing in catalog level to the KV driver level.
Listing works with paginations, while KV with iterators.
In this PR the caller can pass recommended batch size to the KV level so the implementation can utilize our resources better and perform faster for specific actions.

### DIFF
--- a/pkg/graveler/graveler.go
+++ b/pkg/graveler/graveler.go
@@ -390,10 +390,10 @@ type KeyValueStore interface {
 	DeleteBatch(ctx context.Context, repository *RepositoryRecord, branchID BranchID, keys []Key) error
 
 	// List lists values on repository / ref
-	List(ctx context.Context, repository *RepositoryRecord, ref Ref) (ValueIterator, error)
+	List(ctx context.Context, repository *RepositoryRecord, ref Ref, batchSize int) (ValueIterator, error)
 
 	// ListStaging returns ValueIterator for branch staging area. Exposed to be used by catalog in PrepareGCUncommitted
-	ListStaging(ctx context.Context, branch *Branch) (ValueIterator, error)
+	ListStaging(ctx context.Context, branch *Branch, batchSize int) (ValueIterator, error)
 }
 
 type VersionController interface {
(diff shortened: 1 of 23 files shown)


pkg/graveler/graveler.go
```diff
@@ -1872,7 +1873,7 @@ func (g *Graveler) addCommitNoLock(ctx context.Context, repository *RepositoryRe
 }
 
 func (g *Graveler) isStagingEmpty(ctx context.Context, repository *RepositoryRecord, branch *Branch) (bool, error) {
-	itr, err := g.listStagingArea(ctx, branch)
+	itr, err := g.listStagingArea(ctx, branch, 0)
```
guy-har, line 1876:
```suggestion
	itr, err := g.listStagingArea(ctx, branch, 1)
```

pkg/graveler/graveler.go
```diff
@@ -1904,7 +1905,7 @@ func (g *Graveler) isSealedEmpty(ctx context.Context, repository *RepositoryReco
 	if len(branch.SealedTokens) == 0 {
 		return true, nil
 	}
-	itrs, err := g.sealedTokensIterator(ctx, branch)
+	itrs, err := g.sealedTokensIterator(ctx, branch, 0)
```
guy-har, line 1908:
```suggestion
	itrs, err := g.sealedTokensIterator(ctx, branch, 1)
```

pkg/graveler/staging/manager.go
```diff
@@ -94,8 +94,8 @@ func (m *Manager) DropKey(ctx context.Context, st graveler.StagingToken, key gra
 
 // List TODO niro: Remove batchSize parameter post KV
```
guy-har, line 95: Think we can remove this todo

pkg/kv/local/store.go
```diff
@@ -194,7 +194,25 @@ func (s *Store) Scan(ctx context.Context, partitionKey, start []byte) (kv.Entrie
...
+	prefix := partitionRange(partitionKey)
+	txn := s.db.NewTransaction(false)
+	opts := badger.DefaultIteratorOptions
+	opts.PrefetchSize = s.prefetchSize
+	if options.BatchSize > 0 && options.BatchSize < s.prefetchSize {
+		opts.PrefetchSize = options.BatchSize
+	}
+	if opts.PrefetchSize > 0 {
+		opts.PrefetchValues = true
+	}
```
guy-har, line 206: Nice catch

pkg/kv/dynamodb/store.go
```diff
@@ -356,6 +357,9 @@ func (s *Store) scanInternal(ctx context.Context, partitionKey, scanKey []byte,
 	if s.params.ScanLimit != 0 {
 		queryInput.SetLimit(s.params.ScanLimit)
 	}
+	if options.BatchSize != 0 {
+		queryInput.Limit = aws.Int64(int64(options.BatchSize))
+	}
```
guy-har, line 362: We should choose the minimum between ScanLimit and batchSize

guy-har (changes requested):
Great change, looks great.
Added some minor comments
Requesting changes only because that the DynamoDB scan doesn't choose the scan size correctly

pkg/kv/store.go
```diff
@@ -79,7 +84,9 @@ type Store interface {
 
 	// Scan returns entries that can be read by key order, starting at or after the `start` position
 	// partitionKey is optional, passing it might increase performance.
-	Scan(ctx context.Context, partitionKey, start []byte) (EntriesIterator, error)
+	// 'options' holds additional parameters to limit the number of records
+	//  and set the prefix to scan.
```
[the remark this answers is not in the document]
nopcoder: true, fixing the comment.

pkg/graveler/graveler.go
nopcoder: Thanks - good call to optimize these calls too.

pkg/graveler/graveler.go
nopcoder: Thanks - good call to optimize empty checks too

pkg/kv/dynamodb/store.go
nopcoder: Fixing

guy-har (approved):

PR #4878, opened 2022-12-27
Support MinIO

### LINKED ISSUE #4795: Discovery: GC committed for MinIO
Try out CGC with MinIO and estimate the effort of implementing it.

arielshaqed:
hard as fixing, so this PR does the latter.

Most of the code already works, of course.  But GC creates its own S3
client.  It is _not_ the same as the S3A client and these differences
were significant:

* S3 endpoint needs to be configured.  Otherwise, the client accesses
  the wrong "S3".
* For most MinIO installations the client needs to be configured with
  path style.  Otherwise it will attempt host style access by default
  and time out in some cases such as when running in Docker.

  Note that this cannot be configured on Hadoop versions before 2.8.0
  so users will need to use a suitable build to GC on Spark 2.4.7.  A
  mitigating factor is that this is _anyway_ needed to use Spark with
  MinIO -- regardless of GC or indeed lakeFS!

With these in place I was able to GC using the Spark 3.1.2 / Hadoop 3
build.

Fixes #4795.  I will split testing on MinIO in Esti off of that issue
for separate work.

### DIFF
--- a/clients/spark/core/src/main/hadoop2/scala/io/treeverse/clients/conditional/S3ClientBuilder.scala
+++ b/clients/spark/core/src/main/hadoop2/scala/io/treeverse/clients/conditional/S3ClientBuilder.scala
@@ -14,6 +14,7 @@ object S3ClientBuilder extends io.treeverse.clients.S3ClientBuilder {
     import org.apache.hadoop.fs.s3a.Constants
 
     val configuration = new ClientConfiguration().withMaxErrorRetry(numRetries)
+    val s3Endpoint = hc.get(Constants.ENDPOINT)
 
     val credentialsProvider =
       if (hc.get(Constants.ACCESS_KEY) != null) {
@@ -29,9 +30,13 @@ object S3ClientBuilder extends io.treeverse.clients.S3ClientBuilder {
         )
       } else None
 
+    val builder = AmazonS3ClientBuilder.standard()
+      .withPathStyleAccessEnabled(hc.getBoolean("fs.s3a.path.style.access", true))
+
     createAndValidateS3Client(configuration,
                               credentialsProvider,
-                              AmazonS3ClientBuilder.standard(),
+                              builder,
+                              s3Endpoint,
                               region,
                               bucket
                              )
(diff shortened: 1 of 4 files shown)


arielshaqed:
I am investigating test failures on non-Hadoop3.  This is strange, I would not expect AWS S3 to stop working as a result of unrelated changes :-/

clients/spark/core/src/main/hadoop2/scala/io/treeverse/clients/conditional/S3ClientBuilder.scala
```diff
@@ -29,9 +30,14 @@ object S3ClientBuilder extends io.treeverse.clients.S3ClientBuilder {
         )
       } else None
 
+    val builder = AmazonS3ClientBuilder
+      .standard()
+      .withPathStyleAccessEnabled(hc.getBoolean("fs.s3a.path.style.access", true))
```
Jonathan-Rosenberg, line 35: can you make `"fs.s3a.path.style.access"` a constant too? (to keep things unified)

clients/spark/core/src/main/scala/io/treeverse/clients/StorageUtils.scala
```diff
@@ -99,11 +102,20 @@ object StorageUtils {
...
         .withClientConfiguration(configuration)
-        .withRegion(region)
+
+      val builderWithEndpoint =
+        if (endpoint != null)
+          builder.withEndpointConfiguration(
+            new AwsClientBuilder.EndpointConfiguration(endpoint, region)
+          )
+        else
+          builder
```
Jonathan-Rosenberg, line 117:
```suggestion
          builder.withRegion(region)
```

Jonathan-Rosenberg (changes requested):
Awesomeeee
Consider adding the documentation change to this PR (change from hadoop-aws 2.7.7 to 2.8.0).
Also the comment from `StorageUtils.scala` to fix a misconfigured client.

arielshaqed (commented):
Thanks!

I want to update documentation separately, under #4913.

Jonathan-Rosenberg (approved):

PR #4880, opened 2022-12-27
GC: Parallelize metarange fetching

### LINKED ISSUE #4890: Parallelize metarange fetching in GC

johnnyaug:
Resolves #4890.

1. Re-partition commit dataset, so that fetching the ranges will be parallelized.
2. ~Make ranges distinct before getting all their addresses.~
3. ~Change flatMap, which seemed to cause parallelization problems, to mapPartitions + flatMap.~

After testing, it seems that (1) gives a significant performance gain (75% faster) on our E2E monitoring. The other ones did not improve speed in my setting - we will test these in the future in other settings.


arielshaqed (approved):
Neat speedup, thanks!

PR #4884, opened 2022-12-28
Code cleanup

nopcoder:
- remove KV from type that doesn't require it anymore
- file names
- delete unused code


eden-ohana (approved):
LGTM

PR #4886, opened 2022-12-28
API delete will not return object not found

### LINKED ISSUE #4798: Graveler's Delete race condition on branch retry can return invalid response
The issue is with returning 204 or 404 on DeleteObject.
The API requires us to return 204 when deleting an object, if that object exists.
A race between thread A that is deleting an object in staging and thread B that is committing. 

Scenario:

**Thread A:** Get staging token, find object it staging.
Succeed and pause right before checking whether the staging token has changed.

**Thread B:** Commit, clean staging.  Object is now deleted because the tombstone was set in staging.

**Thread A:** See that the staging token has changed, go again.  Get the staging token, find nothing in staging, find nothing in committed, return ErrNotFound - error code (404) to the caller.

**Result:** API returns 404 for a DeleteObject call when the object existed and was successfully deleted by the same call.

nopcoder:
**Change in API `delete object`**

On the following cases the API will not return NotFound error:

- when tombstone found on stage
- when object not found on branch commit

Fix #4798

### DIFF
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -3094,7 +3094,7 @@ paths:
       tags:
         - objects
       operationId: deleteObject
-      summary: delete object
+      summary: delete object. Missing objects will not return a NotFound error.
       responses:
         204:
           description: object deleted successfully
@@ -3121,7 +3121,7 @@ paths:
       tags:
         - objects
       operationId: deleteObjects
-      summary: delete objects
+      summary: delete objects. Missing objects will not return a NotFound error.
       requestBody:
         required: true
         content:
--- a/pkg/graveler/graveler.go
+++ b/pkg/graveler/graveler.go
@@ -1428,32 +1428,30 @@ func (g *Graveler) Set(ctx context.Context, repository *RepositoryRecord, branch
 func (g *Graveler) safeBranchWrite(ctx context.Context, log logging.Logger, repository *RepositoryRecord, branchID BranchID, stagingOperation func(branch *Branch) error) error {
 	var try int
 	for try = 0; try < BranchWriteMaxTries; try++ {
-		branch, err := g.GetBranch(ctx, repository, branchID)
+		branchPreOp, err := g.GetBranch(ctx, repository, branchID)
 		if err != nil {
 			return err
 		}
-		startToken := branch.StagingToken
-
-		if err = stagingOperation(branch); err != nil {
+		// startToken := branchPreOp.StagingToken
+		if err = stagingOperation(branchPreOp); err != nil {
 			return err
 		}
 
 		// Checking if the token has changed.
 		// If it changed, we need to write the changes to the branch's new staging token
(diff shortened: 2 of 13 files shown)


pkg/graveler/graveler.go
```diff
@@ -1428,32 +1428,30 @@ func (g *Graveler) Set(ctx context.Context, repository *RepositoryRecord, branch
...
-				WithField("startToken", startToken).
-				WithField("endToken", endToken).
-				Info("Retrying Set")
 		}
+		// we got a new token, try again
+		log.WithFields(logging.Fields{
+			"try":               try + 1,
+			"branch_token_pre":  branchPreOp.StagingToken,
+			"branch_token_post": branchPostOp.StagingToken,
+		}).Info("Retrying Set")
```
arielshaqed, line 1454: This can _and will_ happen a lot under load!  I think "INFO" is much too high, it is an operation that can repeatedly log but does not log per-request.  So I'd go with "DEBUG'.
```suggestion
		}).Debug("Retrying Set")
```

arielshaqed (changes requested):
Thanks!
This looks great, but I would like it to go much further!  I want us to remove the 404 response _entirely_.  Currently swagger.yml makes me think that I could synchronize multiple workers on 200-vs-404s, but in fact this is not true!  404 is not programmatically usable in the API.

nopcoder:
> Thanks! This looks great, but I would like it to go much further! I want us to remove the 404 response _entirely_. Currently swagger.yml makes me think that I could synchronize multiple workers on 200-vs-404s, but in fact this is not true! 404 is not programmatically usable in the API.

We require 404 for repository and branch not found

api/swagger.yml
```diff
@@ -3094,7 +3094,7 @@ paths:
       tags:
         - objects
       operationId: deleteObject
-      summary: delete object
+      summary: delete object. Missing objects will not return a NotFound error.
```
arielshaqed, line 3097:
```suggestion
      summary: delete object.
```
I believe that this belongs on the descriptions of 204 and of 404, and in a positive not a negative.

```yaml
        404:
          description: Repository or branch do not exist.
```

(The same for deleteObjects, of course!)

api/swagger.yml
```diff
@@ -3094,7 +3094,7 @@ paths:
...
       responses:
         204:
           description: object deleted successfully
```
arielshaqed, line 3100:
```suggestion
          description: Success: Now there is no object at key.  Also returned if the object did not exist in the first place.
```

arielshaqed (approved):
Thanks!

I think we can try to improve the API documentation, and made some suggestions.

In any case it's a really useful change: as someone coding against it, an API that precisely describes what it does is really important to me.

api/swagger.yml
nopcoder: Tried this one first. The description is part of the definition. Set it also on the status code - the code generators just ignore it.

PR #4887, opened 2022-12-28
Consistently shade everything as io.lakefs.<client>.shade.*

### LINKED ISSUE #3823: JVM assemblies: Pick a naming convention for shading and stick to it
We currently have a variety of different formats for shaded names in our JVM assemblies.  Additionally many of these names do not include any organizationally-unique prefix, meaning that they could collide with _other_ shaded names in Spark assemblies from _other_ organizations!

Select a single consistent prefix format that includes either "lakefs" or "treeverse" (suggestion: `io.lakefs.<client>.shaded`), and use that consistently.

arielshaqed:
Apply a single identical shading naming rule to clients: To shade a library
com.bar.baz in a client foo, use io.lakefs.foo.com.bar.baz.  This:

* Is easy to predict and clear to debug
* Guarantees uniqueness among other packages by using a prefix that we own
* Guarantees uniqueness among our clients by using the client name

Fixes #3823.

### DIFF
--- a/clients/hadoopfs/pom.xml
+++ b/clients/hadoopfs/pom.xml
@@ -148,27 +148,27 @@ To export to S3:
                                     <relocations>
                                         <relocation>
                                             <pattern>org.apache.httpcomponents</pattern>
-                                            <shadedPattern>io.lakefs.shaded.apache.httpcomponents</shadedPattern>
+                                            <shadedPattern>io.lakefs.hadoopfs.shade.org.apache.httpcomponents</shadedPattern>
                                         </relocation>
                                         <relocation>
                                             <pattern>okio</pattern>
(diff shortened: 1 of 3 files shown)


lynnro314 (approved):
LGTM, thanks!

arielshaqed:
Thanks!  Pulling...

Release note **FOR _both_ CLIENT RELEASES**

This release changes JVM package shading to be consistent.  There should be no
externally-visible effects: your code should never directly access any package
that has "shade" in its name.  Code that nonetheless does this might break.  A
good fix is to add dependencies directly: such code can break from any package
upgrade.  1292f83060239b851826d15ce1d9d070cd2d8099 changed the package shading
rules.
or preferably a

PR #4888, opened 2022-12-28
UI: added a blame button to path context menu

ozkatz:


webui/src/lib/components/repository/tree.jsx
```diff
@@ -154,6 +171,125 @@ const StatModal = ({ show, onHide, entry }) => {
 };
 
 
+const CommitMetadata = ({ metadata }) => {
+    const keys = Object.getOwnPropertyNames(metadata);
```
eladlachmi, line 175: Instead of this, you can cover the keys retrieval, length check, and `map` with `Object.entries`. It already filters out only own properties. [Ref](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/entries).

You can use it with `for..of` (like in the ref) or just map over the array of arrays.

eladlachmi (approved):
One small improvement suggestion. Not a blocker. A side from that, looks great.

ozkatz:
Thanks @eladlachmi - I've implemented your suggestions!

eladlachmi (approved):
LGTM

PR #4902, opened 2022-12-29
Delta Table Diff - MVP design

### LINKED ISSUE #4832: Delta Lake: Diff service design
Design a service that generates a diff of two given Delta lake table paths.

**Diff scope:**
* Two lines of operation history since the last common transaction (`txnid`).
* Three-way history diff (last common `txnid` as the ancestor).

Jonathan-Rosenberg:
Closes #4832


design/open/delta-diff.md
```diff
@@ -0,0 +1,160 @@
...
+
+### Authentication
+
+The `delta-rs` package generates an S3 client which will be used to communicate back to lakeFS (through the S3 gateway).  
+In order for the S3 client to communicate with lakeFS (or S3 in general) it needs to pass an AWS Access Key Id and Secret access key.  
+Since applicative credentials are not obligatory, the users that sent the request for a diff might not have such, and even if they have,
+they cannot send them through the GUI (which is the UI we chose to implement this feature for the MVP).
+To overcome this scenario, we'll use special diff credentials as follows:
+1. User makes a Delta Diff request.
+2. The DiffService checks if the user has "diff credentials" in the DB:
```
johnnyaug, line 121: This is essentially a session. Maybe we should implement a session mechanism as part of this feature?

design/open/delta-diff.md
```diff
@@ -0,0 +1,160 @@
...
+               "partitionBy":"[]"
+      },...
+  ]
+  ```
+
+---
+
+### Packaging
+
+1. We will package the binary with the lakeFS release.
```
johnnyaug, line 148: Consider publishing it separately, to avoid overloading the tar-ball.

johnnyaug (approved):
Love the plugin idea!

design/open/delta-diff.md
```diff
@@ -0,0 +1,168 @@
...
+---
+
+## Metrics
+
+### Applicative Metrics
+- Diff runtime
+- The total number of versions returned - this is the number of files that were read from lakeFS. This is basically the size of a Delta Log. We can use it later on to optimize reading.
+
+### Business Statistics
+- The number of unique requests - get the number of Delta Lake users.
```
talSofer, line 168: Did you mean "get the number of Delta Lake **diff** users."? 
What defines a unique request?

talSofer (approved):
Great writeup @Jonathan-Rosenberg, thank you!
LGTM, asked some clarification questions.

design/open/delta-diff.md
```diff
@@ -0,0 +1,168 @@
...
+
+The `delta-rs` package generates an S3 client which will be used to communicate back to lakeFS (through the S3 gateway).  
+In order for the S3 client to communicate with lakeFS (or S3 in general) it needs to pass an AWS Access Key Id and Secret access key.  
+Since applicative credentials are not obligatory, the users that sent the request for a diff might not have such, and even if they have,
+they cannot send them through the GUI (which is the UI we chose to implement this feature for the MVP).
+To overcome this scenario, we'll use special diff credentials as follows:
+1. User makes a Delta Diff request.
+2. The DiffService checks if the user has "diff credentials" in the DB:
+    1. If there are such credentials, it will use them.
+    2. If there aren't such, it will generate the credentials and save them: `{AKIA: DIFF-<>, SAK: <>}`. The `DIFF` prefix will be used to identify "diff credentials".
```
arielshaqed, line 128: 1. We should get someone from the auth team to go over this!
1. I would like us to declare this solution is explicitly temporary.  There are at least 2 auth requirements in here:
   * Support a note on credentials
   * Allow temporary credentials.
   * (Possibly...) allow attaching a policy with reduced permissions to such credentials.

design/open/delta-diff.md
```diff
@@ -0,0 +1,168 @@
...
+3. The DiffService will pass the credentials to the Delta Diff plugin during the call as environment variables (or part of the `gRPC` call).
```
arielshaqed, line 129:
```suggestion
3. The DiffService will pass the credentials to the Delta Diff plugin during the call as part of the `gRPC` call.
```

We _cannot_ use environment variables for this, the same plugin instance can be used for multiple calls.

design/open/delta-diff.md
```diff
@@ -0,0 +1,168 @@
...
+  [
+       {
+           "version": 1,
+           "timestamp":1515491537026,
+           "operation":"INSERT",
+           "operationContent":{
+               "operationParameters": {
+                  "mode":"Append",
+                  "partitionBy":"[]"
+                }
```
arielshaqed, line 146: Please document this schema.  If it's just the Delta schema, can you paste a link to it?

design/open/delta-diff.md
```diff
@@ -0,0 +1,168 @@
...
+1. We will package the binary with the lakeFS release.
+2. We will also update the Dockerfile to include the binary within it.
+
+---
+
+## Metrics
+
+### Applicative Metrics
+- Diff runtime
+- The total number of versions returned - this is the number of files that were read from lakeFS. This is basically the size of a Delta Log. We can use it later on to optimize reading.
```
arielshaqed, line 165: Doesn't compaction change this?

arielshaqed (commented):
Extremely neat stuff, thanks!

design/open/delta-diff.md
```diff
@@ -0,0 +1,168 @@
...
+She would like to debug the ETL run. In other words, she would want to see the Delta table history changes inserted applied in branch `exp1` since the commit that she branched from.
+
+---
+
+## Goals and scope
+
+1. For the MVP, support only Delta table diff.
+2. The system should be open for different data comparison implementations.
+3. The diff itself will consist of metadata changes only, in the form of the operation histories of the two tables, and the change in the number of rows.
+4. The diff will be a "three-dots" diff (like `git log branch1...branch2`). Basically showing the log changes that happened in one branch and not in the other.
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: I actually meant two-dots... Fixing

design/open/delta-diff.md
```diff
@@ -0,0 +1,168 @@
...
+5. UI: GUI only.
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: sounds good

design/open/delta-diff.md
```diff
@@ -0,0 +1,168 @@
...
+6. Reduce user friction as much as possible.
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: I guess that it's in the context of the other goals and non-goals specified.

design/open/delta-diff.md
```diff
@@ -0,0 +1,168 @@
...
+
+---
+
+## Non-Goals and off-scope
+
+1.  The Delta diff will be limited to the available Delta Log entries (the JSON files).
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: No, the support of log compaction is actually out of scope.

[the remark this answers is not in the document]
Jonathan-Rosenberg: 1. Adding.
2. I don't really think that it's a goal or a non-goal as it doesn't affect any implementation of the table diff...

design/open/delta-diff.md
```diff
@@ -0,0 +1,168 @@
...
+This is almost where we want to be, except that we'll need to somehow validate that the lakeFS server and the executable are using the same "API version" to communicate, so that the output of the binary would match the expected one from lakeFS. In addition, I would like to decrease the amount of deserialization needed to be implemented to interpret the returned DTO (maintainability and extensibility-wise).
+
+### Chosen architecture
+
+#### The Microkernel/Plugin Architecture
+
+The Microkernel/Plugin architecture is composed of two entities: a single "core system" and multiple "plugins".  
+In our case, the lakeFS server act as the core system, and the different diff implementations, including the Delta diff implementation, will act as plugins.  
+We'll use `gRPC` as the transport protocol, which makes the language of choice almost immaterial (due to the use of protobufs as the transferred data format)
+as long as it's self-contained (theoretically it can also be not system-runtime-dependent but then the cost will be an added requirement for running lakeFS- runtime support).
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: > * I need an external binary, built for the correct architecture.
> * I can simplify deployment by bundling the binary inside the lakeFS executable.
> * I need to serialize and deserialize, and can use protobufs if I like.

This is what I meant by "This is almost where we want to be" in the "diff as an external binary executable" section 🙂 

> * I need some way of sharing the schema (JSON schema, Swagger spec, gRPC Protobuf definition, whatever) between the trees.
>  +We will need to figure out how to share protobuf sources between 2 separate trees

This can be solved by adding the plugin code to lakeFS (the same way we add different KV implementations to lakeFS, but possibly using other languages).

The thing I like about the `go-plugin` library is that it's a very lightweight and maintained library that is used in production for quite some time, offers API protocol validation (although a basic one), wraps the gRPC communication, and attaches the logging of the executable for us.

> we will have to write a gRPC server entirely on our own.

True, I actually did this and it's pretty easy. I guess that in both approaches we'll have to implement some kind of entry point to the "plugin". I personally prefer that it would follow some kind of protocol and typed data, but it's arguable (obviously).

design/open/delta-diff.md
```diff
@@ -0,0 +1,168 @@
...
+## Implementation
+
+![Delta Diff flow](diagrams/delta-diff-flow.png)
+[(excalidraw file)](diagrams/delta-diff-flow.excalidraw)
+
+#### DiffService
+
+The DiffService will be an internal component in lakeFS which will serve as the Core system.  
+In order to realize which diff plugins are available, we shall use new configuration values:
+1. `LAKEFS_PLUGINS_LOCATION`
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: What if there would be additional plugins?

design/open/delta-diff.md
```diff
@@ -0,0 +1,168 @@
...
+   The DiffService will use this location to search for the required plugin. Defaults to lakeFS's binary directory.
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: > running from relative to the configuration file (because we do know where that it!)

That's what I meant. I'll clarify that.
Thanks!

design/open/delta-diff.md
```diff
@@ -0,0 +1,168 @@
...
+2. `LAKEFS_PLUGIN_DIFF_{TYPE OF DIFF}`
+   The DiffService will use the name of binary provided by this env var to perform the diff. For instance, `LAKEFS_PLUGIN_DIFF_DELTA=deltaDiffBinary`
+   The DiffService will use the [plugins' location](https://github.com/hashicorp/terraform/blob/main/plugins.go) and the diff binary to load the plugin and request a diff.
+   The **type** of diff will be sent as part of the request to lakeFS as specified [here](#API).
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: Do you mean adding another endpoint for diff-location validation?

design/open/delta-diff.md
```diff
@@ -0,0 +1,168 @@
...
+   The communication between the DiffService and the plugins, as explained above, will be through the `go-plugin` package (`gRPC`).
+
+#### Delta Diff Plugin
+
+Implemented using [delta-rs](https://github.com/delta-io/delta-rs) (Rust), this plugin will perform the diff operation using table paths provided by the DiffService through a `gRPC` call.  
+To query the Delta Table from lakeFS, the plugin will generate an S3 client (this is a constraint imposed by the `delta-rs` package) and send a request to lakeFS's S3 gateway.
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: The `delta-rs` package doesn't allow you to configure your own client, rather it supports a few configured clients for S3, GCP, and Azure, thus we cannot configure a client like lakeFSFS that directly communicates with the underlying store.

The data that will flow through the lakeFS server will be only delta log files (files under `_delta_log`), and specifically, only the Delta Log entries, i.e. the Delta log JSON version files.

Would you like me to document the above?

design/open/delta-diff.md
```diff
@@ -0,0 +1,168 @@
...
+   The **type** of diff will be sent as part of the request to lakeFS as specified [here](#API).
+   The communication between the DiffService and the plugins, as explained above, will be through the `go-plugin` package (`gRPC`).
+
+#### Delta Diff Plugin
+
+Implemented using [delta-rs](https://github.com/delta-io/delta-rs) (Rust), this plugin will perform the diff operation using table paths provided by the DiffService through a `gRPC` call.  
+To query the Delta Table from lakeFS, the plugin will generate an S3 client (this is a constraint imposed by the `delta-rs` package) and send a request to lakeFS's S3 gateway.  
+The diff algorithm:
+1. Run the Delta [HISTORY command](https://docs.delta.io/latest/delta-utility.html#history-schema) on both table paths.
+2. Traverse through the [returned "commitInfo" entry vector ](https://github.com/delta-io/delta-rs/blob/main/rust/src/delta.rs#L888)
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: Not to self: don't count on a line number under a branch-namespace (as opposed to a commit namespace)

Thanks!
fixing

design/open/delta-diff.md
```diff
@@ -0,0 +1,168 @@
...
+starting from the **last** version of each history vector: 
+    1. While the returned entry versions **aren't** equal:
+        1. If the bigger version is the _"left"'s_ version, add the entry to the returned history list.
+        2. Traverse one version back of the bigger version entry's history.
+    2. While the versions > 0: 
+        1. Create a hash for the entry based on fields: `timestamp`, `operation`, `operationParameters` , and `operationMetrics` values.
+        2. Compare the hashes of the versions.
+        3. If they **aren't equal**, add the "left"'s entry to the returned history list, else break and **return the history vector**.
+        4. Traverse one version back in both vectors.
+3. Return an empty history vector.
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: Right.

design/open/delta-diff.md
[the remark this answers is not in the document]
Jonathan-Rosenberg: No,
These are applicative credentials just as we have now.
The difference is the `DIFF` prefix.

Jonathan-Rosenberg: 1. This solution involved @guy-har. Adding as a reviewer...
2. Absolutely.

design/open/delta-diff.md
Jonathan-Rosenberg: compaction == deleted history -> not part of the history command

design/open/delta-diff.md
Jonathan-Rosenberg: Both.
I would like to understand how many of our users use Delta Lake. I guess that not all of our Delta users will use the Delta Diff, but it might give us a better idea than the tools we currently have.
A unique request is defined by an installation id.

design/open/delta-diff.md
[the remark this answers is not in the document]
talSofer: Thanks! 
IMO it worth documenting this answer 
>The data that will flow through the lakeFS server will be only delta log files (files under _delta_log), and specifically, only the Delta Log entries, i.e. the Delta log JSON version files.

design/open/delta-diff.md
talSofer: Thanks, then consider using - 
```suggestion
- The number of unique requests by installation id - get the lower bound of the number of Delta Lake users.
```

design/open/delta-diff.md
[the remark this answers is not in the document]
itaiad200: It's better if we use the same prefix for all configuration values related to plugins.

design/open/delta-diff.md
```diff
@@ -0,0 +1,186 @@
...
+[(excalidraw file)](diagrams/microkernel-overview.excalidraw)
+
+---
+
+## Implementation
+
+![Delta Diff flow](diagrams/delta-diff-flow.png)
+[(excalidraw file)](diagrams/delta-diff-flow.excalidraw)
+
+#### DiffService
```
itaiad200, line 90: How does one configures the plugins, i.e. passes configuration values?

design/open/delta-diff.md
```diff
@@ -0,0 +1,186 @@
...
+   The DiffService will use the name of binary provided by this env var to perform the diff. For instance, `LAKEFS_PLUGIN_DIFF_DELTA=deltaDiffBinary`
+   The DiffService will use the [plugins' location](https://github.com/hashicorp/terraform/blob/main/plugins.go) and the diff binary to load the plugin and request a diff.
+   The **type** of diff will be sent as part of the request to lakeFS as specified [here](#API).
+   The communication between the DiffService and the plugins, as explained above, will be through the `go-plugin` package (`gRPC`).
+
+#### Delta Diff Plugin
+
+Implemented using [delta-rs](https://github.com/delta-io/delta-rs) (Rust), this plugin will perform the diff operation using table paths provided by the DiffService through a `gRPC` call.  
+To query the Delta Table from lakeFS, the plugin will generate an S3 client (this is a constraint imposed by the `delta-rs` package) and send a request to lakeFS's S3 gateway.  
+The diff algorithm:
```
itaiad200, line 106: What is the expected amount of calls to the gateway? How much data is to be retrieved from the object store? Two things to consider:
1. IIUC we have a request waiting in the background, we should be fast..
2. How we scale.. If users perform several concurrent requests, will we choke?

design/open/delta-diff.md
```diff
@@ -0,0 +1,186 @@
...
+To overcome this scenario, we'll use special diff credentials as follows:
+1. User makes a Delta Diff request.
+2. The DiffService checks if the user has "diff credentials" in the DB:
+    1. If there are such credentials, it will use them.
+    2. If there aren't such, it will generate the credentials and save them: `{AKIA: DIFF-<>, SAK: <>}`. The `DIFF` prefix will be used to identify "diff credentials".
+3. The DiffService will pass the credentials to the Delta Diff plugin during the call as part of the `gRPC` call.
+
+### API
+
+- GET `/repositories/repo/{repo}/otf/refs/{left_ref}/diff/{right_ref}?type={diff_type}`
```
itaiad200, line 135: What about pagination/limitations?

itaiad200 (commented):
Thanks, looks promising and well thought though! Reviewing from phone so apologies if I'm reiterating on answered questions

design/open/delta-diff.md
Jonathan-Rosenberg: As we don't really know the amount of Delta Log files used on average (or median, or max) the amount of data is currently an unknown. There will be two calls to the S3 gateway- one history command for each table.
One of the metrics specified below is getting the number of files That Delta users use.
Regarding point 1, for the MVP, this process can be slow (the whole thing is tagged as experimental), but it will help us evaluate the concurrency and resources needed for the Diff.
Regarding point 2, scaling the requests from the Diff service to the diff plugin is pretty easy as it uses an HTTP/2 connection to communicate (we can run multiple streams on a single connection). If a user makes multiple requests for a diff we can implement a backoff in the plugin itself.

design/open/delta-diff.md
```diff
@@ -0,0 +1,186 @@
...
+
+## Implementation
+
+![Delta Diff flow](diagrams/delta-diff-flow.png)
+[(excalidraw file)](diagrams/delta-diff-flow.excalidraw)
+
+#### DiffService
+
+The DiffService will be an internal component in lakeFS which will serve as the Core system.  
+In order to realize which diff plugins are available, we shall use new configuration values:
```
nopcoder, line 93: - There is no manifest / configuration to map which plugins to load? from reading this section we have location and the specific delta diff information. Suggest you define the plugin configuration without sharing the location.
  Example:
  ```yaml
  plugins:
    diff:
      delta: <location to the delta diff plugin - full path to the binary needed to execute>    
  ```
- In case we like to map multiple plugins to the same format we should consider separating the plugins configuration and the use cases. Like:
  ```yaml
  diff:
    otf:
      delta:
        plugin: diff_delta_v2
  plugins:
    diff_delta_v1: <location to the delta diff plugin - full path to the binary needed to execute>    
    diff_delta_v2: <location to the delta diff plugin - full path to the binary needed to execute>    
  ```
- Another consideration - if we plan to use the plugin later also to support advance merge - we can just drop the `diff` from all the above examples.
- In case we would like to leverage more plugins for different use-cases (ex: merge) will it possible to share the same location for different plugin interfaces or we need to split / catalog them now?

design/open/delta-diff.md
```diff
@@ -0,0 +1,186 @@
...
+   The DiffService will use this location to search for the required plugin. Defaults to lakeFS's configuration file directory.
+2. `LAKEFS_PLUGIN_DIFF_{TYPE OF DIFF}`
+   The DiffService will use the name of binary provided by this env var to perform the diff. For instance, `LAKEFS_PLUGIN_DIFF_DELTA=deltaDiffBinary`
+   The DiffService will use the [plugins' location](https://github.com/hashicorp/terraform/blob/main/plugins.go) and the diff binary to load the plugin and request a diff.
+   The **type** of diff will be sent as part of the request to lakeFS as specified [here](#API).
+   The communication between the DiffService and the plugins, as explained above, will be through the `go-plugin` package (`gRPC`).
+
+#### Delta Diff Plugin
+
+Implemented using [delta-rs](https://github.com/delta-io/delta-rs) (Rust), this plugin will perform the diff operation using table paths provided by the DiffService through a `gRPC` call.
```
nopcoder, line 104: - We should specify the constraint of fetching large objects from delta-rs. Configurable (optional) location to store the temporary data downloaded by the plugin - will probably be required in case we don't want to use the default temporary folder or control/strict the size used by the plugin.
- Cleanup or this storage if it is not managed by the OS temp dir/files API.
- Calling the plugin with 'deadline' (timeout for the grpc call) just to prevent a lock during the call to the plugin.

design/open/delta-diff.md
nopcoder: Generate for missing is good short term solution - it can also be an issue when two services will try to generate while currently we don't lock this option means we will have two sets of credentials.

design/open/delta-diff.md
nopcoder: - Thought we need to provide a path to the table - there are two refs here
- The location of `otf` is after the repository - what does it represents in this level? is it just a different `diff` operation?

design/open/delta-diff.md
```diff
@@ -0,0 +1,186 @@
...
+        - description: the version/snapshot/transaction id of the operation.
+      - timestamp (epoch):
+        - type: long
+        - description: operation's timestamp.
+      - operation:
+        - type: string
+        - description: operation's name.
+      - operationContent:
+        - type: map
+        - description: an operation content specific to the table format implemented.
```
nopcoder, line 151: Just verify - what types the map will hold? we need to serialize any json object?

design/open/delta-diff.md
```diff
@@ -0,0 +1,186 @@
...
+2. The DiffService checks if the user has "diff credentials" in the DB:
+    1. If there are such credentials, it will use them.
+    2. If there aren't such, it will generate the credentials and save them: `{AKIA: DIFF-<>, SAK: <>}`. The `DIFF` prefix will be used to identify "diff credentials".
+3. The DiffService will pass the credentials to the Delta Diff plugin during the call as part of the `gRPC` call.
+
+### API
+
+- GET `/repositories/repo/{repo}/otf/refs/{left_ref}/diff/{right_ref}?type={diff_type}`
+    - Tagged as experimental
+    - **Response**:
```
nopcoder, line 137: The response format render log of operations - maybe we can scope it at this level - so if the next change will return a response with diff - we can just return a empty list/log here.

design/open/delta-diff.md
```diff
@@ -0,0 +1,186 @@
...
+                      "partitionBy":"[]"
+                    }
+          },
+          ...
+      ]
+      ```
+    
+---
+
+### Packaging
```
nopcoder, line 172: **Build* and packaging - we need to address the build part too

nopcoder (commented):
Looks good!  Added some comments and concerns.

design/open/delta-diff.md
itaiad200: Not sure I understand, is "history" a gateway command or a delta one? I'm not familiar with that gw endpoint so I'm guessing it's a delta command that executes s3 requests to the gw. I suggest we have a rough estimate for the number and type of calls now. I'm sure we'll be able to handle it but it's better to understand the costs sooner rather than later. 
By scale I meant the number of calls to the gateway. If a single diff performs 100 calls that fetches data, then 10 diffs performance is something to consider. I agree that the service-plugin communication scale is not a concern at this point.

Jonathan-Rosenberg: Gotcha.
So the **delta** history command is quite an expensive command- it will send a request to fetch each delta log (JSON) file one by one from the gateway. That means that if there are N log entries, then N requests will get sent to the gateway. 
This is surely not the way we would want our diff to eventually work, but at this point (for the case of an MVP, and to verify that a diff is even needed) I don't think we should work towards making it efficient.

design/open/delta-diff.md
```diff
@@ -0,0 +1,186 @@
+# Delta Table Differ
+
+## User Story
+
+Jenny the data engineer runs a new retention-and-column-remover ETL over multiple Delta Lake tables. To test the ETL before running on production, she uses lakeFS (obviously) and branches out to a dedicated branch `exp1`, and runs the ETL pointed at `exp1`.
+The output is not what Jenny planned... The Delta Table is missing multiple columns and the number of rows just got bigger!
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: yes, this is optional (we know we are getting it back as part of the metadata)

design/open/delta-diff.md
```diff
@@ -0,0 +1,186 @@
...
+* The plugin system works by launching subprocesses and communicating over RPC (both `net/rpc` and `gRPC` are supported).
+  A single connection is made between any plugin and the core process. For gRPC-based plugins, the HTTP2 protocol handles [connection multiplexing](https://freecontent.manning.com/animation-http-1-1-vs-http-2-vs-http-2-with-push/).
+* The plugin and core system are separate processes, which means that a crash of a plugin won't cause the core system to crash (lakeFS in our case).
+
+![Microkernel architecture overview](diagrams/microkernel-overview.png)
+[(excalidraw file)](diagrams/microkernel-overview.excalidraw)
+
+---
+
+## Implementation
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: 1. Ok
2. Not sure I understand. What I meant in the diagram is that the request will include the type of diff that the user wishes to see. If it's a diff of CSV data, the user will send `type=csv` in the request, and the DiffService will realize that the CSV diff plugin is needed to create the diff.

design/open/delta-diff.md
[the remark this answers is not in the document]
Jonathan-Rosenberg: The DiffService is not the plugin itself, it's the component within lakeFS that interacts with the plugins.
What about `PlugableDiffService`?

design/open/delta-diff.md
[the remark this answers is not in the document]
arielshaqed: True.  So everything below should also be `LAKEFS_PLUGINS_LOCATIONS`.  That means that everything lives inside the same configuration object.

Also as specificed this might not really be a plugin: I don't know whether Viper will let us configure an "any object" and be able to pass it usefully to the plugin during load.

If this turns out to be impossible, I would suggest configuring a separate "plugins file"; that one can be a general YAML/JSON object with very few parsed fields, and then each plugin gets its configuration subobject from that file (and presumably parses it using mapstructure in Go, or libwhatever in other languages).

design/open/delta-diff.md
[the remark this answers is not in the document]
arielshaqed: Yes!

design/open/delta-diff.md
Jonathan-Rosenberg:
```yaml
diff:
  otf:
    delta:
      plugin: diff_delta_v2
plugins:
  diff_delta_v1: <location to the delta diff plugin - full path to the binary needed to execute>    
  diff_delta_v2: <location to the delta diff plugin - full path to the binary needed to execute>    
```
Really liked this one, although I would still want to have default values so that if the user didn't specify a plugin location in the manifest, lakeFS will try to figure it out using the latest known version of the plugin and try to locate it in some default path. For example, if the user didn't specify the location for `diff.delta.plugin`, lakeFS will try to locate the binary under `~/.lakefs/plugins`.

> * Another consideration - if we plan to use the plugin later also to support advance merge - we can just drop the diff from all the above examples.
> * In case we would like to leverage more plugins for different use-cases (ex: merge) will it possible to share the same location for different plugin interfaces or we need to split / catalog them now?

I think that it's a very possible scenario, yet I think that it might be preferable that each plugin would know how to do one thing. Anyway, we can decide upon it when we get there...

design/open/delta-diff.md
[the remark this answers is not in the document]
arielshaqed: Is this a ***future*** advantage to contributing our own client to Delta?

design/open/delta-diff.md
arielshaqed: +1 :+1: to measure the number of calls a "simple" history performs.

arielshaqed (commented):
T-h-a-n-k-s-!

design/open/delta-diff.md
Jonathan-Rosenberg: The [second applicative metric](https://github.com/treeverse/lakeFS/pull/4902/files#diff-e66e2e9fcb8162bd781db293307eed89340003eac894d788a44ce24341f64018R183) will basically do that.

design/open/delta-diff.md
Jonathan-Rosenberg: Ok,
I'll mark the temporary solutions (for the MVP) all around the document.

design/open/delta-diff.md
Jonathan-Rosenberg: @itaiad200 I'll add a description to your points.

@nopcoder you're right, it should be `otf/table/{left_table_path}/diff/{right_table_path}?type={diff_type}`
The `otf` is to specify that this API is strictly for OTFs

design/open/delta-diff.md
Jonathan-Rosenberg: Through the request to the plugin (think of Terraform providers as an example) and [this](https://github.com/treeverse/lakeFS/pull/4902/files#r1061355569)

N-o-Z (approved):
Love the design - looks well thought through.
I don't have much to add to the already existing comments.
One thing though, maybe you can reduce the images sizes? For example the delta-diff-flow image can't be read from the document and when opening it in a different window it pops up very big making it hard to see the entire flow

design/open/delta-diff.md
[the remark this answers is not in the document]
nopcoder: about item 2 - I mean that a single plugin can serve multiple formats - so the mapping of the plugin system may look different. As we load a single set of plugins and each plugin publish the formats it supports or in the the configuration level map each plugin to one or more formats.

design/open/delta-diff.md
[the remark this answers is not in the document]
nopcoder: ok, got it - just didn't want to couple the plugin system to the DiffService.  I would like that any service in lakeFS and access the plugins. This way we can implement merge and future capabilities as needed without adding another plugin system.

nopcoder (changes requested):
Left couple of comments - the concern I left was related to plugin system we are adding - I think it should serve more than just the diff if possible. I assume the next service will be called MergeService and it may use the same or different set of plugins and I think it is a place to use a single system that can serve different functionality.

design/open/delta-diff.md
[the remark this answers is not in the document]
Jonathan-Rosenberg: I think that this step should be decided by the plugin "port", e.g. "DiffService". 
If plugin `A` can serve multiple formats, then the configurations might look like:
```yaml
diff:
  delta:
    plugin: A
  csv:
    plugin: A
  iceberg:
    plugin: A
merge:
  delta:
    plugin: A
plugins:
  A: ~/.lakefs/plugins/A
```

At this point, every core (port) component can choose according to the specified format of the plugin it needs. And a plugin can be used multiple times for different formats.

Jonathan-Rosenberg:
@nopcoder your concern is a very valid one.
I think that this design still answers it.
Plugins can be developed and used in a variety of ways with different services (Merge, Diff, Rebase...) by setting the correct configuration file fields:
```yaml
diff: // service/functionality that uses the plugins
  delta:
    plugin: A
  csv:
    plugin: A
  iceberg:
    plugin: A
merge: // service/functionality that uses the plugins
  delta:
    plugin: A
rebase: // service/functionality that uses the plugins
  parquet:
    plugin: B
plugins:
  A: ~/.lakefs/plugins/A
  B: ~/.lakefs/plugins/B
```

design/open/delta-diff.md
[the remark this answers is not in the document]
arielshaqed: Long-term we **will** need and have a generic external service interface: think of a user who wants to add or modify a diff/merge service.  I believe they will prefer a looser coupling.
Short- and medium-term I agree with you that using go-plugin makes sense _if_ we believe that it shortens initial development time.

Counterpoint: If we implement the plugin service as a separate REST microservice, can we drive it directly from the GUI, and then we don't need to add anything to lakeFS (server)?  Might skip having to agree with team VE about configuration, adding go-plugin, packaging, etc.

design/open/delta-diff.md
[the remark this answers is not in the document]
arielshaqed: A bigger point here is that we will need some way for lakeFS to detect what type it should offer and/or request.  But this is in no way blocking for the MVP, and I am perfectly happy to skip it for now!  (For instance, we **might** discover that users _want_ a diff but _don't want_ to initiate it from lakeFS.)

arielshaqed (approved):
Thanks!

nopcoder (approved):
Thanks

PR #4907, opened 2022-12-29
fix: ingest range content type is wrong

### LINKED ISSUE #4906: When using `ingest-range`, Content-Type is populated with objects' physical address
When running an Import, the resulting entries end up having a physical address URI as their `Content-Type`. This is improper use of this field and might cause data consumers to wrongfully negotiate the type the returned object.

Example:

image

the address appears under physical address as well as Content-Type.

ozkatz:
Removed the use of the physical address as the content type, as it's not a valid content type.
Since the walker function has no access to content types (nor does S3 when listing objects, for example), I believe it's better to leave the field empty than put something invalid.

Closes #4906


N-o-Z (approved):
LGTM

PR #4910, opened 2023-01-01
Configurable repository and commit cache

### LINKED ISSUE #4637: Allow configuring the commit cache parameters: size, timing
The default commit cache is quite small (1000 entries).  On large
repositories this significantly slows down FindMergeBase operations for
computing diffs and merges.  The 30-second expiry is also much too short:
this cache holds immutable data, it is not clear that it is _ever_ justified
to expired entries from it.  Allow configuring this cache.

nopcoder:
Close #4637

- Configurable cache size for repository and commit information.
- Increase the size of commit cache from 1k to 50k items.
- Increase the expiry of commit cache from 5sec to 10min.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -451,7 +451,7 @@ func (c *Config) GetLoginDuration() time.Duration {
 	return c.values.Auth.LoginDuration
 }
 
-func (c *Config) GravelerRepositoryCacheConfig() ref.CacheConfig {
+func (c *Config) GetGravelerRepositoryCacheConfig() ref.CacheConfig {
 	return ref.CacheConfig{
 		Size:   c.values.Graveler.RepositoryCache.Size,
 		Expiry: c.values.Graveler.RepositoryCache.Expiry,
@@ -459,7 +459,7 @@ func (c *Config) GravelerRepositoryCacheConfig() ref.CacheConfig {
 	}
 }
 
-func (c *Config) GravelerCommitCacheConfig() ref.CacheConfig {
+func (c *Config) GetGravelerCommitCacheConfig() ref.CacheConfig {
 	return ref.CacheConfig{
 		Size:   c.values.Graveler.CommitCache.Size,
 		Expiry: c.values.Graveler.CommitCache.Expiry,
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -21,6 +21,7 @@ import (
 	"github.com/treeverse/lakefs/pkg/block"
 	blockparams "github.com/treeverse/lakefs/pkg/block/params"
 	"github.com/treeverse/lakefs/pkg/graveler/committed"
+	"github.com/treeverse/lakefs/pkg/graveler/ref"
 	kvparams "github.com/treeverse/lakefs/pkg/kv/params"
 	"github.com/treeverse/lakefs/pkg/logging"
 	pyramidparams "github.com/treeverse/lakefs/pkg/pyramid/params"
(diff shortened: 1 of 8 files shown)


pkg/config/config.go
```diff
@@ -449,3 +450,19 @@ func (c *Config) GetUIEnabled() bool {
 func (c *Config) GetLoginDuration() time.Duration {
 	return c.values.Auth.LoginDuration
 }
+
+func (c *Config) GravelerRepositoryCacheConfig() ref.CacheConfig {
```
arielshaqed, line 454:
```suggestion
func (c *Config) GetGravelerRepositoryCacheParams() ref.CacheConfig {
```
for consistency, e.g. like GetBlockAdapterS3Params.  Also below?

pkg/config/defaults.go
```diff
@@ -159,6 +159,19 @@ const (
...
 	UIEnabledKey = "ui.enabled"
+
+	GravelerRepositoryCacheSizeKey       = "graveler.repository_cache.size"
+	DefaultGravelerRepositoryCacheSize   = 1000
+	GravelerRepositoryCacheExpiryKey     = "graveler.repository_cache.expiry"
+	DefaultGravelerRepositoryCacheExpiry = 5 * time.Second
+	GravelerRepositoryCacheJitterKey     = "graveler.repository_cache.jitter"
+	DefaultGravelerRepositoryCacheJitter = DefaultGravelerRepositoryCacheExpiry / 2
+	GravelerCommitCacheSizeKey           = "graveler.commit_cache.size"
+	DefaultGravelerCommitCacheSize       = 50000
```
arielshaqed, line 170: Nit:
```suggestion
	DefaultGravelerCommitCacheSize       = 50_000
```
:-)

pkg/config/defaults.go
```diff
@@ -159,6 +159,19 @@ const (
...
+	GravelerCommitCacheExpiryKey         = "graveler.commit_cache.expiry"
+	DefaultGravelerCommitCacheExpiry     = 30 * time.Second
+	GravelerCommitCacheJitterKey         = "graveler.commit_cache.jitter"
+	DefaultGravelerCommitCacheJitter     = DefaultGravelerCommitCacheExpiry / 2
```
arielshaqed, line 174: This seems very high, we want a long expiry time.  And if it's inconsistent, FindMergeBase may take very different times to do the exact same thing.
```suggestion
	DefaultGravelerCommitCacheJitter     = 2 * time.Second
```

pkg/config/defaults.go
```diff
@@ -159,6 +159,19 @@ const (
...
+	GravelerRepositoryCacheSizeKey       = "graveler.repository_cache.size"
+	DefaultGravelerRepositoryCacheSize   = 1000
+	GravelerRepositoryCacheExpiryKey     = "graveler.repository_cache.expiry"
+	DefaultGravelerRepositoryCacheExpiry = 5 * time.Second
+	GravelerRepositoryCacheJitterKey     = "graveler.repository_cache.jitter"
+	DefaultGravelerRepositoryCacheJitter = DefaultGravelerRepositoryCacheExpiry / 2
+	GravelerCommitCacheSizeKey           = "graveler.commit_cache.size"
+	DefaultGravelerCommitCacheSize       = 50000
+	GravelerCommitCacheExpiryKey         = "graveler.commit_cache.expiry"
+	DefaultGravelerCommitCacheExpiry     = 30 * time.Second
```
arielshaqed, line 172: Nice... but even this seems low!  What's the harm in keeping something in the commit cache?  Even 5 minutes seems reasonable.

arielshaqed (approved):
THANKS!

pkg/config/defaults.go
nopcoder: Agree - I also not sure jitter is needed for these type of cache

pkg/config/defaults.go
nopcoder: No harm - updated the size and forgot to update the expiry. Setting to 10min - comment if you think we should go higher.

pkg/config/config.go
nopcoder: Thanks - I'll address the new/old practice in a different pr

PR #4911, opened 2023-01-01
Proposal to simplify config pkg

nopcoder:
Simplify working with configuration variables by embeding the pkg config's
template structure into config enabling public access to all members, as
alternative to the way we address it today by implementing getting methods.

Use https://go.dev/ref/spec#Conversions to copy configuration structure
to service specific parameters when possible.

Currently some methods are kept not to break existing contracts

Looking forward in future pulls:
- Simplify some block adapter initialization and copy configuration values to the package to initialize
- Define and copy configuration values for catalog package instead of passing the complete config


pkg/config/defaults.go
```diff
@@ -247,6 +247,7 @@ func setDefaults(local bool) {
 
 	viper.SetDefault(DatabaseLocalPathKey, DefaultDatabaseLocalPath)
 	viper.SetDefault(DatabaseLocalPrefetchSizeKey, DefaultDatabaseLocalPrefetchSize)
+	viper.SetDefault("database.local.sync_writes", true)
```
johnnyaug, line 250: Related?

johnnyaug (approved):
Awesome! See one comment.

pkg/config/defaults.go
nopcoder: Yes - config.go's GetKVParams as normalize this to true by treat it as a pointer.
I'll update the code so it will look like all other defaults.

PR #4914, opened 2023-01-01
Report all errors as messages not as Exceptions

### LINKED ISSUE #4912: lakefs_export: Report errors as messages not exceptions
lakefs_export reports errors by raising Python Exceptions.  This is hard for
its users to read, at least for `'@arielshaqed' in users`.  Report errors in
the regular way: write to standard error and then exit with a nonzero status
code.

Discovered while working on #4904, #4841.

arielshaqed:
This improves the UX and makes thing more scriptable and more UN*X-like.

Tested by manually running these failing commands:
```sh
❯ ./lakefs_export.py --branch asdf --commit_id sdf abc def
Cannot set both branch and commit

❯ ./lakefs_export.py abc def
Must set one of branch, commit

❯ ./lakefs_export.py --prev_commit_id asdf --commit_id sdf abc def
Cannot export diff between two commits.
```

Fixes #4912.

### DIFF
--- a/deployments/tools/export/lakefs_export.py
+++ b/deployments/tools/export/lakefs_export.py
@@ -60,6 +60,10 @@ Rewrite lines to explain more and remove weird logging indicators."""
     return
 
 
+def error(msg, statuscode=1):
+    print(msg, file=sys.stderr)
+    exit(statuscode)
+
 def main():
 
     # create rclone configuration file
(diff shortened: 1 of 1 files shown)


deployments/tools/export/lakefs_export.py
```diff
@@ -78,11 +82,13 @@ def main():
     elif not has_branch and has_commit:
         source += args.commit_id + "/"
         reference = args.commit_id
-    else:
-        raise Exception("Should assign branch or commit, but not both.")
+    elif has_branch:            # and has_commit
+        error("Cannot set both branch and commit_id")
+    else:                       # not has_branch and not has_commit
+         error("Must set one of branch, commit_id")
```
johnnyaug, line 88: strange indentation here

johnnyaug (approved):

deployments/tools/export/lakefs_export.py
arielshaqed: Thanks!  Turns out we never flake8 this.  Opening a _new_ bug to do that, but for now fixed _everything_ flake8 had to complain about.  So as punishment PTAL :-/

johnnyaug (approved):
LGTM

arielshaqed:
Thanks muchly!

PR #4918, opened 2023-01-01
Performance: Parallelize DeleteBatch

### LINKED ISSUE #4917: Parallelize DeleteBatch operation
As part of the gateway performance investigation we've identified the batch delete flow as a bottle neck which significantly reduces spark write time when override == true.
`reviews.repartition(4000).write.mode(SaveMode.Overwrite).parquet(path)`
The above command produces delete tasks which are split into batches of 1000 items with the following duration:
```
time="2022-12-27T18:29:01Z" level=info msg=Completed func="pkg/gateway/operations.(*DeleteObjects).Handle" file="build/pkg/gateway/operations/deleteobjects.go:118" Can Batch=true Operation=handleDeleteObjects Time Elapsed=2.991486742s

time="2022-12-27T18:29:04Z" level=info msg=Completed func="pkg/gateway/operations.(*DeleteObjects).Handle" file="build/pkg/gateway/operations/deleteobjects.go:118" Can Batch=true Operation=handleDeleteObjects Time Elapsed=2.908961254s

time="2022-12-27T18:29:07Z" level=info msg=Completed func="pkg/gateway/operations.(*DeleteObjects).Handle" file="build/pkg/gateway/operations/deleteobjects.go:118" Can Batch=true Operation=handleDeleteObjects Time Elapsed=3.09590818s

```

We can reduce time by parallelizing the batch delete

N-o-Z:
Closes #4917


N-o-Z:
Performance improvement:

Before 
`Command took 20.61 seconds -- by nir.ozery@treeverse.io at 01/01/2023, 10:49:26 on Nir Ozery's Cluster`

After
`Command took 18.06 seconds -- by nir.ozery@treeverse.io at 01/01/2023, 14:17:59 on Nir Ozery's Cluster`

pkg/graveler/graveler.go
```diff
@@ -1487,17 +1499,36 @@ func (g *Graveler) DeleteBatch(ctx context.Context, repository *RepositoryRecord
...
+				})
+			}
+
 			for _, key := range keys {
-				err := g.deleteUnsafe(ctx, repository, branch, key, &cachedMetaRangeID)
-				if err != nil {
-					m = multierror.Append(m, &DeleteError{Key: key, Err: err})
+				select {
+				case <-cctx.Done():
+					break
```
arielshaqed, line 1525: This will cause issues if you ever cancel: you never abort any of the planned workers.  So every delete worker will run until keysChan is empty.

We could equally well write
```go
			for _, key := range keys {
				keysChan <- key
			}
```
They are the same effectively because realistically ctx will not be cancelled before every key is loaded on keysChan -- blocking is impossible because keysChan has capacity for all of keys.

I think handling cancellation _inside_ the workers on line 1518 is the important part, and the only bit that will run in real cases.

pkg/graveler/graveler.go
```diff
@@ -1474,6 +1475,17 @@ func (g *Graveler) Delete(ctx context.Context, repository *RepositoryRecord, bra
 		})
 }
 
+func (g *Graveler) deleteConsumer(ctx context.Context, cancel context.CancelFunc, repository *RepositoryRecord, branch *Branch, cachedMetaRangeID *MetaRangeID, keysChan <-chan Key) error {
+	for key := range keysChan {
+		err := g.deleteUnsafe(ctx, repository, branch, key, cachedMetaRangeID)
+		if err != nil {
+			cancel()
+			return err
```
arielshaqed, line 1483:
```suggestion
		if err != nil {
		if !errors.Is(ctx, context.Canceled) {
			cancel()
		}
		return err
```
Otherwise when _one_ error occurs, every call is cancelled.  If there are 100 workers then that's 100 cancellations; [the code](https://cs.opensource.google/go/go/+/refs/tags/go1.19.4:src/context/context.go;l=394-433;drc=db36eca33c389871b132ffb1a84fd534a349e8d8) is tricky for me to grasp, but I'm pretty sure there may be 99 mutex contentions simply due to some layer on the stack having to call ctx.Err to retrieve the cancellation error.

pkg/graveler/graveler.go
```diff
@@ -1487,17 +1499,36 @@ func (g *Graveler) DeleteBatch(ctx context.Context, repository *RepositoryRecord
...
+
+			deleteWorkers := int(math.Min(float64(100), float64(len(keys))))
+			keysChan := make(chan Key, len(keys))
+			var wg multierror.Group
+			cctx, cancel := context.WithCancel(ctx)
+			defer cancel()
+			for i := 0; i < deleteWorkers; i++ {
+				wg.Go(func() error {
+					return g.deleteConsumer(cctx, cancel, repository, branch, &cachedMetaRangeID, keysChan)
+				})
```
arielshaqed, line 1519: I don't understand how this can return the correct errors.  We are supposed to delete _everything_ on the list, and return _all_ the errors.  But if _something_ on the list fails, AFAICT:

1. Each failing goroutine returns the _first_ error, cancels the rest, and stops.
2. Each other goroutine is _cancelled_ and returns a cancellation error for some arbitrary key -- even though that key may have been successfully deleted with no error!
3. Some keys are never tried and therefore _not_ reported as deleted and also _not_ reported as having failed deletion.

arielshaqed (changes requested):
Nice!  Increasing deletion parallelism is probably a great way to speed things up -- do we know by how much?

But I think that this code tries to be too efficient when deletions start failing, and as a result does not fullfil the contract I was expecting it to fullfil from the API: every item is either deleted or returns an error, or, in rare cases, both.

pkg/graveler/graveler.go
N-o-Z: You are right - it's part of the changes I've made while experimenting - will revert

pkg/graveler/graveler.go
```diff
@@ -1487,17 +1499,36 @@ func (g *Graveler) DeleteBatch(ctx context.Context, repository *RepositoryRecord
...
 	err = g.safeBranchWrite(ctx, g.log(ctx).WithField("operation", "delete_keys"),
 		repository, branchID, func(branch *Branch) error {
-			var cachedMetaRangeID MetaRangeID // used to cache the committed branch metarange ID
+			commit, err := g.RefManager.GetCommit(ctx, repository, branch.CommitID)
+			if err != nil {
+				return err
+			}
+			cachedMetaRangeID := commit.MetaRangeID
+
+			deleteWorkers := int(math.Min(float64(100), float64(len(keys))))
```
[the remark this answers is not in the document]
N-o-Z: It was measured and for batches of 1000 items the performance increase from 10 workers to 100 workers was two times

nopcoder (changes requested):
Any concurrent work should be also limited by the process and not by the call itself.
Very similar to what we did in the controller to enable parallel work on commit log.
Also, if the underlaying implementation and do the batch - discovery - maybe we should just push down the request and have the implementation choose the right to perform better.

pkg/graveler/graveler.go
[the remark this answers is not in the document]
arielshaqed: That's disappointing.  I know it takes a while, but I'd tune for a value that gets almost that performance increase with hopefully many fewer workers.

pkg/graveler/graveler.go
```diff
@@ -1487,17 +1498,32 @@ func (g *Graveler) DeleteBatch(ctx context.Context, repository *RepositoryRecord
...
+				return err
+			}
+			cachedMetaRangeID := commit.MetaRangeID
+
+			deleteWorkers := len(keys)
+			if deleteWorkers > maxWorkerNum {
+				deleteWorkers = maxWorkerNum
+			}
+			keysChan := make(chan Key, len(keys))
+			var wg multierror.Group
```
arielshaqed, line 1515: Usually wg is a WaitGroup; is-a multierror.Group a WaitGroup?

If not, I'd probably prefer
```suggestion
			var mg multierror.Group
```

pkg/graveler/graveler.go
```diff
@@ -1474,9 +1474,20 @@ func (g *Graveler) Delete(ctx context.Context, repository *RepositoryRecord, bra
 		})
 }
 
+func (g *Graveler) deleteConsumer(ctx context.Context, repository *RepositoryRecord, branch *Branch, cachedMetaRangeID *MetaRangeID, keysChan <-chan Key) error {
+	for key := range keysChan {
+		err := g.deleteUnsafe(ctx, repository, branch, key, cachedMetaRangeID)
+		if err != nil {
+			return err
+		}
```
arielshaqed, line 1482: AFAICT this will _still_ end up not fulfilling the DeleteObjects contract: all errors returned need to be wrapped in a DeleteError for NewMapDeleteErrors (called from [controller](https://github.com/treeverse/lakeFS/blob/4992d91d15983ea189f3d6025e1211db2fe2ebfa/pkg/api/controller.go#L194) and from the [S3 gateway](https://github.com/treeverse/lakeFS/blob/4992d91d15983ea189f3d6025e1211db2fe2ebfa/pkg/gateway/operations/deleteobjects.go#L127)).  Otherwise the call fails instead of returning success and a list of failures.

Also we end up not even trying to delete everything, which violates the DeleteObjects contract in another way.

Note that we do not have much leeway here to change semantics of the call, because the S3 gateway uses it and we have to be "S3 \"compatible\"".

arielshaqed (changes requested):
I still don't think that it satisfies the DeleteObjects contracts, and in another way that I did not notice at first.

I think we will need to make some way of configuring the number of workers on the call, so the TestDeleteObjects will be able to detect this violation.  I can imagine 2-3 ways in which it can fail, and I probably don't have enough imagination to figure out _all_ of them.

pkg/graveler/graveler.go
```diff
@@ -1474,9 +1474,20 @@ func (g *Graveler) Delete(ctx context.Context, repository *RepositoryRecord, bra
...
+	}
+	return nil
```
arielshaqed, line 1484: I think we might be better off with something like this:
```suggestion
```suggestion
	var retErr error
	for key := range keysChan {
		err := g.deleteUnsafe(ctx, repository, branch, key, cachedMetaRangeID)
		if err != nil {
			// do something with err to make it a DeleteError if it is one
			retErr = multierror.Append(retErr, err)
		}
	}
	return retErr
```

I think that will solve _some_ of the issues (and note that [multierror flattens Errors](https://github.com/hashicorp/go-multierror/blob/9974e9ec57696378079ecc3accd3d6f29401b3a0/append.go#L22-L25)).  But I'd still be happy if a test could catch the issues I think that we had.

pkg/graveler/graveler.go
N-o-Z: It is a waitGroup and this is the convention we are using in all usage instances

pkg/graveler/graveler.go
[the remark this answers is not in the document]
N-o-Z: I performed another check and it seems that that it was a variation in the ecosystem. It does work the same also for 10 workers. I will implement Barak's suggestion

N-o-Z:
Closing for now - reopen in the future if we are convinced that parallelization is the only way to improve performance

Issue #4897, opened 2022-12-29
Add integration test for GC on Spark 2 


lynnro314:
We have GC client for Spark2 but we currently don't test it. We should add integration test for GC on Spark 2.

PR #4920, opened 2023-01-01
Include aws-java-sdk-bundle in assembled jar

### LINKED ISSUE #4919: Error running GC sweep phase in Spark 2.x using hadoop 2.7.7
When running GC and specifiying `hadoop-aws 2.7.7 ` in packages, during the sweep phase we are getting 
```
Exception in thread "main" org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 57.0 failed 1 times, most recent failure: Lost task 0.0 in stage 57.0 (TID 1660, localhost, executor driver): java.lang.NoClassDefFoundError: com/amazonaws/auth/AWSStaticCredentialsProvider
	at io.treeverse.clients.conditional.S3ClientBuilder$.build(S3ClientBuilder.scala:26)
	at io.treeverse.clients.BulkRemoverFactory$S3BulkRemover.getS3Client(BulkRemoverFactory.scala:88)
	at io.treeverse.clients.BulkRemoverFactory$S3BulkRemover.deleteObjects(BulkRemoverFactory.scala:77)
	at io.treeverse.clients.GarbageCollector$$anonfun$bulkRemove$1$$anonfun$apply$2.apply(GarbageCollector.scala:568)
	at io.treeverse.clients.GarbageCollector$$anonfun$bulkRemove$1$$anonfun$apply$2.apply(GarbageCollector.scala:568)
```

This happens because on startup it installed `aws-java-sdk 1.7.4` as a dependency. this bundle does not include AWSStaticCredentialsProvider and our code uses this class [here](https://github.com/treeverse/lakeFS/blob/922ea2b6f5f48578894a5a4a3d824ad0bc1555d9/clients/spark/core/src/main/hadoop2/scala/io/treeverse/clients/conditional/S3ClientBuilder.scala#L27). 

### Steps to reproduce 
1. setup a spark 2.4.7/8 env 
2. setup a repo with objects to get garbage collected 
3. runs gc with --packages org.apache.hadoop:hadoop-aws:2.7.7, and using lakefs-spark-client-247-assembly-0.6.0.jar

talSofer:
Fixes #4919

### Background 
* Our GC code depends on certain classes provided by aws-sdk. One example is `AWSStaticCredentialsProvider` which is part of version [1.11.30](https://github.com/aws/aws-sdk-java/blob/1.11.30/aws-java-sdk-core/src/main/java/com/amazonaws/auth/AWSStaticCredentialsProvider.java) of [aws-java-sdk](https://mvnrepository.com/artifact/com.amazonaws/aws-java-sdk), but is also part of any version of [aws-java-sdk-bundle](https://mvnrepository.com/artifact/com.amazonaws/aws-java-sdk-bundle). 
* lakeFS spark client has three versions:
1. Spark 312 + hadoop 3 
2. Spark 301 + hadoop 2
3. Spark 247 + hadoop 2

### Problem 
While running GC on a **hadoop 2** infrastructure (clients 1 & 2), the provided hadoop-aws package typically installs an aws-java-sdk version < 1.11.30 and leaving the user with classes that are missing at runtime. On the other hand, while running GC on a **hadoop 3** infrastructure (client 3), the provided hadoop-aws package installs an aws-java-sdk version >= 1.11.30 making the required classes available at runtime. 

### Solution 
For hadoop2 GC clients, ensure that the client assembly jar will forever include the classes it needs to run, independently of backward-incompatible changes to AWS libraries. For hadoop 3 clients, leave things as is. 

### How was this tested 
* Tested locally with spark 247 + hadoop 277
* Tested on databricks Spark 301 and the latest spark 
* Used dev-on-prod to test managed GC and ensure it is working
* [This](https://github.com/treeverse/lakeFS/pull/4933) pr introduces an integration test for the current pr - still haven't tested with it.

### DIFF
--- a/clients/spark/build.sbt
+++ b/clients/spark/build.sbt
@@ -62,7 +62,7 @@ def generateCoreProject(buildType: BuildType) =
         // hadoop-aws provides AWS SDK at version >= 1.7.4.  So declare this
         // version, but ask to use whatever is provided so we do not
         // override what it selects.
-        "com.amazonaws" % "aws-java-sdk-bundle" % "1.12.194" % "provided",
+        "com.amazonaws" % "aws-java-sdk-bundle" % "1.12.194",
         "com.azure" % "azure-core" % "1.10.0",
         "com.azure" % "azure-storage-blob" % "12.9.0",
         "com.azure" % "azure-storage-blob-batch" % "12.7.0",
@@ -182,7 +182,8 @@ lazy val assemblySettings = Seq(
     rename("okio.**").inAll,
     rename("okhttp3.**").inAll,
     rename("reactor.netty.**").inAll,
-    rename("reactor.util.**").inAll
+    rename("reactor.util.**").inAll,
+    rename("com.amazonaws.**").inAll
   )
 )
 
--- a/clients/spark/build.sbt
+++ b/clients/spark/build.sbt
@@ -36,6 +36,11 @@ def generateCoreProject(buildType: BuildType) =
   Project(s"${baseName}-client-${buildType.name}", file(s"core"))
     .settings(
       sharedSettings,
+      if (buildType.hadoopFlavour == "hadoop2") {
+        hadoop2AssemblySettings
+      } else {
+        hadoop3AssemblySettings
+      },
       s3UploadSettings,
       settingsToCompileIn("core", buildType.hadoopFlavour),
       scalaVersion := buildType.scalaVersion,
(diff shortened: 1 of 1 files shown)


nopcoder:
@talSofer the PR embed in the assembly the same package we document to use when we submit the job, so how is it going to work? we are going to update the documentation not to sue the package now? how is it that the client we provide doesn't work with the assembly? also, is it not the same setup we running use when we run the tests?

talSofer:
> the PR embed in the assembly the same package we document to use when we submit the job

@nopcoder This is incorrect. the docs say use org.apache.hadoop:hadoop-aws:2.7.7 and we are embedding com.amazonaws:aws-java-sdk-bundle:1.12.194. 

>  how is it that the client we provide doesn't work with the assembly 

since aws-java-sdk-bundle:1.12.194 declared as provided, the assembly does not include it. this bundle has the AWSStaticCredentialsProvider class, which is also included in hadoop-aws > 2.7.7. so I think that without either a hadoop-aws > 2.7.7 or aws-java-sdk-bundle:1.12.194 (shaded, otherwise it clashes with other stuff) the client fails. 

> also, is it not the same setup we running use when we run the tests?

We don't have a section for spark 2 on the metadata client tests, we have this open issue for it https://github.com/treeverse/lakeFS/issues/4897. we have spark 3 + hadoop 2, and spark 3 + hadoop 3 (see https://github.com/treeverse/lakeFS/blob/9b6478ab1518f9d5981c10f2a8271238f8ec3e82/.github/workflows/esti.yaml#L908)

arielshaqed (commented):
This will require a lot of testing, TBH I'm not even sure what.  Can you list what you're planning?

talSofer:
Thanks @arielshaqed. 
Testing done: 
* Run gc locally with a lakeFS installation connected to aws with Spark 2.4.7 and hadoop 
* Integration tests pass (they validate Spark 3 + hadoop 3 and Spark 3 + hadoop 2)

Planned tests: 
* Run in databricks I guess? databricks don't let us run Spark 2. 
* any other suggestions?

Jonathan-Rosenberg:
@arielshaqed I think that it will require as much testing as our current implementation. 
We didn't find the bug that this PR fixes until a user raised a flag.
The only thing that's missing (I think) is testing it with Spark 2, and possibly testing it with different combinations of Spark 3 and the different Hadoop versions.

talSofer:
Thanks @Jonathan-Rosenberg! 
> The only thing that's missing (I think) is testing it with Spark 2

Do you mean in addition to the manual testing? as in add integration tests? 

> and possibly testing it with different combinations of Spark 3 and the different Hadoop versions.

here again, our integration tests validate gc with spark 3/hadoop 2 and spark 3/hadoop 3 combinations. What testing are you referring to?

arielshaqed:
I guess test it on some old DataBricks Spark that uses a Hadoop 2, then.

Jonathan-Rosenberg:
@talSofer 
> Do you mean in addition to the manual testing? as in add integration tests?

Optimally integration tests (but probably not part of this PR).

> our integration tests validate gc with spark 3/hadoop 2 and spark 3/hadoop 3 combinations. What testing are you referring to?

I would probably test the Hadoop versions that I know that something relevant to GC was introduced in them. For example, 2.7.7 (as a base), 2.10.2 (the change in java sdk bundle), 3.2.1 (as a commonly used version?), the latest Hadoop 3 version (to validate that nothing is breaking)

johnnyaug (approved):
LGTM

arielshaqed (approved):
Thanks!

I think we should also test this on DataBricks, if they still have something old enough to have Hadoop 2 :-)

Jonathan-Rosenberg (approved):
COOL

PR #4921, opened 2023-01-01
UncommittedGC fix for azure support

### LINKED ISSUE #4844: Discovery: GC+ support for Azure
2 days discovery to check GC+ on Azure, if it's not working what is missing to support it.

nopcoder:
Close #4844

Changes to support Azure part of discovery - tested manually.

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedAddressLister.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedAddressLister.scala
@@ -23,10 +23,10 @@ class APIUncommittedAddressLister(apiClient: ApiClient) extends UncommittedAddre
     var continuationToken: String = null
     do {
       resp = apiClient.prepareGarbageCollectionUncommitted(repo, continuationToken)
-      continuationToken = resp.getContinuationToken()
+      continuationToken = resp.getContinuationToken
     } while (StringUtils.isNotBlank(continuationToken))
-    val runID = resp.getRunId()
-    val uncommittedLocation = resp.getGcUncommittedLocation()
+    val runID = resp.getRunId
+    val uncommittedLocation = resp.getGcUncommittedLocation
     new UncommittedGCRunInfo(uncommittedLocation, runID)
   }
 }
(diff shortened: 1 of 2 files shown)


clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedAddressLister.scala
```diff
@@ -23,10 +23,10 @@ class APIUncommittedAddressLister(apiClient: ApiClient) extends UncommittedAddre
     var continuationToken: String = null
     do {
       resp = apiClient.prepareGarbageCollectionUncommitted(repo, continuationToken)
-      continuationToken = resp.getContinuationToken()
+      continuationToken = resp.getContinuationToken
```
N-o-Z, line 26: These changes seem unnecessary?

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
```diff
@@ -156,7 +157,11 @@ object UncommittedGarbageCollector {
 
         var uncommittedDF =
           if (uncommittedGCRunInfo.uncommittedLocation != "") {
-            spark.read.parquet(uncommittedGCRunInfo.uncommittedLocation)
+            val uncommittedLocation =
```
N-o-Z, line 160: Can you explain what exactly was the issue with Azure?

N-o-Z (changes requested):
Thanks - some comments

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedGarbageCollector.scala
nopcoder: When we use Hadoop's FileSystem for the repository storage namespace we are getting AbstractFileSystem that can't be used for listing.
By converting the storage namespace address to `abfs://...` scheme and query for FileSystem, we are getting AzureBlobStoreFileSystem that enables the API we use.
Used the same method that the current GC users - fetch the right storage NS address based on the client you use.

clients/spark/core/src/main/scala/io/treeverse/gc/UncommittedAddressLister.scala
nopcoder: Just my IDE's scala recommended way calling java accessor methods:

image

N-o-Z (approved):
Thanks

PR #4922, opened 2023-01-02
API GetCommit return empty array instead of nil for no parent

### LINKED ISSUE #4829: Get commits, without parents error with py client
Calling get-commits , throws when parents field is missing , which is ok for the initial commit-ids ie the commit-id of the initial branch creation. 
```
lakefs_client.exceptions.ApiTypeError: Invalid type for variable 'parents'. Required value type is list and passed type was NoneType at ['received_data']['parents']`
```

nopcoder:
Fix #4829

Added API test to validate the initial commit structure


N-o-Z (approved):
LGTM!

PR #4928, opened 2023-01-04
Graveler Set with IfAbsent implemented as set without override

### LINKED ISSUE #3705: Graveler's Set flow for `IfAbsent` pre condition is wrong
If the key exists in the committed data, Graveler will return `ErrPreconditionFailed`, although it may have been deleted in the in the staging area. Moreover, tombstones are not being overwritten by this condition in the staging manager, although a tombstone meaning is actually: "This key does not exist".

### LINKED ISSUE #4796: Graveler's Set IfAbsent during commit can return incorrect response
The issue occurs while we perform `Set` IfAbsent while `Commit` in progress.
Current implementation "retry" the operation in case of a change in the staging token (like after commit).
In this case we need to verify that the object does not exists in the latest commit.
But if a commit runs during our 'set' it means that the committed data is caused by our operation.

Scenario thanks to @arielshaqed!

**Thread A:**
Get staging token, enter setFunc, see that object does not exist, [update its value](https://github.com/treeverse/lakeFS/blob/89f8c4086afa97afbdc10d1b41cc9913c3e5008a/pkg/graveler/graveler.go#L1417), return successfully, block before reloading the staging token.

**Thread B:**
Commit successfully, object now exists on committed.

**Thread A:**
Reload staging token, see that it's changed, try again.  Enter setFunc again, find that the object on the branch (d'oh!), [return ErrPreconditionFailed](https://github.com/treeverse/lakeFS/blob/89f8c4086afa97afbdc10d1b41cc9913c3e5008a/pkg/graveler/graveler.go#L1407).

**Result:** Object did not exist, we set it, and claimed that it already existed before.

nopcoder:
Close #3705
Close #4796

This change propose to handle Set+IfAbsent behaviour as "set without override".
The call will ensure we will not override existing key value - without error in the case an existing key already set.
Caller will have to manually update the entry and fetch the value after the conditional set to verify that the caller's value was set.


pkg/graveler/graveler.go
```diff
@@ -775,6 +775,7 @@ type StagingManager interface {
 	Set(ctx context.Context, st StagingToken, key Key, value *Value, overwrite bool) error
 
 	// Update updates a (possibly nil) value under the given staging token and key.
+	// Skip update in case value returned from 'updateFunc' is the same as the value passed to call.
```
arielshaqed, line 778:
```suggestion
	// It skips update when the value returned from 'updateFunc' is the same as the value passed to call.
```
Also please clarify what "value" means here.

pkg/graveler/staging/manager.go
```diff
@@ -82,7 +82,8 @@ func (m *Manager) Update(ctx context.Context, st graveler.StagingToken, key grav
 		oldValue = graveler.StagedEntryFromProto(oldValueProto)
 	}
 	updatedValue, err := updateFunc(oldValue)
-	if err != nil {
+	if err != nil || updatedValue == oldValue {
```
arielshaqed, line 85: `updatedValue == oldValue` is a bit subtle: it need to be the same slice, not just the same sequence of bytes.  I completely understand **why**, but I am worried that it will not be clear enough at the call site.

Some options:
* Improve documentation of this method in the interface.  Right now it says:

  ```go
	// Update updates a (possibly nil) value under the given staging token and key.
  ```
* Don't do it this way.  Instead add a new "error" return option for `updateFunc`, call it say `SkipUpdate` -- and document _that_ as something like
  > SkipUpdate is returned from ValueUpdateFunc to request _not_ to update the value.

arielshaqed (approved):
Yeah, this is great stuff!

Similarly to other options, we should additionally improve documentation in swagger.yml.  I think we can and therefore should explicitly remove 412 Precondition Failed, document that with `If-None-Match: *` it succeeds if the object exists after the call, and probably also document it on 201 Created.  I prefer to be able to call the API by using only swagger.yml as reference.

pkg/graveler/staging/manager.go
nopcoder: Using on special error

PR #4930, opened 2023-01-04
Performance: Optimize Delete flow

### LINKED ISSUE #4929: Optimize delete flow
Add option to update KV store key only if exists. Use it to add an optimistic attempt to write tombstone assuming key exists on staging token at the beginning of the delete flow

N-o-Z:
Closes #4929

### DIFF
--- a/pkg/kv/kvtest/store.go
+++ b/pkg/kv/kvtest/store.go
@@ -9,10 +9,9 @@ import (
 	"sync"
 	"testing"
 
-	"github.com/stretchr/testify/require"
-
 	"github.com/hashicorp/go-multierror"
 	nanoid "github.com/matoous/go-nanoid/v2"
+	"github.com/stretchr/testify/require"
 	"github.com/treeverse/lakefs/pkg/kv"
 	_ "github.com/treeverse/lakefs/pkg/kv/mem"
 	kvparams "github.com/treeverse/lakefs/pkg/kv/params"
--- a/pkg/kv/kvtest/store.go
+++ b/pkg/kv/kvtest/store.go
@@ -293,6 +293,30 @@ func testStoreSetIf(t *testing.T, ms MakeStore) {
 			t.Fatalf("SetIf err=%v - key=%s, value=%s, pred=nil, expected err=%s", err, key, val2, kv.ErrPredicateFailed)
 		}
 	})
+
+	t.Run("update_if_exists", func(t *testing.T) {
+		key := uniqueKey("set-if-exists")
+		val1 := []byte("v1")
+		err := store.Set(ctx, []byte(testPartitionKey), key, val1)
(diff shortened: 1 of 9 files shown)


pkg/kv/store.go
```diff
@@ -17,6 +17,8 @@ const (
 	InitialMigrateVersion = 1
 	PathDelimiter         = "/"
 	MetadataPartitionKey  = "kv-internal-metadata"
+
+	ConditionalExists = "dd136a329c1f297d37d039d9ddd99699"
```
nopcoder, line 21: - We need to define a type - Like `Precondition` or `Precond` (can be string or int - iota) and use `var` to define `ConditionExists` with any value.
- Suggestion name the variable PrecondConditionalExists or something that will include the type / parameter type as a prefix, so it will have more context as the precond value type is interface{}.

pkg/kv/local/store.go
```diff
@@ -127,14 +127,16 @@ func (s *Store) SetIf(ctx context.Context, partitionKey, key, value []byte, valu
...
 			}
-			val, err := item.ValueCopy(nil)
-			if err != nil {
-				log.WithError(err).Error("could not get byte value for predicate")
-				return err
-			}
-			if !bytes.Equal(val, valuePredicate.([]byte)) {
-				log.WithField("predicate", valuePredicate).WithField("value", val).Trace("predicate condition failed")
-				return kv.ErrPredicateFailed
+			if valuePredicate != kv.ConditionalExists {
```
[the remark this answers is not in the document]
N-o-Z: Please note that in line 126 we check if error is ErrKeyNotFound and return ErrPredicateFailed

pkg/kv/kvtest/store.go
```diff
@@ -9,6 +9,8 @@ import (
 	"sync"
 	"testing"
 
+	"github.com/stretchr/testify/require"
+
```
nopcoder, line 13: remove ws

pkg/kv/store.go
```diff
@@ -33,6 +33,12 @@ var (
 	ErrTableNotActive      = errors.New("table not active")
 )
 
+// Precond Type for special conditionals provided as predicates for the SetIf method
+type Precond string
+
+// PrecondConditionalExists Conditional for SetIf which performs Set only if key already exists in store
+var PrecondConditionalExists = Precond("dd136a329c1f297d37d039d9ddd99699")
```
nopcoder, line 40: Will be nice when we debug to see the value render with the meaning
```suggestion
var PrecondConditionalExists = Precond("ConditionalExists")
```

nopcoder (approved):
LGTM - minor comments left

idanovo (approved):
LGTM

PR #4933, opened 2023-01-05
Test GC with Spark 3 / Hadoop 2

arielshaqed:
Build a new version (manually for now) and use it to test GC when Spark 3
itself is running with Hadoop 2.  (Currently the Hadoop 2 version is tested
to run on Spark 3 with Hadoop 3, which is nice but not the main use-case!)


arielshaqed:
Should reproduce the #4919 failure.

talSofer (approved):
Super cool! thanks

PR #4943, opened 2023-01-07
Docs: Added Lua library reference

ozkatz:
* Added a full reference of available Lua packages
* Made `hooks` a directory with one file per type
* Aligned Lua function naming convention


itaiad200 (approved):
Thanks for the hooks docs reorg!

PR #4944, opened 2023-01-08
4476/fix hadoopfs close twice

### LINKED ISSUE #4946: fix hadoopfs close twice - linking the same physical address
pr #4900 disable the ability to link to the same physical address twice.
when closing the hadoopfs we also link to the physical address.
hadoopfs requires closing twice to return the same result.

eden-ohana:
closes #4946

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LinkOnCloseOutputStream.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LinkOnCloseOutputStream.java
@@ -18,6 +18,7 @@ class LinkOnCloseOutputStream extends OutputStream {
     private final URI physicalUri;
     private final MetadataClient metadataClient;
     private final OutputStream out;
+    private boolean isLinked;
 
     /**
      * @param lfs LakeFS file system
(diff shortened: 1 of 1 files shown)


N-o-Z (approved):
LGTM!
We should wait for eco to approve as well

Jonathan-Rosenberg (approved):

clients/hadoopfs/src/main/java/io/lakefs/LinkOnCloseOutputStream.java
```diff
@@ -63,11 +64,14 @@ public void close() throws IOException {
...
-            lfs.linkPhysicalAddress(objectLoc, stagingLoc, physicalUri, metadataClient);
-        } catch (io.lakefs.clients.api.ApiException e) {
-            throw new IOException("link lakeFS path to physical address", e);
+        if (!this.isLinked) {
+            try {
+                lfs.linkPhysicalAddress(objectLoc, stagingLoc, physicalUri, metadataClient);
+            } catch (io.lakefs.clients.api.ApiException e) {
+                throw new IOException("link lakeFS path to physical address", e);
+            }
+            this.isLinked = true;
```
arielshaqed, line 73: This is a race, it may fail to trigger in some cases -- for instance when 2 threads both close the same LinkOnCloseOutputStreamOutputStream.  This will occur in some use-cases, and if the API allows multiple calls to close then it makes sense to require it to work.

If we believe that this fixes an actual bug then the fix is bad.  Otherwise it just adds a nicer error report and somewhat reduces the frequency of the bug... but does not remove it.

arielshaqed (commented):
Sorry, late for this one.

This fixes a very particular issue: multiple closes **from the same thread** will now work.  However if they occur from different thread then they might be concurrent and it might fix nothing.

So we need to understand what the requirement is here.

PR #4945, opened 2023-01-08
Return secondary key for secondary iterator

guy-har:


N-o-Z (changes requested):
I'm not really sure why is this change required - The secondary iterator iterates using the secondary key but should return the primary value

nopcoder (approved):
Good catch!!

N-o-Z (approved):
Thanks - approved!

PR #4947, opened 2023-01-08
[lakeFSFS] Use UploadObject IfNotExists to "create" "directories"

### LINKED ISSUE #4722: lakeFSFS: Use If-None-Match in createDirectoryMarkerIfEmptyDirectory
The createDirectoryMarkerIfEmptyDirectory function checks for existence of a
directory marker, then creates it if needed.  lakeFS supports creating an
object only if does not exist.  Use that!

In fact, createDirectoryMarker could just do this directly, and this
function deleted.

Saves 1 API call for every call to createDirectoryMarkerIfEmptyDirectory.
Relevant for #4676.

arielshaqed:
Saves 1 listing operation on every rename() or delete() operation.

Fixes #4722.

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -310,7 +310,7 @@ public class LakeFSFileSystem extends FileSystem {
         }
         if (!src.getParent().equals(dst.getParent())) {
             deleteEmptyDirectoryMarkers(dst.getParent());
-            createDirectoryMarkerIfEmptyDirectory(src.getParent());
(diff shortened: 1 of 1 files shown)


arielshaqed:
### ⚠️ Reviewer please note ⚠️ 

If you run this client against an older lakeFS server then it will fail due to receiving a "conflict" HTTP return response.  I am not sure how best to handle this: the newer client APIs will stop recognizing this return code!  I _suggest_ that I handle it by examining HTTP status codes directly.

nopcoder (approved):
LGTM!

arielshaqed:
Cool, tests pass thanks to #4995.  Thanks @nopcoder!

I can pull, it will be neat.

PR #4952, opened 2023-01-08
Fix test secondary iterator

nopcoder:
Fix secondary index iterator test to use the right proto messages for test data and checking the iterator functionality.

### DIFF
--- a/pkg/kv/kvtest/iterators.go
+++ b/pkg/kv/kvtest/iterators.go
@@ -12,7 +12,6 @@ import (
 const (
 	firstPartitionKey  = "m"
 	secondPartitionKey = "ma"
-	thirdPartitionKey  = "mb"
 )
 
 func testPartitionIterator(t *testing.T, ms MakeStore) {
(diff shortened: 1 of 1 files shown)


N-o-Z (changes requested):
Thanks Barak - If I'm not mistaken it seems that we're missing the validation of the keys themselves which is the issue that @guy-har found with the paignation

nopcoder:
> Thanks Barak - If I'm not mistaken it seems that we're missing the validation of the keys themselves which is the issue that @guy-har found with the paignation

There is no way to validate the value we keep under kv as we do not encode the message metadata, just the data.
The test use the message type for the model and the secondary index as it worked because the first field in the message was the same type.
Update test code uses the right message type and validates iterate after 'key' works as the index and the secondary now holds different values.

N-o-Z:
> > Thanks Barak - If I'm not mistaken it seems that we're missing the validation of the keys themselves which is the issue that @guy-har found with the paignation
> 
> There is no way to validate the value we keep under kv as we do not encode the message metadata, just the data. The test use the message type for the model and the secondary index as it worked because the first field in the message was the same type. Update test code uses the right message type and validates iterate after 'key' works as the index and the secondary now holds different values.

secondary iterator returns MessageEntry which is key + value, in the unit tests we only validate the value part without the key

nopcoder:
> secondary iterator returns MessageEntry which is key + value, in the unit tests we only validate the value part without the key

Key not found does lookup for "just_a_string/e" find the secondary key and fail to find the value. Is this the test we are after?

N-o-Z (approved):
LGTM

PR #4953, opened 2023-01-08
change to dynamodb on-demand capacity

### LINKED ISSUE #4626: Bug: Default configuration of lakefs DynamoDb uses provisioned capacity
While the documentation mentions that the DynamoDb table is created in on-demand mode by default:

> lakeFS will create a table on the DB, with the default on-demand setting. No need to specify how much read and write throughput you expect your application to perform, as DynamoDB instantly accommodates your workloads as they ramp up or down.
cited from [Sizing guide](https://docs.lakefs.io/understand/sizing-guide.html#dynamodb-ram)

It is actually creating it in provisioned mode with settings of 1000 read and write capacity. Here the reference to [AWS Docs](https://docs.aws.amazon.com/wellarchitected/latest/serverless-applications-lens/capacity.html).
This is also mentioned in the [Configuration Reference](https://docs.lakefs.io/reference/configuration.html#reference), conflicting with the other documentation.
This setting creates significant cost in AWS (~24$ per day for an empty database) if you just setup the instance to evaluate the system.

IMO, the default setting should be on-demand, as this will not create any cost while the database is not heavily/regularly used (which I guess is common for deployments with default settings).

ortz:
Closes #4626

Changing to DynamoDB on-demand capacity by default instead of provisioned capacity as it creates unwanted costs


nopcoder (approved):
LGTM!

PR #4955, opened 2023-01-09
[metadata client] Start scalacheck-ing jpebble

arielshaqed:
Start testing the jpebble parser using ScalaCheck.
[ScalaCheck](https://github.com/typelevel/scalacheck/blob/main/doc/UserGuide.md)
generates tests automatically from specifications.  Obviously it cannot find
all bugs, but it *is* a pretty nice directed fuzzer.

I only tested a few low-level routines, but already found some nice issues:

* readFixedInt is indeed equivalent to readInt32.  We suspected as much but
  were unsure how to test it.  I added a ScalaCheck to check that they are
  identical.  It passed so I removed readFixedInt, along with the
  ScalaCheck.  View the progression on the commit history.

  Given that the likely divergence would be on negative (high bit set) Ints,
  I believe this ScalaCheck gives us the confidence to do so.
* readMagic would fail when it could not read enough bytes.  This makes
  failures faster for SSTable "files" which are less than 4 bytes in length.
  It also makes readMagic safe to use elsewhere.

### DIFF
--- a/clients/spark/build.sbt
+++ b/clients/spark/build.sbt
@@ -140,7 +140,8 @@ def getSharedLibraryDependencies(buildType: BuildType) : Seq[ModuleID]  = {
     "org.scalactic" %% "scalactic" % "3.2.9",
     "dev.failsafe" % "failsafe" % "3.2.4",
     "org.apache.hadoop" % "hadoop-distcp" % buildType.hadoopVersion,
-    "org.scalatestplus" %% "mockito-4-6" % "3.2.14.0" % "test",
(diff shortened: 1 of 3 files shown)


johnnyaug (approved):
Thanks!

arielshaqed:
Thanks!

I'll take this as agreement to use ScalaCheck, and pull.  I think we can make good use of this, particularly on the lower levels of stuff: that usually has less intent and more edge cases, so the enhanced safety is worth it.  Plus it's really not that expensive to code these checks!

PR #4960, opened 2023-01-09
Check-in mock gen code into the project

### LINKED ISSUE #4961: Enable go import of lakeFS kv package
Currently the `pkg/kv` package in lakeFS depends on generated code step - mockgen.
Enabling import of this package will require that all code required to build/test the package will be checked-in.

nopcoder:
This change will embed the mockgen go code into the project and change the default `gen` make step not to create these files.

To re-generate the mockgen files use `make gen-code`

- Make gen-code will regenerate the mockgen files
- Makefile will verify that the mockgen generated code is up-to-date
- Default `make gen` will not generate these files
- Make `gen-api` uses `go generate` so in case of `auth` package it will also re-generate the mockgen

Close #4961


itaiad200 (approved):
LGTM!

PR #4985, opened 2023-01-09
Add support for presigned URLs in the lakeFS API

### LINKED ISSUE #3160: Direct-access FileSystem to be credential-free
When using the the lakeFS Hadoop Filesystem, I need to configure the `fs.s3a.access_key` and `fs.s3a.secret_key`.

Use the fact that lakeFS already authenticates with AWS to remove the need for these configurations.

Handle the authentication by means of:

## Option 1: assume role on lakeFS server side

Add a lakeFS server configuration ("role_to_assume" in the same place where AWS credentials are given).
Then return session credentials and use them in the lakeFS filesystem.

The credentials given to lakeFS are expected to be allowed to assume this role.
 
## Option 2: Pre-signed URLs
Use the lakeFS server to sign the URLs returned to the lakeFS filesystem.
The credentials given to lakeFS are expected to be allowed to sign URLs.

ozkatz:
This one requires discussion :)

Currently, lakeFS users have 2 options when reading and writing data:

1. Involve lakeFS in the data path: this typically means either using the s3 gateway to read/write data, or to use the data operations in the lakeFS API (get_object, put_object)
2. Use lakeFS for metadata only, with direct read/write to the underlying object store.

This requires a pretty hard tradeoff: **(1)** has implications regarding security (data going through the lakeFS server) as well as added operational complexity (i.e. the need to size a lakeFS installation for the required bandwidth) and potentially cost implications if the lakeFS installation is outside the AWS account and VPC used by its consumers (such as in lakeFS Cloud).

As for **(2)**, it has a different set of problems: Once the metadata operation is done, direct access to the object store requires authentication and authorization. This means managing 2 different auth* systems: one for lakeFS and one for the underlying object store. Keeping both of them in sync is very hard (i.e. how do I give every user permissions to read only the exact underlying objects in a repository?)

This PR introduces pre-signed URLs as a middle ground. It allows receiving short-lived pre-signed URLs from `GetObject`, `StatObject` and `ListObjects` with the optional `presign=true` query parameter, as well as limited write support using `GetPhysicalAddress` where the address returned is pre-signed for a write operation.

This PR currently *does not* handle Multipart uploads, this could be added later (and is probably a requirement before it resolves #3160) 

---

**This is currently in draft and requires discussion. I don't plan on merging this unless we agree this is desired.**


arielshaqed:
Yay, this is a great direction!

One worry is how to use this from lakeFSFS, because I would really like to use this from lakeFSFS.  While this is certainly not the _only_ client, it's a pretty important one.  Here I am not sure how to use this _and_ get the optimizations from S3A and other underlying FileSystems.

pkg/api/controller.go
```diff
@@ -333,6 +333,21 @@ func (c *Controller) GetPhysicalAddress(w http.ResponseWriter, r *http.Request,
...
+		preSignedURL, err := c.BlockAdapter.GetPreSignedURL(ctx, block.ObjectPointer{
+			StorageNamespace: repo.StorageNamespace,
+			Identifier:       address,
+			IdentifierType:   block.IdentifierTypeRelative,
+		}, block.PreSignModeWrite)
+		if err != nil {
+			writeError(w, r, http.StatusInternalServerError, err)
+			return
+		}
+		response.PresignedUrl = StringPtr(preSignedURL)
```
arielshaqed, line 348: Nit: I think
```suggestion
		response.PresignedUrl = &preSignedURL
```
is more readable as well as more efficient.

pkg/api/controller.go
```diff
@@ -3099,6 +3126,31 @@ func (c *Controller) ListObjects(w http.ResponseWriter, r *http.Request, reposit
...
+				if authResponse.Allowed {
+					objStat.PhysicalAddress, err = c.BlockAdapter.GetPreSignedURL(ctx, block.ObjectPointer{
+						StorageNamespace: repo.StorageNamespace,
+						Identifier:       entry.PhysicalAddress,
+						IdentifierType:   entry.AddressType.ToIdentifierType(),
+					}, block.PreSignModeRead)
+					if c.handleAPIError(ctx, w, r, err) {
+						return
+					}
+				}
```
arielshaqed, line 3152: When I think of it this is really neat: I will be able to read all contents of an entire directory with a single listObjects on lakeFS.

api/swagger.yml
```diff
@@ -2872,6 +2888,12 @@ paths:
         - staging
       operationId: getPhysicalAddress
       summary: get a physical address and a return token to write object to underlying storage
+      parameters:
+        - in: query
+          name: presign
+          required: false
+          schema:
+            type: boolean
```
arielshaqed, line 2896: I think that this should be a POST endpoint that redirects!

arielshaqed (approved):
Neat stuff!  Can you open an issue optionally to rewrite `lakectl fs {upload,car} --direct` to use this redirecting mode?  It will make it so much simpler and more natural!

Approving as a satisfied future user of this, obviously not as a versioning engine team member.

(Still not sure how to use it efficiently in lakeFSFS, but that's a different matter!)

api/swagger.yml
```diff
@@ -953,6 +953,11 @@ components:
         token:
           type: string
           description: opaque staging token to use to link uploaded object
+        presigned_url:
+          type: string
+          nullable: true
```
nopcoder, line 958: Can we remove this one and just keet it 'non required' field? Empty have the same meaning as nil in this case and it will make the client code simple.

esti/presign_test.go
```diff
@@ -0,0 +1,30 @@
+package esti
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/treeverse/lakefs/pkg/api"
```
nopcoder, line 12:
```suggestion
	"github.com/stretchr/testify/require"
	"github.com/treeverse/lakefs/pkg/api"
```

pkg/api/controller.go
```diff
@@ -2990,9 +3005,20 @@ func (c *Controller) GetObject(w http.ResponseWriter, r *http.Request, repositor
...
+	pointer := block.ObjectPointer{StorageNamespace: repo.StorageNamespace, Identifier: entry.PhysicalAddress}
+	if params.Presign != nil && *params.Presign {
+		location, err := c.BlockAdapter.GetPreSignedURL(ctx, pointer, block.PreSignModeRead)
+		if c.handleAPIError(ctx, w, r, err) {
+			return
+		}
+		w.Header().Set("Location", location)
+		w.WriteHeader(http.StatusFound)
+		return
+	}
```
nopcoder, line 3018: Please verify that this flow works from python sdk client - there were issues with different return types, just want to make sure.

pkg/api/controller.go
```diff
@@ -2990,9 +3005,20 @@ func (c *Controller) GetObject(w http.ResponseWriter, r *http.Request, repositor
 		return
 	}
 
+	// if pre-sign, return a redirect
+	pointer := block.ObjectPointer{StorageNamespace: repo.StorageNamespace, Identifier: entry.PhysicalAddress}
+	if params.Presign != nil && *params.Presign {
```
nopcoder, line 3010:
```suggestion
	if swag.BoolValue(params.Presign) {
```

pkg/api/controller.go
```diff
@@ -3099,6 +3126,31 @@ func (c *Controller) ListObjects(w http.ResponseWriter, r *http.Request, reposit
 			if (params.UserMetadata == nil || *params.UserMetadata) && entry.Metadata != nil {
 				objStat.Metadata = &ObjectUserMetadata{AdditionalProperties: entry.Metadata}
 			}
+			if params.Presign != nil && *params.Presign {
```
nopcoder, line 3129:
```suggestion
			if swag.BoolValue(params.Presign) {
```

pkg/block/azure/adapter.go
```diff
@@ -196,6 +214,46 @@ func (a *Adapter) Get(ctx context.Context, obj block.ObjectPointer, _ int64) (io
...
+	}
+	vals := azblob.BlobSASSignatureValues{
+		Protocol:      azblob.SASProtocolHTTPS, // Users MUST use HTTPS (not HTTP)
+		ExpiryTime:    a.preSignedURLDurationGenerator(),
+		ContainerName: qualifiedKey.ContainerName,
+		BlobName:      qualifiedKey.BlobURL,
+		Permissions:   permissions.String(),
+	}
+
+	keyCredentials, ok := a.credentials.(*azblob.SharedKeyCredential)
```
nopcoder, line 238: Can be done once at `NewAdater` and leave here just pointer check.

pkg/block/azure/adapter.go
```diff
@@ -196,6 +214,46 @@ func (a *Adapter) Get(ctx context.Context, obj block.ObjectPointer, _ int64) (io
...
+	sasQueryParams, err := vals.NewSASQueryParameters(keyCredentials)
+	if err != nil {
+		a.log(ctx).WithError(err).Error("could not generate SAS query parameters")
+		return "", err
+	}
+	// Create the URL of the resource you wish to access and append the SAS query parameters.
+	// Since this is a blob SAS, the URL is to the Azure storage blob.
+	qp := sasQueryParams.Encode()
+	return fmt.Sprintf(preSignedBlobPattern,
+		a.credentials, qualifiedKey.ContainerName, qualifiedKey.BlobURL, qp), nil
```
nopcoder, line 254: We depends on `credentials` - string format of a type we don't control? Just making sure we do not need to call any explicit function to format the string.

pkg/gateway/operations/mock_adapter_test.go
```diff
@@ -17,6 +17,11 @@ type mockAdapter struct {
 	lastStorageClass *string
 }
 
+func (a *mockAdapter) GetPreSignedURL(ctx context.Context, obj block.ObjectPointer, mode block.PreSignMode) (string, error) {
+	// TODO implement me
+	panic("implement me")
```
nopcoder, line 22: return not implemented error

nopcoder (changes requested):
Very cool! Added some comments.
And general question - do we want the pre sign duration configuration controlled in one location / settings?

pkg/block/azure/adapter.go
ozkatz: Good catch, it's supposed to be the account name extracted from the credentials! Fixed.

pkg/gateway/operations/mock_adapter_test.go
ozkatz: Done

pkg/block/gs/adapter.go
```diff
@@ -21,6 +21,8 @@ const (
 	delimiter    = "/"
 	partSuffix   = ".part_"
 	markerSuffix = ".multipart"
+
+	defaultPreSignedURLDuration = time.Minute * 15
```
itaiad200, line 25: Same here - if we define it equally in 2 different places, let's move it up the packages stack.

pkg/block/azure/adapter.go
ozkatz: Done

pkg/api/controller.go
ozkatz: Done

pkg/api/controller.go
ozkatz: Done

pkg/api/controller.go
ozkatz: Works great :) good call

pkg/block/s3/adapter.go
```diff
@@ -29,6 +29,9 @@ const (
 
 	// Chunks smaller than that are only allowed for the last chunk upload
 	minChunkSize = 8 * 1024
+
+	// DefaultPreSignedURLDuration determines how long should pre-signed URLs be valid for
+	DefaultPreSignedURLDuration = time.Minute * 15
```
itaiad200, line 34: Here too

esti/presign_test.go
ozkatz: done

api/swagger.yml
ozkatz: Even if not required, Go still treats it as a pointer. When removing the nullable flag, the Python SDK expects a string and crashes when receiving null..

esti/presign_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	"github.com/spf13/viper"
+	"github.com/treeverse/lakefs/pkg/block"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/treeverse/lakefs/pkg/api"
+)
+
+func TestPreSign(t *testing.T) {
```
itaiad200, line 13: Thanks for adding tests for the Stat operation! Can we also add tests to all other new operations?
Also - let's validate that the links work as expected too

api/swagger.yml
```diff
@@ -3286,6 +3318,11 @@ paths:
         schema:
           type: string
         description: path to the table data
+      - in: query
```
itaiad200, line 3321: Why do we need it in symlinks?

api/swagger.yml
```diff
@@ -3245,6 +3272,11 @@ paths:
         schema:
           type: boolean
           default: true
+      - in: query
```
itaiad200, line 3275: I think `ObjectStatsList` description should be modified for the double meaning of the property.

itaiad200 (dismissed):
Thanks - I think this can bring value to our users!

pkg/api/controller.go
```diff
@@ -3099,6 +3126,31 @@ func (c *Controller) ListObjects(w http.ResponseWriter, r *http.Request, reposit
...
+							Action:   permissions.ReadObjectAction,
+							Resource: permissions.ObjectArn(repository, entry.Path),
+						},
+					},
+				})
+				if c.handleAPIError(ctx, w, r, err) {
+					return
+				}
+				if authResponse.Allowed {
+					objStat.PhysicalAddress, err = c.BlockAdapter.GetPreSignedURL(ctx, block.ObjectPointer{
```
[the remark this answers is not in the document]
ozkatz: `GetPreSignedURL` Doesn't do any network calls. 
Pre-signed URLs are constructed locally by the adapter

api/swagger.yml
ozkatz: I added a description paragraph to the field to explain the expected results

api/swagger.yml
ozkatz: Good catch! We do not. removed

pkg/block/azure/adapter.go
```diff
@@ -29,21 +29,37 @@ const (
 	defaultMaxRetryRequests = 0
 	AuthMethodAccessKey     = "access-key"
 	AuthMethodMSI           = "msi"
+
+	defaultPreSignedURLDuration = time.Minute * 15
```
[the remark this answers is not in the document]
ozkatz: Good call - moved

pkg/block/azure/adapter.go
ozkatz: @itaiad200 you caught this too, changed :)

pkg/block/gs/adapter.go
ozkatz: moved

pkg/block/s3/adapter.go
ozkatz: moved

esti/presign_test.go
```diff
@@ -0,0 +1,47 @@
...
+	})
+	require.NoError(t, err, "failed to stat object with presign=true")
+	signedUrl := response.JSON200.PhysicalAddress
+	if !strings.HasPrefix(signedUrl, "http") {
+		t.Errorf("exected an HTTP(s) URL instead of an object store URI, got %s", signedUrl)
+	}
+
+	if !strings.Contains(signedUrl, expected) {
+		t.Errorf("expected a signed URL, got %s", signedUrl)
+	}
```
nopcoder, line 46:
```suggestion
	blockStoreType := viper.GetString("blockstore_type")
	expectedKey := ""
	switch blockStoreType {
	case block.BlockstoreTypeS3:
		expectedKey = "X-Amz-Signature"
	case block.BlockstoreTypeGS:
		expectedKey = "X-Goog-Signature"
	case block.BlockstoreTypeAzure:
		expectedKey = "sv"
	default:
		t.Skip("Only GS, S3 and Azure Blob supported for pre-signed urls")
	}

	ctx, _, repo := setupTest(t)
	defer tearDownTest(repo)

	_, _ = uploadFileRandomData(ctx, t, repo, mainBranch, "foo/bar", true)

	response, err := client.StatObjectWithResponse(ctx, repo, mainBranch, &api.StatObjectParams{
		Path:    "foo/bar",
		Presign: swag.Bool(true),
	})
	require.NoError(t, err, "failed to stat object with presign=true")
	require.NotNil(t, response.JSON200, "successful response")
	signedURL := response.JSON200.PhysicalAddress
	parsedSignedURL, err := url.Parse(signedURL)
	require.NoErrorf(t, err, "failed to parse url - %s", signedURL)
	require.Truef(t, strings.HasPrefix(parsedSignedURL.Scheme, "http"), "URL scheme http(s) - %s", signedURL)
	require.NotEmptyf(t, parsedSignedURL.Query().Get(expectedKey), "signature expected in key '%s' - %s", expectedKey, signedURL)
```

nopcoder (approved):
LGTM! - additional comments for your consideration

itaiad200:
@ozkatz Still missing tests for most of the APIs to get the links, including validations that links actually work.

ozkatz:
You're right @itaiad200 - Merged too early. Tests added in #5035 5035

PR #4986, opened 2023-01-09
Fix graveler to work with empty key

### LINKED ISSUE #3588: Import truncates keys, resulting in empty keys that yields error
If the given URL to import from is itself a file, e.g. `s3://bucket/some/path/to/key/ -> file`, lakeFS will truncate the given URL and will iterate over the keys past it (including it), so it will also try to store the empty key but will fail since pebble does not allow empty keys.
The returned error looks like this:
`writing range from entry iterator: writing record: setting key and value: pebble: keys must be added in order: #0,DEL, #0,SET`

### LINKED ISSUE #4909: [BUG] Error when importing file and directory that share a prefix
I am getting the following error when running a `lakectl import`:

```
writing range from entry iterator: writing record: setting key and value: pebble: keys must be added in order: helloworld/myfile.csv#0,SET, helloworld.csv#0,SET

500 Internal Server Error
```

Other imports I have done work perfectly fine and after some research I could find out the error only happens when I have a file and a directory that share the same prefix. 

I can replicate the error by creating this simple files setup in ADLS Gen 2 and importing it using `lakectl import`

```
-- somepath
    -- helloworld.csv
    -- helloworld
        -- myfile.csv

```

In this case, `somepath` contains a directory named `helloworld` and a file `helloworld.csv`, which somehow breaks the import. I found #3588 which gives a similar error and might be related

My setup is: 

- LakeFS 0.86.0 deployed in Azure Kubernetes with ADLS Gen2 as underlying file system
- lakectl 0.87.1 in Linux
- The source of my import is a different ADLS Gen2 container from the one LakeFS uses as underlying file system

nopcoder:
Allow empty key ("" / not nil) to be kept in Graveler. It fix issues related to import of directory marker into the root.
Note that this change upgrade Graveler's pebble implementation.

Fix #3588
Fix #4909


pkg/graveler/sstable/writer.go
```diff
@@ -138,10 +138,14 @@ func (dw *DiskWriter) Close() (*committed.WriteResult, error) {
...
 	// sstable.Writer constructor and cannot be changed, but we can replace its values
 	// before writing it out.
-	dw.SetMetadata(MetadataFirstKey, string(dw.first))
-	dw.SetMetadata(MetadataLastKey, string(dw.last))
-	dw.SetMetadata(MetadataNumRecordsKey, fmt.Sprint(dw.count))
-	dw.SetMetadata(MetadataEstimatedSizeKey, fmt.Sprint(dw.w.EstimatedSize()))
+	first := dw.first
+	last := dw.last
+	estimatedSize := dw.w.EstimatedSize()
+	count := dw.count
```
itaiad200, line 144: Why set these values now? Does the `Store`/`Close` change them?

pkg/graveler/sstable/range_manager.go
```diff
@@ -86,17 +86,20 @@ func (m *RangeManager) GetValueGE(ctx context.Context, ns committed.Namespace, i
 	defer m.execAndLog(ctx, it.Close, "close iterator")
 
 	// Ranges are keyed by MaxKey, seek to the range that might contain key.
-	key, value := it.SeekGE(lookup)
+	key, value := it.SeekGE(lookup, sstable.SeekGEFlags(0))
```
itaiad200, line 89: Here and elsewhere
```suggestion
	key, value := it.SeekGE(lookup, sstable.SeekGEFlagsNone)
```

itaiad200 (changes requested):
I have some concerns about backward compatibility, how can we make sure old sstables are respected?
Also - consider adding the empty directory marker to our esti import test to cover the bug fix.

pkg/graveler/sstable/writer.go
nopcoder: Yes - panic. The `EstimatedSize` calculate and access underlaying structure that we already closed.

pkg/graveler/sstable/range_manager.go
nopcoder: Can't access `SeekGEFlagsNone` it is part of the internal package.
A lot of the sstable implementation is inside `/internal` which we don't have access - unless an alias or equal value was re-defined under `sstable` package.

nopcoder:
> I have some concerns about backward compatibility, how can we make sure old sstables are respected?

Run latest release of lakeFS, uploaded, committed changes. Run the above version and it works. Assumed we had to read sstable with commit metadata.

> Also - consider adding the empty directory marker to our esti import test to cover the bug fix.

Will add soon - thanks.

itaiad200 (approved):
Thanks for adding the tests!

PR #4993, opened 2023-01-10
Display email when adding users to groups

### LINKED ISSUE #4991: Display email when adding users to groups and clear localstorage on login
When adding users to groups, IDs are always displayed, even when email are available.
This fix will display emails, if they are available.

Also, logging in should clear any residual values in localstorage.

eladlachmi:
Resolves #4991


guy-har (approved):
Nice Clean and Fast
Thanks!

PR #4995, opened 2023-01-10
Reduce the use of specific catalog errors to fix missing checks

### LINKED ISSUE #4984: UploadObject ifAbsent fails with "not found" / 500 ISE when it should succeed
# What

lakeFS fails "upload if absent" when it should succeed.

# How to reproduce

Either look at lakeFSFS in #4947 tests failing to mkdir, or apply [these patches]((https://github.com/treeverse/lakeFS/files/10374445/lakectl.if-absent.patch.gz)) and then try to upload to a non-existent path:
```sh
❯ echo foo | go run ./cmd/lakectl   -c ~/.lakectl.prod-e2e.yaml fs upload --if-absent -s - lakefs://ariels-repo/main/upload/3
not found
500 Internal Server Error
exit status 1
```

If the path _does_ exist, it gives a nice 412 error as it should:
```sh
❯ echo foo | go run ./cmd/lakectl   -c ~/.lakectl.prod-e2e.yaml fs upload --if-absent -s - lakefs://ariels-repo/main/upload/2
path already exists
412 Precondition Failed
exit status 1
```

# Context

Discovered as part of work on lakeFSFS in #4947.  It fails integration tests.  I believe that this is due to a bug in the lakeFS server, and indeed the above patches reproduce it without lakeFSFS.  Also "Internal Server Error" hints at an issue on the server.

nopcoder:
Catalog defined set of error that just map 1:1 errors found in Graveler.
In order to reduce bugs where we compare different type - reuse or map these errors with the ones found in the graveler  package.
+ Add test to make sure it will never happen again.

Fix #4984

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -199,7 +199,7 @@ func (c *Controller) DeleteObjects(w http.ResponseWriter, r *http.Request, body 
 		}
 		lg := c.Logger.WithField("path", objectPath)
 		switch {
-		case errors.Is(err, catalog.ErrNotFound), errors.Is(err, graveler.ErrNotFound):
+		case errors.Is(err, graveler.ErrNotFound):
 			lg.WithError(err).Debug("tried to delete a non-existent object")
 		case errors.Is(err, graveler.ErrWriteToProtectedBranch):
 			errs = append(errs, ObjectError{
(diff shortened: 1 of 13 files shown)


arielshaqed (approved):
It's definitely an improvement!

Unsure about whether this fix is the right way to go: just look at how many `ErrNotFound`s we still have!

In the past we generated all `ErrNotFound`s by wrapping a single base.  That let us have the best of both worlds: we could return package-specific errors (which are useful  because they do give more information -- see how many ErrNotFounds we still have!) but also have a single test.  I am not sure why we went back on this.  I think the current state is still brittle: I have little confidence that we actually cover all ErrNotFounds everywhere.

PR #4997, opened 2023-01-10
Fix: CopyBlob using wrong path format

### LINKED ISSUE #4996: Bug: CopyBlob using wrong path format
CopyBlob creates address using uuid package instead of using the PathProvider object

N-o-Z:
Closes #4996


nopcoder (approved):
Thanks!

PR #5005, opened 2023-01-10
Create CopyObject API

### LINKED ISSUE #5004: Create initial CopyObject API
Create a basic CopyObject API with naive implementation - to unblock tasks and split PR workload

N-o-Z:
Closes #5004


api/swagger.yml
```diff
@@ -274,6 +274,19 @@ components:
...
+      required:
+        - source
+        - destination
+      properties:
+        source:
+          type: string
+          description: path of the copied object relative to the branch
+        destination:
+          type: string
+          description: the new copy path relative to the branch
```
itaiad200, line 288: Why can't I use this API to copy to another branch?

api/swagger.yml
```diff
@@ -3215,6 +3228,45 @@ paths:
...
+      - in: path
+        name: repository
+        required: true
+        schema:
+          type: string
+      - in: path
+        name: branch
+        required: true
+        schema:
+          type: string
```
itaiad200, line 3242: Description here is required. Is this the source/destination repo & branch?

pkg/api/controller.go
```diff
@@ -2304,6 +2306,80 @@ func (c *Controller) StageObject(w http.ResponseWriter, r *http.Request, body St
...
+	}
+	ctx := r.Context()
+	c.LogAction(ctx, "copy_object", r, repository, branch, "")
+
+	repo, err := c.Catalog.GetRepository(ctx, repository)
+	if c.handleAPIError(ctx, w, r, err) {
+		return
+	}
+
+	// TODO (niro): Naive copy object flow - real implementation still missing
```
itaiad200, line 2337: Can you link to the followup issue?

pkg/api/controller.go
```diff
@@ -2304,6 +2306,80 @@ func (c *Controller) StageObject(w http.ResponseWriter, r *http.Request, body St
...
+	blob, err := upload.CopyBlob(ctx, c.BlockAdapter, repo.StorageNamespace, repo.StorageNamespace, src.PhysicalAddress, src.Checksum, destAddress, src.Size)
+	if c.handleAPIError(ctx, w, r, err) {
+		return
+	}
+
+	writeTime := time.Now()
+	entryBuilder := catalog.NewDBEntryBuilder().
+		CommonLevel(false).
+		Path(body.Destination).
+		PhysicalAddress(blob.PhysicalAddress).
```
itaiad200, line 2353: is `blob.PhysicalAddress` relative? Shouldn't we just use `destAddress`?

pkg/api/controller.go
```diff
@@ -2304,6 +2306,80 @@ func (c *Controller) StageObject(w http.ResponseWriter, r *http.Request, body St
...
+	err = c.Catalog.CreateEntry(ctx, repo.Name, branch, entry)
+	if c.handleAPIError(ctx, w, r, err) {
+		return
+	}
+	qk, err := block.ResolveNamespace(repo.StorageNamespace, blob.PhysicalAddress, block.IdentifierTypeRelative)
+	if err != nil {
+		writeError(w, r, http.StatusInternalServerError, err)
+		return
+	}
+	// TODO: end
```
itaiad200, line 2369: End what ;) ?

pkg/api/controller_test.go
```diff
@@ -3713,3 +3713,67 @@ func TestController_PostStatsEvents(t *testing.T) {
...
+	verifyResponseOK(t, uploadResp, err)
+
+	sizeBytes := api.Int64Value(uploadResp.JSON201.SizeBytes)
+
+	if sizeBytes != expectedSize {
+		t.Fatalf("expected %d bytes to be written, got back %d", expectedSize, sizeBytes)
+	}
+
+	t.Run("copy object", func(t *testing.T) {
+		time.Sleep(1 * time.Second)
```
itaiad200, line 3738: Why?

pkg/api/controller_test.go
```diff
@@ -3713,3 +3713,67 @@ func TestController_PostStatsEvents(t *testing.T) {
...
+	uploadResp, err := uploadObjectHelper(t, ctx, clt, "foo/bar", strings.NewReader(content), "repo1", "main")
+	verifyResponseOK(t, uploadResp, err)
+
+	sizeBytes := api.Int64Value(uploadResp.JSON201.SizeBytes)
+
+	if sizeBytes != expectedSize {
+		t.Fatalf("expected %d bytes to be written, got back %d", expectedSize, sizeBytes)
+	}
+
+	t.Run("copy object", func(t *testing.T) {
```
itaiad200, line 3737: suggestion: Test on the same/different branch - they both copy now, but soon that won't be the case :)

itaiad200 (changes requested):
Thanks! See my comments regarding the API definition around different branch usage

pkg/api/controller.go
N-o-Z: 1. yes
2. destAddress == blob.PhysicalAddress

pkg/api/controller.go
N-o-Z: End of TODO - how would you like me to make it more clear?

pkg/api/controller_test.go
N-o-Z: To validate different creation time

api/swagger.yml
N-o-Z: No where did we discuss about enabling copy between branches.
Do we really want to support this?

itaiad200: How can we not? The StageObject API does not contain this limitation right? Shouldn't we support the same use cases?
We can "really" copy if not the same branch, and "fancy-link" if it is. Also - the source may be any kind of ref, like tag or commit.
I understand this is just an initial PR focused on unblocking, so copying everything for now is fine.

N-o-Z: Decided to support only in-branch copy, and take cross-branch copy to discussion

pkg/api/controller_test.go
N-o-Z: When we will support cross branch copy - we will add the relevant tests

N-o-Z:
@nopcoder @arielshaqed would like your input here regarding the correct url path for the new API.
Optimally we would have wanted this to be under `/repositories/{repository}/branches/{branch}/objects` put command but this is already taken by getObject API so `/repositories/{repository}/refs/{ref}/objects/copy` seems like the next best thing, but if you have another idea I would love to hear

nopcoder:
> @nopcoder @arielshaqed would like your input here regarding the correct url path for the new API. Optimally we would have wanted this to be under `/repositories/{repository}/branches/{branch}/objects` put command but this is already taken by getObject API so `/repositories/{repository}/refs/{ref}/objects/copy` seems like the next best thing, but if you have another idea I would love to hear

The endpoint looks good to me - maybe I would make the target branch as optional as the default will be the same branch.

N-o-Z:
As @arielshaqed suggested replacing the url to `/repositories/{repository}/branches/{branch}/objects/copy` so it will be (A bit) more RESTful

pkg/api/controller_test.go
```diff
@@ -3713,3 +3713,74 @@ func TestController_PostStatsEvents(t *testing.T) {
 		})
 	}
 }
+
+func TestController_CopyObjectHandler(t *testing.T) {
```
itaiad200, line 3717: GR8 Tests!

itaiad200 (approved):
Thanks!!

pkg/api/controller.go
```diff
@@ -1818,6 +1818,8 @@ func (c *Controller) handleAPIErrorCallback(ctx context.Context, w http.Response
 	case errors.Is(err, graveler.ErrDirtyBranch),
 		errors.Is(err, graveler.ErrCommitMetaRangeDirtyBranch),
 		errors.Is(err, catalog.ErrNoDifferenceWasFound),
+		errors.Is(err, catalog.ErrInvalidValue),
```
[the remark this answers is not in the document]
N-o-Z: Only duplicates from graveler

pkg/api/controller_test.go
N-o-Z: 🎉

Issue #4974, opened 2023-01-09
Delta Diff MVP - Task: Add makefile target 


Jonathan-Rosenberg:
Add a makefile target specifically for the diff plugin. Don't execute it as part of the `all` target.

- [ ] Lean version
- [ ] Fully functional
- [ ] Tests

PR #5007, opened 2023-01-10
Delta Lake Diff MVP: Add Protobuf

### LINKED ISSUE #4963: Delta Diff MVP - Task: Define the Diff protobuf
Define the protobuf to be used for diffs. Keep monitoring in mind.

- [x] Lean version
- [x] Fully functional

Jonathan-Rosenberg:
Add the protobuf definition and the `protoc-gen-go` to generate the gRPC code.
The response has changed a bit to be more general than the [design](https://github.com/treeverse/lakeFS/blob/master/design/open/delta-diff.md#api).

Closes #4963


pkg/plugins/diff/diff.proto
```diff
@@ -0,0 +1,39 @@
...
+}
+
+message GatewayConfig {
+  string key = 1;
+  string secret = 2;
+  string endpoint = 3;
+}
+
+message DiffResponse {
+  message Diff {
```
nopcoder, line 28: can you move the message definition outside. it will make it simple type in the generated code.

pkg/plugins/diff/diff.proto
```diff
@@ -0,0 +1,39 @@
...
+    string version = 1;
+    int64 timestamp = 2;
```
nopcoder, line 30: use the timestamp google definition

pkg/plugins/diff/diff.proto
```diff
@@ -0,0 +1,39 @@
...
+
+message GatewayConfig {
+  string key = 1;
+  string secret = 2;
+  string endpoint = 3;
+}
+
+message DiffResponse {
+  message Diff {
+    string version = 1;
```
nopcoder, line 29: semver? asking because an integer will be easier to communicate with.

pkg/plugins/diff/diff.proto
```diff
@@ -0,0 +1,39 @@
...
+message DiffRequest {
+  TablePaths paths = 1;
+  GatewayConfig gateway_config = 2;
+}
+
+message TablePaths {
+  string left_table_path = 1;
+  string right_table_path = 2;
+  optional string base_table_path = 3;
+}
```
nopcoder, line 19: '_table_' looks redundant.
consider a comment about the path format.

pkg/plugins/diff/diff.proto
```diff
@@ -0,0 +1,39 @@
...
+syntax = "proto3";
+package diff;
+option go_package = "github.com/treeverse/lakefs/plugins/diff";
+
+message DiffRequest {
+  TablePaths paths = 1;
+  GatewayConfig gateway_config = 2;
+}
+
+message TablePaths {
```
nopcoder, line 15: push the definition up - before we use it.

nopcoder (changes requested):

nopcoder (commented):
Added comments - also missing makefile updates to generate and verify code.

pkg/plugins/diff/diff.proto
```diff
@@ -0,0 +1,39 @@
...
+  string secret = 2;
+  string endpoint = 3;
+}
+
+message DiffResponse {
+  message Diff {
+    string version = 1;
+    int64 timestamp = 2;
+    string description = 3;
+    map<string, string> content = 4;
```
arielshaqed, line 32: Can we define these fields here?  I _assume_ that they are particular fields from the Delta diff.  But that information is available only indirectly when reading lakeFS code.

arielshaqed (approved):
Small request for more field docs.  These docs can be links to DeltaIO diff documentation -- and that might be the **best** way to add them.

pkg/plugins/diff/diff.proto
Jonathan-Rosenberg: done

pkg/plugins/diff/diff.proto
Jonathan-Rosenberg: done

pkg/plugins/diff/diff.proto
Jonathan-Rosenberg: done

pkg/plugins/diff/diff.proto
Jonathan-Rosenberg: It's integer you're right

pkg/plugins/diff/diff.proto
Jonathan-Rosenberg: done

Jonathan-Rosenberg:
@nopcoder the makefile changes will be added on a later PR

pkg/plugins/diff/diff.proto
Jonathan-Rosenberg: This is an attempt to make the protobuf general enough to allow different types of diffs (iceberg, parquet...).
I don't want to limit it to DeltaLake diff only content

nopcoder:
> Added comments - also missing makefile updates to generate and verify code.

Cool - just remember the comment from the `proto` after you add the instructions to the Makefile.

nopcoder (approved):
LGTM - minor comment for a missing change.

Jonathan-Rosenberg:
@nopcoder it will happen in [#4974](https://github.com/treeverse/lakeFS/issues/4974).
THANKS

pkg/plugins/diff/diff.proto
```diff
@@ -0,0 +1,46 @@
...
+import "google/protobuf/timestamp.proto";
+
+/* Examples:
+    s3://repo/branch/path/to/dataset
+    lakefs://repo/branch/path/to/dataset
+ */
+message DiffPaths {
+  string left_path = 1;
+  string right_path = 2;
+  optional string base_table_path = 3;
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: done

PR #5015, opened 2023-01-10
Gateway copy object if object on different repository or branch

### LINKED ISSUE #4478: Modify Gateway CopyObject API
Prerequisite: [Implement CopyObject API](https://github.com/treeverse/lakeFS/issues/4477#top)

For copy operations on different branches, change CopyObject behavior to perform "full copy" using the underlying storage adapter's Copy method.

1. Copy object in the same branch will work the same - creating a new staging entry using the existing entry information.
2. For objects that are not part of the branch, use the underlying adapter copy operation.

nopcoder:
Close #4478

### DIFF
--- a/pkg/gateway/operations/putobject.go
+++ b/pkg/gateway/operations/putobject.go
@@ -5,7 +5,6 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/treeverse/lakefs/pkg/block"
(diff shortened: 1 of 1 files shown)


pkg/gateway/operations/putobject.go
```diff
@@ -144,18 +143,15 @@ func handleCopy(w http.ResponseWriter, req *http.Request, o *PathOperation, copy
 		_ = o.EncodeError(w, req, gatewayErrors.Codes.ToAPIErr(gatewayErrors.ErrInvalidCopySource))
 		return
 	}
+	// check if src and dst are in the same repository and branch
 	var ent *catalog.DBEntry
-	// check if src and dst are in the same repository
-	if strings.EqualFold(o.Repository.Name, p.Repo) {
+	if o.Repository.Name == p.Repo && o.Reference == p.Reference {
```
itaiad200, line 148: Stupid question, which was not introduced in this PR: Where do we validate that the reference is a branch and not a commit/tag?

itaiad200 (approved):
Thanks! Do we have some tests in place where it would be easy to add a check if an actual copy was performed? Also - Do we have a followup issue for the copy-table in the gateway context?

pkg/gateway/operations/putobject.go
nopcoder: Gateway Put handle check if branch/reference exists before branching out to "copy"

PR #5020, opened 2023-01-11
API to expose ugc grace period

### LINKED ISSUE #4990: Sync uncommitted GC client run time to the linked address tokens
Getting a physical address token is valid for 6 hours (LinkAddressTime). make sure the GC is not configured to run less than 6 hours so it won't delete the addresses that haven't been linked.

itaiad200:
closes #4990


pkg/api/controller.go
```diff
@@ -3651,6 +3652,19 @@ func (c *Controller) GetLakeFSVersion(w http.ResponseWriter, r *http.Request) {
...
+func (c *Controller) GetGarbageCollectionConfig(w http.ResponseWriter, r *http.Request) {
+	ctx := r.Context()
+	_, err := auth.GetUser(ctx)
+	if err != nil {
+		writeError(w, r, http.StatusUnauthorized, ErrAuthenticatingRequest)
+		return
+	}
+
+	writeResponse(w, r, http.StatusOK, GarbageCollectionConfig{
+		GracePeriod: aws.Int(int(ref.LinkAddressTime)),
```
N-o-Z, line 3664: Can you add lakeFS version as well

N-o-Z (changes requested):
One comment and missing test

pkg/api/controller.go
itaiad200: You have `/config/version` endpoint for that.

N-o-Z (approved):
LGTM

PR #5024, opened 2023-01-11
Delta Lake Diff MVP: Plugins manager

### LINKED ISSUE #4965: Delta Diff MVP - Task: Create a go-plugin client abstraction
Create a go-plugin client abstraction to use the underlying go-plugin client.

- [x] Lean version
- [x] Fully functional
- [x] Tests

Jonathan-Rosenberg:
This is a lean version of the plugin manager.
The plugin manager holds a mapping of plugin group types (diff, merge, rebase) to plugin types configurations.
It is used to initialize plugins and their clients.
An extra plugin wrapper struct will be used in every plugin type (wrapper) implementation to initialize a GRPC client and destroy the plugin itself.

### Tests
There are no tests. This is agreed upon and by design. Notice it's not merged into the master branch.
**Nothing is merged to master without tests.**

Closes #4965


pkg/plugins/plugins_manager.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+	clientConfig, ok := ptpp[pluginType]
+	if !ok {
+		return nil, fmt.Errorf("unknown plugin type %s under plugin group %s", pluginType, pluginGroupType)
+	}
+	return newPlugin(fmt.Sprintf("%s_%s", pluginGroupType, pluginType), clientConfig)
+}
+
+// newPlugin generates a PluginWrapper that wraps the go-plugin client.
+func newPlugin(clientName string, clientConfig plugin.ClientConfig) (*PluginWrapper, error) {
```
johnnyaug, line 132:
```suggestion
func newPluginWrapper(clientName string, clientConfig plugin.ClientConfig) (*PluginWrapper, error) {
```

pkg/plugins/plugins_manager.go
```diff
@@ -0,0 +1,163 @@
...
+}
+*/
+type PluginGroupTypeMap map[PluginGroup]PluginTypeConfigMap
+
+// The Manager holds the different types of plugins that can be used in the plugin system
+type Manager struct {
+	pluginGroups PluginGroupTypeMap
+}
+
+func (m *Manager) AddPluginType(pluginGroup PluginGroup, pluginType PluginType, clientConfig plugin.ClientConfig) {
```
johnnyaug, line 101: Shouldn't this already return a PluginWrapper for the plugin?
Instead of performing two subsequent calls.

pkg/plugins/plugins_manager.go
```diff
@@ -0,0 +1,163 @@
...
+				plugin.ProtocolGRPC,
+			},
+			Plugins: map[string]plugin.Plugin{
+				"icebergMerge": IcebergGRPCPlugin
+			},
+		}
+	},
+}
+*/
+type PluginGroupTypeMap map[PluginGroup]PluginTypeConfigMap
```
johnnyaug, line 94: Personal taste: don't really need the type definitions for the map types. Simple maps can work here, as these types are not exposed, if I understand correctly.

johnnyaug (changes requested):
That was fast!
Some comments, nothing major.

pkg/plugins/plugins_manager.go
```diff
@@ -0,0 +1,163 @@
+package plugins
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: I think that it's a convention to name the package as the direct directory it's in.
Do you think I should create a directory for it?

pkg/plugins/plugins_manager.go
```diff
@@ -0,0 +1,163 @@
...
+// PluginGroup specifies the group name of related plugins
+// For example, "diff" is a group of plugins that know how to perform a diff.
+type PluginGroup string
+
+// PluginType specifies a type of plugin
+// For example, "delta" is a plugin.
+type PluginType string
+
+// PluginIdentity identifies the plugin using a Key and Secret, and supplies the Version it expects the plugin to be of.
+type PluginIdentity struct {
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: I agree that it's currently the same, it's just that higher-level components shouldn't be coupled with the underlying plugin implementation. 
Can you suggest a way to decouple the layers differently?

pkg/plugins/plugins_manager.go
Jonathan-Rosenberg: Not exactly- I intend this method to add all available plugins in the startup of lakeFS to the Plugin manager, and not getting the PluginWrapper to use

pkg/plugins/plugins_manager.go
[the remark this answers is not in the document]
johnnyaug: I was talking about the filename: plugin_manager.go (without the _s_). Actually, even manager.go?
Not critical, obviously.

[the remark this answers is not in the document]
Jonathan-Rosenberg: oh my!
will do

pkg/plugins/plugins_manager.go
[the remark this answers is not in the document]
johnnyaug: It's not really a layer that you need to decouple, IMO. You can use the existing `plugin.Handshake` struct, like you're using `plugin.ClientConfig`.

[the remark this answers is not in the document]
Jonathan-Rosenberg: Yes I agree.
wait for it...

pkg/plugins/plugin_wrapper.go
```diff
@@ -0,0 +1,36 @@
...
+)
+
+// PluginWrapper wraps the go-plugin client (which is in fact a plugin controller) and the GRPC client.
+// It's used as a convenience struct to call the different types of plugins, and to be included as a subtype of specific
+// plugin client implementations
+type PluginWrapper struct {
+	GRPCClient *plugin.ClientProtocol
+	Client     *plugin.Client
+	Log        logging.Logger
+}
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: These will be used by specific GRPC clients (Delta for instance)

pkg/plugins/plugins_manager.go
Jonathan-Rosenberg: done

talSofer (approved):
Nice! added minor comments before you posted changes some are still relevant

johnnyaug (approved):
Cool!

pkg/plugins/manager.go
```diff
@@ -0,0 +1,170 @@
...
+// PluginIdentity identifies the plugin implementation.
+type PluginIdentity struct {
+	Impl               plugin.Plugin
+	ImplName           string
+	HandshakeConfig    plugin.HandshakeConfig
+	ExecutableLocation string
+}
+type pluginTypeConfigMap map[PluginName]plugin.ClientConfig
+
+// PluginTypeNameMap example:
```
arielshaqed, line 35: Can this be part of the godoc for the type?

pkg/plugins/manager.go
```diff
@@ -0,0 +1,170 @@
...
+
+	"github.com/hashicorp/go-hclog"
+	"github.com/hashicorp/go-plugin"
+	"github.com/treeverse/lakefs/pkg/logging"
+)
+
+var (
+	errPluginGroupNotFound = fmt.Errorf("unknown plugin group")
+	errPluginTypeNotFound  = fmt.Errorf("unknown plugin type")
+)
```
nopcoder, line 16: Returned from public func can we make them public too

pkg/plugins/manager.go
```diff
@@ -0,0 +1,170 @@
...
+// PluginType specifies the type of related plugins
+// For example, "diff" is a type of plugins that know how to perform a diff.
+type PluginType string
+
+// PluginName specifies the name of a plugin
+// For example, "delta" is a plugin.
+type PluginName string
+
+// PluginIdentity identifies the plugin implementation.
+type PluginIdentity struct {
```
nopcoder, line 27: Looks more than identity - it seems we capture the name, implementation and configuration.
Plugin ID can be a unique name or name+type - any reason to include everything?

pkg/plugins/manager.go
```diff
@@ -0,0 +1,170 @@
...
+}
+*/
+type PluginTypeNameMap map[PluginType]pluginTypeConfigMap
+
+// The Manager holds the different types of plugins that can be used in the plugin system
+type Manager struct {
+	pluginTypes PluginTypeNameMap
+}
+
+// WrapPlugin generates a Wrapper that wraps the go-plugin client.
```
nopcoder, line 107: Didn't understand from the description why we need a wrapper or what it is used for?
It seem like it should be called `NewPlugin` or `LoadPlugin`

pkg/plugins/manager.go
```diff
@@ -0,0 +1,170 @@
...
+// For example, "diff" is a type of plugins that know how to perform a diff.
+type PluginType string
+
+// PluginName specifies the name of a plugin
+// For example, "delta" is a plugin.
+type PluginName string
+
+// PluginIdentity identifies the plugin implementation.
+type PluginIdentity struct {
+	Impl               plugin.Plugin
```
arielshaqed, line 28: * IIUC this is not the implementation, this is the plugin for the implementation.  The implementation itself lives in a separate binary.  Consider calling it `Plugin`.
* Given that it has known input and output types, why do we need to configure this field?  I would expect the _same_ `plugin.Plugin` to work for all executables -- and then I'd expect it to be called `plugin` (or be entirely missing, or plugin_ if package name `plugin` cannot be re-used).

pkg/plugins/manager.go
```diff
@@ -0,0 +1,170 @@
...
+	errPluginTypeNotFound  = fmt.Errorf("unknown plugin type")
+)
+
+// PluginType specifies the type of related plugins
+// For example, "diff" is a type of plugins that know how to perform a diff.
+type PluginType string
+
+// PluginName specifies the name of a plugin
+// For example, "delta" is a plugin.
+type PluginName string
```
nopcoder, line 24: Not a fan of too many types - we load the information from the configuration as strings.  It means that we will need to cast each time to work with the types we map.

pkg/plugins/manager.go
```diff
@@ -0,0 +1,170 @@
...
+	pluginTypes PluginTypeNameMap
+}
+
+// WrapPlugin generates a Wrapper that wraps the go-plugin client.
+//
+// It accepts two parameters: the top PluginType and the required PluginName under it.
+// For example, PluginType = "diff", PluginName = "delta" will generate a Wrapper with a client that performs
+// diffs over Delta Lake tables.
+// It also returns a DestroyClientFunc
+func (m *Manager) WrapPlugin(pluginType PluginType, pluginName PluginName) (*Wrapper, error) {
```
nopcoder, line 113: These two args looks like the plugin identity.

pkg/plugins/manager.go
```diff
@@ -0,0 +1,170 @@
...
+// PluginName specifies the name of a plugin
+// For example, "delta" is a plugin.
+type PluginName string
+
+// PluginIdentity identifies the plugin implementation.
+type PluginIdentity struct {
+	Impl               plugin.Plugin
+	ImplName           string
+	HandshakeConfig    plugin.HandshakeConfig
+	ExecutableLocation string
```
arielshaqed, line 31: How is this resolved?  Is it by environment variable `PATH` (which may include working directory), or does it need to be a full path?

pkg/plugins/manager.go
```diff
@@ -0,0 +1,170 @@
...
+func (m *Manager) WrapPlugin(pluginType PluginType, pluginName PluginName) (*Wrapper, error) {
+	ptpp, ok := m.pluginTypes[pluginType]
+	if !ok {
+		return nil, errPluginGroupNotFound
+	}
+	clientConfig, ok := ptpp[pluginName]
+	if !ok {
+		return nil, errPluginTypeNotFound
+	}
+	return newPluginWrapper(fmt.Sprintf("%s_%s", pluginType, pluginName), clientConfig)
```
nopcoder, line 122: Consider using ':' usually it is not used as part of a name or type (or we can block it).
This way the may have less chance to name collide.

pkg/plugins/manager.go
```diff
@@ -0,0 +1,170 @@
...
+{
+	"diff": {
+		"delta": plugin.clientConfig{
+			HandshakeConfig: plugin.HandshakeConfig{
+				ProtocolVersion:  1,
+				MagicCookieKey:   "key",
+				MagicCookieValue: "value",
+			},
+			Cmd: exec.Command("path/to/executable/diff/delta"),
+			AllowedProtocols: []plugin.Protocol{
```
arielshaqed, line 46: `AllowedProtocols` has been deleted, I think.

pkg/plugins/manager.go
```diff
@@ -0,0 +1,170 @@
...
+				plugin.ProtocolGRPC,
+			},
+			Impl: map[string]plugin.Plugin{
```
arielshaqed, line 49: I think this type has also changed.

pkg/plugins/manager.go
```diff
@@ -0,0 +1,170 @@
...
+type Manager struct {
+	pluginTypes PluginTypeNameMap
+}
+
+// WrapPlugin generates a Wrapper that wraps the go-plugin client.
+//
+// It accepts two parameters: the top PluginType and the required PluginName under it.
+// For example, PluginType = "diff", PluginName = "delta" will generate a Wrapper with a client that performs
+// diffs over Delta Lake tables.
+// It also returns a DestroyClientFunc
```
arielshaqed, line 112: Can we rename Wrapper?  Similarly, not sure where DestroyClientFunc is returned.

pkg/plugins/manager.go
```diff
@@ -0,0 +1,170 @@
...
+func (m *Manager) WrapPlugin(pluginType PluginType, pluginName PluginName) (*Wrapper, error) {
+	ptpp, ok := m.pluginTypes[pluginType]
+	if !ok {
+		return nil, errPluginGroupNotFound
```
arielshaqed, line 116: This error is returned so it should probably be exported.

pkg/plugins/manager.go
nopcoder, line 35: Thought we have two sets of configurations - one specify the plugins+config. The second maps the operation to a plugin.

pkg/plugins/manager.go
```diff
@@ -0,0 +1,170 @@
...
+	return newPluginWrapper(fmt.Sprintf("%s_%s", pluginType, pluginName), clientConfig)
+}
+
+// newPluginWrapper generates a Wrapper that wraps the go-plugin client.
+func newPluginWrapper(clientName string, clientConfig plugin.ClientConfig) (*Wrapper, error) {
+	logger := hclog.New(&hclog.LoggerOptions{
+		Name:   fmt.Sprintf("%s_logger", clientName),
+		Output: os.Stdout,
+		Level:  hclog.Debug,
+	})
```
arielshaqed, line 131: I would prefer to wrap our logger here.  Otherwise it cannot be controlled.

arielshaqed (changes requested):
THANKS: First plugin interface to lakeFS -- something that's been a long way coming!

This is a bit confusing.  Most of the changes requested are documentation and naming, which I hope will unconfuse me.

pkg/plugins/manager.go
```diff
@@ -0,0 +1,170 @@
...
+	return &Wrapper{
+		Client: client,
+		Log:    logging.Default(),
+	}, nil
+}
+
+// RegisterPlugin is used to register a new plugin type with a plugin group. It can also introduce a new plugin group.
+func (m *Manager) RegisterPlugin(pluginType PluginType, pluginName PluginName, pluginID PluginIdentity) {
+	if m.pluginTypes == nil {
+		m.pluginTypes = make(PluginTypeNameMap)
```
nopcoder, line 143: Can be done once on NewManager

pkg/plugins/manager.go
```diff
@@ -0,0 +1,170 @@
...
+	handshakeConfig := identity.HandshakeConfig
+	allowedProtocols := []plugin.Protocol{
+		plugin.ProtocolGRPC,
+	}
+	return plugin.ClientConfig{
+		HandshakeConfig:  handshakeConfig,
+		Cmd:              exec.Command(identity.ExecutableLocation), //nolint:gosec
+		AllowedProtocols: allowedProtocols,
+		// "Plugins": the different types of plugins that the Plugin serves/consumes. Basically a plugin per
+		// communication type. There will be a single plugin which is the GRPC implementation of the Plugin.
```
nopcoder, line 165: Think the comment should move to the struct definition and use Plugins (instead of "Plugins") for codedoc.

pkg/plugins/manager.go
```diff
@@ -0,0 +1,170 @@
...
+	ptcm[pluginName] = clientConfig(pluginID)
+}
+
+// clientConfig generates a plugin.ClientConfig struct to be used to configure the Manager's plugins with the
+// corresponding configurations
+func clientConfig(identity PluginIdentity) plugin.ClientConfig {
+	handshakeConfig := identity.HandshakeConfig
+	allowedProtocols := []plugin.Protocol{
+		plugin.ProtocolGRPC,
+	}
```
nopcoder, line 159: Inline?

pkg/plugins/wrapper.go
```diff
@@ -0,0 +1,36 @@
...
+}
+
+// InitClient initializes the inner GRPC client to communicate to the
+func (c *Wrapper) InitClient() error {
+	if c.GRPCClient != nil {
+		return nil
+	}
+	grpcClient, err := c.Client.Client()
+	if err != nil {
+		c.Log.Fatal(err)
```
nopcoder, line 24: Remove the call for fatal - move to the caller if needed

pkg/plugins/wrapper.go
```diff
@@ -0,0 +1,36 @@
...
+	GRPCClient *plugin.ClientProtocol
+	Client     *plugin.Client
+	Log        logging.Logger
+}
+
+// InitClient initializes the inner GRPC client to communicate to the
+func (c *Wrapper) InitClient() error {
+	if c.GRPCClient != nil {
+		return nil
+	}
```
nopcoder, line 21: - Think it will be better to return an error as it will look like a bug calling init more than once. Or remove the check
- Wrote the first item as you also do not protect the Destroy (calling more than once) - why like that? ;)

nopcoder (changes requested):

pkg/plugins/manager.go
Jonathan-Rosenberg: full path

pkg/plugins/manager.go
Jonathan-Rosenberg: sometimes you're just right...

pkg/logging/logger.go
```diff
@@ -274,6 +278,22 @@ func (*logrusEntryWrapper) IsTracing() bool {
 	return defaultLogger.IsLevelEnabled(logrus.TraceLevel)
 }
 
+func (*logrusEntryWrapper) IsDebugging() bool {
+	return defaultLogger.IsLevelEnabled(logrus.DebugLevel)
+}
```
arielshaqed, line 283: All these `Is*` methods seem wrong, they should look something like
```go
func (l *logrusEntryWrapper) IsDebugging() bool {
	return l.e.Logger.IsLevelEnabled(logrus.DebugLevel)
}
```

Currently they may look at the wrong Logger.

@nopcoder: if this is true, separate issue for this perhaps?

pkg/plugins/client.go
```diff
@@ -0,0 +1,32 @@
...
+		client: client,
+		log:    log,
+	}
+}
+
+// Close terminates the plugin
+func (c Client) Close() error {
+	if c.client == nil {
+		return ErrClientAlreadyClosed
+	}
```
arielshaqed, line 29: Nit: Without locking this is thread-unsafe!  [Interface io.Closer](https://pkg.go.dev/io#Closer) says closing twice is undefined, and that implementations may document their own behaviour.  We should document that this Close may only be called once.  Additionally we would additionally ensure that it is thread-safe to rely on this error being returned.

pkg/plugins/manager.go
arielshaqed:
```suggestion
	// ExecutableLocation is the full path to the plugin executable on the local filesystem.
	ExecutableLocation string
```

pkg/plugins/manager.go
```diff
@@ -11,160 +12,104 @@ import (
...
-			Cmd: exec.Command("path/to/executable/merge/iceberg"),
-			AllowedProtocols: []plugin.Protocol{
-				plugin.ProtocolGRPC,
-			},
-			Impl: map[string]plugin.Plugin{
-				"icebergMerge": IcebergGRPCPlugin
-			},
-		}
-	},
+// PluginAuth includes authentication properties for the plugin.
```
arielshaqed, line 30: Can you expand?  It looks like a _single_ property, and it's not clear what the property means.  Maybe if the docs added what "authentication" we're performing it would be easier for me to understand.

pkg/plugins/manager.go
```diff
@@ -11,160 +12,104 @@ import (
...
+	Key   string
+	Value string
 }
-*/
-type PluginTypeNameMap map[PluginType]pluginTypeConfigMap
 
-// The Manager holds the different types of plugins that can be used in the plugin system
-type Manager struct {
-	pluginTypes PluginTypeNameMap
+// The Manager holds a map for different kinds of possible plugin clients.
```
arielshaqed, line 36:
```suggestion
// Manager maps the available plugin names to the different kinds of possible plugin clients.
```

pkg/plugins/manager.go
```diff
@@ -11,160 +12,104 @@ import (
...
+// For example, the clients map might contain a mapping of "delta" -> plugin.Client to communicate with the Delta
+// plugin.
+type Manager[T any] struct {
```
arielshaqed, line 39: What is T?  `plugin.Client` is not parametrized on it, which makes it less clear.

pkg/plugins/manager.go
```diff
@@ -11,160 +12,104 @@ import (
...
+
+// RegisterPlugin is used to register a new plugin client with the corresponding plugin type.
+func (m *Manager[T]) RegisterPlugin(name string, id PluginIdentity, auth PluginAuth, p plugin.Plugin) error {
+	if m == nil {
+		return ErrUninitializedManager
+	}
+	hc := plugin.HandshakeConfig{
+		ProtocolVersion:  uint(id.Version),
+		MagicCookieKey:   auth.Key,
+		MagicCookieValue: auth.Value,
```
arielshaqed, line 57: Not sure this is authentication; it looks like a magic cookie to me.

pkg/plugins/manager.go
```diff
@@ -11,160 +12,104 @@ import (
...
 	}
-	return newPluginWrapper(fmt.Sprintf("%s_%s", pluginType, pluginName), clientConfig)
+	m.clients[name] = c
+	return nil
 }
 
-// newPluginWrapper generates a Wrapper that wraps the go-plugin client.
-func newPluginWrapper(clientName string, clientConfig plugin.ClientConfig) (*Wrapper, error) {
-	logger := hclog.New(&hclog.LoggerOptions{
+func pluginClient(name string, p plugin.Plugin, hc plugin.HandshakeConfig, cmd *exec.Cmd) (*plugin.Client, error) {
```
arielshaqed, line 72: Having both `newPluginClient` and `pluginClient` confuses me.  Renaming or documentation will help me.

pkg/plugins/manager.go
```diff
@@ -11,160 +12,104 @@ import (
...
 		Output: os.Stdout,
 		Level:  hclog.Debug,
 	})
-	clientConfig.Logger = logger
-	client := plugin.NewClient(&clientConfig)
-	return &Wrapper{
-		Client: client,
-		Log:    logging.Default(),
-	}, nil
+	l := logging.Default()
```
arielshaqed, line 90: It would help me read logs if we added at least the plugin name as a field to the logger.  Similarly, we should output a log here giving _all_ the properties of the plugin: name, protocol, executable location, and that it's being used.

pkg/plugins/manager.go
```diff
@@ -11,160 +12,104 @@ import (
...
+		AllowedProtocols: allowedProtocols,
+		HandshakeConfig:  hc,
+		Cmd:              cmd,
+	}
+	return newPluginClient(name, clientConfig)
+}
+
+func newPluginClient(clientName string, clientConfig plugin.ClientConfig) (*plugin.Client, error) {
+	hl := hclog.New(&hclog.LoggerOptions{
 		Name:   fmt.Sprintf("%s_logger", clientName),
```
arielshaqed, line 86: The [documentation of "Name" suggests](https://pkg.go.dev/github.com/hashicorp/go-hclog#LoggerOptions) it should not end in "_logger".

pkg/plugins/manager.go
```diff
@@ -11,160 +12,104 @@ import (
...
-		// "Plugins": the different types of plugins that the Plugin serves/consumes. Basically a plugin per
-		// communication type. There will be a single plugin which is the GRPC implementation of the Plugin.
-		Plugins: map[string]plugin.Plugin{
-			identity.ImplName: identity.Impl,
-		},
+	stub, err := grpcClient.Dispense(name)
+	if err != nil {
+		return nil, err
 	}
+	cd, _ := stub.(T)
```
arielshaqed, line 113: IIUC we really _know_ the desired interface of the gRPC client: it has to support exactly the calls that lakeFS knows about and can perform.  So I don't understand the genericity: why not just cast to interface LakeFSOperatorsOnPluginer (possibly with a better name) here?

arielshaqed (changes requested):
Thanks!

Getting much clearer.  Only requested explanation is why not cast all plugins to a single interface, given that lakeFS knows what it will call.

pkg/plugins/client.go
Jonathan-Rosenberg: this code doesn't exist in the next PR

pkg/plugins/manager.go
Jonathan-Rosenberg: It's a go-plugin thing for insuring that the client and server of the plugin identify one another

pkg/plugins/manager.go
Jonathan-Rosenberg: Each gRPC client is configured with a specific interface to interact with the gRPC server, e.g. Diff interface, Merge interface, etc'
Each plugin manager will hold the interface type that its corresponding plugin types implement, e.g. `Manager[Diff]` will hold the Delta, Iceberg, and Parquet plugins to perform a diff respectively, all of them implementing the `Diff` interface.

PR #5027, opened 2023-01-12
Support ADLS Gen2

### LINKED ISSUE #5037: Support Azure Data Lake Gen2
Change underlying Azure SDK from azure-storage-blob-go to the newer azure-sdk-for-go which supports Hierarchical Namespace (HNS aka ADLS Gen2)

N-o-Z:
Closes #5037


N-o-Z:
PR ready to review.
Added an Esti Azure with HNS flavor job to the pipeline which runs our integration tests with ADLS Gen2.

@ozkatz adding you to reviewers as I had to make some changes to the pre signed URL code

pkg/block/azure/adapter.go
```diff
@@ -191,67 +178,56 @@ func (a *Adapter) Put(ctx context.Context, obj block.ObjectPointer, sizeBytes in
...
-	}
-	uploadOpts := a.translatePutOpts(ctx, opts)
-	uploadOpts.TransferManager = transferManager
-	defer transferManager.Close()
-	resp, err := copyFromReader(ctx, reader, blobURL, uploadOpts)
-	if err != nil {
-		return err
-	}
-	_ = resp == nil // this is done in order to ignore "result 0 is never used" error ( copyFromReader is copied from azure, and we want to keep it with minimum changes)
-	return nil
```
guy-har, line 210: 🙏 😄

pkg/block/azure/adapter.go
```diff
@@ -191,67 +178,56 @@ func (a *Adapter) Put(ctx context.Context, obj block.ObjectPointer, sizeBytes in
...
 		return "", err
 	}
-	// Create the URL of the resource you wish to access and append the SAS query parameters.
-	// Since this is a blob SAS, the URL is to the Azure storage blob.
-	qp := sasQueryParams.Encode()
+
+	qp, err := client.GetAccountSASURL(aztables.AccountSASResourceTypes{
+		Container: true,
+		Object:    true,
+	}, permissions, time.Now(), a.preSignedURLDurationGenerator())
```
guy-har, line 226: Not sure we should use time.Now() it seems like we pass a time that "IsZero" it isn't specified, which is what I believe we want in this case.

pkg/block/azure/adapter.go
```diff
@@ -293,26 +274,35 @@ func (a *Adapter) Walk(ctx context.Context, walkOpt block.WalkOpts, walkFn block
...
-	for marker := (azblob.Marker{}); marker.NotDone(); {
-		listBlob, err := containerURL.ListBlobsFlatSegment(ctx, marker, azblob.ListBlobsSegmentOptions{Prefix: qualifiedPrefix.Prefix})
-		if err != nil {
-			return err
-		}
-		marker = listBlob.NextMarker
-		for _, blobInfo := range listBlob.Segment.BlobItems {
-			if err := walkFn(blobInfo.Name); err != nil {
+		for listBlob.More() {
+			resp, err := listBlob.NextPage(context.Background())
```
guy-har, line 286:
```suggestion
			resp, err := listBlob.NextPage(ctx)
```

guy-har (changes requested):
Great addition, the new SDK seems much better than the previous one. Moving to the new SDK should even solve more issues other than support ADLS Gen2. Thanks!
Added some comments

pkg/block/azure/chunkwriting.go
```diff
@@ -8,46 +8,70 @@ import (
...
-		o:      o,
-		errCh:  make(chan error, 1),
+		ctx:     ctx,
+		cancel:  cancel,
+		reader:  from,
+		to:      to,
+		id:      newID(),
+		o:       o,
+		errCh:   make(chan error, 1),
+		buffers: buffers,
```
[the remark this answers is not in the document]
N-o-Z: The new implementation is using a tracker to track the parts which we are already doing in a different way to support multiple requests. I did use the new mmbPool struct they are using and modified the old code to work similarly to how they do it. I found it difficult to align to their implementation but if you have any suggestions - I would love to hear

pkg/block/azure/multipart_block_writer.go
```diff
@@ -20,46 +23,58 @@ import (
...
 	}
 
-	return &azblob.BlockBlobCommitBlockListResponse{}, err
+	return blockblob.CommitBlockListResponse{}, err
 }
 
-func completeMultipart(ctx context.Context, parts []block.MultipartPart, container azblob.ContainerURL, objName string, retryOptions azblob.RetryReaderOptions) (*block.CompleteMultiPartUploadResponse, error) {
+func (m *MultipartBlockWriter) Upload(_ context.Context, _ io.ReadSeekCloser, _ *blockblob.UploadOptions) (blockblob.UploadResponse, error) {
+	panic("Should not be called")
+}
```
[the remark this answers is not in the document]
N-o-Z: We needed to implement the interface - but we don't really need to use it.
I'm not sure what is the right implementation as I do not know how this will be used.

pkg/block/s3/adapter.go
```diff
@@ -69,12 +68,6 @@ type Adapter struct {
 	ServerSideEncryptionKmsKeyID string
 }
 
-func WithHTTPClient(c *http.Client) func(a *Adapter) {
-	return func(a *Adapter) {
-		a.httpClient = c
-	}
-}
-
```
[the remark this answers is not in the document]
N-o-Z: Unused

pkg/block/gs/adapter.go
```diff
@@ -37,12 +37,6 @@ type Adapter struct {
 	presignDurationGenerator func() time.Time
 }
 
-func WithPreSignedURLDurationGenerator(f func() time.Time) func(a *Adapter) {
-	return func(a *Adapter) {
-		a.presignDurationGenerator = f
-	}
-}
-
```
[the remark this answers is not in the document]
N-o-Z: unused

pkg/block/factory/build.go
```diff
@@ -150,28 +148,24 @@ func buildGSAdapter(ctx context.Context, params params.GS) (*gs.Adapter, error)
...
-	accountKey := params.StorageAccessKey
-	var credentials azblob.Credential
-	var err error
-	switch params.AuthMethod {
-	case azure.AuthMethodAccessKey:
-		credentials, err = azure.GetAccessKeyCredentials(accountName, accountKey)
-	case azure.AuthMethodMSI:
-		credentials, err = azure.GetMSICredentials()
-	default:
-		err = ErrAuthMethodNotSupported
```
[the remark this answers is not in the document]
N-o-Z: You are right - I originally added the code which checks for default credentials (CLI, ENV etc..)
Re-added it - let me know if it solves the issue

pkg/block/azure/credentials.go
```diff
@@ -1,45 +0,0 @@
...
-	}, nil
-}
-
-func GetAccessKeyCredentials(accountName, accountKey string) (azblob.Credential, error) {
-	if len(accountName) == 0 && len(accountKey) == 0 {
-		// fallback to Azure environment variables
-		accountName, accountKey = os.Getenv("AZURE_STORAGE_ACCOUNT"), os.Getenv("AZURE_STORAGE_ACCESS_KEY")
-	}
-	return azblob.NewSharedKeyCredential(accountName, accountKey)
-}
```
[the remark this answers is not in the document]
N-o-Z: Fixed

pkg/block/factory/build.go
```diff
@@ -150,28 +152,35 @@ func buildGSAdapter(ctx context.Context, params params.GS) (*gs.Adapter, error)
...
 	if err != nil {
 		return nil, nil, fmt.Errorf("invalid credentials: %w", err)
 	}
-	return azblob.NewPipeline(credentials, azblob.PipelineOptions{Retry: azblob.RetryOptions{TryTimeout: params.TryTimeout}}), credentials, nil
+	if defaultCreds == nil {
+		cred, err := service.NewSharedKeyCredential(params.StorageAccount, params.StorageAccessKey)
+		if err != nil {
+			return nil, nil, fmt.Errorf("invalid credentials: %w", err)
+		}
+		client, err = service.NewClientWithSharedKeyCredential(url, cred, &options)
```
nopcoder, line 180: The err is set to the scope of the if - need to fix the first assignment.

pkg/ingest/store/factory.go
```diff
@@ -133,24 +133,24 @@ func (f *walkerFactory) buildGCSWalker(ctx context.Context) (*gcsWalker, error)
...
 	} else {
 		var err error
-		p, err = getAzureClient()
+		c, err = getAzureClient()
 		if err != nil {
 			return nil, err
 		}
 	}
-	return NewAzureBlobWalker(p)
+	return NewAzureBlobWalker(*c)
```
nopcoder, line 153: I write the comment here but it is relevant for factory build also - we keep a copy of the client and not keeping the new client pointer.
Every time we create a client we copy the structure and not the pointer to the client we created.
In general if the function that creates the client can return an error which means that we keep a pointer (optional nil if we got an error) - I would keep that pointer and not copies.

pkg/ingest/store/factory.go
```diff
@@ -133,24 +133,24 @@ func (f *walkerFactory) buildGCSWalker(ctx context.Context) (*gcsWalker, error)
...
 		if err != nil {
 			return nil, err
 		}
 	} else {
 		var err error
-		p, err = getAzureClient()
+		c, err = getAzureClient()
 		if err != nil {
 			return nil, err
 		}
```
nopcoder, line 151: can extract out the `if err != nil` block

nopcoder (changes requested):

pkg/block/azure/adapter.go
```diff
@@ -386,21 +373,19 @@ func (a *Adapter) Copy(ctx context.Context, sourceObj, destinationObj block.Obje
...
-	// validate copy is not asynchronous
-	copyStatus := resp.CopyStatus()
-	if copyStatus == "pending" {
-		return ErrAsyncCopy
-	}
-	return nil
+
+	// TODO (niro): copy is limited to 256MB, should we handle it somehow?
+	_, err = destClient.CopyFromURL(ctx, sasKey, nil)
+	return err
```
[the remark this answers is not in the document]
N-o-Z: It was wrong to use StartCopyFromURL. We anyway error on async copy so we might as well get the error from the sdk and not create it

pkg/block/azure/chunkwriting.go
```diff
@@ -212,3 +262,94 @@ func (id *id) next() string {
 func (id *id) issued() []string {
 	return id.all
 }
+
+// Code taken from Azure SDK for go blockblob/chunkwriting.go
```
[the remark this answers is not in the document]
N-o-Z: Discussed also with @guy-har, we currently have a working solution for multipart upload.
I'll open a separate issue for this so we can improve the implementation and use the new sdk conventions

pkg/block/azure/adapter.go
```diff
@@ -190,46 +185,32 @@ func (a *Adapter) Get(ctx context.Context, obj block.ObjectPointer, _ int64) (io
...
-	if err != nil {
-		return "", err
-	}
-
-	qp, err := client.GetAccountSASURL(aztables.AccountSASResourceTypes{
-		Container: true,
-		Object:    true,
-	}, permissions, time.Now(), a.preSignedURLDurationGenerator())
+	blobURL := a.client.NewContainerClient(qualifiedKey.ContainerName).NewBlobClient(qualifiedKey.BlobURL)
+	u, err := blobURL.GetSASURL(permissions, time.Unix(0, 0), a.preSignedURLDurationGenerator())
```
guy-har, line 208:
```suggestion
	u, err := blobURL.GetSASURL(permissions, time.Time{}, a.preSignedURLDurationGenerator())
```
This should also work

pkg/block/factory/build.go
```diff
@@ -148,24 +149,27 @@ func buildGSAdapter(ctx context.Context, params params.GS) (*gs.Adapter, error)
...
+			return nil, fmt.Errorf("invalid credentials: %w", err)
+		}
+		return service.NewClientWithSharedKeyCredential(url, cred, &options)
 	}
-	keyCred, err := aztables.NewSharedKeyCredential(params.StorageAccount, params.StorageAccessKey)
+
+	defaultCreds, err := azidentity.NewDefaultAzureCredential(nil)
 	if err != nil {
-		return nil, nil, fmt.Errorf("invalid credentials: %w", err)
+		return nil, fmt.Errorf("invalid credentials: %w", err)
```
guy-har, line 172: Change error message

pkg/block/factory/build.go
```diff
@@ -148,24 +149,27 @@ func buildGSAdapter(ctx context.Context, params params.GS) (*gs.Adapter, error)
...
 }
 
-func BuildAzureServiceClient(params params.Azure) (*service.Client, *aztables.SharedKeyCredential, error) {
-	cred, err := azblob.NewSharedKeyCredential(params.StorageAccount, params.StorageAccessKey)
-	if err != nil {
-		return nil, nil, fmt.Errorf("invalid credentials: %w", err)
+func BuildAzureServiceClient(params params.Azure) (*service.Client, error) {
+	url := fmt.Sprintf(azure.AzURLTemplate, params.StorageAccount)
+	options := service.ClientOptions{ClientOptions: azcore.ClientOptions{Retry: policy.RetryOptions{TryTimeout: params.TryTimeout}}}
+	if params.StorageAccessKey != "" {
```
guy-har, line 162: We should change the documentation to reflect this

guy-har (approved):

pkg/block/factory/build.go
N-o-Z: Opened https://github.com/treeverse/lakeFS/issues/5044

nopcoder (approved):
LGTM! nice upgrade from the previous SDK.

PR #5028, opened 2023-01-12
Use copy instead of stage object in lakeFSFS rename

### LINKED ISSUE #4479: GC+: Modify lakeFSFS renameObject method
- Use CopyObject + DeleteObject when working via Gateway
- Use RenameObject when working with OpenAPI

- Prerequisite: #4477 

Opens:

- Any action required on clients?
- Do we need to make them version aware?
- Should we issue a notification to upgrade clients?

itaiad200:
close #4479 

Context: UGC milestone 2 aims to enable "safe" runs of the UGC. "safe" means that lakeFS is accepting writes and that they don't get accidentally picked up by UGC. To ensure that, we need to block `StageObject` from staging objects in the repo's storage namespace. The new `CopyObject` API will create the link (without physical copying) for uncommitted objects on the same branch, hence is equal to the current `StageObject` behaviour.

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -503,10 +503,6 @@ public class LakeFSFileSystem extends FileSystem {
                 return (FileNotFoundException) new FileNotFoundException(msg).initCause(e);
             case HttpStatus.SC_FORBIDDEN:
                 return (AccessDeniedException) new AccessDeniedException(msg).initCause(e);
-            case HttpStatus.SC_INTERNAL_SERVER_ERROR:
-                if (msg.contains("invalid API endpoint")){
-                    return (AccessDeniedException) new AccessDeniedException(msg).initCause(e);
-                }
             default:
                 return new IOException(msg, e);
         }
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -431,7 +431,6 @@ public class LakeFSFileSystem extends FileSystem {
      */
     private boolean fallbackToStage = false;
 
-
     /**
      * Non-atomic rename operation.
      *
@@ -447,21 +446,23 @@ public class LakeFSFileSystem extends FileSystem {
 
         ObjectsApi objects = lfsClient.getObjects();
         //TODO (Tals): Can we add metadata? we currently don't have an API to get the metadata of an object.
-        ObjectCopyCreation creationReq = new ObjectCopyCreation()
-                .srcRef(srcObjectLoc.getRef())
-                .srcPath(srcObjectLoc.getPath());
 
-        try {
-            objects.copyObject(dstObjectLoc.getRepository(), dstObjectLoc.getRef(), dstObjectLoc.getPath(),
-                    creationReq);
-        } catch (ApiException e) {
-            if (e.getCode() != HttpStatus.SC_INTERNAL_SERVER_ERROR ||
-                    !e.getMessage().contains("invalid API endpoint")){
-                throw translateException("renameObject: src:" + srcStatus.getPath() + ", dst: " + dst + ", failed to copy object", e);
+        if (!fallbackToStage) {
+            try {
+                ObjectCopyCreation creationReq = new ObjectCopyCreation()
+                        .srcRef(srcObjectLoc.getRef())
+                        .srcPath(srcObjectLoc.getPath());
+                objects.copyObject(dstObjectLoc.getRepository(), dstObjectLoc.getRef(), dstObjectLoc.getPath(),
+                        creationReq);
+            } catch (ApiException e) {
+                if (e.getCode() != HttpStatus.SC_INTERNAL_SERVER_ERROR ||
+                        !e.getMessage().contains("invalid API endpoint")) {
+                    throw translateException("renameObject: src:" + srcStatus.getPath() + ", dst: " + dst + ", failed to copy object", e);
+                }
+
+                LOG.warn("Copy API doesn't exist, falling back to stageObject");
+                fallbackToStage = true;
             }
-
-            LOG.warn("Copy API doesn't exist, falling back to stageObject");
-            fallbackToStage = true;
         }
 
         if (fallbackToStage){
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -426,6 +426,13 @@ public class LakeFSFileSystem extends FileSystem {
     }
 
     /**
+     * fallbackToStage determines whether the old StageObject API should be use,
+     * turn true when CopyObject API is not supported.
+     */
+    private boolean fallbackToStage = false;
+
+
+    /**
      * Non-atomic rename operation.
      *
      * @return true if rename succeeded, false otherwise
@@ -444,13 +451,12 @@ public class LakeFSFileSystem extends FileSystem {
                 .srcRef(srcObjectLoc.getRef())
                 .srcPath(srcObjectLoc.getPath());
 
-        Boolean fallbackToStage = false;
         try {
             objects.copyObject(dstObjectLoc.getRepository(), dstObjectLoc.getRef(), dstObjectLoc.getPath(),
                     creationReq);
         } catch (ApiException e) {
-            if (e.getCode() !=HttpStatus.SC_INTERNAL_SERVER_ERROR ||
-                    e.getMessage().contains("invalid API endpoint")){
+            if (e.getCode() != HttpStatus.SC_INTERNAL_SERVER_ERROR ||
+                    !e.getMessage().contains("invalid API endpoint")){
                 throw translateException("renameObject: src:" + srcStatus.getPath() + ", dst: " + dst + ", failed to copy object", e);
             }
 
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -444,11 +444,32 @@ public class LakeFSFileSystem extends FileSystem {
                 .srcRef(srcObjectLoc.getRef())
                 .srcPath(srcObjectLoc.getPath());
 
+        Boolean fallbackToStage = false;
         try {
             objects.copyObject(dstObjectLoc.getRepository(), dstObjectLoc.getRef(), dstObjectLoc.getPath(),
                     creationReq);
         } catch (ApiException e) {
-            throw translateException("renameObject: src:" + srcStatus.getPath() + ", dst: " + dst + ", failed to copy object", e);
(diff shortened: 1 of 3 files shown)


N-o-Z (commented):
@itaiad200 I think we should wait with this until the optimized copy is implemented - otherwise this could cause performance degradation for users who decide to upgrade their client

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
```diff
@@ -821,13 +821,13 @@ private void verifyObjDeletion(ObjectLocation srcObjLoc) throws ApiException {
     }
 
     private boolean dstPathLinkedToSrcPhysicalAddress(ObjectLocation srcObjLoc, ObjectLocation dstObjLoc) throws ApiException {
-        ArgumentCaptor<ObjectStageCreation> creationReqCapture = ArgumentCaptor.forClass(ObjectStageCreation.class);
-        verify(objectsApi).stageObject(eq(dstObjLoc.getRepository()), eq(dstObjLoc.getRef()), eq(dstObjLoc.getPath()),
+        ArgumentCaptor<ObjectCopyCreation> creationReqCapture = ArgumentCaptor.forClass(ObjectCopyCreation.class);
```
N-o-Z, line 824: We should have a fallback to stageobject in case the API doesn't exist

N-o-Z (changes requested):
Fallback to stageObject

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -440,16 +440,15 @@ private boolean renameObject(LakeFSFileStatus srcStatus, Path dst) throws IOExce
...
-                .checksum(srcStatus.getChecksum())
-                .sizeBytes(srcStatus.getLen())
-                .physicalAddress(srcStatus.getPhysicalAddress());
+        ObjectCopyCreation creationReq = new ObjectCopyCreation()
+                .srcRef(srcObjectLoc.getRef())
+                .srcPath(srcObjectLoc.getPath());
 
         try {
-            objects.stageObject(dstObjectLoc.getRepository(), dstObjectLoc.getRef(), dstObjectLoc.getPath(),
+            objects.copyObject(dstObjectLoc.getRepository(), dstObjectLoc.getRef(), dstObjectLoc.getPath(),
```
arielshaqed, line 448: As noted by @N-o-Z , this fails when the server is at an older version. So we need to catch the failure in that case and fall back to the old code.

arielshaqed (changes requested):

N-o-Z (approved):
Thanks Itai, looks good. Did you test the fallback scenario manually to see that it goes through the correct flow?

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -440,16 +440,36 @@ private boolean renameObject(LakeFSFileStatus srcStatus, Path dst) throws IOExce
...
+                .srcPath(srcObjectLoc.getPath());
 
+        Boolean fallbackToStage = false;
         try {
-            objects.stageObject(dstObjectLoc.getRepository(), dstObjectLoc.getRef(), dstObjectLoc.getPath(),
+            objects.copyObject(dstObjectLoc.getRepository(), dstObjectLoc.getRef(), dstObjectLoc.getPath(),
                     creationReq);
         } catch (ApiException e) {
-            throw translateException("renameObject: src:" + srcStatus.getPath() + ", dst: " + dst + ", failed to stage object", e);
+            if (e.getCode() !=HttpStatus.SC_INTERNAL_SERVER_ERROR ||
```
arielshaqed, line 452: Nit:
```suggestion
            if (e.getCode() != HttpStatus.SC_INTERNAL_SERVER_ERROR ||
```

arielshaqed, line 452: **NOT** nit: Is this the only way to detect that copy is not supported?  Because there are many possible causes for ISE, and we will double our API calls for _each one_.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -440,16 +440,36 @@ private boolean renameObject(LakeFSFileStatus srcStatus, Path dst) throws IOExce
...
         //TODO (Tals): Can we add metadata? we currently don't have an API to get the metadata of an object.
-        ObjectStageCreation creationReq = new ObjectStageCreation()
-                .checksum(srcStatus.getChecksum())
-                .sizeBytes(srcStatus.getLen())
-                .physicalAddress(srcStatus.getPhysicalAddress());
+        ObjectCopyCreation creationReq = new ObjectCopyCreation()
+                .srcRef(srcObjectLoc.getRef())
+                .srcPath(srcObjectLoc.getPath());
 
+        Boolean fallbackToStage = false;
```
arielshaqed, line 447: I don't think this needs to be boxed; we can use
```suggestion
        boolean fallbackToStage = false;
```

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -440,16 +440,36 @@ private boolean renameObject(LakeFSFileStatus srcStatus, Path dst) throws IOExce
...
                     creationReq);
         } catch (ApiException e) {
-            throw translateException("renameObject: src:" + srcStatus.getPath() + ", dst: " + dst + ", failed to stage object", e);
+            if (e.getCode() !=HttpStatus.SC_INTERNAL_SERVER_ERROR ||
+                    e.getMessage().contains("invalid API endpoint")){
+                throw translateException("renameObject: src:" + srcStatus.getPath() + ", dst: " + dst + ", failed to copy object", e);
+            }
+
+            LOG.warn("Copy API doesn't exist, falling back to stageObject");
+            fallbackToStage = true;
```
arielshaqed, line 458: This **must** be stored for future calls.  Otherwise whenever we run a new client with an old version, we will double the number of API calls.  That gives poor performance, spams logs on client and on server, and eats away from any API call quota.

arielshaqed (changes requested):
Thanks!

Requested changes:
* If we've detected copy is not supported, don't try it again.
* Use something more accurate than ISE to detect using the copy API.  ISEs can be caused by many things.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
arielshaqed: The full response on failure is:
```
❯ http get https://treeverse.us-east-1.lakefscloud.io/api/v1/a/b/c
HTTP/1.1 500 Internal Server Error
Connection: keep-alive
Content-Length: 35
Content-Type: application/json
Date: Sun, 15 Jan 2023 07:59:26 GMT
X-Content-Type-Options: nosniff

{
    "message": "invalid API endpoint"
}
```

Is "invalid API endpoint" on the JSON message a part of the API?  (Even if it isn't, it won't change on _old_ server versions, so it's probably safe to use!)

itaiad200:
> Thanks Itai, looks good. Did you test the fallback scenario manually to see that it goes through the correct flow?

Better - added a test to check it.

itaiad200 (commented):
Added a test and fixed comments, PTAL!

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -476,6 +502,10 @@ private IOException translateException(String msg, ApiException e) {
                 return (FileNotFoundException) new FileNotFoundException(msg).initCause(e);
             case HttpStatus.SC_FORBIDDEN:
                 return (AccessDeniedException) new AccessDeniedException(msg).initCause(e);
+            case HttpStatus.SC_INTERNAL_SERVER_ERROR:
+                if (msg.contains("invalid API endpoint")){
```
arielshaqed, line 506: I don't understand, why make this an AccessDeniedException?

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -476,6 +502,10 @@ private IOException translateException(String msg, ApiException e) {
...
+                    return (AccessDeniedException) new AccessDeniedException(msg).initCause(e);
```
arielshaqed, line 507: Cast does nothing here.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -440,16 +447,35 @@ private boolean renameObject(LakeFSFileStatus srcStatus, Path dst) throws IOExce
...
-                .checksum(srcStatus.getChecksum())
-                .sizeBytes(srcStatus.getLen())
-                .physicalAddress(srcStatus.getPhysicalAddress());
+        ObjectCopyCreation creationReq = new ObjectCopyCreation()
+                .srcRef(srcObjectLoc.getRef())
+                .srcPath(srcObjectLoc.getPath());
 
         try {
-            objects.stageObject(dstObjectLoc.getRepository(), dstObjectLoc.getRef(), dstObjectLoc.getPath(),
+            objects.copyObject(dstObjectLoc.getRepository(), dstObjectLoc.getRef(), dstObjectLoc.getPath(),
```
arielshaqed, line 455: But this is always tried, even after it's already failed.  That's still 2 API calls each time.

arielshaqed (changes requested):
:-(

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
itaiad200: 🤦

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
itaiad200: Mistake.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
itaiad200: Removed.

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -440,16 +446,37 @@ private boolean renameObject(LakeFSFileStatus srcStatus, Path dst) throws IOExce
...
+                        !e.getMessage().contains("invalid API endpoint")) {
+                    throw translateException("renameObject: src:" + srcStatus.getPath() + ", dst: " + dst + ", failed to copy object", e);
+                }
+
+                LOG.warn("Copy API doesn't exist, falling back to stageObject");
+                fallbackToStage = true;
+            }
+        }
+
+        if (fallbackToStage){
```
arielshaqed, line 468: Nit:
```suggestion
        if (fallbackToStage) {
```

arielshaqed (approved):
Great, thanks!

One really cool thing this opens is _we don't need to statObject on S3_ to rename!  This will matter in future, and also we might be able to improve performance here in this case.  I'll open an issue after you pull this -- we're now so far beyond developing lakeFS server it's not fun or funny.

PR #5046, opened 2023-01-16
Stats collector handle post with retry and graceful shutdown

### LINKED ISSUE #5048: Improve stats posting
Handle posting stats better by using retry. Make sure we post data while we shutdown in order to prevent stats lost.

nopcoder:
Close #5048

- retry on client error while sending metrics (configured to have 2 more tries)
- remove low timeout while posting events.
- sending metrics will not abort on cancellation.
- separate background processing of metrics and one sending metrics.
- shutdown using context cancellation
- first audit check done in the background

### DIFF
--- a/pkg/stats/collector.go
+++ b/pkg/stats/collector.go
@@ -103,7 +103,7 @@ func (t *TimeTicker) Tick() <-chan time.Time {
 
 type BufferedCollector struct {
 	cache            keyIndex
-	writes           chan Metric
+	updatesCh        chan Metric
 	sender           Sender
 	flushTicker      FlushTicker
 	flushSize        int
@@ -112,20 +112,22 @@ type BufferedCollector struct {
 	processID        string
 	runtimeCollector func() map[string]string
 	runtimeStats     map[string]string
-	pendingWrites    sync.WaitGroup
+	inFlight         sync.WaitGroup
 	pendingRequests  sync.WaitGroup
-	ctxCancelled     int32
+	closed           int32
 	done             chan bool
-	runCalled        int32
 	extended         bool
 	log              logging.Logger
+	wg               sync.WaitGroup
+	cancel           context.CancelFunc
+	sendCh           chan *InputEvent
 }
 
 type BufferedCollectorOpts func(s *BufferedCollector)
 
 func WithWriteBufferSize(bufferSize int) BufferedCollectorOpts {
 	return func(s *BufferedCollector) {
-		s.writes = make(chan Metric, bufferSize)
+		s.updatesCh = make(chan Metric, bufferSize)
 	}
 }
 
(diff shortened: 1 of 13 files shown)


pkg/stats/collector.go
```diff
@@ -104,30 +103,29 @@ func (t *TimeTicker) Tick() <-chan time.Time {
...
-	ctxCancelled     int32
-	done             chan bool
-	runCalled        int32
+	inFlight         sync.WaitGroup
+	closed           int32
 	extended         bool
 	log              logging.Logger
+	wg               sync.WaitGroup
+	cancel           context.CancelFunc
+	sendCh           chan *InputEvent
```
itaiad200, line 121: Seems like we added or replaced many synchronization protocols. Can you document each one for the next developer?

pkg/stats/collector.go
```diff
@@ -173,20 +171,18 @@ func NewBufferedCollector(installationID string, cfg Config, opts ...BufferedCol
...
-		pendingWrites:   sync.WaitGroup{},
-		pendingRequests: sync.WaitGroup{},
-		ctxCancelled:    0,
-		done:            make(chan bool),
-		sendTimeout:     sendTimeout,
+		inFlight:        sync.WaitGroup{},
+		closed:          0,
 		extended:        cfg.Extended,
 		log:             logging.Default(),
+		sendCh:          make(chan *InputEvent, 1), // buffered as like to check if sender is free
```
itaiad200, line 185: Didn't understand your comment.

pkg/stats/collector.go
```diff
@@ -211,154 +207,187 @@ func (s *BufferedCollector) getInstallationID() string {
...
+func (s *BufferedCollector) processSendCh(ctx context.Context) {
+	for event := range s.sendCh {
+		s.sendEventHelper(ctx, event)
+	}
+}
 
-	// wait for http requests to complete
-	s.pendingRequests.Wait()
+func (s *BufferedCollector) sendEventHelper(ctx context.Context, event *InputEvent) {
+	err := s.sender.SendEvent(ctx, event)
```
itaiad200, line 358: pass context.Background to avoid throwing away requests.

pkg/stats/collector.go
```diff
@@ -211,154 +207,187 @@ func (s *BufferedCollector) getInstallationID() string {
...
+	// stop main loop processing
+	s.cancel()
+	s.wg.Wait()
 
-	// drain writes
-	close(s.writes)
-	for w := range s.writes {
+	// drain events
+	ctx := context.Background()
+	s.processSendCh(ctx)
```
itaiad200, line 339: redundant, that function is self draining

pkg/stats/collector.go
```diff
@@ -211,154 +207,187 @@ func (s *BufferedCollector) getInstallationID() string {
...
-	if atomic.LoadInt32(&s.runCalled) == 0 {
-		// nothing to do
+	// mark as closed
+	if !atomic.CompareAndSwapInt32(&s.closed, 0, 1) {
 		return
 	}
 
-	// wait for main loop to exit
-	<-s.done
+	// wait until no more in flight requests
```
itaiad200, line 325: What if there's always more incoming requests?

itaiad200 (approved):
Hopefully you found it 🤞

itaiad200 (approved):

pkg/stats/collector.go
nopcoder: fixed

pkg/stats/collector.go
nopcoder: single item channel to handle one post at a time

pkg/stats/collector.go
nopcoder: fixed

pkg/stats/collector.go
nopcoder: added comments

pkg/stats/collector.go
nopcoder: we set the 'stop' flag so no new request will send anything on this channel

PR #5049, opened 2023-01-16
Delta Diff MVP: Diff Service

### LINKED ISSUE #4964: Delta Diff MVP - Task: Create a Diff Service for OTFs
Add a new Diff service for handling diff requests for OTFs. 

- [ ] Lean version
- [ ] Fully functional
- [ ] Tests

Jonathan-Rosenberg:
This PR includes the Diff service implementation and its initialization in the controller.

## Notice
The implementation in the controller is currently hard-coded! This is not the final version and will be revisited in a later PR

Closes #4964


johnnyaug (approved):
Some comments, nothing major

talSofer (approved):

cmd/lakefs/cmd/run.go
```diff
@@ -531,6 +536,17 @@ func enableKVParamsMetrics(p params.Config) params.Config {
 	return p
 }
 
+func initOTFService() *diff.Service {
+	pluginsManager := plugins.NewManager[diff.Differ]()
+	pid := plugins.PluginIdentity{Version: 0, ExecutableLocation: "~/.lakefs/plugins/delta"}
+	pa := plugins.PluginAuth{Key: "deltaKey", Value: "7a44968c-5076-4a90-9a0e-0249cc6c119a"}
```
arielshaqed, line 542: 1. Please add a comment how to find / generate this value.
1. Could it be just a readable string such as "kind: diff; type: delta; version: 0.0.1" (I don't have any particular preference for the format to use!)?

cmd/lakefs/cmd/run.go
```diff
@@ -531,6 +536,17 @@ func enableKVParamsMetrics(p params.Config) params.Config {
 	return p
 }
 
+func initOTFService() *diff.Service {
+	pluginsManager := plugins.NewManager[diff.Differ]()
+	pid := plugins.PluginIdentity{Version: 0, ExecutableLocation: "~/.lakefs/plugins/delta"}
```
arielshaqed, line 541: We documented `ExecutableLocation` as a full path.  So you'll need to call some function to expand the "~" here.  Also, this must be configurable: it may be hard / inconvenient to write the home directory in some setups.  In a containerized setting I would probably prefer to mount it into someplace such as /etc/lakefs/plugins/delta.

pkg/api/controller.go
```diff
@@ -96,6 +98,7 @@ type Controller struct {
 	sessionStore          sessions.Store
 	oidcAuthenticator     *oidc.Authenticator
 	PathProvider          upload.PathProvider
+	OTFDiffService        *diff.Service
```
arielshaqed, line 101: Nit: Why is this exported?

pkg/plugins/diff/service.go
```diff
@@ -0,0 +1,81 @@
...
+	Secret   string
+	Endpoint string
+}
+
+type Params struct {
+	TablePaths TablePaths
+	S3Creds    S3Creds
+}
+
+type Entry struct {
```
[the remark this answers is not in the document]
arielshaqed: Please don't.  This is already `diff.Entry`.

pkg/plugins/diff/service.go
```diff
@@ -0,0 +1,81 @@
...
+	Version          string
+	Timestamp        time.Time
+	Operation        string
+	OperationContent map[string]string
```
arielshaqed, line 41: Please document these fields.

Also, can we make Operation a special type, with a closed list of known operations?

pkg/plugins/diff/service.go
```diff
@@ -0,0 +1,81 @@
...
+	defer closeClient()
+	diffs, err := (*d).Diff(ctx, diffParams.TablePaths, diffParams.S3Creds)
+	if err != nil {
+		return nil, err
+	}
+	return buildEntries(diffs), nil
+}
+
+func buildEntries(diffs []*Diff) []Entry {
+	result := make([]Entry, len(diffs))
```
arielshaqed, line 71: I believe this is incorrect: it will give `len(result) == 2*len(diffs)`.  You probably need something like
```suggestion
	result := make([]Entry, 0, len(diffs))
```
which makes `result` a slice of length `0` with capacity `len(diffs).`

pkg/plugins/diff/service.go
```diff
@@ -0,0 +1,81 @@
...
+	for _, diff := range diffs {
+		result = append(result, Entry{
+			Version:          strconv.Itoa(int(diff.Version)),
```
arielshaqed, line 74: This is scary: `diff.Version` is not an int, and we will lose some of its digits?  Let's pick one type for version and use that!

arielshaqed (changes requested):
Thanks!

Blocking on:
* Configurable ExecutableLocation and calling a tilde expansion function.
* Strange versions type in Diff.

pkg/plugins/diff/diff_grpc_client.go
```diff
@@ -0,0 +1,49 @@
+package diff
+
+import (
+	"context"
+
+	"github.com/hashicorp/go-plugin"
+	"google.golang.org/grpc"
+)
+
+type DeltaDiffGRPCClient struct {
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: basically not, but this struct implements the [`Differ` interface](https://github.com/treeverse/lakeFS/pull/5049/files#diff-ad1e64f96e7ccb094a477097648805062420a0b980587acfb98745e387a6c02cR12), which in the future might be implemented differently for different diff types.
I'll change it to be a general impl for now

cmd/lakefs/cmd/run.go
```diff
@@ -524,6 +528,17 @@ func enableKVParamsMetrics(p params.Config) params.Config {
 	return p
 }
 
+func initOTFService() *diff.Service {
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: Yes, it's 1... Sorry for the name mismatch

cmd/lakefs/cmd/run.go
Jonathan-Rosenberg: 1. Adding the location to the design. Will be updated in the final version to have 
2. It would include data duplication if it would be of this format.

cmd/lakefs/cmd/run.go
Jonathan-Rosenberg: This is configurable. Haven't been implemented obviously...

pkg/api/controller.go
Jonathan-Rosenberg: fixing

pkg/plugins/diff/service.go
Jonathan-Rosenberg: Why is twice the size of the diffs?

PR #5060, opened 2023-01-18
UI: allow filtering tags and branches by prefix

ozkatz:
Allow filtering by prefix in the Tags and Branches pages!

### DIFF
--- a/webui/src/lib/components/controls.jsx
+++ b/webui/src/lib/components/controls.jsx
@@ -286,12 +286,12 @@ export const PrefixSearchWidget = ({ onFilter, text = "Search by Prefix", defaul
         })
     }, [setExpanded])
 
-    const handleSubmit =(e) => {
+    const ref = useRef(null);
+
+    const handleSubmit = useCallback((e) => {
         e.preventDefault()
         onFilter(ref.current.value)
-    }
-
-    const ref = useRef(null);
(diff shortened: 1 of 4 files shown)


webui/src/lib/components/controls.jsx
```diff
@@ -274,6 +275,56 @@ export const ClipboardButton = ({ text, variant, onSuccess, icon = <PasteIcon/>,
...
+    const [expanded, setExpanded] = useState(!!defaultValue)
+
+    const toggle = useCallback((e) => {
+        e.preventDefault()
+        setExpanded((prev) => {
+            return !prev
+        })
+    }, [setExpanded])
+
+    const handleSubmit =(e) => {
```
eladlachmi, line 289: Why not add a `useCallback` here? The idea of `useCallback` is to memoize the function instance/reference so that it doesn't change between renders. If you pass the function instance as a prop and the ref changes, the component will re-render. Usually, React does this fast enough, but these things tend to pile on.

eladlachmi (approved):
One minor comment. Not a blocker. Other than that, LGTM.

webui/src/lib/components/controls.jsx
ozkatz: Right! Added

PR #5062, opened 2023-01-18
CopyObject api and gateway using copy-table tracking

### LINKED ISSUE #4477: Implement CopyObject API
Clients working through the S3 Gateway can use the CopyObject + DeleteObject to perform a Rename or Move operation.
For clients using the OpenAPI this could have been done using StageObject + DeleteObject.
To continue support of this operation, introduce a new API to Copy an object

**Input:** branch_id, old_path, new_path

**Flow:**
- Get branch staging token
- Get entry from branch
- If entry committed - perform a full copy object operation
- If entry is staged - Stage new entry using old entry data and new path
- Retry if staging token changed

nopcoder:
Close #4477


pkg/catalog/catalog.go
```diff
@@ -1923,6 +1930,28 @@ func (c *Catalog) PrepareGCUncommitted(ctx context.Context, repositoryID string,
...
+//
+//	time based order to later delete old ones
+func (c *Catalog) TrackPhysicalAddress(ctx context.Context, repository, physicalAddress string) (string, error) {
+	repo, err := c.getRepository(ctx, repository)
+	if err != nil {
+		return "", err
+	}
+	repoPartition := graveler.RepoPartition(repo)
+	id := xid.New().String()
+	key := kvTrackPrefix + id
```
itaiad200, line 1943: Why can't the physical address (prefixed by the repo) be the key?

pkg/catalog/catalog.go
```diff
@@ -1940,29 +1969,116 @@ func (c *Catalog) VerifyLinkAddress(ctx context.Context, repository, token strin
...
+		if err := c.deleteRepoTrackedPhysicalAddresses(ctx, repo); err != nil {
+			c.log.WithError(err).
+				WithField("repository", repo.RepositoryID).
+				Warn("Repository delete tracked physical addresses failed")
+		}
+	}
+}
+
+func (c *Catalog) deleteRepoTrackedPhysicalAddresses(ctx context.Context, repo *graveler.RepositoryRecord) error {
+	// collect and delete old keys - work in 'kvTrackDeleteBatchSize' and interval of 5 min between if batch is bigger
```
itaiad200, line 2020: You can skip the optimizations for kv calls at the moment, we need to handle this as part of the bigger picture. This is not the only place where this pose a problem.

pkg/catalog/catalog.go
```diff
@@ -1940,29 +1969,116 @@ func (c *Catalog) VerifyLinkAddress(ctx context.Context, repository, token strin
...
+			return ctx.Err()
+		}
+	}
+	return nil
+}
+
+func (c *Catalog) deleteRepoTrackedPhysicalAddressHelper(ctx context.Context, repo *graveler.RepositoryRecord) (bool, error) {
+	repoPartition := graveler.RepoPartition(repo)
+	msgType := (&Entry{}).ProtoReflect().Type()
+	it, err := kv.NewPrimaryIterator(ctx, c.kvStore.Store, msgType, repoPartition, []byte(kvTrackPrefix), kv.IteratorOptionsFrom([]byte{}))
```
itaiad200, line 2043: You're always starting from the top. 
Every 2 seconds you will be getting the same results which should not be deleted, it's a waste in kv calls

itaiad200: Oh, is that the reason for the sorted keys?

pkg/api/controller.go
```diff
@@ -2334,85 +2338,156 @@ func (c *Controller) StageObject(w http.ResponseWriter, r *http.Request, body St
...
+func (c *Controller) copyObjectShallow(ctx context.Context, repository, branch string, srcEntry *catalog.DBEntry, destPath string) (*catalog.DBEntry, error) {
+	// track physical address (copy-table)
+	_, err := c.Catalog.TrackPhysicalAddress(ctx, repository, srcEntry.PhysicalAddress)
+	if err != nil {
+		return nil, err
+	}
+
+	// create entry (copy) - try only once
+	dstEntry := *srcEntry
+	dstEntry.CreationDate = time.Now().UTC()
```
itaiad200, line 2450: Isn't the `.UTC()` redundant?

pkg/api/controller.go
```diff
@@ -2334,85 +2338,156 @@ func (c *Controller) StageObject(w http.ResponseWriter, r *http.Request, body St
...
+	dstEntry.Path = destPath
+	err = c.Catalog.CreateEntry(ctx, repository, branch, dstEntry, graveler.WithMaxTries(1))
```
itaiad200, line 2452: Can you document the `1` try limitation?

pkg/api/controller.go
```diff
@@ -2334,85 +2338,156 @@ func (c *Controller) StageObject(w http.ResponseWriter, r *http.Request, body St
...
+	if srcEntry == nil {
+		srcEntry, err = c.Catalog.GetEntry(ctx, repo.Name, branch, srcPath, catalog.GetEntryParams{ReturnExpired: true})
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	// copy data to a new physical address
+	dstEntry := *srcEntry
+	dstEntry.CreationDate = time.Now().UTC()
```
itaiad200, line 2472: Same here

pkg/api/controller.go
```diff
@@ -2334,85 +2338,156 @@ func (c *Controller) StageObject(w http.ResponseWriter, r *http.Request, body St
...
+	}
+	return &dstEntry, nil
+}
+
+// copyObjectFull copy data from srcEntry's physical address (if set) or srcPath into destPath
+func (c *Controller) copyObjectFull(ctx context.Context, repo *catalog.Repository, branch string, srcPath string, destPath string, srcEntry *catalog.DBEntry) (*catalog.DBEntry, error) {
+	var err error
+	// fetch src entry if needed - optimization in case we already have the entry
+	if srcEntry == nil {
+		srcEntry, err = c.Catalog.GetEntry(ctx, repo.Name, branch, srcPath, catalog.GetEntryParams{ReturnExpired: true})
```
itaiad200, line 2464: Why do we return expired?

pkg/api/controller_test.go
```diff
@@ -3860,64 +3860,157 @@ func TestController_CopyObjectHandler(t *testing.T) {
...
-		time.Sleep(1 * time.Second)
-		copyResp, err := clt.CopyObjectWithResponse(ctx, "repo1", "main", &api.CopyObjectParams{
-			DestPath: "bar/foo",
+		// get back info
+		statResp, err := clt.StatObjectWithResponse(ctx, repo, "main", &api.StatObjectParams{Path: destPath})
+		verifyResponseOK(t, statResp, err)
+		require.Nil(t, deep.Equal(statResp.JSON200, copyStat))
+	})
+
+	t.Run("full_diff_branch", func(t *testing.T) {
```
itaiad200, line 3914: diff has many meanings
```suggestion
	t.Run("full_different_branch", func(t *testing.T) {
```

pkg/gateway/operations/putobject.go
```diff
@@ -136,38 +96,115 @@ func getPathFromSource(copySource string) (path.ResolvedAbsolutePath, error) {
...
+	o.Incr("copy_object", o.Principal, repository, branch)
+	srcPath, err := getPathFromSource(copySource)
 	if err != nil {
 		o.Log(req).WithError(err).Error("could not parse copy source path")
 		_ = o.EncodeError(w, req, gatewayErrors.Codes.ToAPIErr(gatewayErrors.ErrInvalidCopySource))
 		return
 	}
+	ctx := req.Context()
+
+	// TODO(barak): verify that o.Reference must be a branch
```
itaiad200, line 110: Delete or fix

pkg/gateway/operations/putobject.go
```diff
@@ -136,38 +96,115 @@ func getPathFromSource(copySource string) (path.ResolvedAbsolutePath, error) {
...
 
 	o.EncodeResponse(w, req, &serde.CopyObjectResult{
-		LastModified: serde.Timestamp(ent.CreationDate),
-		ETag:         httputil.ETag(ent.Checksum),
+		LastModified: serde.Timestamp(entry.CreationDate),
+		ETag:         httputil.ETag(entry.Checksum),
 	}, http.StatusOK)
 }
 
+func copyObjectShallow(ctx context.Context, c catalog.Interface, repository string, branch string, srcEntry *catalog.DBEntry, destPath string) (*catalog.DBEntry, error) {
```
itaiad200, line 154: Can't we reuse `copyObjectShallow` & `copyObjectFull`?
Not blocking but I think it's worth a scoped little effort if possible

pkg/gateway/operations/base.go
```diff
@@ -39,7 +39,7 @@ const (
 	OperationIDOperationNotFound    OperationID = "not_found"
 )
 
-type ActionIncr func(action, repository, ref, userID string)
+type ActionIncr func(action, userID, repository, ref string)
```
itaiad200, line 42: Did you change all the usages? I missed it

itaiad200 (changes requested):
Thanks for this awesome almost-last-piece of the ugc#2 puzzle! Great tests too!
Most comments around questions and clarifications, nothing too serious but they piled up so requesting changes.

pkg/graveler/graveler.go
```diff
@@ -1522,17 +1553,16 @@ func (g *Graveler) DeleteBatch(ctx context.Context, repository *RepositoryRecord
...
 	var m *multierror.Error
-	err = g.safeBranchWrite(ctx, g.log(ctx).WithField("operation", "delete_keys"),
-		repository, branchID, func(branch *Branch) error {
-			var cachedMetaRangeID MetaRangeID // used to cache the committed branch metarange ID
-			for _, key := range keys {
-				err := g.deleteUnsafe(ctx, repository, branch, key, &cachedMetaRangeID)
-				if err != nil {
-					m = multierror.Append(m, &DeleteError{Key: key, Err: err})
-				}
+	err = g.safeBranchWrite(ctx, g.log(ctx).WithField("operation", "delete_keys"), repository, branchID, 0, func(branch *Branch) error {
```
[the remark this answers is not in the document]
nopcoder: I'll pass options struct so it will be clear to the reader

pkg/catalog/catalog.go
nopcoder: wanted to sort order to be ascending order - our physical addresses ordered by desc.

pkg/catalog/catalog.go
nopcoder: Suggest to remove the sleep or make the delete unbound to any size?

pkg/catalog/catalog.go
nopcoder: yes, I want to delete old 1k keys and repeat the action without overload the system with delete requests.

pkg/api/controller.go
nopcoder: Removed as CreateEntry handles that - thanks.

pkg/api/controller.go
nopcoder: This is the default for get object - but looking into it, no one is using this field so I'll handle it in a different PR.

pkg/gateway/operations/base.go
nopcoder: This is a fix - all the calls use this order of arguments.

nopcoder:
@itaiad200 addressed all the comments, except the one related to the delete of expired tracked physical addresses.
waiting for feedback.
Note that this pr is based on @N-o-Z 's work - thanks for the kind words.

itaiad200 (approved):
Thanks for making the changes!

pkg/api/controller.go
nopcoder: Handled in this PR

PR #5063, opened 2023-01-18
Deprecate OIDC

### LINKED ISSUE #5061: Deprecate embedded OIDC support

arielshaqed:
Closes #5061.

### DIFF
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -612,9 +612,11 @@ components:
         oidc_enabled:
           deprecated: true
           type: boolean
+          description: "Deprecated: embedded OIDC support is now deprecated"
         oidc_default_login:
           deprecated: true
           type: boolean
+          description: "Deprecated: embedded OIDC support is now deprecated"
     AccessKeyCredentials:
       type: object
       properties:
@@ -1297,6 +1299,7 @@ paths:
   /oidc/callback:
     get:
       deprecated: true
+      description: "Deprecated: embedded OIDC support is now deprecated"
       tags:
         - auth
       operationId: oauthCallback
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -610,8 +610,10 @@ components:
           type: string
           enum: [initialized, not_initialized, comm_prefs_done]
         oidc_enabled:
+          deprecated: true
           type: boolean
         oidc_default_login:
+          deprecated: true
           type: boolean
     AccessKeyCredentials:
       type: object
@@ -1294,6 +1296,7 @@ paths:
 
   /oidc/callback:
     get:
+      deprecated: true
       tags:
         - auth
       operationId: oauthCallback
--- a/cmd/lakefs/cmd/run.go
+++ b/cmd/lakefs/cmd/run.go
@@ -257,6 +257,9 @@ var runCmd = &cobra.Command{
 		var oauthConfig *oauth2.Config
 		var oidcProvider *oidc.Provider
 		if cfg.Auth.OIDC.Enabled {
+			logger.
+				WithField("feature", "oidc").
+				Warn("Enabling OIDC on lakeFS server, but this functionality is deprecated")
 			oidcProvider, err = oidc.NewProvider(
 				ctx,
 				cfg.Auth.OIDC.URL,
--- a/docs/assets/js/swagger.yml
+++ b/docs/assets/js/swagger.yml
@@ -610,9 +610,13 @@ components:
           type: string
           enum: [initialized, not_initialized, comm_prefs_done]
         oidc_enabled:
+          deprecated: true
           type: boolean
+          description: "Deprecated: embedded OIDC support is now deprecated"
         oidc_default_login:
+          deprecated: true
           type: boolean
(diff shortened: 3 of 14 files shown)


arielshaqed:
@Isan-Rivkin PTAL!

api/swagger.yml
```diff
@@ -610,9 +610,13 @@ components:
           type: string
           enum: [initialized, not_initialized, comm_prefs_done]
         oidc_enabled:
+          deprecated: true
           type: boolean
+          description: "Deprecated: embedded OIDC support is now deprecated"
```
Isan-Rivkin, line 615: Consider adding "... and will be removed in the upcoming versions"

cmd/lakefs/cmd/run.go
```diff
@@ -257,6 +257,9 @@ var runCmd = &cobra.Command{
 		var oauthConfig *oauth2.Config
 		var oidcProvider *oidc.Provider
 		if cfg.Auth.OIDC.Enabled {
+			logger.
+				WithField("feature", "oidc").
+				Warn("Enabling OIDC on lakeFS server, but this functionality is deprecated")
```
Isan-Rivkin, line 262: Consider adding "... and will be removed in the upcoming versions"

docs/assets/js/swagger.yml
```diff
@@ -246,6 +246,10 @@ components:
           enum: [common_prefix, object]
         physical_address:
           type: string
+          description: |
+            The location of the object on the underlying object store.
+            Formatted as a native URI with the object store type as scheme ("s3://...", "gs://...", etc.)
+            Or, in the case of presign=true, will be an HTTP URL to be consumed via regular HTTP GET
```
Isan-Rivkin, line 252: Is it related to deprecation?

docs/assets/js/swagger.yml
```diff
@@ -874,6 +882,12 @@ components:
           type: boolean
         upgrade_url:
           type: string
+    GarbageCollectionConfig:
+      type: object
+      properties:
+        grace_period:
+          description: Duration in seconds. Objects created in the recent grace_period will not be collected.
+          type: integer
```
Isan-Rivkin, line 890: Related to deprecation?

docs/assets/js/swagger.yml
```diff
@@ -3868,6 +3879,21 @@ paths:
...
+      description: get information of gc settings
+      responses:
+        200:
+          description: lakeFS garbage collection config
+          content:
+            application/json:
+              schema:
+                $ref: "#/components/schemas/GarbageCollectionConfig"
+        401:
+          $ref: "#/components/responses/Unauthorized"
```
Isan-Rivkin, line 3896: Is this related in anyway to the OIDC deprecation?

Isan-Rivkin (changes requested):

api/swagger.yml
arielshaqed: Not done: this is a product decision.  Our [communication on CHANGELOG](https://github.com/treeverse/lakeFS/blob/bf692ea20db7e9f9f48ef646a3eb36afbed53c33/CHANGELOG.md?plain=1#L7) is merely "deprecated".

docs/assets/js/swagger.yml
arielshaqed: No, just a missing "make gen".  I'll rebase before pushing.  The automated generated copy of Swagger.yml must be generated automatically.

docs/assets/js/swagger.yml
arielshaqed: As above.

arielshaqed (commented):
Thanks!

Didn't do either thing (add to deprecation, fix automatic copies) as explained inside.

arielshaqed (commented):
Other changes to swagger.yml copy went away after rebasing  🎆

api/swagger.yml
ozkatz: If we expect OIDC to remain in lakeFS cloud and for lakeFS cloud to be compatible with existing clients, I would suggest not marking it as deprecated in the API itself (docs changes that were already pushed should clarify and suffice)

ozkatz (changes requested):
I believe this PR should be closed since the deprecated features are only deprecated for the lakeFS OSS server. Clients should still support this functionality.

api/swagger.yml
arielshaqed: 1. It is hard to imagine a client using getSetupState _and_ a generated OpenAPI client.  webui does use the endpoint, but needs to change and anyway doesn't use a generated client.
1. If there is such a client, it probably doesn't use `oidc_enabled` -- and will indeed be unaffected by this change.
1. If there is such a client that _does_ use `oidc_enabled` then it will stop working, because this property is indeed going away in lakeFS and also in Cloud.

ozkatz: Thanks for the context. I agree it should be marked as deprecated then.

ozkatz (approved):

Isan-Rivkin (approved):

arielshaqed:
Thanks!

Rebased on trunk, pulling once it's green again.

PR #5066, opened 2023-01-18
Fix Azure walker

### LINKED ISSUE #5065: [Bug]: Large Azure imports become unmanageably slow
## Current Behavior:

When running `lakectl import` from an Azure ADLS Gen2 into LakeFS using 0.90.0, the **import speed goes down with time, taking forever** when there is a large number of files to import.

On my LakeFS deployment, the first ~150k files are imported at good speed (>1200 objects/s), but after ~250k files the speed starts going down to ~300 objects/s. At ~450k files the speed is ~80 objects/s

This makes imports of more than 500k files unmanageable. 

I can work around this by importing files in smaller batches of 100-200k objects each, but it would be great if it worked out of the box.

I believe this might be caused by the Azure SDK upgrade that took place in LakeFS 0.90.0 because **with LakeFS 0.86.0 I had run imports of more than 500k files and they finished correctly** with a steady speed of ~1000 objects/s
        
## Steps to Reproduce:
1. Install LakeFS 0.90.0 using Azure ADLS Gen2 as underlying storage
2. Run a lakectl import of more than 600k files

### lakeFS Version

0.90.0

### Deplyoment

Azure helm

guy-har:
Close #5065

### DIFF
--- a/pkg/ingest/store/azure.go
+++ b/pkg/ingest/store/azure.go
@@ -82,6 +82,7 @@ func (a *azureBlobWalker) Walk(ctx context.Context, storageURI *url.URL, op Walk
 	container := a.client.NewContainerClient(qk.ContainerName)
 	listBlob := container.NewListBlobsFlatPager(&azblob.ListBlobsFlatOptions{
 		Prefix: &prefix,
+		Marker: &op.ContinuationToken,
 	})
 
 	for listBlob.More() {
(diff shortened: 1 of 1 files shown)


nopcoder (approved):
LGTM!

pkg/ingest/store/azure.go
```diff
@@ -82,19 +82,22 @@ func (a *azureBlobWalker) Walk(ctx context.Context, storageURI *url.URL, op Walk
...
 		if err != nil {
 			return err
 		}
-
 		for _, blobInfo := range resp.Segment.BlobItems {
 			// skipping everything in the page which is before 'After' (without forgetting the possible empty string key!)
 			if op.After != "" && *blobInfo.Name <= op.After {
 				continue
 			}
+			if resp.Marker != nil {
```
N-o-Z, line 98: @guy-har I think we should use resp.NextMarker instead.
WDYT?

PR #5078, opened 2023-01-22
fix: Chi breaks lua client routing

### LINKED ISSUE #5077: [Bug]: lakeFS client in Lua raises "invalid API endpoint" error
### What happened?

Current Behavior:
Using a Lua hook that uses the embedded lakeFS Lua client will return "invalid API endpoint" for legitimate client requests.
        
Steps to Reproduce:
1. Set up any Lua hook that uses the embedded client such as [examples/hooks/parquet_schema_validator.lua](https://github.com/treeverse/lakeFS/blob/master/examples/hooks/parquet_schema_validator.lua)
2. Trigger the hook by e.g. performing a merge
3. An error will be returned: "invalid API endpoint"

### Expected Behavior

Expected Behavior:
The client will succeed to call the API endpoint (or fail for other, legitimate reasons)

### lakeFS Version

0.90.0

### Deplyoment

local

### Affected Clients

_No response_

### Relevant logs output

```shell
ApiException: (412)
Reason: Precondition Failed
HTTP response headers: HTTPHeaderDict({'Date': 'Sat, 21 Jan 2023 02:13:40 GMT', 'Content-Type': 'application/json', 'Content-Length': '285', 'Connection': 'keep-alive', 'X-Content-Type-Options': 'nosniff', 'X-Request-Id': '91a6b1b8-392f-4ec8-a666-c40448548c24'})
HTTP response body: {"message":"update branch main: pre-merge hook aborted, run id '5lkl3b47ifk0v5jqvr7g': 1 error occurred:\n\t* hook run id '0000_0000' failed on action 'pre merge format check on main' hook 'check_formats': runtime error: [string \"lua\"]:49: could not diff: invalid API endpoint\n\n"}
```

### Contact Details

_No response_

ozkatz:
Closes #5077

### DIFF
--- a/pkg/actions/lua/lakefs/client.go
+++ b/pkg/actions/lua/lakefs/client.go
@@ -10,6 +10,8 @@ import (
 	"net/http/httptest"
 	"strings"
 
+	"github.com/go-chi/chi/v5"
+
 	"github.com/Shopify/go-lua"
 	"github.com/treeverse/lakefs/pkg/actions/lua/util"
 	"github.com/treeverse/lakefs/pkg/auth"
@@ -39,12 +41,15 @@ func newLakeFSRequest(ctx context.Context, user *model.User, method, url string,
 	if data == nil {
 		body = bytes.NewReader(data)
 	}
-	req, err := http.NewRequestWithContext(ctx, method, url, body)
+
+	// Chi stores its routing information on the request context which breaks this sub-request's routing.
+	// We explicitly nullify any existing routing information before creating the new request
+	subCtx := auth.WithUser(context.WithValue(ctx, chi.RouteCtxKey, nil), user)
+	req, err := http.NewRequestWithContext(subCtx, method, url, body)
 	req.Header.Set("User-Agent", LuaClientUserAgent)
 	if err != nil {
 		return nil, err
 	}
-	req = req.WithContext(auth.WithUser(req.Context(), user)) // add user
 	return req, nil
 }
 
--- a/pkg/actions/lua/lakefs/client.go
+++ b/pkg/actions/lua/lakefs/client.go
@@ -10,9 +10,8 @@ import (
 	"net/http/httptest"
 	"strings"
 
-	"github.com/go-chi/chi/v5"
-
 	"github.com/Shopify/go-lua"
+	"github.com/go-chi/chi/v5"
 	"github.com/treeverse/lakefs/pkg/actions/lua/util"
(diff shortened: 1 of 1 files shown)


pkg/actions/lua/lakefs/client.go
```diff
@@ -39,12 +41,15 @@ func newLakeFSRequest(ctx context.Context, user *model.User, method, url string,
 	if data == nil {
 		body = bytes.NewReader(data)
 	}
-	req, err := http.NewRequestWithContext(ctx, method, url, body)
+
+	// Chi stores its routing information on the request context which breaks this sub-request's routing.
+	// We explicitly nullify any existing routing information before creating the new request
+	subCtx := auth.WithUser(context.WithValue(ctx, chi.RouteCtxKey, nil), user)
+	req, err := http.NewRequestWithContext(subCtx, method, url, body)
```
nopcoder, line 48: 1. Suggested small re-write to make the code more clear.
2. If our context already passed the auth middleware - the user already set, right?
```suggestion
	// Chi stores its routing information on the request context which breaks this sub-request's routing.
	// We explicitly nullify any existing routing information before creating the new request
	ctx = context.WithValue(ctx, chi.RouteCtxKey, nil)
	// Add user to the request context
	ctx = auth.WithUser(ctx, user)
	req, err := http.NewRequestWithContext(ctx, method, url, body)
```

pkg/actions/lua/lakefs/client.go
```diff
@@ -10,6 +10,8 @@ import (
 	"net/http/httptest"
 	"strings"
 
+	"github.com/go-chi/chi/v5"
+
 	"github.com/Shopify/go-lua"
```
nopcoder, line 15: delete the blank line

nopcoder (changes requested):
Question and a suggestion

pkg/actions/lua/lakefs/client.go
ozkatz: Thanks, modified.
I prefer to explicitly set the given user - in cases where we decide to execute a hook without being triggered by another HTTP request, this assumption will break. 

I prefer to treat context as an opaque value without assuming anything about it (except of course, to clean any Chi related things that will break this execution)

pkg/actions/lua/lakefs/client.go
ozkatz: Done

nopcoder (approved):
LGTM!

PR #5083, opened 2023-01-23
Automated playground deployment trigger for new lakeFS release

idanovo:


.github/workflows/docker-publish.yaml
```diff
@@ -90,3 +90,21 @@ jobs:
...
+          https://api.github.com/repos/${{ secrets.REPOSITORY_NAME }}/actions/workflows/${{ secrets.WORKFLOW_ID_STAGING }}/dispatches \
+          --data '{"ref":"main", "inputs": { "lakefs_image": ${{ steps.version.outputs.tag }} }}'
+
+          curl \
+          -X POST \
+          -H "Authorization: Bearer ${{ secrets.PAT_TOKEN }}"\
+          -H "Accept: application/vnd.github+json" \
+          -H "Content-Type: application/json" \
+          https://api.github.com/repos/${{ secrets.REPOSITORY_NAME }}/actions/workflows/${{ secrets.WORKFLOW_ID_PRODUCTION }}/dispatches \
+          --data '{"ref":"main", "inputs": { "lakefs_image": ${{ steps.version.outputs.tag }} }}'
```
ortz, line 110: @idanovo this doesn't allow you to create a dependency between the workflows (and test the outcome of the change).

idanovo: @ortz you are right. changed to one one job with dependency

ortz (approved):

Jonathan-Rosenberg (changes requested):
shouldn't you use [Reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows)?

idanovo:
> shouldn't you use [Reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows)?

Looks cool, but I'm not sure how it's related.
I'm triggering here workflow of another repository.

Jonathan-Rosenberg:
@idanovo oh miss that detail...
Thanks!

Jonathan-Rosenberg (approved):

PR #5084, opened 2023-01-23
Delta diff prototype: Prettify UI 

talSofer:


nopcoder:
@Jonathan-Rosenberg is this still relevant or can we close it?

PR #5086, opened 2023-01-24
Async comm prefs

johnnyaug:


pkg/stats/collector.go
```diff
@@ -405,9 +405,15 @@ func (s *BufferedCollector) CollectCommPrefs(email, installationID string, featu
...
 	err := s.sender.UpdateCommPrefs(ctx, commPrefs)
+	status := "success"
 	if err != nil {
 		s.log.WithError(err).Info("could not update comm prefs")
+		status = "failed"
 	}
+	s.CollectEvent(Event{
+		Class: "global",
+		Name:  fmt.Sprintf("update_comm_prefs_%s", status),
+	})
```
nopcoder, line 416: Suggestion - it will be easy to lookup later
```suggestion
	name := "update_comm_prefs_success"
	if err != nil {
		name = "update_comm_prefs_failed"
		s.log.WithError(err).Info("could not update comm prefs")
	}
	s.CollectEvent(Event{
		Class: "global",
		Name:  name,
	})
```

nopcoder (approved):
Added small suggestion to make it easy to code search later

PR #5089, opened 2023-01-24
UGC docs update for milesone 2

nopcoder:


itaiad200 (approved):
Thanks! Let's merge it once we release the next metadata client version

PR #5092, opened 2023-01-25
Github Workflows: docker publish manual workflow

Jonathan-Rosenberg:
Added a manual docker hub publish workflow.
This will be used to publish docker images to Docker hub (notice- not to ECR), and will require an explicit name provided by the user.

For the name "exp", `$(echo $(git describe --abbrev=1 --tags)-$(echo ${{ inputs.name }}))` will output something like:
`v0.90.1-4-g3ceb5-exp`


.github/workflows/docker-publish-manual.yaml
```diff
@@ -0,0 +1,69 @@
+name: Docker manual
```
nopcoder, line 1: Just name it in a different way (filename too) to make it clear it is not the manual workflow of our release docker flow

nopcoder (approved):

.github/workflows/docker-publish-manual.yaml
Jonathan-Rosenberg: done
thanks!

PR #5093, opened 2023-01-25
Configure webui login URLs and details dynamically

### LINKED ISSUE #5085: Optionally configure log{in,out} URLs and cookies in web UI
Replace specific hard-coded OIDC values in web UI and a general response
"OIDC enabled" from lakeFS with generic URLs.  If nothing is configured,
support the non-SSO URLs by default.

arielshaqed:
Fixes #5085.

### DIFF
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -618,6 +618,32 @@ components:
             with an external auth service.
           type: string
           enum: [simplified, external]
+        login_url:
+          description: primary URL to use for login.
+          type: string
+        login_failed_message:
+          description: |
+            message to display to users who fail to login; a full sentence that is rendered
+            in HTML and may contain a link to a secondary login method
+          type: string
+        fallback_login_url:
+          description: secondary URL to offer users to use for login.
+          type: string
+        fallback_login_label:
+          description: label to place on fallback_login_url.
+          type: string
+        login_cookies:
+          description: cookies to store JWT
+          type: array
+          items:
+            type: string
+        logout_url:
+          description: URL to use for logging out.
+          type: string
+      required:
+      - login_url
+      - login_cookies
+      - logout_url
 
     SetupState:
       type: object
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -632,8 +632,8 @@ components:
         fallback_login_label:
           description: label to place on fallback_login_url.
           type: string
-        login_cookies:
-          description: cookies to store JWT
+        login_cookie_names:
+          description: cookie names used to store JWT
           type: array
           items:
             type: string
@@ -642,7 +642,7 @@ components:
           type: string
       required:
       - login_url
-      - login_cookies
+      - login_cookie_names
       - logout_url
 
     SetupState:
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -3523,8 +3523,15 @@ func (c *Controller) GetSetupState(w http.ResponseWriter, r *http.Request) {
 		State:            swag.String(state),
 		OidcEnabled:      swag.Bool(c.Config.Auth.OIDC.Enabled),
 		OidcDefaultLogin: swag.Bool(c.Config.Auth.OIDC.IsDefaultLogin),
+		// TODO(ariels): Merge defaults and configuration
 		LoginConfig: &LoginConfig{
 			RBAC: &c.Config.Auth.UIConfig.RBAC,
+			LoginCookies: []string{
+				"internal_auth_session",
+				// Add sso_auth_session for SSO
+			},
+			LoginUrl:  "/auth/login",
+			LogoutUrl: "/logout",
 		},
 	}
 	writeResponse(w, r, http.StatusOK, response)
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -3523,7 +3523,7 @@ func makeLoginConfig(c *config.Config) *LoginConfig {
 	}
 
 	return &LoginConfig{
-		RBAC:               &c.Config.Auth.UIConfig.RBAC,
+		RBAC:               &c.Auth.UIConfig.RBAC,
 		LoginUrl:           c.Auth.UIConfig.LoginURL,
 		LoginFailedMessage: &loginFailedMessage,
 		FallbackLoginUrl:   fallbackLoginURL,
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -3504,28 +3504,24 @@ func (c *Controller) GetTag(w http.ResponseWriter, r *http.Request, repository s
 }
 
 func makeLoginConfig(c *config.Config) *LoginConfig {
-	// TODO(ariels): Configure by c.Auth.OIDC.Enabled if set, otherwise
-	// from C.Auth.UIConfig
 	var (
-		cookieNames = c.Auth.UIConfig.LoginCookieNames
-		// []string{
-		// 	"internal_auth_session",
-		// }
-		loginFailedMessage = c.Auth.UIConfig.LoginFailedMessage // "The credentials don't match."
-		fallbackLoginURL   = c.Auth.UIConfig.FallbackLoginURL   // nil
-		fallbackLoginLabel = c.Auth.UIConfig.FallbackLoginLabel // nil
-
-		defaultFallbackLoginURL   = "/oidc/login?prompt=login"
-		defaultFallbackLoginLabel = "Sign in with SSO provider"
+		cookieNames        = c.Auth.UIConfig.LoginCookieNames
+		loginFailedMessage = c.Auth.UIConfig.LoginFailedMessage
+		fallbackLoginURL   = c.Auth.UIConfig.FallbackLoginURL
+		fallbackLoginLabel = c.Auth.UIConfig.FallbackLoginLabel
 	)
 	if c.Auth.OIDC.Enabled {
+		var (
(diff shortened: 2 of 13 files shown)


arielshaqed:
Manual verifications performed:

* [ ]: Nothing configured in login_config, everything works normally.
* [ ]: OIDC configured and enabled, everything continues to work.  (But this
     option **WILL** go away shortly!)
* [ ]: login_config configured, all fields work.  Full end-to-end tests will
     have to wait until we have an alternative login controller service, but
     we can already validate the UI works according to the login_config that
     it receives from lakeFS.

arielshaqed:
# Validation with OIDC configuration

I used this configuration snippet:

``` yaml
  oidc:
    enabled: true
    url: https://XXX.YYY.IDP.com/   # elided
    client_id: abcdefg
    client_secret: hijklmnoprshhhhhh
    callback_base_url: http://localhost:8000/
```

and login still works.

arielshaqed:
# Validation of ui_config.login_failed_message

Added this snippet:

``` yaml
  ui_config:
    login_failed_message: That didn't work.  Does a link to <a href="https://cnn.com">CNN</a> work?
    # ...
```

Failed to login, and got a text with a link to CNN.

arielshaqed:
# Validation of fallback configuration

With this configuration:

``` yaml
  ui_config:
    fallback_login_url: https://login.example.com
    fallback_login_label: |
      Didn't work?  Maybe try SSL Stone Souper Login, "just add water  and login!"
```

the fallback button has the desired label and tries to redirect to the
server.

arielshaqed:
# Validation of login redirection

With this configuration:

``` yaml
  ui_config:
    login_url: https://another.example.com/go/login?secret=1234
```

the URL localhost:8000/auth/login?redirected=true&next=%2Frepositories
redirects to the pretend login server.

johnnyaug:
> Note that it is only relevant when redirected=true; we never go directly
to a login URL. @johnnyaug amirite?

Yes, this is to avoid jumping to this URL when the user explicitly navigates to the "internal" login page. It may or may not still be relevant today.

api/swagger.yml
```diff
@@ -603,12 +603,45 @@ components:
...
+            in HTML and may contain a link to a secondary login method
+          type: string
+        fallback_login_url:
+          description: secondary URL to offer users to use for login.
+          type: string
+        fallback_login_label:
+          description: label to place on fallback_login_url.
+          type: string
+        login_cookies:
+          description: cookies to store JWT
```
johnnyaug, line 625:
```suggestion
          description: cookie names used to store JWT
```

johnnyaug: Possibly also change the property name to `login_cookie_names`. Not feeling too strongly about this one.

pkg/api/controller.go
```diff
@@ -3493,6 +3493,39 @@ func (c *Controller) GetTag(w http.ResponseWriter, r *http.Request, repository s
 	writeResponse(w, r, http.StatusOK, response)
 }
 
+func makeLoginConfig(c *config.Config) *LoginConfig {
+	// TODO(ariels): Configure by c.Auth.OIDC.Enabled if set, otherwise
+	// from C.Auth.UIConfig
```
johnnyaug, line 3498: Seems like this TODO is already done?

pkg/api/controller.go
```diff
@@ -3493,6 +3493,39 @@ func (c *Controller) GetTag(w http.ResponseWriter, r *http.Request, repository s
...
+		cookies = c.Auth.UIConfig.LoginCookies
+		// []string{
+		// 	"internal_auth_session",
+		// }
+		loginFailedMessage = c.Auth.UIConfig.LoginFailedMessage // "The credentials don't match."
+		fallbackLoginURL   = c.Auth.UIConfig.FallbackLoginURL   // nil
+		fallbackLoginLabel = c.Auth.UIConfig.FallbackLoginLabel // nil
+
+		defaultFallbackLoginURL   = "/oidc/login?prompt=login"
+		defaultFallbackLoginLabel = "Sign in with SSO provider"
```
johnnyaug, line 3509: These are only used when OIDC is enabled, so they are not really default. Can hard-code them in the `if` below.

pkg/api/controller.go
```diff
@@ -3493,6 +3493,39 @@ func (c *Controller) GetTag(w http.ResponseWriter, r *http.Request, repository s
...
+	// TODO(ariels): Configure by c.Auth.OIDC.Enabled if set, otherwise
+	// from C.Auth.UIConfig
+	var (
+		cookies = c.Auth.UIConfig.LoginCookies
+		// []string{
+		// 	"internal_auth_session",
+		// }
+		loginFailedMessage = c.Auth.UIConfig.LoginFailedMessage // "The credentials don't match."
+		fallbackLoginURL   = c.Auth.UIConfig.FallbackLoginURL   // nil
+		fallbackLoginLabel = c.Auth.UIConfig.FallbackLoginLabel // nil
```
johnnyaug, line 3506: Are the comments here the default values? Because I don't see them set anywhere.

webui/src/pages/auth/login.tsx
```diff
@@ -10,9 +10,18 @@ import {Error} from "../../lib/components/controls"
 import {useRouter} from "../../lib/hooks/router";
 import {useAPI} from "../../lib/hooks/api";
 
-const OIDC_LOGIN_URL = "/oidc/login?prompt=login";
+//const OIDC_LOGIN_URL = "/oidc/login?prompt=login";
```
johnnyaug, line 13: /dd?

johnnyaug (approved):
Cool! Some comments, nothing major.

pkg/api/controller.go
```diff
@@ -3493,6 +3493,39 @@ func (c *Controller) GetTag(w http.ResponseWriter, r *http.Request, repository s
...
+		defaultFallbackLoginLabel = "Sign in with SSO provider"
+	)
+	if c.Auth.OIDC.Enabled {
+		cookies = append(cookies, "oidc_auth_session")
+		loginFailedMessage = `The credentials don&apos;t match. You may be registered through our <a href={"/oidc/login?prompt=login"}>SSO Provider.</a>`
+		fallbackLoginURL = &defaultFallbackLoginURL
+		fallbackLoginLabel = &defaultFallbackLoginLabel
+	}
+
+	// "/oidc/login?prompt=login"
```
Isan-Rivkin, line 3518: remove comment? :)

pkg/api/controller.go
```diff
@@ -3506,10 +3539,13 @@ func (c *Controller) GetSetupState(w http.ResponseWriter, r *http.Request) {
 	if c.Config.IsAuthTypeAPI() {
 		state = string(auth.SetupStateInitialized)
 	}
+	lc := makeLoginConfig(c.Config)
 	response := SetupState{
 		State:            swag.String(state),
 		OidcEnabled:      swag.Bool(c.Config.Auth.OIDC.Enabled),
 		OidcDefaultLogin: swag.Bool(c.Config.Auth.OIDC.IsDefaultLogin),
```
Isan-Rivkin, line 3546: Those fields will be removed after deprecation from this struct - right?

Isan-Rivkin (changes requested):

pkg/api/controller.go
```diff
@@ -3493,6 +3493,39 @@ func (c *Controller) GetTag(w http.ResponseWriter, r *http.Request, repository s
 	writeResponse(w, r, http.StatusOK, response)
 }
```
Isan-Rivkin, line 3494: Regarding [logout_handler](https://github.com/treeverse/lakeFS/blob/master/pkg/api/logout_handler.go#L11) handler, if we handle some logic in lakeFS and only then redirect to external provider (instead of directly redirecting from UI to logout provider skipping lakeFS) it's also need to be modified.

arielshaqed:
Thanks!

PTAL...  (Will rebase on trunk after, to keep ease of commenting on this PR)

arielshaqed (commented):
THANKS!

PTAL...

pkg/api/controller.go
Isan-Rivkin: We're already aligned, maybe the OIDC should be removed in the future, but nothing else.

Isan-Rivkin (approved):
LGTM Great work!

arielshaqed:
Reverify after all changes: unchecked above boxes, re-checking now before pulling.

> Manual verifications performed:
> 
>     * [ ]  Nothing configured in login_config, everything works normally.
> 
>     * [ ]  OIDC configured and enabled, everything continues to work.  (But this
>       option **WILL** go away shortly!)
> 
>     * [ ]  login_config configured, all fields work.  Full end-to-end tests will
>       have to wait until we have an alternative login controller service, but
>       we can already validate the UI works according to the login_config that
>       it receives from lakeFS.

arielshaqed:
> * [ ] login_config configured, all fields work. Full end-to-end tests will
>   have to wait until we have an alternative login controller service, but
>   we can already validate the UI works according to the login_config that
>   it receives from lakeFS.

Everything works except for logout_url.  Opening a separate issue #5195 to support that, as requirements will be part of building a complete system.  @Isan-Rivkin FYI, please add a comment there about actual requirements :-)

arielshaqed:
Thanks!  Pulling...

Only required check not reported is **license/cla**.  But it's me, guv, honest!  So pulling by adminiforce.

1 "expected" test is license/cla, not run for reasons unknown

PR #5096, opened 2023-01-25
Support multiple storage accounts in Azure

### LINKED ISSUE #5047: [Bug]: `lakectl ingest` for Azure doesn't allow reading ingested files
### TL;DR:

`lakectl ingest` purpose is to add Data from external sources into LakeFS, but for Azure it does not work. 

## Current Behavior:

Azure storage accounts need an authentication key to read data. [According to LakeFS documentation](https://docs.lakefs.io/howto/import.html#using-the-lakectl-ingest-command) when doing a `lakectl ingest` I should add the storage account key as a local environment variable. That makes the `ingest` command authenticate and seems to ingest the data successfully.
        
However, when trying to read that data using LakeFS `s3a://` afterwards I get an error, because the key of the storage account hasn't been stored anywhere and I have no way of passing it to LakeFS. 

### Why is this important? 

Currently I can do `lakectl import` to load data from the same storage account that LakeFS uses as underlying storage, but there is no way to load data from other storage accounts, **For organizations with multiple Azure storage accounts this becomes a blocker for LakeFS adoption**

## Steps to Reproduce:
1. Do a `lakectl ingest` from a different Azure storage account than the one LakeFS uses as underlying storage
2. Try to read the data, e.g. with a `spark.read.csv("s3a://<some-lakefs-path>")` 

## Additional comments

- The `lakefs ingest` [documentation](https://docs.lakefs.io/howto/import.html#using-the-lakectl-ingest-command) is very misleading: in the prerequisites it states that the user and LakeFS installation should have permissions to list and read the objects being ingested, but there's no way I can give LakeFS permission to an Azure storage account
- I guess the easiest way of solving this is allowing LakeFS to store multiple keys for Azure Storage accounts and automatically use the corresponsing key to authenticate whenever I ingest or read data from any of them

### Expected Behavior

_No response_

### lakeFS Version

0.86.0

### Deplyoment

Kubernetes/Helm in Azure

### Affected Clients

_No response_

### Relevant logs output

_No response_

### Contact Details

Iddo and Amit have my email. You can also reach me in LakeFS community Slack

itaiad200:
closes #5047

### DIFF
--- a/pkg/block/azure/adapter.go
+++ b/pkg/block/azure/adapter.go
@@ -163,7 +163,7 @@ func (a *Adapter) Put(ctx context.Context, obj block.ObjectPointer, sizeBytes in
 		return err
 	}
 	o := a.translatePutOpts(ctx, opts)
-	containerClient, err := a.clientCache.NewContainerClient(obj.StorageNamespace, qualifiedKey.ContainerName)
+	containerClient, err := a.clientCache.NewContainerClient(qualifiedKey.ContainerURL, qualifiedKey.ContainerName)
 	if err != nil {
 		return err
 	}
@@ -196,7 +196,7 @@ func (a *Adapter) getPreSignedURL(obj block.ObjectPointer, permissions sas.BlobP
 		return "", err
 	}
 
-	containerClient, err := a.clientCache.NewContainerClient(obj.StorageNamespace, qualifiedKey.ContainerName)
+	containerClient, err := a.clientCache.NewContainerClient(qualifiedKey.ContainerURL, qualifiedKey.ContainerName)
 	if err != nil {
 		return "", err
 	}
@@ -222,7 +222,7 @@ func (a *Adapter) Download(ctx context.Context, obj block.ObjectPointer, offset,
 	if err != nil {
 		return nil, err
 	}
-	container, err := a.clientCache.NewContainerClient(obj.StorageNamespace, qualifiedKey.ContainerName)
+	container, err := a.clientCache.NewContainerClient(qualifiedKey.ContainerURL, qualifiedKey.ContainerName)
 	if err != nil {
 		return nil, err
 	}
@@ -254,7 +254,7 @@ func (a *Adapter) Walk(ctx context.Context, walkOpt block.WalkOpts, walkFn block
 		return err
 	}
 
-	containerClient, err := a.clientCache.NewContainerClient(walkOpt.StorageNamespace, qualifiedPrefix.ContainerName)
+	containerClient, err := a.clientCache.NewContainerClient(qualifiedPrefix.ContainerURL, qualifiedPrefix.ContainerName)
 	if err != nil {
 		return err
 	}
@@ -292,7 +292,7 @@ func (a *Adapter) Exists(ctx context.Context, obj block.ObjectPointer) (bool, er
 		return false, err
 	}
 
-	containerClient, err := a.clientCache.NewContainerClient(obj.StorageNamespace, qualifiedKey.ContainerName)
+	containerClient, err := a.clientCache.NewContainerClient(qualifiedKey.ContainerURL, qualifiedKey.ContainerName)
 	if err != nil {
 		return false, err
 	}
@@ -318,7 +318,7 @@ func (a *Adapter) GetProperties(ctx context.Context, obj block.ObjectPointer) (b
 		return block.Properties{}, err
 	}
 
-	containerClient, err := a.clientCache.NewContainerClient(obj.StorageNamespace, qualifiedKey.ContainerName)
+	containerClient, err := a.clientCache.NewContainerClient(qualifiedKey.ContainerURL, qualifiedKey.ContainerName)
 	if err != nil {
 		return block.Properties{}, err
 	}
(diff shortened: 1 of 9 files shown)


pkg/block/azure/client_cache.go
```diff
@@ -0,0 +1,91 @@
...
+		containerToClient: xsync.NewMapOf[*container.Client](),
+		params:            p,
+	}
+}
+
+// storageAccountRegex is used to extract the storage account from azure storage url.
+// Following rules from https://learn.microsoft.com/en-us/azure/storage/common/storage-account-overview#storage-account-name
+var storageAccountRegex = regexp.MustCompile("https://([a-z0-9]{3,24}).blob.core.windows.net/.*")
+
+func ExtractStorageAccount(storageAccountURL string) (string, error) {
```
N-o-Z, line 35: This can be apart of the BlobURLInfo or PrefixURLInfo and extracted usng the ResolveBlobURLInfoFromURL.
I don't think we need a special function just for the storage account

pkg/ingest/store/factory.go
```diff
@@ -10,11 +10,14 @@ import (
 	"cloud.google.com/go/storage"
 	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/service"
 	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/treeverse/lakefs/pkg/block/azure"
 	"github.com/treeverse/lakefs/pkg/block/factory"
 	"github.com/treeverse/lakefs/pkg/block/params"
 )
 
-var ErrNotSupported = errors.New("no storage adapter found")
+var (
```
N-o-Z, line 18: I think this can be reverted?

pkg/block/azure/client_cache.go
```diff
@@ -0,0 +1,91 @@
...
+type ClientContainerCache struct {
+	containerToClient *xsync.MapOf[string, *container.Client]
+	params            params.Azure
+}
+
+var ErrUnknownAzureURL = errors.New("unable to parse storage account")
+
+func NewCache(p params.Azure) *ClientContainerCache {
+	return &ClientContainerCache{
+		containerToClient: xsync.NewMapOf[*container.Client](),
```
N-o-Z, line 26: Is there a cache invalidation / expiry policy?
Theoretically if the cache is never evacuated this could explode over time (or become less efficient?)

N-o-Z (changes requested):
Thanks @itaiad200  - looks awesome.

Some questions / comments to address

pkg/block/azure/adapter.go
```diff
@@ -347,8 +361,18 @@ func (a *Adapter) Copy(ctx context.Context, sourceObj, destinationObj block.Obje
...
+		return err
+	}
+	destClient := destContainerClient.NewBlobClient(qualifiedDestinationKey.BlobURL)
+
+	srcContainerClient, err := a.clientCache.NewContainerClient(qualifiedSourceKey.ContainerURL, qualifiedSourceKey.ContainerName)
+	if err != nil {
+		return err
+	}
+	sourceClient := srcContainerClient.NewBlobClient(qualifiedSourceKey.BlobURL)
+
```
N-o-Z, line 375: I think that creating a presigned URL is not needed anymore - we should be able to perform a copy using the credentials we have

pkg/block/azure/client_cache.go
itaiad200: Done. The function is still needed for the walker flow in ingest.

pkg/ingest/store/factory.go
itaiad200: Done.

pkg/block/azure/adapter.go
itaiad200: That's not always true. Using StorageAccessKey is still supported and this workaround is needed for that.

itaiad200:
Tested on Azure. Both Import & Ingest works as expected from a different storage account

pkg/block/azure/client_cache.go
```diff
@@ -28,35 +30,27 @@ func NewCache(p params.Azure) *ClientContainerCache {
...
 
-// storageAccountRegex is used to extract the storage account from azure storage url.
-// Following rules from https://learn.microsoft.com/en-us/azure/storage/common/storage-account-overview#storage-account-name
-var storageAccountRegex = regexp.MustCompile("https://([a-z0-9]{3,24}).blob.core.windows.net/.*")
-
-func ExtractStorageAccount(storageAccountURL string) (string, error) {
-	matches := storageAccountRegex.FindAllStringSubmatch(storageAccountURL, -1)
-	if len(matches) != 1 || len(matches[0]) != 2 {
-		return "", ErrUnknownAzureURL
+func ExtractStorageAccount(storageAccount *url.URL) (string, error) {
```
N-o-Z, line 33: Can we move it to the adapter module - as this is not directly use in this module

N-o-Z, line 33: Can we move it to the adapter module - as this is not directly use in this module

N-o-Z (approved):
Looks great! Thanks for the code cleanup - approved with one comment

PR #5109, opened 2023-01-26
doc: Minor tweaks to clarify and update `contributing.html`

rmoff:
- State required Java version (8)
- Add some useful hyperlinks, tidy up a couple of notes for readability.


docs/contributing.md
```diff
@@ -35,25 +35,27 @@ Our [Go release workflow](https://github.com/treeverse/lakeFS/blob/master/.githu
...
 
-1. Clone the repository from https://github.com/treeverse/lakeFS (gives you read-only access to the repository. To contribute, see the next section).
 1. Build the project:
 
    ```shell
    make build
    ```
 
-Note: Make build won't work for Windows user for building lakeFS.
+   _Note: Make build won't work for Windows user for building lakeFS._
```
arielshaqed, line 56: Ambivalent about this note, but as long as we have it I'd prefer to phrase it more clearly:
```suggestion
   _Note: `make build` won't work for Windows users._
```

docs/contributing.md
```diff
@@ -35,25 +35,27 @@ Our [Go release workflow](https://github.com/treeverse/lakeFS/blob/master/.githu
 
 1. Install the required dependencies for your OS:
    1. [Git](https://git-scm.com/downloads)
-   1. [GNU make](https://www.gnu.org/software/make/) (probably best to install from your OS package manager such as apt or brew)
+   1. [GNU make](https://www.gnu.org/software/make/) (probably best to install from your OS package manager such as [apt](https://en.wikipedia.org/wiki/APT_(software)) or [brew](https://brew.sh/))
    1. [Docker](https://docs.docker.com/get-docker/)
    1. [Go](https://golang.org/doc/install)
    1. [Node.js & npm](https://www.npmjs.com/get-npm)
    1. [Maven](https://maven.apache.org/) to build and test Spark client codes.
+   1. Java 8, which you can install from [Azul Zulu Builds for Java JDK](https://www.azul.com/downloads/?package=jdk)
```
arielshaqed, line 43: Why recommend Azul for non-Mac users?  E.g. [Fedora has current support for Java 8](https://docs.fedoraproject.org/en-US/quick-docs/installing-java/).  Linux users (let alone GNU/Linux users...) might not like Azul as much; there is no requirement for this particular vendor and I'd rather they got a package with an easier upgrade path.

arielshaqed (changes requested):
Thanks, this will be _much better_!

Only response required for approval is about requiring Azul Java.  All else is optional.

docs/contributing.md
rmoff: Ah good point - it was just the Apple M1 comment that linked to this. I'll rephrase.

docs/contributing.md
```diff
@@ -40,7 +40,8 @@ Our [Go release workflow](https://github.com/treeverse/lakeFS/blob/master/.githu
    1. [Go](https://golang.org/doc/install)
    1. [Node.js & npm](https://www.npmjs.com/get-npm)
    1. [Maven](https://maven.apache.org/) to build and test Spark client codes.
-   1. Java 8, which you can install from [Azul Zulu Builds for Java JDK](https://www.azul.com/downloads/?package=jdk)
+   1. Java 8
+     * Apple M1 users can install this from [Azul Zulu Builds for Java JDK](https://www.azul.com/downloads/?package=jdk). Builds for Intel-based Macs are available from [java.com](https://www.java.com/en/download/help/mac_install.html).
```
rmoff, line 44: @arielshaqed is this better?

arielshaqed: Yes! But there are also Linux users. Maybe just say to use your package manager?

arielshaqed (approved):
Neat, thanks!

PR #5115, opened 2023-01-26
Merge API response - remove summary field

### LINKED ISSUE #3978: Merge response contains zero values
We don't calculate the statistics of the merge operation and return a zero list regardless of the real result:
```
{'summary': {'added': 0, 'changed': 0, 'conflict': 0, 'removed': 0}}
```

If we want to continue supporting that response body we ought to return the real numbers.

nopcoder:
**This is API breaking change**

Remove merge response summary as we do not reflect the actual numbers.
In case caller would like to have data about the merge, it should be reflected by calling diff after getting the new merge reference.

Close #3978

> This remains a very far-reaching change. For instance: say I upgrade my lakeFS. Then my Airflow DAGs stop working!
> 
> Let's add a note to the CHANGELOG _now_ (could be in the next PR, of course, rather than wait for whoever next releases lakeFS).


arielshaqed:
Does this break existing code? I would prefer to return junk numbers, or an arbitrary map, than do that.

If we do decide to break b/c, we will need to communicate to our users. We are one of those users, I think [our Airflow provider provides merge](https://github.com/treeverse/airflow-provider-lakeFS/blob/main/lakefs_provider/operators/merge_operator.py). It mustn't break.

nopcoder:
It will break clients with old API like other breaking changes. No existing code in lakeFS uses this response.
Will check other integrations, and give heads up before it will become part of lakeFS.

nopcoder:
@arielshaqed created https://github.com/treeverse/lakeFS/pull/5119 to have a version where the field is optional while the server will return a valid answer until we remove the summary field.

N-o-Z (approved):
LGTM

arielshaqed:
This remains a very far-reaching change.  For instance: say I upgrade my lakeFS.  Then my Airflow DAGs stop working!

Let's add a note to the CHANGELOG _now_ (could be in the next PR, of course, rather than wait for whoever next releases lakeFS).

nopcoder:
> This remains a very far-reaching change. For instance: say I upgrade my lakeFS. Then my Airflow DAGs stop working!
> 
> Let's add a note to the CHANGELOG _now_ (could be in the next PR, of course, rather than wait for whoever next releases lakeFS).

Thanks @arielshaqed!! I'll fix the client version used by Airflow before merging this one and add a note.

PR #5116, opened 2023-01-26
docs: Expand instructions on building docs locally

rmoff:
1. Expand instructions on how to build docs locally
2. Add Live Reload
3. Add local caching to speed up subsequent local builds (removes the ~4 min fetch of dependencies next time)

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -134,6 +134,7 @@ To render the documentation locally and preview changes you can run the Jeykll s
 
    ```sh
    docker run --rm \
+              --name lakefs_docs \
               --publish 4000:4000 --publish 35729:35729 \
               --volume="$PWD/docs:/srv/jekyll:Z" \
               --volume="$PWD/docs/.jekyll-bundle-cache:/usr/local/bundle:Z" \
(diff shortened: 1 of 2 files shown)


docs/contributing.md
```diff
@@ -126,16 +126,47 @@ Customizing the lakeFS docs site should follow the following guidelines: [Just T
...
+
+   This can take a short while—you'll see something like this in the server's output when it's done. 
+   
+   ```
+   ...done in 34.714073461 seconds.
+   ```
 
+   Your page will automatically reload to show the changes.
 
+_If you are doing lots of work on the docs you may want to leave the Docker container in place (so that you don't have to wait for the dependencies to load each time you re-create it). To do this remove the `--rm` from the `docker run` command._
```
nopcoder, line 169: Do you mean to recommend keep running the service in the background? Removing the '--rm' flag and run as a daemon using `-d`? We can also recommend adding a name like `--name docs`, so in case we like to stop it, we can use the name and not try to lookup for the right container.

rmoff: thanks @nopcoder, added.

nopcoder (approved):

PR #5119, opened 2023-01-27
Merge result summary deprecation

nopcoder:
Merge API change and deprecation waring - step before we  can remove the summary fields from merge result.

This version will always return the summary field and define the field as optional.
It will enable users to switch to a version which will not break when the field is gone.

Related to #3978

### DIFF
--- a/api/swagger.yml
+++ b/api/swagger.yml
@@ -176,28 +176,26 @@ components:
           items:
             $ref: "#/components/schemas/Repository"
 
+    MergeResultSummary:
+      description: "Deprecated: merge summary will be removed"
+      type: object
+      properties:
+        added:
+          type: integer
+        removed:
+          type: integer
+        changed:
+          type: integer
+        conflict:
+          type: integer
+
     MergeResult:
       type: object
       required:
-        - summary
         - reference
       properties:
         summary:
-          type: object
-          required:
-            - added
-            - removed
-            - changed
-            - conflict
-          properties:
-            added:
-              type: integer
-            removed:
-              type: integer
-            changed:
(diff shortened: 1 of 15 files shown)


api/swagger.yml
```diff
@@ -176,28 +176,26 @@ components:
           items:
             $ref: "#/components/schemas/Repository"
 
+    MergeResultSummary:
+      description: "Deprecated: merge summary will be removed"
```
arielshaqed, line 180: I would rather only deprecate all the _fields_ and keep a MergeSummary, even empty.  That will prevent _future_ backwards-compatibility failures -- otherwise when we will want to add a response type, that's another breaking change :-(

api/swagger.yml
```diff
@@ -176,28 +176,26 @@ components:
...
+      type: object
+      properties:
+        added:
+          type: integer
```
arielshaqed, line 184:
```suggestion
        added:
          type: integer
          deprecated: true
          description: Deprecated; inaccurate and will be removed.
```

(Also on all other properties, of course.)

arielshaqed (approved):
Thanks!

I prefer to deprecate the fields rather than the message, because one day we might want to return some value.  At the same time, please also mark all these fields as "deprecated", which issues warnings on Python I think, and add a scary "description".

THANKS!

api/swagger.yml
nopcoder: @arielshaqed updated the PR with field level deprecation - note that I kept the summary as optional as I don't think we will set an empty structure if we keep it as required.

PR #5121, opened 2023-01-28
Improve ingest walkter and cover import from parent directory test

nopcoder:
Test code should break until the fix for GS walker will be merged.


itaiad200 (approved):
LGTM

PR #5126, opened 2023-01-29
Allow stageObject without checksum and size

### LINKED ISSUE #5129: Optional size & checksum field in StageObject
These fields are required in the StageObject request. It's possible to make these required and Stat the object from the object store in the server side.

itaiad200:
closes #5129

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -2308,14 +2308,29 @@ func (c *Controller) StageObject(w http.ResponseWriter, r *http.Request, body St
 
 	physicalAddress, addressType := normalizePhysicalAddress(repo.StorageNamespace, body.PhysicalAddress)
 
+	var sizeBytes int64
+	var checksum string
+	if body.SizeBytes == nil || body.Checksum == nil {
+		props, err := c.BlockAdapter.GetProperties(ctx, block.ObjectPointer{
+			StorageNamespace: qk.StorageNamespace,
+			Identifier:       qk.Key,
+		})
+		if err != nil {
+			writeError(w, r, http.StatusInternalServerError, fmt.Sprintf("Unable to get object properties: %s", err))
+			return
+		}
+		sizeBytes = swag.Int64Value(props.SizeBytes)
+		checksum = swag.StringValue(props.Checksum)
+	}
+
 	entryBuilder := catalog.NewDBEntryBuilder().
 		CommonLevel(false).
 		Path(params.Path).
 		PhysicalAddress(physicalAddress).
 		AddressType(addressType).
 		CreationDate(writeTime).
-		Size(body.SizeBytes).
-		Checksum(body.Checksum).
+		Size(sizeBytes).
+		Checksum(checksum).
 		ContentType(StringValue(body.ContentType))
 	if body.Metadata != nil {
 		entryBuilder.Metadata(body.Metadata.AdditionalProperties)
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -2307,9 +2307,8 @@ func (c *Controller) StageObject(w http.ResponseWriter, r *http.Request, body St
 	}
 
 	physicalAddress, addressType := normalizePhysicalAddress(repo.StorageNamespace, body.PhysicalAddress)
-
-	var sizeBytes int64
-	var checksum string
+	sizeBytes := swag.Int64Value(body.SizeBytes)
+	checksum := swag.StringValue(body.Checksum)
 	if body.SizeBytes == nil || body.Checksum == nil {
 		props, err := c.BlockAdapter.GetProperties(ctx, block.ObjectPointer{
 			StorageNamespace: qk.StorageNamespace,
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -2314,11 +2314,13 @@ func (c *Controller) StageObject(w http.ResponseWriter, r *http.Request, body St
 			StorageNamespace: qk.StorageNamespace,
 			Identifier:       qk.Key,
 		})
-		if err != nil {
-			writeError(w, r, http.StatusInternalServerError, fmt.Sprintf("Unable to get object properties: %s", err))
+		if c.handleAPIError(ctx, w, r, err) {
 			return
 		}
-		sizeBytes = swag.Int64Value(props.SizeBytes)
+		if err != nil {
+			writeError(w, r, http.StatusInternalServerError, fmt.Sprintf("Unable to get object properties: %s", err))
+		}
+		sizeBytes = props.SizeBytes
 		checksum = swag.StringValue(props.Checksum)
 	}
 
--- a/pkg/api/controller_test.go
+++ b/pkg/api/controller_test.go
@@ -2368,7 +2368,7 @@ func TestController_ObjectsStageObjectHandler(t *testing.T) {
 		}
 		content := "this is some content"
 		fullAddr := obj.StorageNamespace + obj.Identifier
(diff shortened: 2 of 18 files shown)


pkg/block/mem/adapter.go
```diff
@@ -76,7 +76,7 @@ func getKey(obj block.ObjectPointer) string {
 	if obj.IdentifierType == block.IdentifierTypeFull {
 		return obj.Identifier
 	}
-	return fmt.Sprintf("%s:%s", obj.StorageNamespace, obj.Identifier)
+	return fmt.Sprintf("%s:%s", strings.TrimRight(obj.StorageNamespace, "/"), obj.Identifier)
```
nopcoder, line 79: there was an issue with the cache key?

pkg/block/adapter.go
```diff
@@ -126,6 +126,8 @@ type CreateMultiPartUploadOpts struct {
 // actually reported.
 type Properties struct {
 	StorageClass *string
+	SizeBytes    *int64
+	Checksum     *string
```
nopcoder, line 130: StorageClass is pointer as it is optional - by if we set size and checksum as additional properties, I think that all underlying storage should return these values. 
Is it the case or we keep them optional because there are other uses for this struct?

nopcoder (changes requested):
Look good. Couple of comments and questions

pkg/api/controller_test.go
```diff
@@ -2350,15 +2350,45 @@ func TestController_ObjectsStageObjectHandler(t *testing.T) {
...
+
+	t.Run("stage object without props", func(t *testing.T) {
+		obj := block.ObjectPointer{
+			StorageNamespace: "some-bucket/",
+			Identifier:       "this/is/the/key",
+			IdentifierType:   block.IdentifierTypeRelative,
+		}
+		content := "this is some content"
+		fullAddr := obj.StorageNamespace + obj.Identifier
+		err := deps.blocks.Put(ctx, obj, int64(len(content)), bytes.NewReader([]byte(content)), block.PutOpts{})
```
nopcoder, line 2371: Use strings reader instead of bytes + cast

pkg/api/controller.go
```diff
@@ -2307,15 +2307,29 @@ func (c *Controller) StageObject(w http.ResponseWriter, r *http.Request, body St
...
 	physicalAddress, addressType := normalizePhysicalAddress(repo.StorageNamespace, body.PhysicalAddress)
+	sizeBytes := swag.Int64Value(body.SizeBytes)
+	checksum := swag.StringValue(body.Checksum)
+	if body.SizeBytes == nil || body.Checksum == nil {
+		props, err := c.BlockAdapter.GetProperties(ctx, block.ObjectPointer{
+			StorageNamespace: qk.StorageNamespace,
+			Identifier:       qk.Key,
+		})
+		if err != nil {
+			writeError(w, r, http.StatusInternalServerError, fmt.Sprintf("Unable to get object properties: %s", err))
```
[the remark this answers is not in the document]
itaiad200: Switched to handleAPIError

pkg/api/controller.go
```diff
@@ -2307,15 +2307,29 @@ func (c *Controller) StageObject(w http.ResponseWriter, r *http.Request, body St
...
+			return
+		}
+		sizeBytes = swag.Int64Value(props.SizeBytes)
+		checksum = swag.StringValue(props.Checksum)
```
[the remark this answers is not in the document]
itaiad200: I was asking my self the same question. We have the "truth" from the object-store, why risk it by using unvalidated user defined data?
Also - what do you think about asserting this flow? You either pass both or none of them, don't give me partial information.

pkg/block/adapter.go
itaiad200: I can change `SizeBytes` to `int64` since it's not optional. `Checksum` isn't available for all block adapters.

pkg/block/gs/adapter.go
```diff
@@ -219,14 +219,17 @@ func (a *Adapter) GetProperties(ctx context.Context, obj block.ObjectPointer) (b
...
 		Attrs(ctx)
 	if err != nil {
-		return props, err
+		return block.Properties{}, err
 	}
-	return props, nil
+	return block.Properties{
+		SizeBytes: &attr.Size,
+		Checksum:  &attr.Etag,
+	}, nil
```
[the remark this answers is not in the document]
itaiad200: Done.

pkg/block/mem/adapter.go
itaiad200: Yap, we would sometime get here with a trailing slash and sometimes don't. It points to the same object, so I made this alignment.

pkg/block/s3/adapter.go
```diff
@@ -453,7 +453,11 @@ func (a *Adapter) GetProperties(ctx context.Context, obj block.ObjectPointer) (b
 	if err != nil {
 		return block.Properties{}, err
 	}
-	return block.Properties{StorageClass: s3Props.StorageClass}, nil
+	return block.Properties{
+		StorageClass: s3Props.StorageClass,
+		Checksum:     s3Props.ETag,
+		SizeBytes:    *s3Props.ContentLength,
```
nopcoder, line 459: need to check for nil before dereference content length

pkg/block/mem/adapter.go
```diff
@@ -88,7 +88,11 @@ func (a *Adapter) Put(_ context.Context, obj block.ObjectPointer, _ int64, reade
 	}
 	key := getKey(obj)
 	a.data[key] = data
-	a.properties[key] = block.Properties(opts)
+
+	a.properties[key] = block.Properties{
+		StorageClass: opts.StorageClass,
+		SizeBytes:    int64(len(data)),
```
nopcoder, line 94: missing checksum

pkg/block/local/adapter.go
```diff
@@ -302,12 +302,14 @@ func (l *Adapter) GetProperties(_ context.Context, obj block.ObjectPointer) (blo
...
-	_, err = os.Stat(p)
+	stats, err := os.Stat(p)
 	if err != nil {
 		return block.Properties{}, err
 	}
 	// No properties, just return that it exists
-	return block.Properties{}, nil
+	return block.Properties{
+		SizeBytes: stats.Size(),
+	}, nil
```
nopcoder, line 312: Missing checksum

pkg/block/azure/adapter.go
```diff
@@ -349,7 +349,17 @@ func (a *Adapter) GetProperties(ctx context.Context, obj block.ObjectPointer) (b
...
-	return block.Properties{StorageClass: props.AccessTier}, nil
+
+	var checksum *string
+	if props.ETag != nil {
+		c := string(*props.ETag)
+		checksum = &c
+	}
+	return block.Properties{
+		StorageClass: props.AccessTier,
+		SizeBytes:    *props.ContentLength,
```
nopcoder, line 360: check if content length is nil before dereference

nopcoder (changes requested):

itaiad200:
Thanks for the review @nopcoder !
I'm putting this aside at the moment until we figure out if we need this.
FYI @ozkatz @kesarwam

ozkatz (changes requested):
Sorry for blocking :)
The implementation looks great, it's the feature I have issues with.

As it stands, this doesn't provide any value over the current API in terms of latency or throughput, while at the same time causing the lakeFS server to accept arbitrary paths and expose even more information about them.

pkg/block/local/adapter.go
nopcoder: Still my concern as we start to treat local adapter like any other one

nopcoder (changes requested):

itaiad200:
Not following in this direction.

PR #5133, opened 2023-01-30
Limited kv store to control background work rate limit

### LINKED ISSUE #4959: Handle async KV calls better
We have some offline processes that perform kv calls. For example, cleaning up a staging area after a commit or a revert, and maintaining the tokens for get/link. While it's the right decision to not block on cleaning operations, they might perform a lot of kv calls and limit more important calls performed in the system. We need to have some pattern in place to limit the number of calls the offline operations perform.

nopcoder:
Close #4959

### DIFF
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -13,8 +13,6 @@ import (
 	"strings"
 	"time"
 
-	"go.uber.org/ratelimit"
-
 	"github.com/alitto/pond"
 	"github.com/cockroachdb/pebble"
 	"github.com/hashicorp/go-multierror"
@@ -41,6 +39,7 @@ import (
 	"github.com/treeverse/lakefs/pkg/upload"
 	"github.com/treeverse/lakefs/pkg/validator"
 	"go.uber.org/atomic"
+	"go.uber.org/ratelimit"
 	"google.golang.org/protobuf/types/known/timestamppb"
 )
 
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -13,6 +13,8 @@ import (
 	"strings"
 	"time"
 
+	"go.uber.org/ratelimit"
+
 	"github.com/alitto/pond"
 	"github.com/cockroachdb/pebble"
 	"github.com/hashicorp/go-multierror"
@@ -136,6 +138,7 @@ type Config struct {
 	SettingsManagerOption    settings.ManagerOption
 	GCMaxUncommittedFileSize int // The maximum file size for uncommitted dump created during PrepareUncommittedGC
 	PathProvider             *upload.PathPartitionProvider
+	Limiter                  ratelimit.Limiter
 }
 
 type Catalog struct {
@@ -146,8 +149,10 @@ type Catalog struct {
 	walkerFactory            WalkerFactory
 	managers                 []io.Closer
 	workPool                 *pond.WorkerPool
-	StoreMessage             *kv.StoreMessage
 	PathProvider             *upload.PathPartitionProvider
+	BackgroundLimiter        ratelimit.Limiter
+	KVStore                  kv.Store
+	KVStoreLimited           kv.Store
 }
 
 const (
@@ -233,10 +238,12 @@ func New(ctx context.Context, cfg Config) (*Catalog, error) {
 	executor := batch.NewConditionalExecutor(logging.Default())
 	go executor.Run(ctx)
 
+	storeLimiter := kv.NewStoreLimiter(cfg.KVStore.Store, cfg.Limiter)
 	refManager := ref.NewRefManager(
 		ref.ManagerConfig{
 			Executor:              executor,
-			KvStore:               cfg.KVStore,
+			KVStore:               cfg.KVStore,
+			KVStoreLimited:        storeLimiter,
 			AddressProvider:       ident.NewHexAddressProvider(),
 			RepositoryCacheConfig: ref.CacheConfig(cfg.Config.Graveler.RepositoryCache),
 			CommitCacheConfig:     ref.CacheConfig(cfg.Config.Graveler.CommitCache),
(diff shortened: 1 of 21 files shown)


pkg/config/config.go
```diff
@@ -9,6 +9,8 @@ import (
 	"strings"
 	"time"
 
+	"go.uber.org/ratelimit"
+
```
N-o-Z, line 13: Newline

pkg/graveler/ref/main_test.go
```diff
@@ -7,6 +7,8 @@ import (
 	"testing"
 	"time"
 
+	"go.uber.org/ratelimit"
+
 	"github.com/treeverse/lakefs/pkg/batch"
```
N-o-Z, line 12: Newline?

pkg/graveler/ref/manager.go
```diff
@@ -575,12 +568,12 @@ func (m *Manager) VerifyLinkAddress(ctx context.Context, repository *graveler.Re
 	if err != nil {
 		return err
 	}
-	return m.deleteLinkAddress(ctx, repository, token)
+	return deleteLinkAddress(ctx, m.kvStore.Store, repository, token)
 }
 
-func (m *Manager) deleteLinkAddress(ctx context.Context, repository *graveler.RepositoryRecord, token string) error {
+func deleteLinkAddress(ctx context.Context, kvStore kv.Store, repository *graveler.RepositoryRecord, token string) error {
```
N-o-Z, line 574: Can you please explain why you decided to modify the function to buy be a member method?

pkg/graveler/staging/manager.go
```diff
@@ -67,16 +69,15 @@ func (m *Manager) Set(ctx context.Context, st graveler.StagingToken, key gravele
 	pb := graveler.ProtoFromStagedEntry(key, value)
 	stPartition := graveler.StagingTokenPartition(st)
 	if requireExists {
-		return m.store.SetMsgIf(ctx, stPartition, key, pb, kv.PrecondConditionalExists)
+		return kv.SetMsgIf(ctx, m.kvStore, stPartition, key, pb, kv.PrecondConditionalExists)
```
N-o-Z, line 72: Manager has both stores no? Why change the signature?

pkg/graveler/staging/manager_test.go
```diff
@@ -13,13 +13,14 @@ import (
 	"github.com/treeverse/lakefs/pkg/kv"
 	"github.com/treeverse/lakefs/pkg/kv/kvtest"
 	"github.com/treeverse/lakefs/pkg/testutil"
+	"go.uber.org/ratelimit"
```
N-o-Z, line 16: New line

pkg/kv/store_limiter.go
```diff
@@ -0,0 +1,48 @@
+package kv
+
+import (
+	"context"
+
+	"go.uber.org/ratelimit"
```
N-o-Z, line 6: New line

N-o-Z (commented):
@nopcoder thank you.
I was just skimming through the code and one question came up to my mind:
Did you check that when rate is unlimited there's no performance degradation in comparison to current code?

pkg/graveler/ref/manager.go
nopcoder: This helper func was use from two places and I wanted to control which store is used for each case.
Called by 'VerifyLinkAddress' which is part of the public interface we should use the kv store.
Called by 'DeleteExpiredLinkAddresses' which is part of the background cleanup wanted to use the limited kv store.

pkg/graveler/staging/manager.go
nopcoder: The rate-limited store uses Store interface not a message store interface.
Didn't want to apply a complete makeover of using the msg store in this PR. I think we can remove the wrapper and just keep the functions as it extend the store interface and not wrap it.

pkg/kv/store_limiter.go
nopcoder: need this new line - built in packages vs external

nopcoder:
> @nopcoder thank you. I was just skimming through the code and one question came up to my mind: Did you check that when rate is unlimited there's no performance degradation in comparison to current code?

The current PR didn't apply any rate limit - need to add it to this PR + docs.
I've tested without rate (the new unlimited just make 'Take' method return time.Now()) and with 1 and 5 request per seconds and tested it manually.
The current tests also uses unlimited store - plan to add to this PR test on the kv level first.

pkg/catalog/catalog.go
```diff
@@ -13,6 +13,8 @@ import (
 	"strings"
 	"time"
 
+	"go.uber.org/ratelimit"
+
```
N-o-Z, line 17: Newline

pkg/graveler/staging/manager.go
N-o-Z: Maybe it's a silly question, but can't StoreLimiter accept a StoreMessage and implement the StoreMessage interface as well?

N-o-Z (changes requested):
Continuing the discussion - I'm convinced we can't avoid taking what seems to be a member method and making it a package method and passing the store object to it

N-o-Z (approved):
LGTM, can you please add the follow up issue regarding refactoring of StoreMessage

itaiad200 (approved):
Thanks, LGTM
Only comment is about the configuration, I don't think we want to let the users define it at this stage.

pkg/config/config.go
```diff
@@ -238,6 +239,9 @@ type Config struct {
 			Expiry time.Duration `mapstructure:"expiry"`
 			Jitter time.Duration `mapstructure:"jitter"`
 		} `mapstructure:"commit_cache"`
+		Background struct {
+			RateLimit int `mapstructure:"rate_limit"`
```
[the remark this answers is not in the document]
nopcoder: The configure is there but not set (unlimited) as it depends on the use-case.
Plan to leave it as is - so in case there is a problem running the backgound load, we can configure lakeFS with slower rate.

nopcoder:
> LGTM, can you please add the follow up issue regarding refactoring of StoreMessage

Will have followup PR to remove `StoreMessage` from the code

nopcoder:
Documented the new configuration

PR #5137, opened 2023-01-30
Azure: Fix Azure ingest walker

### LINKED ISSUE #5136: Fix Azure ingest walker
Prefix is being redacted and passed to the pager empty

N-o-Z:
Close #5136


nopcoder:
@N-o-Z can you verify this is a fix - the current test cover the case the base dir can be an empty path.

N-o-Z:
> @N-o-Z can you verify this is a fix - the current test cover the case the base dir can be an empty path.

@nopcoder look at the example of gcs.
What happens is that you are importing the entire container without taking into account the requested prefix (folder).

N-o-Z:
@nopcoder  Added an additional test validation the demonstrates the issue

nopcoder (approved):
Thanks, I missed this one when aligned the implementaions and added the additional test code.

PR #5139, opened 2023-01-30
[docs] Write up a clear comparison of commands and workings of git vs lakeFS

### LINKED ISSUE #5138: Understanding basic differences between using lakeFS and Git
Sometimes users first interacting with lakeFS can't clearly understand the differences between working locally against a git repository and using lakeFS to work in isolation against the data, without copying it locally. 
Better documentation would help.

iddoavn:
FIX #5138 

Better explanation of differences between working with Git and working with lakeFS while the data stays in place.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -174,6 +174,16 @@ Git had conquered the world of code because it had best supported engineering be
 
 lakeFS provides these exact benefits, that are data practitioners are missing today, and enables them a clear intuitive git-like inetrface to easily manage their data like they manage code. Therefore, lakeFS can definitely be regarded as git for data.
 
+### How is lakeFS different from Git?
+
+An obvious question might be: Why not actually use “Git”, for data (...as opposed to a solution that provides “Git for data” :) )
+
+A basic requirement of working with big data, is that the data stays in place. When writing code, it is common to “check out” a local version of the code to your desktop and work against a copy of the code. However, when working with data, it is unreasonable to copy terabytes (or more) of files locally to experiment, develop, test or transform the data.
+
+Working with lakeFS, the data stays in place and the [branch isolation is achieved via pointers manipulations](https://docs.lakefs.io/understand/model.html#objects), pointing to the data staying on the object store. Meaning, you won’t check out a local copy of the data, but rather create a separate branch and work against that branch. Once done, if you want to “push” the changes back, merge back your isolated branch.
+
+On the other hand, unlike Git, lakeFS is not typically used to version control code. Instead, add metadata to your commit to link a data version (commit) with a code version (or an execution step in a DAG, or anything else that could be useful to understand the context of the data).
+
 ### Downloads
 
 #### Binary Releases


docs/index.md
```diff
@@ -174,6 +174,16 @@ Git had conquered the world of code because it had best supported engineering be
 
 lakeFS provides these exact benefits, that are data practitioners are missing today, and enables them a clear intuitive git-like inetrface to easily manage their data like they manage code. Therefore, lakeFS can definitely be regarded as git for data.
 
+### How is lakeFS different from Git?
+
+An obvious question might be: Why not actually use “Git”, for data (...as opposed to a solution that provides “Git for data” :) )
+
+A basic requirement of working with big data, is that the data stays in place. When writing code, it is common to “check out” a local version of the code to your desktop and work against a copy of the code. However, when working with data, it is unreasonable to copy terabytes (or more) of files locally to experiment, develop, test or transform the data.
```
rmoff, line 181: It feels like this para and the next conflate concern with where the data is (local vs object store) with whether a copy is made of it or not (regardless of where that copy it). 
Since we're trying to answer "how is lakeFS different from Git" maybe we need to address (a) why not just use git [git makes local copies, which we don't want to do because it's impractical & inefficient], and then (b) at a simple level how does lakeFS differ in its approach?

ortz:
@rmoff I started working on a draft for this, I'm adding it below, maybe it'll help and give ideas.
If you like, I have a example of this running within the docs (also attached, below)

# lakeFS vs. Git

## In a Nutshell

lakeFS is built on top of Git's primitives, and therefore, provides similarities for some of the lifecycle management commands.
Due to it's nature, and the fact that lakeFS is designed to handle large binary files and datasets, while Git is optimized for text files and source code, there are also differences that are important to understand.

Interested in understanding more on lakeFS model? [Read the documentation](../understand/model.md) 

## Similar commands

| lakeFS Command   | Git Command  | Description                                                       |
|------------------|--------------|-------------------------------------------------------------------|
| `lakectl branch` | `git branch` | Create and manage branches                                        |
| `lakectl commit` | `git commit` | Commit changes on a given branch                                  |
| `lakectl log`    | `git log`    | Show log of commits                                               |
| `lakectl show`   | `git show`   | Show detailed information about an entity by ID                   |
| `lakectl tag`    | `git tag`    | Create and manage tags                                            |
| `lakectl merge`  | `git merge`  | Merge & commit changes from source branch into destination branch |

## Differences

| lakeFS Command      | Git Command | Description                                                       | Why?                                                                                                                                                                                                                  |
|---------------------|-------------|-------------------------------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| `lakectl fs upload` | `git push`  | Update remote refs along with associated objects                  | Unlike Git, which is VCS for local development, lakeFS is a VCS for data-lakes, which are hosted on remote servers, therefore, there's no "local" staging to push.                                                    |
| N/A                 | `git clone` | Clone a repository into a new directory                           | lakeFS doesn't support cloning a repository as in lakeFS it represent a data-lake, which is usually VERY large and makes no sense to clone, if you like to copy specific objects, `lakectl fs download` is available. |
| `lakectl actions`   | N/A         | Manage Actions commands                                           | lakeFS provides built in capabilities to execute commands for governance, quality and assurance.                                                                                                                      |


# Example Screenshot
![lakefs-and-git](https://user-images.githubusercontent.com/1756345/216057660-ed638019-9089-4679-aa7e-7cea42f61efa.gif)

adipolak:
This looks good to me. wdyt? @ortz @iddoavn @rmoff ?

ortz (approved):
Overall looks good to me.
I do think that we'll benefit from a side-by-side comparison of the different commands, at least the ones we share.

rmoff:
add this too from Community Slack thread

> lakeFS does not have the equivalent of `git mv`.
> The main reason for this is that S3 itself does not support this operation. The equivalent in S3 is to perform a copy and delete.
> We played around with the idea and realized the complexities this creates far outweighs its advantages, mostly due to the fact that this flow cannot be implemented atomically with the KV interface we provide today (edited)

rmoff:
[from community slack](https://lakefs.slack.com/archives/C02CV7MUV4G/p1682529097277269)

> When there's an _uncomitted change_ on a branch, is that readable by anyone reading that branch?And so a _commit_ is just a way of delineating and referencing particular changes to a branch, rather than anything to do with making visible the change, is that right?

That's correct. Additionally, merging from `source` to `destination` will only merge the latest committed data on `source`.

This is also where [protected branches](https://docs.lakefs.io/howto/protect-branches.html) come in handy - they stop you from writing directly to a branch, effectively not allowing uncommitted changes.

nopcoder:
@ozkatz do we like to continue with this one or close it?

nopcoder:
Closing this one, please reopen if you think it is still relevant.

PR #5145, opened 2023-01-31
Email subscription drop rate

### LINKED ISSUE #4958: Email subscription drop rate
We'd like to know how many people churn due to the email subscription feature.

eladlachmi:
Resolves #4958


pkg/api/controller.go
```diff
@@ -3511,9 +3511,13 @@ func (c *Controller) GetSetupState(w http.ResponseWriter, r *http.Request) {
 		writeError(w, r, http.StatusInternalServerError, err)
 		return
 	}
+
+	if savedState == auth.SetupStateNotInitialized {
+		c.Collector.CollectEvent(stats.Event{Class: "global", Name: "preinit", Client: httputil.GetRequestLakeFSClient(r)})
```
nopcoder, line 3516: This is a new metric you understand how many users are still in preinit state?
The call for GetSetupState is called many names - each login if I remember to understand if we need to redirect the user to the setup flow.
The problem will be that for external API we always in not initialized mode and we will get this event on all checks.

pkg/api/controller.go
```diff
@@ -3511,6 +3511,10 @@ func (c *Controller) GetSetupState(w http.ResponseWriter, r *http.Request) {
 		writeError(w, r, http.StatusInternalServerError, err)
 		return
 	}
+
+	if !c.Config.IsAuthTypeAPI() && savedState == auth.SetupStateNotInitialized {
```
nopcoder, line 3515: Suggest to merge the two condition into if/else
```go
state := string(savedState)
// no need to create an admin user if users are managed externally
if c.Config.IsAuthTypeAPI() {
    state = string(auth.SetupStateInitialized)
} else if savedState == auth.SetupStateNotInitialized {
    c.Collector.CollectEvent(stats.Event{Class: "global", Name: "preinit", Client: httputil.GetRequestLakeFSClient(r)})
}
```

nopcoder (approved):
LGTM - added suggestion for the condition check

pkg/api/controller.go
eladlachmi: Done. Thanks!

PR #5149, opened 2023-01-31
Add compatibility checks for lakeFSFS

### LINKED ISSUE #4668: Discovery: How to test generated clients compatibility

itaiad200:
closes #4668 

The new compatibility tests will run on every pull request. They will:
1. Build lakeFS.
2. Build Hadoop lakeFS client.
3. Test the built lakeFS against all compatible client versions.
4. Test the built client against all compatible server versions.

How should this be maintain:
1. Release of a new lakeFS version -> Add that version to client compatibility tests.
2. Release of a new lakeFSFS version -> Add that version to the server compatibility tests.
3. The workflow fail?
	- If it's reasonable to deprecate the failing version, remove it from the list.
	- If it isn't, make the effort to support the older versions.

### DIFF
--- a/.github/workflows/compatability-tests.yaml
+++ b/.github/workflows/compatability-tests.yaml
@@ -66,16 +66,6 @@ jobs:
             name: generated-code
             path: /tmp/generated.tar.gz
 
-#      - name: Package Spark App
-#        working-directory: test/spark/app
-#        run: sbt sonnets-311/package
-#
-#      - name: Store Sonnets Spark app
-#        uses: actions/upload-artifact@v3
-#        with:
-#          name: generated-sonnets
-#          path: test/spark/app/target/sonnets-311/target/sonnets-311/scala-2.12/sonnets-311_2.12-0.1.0.jar
-
       - name: Build Spark direct-access client
         working-directory: clients/hadoopfs
         run: |
@@ -216,7 +206,7 @@ jobs:
         run: docker-compose logs --tail=15000 lakefs
 
   compatibility-checks-server:
-    name: Test lakeFS server FileSystem ccompatibility
+    name: Test lakeFS server FileSystem compatibility
     needs: [gen-code, deploy-image]
     strategy:
       fail-fast: false
--- a/.github/workflows/compatability-tests.yaml
+++ b/.github/workflows/compatability-tests.yaml
@@ -1,6 +1,7 @@
 name: lakeFS HadoopFS Compatability Tests
 
 on:
+  pull_request:
   push:
     branches:
       - master
@@ -216,9 +217,7 @@ jobs:
           path: /tmp
 
       - name: Load Docker image
-        run: |
-          docker load --input /tmp/lakefs.tar
-          docker image ls -a
+        run: docker load --input /tmp/lakefs.tar
 
       - uses: actions/setup-java@v3
         with:
@@ -284,5 +283,5 @@ jobs:
         payload: |
           {
             "mrkdwn": true,
-            "text": "Compatibility tests failure in master branch: https://github.com/treeverse/lakeFS/actions/workflows/compatability-tests.yaml"
+            "text": "Compatibility tests failure in master branch: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}"
           }
--- a/.github/workflows/compatability-tests.yaml
+++ b/.github/workflows/compatability-tests.yaml
@@ -1,7 +1,9 @@
 name: lakeFS HadoopFS Compatability Tests
 
 on:
-  pull_request:
+  push:
+    branches:
+      - master
 
 # These permissions are needed to interact with GitHub's OIDC Token endpoint.
 permissions:
@@ -282,5 +284,5 @@ jobs:
         payload: |
           {
             "mrkdwn": true,
-            "text": "Compatibility tests failure in master branch, [link](https://github.com/treeverse/lakeFS/actions/workflows/compatability-tests.yaml)"
+            "text": "Compatibility tests failure in master branch: https://github.com/treeverse/lakeFS/actions/workflows/compatability-tests.yaml"
           }
--- a/.github/workflows/compatability-tests.yaml
+++ b/.github/workflows/compatability-tests.yaml
@@ -279,4 +279,8 @@ jobs:
       env:
         SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
       with:
-        slack-message: "Compatibility tests failure in master branch: https://github.com/treeverse/lakeFS/actions/workflows/compatability-tests.yaml"
+        payload: |
+          {
+            "mrkdwn": true,
+            "text": ("Compatibility tests failure in master branch <https://github.com/treeverse/lakeFS/actions/workflows/compatability-tests.yaml+"|View>")
+          }
--- a/.github/workflows/compatability-tests.yaml
+++ b/.github/workflows/compatability-tests.yaml
@@ -272,7 +272,7 @@ jobs:
     name: Notify slack on workflow failures
     needs: [compatibility-checks-client, compatibility-checks-server]
     runs-on: ubuntu-20.04
-#    if: ${{ always() && contains(needs.*.result, 'failure') }}
+    if: ${{ always() && contains(needs.*.result, 'failure') }}
     steps:
     - name: slack-send
       uses: slackapi/slack-github-action@v1.23.0
--- a/.github/workflows/compatability-tests.yaml
+++ b/.github/workflows/compatability-tests.yaml
@@ -1,7 +1,6 @@
 name: lakeFS HadoopFS Compatability Tests
 
 on:
-  pull_request:
   push:
     branches:
       - master
--- a/.github/workflows/compatability-tests.yaml
+++ b/.github/workflows/compatability-tests.yaml
@@ -1,0 +1,300 @@
+name: Compatability Tests
+
+on:
+  pull_request:
+
+# These permissions are needed to interact with GitHub's OIDC Token endpoint.
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  gen-code:
+    name: Generate code from latest lakeFS app
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Check-out code
+        uses: actions/checkout@v3
+
+      # No way to share code between workflows :-( If you change this, find and change the
+      # same code wherever "Find Go module and build caches" appears!
+      - name: Find Go module and build caches
+        run: |
+          echo GOMODCACHE=`go env GOMODCACHE` >> $GITHUB_ENV
+          echo GOCACHE=`go env GOCACHE` >> $GITHUB_ENV
+          cat $GITHUB_ENV
+
+      - name: Setup Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.19.2
+        id: go
+
+      - name: Cache Go modules and builds
+        uses: actions/cache@v3
+        env:
+          cache-name: cache-go-modules
+        with:
+          path: |
+            ${{ env.GOMODCACHE }}
+            ${{ env.GOCACHE }}
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('go.mod', 'go.sum') }}
+          restore-keys: ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+
+      - uses: actions/setup-node@v3
+        with:
+          node-version: "16.17.1"
+
+      - uses: actions/setup-java@v3
+        with:
+          distribution: "adopt-hotspot"
+          java-version: "8"
+          cache: "sbt"
+
+      - name: Generate code
+        run: |
+          make -j3 gen-api VERSION=${{ steps.version.outputs.tag }}
+          mkdir webui/dist
+          touch webui/dist/index.html
+          tar -cf /tmp/generated.tar.gz .
+
+      - name: Store generated code
+        uses: actions/upload-artifact@v3
+        with:
+            name: generated-code
+            path: /tmp/generated.tar.gz
+
+#      - name: Package Spark App
+#        working-directory: test/spark/app
+#        run: sbt sonnets-311/package
+#
+#      - name: Store Sonnets Spark app
+#        uses: actions/upload-artifact@v3
+#        with:
+#          name: generated-sonnets
+#          path: test/spark/app/target/sonnets-311/target/sonnets-311/scala-2.12/sonnets-311_2.12-0.1.0.jar
+
+      - name: Build Spark direct-access client
+        working-directory: clients/hadoopfs
+        run: |
+          mvn -Passembly -Djar.finalName=client -DskipTests --batch-mode --update-snapshots package
+
+      - name: Store client assembly
+        uses: actions/upload-artifact@v3
+        with:
+          name: client-assembly
+          path: clients/hadoopfs/target/client.jar
+
+  deploy-image:
+    name: Build and cache Docker image
+    needs: [gen-code]
+    runs-on: ubuntu-20.04
+    outputs:
+      tag: ${{ steps.version.outputs.tag }}
+      image_id: ${{ steps.build_export.outputs.ImageID }}
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Setup NodeJS
+        uses: actions/setup-node@v3
+        with:
+          node-version: "16.17.1"
+
+      - name: Setup Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: 1.19.2
+        id: go
+
+      - name: Retrieve generated code
+        uses: actions/download-artifact@v3
+        with:
+          name: generated-code
+          path: /tmp/
+
+      - name: Unpack generated code
+        run: tar -xf /tmp/generated.tar.gz
+
+      - name: Extract version
+        shell: bash
+        run: echo "tag=sha-$(git rev-parse --short HEAD | sed s/^v//g)" >> $GITHUB_OUTPUT
+        id: version
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Build and export
+        uses: docker/build-push-action@v2
+        id: build_export
+        with:
+          context: .
+          file: ./Dockerfile
+          tags: treeverse/lakefs:${{ steps.version.outputs.tag }}
+          outputs: type=docker,dest=/tmp/lakefs.tar
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: lakefs-image
+          path: /tmp/lakefs.tar
+
+  compatibility-checks-client:
+    name: Test lakeFS Hadoop FileSystem compatibility
+    needs: gen-code
+    strategy:
+      fail-fast: false
+      matrix:
+        # Removing a version from this list means the published client is no longer compatible with
+        # that lakeFS version.
+        lakeFS_version: [ 0.89.1, 0.90.0, 0.91.0, 0.92.0 ]
+    runs-on: ubuntu-20.04
+    env:
+      TAG: ${{ matrix.lakeFS_version }}
+      REPO: treeverse
+      SPARK_TAG: 3
+    steps:
+      - name: Check-out code
+        uses: actions/checkout@v3
+
+      - uses: actions/setup-java@v3
+        with:
+          distribution: "adopt-hotspot"
+          java-version: "8"
+          cache: "sbt"
+
+      - name: Package Spark App
+        working-directory: test/spark/app
+        run: sbt sonnets-311/package
+
+      - name: Generate uniquifying value
+        id: unique
+        run: echo "value=$RANDOM" >> $GITHUB_OUTPUT
+
+      - name: Start lakeFS for Spark tests
+        uses: ./.github/actions/bootstrap-test-lakefs
+        with:
+          compose-directory: test/spark
+        env:
+          REPO: treeverse
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          LAKEFS_DATABASE_TYPE: postgres
+          LAKEFS_BLOCKSTORE_TYPE: s3
+          LAKEFS_BLOCKSTORE_S3_CREDENTIALS_ACCESS_KEY_ID: ${{ secrets.ESTI_AWS_ACCESS_KEY_ID }}
+          LAKEFS_BLOCKSTORE_S3_CREDENTIALS_SECRET_ACCESS_KEY: ${{ secrets.ESTI_AWS_SECRET_ACCESS_KEY }}
+
+      - name: Setup lakeFS for tests
+        working-directory: test/spark
+        run: ./setup-test.sh
+
+      - name: Retrieve client
+        uses: actions/download-artifact@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/compatability-tests.yaml
```diff
@@ -0,0 +1,300 @@
+name: Compatability Tests
+
+on:
+  pull_request:
```
itaiad200, line 4: Another option is to trigger this only on push to `master`. The incentive is to save resources, this action alone is probably ~half of an Esti run.

.github/workflows/compatability-tests.yaml
```diff
@@ -0,0 +1,300 @@
...
+          touch webui/dist/index.html
+          tar -cf /tmp/generated.tar.gz .
+
+      - name: Store generated code
+        uses: actions/upload-artifact@v3
+        with:
+            name: generated-code
+            path: /tmp/generated.tar.gz
+
+#      - name: Package Spark App
```
talSofer, line 69: Is this intentionally kept here?

.github/workflows/compatability-tests.yaml
```diff
@@ -0,0 +1,300 @@
...
+        run: ./run-test.sh
+
+      - name: lakeFS Logs on Spark with client failure
+        if: ${{ failure() }}
+        continue-on-error: true
+        working-directory: test/spark
+        run: docker-compose logs --tail=15000 lakefs
+
+  compatibility-checks-server:
+    name: Test lakeFS server FileSystem ccompatibility
```
talSofer, line 219: nit;
```suggestion
    name: Test lakeFS server FileSystem compatibility
```

.github/workflows/compatability-tests.yaml
```diff
@@ -0,0 +1,300 @@
+name: Compatability Tests
```
talSofer, line 1: Can you please edit both release checklists (lakefs and hadoopfs) to include a step to update the matrixes in this test?

.github/workflows/compatability-tests.yaml
```diff
@@ -0,0 +1,300 @@
...
+        working-directory: test/spark
+        run: ./setup-test.sh
+
+      - name: Retrieve client
+        uses: actions/download-artifact@v3
+        with:
+          name: client-assembly
+          path: clients/hadoopfs/target
+
+      - name: Test lakeFS S3 with Spark 3.x thick client
```
talSofer, line 199: Can you please also test with Spark 2.x? see https://github.com/treeverse/lakeFS/blob/d996d9eb603ec13cf5499380b4ec4078dfd6b7de/.github/workflows/esti.yaml#L709 for a reference

.github/workflows/compatability-tests.yaml
talSofer: I like the option of running only on push to master, I'm thinking about how we can enforce a failure is handled - maybe we can automate a bug creation and assign it to the pr owner? 

Another suggestion is to run it only on pull requests that include changes to hadoopfs and lakeFS API client and on push to master.

talSofer (commented):
Thanks @itaiad200, this is a VERY valuable test!! thanks for working on it :) 

Added my thoughts on how frequently we should run this and a few more comments.

.github/workflows/compatability-tests.yaml
Jonathan-Rosenberg, line 219:
```suggestion
    name: Test lakeFS server FileSystem compatibility
```

.github/workflows/compatability-tests.yaml
```diff
@@ -0,0 +1,300 @@
...
+
+  compatibility-checks-client:
+    name: Test lakeFS Hadoop FileSystem compatibility
+    needs: gen-code
+    strategy:
+      fail-fast: false
+      matrix:
+        # Removing a version from this list means the published client is no longer compatible with
+        # that lakeFS version.
+        lakeFS_version: [ 0.89.1, 0.90.0, 0.91.0, 0.92.0 ]
```
Jonathan-Rosenberg, line 152: so if there's a new version we need to remember to add it to the list?
can this be automated?

.github/workflows/compatability-tests.yaml
```diff
@@ -0,0 +1,300 @@
...
+
+#      - name: Package Spark App
+#        working-directory: test/spark/app
+#        run: sbt sonnets-311/package
+#
+#      - name: Store Sonnets Spark app
+#        uses: actions/upload-artifact@v3
+#        with:
+#          name: generated-sonnets
+#          path: test/spark/app/target/sonnets-311/target/sonnets-311/scala-2.12/sonnets-311_2.12-0.1.0.jar
```
ortz, line 77: forgot to remove?

ortz (commented):
Nice!

Some questions :)

.github/workflows/compatability-tests.yaml
```diff
@@ -0,0 +1,300 @@
...
+
+  compatibility-checks-server:
+    name: Test lakeFS server FileSystem ccompatibility
+    needs: [gen-code, deploy-image]
+    strategy:
+      fail-fast: false
+      matrix:
+        # Removing a version from this list means the current lakeFS is no longer compatible with
+        # that Hadoop lakeFS client version.
+        client_version: [ 0.1.10, 0.1.11, 0.1.12 ]
```
Jonathan-Rosenberg, line 226: same as the [above](https://github.com/treeverse/lakeFS/pull/5149/files#r1096720620) comment

.github/workflows/compatability-tests.yaml
itaiad200: I will as soon as this is merged!

.github/workflows/compatability-tests.yaml
itaiad200: Yeah, I was also thinking of sending a war-room slack message when failure happens.
Regarding your other suggestion, that won't catch all breaking changes. Consider
```
func (c *Controller) PutObject(...) (...){
    panic("I didn't change the API!!!")
}
```

.github/workflows/compatability-tests.yaml
itaiad200: Nope, removing

.github/workflows/compatability-tests.yaml
itaiad200: Done.

.github/workflows/compatability-tests.yaml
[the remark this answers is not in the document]
itaiad200: I'm looking for a sane way to do it. Releases don't have tags 🤷 
Doing a `>=V0.x.y` is slightly better than what we have here, but sometimes the breaking changes aren't linearly increasing (For example, a known bad version with a breaking change that was fixed).

itaiad200: Currently yes. It can be automated, I'm open to suggestions. 
On the other hand, I kind of feel this should be explicit "This version is compatible with master. This version isn't".

.github/workflows/compatability-tests.yaml
itaiad200: I can, but should we? I was under the impression that lakeFSFS uses the API client in hadoop2 the same as it uses it in hadoop3. Can the compatibility of the API break in one and not the other?
Hadoop2 testing keeps on running in Esti, this was not intended to replace it.

.github/workflows/compatability-tests.yaml
itaiad200: Done.

.github/workflows/compatability-tests.yaml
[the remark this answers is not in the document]
itaiad200: See previous response regarding automations.

Not sure I follow the ua part. When the client breaks you'll get `Test lakeFS server FileSystem ccompatibility (client_version 0.1.11)`, what do you need to ua for?

.github/workflows/compatability-tests.yaml
talSofer: It's Spark2 rather than hadoop2, but you are making a good point the client usage is the same for both Spark versions so no need to add this.

.github/workflows/compatability-tests.yaml
talSofer: I know that it won't catch all breaking changes but will catch some of them before they are merged

.github/workflows/compatability-tests.yaml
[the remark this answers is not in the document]
ortz: I understand that the test will fail, I think that the client version and the reported version in UA should match for better visibility.
It'll help other operations (such as lakeFS Cloud) to determine beforehand which version is being used.

.github/workflows/compatability-tests.yaml
Jonathan-Rosenberg: it's just so prone to be forgotten... 
I wonder if, upon a new lakeFS release, this list can be updated automatically. 
We can save that list as a GitHub secret and append to it the most recent version when released. 
If any version becomes incompatible, it can be removed (by a workflow or by hand) from the list.

.github/workflows/compatability-tests.yaml
itaiad200: But then we're sacrificing coverage. Our coverage is already limited as the Sonnets app is far from covering the entire lakeFS API. I prefer the tests we have for the endpoints is running and catching breaking changes instead of skipping them.

.github/workflows/compatability-tests.yaml
itaiad200: A Github secret means that no one can see the current value of the secret and it's very hard to maintain. What value does the secret add on top of the list here? 

Version becoming incompatible is now a decision. How can a workflow decide if a version was broken due to a bug or known deprecation?

.github/workflows/compatability-tests.yaml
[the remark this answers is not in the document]
itaiad200: Not sure I follow. What information is missing from the client_version that the UA will provide? How will the cloud leverage that?

.github/workflows/compatability-tests.yaml
Jonathan-Rosenberg: It doesn't have to be a secret, it can be a GitHub variable and run as specified [here](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-using-contexts-to-create-matrices) (only for vars).
The value that I see being added is the automated part that is adding the **new** version to the list instead of us humans needing to remember or to follow a step in the release process spec which we'll, of course, forget at some point.
Version becoming incompatible will remain a decision: if an older version will need to be deleted this can be done manually or via another workflow.

Having said all that, it might be done as part of a new PR and not this one...

.github/workflows/compatability-tests.yaml
```diff
@@ -0,0 +1,290 @@
...
+          LAKEFS_DATABASE_TYPE: postgres
+          LAKEFS_BLOCKSTORE_TYPE: s3
+          LAKEFS_BLOCKSTORE_S3_CREDENTIALS_ACCESS_KEY_ID: ${{ secrets.ESTI_AWS_ACCESS_KEY_ID }}
+          LAKEFS_BLOCKSTORE_S3_CREDENTIALS_SECRET_ACCESS_KEY: ${{ secrets.ESTI_AWS_SECRET_ACCESS_KEY }}
+
+      - name: Setup lakeFS for tests
+        working-directory: test/spark
+        run: ./setup-test.sh
+
+      - name: Retrieve client
```
nopcoder, line 183: spark-submit + --package will download the lakeFSFS for us, why do we need this step?

.github/workflows/compatability-tests.yaml
```diff
@@ -0,0 +1,290 @@
...
+          tags: treeverse/lakefs:${{ steps.version.outputs.tag }}
+          outputs: type=docker,dest=/tmp/lakefs.tar
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: lakefs-image
+          path: /tmp/lakefs.tar
+
+  compatibility-checks-client:
```
nopcoder, line 134: Suggestion - use services instead of the docker-compose. We can have spark run once for example and use it for testing all the clients.

.github/workflows/compatability-tests.yaml
```diff
@@ -0,0 +1,290 @@
...
+
+      - name: Start lakeFS for Spark tests
+        uses: ./.github/actions/bootstrap-test-lakefs
+        with:
+          compose-directory: test/spark
+        env:
+          REPO: treeverse
+          AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          LAKEFS_DATABASE_TYPE: postgres
```
nopcoder, line 264: we can work local too

nopcoder (changes requested):
Nice matrix - I think we can improve the flow to run with less steps and make it more dynamic for changes/releases.

.github/workflows/compatability-tests.yaml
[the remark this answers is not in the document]
ortz: I see, so at least in the meantime, I suggest adding it to the lakeFS/client release process (to make sure this list is up to date, and maybe with some guidelines next to it; e.g how many versions we support backward?)

.github/workflows/compatability-tests.yaml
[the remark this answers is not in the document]
ortz: it's the other way around, I'd like to make sure that current cloud clients' versions (which are reported using the UA) are aligned with this list.
anyway, if you don't think that's an issue or something that we should address in this scope, let's defer it.

.github/workflows/compatability-tests.yaml
itaiad200: That's a test for the latest client with the server versions. The latest client is not published.

.github/workflows/compatability-tests.yaml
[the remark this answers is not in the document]
itaiad200: I would like to help, just not sure I understand how. UA gives you the client version, right? It's written here directly, so I'm confused on what the UA provides that's not already known.

.github/workflows/compatability-tests.yaml
itaiad200: Correct, missed that. I think I'll keep it this way for parity with the same Esti test.

test/spark/docker-compose.yaml
```diff
@@ -24,7 +24,7 @@ services:
       - LAKEFS_AUTH_ENCRYPT_SECRET_KEY=some random secret string
       - LAKEFS_DATABASE_TYPE=postgres
       - LAKEFS_DATABASE_POSTGRES_CONNECTION_STRING=postgres://lakefs:lakefs@postgres/postgres?sslmode=disable
-      - LAKEFS_LOGGING_LEVEL=DEBUG
+      - LAKEFS_LOGGING_LEVEL=TRACE
```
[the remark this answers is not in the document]
itaiad200: Done.

test/spark/run-test.sh
```diff
@@ -14,6 +14,12 @@ STORAGE_NAMESPACE=${STORAGE_NAMESPACE:-local://}
 REPOSITORY=${REPOSITORY:-example}
 SONNET_JAR=${SONNET_JAR:-sonnets-246/target/sonnets/246/scala-2.12/sonnets-246_2.12-0.1.jar}
 
+if [ -v CLIENT_VERSION ]; then # Direct access using thick Spark client.
```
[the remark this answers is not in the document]
itaiad200: Done.

test/spark/run-test.sh
```diff
@@ -14,6 +14,12 @@ STORAGE_NAMESPACE=${STORAGE_NAMESPACE:-local://}
...
+    CLIENT_PATH="--packages io.lakefs:hadoop-lakefs-assembly:"$CLIENT_VERSION
```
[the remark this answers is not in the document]
itaiad200: Done.

test/spark/run-test.sh
```diff
@@ -14,6 +14,12 @@ STORAGE_NAMESPACE=${STORAGE_NAMESPACE:-local://}
...
+else
+    CLIENT_PATH="--jars /target/client.jar"
```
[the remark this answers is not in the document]
itaiad200: Done.

test/spark/run-test.sh
```diff
@@ -22,7 +28,7 @@ docker-compose exec -T lakefs lakectl repo create "lakefs://${REPOSITORY}" ${STO
...
+    docker-compose run -v $PWD/../../clients/hadoopfs/target/:/target/ -T --no-deps --rm spark-submit sh -c 'spark-submit --master spark://spark:7077 '"${CLIENT_PATH}"' -c "spark.hadoop.fs.lakefs.impl=io.lakefs.LakeFSFileSystem" --conf=spark.driver.extraJavaOptions=-Dcom.amazonaws.services.s3.enableV4=true --conf=spark.executor.extraJavaOptions=-Dcom.amazonaws.services.s3.enableV4=true  -c spark.hadoop.fs.lakefs.endpoint=http://lakefs:8000/api/v1 -c "spark.hadoop.fs.lakefs.access.key=${TESTER_ACCESS_KEY_ID}" -c "spark.hadoop.fs.lakefs.secret.key=${TESTER_SECRET_ACCESS_KEY}" -c "spark.hadoop.fs.s3a.access.key=${AWS_ACCESS_KEY_ID}" -c "spark.hadoop.fs.s3a.secret.key=${AWS_SECRET_ACCESS_KEY}" -c "spark.hadoop.fs.s3a.region=${AWS_REGION}" --class Sonnets /local/app/target/'${SONNET_JAR}' ${input} ${output}'
```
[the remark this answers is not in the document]
itaiad200: Dropping. It took me forever to make it work, so I'll revert if it doesn't.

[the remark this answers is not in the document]
itaiad200: It's needed, reverting.

.github/workflows/compatability-tests.yaml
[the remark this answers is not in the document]
ortz: It does, and it's not aligned with this list.
For instance, `0.1.12` is one of the client versions listed here.
Looking at the client user-agent reported in our BI, I can see the following:

- lakefs-webui/0.90.0
- lakefs-webui/0.85.0
- lakefs-webui/0.88.0
- OpenAPI-Generator/0.82.0/python
- Hadoop 3.3.4, aws-sdk-java/1.12.189 Linux/5.4.0-1088-aws OpenJDK_64-Bit_Server_VM/25.302-b08 java/1.8.0_302 scala/2.12.14 vendor/Azul_Systems,_Inc. cfg/retry-mode/legacy
- managed/gc
- lakefs-webui/dev
- lakefs-python-sdk/0.83.3
- lakefs-python-sdk/0.83.4
- lakefs-python-sdk/0.88.0

As you can see, none of these match the client's list, and it seems like it reports the server version.
Now, when an operator would like to upgrade lakeFS version, they'll have to look at this list and compare it to what we support, and they'll be blocked.
How would you suggest to the operator (or any lakeFS client TBH) to continue?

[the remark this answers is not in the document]
itaiad200: I got what you mean. You're referring to the generated client SDK version, which are released with each server versions (but are not the server versions!). Notice that the compatibility tests added here are just for the Hadoop lakeFS implementation and cannot provide compatibility gurantees for Python, GC, webui, etc. 
The generated client used by it is determined [here](https://github.com/treeverse/lakeFS/blob/67854f201eccce26feda9996ff06fcdebee36217/clients/hadoopfs/pom.xml#L251). If the HadoopFS doesn't report its version in the UA we'll need to fix that.

.github/workflows/compatability-tests.yaml
```diff
@@ -0,0 +1,288 @@
...
+    steps:
+    - name: slack-send
+      uses: slackapi/slack-github-action@v1.23.0
+      env:
+        SLACK_WEBHOOK_URL: ${{ secrets.SLACK_WEBHOOK_URL }}
+      with:
+        payload: |
+          {
+            "mrkdwn": true,
+            "text": "Compatibility tests failure in master branch: https://github.com/treeverse/lakeFS/actions/workflows/compatability-tests.yaml"
```
nopcoder, line 287: Check if this will link directly to the action - `${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}`

.github/workflows/compatability-tests.yaml
```diff
@@ -0,0 +1,288 @@
...
+      - name: Download artifact
+        uses: actions/download-artifact@v3
+        with:
+          name: lakefs-image
+          path: /tmp
+
+      - name: Load Docker image
+        run: |
+          docker load --input /tmp/lakefs.tar
+          docker image ls -a
```
nopcoder, line 221: no need to list images
```suggestion
        run: docker load --input /tmp/lakefs.tar
```

.github/workflows/compatability-tests.yaml
```diff
@@ -0,0 +1,288 @@
...
+      TAG: ${{ needs.deploy-image.outputs.tag }}
+      IMAGE_ID: ${{ needs.deploy-image.outputs.image_id }}
+      REPO: treeverse
+      SPARK_TAG: 3
+    steps:
+      - name: Check-out code
+        uses: actions/checkout@v3
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
```
nopcoder, line 210: Don't think we need this one - not using buildx here

nopcoder (changes requested):
Thanks for addressing the issues. Added more feedback.

.github/workflows/compatability-tests.yaml
```diff
@@ -0,0 +1,290 @@
...
+on:
+  pull_request:
+
+# These permissions are needed to interact with GitHub's OIDC Token endpoint.
+permissions:
+  id-token: write
+  contents: read
+
+jobs:
+  gen-code:
```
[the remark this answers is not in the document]
itaiad200: Done.

.github/workflows/compatability-tests.yaml
itaiad200: Removed.

.github/workflows/compatability-tests.yaml
itaiad200: Done.

.github/workflows/compatability-tests.yaml
itaiad200: Done.

.github/workflows/compatability-tests.yaml
itaiad200: Nope, it's needed.

nopcoder (approved):
LGTM - much shorter :) - There was additional dup that we need to remove.

.github/workflows/compatability-tests.yaml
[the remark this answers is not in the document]
itaiad200: Sorry, had to revert that. All jobs rely on the bootstrap lakeFS action which needs to generated code.

.github/workflows/compatability-tests.yaml
```diff
@@ -0,0 +1,258 @@
...
+          distribution: "adopt-hotspot"
+          java-version: "8"
+          cache: "sbt"
+
+      - name: Generate code
+        run: |
+          make -j3 gen-api VERSION=${{ steps.version.outputs.tag }}
+          mkdir webui/dist
+          touch webui/dist/index.html
+          tar -cf /tmp/generated.tar.gz .
```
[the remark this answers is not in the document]
itaiad200: Now I do :|

.github/workflows/compatability-tests.yaml
```diff
@@ -0,0 +1,258 @@
...
+
+      - name: Setup NodeJS
+        uses: actions/setup-node@v3
+        with:
+          node-version: "16.17.1"
```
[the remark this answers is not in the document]
itaiad200: Separated again..

PR #5152, opened 2023-01-31
proposal: branch checks

### LINKED ISSUE #3020: Support long running hooks / Branch status
The ability to support hooks that run for a long period of time (hours instead of seconds) and to do so without blocking branches for writes.

ozkatz:
co-authored by @nopcoder (THANK YOU!)

Implementing this closes #3020

### DIFF
--- a/design/open/branch_checks.md
+++ b/design/open/branch_checks.md
@@ -154,7 +154,7 @@ This could be done using a pre-merge hook that runs this validation, or at a hig
 
 #### CLI
 
-`lakectl check <ref uri>` to run the checks for the given ref
+`lakectl checks run <ref uri> [--id <check id>] ` to run the checks for the given ref
 
 `lakectl checks list <ref uri>` to show checks and their state for the given ref
 
--- a/design/open/branch_checks.md
+++ b/design/open/branch_checks.md
@@ -36,7 +36,7 @@ sequenceDiagram
     user1->>lakeFS: commit1(branch_a)
     user1->>lakeFS: commit2(branch_a)
     user1->>lakeFS: commitN(branch_a)
-    user1->>lakeFS: check(branch_a)
+    user1->>lakeFS: check(branch_a) // will be dereferenced
     loop For every configured check
       lakeFS->>+External Executor: execute(event details, callback URL)
       External Executor->>-lakeFS: callback(status)
@@ -82,9 +82,9 @@ checks:
 
 #### Check triggering
 
-Checks are triggered when the `check(repository_id, source_branch_id)` endpoint is called in the lakeFS API.
+Checks are triggered when the `check(repository_id, ref_id)` endpoint is called in the lakeFS API.
+When running check() with a branch or tag, it will be dereferenced: checks are valid for a given commit ID.
 
-lakeFS will store the state of all checks that were initiated by this action for the given HEAD of `source_branch_id`.
 
 #### Check execution
 
@@ -110,7 +110,7 @@ stateDiagram-v2
 
 #### Retrying checks
 
-From the `lost` and `failed` statuses, a retry operation is available that will cause the execute function to run again.
+From the `lost` and `failed` statuses, users can trigger a retry operationthat will cause the execute function to run again for the given check.
 This is used when transient errors occur (or a timeout).
 
 #### Getting check status
@@ -141,6 +141,9 @@ Each execution will send a `callback_token` to the executing system (Airflow, fo
 ```
 
 Where metadata is a simple `Map<string, string>` mapping to be sent by the executing system.
+
+When lakeFS receives a callback, it checks that the provided token is the last one issued for that (check, commit). 
+Any previous token issued will be ignored.
 
 #### Merging into a protected branch
 
--- a/design/open/branch_checks.md
+++ b/design/open/branch_checks.md
@@ -96,25 +96,17 @@ Like hooks, checks could have multiple type implementations (Airflow, Webhooks, 
 
 ```mermaid
 stateDiagram-v2
-    state start_failed <<choice>>
-    state execute_failed <<choice>>
-    state timed_out <<choice>>
-    [*] --> starting
-    starting --> start_failed
-    starting --> timed_out: no response from Airflow
-    start_failed --> executing: OK from airflow
-    start_failed --> failed: error from airflow
-    executing --> execute_failed
-    execute_failed --> success: execution callback returns success
-    execute_failed --> failed: execution callback returns failure
-    executing --> timed_out
-    timed_out --> lost: no response from Airflow within execute_timeout
-    failed --> retry
-    lost --> retry
-    retry --> starting
-    success --> [*]
+    direction LR
+    [*] --> STARTING: execute()
+    STARTING --> FAILED: error calling Airflow
+    STARTING --> EXECUTING
+    EXECUTING --> SUCCESS: callback(success)
+    EXECUTING --> FAILED: callback(error)
+    EXECUTING --> LOST: timeout
+    LOST --> STARTING: retry
+    FAILED --> STARTING: retry
+    SUCCESS --> [*]
 ```
-
 
 #### Retrying checks
 
--- a/design/open/branch_checks.md
+++ b/design/open/branch_checks.md
@@ -1,0 +1,176 @@
+# Branch Checks
+
+## User Story
+
+In order to achieve CI/CD for data, data engineering teams would like to validate the data itself before allowing to merge into a production branch.
+These validations might take longer than a typical HTTP request.
+
+## Goals
+
+Currently, lakeFS hooks allow running `pre-*` type hooks as blocking operations: They execute in a manner that is tied to an HTTP request.
+This puts a pretty short time window for hooks to complete in.
+
+Distributed processing frameworks such as Apache Spark typically have a long startup time and might take minutes (or even hours) to complete.
+
+The main goal of this proposal is to allow such validations to occur with a very long timeout (configurable, but could be days).
+
+## Non-Goals
+
+Out of scope for this proposal is any collaboration features that are typically associated with PRs (that provide similar functionality to the one being proposed).
+
+## High Level design
+
+This proposal adds a `Check` primitive to lakeFS branches. a Check is similar to a hook: it calls out to some external component with information about a modification or event that occurred in the system.
+
+The user flow for checks is the following:
+
+1. configure checks
+1. ensure all *mandatory* checks have passed for the given branch's HEAD
+1. merge said branch into a protected one.
+
+Merging into any protected branch requires all mandatory checks to have completed successfully for the current HEAD:
+
+```mermaid
+sequenceDiagram
+    Actor user1
+    user1->>lakeFS: commit1(branch_a)
+    user1->>lakeFS: commit2(branch_a)
+    user1->>lakeFS: commitN(branch_a)
+    user1->>lakeFS: check(branch_a)
+    loop For every configured check
+      lakeFS->>+External Executor: execute(event details, callback URL)
+      External Executor->>-lakeFS: callback(status)
+      lakeFS->>lakeFS: update check status (SUCCESS)
+    end
+    user1->>lakeFS: merge(branch_a, branch_b_protected)
+    lakeFS-->>user1: success!
+```
+
+## Implementation
+
+#### Check configuration
+
+Checks are configured under `_lakefs_actions/` within a lakeFS repository's root. Any YAML file within the directory can contain checks, similarly to hooks.
+Here's what a check would look like:
+
+```yaml
+name: Ensure Event counts are within bounds
+checks:
+  - id: great_expectations_validate_events
+    type: webhook
+    properties:
+      url: "http://<host:port>/checks/great_expectations"
+      query_params:
+        condition: expect_table_row_count_to_be_between(min_value=2000, max_value=5000)
+      headers:
+        secret_header: "{{ ENV.MY_SECRET }}"
+```
+
+#### Check event syntax
+
+```json
+{
+  "repository_id": "example-repo",
+  "branch_id": "branch_a",
+  "source_ref": "0307e79ab5afef2b0726a8c13896f716033b1acd8454e22210bd43a2f72a94a9",
+  "check_id": "great_expectations_validate_events",
+  "storage_namespace": "s3://path/to/storage/",
+  "callback_token": "abc123",
+  "output_url": "presigned URL or another HTTP POST-able URL"
+}
+```
+
+#### Check triggering
+
+Checks are triggered when the `check(repository_id, source_branch_id)` endpoint is called in the lakeFS API.
+
+lakeFS will store the state of all checks that were initiated by this action for the given HEAD of `source_branch_id`.
+
+#### Check execution
+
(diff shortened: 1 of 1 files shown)


design/open/branch_checks.md
```diff
@@ -0,0 +1,168 @@
...
+#### Merging into a protected branch
+
+A branch may define 0 or more checks that must be in a state of SUCCESS for the ref being merged into it.
+Once a merge is issued, the server will dereference the source branch and get back its list of checks to validate this status.
+
+This could be done using a pre-merge hook that runs this validation, or at a higher level using the branch protection settings.
+
+#### CLI
+
+`lakectl check <ref uri>` to run the checks for the given ref
```
N-o-Z, line 154:
```suggestion
`lakectl checks <ref uri>` to run the checks for the given ref
```

design/open/branch_checks.md
```diff
@@ -0,0 +1,171 @@
...
+  "metadata": {},
+}
+```
+
+Where metadata is a simple `Map<string, string>` mapping to be sent by the executing system.
+
+When lakeFS receives a callback, it checks that the provided token is the last one issued for that (check, commit). 
+Any previous token issued will be ignored.
+
+#### Merging into a protected branch
```
N-o-Z, line 148: Successful merges to the protected branch may introduce / remove checks to the list
These check might have not been run by potential merge branches or might have failed during their checks execution.
How do we plan to handle these kind of situations? What is the wanted behavior in these cases?

design/open/branch_checks.md
```diff
@@ -0,0 +1,171 @@
...
+}
+```
+
+#### Check triggering
+
+Checks are triggered when the `check(repository_id, ref_id)` endpoint is called in the lakeFS API.
+When running check() with a branch or tag, it will be dereferenced: checks are valid for a given commit ID.
+
+
+#### Check execution
```
N-o-Z, line 89: If I understand correctly - checks are run under "branch_a" code base - while status check is done from "branch_b" context.
Some questions:
1. If I want to remove a check from "branch_b" how can I do it and still be able to provided a successful status for it.
2. What is stopping an "uneducated" user (or malicious?) to modify checks to return "always true" and bypass the checks mechanism?

ozkatz: Good questions :)
1. pre-merge, the destination branch can validate that the required check ids have passed. if source branch didn't run them, they can merge destination into source (or rebase in the future) to make sure they have the correct checks.
2. use a policy that doesn't allow uneducated users (or say, only allows a specific power group) to write to `_lakefs_actions/` - this is the same answer for hooks.

design/open/branch_checks.md
ozkatz: Thanks, fixed

design/open/branch_checks.md
ozkatz: validation of required checks is done on the destination branch pre-merge. If a merge removes or changes the required checks, the new state will have to conform to whatever was merged. Does that make sense?

nopcoder (approved):

design/open/branch_checks.md
N-o-Z: I seem to be confused a bit on no.1
Let's say for example that I want to remove a required check on "branch_b". I create a branch called "branch_a" which removes a required check from "branch_b" and commit the changes. Now I run the check API against "branch_a" which creates a list of check results on the given head commit ref that is missing the specific check. Now when I want to merge "branch_a" to "branch_b" I have a problem since there is no status for a required check in "branch_b" for that given reference

design/open/branch_checks.md
```diff
@@ -0,0 +1,171 @@
...
+    end
+    user1->>lakeFS: merge(branch_a, branch_b_protected)
+    lakeFS-->>user1: success!
+```
+
+## Implementation
+
+#### Check configuration
+
+Checks are configured under `_lakefs_actions/` within a lakeFS repository's root. Any YAML file within the directory can contain checks, similarly to hooks.
```
N-o-Z, line 53: Suggest separating checks and hooks files either by a different path or a subdir. If it was up to me, I would allow only a single checks file with a specific path.
Putting the checks and hooks together creates an overhead of reading the entire actions path to filter only for the checks. You can have dozen of action files which may contain only a single check in one of them and you will need to read all the files from storage and parse them.

design/open/branch_checks.md
```diff
@@ -0,0 +1,171 @@
...
+Out of scope for this proposal is any collaboration features that are typically associated with PRs (that provide similar functionality to the one being proposed).
+
+## High Level design
+
+This proposal adds a `Check` primitive to lakeFS branches. a Check is similar to a hook: it calls out to some external component with information about a modification or event that occurred in the system.
+
+The user flow for checks is the following:
+
+1. configure checks
+1. ensure all *mandatory* checks have passed for the given branch's HEAD
```
N-o-Z, line 28: You are mentioning "mandatory" checks - are there also checks which are not mandatory? If so - where do we configure a check to be mandatory or optional?

N-o-Z (commented):
Sorry for the intermittent review, I added some more comments

design/open/branch_checks.md
```diff
@@ -0,0 +1,171 @@
...
+
+The main goal of this proposal is to allow such validations to occur with a very long timeout (configurable, but could be days).
+
+## Non-Goals
+
+Out of scope for this proposal is any collaboration features that are typically associated with PRs (that provide similar functionality to the one being proposed).
+
+## High Level design
+
+This proposal adds a `Check` primitive to lakeFS branches. a Check is similar to a hook: it calls out to some external component with information about a modification or event that occurred in the system.
```
N-o-Z, line 23: Are checks exclusive only to protected branches? Can they be enabled / disabled?
Similar to hooks we should consider using a configuration parameter to avoid unnecessary access to storage.

design/open/branch_checks.md
ozkatz: As long as mandatory check validation happens as a pre-merge hook (either directly or behind the scenes), hooks should run using a the merged state of the `_lakefs_actions/` directory. 

I'm not sure this is currently the case - but if not, it means that the problem is how we can update pre-merge hooks on protected branches. Checks should be mostly orthogonal to this..

design/open/branch_checks.md
ozkatz: Allowing users to decide how to manage these files is a property I'd like to retain. I can share a use case from existing users where they provision repositories automatically and have a process that writes a couple of post-commit hooks into `_lakefs_actions/`. 

Because each file is independent, automating such processes is easy - no need to read or parse the existing checks. AFAIK lakeFS doesn't care (or at least, shouldn't care) if objects under `_lakefs_actions/` include a `/` delimiter - so users can decide - and we can even recommend, that they store checks under `_lakefs_actions/checks/` and hooks under `_lakefs_actions/hooks/<event-type>/` to make it easier for humans to consume.

design/open/branch_checks.md
ozkatz: Mandatory checks are enforced by the destination branch.

I would suggest - and perhaps its a good idea to cover it in the proposal - that this would happen using a pre-merge hook.

design/open/branch_checks.md
ozkatz: Checks can happen on any branch. No need to enable or disable them - it's up to the user to call check() and ensure they run, if they so wish.

Using a pre-merge hook to enforce that a ref has been validated by a check of a certain ID allows users to create mandatory checks - but it's still up to the user to trigger a check.

design/open/branch_checks.md
```diff
@@ -0,0 +1,171 @@
...
+
+#### Check triggering
+
+Checks are triggered when the `check(repository_id, ref_id)` endpoint is called in the lakeFS API.
+When running check() with a branch or tag, it will be dereferenced: checks are valid for a given commit ID.
+
+
+#### Check execution
+
+Like hooks, checks could have multiple type implementations (Airflow, Webhooks, etc). Initially, only Airflow is supported.
```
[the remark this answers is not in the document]
ozkatz: By all means let's support them :) I feel that at this point, starting with a single implementation would be easier but if that's not the case, let's support all of them :)
AFAIK Airflow might need some more thought as it's currently only designed for post-* hooks, but i'll leave the implementation details up to you.

design/open/branch_checks.md
```diff
@@ -0,0 +1,171 @@
...
+
+Merging into any protected branch requires all mandatory checks to have completed successfully for the current HEAD:
+
+```mermaid
+sequenceDiagram
+    Actor user1
+    user1->>lakeFS: commit1(branch_a)
+    user1->>lakeFS: commit2(branch_a)
+    user1->>lakeFS: commitN(branch_a)
+    user1->>lakeFS: check(branch_a) // will be dereferenced
```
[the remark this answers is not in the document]
ozkatz: a check's status hold state for a given commit, not to a pointer to one. If I commit, check and then commit again, my check's state is no longer relevant.

design/open/branch_checks.md
```diff
@@ -0,0 +1,171 @@
...
+When lakeFS receives a callback, it checks that the provided token is the last one issued for that (check, commit). 
+Any previous token issued will be ignored.
+
+#### Merging into a protected branch
+
+A branch may define 0 or more checks that must be in a state of SUCCESS for the ref being merged into it.
+Once a merge is issued, the server will dereference the source branch and get back its list of checks to validate this status.
+
+This could be done using a pre-merge hook that runs this validation, or at a higher level using the branch protection settings.
+
```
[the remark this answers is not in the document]
ozkatz: 1. Checks don't run unless you trigger them.
2. When running checks, you don't yet know which branch you're merging to. Mandatory checks are defined on the receiver to separate concerns. Let's take a real world example:

Let's say I have a `users` table that has an `age` column. I can define a check on `main` whose ID is `ensure_age_is_between_0_and_125`. I can check this with a Spark job, with Great Expectations or with DuckDB, that's an implementation detail. It's up to the source branch to make sure this check passes and how it is implemented. 

This is also the case for e.g. GitHub actions - defining the mandatory check is a setting on the receiving branch, not within the action file itself.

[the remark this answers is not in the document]
ozkatz: 3. For GitHub PR's that's not the case. If you create a PR, all defined actions that trigger before a merge would happen. If you push another commit to the source branch of the PR - your actions would run again. 

Branch checks in lakeFS have the potential of being much more expensive - running a distributed system and potentially scanning large datasets. I don't think this behavior makes much sense there - this is why check() is an independent API and doesn't trigger automatically.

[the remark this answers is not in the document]
itaiad200: Thanks for clarifying, I still don't get on how the destination branch enforces the check on the source branch if they may diverge and apply different logic.
Example:
1. `main`: check `ensure_age`: validates the age is between 0 and 125.
2. created a branch `feature_branch`.
3. `main`: committed a change to the check that the age is now between 0 and 100.
4. `feature_branch`: committed insertion of ages between 100-125 and triggered `check()`. It passed.
5. Merge from `feature_branch` to `main` is successful since the required check passed.

design/open/branch_checks.md
```diff
@@ -0,0 +1,171 @@
...
+In order to achieve CI/CD for data, data engineering teams would like to validate the data itself before allowing to merge into a production branch.
+These validations might take longer than a typical HTTP request.
+
+## Goals
+
+Currently, lakeFS hooks allow running `pre-*` type hooks as blocking operations: They execute in a manner that is tied to an HTTP request.
+This puts a pretty short time window for hooks to complete in.
+
+Distributed processing frameworks such as Apache Spark typically have a long startup time and might take minutes (or even hours) to complete.
```
arielshaqed, line 13: Even short-lived tasks can end on a different node than where they started, or might run on a framework with an asynchronous API[^1], or _may_ be delayed.  This is really cool because it makes deploying reliable hooks a lot easier.

Also the experience it yields is a lot like GitHub actions.  An action _runs_ because something happened.  Eventually it reports a status, and then merging might be allowed.

[^1] E.g. in Airflow, it's really easy to kick off a hook and report a status at the last stage.  Arguably simpler than anything we already have, even _with_ our current Airflow support.

design/open/branch_checks.md
```diff
@@ -0,0 +1,171 @@
...
+
+```mermaid
+sequenceDiagram
+    Actor user1
+    user1->>lakeFS: commit1(branch_a)
+    user1->>lakeFS: commit2(branch_a)
+    user1->>lakeFS: commitN(branch_a)
+    user1->>lakeFS: check(branch_a) // will be dereferenced
+    loop For every configured check
+      lakeFS->>+External Executor: execute(event details, callback URL)
```
arielshaqed, line 41: More radical suggestion: separate concerns!  Each of these is useful on its own without necessarily all of the others.

* An _action_ starts executing when some trigger occurs: a new commit on a branch, or a new merge to the branch, or an explicit endpoint.  In future we may add new event types, just like GitHub.  This is a lot like current hooks/actions, except that it is asynchronous and there is no return value.

  Not all such actions will report a status.  Right now we call them post-* hooks IIUC, but they are equally valid as pre-* hooks.  If an action uses a reliable runner, it does not need to return a status.  For instance an "export" action just needs to trigger on commit, it's a post-commit.A "send email that a merge is starting" is a pre-merge action that does not return a status.
* An action can eventually _report status_ for a commit if it likes.  lakeFS associates this status with the commit.

  TBD: we may want to allow _anything_ to report status for a commit, regardless of whether it started as an action or not.  For example, an Airflow DAG may commit and then start validation.  It does not need lakeFS to schedule validation, indeed it may be better structured as a single DAG.  Now it can report status!
* A _protected branch_ may _require_ certain checks to pass in order to report status.

design/open/branch_checks.md
```diff
@@ -0,0 +1,171 @@
...
+    EXECUTING --> FAILED: callback(error)
+    EXECUTING --> LOST: timeout
+    LOST --> STARTING: retry
+    FAILED --> STARTING: retry
+    SUCCESS --> [*]
+```
+
+#### Retrying checks
+
+From the `lost` and `failed` statuses, users can trigger a retry operationthat will cause the execute function to run again for the given check.
```
arielshaqed, line 113:
```suggestion
From the `lost` and `failed` statuses, users can trigger a retry operation that will cause the execute function to run again for the given check.
```

design/open/branch_checks.md
```diff
@@ -0,0 +1,171 @@
...
+{
+  "status": "SUCCESS",
+  "metadata": {},
+}
+```
+
+Where metadata is a simple `Map<string, string>` mapping to be sent by the executing system.
+
+When lakeFS receives a callback, it checks that the provided token is the last one issued for that (check, commit). 
+Any previous token issued will be ignored.
```
arielshaqed, line 146: We may want to provide the opportunity to _cancel_ a previously-started check.  It's an effort to do it right.  But we're talking about actions that can run for _days_.  If I commit and 10 minutes later add a new commit "fix mess", there's no point wasting resourced for days.

Probably not required initially, but we should have the ability.  For instance we may want to allow the response to a "start" message to record a run token for lakeFS, and  use it to cancel that run later!

arielshaqed (approved):
As a lakeFS user, this will be _very_ useful.  NEAT!

nopcoder:
Closing - re-open if still relevant.

PR #5153, opened 2023-01-31
Delta Diff MVP: Plugins manager

### LINKED ISSUE #4965: Delta Diff MVP - Task: Create a go-plugin client abstraction

Jonathan-Rosenberg:
The plugin manager will be used to store `go-plugin` application clients and to load gRPC clients needed to communicate with gRPC server.
Each type of "plugin service" (like Diff Service for example) will initialize its own plugin manager with its plugin type.
This PR is the final version of #5024

---

**Optional**
This is a "`go-plugin` implementation" diagram to help understand this package:

go-plugin

- _Generated code_: Code generated by running `protoc`
- _Implemented code_: Code needed to be implemented by us.
- _go-plugin code_: Some related `go-plugin` code

Closes #4965


pkg/plugins/manager.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+func NewManager[T any]() *Manager[T] {
+	return &Manager[T]{
+		pluginApplicationClients: make(map[string]*plugin.Client),
+	}
+}
+
+// RegisterPlugin is used to register a new plugin client with the corresponding plugin type.
+func (m *Manager[T]) RegisterPlugin(name string, id PluginIdentity, auth PluginHandshake, p plugin.Plugin) error {
```
arielshaqed, line 55:
```suggestion
func (m *Manager[T]) RegisterPlugin(name string, id PluginIdentity, handshake PluginHandshake, p plugin.Plugin) error {
```

It's a handshake not an auth:

* No secrets, can _and should_ log it.
* Comes from regular configuration (e.g. K8s configmap not secret).

pkg/plugins/manager.go
```diff
@@ -0,0 +1,113 @@
...
+}
+
+// PluginIdentity identifies the plugin's version and executable location.
+type PluginIdentity struct {
+	Version int
+	// Cmd is the command that is used to run the plugin executable on the local filesystem.
+	Cmd *exec.Cmd
+}
+
+// PluginHandshake includes handshake properties for the plugin.
```
arielshaqed, line 30: I wonder, isn't this tied 1:1 with PluginIdentity?  I know we've been over this before, I don't see in the code why they need to be separate.

pkg/plugins/manager.go
```diff
@@ -0,0 +1,113 @@
...
+		Level:  hclog.Debug,
+	})
+	l := logging.Default()
+	hcl := NewHClogger(hl, l)
+	clientConfig.Logger = hcl
+	return plugin.NewClient(&clientConfig)
+}
+
+// LoadPluginClient loads a Client of type T.
+func (m *Manager[T]) LoadPluginClient(name string) (*T, error) {
```
arielshaqed, line 91: Consider spec'cing the manager on `*T` when you create it, and returning `T` here.  For instance, if `T` is an interface (it often will be!) then this is a pointer-to-interface, which is really weird.  Even if the _current_ implementation in the library returns a struct, there is no requirement for it to stay that way.

pkg/plugins/manager.go
```diff
@@ -0,0 +1,113 @@
...
+		return nil, ErrPluginNameNotFound
+	}
+	grpcPluginClient, err := c.Client()
+	if err != nil {
+		return nil, err
+	}
+	rawGrpcClientStub, err := grpcPluginClient.Dispense(name) // Returns an implementation of the stub service.
+	if err != nil {
+		return nil, err
+	}
```
arielshaqed, line 106: This code needs a mutex to be thread-safe.

pkg/plugins/manager_test.go
```diff
@@ -0,0 +1,145 @@
...
+	cmd := exec.Command(os.Args[0], cs...)
+
+	env := []string{
+		"RUN_PLUGIN_SERVER=1",
+	}
+	cmd.Env = append(env, os.Environ()...)
+	return cmd
+}
+
+// This is not a real test. This is the plugin server that will be initialized by the tests
```
arielshaqed, line 97: Then it shouldn't be a test.  Perhaps this needs to be [`TestMain`](https://pkg.go.dev/testing#hdr-Main)?

arielshaqed (changes requested):
Thanks!  Very nice & useful.

A bunch of minor-but-important things that I think we should improve.

pkg/plugins/manager.go
Jonathan-Rosenberg: the previous argument was that a handshake is more of an auth entity rather than an identifier.

pkg/plugins/manager.go
Jonathan-Rosenberg: it kinda is- the plugin client and server agree on a certain key and value that identify one another

pkg/plugins/manager.go
Jonathan-Rosenberg: True, it's implemented on the next PR 🤫

pkg/plugins/manager_test.go
Jonathan-Rosenberg: It's triggered by the tests themselves, not initialized at the beginning of the tests

pkg/plugins/manager.go
Jonathan-Rosenberg: What do you mean by `Consider **spec'cing** the manager on *T `

talSofer (approved):
Thanks! added a few minor comments

pkg/plugins/manager.go
```diff
@@ -0,0 +1,113 @@
...
+	Key   string
+	Value string
+}
+
+// Manager maps the available plugin names to the for different kinds of plugin.Client plugin controllers.
+// T is the custom interface type that the returned GRPCClient implementation implements, e.g. "Differ" for `plugin.Client`s that
+// include a GRPCClient that implements the "Differ" interface:
+// grpcPluginClient, err := c.Client() // Returns a plugin.GRPCClient
+// rawGrpcClientStub, err := grpcPluginClient.Dispense(name) // Calls grpcPluginClient's GRPCClient method and returns the gRPC stub.
+// grpcClient, ok := rawGrpcClient.(Differ) // Asserts the expected type of stub client.
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: yas

pkg/plugins/manager.go
arielshaqed: The type `Manager[T]` returns clients of type `*T`.  This is technically correct if the client is a struct, and we instantiate it on the struct type:

```go
type MyClient struct { /* stuff */ } // But really the client is *MyClient

func a(m Manager[MyClient], name string) {
	var client *MyClient = m.LoadPluginClient(name)
	// client stuff
```

This works mostly because `*MyClient` satisfies the `MyClient` interface (the converse is not really true!).

BUT if `T` is an interface type, then `Manager[T].LoadPluginClient` returns a `*T`, which is a pretty useless type.

AFAIU the more Go-like interface would be to return a `T` from `Manager[T].LoadPluginClient`.  In the `*MyClient` case above, you would need to use `Manager[*MyClient]` as the manager type.  But that makes sense: the client type really is `*MyClient` there!

pkg/plugins/manager.go
```diff
@@ -0,0 +1,112 @@
...
+	return &Manager[T]{
+		pluginApplicationClients: make(map[string]*plugin.Client),
+	}
+}
+
+// RegisterPlugin is used to register a new plugin client with the corresponding plugin type.
+func (m *Manager[T]) RegisterPlugin(name string, id PluginIdentity, auth PluginHandshake, p plugin.Plugin) {
+	if m == nil || m.pluginApplicationClients == nil {
+		return
+	}
```
arielshaqed, line 58: Not sure when this is useful.

pkg/plugins/manager_test.go
```diff
@@ -0,0 +1,145 @@
...
+
+	env := []string{
+		"RUN_PLUGIN_SERVER=1",
+	}
+	cmd.Env = append(env, os.Environ()...)
+	return cmd
+}
+
+// This is not a real test. This is the plugin server that will be triggered by the tests
+func TestPluginServer(*testing.T) {
```
arielshaqed, line 98: If this is not a test, it should start with something other than `Test`.  Also not sure why it takes `*testing.T` parameter.

pkg/plugins/manager_test.go
```diff
@@ -0,0 +1,145 @@
...
+	if os.Getenv("RUN_PLUGIN_SERVER") != "1" {
+		return
+	}
+
+	defer os.Exit(0)
```
arielshaqed, line 103: Not sure why a test helper should ever exit.  This is even worse, it exits successfully when something strange happens.  But if that happens, the test succeeds _even though it should fail_.

arielshaqed (changes requested):
The TestPluginServer hook is a really strange way to run the test plugin server.  It should just be a separate program.  Failing that, if building 2 binaries for a test is too hard because Go, it might execute as a TestMain and then do one of 2 things.

But I want to question even that.  Why run the plugin server as a separate process?  It's just a web server, can't we run it in a separate goroutine?

pkg/plugins/manager.go
Jonathan-Rosenberg: got it

pkg/plugins/manager_test.go
Jonathan-Rosenberg: true this is useless

pkg/plugins/manager_test.go
Jonathan-Rosenberg: It's not a test, yet the command that runs it is `go test` which accepts the `-test.run` flag that specifies the tests it should run. Unfortunately, golang don't allow tests to be named other than `Test...` and must accept `*testing.T` as an argument:

> You write a test by creating a file with a name ending in `_test.go` that contains functions named `TestXXX` with signature `func (t *testing.T)`.

Jonathan-Rosenberg:
@arielshaqed, the plugin manager is dependent on the `go-plugin` (Hashicorp's) package which receives an `exec.Command` as the plugin server's binary it should run. It also kills it when the plugin's job is over. 
It means that a running go routine is not an option.
We can implement a separate process, but I think it's quite an overkill.

pkg/plugins/manager.go
```diff
@@ -0,0 +1,106 @@
...
+	pluginApplicationClients map[string]*plugin.Client
+}
+
+func NewManager[T any]() *Manager[T] {
+	return &Manager[T]{
+		pluginApplicationClients: make(map[string]*plugin.Client),
+	}
+}
+
+// RegisterPlugin is used to register a new plugin client with the corresponding plugin type.
```
arielshaqed, line 53: nit:
```suggestion
// RegisterPlugin registers a new plugin client with the corresponding plugin type.
```

pkg/plugins/manager.go
```diff
@@ -0,0 +1,106 @@
...
+	Key   string
+	Value string
+}
+
+// Manager maps the available plugin names to the different kinds of plugin.Client plugin controllers.
+// T is the custom interface type that the returned GRPCClient implementation implements, e.g. "Differ" for `plugin.Client`s that
+// include a GRPCClient that implements the "Differ" interface:
+// grpcPluginClient, err := c.Client() // Returns a plugin.GRPCClient
+// rawGrpcClientStub, err := grpcPluginClient.Dispense(name) // Calls grpcPluginClient's GRPCClient method and returns the gRPC stub.
+// grpcClient, ok := rawGrpcClientStub.(Differ) // Asserts the expected type of stub client.
```
arielshaqed, line 40: Consider:

```suggestion
//
// 	include a GRPCClient that implements the "Differ" interface:
// 	grpcPluginClient, err := c.Client() // Returns a plugin.GRPCClient
// 	rawGrpcClientStub, err := grpcPluginClient.Dispense(name) // Calls grpcPluginClient's GRPCClient method and returns the gRPC stub.
// 	grpcClient, ok := rawGrpcClientStub.(Differ) // Asserts the expected type of stub client.
```

which may produce a neater Godoc.

arielshaqed (approved):
Thanks for your patience explaining the go-plugin test horridness to me!

arielshaqed:
> Thanks for your patience explaining the go-plugin test horridness to me!

While we're there, can you use one of the go-plugin test plugin plugins, maybe [this one](https://github.com/hashicorp/go-plugin/blob/84014b780e9b5a53e3916195b32be8e183e67214/testing.go#L110)?

pkg/plugins/manager_test.go
johnnyaug: I like the ping-pong protobuf and additional classes, since it explains very well how plugins and grcp work. However, I'm wondering about this methodology, because this test doesn't (and shouldn't) test the logic of the plugins. I would expect this test to mock both `plugin.Plugin` and `plugin.Client`, so it can become a strict unit test.

One problem is the `plugin.Client` is a struct and not an interface, but I suggest creating an interface on our side in order to allow mocking it as well:

```golang
type Client interface {
    Client() (ClientProtocol, error)
}
```

pkg/plugins/manager.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// RegisterPlugin registers a new plugin client with the corresponding plugin type.
+func (m *Manager[T]) RegisterPlugin(name string, id PluginIdentity, auth PluginHandshake, p plugin.Plugin) {
+	hc := plugin.HandshakeConfig{
+		ProtocolVersion:  id.ProtocolVersion,
+		MagicCookieKey:   auth.Key,
+		MagicCookieValue: auth.Value,
+	}
+	cmd := id.Cmd
```
johnnyaug, line 61: It's confusing to me that an "id" contains a command.

pkg/plugins/manager.go
```diff
@@ -0,0 +1,107 @@
...
+		Level:  hclog.Debug,
+	})
+	l := logging.Default()
+	hcl := NewHClogger(hl, l)
+	clientConfig.Logger = hcl
+	return plugin.NewClient(&clientConfig)
+}
+
+// LoadPluginClient loads a Client of type T.
+func (m *Manager[T]) LoadPluginClient(name string) (T, error) {
```
johnnyaug, line 87:
```suggestion
func (m *Manager[T]) DispensePluginClient(name string) (T, error) {
```
I think is more accurate?

pkg/plugins/manager.go
```diff
@@ -0,0 +1,107 @@
...
+	Value string
+}
+
+// Manager maps the available plugin names to the different kinds of plugin.Client plugin controllers.
+// T is the custom interface type that the returned GRPCClient implementation implements, e.g. "Differ" for `plugin.Client`s that
+//
+//	include a GRPCClient that implements the "Differ" interface:
+//	grpcPluginClient, err := c.Client() // Returns a plugin.GRPCClient
+//	rawGrpcClientStub, err := grpcPluginClient.Dispense(name) // Calls grpcPluginClient's GRPCClient method and returns the gRPC stub.
+//	grpcClient, ok := rawGrpcClientStub.(Differ) // Asserts the expected type of stub client.
```
johnnyaug, line 41: This is an explanation of an implementation detail.

johnnyaug (changes requested):
Mainly cleanups and naming. Also, I really enjoyed reading the tests, they helped me understand the world of go-plugins. However, not sure the methodology is optimal for these specific tests.

PR #5154, opened 2023-01-31
Fix configure aws credentials in system tests secrets check

nopcoder:
Check secrets required for login to ecr on system tests


rmoff (approved):
LGTM!

PR #5161, opened 2023-02-01
Fix lakeFSFS stage<->copy fallback mechanism

### LINKED ISSUE #5160: Bug: Fallback mechanism from copyObject to stageObject is broken
In lakeFSFS, during renameObject, there's a fallback mechanism to support older lakeFS server versions.
That mechanism is broken due to a bug in error comparison.

itaiad200:
closes #5160

### DIFF
--- a/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
+++ b/clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
@@ -456,6 +456,7 @@ public class LakeFSFileSystem extends FileSystem {
                         creationReq);
             } catch (ApiException e) {
                 if (e.getCode() != HttpStatus.SC_INTERNAL_SERVER_ERROR ||
+                        e.getResponseBody() == null ||
                         !e.getResponseBody().contains("invalid API endpoint")) {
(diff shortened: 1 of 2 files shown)


clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
```diff
@@ -456,7 +456,7 @@ private boolean renameObject(LakeFSFileStatus srcStatus, Path dst) throws IOExce
                         creationReq);
             } catch (ApiException e) {
                 if (e.getCode() != HttpStatus.SC_INTERNAL_SERVER_ERROR ||
-                        !e.getMessage().contains("invalid API endpoint")) {
+                        !e.getResponseBody().contains("invalid API endpoint")) {
```
nopcoder, line 459: The `getReponseBody()` method can return nil can you verify before calling `contains`.

nopcoder (changes requested):

clients/hadoopfs/src/main/java/io/lakefs/LakeFSFileSystem.java
itaiad200: Done.

nopcoder (approved):

PR #5165, opened 2023-02-01
Remove kv's StoreMessage

### LINKED ISSUE #5192: Refactor out StoreMessage from kv package
Using set of functions for set/get that support marshal proto message can replace the use of a wrapper for kv.Store.

nopcoder:
Using the set/get msg funcs instead of embedding the store into a message store.

Close #5192


pkg/actions/store.go
```diff
@@ -25,17 +25,17 @@ type Store interface {
 }
 
 type KVStore struct {
-	store kv.StoreMessage
+	store kv.Store
 }
 
-func NewActionsKVStore(store kv.StoreMessage) Store {
+func NewActionsKVStore(store kv.Store) Store {
```
N-o-Z, line 31:
```suggestion
func NewActionsStore(store kv.Store) Store {
```

pkg/actions/store.go
```diff
@@ -25,17 +25,17 @@ type Store interface {
 }
 
 type KVStore struct {
-	store kv.StoreMessage
+	store kv.Store
```
N-o-Z, line 28: Maybe we want to change the type from `KVStore` to `ActionStore`? This might be a bit confusing

pkg/auth/service.go
```diff
@@ -955,7 +955,7 @@ func (s *AuthService) GetCredentialsForUser(ctx context.Context, username, acces
 func (s *AuthService) GetCredentials(ctx context.Context, accessKeyID string) (*model.Credential, error) {
 	return s.cache.GetCredential(accessKeyID, func() (*model.Credential, error) {
 		m := &model.UserData{}
-		itr, err := s.store.Scan(ctx, m.ProtoReflect().Type(), model.PartitionKey, model.UserPath(""), []byte(""))
+		itr, err := kv.NewPrimaryIterator(ctx, s.store, m.ProtoReflect().Type(), model.PartitionKey, model.UserPath(""), kv.IteratorOptionsAfter([]byte("")))
```
N-o-Z, line 958: Why not use Scan? Do we not need Scan in the interface anymore?

pkg/kv/store_message_test.go
```diff
@@ -25,27 +25,24 @@ func TestStoreMessage(t *testing.T) {
...
-		testStoreMessageSetIf(t, ctx, sm)
+		testStoreMessageSetIf(t, ctx, store)
 	})
 	t.Run("delete test", func(t *testing.T) {
-		testStoreMessageDelete(t, ctx, sm)
+		testStoreMessageDelete(t, ctx, store)
 	})
 	t.Run("scan test", func(t *testing.T) {
-		testStoreMessageScan(t, ctx, sm)
+		testStoreMessageScan(t, ctx, store)
```
N-o-Z, line 38: Redundant - we have per driver tests for this operation

pkg/kv/store_message_test.go
```diff
@@ -25,27 +25,24 @@ func TestStoreMessage(t *testing.T) {
...
 	})
 	t.Run("scan wrong format test", func(t *testing.T) {
-		testStoreMessageScanWrongFormat(t, ctx, sm)
+		testStoreMessageScanWrongFormat(t, ctx, store)
```
N-o-Z, line 41: Redundant - we have per driver tests for this operation

pkg/kv/store_message_test.go
```diff
@@ -25,27 +25,24 @@ func TestStoreMessage(t *testing.T) {
...
-		testStoreMessageSetGet(t, ctx, sm)
+		testStoreMessageSetGet(t, ctx, store)
 	})
 	t.Run("set-if test", func(t *testing.T) {
-		testStoreMessageSetIf(t, ctx, sm)
+		testStoreMessageSetIf(t, ctx, store)
 	})
 	t.Run("delete test", func(t *testing.T) {
-		testStoreMessageDelete(t, ctx, sm)
+		testStoreMessageDelete(t, ctx, store)
```
N-o-Z, line 35: Redundant - we have per driver tests for this operation

pkg/actions/store.go
nopcoder: The package is 'actions' so ActionStore will just repeat the name. The 'KVStore' is the implementation of the 'Store' interface used by the actions package.
Will change it to `kvStore` as it is an implementation of `Store` interface using the kv.

pkg/auth/service.go
nopcoder: We have 'Scan' on the kv level which doesn't support messages.
The 'Scan' we used to call on the message store was just returning a new iterator as we do here.

pkg/kv/store_message_test.go
nopcoder: Removed - thanks!

N-o-Z:
@nopcoder not sure `minor-change` applies here - can you please link an issue

N-o-Z (approved):
LGTM!

PR #5169, opened 2023-02-01
Allow configuration of GUI warning message for deprecating RBAC

arielshaqed:
You can switch it off by setting auth.ui_config.RBAC to "external" (default
"simplified"), which will be required to use an external RBAC server later.


arielshaqed:
Tested by running lakeFS locally with and without this config option.

docs/reference/configuration.md
```diff
@@ -100,6 +100,10 @@ This reference uses `.` to denote the nesting of values.
 * `auth.oidc.authorize_endpoint_query_parameters` `(map[string]string : )` - Add these parameters when calling the provider's `/authorize` endpoint
 * `auth.oidc.validate_id_token_claims` `(map[string]string : )` - When a user tries to access lakeFS, validate that the ID token contains these claims with the corresponding values.
 * `auth.oidc.additional_scope_claims` `(string[]: [])` - Add these additional scopes/claims to the OIDC user token. `openid` and `profile` are provided already. 
+* `auth.ui_config.RBAC` `(string: "simplified")` - "simplified" or
```
ortz, line 103: I wonder if this can be associated with the external auth configuration instead of maintaining a couple of config params

ortz (approved):
One question, not a blocker :)

webui/src/lib/hooks/conf.jsx
```diff
@@ -0,0 +1,17 @@
...
+
+export const WithLoginContext = ({children}) => {
+    const { response, error, loading } = useAPI(() => setup.getState());
+    const value = (error || loading) ? null : response?.login_config;
+    return <LoginConfigContext.Provider value={value}>
+             {children}
+           </LoginConfigContext.Provider>;
+};
+
+export const useLoginContext = () => useContext(LoginConfigContext);
```
eladlachmi: I'd consider renaming to `useLoginConfigContext`

webui/src/lib/hooks/conf.jsx
```diff
@@ -0,0 +1,17 @@
...
+
+import {useAPI} from "./api";
+import {setup} from "../api";
+
+
+export const LoginConfigContext = createContext(null);
+
+export const WithLoginContext = ({children}) => {
+    const { response, error, loading } = useAPI(() => setup.getState());
+    const value = (error || loading) ? null : response?.login_config;
```
eladlachmi: `useAPI` promises a stable response. However, `value` isn't a stable reference.  
If you assume that the provider is at the top of the tree, it might not be a huge deal, but I'd still recommend `useMemo` or `useRef` for `value` to keep referential stability as it's a prop of the provider.

docs/reference/configuration.md
arielshaqed: It can, but I would rather not.  The upcoming RBAC removal also makes this distinction.  It can make sense to provide an external auth server and still not provide RBAC in the GUI.  For instance when the external auth server might be shared between multiple lakeFS installations.

webui/src/pages/auth/policies/index.jsx
```diff
@@ -43,6 +44,8 @@ const PoliciesContainer = () => {
     if (error) return <Error error={error}/>;
     if (loading) return <Loading/>;
 
+    const showRBACWarning = useLoginContext()?.RBAC === 'simplified';
```
eladlachmi, line 47: It's subtle, but this is still a conditional use of a hook. This needs to move above line 44.

webui/src/lib/hooks/conf.jsx
```diff
@@ -0,0 +1,17 @@
+import React, {createContext, useContext} from "react";
+
+import {useAPI} from "./api";
+import {setup} from "../api";
+
+
+export const LoginConfigContext = createContext(null);
+
+export const WithLoginContext = ({children}) => {
```
eladlachmi: I'd consider renaming to `WithLoginConfigContext`

eladlachmi (changes requested):
Looks good! Got a few comments and suggestions.

webui/src/lib/hooks/conf.jsx
arielshaqed: Ouch, great catch --  thanks!

webui/src/lib/hooks/conf.jsx
arielshaqed: Done.

webui/src/pages/auth/policies/index.jsx
arielshaqed: Of course.  Done.

arielshaqed (commented):
Thanks!

PTAL...

eladlachmi (approved):
Thanks for taking the time to implement the changes.
LGTM ✌🏻

arielshaqed:
Thanks!

This time pulling for sure :-)

PR #5176, opened 2023-02-02
UGC custom jar for cloud testing

eden-ohana:


nopcoder:
Hope we can trust our `if` statements.
If we like to be sure we have two separate jobs that one can only mark - we can add another main entry point and refactor the code.

nopcoder (changes requested):
would refactor the code to enable different entry point and reuse the code for this entry point.

eden-ohana:
@nopcoder It's not only the if statement, I removed the code that can delete objects.
I'm not sure I understand what you are suggesting.

nopcoder:
@eden-ohana all the deleted code is inside a scope that runs when sweep is true. The value of sweep is controlled by a configuration. 
So why do we need this change?

eden-ohana:
> @eden-ohana all the deleted code is inside a scope that runs when sweep is true. The value of sweep is controlled by a configuration. So why do we need this change?

for safety, even if somehow wrong configuration is passed, we still won't delete anything.

PR #5184, opened 2023-02-05
Export gob registration to let external services to use auth.Claims

arielshaqed:
Add a new package github.com/treeverse/lakefs/pkg/auth/oidc/encoding that
holds Claims and registers it for use in JWTs.  External auth services may
use these Claims to be interoperable with lakeFS.

This separate package allows a small minimal import.

### DIFF
--- a/pkg/auth/oidc/authenticator.go
+++ b/pkg/auth/oidc/authenticator.go
@@ -4,11 +4,18 @@ import (
 	"context"
 	"errors"
 
+	"github.com/treeverse/lakefs/pkg/auth/oidc/encoding"
+
 	"github.com/coreos/go-oidc/v3/oidc"
 	"golang.org/x/oauth2"
 )
 
 var ErrTokenExtract = errors.New("failed to extract id token")
+
+// Claims is encoding.Claims.  To preserve serialization, it must be an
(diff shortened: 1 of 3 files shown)


pkg/auth/oidc/authenticator.go
```diff
@@ -4,12 +4,19 @@ import (
 	"context"
 	"errors"
 
+	"github.com/treeverse/lakefs/pkg/auth/oidc/encoding"
```
nopcoder, line 7: delete was after this one

nopcoder (approved):
LGTM. probably related to previous pr - why we went with gob and not JSON which will enable better interoperability in parse the claims?

arielshaqed:
> LGTM. probably related to previous pr - why we went with gob and not JSON which will enable better interoperability in parse the claims?

Not sure of history for these Claims.  Just the minimal change to make it work as part of moving OIDC support from SSO to external auth servers.  Only interested in producing a JWT the same as before, so no reason to change its encoding here.

THANKS!  Pulling once @Isan-Rivkin (the upcoming user) can also approve.

Isan-Rivkin (approved):
LGTM TY!

PR #5189, opened 2023-02-06
fix/ UI import compare link

### LINKED ISSUE #5190: UI import done page broken compare link

eden-ohana:
closes #5190


eladlachmi (approved):
It'll get the job done. As long as you're aware of the risk in this approach to building the URL, it's okay.

PR #5193, opened 2023-02-06
KV postgres use advisory lock while performing table creation

### LINKED ISSUE #5196: Docker compose using postgres and number of replicas fail with error on relation creation
While using docker-compose with lakefs configured to use 3 replicas and postgres database, startup fail with the following error:

```
setup failed: ERROR: relation \"kv_49\" already exists (SQLSTATE 42P07)
```

nopcoder:
Fix #5196

### DIFF
--- a/pkg/kv/postgres/store.go
+++ b/pkg/kv/postgres/store.go
@@ -5,7 +5,6 @@ import (
 	"errors"
 	"fmt"
 	"hash/fnv"
-	"os"
 	"strconv"
 
 	"github.com/IBM/pgxpoolprometheus"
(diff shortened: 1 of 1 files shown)


johnnyaug (approved):
LGTM, thanks for fixing quickly!

pkg/kv/postgres/store.go
```diff
@@ -152,26 +153,45 @@ func parseStoreConfig(runtimeParams map[string]string, pgParams *kvparams.Postgr
...
-func setupKeyValueDatabase(ctx context.Context, conn *pgxpool.Conn, table string, partitionsAmount int) error {
+func setupKeyValueDatabase(ctx context.Context, conn *pgxpool.Conn, table string, partitionsAmount int) (err error) {
+	var aid string
+	aid, err = generateAdvisoryLockID("lakefs:" + table)
+	if err != nil {
+		return err
+	}
+
+	// This will wait indefinitely until the lock can be acquired.
+	_, err = conn.Exec(ctx, `SELECT pg_advisory_lock($1)`, aid)
```
N-o-Z, line 164: @nopcoder can we add a lock_timeout and provide a detailed error message.
We should want to know when a lock was acquired and not released - and add details of required action on the event of.

N-o-Z (changes requested):
Thanks! One comment

pkg/kv/postgres/store.go
nopcoder: @N-o-Z This is the same technique we used when `go-migrate` was used. (https://github.com/golang-migrate/migrate/blob/dc26c41ac9dac4a1023c87723ffda644e41af1fe/database/pgx/pgx.go#L226)
The lock timeout is specified in the the database configuration (lock_timeout) the default is unlimited.

N-o-Z (approved):
Thanks for addressing the comment!

PR #5198, opened 2023-02-06
Delta Diff: Plugin manager concurrency

Jonathan-Rosenberg:
Add concurrency handling to the plugin manager.
Notice that the re-registration of the client is necessary because of the `exec.Cmd` struct that cannot be restarted (not reused) after it was used.


pkg/plugins/client_store.go
```diff
@@ -0,0 +1,59 @@
...
+type Client struct {
+	PluginClient *plugin.Client
+	ClientProps  *clientProps
+}
+
+// clientStore maps the available plugin names to the different kinds of plugin.Client plugin controllers and their properties.
+//
+// The map might include a mapping of "delta" -> `{ plugin.Client, "deltaPluginLocation", ["arg1"], ["env1"] }` to
+// communicate with the Delta plugin.
+type clientStore struct {
```
arielshaqed, line 24: Can we use [sync.Map](https://pkg.go.dev/sync#Map) instead?

pkg/plugins/manager.go
```diff
@@ -84,24 +93,36 @@ func newPluginClient(name string, p plugin.Plugin, hc plugin.HandshakeConfig, cm
 }
 
 // LoadPluginClient loads a Client of type T.
-func (m *Manager[T]) LoadPluginClient(name string) (T, error) {
-	var zero T
-	c, ok := m.pluginApplicationClients[name]
-	if !ok {
-		return zero, ErrPluginNameNotFound
+// Also returns a function used to close the Client and reset it. The reset is needed because after
```
arielshaqed, line 96: If the close func must be called, please document it here.

arielshaqed (approved):
Thanks!

I think this does more than add concurrency; consider retitling.

Not a fan of reimplementing concurrent maps.  But if we _do_, please consider using a plain Mutex:
* Actual performance is usually higher
* Performance doesn't really matter here
* Using RWMutex requires a bit more care than Mutex.

THANKS!

pkg/plugins/client_store.go
Jonathan-Rosenberg: Thought about doing that, but then read their documentation:

> The Map type is specialized. Most code should use a plain Go map instead, with separate locking or coordination, for better type safety and to make it easier to maintain other invariants along with the map content.

WDYT?

PR #5204, opened 2023-02-07
Lua hook encode URL parameters for internal server requests

### LINKED ISSUE #5207: Lua hook request to lakefs does not escape URL arguments
The following Lua APIs concatenate the some of the call arguments as part of internal request URL.
Each argument should be escaped in order to keep a valid URL and risk free processing.

- create_tag
- diff_refs
- list_objects  
- get_object

nopcoder:
Fix #5207


itaiad200 (approved):
Thanks for the important fix!

PR #5206, opened 2023-02-08
Reuse changes tree container in compare, commit, and uncommitted changes pages

### LINKED ISSUE #5205: Refactor Compare, commit changes and uncommitted changes page to share changes tree container
Currently, the compare, changes done in a commit, and uncommitted changes pages show a lost of changes which is essentially the diff between two refs. the three pages use the same tree-container structure to show the diff but are not sharing code for that. This task is to refactor the code to reuse the shared parts.

talSofer:
Closes #5205

### How was this tested 
* manually tested that compare, commit changes and uncommitted changes view work as before 
* Validated that pagination works


webui/src/lib/components/repository/changes.jsx
```diff
@@ -145,3 +151,103 @@ function useTreeItemType(entry, repo, leftDiffRefID, rightDiffRefID) {
     }, [leftResult, rightResult]);
     return treeItemType;
 }
+
+/**
+ * A container component for entries that represent diff between refs. this container is use by compare, commit changes,
```
Jonathan-Rosenberg, line 156:
```suggestion
 * A container component for entries that represent a diff between refs. this container is used by the compare, commit changes,
```

webui/src/lib/components/repository/changes.jsx
```diff
@@ -145,3 +151,103 @@ function useTreeItemType(entry, repo, leftDiffRefID, rightDiffRefID) {
...
+ * @param repo Repository
+ * @param reference commitID / branch
+ * @param internalRefresh to be called when the page refreshes manually
+ * @param prefix for which changes are displayed
+ * @param getMore to be called when requesting more diff results for a prefix
+ * @param loading of API response state to get changes
+ * @param nextPage of API response state to get changes
+ * @param setAfterUpdated state of pagination of the item's children
+ * @param onNavigate to be called when navigating to a prefix
+ * @param onRevert to be called when an object/prefix is requested to be reverted
```
Jonathan-Rosenberg, line 174: oh my

webui/src/lib/components/repository/changes.jsx
```diff
@@ -145,3 +151,103 @@ function useTreeItemType(entry, repo, leftDiffRefID, rightDiffRefID) {
...
+                                        disabled={false}
+                                        onClick={async () => {
+                                            setShowComingSoonModal(true);
+                                            await sendDeltaDiffStats();
+                                        }}>
+                                    <DiffIcon/> Compare Delta Lake tables
+                                </Button>
+                            </ExperimentalOverlayTooltip>
+                            </>
+                            : ""}</>
```
Jonathan-Rosenberg, line 216: nit:
```suggestion
                            : ""}
                        </>
```

webui/src/lib/components/repository/changes.jsx
```diff
@@ -145,3 +151,103 @@ function useTreeItemType(entry, repo, leftDiffRefID, rightDiffRefID) {
...
+        ];
+        await statistics.postStatsEvents(deltaDiffStatEvents);
+    }
+
+    return <>
+        <div className="tree-container">
+            {(results.length === 0) ? <Alert variant="info">No changes</Alert> : (
+                <>
+                    {!enableDeltaDiff ?
+                        <>
```
Jonathan-Rosenberg, line 197:
```suggestion
                    { !enableDeltaDiff && <>
```

webui/src/lib/components/repository/changes.jsx
```diff
@@ -145,3 +151,103 @@ function useTreeItemType(entry, repo, leftDiffRefID, rightDiffRefID) {
...
+                        {showExperimentalDeltaDiffButton
+                            ? <>
```
Jonathan-Rosenberg, line 199: This should work, right?
```suggestion
                        { showExperimentalDeltaDiffButton && <>
```

webui/src/lib/components/repository/changes.jsx
```diff
@@ -145,3 +151,103 @@ function useTreeItemType(entry, repo, leftDiffRefID, rightDiffRefID) {
...
+                "count": 1,
+            }
+        ];
+        await statistics.postStatsEvents(deltaDiffStatEvents);
+    }
+
+    return <>
+        <div className="tree-container">
+            {(results.length === 0) ? <Alert variant="info">No changes</Alert> : (
+                <>
```
Jonathan-Rosenberg, line 195: I think it's redundant

webui/src/lib/components/repository/changes.jsx
```diff
@@ -145,3 +151,103 @@ function useTreeItemType(entry, repo, leftDiffRefID, rightDiffRefID) {
...
+ * @param internalRefresh to be called when the page refreshes manually
+ * @param prefix for which changes are displayed
+ * @param getMore to be called when requesting more diff results for a prefix
+ * @param loading of API response state to get changes
+ * @param nextPage of API response state to get changes
+ * @param setAfterUpdated state of pagination of the item's children
+ * @param onNavigate to be called when navigating to a prefix
+ * @param onRevert to be called when an object/prefix is requested to be reverted
+ */
+export const ChangesTreeContainer = ({results, showExperimentalDeltaDiffButton = false, delimiter, uriNavigator,
```
Jonathan-Rosenberg, line 176: It seems like this container is only for OTF diffs. Consider naming it accordingly.

webui/src/lib/components/repository/changes.jsx
```diff
@@ -145,3 +151,103 @@ function useTreeItemType(entry, repo, leftDiffRefID, rightDiffRefID) {
...
+                    }
+                    <Card>
+                        <Card.Header>
+                                <span className="float-start">
+                                    {(delimiter !== "") && uriNavigator}
+                                </span>
+                        </Card.Header>
+                        <Card.Body>
+                            <Table borderless size="sm">
+                                <tbody>
```
Jonathan-Rosenberg, line 230: Is this supposed to be a table? It looks like a table with a body containing a list of `TreeItem`s

Jonathan-Rosenberg (approved):
Very cool!
Had a few question.

webui/src/lib/components/repository/changes.jsx
eladlachmi: I second this 👆🏻
There's at least one case where `TreeItem` doesn't return a `tr`
If you want to make it more apparent, you could name it `TreeTableRow` or something like that. You can also change the `.jsx` to `.tsx` and TS will make it clear that whatever is returned from `TreeItem` has to be a `tr` or `tr`-adjacent.

webui/src/pages/repositories/repository/changes.jsx
```diff
@@ -251,50 +259,13 @@ const ChangesBrowser = ({repo, reference, prefix, onSelectRef, }) => {
...
-                                </tbody>
-                            </Table>
-                        </Card.Body>
-                    </Card>
-                )}
-            </div>
+            <ChangesTreeContainer results={results} delimiter={delimiter}
+                                  uriNavigator={uriNavigator} leftDiffRefID={committedRef} rightDiffRefID={uncommittedRef}
+                                  repo={repo} reference={reference} internalReferesh={internalRefresh} prefix={prefix}
+                                  getMore={(afterUpdated, path, useDelimiter= true, amount = -1) => {
```
eladlachmi, line 265: Make the anonymous function passed to the `getMore` prop a named function, stop with the ternary ifs everywhere (!!), and `useCallback` as needed.

eladlachmi: Since you're re-using the function, move it to the same place where `ChangesTreeContainer` is and provide that default implementation as a default prop value. If another implementation is needed, it can always be overwritten.
If you want the consumer to provide `refs`, you can do something like this
```JavaScript
export const defaultGetMore = (refs) => 
  (afterUpdated, path, useDelimiter= true, amount = -1) => { ... }; 
```
and you can use `refs` from the closure <-- functional programming pro tip 😁

eladlachmi (changes requested):

webui/src/pages/repositories/repository/changes.jsx
talSofer: done, but - 
- I think that ternary ifs are fine here as they make things short
- I don't see how I can use useCallback here because it will make me violate hooks rules. 

let me know if the current implementation looks ok or still create issues

webui/src/lib/components/repository/changes.jsx
talSofer: I agree this is not a real table, if I had to guess why this was implemented as a table I'd say it is here for creating an easy-to-organize grid. any other suggestions for react component I can use?

eladlachmi: @talSofer I don't have a problem with the table itself. There are other ways of building something similar, but you could make a case for using a table.  
My main issue is that `tbody` should include `tr` tags only at the top level, and in this case, `TreeItem` includes at least one path where it doesn't return a `tr`.  
As a side note, I'd suggest renaming `TreeItem` to something like `TreeItemRow` or `TreeTableRow`, so it's more obvious that it's intended to return something with a `tr` at the top level. Hope that makes more sense.

eladlachmi (approved):
Much cleaner and easier to read IMHO ✌🏻

webui/src/lib/components/repository/changes.jsx
talSofer: thanks @eladlachmi!

> My main issue is that tbody should include tr tags only at the top level, and in this case, TreeItem includes at least one path where it doesn't return a tr.

So in that case, would you just remove the table body? in addition to renaming TreeItem

eladlachmi: @talSofer Probably not. I suggest making sure that `TreeItem` (or whatever its new name will be) always returns a component with a single (or maybe even multiple, if there's a reason to) `<tr>` at the top level.

You can't just remove `tbody`, because the issue cascades. `<table>` also can't have just any random child tags. If you want to get rid of the table completely, that's a bigger refactor, and I wouldn't go there right now if you ask me.

talSofer: got it, fixed

PR #5208, opened 2023-02-08
Delta Lake Diff MVP: Diff Service

### LINKED ISSUE #4964: Delta Diff MVP - Task: Create a Diff Service for OTFs

Jonathan-Rosenberg:
First of all, don't be alarmed by the size of the PR. 8/13 files are a result of moving the plugin manager to an internal directory, as it should only be in the context of the `plugins` package.

---

This PR includes the Diff Service implementation.
As part of it, I introduced a new interface called `Handler`, and this interface is implemented by the plugin `Manager` struct.
This is done so that services could register different handlers for different plugin types (i.e. not only Hashicorp plugins).

Closes #4964


pkg/plugins/diff/service_test.go
```diff
@@ -0,0 +1,103 @@
...
+		return nil, nil, errNotFound
+	}
+	return TestDiffer{}, nil, nil
+}
+
+func NewMockHandler[T Differ]() *MockHandler[T] {
+	return &MockHandler[T]{PropertyMap: make(map[string]internal.HCPluginProperties)}
+}
+
+type TestDiffer Response
```
talSofer, line 79: Why is TestDiffer or type Response?

pkg/plugins/diff/service.go
```diff
@@ -0,0 +1,98 @@
...
+
+type Differ interface {
+	Diff(context.Context, Params) (Response, error)
+}
+
+// Handler is an interface used to handle the different plugin implementation.
+// T is the custom interface that the plugins implement.
+// I is the type of input properties needed to support the plugin.
+// Look at internal.Manager to get an example of an implementation for this interface.
+type Handler[T, I any] interface {
```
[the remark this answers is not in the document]
Jonathan-Rosenberg: You're right. I thought about "defining interfaces where you consume them" but this is just not right here...

pkg/plugins/diff/service_test.go
Jonathan-Rosenberg: just so it could return itself and also be the differ.
It can basically be anything as long as it implements the interface.

Jonathan-Rosenberg: Done

pkg/plugins/diff/service.go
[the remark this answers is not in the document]
Jonathan-Rosenberg: Done

talSofer (approved):
LGTM!

PR #5216, opened 2023-02-11
Log source IP in operations logs

johnmantios:
Not sure how to test this new function, please advise. Also, I am not sure how to handle a possible error in the function, just passed a generic error string for the time being

### DIFF
--- a/pkg/httputil/logging.go
+++ b/pkg/httputil/logging.go
@@ -58,9 +58,9 @@ func SourceIP(r *http.Request) string {
 	sourceIP, sourcePort, err := net.SplitHostPort(r.RemoteAddr)
 
 	if err != nil {
-		return "Could not get IP address" //not sure of the error handling here
+		return err.Error()
 	}
-	return sourceIP + sourcePort
+	return sourceIP + ":" + sourcePort
 }
 
 func DefaultLoggingMiddleware(requestIDHeaderName string, fields logging.Fields, middlewareLogLevel string) func(next http.Handler) http.Handler {
--- a/pkg/httputil/logging.go
+++ b/pkg/httputil/logging.go
@@ -54,17 +54,13 @@ func RequestID(r *http.Request) (*http.Request, string) {
 	return r, reqID
 }
 
-func SourceIP() string {
-	conn, err := net.Dial("udp", "8.8.8.8:80")
+func SourceIP(r *http.Request) string {
+	sourceIP, sourcePort, err := net.SplitHostPort(r.RemoteAddr)
+
 	if err != nil {
-		return "Could not get IP address" //TODO: do we need better handling for this one?
+		return "Could not get IP address" //not sure of the error handling here
 	}
-	defer conn.Close()
-
-	localAddress := conn.LocalAddr().(*net.UDPAddr)
-
-	return localAddress.IP.String()
-
+	return sourceIP + sourcePort
 }
 
 func DefaultLoggingMiddleware(requestIDHeaderName string, fields logging.Fields, middlewareLogLevel string) func(next http.Handler) http.Handler {
@@ -74,7 +70,7 @@ func DefaultLoggingMiddleware(requestIDHeaderName string, fields logging.Fields,
 			writer := &ResponseRecordingWriter{Writer: w, StatusCode: http.StatusOK}
 			r, reqID := RequestID(r)
 			client := GetRequestLakeFSClient(r)
-			sourceIP := SourceIP()
+			sourceIP := SourceIP(r)
 
 			// add default fields to context
 			requestFields := logging.Fields{
--- a/pkg/httputil/logging.go
+++ b/pkg/httputil/logging.go
@@ -2,6 +2,7 @@ package httputil
 
 import (
 	"context"
+	"net"
 	"net/http"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


pkg/httputil/logging.go
```diff
@@ -53,13 +54,27 @@ func RequestID(r *http.Request) (*http.Request, string) {
 	return r, reqID
 }
 
+func SourceIP() string {
```
arielshaqed, line 57: This function will perform a DNS lookup per log line. There are multiple reasons not to do this: it's slow, leaks information, can fail, won't yield information for internal IPs, loss information for external IPs, needs to be configured to use a different server from some networks, etc.

Could we just not do this, and instead use the text format of the IP address? It will look like "10.20.30.40", which is now than adequate.

arielshaqed (changes requested):
Thanks for taking this issue!

I fear there may be a misunderstanding here about what to log. We would like to log an IP of the remote side the http connection. This code seems to try to log the local address of some interface on the machine that is used on another connection.

That said, the IP can be logged in numerical form.

pkg/httputil/logging.go
```diff
@@ -53,13 +54,23 @@ func RequestID(r *http.Request) (*http.Request, string) {
 	return r, reqID
 }
 
+func SourceIP(r *http.Request) string {
+	sourceIP, sourcePort, err := net.SplitHostPort(r.RemoteAddr)
+
+	if err != nil {
+		return "Could not get IP address" //not sure of the error handling here
+	}
+	return sourceIP + sourcePort
```
arielshaqed, line 63: Not sure this will be properly formatted anyway, I'd expect needing to add a colon char as a separator. Doesn't matter if we do the above suggestion, wdyt?

pkg/httputil/logging.go
```diff
@@ -53,13 +54,23 @@ func RequestID(r *http.Request) (*http.Request, string) {
 	return r, reqID
 }
 
+func SourceIP(r *http.Request) string {
+	sourceIP, sourcePort, err := net.SplitHostPort(r.RemoteAddr)
+
+	if err != nil {
+		return "Could not get IP address" //not sure of the error handling here
```
arielshaqed, line 61: Not sure why we need to parse the string here: we either reconstruct soon-ish the same string on success, or return an error during that will be less useful than the original string.

What would we lose if we dropped this function and just used r.RemoteAddr directly?

arielshaqed (commented):
Nearly there, thanks!

pkg/httputil/logging.go
johnmantios: Using just r.RemoteAddr will return the following (tested on localhost in my machine):
```
[::1]:60046
```
Notice the weird brackets? As a solution to that the go team proposes using the net.SplitHostPort function that returns the IP and the Port separately. We can just use the IP that is returned by net.SplitHostPort which in this case would be ``` ::1 ```.  But if you're ok with the brackets thing then we should get rid of the whole SourceIP function and just use r.RemoteAddr.

pkg/httputil/logging.go
johnmantios: You're absolutely right.

pkg/httputil/logging.go
arielshaqed: Yikes!  Looks like a classic Go move :-/

I'm fine with brackets.  As yet another alternative, net.GetHostPort seems to return [really neat error messages](https://go.dev/play/p/hqqRs8WlF2h) -- as that example shows, we could use _those_.

```suggestion
		return err.Error()
```

pkg/httputil/logging.go
arielshaqed:
```suggestion
	return sourceIP + ":" + sourcePort
```

arielshaqed (commented):
Great, thanks for explaining about Go IPv6 endpoints.

Let's just get the formatting done either which way and I will be very glad to pull.

arielshaqed:
Hi @johnmantios ,

I'd like to wrap this PR up and pull:

* Some minor code review comments remain.
* As regards testing, I would be happy to quote a single log line that this generates: just run the code with logging.level set to "debug" and access lakeFS, something should show up.

  For the record, I find unit tests for logs more trouble than they are worth: they are brittle and give little value.

If you like, I can do these for you.

THANKS!

arielshaqed (approved):
Thanks!

Here's what 2 logs look like.  Both are from loopback, one v6 and one v4.  But they do show what things will look like:
```
DEBUG  [2023-02-27T15:04:36+02:00]home/ariels/sdk/go1.19/src/net/http/server.go:2109 net/http.HandlerFunc.ServeHTTP HTTP call ended                               client=lakectl/dev host="localhost:8000" log_audit=true method=GET operation_id=ListRepositories path="/api/v1/repositories?after=&amount=100" request_id=9f7ea60c-1da6-46a9-bb7a-0c9fbd6f8d97 sent_bytes=354 service_name=rest_api source_ip="::1:35748" status_code=200 took=8.325235ms user=admin
DEBUG  [2023-02-27T15:05:17+02:00]pkg/api/controller.go:4099 pkg/api.(*Controller).LogAction performing API action                         class=api_server client=lakectl/dev host="127.0.0.1:8000" method=GET name=list_repos operation_id=ListRepositories path="/api/v1/repositories?after=&amount=100" ref="" repository="" request_id=79300b1b-dbfc-470d-a733-c584a5d18754 service=api_gateway service_name=rest_api source_ip="127.0.0.1:32964" source_ref="" user=admin user_id=""
```

Pulling... 😋

PR #5225, opened 2023-02-13
Add number of marked/sweeped objects to GC stdout

### LINKED ISSUE #5226: GC: print number of objects marked and sweeped
Add to the output of the GC process:
1. How many objects were marked for deletion.
2. How many objects were actually deleted.

johnnyaug:
Resolves #5226.

This one is for the Spark client's changelog, not the lakeFS one.
**Breaking change**: the path for the report containing the objects deleted by GC has changed to:

`[repository storage namespace]/_lakefs/logs/gc/deleted_objects/dt=[run timestamp]`

(the old one was:)
`[repository storage namespace]/_lakefs/logs/gc/deleted_objects/[run timestamp]/deleted.parquet`


lynnro314 (approved):

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -528,20 +527,33 @@ object GarbageCollector {
   ) = {
     val reportLogsDst = concatToGCLogsPrefix(storageNSForHadoopFS, "summary")
     val reportExpiredDst = concatToGCLogsPrefix(storageNSForHadoopFS, "expired_addresses")
+    val deletedObjectsDst = concatToGCLogsPrefix(storageNSForHadoopFS, s"deleted_objects")
```
Jonathan-Rosenberg, line 530: nit:
```suggestion
    val deletedObjectsDst = concatToGCLogsPrefix(storageNSForHadoopFS, "deleted_objects")
```

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -528,20 +527,33 @@ object GarbageCollector {
...
+
     val time = DateTimeFormatter.ISO_INSTANT.format(java.time.Clock.systemUTC.instant())
     writeParquetReport(commitsDF, reportLogsDst, time, "commits.parquet")
-    writeParquetReport(expiredAddresses, reportExpiredDst, time)
+    try {
+      writeParquetReport(expiredAddresses, reportExpiredDst, time)
+      val expiredDF = spark.read.parquet(f"${reportExpiredDst}/dt=${time}/")
+      println(f"Total of ${expiredDF.count()} addresses marked for deletion")
```
Jonathan-Rosenberg, line 537: if this is a sweep-only run then this log isn't true (and was already logged in the mark run)

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -528,20 +527,33 @@ object GarbageCollector {
...
     val reportLogsDst = concatToGCLogsPrefix(storageNSForHadoopFS, "summary")
     val reportExpiredDst = concatToGCLogsPrefix(storageNSForHadoopFS, "expired_addresses")
+    val deletedObjectsDst = concatToGCLogsPrefix(storageNSForHadoopFS, s"deleted_objects")
+
     val time = DateTimeFormatter.ISO_INSTANT.format(java.time.Clock.systemUTC.instant())
     writeParquetReport(commitsDF, reportLogsDst, time, "commits.parquet")
-    writeParquetReport(expiredAddresses, reportExpiredDst, time)
+    try {
+      writeParquetReport(expiredAddresses, reportExpiredDst, time)
+      val expiredDF = spark.read.parquet(f"${reportExpiredDst}/dt=${time}/")
```
Jonathan-Rosenberg, line 536: wouldn't it be sufficient to read the `expiredAddresses` DF?
I think that the source of truth for the marked addresses is: `/_lakefs/retention/gc/addresses/mark_id=<>/`

Jonathan-Rosenberg (changes requested):
Noice!
Just a few clarifying questions and nits

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
johnnyaug: I just didn't want to take any risks with trying to cache this possibly huge DataFrame (or worse, re-computing it). Therefore I chose to read the Parquet from the storage. Thoughts?

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
johnnyaug: Good point, changed the message.

johnnyaug:
@Jonathan-Rosenberg - thanks for the review, PTAL. Also note two additional changes:
1. [Breaking] Removed "deleted.parquet" from the path of the saved report (it causes issues when trying to read it in Spark).
2. Not failing on any exception when printing the count.

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -528,20 +531,39 @@ object GarbageCollector {
...
       .write
       .partitionBy(MARK_ID_KEY, RUN_ID_KEY)
       .mode(SaveMode.Overwrite)
-      .parquet(
-        concatToGCLogsPrefix(storageNSForHadoopFS, s"deleted_objects/$time/deleted.parquet")
-      )
+      .parquet(f"${deletedObjectsDst}/dt=$time")
+    try {
+      val deleteDF = spark.read.parquet(f"${deletedObjectsDst}/dt=$time")
+      println(f"Total objects deleted (or already had been deleted): ${deleteDF.count()}")
```
talSofer, line 559: line 549 counts removed objects, can we use it instead of reading deleted and counting again?

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
talSofer: i'm also not sure how is the current approach of re-reading expired addresses from storage is different than what we do in readExpiredAddresses. can you please help me understand?

talSofer (changes requested):

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
johnnyaug: Nice catch, will do

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
johnnyaug: readExpiredAddresses will read the addresses from a previous, mark-only run. This is not necessarily the case here.

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
```diff
@@ -528,20 +527,37 @@ object GarbageCollector {
...
+      val removedCount = removed.count()
+      writeJsonSummary(configMapper, reportLogsDst, removedCount, gcRules, time)
+      removed
+        .withColumn(MARK_ID_KEY, lit(markID))
+        .withColumn(RUN_ID_KEY, lit(runID))
+        .write
+        .partitionBy(MARK_ID_KEY, RUN_ID_KEY)
+        .mode(SaveMode.Overwrite)
+        .parquet(f"${deletedObjectsDst}/dt=$time")
+      println(f"Total objects deleted (or already had been deleted): ${removedCount}")
```
talSofer, line 554: >or already had been deleted

Isn't the removed count represent the exact number of objects that were deleted? I mean, given that gc is not incremental I expect expired addresses to include addresses of already deleted objects but for removed objects I expect is to be reliable for the run.

talSofer (approved):
Thanks! Added a clarification question about the logged messages

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
johnnyaug: The count represent all the expired addresses that, at the end of this run, were not present in the storage. It does not know if they were deleted in this run, in previous runs, or by anyone else for that matter.

IMO, it's good that we are being consistent with how the storage reports deletion. The only problem here is that it may not be consistent with Azure (I'm not sure whether it is or not).

clients/spark/core/src/main/scala/io/treeverse/clients/GarbageCollector.scala
Jonathan-Rosenberg: Not quite sure what you mean by "caching or recomputing the DF". 
You anyway write the `expiredAddresses` DF in line 535 which means that it will be computed, no?

johnnyaug: @Jonathan-Rosenberg - right. We may have another performance bug here, I just didn't want to add another one.

arielshaqed: Only getting here today, sorry.

It's really annoying.  I think fastest would be to cache the DF before writing it, and then count() that.  But then you need local disk space on the workers.  Rereading Parquet might be slower because I think it will need to open each object written.  But not by a lot: IIRC Parquet has a row count in every file

SO has several answers, all very old, none particularly informative.  You can pipe the results through something that counts, but that is quite intrusive.

PR #5227, opened 2023-02-13
Delta Diff: Add dummy OTF endpoint

### LINKED ISSUE #4967: Delta Diff MVP - Task: Endpoint
Add an endpoint to the swagger file and an implementation to the controller.

- [x] Lean version
- [ ] Fully functional
- [ ] Tests

Jonathan-Rosenberg:
Add OTF endpoint to the controller.
Notice that this PR doesn't contain the implementation of the endpoint, only its definition.

Closes #4967


pkg/plugins/diff/service.go
```diff
@@ -0,0 +1,89 @@
...
+
+type RefPath struct {
+	Ref  string
+	Path string
+}
+
+type TablePaths struct {
+	LeftTablePath  RefPath
+	RightTablePath RefPath
+	BaseTablePath  RefPath
```
johnnyaug, line 29:
```suggestion
	Left  RefPath
	Right RefPath
	Base  RefPath
```

pkg/plugins/diff/service.go
```diff
@@ -0,0 +1,89 @@
...
+	return service, service.Close
+}
+
+func (s *Service) RunDiff(ctx context.Context, diffType string, diffParams Params) (Response, error) {
+	d, closeClient, err := s.pluginHandler.LoadPluginClient(diffType)
+	if err != nil {
+		return Response{}, err
+	}
+	if closeClient != nil {
+		s.closeFunctions = append(s.closeFunctions, closeClient)
```
johnnyaug, line 70: This seems thread-unsafe. Also, for every diff type do we have multiple clients? Because we _will_ add multiple close functions for every diff type.

pkg/plugins/internal/manager.go
```diff
@@ -54,8 +63,17 @@ func NewManager[T any]() *Manager[T] {
...
+	Handshake PluginHandshake
+	P         plugin.Plugin
+}
+
 // RegisterPlugin registers a new plugin client with the corresponding plugin type.
-func (m *Manager[T]) RegisterPlugin(name string, id PluginIdentity, auth PluginHandshake, p plugin.Plugin) {
+func (m *Manager[T]) RegisterPlugin(name string, pp HCPluginProperties) {
+	id := pp.ID
+	auth := pp.Handshake
+	p := pp.P
```
johnnyaug, line 76: This is art

pkg/plugins/internal/manager.go
```diff
@@ -54,8 +63,17 @@ func NewManager[T any]() *Manager[T] {
 	}
 }
 
+type HCPluginProperties struct {
```
johnnyaug, line 66: What is HC? HashiCorp? Strange to include it as part of the name.

pkg/plugins/diff/service_test.go
```diff
@@ -0,0 +1,103 @@
...
+		return Response{}, errDiffFailed
+	}
+	return Response(ed), nil
+}
+
+type failureKey string
+
+const key failureKey = "fail"
+
+func ContextWithFailDiff(ctx context.Context, shouldFail bool) context.Context {
```
johnnyaug, line 93:
```suggestion
func contextWithFailDiff(ctx context.Context, shouldFail bool) context.Context {
```

pkg/plugins/diff/service_test.go
```diff
@@ -0,0 +1,103 @@
...
+	return context.WithValue(ctx, key, shouldFail)
+}
+
+func FailDiffFromContext(ctx context.Context) bool {
```
johnnyaug, line 97:
```suggestion
func failDiffFromContext(ctx context.Context) bool {
```

pkg/plugins/diff/service_test.go
```diff
@@ -0,0 +1,103 @@
...
+	shouldFail, ok := ctx.Value(key).(bool)
+	if !ok {
+		return ok
+	}
+	return shouldFail
```
johnnyaug, line 102: return ok && shouldFail

johnnyaug, line 102:
```suggestion
	return ok && shouldFail
```

johnnyaug (approved):
See comment about thread safety.

api/swagger.yml
```diff
@@ -386,6 +386,36 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/Diff"
+    OtfDiff:
+      type: object
+      required:
+        - version
+        - operation
+        - timestamp
+        - operation_content
```
talSofer, line 395: TBD: do we want to add an operation-level diff type to this structure instead of mapping table operations into diff types on the client side?

api/swagger.yml
```diff
@@ -386,6 +386,36 @@ components:
           type: array
           items:
             $ref: "#/components/schemas/Diff"
+    OtfDiff:
```
talSofer, line 389: a suggestion - up to you, I think that it makes it clearer that this object includes a diff record rather than the whole diff
```suggestion
    OtfDiffRecord:
```

talSofer (changes requested):
Thanks! it seems like this pr includes multiple changes that are out of the dummy endpoint scope. I guess this is not intentional? 

Also, look good but let's discuss adding an operation-level diff type before we merge the pr so blocking for that reason.

Jonathan-Rosenberg:
@talSofer thanks for the heads up.
It was a git merge thing (because of a change for the destination branch)

PR #5230, opened 2023-02-14
Fix doc images: branch protection rules

### LINKED ISSUE #3509: Docs: branch protection examples use illegal branch names
The [documentation for protected branches](https://docs.lakefs.io/reference/protected_branches.html) shows examples with a `/` in the branch name (see the screenshots with `stable/*` in them), which is an illegal, so no such branch could exist.

johnnyaug:
Resolves #3509.


ozkatz (approved):
Thank you!

PR #5236, opened 2023-02-14
Delta Lake Diff MVP: Add Delta Lake differ

### LINKED ISSUE #4966: Delta Diff MVP - Task: Business Logic
Add the business logic to the diff service: according to the received type (delta), use the client abstraction to call the corresponding plugin, receive the answer and compute a response.
- Create a Delta Diff gRPC client.

- [x] Lean version
- [ ] Fully functional
- [ ] Tests

Jonathan-Rosenberg:
### Linked Issue

Closes #4966

---

## Change Description

- Add a Delta Lake differ and plugin implementation.
- Extracted the plugin properties outside of the internal package so that they could be configured...


johnnyaug (approved):

pkg/plugins/diff/delta_differ.go
```diff
@@ -0,0 +1,96 @@
...
+		}
+		return Response{}, err
+	}
+	return Response{
+		Diffs:      buildDiffEntries(dr),
+		ChangeType: ChangeType(dr.ChangeType),
+	}, nil
+}
+
+func buildDiffEntries(dr *DiffResponse) []DiffEntry {
```
talSofer, line 58: nit; should we decide between record/entry and go with one term?

talSofer (commented):
Looks good, what is the plan for testing delta lake differ?

pkg/plugins/diff/delta_differ.go
Jonathan-Rosenberg: decided 😄

talSofer:
> @talSofer, regarding:
> 
> > what is the plan for testing delta lake differ?
> 
> I don't see a lot of value in unit testing this code as it has a very basic contract. Would you feel more confident if I'll add this test?

@Jonathan-Rosenberg adding a test will make sure that we don't break the contract even if it is simple (i.e. the response structure etc), so I think it is worth adding it.

pkg/plugins/diff/delta_differ_test.go
```diff
@@ -0,0 +1,158 @@
...
+		{
+			name:   "success",
+			client: TableDifferClientMock{},
+			diffResponse: &DiffResponse{
+				Entries: []*TableOperation{
+					{
+						Id:        "id1",
+						Timestamp: timestamppb.Now(),
+						Operation: "OPA",
+						Content: map[string]string{
```
talSofer, line 35: I guess that an updated version will include operation type?

talSofer (approved):
Looks good, thanks for adding tests! 
please look into adding the operation type into the diff entry

pkg/plugins/diff/delta_differ_test.go
Jonathan-Rosenberg: good catch

PR #5240, opened 2023-02-15
Delete temporary copies of ranges on closing their SSTables

### LINKED ISSUE #5241: [Bug]: Scala SSTableReader deletes local files only on exist
### What happened?

Current Behavior:
When reading ranges and meta ranges we copy the files to local. The files will be deleted only on the exit of the machine.
        
Steps to Reproduce:
1. When running UGC on scale (repository with large `_lakefs` metadata), reading a lot of ranges and not deleting them, spark exist on no space left on the device.

### Expected Behavior

Expected Behavior:
Delete the local file after reading the range.

arielshaqed:
Otherwise deleteOnExit only removes them _when the JVM exits_, which can
take a long while on Spark!

Why deleteOnExit in the first place?  File.createTempFile actually says to
do that, without explaining when to improve it...

Fixes #5241.

### DIFF
--- a/clients/spark/core/src/main/scala/io/treeverse/clients/SSTableReader.scala
+++ b/clients/spark/core/src/main/scala/io/treeverse/clients/SSTableReader.scala
@@ -74,19 +74,18 @@ object SSTableReader {
 }
 
 class SSTableReader[Proto <: GeneratedMessage with scalapb.Message[Proto]] private (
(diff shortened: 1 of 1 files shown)


arielshaqed:
Fixes #5241.

eden-ohana (approved):
Thanks!

PR #5249, opened 2023-02-15
Fix: Search Widget

### LINKED ISSUE #5178: [Bug]: Search results split filename incorrectly
### What happened?

When I search for an expression that partially matches a path, lakeFS returns a path with the partial match, and then a file with the remainder: 

For a file called `catjam.gif`: 

image

Search for `cat`, expect it to match `catjam.gif`

Instead it shows […]`cat/jam.gif` as the match: 

image

If I click on the path that doesn't actually exist (`jam.gif` in this case) it goes to the complete version (`catjam.gif`) which is good, but also confusing since I clicked on a link for something called `jam.gif`. 
image

### Expected Behavior

I'd expect `cat` to return `catjam.gif` as a complete filename, under its correct parent (instead of `jam.gif` under a parent of `cat/`

### lakeFS Version

0.91

### Deplyoment

Docker (also happens on demo.lakeFS.io)

### Affected Clients

_No response_

### Relevant logs output

_No response_

### Contact Details

robin.moffatt@treeverse.io

N-o-Z:
### Linked Issue

Closes #5178
---

### Background

Searching for a file in the objects browser by partial name resulted in trimmed file name containing only the suffix after the search term.
Also search was not performed on the current lakeFS URI context

### Bug Fix

Fixed by parsing URI and search term and separate use cases for file and path. Used router path for search within context
 
### Testing Details

Tested manually 

### Breaking Change?

No breaking changes

---

## Additional info

image

### DIFF
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -3172,6 +3172,11 @@ func (c *Controller) ListObjects(w http.ResponseWriter, r *http.Request, reposit
 	user, _ := auth.GetUser(ctx)
 	c.LogAction(ctx, "list_objects", r, repository, ref, "")
 
+	repo, err := c.Catalog.GetRepository(ctx, repository)
+	if c.handleAPIError(ctx, w, r, err) {
+		return
+	}
+
 	res, hasMore, err := c.Catalog.ListEntries(
 		ctx,
 		repository,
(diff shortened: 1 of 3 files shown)


eladlachmi (approved):
LGTM

N-o-Z:
> LGTM

Thanks!

PR #5252, opened 2023-02-15
Add merge strategy to commit data

### LINKED ISSUE #5222: Keep and show the merge strategy used in merge commit
Merge strategy is not logged or stored today. This information is useful in cases we like to explain how data managed to get to a specific commit.

N-o-Z:
### Linked Issue

Closes #5222
---

### Background

During investigation of a potential issue - we found the need to understand what merge strategy was performed for a given merge commit
      
### New Feature

On new merge commit - added the merge strategy used to the commit metadata

### Testing Details

Add controller unit test
Modified esti integration tests
Tested manually

### Breaking Change?

No

### DIFF
--- a/cmd/lakectl/cmd/log.go
+++ b/cmd/lakectl/cmd/log.go
@@ -17,11 +17,12 @@ Date:          {{ $val.CreationDate|date }}
 Merge:         {{ $val.Parents|join ", "|bold }}
 {{ end }}
 	{{ $val.Message }}
-
+{{ if $val.Metadata.AdditionalProperties }}
 Metadata:
 	{{ range $key, $value := $val.Metadata.AdditionalProperties }}
 	{{ $key | printf "%-18s" }} = {{ $value }}
 	{{ end -}}
+{{ end -}}
 {{ end }}{{ if .Pagination  }}
 {{.Pagination | paginate }}{{ end }}`
 
--- a/cmd/lakectl/cmd/log.go
+++ b/cmd/lakectl/cmd/log.go
@@ -20,7 +20,7 @@ Merge:         {{ $val.Parents|join ", "|bold }}
 
 Metadata:
 	{{ range $key, $value := $val.Metadata.AdditionalProperties }}
-	{{ $key }}	=	{{ $value }}
+	{{ $key | printf "%-18s" }} = {{ $value }}
 	{{ end -}}
 {{ end }}{{ if .Pagination  }}
 {{.Pagination | paginate }}{{ end }}`
--- a/cmd/lakectl/cmd/log.go
+++ b/cmd/lakectl/cmd/log.go
@@ -17,9 +17,10 @@ Date:          {{ $val.CreationDate|date }}
 Merge:         {{ $val.Parents|join ", "|bold }}
 {{ end }}
 	{{ $val.Message }}
-	
+
+Metadata:
 	{{ range $key, $value := $val.Metadata.AdditionalProperties }}
-		{{ $key }} = {{ $value }}
+	{{ $key }}	=	{{ $value }}
 	{{ end -}}
 {{ end }}{{ if .Pagination  }}
 {{.Pagination | paginate }}{{ end }}`
--- a/pkg/api/controller.go
+++ b/pkg/api/controller.go
@@ -1856,7 +1856,8 @@ func (c *Controller) handleAPIErrorCallback(ctx context.Context, w http.Response
 		errors.Is(err, model.ErrValidationError),
 		errors.Is(err, graveler.ErrInvalidRef),
 		errors.Is(err, actions.ErrParamConflict),
-		errors.Is(err, graveler.ErrDereferenceCommitWithStaging):
+		errors.Is(err, graveler.ErrDereferenceCommitWithStaging),
+		errors.Is(err, graveler.ErrInvalidMergeStrategy):
 		cb(w, r, http.StatusBadRequest, err)
 
 	case errors.Is(err, graveler.ErrNotUnique),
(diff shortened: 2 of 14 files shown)


cmd/lakectl/cmd/log.go
```diff
@@ -17,9 +17,10 @@ Date:          {{ $val.CreationDate|date }}
 Merge:         {{ $val.Parents|join ", "|bold }}
 {{ end }}
 	{{ $val.Message }}
-	
+
+Metadata:
 	{{ range $key, $value := $val.Metadata.AdditionalProperties }}
-		{{ $key }} = {{ $value }}
+	{{ $key }}	=	{{ $value }}
```
arielshaqed, line 23: I actually like this form less, it is harder to understand what is going on.

Options:
* Have JSON output format (for all of lakectl...) for people who want to know what's really going on.
* Put metadata in a table, I think we have enough support in our formatting for that.
* Try to align keys but don't worry if that doesn't work.

  Something like this might be good enough:
  ```suggestion
  	{{ $key | printf "%-18s" }} = {{ $value }}
  ```

pkg/graveler/errors.go
```diff
@@ -24,6 +24,7 @@ var (
 	ErrInvalidMergeBase             = fmt.Errorf("only 2 commits allowed in FindMergeBase: %w", ErrInvalidValue)
 	ErrNoCommitGeneration           = errors.New("no commit generation")
 	ErrNoMergeBase                  = errors.New("no merge base")
+	ErrInvalidMergeStrategy         = errors.New("invalid merge strategy")
```
arielshaqed, line 27: If this error actually reaches the user to be returned by HTTP, we should strip accumulated messages from it on the HTTP response.  So consider
```suggestion
	ErrInvalidMergeStrategy         = wrapError(ErrUserVisible, "invalid merge strategy")
```

pkg/graveler/graveler.go
```diff
@@ -111,9 +108,22 @@ type MergeStrategy int
...
 	MergeStrategyNone MergeStrategy = iota
 	MergeStrategyDest
-	MergeStrategySource
+	MergeStrategySrc
+
+	MergeStrategyNoneStr     = "default"
+	MergeStrategyDestWinsStr = "dest-wins"
+	MergeStrategySrcWinsStr  = "source-wins"
+
+	MergeStrategyMetadataKey = "lakefs.merge.strategy"
```
arielshaqed, line 117: Nice!

I would like us to take the opportunity to announce that some metadata keys are reserved for lakefs, and to **document** this fact.  At the same time, we might prefer to use `_lakefs` or `%lakefs` or `.lakefs`, to help people acknowledge that it is ours.

arielshaqed (commented):
Nice, thanks!

I really think we should stake a claim to a particular metadata key prefix.  That will involve documenting it somewhere (possibly not in this PR).  Also we might prefer to pick something _other_ than `lakefs.` by prefixing it with a special character: some people might already be using this prefix for some reason.

pkg/graveler/graveler.go
```diff
@@ -2265,14 +2275,24 @@ func (g *Graveler) Merge(ctx context.Context, repository *RepositoryRecord, dest
...
+			"strategy":               strategy,
 		}).Trace("Merge")
-		mergeStrategy := MergeStrategyNone
-		if strategy == MergeStrategyDestWins {
+
+		var mergeStrategy MergeStrategy
+		switch strategy {
+		case MergeStrategyDestWinsStr:
 			mergeStrategy = MergeStrategyDest
+
```
itaiad200, line 2285: nit: remove empty lines between cases

pkg/graveler/graveler.go
```diff
@@ -111,9 +108,22 @@ type MergeStrategy int
...
+	MergeStrategySrc
+
+	MergeStrategyNoneStr     = "default"
+	MergeStrategyDestWinsStr = "dest-wins"
+	MergeStrategySrcWinsStr  = "source-wins"
+
+	MergeStrategyMetadataKey = "lakefs.merge.strategy"
 )
 
+// mergeStrategyString String representation for MergeStrategy consts. Pay attention to the order!
```
itaiad200, line 120: Why does the order matter?

pkg/graveler/committed/merge_test.go
```diff
@@ -131,7 +131,7 @@ func Test_merge(t *testing.T) {
 				{rng: committed.Range{ID: "base:k3-k6", MinKey: committed.Key("k3"), MaxKey: committed.Key("k6"), Count: 2, EstimatedSize: 1234}},
 			}),
 			expectedResult: []testRunResult{{
-				mergeStrategies: []graveler.MergeStrategy{graveler.MergeStrategyNone, graveler.MergeStrategyDest, graveler.MergeStrategySource},
```
itaiad200, line 134: Can we add a check to some of theses tests that the new metadata key was added correctly?

pkg/api/controller.go
```diff
@@ -1856,7 +1856,8 @@ func (c *Controller) handleAPIErrorCallback(ctx context.Context, w http.Response
 		errors.Is(err, model.ErrValidationError),
 		errors.Is(err, graveler.ErrInvalidRef),
 		errors.Is(err, actions.ErrParamConflict),
-		errors.Is(err, graveler.ErrDereferenceCommitWithStaging):
+		errors.Is(err, graveler.ErrDereferenceCommitWithStaging),
+		errors.Is(err, graveler.ErrInvalidMergeStrategy):
```
itaiad200, line 1860: Nice one!

pkg/graveler/committed/merge_test.go
itaiad200: I saw you added that in esti, so this comment becomes less critical.

itaiad200 (approved):
LGTM
No blockers

pkg/graveler/graveler.go
N-o-Z: Modified to .lakefs.
Will open a separate issue for documentation

pkg/graveler/graveler.go
N-o-Z: Because the const indexes are coupled to the string representation array

PR #5254, opened 2023-02-15
Modified openapi hadoop client

### LINKED ISSUE #4938: Discover API option: OpenAPI with modified template or wrapper
## Definition of done

* [ ]Implement:
  * [ ] GetObject
  * [ ] UploadObject
  * [ ] (Optionally) something with streaming or pagination
* [ ] Use a single port on the lakeFS side
* [ ] Show calls from:
  * [ ] Web browser (GUI) [TypeScript]
  * [ ] Hadoop client (**without JNI!**) [Java]
* [ ] Object validation on client must be optional

eladlachmi:
Resolves #4938

### DIFF
--- a/clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
+++ b/clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
@@ -1,6 +1,8 @@
 package io.lakefs;
 
 import io.lakefs.clients.api.*;
+import io.lakefs.clients.api.StagingApi.APIgetPhysicalAddressRequest;
+import io.lakefs.clients.api.StagingApi.APIlinkPhysicalAddressRequest;
 import io.lakefs.clients.api.model.*;
 import io.lakefs.clients.api.model.ObjectStats.PathTypeEnum;
 import io.lakefs.utils.ObjectLocation;
@@ -64,6 +66,8 @@ public class LakeFSFileSystemTest {
     protected BranchesApi branchesApi;
     protected RepositoriesApi repositoriesApi;
     protected StagingApi stagingApi;
+    protected APIgetPhysicalAddressRequest getPhysicalAddressRequest;
+    protected APIlinkPhysicalAddressRequest linkPhysicalAddressRequest;
 
     protected AmazonS3 s3Client;
 
@@ -150,6 +154,8 @@ public class LakeFSFileSystemTest {
         when(lfsClient.getRepositories()).thenReturn(repositoriesApi);
         stagingApi = mock(StagingApi.class, Answers.RETURNS_SMART_NULLS);
         when(lfsClient.getStaging()).thenReturn(stagingApi);
+        getPhysicalAddressRequest = mock(APIgetPhysicalAddressRequest.class);
+        linkPhysicalAddressRequest = mock(APIlinkPhysicalAddressRequest.class);
 
         when(repositoriesApi.getRepository("repo"))
             .thenReturn(new Repository().storageNamespace(s3Url("/repo-base")));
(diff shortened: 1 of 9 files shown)


clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
```diff
@@ -304,7 +310,9 @@ public void testDelete_FileExists() throws ApiException, IOException {
                     .thenReturn(new ObjectStatsList().results(Collections.emptyList()).pagination(new Pagination().hasMore(false)));
         }
         StagingLocation stagingLocation = new StagingLocation().token("foo").physicalAddress(s3Url("/repo-base/dir-marker"));
-        when(stagingApi.getPhysicalAddress("repo", "main", "no/place/", false))
+        when(stagingApi.getPhysicalAddress("repo", "main", "no/place/"))
+                .thenReturn(getPhysicalAddressRequest);
+        when(getPhysicalAddressRequest.execute())
```
arielshaqed, line 315: AFAICT these calls show that `getPhysicalAddressRequest` should be a local variable here rather than a member.  Consider a test that causes `getPhysicalAddress` to run twice.

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
```diff
@@ -496,9 +504,21 @@ public void testCreate() throws ApiException, IOException {
...
 
         StagingLocation stagingLocation = new StagingLocation().token("foo").physicalAddress(s3Url("/repo-base/create"));
-        when(stagingApi.getPhysicalAddress("repo", "main", "sub1/sub2/create.me", false))
+        when(stagingApi.getPhysicalAddress("repo", "main", "sub1/sub2/create.me"))
+                .thenReturn(getPhysicalAddressRequest);
+        when(getPhysicalAddressRequest.execute())
                 .thenReturn(stagingLocation);
 
+        when(stagingApi.linkPhysicalAddress(eq("repo"), eq("main"), eq("sub1/sub2/create.me"), any(StagingMetadata.class)))
+                .thenReturn(linkPhysicalAddressRequest);
```
arielshaqed, line 513: If you do this, why do you still verify at l. 532?  I think it _used to_ work because of some Mockito weirdness here where it's configured to allow _any_ calls (and return a "smart" `null`); now you need to mock it, but that probably means you don't have to verify.

arielshaqed (commented):
Nice, thanks!

Please verify that client/python is unchanged by this OpenAPI spec change.  Or if it is, let's review that change too.

arielshaqed:
Not approving because it's a draft that we've not picked yet.

eladlachmi:
> Nice, thanks!
> 
> Please verify that client/python is unchanged by this OpenAPI spec change. Or if it is, let's review that change too.

@arielshaqed The python client isn't affected. This feature is currently implemented only in the Java generator. We could implement something similar in the Python client, but we can do that if/when we decide to go with this approach.

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
eladlachmi: Yeah, you're right. However, I don't want to invest more time into this before we decide if we want to move forward with this approach. If we do, I'll fix the test as you suggested.

clients/hadoopfs/src/test/java/io/lakefs/LakeFSFileSystemTest.java
eladlachmi: @arielshaqed You're not wrong, but the difference IMHO is in the feedback you get. If you use the capture in `when`, you'll get an error that no value was captured, so it looks like the test method is broken while actually the tested method might be broken. That's at least how I understand it.

nopcoder:
@eladlachmi can I close this one?

PR #5270, opened 2023-02-16
Remove unused table operation type enum

talSofer:
Remove unused enum created to represent a table operation type that's sent from the server.


Jonathan-Rosenberg (approved):
LGTM, yet you still have two warnings.
I don't think it's urgent because it will be used soon...

talSofer:
> LGTM, yet you still have two warnings. I don't think it's urgent because it will be used soon...

thanks, agreed that's why I only handled this warning

PR #5273, opened 2023-02-16
Branch protection status code should be 403

### LINKED ISSUE #5272: Put object branch protection status code should be 403 Forbidden
Trying to update protected branch returns status code 500.
Expected to return 403 Forbidden.

nopcoder:
Close #5272


clients/java/src/main/java/io/lakefs/clients/api/BranchesApi.java
```diff
@@ -213,6 +213,7 @@ public okhttp3.Call createBranchAsync(String repository, BranchCreation branchCr
         <tr><td> Status Code </td><td> Description </td><td> Response Headers </td></tr>
         <tr><td> 204 </td><td> branch deleted successfully </td><td>  -  </td></tr>
         <tr><td> 401 </td><td> Unauthorized </td><td>  -  </td></tr>
+        <tr><td> 403 </td><td> Internal Server Error </td><td>  -  </td></tr>
```
ozkatz, line 216: Example of why this is confusing :)

nopcoder:
> * Create a swagger error type for `Forbidden` errors (currently the errors are returned with a description of `Internal Server Error`)
> * Handle the other error that may rise from committing to a protected branch
> * Verify that merging could fail due to branch protection? if so, in which case? (i.e. if this is really the case, this requires documentation)

Thanks for the input - did better work in the next commit:
1. Added Fobidden error
2. Have common error for both cases of branch protection
3. Verified - commit protection doesn't block merge into branch. Think it will be helpful to add a "complete" protection option to enable read-only repositories.

ozkatz (approved):
Great! thank you!

